Issue #1938, opened 2020-03-11
Add endpoint for creating and processing order.


nerrad:
Essentially this issue is composed of two parts:

- Moving the draft order creation to the cart controller and exposing `orderId` on every cart route response (it would be `0` in the cart context and only have a value when an order is created.  Draft orders would probably be created as soon as there's enough billing info to do so in the checkout context.
- Creating an endpoint in the cart controller for processing the order. 

## Processing Order

Ideally, this endpoint would:

- Convert the draft order to pending.
- Then call `WC->checkout->process_checkout()` so all existing checkout functionality is just implemented.

However, it's more likely that we'll have to do all the same logic as `process_checkout` as a part of the endpoints callback because of the coupling `process_checkout()` has with an ajax request.

This issue only has to deal with getting the endpoint (and related php side tests) in place.

PR #1957, opened 2020-03-12
Break down Store API controllers into specific routes

mikejolley:
#### What is this? 

This PR converts our Store API controllers into classes for individual routes.

#### Why do this?

#1938 needs another route under `/cart`. The `/cart/order` route requires many methods so adding this to the existing cart controller would make things difficult to maintain.

Each controller was responsible for all endpoints and sub-routes under a route. For example, `/cart` had a single class handling the index route, apply coupon, update item, update shipping and so on. This lead to a monster of a class doing many things at once. 

Having single controllers also opened up the possibility of accidentally implementing incorrect schema for a route which almost happened recently when adding routes under `/cart`.

The code in this PR effectively:

- Abstracts out routes to create shared methods
- Removes the reliance on the core controller class
- Enforces a single route per class. A route can however be made up of multiple endpoints (PUT POST GET etc)
- Enforces a single **schema** per route
- Makes each class smaller and more focussed 
- Makes it simpler to add new routes or endpoints under an existing route

Whilst refactoring this, I removed the now unused `/customer` and `/shipping-rates` routes.

A follow up PR is incoming for the changes in #1938.

1. Smoke test various Store API endpoints
2. Test the Product Block and cart since those use Store API endpoints
3. Ensure tests continue to pass


src/RestApi/StoreApi/Routes/Cart.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @param \WP_REST_Request $request Request object.
+	 * @return \WP_Error|\WP_REST_Response
+	 */
+	public function get_response( \WP_REST_Request $request ) {
+		$controller = new CartController();
+		$cart       = $controller->get_cart_instance();
+
+		if ( ! $cart || ! $cart instanceof \WC_Cart ) {
+			return new \WP_Error( 'woocommerce_rest_cart_error', __( 'Unable to retrieve cart.', 'woo-gutenberg-products-block' ), array( 'status' => 500 ) );
+		}
```
nerrad: I realize we throw an error here because `\WP_Rest_Server` handles WP_Error, not exceptions.  If we have a common `get_response` callback wrapper on the `AbstractRoute`, it could have a single try catch that converts caught exceptions to `WP_Error`.  Then all our logic can just throw exceptions.  That way in this particular case (and anywhere the `CartController` is invoked, you could have the exception happen in the `CartController::get_cart_instance` method and that would save duplicating this same error logic everywhere (i.e. centralize the exception where it occurs, if `get_cart_instance` can't retrieve a cart instance, it throws).

So `AbstractRoute`, would have something like:

```php
public function get_response( \WP_REST_Request $request ) {
    $response = null;
    try { 
        $response = $this->get_route_response( $request );
    } catch ( WC_Rest_Exception $error ) {
        $response = new WP_Error(
          $error->code,
          $error->message,
          [ 'status' => $error->status ]
        )
    } catch ( Exception $error ) {
      $response = new WP_Error(
        'unknown_server_error',
        $error->message,
        [ 'status' => $error->status ]
      )
    }
    return $response;
}

abstract protected function get_route_response( $request );
```

src/RestApi/StoreApi/Routes/AbstractRoute.php
```diff
@@ -0,0 +1,135 @@
...
+	 * @var AbstractSchema
+	 */
+	protected $schema;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param AbstractSchema $schema Schema class for this route.
+	 */
+	public function __construct( AbstractSchema $schema ) {
```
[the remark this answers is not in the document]
nerrad: I realize there may be some routes that don't need a controller. You _could_ make it optional via the signature here: `public function __construct( AbstractSchema $schema, ControllerInterface $controller = null );` but I'd prefer to see us _require_ a controller (because it would help enforce keeping the logic separate a bit - controllers _could_ be tested independently of the route class for complicated logic).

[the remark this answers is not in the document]
mikejolley: I think the only endpoints that use a controller are the Cart ones. I'm not really happy with the Cart Controller itself and as you point out, not all endpoints need them. We could also consider using a AbstractCartRoute to house some of the cart functionality, since the endpoints are quite coupled to that controller anyway.

I think let's leave that out of this PR and come back around, possibly during cooldown.

src/RestApi/StoreApi/Routes/Cart.php
mikejolley: @nerrad So because we have a callback per method, the only way to do something like this would be to have a try catch in each and a handler in abstract.

However, by using the request method we can have a single callback like you suggested. I ran with that. Just updated - can you take a look?

src/RestApi/StoreApi/Routes/Cart.php
```diff
@@ -54,15 +54,16 @@ public function get_args() {
...
+	 * @return \WP_REST_Response
 	 */
-	public function get_response( \WP_REST_Request $request ) {
+	protected function get_route_response( \WP_REST_Request $request ) {
 		$controller = new CartController();
 		$cart       = $controller->get_cart_instance();
 
 		if ( ! $cart || ! $cart instanceof \WC_Cart ) {
-			return new \WP_Error( 'woocommerce_rest_cart_error', __( 'Unable to retrieve cart.', 'woo-gutenberg-products-block' ), array( 'status' => 500 ) );
+			throw new RouteException( 'woocommerce_rest_cart_error', __( 'Unable to retrieve cart.', 'woo-gutenberg-products-block' ), 500 );
```
nerrad: I was thinking you could have this in the `$controller->get_cart_instance()` because `get_cart_instance` is called in a lot of places. That would centralize this exception in one place?

src/RestApi/StoreApi/Routes/RouteException.php
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * @package WooCommerce/Blocks
+ */
+
+namespace Automattic\WooCommerce\Blocks\RestApi\StoreApi\Routes;
+
+/**
+ * ReserveStockRouteExceptionException class.
+ */
+class RouteException extends \Exception {
```
nerrad: Love it 👏

PR #1959, opened 2020-03-12
Add Order Summary card to Checkout sidebar

### LINKED ISSUE #1304: Checkout Block: Sidebar Section of Checkout Template - Edit Context
Note: when working on this issue, please give due consideration to:

- Are there components that can be pulled from `@wordpress/components` (i.e. `<Card>`)?
- Are there components that can be pulled from `@woocommerce/components`?
- Are there components/logic that can be pulled from [`@automattic/components`](https://github.com/Automattic/wp-calypso/tree/master/packages/components/src) or [`@automattic/composite-checkout`](https://github.com/Automattic/wp-calypso/tree/master/packages/composite-checkout)

**Edit:** Related to the above, it looks like #1423 can be used as the container for this area.

We want to avoid duplication of work/logic already done elsewhere as much as possible.

As per the exploration done by @Aljullu in #1208 this issue is concerned with the sidebar area of the checkout template.

In context of entire view:

checkout

## Sidebar

image
* CheckoutSidebar (`/base/components/checkout/sidebar`)
	* All contents are already defined as Cart Block components.

### LINKED ISSUE #1486: Checkout (frontend context): Wire up order summary to data.
The following needs to be wired up to data in the frontend (some of this could be split out to separate pulls):

* Update totals and product details based on the contents of the cart.
* Update totals when shopper adds a valid coupon (and add appropriate notices/messaging when invalid coupon is added).
* Update the totals when the shopper changes the shipping option.
* Update the totals when the shopper adds a valid coupon
* Update the shipping options and totals when the shopper changes their shipping address.

Aljullu:
Closes #1486.
Closes #1304.

### Screenshots
imatge

1. Create a page with the _Checkout_ block and verify it has the _Order Summary_ in the sidebar.
2. Verify there are no visual regressions in the _Cart_ block.

### DIFF
--- a/assets/js/base/components/cart-checkout/index.js
+++ b/assets/js/base/components/cart-checkout/index.js
@@ -1,1 +1,8 @@
 export { default as Button } from './button';
+export { default as ProductImage } from './product-image';
+export { default as ProductLowStockBadge } from './product-low-stock-badge';
+export { default as ProductMetadata } from './product-metadata';
+export { default as ProductName } from './product-name';
+export { default as ProductPrice } from './product-price';
+export { default as ProductSaleBadge } from './product-sale-badge';
+export { default as ProductVariationData } from './product-variation-data';
--- a/assets/js/base/components/cart-checkout/product-sale-badge/index.js
+++ b/assets/js/base/components/cart-checkout/product-sale-badge/index.js
@@ -10,9 +10,6 @@ import PropTypes from 'prop-types';
  */
 import './style.scss';
 
-/**
- * Returns a low stock badge for a line item.
- */
 const ProductSaleBadge = ( { currency, saleAmount } ) => {
 	return (
 		saleAmount > 0 && (
--- a/assets/js/base/components/cart-checkout/product-sale-badge/index.js
+++ b/assets/js/base/components/cart-checkout/product-sale-badge/index.js
@@ -1,0 +1,35 @@
+/**
+ * External dependencies
+ */
+import { __, sprintf } from '@wordpress/i18n';
+import { formatPrice } from '@woocommerce/base-utils';
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+/**
+ * Returns a low stock badge for a line item.
+ */
+const ProductSaleBadge = ( { currency, saleAmount } ) => {
+	return (
+		saleAmount > 0 && (
+			<div className="wc-block-sale-badge">
+				{ sprintf(
+					/* translators: %s discount amount */
+					__( 'Save %s!', 'woo-gutenberg-products-block' ),
+					formatPrice( saleAmount, currency )
+				) }
+			</div>
+		)
+	);
+};
+
+ProductSaleBadge.propTypes = {
+	currency: PropTypes.object,
+	saleAmount: PropTypes.number,
+};
+
+export default ProductSaleBadge;
--- a/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
@@ -113,32 +113,6 @@ table.wc-block-cart-items {
 			width: 100%;
 			margin: 0;
 		}
-		.wc-block-cart-item__product {
-			.wc-block-cart-item__product-name {
-				color: $core-grey-dark-600;
-				font-size: 16px;
-				display: block;
-			}
-
-			.wc-block-cart-item__low-stock-badge {
-				display: inline-block;
-
-				background-color: $white;
-				border-radius: 3px;
-				border: 1px solid $black;
-				color: $black;
-				font-size: 12px;
-				padding: 0 1em;
-				text-transform: uppercase;
-				white-space: nowrap;
-			}
-
-			.wc-block-cart-item__product-metadata {
-				color: $core-grey-dark-400;
-				font-size: 12px;
-				margin-top: 0.25em;
-			}
-		}
 		.wc-block-cart-item__quantity {
 			.wc-block-cart-item__remove-link {
 				@include link-button;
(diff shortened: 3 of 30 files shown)


assets/js/base/components/cart-checkout/index.js
```diff
@@ -0,0 +1,7 @@
+export { default as ProductImage } from './product-image';
+export { default as ProductLowStockBadge } from './product-low-stock-badge';
+export { default as ProductMetadata } from './product-metadata';
+export { default as ProductName } from './product-name';
+export { default as ProductPrice } from './product-price';
+export { default as ProductSaleBadge } from './product-sale-badge';
+export { default as ProductVariationData } from './product-variation-data';
```
Aljullu: I noticed some of these blocks are very similar to the ones in `atomic/components/product/`. I think it might be worth checking if they can be merged, but I left it out of this PR because it was already quite big and I don't think that's something blocking.

assets/js/base/components/cart-checkout/product-sale-badge/index.js
```diff
@@ -0,0 +1,35 @@
...
+import { formatPrice } from '@woocommerce/base-utils';
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+/**
+ * Returns a low stock badge for a line item.
```
haszari: Comment pasteo :)

assets/js/blocks/cart-checkout/cart/full-cart/style.scss
```diff
@@ -321,7 +273,7 @@ table.wc-block-cart-items {
 					display: inline-block;
 				}
 
-				.wc-block-cart-item__sale-badge {
+				.wc-block-sale-badge {
```
haszari: Should some of these styles be in the components directly? Probably not a big deal, there's a little mixing of concerns here - the cart is overriding/tweaking component styles.

assets/js/blocks/cart-checkout/checkout/order-summary-item.js
```diff
@@ -0,0 +1,86 @@
...
+						value={ purchasePrice }
+					/>
+				</div>
+				<ProductLowStockBadge lowStockRemaining={ lowStockRemaining } />
+				<ProductMetadata summary={ summary } variation={ variation } />
+			</div>
+			<div className="wc-block-order-summary-item__total-price">
+				<ProductPrice
+					currency={ currency }
+					value={ purchasePrice * quantity }
```
haszari: Does this work correctly with the different tax options?

assets/js/blocks/cart-checkout/checkout/order-summary-item.js
```diff
@@ -0,0 +1,86 @@
...
+		images,
+		low_stock_remaining: lowStockRemaining = null,
+		name,
+		permalink,
+		prices,
+		summary,
+		variation,
+	} = cartItem;
+
+	const { quantity } = useStoreCartItemQuantity( cartItem );
```
haszari: I don't think you need this hook - you can just accept a static quantity prop from parent component (which should have all the item details).

`useStoreCartItemQuantity` hook is for changing quantity or removing item :)

assets/js/base/components/cart-checkout/index.js
haszari: Agree, this is good follow up for later. Great that cart & checkout are sharing these components now though :)

haszari:
@Aljullu One more thing .. is it possible to give the description / summary text more room? They all wrap and it looks like this is unnecessary for the data I have. (e.g. `This is a simple product`)

Screen Shot 2020-03-13 at 12 41 52 PM

garymurray:
> The circled quantity numbers seem a little overkill when there's 1x of a product, I'm guessing this is the intended design. Could consider showing number only when there are multiple of the item cc @garymurray. (Note this is exacerbated by my red customisation!!)

Let's leave that as is for now. Any changes here can be made later in future iterations.

assets/js/base/components/cart-checkout/product-sale-badge/index.js
Aljullu: Removed in 4a60a67b57a8d8d5f40fc7467b79c69c973fd942.

assets/js/blocks/cart-checkout/cart/full-cart/style.scss
Aljullu: I don't think so, I left them here in purpose. Even though they refer to the `ProductSaleBadge` component, they are only relevant under the _Cart_ block, so I think they belong to the _Cart_ CSS.

assets/js/blocks/cart-checkout/checkout/order-summary-item.js
Aljullu: I think so, those prices honor the tax settings.

I found an issue, though: #1963. But since it's affecting the _Cart_ too, I don't think it should block this PR.

assets/js/blocks/cart-checkout/checkout/order-summary-item.js
Aljullu: Oh, good catch. I missed that `cartItem` has the `quantity` attribute. Fixed in 87214e31c0bb18ccdcd9e22ab76ab278f95e48eb.

Aljullu:
> Focus highlight on order summary "collapse" heading element has different padding/margin on right/left.

Fixed in 7d4fd6eb2eb9458ef01b38cdfcd8362115b0fa22.

> Hard-coded text color - it would be great if we can avoid this and/or use theme colors. In my cart I'm seeing a mix of black-on-white (opinionated blocks) vs. red (storefront customiser).

That's a tricky issue that I would like to handle in another PR, the hardcoded colors are there because in the designs they differ from the default color. 

I also noticed the _Cart_ block completely ignores the _Text color_ preference, while the _Checkout_ block applies it inconsistently. I created an issue to discuss what's the best solution for this: #1964.

> Hard error when a bad coupon is entered - I think this is already logged as an issue.

Right, I think that's not related to this PR.

> One more thing .. is it possible to give the description / summary text more room? 

Good idea. I refactored the CSS in c575d54564b920b26866195a40bd7937aea27e7e.

Issue #1947, opened 2020-03-11
Expose default `country` for payment methods.


nerrad:
While working on the stripe implementation (apple pay), I noticed that the _existing_ stripe extension relies on the [default country](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/9d7bce7de4817ab2286640f37789c3e596c6b406/includes/payment-methods/class-wc-stripe-payment-request.php#L95) both for determining whether stripe is supported and also to pre-select the country in any shipping/billing fields found in the payment method interface client side.

I suspect this will be similar for some other payment methods integrating with the cart/checkout so we'll need to [expose this value as a part of `usePaymentMethodInterface`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/82cc6c6dbb0b6c4b4d2ef3d4da1143907ec5fa06/assets/js/base/hooks/payment-methods/use-payment-method-interface.js#L156-L158). 

It's possible this will already be available via the shipping rates value on the cart endpoint but I'm uncertain (having not looked closely) whether this is coupled more to the logged in customer or not. So this issue is to ensure that this value is exposed so that payment methods have access to this.

PR #1956, opened 2020-03-12
Implement error handling for shipping rates and selected shipping rates.

nerrad:
Fixes: #1953

This pull adds some error handling for the shipping rates requests and select shipping rate requests.

Currently this will correctly throw errors for non-recoverable states. A couple of follow-up issues are open for validating and improving validation signals to customers: #1969, and #1966 

All errors will trigger something like this:

Image 2020-03-12 at 12 00 58 PM

In this pull:

- Added a `useThrowError` hook that is the workaround currently needed for throwing errors in hook callbacks that get caught by error boundaries (replaced existing logic in `useCollection` with this too).
- update error handling for `useSelectShippingRate` and `useShippingAddress`.

## To Test

This impacts coupon handling (invalid and valid coupons), shipping rate calculations and selecting shipping rates.

- First test that existing behaviour for the above works as expected.
- To test errors with shipping rates and selected shipping rates, you'll need to tweak the endpoints temporarily manually directly in the code to trigger an error on the rest api responses.

In polish work after we get the blocks done, I want to add tests for this error handling. Adding tests at this point extends how long it will take to get this pull merged (a number of mocks will be needed) so not doing now.

PR #1943, opened 2020-03-11
Refactor cart shipping settings and shipping calculator

mikejolley:
Updates the shipping calculator and shipping total row logic to match the designs and feedback in #1787. This turned out to be a complex refactor, but it does resolve a few issues.

Fixes #1835
Fixes #1787

This PR also:

- introduces context to handle sample cart data/rates avoiding the need to pass data down the component tree.
- (#1787) Makes the two shipping block options unrelated to each other so each option can be toggled individually.
- Tidies up the shipping calculator logic so it's clearer where addresses are updated
- Fixes some shapes/missing default data preventing cart notices (found this when implementing preview data).
- Allows `TotalsItem` to display regular text (required for #1787)
- Small update to `withFeedbackPrompt` to avoid it jumping around in position when toggling the block options on/off. Made the text consistent between the feedback prompt for cart and checkout to cut down i18n.

### Screenshots

No rates, with address:
no rates found with address entered

Rates found:
with rates

No address entered yet, and no calculator enabled:
no address no rates no calculator

No address entered yet, and no rates found:
no address and no rates

### How to test the changes in this Pull Request:

1. Best to test in a private window. 
2. In wc settings, set the default address to "no location" so when you enter the cart page no address data will be set.
3. Ensure that when "Hide shipping costs until an address is entered" is enabled, rates only ever show after entering an address (country minimum)
4. Test that "Shipping calculator" option hides/shows the calculator respectively.

PR #1960, opened 2020-03-12
Refactor Cart to include shippingAddress

nerrad:
Fixes: #1947

The original purpose of #1947 was to expose the default country from the WooCommerce store on the payment method interface so payment methods could use that to pre-populate any requirements for that country. However, after some investigation on where I could get this data:

- I realized that `WC()->customer` will automatically be populated with default locale according to the WC site (if the current session isn't for a customer). 
- I noticed, `shippingAddress` client side was being derived from the destination on the first shipping package. This to me seems fragile because although it appears woo only supports one destination per package, the very fact that it _is_ per package suggests it might not always be the case. So it seems to me, deriving the main customer address from the first package is fragile. 
- The logic in `useShippingRates` (although it's not the first time I've seen this) seems overly complicated to allow for deriving the shipping address from the first package.

Based on the above, I decided to give a go at doing a bit of refactoring to simplify things and ensure that we _always_ have a default address in the cart depending on the current session. So this pull:

- adds `shippingAddress` to the cart response (which are derived from `WC()->customer`).
- adds a `ShippingAddressSchema` class server side (since shipping address is shared between cart and order routes).
- modifies `useStoreCart` to expose `shippingAddress`  (and updates all the related typedefs).
- updates the reducer in the cart store so the default state includes `shippingAddress` and `billingAddress`.
- adds a new hook `useShippingAddress` which takes some of the address logic from `useShippingRates` and simplifies it a bit. 
- implements the new hook in `useShippingRates` (which simplifies that hook greatly) - no longer require any arguments.
- updates any places `useShippingRates` was implemented (including tests).
- removes the `country` property exposed on `usePaymentMethodInterface` since payment methods can now reliably get that value from the `shippingAddress` in the interface (which will always be populated with any defaults that might exist according to the settings on the site or the current customer session).

## To Test

- phpunit tests for the cart endpoint should pass (they cover these changes)
- js unit tests for `useShippingRates` should pass (which are impacted).
- smoke test address changes etc in the cart and ensure that there are no errors with getting new rates etc.

## Conflicting/related pulls

There are existing pulls that will impact (or be impacted by) this pull because of in progress work for awareness: 

- #1943 : refactoring of cart shipping settings and calculator. This mostly just involves files getting moved around (in terms of impact).
- #1956 : adds error handling for the shipping rate and shipping rate selection hooks. I think I may want to merge this pull first and then deal with the conflicts in merging into the other pull (likely just copying directly the error handling since it will move around hooks).
- #1957: This is breaking down cart api controller into separate controllers and thus the new properties will be affected.


assets/js/base/hooks/shipping/use-shipping-address.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Internal dependencies
+ */
+import { useStoreCart } from '../cart/use-store-cart';
+import { pluckAddress } from '../../utils';
+
+const reducer = ( state, address ) =>
+	isShallowEqual( pluckAddress( address ), pluckAddress( state ) )
+		? state
+		: { ...state, ...address };
```
nerrad: I moved this logic into the reducer and out of the effect because this simplifies the hook implementation. With this change `debouncedShippingAddress` only updates when the `useReducer` state updates and that state will only update when there is a difference between the incoming pluckedAddress and the current address in the state (rather than change the state everytime the reducer was invoked).

assets/js/base/hooks/payment-methods/use-payment-method-interface.js
```diff
@@ -169,9 +169,6 @@ export const usePaymentMethodInterface = () => {
 			orderLoading,
 			cartTotal: currentCartTotal.current,
 			currency: getCurrencyFromPriceResponse( cartTotals ),
-			// @todo need to pass along the default country set for the site
-			// if it's available.
-			country: '',
```
nerrad: This can now be derived from `shippingData.shippingAddress`.

assets/js/base/hooks/shipping/use-shipping-address.js
senadir: `useReducer` also controls the local state of that form (as well as cart local state), by having this there 
`isShallowEqual( pluckAddress( address ), pluckAddress( state ) )` we refuse all state changes that are not related to shipping fields (country, state, city, postcode).

the goal of that line was to simply cancel API actions if those fields didn’t change

assets/js/base/hooks/shipping/use-shipping-address.js
```diff
@@ -0,0 +1,39 @@
...
+
+export const useShippingAddress = () => {
+	const { shippingAddress: initialAddress } = useStoreCart();
+	const [ shippingAddress, setShippingAddress ] = useReducer(
+		reducer,
+		initialAddress
```
senadir: now that we receive shipping address schema from the backend, we don’t need to pass it down. I love this, I didn’t like the previous solution I implemented.

assets/js/base/hooks/shipping/use-shipping-rates.js
```diff
@@ -1,64 +1,33 @@
...
  *                   update shipping address and dispatches async rate fetching.
  *                 - {Object} shippingAddress       An object containing shipping address.
  */
-export const useShippingRates = ( addressFieldsKeys ) => {
+export const useShippingRates = () => {
 	const { cartErrors, shippingRates } = useStoreCart();
-	const addressFields = Object.fromEntries(
-		addressFieldsKeys.map( ( key ) => [ key, '' ] )
-	);
-	const derivedAddress = shippingRates[ 0 ]?.destination;
```
senadir: I do agree that this is fragile, when I first implemented it, I didn’t find any other solutions, exposing yet other values from the cart endpoint didn’t feel right to me, I feel we're shipping too much data from that endpoint.

assets/js/base/hooks/shipping/use-shipping-address.js
```diff
@@ -0,0 +1,39 @@
...
+import { pluckAddress } from '../../utils';
+
+const reducer = ( state, address ) =>
+	isShallowEqual( pluckAddress( address ), pluckAddress( state ) )
+		? state
+		: { ...state, ...address };
+
+export const useShippingAddress = () => {
+	const { shippingAddress: initialAddress } = useStoreCart();
+	const [ shippingAddress, setShippingAddress ] = useReducer(
```
senadir: what I do love about this new approach is that is somehow solved an issue with the previous approach, shipping fields in checkout didn’t populate if Cart endpoint returned an address, now it populates it 🙌

assets/js/base/hooks/shipping/use-shipping-address.js
nerrad: Ya good catch, this will have more importance in the checkout when there are more fields that could change, so good call there.

assets/js/base/hooks/shipping/use-shipping-address.js
```diff
@@ -13,22 +13,22 @@ import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
...
+const shouldUpdateStore = ( oldAddress, newAddress ) =>
+	! isShallowEqual( pluckAddress( oldAddress ), pluckAddress( newAddress ) );
 
 export const useShippingAddress = () => {
 	const { shippingAddress: initialAddress } = useStoreCart();
-	const [ shippingAddress, setShippingAddress ] = useReducer(
-		reducer,
-		initialAddress
-	);
+	const [ shippingAddress, setShippingAddress ] = useState( initialAddress );
```
senadir: I usually prefer `useReducer` for form handling, but since we didn’t update the checkout form updating method (still works as if this is `useState`) I think this is fine, we do have one use case that could benefit from `useReducer` (updating state select when you a country is set) but it feels like over engineering, so this is good.

assets/js/base/hooks/shipping/use-shipping-address.js
```diff
@@ -13,22 +13,22 @@ import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
...
 	const [ debouncedShippingAddress ] = useDebounce( shippingAddress, 400 );
 	const { updateShippingAddress } = useDispatch( storeKey );
 
+	// Note, we're intentionally not using initialAddress as a dependency here
+	// so that the stale (previous) value is being used for comparison.
```
senadir: ✅

assets/js/base/hooks/shipping/use-shipping-address.js
nerrad: `useReducer` is mostly for when you have multiple conditions in which the state changes, or when the state has different slices that get updated instead of the _entire state_.  Since we're dealing with a simple update replacing the entire state, `useReducer` for this use-case was definitely overkill.

senadir:
again, this is still testing fine for me, all behavior is working fine and clearing fields is working fine as well, the address is still preserved, I don’t see any issues in the console. Didn’t check the grammar much of the new fields since I trusted that you simply copied them from the previous place

PR #1780, opened 2020-02-19
Cart Checkout Flow Architectural Update (reference pull). Includes Apple Pay Integration POC

nerrad:
# Cart Checkout Flow Architectural Update

This pull is a reference pull for some proposed changes to checkout block architecture structure and flow.  **I reference Cart block as well** because I think much of this _can_ apply to the cart context as well considering express payment methods _might_ get implemented in the cart at some point. So what is being proposed here could be adapted for the cart relatively trivially. However, the perspective of this pull is orientated at least initially, towards the checkout block implementation. It is derived out of the following guiding principles:

- Having a single source of "truth" for data within the checkout flow.
- Providing a consistent interface for payment method integrations (and potentially other important extensibility pieces in the future). This interface protects the integrity of the checkout process and isolates extension processing from checkout processing. Checkout handles _all_ processing of the order, extensions react to and communicate data for checkout to pass along to the server where server side extensibility allows extensions to continue necessary processing there.
- Checkout flow state is tracked by status.
- Extensions (beginning with payment methods) are able to interact through checkout flow via the usage of emitted events.

Ultimately, this is a **proposal**, however it does hold up finishing off the cart and checkout blocks. Since we've already had some sign off from #1349 with regard to payment method integration, I don't think it's a _requirement_ to get this reviewed by payment method developers before moving ahead with this, but it'd still be worthwhile having this reviewed at some point to catch any potential red flags (I'll be posting an internal post to help gather this feedback). 

However, I do want feedback from this repos maintainers and team:

- Are there concerns with any of the general architectural proposals here?
- What red flags immediately pop up that are concerns?
- When reviewing in particular the plan section (which outlines how we'd implement the work in this pull and remaining todos) are there any adjustments to the plan that you forsee? Any additions that you think should be included?

## Bringing it all together

Many of the component parts of the checkout flow have already been developed and implemented over the last few weeks of cart and checkout block work. For the most part, this proposal does not replace any work that has already been done. It simply integrates these disparate parts together into a cohesive flow that provides consistent status state and extensibility points (which focus initially on payment method extensions) to support iteration one of the blocks.

As a part of working out this architecture proposal, I built an example integration of Apple Pay. I was able to get it to the point of actually triggering the payment interface and asserting it received totals as expected. Essentially, this validated that the approach was _workable_. However, there are still a number of pieces needing complete before I can do a complete transaction. Still, the validation was enough for me to get this reference pull up for review and comments. This should help us with the final push to finishing off the blocks.

[...]

Issue #1471, opened 2020-01-03
Create Payment Methods for Stripe


nerrad:
Using the api introduced in #1349, this pull will create the `ExpressPaymentMethod` and `PaymentMethod` integrations for Stripe.  Some notes:

* should use as much of the existing stripe extension logic as possible (so will need to detect if the extension is installed).
* should be created with a goal of moving this logic to the actual payment method extensions themselves (possibly before release). We may need to talk more about this.
* Will need to expose icons (on a yet to be built api) for the cart payment icons. (see #1472)

Here's an updated mockup provided by @LevinMedia  that shows a few changes:

<img width="775" alt="Image 2020-03-25 at 11 52 31 AM" src="https://user-images.githubusercontent.com/1429108/77556703-382be180-6e8f-11ea-90f7-1a46a734087f.png">

- label for stripe tab is `Credit/Debit Card`
- Credit card icons get moved to below save payment information checkbox.
- Saved payment methods are moved outside of the PaymentMethod component and are a part of the payment step (saved payment methods are provided by checkout).
- Payment Method component only gets displayed if there are no saved payment methods or "Use a new payment method" is selected.

PR #1961, opened 2020-03-13
Implement Stripe Apple Pay into checkout

nerrad:
Part of: #1471 

The majority of the work in this pull was sonde as a part of #1780 and implements the Stripe Apple Pay payment method into the checkout block.

When complete, this will be a functional payment method for making payments via checkout.

I'll be adding to this pull as various pieces are updated in the master branch.


nerrad:
I'm closing and deleting this pull in favor of #1983 which has all the changes in this pull as well. Maintaining the two branches is taking too much time so this will simplify things.

PR #1968, opened 2020-03-13
Fix layout issues after merges

Aljullu:
Fixes some issues we found after the recent merges.

* 5445e707b127596ff798adb7598b2b49ae26eb6f fixes a misalignment in the address form which was introduced in #1941. I think we initially missed it because it was only reproducible in Chrome.
* 8be46078d1e9257d3fd797f0b4f6b91775f5dbc7 prevents the Shipping Calculator from appearing in the _Checkout_ sidebar. I think we merged two PRs in parallel and they both conflicted in the way `<TotalsShippingItem />` was used.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. In a post with the _Checkout_ block.
2. Verify there is no button to _Change address_ in the sidebar neither the list of shipping rates.
3. With a Chromium-based browser, verify the address form doesn't break into several rows.

### DIFF
--- a/assets/js/base/components/cart-checkout/button/style.scss
+++ b/assets/js/base/components/cart-checkout/button/style.scss
@@ -2,14 +2,16 @@
 .wc-block-cart,
 .wc-block-checkout {
 	.button.wc-block-button {
+		align-items: center;
 		background-color: #000 !important;
 		color: #fff !important;
-		display: block;
+		display: flex;
 		font-size: inherit;
 		font-weight: bold;
 		height: 48px; // same height as text-input
+		justify-content: center;
 		line-height: 1;
-		padding: $gap-small;
+		padding: 0 $gap-small;
 		text-align: center;
 		text-decoration: none;
 		text-transform: none;
--- a/assets/js/base/components/select/style.scss
+++ b/assets/js/base/components/select/style.scss
@@ -1,4 +1,5 @@
 .wc-block-select {
+	height: 48px;
 	position: relative;
 	margin-top: $gap;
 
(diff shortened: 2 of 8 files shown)


senadir:
> 5445e70 fixes a misalignment in the address form which was introduced in #1941. I think we initially missed it because it was only reproducible in Chrome.

It’s weird because this was not reproducible to me and I use Chrome

nerrad:
> It’s weird because this was not reproducible to me and I use Chrome

It only happened in the editor, does your observation include the editor or just frontend?

senadir:
> > It’s weird because this was not reproducible to me and I use Chrome
> 
> It only happened in the editor, does your observation include the editor or just frontend?

Yeah I only tested in the frontend, my bad, didn’t assume much will change.

senadir:
So there is this regression that you might add to this PR Albert?
image

in a previous PR, it seems you set line-height to 1 and added a padding, to ensure the buttom is well aligned we can remove the top and bottom padding and have a fixed size line height.

```scss
.wc-block-cart,
.wc-block-checkout {
	.button.wc-block-button {
		height: 48px; // same height as text-input
		line-height: 48px;
		padding: 0 $gap-small;
	}
}

```

Aljullu:
> So there is this regression that you might add to this PR Albert?

Good catch! My only concern with setting the `line-height` to the button `height` is that it wouldn't play well when text breaks in two lines. Instead, in 7e5393f08270c01add365c6108b2d10ae6601b8f I tried refactoring the CSS to use flexbox, so it will handle multi-line buttons correctly:
imatge

Let me know how it looks. :slightly_smiling_face:

Issue #1969, opened 2020-03-13
(Cart) Add validation for empty country to shipping address field.


nerrad:
Right now the following happens on the shipping fields (in shipping calculator).

- We don't submit any update to the shipping rates endpoint until there is a country.
- So someone can fill out any of the other fields and not country, and there will be no update sent to the server.
- So customers could fill out everything _but_ country and think they are seeing the correct rates when they aren't.

To reproduce this, the site must be configured to have no default address.

This issue is for adding inline validation to the shipping calculator fields so that there is an inline validation message when any of the fields are filled out _except_ the country.

Issue #1966, opened 2020-03-13
Add inline validation error and snackbar notices for coupon ui/ux


senadir:
**Describe the bug**
Currently, if you enter a non-valid coupon, a top level error notice will show up, this is not convenient because it’s hard to immediately make that relation, and you would need to scroll to notice it, so a better design (something inline within the coupon should do it)

Example:
![image](https://user-images.githubusercontent.com/6165348/76630466-5533ec80-6540-11ea-8b5e-9dbf08d7b817.png)

**Edit** (added by Darren): 

Also add snackbar notices for the coupon add success and remove coupon actions.

Issue #1939, opened 2020-03-11
Checkout Context and Validation: Improve `onCheckoutProcessing` event emitter to allow for aborted subscribers.


nerrad:
The `onCheckoutProcessing` event handler in the checkout context will get used by validation routines or processes that _can abort the checkout being processed_.

For reference, see the flow as outlined here:

<img width="1012" alt="Image 2020-03-11 at 9 32 28 AM" src="https://user-images.githubusercontent.com/1429108/76422218-3d5f3b80-637b-11ea-9b99-2e04ed24f777.png">

This means:

- eventEmitter must handle aborting running any subsequent subscribed observers as soon as an observer returns an error object (anything but `true`).
- `onCheckoutProcessing` will handle error objects that are returned from aborted events and update relevant checkout statuses and error states.

## Suggested apis (roughly):

This has the overall idea, but there are still some specifics that likely will need worked out as a part of the implementation.

A subscriber triggering a processing abort would be expected to return an error object in the following shape:

```js
const error = {
    message: 'Main error message',
    validation: {
        propertyFailingValidation: 'error message for field',
        anotherPropertyFailingValidation: 'error message for field',
    }
};
```

The `message` property will be output in an error notice by the checkout context.
The `validation` property is for errors triggered by validation fails. The value is an object where properties correspond to field properties/ids for existing form fields and values correspond to the validation error message to accompany the field error.

Implementation wise, the errors can be exposed on the checkout context and field components can grab that (via context hook) to determine whether there is an error and whether there are any matching validation properties for the field the component outputs.

## Additional considerations

- I don't anticipate there would be a lot of subscribers hooked into the `onCheckoutProcessing` hook, but I do wonder if instead of aborting after a single subscriber fail, if we should accumulate all the error objects and aggregate them? That might be a better customer experience for fixing all the potential issues rather than fixing for one subscriber and then getting hit by the next subscriber after?

nerrad:
Followups from #1972:

- increase [specificity of the selector](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1972#discussion_r393515086) in `text-input/style.css` to remove the need for `!important`
- As [noted by Nadir](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1972#pullrequestreview-375959258), there's an alignment issue with the invalid coupon message in the context of the checkout.

PR #1972, opened 2020-03-14
Add validation context provider and implement validation for shipping country and coupons.

nerrad:
Fixes: #1969
Fixes: #1966

Some of the work in this pull can also help with #1939 (cc @Aljullu )

As a part of our work on the checkout and cart blocks, we need a way to validate fields, show validation errors, and track validation state. In this pull I've tackled validation for two things:

- Give an indication to customer when they fill our address fields but there is no country selected.
- Give an indication to customer when invalid coupon is applied (currently this is via notice, but we can do this via inline validation to avoid scroll).

I thought these two use-cases would be involve straightforward validation logic giving opportunity to validate (!pun) an approach I had in mind for using context to help with validation.

So in this pull is the following:

- Addition `ValidationContextProvider` and `useValidationContext`. This provides state for validation errors which are stored as an object where properties are the field the validation error applies to and values are the user friendly message to show to customer's when a validation fails.
- Addition of `ValidationInputError` component. This component is used in field components wherever you want a validation error to appear. It applies a consistent style and format to the error message.
- Implemented the new `ValidationInputError` component with the `CountryInput` and also created a storybook for the `CountryInput` so that the we could see the validation at work!  Note: There is a console.error with the story whenever a country is selected. The story is still completely functional and this console error doesn't appear in actual use. I think it's a storybook bug because the error doesn't make sense (might have to do with decorators being used).

Screen Recording 2020-03-13 at 10 09 PM

- Implemented the new `ValidationInputError` component with the `TotalsCouponCodeInput` component and also created a story book for this.
- Various improvements in related code touched in this pull (I'll note some of them via inline comments in the code).

Screen Recording 2020-03-14 at 06 21 PM

## To Test

- Verify cart loads in editor and frontend without errors
- Verify checkout loads in editor and frontend without errors.
- **Country validation:** Set your site so that "Default Customer Location" is set to "No location by default" and then load an incognito browser view. In the shipping calculator select/enter in a field without selecting country. A validation error should appear and the update button should be disabled. As soon as you select country the validation error should disappear and the button re-enabled. Verify shipping rates etc work as before.
- **Coupon validation:** Coupon notices for coupon errors have been replaced by the validation error. Verify this works as expected and coupon behaviour for valid coupons works as expected too.
- start up storybook and you can see components and their behaviour in storybook as well (`CountryInput` and `TotalsCouponCodeInput`)

Observer styles etc and report on any issues or unexpected behaviour (see gifs for what is expected).


assets/js/base/components/text-input/style.scss
```diff
@@ -2,6 +2,18 @@
 	position: relative;
 	margin-top: $gap;
 	white-space: nowrap;
+
+	input {
+		border: 1px solid $input-border-gray;
+	}
+
+	&.has-error input {
+		border: 1px solid $error-red;
```
nerrad: This is needed because storefront sets the border to 0 and adds a shadow to inputs.  So I think this should be an explicit rule that overrides _any_ themes setting for input border (on errors).

assets/js/base/components/totals/totals-coupon-code-input/index.js
```diff
@@ -62,18 +67,29 @@ const TotalsCouponCodeInput = ( {
...
-							label={ __(
-								'Enter code',
-								'woo-gutenberg-products-block'
-							) }
-							value={ couponValue }
-							onChange={ ( newCouponValue ) =>
-								setCouponValue( newCouponValue )
-							}
-						/>
+						<div>
```
nerrad: needed to add the extra container here so that the validation error would go below the text input for this row.

assets/js/base/hooks/cart/use-store-cart-coupons.js
```diff
@@ -56,9 +62,9 @@ export const useStoreCartCoupons = () => {
 						}
 					} )
 					.catch( ( error ) => {
-						addErrorNotice( error.message, {
-							id: 'coupon-form',
-						} );
+						setValidationErrors( { coupon: error.message } );
+						// Finished handling the coupon.
+						receiveApplyingCoupon( '' );
```
nerrad: I needed to move setting the loading state for `applyingCoupon` outside of the `applyCoupon` action (for errors only) because throwing an error _after_ setting the loading state in the action meant that validation errors were set _after_ the the loading is `false`.  In turn this meant that in the `TotalsCouponCodeInput` logic, the invalid coupon code would always be cleared from the input.

I tried various different approaches (including using the cartData error state) but this seemed to be the cleanest. The downside is any code catching errors will _have_ to explicitly cancel the `receiveApplyingCoupon` error state. I think that's okay because `useStoreCartCoupons` is generally the interface components will go through anyways.

storybook/style.scss
```diff
@@ -7,3 +7,26 @@
...
+	overflow: hidden;
+	padding: 0;
+	position: absolute;
+	width: 1px;
+	word-wrap: normal !important; /* many screen reader and browser combinations announce broken words as they would appear visually */
+}
+
+.button .screen-reader-text {
+	height: auto; /* Fixes a Safari+VoiceOver bug, see ticket #42006 */
+}
```
nerrad: This is a css class defined in the WP context (which storybook doesn't have access too) so this ensures all storybook components have access to this css.

storybook/webpack.config.js
```diff
@@ -2,18 +2,31 @@
...
+	( acc, key ) => {
+		const currentPath = tsConfig.compilerOptions.paths[ key ][ 0 ];
+		acc[ key.replace( '/*', '' ) ] = path.resolve(
+			__dirname,
+			'../' + currentPath.replace( '/*', '/' )
+		);
+		return acc;
+	},
+	{}
+);
```
nerrad: I switched to this configuration for getting aliases in the storybook webpack config because it's a bit more resilient than the other. I kept getting errors from using the base webpack config because it doesn't define all aliases (due to built bundles on externals). Whereas typescript does.  Typescript configuration for aliases is also pretty much the same as what storybook needs, so this aligns with that.

assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -92,251 +95,255 @@ const Block = ( {
...
-										...contactFields,
-										keepUpdated: ! contactFields.keepUpdated,
-									} )
-								}
-							/>
-						</FormStep>
-						{ SHIPPING_ENABLED && (
+		<ValidationContextProvider>
+			<CheckoutProvider isEditor={ isEditor }>
+				<SidebarLayout className="wc-block-checkout">
```
nerrad: `ValidationContextProvider` is wrapping `CheckoutProvider` so that the validation context is available to the checkout processing event.

assets/js/base/components/text-input/style.scss
mikejolley: How about trying outline instead? Outlines don't affect layout.

assets/js/base/components/totals/totals-coupon-code-input/index.js
mikejolley: Why can't this be solved with CSS display block? Is it due to flexbox? There are ways to work around that even with flexbox if we want cleaner markup.

That said, from accessibility standpoint we may want to link these elements somehow. Could we pass it down to TextInput to render?

mikejolley: That would also let us handle classnames in the TextInput component instead of doing it here.

assets/js/blocks/cart-checkout/cart/frontend.js
```diff
@@ -32,12 +35,16 @@ const Block = ( { emptyCart, attributes } ) => {
 				<RawHTML>{ emptyCart }</RawHTML>
 			) : (
 				<LoadingMask showSpinner={ true } isLoading={ cartIsLoading }>
-					<FullCart
-						isShippingCalculatorEnabled={
-							attributes.isShippingCalculatorEnabled
-						}
-						isShippingCostHidden={ attributes.isShippingCostHidden }
-					/>
+					<ValidationContextProvider>
```
mikejolley: I wonder if we should somehow link this context provider to our `<form>` components via a new component that implements it. I'm assuming validation and forms go hand in hand.

assets/js/data/cart/actions.js
```diff
@@ -98,8 +98,6 @@ export function* applyCoupon( couponCode ) {
 	} catch ( error ) {
 		// Store the error message in state.
 		yield receiveError( error );
-		// Finished handling the coupon.
-		yield receiveApplyingCoupon( '' );
```
mikejolley: Why was this removed?

assets/js/type-defs/contexts.js
```diff
@@ -215,4 +215,34 @@
  * @property {number}                  currentPostId      The post ID being edited.
  */
 
+/**
+ * @typedef {Object} ValidationContext
+ *
+ * @property {function(string):string}  getValidationError       Return
```
mikejolley: The column widths really make this harder to read :)

assets/js/base/context/cart-checkout/validation/index.js
```diff
@@ -0,0 +1,92 @@
...
+	 * This retrieves any validation error message that exists in state for the
+	 * given property name.
+	 *
+	 * @param {string} property The property the error message is for.
+	 *
+	 * @return {string} Either the error message for the given property or an
+	 *                  empty string.
+	 */
+	const getValidationError = ( property ) =>
+		validationErrors[ property ] || '';
```
Aljullu: No need to change it, but just curious, why returning an empty string instead of `null` when there is no error?

assets/js/base/components/validation/style.scss
```diff
@@ -0,0 +1,6 @@
+.wc-block-form-input-validation-error {
+	color: $error-red;
+	@include font-size(12);
+	padding-left: $gap-small;
```
Aljullu: Currently it looks a bit misaligned, the error notice left padding size is between the border and the input text:
imatge
I would either remove the `padding-left` completely, or make it like it is in the text input CSS (`$gap` ).

assets/js/base/components/totals/totals-coupon-code-input/index.js
Aljullu: Actually, I would move it below so the validation error spans under the button. This way I think text would not break in two lines.
imatge

assets/js/base/components/text-input/style.scss
nerrad: Outline looks bad (focus is lost on applying the coupon so it will be with non-rounded corners initially).

Also tried shadow as per Mike's suggestion but I think it looks bad too so just went with forcing the border using `!important`.  We can revisit styles in our styles review as necessary but this seems to be the better option here because I think we want to make sure this error condition is clear (and make it harder for themes to override).

assets/js/base/components/totals/totals-coupon-code-input/index.js
nerrad: Done.

assets/js/blocks/cart-checkout/cart/frontend.js
nerrad: I like the idea and it makes sense, however that would prevent the checkout context (for processing checkout) to be aware of validation. So to implement this we'd probably have to move validation to a global store.  For now, I'd like to roll with this because it has flexibility across the Cart (which doesn't have checkout context) and checkout (which does have checkout context).  

We can look at iterating on this to make it more form specific (I'd like to get a better idea of how payment methods might implement validation and make checkout aware of that before committing to another approach).

assets/js/data/cart/actions.js
nerrad: I'm assuming you saw the comment I [left here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1972#discussion_r392603600) about this (since you acknowledged it with a thumbs up emojii). Is that sufficient explanation?

assets/js/type-defs/contexts.js
nerrad: Agreed. I've added another editor ruler at the `100` column in my IDE so that when 80 is hit for comments and it's like this I can extend the comments to the 100 ruler instead (which I've applied as a commit here).

assets/js/base/context/cart-checkout/validation/index.js
nerrad: Mostly to preserve the type. Since empty string can be used for truthy tests I prefer to keep the return value the same type.

assets/js/base/components/validation/style.scss
nerrad: Agreed, and fixed.

nerrad:
> Should this behavior also be applied to the checkout form? (not a blocker) since the coupon one is also applied there.

It should, I reworked where the validation happens so it works for both checkout and cart now. However, I'm not sure how we should be handling the css for this:

Image 2020-03-16 at 1 39 41 PM

Ideas welcome.

LevinMedia:
Looks good to me for now, we can fiddle with the styles to see if we can shrink the gaps in a follow up.

assets/js/base/components/shipping-calculator/address.js
```diff
@@ -19,17 +20,25 @@ const ShippingCalculatorAddress = ( {
 	addressFields,
 } ) => {
 	const [ address, setAddress ] = useState( initialAddress );
+	const { getValidationError } = useValidationContext();
+
+	const updateAddress = ( newAddress ) => {
+		setAddress( newAddress );
+	};
```
Aljullu: Is this function needed anymore?

assets/js/base/components/text-input/style.scss
```diff
@@ -1,7 +1,23 @@
 .wc-block-text-input {
 	position: relative;
-	margin-top: $gap;
+	margin-bottom: $gap-large;
 	white-space: nowrap;
+
+	&.has-error input {
```
Aljullu: Just an idea, but what do you think about increasing the specificity of this selector like this:
```suggestion
	&.has-error input[type] {
```
This way we can remove the `!important` below. We would still need a comment explaining `[type]` is there for specificity, but that way we avoid an `!important` rule.

assets/js/base/components/shipping-calculator/address.js
nerrad: Nope, good catch, we could just use `setAddress` directly.

assets/js/base/components/text-input/style.scss
nerrad: Ya that makes sense, I'll leave for #1939

nerrad:
> In a more personal side, I feel the error in the Country input should appear when you hit Place order instead of happening when filling other address fields.

So I'm assuming you'd feel the same way about the behaviour in the shipping calculator too?  

Keep in mind in the checkout, the shipping field values will trigger a shipping rates update. So if we don't give feedback, how will the user know to add a country?

PR #1973, opened 2020-03-15
Add snackbar notices to notices system and implement for coupon ui/ux

nerrad:
Fixes: #1966 (in part)

This finishes off the remainder of #1966.  In this pull:

- I added Snackbar notices for the Notices system. This utilizes the `SnackbarList` component from `@wordpress/components` (aliased in directly).
- I implemented the new snackbar notices with the coupon ui/ux. Snackbar notices are now used for a coupon added successfully, and a coupon removed successfully.

Screen Recording 2020-03-14 at 09 44 PM

Mobile:

Screen Recording 2020-03-15 at 09 13 AM

## To test:

- verify coupon logic works as expected
- review notice behaviour - any tweaks, changes needed? 
- this should probably be tested across browsers (I tested Chrome).
- Accessibility should be tested (I think this applies a11y speak automatically - "polite").
- test notices on mobile.


nerrad:
I realize this bumps up bundlesizes, I think this is mostly because of the inclusion of react-spring as a part of the `SnackbarList` components (and vendors jumps because of that component pulling in `react-spring`). We will have to evaluate whether we ship with this - I'm not sure if I want to make that decision with this pull.

assets/js/base/components/store-notices-container/style.scss
```diff
@@ -24,3 +24,88 @@
 		margin-top: 1em;
 	}
 }
+
+/** Snackbar notices styles **/
+.components-snackbar {
```
haszari: Is this styling for (cart/checkout) notice snackbars, or all snackbars anywhere? Should probably have a class targeting these styles to `store-notices-container`, or factor out our own woo frontend snackbar component (if this is intended to be used for things other than notices).

Also, do we get any styling for free from the `SnackbarList` component? It looks like we're reinventing snackbar styling here.

assets/js/base/context/store-notices-context.js
```diff
@@ -52,6 +56,16 @@ export const StoreNoticesProvider = ( {
 		[ createNotice, context ]
 	);
 
+	const createSnackbarNotice = useCallback(
```
haszari: I'm wondering about whether it's nicer to expose snackbars using the same API as the existing notices - i.e. do we need `createSnackbarNotice` as a convenience?

assets/js/base/hooks/use-store-notices.js
```diff
@@ -2,44 +2,49 @@
...
+		addWarningNotice: ( text, noticeProps = {} ) =>
+			void createNotice( 'warning', text, {
+				...noticeProps,
+			} ),
+		addInfoNotice: ( text, noticeProps = {} ) =>
+			void createNotice( 'info', text, {
+				...noticeProps,
+			} ),
+		addSuccessNotice: ( text, noticeProps = {} ) =>
+			void createNotice( 'success', text, {
```
haszari: Related to above (single API for all notices) – I'm wondering if we can smart-default the UI based on severity here, to encourage consistent use. So `addErrorNotice` would default to notice, `addSuccessNotice` would default to snackbar, etc. This could still potentially be overridden if needed, by passing `noticeProps = { type: 'snackbar' }` or `noticeProps = { type: 'notice' }`.

This interface is probably stretching the meaning of the term "notice" though.

Not a blocker, just exploring this as a potential improvement to the API we expose to clients.

nerrad: Currently I'm following the pattern established by Gutenberg for the [usage of Snackbars](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/snackbar) which doesn't involve varying degrees of styles or behaviour.

So with that in mind,  it seems pretty clear to me that snackbar notices have an explicit purpose that differs from that of regular notices.  Whether that means all success and/or info notices will always be snackbar notices is not so clear to me so I think it's best to keep them distinct for now.

assets/js/base/context/store-notices-context.js
nerrad: See my comment [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1973#discussion_r392724837)

assets/js/base/components/store-notices-container/style.scss
nerrad: > Is this styling for (cart/checkout) notice snackbars, or all snackbars anywhere?

This isn't a component specific to cart/checkout, so everywhere (in the frontend).

> Also, do we get any styling for free from the SnackbarList component? It looks like we're reinventing snackbar styling here.

Ya I thought we'd get that styling for free with our current webpack config but it doesn't look like we are.  I'm guessing because gutenberg doesn't directly import the styles but has them pulled in via build process instead (so our webpack config isn't picking them up).  

I definitely would rather pull in the styles from the package. @Aljullu do you have any ideas around this?

nerrad:
> There is a small scroll jump when a snackbar appears or disappears (in Safari and Firefox).

Ya I've seen this too, my css fu isn't really great at addressing this, not sure what to do here (@Aljullu do you have any ideas?)

> Error notice for bad coupon is not visible on mobile (I'm guessing this is existing behaviour).

Ya existing, is fixed in #1972 

> On macOS Safari, I couldn't get VoiceOver to read out the snackbars. (I don't have a lot of experience with VoiceOver / speak-a11y testing).

Hmm, snackbars _should_ be [using `wp.a11y.speak`](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/snackbar/index.js#L53).  I confirmed the notices weren't spoken - I think it's because it defaults to `polite` for the politeness attribute which means it only is spoken if there is no other activity being spoken.  However, I tried forcing it to `assertive` and voiceover still didn't pick up the snackbar content. So I'm not sure why here. The `aria-live` containers _are_ on the page.  One thing I did notice is that if I do `wp.a11y.speak( 'test content', 'assertive' );` in the browser console, voiceover isn't picking that up either (even though I see the aria-live container updated). So this suggests that maybe voiceover doesn't pick up aria-live regions? (possibly confirmed by this [stack overflow article](https://stackoverflow.com/questions/45821764/assertive-aria-live-region-not-read-by-voice-over-on-page-load-form-submit))

assets/js/base/components/store-notices-container/style.scss
Aljullu: @nerrad you need to specify the component's CSS file in the webpack entry config:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/bin/webpack-helpers.js#L107-L114

assets/js/base/components/store-notices-container/store-notices-container.js
```diff
@@ -0,0 +1,69 @@
...
+
+	if ( ! notices.length ) {
+		return null;
+	}
+
+	const regularNotices = filter(
+		notices,
+		( notice ) => notice.type !== 'snackbar'
+	);
+
```
Aljullu: The issue with scroll jumping when a snackbar appeared or disappeared came from here: after the `filter` we need to return if `regularNotices.length === 0`. Otherwise, we are rendering an empty `.wc-block-components-notices`.

assets/js/base/components/store-notices-container/store-notices-container.js
```diff
@@ -0,0 +1,69 @@
...
+const StoreNoticesContainer = ( { className, notices } ) => {
+	const { removeNotice } = useStoreNoticesContext();
+	const wrapperClass = classnames( className, 'wc-block-components-notices' );
+
+	if ( ! notices.length ) {
+		return null;
+	}
+
+	const regularNotices = filter(
```
Aljullu: I know we get lodash for free, but for methods which are equivalent in raw JavaScript, I usually feel it's better to use the standard method instead of lodash's.

assets/js/base/components/store-notices-container/store-notices-container.js
```diff
@@ -0,0 +1,69 @@
...
+
+	return (
+		<div className={ wrapperClass }>
+			{ regularNotices.map( ( props ) => (
+				<Notice
+					key={ 'store-notice-' + props.id }
+					{ ...props }
+					className={ classnames(
+						'wc-block-components-notices__notice',
+						getWooClassName( props )
```
Aljullu: This method always returns the class `woocommerce-message` + another one.

What would you think about declaring `woocommerce-message` here and renaming `getWooClassName` to `getStatusClassName`?

assets/js/base/components/store-notices-container/store-notices-container.js
```diff
@@ -0,0 +1,69 @@
...
+					{ props.content }
+				</Notice>
+			) ) }
+		</div>
+	);
+};
+
+StoreNoticesContainer.propTypes = {
+	className: PropTypes.string,
+	notices: PropTypes.array,
```
Aljullu: This propType could be a bit more precise:
```suggestion
	notices: PropTypes.arrayOf(
		PropTypes.shape( {
			content: PropTypes.string.isRequired,
			id: PropTypes.string.isRequired,
			status: PropTypes.string.isRequired,
			isDismissible: PropTypes.bool,
			type: PropTypes.oneOf( [ 'default', 'snackbar' ] ),
		} )
	),
```

Aljullu:
> On macOS Safari, I couldn't get VoiceOver to read out the snackbars. (I don't have a lot of experience with VoiceOver / speak-a11y testing).

I tested with Firefox + Orca (Linux screen reader) and snackbars and default notices were both read properly.

LevinMedia:
This is looking great! I haven't had a chance to get this set up locally yet, how long does the snackbar persist until it disappears?

Also noticing the small jump behavior that @haszari noted in the GIF above. Would be great to smooth that out.

assets/js/base/components/store-notices-container/style.scss
nerrad: Thanks for the tip Albert, that did the trick 🌮

assets/js/base/components/store-notices-container/store-notices-container.js
nerrad: Yup another good catch! Thanks :)

assets/js/base/components/store-notices-container/store-notices-container.js
nerrad: Agreed. I had already replaced this but forgot to commit and push. Thanks for pointing this out though because we definitely should avoid lodash if it has a clear replacement like this :)

assets/js/base/components/store-notices-container/store-notices-container.js
nerrad: Works for me!

nerrad:
All feedback is addressed. Thanks for the help with things as well Albert, the scroll jump is resolved and the css import from `wordpress-components` works nice! Also thanks for validating that a11y.speak behaves as expected for the snackbar.

nerrad:
> how long does the snackbar persist until it disappears?

@LevinMedia It's currently the [default timeout](https://github.com/WordPress/gutenberg/blob/master/packages/components/src/snackbar/index.js#L20) which is 10s.

bin/webpack-helpers.js
```diff
@@ -112,6 +112,8 @@ const stableMainEntry = {
 		'./node_modules/@wordpress/components/src/checkbox-control/style.scss',
 	'spinner-style':
 		'./node_modules/@wordpress/components/src/spinner/style.scss',
+	'snackbar-notices':
+		'./node_modules/@wordpress/components/src/snackbar/style.scss',
```
haszari: The name for this isn't following the pattern of the others (e.g. `snackbar-style`), but I don't think that matters too much.

nerrad: Doesn't matter but I favor consistency (yet somehow missed this), so thanks for pointing it out 👍

PR #1926, opened 2020-03-10
Implement "return to cart" and "place order" buttons on checkout

mikejolley:
Adds a return to cart link, and the place order button.

**Return to cart**
- Can be disabled in block options
- URL can be controlled in block options
- URL defaults to cart page set in WC

**Place order**
- Standard button with styling and callback
- Gets the callback action and text from checkout context
- I have added a TODO for the actual action. The button is non-functional in this PR

**Misc**
Whilst styling these new buttons, I also styled some of the margins on the rest of the form to match the mockup. This was mostly needing to align things vertically.

Fixes #1303

### Screenshots

Screenshot 2020-03-10 at 14 19 34
Screenshot 2020-03-10 at 14 19 17

### How to test the changes in this Pull Request:

1. Go to the checkout
2. Check appearance of link and button
3. Edit block (might be invalidated)
4. Confirm you can toggle the cart link on/off and change the URL.

PR #1975, opened 2020-03-16
Record a tracks event when merchant toggles cart shipping calculator:

### LINKED ISSUE #1273: Cart & Checkout Block: Add Data Tracking
In order to quickly and properly measure the cart & checkout blocks impact it would be important to have an event triggered when the cart and checkout pages are published with the blocks.

If it's not possible to add the tracking we'll need to rethink the release strategy of v1.

- There's a good guide about Tracks in the Fieldguide: /tracks
- A couple of months ago it was implemented an event tracking library for the block editor, but it seems to be exclusive to .com. More info here: p7H4VZ-2hb-p2

haszari:
Related #1273

This PR adds a utility function for triggering tracks events in blocks plugin, specifically for editor events for our blocks. This a wrapper for `window.wcTracks.recordEvent` which adds a standard prefix (so we can easily wrangle blocks events), and some standard props. I've added `post_type` and `post_id` props as an example; we can iterate on what standard props we want as we add events.

I've added one example event `cart_shipping_calculator_toggled`, with a boolean prop for whether the post author is toggling it on or off.

#### Test `cart_shipping_calculator_toggled` event

1. Enable usage tracking: `WooCommerce > Settings > Advanced > WooCommerce.com`.
2. Add a page and add cart block (or edit page containing cart block). 
3. Open dev tools Network tab, search for `pixel.wp.com` (so you can see events).
2. Toggle the shipping calculator option. You should see a network request for `cart_shipping_calculator_toggled` event; check the props are all correct and sensible.

Screen Shot 2020-03-23 at 5 54 43 PM

Note: screenshot taken with Woo core https://github.com/woocommerce/woocommerce/pull/25977 - so the event also has new standard props `woo_version` and `url`.

#### Test the `recordEditorEvent` interface (optional)
- Try adding an event for some aspect of one of the Woo blocks. 
- Does `recordEditorEvent` work for your use case? 
- What custom props does your event need?
- What default props should we add to all events?

### DIFF
--- a/assets/js/base/utils/tracks.js
+++ b/assets/js/base/utils/tracks.js
@@ -3,19 +3,17 @@
  */
 import { select } from '@wordpress/data';
 
-function isTracksAvailable() {
-	return typeof window.wcTracks.recordEvent === 'function';
-}
+// Stand-in wcTracks.recordEvent in case tracks is not available (for any reason).
+window.wcTracks = window.wcTracks || {};
+window.wcTracks.recordEvent = window.wcTracks.recordEvent || function() {};
 
 export const recordEditorEvent = function( event, props ) {
-	if ( ! isTracksAvailable() ) {
-		return;
-	}
-
 	// Force prefix - our editor events will be 'wcadmin_blocks_*'.
 	const blocksPrefix = 'blocks_';
 
 	const postData = select( 'core/editor' );
+
+	// If tracks is not available (e.g. opt-out), recordEvent is a no-op.
 	window.wcTracks.recordEvent( blocksPrefix + event, {
 		// Automatically include post id and post type props.
 		post_id: postData.getCurrentPostId(),
--- a/assets/js/base/utils/tracks.js
+++ b/assets/js/base/utils/tracks.js
@@ -8,7 +8,7 @@ function isTracksAvailable() {
 }
 
 export const recordEditorEvent = function( event, props ) {
-	if ( ! isTracksAvailable ) {
+	if ( ! isTracksAvailable() ) {
 		return;
 	}
 
--- a/assets/js/base/utils/tracks.js
+++ b/assets/js/base/utils/tracks.js
@@ -1,0 +1,25 @@
+/**
+ * External dependencies
+ */
+import { select } from '@wordpress/data';
+
+function isTracksAvailable() {
+	return typeof window.wcTracks.recordEvent === 'function';
+}
+
+export const recordEditorEvent = function( event, props ) {
+	if ( ! isTracksAvailable ) {
+		return;
+	}
+
+	// Force prefix - our editor events will be 'wcadmin_blocks_*'.
+	const blocksPrefix = 'blocks_';
+
+	const postData = select( 'core/editor' );
+	window.wcTracks.recordEvent( blocksPrefix + event, {
+		// Automatically include post id and post type props.
+		post_id: postData.getCurrentPostId(),
+		post_type: postData.getCurrentPostType(),
+		...props,
+	} );
+};
--- a/assets/js/blocks/cart-checkout/cart/edit.js
+++ b/assets/js/blocks/cart-checkout/cart/edit.js
@@ -87,9 +87,12 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 					) }
 					checked={ isShippingCalculatorEnabled }
 					onChange={ () => {
-						recordEditorEvent( 'cart_shipping_calculator_toggled', {
-							enabled: ! isShippingCalculatorEnabled,
-						} );
+						recordEditorEvent(
+							'cart_settings_shipping_calculator_toggle',
+							{
+								enabled: ! isShippingCalculatorEnabled,
+							}
+						);
 						setAttributes( {
 							isShippingCalculatorEnabled: ! isShippingCalculatorEnabled,
 						} );
--- a/assets/js/blocks/cart-checkout/cart/edit.js
+++ b/assets/js/blocks/cart-checkout/cart/edit.js
@@ -19,6 +19,7 @@ import { EditorProvider, useEditorContext } from '@woocommerce/base-context';
 import { useSelect } from '@wordpress/data';
 import { __experimentalCreateInterpolateElement } from 'wordpress-element';
 import { getAdminLink } from '@woocommerce/settings';
+import { recordEditorEvent } from '@woocommerce/base-utils';
 
 /**
  * Internal dependencies
@@ -85,11 +86,15 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 						'woo-gutenberg-products-block'
 					) }
 					checked={ isShippingCalculatorEnabled }
-					onChange={ () =>
+					onChange={ () => {
+						const newEnabled = ! isShippingCalculatorEnabled;
+						recordEditorEvent( 'cart_shipping_calculator_toggled', {
+							enabled: newEnabled,
+						} );
 						setAttributes( {
-							isShippingCalculatorEnabled: ! isShippingCalculatorEnabled,
-						} )
-					}
+							isShippingCalculatorEnabled: newEnabled,
+						} );
+					} }
 				/>
 				<ToggleControl
 					label={ __(
--- a/assets/js/blocks/cart-checkout/cart/edit.js
+++ b/assets/js/blocks/cart-checkout/cart/edit.js
@@ -87,12 +87,11 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 					) }
 					checked={ isShippingCalculatorEnabled }
 					onChange={ () => {
-						const newEnabled = ! isShippingCalculatorEnabled;
 						recordEditorEvent( 'cart_shipping_calculator_toggled', {
-							enabled: newEnabled,
+							enabled: ! isShippingCalculatorEnabled,
 						} );
 						setAttributes( {
-							isShippingCalculatorEnabled: newEnabled,
+							isShippingCalculatorEnabled: ! isShippingCalculatorEnabled,
 						} );
 					} }
 				/>
--- a/assets/js/base/utils/index.js
+++ b/assets/js/base/utils/index.js
@@ -1,3 +1,4 @@
 export * from './errors';
 export * from './price';
 export * from './address';
+export * from './tracks';


haszari:
When #1926 lands we might be able to use [`storePages`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1926/files#diff-d8fbf540e343717e9a56b10e0c5de5b3R146) to add a tracks prop for if the current editor page is the woo cart page (etc).

This could be a useful prop on cart/checkout tracks events, for example:

```
event_name: 'blocks_cart_added',
is_woo_cart_page: true,
```

assets/js/base/utils/tracks.js
```diff
@@ -0,0 +1,25 @@
...
+ * External dependencies
+ */
+import { select } from '@wordpress/data';
+
+function isTracksAvailable() {
+	return typeof window.wcTracks.recordEvent === 'function';
+}
+
+export const recordEditorEvent = function( event, props ) {
+	if ( ! isTracksAvailable ) {
```
nerrad: Should be calling a function here?

assets/js/base/utils/tracks.js
```diff
@@ -0,0 +1,25 @@
...
+	const blocksPrefix = 'blocks_';
+
+	const postData = select( 'core/editor' );
+	window.wcTracks.recordEvent( blocksPrefix + event, {
+		// Automatically include post id and post type props.
+		post_id: postData.getCurrentPostId(),
+		post_type: postData.getCurrentPostType(),
+		...props,
+	} );
+};
```
nerrad: Some things to consider here:

- This utility is tightly coupled to recording events in the context of the `core/editor` data store and the post/post-type values.  As we start building interfaces outside of the WP editor are we going to build utility functions like this for every interface? What if we want to track other things about the editor instance beyond post_id and post_type?
- This is using a data selector that is not subscribed. In most cases (like this one), this will probably be okay (because we only _want_ the values at the time of select). But we do need to keep in mind it will only select the data at the time of tracking the event.

With those considerations, I think it might be good to implement tracks as a custom React hook. You could have something like this:

```jsx
export const useTracks = ( event, trackCallback ) => {
  if ( typeof trackCallback !== 'function' ) {
    throw new Error( 'trackCallback must be a function' );
  }
  const recordEvent = useCallback( (...data ) => {
    const recordedData = trackCallback( ...data );
    if ( ! isTracksAvailable() ) {
      return;
    }
    window.wcTracks.recordEvent( `blocks_${ event }`, recordedData );
  }, [ event, trackCallback ] );
  return recordEvent;
}

// Example implementation of hook
const MyComponentWithTracks = ( props ) => {
  const recordEvent = useTracks("cart_shipping_calculator_toggled", ( trackedProps ) => {
    const postData = select( 'core/editor' );
    return {
      post_id: postData.getCurrentPostId(),
      post_type: postData.getCurrentPostType(),
      ...trackedProps
    };
  });
  /** where tracking happens later in component */
  recordEvent( { enabled: ! isShippingCalculatorEnabled } );
```

The `trackCallback` then can be something that is extracted out for re-usability if needed but the hook allows for composition of various track events.

Unless there are reasonable objections, I'd like us to try this route (custom hook) and see how it works. If it works well, we should create a tracks package in the wc-admin repo that exposes the hook (and possibly other tracks utils) for use by all Woo projects so that there's a common mechanism for implementing tracks in our projects.

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -39,11 +40,15 @@ const CartEditor = ( { className, attributes, setAttributes } ) => {
 						'woo-gutenberg-products-block'
 					) }
 					checked={ isShippingCalculatorEnabled }
-					onChange={ () =>
+					onChange={ () => {
+						const newEnabled = ! isShippingCalculatorEnabled;
```
nerrad: I don't think this const declaration is necessary. You can just use `! isShippingCalculatorEnabled` in both places it is used.

assets/js/base/utils/tracks.js
haszari: I had a quick look at setting up a custom hook, but it didn't seem to be offering anything useful. I can imagine this might be a useful approach as we iterate on this, but am struggling to justify this approach right now.

Responding to specific points:

> This utility is tightly coupled to recording events in the context of the core/editor data store and the post/post-type values. As we start building interfaces outside of the WP editor are we going to build utility functions like this for every interface? 

This is intentional - this API is just for Woo Blocks plugin for now, and for triggering events in the block editor (which is the bulk of what we need to do). I.e. scoped to this plugin and its needs for now.

I like the idea of generalising this to other use cases and exporting an API from wc-admin, but I'm not convinced we need to solve this / prototype this now. I'm focused on the immediate need of sending events for woo blocks editor features.

And just to clarify: the common props (e.g. `post_id`) and custom per-event props (`enabled`) in this PR are not set in stone, these are some examples to get us started. As we flesh out the events it will become clearer what consistent data we want on all block-editor events, vs what's custom per event. We may also want to add common props to all tracks events, further up the chain in Woo core (e.g. `url`, `woo_version`).

> As we start building interfaces outside of the WP editor are we going to build utility functions like this for every interface?

Yes ish – wc-admin already has a working solution for much of the rest of admin screens. This adds a specific wrapper to that for blocks editor events.

> This is using a data selector that is not subscribed. In most cases (like this one), this will probably be okay (because we only want the values at the time of select). But we do need to keep in mind it will only select the data at the time of tracking the event.

I don't see this as a problem – the event props should be populated with data that matches what's happening at the time of the event. Do you have an event or use-case in mind? With `cart_shipping_calculator_toggled` example, the props could be passed in at event time, or supplied as defaults, I don't think there's a need to supply these via callback.

I can imagine that patterns might emerge as we implement more events – perhaps we can iterate to a custom hook or other abstraction as we go.

Let's have a chat about this – I want to understand better what the advantages would be of using a hook (I may be missing something).

assets/js/base/utils/tracks.js
haszari: Yes!! Good spotting, fixed in 05fe5a98b42bd1fe61698c68435b57321180aa9d

assets/js/blocks/cart-checkout/cart/edit.js
haszari: I've made that change here: b7e11874b95b14ca018b3ae2bf14d50b86cbffbf

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -85,11 +86,14 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 						'woo-gutenberg-products-block'
 					) }
 					checked={ isShippingCalculatorEnabled }
-					onChange={ () =>
+					onChange={ () => {
+						recordEditorEvent( 'cart_shipping_calculator_toggled', {
+							enabled: ! isShippingCalculatorEnabled,
+						} );
```
mikejolley: This would have to be implemented inline for everything we want to track right? Right now it's only one attribute, but what if there are many attributes in the same component to track?

One solution would be to use `useEffect` and watch for changes to attributes but that might be tricky to pick out specific changes.

Another idea would be a wrapper for `setAttributes` so that anything using it could fire a tracks event then pass through the changes to `setAttributes`. That way the inline code changes are minimal and all tracking code would be together.

senadir: setAttribute is essentially a wrapper for setState, if you can possibly hook into the save function of the block or the post content and report from there, this could give more concrete results and could be implemented in the backend (avoid being blocked)

assets/js/base/utils/tracks.js
```diff
@@ -0,0 +1,25 @@
...
+	}
+
+	// Force prefix - our editor events will be 'wcadmin_blocks_*'.
+	const blocksPrefix = 'blocks_';
+
+	const postData = select( 'core/editor' );
+	window.wcTracks.recordEvent( blocksPrefix + event, {
+		// Automatically include post id and post type props.
+		post_id: postData.getCurrentPostId(),
+		post_type: postData.getCurrentPostType(),
```
mikejolley: I know this are examples, but whats the purpose in tracking Post ID and type? Is this the type of usage data we actually need? I thought we'd be more interested in how blocks themselves are interacted with. Including editor specific selects also limits future usage right?

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -85,11 +86,14 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 						'woo-gutenberg-products-block'
 					) }
 					checked={ isShippingCalculatorEnabled }
-					onChange={ () =>
+					onChange={ () => {
+						recordEditorEvent( 'cart_shipping_calculator_toggled', {
```
mikejolley: Do we have a standard in mind for event names? Good to get that in place asap.

assets/js/base/utils/tracks.js
mikejolley: Just looking at the function vs the hook suggestion, I immediately see a pro to the hook; `recordEditorEvent` forces the use of `const postData = select( 'core/editor' );`. If we ever want different data for the event you either have to edit this and affect all tracked events, or create more similar methods grabbing different sets of data.

With the hook, when you use it you can immediately override how the tracking event should work:

```
( trackedProps ) => {
    const postData = select( 'core/editor' );
    return {
      post_id: postData.getCurrentPostId(),
      post_type: postData.getCurrentPostType(),
      ...trackedProps
    };
  }
```
That's more flexible. Also to note, we wouldn't need to do this everywhere. We could have a default `trackCallback` that gets used when a custom one isn't provided.

Additionally, the suggested hook is currently just returning a single function:

```
return recordEvent;
```

This could be expanded to return other tracks related functions from the hook if we wanted. For example, one could create a wrapper for setAttributes and return that here to simplify some of the integration with components.

> As we flesh out the events it will become clearer what consistent data we want on all block-editor events, vs what's custom per event. We may also want to add common props to all tracks events, further up the chain in Woo core (e.g. url, woo_version).

That could be in the hook, given the suggested approach.

assets/js/base/utils/tracks.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * External dependencies
+ */
+import { select } from '@wordpress/data';
+
+function isTracksAvailable() {
+	return typeof window.wcTracks.recordEvent === 'function';
```
mikejolley: Just to check, is this only available after opt-in?

assets/js/base/utils/tracks.js
nerrad: Also to be clear, we could have a set of common track actions exported from the tracks package, for instance:

```js
const includePostdata = ( trackedProps ) => {
    const postData = select( 'core/editor' );
    return {
      post_id: postData.getCurrentPostId(),
      post_type: postData.getCurrentPostType(),
      ...trackedProps
    };
  };
```

And then implementation would be something like:

```jsx
import { useTracks, includePostData } from '@woocommerce/tracks';

const MyTrackedComponent = ( props ) => {
  const recordEvent = useTracks( 'cart_shipping_calculator_toggled', includePostData )
  const onClick = () => {
    recordEvent( { enabled: ! isShippingCalculatorEnabled } );
    props.onClick;
  }
  return <button onClick={ onClick }></button>;
}
```

This could even allow for composing various data we want included in the track in future iterations:

```jsx
const recordEvent = useTracks( 'some_event_slug', [ 'includePostData', 'includeSiteData' ] );
```

assets/js/base/utils/tracks.js
haszari: Good catch – if store has opted out, this is still a function, just one that does nothing. Ideally we have a way to inhibit any tracks-related setup code from running when store has opted out.

I'll need to revisit this - this if is not working as intended right now.

assets/js/blocks/cart-checkout/cart/edit.js
haszari: Good point – I hadn't checked this against guidelines. There are clear guidelines in Field Guide. In d414e28e9c81c3dd18ce225758705adf12f176bc I've updated the event name per guidelines.

- `wcadmin_` service context is provided by core tracks
- `blocks_` context associates this event with Woo Blocks repo/plugin
- `cart_` - this is a cart block event
- `settings_` in the block settings area of the UI
- `shipping_calculator_` the shipping calculator setting
- `toggle` the user action (present tense as per guidelines)

So the full event name in this case is `wcadmin_blocks_cart_settings_shipping_calculator_toggle`.

assets/js/base/utils/tracks.js
haszari: This is thinking ahead to an important event we want to track – when the merchant adds the cart or checkout to a page. One interesting prop that we'd like to track is whether this is the woo cart page (vs. a post, or a "test" page). 

These props are initial work towards this:

- We want to confirm that the block is added to a page (as opposed to blog post or other type).
- If we have the post ID, we can compare this to the ID for the cart page, and then we can add a prop for if merchant is editing the real cart page, which will help us understand how the blocks are being used – are merchants using cart block for real, or experimenting with it on a test page?

Having said that, the details of these standard props are something that we should iterate on as we add events. For example, `post_id` may not be that useful on its own, possibly post slug might be more useful.

assets/js/blocks/cart-checkout/cart/edit.js
haszari: I believe that it's best practise to trigger events directly from user actions, as close to the action as possible (I'm looking for guidelines / docs to back up this belief!). 

One risk with watching changes to attributes / state is that you risk firing events that are not tracking user intentions – e.g. default values getting overwritten with real state after a fetch.

haszari: There is one specific guideline related to this – events should not be fired generically, with the event name dynamically generated (with the exception of prefixing). This is so that you can easily search the code base for an event name, to see how it's fired (for what user action, where in the UI etc).

assets/js/base/utils/tracks.js
haszari: I've revised this (c7400b8af1a0a8e922032d47ad95a779cdeae9f9), using similar approach as core - `recordEvent` is a no-op, no mechanism to inhibit setup code related to tracks events (as far as I'm aware). 

If (in future) we find we need this we could look at adding info about tracks availability to wcSettings.

haszari:
Thanks for the reviews everyone (@mikejolley @nerrad) - your reviews have caught a few key details and helped clarify intentions here.

I'm going to try the hook / callback approach in separate branch now, to explore and better understand the tradeoffs. My understanding right now is that this would provide extra flexibility to allow client components to add props at event time; in my view, this is not needed. All the per-event props should be generated and passed in by the client code at time of event anyway (and of course the component can send "cached" (component instance local variables) or static state if it so chooses). 

Regarding concerns about how this will work for non-block-editor events – this is out of scope for now. The existing support in wc-admin can be used for firing events elsewhere in admin. 

This PR builds on the existing support and is intentionally scoped to a specific need – _sending events about usage of WooCommerce blocks in the editor_. The common props and the event prefix should be reviewed in this context, and we will continue to review and iterate these as we add more event coverage. 

As we implement more events we may find we need an abstraction to help generate some props that are shared by a subset of blocks events – I think we can address that need when it comes up, rather than add a specific solution now.

Also pinging data expert @deltaWhiskey for a review.

assets/js/base/utils/tracks.js
haszari: Here's an exploration of the hook approach. https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/try/woo-blocks-editor-tracks-events-hook

It looks to me like we'd be better off just providing useful `select` utils like `selectPostInfoProps()`. 

One specific thing I want to get in place now is a way to standardise common props for all woo blocks editor events - wrapping `wcTracks.recordEvent` seems the simplest way to do that.

nerrad: Thanks for doing the exploration of the hook approach.  

> If we go that way, it looks to me like we'd be better off just providing useful select utils like selectPostInfoProps()

That is basically what `includePostData` was in my examples above.

> the hook isn't providing anything particularly useful.

Can you expand on this please? I thought Mike and I gave some reasons why the hook has value (both for the work in this pull and also for needed future use as we expand out the implementation of data tracking) but I don't see where you've addressed that feedback.

PR #1977, opened 2020-03-16
disable coupon submit if the field is empty

senadir:
This PR disables the add coupon button if the field is empty, this is to prevent weird errors like `"" is already added` or `"" is not a valid coupon`.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1965

1. in a cart or checkout block, trying submitting an empty coupon
2. the button shouldn’t work
3. try writing something
4. button is now enabled.


nerrad:
Ya awesome catch!

PR #1978, opened 2020-03-16
disable proceed to checkout if rates are loading or being selected

senadir:
This PR disables `Proceed to checkout` if `isCalculating` from `useCheckoutContext` is set to true, this will be fully functional when we hook all of our hooks to that context (lineItem quantity change, removing items, selecting rates, fetching rates, and updating address, removing and adding coupons).

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1882

Currently, I couldn’t test this, I tried hooking `useShippingRates` to the context, but the value never seems to change cc @nerrad 

```js
useEffect( () => {
	if ( shippingRatesLoading ) {
		dispatchActions.incrementCalculating();
	} else {
		dispatchActions.decrementCalculating();
	}
}, [ shippingRatesLoading ] );
```

### DIFF
--- a/assets/js/base/context/cart-checkout/shipping/index.js
+++ b/assets/js/base/context/cart-checkout/shipping/index.js
@@ -119,7 +119,6 @@ export const ShippingDataProvider = ( { children } ) => {
 	};
 
 	// emit events.
-	// @todo add emitters for shipping rate selection.
 	useEffect( () => {
 		if ( ! shippingRatesLoading && currentErrorStatus.hasError ) {
 			emitEvent(
--- a/assets/js/base/hooks/shipping/use-shipping-rates.js
+++ b/assets/js/base/hooks/shipping/use-shipping-rates.js
@@ -2,9 +2,7 @@
  * External dependencies
  */
 import { useSelect } from '@wordpress/data';
-import { useEffect } from '@wordpress/element';
 import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
-import { useCheckoutContext } from '@woocommerce/base-context';
 
 /**
  * Internal dependencies
@@ -25,20 +23,12 @@ import { useShippingAddress } from '../shipping/use-shipping-address';
  *                 - {Object} shippingAddress       True when address data exists.
  */
 export const useShippingRates = () => {
-	const { dispatchActions } = useCheckoutContext();
 	const { cartErrors, shippingRates } = useStoreCart();
 	const { shippingAddress, setShippingAddress } = useShippingAddress();
 	const shippingRatesLoading = useSelect(
 		( select ) => select( storeKey ).areShippingRatesLoading(),
 		[]
 	);
-	useEffect( () => {
-		if ( shippingRatesLoading ) {
-			dispatchActions.incrementCalculating();
-		} else {
-			dispatchActions.decrementCalculating();
-		}
-	}, [ shippingRatesLoading ] );
 	return {
 		shippingRates,
 		shippingAddress,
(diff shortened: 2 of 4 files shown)


senadir:
for this to work, we will need to wrap `Cart` in a checkout provider, which I'm not sure makes much sense, also I pushed to see how much will it affect the bundle size @nerrad.

However, disregarding this issue, there is another issue, hooks seem to run multiple times (not sure what is causing this yet, my guess is the context is causing reruns), causing them to recreate again (and rerunning the useEffect) which causes the `calculationCount` to increase to 3 but only decrease back to 2 when it finishes, making this unreliable, if we’re not able to fix this, one solution I might suggest is moving to a keyed calculation logic, so instead of a counter, we set up keys

```js
dispatchActions.incrementCalculating('shipping-rates');
dispatchActions.decrementCalculating('shipping-rates');

// calculation shape
{
    "shipping-rates": true,
    "something-else": false
}

const isCalculating = Object.keys(checkoutState.calculation).some(Boolean);
```

nerrad:
I pushed a couple commits. The reason why you were seeing the calculating increment multiple times is because it was already incrementing in the `ShippingDataProvider`.  

> for this to work, we will need to wrap Cart in a checkout provider, which I'm not sure makes much sense, also I pushed to see how much will it affect the bundle size @nerrad.

Right, the idea I had planned on was either:

- rename Checkout context and it would receive props to help with re-using it in either Checkout or Cart context (for instance I would see the process component as something being passed into the provider as a prop so Cart would have a different processor than checkout - in this case null). That would also make the processor extendable/customizable by stores.  This is the preferred option I have. 
- OR create a new Cart context that just implements the providers it needs and otherwise has the same logic.  However, this route would likely result in duplicate logic between checkout and cart context providers (unless we could abstract some more of it out)... likely doable.

So before this pull can be merged, we have to decide what we're going to do with the contexts. What are your thoughts?

senadir:
> OR create a new Cart context that just implements the providers it needs and otherwise has the same logic. However, this route would likely result in duplicate logic between checkout and cart context providers (unless we could abstract some more of it out)... likely doable.

I'd rather go with this, I prefer we avoid having a monster-sized context, also the needs of Cart might differ with time from the needs of Checkout, duplicating code is a preferable approach to over abstracting to avoid over-engineering some parts to work for both.

however, this is just a guess, I don't have real data to say for sure that extending checkoutContext will result in this since I'm not yet fully aware of what we should copy (expect the calculation logic and possibly some event listeners).

nerrad:
I'm going to do a pull that moves the checkout logic from the checkout context provider into a custom hook and then we can probably easily use that in a Cart specific context.

nerrad:
Actually, we can't use a separate provider. The reason is that if we want to use the context hook (`useCheckoutContext`) anywhere in the components of the tree, then they have to be within the context provider. For components that might be shared between cart and checkout this could be problematic. I'm going to try instead abstracting some things outside of the context provider to see if I make it more re-usable.

senadir:
Currently, most of our hooks (except billing) are standard hooks, they don't deal with context, the context use them to watch for incrementing the status, what I meant initially is it possible to have things like increment calculation and decrement calculation in a Cart context

nerrad:
Right, but the plan is to have most components read data from the _context_ instead of the hook and have the context providers implement all the hooks.  I'm giving a try in a pull to see how things go and then we can go from there.

senadir:
So this is working fine right now for things like shipping thanks to Darren work, to extend this over to coupons and line quantity will require creating new context provider for those and moving functionality to them, I felt that this is a bit out of scope for this PR and that I prefer to open a new PR that will get better discussion there.

PR #1979, opened 2020-03-16
Add line pricing for cart items in the API

### LINKED ISSUE #1963: Totals don't add up when prices are displayed without taxes
**To Reproduce**
1. Go to _WooCommerce_ > _Settings_ > _Taxes_ and set _Display prices during cart and checkout_ to _Excluding Tax_.
2. Add some products to your cart and visit a page with the _Cart_ block.
3. Depending on the items you have in the cart, totals might not add up correctly:
imatge
_(line items are 170,93€, while subtotal is 170,91€)_

This is how it looks in the _Cart_ page, which doesn't have this issue:
imatge
_(notice the subtotal of _Cap_ is different than before)_

mikejolley:
#1963 is caused by rounding; the API returns rounded prices. Hard to explain but to summarise:

- Taxes are calculated per line, not per item
- If you round item prices first before multiplying by qty, you end up with a higher amount. 
- We need to use unrounded values, or return line prices via the API 

In the example @Aljullu gave in #1963 the cart page was doing this:

14.55 * 5 = 72.75

The cart page (core) is doing this:

14.54545500 * 5 = 72.727 (rounds to 72.73)

Hence the mismatch. This PR solves it by returning line_price properties with the qty factored in already.

Fixes #1963

1. Input prices inc tax. 
2. Setup product costing 16.
3. Setup tax rate of 10%
4. Add 5 of those items to cart.
5. Subtotal and line total should match.


senadir:
by doing that we lose live line price calculations (changing quantity won’t change the line price directly until the API response)

mikejolley:
@senedir then you'd need the raw unrounded price but it won't be an INT. Ideas?

mikejolley:
I wonder how critical this is. We lack a loading indicator for totals too semi related. 

We could introduce raw props in the api but given how we serve integers now I'm not sure. Perhaps a separate response with different precision?

@nerrad any thoughts for morning?

nerrad:
I think it's preferable to have accurate totals over immediate feedback at this point. So I'm currently leaning to merge with this fix and then look at potentially doing more immediate feedback on line item totals (or some sort of indicator something is loading) as a follow-up (in which case we'd probably need to return raw line item price values back from the api as well - to the extra decimal places right?).

senadir:
I think the less immediate response we get from the cart the slower it will be perceived, but that’s just a speculation, putting more loading indicators everywhere won’t help change that presumption, I do have a hard time seeing emergency in this PR since it the effect is really minimal (a couple of cents that you need to add up by yourself to find), but I don’t want to block this PR as well so feel free to merge and can pounder on this more in the future.

nerrad:
Just noting the following after doing some tests against this pull (I haven't reviewed any code):

I get this error on the frontend when loading the cart (does not happen in master):

```
Warning: Failed prop type: Invalid prop `shippingRates[0].items[2].quantity` of type `string` supplied to `Packages`, expected `number`.
    in Packages (created by ShippingRatesControl)
    in ShippingRatesControl (created by ShippingRateSelector)
    in fieldset (created by ShippingRateSelector)
    in ShippingRateSelector (created by TotalsShippingItem)
    in TotalsShippingItem (created by Cart)
    in div (created by Context.Consumer)
    in BodyUI (created by CardBody)
    in CardBody (created by Cart)
    in div (created by Context.Consumer)
    in CardUI (created by Card)
    in Card (created by Cart)
    in div (created by Sidebar)
    in Sidebar (created by Cart)
    in div (created by SidebarLayout)
    in SidebarLayout (created by Cart)
    in Cart (created by Block)
    in ValidationContextProvider (created by Block)
    in LoadingMask (created by Block)
    in Block (created by CartFrontend)
    in StoreNoticesProvider (created by CartFrontend)
    in CartFrontend
    in Unknown
    in Unknown
    in BlockErrorBoundary
```

Also I get an error in the editor for the cart block.

PR #1981, opened 2020-03-16
Configure link for the proceed to checkout button

mikejolley:
As a follow up to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1926, this implements a way to select the checkout page from the cart block. This option is only shown if the current page is _not_ the cart page set in settings (or if the option was previously set).

This also implements the notice to inform the user that the page is not set as the default, to match checkout.

Screenshot 2020-03-16 at 17 24 43

1. Edit the cart block page
2. Confirm the feedback notice and page notice is shown correctly
3. Try setting the button option to point to a different page and test frontend


mikejolley:
👍 I was working on it and didn't think I would finish, but I did :)

Issue #1474, opened 2020-01-03
Improve or replace Tabs Component.


nerrad:
I introduced a tabs component in #1349 to get the ball rolling but I'm not sure if we should roll with it because there's some accessibility issues and I also think we should look at using existing solutions out there. Possibilities are:

### `@wordpress/components` [tab-panel](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/tab-panel) component. 

This is what I loosely based the existing implementation off of.  Implementing this will require something like #1371 (along with tree-shaking) to be implemented to reduce weight on the frontend.

### [reakit](https://reakit.io/docs/tab/)

I'm leaning towards trying to implement this one because it was built accessibility first in mind and likely will result in less weight overall.  If it works well, we could consider contributing it back to the WordPress component package.

nerrad:
This is a candidate for followup iteration.

senadir:
I think we should go for `reakit`

senadir:
I'm returning this to the current sprint (feel free to move), I did some initial playing around in this branch, sadly it didn't work, something about how tabs content render.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/compare/add/reakit
I will revisit back in the future if anyone wants to take this go-ahead.

diegohaz:
@senadir Took a quick look at the code and noticed that you're not passing `selectedId` to the components, only the rest of the state:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1666671f274c0ff8b088f10fee1b1a4d20ed9acb/assets/js/base/components/tabs/index.js#L23

Not sure if that's the only problem, but that will definitely make it not work.

senadir:
Thanks @diegohaz that actually did help me solve the initial problem and discover a second issue in our code, must have took it off and never returned the `selectedId`

PR #1983, opened 2020-03-17
Implement Stripe CC and Stripe ApplePay payment methods

nerrad:
Fixes: #1471 

This pull implements the Stripe ApplePay and Stripe CC payment methods in the Checkout block and gets us one step closer to having the blocks fully functional.  There are still some todos after merging this pull but I'm submitting it for review now because:

- The payment methods have their logic pretty much wired up.
- UI/UX behaviour is functionally complete.
- Existing behaviour for the checkout block is not compromised, so we can merge and still test while developing.
- The work is at a point where others can jump in to fine tune and polish as we complete remaining tasks for the checkout block. Delaying merge until that's all complete makes it more difficult to accomplish those remaining tasks.
- This pull is already large. I'm wary of doing all the work needed in this pull making it that much harder to review (it's already larger than I'd like).

## Incomplete tasks:

This pull has some todos that are not complete. Once this pull is merged, I will go through this list (and inline todos) and determine what needs documented as issues.

- There is no server side handling for the selected payment method _when processing checkout_.
- While the payment methods are wired up to listen on the appropriate checkout emitted events, there has been no extensive testing that things are functioning as expected. I fully expect there may still be some buggy portions that need fixed/worked out once we get checkout processing and the server side part of payment integration completed.
- Neither of the payment methods account for zero total checkouts yet. 
- Some work is still needed to work on what components get exposed on payment methods via the api (the Save Payment Method checkbox for instance can likely be a component that is provided to via the payment method interface).
- Some work is needed on a preview view for the stripe cc active view. stripe.js is not loaded in the backend for performance so the elements as rendered on the frontend are not available. 
- We should probably fix the `SavedPaymentOptions` component so it doesn't pull actual saved payments (if they exist) for the logged in user when in the editor. Instead, it should use preview data and also _default_ to create a new payment option (for visibility).
- While payment methods do handle validation issues inline, they don't signal that to the checkout context yet. That needs wired up so the checkout button is disabled when there are validation errors.
- A review of inline todos and comment blocks needs reviewed to determine what is still valid and what needs removed. I'd prefer to do that in a follow-up.
- refactor some code out of the main files to reduce the code lines in a single file and help with maintenance/future review by other developers learning the implementation.
- refactor so that the stripe integration is built to it's own bundle and loaded in the same way the extension would, that can help us identify any additional loading points we may need to put in place for easier registration of payment methods.
- documentation (both inline and externally)
- Tests! I've deliberately held off writing tests, what you see in this pull is the end result of weeks of various experiments and refactors. Writing tests will be done as follow-ups.
- Mobile views are not completely tested and may need fine-tuning (in followups)
- Accessibility is not tested completely (however _some_ of that may be out of our control given we're using stripe elements).
- Are accepted credit cards pulled from the gateway?  (currently I'm just outputting an svg) - see `abstract-wc-stripe-payment-gateway::get_icon()` for how the payment method extension currently determines what icons to show.
- Do we need to launch using Stripe intents?
- Do we need to support other Stripe payment methods (Sepa, Bancontact, GiroPay, AliPay, Ideal etc?)
- Stripe.loadStripe functionality for the StripePromise resolution silently discards any exceptions. This needs reviewed and revisited. I think the reason why I did this is because the only time it was throwing for me was when things were loaded in the editor (because we don't load the stripe third party script in the editor - which causes the fail). So we might need a way to determine we're in the editor context here and use that to determine whether to throw the error or not.
- Implement a php api interface for payment methods to register data on the server to pass through to the payment registration client side. That way payment methods have a consistent interface for passing along (and implementing) that data.

## Complete tasks

- UI/UX for payment methods are complete and functional (up to the point of actual processing).
- `stripe.js` is wired up and hooked in to the current event emitters.
- Payment method api and interface has been fine-tuned to expose needed elements to the payment methods.
- When express payment methods are initialized, if there are no payment methods reporting that they can be used for payment on the frontend, don't show the express payments section.
- Express payment methods will show previews in the editor for the buttons.
- The payment method api now receives an `edit` node so payment methods can indicate what gets shown when in the editor view.
- Numerous type-defs have been created for Stripe (but there may need to be some fixes in followups).
- removed demo paypal logic. 
- Add actions server side to blocks that payment methods can hook into.
- Add a `Stripe` php class server side that is used for the Stripe payment method integration server side.

## Screenshots

Here's some screenshots/gifs to give an overview of completed things (and to verify I tested what I could 😄 )

**When there are no express payment methods that can pay (in this case Apple Pay is configured but I'm viewing checkout in the frontend in Chrome)**

Image 2020-03-26 at 4 07 17 PM

**Field behaviour for Stripe Credit Card**
Screen Recording 2020-03-26 at 04 07 PM

**Example of inline validation errors**

Image 2020-03-26 at 4 08 13 PM

**Apple Pay in Safari on desktop**

Image 2020-03-26 at 4 31 39 PM

**Apple Pay, Safari desktop, browser payment interface opened.**

Image 2020-03-26 at 5 05 02 PM

**Apple Pay, Safari - iPhone X**
Image 2020-03-26 at 4 35 20 PM

**ApplePay, Safari - iPhone X - browser payment interface opened.**

Image 2020-03-26 at 5 05 57 PM

## Testing

The following are the important things you are testing:

- When viewing checkout block in the editor or the frontend, and when interacting with existing ui, there are no console errors, and the block itself is functional.
- When payment methods are configured, they show up in the block as expected in the frontend and editor (please do ask if behaviour is expected or not if unsure).
- ApplePay button _should_ show up when the payment method is configured and you are testing on a publicly accessible website with ssl setup. Clicking the button _should_ open the browser payment request interface for ApplePay and have the correct totals. Behaviour after completing payment is _not_ wired up or supported yet, so we're not testing that.
- Stripe CC can be tested in local dev environment. Inputs should respond and update as expected for values entered. Submitting the checkout after entering values in the payment method inputs _are not expected to work yet_ as that has not been wired up to the server.

**To setup for testing**

- you will need Stripe Payment extension installed. You can find setup information etc in the fieldguide article: PCYsg-85x-p2 - In this article you are only concerned about Stripe CC and ApplePay instructions.
- For Apple Pay, you must test on a publicly accessible site that has an SSL cert (I used Atomic Ephemeral sites). Remember, Apple Pay _only works in Safari_ and on supported devices (you wouldn't believe how many times I forgot this)!
- For easier installation, I used the `package-plugin` npm script included in our build tools.


assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
```diff
@@ -34,8 +34,8 @@ import {
 import { __ } from '@wordpress/i18n';
 
 /**
- * @typedef {import('./type-defs').Stripe} Stripe
- * @typedef {import('./type-defs').StripePaymentRequest} StripePaymentRequest
+ * @typedef {import('../../stripe-utils/type-defs').Stripe} Stripe
```
mikejolley: Will these typedefs be shared? `StripePaymentRequest` in Apple Pay.

nerrad: I might not be addressing your question but if you're asking if the type-defs will be shared across payment methods, Apple Pay is a part of the stripe gateway plugin, so the type-defs will just be a part of that plugin (whether that team actually uses the type-defs or not is up to them)

assets/js/payment-methods-demo/payment-methods/stripe/payment-method.js
```diff
@@ -0,0 +1,317 @@
...
+		const response = createSource( stripeBilling );
+		if ( response.error ) {
+			// @todo set payment error
+			// return error object including any validation errors if present?
+		}
+		paymentStatus.setPaymentStatus().success( billingData, {
+			sourceId: response.source.id,
+		} );
+
+		// this should hook into checkout processing? Basically when the checkout
```
mikejolley: Am I reading this comment correctly and you want to do 2 actions following this; 1. Process payment, then 2. Update order/set paid?

If so I think /checkout would be a single action with response. Stripe would update the order on the server side after payment, or return an error.

assets/js/payment-methods-demo/payment-methods/stripe/payment-method.js
```diff
@@ -0,0 +1,317 @@
...
+		// if saved source (card) is selected by user, then there should be no
+		// processing/validation by stripe.
+		// as a part of the submission, go to `createSource`
+
+		// if submission completes okay, then set payment status to complete
+		// and that's where billing data and payment data are passed through.
+	};
+	const onComplete = () => {
+		// @todo this is invoked on checkout complete success
+		// in this place just set the currentPaymentStatus to completed.
```
mikejolley: and redirect somewhere?

assets/js/payment-methods-demo/stripe-utils/index.js
```diff
@@ -151,11 +153,106 @@ const canDoApplePay = ( paymentRequest ) => {
...
+const isNonFriendlyError = ( type ) =>
+	[
+		errorTypes.INVALID_REQUEST,
+		errorTypes.API_CONNECTION,
+		errorTypes.API_ERROR,
+		errorTypes.AUTHENTICATION_ERROR,
+		errorTypes.RATE_LIMIT_ERROR,
+	].includes( type );
+
+const getErrorMessageForCode = ( code ) => {
```
mikejolley: These codes and messages, as well as the CC form fields, could definitely be reusable across gateways and imported into Stripe.

assets/js/payment-methods-demo/stripe-utils/index.js
```diff
@@ -151,11 +153,106 @@ const canDoApplePay = ( paymentRequest ) => {
...
+		),
+	};
+	return messages[ code ] || '';
+};
+
+const getErrorMessageForTypeAndCode = ( type, code = '' ) => {
+	switch ( type ) {
+		case errorTypes.INVALID_EMAIL:
+			return __(
+				'Invalid email address, please correct and try again.',
```
mikejolley: Are gateways validating checkout fields rather than checkout itself?

assets/js/payment-methods-demo/payment-methods/stripe/payment-method.js
nerrad: > Am I reading this comment correctly and you want to do 2 actions following this; 1. Process payment, then 2. Update order/set paid?

No. What happens in this logic is the `createSource`  method is called to get the stripe _source_ (which is basically the token for completing the payment server side). If the source is successfully created then this `onSubmit` method (which is hooked into the checkout context `onCheckoutProcessing` event, will just return `true`.  

The `setPaymentStatus` is setting the source id (and I have to set the `payment_method` property yet too). 

Then after all the subscribers hooked into the checkout context `onCheckoutProcessing` event are done, checkout hit the server (the new endpoint you are doing) with the package for finishing the checkout processing server side.

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
mikejolley: ah sorry my bad :)

assets/js/payment-methods-demo/payment-methods/stripe/payment-method.js
nerrad: Checkout handles the redirect in this case. `onComplete` is hooked into the `onCheckoutCompleteSuccess` event checkout fires as a final set of things to fire before checkout redirects.  It's quite possible for this particular payment method I don't _need_ to do anything here and can remove this (since checkout won't care if the payment status is complete or not).

assets/js/payment-methods-demo/payment-methods/stripe/payment-method.js
```diff
@@ -0,0 +1,317 @@
...
+	// hook into and register callbacks for events.
+	useEffect( () => {
+		// validation and stripe processing (get source etc).
+		const unsubscribeProcessing = eventRegistration.onCheckoutProcessing(
+			onSubmit
+		);
+		const unsubscribeCheckoutComplete = eventRegistration.onCheckoutCompleteSuccess(
+			onComplete
+		);
+		const unsubscribeCheckoutCompleteError = eventRegistration.onCheckoutCompleteError(
```
nerrad: I'm considering changing the name of this event in the checkout context to `onCheckoutProcessingError`, and then `onCheckoutCompleteSuccess/Error` would just need to be `onCheckotuComplete`.

assets/js/payment-methods-demo/stripe-utils/index.js
nerrad: Generally, agreed, however some code strings are gateway specific (in this case these codes are). Once I get everything functional, I do plan on going through and reviewing what components/utilities etc should be exposed by checkout for gateways to use as opposed to something the gateway provides.

assets/js/payment-methods-demo/stripe-utils/index.js
nerrad: This was pulled directly from the stripe payment gateway and I _think_ there are some stripe payment method variations the stripe form has an option for an email field (which is why they have it). In our specific implementation this code would probably never get fired (b/c we're not using that variation) but I just copied all codes in.

nerrad: The other possibility is that this may be present because at one point Woo Core didn't have great email field validation and sometimes the gateway would reject the email submitted with the payment? IF this error gets returned when trying to create the stripe source, I can still have checkout handle invalidating the email field during process (with a message for why it was invalidated).

senadir:
> For Apple Pay, you must test on a publicly accessible site that has an SSL cert (I used Atomic Ephemeral sites). Remember, Apple Pay only works in Safari and on supported devices (you wouldn't believe how many times I forgot this)!

https://support.apple.com/en-us/HT207957
and countries, don’t forget countries 😢 well at least I will get to test when it doesn’t show up.

haszari:
Tip for reviewers: make sure you `npm install` and `composer install` before testing!

haszari:
For some reason the block never goes out of placeholder mode for me. I have stuff in my cart and Stripe configured (with a saved card for current user, from testing previous PR), no other payment methods enabled.

Screen Shot 2020-03-27 at 2 46 24 PM

I'll try disabling stripe, clearing cart and generally clearing various settings to see if I can figure out what's causing this.

nerrad:
@haszari The issues you were experiencing were due to the production build being broken (I'm assuming that's what you were testing with). The reason why I didn't see those issues is because I build my plugin zip based on the dev build (so I could see error stacks). It turns out that production was broken in master (not introduced by this branch) and thanks to help from @senadir , fixed in #2042.

nerrad:
@senadir regarding your testing (thanks for the notes!).  I'm noting what I experienced when trying to reproduce. I'm thinking these can be handled in followups (those more skilled with css can tackle them?)

> tabbing between inputs doesn’t focus them, so I can’t write until I click on the the input to write

It wasn't clear what browser this was happening for you, so I tested Chrome, Safari and Firefox.  The only browser I could reproduce this in was Safari. Not sure yet how we'll fix this for Safari.

> after writing and deleting, we end up with double placeholder + label, seems to only happen for the first two inputs.

The only browser I could reproduce this in was Firefox and even then, I can't reproduce reliably, sometimes it happens, other times it doesn't (no matter how I try to select and delete).

> I'm not sure if it also covers what happen when you click "Place order" it’s not working for me well
TypeError: undefined is not an object (evaluating 'stripe.createSource')

Logic handling when the button clicked is not expected to be functional in this pull, I'd planned on doing that as a followup.

haszari:
> @haszari The issues you were experiencing were due to the production build being broken (I'm assuming that's what you were testing with).

@nerrad I was testing with a debug build this time. I think all the issues I noted were minor style fixes for Twenty Twenty - we can follow up separately.

assets/js/base/components/payment-methods/style.scss
```diff
@@ -36,6 +36,7 @@
 			width: 50%;
 			> img {
 				width: 100%;
+				height: 48px;
```
mikejolley: Should this be max-height? Are we setting an expected/standard dimensions for gateways to use?

assets/js/base/components/payment-methods/style.scss
```diff
@@ -46,3 +47,107 @@
...
+		.wc-block-gateway-input.focused.empty,
+		.wc-block-gateway-input:not(.empty) {
+			& + label {
+				margin-left: $gap;
+				transform: translateY(#{$gap-smallest}) scale(0.75);
+			}
+		}
+		& + .wc-block-form-input-validation-error {
+			position: static;
+			margin-top: -$gap-large;
```
mikejolley: Is this working? I've had issues with this before (values became positive) and found a SO thread saying interpolation should be used. `-#{$gap-large}` I didn't test, just stood out.

assets/js/base/components/tabs/index.js
```diff
@@ -21,6 +21,7 @@ const Tabs = ( {
 	instanceId,
 	ariaLabel = __( 'Tabbed Content', 'woo-gutenberg-products-block' ),
 	children,
+	id,
```
mikejolley: Should we generate a default fallback for ID like on `selectedId`.

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
```diff
@@ -0,0 +1,380 @@
...
+	getStripeServerData,
+	getPaymentRequest,
+	updatePaymentRequest,
+	canDoApplePay,
+	getTotalPaymentItem,
+	stripePromise,
+} from '../../stripe-utils';
+
+/**
+ * External dependencies
```
mikejolley: I've noticed in recent files you've worked on you've started defining internal before external - is there a reason for this?

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
```diff
@@ -0,0 +1,380 @@
...
+	billing,
+	eventRegistration,
+	onSubmit,
+	activePaymentMethod,
+	setActivePaymentMethod,
+} ) => {
+	/**
+	 * @type {[ StripePaymentRequest|null, function( StripePaymentRequest ):StripePaymentRequest|null]}
+	 */
+	// @ts-ignore
```
mikejolley: Is this temporary?

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
```diff
@@ -0,0 +1,380 @@
...
+	const [ canMakePayment, setCanMakePayment ] = useState( false );
+	const eventHandlers = useRef( DEFAULT_STRIPE_EVENT_HANDLERS );
+	const currentBilling = useRef( billing );
+	const currentShipping = useRef( shippingData );
+	const currentPaymentRequest = useRef( paymentRequest );
+	const currentPaymentStatus = useRef( paymentStatus );
+	const currentEventRegistration = useRef( eventRegistration );
+	const isActive = useRef( activePaymentMethod === PAYMENT_METHOD_NAME );
+
+	// update refs when any change
```
mikejolley: nit. inline comments should end with full stop - php linters catch this, not sure about js

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
```diff
@@ -0,0 +1,380 @@
...
+		currentPaymentRequest.current = paymentRequest;
+		currentPaymentStatus.current = paymentStatus;
+		currentEventRegistration.current = eventRegistration;
+		isActive.current = activePaymentMethod === PAYMENT_METHOD_NAME;
+	}, [ billing, shippingData, paymentRequest, paymentStatus, isActive ] );
+
+	//set paymentRequest
+	useEffect( () => {
+		// can't do anything if stripe isn't available yet or we have zero total
+		if ( ! stripe || ! billing.cartTotal.value ) {
```
mikejolley: We should probably expose order and payment info on the CheckoutProvider rather than cart at some point.

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}, [ paymentRequest ] );
+
+	// kick off payment processing
+	const onButtonClick = () => {
+		setActivePaymentMethod( PAYMENT_METHOD_NAME );
+		currentPaymentStatus.current.setPaymentStatus().processing();
+	};
+
+	const abortPayment = ( paymentMethod, message ) => {
```
mikejolley: Not needed for this PR but we definitely should start adding docblocks to new methods.

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
```diff
@@ -0,0 +1,380 @@
...
+//@todo suggested endpoint for the checkout to do it's processing on submit:
+/**
+/store/checkout/payment/{order_id}
+body: {
+  paymentMethod: 'applePay',
+  // arbitrary things the payment needs to process - eg. token, payment.source.order_id
+  // that have no totals or anything and are simply used for processing the
+  // payment server side.
+}
+ */
```
mikejolley: Nothing blocking here yet but there is a lot of code and events to track for express methods - maybe we can reduce whats needed when we do some more express methods and see what gets duplicated? Before it's made public I mean.

assets/js/payment-methods-demo/express-payment/apple-pay/normalize.js
```diff
@@ -0,0 +1,169 @@
...
+};
+
+export {
+	normalizeLineItems,
+	normalizeShippingOptions,
+	normalizeShippingAddressForCheckout,
+	normalizeShippingOptionSelectionsForCheckout,
+	getBillingData,
+	getPaymentMethodData,
+};
```
mikejolley: This all makes sense 👍

assets/js/payment-methods-demo/express-payment/apple-pay/normalize.js
```diff
@@ -0,0 +1,169 @@
...
+const getBillingData = ( paymentResponse ) => {
+	const source = paymentResponse.source;
+	const name = source && source.owner.name;
+	const billing = source && source.owner.address;
+	const payerEmail = paymentResponse.payerEmail || '';
+	const payerPhone = paymentResponse.payerPhone || '';
+	return {
+		first_name: name
+			? name
+					.split( ' ' )
```
mikejolley: This might be common across gateways so extracting this type of thing to a util would probably be worthwhile

assets/js/payment-methods-demo/payment-methods/index.js
```diff
@@ -1,41 +1,25 @@
...
-	label: <img src={ ccSvg } alt="" />,
-	stepContent: null,
-	activeContent: <CreditCardActivePaymentMethod />,
-	edit: <CreditCardActivePaymentMethod />,
-	canMakePayment: Promise.resolve( true ),
-	ariaLabel: 'credit-card-payment-method',
+export const stripeCcPaymentMethod = {
+	id: PAYMENT_METHOD_NAME,
+	label: (
+		<strong>
```
mikejolley: Maybe this (strong) should be moved into where this method is rendered so gateways are not responsible.

assets/js/payment-methods-demo/payment-methods/index.js
```diff
@@ -1,41 +1,25 @@
...
+			{ __( 'Credit/Debit Card', 'woo-gutenberg-products-block' ) }
+		</strong>
+	),
+	activeContent: <StripeCreditCard />,
+	edit: <EditPlaceHolder />,
+	canMakePayment: stripePromise,
+	ariaLabel: 'Stripe Credit Card payment method',
```
mikejolley: Missing i18n

assets/js/payment-methods-demo/payment-methods/index.js
```diff
@@ -1,41 +1,25 @@
...
-	canMakePayment: Promise.resolve( true ),
-	ariaLabel: 'credit-card-payment-method',
+export const stripeCcPaymentMethod = {
+	id: PAYMENT_METHOD_NAME,
+	label: (
+		<strong>
+			{ __( 'Credit/Debit Card', 'woo-gutenberg-products-block' ) }
+		</strong>
+	),
+	activeContent: <StripeCreditCard />,
```
mikejolley: I think content would be fine for naming - I think you used active because it's a tab, but I don't think thats important when it comes to naming things in the common API.

assets/js/payment-methods-demo/payment-methods/index.js
```diff
@@ -1,41 +1,25 @@
...
+	edit: <EditPlaceHolder />,
+	canMakePayment: stripePromise,
```
mikejolley: The naming of this one is slightly odd. I assume this is basically whether or not the method will be enabled or visible? Remember some gateways (offline, invoice etc) don't technically take payment.

assets/js/payment-methods-demo/stripe-utils/load-stripe.js
```diff
@@ -0,0 +1,22 @@
...
+ * Internal dependencies
+ */
+import { getApiKey } from './utils';
+
+const stripePromise = new Promise( ( resolve ) => {
+	let stripe = null;
+	try {
+		stripe = loadStripe( getApiKey() );
+	} catch ( error ) {
+		// eslint-disable-next-line no-console
```
mikejolley: Will be be a customer facing error or a silent one?

assets/js/payment-methods-demo/stripe-utils/utils.js
```diff
@@ -0,0 +1,258 @@
...
+// rendered! (which also means it won't get passed from the server until it is rendered).
+// so we need to work out a loading process for when the cart or checkout hasn't been added to the
+// content yet - definitely don't want to load stripe without the block in the page. So that means
+// checkout will need to have some sort of editor preview for stripe.
+/**
+ * Stripe data comes form the server passed on a global object.
+ *
+ * @return  {StripeServerData}
+ */
+const getStripeServerData = () => {
```
mikejolley: So if I understand correctly, gateways might need to pass data to themselves via settings,  but we need to ensure block checkout is loaded first?

How about a common API on the PHP side for gateways to insert variables, and then have either the checkout or payment provider load them in somehow and pass through to payment method on initialisation.

assets/js/type-defs/contexts.js
```diff
@@ -212,17 +213,17 @@
  * @property {string}                  redirectUrl        This is the url that
  *                                                        checkout will redirect
  *                                                        to when it's ready.
- * @property {function()}              onCheckoutCompleteSuccess Used to register a
+ * @property {function(function())}    onCheckoutCompleteSuccess Used to register a
```
mikejolley: What does this mean?

assets/js/base/components/payment-methods/style.scss
nerrad: Working for me, maybe the webpack plugin handles it better than it used to?

assets/js/base/components/tabs/index.js
nerrad: This is mostly just an optional id for the entire tabs container. There's an issue (#1474) for reviewing/improving/replacing this tabs component I added, when we get to that at some point, we can use the opportunity to review whether this id is even needed (or improve).

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
nerrad: Ha! I always get mixed up and go with whatever I remember doing last. It didn't even occur to me I was doing it wrong. If the convention is external first (which makes sense) I should definitely be consistent with that. I created an issue (#2052)

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
nerrad: Yes. I have a personal todo (not an issue) to review all @ts-ignore usages and see about fixing. I just don't want to waste time on fixing things that is because I can't figure out how to properly document a particular type.

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
nerrad: Hmm js linters probably should too... I'm sure I've seen it caught before in the GB project.  I'll fix before merging.

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
nerrad: Agreed, but I also want to try to keep the payment interface usable for the cart context too (even though we won't implement that yet). Reason being is that _express payment methods_ should be usable for cart or "pay now" type buttons in the shop.

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
nerrad: Agreed.

assets/js/payment-methods-demo/express-payment/apple-pay/apple-pay-express.js
nerrad: Agreed

assets/js/base/components/payment-methods/style.scss
```diff
@@ -46,3 +47,107 @@
...
+		top: 0;
+		transform-origin: top left;
+		font-size: 16px;
+		line-height: 22px;
+		color: $gray-50;
+		transition: transform 200ms ease;
+		margin: 0 $gap;
+		overflow: hidden;
+		text-overflow: ellipsis;
+		max-width: calc(100% - #{  2 * $gap });
```
Aljullu: There seems to be a double space before the `2`, I think starting/ending spacing in curly braces should be removed?
```suggestion
		max-width: calc(100% - #{2 * $gap});
```
I don't know why stylelint didn't catch those.

assets/js/base/components/payment-methods/style.scss
nerrad: I _think_ we want to ensure that there's some consistency in styling here. Otherwise multiple express payment methods could risk being varying heights?

assets/js/base/components/payment-methods/style.scss
nerrad: > I don't know why stylelint didn't catch those.

Yup no idea.

assets/js/payment-methods-demo/express-payment/apple-pay/normalize.js
nerrad: The need for mapping will probably be common, but the values mapped will likely differ between gateways. Also, this is only used for the stripe apple pay gateway (because Apple Pay exposes the billing data from the customer's Apple Pay account).

assets/js/payment-methods-demo/payment-methods/index.js
nerrad: The `label` property for the config can receive any React node. This allows for gateways that might pass in an image instead of text (i.e. Paypal) for the label.  So while I agree somewhat with what you're saying here, I think we have to decide first about how restrictive we want to be for payment methods registering nodes for the tab label.

assets/js/payment-methods-demo/payment-methods/index.js
nerrad: It's for a consistent api, because some payment methods can only make a payment given certain conditions on the environment (browser in use, configured properly, third party library loaded etc).  This allows for a consistent way for gateways to indicate they can make payment.  For offline, or invoice payment methods they could just do `Promise.resolve( true )` as the value for this field.

assets/js/payment-methods-demo/payment-methods/index.js
nerrad: I didn't just use `activeContent` because it's in a tab. I used it because it's the component to render when the _payment method_ is selected (and thus made active).  However, I do realize that we could probably just get away with `content` here since there is no other "content" type (other than label which is already distinctive). I'll change.

assets/js/payment-methods-demo/payment-methods/index.js
nerrad: Good catch! Fixed and pushed.

assets/js/payment-methods-demo/stripe-utils/load-stripe.js
nerrad: If stripe fails to load here, then the payment method itself will not load or display anything.  The reason why I commented this out is because it does a console error in the admin.... I need to revisit though, I'm going to add a @todo for a followup.

assets/js/payment-methods-demo/stripe-utils/utils.js
nerrad: I like that idea. Adding as a follow-up.

assets/js/type-defs/contexts.js
nerrad: It means that the "type" for this value is a function that receives a function as it's argument.

PR #1984, opened 2020-03-17
Switch Add to Cart buttons to useStoreCart

mikejolley:
@nerrad I did this thinking it would allow us to track cart changes from the Empty Cart state (add to cart->updates cart store). But we're using the old blocks...

Discard or polish/keep? I think it still might be a useful refactor.


nerrad:
I think this would be good to keep if only for making any blocks interacting with the cart (WP 5.3+ blocks that is) using the same api (so updates/fixes to that api will automatically propagate).

I didn't do an in depth review, but just did a cursory review.

mikejolley:
Closing in favour of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2103

PR #1993, opened 2020-03-18
Add Checkout form validation

### LINKED ISSUE #1848: CountryInput and StateInput use CustomControlSelect, which doesn't support a required prop
Currently, there is no way to mark _Country_ and _State_ input fields as required, but we would like to do so in the _Checkout_ form. That's because internally they use Gutenberg's `<CustomControlSelect>`, which can't handle `required` cases.

### LINKED ISSUE #1994: Country Validation triggered when writing in any input in the Shipping form
**Describe the bug**
When writing in any field in the checkout shipping form, country validation is immediately triggered, even if that field is not relevant for rates (name, address...).

**To Reproduce**
0. Make sure your store default location is no location
1. In a private Window, go to checkout block
2. try on any field like name or last name
3. country validation will trigger.

**Expected behavior**
It should only trigger when you change a relevant field (state, city, postcode).

**Screenshots**
ezgif com-video-to-gif (3)

Aljullu:
Part of #1939.
Fixes #1848.
Fixes #1994.

This PR extends the validation context introduced in #1972 with some new methods:

* `hideValidationError` and `showValidationError` allow marking an error as visible or hidden to the user.
* `showAllValidationErrors` does the same but with all errors.
* `areThereValidationErrors` returns true whether there is at least one error.

It also changes the shape of the stored errors to `{ message: 'Test error', hidden: false }`. This allows adding errors in the context but displaying them conditionally. This way we can create validation errors when an input field is mounted, so if the user tries to submit the form without interacting with any field, those errors already exist and we can prevent the submit event and display the errors.

There are two new components that take care of creating validation errors on mount:
* `ValidatedTextInput` and `ValidatedSelect`, those components extend the `TextInput` and `Select` component we were currently using, but add validation on top of them. They also take care of clearing/hiding the errors when values are updated, the input is blurred, etc.

`ValidatedTextInput` relies on the native browser `<input>` validation rules. This has some advantages:
* We get email validation for free for the email input, required fields, etc. If we ever wanted to add new rules (min size for the address field, for example), we could do that with plain HTML.
* When a form fails to submit because there are validation errors, we can directly scroll to and focus the first invalid input.

That doesn't mean we can't build more complex validation rules. The validation check introduced in #1972 that verified a country was set when filling the address form in the shipping calculator, still works (and it works in the _Checkout_ block too).

### Screenshots

Peek 2020-03-18 17-11

1. Create a page with the _Checkout_ block and try filling the form leaving some fields empty, entering wrong emails in the email field, submitting the form without interacting with any field, etc.
2. Try entering the city, state or postal code before entering the Country and verify an error appears (only in the shipping form, it shouldn't happen in the billing form!).
3. Verify the coupon input field and the shipping calculator of the _Cart_ block still report errors properly like in #1972.

### DIFF
--- a/assets/js/base/components/cart-checkout/place-order-button/index.js
+++ b/assets/js/base/components/cart-checkout/place-order-button/index.js
@@ -2,23 +2,33 @@
  * External dependencies
  */
 import { useCheckoutContext } from '@woocommerce/base-context';
+import PropTypes from 'prop-types';
 
 /**
  * Internal dependencies
  */
 import Button from '../button';
 
-const PlaceOrderButton = () => {
+const PlaceOrderButton = ( { validateSubmit } ) => {
 	const { submitLabel, onSubmit } = useCheckoutContext();
 
 	return (
 		<Button
 			className="wc-block-components-checkout-place-order-button"
-			onClick={ onSubmit }
+			onClick={ () => {
+				const isValid = validateSubmit();
+				if ( isValid ) {
+					onSubmit();
+				}
+			} }
 		>
 			{ submitLabel }
 		</Button>
 	);
+};
+
+PlaceOrderButton.propTypes = {
+	validateSubmit: PropTypes.func.isRequired,
 };
 
 export default PlaceOrderButton;
--- a/assets/js/base/context/cart-checkout/validation/index.js
+++ b/assets/js/base/context/cart-checkout/validation/index.js
@@ -38,11 +38,11 @@ export const ValidationContextProvider = ( { children } ) => {
 	 *
 	 * @param {string} property The property the error message is for.
 	 *
-	 * @return {string} Either the error message for the given property or an
-	 *                  empty string.
+	 * @return {Object} The error object for the given property.
 	 */
-	const getValidationError = ( property ) =>
-		validationErrors[ property ] || '';
+	const getValidationError = ( property ) => {
+		return validationErrors[ property ];
+	};
 
 	/**
 	 * Clears any validation error that exists in state for the given property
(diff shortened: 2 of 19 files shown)


assets/js/base/components/cart-checkout/place-order-button/index.js
```diff
@@ -2,23 +2,33 @@
...
+const PlaceOrderButton = ( { validateSubmit } ) => {
 	const { submitLabel, onSubmit } = useCheckoutContext();
 
 	return (
 		<Button
 			className="wc-block-components-checkout-place-order-button"
-			onClick={ onSubmit }
+			onClick={ () => {
+				const isValid = validateSubmit();
+				if ( isValid ) {
```
nerrad: The idea I had was that checkout context would register a default callback with the `onCheckoutProcessing` event subscriber and then the callback would return the error object with the validation errors in it. Then checkout can take care of setting validation errors based on that response and abort the checkout setting checkout status appropriately etc. 

That way our components only have to:

- output a validation error if there is one in the validation state.
- clear the error once conditions are satisfied.

So with that in mind, I think the `PlaceOrderButton` should be disabled if the checkout is in error state and submission should only be blocked by that.  The current behaviour in `validateSubmit()` (scrolling to first invalid submit) could be triggered via the processing event callback.

If you're uncertain on how this could be done let me know and I can share more details :)

Aljullu:
As discussed in p1584615028124900-slack-C8X6Q7XQU, we will go for this solution for now. In a follow-up we will wire this up with the Checkout Context.

assets/js/base/context/cart-checkout/validation/index.js
```diff
@@ -98,6 +165,10 @@ export const ValidationContextProvider = ( { children } ) => {
 		clearValidationError,
 		clearAllValidationErrors,
 		getValidationErrorId,
+		hideValidationError,
+		showValidationError,
+		showAllValidationErrors,
+		areThereValidationErrors,
```
nerrad: Maybe rename this to `hasValidationErrors`?

Aljullu: Sure thing! Renamed in a21df92d6cd5ee85b2c7b66298205e479f1ad499.

PR #1995, opened 2020-03-18
Add checkout skeleton whilst loading

mikejolley:
Following on from the cart skeleton I added (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1866) this adds a loading Skeleton to the checkout and removes the 'coming soon' placeholder text we had in place.

### Screenshots

Screenshot 2020-03-18 at 16 36 39

1. Re-save or re-create the checkout block - the markup needs updating/invalidating
2. View the page with throttling
3. See the skeleton until the real checkout loads and replaces it.

### DIFF
--- a/assets/js/base/components/cart-checkout/button/style.scss
+++ b/assets/js/base/components/cart-checkout/button/style.scss
@@ -3,8 +3,8 @@
 .wc-block-checkout {
 	.button.wc-block-button {
 		align-items: center;
-		background-color: #000 !important;
-		color: #fff !important;
+		background-color: $black;
+		color: $white;
 		display: flex;
 		font-size: inherit;
 		font-weight: bold;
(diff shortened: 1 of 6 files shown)


mikejolley:
> One thing - I don't know how useful it is to have four grey boxes for steps – these expand out when the real steps are rendered, so perhaps could just have one circle + rect representing the form.

Those represent steps so having one would be weird. This doesn't display for long and does accurately match the shape of checkout. It's probably fine to ship as is. This won't invalidate blocks if changed.

On that subject (not related to this pull) @Aljullu @nerrad Would it be possible to inject attributes into the markup similar to how I'm injecting skeleton? This would solve the issue of invalidation due to adding attributes via the `save()` method using `data-` attributes. They'd be injected by the server.

nerrad:
> Would it be possible to inject attributes into the markup similar to how I'm injecting skeleton? This would solve the issue of invalidation due to adding attributes via the save() method using data- attributes. They'd be injected by the server. Just a thought since that happens a lot :)

Try it! I like that idea for reducing the chance for invalidation. I think this is what I might have had in mind originally for the All Products block too.

PR #1178, opened 2019-11-15
Add no result templates

mikejolley:
Part of #1138 cc @Aljullu @jwold 

Implements no products placeholders based on results.

Screenshot 2019-11-15 at 10 06 46

Screenshot 2019-11-15 at 10 07 35

Note: The reset button works but not with filter blocks correctly, but this will be resolved with #1168

To test, choose a price range which matches no products. For the other scenario, delete all products or hardcode the `hasProducts` variable.

PR #1996, opened 2020-03-18
Update empty states

### LINKED ISSUE #1976: Checkout should not work if you don’t have any products in your cart
**Describe the bug**
**I'm putting this in the next so that we can handle it in the next cycle.**

Currently, when you visit the checkout block when you don’t have any products in your cart, it still works but we no data and no interactions what so ever, so it’s just a useless confusing form, we should try to hide it until something is added to the Cart, ore simply redirect to something else.

**To Reproduce**

1. Empty your cart
2. Visit Checkout Block

**Expected behavior**
Nothing should show up.

mikejolley:
This PR introduces an empty state for the checkout block (when no items are found) and updates the product block empty states based on the feedback in #1178 (no issue was logged for 1178 but I had it on my list to revisit).

Ref #1178
Fixes #1976

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1995 should be reviewed first becasue I based this PR on the other branch.

### Screenshots

Screenshot 2020-03-18 at 21 29 03
Screenshot 2020-03-18 at 21 24 37
Screenshot 2020-03-18 at 21 21 37

1. Visit checkout page with an empty cart - see state.
2. Search for something on all products which you know returns no products e.g. price filter 89-90, and view icon.
3. Delete products. View all products. See the final icon.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/empty-cart.js
+++ b/assets/js/blocks/cart-checkout/checkout/empty-cart.js
@@ -10,7 +10,7 @@ const EmptyCart = () => {
 		<div className="wc-block-checkout-empty">
 			<Icon
 				className="wc-block-checkout-empty__image"
-				alt={ __( 'Empty cart', 'woo-gutenberg-products-block' ) }
+				alt=""
 				srcElement={ cart }
 				size={ 100 }
 			/>
--- a/assets/js/blocks/cart-checkout/checkout/empty-cart.js
+++ b/assets/js/blocks/cart-checkout/checkout/empty-cart.js
@@ -2,22 +2,23 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { WC_BLOCKS_ASSET_URL, SHOP_URL } from '@woocommerce/block-settings';
+import { SHOP_URL } from '@woocommerce/block-settings';
+import { Icon, cart } from '@woocommerce/icons';
 
 const EmptyCart = () => {
 	return (
 		<div className="wc-block-checkout-empty">
-			<img
-				src={ WC_BLOCKS_ASSET_URL + 'img/no-products.svg' }
-				alt={ __( 'No products', 'woo-gutenberg-products-block' ) }
+			<Icon
 				className="wc-block-checkout-empty__image"
+				srcElement={ cart }
+				size={ 100 }
 			/>
 			<strong className="wc-block-checkout-empty__title">
 				{ __( 'Your cart is empty!', 'woo-gutenberg-products-block' ) }
 			</strong>
 			<p className="wc-block-checkout-empty__description">
 				{ __(
-					'Checkout is not available whilst your cart is empty. Add some products to your cart and come back later!',
+					"Checkout is not available whilst your cart is empty—please take a look through our store and come back when you're ready to place an order.",
 					'woo-gutenberg-products-block'
 				) }
 			</p>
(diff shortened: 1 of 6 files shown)


assets/js/blocks/cart-checkout/checkout/empty-cart.js
```diff
@@ -0,0 +1,35 @@
...
+import { SHOP_URL } from '@woocommerce/block-settings';
+import { Icon, cart } from '@woocommerce/icons';
+
+const EmptyCart = () => {
+	return (
+		<div className="wc-block-checkout-empty">
+			<Icon
+				className="wc-block-checkout-empty__image"
+				alt={ __( 'Empty cart', 'woo-gutenberg-products-block' ) }
+				srcElement={ cart }
```
haszari: Aha ! I see there is an empty cart icon intended for the "empty checkout" state, this might work for empty cart too.

mikejolley:
I agree this could use some additional feedback from @garymurray however to answer some points:

> I noticed one issue which might be blocking – when I view /checkout and the cart is empty, I get redirected to cart with a notice. So possibly some of this markup isn't getting used?

This is core checkout behaviour when you have the pages setup correctly in woo core. If you're using blocks on other pages, there is no redirect. Hence why the empty state is useful.

> The empty cart message is a little fussy, can we simplify it? (Especially the one in code.) The whilst feels a bit nonstandard to me anyway. Possible alternative: There are no products in your cart. Add some products to continue to checkout.

Same wording in core probably because I wrote it and I love using the word 🧐
https://github.com/woocommerce/woocommerce-gutenberg-products-block/search?q=whilst&type=
https://github.com/woocommerce/woocommerce/search?q=whilst&type=

> When combined with the empty cart state, there are lots of redundant/similar things - e.g. Browse store / New in store, the "empty" in notice and in empty state warning. Put another way, the empty cart state header and the notice are essentially saying the same thing.

> The empty cart icon is sad! Perhaps we can use an empty cart! Or stacked shelves, or something :)

Note these are inner blocks, not a standard message we're outputting.

garymurray:
So just to check - we dealing with a few things here:

1. What does the cart look and what is the message if you try go to checkout prior to adding anything items to your cart?
2. What does the empty cart look like if the `New Products block` or any block a customer chooses to put in there returns no products?
3. Changing the empty cart icon to something less sad?

Is that a correct breakdown of what needs to be addressed?

mikejolley:
@garymurray The comments from @haszari basically increased scope - he is asking if all empty states should follow a similar design pattern. The outlier here is the "cart" empty state, which is using inner blocks and allows the merchant to control.

The other 3 empty states are fixed and rendered by us; for this I basically just want a 👍 on the icon used, and the wording for each. Those are the ones addressed in this PR.

I think if the cart inner blocks need adjustment it needs a separate issue logged.

garymurray:
@mikejolley thanks for clarifying.

> The other 3 empty states are fixed and rendered by us; for this I basically just want a 👍 on the icon used, and the wording for each. Those are the ones addressed in this PR.

Icons are good, and let's stick with the wording as is for now. That can always be reworked later, but should be consistent across what is in core and our blocks.

> I think if the cart inner blocks need adjustment it needs a separate issue logged.

Anything else can be logged as separate issues and dealt with in future iterations.

mikejolley:
Thanks @garymurray Ill leave existing states text alone in that case. Just waiting for code review approval and :shipit:

mikejolley:
Thanks for reviewing everyone.

PR #1998, opened 2020-03-19
Add no payment methods placeholder

### LINKED ISSUE #1478: Add placeholder for when no payment methods are installed and available.
This is a followup to #1349.  There is no design yet for this.

cc @garymurray, @jwold

mikejolley:
This implements a placeholder element in the checkout block when no payment methods are registered.

In addition, this PR fixes class names (to match standards) and updates the _existing_ shipping method placeholders to match style and structure.

Fixes #1478

(@nerrad you mentioned detecting if gateways supported blocks, but given the current registration system it didn't seem relevant to 1478 itself).

### Screenshots

This is a placeholder in editor context:

Screenshot 2020-03-19 at 14 10 08

And this is what a customer would see on frontend:

Screenshot 2020-03-19 at 14 22 21

1.  Delete all shipping methods and force `usePaymentMethodState` to return no methods.
2. Edit checkout block - see placeholders
3. View checkout block - see notices

### DIFF
--- a/assets/js/base/components/checkout/no-shipping/style.scss
+++ b/assets/js/base/components/checkout/no-shipping/style.scss
@@ -1,6 +1,5 @@
-.wc-block-checkout__no-shipping {
-	display: block;
-	margin-bottom: $gap !important;
+.components-placeholder.wc-block-checkout__no-shipping {
+	margin-bottom: $gap;
 
 	.components-placeholder__fieldset {
 		display: block;
--- a/assets/js/base/components/payment-methods/no-payment-methods/index.js
+++ b/assets/js/base/components/payment-methods/no-payment-methods/index.js
@@ -2,10 +2,11 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Placeholder, Button } from 'wordpress-components';
+import { Placeholder, Button, Notice } from 'wordpress-components';
 import { Icon, card } from '@woocommerce/icons';
 import { ADMIN_URL } from '@woocommerce/settings';
 import { useCheckoutContext } from '@woocommerce/base-context';
+import classnames from 'classnames';
 
 /**
  * Internal dependencies
@@ -49,7 +50,21 @@ const NoPaymentMethodsPlaceholder = () => {
 };
 
 const NoPaymentMethodsNotice = () => {
-	return <div>Hello</div>;
+	return (
+		<Notice
+			isDismissible={ false }
+			className={ classnames(
+				'wc-block-checkout__no-payment-methods-notice',
+				'woocommerce-message',
+				'woocommerce-error'
+			) }
+		>
+			{ __(
+				'There are no payment methods available. This may be an error on our side. Please contact us if you need any help placing your order.',
+				'woo-gutenberg-products-block'
+			) }
+		</Notice>
+	);
 };
 
 export default NoPaymentMethods;
--- a/assets/js/base/components/payment-methods/no-payment-methods/index.js
+++ b/assets/js/base/components/payment-methods/no-payment-methods/index.js
@@ -38,7 +38,7 @@ const NoPaymentMethodsPlaceholder = () => {
 		>
 			<span className="wc-block-checkout__no-payment-methods-description">
 				{ __(
-					'Your store does not have any payment methods configured. Once you have configured your payment methods they will appear here.',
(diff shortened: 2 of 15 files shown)


assets/js/base/components/checkout/no-shipping/style.scss
```diff
@@ -1,9 +1,19 @@
-.wc-block-checkout__no-shipping .components-placeholder__fieldset {
+.wc-block-checkout__no-shipping {
 	display: block;
```
Aljullu: Is `display: block` needed? I think it's always overridden in the last versions of Gutenberg. In old versions, it doesn't play well with the placeholder (screenshot from WP 5.3 with Gutenberg disabled):
imatge

assets/js/base/components/checkout/no-shipping/style.scss
```diff
@@ -1,9 +1,19 @@
...
-}
+	margin-bottom: $gap !important;
```
Aljullu: We could remove the `!important` here increasing the specificity of the selector as follows:
```CSS
.components-placeholder.wc-block-checkout__no-shipping {
```
When possible, I think it's better to increase specificity with extra classes in the selectors than using `!important`, this way it's easier to add styling on top of it if themes want to.

assets/js/base/components/payment-methods/no-payment-methods/index.js
```diff
@@ -0,0 +1,79 @@
...
+		>
+			<span className="wc-block-checkout__no-payment-methods-description">
+				{ __(
+					'Your store does not have any payment methods configured. Once you have configured your payment methods they will appear here.',
+					'woo-gutenberg-products-block'
+				) }
+			</span>
+			<Button
+				isDefault
+				href={ `${ ADMIN_URL }admin.php?page=wc-settings&tab=checkout` }
```
Aljullu: Thinking out loud, but I wonder whether we should add `target="_blank"` here. I'm usually against forcing a link to open in a new tab, but I think this a good use-case for it. So I'm 50%-50% about adding it or leaving it as-is.

assets/js/base/components/payment-methods/no-payment-methods/style.scss
```diff
@@ -0,0 +1,24 @@
+
```
Aljullu: No need for an empty line at the beginning of the file:
```suggestion
```

assets/js/base/components/payment-methods/no-payment-methods/style.scss
```diff
@@ -0,0 +1,24 @@
...
+.wc-block-checkout__no-payment-methods {
+	display: block;
+	margin-bottom: $gap !important;
```
Aljullu: All comments I did above about removing `display: block` and exchanging `!important` for a more specific selector apply here too.

assets/js/base/components/shipping-rates-control/style.scss
```diff
@@ -27,9 +27,8 @@
 	color: $core-grey-dark-400;
 }
 
-.wc-block-shipping-rates-control__no-results {
-	margin-bottom: 0;
-	font-style: italic;
+.wc-block-shipping-rates-control__no-results-notice {
+	margin-bottom: $gap !important;
```
Aljullu: Here again, I think we could remove `!important` adding an extra class to the selector.

mikejolley:
Thanks @Aljullu should be taken care of.

mikejolley:
Bundlesize above 155 something to worry about?

nerrad:
bump it to 160kb

nerrad:
> @nerrad you mentioned detecting if gateways supported blocks, but given the current registration system it didn't seem relevant to 1478 itself

I was mostly thinking that the current message could be confusing to merchants that _do_ have a payment method configured, but just not one that the blocks support.  So my idea was that in our messaging we make that clear.

assets/js/base/components/shipping-rates-control/style.scss
Aljullu: @mikejolley this was marked as solved but I think those lines didn't change, right? Was that an overlook?

mikejolley:
Bundle size increased, just needs approval.

mikejolley:
@nerrad I think we should hide that element if there are no express methods - it's static right now.

nerrad:
> I think we should hide that element if there are no express methods - it's static right now.

Let's roll with that for now and we can revisit in the future if needed.

PR #1999, opened 2020-03-19
Storefront compatibility for sticky proceed to checkout button on mobile 

### LINKED ISSUE #1930: Storefront sticky footer overlaps sticky cart submit button on mobile
Follow up issue to #1917

The cart block has a sticky checkout button. Storefront has a sticky footer on all pages, which obscures the checkout button. 

Screen Shot 2020-03-11 at 9 45 33 AM

We'll need to either override storefront styling (to hide the footer), or get this addressed in Storefront directly. 

This may also affect checkout (assuming it has fixed footer for submit button on mobile).

mikejolley:
Adds compatibility between the sticky checkout button and Strorefront by removing the sticky Storefront bar when rendering the cart and checkout blocks. This uses `remove_action`.

In a similar fashion, this also renders a 'push' div in the footer so that our sticky element does not overlap the footer content on the website.

Fixes #1930

### Screenshots

Screenshot 2020-03-19 at 15 20 28

1. View cart/checkout in mobile view whilst running Storefront
2. Check it doesn't overlap the footer, and that the Storefront bar does not render.

### DIFF
--- a/src/BlockTypes/Cart.php
+++ b/src/BlockTypes/Cart.php
@@ -52,6 +52,20 @@ class Cart extends AbstractBlock {
 		do_action( 'woocommerce_blocks_enqueue_cart_block_scripts_before' );
 		$this->enqueue_assets( $attributes );
 		do_action( 'woocommerce_blocks_enqueue_cart_block_scripts_after' );
+
+		// Disable storefront footer bar when using this block.
+		if ( has_action( 'storefront_footer', 'storefront_handheld_footer_bar' ) ) {
+			remove_action( 'storefront_footer', 'storefront_handheld_footer_bar', 999 );
+		}
+
+		// Add placeholder element to footer to push content for the sticky bar on mobile.
+		add_action(
+			'wp_footer',
+			function() {
+				echo '<div class="wc-block-cart__submit-container-push"></div>';
+			}
+		);
+
 		return $content . $this->get_skeleton();
 	}
 
--- a/src/BlockTypes/Cart.php
+++ b/src/BlockTypes/Cart.php
@@ -53,11 +53,6 @@ class Cart extends AbstractBlock {
 		$this->enqueue_assets( $attributes );
 		do_action( 'woocommerce_blocks_enqueue_cart_block_scripts_after' );
 
-		// Disable storefront footer bar when using this block.
-		if ( has_action( 'storefront_footer', 'storefront_handheld_footer_bar' ) ) {
-			remove_action( 'storefront_footer', 'storefront_handheld_footer_bar', 999 );
-		}
-
 		// Add placeholder element to footer to push content for the sticky bar on mobile.
 		add_action(
 			'wp_footer',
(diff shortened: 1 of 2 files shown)


src/BlockTypes/Cart.php
```diff
@@ -78,6 +78,20 @@ public function render( $attributes = array(), $content = '' ) {
 			$this->block_name . '-frontend',
 			$this->block_name . '-block-frontend'
 		);
+
+		// Disable storefront footer bar when using this block.
+		if ( has_action( 'storefront_footer', 'storefront_handheld_footer_bar' ) ) {
+			remove_action( 'storefront_footer', 'storefront_handheld_footer_bar', 999 );
+		}
```
nerrad: I'm wary that we're establishing a precedent here for handling theme specific customizations in the block plugin. What if we encounter other themes in the wild that add a similar mobile footer?

Some possible options:

a.  Leave this for now, deal with another approach if/when we find other themes that need dealt with.
b.  Create a theme compat file/class so we keep all theme specific compatibility things separate so as themes account for them we can remove the compat.  This might warrant adding action hooks to the blocks for hooking this compat file in.
c.  Put the compat code in the storefront theme (establishing the precedent that _themes_ have to account for the blocks).

I personally lean to b or c here. Thoughts?

mikejolley: So the reason I put it here instead of directly in Storefront was, this action needs to run only when the block is used. Storefront has no way of knowing this, unless we add an action or something here for it to listen out for.

Alternatively we can just go with z-index for now. It's hacky but it will cover the Storefront footer.

nerrad: > Storefront has no way of knowing this, unless we add an action or something here for it to listen out for.

I think this may be a good signal of the need to _add_ an action here. In #1961 I added a set of actions to both the cart block and the checkout block that would work well for this.  Although we may want to look at automatically adding dynamic actions  on the via the render method in the abstract class so that every child block type already has them in place.

mikejolley: Yes those 2 actions would be enough. Ill hold off for those to merge - we can raise PR in storefront.

haszari: In this case I think a temporary hook on `storefront_footer` is ok, since it's our default theme, and these blocks are still in development.

Long term adding hooks for blocks and implementing these kinds of fixes in themes is the best approach, happy if we go that approach now. Related, we may need to fire certain existing cart/checkout hooks to ensure things like `woocommerceanalytics` events fire for block-based cart/checkout - e.g. [`woocommerce_after_checkout_form`](https://github.com/Automattic/jetpack/blob/master/modules/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php#L46). I'm guessing there are other hooks which we'll need to review/fire when cart & checkout blocks are fully functional.

mikejolley:
@nerrad This PR only contains the style fixes now so can be merged? Storefront fix is raised here: https://github.com/woocommerce/storefront/pull/1307

PR #2000, opened 2020-03-20
Load product grid styles in WP <=5.2

Aljullu:
We moved product grid styles to the `ProductList` component, but that component is not loaded in WP <= 5.2 because  the _All Product_ block is restricted to WP >= 5.3, so product grid styles were not loading in WP <= 5.2.

This PR explicitly declares the `ProductList` styles as an entry file in Webpack config, so it's always loaded no matter if it's the legacy build or the default build.

Notice this issue is affecting the last release of blocks (2.5.x), so I think it might be worth doing a point release for it.

### Screenshots

_Before:_
imatge

_After:_
imatge

1. In a WP 5.0, 5.1 or 5.2 environment add a _Products by Tag_ block, for example.
2. Preview it in the frontend and verify it's correctly styled.

### Changelog

> Fix broken product grid blocks styles in old versions of WordPress.

### DIFF
--- a/bin/webpack-helpers.js
+++ b/bin/webpack-helpers.js
@@ -115,6 +115,10 @@ const stableMainEntry = {
 	'snackbar-notice-style':
 		'./node_modules/@wordpress/components/src/snackbar/style.scss',
 
+	// Styles for grid blocks. WP <=5.2 doesn't have the All Products block,
+	// so this file would not be included if not explicitly declared here.
+	'product-list-style': './assets/js/base/components/product-list/style.scss',
+
 	// Blocks
 	'handpicked-products': './assets/js/blocks/handpicked-products/index.js',
 	'product-best-sellers': './assets/js/blocks/product-best-sellers/index.js',
--- a/bin/webpack-helpers.js
+++ b/bin/webpack-helpers.js
@@ -117,6 +117,8 @@ const stableMainEntry = {
 
 	// Styles for grid blocks. WP <=5.2 doesn't have the All Products block,
 	// so this file would not be included if not explicitly declared here.
+	// This file is excluded from the default build so CSS styles are included
+	// in the other the components are imported.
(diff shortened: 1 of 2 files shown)


Aljullu:
I'm a bit confused by this. I checked it locally and `style.css` from master and `style.css` from this branch had exactly the same size and even the same number of characters (109561). The contents inside them were in a different order, so I thought that might be the cause of the size difference in the gzipped files reported by bundlewatch.

So in 2a651684152b39b1f4eb48d4b001e3100895381b I explicitly excluded `product-list-style` from the default build, this way the `style.css` generated from master and the one generated from this branch are exactly the same: same file size, same number of characters and no diff between the two files. Still, bundlewatch seems to report a size difference.

I see it also [reports](https://travis-ci.com/github/woocommerce/woocommerce-gutenberg-products-block/jobs/300367241) file differences for other files which are completely unrelated to this pull:
```
PASS ./build/all-products-frontend.js: 16.28KB < 60KB (578B larger than master) (gzip)
PASS ./build/all-products.js: 64.07KB < 200KB (152B larger than master) (gzip)
PASS ./build/all-reviews-legacy.js: 10.07KB < 200KB (1B larger than master) (gzip)
PASS ./build/all-reviews.js: 10.34KB < 200KB (22B larger than master) (gzip)
PASS ./build/attribute-filter-frontend.js: 16.13KB < 60KB (11B larger than master) (gzip)
PASS ./build/attribute-filter.js: 11.21KB < 200KB (4B smaller than master) (gzip)
```

So I'm even more confused now. Any clues @nerrad?

nerrad:
Ya those additional size differences are puzzling. My main concern was potential duplication of css styles in the builds themselves though, so if you've verified there's no duplication, I'm fine with ignoring these apparent size discrepancies. 

Sidenote: On the weekend I'm planning on experimenting to switch away from bundlesize to the same action Gutenberg is using for monitoring file sizes. I'm hoping that will help improve the reporting.

PR #2001, opened 2020-03-20
Fix blocks not being added in WP <= 5.2

### LINKED ISSUE #1914: All blocks are broken in master and WP 5.0
**To Reproduce**
1. In a WP 5.0 install, try adding any block from WooCommerce Blocks to a post.
2. Notice it crashes with this error in the console: `TypeError: o is not a function`.

The issue seems to happen here:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/filters/block-list-block.js#L23

**Screenshots**
Peek 2020-03-09 21-23

I didn't test other versions of WordPress, which might be affected as well.

WooCommerce Blocks 2.5.14 is not affected and it works properly.

Aljullu:
Fixes #1914.

Notice you might need to build the project with `npm run build` instead of `npm start`. There are some side effects issues with old dependencies, so a production build is needed to have all blocks available.

1. In a WP 5.0, 5.1 or 5.2 environment try to add a block to a post.
2. Verify the block is added without error.
3. Try adding the _All Reviews_ block and verify it doesn't crash.

### DIFF
--- a/assets/js/base/components/sort-select/index.js
+++ b/assets/js/base/components/sort-select/index.js
@@ -4,7 +4,7 @@
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Label from '@woocommerce/base-components/label';
-import { withInstanceId } from '@wordpress/compose';
+import { withInstanceId } from '@woocommerce/base-hocs/with-instance-id';
 
 /**
  * Internal dependencies
--- a/assets/js/base/components/sort-select/index.js
+++ b/assets/js/base/components/sort-select/index.js
@@ -4,7 +4,7 @@
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Label from '@woocommerce/base-components/label';
-import { withInstanceId } from 'wordpress-compose';
+import { withInstanceId } from '@wordpress/compose';
 
 /**
  * Internal dependencies
--- a/assets/js/components/view-switcher/index.js
+++ b/assets/js/components/view-switcher/index.js
@@ -6,7 +6,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { ButtonGroup, Button } from '@wordpress/components';
 import { useState, Fragment } from '@wordpress/element';
-import { withInstanceId } from 'wordpress-compose';
+import { withInstanceId } from '@wordpress/compose';
 
 /**
  * Internal dependencies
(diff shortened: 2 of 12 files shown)


assets/js/base/components/sort-select/index.js
```diff
@@ -4,7 +4,7 @@
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import Label from '@woocommerce/base-components/label';
-import { withInstanceId } from 'wordpress-compose';
+import { withInstanceId } from '@wordpress/compose';
```
Aljullu: This component is used in _Reviews_ block, which are available in legacy environments, so we can't use the lastest version of `@wordpress/compose` because it relies on hooks which are not available in WP <= 5.2.

The change included in this PR makes it so we load the external instead of the latest version of the library.

mikejolley:
@nerrad I think that change was to make it so the attributes (defaults) were always stored to source. Might be separate to that suggestion depending on where the server side list of attributes come from.

mikejolley:
@nerrad @Aljullu If we're reverting to `@wordpress/compose` in the base components we may have to put the instance ID HOC back if this adds extra bundles to frontend right? Or find a way to not use instance IDs if that is possible...

nerrad:
>  we may have to put the instance ID HOC back if this adds extra bundles to frontend right?

Assuming you mean our former `withComponentId` component? If so, yes.  The other option here is to use the [legacy build system](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/master/assets/js/legacy) for the affected component so that on legacy builds it uses the external and on recent (5.3+) builds it uses the direct import. Would be a good opportunity to validate the legacy build system works (and fix any bugs if necessary).

Aljullu:
In the last commits I created `@woocommerce/base-hocs/with-instance-id`, which imports the HOC from the external in WP <= 5.2 and imports it directly for WP >= 5.3. I also moved all `withInstanceId` imports around the codebase to this new file.

So this PR should be ready for another review.

assets/js/components/view-switcher/index.js
```diff
@@ -6,7 +6,7 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { ButtonGroup, Button } from '@wordpress/components';
 import { useState, Fragment } from '@wordpress/element';
-import { withInstanceId } from 'wordpress-compose';
+import { withInstanceId } from '@wordpress/compose';
```
Aljullu: In the editor, I think we should always use the external.

PR #2003, opened 2020-03-20
Convert emitters to promise

nerrad:
While working on #1983 I realized that the event emitters api needs to be implemented using Promises as registered callbacks might need to be promises themselves.  For example, Stripe CC interacts with promises for creating the stripe source.

This also allows the observers to be more asynchronous (even though the implementation here still waits for each observer to resolve).

As a part of this pull:

- I did a bit of restructuring of the `event-emit` directory (moving emit functions outside of the reducer file).
- Added a new `emitEventWithAbort` function that will abort looping through the registered observers on the event if any of them return a non-truthy value. This will help with the handling of validation/errors in the checkout context (or any other future types of events that need similar behaviour).
- Converted `emitEvent` to be a promise.
- updated `CheckoutContextProvider` to correctly implement the emit event changes (`ShippingDataContextProvider` implements `emitEvent`, but no changes were necessary).
- added js unit tests for the `emitEvent` and `emitEventWithAbort` functions.

## To Test

One reason I added the unit tests in this pull is because nothing is currently implementing the event system - so tests verify these functions work as expected. Until we implement observers on the event system there's no user facing tests.

So the testing needed is just doing smoke tests to ensure the Checkout and Shipping contexts don't affect any existing behaviour of the checkout block.


assets/js/base/context/cart-checkout/event_emit/emitters.js
```diff
@@ -0,0 +1,63 @@
...
+ *                           invoked.
+ *
+ * @return {Promise} A promise that resolves to true after all observers have
+ *                   executed.
+ */
+export const emitEvent = async ( observers, eventType, data ) => {
+	const observersByType = observers[ eventType ] || [];
+	for ( let i = 0; i < observersByType.length; i++ ) {
+		try {
+			await Promise.resolve( observersByType[ i ]( data ) );
```
nerrad: `Promise.resolve` is used here to ensure that every observer is handled like a promise.

assets/js/base/context/cart-checkout/event_emit/emitters.js
```diff
@@ -0,0 +1,63 @@
...
+ */
+export const emitEventWithAbort = async ( observers, eventType, data ) => {
+	const observersByType = observers[ eventType ] || [];
+	let response;
+	for ( let i = 0; i < observersByType.length; i++ ) {
+		try {
+			response = await Promise.resolve( observersByType[ i ]( data ) );
+			if ( response !== true ) {
+				// abort we have an error
+				i = observersByType.length;
```
Aljullu: Would returning the `response` here make the code simpler? I needed to think twice to understand what this line was doing. :sweat_smile: 
```suggestion
				return response;
```
If we do this, the `response` could be a const declared inside the `try` block.

nerrad:
bundlesize budget is being bumped in one of Mike's pulls, so I'll ignore that for merging this pull.

assets/js/base/context/cart-checkout/event_emit/emitters.js
nerrad: oh ya! definitely.

Aljullu:
Oh, and something that surprised me even though it's not added in this PR. Why the folder is named `event_emitters` (notice the underscore) instead of `event-emitters`? Was that in purpose?

nerrad:
> Why the folder is named event_emitters (notice the underscore) instead of event-emitters? Was that in purpose?

🤦‍♂ not on purpose. I'll fix.

PR #2004, opened 2020-03-20
[Experiment] Add container to run php unit tests in docker

mikejolley:
This PR makes some (possible) improvements to phpunit tests.

If using Docker, you can run `npm run phpunit` which will setup a container with WordPress + WooCommerce and run phpunit. @senadir You said you could not run tests locally. Now you can.

For Travis, I've made a small improvement to copy WooCommerce from .org which avoids the need to run `npm install` and `composer install` everytime tests run. It's way faster. cc @nerrad 

Docker was a massive PITA but I think I've worked it out :) Some of this new code, such as the entry point, is copied from the WordPress docker package. I needed to make a custom Dockerfile to ensure WooCommerce was downloaded and installed correctly. There are probably ways to optimise this further but it works.


mikejolley:
Happy with this now. Simplified the logic somewhat.

PR #2005, opened 2020-03-21
Add bundle size github workflow

nerrad:
This workflow implements a bundle-size monitoring utility for our repository to give better metrics on increases in bundle-sizes. I basically copied the configuration for the WordPress gutenberg project workflow and it utilizes the [`preactjs/compressed-size-action`](https://github.com/preactjs/compressed-size-action).

This will replace our existing bundlewatch implementation.

### Benefits of the switch:

- We get inline feedback (as a comment) in the pull on any size changes. 
- This compares overall changes against the base branch for the pull (which we weren't getting with bundlewatch).
- A list of all files and their changes is shown in a collapsed list in the posted comment. With bundlewatch, their linking system was broken, so if we wanted that granularity we had to go into the travis-ci job (which only showed us the sizes, not the difference from the base branch).
- Removes the time for the bundlewatch checks from the linting travis job (not a huge difference but helps).

The only real downside to this switch is that we don't block merging pull with this. So there's no _enforced_ size budgets. I'm not sure that's much of a downside though because our current pattern has been mostly to use this for awareness and we've generally bumped bundets to merge pulls (while having an issue for following up on size reviews).

So the big benefit here is getting _actionable_ feedback on significant change (or improvements!!) on bundle sizes in context of the pull making those changes.

### DIFF
--- a/.github/workflows/bundle-size.yml
+++ b/.github/workflows/bundle-size.yml
@@ -1,0 +1,17 @@
+name: Bundle Size
+
+on: [pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2-beta
+      with:
+        fetch-depth: 1
+    - uses: preactjs/compressed-size-action@v1
+      with:
+        repo-token: "${{ secrets.GITHUB_TOKEN }}"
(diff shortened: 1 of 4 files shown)


nerrad:
This is a build tool only, so I'm going to go ahead and merge this.

PR #2006, opened 2020-03-22
Build storybook to github pages on pushes to master

nerrad:
This pull adds configuration for deploying a build of storybook to Github pages on any push/merge to master. 

- Builds are output to `storybook/dist` (but only during the deploy)
- Travis is used for deploying.

Using this pull to document and verify travis config isn't borked. Otherwise, it needs to be merged to master to verify the build works as expected.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,3 +85,20 @@ jobs:
               - npm run test:e2e
           env:
               - WOOCOMMERCE_BLOCKS_PHASE=experimental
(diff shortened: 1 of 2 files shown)


nerrad:
For those looking at this pull as a resource, there were a few commits I made directly on master fixing things for the travis deploy configuration. So make sure you take a look at the latest travis config in master for that part of things.

PR #2009, opened 2020-03-23
Decode address entities in shipping calculator

Aljullu:
### Screenshots

_Before:_
imatge

_After:_
imatge

1. Go to the _Cart_ block and add a city with a special character in the shipping calculator.
2. Update the address and reload the page.
3. Verify the address appears correctly.

### DIFF
--- a/assets/js/base/hooks/shipping/use-shipping-address.js
+++ b/assets/js/base/hooks/shipping/use-shipping-address.js
@@ -6,6 +6,7 @@ import { useEffect, useState } from '@wordpress/element';
 import isShallowEqual from '@wordpress/is-shallow-equal';
 import { useDebounce } from 'use-debounce';
 import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
+import { decodeEntities } from '@wordpress/html-entities';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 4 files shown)


assets/js/base/hooks/shipping/use-shipping-rates.js
```diff
@@ -29,9 +30,15 @@ export const useShippingRates = () => {
 		( select ) => select( storeKey ).areShippingRatesLoading(),
 		[]
 	);
+	const decodedShippingAddress = {};
+	Object.keys( shippingAddress ).forEach( ( key ) => {
+		decodedShippingAddress[ key ] = decodeEntities(
+			shippingAddress[ key ]
+		);
+	} );
```
nerrad: This looks like something that should be implemented in `useShippingAddress` instead of `useShippingRates`?

Aljullu: Absolutely! Done in d237e4d6c6c2d80fb41cf6eb2dedf54467f865dc.

assets/js/base/hooks/shipping/use-shipping-address.js
```diff
@@ -39,8 +40,16 @@ export const useShippingAddress = () => {
...
 		}
 	}, [ debouncedShippingAddress ] );
+
+	const decodedShippingAddress = {};
+	Object.keys( shippingAddress ).forEach( ( key ) => {
+		decodedShippingAddress[ key ] = decodeEntities(
+			shippingAddress[ key ]
+		);
+	} );
+
```
senadir: we should start investing in a `fromEntities`  pollyfill later

Aljullu: Agree, I missed being able to use it here. :smile:

PR #2011, opened 2020-03-23
Update checkout API endpoint to process orders

### LINKED ISSUE #1938: Add endpoint for creating and processing order.

mikejolley:
This might be more work when we integrate it, but this should be good enough to merge for now to unblock other tasks.

Creates the `wc/store/checkout` endpoint. This endpoint takes an existing order and processes payment using a gateway.

To make this easier to test, add this to wp-config:

```
define( 'WOOCOMMERCE_STORE_API_DISABLE_NONCE_CHECKS', true );
```

This disables nonce verification so you can use basic auth from a REST client to test the endpoints.

After doing this:

1. Enable the bacs gateway
2. Add some things to your cart logged in
3. Using basic auth, POST to `/wc/store/cart/create-order`. Note the order ID and key in the response.
4. Use the order ID and key to POST to `wc/store/checkout`. The request will be something like this:

```json
{
  "order_id": 100,
  "order_key": "xxxxxx",
  "payment_method": "bacs"
}
```

Check the response and confirm it was successful.

Closes #1938

### DIFF
--- a/src/Library.php
+++ b/src/Library.php
@@ -9,6 +9,9 @@ namespace Automattic\WooCommerce\Blocks;
 
 defined( 'ABSPATH' ) || exit;
 
+use Automattic\WooCommerce\Blocks\Payments\PaymentResult;
+use Automattic\WooCommerce\Blocks\Payments\PaymentContext;
+
 /**
  * Library class.
  */
@@ -26,6 +29,7 @@ class Library {
 		add_filter( 'woocommerce_register_shop_order_post_statuses', array( __CLASS__, 'register_draft_order_post_status' ) );
 		add_filter( 'woocommerce_valid_order_statuses_for_payment', array( __CLASS__, 'append_draft_order_post_status' ) );
 		add_action( 'woocommerce_cleanup_draft_orders', array( __CLASS__, 'delete_expired_draft_orders' ) );
+		add_action( 'woocommerce_rest_checkout_process_payment_with_context', array( __CLASS__, 'process_legacy_payment' ), 10, 2 );
 	}
 
 	/**
@@ -193,6 +197,49 @@ class Library {
 			WHERE posts.post_status = 'wc-checkout-draft'
 			AND posts.post_modified <= ( NOW() - INTERVAL 1 DAY )
 			"
+		);
+	}
+
+	/**
+	 * Attempt to process a payment for the checkout API if no payment methods support the
+	 * woocommerce_rest_checkout_process_payment_with_context action.
+	 *
+	 * @param PaymentContext $context Holds context for the payment.
+	 * @param PaymentResult  $result  Result of the payment.
+	 */
+	public static function process_legacy_payment( PaymentContext $context, PaymentResult &$result ) {
+		if ( $result->status ) {
+			return;
+		}
+
+		// phpcs:ignore WordPress.Security.NonceVerification
+		$post_data = $_POST;
+
+		// Set constants.
+		wc_maybe_define_constant( 'WOOCOMMERCE_CHECKOUT', true );
+
+		// Add the payment data from the API to the POST global.
+		$_POST = $context->payment_data;
+
+		// Call the process payment method of the chosen gatway.
+		$available_gateways = WC()->payment_gateways->get_available_payment_gateways();
+
+		if ( ! isset( $available_gateways[ $context->payment_method ] ) ) {
+			return;
+		}
+
+		// Process Payment.
+		$gateway_result = $available_gateways[ $context->payment_method ]->process_payment( $context->order->get_id() );
+
+		// Restore $_POST data.
+		$_POST = $post_data;
+
+		// Handle result.
+		$result->set_status( isset( $gateway_result['result'] ) && 'success' === $gateway_result['result'] ? 'success' : 'failure' );
+		$result->set_payment_details(
+			[
+				'redirect' => $gateway_result['redirect'],
+			]
 		);
 	}
 }
--- a/src/Library.php
+++ b/src/Library.php
@@ -24,6 +24,7 @@ class Library {
 		add_action( 'init', array( __CLASS__, 'maybe_create_cronjobs' ) );
 		add_filter( 'wc_order_statuses', array( __CLASS__, 'register_draft_order_status' ) );
 		add_filter( 'woocommerce_register_shop_order_post_statuses', array( __CLASS__, 'register_draft_order_post_status' ) );
+		add_filter( 'woocommerce_valid_order_statuses_for_payment', array( __CLASS__, 'append_draft_order_post_status' ) );
 		add_action( 'woocommerce_cleanup_draft_orders', array( __CLASS__, 'delete_expired_draft_orders' ) );
 	}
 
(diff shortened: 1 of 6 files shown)


src/Library.php
```diff
@@ -183,4 +199,54 @@ public static function delete_expired_draft_orders() {
...
+	 * @param PaymentContext $context Holds context for the payment.
+	 * @param PaymentResult  $result  Result of the payment.
+	 */
+	public static function process_legacy_payment( PaymentContext $context, PaymentResult &$result ) {
+		if ( $result->status ) {
+			return;
+		}
+
+		// phpcs:ignore WordPress.Security.NonceVerification
+		$post_data = $_POST;
```
nerrad: Currently payment methods in Woo Core probably rely on the nonce verification that happens in the `WC()->checkout->process_checkout()` method right?  Do we not care about that because this is covered by nonce validation in our REST endpoint?

src/Library.php
```diff
@@ -183,4 +199,54 @@ public static function delete_expired_draft_orders() {
...
+		if ( ! isset( $available_gateways[ $context->payment_method ] ) ) {
+			return;
+		}
+
+		$payment_method_object = $available_gateways[ $context->payment_method ];
+		$payment_method_object->validate_fields();
+
+		if ( 0 !== wc_notice_count( 'error' ) ) {
+			return;
+		}
```
nerrad: This is something we probably will have to handle somehow right? If there is a notice, then there was an error, so we'll need some way of communicating that error back to the client?

src/Payments/PaymentContext.php
```diff
@@ -0,0 +1,79 @@
...
+	/**
+	 * Magic getter for protected properties.
+	 *
+	 * @param string $name Property name.
+	 */
+	public function __get( $name ) {
+		if ( in_array( $name, [ 'payment_method', 'order', 'payment_data' ], true ) ) {
+			return $this->$name;
+		}
+		return null;
```
nerrad: What about throwing an error if something tries to access a property that isn't exposed?

src/Payments/PaymentResult.php
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * PaymentResult class.
+ */
+class PaymentResult {
+	/**
+	 * List of valid payment statuses.
+	 *
+	 * @var array
+	 */
+	protected $valid_statuses = [ 'success', 'failure', 'pending', 'error' ];
```
nerrad: Are these defined somewhere in WooCommerce core as a list of payment statuses that payment methods can use? Can we reference that in the comment doc so if that changes we know it needs updated (or better yet, this is something that maybe could be added to WooCore at some point).

src/RestApi/StoreApi/Routes/Checkout.php
```diff
@@ -43,28 +45,289 @@ public function get_args() {
...
+				$response->set_status( 202 );
+				break;
+			case 'failure':
+				$response->set_status( 400 );
+				break;
+			case 'error':
+				$response->set_status( 500 );
+				break;
+		}
+
```
nerrad: It appears this is the replacement for the `WC_Checkout::process_checkout`  behaviour right? Should we make note somewhere (maybe in a followup issue?) of all the things that are _not_ running through this logic that _are_ running with the logic on `WC_Checkout::process_checkout`? For instance, from a brief look this is a list I came up with:

- `wc_set_time_limit`,
- `woocommerce_before_checkout_process` action hook
- handling if cart is empty (session expired?)
- `woocommerce_checkout_process` action hook
- `WC_Checkout::validate_checkout` it looks like we do the same logic in this endpoint, but just checking.
- `woocommerce_checkout_order_processed` action

src/RestApi/StoreApi/Schemas/CheckoutSchema.php
```diff
@@ -40,16 +63,23 @@ public function get_properties() {
...
+		if ( ! $order || ! $payment_result ) {
+			throw new Exception( 'Invalid response.' );
+		}
+
 		return [
-			'note'  => 'This is a placeholder',
-			'order' => $order_schema->get_item_response( $checkout_result['order'] ),
+			'payment_status'  => $payment_result->status,
+			'payment_details' => $payment_result->payment_details,
+			'order'           => ( new OrderSchema() )->get_item_response( $order ),
```
nerrad: If payment method returns a redirect url, should we require a standard key for that and return that on the root of the response object so that checkout processing can use that to know where to redirect to?

senadir:
I did a functionality test of this and it worked fine for me!

src/Payments/PaymentResult.php
senadir: `wc_get_order_statuses` provides that (with the `wc-` prefix), and plugins can extend into it new status.

src/Library.php
mikejolley: Correct; we have our own nonce protection and gateways are not responsible for this either.

src/Library.php
mikejolley: Yes. I'm not sure how yet, I just added it here "just in case". I'd consider this BW compat code experimental.

src/RestApi/StoreApi/Schemas/CheckoutSchema.php
mikejolley: Possibly but I'm not sure if we want to enforce it for all gateways as as standard key. We may for instance what to simply default to the thanks page, or replace that at some point. We should revisit this after implementing some gateways.

src/RestApi/StoreApi/Routes/Checkout.php
mikejolley: Yes we need a follow up to discuss the hooks in particular. I am concerned that throwing in hooks may further complicate things under this context.

src/Payments/PaymentResult.php
mikejolley: These are not core _order_ statuses. These are for payment status from the gateway. This is specific to the PaymentResult class and can be adjusted.

src/Payments/PaymentContext.php
mikejolley: PHP doesn't throw exceptions if you access non-defined variables so I don't think we need to here. This is just for convenience -- I was going to just use stdClass but wanted a little extra flexibility just in case.

src/Payments/PaymentResult.php
nerrad: To be clear, I'm asking if Woo Core has _payment_ statuses defined somewhere?  If so, we should probably pull them from where they are defined (if possible). If not, then we should make it really clear we're defining canonical statuses here (especially since these are functionally enums for the logic in this class).

mikejolley: No they don't exist. The ajax response for payments in core is just success or fail. Nothing else.

mikejolley:
Merging this as I need to make some changes in the processing PR.

PR #2012, opened 2020-03-24
Reinitialise useSelectShippingRate when rates change.

### LINKED ISSUE #1970: Removing a selected rate from Cart will leave Cart with no selected rate.
So this is a bit tricky to set up and explain, Imagine you have free shipping initiated by a coupon or min order amount, if you select it, then remove the condition that initiated it, it will leave you with unselected rate, and it won’t invalidate the previous one

image

image

I think we should do some kind of check in the REST API (seeing if the selected rate exists within presented rates)

senadir:
This PR fixes a small issue when selecting shippingRates, currently, we depend on `useState` to maintain the selected shipping rates, and we initialize `useState` with selected rates from the server, the problem is that, if rates changes, useState is not recreated again (and the initial value is not changed), weirdly enough, this issue is not present if you change the country.

so overall, this PR adds a useEffect that watches a memoized initial rates and updates the state if that value changes.

This could also be fixed if `useState` took a dependency array like other hooks (but this won't happen https://github.com/facebook/react/issues/14738), some solutions exists like [`useStateWithDeps`](https://www.npmjs.com/package/use-state-with-deps) but I didn’t want to introduce a new thing.

Fixes #1970

1. setup a deletable shipping rate (free shipping with a coupon, or free shipping above a cart amount) 
2. meet the criteria for that rate and select it.
3. remove coupon and reduce amount so the rates gets removed.
4. see that another rate (depends on the server) will be selected and applied.

### DIFF
--- a/assets/js/base/hooks/shipping/use-select-shipping-rate.js
+++ b/assets/js/base/hooks/shipping/use-select-shipping-rate.js
@@ -2,10 +2,9 @@
  * External dependencies
  */
 import { useDispatch } from '@wordpress/data';
-import { useState, useEffect } from '@wordpress/element';
+import { useState, useEffect, useMemo } from '@wordpress/element';
 import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
-import { useThrowError, usePrevious } from '@woocommerce/base-hooks';
-import isShallowEqual from '@wordpress/is-shallow-equal';
+import { useThrowError } from '@woocommerce/base-hooks';
 
 /**
  * This is a custom hook for loading the selected shipping rate from the cart store and actions for selecting a rate.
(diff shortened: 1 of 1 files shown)


senadir:
> And if there are multiple alternative rates, then you see no radio button selected for the shipping method (the bug, #1970), but the cart still functions correctly and a method is selected when you proceed to cart.

yeah this was not an issue with the backend since the API returns the now selected rate and calculates everything, so prices are all set, it’s just the internal state in the hook is using stale values.

this has to do with how react works, so when you change a country, the whole component (shipping rates) re-render, causing the useState to initialise again, the same when there is 1 shipping rates (it uses a whole different render up in the tree (does not pass by `Packages`) so the issue is not present there.
but in our case, only a single rate disappears, the other rates preserve their keys, so react doesn’t rerender `Packages` it only removes that single package, which means useState in our hook keeps the same stale value.

PR #2023, opened 2020-03-25
Make phone number option respected in billing field

### LINKED ISSUE #2020: Add url to ideas board for feedback prompt
See @garymurray 's [comment here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1356#issuecomment-603729384):

> We can now link this to the Ideas board - I have created a category for WooCommerce Blocks - this link should pre-filter it: https://ideas.woocommerce.com/forums/133476-woocommerce?category_id=384565

> We can then also use that link for any future/past blocks that we want to capture feedback on.

### LINKED ISSUE #2022: Phone form input option not respected by billing
**Describe the bug**
According to Checkout user story
> If the user disables [shipping as billing] – we add a new section to the form for a custom billing address – which inherits the same fields (with settings) as the Shipping address above it.

Phone is one of those options but it’s only respected within Shipping fields, not billing fields.

**To Reproduce**
Enable phone option in checkout.
Shipping form will now have it.
uncheck use shipping as billing.
billing form does not have it.

**Expected behavior**
the same option should be reflected.

senadir:
Phone number options only existed in shipping form, this PR move the input to `defaultAddressFields` so both shipping and billing can use it.

Fixes #2022

1. toggle phone form option.
2. untoggle use shipping as billing
3. see the input in billing as well.

### DIFF
--- a/assets/js/base/components/cart-checkout/address-form/default-address-fields.js
+++ b/assets/js/base/components/cart-checkout/address-form/default-address-fields.js
@@ -17,6 +17,7 @@ import { __ } from '@wordpress/i18n';
  * @property {AddressField} city       City name.
  * @property {AddressField} state      State name or code.
  * @property {AddressField} postcode   Postal code.
+ * @property {AddressField} phone      Phone number.
  */
 const AddressFields = {
 	first_name: {
@@ -114,6 +115,14 @@ const AddressFields = {
 		required: true,
 		hidden: false,
 		index: 9,
+	},
+	phone: {
+		label: __( 'Phone', 'woo-gutenberg-products-block' ),
+		optionalLabel: __( 'Phone (optional)', 'woo-gutenberg-products-block' ),
+		autocomplete: 'phone',
+		required: false,
+		hidden: false,
+		index: 10,
 	},
 };
 
--- a/assets/js/base/components/cart-checkout/address-form/default-address-fields.js
+++ b/assets/js/base/components/cart-checkout/address-form/default-address-fields.js
@@ -119,10 +119,11 @@ const AddressFields = {
 	phone: {
 		label: __( 'Phone', 'woo-gutenberg-products-block' ),
 		optionalLabel: __( 'Phone (optional)', 'woo-gutenberg-products-block' ),
-		autocomplete: 'phone',
+		autocomplete: 'tel',
 		required: false,
 		hidden: false,
 		index: 10,
+		type: 'tel',
 	},
 };
 
--- a/assets/js/base/components/cart-checkout/address-form/index.js
+++ b/assets/js/base/components/cart-checkout/address-form/index.js
@@ -164,6 +164,7 @@ const AddressForm = ( {
 
 				return (
 					<ValidatedTextInput
+						type={ field.type || 'text' }
 						key={ field.key }
 						className={ `wc-block-address-form__${ field.key }` }
 						label={
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -108,6 +108,11 @@ const Checkout = ( {
 			...defaultAddressFields.address_2,
 			hidden: ! attributes.showAddress2Field,
 		},
+		phone: {
+			...defaultAddressFields.phone,
+			hidden: ! attributes.showPhoneField,
+			required: attributes.requirePhoneField,
+		},
 	};
 
 	const setShippingFields = useCallback(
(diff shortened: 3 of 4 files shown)


mikejolley:
I would not merge this yet. @Aljullu @senadir are you both sure this fixes the phone field? This PR adds it to the default fields array but it's only present in billing. There is no "phone" field in core for shipping at all. I noticed this earlier and moved to the contact section fwiw https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2027

Aljullu:
Yes, that was something that confused me as well. Designs show the phone as a shipping field, but maybe that was an overlook?

senadir:
so if I understand correctly, it should only be present in the shipping form, not the billing form? we can still keep this update and remove it from billing (rather than having it inline in the checkout)

Aljullu:
The checkout shortcode shows the Phone field only in the Billing section.

mikejolley:
Whilst it's billing data, I think it belongs in contact like email, which is also billing data. See #2027

senadir:
> The checkout shortcode shows the Phone field only in the Billing section.
Isn’t because billing is the default form.
Phone number has more value in shipping than billing, but I had clients use phone number as part of their sign up or delivery, where do you think it fits best?
we currently don’t send or track some values like 'keep me updated' thing, email is part of billing and shipping is still not set.

garymurray:
If enabled via the block setting - the phone number field will show in the shipping address - as per the design here:

Screenshot 2020-03-26 at 09 09 30

In this case above, if the customer wanted to enter a different billing address - the phone number field will show in the billing address area too.

So basically, both the shipping address and billing address section can show a phone number field based on the setting the merchant chooses for the billing address (disabled, enabled and optional, enabled and required).

mikejolley:
@garymurray if the user has a different billing and shipping number, what we are saying is core only stores billing phone. There is no shipping phone number unless we find a way to add it to core.

garymurray:
> @garymurray if the user has a different billing and shipping number, what we are saying is core only stores billing phone. There is no shipping phone number unless we find a way to add it to core.

@mikejolley let's remove the option to input a different phone number for the billing address then seeing it is not supported by core.

mikejolley:
@garymurray If shipping address is hidden we need to place it somewhere too.

Is contact details (where email is) not appropriate? Or put it there when shipping address is disabled?

garymurray:
> @garymurray If shipping address is hidden we need to place it somewhere too.

My assumption is that the hiding of a shipping address would be for stores selling digital products. 

So to check - why would they want/need to capture a phone number?

mikejolley:
> So to check - why would they want/need to capture a phone number?

Presumably you'd still have billing data. So back at you; why did we want a billing phone number as well as shipping phone number?

garymurray:
> Presumably you'd still have billing data. So back at you; why did we want a billing phone number as well as shipping phone number?

If a store is not using/needing the shipping address, then the telephone number should be part of the billing address details imo. 

With that said, we are then getting into the details of what happens when someone is using these blocks on a store selling digital only products, which was outside of the scope of the MVP.

senadir:
giving the discussion, I've deleted the phone field from the billing address and only left it for the shipping one.

senadir:
waiting for tests to pass to merge so I can continue https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2057

mikejolley:
@senadir I don't understand why you bothered adding this to the _default_ fields only to remove it later. Surely those fields need to be only fields common to both.

Would it not make more sense to just show with shipping address as normal, unless shipping is disabled in which case render it in billing instead. The field is identical in both places - stored to billing address only. Like billing email is.

senadir:
I added it to `defaultFields` so that it renders with the rest of shipping fields as normal (currently in master, it renders outside the form div, causing styling issues), removing it from billing insures it doesn’t show up there (currently in master, not rendering shipping form will also not render the phone since it’s inside the condition, it also doesn’t render billing form but that’s outside the scope).

What I failed to understand is where should this field live, and where should it persist, so that we can move it there, simply keeping an extra chunk of code in the block in not ideal and I would rather have it with the rest of fields.

mikejolley:
> I added it to defaultFields so that it renders with the rest of shipping fields as normal 

This will therefore mix billing and shipping data right?

Whilst it's showing with shipping fields, it's actually billing data.

senadir:
> Whilst it's showing with shipping fields, it's actually billing data.

I initially assumed it’s a shipping field, in that case, you want it to live within the shipping form visually but controlled by billing data?

mikejolley:
@senadir Thats how it is, which is why it's probably best to render it separately like we do billing email.

senadir:
okay so I guess I will close this PR, and let the current behavior as is.

PR #2026, opened 2020-03-25
fix rates not updating if you change coupon

### LINKED ISSUE #2025: Coupons do not affect rates in checkout
**Describe the bug**
slightly similar to https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1970 this seems like another stale prop issue, it seems when you setup a free shipping rate using a coupon, inserting that coupon in checkout will not update the current rates, you would have to select a rate, change shipping fields or something to initiate a change.

**To Reproduce**

1. setup a free shipping rate + free shipping coupon 
2. in checkout, use that coupon to get free shipping 
3. scrolling back to shipping rates, that free rate is not there
4. you have to select another rate, or change the shipping for it to show up.
5. same if you remove the coupon, the rate will stay.

**Expected behavior**
rates should update as well.

senadir:
This was a bit tricky, but it seems `usePrevious` was not updating well, and was working correctly to begin with (the condition was to use previous if current rates are 0, this could be misleading if the current address has no rates associated with it, we never reached that case but changing address always caused the whole tree to render).

Deleting that code fixed the issue.

Fixes #2025

1. setup a free shipping rate + free shipping coupon
2. in checkout, use that coupon to get free shipping
3. scrolling back to shipping rates, that free rate is now visible
4. deleting the coupon will also delete the rate.

You can also test the Cart block to see if nothing changed there.

### DIFF
--- a/assets/js/base/components/shipping-rates-control/index.js
+++ b/assets/js/base/components/shipping-rates-control/index.js
@@ -3,7 +3,6 @@
  */
 import { __ } from '@wordpress/i18n';
 import PropTypes from 'prop-types';
-import { usePrevious } from '@woocommerce/base-hooks';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


senadir:
> Not blocking this PR, but I noticed that if I remove the coupon, I get no shipping rates selected by default

this is reported in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1970 and fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2012 (which you reviewed 😅)

Aljullu:
Oh right, never mind. I assumed that one was already merged and it was a regression. :sweat:

PR #2029, opened 2020-03-25
Move payment method registration state to payment data context and add saved payment methods handling

nerrad:
This work was initially done as a part of #1983 and I've pulled out into this separate pull so it can be reviewed on it's own merit.

This pull does the following:

- Adds a `SavedPaymentMethodOptions` component and related logic for outputting a ui for saved payment methods within the payment methods step if there are any saved payment methods for the logged in user.
- Refactors `usePaymentMethods` so the state and logic handling initialization of registered payment methods is moved to the payment method data context. The current behaviour was poor because it resulted in unnecessary effect updates and the state needs to be persistent through the tree, not local to the hook as implemented in a component.

Image 2020-03-25 at 5 57 21 PM

Note: the above screenshot is specific to the master branch. 

## To Test

- First verify that things work without saved payment methods for the logged in user (or non-logged in user). ("Things work", meaning the checkout block loads without errors and has expected functionality)
- Next, you'll have to use the existing wc core checkout to save a payment method - and complete payment (probably easiest to setup stripe gateway for this test).
- Once your logged in user has a saved payment method, test the checkout block to ensure the new options appear as expected and you can toggle between them.

### DIFF
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -1,0 +1,137 @@
+/**
+ * External dependencies
+ */
+import { useEffect, useState, useRef, useCallback } from '@wordpress/element';
+import { __, sprintf } from '@wordpress/i18n';
+import { usePaymentMethodDataContext } from '@woocommerce/base-context';
+import RadioControl from '@woocommerce/base-components/radio-control';
+
+/** @typedef {import('@woocommerce/type-defs/contexts').CustomerPaymentMethod} CustomerPaymentMethod */
+
+/**
+ * Returns the option object for a cc or echeck saved payment method token.
+ *
+ * @param {CustomerPaymentMethod} savedPaymentMethod
+ *
+ * @return {Object} An option objects to use for RadioControl.
+ */
+const getCcOrEcheckPaymentMethodOption = ( { method, expires, tokenId } ) => {
+	return {
+		value: tokenId,
+		label: sprintf(
+			__(
+				'%1$s ending in %2$s (expires %3$s)',
+				'woo-gutenberg-product-blocks'
+			),
+			method.brand,
+			method.last4,
+			expires
+		),
+		name: `wc-saved-payment-method-token-${ tokenId }`,
+	};
+};
+
+/**
+ * Returns the option object for any non specific saved payment method.
+ *
+ * @param {CustomerPaymentMethod} savedPaymentMethod
+ *
+ * @return {Object} An option objects to use for RadioControl.
+ */
+const getDefaultPaymentMethodOptions = ( { method, tokenId } ) => {
+	return {
+		value: tokenId,
+		label: sprintf(
+			__( 'Saved token for %s', 'woo-gutenberg-products-block' ),
+			method.gateway
+		),
+		name: `wc-saved-payment-method-token-${ tokenId }`,
+	};
+};
+
+const SavedPaymentMethodOptions = ( { onSelect } ) => {
+	const {
+		setPaymentStatus,
+		customerPaymentMethods,
+		activePaymentMethod,
+	} = usePaymentMethodDataContext();
+	const [ selectedToken, setSelectedToken ] = useState( null );
+
+	/**
+	 * @type      {Object} Options
+	 * @property  {Array}  current  The current options on the type.
+	 */
+	const currentOptions = useRef( [] );
+	useEffect( () => {
+		let options = [];
+		const paymentMethodKeys = Object.keys( customerPaymentMethods );
+		if ( paymentMethodKeys.length > 0 ) {
+			paymentMethodKeys.forEach( ( type ) => {
+				const paymentMethods = customerPaymentMethods[ type ];
+				if ( paymentMethods.length > 0 ) {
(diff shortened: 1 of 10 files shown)


assets/js/base/components/payment-methods/saved-payment-method-options.js
```diff
@@ -0,0 +1,137 @@
...
+		value: tokenId,
+		label: sprintf(
+			__( 'Saved token for %s', 'woo-gutenberg-products-block' ),
+			method.gateway
+		),
+		name: `wc-saved-payment-method-token-${ tokenId }`,
+	};
+};
+
+const SavedPaymentMethodOptions = ( { onSelect } ) => {
```
haszari: Is this component specific to Stripe or should it work with multiple payment method providers (i.e. extensions)? 

My understanding is that it's a saved stripe token for a shopper's specific card.

If this is specific to stripe, should this be in a stripe subfolder?

assets/js/base/context/cart-checkout/payment-methods/actions.js
```diff
@@ -80,3 +86,27 @@ export const success = ( { billingData, paymentMethodData } ) => ( {
...
+} );
+
+/**
+ * Used to dispatch an action for updating a registered express payment
+ * method in the state.
+ *
+ * @param {Object} paymentMethod Payment method to register.
+ * @return {Object} An action object.
+ */
+export const setRegisteredExpressPaymentMethod = ( paymentMethod ) => ( {
```
haszari: Should I be testing express payment flows? I was testing with this area at the bottom of checkout, this is regular payment right (not "express")? 

Screen Shot 2020-03-26 at 3 03 15 PM

garymurray:
> Another consideration – the term "payment method" may mean two different things:

> We'll need to be clear on these two different concepts in the UX and in the APIs. Perhaps we need to define these terms now, based on patterns in core or payments extensions. cc @garymurray @LevinMedia - you may have already covered this in the design?

@haszari where are you seeing the confusion in the use of the term "Payment Method" - in the actual name of that section of the cart - or alongside the actual payment option?

If it is for the heading of the section - we could just simplify it to be `Payment` as per the Shopify checkout:

Screenshot 2020-03-26 at 09 15 51

nerrad:
> Save payment information to my account for future purchases. checkbox – in checkout shortcode, this is associated with Stripe, when I select Cash on delivery it disappears. In the checkout block this is always displayed – I think it should be associated with payment methods.

It is in the stripe branch I'm working on.

> I only see the fake payment method content until I've saved a card to stripe. I'm assuming that work is in a separate PR coming soon, though it's strange to test saved card functionality before allowing user to select a payment method. To clarify, this is what I see when I have Stripe & COD set up, but the user has no saved card:

It is expected to not see the radios if there are no saved cards.  Other than that I'm not certain what this comment refers to. If it refers to the checkbox for saving a payment, then as noted above, that checkbox is moved to be something payment methods implement as a part of their integration in the stripe branch I have in progress.

> Another consideration – the term "payment method" may mean two different things:

Personally, I don't see how that's confusing. It literally means _a method of payment_. From the customer's perspective that's all that matters, "How do I pay for these things?"... here's some payment methods.  From the developer standpoint we may care more about the specifics (payment gateway, vs payment processor etc) but I think this is overthinking things.  I'm okay with shortening it to "Payment" as @garymurray suggested but I think that's a bit ambiguous because I think "Payment" is more referring to the actual amount paid, i.e. "Susan made a payment on her bill" vs "Susan used the Credit Card payment method to make a payment on her bill".

Anyways, it's semantics so I'm happy to roll with whatever is desired here.

assets/js/base/components/payment-methods/saved-payment-method-options.js
nerrad: > If this is specific to stripe, should this be in a stripe subfolder?

No. It's modelled after functionality that _exists in WooCommerce core_ currently (php side).

assets/js/base/context/cart-checkout/payment-methods/actions.js
nerrad: Ya the screenshot is regular payment methods. There are no explicit tests for express payment flows other than if the payment method icons appeared in express payment container, then the this code is covered by testing.  I verified that in my testing so we should be good.

haszari:
> @haszari where are you seeing the confusion in the use of the term "Payment Method" - in the actual name of that section of the cart - or alongside the actual payment option?

My confusion was mostly around the code – understanding the role of each API/module. 

I only mentioned this because the model in the code will often "leak" into the UX at some point, or worse, if the model in the code is confused, it's harder to build an intuitive UI.

Let's keep an eye on this as we iterate – no need to rename anything now, I just wanted to raise this as a consideration. How we model payment "types" (e.g. COD vs Stripe vs Apple) vs saved methods (card ending in 4222) is important for understanding the code base (maintainability/onboarding), and building the right APIs for payments devs.

assets/js/base/components/payment-methods/saved-payment-method-options.js
haszari: Thanks for clarifying. Perhaps this is a saved "instance" of a payment method with details, e.g. "stripe + card 1234"?

PR #2033, opened 2020-03-26
Hook up checkout validation with context

### LINKED ISSUE #1939: Checkout Context and Validation: Improve `onCheckoutProcessing` event emitter to allow for aborted subscribers.

Aljullu:
Closes #1939.

### Screenshots
imatge
_(:point_up_2: errors added by the payment method)_

In order to test it, let's force the payment method to return an error in certain circumstances. In `payment-method-extensions/payment-methods/stripe/payment-method.js` L289, paste something like this:
```JS
if ( billingData.postcode !== 'ABCD' ) {
	paymentStatus
		.setPaymentStatus()
		.error( 'Stripe found an error with your postcode.' );
	const key = billingData.shippingAsBilling
		? 'shipping-postcode'
		: 'billing-postcode';
	return {
		[ key ]: {
			message: 'Postcode must be ABCD.',
		},
	};
}
```
That will make Stripe to return an error if the billing postcode is different from `ABCD`.
1. Go to the _Checkout_ block and enter some data.
2. Feel free to leave some required fields empty or introduce an invalid email. We are testing validation!
3. Click on `Place Order` and verify errors are displayed properly.

### Notes
Currently, not all errors will appear at once: first Checkout validation errors and then, errors created by the payment method. I think this can be worked on in a follow-up and doesn't have to block this PR.

### DIFF
--- a/assets/js/base/components/select/validated.js
+++ b/assets/js/base/components/select/validated.js
@@ -52,6 +52,13 @@ const ValidatedSelect = ( {
 		validateSelect();
 	}, [ value ] );
 
+	// Remove validation errors when unmounted.
+	useEffect( () => {
+		return () => {
+			clearValidationError( errorId );
+		};
+	}, [] );
+
 	const error = getValidationError( errorId ) || {};
 
 	return (
--- a/assets/js/base/context/cart-checkout/checkout/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/index.js
@@ -45,8 +45,8 @@ export const CheckoutProvider = ( {
 					activePaymentMethod={ initialActivePaymentMethod }
 				>
 					<ShippingDataProvider>
-						{ children }
 						<CheckoutProcessor />
+						{ children }
 					</ShippingDataProvider>
 				</PaymentMethodDataProvider>
 			</BillingDataProvider>
--- a/assets/js/base/context/cart-checkout/checkout/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/index.js
@@ -45,8 +45,8 @@ export const CheckoutProvider = ( {
 					activePaymentMethod={ initialActivePaymentMethod }
 				>
 					<ShippingDataProvider>
-						<CheckoutProcessor />
 						{ children }
+						<CheckoutProcessor />
 					</ShippingDataProvider>
 				</PaymentMethodDataProvider>
 			</BillingDataProvider>
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -49,10 +49,6 @@ const CheckoutProcessor = ( { scrollToTop } ) => {
 	 * @return {boolean} True if everything was successful.
 	 */
 	const processCheckout = useCallback( async () => {
-		if ( withErrors ) {
-			return false;
-		}
-
 		await triggerFetch( {
 			path: '/wc/store/checkout',
 			method: 'POST',
@@ -95,8 +91,6 @@ const CheckoutProcessor = ( { scrollToTop } ) => {
 					dispatchActions.setRedirectUrl(
 						response.payment_result.redirect_url
 					);
-
-					// @todo do we need to trigger more handling here?
 				} );
 			} )
 			.catch( ( error ) => {
@@ -114,6 +108,10 @@ const CheckoutProcessor = ( { scrollToTop } ) => {
 		withErrors,
 	] );
 
+	const checkValidation = useCallback( () => {
+		return ! withErrors;
+	}, [ withErrors ] );
+
 	/**
 	 * When the checkout is processing, process the order.
 	 */
@@ -123,6 +121,15 @@ const CheckoutProcessor = ( { scrollToTop } ) => {
 			unsubscribeProcessing();
 		};
 	}, [ onCheckoutProcessing, processCheckout ] );
+
+	useEffect( () => {
+		const unsubscribeCompleteSuccess = onCheckoutProcessing(
+			checkValidation
+		);
+		return () => {
+			unsubscribeCompleteSuccess();
+		};
+	}, [ onCheckoutProcessing, checkValidation ] );
 
 	useEffect( () => {
 		const unsubscribeCompleteError = onCheckoutCompleteError( () => {
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -8,23 +8,35 @@ import {
 	useShippingDataContext,
 	useBillingDataContext,
 	usePaymentMethodDataContext,
+	useValidationContext,
 } from '@woocommerce/base-context';
 import { useEffect, useRef, useCallback } from '@wordpress/element';
 import { useStoreNotices } from '@woocommerce/base-hooks';
+import withScrollToTop from '@woocommerce/base-hocs/with-scroll-to-top';
 
 /**
  * CheckoutProcessor component. @todo Needs to consume all contexts.
  *
  * Subscribes to checkout context and triggers processing via the API.
  */
-const CheckoutProcessor = () => {
-	const { onCheckoutProcessing, dispatchActions } = useCheckoutContext();
+const CheckoutProcessor = ( { scrollToTop } ) => {
+	const {
+		onCheckoutProcessing,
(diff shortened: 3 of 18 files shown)


assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -8,36 +8,73 @@ import {
...
+				id: 'payment-method-error',
+			} );
+		} else {
+			removeNotice( 'payment-method-error' );
+		}
+	}, [ errorMessage ] );
+
+	const checkValidation = useCallback( () => {
+		return ! withErrors;
+	}, [ withErrors ] );
```
nerrad, line 65: Genius! I love how this removes the logic from the checkout button component!

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -79,35 +116,39 @@ const CheckoutProcessor = () => {
...
+
+	useEffect( () => {
+		const unsubscribeCompleteError = onCheckoutCompleteError( () => {
+			showAllValidationErrors();
+			scrollToTop( { focusableSelector: 'input:invalid' } );
+		} );
+		return () => {
+			unsubscribeCompleteError();
+		};
+	}, [ onCheckoutCompleteError ] );
```
nerrad, line 149: Nice!

assets/js/base/context/cart-checkout/checkout/index.js
```diff
@@ -45,8 +45,8 @@ export const CheckoutProvider = ( {
 					activePaymentMethod={ initialActivePaymentMethod }
 				>
 					<ShippingDataProvider>
-						{ children }
 						<CheckoutProcessor />
+						{ children }
```
nerrad, line 49: It's odd how you had to move this (I saw the commit message). I would have expected the reverse to be needed...

assets/js/base/context/cart-checkout/event-emit/emitters.js
```diff
@@ -16,11 +16,11 @@
...
 export const emitEvent = async ( observers, eventType, data ) => {
 	const observersByType = observers[ eventType ]
-		? Object.values( observers[ eventType ] )
+		? observers[ eventType ].values()
 		: [];
-	for ( let i = 0; i < observersByType.length; i++ ) {
+	for ( const observer of observersByType ) {
 		try {
-			await Promise.resolve( observersByType[ i ]( data ) );
+			await Promise.resolve( observer( data ) );
```
nerrad, line 23: Oh! nice improvements here!

assets/js/base/context/cart-checkout/event-emit/reducer.js
```diff
@@ -33,19 +33,19 @@ export const actions = {
...
 			return {
 				...state,
-				[ eventType ]: {
-					...state[ eventType ],
-					[ id ]: callback,
-				},
+				[ eventType ]: newEvents,
 			};
 		case TYPES.REMOVE_EVENT_CALLBACK:
+			newEvents.delete( id );
```
nerrad, line 45: I love how switching to `Map` removes a dependency, it seems to make the logic clearer to and reduces some code.

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -165,8 +165,10 @@ export const ShippingDataProvider = ( { children } ) => {
 
 	// dispatch checkout error if there's a shipping error.
 	useEffect( () => {
-		dispatchActions.setHasError( currentErrorStatus.hasError );
-	}, [ currentErrorStatus, dispatchActions.setHasError ] );
+		if ( currentErrorStatus.hasError ) {
+			dispatchActions.setHasError( true );
+		}
+	}, [ currentErrorStatus.hasError, dispatchActions.setHasError ] );
```
nerrad, line 171: A problem with this change is it means if the shipping error is cleared, then the checkout error status won't get cleared. If there is potential for race conditions here, we may have to (not in this pull) add similar treatment for checkout hasError as what I did for the `isCalculating` state (keep count of things causing an error and increment/decrement count on those individual error state changes).

Aljullu: Yeah... makes sense. I removed the conditional for now (dc1901580e7812b1eef6585b32d3e5f19ac21a62) and created a follow-up, so we don't block this PR: #2096. I can start working on that one once this PR is merged.

I just wonder what's the benefit on having an error counter with increment/decrement instead of an array of error ids (or even better, a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)). Both solutions would work, but a set of error ids would make our lives easier when debugging IMO. What do you think?

assets/js/base/context/cart-checkout/checkout/index.js
Aljullu: Yeah, that's because I moved the `onCheckoutCompleteError` hook to the CheckoutProcessor instead of the CheckoutBlock, so when calling `scrollToTop`, it needed to be on top.

But thinking about it twice, I think that effect belongs more to the block than the processor, so I moved it back in 56164430a05ea5ca5587bd5e2462d453496d43c1.

assets/js/base/components/select/validated.js
```diff
@@ -52,6 +52,13 @@ const ValidatedSelect = ( {
 		validateSelect();
 	}, [ value ] );
 
+	// Remove validation errors when unmounted.
+	useEffect( () => {
+		return () => {
+			clearValidationError( errorId );
```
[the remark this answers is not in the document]
Aljullu: Does it make sense leaving this as-is for now and working on this in parallel with #2096?

assets/js/base/context/cart-checkout/shipping/index.js
nerrad: > Both solutions would work, but a set of error ids would make our lives easier when debugging IMO. What do you think?

Good thoughts, depending on what you end up storing as references for tracking, you may want to use a WeakSet to protect against potential memory leaks (from unmounts etc). The only reason I suggested an error counter is because it's trivial to implement and memory impact is insignificant.

assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -126,10 +114,22 @@ const Checkout = ( {
 	);
 	useEffect( () => {
 		if ( shippingAsBilling ) {
-			setBillingData( shippingAddress );
+			setBillingData( { ...shippingAddress, shippingAsBilling } );
+		} else {
+			setBillingData( { shippingAsBilling } );
 		}
 	}, [ shippingAsBilling, setBillingData ] );
```
senadir, line 122: what was the point of adding shippingAsBilling to shipping data? It’s not paired with the backend and I don’t see it being used anywhere else

Aljullu: It's used in the example above. I thought it would be useful for extensions to know if the shipping and billing addresses are the same. That would also allow them to display the errors messages in the correct fields.

PR #2038, opened 2020-03-27
Ensure woocommerceanalytics checkout event fires for checkout block

### LINKED ISSUE #2031: Ensure woocommerceanalytics_product_checkout fires for stores that use blocks
We'll use the `woocommerceanalytics` events (from [Jetpack](https://github.com/Automattic/jetpack/tree/master/modules/woocommerce-analytics)) for measuring conversion rate of the new block-based cart & checkout, and comparing with existing (shortcode) flow.

We need the `checkout` event to fire correctly for stores using blocks, in particular when the checkout block is viewed.

See also #2032 #2030

haszari:
Fixes #2031

This fires the existing woocommerce core [`woocommerce_after_checkout_form` hook](https://github.com/woocommerce/woocommerce/blob/master/templates/checkout/form-checkout.php#L66) after the checkout block.

The analytics event is [fired off this hook](https://github.com/Automattic/jetpack/blob/master/modules/woocommerce-analytics/classes/wp-woocommerce-analytics-universal.php#L46), so firing the hook ensures the event is triggered. At face value this seems appropriate – this is after the basic checkout render – but we may need to review other uses of the hook to check that this approach works well and doesn't cause issues with other extensions/themes using this hook.

### Screenshots

Screen Shot 2020-03-27 at 4 39 18 PM

0. Test on incognito window, or logged in as customer - analytics aren't fired for admins.
3. Install & activate Jetpack plugin in your dev site (running this branch of blocks plugin). To do this you may need to set up an ngrok tunnel or similar; see [Jetpack repo for more info](https://github.com/Automattic/jetpack/blob/master/docs/development-environment.md).
2. View network tab in dev tools, search for `pixel`.
1. Add stuff to your cart, view checkout block.
3. Confirm that `woocommerceanalytics_product_checkout` event is logged, and parameters are correct (e.g. product info, quantity).

Also check that this event still fires on shortcode based checkout, and doesn't fire when it shouldn't :)

### DIFF
--- a/src/BlockTypes/Checkout.php
+++ b/src/BlockTypes/Checkout.php
@@ -82,10 +82,17 @@ class Checkout extends AbstractBlock {
 			$data_registry->add( 'shippingMethodsExist', $methods_exist );
 		}
 
+		\Automattic\WooCommerce\Blocks\Assets::register_block_script( $this->block_name . '-frontend', $this->block_name . '-block-frontend' );
+
 		do_action( 'woocommerce_blocks_enqueue_checkout_block_scripts_before' );
 		Assets::register_block_script( $this->block_name . '-frontend', $this->block_name . '-block-frontend' );
 		do_action( 'woocommerce_blocks_enqueue_checkout_block_scripts_after' );
-		return $content . $this->get_skeleton();
+		
+		$skeleton = $this->get_skeleton();
+		$checkout = WC()->checkout();
+		do_action( 'woocommerce_after_checkout_form', $checkout );
+
+		return $content . $skeleton;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


nerrad:
@haszari I had to revert this from master. `woocommerce_after_checkout_form` is an action hook that fires in a _template_ file found in WooCommerce core which means that anything hooking into it could be expected to _output_ content. Implementing it in the block render method is a no-go because that method is expected to _return_ content, so outputting headers in the action would break things.

Instead, it's possible the following will work:

- create a new action hook to add in the place you added this one.
- in the plugin adding the analytics event, hook into the new hook.

PR #2041, opened 2020-03-27
Cart and Checkout accessibility fixes

### LINKED ISSUE #1538: Cart block - accessibility review and related fixes
Some accessibility issues in the cart line items UX were noted during #1333. This issue covers reviewing the cart UX for any accessibility problems and fixing them.

Note that PRs for cart work should not ignore accessibility – this issue is here to ensure we review the whole block once the UX comes together.

Aljullu:
Fixes #1538.

This PR fixes several accessibility issues in the _Cart_ and _Checkout_ blocks.

#### Accessibility

I think each commit description is quite self-explanatory of what needs to be tested but...

Instead of testing each individual commit one-by-one, I think it's better just to test the entire block with a screen reader & keyboard navigation. If you find any issue that is not fixed in this PR or has an issue, report it! :slightly_smiling_face:

### DIFF
--- a/assets/js/base/context/cart-checkout/validation/index.js
+++ b/assets/js/base/context/cart-checkout/validation/index.js
@@ -151,12 +151,13 @@ export const ValidationContextProvider = ( { children } ) => {
 	 * Provides an id for the validation error that can be used to fill out
 	 * aria-describedby attribute values.
 	 *
-	 * @param {string} inputId The input css id the validation error is related
+	 * @param {string} errorId The input css id the validation error is related
 	 *                         to.
 	 * @return {string} The id to use for the validation error container.
 	 */
-	const getValidationErrorId = ( inputId ) => {
-		return inputId ? `validate-error-${ inputId }` : '';
+	const getValidationErrorId = ( errorId ) => {
+		const error = getValidationError( errorId );
+		return error && ! error.hidden ? `validate-error-${ errorId }` : '';
 	};
 
 	const context = {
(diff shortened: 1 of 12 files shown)


assets/js/base/context/cart-checkout/validation/index.js
```diff
@@ -151,12 +151,13 @@ export const ValidationContextProvider = ( { children } ) => {
...
-	 * @param {string} inputId The input css id the validation error is related
+	 * @param {string} errorId The input css id the validation error is related
 	 *                         to.
 	 * @return {string} The id to use for the validation error container.
 	 */
-	const getValidationErrorId = ( inputId ) => {
-		return inputId ? `validate-error-${ inputId }` : '';
+	const getValidationErrorId = ( errorId ) => {
+		const error = getValidationError( errorId );
+		return error && ! error.hidden ? `validate-error-${ errorId }` : '';
```
senadir: this seems hard to understand and I'm not sure if it will return `validate-error-${ errorId }` or `''` or will return the expression `error && resultOfTernary` 
```suggestion
		return ( error && ! error.hidden ) ? `validate-error-${ errorId }` : '';
```

Aljullu: Those parentheses are removed by ESLint. I refactored it to avoid using a ternary here, hope it makes code clearer: 5c2ff5597c2eda1c653052640973ebfd1d17cedc.

PR #2044, opened 2020-03-27
Checkout processing and order creation APIs

mikejolley:
Updates the `CheckoutProcessor` component to take details from our contexts and send to the API. The API processes the checkout/updates the order and returns the response.

On the API side this:
- Makes `/checkout` responsible for reading/creating a draft order, as well as handling payment
- Formalises billing address schema
- Removes draft order ID/create order from `cart` endpoints

The draft order is passed to `CheckoutContext` via a setting. We need error handling if this fails.

Follow ups:
- Create error state if no draft order is created.
- Check post-order creation handling and order creation. This creates the order but the checkout does not proceed.
- Update api docs and unit tests.


assets/js/base/context/cart-checkout/checkout/index.js
```diff
@@ -195,6 +199,8 @@ export const CheckoutProvider = ( {
 		onCheckoutProcessing,
 		dispatchActions,
 		isEditor,
+		orderId: checkoutState.orderId,
+		hasDraftOrder: !! checkoutState.orderId,
```
nerrad: So when `orderId === 0` that means this is a draft order?

mikejolley: No, draft orders have ids. !! just ensures we have a bool. Could have wrote it like `checkoutState.orderId > 0`

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -1,12 +1,111 @@
...
+	 *
+	 * @return {boolean} True if everything was successful.
+	 */
+	const processCheckout = useCallback( async () => {
+		await triggerFetch( {
+			path: '/wc/store/checkout',
+			method: 'POST',
+			data: {
+				payment_method: activePaymentMethod,
+				payment_data: [],
```
nerrad: Is including the payment data  in the checkout processing to be done in a follow-up?

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -1,12 +1,111 @@
...
+								}
+							);
+						}
+					}
+
+					dispatchActions.setRedirectUrl(
+						response.payment_result.redirect_url
+					);
+
+					// @todo do we need to trigger more handling here?
```
nerrad: The checkout processing event emitter responds to what gets returned from subscribers. We can use this for error handling in the followup but basically if you return _anything_ but `true` then the processing aborts on that subscriber and sets the checkout state to `hasError`. The idea is if you return an error object like this:

```jsx
const error  =  {
        message: 'Main error message for top level notice'
        validation: [
             { propertyFailingValidation: 'error message for field' },
             { anotherPropertyFailingValidation: 'error message for field' },
        ]
};
```
Then checkout will  use that both to determine the type of error notices shown and/or any inline validation error notices to show.  This has overlap with the work @Aljullu is doing on #1939 and will likely also require some thought into how payment methods will pick up inline validation errors (or at a minimum we should probably display payment method related errors in context of the payment method step).

assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -126,6 +126,11 @@ const Checkout = ( {
 			setBillingData( shippingAddress );
 		}
 	}, [ shippingAsBilling, setBillingData ] );
+
+	if ( ! isEditor && ! hasDraftOrder ) {
+		return <div>No draft order - add error state.</div>;
+	}
```
nerrad: Add a @todo comment here so we don't miss?

src/BlockTypes/Checkout.php
```diff
@@ -48,63 +50,74 @@ public function register_block_type() {
...
+		$block_data = [
+			'allowedCountries'  => [ WC()->countries, 'get_allowed_countries' ],
+			'shippingCountries' => [ WC()->countries, 'get_shipping_countries' ],
+			'allowedStates'     => [ WC()->countries, 'get_allowed_country_states' ],
+			'shippingStates'    => [ WC()->countries, 'get_shipping_country_states' ],
+		];
+
+		foreach ( $block_data as $key => $callback ) {
+			if ( ! $data_registry->exists( $key ) ) {
+				$data_registry->add( $key, call_user_func( $callback ) );
```
nerrad: nice 👏

src/BlockTypes/Checkout.php
```diff
@@ -48,63 +50,74 @@ public function register_block_type() {
...
-		if ( is_user_logged_in() && ! $data_registry->exists( 'customerPaymentMethods' ) ) {
-			add_filter( 'woocommerce_payment_methods_list_item', [ $this, 'include_token_id_with_payment_methods' ], 10, 2 );
-			$data_registry->add(
-				'customerPaymentMethods',
-				wc_get_customer_saved_methods_list( get_current_user_id() )
-			);
-			remove_filter( 'woocommerce_payment_methods_list_item', [ $this, 'include_token_id_with_payment_methods' ], 10, 2 );
+
+		$this->hydrate_from_api( $data_registry );
+		$this->hydrate_customer_payment_methods( $data_registry );
```
nerrad: Happy to see these being extracted as well 👍

src/BlockTypes/Checkout.php
```diff
@@ -48,63 +50,74 @@ public function register_block_type() {
...
+
+		$permalink = ! empty( $attributes['cartPageId'] ) ? get_permalink( $attributes['cartPageId'] ) : false;
+
+		if ( $permalink && ! $data_registry->exists( 'page-' . $attributes['cartPageId'] ) ) {
+			$data_registry->add( 'page-' . $attributes['cartPageId'], get_permalink( $attributes['cartPageId'] ) );
```
nerrad: Does there need to be another call to `get_permalink` here? Could you use `$permalink`?

src/RestApi/StoreApi/Routes/Checkout.php
```diff
@@ -43,63 +45,116 @@ public function get_path() {
...
 	 */
 	protected function get_route_post_response( \WP_REST_Request $request ) {
-		$order = $this->get_request_order_object( $request );
+		$order_controller = new OrderController();
+		$order_object     = $this->get_draft_order_object( $this->get_draft_order_id() );
+
+		if ( ! $order_object ) {
+			throw new RouteException(
+				'woocommerce_rest_checkout_invalid_order',
+				__( 'No order pending payment found.', 'woo-gutenberg-products-block' ),
```
nerrad: Should this be `No draft order, or order needing payment found.`?

src/RestApi/StoreApi/Routes/Checkout.php
```diff
@@ -164,7 +326,7 @@ protected function process_payment( \WC_Order $order, \WP_REST_Request $request
 			 * @param PaymentContext $context Holds context for the payment, including order ID and payment method.
 			 * @param PaymentResult  $result Result object for the transaction.
 			 */
-			do_action( 'woocommerce_rest_checkout_process_payment_with_context', $context, $result );
+			do_action_ref_array( 'woocommerce_rest_checkout_process_payment_with_context', [ $context, &$result ] );
```
nerrad: It shouldn't be necessary to pass `PaymentResult` as a reference value since in PHP all objects are automatically references anyways right?

assets/js/base/context/cart-checkout/checkout/processor/index.js
mikejolley: Depends; is it ready to go now via the provider?

src/PaymentMethodIntegrations/Stripe.php
```diff
@@ -77,7 +77,7 @@ public function register_assets() {
 	 */
 	public function enqueue_data() {
 		$available_gateways = WC()->payment_gateways->get_available_payment_gateways();
-		$stripe_gateway     = $available_gateways['stripe'] ? $available_gateways['stripe'] : null;
+		$stripe_gateway     = isset( $available_gateways['stripe'] ) ? $available_gateways['stripe'] : null;
```
nerrad: Since `::get_available_gateways()` returns a filtered value, I think we should also be doing an `instanceof` check here to make sure it's the Stripe gateway we expect here.

src/RestApi/StoreApi/Routes/Checkout.php
mikejolley: That is what I thought but I spotted some notices in my error log which went away with this change. WP core does this when it explicitly needs references.

assets/js/base/context/cart-checkout/checkout/index.js
nerrad: Hmm, the name of the property seems to suggest that there might be an orderId but it might not be for a draft order.  Could this just be `hasOrder`?

assets/js/base/context/cart-checkout/checkout/processor/index.js
nerrad: Well I'm mostly concerned that we could _miss_ this. As long as it's documented for a follow-up I'm fine with it happening after this pull is merged.

src/RestApi/StoreApi/Routes/Checkout.php
nerrad: Odd, and concerning.  I assumed WP core just has it because of legacy code.

src/PaymentMethodIntegrations/Stripe.php
mikejolley: `is_a`, because `instanceof` would fail if the class didn't exist.

PR #2069, opened 2020-03-30
Add checkout error state

### LINKED ISSUE #2060: Missing draft order error state handling
Follow up from #2044

When no draft order exists, or failed to be created for some reason, we need to show a user facing error message. It's likely the cart contained something invalid or out of stock.

See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2044/files#diff-a6e2415457b69288f082ab8bd1837c96R131

mikejolley:
Shows an error state if a draft order cannot be created, for example, if it's out of stock.

Fixes #2060

### Screenshots

Screenshot 2020-03-30 at 16 21 01

1. Add item to cart.
2. Edit the item and make it out of stock.
3. Go to block checkout. See error message.


assets/js/blocks/cart-checkout/checkout/checkout-error/index.js
```diff
@@ -0,0 +1,61 @@
...
+				{ __(
+					'There was a problem with your cart',
+					'woo-gutenberg-products-block'
+				) }
+			</strong>
+			<CheckoutErrorMessage />
+			<span className="wp-block-button">
+				<a href={ CART_URL } className="wp-block-button__link">
+					{ __( 'View your cart', 'woo-gutenberg-products-block' ) }
+				</a>
```
nerrad: The `CheckoutError` component appears to be setup to handle _any_ error that prevents checkout from rendering. Given that, will a `View Cart` link always be relevant to the error?  Will the "...problem with your cart" message always be relevant?

It _appears_ that this is tailored to error's having cart related issues as the root cause. With that in mind, I wonder if I'm getting tripped up more by the component name and a different name might work better? Something like `CheckoutCartError` or `CartError`?

mikejolley: Ill adjust this and add some more generic button/message handling based on error code.

mikejolley:
Screenshot 2020-03-31 at 11 24 19
Better :)

Issue #1944, opened 2020-03-11
Checkout: Finish integrating shipping store into the ShippingMethodDataProvider


nerrad:
The `ShippingMethodDataProvider` has some initial work for this implementation but the pieces need wiring up:

### Shipping Rate Selection

- values and setters for need connected to the store for the context provider
- event emitters for shipping rate need connected to store state.

### Shipping Rates

- The current logic found in the context is out of date. It will need updated to connect with the current `useShippingRates` hook (including error handling etc).

### Wiring up checkout.calculating state dispatches.

- Checkout status for `isCalculating` is derived from the checkout.calculating state ( updated via [calculating incrementing/decrementing dispatches](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/82cc6c6dbb0b6c4b4d2ef3d4da1143907ec5fa06/assets/js/base/context/cart-checkout/checkout/index.js#L115-L118)). This will need wired up to the shipping rate loading state so that calculating increments/decrements accordingly.
- Checkout status for when there's an error should be wired up to the shipping rate error state.

PR #2070, opened 2020-03-30
Finish wiring up shipping data context provider and add disabled state to checkout submit button

nerrad:
Fixes: #1944 

This pull finishes off the `ShippingDataProvider` component and implements the following:

- Adds selecting rate resolution state to the cart store and exposes on `useSelectShippingRate` hook.
- Refactors `ShippingDataProvider` so that:
    - remove un-needed component and move shipping rates into context provider (via `useShippingRates`).
    - remove error handling (currently the ui/ux for the checkout should result in no errors from ui/ux and any api errors are non-recoverable - thus thrown from the hooks). This may need revisited as we test but for now I'm removing the code that does nothing. 
    - Increment/decrement checkout calculation counts when shipping rates are being selected (shipping rate retrieval is already doing this).
- dispatch checkout error state update if there is a shipping rate error.
- disable the place order button component if checkout is calculating or checkout has an error.

## To Test

- This impacts the cart because of the hook changes. So verify that existing behaviour for shipping related ui/ux in the cart remains unaffected.
- This impacts anything to do with shipping in the Checkout block. So verify that behaviour for shipping related calculations and total updates continue to work as expected.
- Also verify in the Checkout block that when shipping rates are being changed, or selected that the place order button is disabled (and enables again after network request is done).


assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -61,33 +57,11 @@ export const useShippingDataContext = () => {
...
-		'country',
-		'state',
-		'city',
-		'postcode',
-	] );
-	useEffect( () => {
-		onChange( shippingRates, shippingRatesLoading );
-	}, [ shippingRates, shippingRatesLoading ] );
-	return null;
-};
```
nerrad: This component was initially implemented before the shipping address refactor and now it's obsolete so I've removed it and moved `useShippingRates` into the context provider.

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -119,28 +92,22 @@ export const ShippingDataProvider = ( { children } ) => {
...
-	// emit the events (see emit events block)
-	const onRateChange = ( shippingRates, shippingRatesLoading, error ) => {
-		setShippingOptions( shippingRates );
-		setShippingOptionsLoading( shippingRatesLoading );
-		if ( ! shippingRatesLoading && error && error.type ) {
-			// @todo this type might need normalizing to something recognized by
-			// the ERROR_TYPE constants.
-			dispatchErrorStatus( setStatusAction( error.type ) );
-		} else if ( ! shippingRatesLoading && shippingRates ) {
-			dispatchErrorStatus( NONE );
```
nerrad: Originally I thought we'd be potentially getting errors like invalid address/country type errors from the api. However, with the current error handling in `useShippingRates` and `useSelectShippingRate` hooks, the assumption is being made that any error coming from the api is a non-recoverable error because of the current ui/ux setup (known countries, states etc).

This is subject to being revisited at some point because it may still be possible for extensions to hook in eventually and they might provide invalid values (from a ui/ux we have less control over).  If needed we can update in future work (which will include handling appropriate error types in the relevant hooks).

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -154,36 +121,71 @@ export const ShippingDataProvider = ( { children } ) => {
...
+		if ( currentErrorStatus.hasError ) {
+			dispatchActions.setHasError();
+		} else {
+			dispatchActions.clearError();
+		}
+	}, [
+		currentErrorStatus,
+		dispatchActions.setHasError,
+		dispatchActions.clearError,
+	] );
```
nerrad: This is a bit iffy.  There's _potential_ for a race condition if there are other things that might trigger checkout error (i.e. validation error/payment error etc).  We may want checkout error state to be something that is controlled in a child component (`CheckoutProcessing`?) that can set the checkout error based on a combination of any known error statuses from the other providers (shipping/billing/payment-method, validation etc).

assets/js/data/cart/selectors.js
```diff
@@ -147,11 +137,18 @@ export const isItemQuantityPending = ( state, cartItemKey ) => {
...
  * @return {boolean} are shipping rates loading.
  */
-export const areShippingRatesLoading = createRegistrySelector(
-	( select ) => ( state ) => {
-		return !! (
-			state.metaData.updatingShipping ||
-			select( STORE_KEY ).isResolving( 'getCartData', [] )
-		);
-	}
-);
```
nerrad: I think I originally suggested `areShippingRatesLoading` considering whether cart data is loading.  However I _think_ that was before the shipping address refactor when cart loading state could be triggered by shipping rate update state.  However, we need a _distinct_ loading state here for loading shipping rates so I reverted to only return that value. Testing seemed to be okay with this change.

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -154,36 +121,71 @@ export const ShippingDataProvider = ( { children } ) => {
...
 	}, [
-		shippingOptions,
-		shippingOptionsLoading,
+		shippingRates,
+		shippingRatesLoading,
 		currentErrorStatus,
 		shippingErrorStatus,
 	] );
 
+	//emit shipping rate selection events
```
mikejolley: comment formatting. 
```suggestion
	// emit shipping rate selection events.
```

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -154,36 +121,71 @@ export const ShippingDataProvider = ( { children } ) => {
...
+		currentErrorStatus,
+		shippingErrorStatus,
+	] );
+
+	// dispatch checkout error if there's a shipping error.
+	useEffect( () => {
+		if ( currentErrorStatus.hasError ) {
+			dispatchActions.setHasError();
+		} else {
+			dispatchActions.clearError();
```
mikejolley: Just a comment on these events; I checked and we're not storing errors here, just whether or not there is an error. Whats the point in having 2 sets of ACTIONS and EVENTS to `setHasError` and `clearError` (which has a differently named action set_no_error) when we could just have  function accepting a boolean? `dispatchActions.hasError( true | false );` Would make this easier to understand.

assets/js/base/context/cart-checkout/shipping/index.js
mikejolley: So move hasError to the CheckoutProvider, since it's the outermost provider? Is that what you mean?

assets/js/base/components/cart-checkout/place-order-button/index.js
```diff
@@ -21,6 +26,7 @@ const PlaceOrderButton = ( { validateSubmit } ) => {
 					onSubmit();
 				}
 			} }
+			disabled={ hasError || isCalculating }
```
senadir: nice!

assets/js/data/cart/actions.js
```diff
@@ -258,6 +273,7 @@ export function* changeCartItemQuantity( cartItemKey, quantity ) {
  */
 export function* selectShippingRate( rateId, packageId = 0 ) {
 	try {
+		yield shippingRatesBeingSelected( true );
```
senadir: so a question, can’t we rely on functions like `hasFinishedResolution` never understood when can we use them.

assets/js/base/context/cart-checkout/shipping/index.js
nerrad: No, I mean in the `CheckoutProcessor`. The checkout processor component will have access to all the error states from the various providers and could use those to derive (and update) the checkout provider error state (so there'd be one place that checkout error state is maintained).

I'm still on the fence with this though, I think we could merge this and test how things work once everything else is wired up.

assets/js/base/context/cart-checkout/shipping/index.js
nerrad: I'll take care of in a followup (see #2081)

assets/js/data/cart/actions.js
nerrad: The resolution functions from `wp.data` are only useful for `resolvers` attached to selectors. There is no automatic enhancement/detection of action generators that use the api. That might be a good thing to add to `wp.data` some day but for now if we need to track resolution state of action generators we need to implement our own state for that.

PR #2072, opened 2020-03-30
add extra checks to checkout/cart/order APIs to avoid editor errors

### LINKED ISSUE #2071: Fatal error with checkout block in editor
## to reproduce
- Edit a page with checkout block. Or add checkout block to a page, save and reload.
- Fatal error!

Looks like the new cart apis (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2044) are accessing a nonexistent `WC()->session`.

```
2020-03-30T21:17:35+00:00 CRITICAL Uncaught Error: Call to a member function get() on null in /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block/src/RestApi/StoreApi/Routes/Checkout.php:187
Stack trace:
#0 /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block/src/RestApi/StoreApi/Routes/Checkout.php(233): Automattic\WooCommerce\Blocks\RestApi\StoreApi\Routes\Checkout->get_draft_order_id()
#1 /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block/src/RestApi/StoreApi/Routes/Checkout.php(95): Automattic\WooCommerce\Blocks\RestApi\StoreApi\Routes\Checkout->create_or_update_draft_order()
#2 /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block/src/RestApi/StoreApi/Routes/AbstractRoute.php(68): Automattic\WooCommerce\Blocks\RestApi\StoreApi\Routes\Checkout->get_route_response()
#3 /var/www/html/wp-includes/rest-api/class-wp-rest-server.php(946): Automattic\WooCommerce\Blocks\RestApi\StoreApi\Routes\AbstractRoute->get_response()
#4 /var/www/html/wp-includes/rest-api.php(429): WP_REST_ in /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block/src/RestApi/StoreApi/Routes/Checkout.php on line 187
```

haszari:
Fixes #2071

This PR fixes an issue with checkout block in editor – various order-related state is not available in editor/admin.

- Add escape clause to `Checkout::get_route_response()` - bail if `is_admin()`. Also add some redundant paranoid checks on `WC()` `cart` & `session` within methods used in that API response.
- Adds some default state to cart-checkout context constants where needed. 

Also an unrelated fix, issue introduced when I rebased a previous PR (missed before merging).

- Removes a redundant `register_block_script()` call in `src/BlockTypes/Checkout.php` `render()`.

1. Test checkout block in editor.
2. Test checkout block in front end.

Should be able to add the block, edit settings, publish, reload editor, etc. And should function correctly on front end.

### DIFF
--- a/assets/js/base/context/cart-checkout/billing/constants.js
+++ b/assets/js/base/context/cart-checkout/billing/constants.js
@@ -8,13 +8,7 @@
  */
 import { getSetting } from '@woocommerce/settings';
 
-const defaultCheckoutData = {
-	billing_address: {},
-};
-
-// Checkout data may be null in editor/admin.
-const checkoutData =
-	getSetting( 'checkoutData', defaultCheckoutData ) || defaultCheckoutData;
+const checkoutData = getSetting( 'checkoutData', {} );
 
 /**
  * @type {BillingData}
--- a/assets/js/base/context/cart-checkout/billing/constants.js
+++ b/assets/js/base/context/cart-checkout/billing/constants.js
@@ -8,7 +8,13 @@
  */
 import { getSetting } from '@woocommerce/settings';
 
-const checkoutData = getSetting( 'checkoutData', {} );
+const defaultCheckoutData = {
+	billing_address: {},
+};
+
+// Checkout data may be null in editor/admin.
+const checkoutData =
+	getSetting( 'checkoutData', defaultCheckoutData ) || defaultCheckoutData;
 
 /**
  * @type {BillingData}
--- a/src/BlockTypes/Checkout.php
+++ b/src/BlockTypes/Checkout.php
@@ -66,20 +66,23 @@ class Checkout extends AbstractBlock {
 			}
 		}
 
-		$this->hydrate_from_api( $data_registry );
-		$this->hydrate_customer_payment_methods( $data_registry );
-
 		$permalink = ! empty( $attributes['cartPageId'] ) ? get_permalink( $attributes['cartPageId'] ) : false;
 
 		if ( $permalink && ! $data_registry->exists( 'page-' . $attributes['cartPageId'] ) ) {
 			$data_registry->add( 'page-' . $attributes['cartPageId'], $permalink );
 		}
 
-		$screen = function_exists( 'get_current_screen' ) ? get_current_screen() : false;
+		// Hydrate the following data depending on admin or frontend context.
+		if ( is_admin() ) {
+			$screen = function_exists( 'get_current_screen' ) ? get_current_screen() : false;
 
-		if ( $screen && $screen->is_block_editor() && ! $data_registry->exists( 'shippingMethodsExist' ) ) {
-			$methods_exist = wc_get_shipping_method_count() > 0;
-			$data_registry->add( 'shippingMethodsExist', $methods_exist );
+			if ( $screen && $screen->is_block_editor() && ! $data_registry->exists( 'shippingMethodsExist' ) ) {
+				$methods_exist = wc_get_shipping_method_count() > 0;
+				$data_registry->add( 'shippingMethodsExist', $methods_exist );
+			}
+		} else {
+			$this->hydrate_from_api( $data_registry );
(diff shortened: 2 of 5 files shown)


assets/js/base/context/cart-checkout/billing/constants.js
```diff
@@ -8,7 +8,13 @@
...
 import { getSetting } from '@woocommerce/settings';
 
-const checkoutData = getSetting( 'checkoutData', {} );
+const defaultCheckoutData = {
+	billing_address: {},
+};
+
+// Checkout data may be null in editor/admin.
+const checkoutData =
+	getSetting( 'checkoutData', defaultCheckoutData ) || defaultCheckoutData;
```
senadir: Shouldn’t the first default be enough?

src/BlockTypes/Checkout.php
```diff
@@ -82,8 +82,6 @@ public function render( $attributes = array(), $content = '' ) {
 			$data_registry->add( 'shippingMethodsExist', $methods_exist );
 		}
 
-		\Automattic\WooCommerce\Blocks\Assets::register_block_script( $this->block_name . '-frontend', $this->block_name . '-block-frontend' );
```
senadir: Why is this redundant? where are we also including this?

assets/js/base/context/cart-checkout/billing/constants.js
haszari: Unfortunately it's coming back null. Not sure if this is a bug with `getSetting`, but that's what's happening. I agree would be better to simplify this.

I just tried `const checkoutData = getSetting( 'checkoutData', defaultCheckoutData );` and saw the js error - `Cannot read property 'billing_address' of null`.

src/BlockTypes/Checkout.php
haszari: See next few lines in the file, when I did a rebase I got two copies of the same code. (Always look past the raw diff when reviewing!)

assets/js/base/context/cart-checkout/billing/constants.js
senadir: yeah it only seems it check if the value is set, not the value by its own
https://github.com/woocommerce/woocommerce-gutenberg-products-block//blob/a1717cbbe811466fefac04b141be60128e56c356/assets/js/settings/shared/get-setting.js#L21-L27
we can either pass a filter function but that just complicates it, or we can add a check in PHP when we're defining this setting

senadir: something like this
https://github.com/woocommerce/woocommerce-gutenberg-products-block//blob/ee73458babfb6ef8d35987491d2f5900280f6255/src/BlockTypes/Checkout.php#L74-L76

senadir:
I'd prefer that put better guards to unset values than to write early bailouts in code, preferably, if the block is initiating an API request on mount that is causing it to break on the editor, we might search and eliminate that initial request (or make it happen depending on a user input). however I understand that master is currently broken and this fix might take time, so we can possibly merge this but work on a higher solution.

senadir:
as I suggested yesterday in slack, it's better to skip those early bailouts and wrap the data hydration in a class check, so wrapping those lines
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/src/BlockTypes/Checkout.php#L121-L126

```php
if ( WC()->customer instanceof \WC_Customer && WC()->session instanceof \WC_Session ) {
	if ( ! $data_registry->exists( 'cartData' ) ) {
		$data_registry->add( 'cartData', WC()->api->get_endpoint_data( '/wc/store/cart' ) );
	}
	if ( ! $data_registry->exists( 'checkoutData' ) ) {
		$data_registry->add( 'checkoutData', WC()->api->get_endpoint_data( '/wc/store/checkout' ) );
	}
}
```

PR #2076, opened 2020-03-31
Implement mechanism for reliably passing through server data in editor context.

nerrad:
Currently in the editor context if you have a registered block already in the post content, when you load the editor, the blocks `render` method will fire.  The render method is what is currently used to pass through server data for the block to consume. 

However, one flaw with this approach is that when the post content does _not_ have a block yet, the blocks `render` method won't have fired and thus anything in that logic will not get loaded in the editor instance.

The fix is to use the `enqueue_block_editor_assets` action to load the server data related things in the context of the block editor.

## To test

Without the work in this pull, the shipping options won't show in the editor context when adding a _new_ checkout (or cart block). So you can verify this fixes that by creating a new page and adding either the cart of checkout block and making sure shipping options are available.

- Verify adding cart or checkout blocks in new editor instance displays/works as expected. 
- Verify loading an editor instance with _existing_ cart or checkout blocks in the content works as expected.
- Verify loading cart or checkout block in the frontend works as expected.

### DIFF
--- a/src/BlockTypes/AbstractBlock.php
+++ b/src/BlockTypes/AbstractBlock.php
@@ -53,6 +53,17 @@ abstract class AbstractBlock {
 	 * Will enqueue any editor assets a block needs to load
 	 */
 	public function enqueue_editor_assets() {
-		// noop, expected that children will override if needed.
+		$this->enqueue_data();
+	}
+
+	/**
+	 * Extra data passed through from server to client for block.
+	 *
+	 * @param array $attributes  Any attributes that currently are available from the block.
+	 *                           Note, this will be empty in the editor context when the block is
+	 *                           not in the post content on editor load.
+	 */
+	protected function enqueue_data( array $attributes = [] ) {
+		// noop. Child classes should override this if needed.
 	}
 }
--- a/src/BlockTypes/AbstractBlock.php
+++ b/src/BlockTypes/AbstractBlock.php
@@ -53,7 +53,18 @@ abstract class AbstractBlock {
 	 * Will enqueue any editor assets a block needs to load
 	 */
 	public function enqueue_editor_assets() {
-		$this->enqueue_data();
+		$this->enqueue_data_once();
+	}
+
+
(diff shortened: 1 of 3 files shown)


src/BlockTypes/Checkout.php
```diff
@@ -49,6 +50,21 @@ public function register_block_type() {
 	 * @return string Rendered block type output.
 	 */
 	public function render( $attributes = array(), $content = '' ) {
+		$this->enqueue_data( $attributes );
```
senadir: why is this only happening in Checkout and not Cart?

nerrad: oversight :)

nerrad: Fixed in latest commits

senadir: yeah it was breaking Cart in frontend

src/BlockTypes/AbstractBlock.php
```diff
@@ -41,4 +48,33 @@ public function register_block_type() {
...
+	public function enqueue_editor_assets() {
+		$this->enqueue_data_once();
+	}
+
+
+	/**
+	 * Ensures the enqueue_data method is only called once in a request.
+	 */
+	public function enqueue_data_once() {
+		static $has_run = false;
```
senadir: I'm not strong in PHP, but as far as I understand, static member variable is linked to the class, not instances of that class, so this will be linked to `AbstractBlock` not classes that it inherits? if that’s the case, then it will only work for the first block that runs it, but I don’t think this is the case because I tried adding both cart and checkout and nothing happend

nerrad: Ya actually this logic I added is flawed in numerous ways. I'm going to remove it and try something else.

nerrad: I've removed it and created #2077 to handle in a followup.  The way this is implemented currently doesn't really do anything.

PR #2079, opened 2020-03-31
Cart item validation and notices in cart/checkout

### LINKED ISSUE #2068: Cart does not report item errors
**Describe the bug**
The cart does not show item errors if they change state.

**To Reproduce**

1. Add an item to the cart.
2. Edit the item and change the inventory options. Make it out of stock.
3. See block cart. No errors.
4. See core cart. Error notice.

**Expected behavior**
There should be some type of warning or notice in the cart view.

mikejolley:
Adds validation of cart items to the API and then displays any returned errors in the cart and checkout.
Fixes #2068

### Screenshots

Screenshot 2020-03-31 at 14 37 34
Screenshot 2020-03-31 at 14 37 20

1. Add items to cart, then edit them to either make out of stock, or so there isn't enough stock qty available.
2. View cart - see notice
3. Go to checkout - see notice
4. Request GET `wc/store/cart` and see `errors` array in response.


assets/js/blocks/cart-checkout/checkout/checkout-order-error/index.js
```diff
@@ -48,7 +49,11 @@ const CheckoutError = () => {
 const ErrorTitle = ( { errorData } ) => {
 	let heading = __( 'Checkout error', 'woo-gutenberg-products-block' );
 
-	if ( errorData.code === 'woocommerce_product_out_of_stock' ) {
+	if (
+		errorData.code === 'woocommerce_product_out_of_stock' ||
+		errorData.code === 'woocommerce_rest_cart_product_is_not_purchasable' ||
+		errorData.code === 'woocommerce_rest_cart_product_no_stock'
```
nerrad: I wonder if we should define (and export) constants for these error types?

mikejolley:
@nerrad 

> While this is better than what we have currently I wonder if user experience would be better if we automatically reduce quantities in the cart for an item and then give a notification on why that was done?

If that's something we need to do it's extra scope and needs defining - I'm also not sure where that logic should live, and when it should be done. This is just there for parity with core cart.

assets/js/blocks/cart-checkout/checkout/checkout-order-error/index.js
mikejolley: What would the benefits be of doing that?  Is this like the Action Names in wp.data being constants? Also not sure of the benefits of that either since the constant names are identical to the contents.

senadir: the point of constants is that you can’t typo them, and you can autocomplete them, if you do a typo, it will throw an error.
for this however, I think the whole expression can be moved to a variable since it’s the same three times

nerrad: > What would the benefits be of doing that?

As Nadir mentioned, biggest benefit is code auto-completion and preventing typos. This will be useful when we add test coverage for this as well because we can use the constants (again for auto-complete and to remove typos).

> Is this like the Action Names in wp.data being constants? Also not sure of the benefits of that either since the constant names are identical to the contents.

In the data stores the constants are referenced in the action creators and again in the reducers. So having the constants helps with code auto-completion and to guard against typos.

Issue #1882, opened 2020-03-05
Disable "Continue to checkout" button if any resource is currently resolving.


senadir:
Part of optimistic UI is to directly load results to the user even if the server didn’t respond yet, this is meant to allow him to continue using the cart while things load in the background and to make an experience seem fast, however, every action will update parts of the cart like totals and shipping and prices, "continue to checkout" is the final step of this formula and we want to make sure servers responded before producing, so this button should be disabled until everything has resolved.

I'm not yet aware of a catch-all `isResolving` selector so we might need to construct it from various actions `isResolving`s

haszari:
Is is this necessary? Just thinking about this from the user point of view...

- Does this mean proceed button will be disabled if user makes a change (e.g. add coupon)? 
- Is it possible the button would be disabled while some content loads initially, or is this covered by server-side hydrate?
- What if users click some other link to checkout – e.g. checkout page link in header/menu - what would be broken? 
- If there is an issue, should we prevent navigating away from cart page? Is it that serious?

I just want to better understand how this could cause an issue/problem for users. 

Here's an example flow: 

1. User adds something to cart.
2. User clicks through to view cart. Something is loading .. shipping rates?
3. User doesn't care about the information presented in the cart – e.g. prices, shipping etc – they just want to complete purchase (NOW!), so they click through to checkout immediately.

Put another way, do we need to enforce that users see cart info and engage with the cart in a certain way, or do we let them proceed to checkout whenever they want?

cc @nerrad  @garymurray

nerrad:
Button will only be disabled if shipping rates or quantity changes result in total updates after the cart is loaded. So if the user decides to just click through this will not be blocking.  

@haszari did you take a look at the conversation in any of the linked pulls before leading your comment? I just want to ensure you have full awareness of the discussions that have happened there.

haszari:
> Button will only be disabled if shipping rates or quantity changes result in total updates after the cart is loaded. So if the user decides to just click through this will not be blocking.

Right - similarly, if user updates quantity or adds a coupon, do we care if they click through to checkout via menu? Or navigate away? I'm trying to understand how the UI will break. 

I agree that usually it's best to indicate to the user in some way that their changes are persisting, and they need to wait / not navigate away - just exploring the implications of this for a store.

> did you take a look at the conversation in any of the linked pulls before leading your comment? I just want to ensure you have full awareness of the discussions that have happened there.

[...]

PR #2083, opened 2020-03-31
Refactor checkout context provider to be more flexible and implement CartProvider

nerrad:
Contributes towards: #1882

The plan I had in mind when creating the `CheckoutProvider` was that the logic from the different providers could be re-used for Cart as well. The idea being that we could have a`CartProvider` that wraps the cart and exposing similar interface for the cart as that exposed to the checkout. This would avoid duplication in code since Cart and Checkout share so many components and logic.

The need for this became more apparent with @senadir 's work on #1882 (see #1978 and the discussion there).

So in this pull the following is implemented:

- The interface in `CheckoutProvider` was extracted and moved to `CheckoutStateProvider` (not settled on the name but I've left it for now, suggestions welcome).
- `CheckoutProvider` now implements `CheckoutStateProvider` and the various providers wrapping the checkout block components.
- `CartProvider` has been created that implements `CheckoutStateProvider` and _just_ the `ShippingDataProvider` since that's all Cart needs currently.
- `useCheckoutContext` exposes the interface from `CheckoutStateProvider` so it can be used in any component that is implemented in either the cart or block. However, the name will likely be confusing so ideas are welcome for renaming this to something more generic for use in components implemented by checkout and/or cart.
- I cleaned up some type-defs (not exhaustive, just a few).
- We have `EditorProvider` for exposing the `isEditor` boolean to components, so I dropped `isEditor` flag from checkout context and refactored all implementations of `isEditor` coming from `useCheckoutContext` to be pulled from `useEditorContext` instead.

With this pull merged, we can continue with switching hook implementations in cart and checkout components to use the relevant context hooks instead.  @senadir should also be able to finish off #1882.

## To test

This impacts both cart and checkout blocks (mostly checkout). So testing should involve:

- verify frontend and editor instance of both blocks work as expected (with no console errors).
- verify inserting either block in a new block-editor instance inserts the block as expected and there are no errors in the console.


assets/js/base/context/cart-checkout/checkout-state/index.js
```diff
@@ -0,0 +1,203 @@
...
+ * @param {Object}  props.children            The children being wrapped.
+ * @param {string}  props.redirectUrl         Initialize what the checkout will
+ *                                            redirect to after successful
+ *                                            submit.
+ * @param {string}  props.submitLabel         What will be used for the checkout
+ *                                            submit button label.
+ * @param {boolean} props.isCart              If context provider is being used
+ *                                            in cart context.
+ */
+export const CheckoutStateProvider = ( {
```
haszari, line 80: We could maybe call this CheckoutOrderProvider? Most of the interfaces are about preparing data for an order or the order itself  .. not sure if this is a better name.

assets/js/base/context/cart-checkout/checkout-state/index.js
```diff
@@ -0,0 +1,203 @@
...
+ * cart or checkout blocks.
+ *
+ * @param {Object}  props                     Incoming props for the provider.
+ * @param {Object}  props.children            The children being wrapped.
+ * @param {string}  props.redirectUrl         Initialize what the checkout will
+ *                                            redirect to after successful
+ *                                            submit.
+ * @param {string}  props.submitLabel         What will be used for the checkout
+ *                                            submit button label.
+ * @param {boolean} props.isCart              If context provider is being used
```
haszari, line 77: How does `isCart` get used inside the provider/context? This feels a little suspect, maybe just the name. If it's to disable certain things, can we expose options for disabling them?

assets/js/base/context/editor/index.js
```diff
@@ -23,11 +23,14 @@ export const useEditorContext = () => {
 /**
  * Editor provider
  *
- * @param {Object}  props              Incoming props for the provider.
- * @param {*}       props.children     The children being wrapped.
- * @param {number}  props.currentPostId The post being edited.
+ * @param {Object}  props                 Incoming props for the provider.
+ * @param {*}       props.children        The children being wrapped.
+ * @param {number}  [props.currentPostId] The post being edited.
```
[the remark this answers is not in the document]
nerrad: Means the argument/property is optional.

assets/js/base/context/cart-checkout/checkout-state/index.js
nerrad: Hmm, ya there _is_ order data exposed on it (if available) but that's not the primary purpose of this provider.

assets/js/base/context/cart-checkout/checkout-state/index.js
nerrad: > How does isCart get used inside the provider/context? 

`CartProvider` and `CheckoutProvider` are components that implement this particular provider along with any other providers they need to use. The challenge here is that children components within the providers _may_ need to know what context they are in (cart or checkout) so this property provides an easy way to expose that context. There are no implementations of this flag yet but one implementation where I could see this being useful is for customizing css classes on shared components.

>  If it's to disable certain things, can we expose options for disabling them?

Sorry I don't know what you mean by "expose options for disabling them".

haszari: Ah so to be clear, this is not currently used?

assets/js/base/context/editor/index.js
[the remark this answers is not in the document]
haszari: TIL :)

assets/js/base/context/cart-checkout/checkout-state/index.js
nerrad: No.

haszari: I see! Usually I'd recommend not adding this until it's needed / used. No biggie since is already merged.

nerrad: Just to followup, there actually [is a place where this can be implemented now](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2101#discussion_r402440162).  _I thought I had seen this kind of flag in use somewhere_ :)

Issue #2063, opened 2020-03-30
Clean up main payment extension files (abstract code into their own files for readability)


nerrad:
The line count in the existing main payment extension component file is fairly large and it makes it harder to follow the logic. So this issue is for:

- refactoring logic into encapsulated functions/hooks for easier readability.
- move code (when/where it makes sense) into separate files.
- add appropriate doc blocks.

PR #2089, opened 2020-04-01
Refactor: split up payment-method.js for stripe extension into smaller files.

nerrad:
Fixes: #2063 

In this pull, the stripe client extension has been reorganized into smaller files.  This helps with:

- reviewing code (less lines to look over)
- establishing what are dependencies.
- easier maintenance.

As a part of this work I also added more internal documentation to describe the various components and custom hooks used by the stripe client and I fixed some bugs noticed in the `CardElements` component.

## To Test

This impacts the stripe payment method extension. You'll need the Stripe Gateway plugin installed, active, and configured in order to test.

- Verify the stripe payment option shows on the frontend and you can interact with the fields and validation errors work as expected for the fields (i.e. leave a field only partially filled).
- Verify the stripe payment option shows in the editor as expected (currently just placeholder text).

There should be no console errors when testing the above.

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/elements.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/elements.js
@@ -1,0 +1,144 @@
+/**
+ * External dependencies
+ */
+import { useState } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import {
+	CardElement,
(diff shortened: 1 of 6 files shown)


nerrad:
I'm amazed the bundle size gh action is reporting no size change! That seems, odd.

PR #2090, opened 2020-04-01
Add Payment Methods Registry for registering scripts and data

### LINKED ISSUE #2067: Implement php api interface for payment methods data from server to client
Currently payment methods would pass through and consume their own necessary data from server to client (api key, js initialization settings etc). This issue is for implementing a php api interface for payment methods to register data on the server to pass through to the payment registration client side. 

The goal here is for payment methods to have a consistent interface for passing along (and implementing) that data eliminating duplication across all the payment method extensions.

mikejolley:
Adds a payment method registry so payment integrations can use a common interface. Modelled like Block registration.

Payment classes on the PHP side register like this:

```php
PaymentMethodRegistry::get_instance()->register(
	$this->payment_method_name,
	array(
		'scripts' => $this->get_payment_method_scripts(),
		'data'    => $this->get_payment_method_script_data(),
	)
);
```

`data` contains the key value pairs to add to the settings array. The key of the data is set to `GATEWAY_data`.

`scripts` are payment scripts/handles which get passed to the Assets class for registration.

The `PaymentMethodAssets` class takes registered payment methods and enqueues their scripts and data when needed. They are enqueued in this case when the checkout is rendered.

**Additional:**
- Using this new system, I also added a **Cheque** gateway which does little processing, just processes orders.
- Each payment method integration needs its own handle so I updated webpack config to split them (stripe and cheque)
- I refactored some of the apple pay/stripe code and directory structure to better organise things

Fixes #2067

Build then go to checkout and confirm you see Stripe (if setup) and cheque. You may need to run `composer dump-autoload` first.


nerrad:
We're going to have to test the production builds on this branch because bundlesize is not reporting the new bundles. That suggests they aren't getting built.

src/Assets/PaymentMethodAssets.php
```diff
@@ -0,0 +1,77 @@
...
+		$payment_methods = $this->payment_method_registry->get_all_registered();
+
+		foreach ( $payment_methods as $payment_method ) {
+			$script_handles = array_merge(
+				$script_handles,
+				is_admin() ? $payment_method->get_payment_method_script_handles_for_admin() : $payment_method->get_payment_method_script_handles()
+			);
+		}
+
+		return array_unique( array_filter( $script_handles ) );
```
nerrad, line 60: Good call on simplifying things here . I get why `array_unique`, but why `array_filter`?  Is it used to guard against non truthy array values that might have been provided by the payment method?

src/BlockTypes/Checkout.php
```diff
@@ -100,6 +98,21 @@ protected function enqueue_data( array $attributes = [] ) {
 			$this->hydrate_from_api( $data_registry );
 			$this->hydrate_customer_payment_methods( $data_registry );
 		}
+
+		Package::container()->get( PaymentMethodAssets::class )->register_payment_method_data();
```
nerrad, line 102: Should we still keep an action here for extensions to have a place to enqueue additional data for the block? Also, could `PaymentMethodAssets` just hook _into_ the action?

nerrad: Actually we'll **need to keep** the old action name for checkout and cart blocks (`woocommerce_blocks_enqueue_checkout_block_scripts_after`) because @haszari merged a [pull in jetpack](https://github.com/Automattic/jetpack/pull/15223) that utilizes the hook for tracks.  We can keep the new hook name for the `enqueue_data` methods, but we'll need to keep the old hook name in `enqueue_scripts`.

src/BlockTypes/Checkout.php
```diff
@@ -100,6 +98,21 @@ protected function enqueue_data( array $attributes = [] ) {
...
+	 * Register/enqueue scripts used for this block.
+	 *
+	 * @param array $attributes  Any attributes that currently are available from the block.
+	 *                           Note, this will be empty in the editor context when the block is
+	 *                           not in the post content on editor load.
+	 */
+	protected function enqueue_scripts( array $attributes = [] ) {
+		$dependencies = Package::container()->get( PaymentMethodAssets::class )->get_all_registered_payment_method_script_handles();
+
+		Assets::register_block_script( $this->block_name . '-frontend', $this->block_name . '-block-frontend', $dependencies );
```
nerrad, line 115: This is where we'll need to put back in `woocommerce_blocks_enqueue_checkout_block_scripts_after` so the tracking hooked in on jetpack recently isn't broken.

src/BlockTypes/Cart.php
```diff
@@ -65,33 +62,70 @@ public function render( $attributes = array(), $content = '' ) {
...
+	 * Register/enqueue scripts used for this block.
+	 *
+	 * @param array $attributes  Any attributes that currently are available from the block.
+	 *                           Note, this will be empty in the editor context when the block is
+	 *                           not in the post content on editor load.
+	 */
+	protected function enqueue_scripts( array $attributes = [] ) {
+		Assets::register_block_script( $this->block_name . '-frontend', $this->block_name . '-block-frontend' );
+
+		do_action( 'woocommerce_blocks_cart_enqueue_scripts' );
```
nerrad, line 118: We'll need to rename this back to `woocommerce_blocks_enqueue_cart_block_scripts_after` because there is tracking that was hooked into it I think.

src/Library.php
```diff
@@ -137,6 +139,13 @@ public static function register_blocks() {
 		}
 	}
 
+	/**
+	 * Register payment methods.
+	 */
+	public static function register_payment_methods() {
+		$payment_method_registry = Package::container()->get( PaymentMethodRegistry::class )->initialize();
```
nerrad, line 146: Don't need to assign this to a variable right?

src/Assets/PaymentMethodAssets.php
mikejolley: yeah; this is a shortcut to remove empty values too.

src/BlockTypes/Checkout.php
mikejolley: lol we're breaking compatibility before we even ship it 🚢 I hate hooks.

nerrad: > I hate hooks.

I share your sentiment.

src/BlockTypes/Checkout.php
mikejolley: We shouldn't need to because extensions _could_ interact with the registry themselves. I think we should keep this contained within the API we're providing tbh

PR #2091, opened 2020-04-01
Tweaks to #2090

nerrad:
@mikejolley, #2090 is good, but I have some suggested tweaks outlined in this pull (with in line comments) that were too hard to outline in the review of #2090, so I decided to put up this pull and we can discuss the proposed changes a bit easier inline here.

The major change in this pull from yours is eliminating a layer (`PaymentMethodType`) and child payment method type classes will take care of registering their assets and just exposing the handles. So our code still takes care of enqueueing assets and data correctly but doesn't worry about registration.


src/Assets/PaymentMethodAssets.php
```diff
@@ -30,24 +30,15 @@ class PaymentMethodAssets {
...
-
 	/**
 	 * Constructor for class
 	 *
 	 * @param PaymentMethodRegistry $payment_method_registry An instance of Payment Method Registry.
 	 * @param AssetDataRegistry     $asset_registry  Used for registering data to pass along to the request.
-	 * @param Api                   $asset_api       Used for registering scripts and styles.
 	 */
-	public function __construct( PaymentMethodRegistry $payment_method_registry, AssetDataRegistry $asset_registry, Api $asset_api ) {
+	public function __construct( PaymentMethodRegistry $payment_method_registry, AssetDataRegistry $asset_registry ) {
```
nerrad, line 39: Removes the Assets/Api class because this is something that is coupled to the plugin path for blocks. So it's interfaces will not work for extensions registering payment method types.

src/Assets/PaymentMethodAssets.php
```diff
@@ -60,38 +51,45 @@ protected function init() {
...
 	}
 
+	/**
+	 * Register payment method styles for usage.
+	 */
+	public function register_payment_method_styles() {
+		// @todo not sure where/how we should enqueue these yet. They can't
+		// be enqueued in the block render function.
+	}
+
```
nerrad, line 81: We'll have to decide how we'll enqueue styles for the payment methods extensions (if they have any). We _might_ be able to use the `enqueue_block_assets` action?

src/Domain/Bootstrap.php
```diff
@@ -183,4 +187,44 @@ protected function define_feature_flag() {
...
+				if ( class_exists( 'WC_Stripe' ) && ! $payment_method_registry->is_registered( 'stripe' ) ) {
+					$payment_method_registry->register(
+						$this->container->get( Stripe::class )
+					);
+				}
+				$payment_method_registry->register(
+					$this->container->get( Cheque::class )
+				);
+			}
+		);
```
nerrad, line 227: For demonstration purposes, I'm hooking into the new action, however for built-ins we can probably just pull the `$payment_method_registry` directly from the container and register payment built-in payment method types that way.

src/Library.php
```diff
@@ -143,15 +143,9 @@ public static function register_blocks() {
...
-			'Cheque',
-		];
-		foreach ( $integrations as $class ) {
-			$class    = __NAMESPACE__ . '\\Payments\\Integrations\\' . $class;
-			$instance = new $class();
-			$instance->register_payment_method();
-		}
+		$payment_method_registry = Package::container()->get( PaymentMethodRegistry::class );
+		do_action( 'wc_blocks_library_payment_method_type_registration', $payment_method_registry );
+		$payment_method_registry->initialize();
```
nerrad, line 148: This is the mechanism by which payment method extensions will register their types.  They hook into the action and then receive an reference to the `$payment_method_registry`  from which they can register their instantiated classes (extending `AbstractPaymentMethodType` or implementing `PaymentMethodTypeInterface`).

src/Payments/Integrations/AbstractPaymentMethodType.php
```diff
@@ -0,0 +1,97 @@
...
+defined( 'ABSPATH' ) || exit;
+
+use Automattic\WooCommerce\Blocks\Payments\PaymentMethodTypeInterface;
+
+/**
+ * AbstractPaymentMethodType class.
+ *
+ * @since $VID:$
+ */
+abstract class AbstractPaymentMethodType implements PaymentMethodTypeInterface {
```
nerrad, line 19: You may wonder why have both an interface and an AbstractPaymentMethodType?  The primary reason is flexibility for how payment method extensions set things up. We _may_ add helpers or extra things to the AbstractPaymentMethodType, but the interface defines the contract needed for the api. So payment method extensions can just implement the interface if they want to do their own thing.

src/Payments/Integrations/Cheque.php
```diff
@@ -13,56 +13,65 @@
...
+	 * Constructor
+	 *
+	 * @param string $name The name of the payment method type.
+	 * @param Api    $asset_api An instance of Api.
 	 */
-	public function register_payment_method() {
+	public function __construct( $name, Api $asset_api ) {
+		$this->asset_api = $asset_api;
+		parent::__construct( $name );
+	}
```
nerrad, line 47: We have a constructor for Stripe and Cheque payment method types here that receives the Api, because _locally_ we can use that for registering the assets. Payment method extensions won't need a constructor if they extend `AbstractPaymentMethodType`.

src/Payments/Integrations/Cheque.php
```diff
@@ -13,56 +13,65 @@
...
+
+	/**
+	 * Initializes the payment method type.
+	 */
+	public function initialize() {
```
nerrad, line 52: We still delay initialization and hitting options until needed. This probably could even be delayed until sometime in `PaymentAssets` (instead of `Library`)?

src/Payments/Integrations/Cheque.php
```diff
@@ -13,56 +13,65 @@
...
-				'src'    => 'build/wc-payment-method-cheque.js',
-				'deps'   => [],
-			],
-		];
+	public function get_payment_method_script_handles() {
+		$this->asset_api->register_script(
+			'wc-payment-method-cheque',
+			'build/wc-payment-method-cheque.js'
+		);
+		return [ 'wc-payment-method-cheque' ];
```
nerrad, line 66: As you can see, we leave script/style _registration_ up to the payment methods. Implementing an interface for that in our code seems like it'd be just duplicating what `wp_register_script` already does. So let's just hand that off to payment methods to deal with. All we care about are handles for registered assets.

src/Payments/PaymentMethodRegistry.php
```diff
@@ -17,60 +17,46 @@ final class PaymentMethodRegistry {
...
+	public function register( PaymentMethodTypeInterface $payment_method_type ) {
+		$name = $payment_method_type->get_name();
 		if ( $this->is_registered( $name ) ) {
 			/* translators: %s: Payment method name. */
 			_doing_it_wrong( __METHOD__, esc_html( sprintf( __( 'Payment method "%s" is already registered.', 'woo-gutenberg-products-block' ), $name ) ), '2.5.0' );
 			return false;
 		}
+		$this->registered_payment_methods[ $name ] = $payment_method_type;
+		return true;
+	}
```
nerrad, line 40: By typecasting the expected value here, we are providing a strong contract which eliminates a lot of conditionals and code paths. I think this simplifies things somewhat.

src/Payments/PaymentMethodRegistry.php
```diff
@@ -17,60 +17,46 @@ final class PaymentMethodRegistry {
...
-		if ( ! $payment_method ) {
-			$payment_method = new PaymentMethodType( $name, $args );
+	/**
+	 * Initializes all payment method types.
+	 */
+	public function initialize() {
+		$registered_payment_method_types = $this->get_all_registered();
+		foreach ( $registered_payment_method_types as $registered_type ) {
+			$registered_type->initialize();
 		}
```
nerrad, line 49: One initialization method to handle initializing all registered payment methods.

src/Payments/PaymentMethodRegistry.php
```diff
@@ -118,19 +104,4 @@ public function get_all_registered() {
...
-	 *
-	 * @return PaymentMethodRegistry The main instance.
-	 */
-	public static function get_instance() {
-		if ( null === self::$instance ) {
-			self::$instance = new self();
-		}
-
-		return self::$instance;
-	}
```
nerrad, line 135: Now we can get rid of the singleton!

src/Assets/PaymentMethodAssets.php
mikejolley: I think styles need to be global due to needing including in header so we don't need to enqueue them for extensions.

Thats why its important we have default styles in place for common elements that gateways can utilise where possible.

PR #2093, opened 2020-04-02
fix nonce issue when logged out user adds to cart from all products

### LINKED ISSUE #2092: nonce error when add-to-cart from All Products block with anon/logged out user
To reproduce:

- Clone master branch (this is a regression on master, not reproducible in current release)
- Add all products block to a page
- View page in incognito window
- Add two different products to cart

Should get a nonce error when adding the second product.

haszari:
Fixes #2092

This PR switches collections store action `__experimentalPersistItemToCollection` to using `apiFetchWithHeaders` (from shared controls). This supplies and updates the Store API nonce in the store, so subsequent add-to-cart requests have a valid nonce.

1. Set up a page with All Products block (if needed).
2. View the page in an incognto window.
3. Add multiple products to cart - should work correctly.

Optional: also test adding to cart when logged in as a regular customer, and other add to cart flows. 

This affects the ["atomic" product button component](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/master/assets/js/atomic/components/product/button), so if this can be used in places other than All Products block, test those flows too :)

I've added `skip-changelog` as I don't think bug affects release versions? If this needs changelog reviewers please update :)

### DIFF
--- a/assets/js/data/collections/actions.js
+++ b/assets/js/data/collections/actions.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { select } from '@wordpress/data-controls';
+
 /**
  * Internal dependencies
  */
--- a/assets/js/data/collections/actions.js
+++ b/assets/js/data/collections/actions.js
@@ -1,11 +1,11 @@
 /**
  * External dependencies
  */
-import { apiFetch, select } from '@wordpress/data-controls';
-
+import { select } from '@wordpress/data-controls';
 /**
  * Internal dependencies
  */
+import { apiFetchWithHeaders } from '../shared-controls';
 import { ACTION_TYPES as types } from './action-types';
 import { STORE_KEY as SCHEMA_STORE_KEY } from '../schema/constants';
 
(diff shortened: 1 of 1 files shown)


haszari:
Saw a different issue when testing this, logged here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2094

mikejolley:
> I'm getting Cookie nonce is invalid when adding multiple products:

I think thats because this is adding the nonce for the request, but not updating the nonce from the response.

mikejolley:
This overlaps with https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1984/files I started a few weeks ago; let me try merging them and see if it fixes things :)

mikejolley:
Related to logged out users https://github.com/woocommerce/woocommerce/pull/25971

Nonces can change after sessions are established.

haszari:
Marking this as `Blocked` for now - I'm betting on #2103, but leaving this open until #2103 is wrapped up and merged.

PR #2101, opened 2020-04-02
Respect needs_shipping in Cart and Checkout

senadir:
This PR builds on the previously introduced `needs_shipping` and `needsShipping` data, it toggles off some shipping sections from Cart and Checkout when the current cart doesn't need shipping, the value of needs_shipping is based on `WC_Cart::needs_shipping()`.
Some code that was previously checked only against `SHIPPING_ENABLED` is now using 'needsShipping` as it can give false positive (cart only containing digital products) while `SHIPPING_ENABLED` watches the whole site state, in places when we used the constant, `needs_shipping` cover both cases so it became redundant (I left because we still use it in the editor to show certain settings if shipping is enabled).
`needs_shipping` only checks for if shipping zones exists.

1. Make sure shipping is enabled, add to your cart a digital product
2. shipping should not be present.
3. add a physical product
4. shipping should show up now.
5. if you don't have shipping zones, shipping fees and row won't show up.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/sidebar/index.js
+++ b/assets/js/blocks/cart-checkout/checkout/sidebar/index.js
@@ -10,6 +10,7 @@ import {
 	TotalsShippingItem,
 	TotalsTaxesItem,
 } from '@woocommerce/base-components/cart-checkout';
+import { useShippingDataContext } from '@woocommerce/base-context';
 import { getCurrencyFromPriceResponse } from '@woocommerce/base-utils';
 import {
 	COUPONS_ENABLED,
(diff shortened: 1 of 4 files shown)


assets/js/blocks/cart-checkout/checkout/sidebar/index.js
```diff
@@ -47,13 +50,15 @@ const CheckoutSidebar = ( {
...
-				noResultsMessage={ null }
-				isCheckout={ true }
-				showCalculator={ false }
-				values={ cartTotals }
-			/>
+			{ needsShipping && (
+				<TotalsShippingItem
+					currency={ totalsCurrency }
+					noResultsMessage={ null }
+					isCheckout={ true }
```
nerrad, line 57: Anyplace we have an `isCheckout` prop can now be removed as we have a `isCart` flag returned from the `useCheckoutContext` hook (which will automatically be true if in the `CartProvider` tree and false if in the `CheckoutProvider` tree). No need to handle in this pull though.

PR #2103, opened 2020-04-02
Update add to cart to useStoreCart hook

### LINKED ISSUE #2092: nonce error when add-to-cart from All Products block with anon/logged out user

mikejolley:
Reworked #1984 and I think this also covers the fix in #2093 due to using the apiFetchWithHeaders action.

Fixes #2092

Converts the add to cart functionality which was previously using `__experimentalPersistItemToCollection` to instead use the `useStoreCart` hook.

To support this, a new `cart/add-to-cart` endpoint has been added to the store API. I've reworked some of the routes/schemas to support this and ensure each route returns the correct schema.

**How to test the changes in this Pull Request:**
Copied from #2093:

- Set up a page with All Products block (if needed).
- View the page in an incognto window.
- Add multiple products to cart - should work correctly.
- Optional: also test adding to cart when logged in as a regular customer, and other add to cart flows.

@Aljullu do you get the same issue with this branch?

### DIFF
--- a/assets/js/atomic/components/product/button/index.js
+++ b/assets/js/atomic/components/product/button/index.js
@@ -4,19 +4,17 @@
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { _n, sprintf } from '@wordpress/i18n';
-import {
-	useMemo,
-	useCallback,
-	useState,
-	useEffect,
-	useRef,
-} from '@wordpress/element';
+import { useMemo, useState, useEffect, useRef } from '@wordpress/element';
 import { useDispatch } from '@wordpress/data';
 import { find } from 'lodash';
-import { useCollection } from '@woocommerce/base-hooks';
-import { COLLECTIONS_STORE_KEY as storeKey } from '@woocommerce/block-data';
+import { useStoreCart } from '@woocommerce/base-hooks';
+import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
 import { useProductLayoutContext } from '@woocommerce/base-context';
 import { decodeEntities } from '@wordpress/html-entities';
+
+/**
+ * @typedef {import('@woocommerce/type-defs/hooks').StoreCartItemAddToCart} StoreCartItemAddToCart
+ */
 
 /**
  * A custom hook for exposing cart related data for a given product id and an
@@ -28,51 +26,34 @@ import { decodeEntities } from '@wordpress/html-entities';
  * @param {number} productId  The product id for the product connection to the
  *                            cart.
  *
- * @return {Object} Returns an object with the following properties:
- *    @type {number}   cartQuantity  The quantity of the product currently in
- *                                   the cart.
- *    @type {boolean}  addingToCart  Whether the product is currently being
- *                                   added to the cart (true).
- *    @type {boolean}  cartIsLoading Whether the cart is being loaded.
- *    @type {Function} addToCart     An action dispatcher for adding a single
- *                                   quantity of the product to the cart.
- *                                   Receives no arguments, it operates on the
- *                                   current product.
+ * @return {StoreCartItemAddToCart} An object exposing data and actions relating
+ *                                  to add to cart functionality.
  */
 const useAddToCart = ( productId ) => {
-	const { results: cartResults, isLoading: cartIsLoading } = useCollection( {
-		namespace: '/wc/store',
-		resourceName: 'cart/items',
-	} );
-	const currentCartResults = useRef( null );
-	const { __experimentalPersistItemToCollection } = useDispatch( storeKey );
-	const cartQuantity = useMemo( () => {
-		const productItem = find( cartResults, { id: productId } );
-		return productItem ? productItem.quantity : 0;
-	}, [ cartResults, productId ] );
 	const [ addingToCart, setAddingToCart ] = useState( false );
-	const addToCart = useCallback( () => {
+	const { cartItems, cartIsLoading } = useStoreCart();
+	const currentCartResults = useRef( cartItems );
+	const { addItemToCart } = useDispatch( storeKey );
+
+	const addToCart = () => {
 		setAddingToCart( true );
-		// exclude this item from the cartResults for adding to the new
-		// collection (so it's updated correctly!)
-		const collection = cartResults.filter( ( cartItem ) => {
-			return cartItem.id !== productId;
-		} );
-		__experimentalPersistItemToCollection(
-			'/wc/store',
-			'cart/items',
-			collection,
-			{ id: productId, quantity: 1 }
-		);
-	}, [ productId, cartResults ] );
+		addItemToCart( productId );
+	};
+
+	const cartQuantity = useMemo( () => {
+		const productItem = find( cartItems, { id: productId } );
+		return productItem ? productItem.quantity : 0;
+	}, [ cartItems, productId ] );
+
 	useEffect( () => {
-		if ( currentCartResults.current !== cartResults ) {
+		if ( currentCartResults.current !== cartItems ) {
 			if ( addingToCart ) {
 				setAddingToCart( false );
 			}
-			currentCartResults.current = cartResults;
+			currentCartResults.current = cartItems;
 		}
-	}, [ cartResults, addingToCart ] );
+	}, [ cartItems, addingToCart ] );
+
 	return {
 		cartQuantity,
 		addingToCart,
--- a/assets/js/atomic/components/product/button/index.js
+++ b/assets/js/atomic/components/product/button/index.js
@@ -4,63 +4,10 @@
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { _n, sprintf } from '@wordpress/i18n';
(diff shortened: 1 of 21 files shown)


senadir:
just a side note, this is causing an alarming increase in bundle size for the all product
```
69.4 kB | +52.2 kB (75%)
```

src/RestApi/StoreApi/Routes/CartAddItem.php
```diff
@@ -0,0 +1,116 @@
...
+		return 'wc/store';
+	}
+
+	/**
+	 * Get the path of this REST route.
+	 *
+	 * @return string
+	 */
+	public function get_path() {
+		return '/cart/add-item';
```
haszari, line 33: Do we need to add this to the API docs? I'm getting a bit lost - there are a few routes related to cart items now.

assets/js/data/cart/actions.js
```diff
@@ -206,6 +206,34 @@ export function* removeCoupon( couponCode ) {
...
+ * @param {number} productId Product ID to add to cart.
+ * @param {number} quantity Number of product ID being added to cart.
+ */
+export function* addItemToCart( productId, quantity = 1 ) {
+	try {
+		const { response } = yield apiFetchWithHeaders( {
+			path: `/wc/store/cart/add-item`,
+			method: 'POST',
+			data: {
+				id: productId,
```
haszari, line 225: Should this API use a `cartItemKey` rather than `productId`, e.g. for consistency with the `remove-item` API?

assets/js/data/cart/actions.js
```diff
@@ -206,6 +206,34 @@ export function* removeCoupon( couponCode ) {
 	return true;
 }
 
+/**
+ * Adds an item to the cart:
+ * - Calls API to add item.
+ * - If successful, yields action to add item from store.
+ * - If error, yields action to store error.
+ * - Sets cart item as pending while API request is in progress.
```
haszari, line 214: This doesn't set the item as pending. Not sure if that's needed or not – either delete comment or add `itemQuantityPending()`s :)

assets/js/data/cart/actions.js
haszari: Doh, perhaps this is because we are making a new item, the new key will be returned :) (Ignore previous comment)

assets/js/data/cart/actions.js
haszari: Probably not needed if this is used for All Products block only - unless the All Products block and Cart block are on the same page … 🤔

assets/js/atomic/components/product/button/index.js
```diff
@@ -28,51 +26,34 @@ import { decodeEntities } from '@wordpress/html-entities';
...
- *                                   added to the cart (true).
- *    @type {boolean}  cartIsLoading Whether the cart is being loaded.
- *    @type {Function} addToCart     An action dispatcher for adding a single
- *                                   quantity of the product to the cart.
- *                                   Receives no arguments, it operates on the
- *                                   current product.
+ * @return {StoreCartItemAddToCart} An object exposing data and actions relating
+ *                                  to add to cart functionality.
  */
 const useAddToCart = ( productId ) => {
```
haszari, line 32: I was secretly hoping we could use an existing hook as part of fixing this – unfortunate having a custom hook just for this add-to-cart button.

haszari: Nice that this now leverages `useStoreCart` and has typedef; perhaps we can move this out of product button soon.

src/RestApi/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -56,18 +42,13 @@ public function get_properties() {
...
+				'readonly'    => true,
 			],
 			'quantity'            => [
 				'description' => __( 'Quantity of this item in the cart.', 'woo-gutenberg-products-block' ),
 				'type'        => 'integer',
 				'context'     => [ 'view', 'edit' ],
-				'arg_options' => [
-					'sanitize_callback' => 'wc_stock_amount',
-				],
+				'readonly'    => true,
```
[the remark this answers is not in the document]
mikejolley: These endpoints have no UPDATE or POST methods to update this data, so it's readonly. Same for the callbacks; if we're not accepting data, we don't need to sanitize anything.

src/RestApi/StoreApi/Routes/CartAddItem.php
```diff
@@ -0,0 +1,116 @@
...
+					],
+					'variation' => [
+						'description' => __( 'Chosen attributes (for variations).', 'woo-gutenberg-products-block' ),
+						'type'        => 'array',
+						'context'     => [ 'view', 'edit' ],
+						'items'       => [
+							'type'       => 'object',
+							'properties' => [
+								'attribute' => [
+									'description' => __( 'Variation attribute name.', 'woo-gutenberg-products-block' ),
```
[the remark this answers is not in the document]
mikejolley: This was moved from schema so matches - I think it is name (humans use APIs too). It's all handled in the cart controller.

mikejolley:
> We'll need to update the API docs – since we're not using a standard patterns (e.g. REST) we need to take care when designing & documenting the API. Can handle that as follow up.

Docs will be a follow up because so much has changed since initial revision. Not sure what you mean about it not being REST though :)

src/RestApi/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -369,8 +352,11 @@ public function get_prices( \WC_Product $product, $tax_display_mode = '', $qty =
 		$tax_display_mode = in_array( $tax_display_mode, [ 'incl', 'excl' ], true ) ? $tax_display_mode : get_option( 'woocommerce_tax_display_shop' );
 		$price_function   = 'incl' === $tax_display_mode ? 'wc_get_price_including_tax' : 'wc_get_price_excluding_tax';
 
-		// Get individual product prices from product schema.
-		$prices = $this->product_schema->get_prices( $product, $tax_display_mode );
+		// Get individual product prices.
+		$prices['price']         = $this->prepare_money_response( $price_function( $product ), wc_get_price_decimals() );
```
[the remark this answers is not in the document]
mikejolley: It's not completely the same because of the raw prices needed here. It didn't seem worth constructing the other class to just take a few lines of logic when this class already has the $product object and all other values.

src/RestApi/StoreApi/Schemas/CheckoutSchema.php
```diff
@@ -22,6 +22,31 @@ class CheckoutSchema extends AbstractSchema {
...
+	/**
+	 * Constructor.
+	 *
+	 * @param BillingAddressSchema  $billing_address_schema Billing address schema instance.
+	 * @param ShippingAddressSchema $shipping_address_schema Shipping address schema instance.
+	 */
+	public function __construct( BillingAddressSchema $billing_address_schema, ShippingAddressSchema $shipping_address_schema ) {
+		$this->billing_address_schema  = $billing_address_schema;
+		$this->shipping_address_schema = $shipping_address_schema;
+	}
```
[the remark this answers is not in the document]
mikejolley: Everytime I add endpoints I notice things that need revision 😅

src/RestApi/StoreApi/Routes/CartAddItem.php
mikejolley: All endpoints need docs when we're done, during cooldown.

src/RestApi/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -383,6 +369,47 @@ public function get_prices( \WC_Product $product, $tax_display_mode = '', $qty =
...
 	}
 
+	/**
+	 * Get price range from certain product types.
+	 *
+	 * @param \WC_Product $product Product instance.
+	 * @return object|null
+	 */
+	protected function get_price_range( \WC_Product $product ) {
+		$tax_display_mode = get_option( 'woocommerce_tax_display_shop' );
```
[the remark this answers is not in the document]
mikejolley: Added inheritance to solve.

assets/js/atomic/components/product/button/index.js
mikejolley: Done and done :)

mikejolley:
This PR has been updated based on feedback.

@haszari The cart issue you mentioned I cannot reproduce now.

The multiple-add-to-carts however is not something new in this PR afaik. It's a session issue/race condition. The only way to prevent that would be to disable all add to cart buttons whilst requests are in progress, or queue them somehow so they run in sequence. I think thats how core does it. 

cc'ing @woocommerce/merch-squad for ideas on a queue for follow up? It doesn't seem blocking.

mikejolley:
@haszari we had that pattern initially :) It was changed as you know to allow carts to be returned in the response.

PR #2106, opened 2020-04-02
Disable continue to checkout if Item quantity is being updated and immediately remove items.

### LINKED ISSUE #2126: Improve UI signal that a cart item is being removed.
Currently, when we delete a cart item, we block the item and wait until the server response for it be deleted, I think it would be better if we directly delete it from local state, the server response will confirm this by having it deleted as well, if it failed being deleted for some reason, the item will pop up again.

senadir:
### Edit 3:
This PR add checkout status watching to `useStoreCartItemQuantity` so that we can disable continue to checkout if an item quantity is being updated, I also plan to extend this to also watching when an element is being removed (waiting for the server to respond, possibly adding a flag and updating from the global context), it also adds a minor refactor
it renames `isPending` flag and all related actions/selectors/action types to `isPendingQuantity` and update related docs, it seems this flag was introduced first when we didn’t change quantity locally, now it reflects its actual state.

This also fixes #2126 being some unused actions and directly removing items, however, it does not signal to checkout if an item is being deleted.
I tried using a flag for removing quantity as well but the problem is once you delete the item, the hook is also deleted and all related code, so the useEffect is never run again, what I did is that I put the logic in the `useEffect` return function, but it ended up incrementing and but never decrementing so we might want to watch somewhere else.

### How to test:
1- In Cart, try removing an item, it will immediately delete, cart total will update after the server respond.
2- Try updating quantity, items will work fine and will never get blocked, the button will get blocked when we dispatch and will get released once we get a respond.


assets/js/data/cart/actions.js
```diff
@@ -85,17 +85,17 @@ export function receiveCartItem( response = {} ) {
...
+ * is being removing.
  *
- * @param {string} cartItemKey Cart item being updated.
- * @param {boolean} isQuantityPending Flag for update state; true if API request is pending.
+ * @param {boolean} isPendingDelete Flag for update state; true if API request is pending.
+ * @param {string} cartItemKey Cart item being deleted.
  * @return {Object} Object for action.
  */
-export function itemQuantityPending( cartItemKey, isQuantityPending ) {
+export function itemPendingDelete( isPendingDelete, cartItemKey ) {
```
senadir, line 94: Switching the order of this function is a personal preference, in my opinion, flag function or flag parameters (`shouldCondition` or `shouldAction`) should have either:
1- the flag as the first parameter.
2- the param as an object and the flag as a named value in it, so that you know what is true or not, example:
```js
deleteStatus( { isDeleting: true } )
```
or
```js
isDeletingItem( true, itemKey )
```

senadir:
Updated tests by deleting unneeded test for `isPendingQuantity` since we no longer expose it on the hook.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -82,12 +82,9 @@ const CartLineItemRow = ( { lineItem } ) => {
...
 
-	const {
-		quantity,
-		changeQuantity,
-		removeItem,
-		isPending: itemQuantityDisabled,
-	} = useStoreCartItemQuantity( lineItem );
+	const { quantity, changeQuantity, removeItem } = useStoreCartItemQuantity(
+		lineItem
+	);
```
nerrad, line 87: what's the rationale for no longer exposing an `isPending` property? Do we not want to disable the quantity input?

assets/js/data/cart/actions.js
```diff
@@ -85,17 +85,17 @@ export function receiveCartItem( response = {} ) {
...
- * is being updated; i.e. removing, or changing quantity.
+ * quantity is being updated.
  *
+ * @param {boolean} isPendingQuantity Flag for update state; true if API request is pending.
  * @param {string} cartItemKey Cart item being updated.
- * @param {boolean} isQuantityPending Flag for update state; true if API request is pending.
  * @return {Object} Object for action.
  */
-export function itemQuantityPending( cartItemKey, isQuantityPending ) {
+export function itemPendingQuantity( isPendingQuantity, cartItemKey ) {
```
nerrad, line 94: Why change the order of the arguments?

In order for anything to happen we _require_ a cartItemKey otherwise the reducer has no clue what item to apply the "pending" state to. Convention is to put _required_ arguments first and optional arguments last (yes, I consider `isPendingQuantity` optional, and we could default it to a sensible default here - maybe true?)

So please return this to it's original order.

senadir:
> You created 2126 and didn't leave any room for discussing whether we even want to do that before fixing it in this pull. So please remove this additional logic from this pull so we can handle it separately.

yes handling that was:
- a proof of concept to test this behavior before merging (Hence why I pinged you about that point).
- it was straightforward, all the logic was already there, I simply called it.

assets/js/data/cart/actions.js
nerrad: I didn't see that you had left this comment in my original review (because the comment is on outdated code).  Just want to note, that I don't think personal preference should play a role here (you can read my inline review comment for more details).

nerrad:
> a proof of concept to test this behavior before merging (Hence why I pinged you about that point).

If you're referring to what you pinged about in slack, I thought it had more to do with the issue of decrementing loading when an item is removed... not that you wanted to change how items are removed altogether (which is a different issue entirely imo).  If this was a proof of concept you were looking for more feedback on before committing to in this pull, it would have helped to make that clear in the pull description. Even just writing something like "I also tried experimenting with fixing #2126 in this pull but can remove and do separately if needed".

assets/js/data/cart/actions.js
senadir: I did consider both of them required, my initial problem is that if you have a flag parameter, it should be named or it should be the first  (and only flag) parameter, because we end up risking with functions like in which we have no idea what that flag means.
```js
apiRequest(
  'products',
  'GET',
  { category: 3 },
  false,
  null,
  true
)
```

granted this is not the case here and might never be the case, but possibly we should enforce a certain default in the future (if the function has 3 parameters, it should be turned to a deconstructed object, as in, named parameters).

> So please return this to it's original order.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
senadir: `isPending` is now `isPendingQuantity` and it tells us if the quantity update has been dispatched or not, we don’t need that.
previously `isPending` was actually `isPendingDelete` but since we're now deleting them immediately, it is not needed and was not implemented.

assets/js/data/cart/actions.js
senadir: That’s why I commented on it here so that we can discuss it, personal preferences should be discussed before making them a general rule since we all share this code.

assets/js/data/cart/actions.js
nerrad: >  my initial problem is that if you have a flag parameter, it should be named or it should be the first (and only flag) parameter, because we end up risking with functions like in which we have no idea what that flag means.

In general I think if there's multiple arguments having different flag parameters there's a larger problem that should be dealt with (and having a defined object as the argument with named arguments is probably the better way to go).  So as you noted, the argument doesn't really apply here.

I still stand by my original comment that `cartItemKey` is more important than the flag argument and should be first.  I also would suggest that some improvements could be made to the action creator name to make it a bit clearer it is actually setting a flag for the item - maybe `itemIsPendingQuantity( cartItemKey, isPendingQuantity = true )`

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
nerrad: So there's a few things that have happened here then.

- the logic for tracking and acting on whether an item is _being removed_ (the former pending quantity logic) has been deleted.
- this pull _adds_ (via repurposing existing logic) for tracking quantity changes and using that internally in the hook to increment/decrement the calculating count.
- this pull implemented removing items instantly from the ui instead of waiting for the server response that it was successfully removed which negated the need to disable quantity selection field on removal.

Is that an accurate description of what is happening in this pull?  If so, I think you could do either of the following:

-  finish #2126 which takes care of the item removing bit (and can disable any inputs along with applying opacity while an item is being removed).  You'll still need a flag for the pending removal resolution state (and you can apply the renames etc in the pull for that issue).
- once that is done, you can do this pull which will _add_ pending resolution state for quantities that you are able to use for tracking increment/decrementing the calculating counts.

OR

- implement resolution state tracking for pending removal in this pull (along with the new pending quantity resolution state) and use the pending removal resolution state to disable quantity select fields (so there's no regression with existing work).
- work on #2126 after this pull gets merged to add opacity on the removal of items (using the new resolution state added here).

If it was up to me, I'd probably take the second approach.

senadir: I'm working on the second already and will push it now (applying styles to the item while loading).

assets/js/data/cart/actions.js
senadir: this is fixed in 0aaf2801e2bc96ba0aa3bccb981e49780bce03b9

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
senadir: this is fixed in a7ba54c45e9bd3807409448120e0e9890457ace5

senadir: I went with not disabling them using that flag, but by using `pointer-events: none` on the parent.

assets/js/base/hooks/cart/use-store-cart-item-quantity.js
```diff
@@ -62,9 +77,25 @@ export const useStoreCartItemQuantity = ( cartItem ) => {
...
+			dispatchActions.decrementCalculating();
+		}
+	}, [ isPendingQuantity ] );
+
+	useEffect( () => {
+		if ( isPendingDelete ) {
+			dispatchActions.incrementCalculating();
+		}
+		return () => {
+			dispatchActions.decrementCalculating();
```
senadir, line 93: running this directly in the return because the hook might unmount before `isPendingQuantity` change, so on unmount we're sure the element has been deleted.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -102,7 +103,11 @@ const CartLineItemRow = ( { lineItem } ) => {
 	const firstImage = images.length ? images[ 0 ] : {};
 
 	return (
-		<tr className="wc-block-cart-items__row">
+		<tr
+			className={ classnames( 'wc-block-cart-items__row', {
```
senadir, line 107: I would have gone with `LoadingMask` but this is a table row and we can’t use divs.

assets/js/blocks/cart-checkout/cart/full-cart/style.scss
```diff
@@ -91,6 +91,12 @@ table.wc-block-cart-items {
 				display: block;
 			}
 		}
+
+		&.is-loading {
+			opacity: 0.5;
+			pointer-events: none;
```
senadir, line 97: should disable all interactions with elements.

assets/js/base/hooks/cart/use-store-cart-item-quantity.js
nerrad: Yup this makes sense, the only problem is if the component _doesn't_ unmount because there was an error with the server removing. So you probably will need to account for whether there was a server error and make sure to decrement the count if `! isPendingDelete` and `whateverIsUsedToIndicateError` in the main body of the callback too.

Also `decrementCalculating` should only happen in the returned callback if `! isPendingDelete` otherwise if the component is unmounted for some other reason, the count could decrement unnecessarily.

senadir:
> Also decrementCalculating should only happen in the returned callback if ! isPendingDelete otherwise if the component is unmounted for some other reason, the count could decrement unnecessarily.

The point of having the decrement in the callback is that because `isPendingDelete` will never switch back from `true` to `false` since it will be deleted before that switch, wrapping it in `isPendingDelete` as true does however work since it ensures that item was being deleted and is now deleted.

> if the component doesn't unmount because there was an error with the server removing. So you probably will need to account for whether there was a server error and make sure to decrement the count if ! isPendingDelete and whateverIsUsedToIndicateError in the main body of the callback too.

Will do, however we don't currently show such errors in the frontend, but I implemented that handling anyway.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -120,7 +125,6 @@ const CartLineItemRow = ( { lineItem } ) => {
 			</td>
 			<td className="wc-block-cart-item__quantity">
 				<QuantitySelector
-					disabled={ itemQuantityDisabled }
```
haszari, line 123: Curious - is it worth keeping the `disabled` attribute on each of these form controls? I realise you're applying an `.is-loading` overlay/mask to disable these, but perhaps it's useful to keep the semantic `disable` as well (accessibility?). Not a blocker just clarifying.

assets/js/base/hooks/cart/use-store-cart-item-quantity.js
```diff
@@ -62,9 +77,29 @@ export const useStoreCartItemQuantity = ( cartItem ) => {
 			changeCartItemQuantity( cartItemKey, debouncedQuantity );
 		}
 	}, [ debouncedQuantity, cartItemKey ] );
+	useEffect( () => {
+		if ( isPendingQuantity ) {
```
haszari, line 81: It looks like there are no components that use `isPendingQuantity` - do we need it in the store, or is it just used to for the `incrementCalculating`/`decrementCalculating`. Curious if we can simplify that code, since the row component only needs `isPendingDelete`, and cart overall needs `isCalculating`.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
Aljullu: Agree with @haszari, if keyboard navigation allows focusing the quantity input, it should be explicitly disabled.

senadir: True, pointer-events only prevent pointer events (🤦🏻‍♂️) thanks for pointing out, will redo them.

assets/js/base/hooks/cart/use-store-cart-item-quantity.js
senadir: we need it for the `incrementCalculating`/`decrementCalculating` yeah, currently I see no way to communicate between action dispatchers and our hook except via the store, I would be happy to reduce this it's possible.

PR #2108, opened 2020-04-03
Fix flaw with event emitters and complete checkout flow for the cheque payment method.

nerrad:
With master having the necessary pieces in place to theoretically complete a payment using the cheque payment method and fulfill the checkout flow (including a redirect to order receipt page), I took some time to test things out and fix bugs that popped up.

In the process I discovered a _serious_ flaw with the event emitters that I overlooked when designing the system. The flaw is that all the registered observers on the event emitters _cannot react to any state changes_ because they are all invoked in one render process.  This also means that any state changes in observers can be problematic because subsequent observers would not know of those state changes.

The solution involves recognizing that we are using the event emitters for tracking flow through the checkout and to limit _status_ and _state_ changes to after the event has emitted.  By adding event emitters to the PaymentMethodData context, and adding another Checkout status it actually ended up simplifying the flow logic because now there is a stable point for payment methods to hook in their behaviour and for the `CheckoutProcessor` component to hook in it's logic. 

> Sidenote: I realize "simplifying" seems to be an oxymoron here because it's still relatively complex. But the complexity is absorbed by the providers instead of passed on to extension components to manage (which can increase complexity exponentially if extensions have to check constantly for status changes among all the elements in the checkout flow).

With the changes in this pull:

- Payment method extensions will not be responsible for updating the payment status. The payment method provider context handles that. Payment method components will register observers on the events they need to react to and return appropriate values for communicating with checkout.
- Convention (not enforced) will be that observers _must_ not set any state in checkout (they can set their own local state if they want). Provider state (payment, checkout, shipping) is handled by provider's emitters after all observers for an event have fired.

Here's a _rough_ diagram of the flow (not including shipping events):

CheckoutFlow@2x

In this pull:

- new payment method event emitters are added,
- new actions/status to checkout state context added (checkoutProcessingComplete)
- refactor checkout state provider.
- refactor checkout processor.
- wire up cheque payment method so it's hooked into the `onPaymentProcessing` event emitter.

## To Test

It is expected that Cart/Checkout behaviour up to the point of clicking the related submit button should continue to work as expected outside what was done in this branch.

You should be able to finish a complete checkout flow using the cheque payment method without errors.

There should be no errors in the console while testing.

Stripe is _not_ expected to work.

Validation errors after submit button is clicked _may_ have regressed. Please do test, but we can fix in follow-ups unless the cause is clear.

## Followups:

- since payment status is now controlled by checkout, I'm going to remove the payment status setter from being exposed on the `usePaymentMethodInterface` hook. That forces payment methods to use the appropriate api for hooking into the checkout flow (events).
- stripe client side will need modified to work with the new payment event emitters.
- I _may_ have broken some validation behaviour. If so, let's fix in follow-ups to avoid this pull becoming stale.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/event-emit.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/event-emit.js
@@ -1,0 +1,99 @@
+/**
+ * Internal dependencies
+ */
+import { actions, reducer, emitEvent, emitEventWithAbort } from '../event-emit';
+
+const EMIT_TYPES = {
+	PAYMENT_PROCESSING: 'payment_processing',
+	PAYMENT_SUCCESS: 'payment_success',
+	PAYMENT_FAIL: 'payment_fail',
+	PAYMENT_HAS_ERROR: 'payment_has_error',
+};
+
+/**
+ * Receives a reducer dispatcher and returns an object with the
+ * various event emitters for the payment processing events.
+ *
+ * Calling the event registration function with the callback will register it
+ * for the event emitter and will return a dispatcher for removing the
+ * registered callback (useful for implementation in `useEffect`).
+ *
+ * @param {Function} dispatcher The emitter reducer dispatcher.
+ *
+ * @return {Object} An object with the various payment event emitter
+ *                  registration functions
+ */
+const emitterSubscribers = ( dispatcher ) => ( {
+	onPaymentProcessing: ( callback, priority ) => {
+		const action = actions.addEventCallback(
+			EMIT_TYPES.PAYMENT_PROCESSING,
+			callback,
+			priority
+		);
+		dispatcher( action );
+		return () => {
+			dispatcher(
+				actions.removeEventCallback(
+					EMIT_TYPES.PAYMENT_PROCESSING,
+					action.id
+				)
+			);
+		};
+	},
+	onPaymentSuccess: ( callback, priority ) => {
+		const action = actions.addEventCallback(
+			EMIT_TYPES.PAYMENT_SUCCESS,
+			callback,
+			priority
+		);
+		dispatcher( action );
+		return () => {
+			dispatcher(
+				actions.removeEventCallback(
+					EMIT_TYPES.PAYMENT_SUCCESS,
+					action.id
+				)
+			);
+		};
+	},
+	onPaymentFail: ( callback, priority ) => {
+		const action = actions.addEventCallback(
+			EMIT_TYPES.PAYMENT_FAIL,
+			callback,
+			priority
+		);
+		dispatcher( action );
+		return () => {
+			dispatcher(
+				actions.removeEventCallback(
+					EMIT_TYPES.PAYMENT_FAIL,
+					action.id
+				)
+			);
+		};
+	},
+	onPaymentError: ( callback, priority ) => {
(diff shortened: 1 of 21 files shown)


nerrad:
I have no idea why bundlesize is reporting 0B for the payment method builds. I tried a production build of them locally and didn't have any issue.

assets/js/base/context/cart-checkout/payment-methods/event-emit.js
```diff
@@ -0,0 +1,99 @@
...
+		dispatcher( action );
+		return () => {
+			dispatcher(
+				actions.removeEventCallback(
+					EMIT_TYPES.PAYMENT_HAS_ERROR,
+					action.id
+				)
+			);
+		};
+	},
```
Aljullu, line 90: No need to do it in this pull, but this looks like something that could be abstracted. These four methods are the same just changing the type. Same for `checkout-state/event-emit.js` and `shipping/event-emit.js`.

assets/js/payment-method-extensions/payment-methods/cheque/index.js
```diff
@@ -16,9 +17,25 @@ const settings = getSetting( 'cheque_data', {} );
...
+ */
 
+/**
+ * Cheque content component
+ *
+ * @param {RegisteredPaymentMethodProps|Object} props Incoming props
+ */
+const Content = ( { activePaymentMethod, eventRegistration } ) => {
+	// hook into payment processing event.
+	useEffect( () => {
```
mikejolley, line 31: Can you help me understand what this does exactly?

Can we abstract/import it so gateways have less to worry about?

assets/js/base/context/cart-checkout/payment-methods/event-emit.js
nerrad: Ya this is a good idea. Done in 01eb74cca

assets/js/payment-method-extensions/payment-methods/cheque/index.js
nerrad: Basically every gateway will have to at a _minimum_ register an observer on the `onPaymentProcessing` event.  This event fires when checkout has initiated payment processing as a part of the flow.  What gateways return from this callback when it's invoked determines whether the gateway was succesful, failed, or errored for it's processing. The possible shapes of the return value are:

### Success
Can return:
- `true`
- OR an object: `{ paymentMethodData, billingData }`

### Fail:
Return an object: `{ fail: { errorMessage, paymentMethodData } }`

### Error:
Return an object: `{ errorMessage, validationErrors }`

- errorMessage is used as just a general error message to display in the payment method step.
- validationErrors is an object where keys are field ids and values are the error message for that field (to help with displaying field validation errors for for the payment method).

nerrad: > Can we abstract/import it so gateways have less to worry about?

So one thing I was thinking we could do is have a function for registering really simple gateways that are similar to cheque in that they all the processing happens server side. So it'd have a fairly simple signature (gateway name, maybe icon, and text for the content window). Then this function would take care of the payment method registration (with `Config` etc) using components common to all these payment methods.  We could export this as a helper on the block-registry global.

Issue #2111, opened 2020-04-03
Update stripe payment method extension (client) to work with new event emitters.


nerrad:
In #2108 new event emitters were added for payment method status changes. In #2110 the payment status setter is being removed from the payment method interface. This issue is for ensuring the stripe payment method integration (client side) is updated to work with the api interface changes.

PR #2115, opened 2020-04-03
Update stripe payment methods to work with new event emitters

nerrad:
Fixes: #2111 (see issue for description).

In this pull I added a new interface for setting an express payment error (because there are a class of errors that will need displayed in the editor if the express payment method has an issue).  Here's a screenshot of how the error is displayed:

Image 2020-04-03 at 11 23 08 AM

The `setExpressPaymentError` function does not change payment status as effectively the only way to handle the error is to restart the payment process.  This function will clear the existing notice if invoked with a falsey value as the argument.

Also in this pull:

- refactored stripe cc payment method to remove all usages of payment status dispatcher.
- refactored stripe apple pay method to remove all usages of payment status dispatcher.

## To test

You can test the new notice function by replacing the existing `payment-extensions/payment-methods/stripe/index.js` file with the following temporary code:

```jsx
/**
 * External dependencies
 */
import {
	registerPaymentMethod,
	registerExpressPaymentMethod,
} from '@woocommerce/blocks-registry';

/**
 * Internal dependencies
 */
import stripeCcPaymentMethod from './credit-card';
import ApplePayPaymentMethod from './apple-pay';
import { applePayImage } from './apple-pay/apple-pay-preview';

registerPaymentMethod( ( Config ) => new Config( stripeCcPaymentMethod ) );
registerExpressPaymentMethod(
	( Config ) => new Config( ApplePayPaymentMethod )
);

const DummyApplePay = ( { setExpressPaymentError } ) => {
	return (
		<a onClick={ () => setExpressPaymentError( 'Testing Error' ) }>
			<img
				src={ applePayImage }
				onClick={ () => setExpressPaymentError( 'Testing Error' ) }
				alt="apple pay dummy"
			/>
		</a>
	);
};

registerExpressPaymentMethod(
	( Config ) =>
		new Config( {
			id: 'dummy-apple-pay',
			content: <DummyApplePay />,
			edit: <DummyApplePay />,
			canMakePayment: Promise.resolve( true ),
		} )
);
```

That will add a dummy apple pay button to test on the frontend and it basically just triggers the notice when the button is clicked.

- Verify checkout and cart blocks load in the editor and frontend with no console errors.

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
@@ -155,6 +155,7 @@ const ApplePayExpressComponent = ( {
 	// kick off payment processing.
 	const onButtonClick = () => {
 		setActivePaymentMethod( PAYMENT_METHOD_NAME );
+		setExpressPaymentError( '' );
 		setApplePayProcessing( true );
 	};
 
--- a/assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
@@ -68,6 +68,7 @@ const ApplePayExpressComponent = ( {
 	onSubmit,
 	activePaymentMethod,
 	setActivePaymentMethod,
+	setExpressPaymentError,
 } ) => {
 	/**
 	 * @type {[ StripePaymentRequest|null, function( StripePaymentRequest ):StripePaymentRequest|null]}
@@ -76,6 +77,7 @@ const ApplePayExpressComponent = ( {
 	const [ paymentRequest, setPaymentRequest ] = useState( null );
 	const stripe = useStripe();
 	const [ canMakePayment, setCanMakePayment ] = useState( false );
+	const [ applePayProcessing, setApplePayProcessing ] = useState( false );
 	const eventHandlers = useRef( DEFAULT_STRIPE_EVENT_HANDLERS );
 	const currentBilling = useRef( billing );
 	const currentShipping = useRef( shippingData );
@@ -116,7 +118,11 @@ const ApplePayExpressComponent = ( {
 		}
 		// otherwise we just update it (but only if payment processing hasn't
 		// already started).
-		if ( ! paymentStatus.isPristine && currentPaymentRequest.current ) {
+		if (
+			! paymentStatus.isPristine &&
+			! applePayProcessing &&
+			currentPaymentRequest.current
+		 ) {
 			updatePaymentRequest( {
 				// @ts-ignore
 				paymentRequest: currentPaymentRequest.current,
@@ -133,6 +139,7 @@ const ApplePayExpressComponent = ( {
 		billing.cartTotalItems,
 		paymentStatus.isPristine,
 		stripe,
+		applePayProcessing,
 	] );
 
 	// whenever paymentRequest changes, then we need to update whether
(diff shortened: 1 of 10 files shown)


assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
```diff
@@ -223,17 +221,33 @@ const ApplePayExpressComponent = ( {
...
+		}
+		return true;
+	};
+
 	const onCheckoutComplete = ( forSuccess = true ) => () => {
 		const handlers = eventHandlers.current;
 		if (
 			typeof handlers.sourceEvent === 'function' &&
-			currentPaymentStatus.current.currentStatus.isSuccessful
+			applePayProcessing
```
Aljullu, line 245: I'm a bit confused, shouldn't it be:
```suggestion
			! applePayProcessing
```

assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
```diff
@@ -223,17 +221,33 @@ const ApplePayExpressComponent = ( {
...
 		) {
 			if ( forSuccess ) {
 				completePayment( handlers.sourceEvent );
-			} else {
-				abortPayment( handlers.sourceEvent );
 			}
+			abortPayment( handlers.sourceEvent );
```
Aljullu, line 250: Why is `abortPayment` outside the `else` clause, now? Shouldn't it be called only when `forSuccess` is false?

assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
mikejolley: @Aljullu It looks like `applePayProcessing` might be to track if the button to process applepay was pressed, so that this logic only runs on applepay requests? Docblocks would help clear this up perhaps.

assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
mikejolley: My guess was to reuse `setApplePayProcessing` but the side effect is that payment would complete, then fail. So looks like a bug.

senadir: yeah the function should return or exist at `completePayment( handlers.sourceEvent );` so that this doesn't get called.

assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
nerrad: Mike is correct here. Naming it `applePayIsProcessing` might also help.   What may be confusing is that originally payment methods took care of updating status and at this point, the payment method would have been set to successful. But now payment method status is set by checkout and internally for the payment method at this point the payment method just cares whether it is in an internal processing state (i.e. the button has been clicked and the browser payment window hasn't closed yet).

assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
nerrad: yup a bug, I noticed and fixed in another in progress pull, but I'll add the fix here too.

nerrad: See e04b413c6 for fix. Also noticed that the completePayment call wasn't setting `applePayProcessing` to false.

assets/js/payment-method-extensions/payment-methods/stripe/apple-pay/apple-pay-express.js
nerrad: I'm going to leave this as is for this pull. I have an in progress pull (just waiting for other pulls to get merge before I wind up) that changes apple-pay to payment-request everywhere because with a few tweaks I was able to just support the payment request api in it's entirety (so includes apple pay, google pay, microsoft pay and any other browsers using the payment request api). So there's some naming changes etc happening in that pull.

PR #2117, opened 2020-04-03
Add forcing source map on production to analyze tree-shaking

senadir:
This PRs adds a couple of changes to our command package.json and a way to force webpack to build maps in production.
Now by running `npm run build:map` you will build tree-shaked production files with maps included, previously maps only got generated in dev.
It also adds a command `npm run explore` that you can run along side a glob or a file path, it uses the file source map to analyze the file dependency tree and size, it helps us catch (manually) large dependency that should be treeshaked, examples of the command:
```
npm run explore build/*-frontend.js
```
analyze all frontend packages, it gives this rather not-so-beautiful-and-quite-dense page with handy select to change files, when you click on a dependency, it zooms in.
image

Note that all sizes are normal, not compressed, to force compression, pass `-- --gzip` by the end of the command, other commands can [be found here](https://github.com/danvk/source-map-explorer#options).

This should not mess with anything else, and it should help debug, usually, when I need to debug bundle size, I use my globally installed version of `source-map-explorer` and short-circuit webpack to force it to produce source maps in production.


senadir:
I also ordered our commands, well, a command did and I just pasted the result.

Issue #2058, opened 2020-03-30
CheckoutProcessor - Hook up payment gateway `payment_data` to request


mikejolley:
Follow up from #2044

The `CheckoutProcessor` component  can pass data from gateways to the server. Therefore, we need to expose and format `payment_data` so it can be sent along with the request.

nerrad:
As noted this will be fixed by #2120

Issue #2054, opened 2020-03-30
Implement server side handling for Stripe Payment Methods


nerrad:
Using the existing [Stripe payment gateway extension](https://github.com/woocommerce/woocommerce-gateway-stripe) as a guide, we need to wire up the stripe payment methods to the checkout processing endpoint server side.

Once things are wired up server side, this will also involve testing the purchase flow and fixing any bugs remaining in how the payment methods are wired up client side.

PR #2120, opened 2020-04-03
Implement stripe server-side handling (stripe cc)

nerrad:
Fixes: #2054
Fixes: #2058 

With the work in this pull, one can use stripe CC payment method to complete a checkout purchase!

Image 2020-04-03 at 4 15 16 PM

There was actually no server side code needed to implement this since the checkout endpoint has some nifty logic to enable existing server side gateway extension code to "just work" (ht @mikejolley ). Instead, I just had to finish wiring up the checkout processor to use the payment method provided payment data (while also making sure it's in the shape the endpoint expects) and also make sure the client side payment method component was also providing the data in the expected shape the gateway extension `process_payment` function was expecting.

**Note:** this pull has a dependency on #2119 (which I have the fix included in this pull to make it easier to test) and #2115  (which this pull's branch is based off of). So assuming testing goes well, we'll want to merge the dependent pulls first then rebase into this branch (basing it off master) before merging.

## To test

- Setup the WooCommerce Stripe Extension on your test site and add test credentials.
- Add some products to your cart.
- Fill out test credit card information.
- Click to make payment.
- Win.

## Caveats

This pull does not:

- handle failed/error payment responses from the server.
- handle surfacing validation issues for the payment method.

So conditions for merge (besides good code review) are that existing checkout block behaviour (compared with master) is not broken and that you can complete a _successful_ stripe payment without any console errors or user facing errors. A successful payment is considered one that results in making a purchase from the checkout block using stripe cc and being redirected to the order details page after purchase automatically.

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -14,6 +14,26 @@ import { useEffect, useRef, useCallback } from '@wordpress/element';
 import { useStoreNotices } from '@woocommerce/base-hooks';
 
 /**
+ * @typedef {import('@woocommerce/type-defs/payments').PaymentDataResponse} PaymentDataResponse
+ */
+
+/**
+ * Utility function for preparing payment data for the request.
+ *
+ * @param {Object} paymentData Arbitrary payment data provided by the payment
+ *                             method.
+ *
+ * @return {PaymentDataResponse[]} Returns the payment data as an array of
+ *                                 PaymentDataResponse objects.
+ */
+const preparePaymentData = ( paymentData ) => {
+	return Object.keys( paymentData ).map( ( property ) => {
+		const value = paymentData[ property ];
+		return { key: property, value };
+	}, [] );
+};
+
+/**
  * CheckoutProcessor component. @todo Needs to consume all contexts.
  *
  * Subscribes to checkout context and triggers processing via the API.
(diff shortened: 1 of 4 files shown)


assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -13,6 +13,26 @@ import {
...
+/**
+ * Utility function for preparing payment data for the request.
+ *
+ * @param {Object} paymentData Arbitrary payment data provided by the payment
+ *                             method.
+ *
+ * @return {PaymentDataResponse[]} Returns the payment data as an array of
+ *                                 PaymentDataResponse objects.
+ */
+const preparePaymentData = ( paymentData ) => {
```
haszari, line 29: On the fence about whether this adds value as a separate function, since it's so generic - i.e. mapping to `[ { key:, value: }]` format. There's probably a lodash routine that does exactly this; I don't think this is customised for our payment use case (apart from the name & typedef). Just a thought - happy for this to merge as is :)

nerrad: Even if there was a lodash function for this (which I'm not aware of their being - did you know of one?) - what value is there introducing a dependency here on lodash just for this specific use case?

> I don't think this is customised for our payment use case (apart from the name & typedef).

I'm not sure what you mean by this exactly, but it _is_ specifically customized for payment use case because the `checkout` processing endpoint [_requires_ payment data to be sent in the body in this format](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/501296234d1903eb9df2645a57df602175af4616/src/RestApi/StoreApi/Routes/Checkout.php#L60-L80).

PR #2121, opened 2020-04-05
Remove prop types on production to reduce bundle size

### LINKED ISSUE #1738: Consider wrapping PropTypes in a __DEV__ flag
We should consider adding a `__DEV__` flag that wraps development only code in it that will be disregarded in production, this should save some minor weight and performance and cleaner console for clients, e2e should in theory catch propType console errors.

for reference of how to implement this:
https://overreacted.io/how-does-the-development-mode-work/

Our own feature flag implementation to see how to introduce build-time flags:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1631

senadir:
This felt like a very easy win so I tackled it now, it should remove all prop types from production builds, it uses a babel plugin that automatically removes it on production.

Fixes #1738 however in a less manual way.

This should not break anything, we will if the bundle size changes. if it didn't then this didn't work and should be merged.

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -9,4 +9,9 @@ module.exports = {
 			},
 		],
 	],
+	env: {
+		production: {
+			plugins: [ 'react-remove-prop-types' ],
+		},
+	},
 };
--- a/package-lock.json
+++ b/package-lock.json
@@ -7943,6 +7943,12 @@
 				}
 			}
 		},
+		"babel-plugin-react-remove-prop-types": {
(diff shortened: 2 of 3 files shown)


senadir:
🤦🏻‍♂️🤦🏻‍♂️🤦🏻‍♂️🤦🏻‍♂️🤦🏻‍♂️
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/839

PR #2124, opened 2020-04-06
Show validation errors again

### LINKED ISSUE #2096: Checkout Context: replace `hasError` with an error counter or set
Discussed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2033#discussion_r401694931.

Currently, the Checkout Context has a property named `hasError` which is a boolean. The issue with the current implementation is that if several different consumers have an error, it's not possible to clear only one of them, so we would need a different system to be able to add/remove errors dynamically.

### LINKED ISSUE #2112: Fix validation error handling regression in #2108
With the work in #2108 it seems there's a regression with the _display_ of errors.  Validation error detection will still kick in, it's just not displaying the errors when the checkout process order button is pressed.

Aljullu:
Fixes #2112.
Closes #2096.

This PR makes it so validation errors show again in the _Checkout_ block, it also makes validation errors added by extensions to be displayed at the same time default validation errors are shown.

In order to test it, let's force the payment method to return an error in certain circumstances. In `payment-method-extensions/payment-methods/stripe/credit-card/use-checkout-subscriptions.js` [L111](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-checkout-subscriptions.js#L111), paste something like this:
```JS
if ( billingData.postcode !== 'ABCD' ) {
	paymentStatus
		.setPaymentStatus()
		.error( 'Stripe found an error with your postcode.' );
	const key = billingData.shippingAsBilling
		? 'shipping-postcode'
		: 'billing-postcode';
	return {
		[ key ]: {
			message: 'Postcode must be ABCD.',
		},
	};
}
```
1. Go to the _Checkout_ block and click on _Place Order_ when you have some validation errors (and the postal code is different from `ABCD`).
2. Make sure the validation errors + the payment method errors appear all at once.
3. Make sure there were no API calls before all validation errors were fixed.


assets/js/base/context/cart-checkout/checkout-state/index.js
```diff
@@ -135,22 +134,25 @@ export const CheckoutStateProvider = ( {
 
 	// emit events.
 	useEffect( () => {
-		const status = checkoutState.status;
+		const { hasError, status } = checkoutState;
 		if ( status === STATUS.PROCESSING ) {
-			emitEventWithAbort(
+			emitEvent(
```
Aljullu, line 139: If we go with these changes, `emitEventWithAbort` won't be used anywhere, so we could remove that method if we don't see a usage in the near future.

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -39,10 +43,28 @@ const CheckoutProcessor = () => {
...
+			( checkoutIsProcessing || checkoutIsProcessingComplete )
+		) {
+			dispatchActions.setHasError( checkoutWillHaveError );
+		}
+	}, [
+		checkoutWillHaveError,
+		checkoutHasError,
+		checkoutIsProcessing,
+		checkoutIsProcessingComplete,
+	] );
```
Aljullu, line 63: Instead of every consumer dispatching `setHasError` to update Checkout's error state, the Checkout processor subscribes to the consumers and dispatches error state updates internally. In my opinion it simplifies the logic because everything is more self-contained. I think this change makes #2096 irrelevant now, but we can reopen it if we still think it would be good for extensions to have a finer control on the number of errors in the Checkout state.

assets/js/base/context/cart-checkout/event-emit/emitters.js
```diff
@@ -24,17 +24,21 @@ const getObserversByPriority = ( observers, eventType ) => {
  */
 export const emitEvent = async ( observers, eventType, data ) => {
 	const observersByType = getObserversByPriority( observers, eventType );
+	let emitterResponse = {};
 	for ( const observer of observersByType ) {
 		try {
-			await Promise.resolve( observer.callback( data ) );
+			const response = await Promise.resolve( observer.callback( data ) );
+			if ( typeof response === 'object' ) {
+				emitterResponse = { ...emitterResponse, ...response };
```
Aljullu, line 32: This change is in order to collect errors produced by all observers without aborting on the first one that fails. This allows us to show validation errors + payment method errors at once. I looked at how `emitEvent` is used around the codebase and I think this change should be safe, but would like some feedback if somebody thinks it will break something now or in the future.

nerrad: Looks okay to me.

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -186,6 +182,7 @@ export const ShippingDataProvider = ( { children } ) => {
 		onShippingRateSelectSuccess,
 		onShippingRateSelectFail,
 		needsShipping,
+		hasError,
```
nerrad, line 185: I'm not sure I see the need for splitting out `hasError` from `shippingErrorStatus` (which already contains it).

assets/js/base/context/cart-checkout/validation/index.js
```diff
@@ -178,7 +169,7 @@ export const ValidationContextProvider = ( { children } ) => {
 		hideValidationError,
 		showValidationError,
 		showAllValidationErrors,
-		hasValidationErrors,
+		hasValidationErrors: Object.keys( validationErrors ).length > 0,
```
nerrad, line 172: Nice bit of improvement here 👍

assets/js/base/context/cart-checkout/shipping/index.js
Aljullu: `shippingErrorStatus` is a string with the values `none`, `invalid_address` or `unknown_error`. So it's difficult to use in other contexts because we need to compare that value with the shipping constants.

Alternatively, we could make the shipping context to expose `currentErrorStatus`, which it does expose a `hasError` property. Would that make more sense?

assets/js/base/context/cart-checkout/validation/index.js
Aljullu: Kudos to @mikejolley. I think he suggested this improvement some time ago in Slack and this pull was a good opportunity to implement it.

assets/js/base/context/cart-checkout/shipping/index.js
nerrad: oh shoot right, I was confusing `currentErrorStatus` with `shippingErrorStatus`.  I think I had always originally intended to export `currentErrorStatus` as `shippingErrorStatus` and I obviously did not. The intent was not to expose the raw status but just the `is*` equivalents, that way consumers wouldn't need to do it's own conditional checks.

So with that in mind, let's:

- expose currentErrorStatus via aliasing `shippingErrorStatus: currentErrorStatus` in the exported `ShippingData` 
- we should memoize `currentErrorStatus` via `useMemo`.
- switch the emmitters passing along `shippingErrorStatus` to pass along `currentErrorStatus` to observers instead.

Aljullu:
I needed to do some more changes after rebase to get validation errors to work. Even though this PR was already approved, feel free to take another look at commit fa53f379f28bf5efe32295909368759d1d0c558e and onwards if you want.

Also the example code to add validation errors to payment methods has been updated.

assets/js/base/context/cart-checkout/event-emit/emitters.js
```diff
@@ -25,11 +25,16 @@ const getObserversByPriority = ( observers, eventType ) => {
...
-			if ( typeof response === 'object' ) {
-				emitterResponse = { ...emitterResponse, ...response };
+			const observerResponse = await Promise.resolve(
+				observer.callback( data )
+			);
+			if ( typeof observerResponse === 'object' ) {
+				emitterResponse = { ...emitterResponse, ...observerResponse };
+			} else {
+				observerResponses.push( observerResponse );
 			}
```
nerrad: I think if observerResponse is not an object, we'd have to assume it's okay and just not push to the collected array. There is no aborting so an observer returning anything other than an object has nothing useful to signal to the emitter right? So I don't think status changes should happen if there's nothing useful to indicate _why_ status is changing.

Aljullu:
In the last commit 35edcb3b5ca090dd53619f28af50fdd3abfabb46 I undid some of the changes to the event emitter, so everything that is not an object behaves as if it was `true`. In addition to that, I added an `errorMessage` property in parallel to a `validationErrors` one, so observers can add notice errors, as discussed in Slack:

imatge
 
@nerrad this is ready for another look.

PR #2127, opened 2020-04-06
Convert apple pay integration to payment request integration and finish implementation

nerrad:
Over the weekend I did an experiment to see how hard it would be to just implement the Stripe payment request api completely instead of just apple pay. Turns out, it wasn't hard at all (just needed to tweak things a bit).

In this pull:

### The full stripe PaymentRequest api is supported (for express payments)

This means we have full support for ApplePay, GooglePay, Microsoft pay and any payment method that supports the payment request api (as made available via stripe).  In particular the utility of having google pay support means for easier local development/testing as long as you have https:// enabled.

### Complete checkout flow is supported for express payments

Here's a list of expectations that are supported by the work in this pull.

- All _successful_ payments are tested and work. I have not verified failures or error handling in payment methods yet because that increases the scope of testing for this pull, followups will need to test and make any necessary fixes. Error handling and fail logic _is_ in place, it's just not tested.
- Shipping is supported for express payment methods (when shipping is required for the order). So selecting an address should result in rates getting updated in the payment request interface and selecting rates should update the totals in the payment request interface.
- Making a payment using the payment request interface should complete successfully and update the order with the correct details (except for the known issue as outlined in #2139)
- Shipping package support is _not_ available with this work (see followups section).
- Server side processing utilizes the stripe payment gateway (request payment method) for handling the extra details for express payments passed through via checkout.
- **Note:** When express payment methods are being used to make a payment, checkout field validation is disabled. This is because the express payment method has its own internal validation and things break with the current order of processing that includes the checkout form validation (the form is typically _empty_ and thus fails validation before payment processing is triggered for express payment method).

~**Known issue with ApplePay and purchases involving shipping**~ (fixed by #2163)

It looks like applePay doesn't do any validation of state/province or postal code, so it's possible for user's to type whatever they want in the field (at least applePay in the browser). Since our shipping address hook (useShippingAddress) uses a comparison of the previous address to the new address (specific address fields) to determine whether to update shipping rates, this means on successful payment, the passed through shipping field will trigger a new shipping rate request and this causes checkout state to go back to idle (which eventually triggers a bunch of shipping rate and checkout processing requests as states flip around).  

So after more testing, I discovered that the issue here is the `shippingaddresschange` event in stripe may not contain all address details - and even could contain partial/redacted address details (see [here](https://stripe.com/docs/js/payment_request/events/on_shipping_address_change)). I was using that to get shipping rates (and it will only work partially - rates specific to zip code won't be guaranteed to work). It does look like the address provided on the `source` event (via the source object) _is_ correct except ApplePay provides the full state/province name, not the state code. Another issue is that when the shipping request happens after payment is successful, the checkout calculating state is incremented which changes checkout status to calculating. After calculating is decremented, the status goes back to the previous state. This is clever and allows for less logic in the code for setting state, but the problem here is that the previous state becomes `checkoutProcessingComplete` and this causes the `processCheckout` action in the `CheckoutProcessor` component to fire again (which in turn will trigger the payment success event in the payment method again which in turn causes shipping rates to recalculate etc (creating a bit of a loop). So I may have to separate calculating state from the actual checkout processing status so that it's possible for checkout to be processing and calculating or not calculating (similar to how checkout has has an error state separate from it's status). So essentially ending up with three checkout tracked states:

- Error state: checkout either has an error (derived from other error states) or not.
- Calculating state: derived from all loading states that contribute to total changes (`isCalculating` would simply be whether there is a positive value for the calculating increments).
- Flow state: which is one of `PRISTINE`, `IDLE`, `PROCESSING`, `PROCESSING_COMPLETE`, `COMPLETE`. At any point in the checkout flow the checkout could be also calculating or have an error. I think that will make things a bit more clear and precise (and prevent some of the loops I'm seeing).

## To Test

> Note: you can use the ephemeral site I setup for testing if you want: https://ephemeral-nerrad-20200302.atomicsites.blog/

For the express payment methods, probably the easiest to test is using Google Pay (chrome browser) as it can be tested locally as long as you have ssl enabled for your local dev site.  I'm fine with merging if Google Pay works okay (but I am going to test Apple Pay today as well).

- See the list of expectations and verify that it works as expected for making payments with express payment methods. Keep the console open while testing and watch for errors. Also watch that there aren't more network requests than expected.
- Verify that cc and cheque payment methods still process payments successfully and that validation works as expected for non express payments.

## Followups

- stripe.paymentRequest payments do not handle multiple packages. This will need to be handled in a followup. We'll either have to consider disabling the payment method if there are multiple packages in the order/cart, or setup the [shipping options](https://stripe.com/docs/js/appendix/shipping_option) for the payment method so each option is a combination of the various package selections (since only one option can be selected). The latter can be super complex and grow exponentially depending on the number of packages might exist.  I'll create an issue as a followup after this is merged.  I should note that I did look at the existing stripe payment gateway extension and how they handle packages is a bit puzzling too...so this may be something that's already a bit of a known issue for express payment methods (might need a p2).
- the `canPay` config option currently is using hardcoded country and currency code values, this will need updated to use something provided by the server so it's more context aware of the shopper in the checkout.
- currently express payment methods need to take care of setting active payment method when their payment request button is clicked and also restore active payment method when the payment request modal is closed. In this pull the payment request express component is handling this logic but ideally it would be moved somehow to checkout so that payment methods don't have to manage (or forget!) that complexity.
- as discussed in [this comment thread](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2127#discussion_r404891414) the current implementation of deriving server side payment_method id needs reworked. This will involve ensuring client side is always passing along the correct payment_method for executing payment processing (instead of leaving it up to payment methods to correctly define as a part of the `paymentData` package in _some_ scenarios - namely stripe express). There's potentially some overlap with this item and the previous one.


assets/js/base/components/payment-methods/saved-payment-method-options.js
```diff
@@ -103,7 +103,7 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 	}, [ customerPaymentMethods, selectedToken ] );
 	const updateToken = useCallback(
 		( token ) => {
-			if ( token !== 0 ) {
+			if ( token !== '0' ) {
```
nerrad, line 106: Input values are strings, hence this change.

assets/js/base/context/cart-checkout/checkout-state/index.js
```diff
@@ -137,15 +137,15 @@ export const CheckoutStateProvider = ( {
...
+		const { status } = checkoutState;
 		if ( status === STATUS.PROCESSING ) {
 			removeNotices( 'error' );
 			emitEvent(
 				currentObservers.current,
 				EMIT_TYPES.CHECKOUT_PROCESSING,
 				{}
 			).then( ( response ) => {
-				if ( response !== true || hasError ) {
+				if ( response !== true ) {
```
nerrad, line 148: hasError doesn't matter in this context because if there's an error from somewhere else, it's already handled elsewhere. Here we only care about any errors passed through on the event emitter.

assets/js/base/context/cart-checkout/checkout-state/index.js
```diff
@@ -160,8 +160,11 @@ export const CheckoutStateProvider = ( {
 				}
 			} );
 		}
-		if ( status === STATUS.COMPLETE ) {
-			if ( hasError ) {
+	}, [ checkoutState.status, setValidationErrors ] );
+
+	useEffect( () => {
+		if ( checkoutState.status === STATUS.COMPLETE ) {
+			if ( checkoutState.hasError ) {
```
nerrad, line 167: I split up the event emitters into separate effects because in my testing it seemed to reduce the amount of times the effects were called.

assets/js/base/context/cart-checkout/checkout/index.js
```diff
@@ -41,14 +41,14 @@ export const CheckoutProvider = ( {
...
+				<ShippingDataProvider>
+					<PaymentMethodDataProvider
+						activePaymentMethod={ initialActivePaymentMethod }
+					>
 						{ children }
 						<CheckoutProcessor />
-					</ShippingDataProvider>
-				</PaymentMethodDataProvider>
+					</PaymentMethodDataProvider>
+				</ShippingDataProvider>
```
nerrad, line 51: Due to the PaymentMethodDataProvider now consuming shipping data context (so `setPaymentStatus.success()` also receives shipping data to update in the state), I had to change the nesting order of the providers.

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -56,20 +56,25 @@ const CheckoutProcessor = () => {
 		currentStatus: currentPaymentStatus,
 		errorMessage,
 		paymentMethodData,
+		expressPaymentMethods,
```
nerrad, line 59: This is used to disable validation when an express payment method is active.

assets/js/base/context/cart-checkout/event-emit/emitters.js
```diff
@@ -1,7 +1,7 @@
 const getObserversByPriority = ( observers, eventType ) => {
 	return observers[ eventType ]
 		? Array.from( observers[ eventType ].values() ).sort( ( a, b ) => {
-				return b.priority - a.priority;
+				return a.priority - b.priority;
```
nerrad, line 4: This order was wrong (and I noticed that the test expectation was wrong!). The intent is that _lower_ priority observers get emitted before higher priority.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -124,10 +126,13 @@ export const PaymentMethodDataProvider = ( {
...
-		setActive( paymentMethodSlug );
-		dispatch( statusOnly( PRISTINE ) );
-	};
+	const setActivePaymentMethod = useCallback(
+		( paymentMethodSlug ) => {
+			setActive( paymentMethodSlug );
+			dispatch( statusOnly( PRISTINE ) );
+		},
+		[ setActive, dispatch ]
+	);
```
nerrad, line 135: Memoizing things in the context is more important (especially for nested context providers like we're doing) to limit the amount of `useRef` usage which is code smell.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -138,13 +143,15 @@ export const PaymentMethodDataProvider = ( {
...
-			context: 'wc/express-payment-area',
-			id: 'wc-express-payment-error',
-		} );
-		if ( ! message ) {
+		if ( message ) {
+			addErrorNotice( message, {
+				context: 'wc/express-payment-area',
+				id: 'wc-express-payment-error',
+			} );
+		} else if ( ! message ) {
```
nerrad, line 154: This fixes a bug where an error notice with no content could get set (when the intention was to clear the notice)!

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -241,12 +248,21 @@ export const PaymentMethodDataProvider = ( {
...
 			 */
-			success: ( paymentMethodData = {}, billingData = null ) => {
+			success: (
+				paymentMethodData = {},
+				billingData = null,
+				shippingData = null
+			) => {
+				if ( shippingData !== null && shippingData?.address ) {
+					setShippingAddress( shippingData.address );
+				}
```
nerrad, line 265: The payment success status dispatcher now will also set shipping address if present (because express payment methods could provide shipping address).

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -130,14 +135,27 @@ export const ShippingDataProvider = ( { children } ) => {
...
+		currentErrorStatus.hasError,
+		shippingErrorStatus,
+	] );
+
+	useEffect( () => {
+		if (
+			! shippingRatesLoading &&
+			shippingRates &&
+			! currentErrorStatus.hasError
+		) {
```
nerrad, line 151: We should only emit this event when rates are not loading.

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -147,14 +165,27 @@ export const ShippingDataProvider = ( { children } ) => {
...
+		currentErrorStatus,
+		shippingErrorStatus,
+	] );
+
+	useEffect( () => {
+		if (
+			! isSelectingRate &&
+			selectedRates &&
+			! currentErrorStatus.hasError
+		) {
```
nerrad, line 181: Similar to elsewhere, I found splitting out the emitters into their own effects helps reduce the amount of times the emitters (and effects) fire.

assets/js/base/utils/address.js
```diff
@@ -10,5 +10,5 @@ export const pluckAddress = ( { country, state, city, postcode } ) => ( {
 	country,
 	state,
 	city,
-	postcode,
+	postcode: postcode.replace( ' ', '' ),
```
nerrad, line 13: Since this function is used to pluck address data for comparison (which in turn is used to determine whether to do a http request to update shipping rates), normalization is needed here to cover cases where postal code might have spaces in it. This was a not so fun bug I had to troubleshoot while developing this pull.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
```diff
@@ -6,20 +6,22 @@ import { __ } from '@wordpress/i18n';
...
+
 const stripeCcPaymentMethod = {
 	id: PAYMENT_METHOD_NAME,
 	label: (
 		<strong>
 			{ __( 'Credit/Debit Card', 'woo-gutenberg-products-block' ) }
 		</strong>
 	),
-	content: <StripeCreditCard />,
+	content: <StripeCreditCard stripe={ stripePromise } />,
```
nerrad, line 24: The `stripePromise` object is now a `unique` instance of `Stripe` for each stripe payment method (cc vs payment request). This is important to ensure there is no lingering "session" info between payment methods if they are switched. We keep the same promise for determining whether payment can be made or not though.

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
```diff
@@ -0,0 +1,41 @@
...
+		// done.
+		// @todo, initial country and currency needs to server.
+		const paymentRequest = stripe.paymentRequest( {
+			total: {
+				label: 'Test total',
+				amount: 1000,
+			},
+			country: 'CA',
+			currency: 'cad',
+		} );
```
nerrad, line 31: This needs to be handled in a follow-up so that `country` and `currency` are provided by the server.

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/payment-request-express.js
```diff
@@ -0,0 +1,380 @@
...
+	/**
+	 * @type {[ StripePaymentRequest|null, function( StripePaymentRequest ):StripePaymentRequest|null]}
+	 */
+	// @ts-ignore
+	const [ paymentRequest, setPaymentRequest ] = useState( null );
+	const stripe = useStripe();
+	const [ canMakePayment, setCanMakePayment ] = useState( false );
+	const [ paymentRequestType, setPaymentRequestType ] = useState( '' );
+	const [ isProcessing, setIsProcessing ] = useState( false );
+	const [ isFinished, setIsFinished ] = useState( false );
```
nerrad, line 73: This payment method needs to track it's internal state of processing or finished. It can't use `paymentStatus` because that has potentially different meanings in the checkout context. Also payment status is for whatever the active payment method is and we still need to initialize some things in this payment method regardless of the payment status. 

Also, event observers _cannot update context state_ but only local state, so that has a huge influence on this approach.

This potentially can be revisited (and maybe use a combination of `activePaymentMethod` along with `paymentStatus`) but for now, these states make things simpler to reason about.

assets/js/base/context/cart-checkout/checkout/index.js
mikejolley: Does it make sense to nest in order of fields on checkout? Shipping>Billing>Payment? Does billing and shipping being nested potentially cause issues down the road if they depend on each other?

assets/js/base/utils/address.js
mikejolley: Qu: does this also need converting to upper case?

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
mikejolley: This can either use store settings (base country, currency) or we can add some extra data to checkout endpoint (billing country, with base country fallback, and order currency)

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
```diff
@@ -0,0 +1,41 @@
...
+		return paymentRequest.canMakePayment().then( ( result ) => {
+			if ( result ) {
+				return true;
+			}
+			return false;
```
mikejolley, line 36: Could simplify this ```return !! result;``

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/payment-request-express.js
```diff
@@ -0,0 +1,380 @@
...
+	const eventHandlers = useRef( DEFAULT_STRIPE_EVENT_HANDLERS );
+	const currentBilling = useRef( billing );
+	const currentShipping = useRef( shippingData );
+	const currentPaymentRequest = useRef( paymentRequest );
+	// for keeping track of what the active payment method was before this
+	// payment button was clicked.
+	const originalActivePaymentMethod = useRef( activePaymentMethod );
```
mikejolley, line 80: Struggling to understand why this is needed.

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/payment-request-express.js
```diff
@@ -0,0 +1,380 @@
...
+
+	// update refs when any change
+	useEffect( () => {
+		currentBilling.current = billing;
+		currentShipping.current = shippingData;
+		currentPaymentRequest.current = paymentRequest;
+	}, [ billing, shippingData, paymentRequest ] );
+
+	//set paymentRequest
```
mikejolley, line 89:
```suggestion
	// set paymentRequest
```

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
Aljullu: I was going to say the same. :slightly_smiling_face:  Even the entire function can be inlined:
```JS
return paymentRequest.canMakePayment().then( ( result ) => !! result );
```

src/Payments/Integrations/Stripe.php
```diff
@@ -49,6 +51,7 @@ final class Stripe extends AbstractPaymentMethodType {
 	 */
 	public function __construct( Api $asset_api ) {
 		$this->asset_api = $asset_api;
+		add_action( 'woocommerce_rest_checkout_process_payment_with_context', [ $this, 'add_payment_request_order_meta' ], 8, 2 );
```
mikejolley, line 54: Hey this is cheating :D I think this is ok for the beta this week but we should probably use this hook correctly in the integration before "public" launch and avoid the legacy handler.

assets/js/base/context/cart-checkout/checkout/index.js
nerrad: Payment is fine as is. We may need to switch Billing and Shipping around but that can be determined in followup?  I think the only place where it's potentially problematic is for copying data between shipping and billing (in which case the current order should be sufficient right (because shipping needs to write to billing).

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -135,14 +141,16 @@ const CheckoutProcessor = () => {
 	] );
 
 	useEffect( () => {
-		const unsubscribeProcessing = onCheckoutProcessing(
-			checkValidation,
-			0
-		);
+		let unsubscribeProcessing;
+		if ( ! expressPaymentMethodActive ) {
+			unsubscribeProcessing = onCheckoutProcessing( checkValidation, 0 );
```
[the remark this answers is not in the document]
nerrad: In the case of the `onCheckoutProcessing` event emitter, I was thinking we'd want the checkout validator to run first. I could be wrong though, why were you thinking it should run last?  Also, does it matter if all observer return values are going to be merged (or is that _precisely the reason it should be last_, so checkout validation errors have priority)?

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
nerrad: 🤦‍♂ Of course!

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -130,14 +135,27 @@ export const ShippingDataProvider = ( { children } ) => {
 				EMIT_TYPES.SHIPPING_RATES_FAIL,
 				currentErrorStatus
 			);
-		} else if ( ! shippingRatesLoading && shippingRates ) {
+		}
+	}, [
+		shippingRates,
+		shippingRatesLoading,
+		currentErrorStatus.hasError,
+		shippingErrorStatus,
```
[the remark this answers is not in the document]
nerrad: Good catch. Ya that's an oversight.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
nerrad: fixed in 758d6a2e

assets/js/base/context/cart-checkout/shipping/index.js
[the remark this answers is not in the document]
nerrad: Fixed in b671e9609706d4

assets/js/base/context/cart-checkout/shipping/index.js
```diff
@@ -147,14 +165,27 @@ export const ShippingDataProvider = ( { children } ) => {
 				EMIT_TYPES.SHIPPING_RATE_SELECT_FAIL,
 				currentErrorStatus
 			);
-		} else if ( ! isSelectingRate && selectedRates ) {
+		}
+	}, [
+		selectedRates,
+		isSelectingRate,
+		currentErrorStatus,
+		shippingErrorStatus,
```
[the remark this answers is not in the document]
nerrad: yup good catch - fixed in 17de9868e

assets/js/base/utils/address.js
nerrad: Ya probably should, good catch! (fixed in 17de9868e9)

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
nerrad: Ya, good catch guys. I eliminated some other code and totally missed the further simplification :) - fixed in 17de9868

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/payment-request-express.js
nerrad: When the checkout form is loaded, the active payment method is whatever is set for the payment methods tab embedded in the checkout. However, when express payment method button is clicked, it needs to set the active payment method to itself so checkout knows what payment method is active. The original active payment method needs to be kept so that when the express payment interface is closed by the express payment method, the active payment method can be reverted back to what it was in checkout (which is important if shopper changes their mind and doesn't want to - or can't - pay with the express payment method).

Ideally though, this logic would be handled by checkout so I'm going to add that to followup.

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/payment-request-express.js
nerrad: fixed this and other things (missing period) in 0b03b0b

src/Payments/Integrations/Stripe.php
nerrad: Hmm... what about this is cheating? I'm not sure what the _correct_ use of this hook should be?

src/RestApi/StoreApi/Routes/Checkout.php
```diff
@@ -357,7 +357,16 @@ protected function process_payment( \WC_Order $order, \WP_REST_Request $request
 	 * @return string
 	 */
 	protected function get_request_payment_method_id( \WP_REST_Request $request ) {
-		$payment_method = wc_clean( wp_unslash( $request['payment_method'] ) );
+		$payment_data = $this->get_request_payment_data( $request );
```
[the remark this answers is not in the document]
nerrad: The issue I'm working around here is that in the client the `active_payment_method` is set to a unique id across all payment methods (including express).  So we can't have both `stripe` as the id for the stripe cc payment method and `stripe` as the id in the client for the payment request payment method. Instead the payment methods are registered as `stripe` for stripe cc and `payment_request` for express payment.  This could be further compounded when the blocks get support for `stripe SEPA` etc payment methods.

So, we either fix this client side or server side.  I chose server side for now because it's trivial for payment methods to include what the payment method id is in the payment_data package. 

I'm open to other suggestions here :)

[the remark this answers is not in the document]
nerrad: Ya as mentioned in my reply on the other comment, I'm open to suggestions. I hope the explanation I gave in my other reply is sufficient, if not, I'm happy to clarify.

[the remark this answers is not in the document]
mikejolley: That's what I'd expect; stripe cc would have ID `stripe`, the express one would have something like `stripe_apple_pay`, and sepa would be `stripe_sepa` -- but what I don't see is why on the server side, these unique IDs cannot be handled together? Maybe Im missing a piece of the puzzle.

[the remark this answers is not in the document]
nerrad: Some additional info.  Currently in WooCommerce core (and payment extensions) server side:

- class-wc-gateway-stripe is used for stripe cc and stripe payment request, but stripe payment request adds details on the request type (apple-pay, chrome pay etc) via order-meta before passing off the checkout for processing.
- payment request payment method actually bypasses the checkout form submission and does it's own submission via ajax.
- Other stripe payment methods have their own gateway files (so you're correct there, having the ids just be `stripe_sepa` etc will map fine for dynamic gateway loading.

So the primary issue here is that the stripe express payment methods are a "special class" of payment method that still use the `class-wc-gateway-stripe` for the main payment processing but also add additional order meta information.

So the active payment id for both stripe cc and stripe payment request is `stripe`.  

I also think I implemented this initially because with express payment methods the "active payment method" in the checkout form might differ from the payment-method-id used for an express payment so this ensures that checkout processing is definitely using the correct id.

Even as i write this, I know the logic probably can be reworked.  I just don't have a good idea of what to rework it to yet. Maybe, if the express payment modal is closed because it's payment was successful we don't restore active payment method to the checkout payment method? Maybe we always have `payment_method` be a property of `payment_data` instead of a top-level property in the request (so express payment methods can just overwrite it) and detached from `activePaymentMethod` in the checkout state (and that prop is simply used for indicating what payment method is active in the embedded payment method options)?

src/Payments/Integrations/Stripe.php
mikejolley: Referring to `add_payment_request_order_meta` and this which appears to be doing something similar to legacy handling (updating POST and doing it's own legacy handling).

`woocommerce_rest_checkout_process_payment_with_context` was intended to be used to handle the payment itself.

src/RestApi/StoreApi/Routes/Checkout.php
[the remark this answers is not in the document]
mikejolley: > So the primary issue here is that the stripe express payment methods are a "special class" of payment method that still use the class-wc-gateway-stripe for the main payment processing but also add additional order meta information.

The confusing thing for me is why then this needs to be munged at all. If the method is `stripe`, `get_request_payment_method_id` is going to return that and validation will pass because `stripe` is valid. 

`woocommerce_rest_checkout_process_payment_with_context` is going to give you both the ID (`stripe`) and the `payment_data` (which may contain this sub-payment method ID and other data) - so why isn't that enough to handle the request and/or add additional data?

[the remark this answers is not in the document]
nerrad: > The confusing thing for me is why then this needs to be munged at all. If the method is stripe, get_request_payment_method_id is going to return that and validation will pass because stripe is valid.

it's because the payment_id coming in through the api is `payment_request` (being `$request['payment_method']` is set from `activePaymentMethod` payment side. So it's not `stripe`.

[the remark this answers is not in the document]
nerrad: To be clear, I'm in agreement that it'd be better to avoid this logic server side.  Would you be okay with me adding this as a follow-up to improve or are you wanting to handle this now?

src/Payments/Integrations/Stripe.php
nerrad: Oh gotcha, I saw that `Library::process_legacy_payment` is currently what is handling server side payments right now (and is using the same filter) so I was puzzled by why what I was doing is considered "cheating" :).  I guess ultimately, `Library::process_legacy_payment` implementation is also cheating.

So next steps would be to work out what needs to happen for payment methods in core and payment methods in extensions to remove reliance on the $_POST globals and be able to just hook into the `woocommerce_rest_checkout_process_payment_with_context` hook (as well as still support shortcode based checkout)?

Also, to clarify, the major problem we have here is the manipulation of `$_POST` right? Assuming we got rid of that dependency in the source (extensions and core) code, it's not really utilizing legacy handling right?

src/RestApi/StoreApi/Routes/Checkout.php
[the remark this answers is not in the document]
mikejolley: @nerrad Depends how hard to resolve and for when; if it's a follow up we should definitely tackle before making the API public because I wouldn't want this to ship with the final version.

> t's because the payment_id coming in through the api is payment_request (being $request['payment_method'] is set from activePaymentMethod payment side. So it's not stripe.

I see. So is it best to let the payment method change that client side before firing the request, or allow some mechanism to swap it server side? On the server side we could add some filter within `get_request_payment_method_id` that lets you change it e.g. `woocommerce_rest_checkout_request_payment_method_id`, or we could lax the validation so that any ID gets passed to `woocommerce_rest_checkout_process_payment_with_context` and if unhandled, then it throws the error.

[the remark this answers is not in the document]
mikejolley: Having `woocommerce_rest_checkout_process_payment_with_context` run even if the method is not recognised might be the best way forward here as it requires less code on the payment method side. We would just want to provide a better error response if the `woocommerce_rest_checkout_process_payment_with_context` action yields no results.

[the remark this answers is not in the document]
nerrad: >  if it's a follow up we should definitely tackle before making the API public because I wouldn't want this to ship with the final version.

I agree.  I just don't want the headache of rebasing on this pull and assuming I can get tests to pass, merging enables express checkout for testing and further work.

src/Payments/Integrations/Stripe.php
mikejolley: > So next steps would be to work out what needs to happen for payment methods in core and payment methods in extensions to remove reliance on the $_POST globals and be able to just hook into the woocommerce_rest_checkout_process_payment_with_context hook (as well as still support shortcode based checkout)?

Exactly. What we document needs to do this.

> Also, to clarify, the major problem we have here is the manipulation of $_POST right?

Yes, we really should be avoiding that going forward, passing only the data that the gateways need. That also makes it more flexible in future if we need to take payments in other contexts.

src/RestApi/StoreApi/Routes/Checkout.php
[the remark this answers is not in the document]
nerrad: > We would just want to provide a better error response if the woocommerce_rest_checkout_process_payment_with_context action yields no results

So would this be a matter of checking that `PaymentResult` reference has the expected internal state?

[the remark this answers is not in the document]
mikejolley: Possibly yes; detect if the result's status is the default so we know it wasn't touched.

What does come through as method ID though? Does it matter that it's saved to the order object? Do we need to adjust that too?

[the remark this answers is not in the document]
nerrad: Method ID is used to determine what gateway class to load for payment processing.  `payment_request_type` is specific to stripe (well at least what I've seen so far) and is used to record what specific payment_request api type was used to make the payment (apple pay, chrome pay etc) which is then displayed (I'm assuming by stripe gateway extension hooking into Woo) in various order details.

assets/js/base/context/cart-checkout/checkout/processor/index.js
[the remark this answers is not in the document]
Aljullu: I thought the `checkValidation` function as a wrap up that would make sure that we invalidate if something has errored, that's why it checks if `currentPaymentStatus.hasError` or `shippingErrorStatus.hasError` are true and returns an error in that case. Just to make sure we abort if any of those is in an error state.

But if we are confident shipping and payment methods will always hook up into `onCheckoutProcessing` and return an error there, it should be save to remove these lines, no?
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/context/cart-checkout/checkout/processor/index.js#L113-L128

In that case, yes, probably running `checkValidation` with priority `0` as you said makes more sense so it runs at the beginning.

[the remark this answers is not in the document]
nerrad: With recent changes, payment processing happens on the `onPaymentProcessing` event which fires when the checkout status is `checkoutIsProcessingComplete`.  Which is set when checkout has successfully processed.

This change was needed so that _Payment Status_ could be controlled by checkout instead of payment methods and related to the fact that event observers cannot change context state.

However shipping error and payment errors might still get picked up by the logic you added in the linked lines because the checkout processing could be reverted to "IDLE" and have to be triggered again (this likely still will need adjusted).

I think we could leave things as is for now here priority wise and we can revisit if needed as we implement more things.

PR #2130, opened 2020-04-06
Implement makeTrashable to cancel inflight requests in `CheckoutProcessor`

mikejolley:
This is for https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2100 but needs tests since the unmount case isn't easily replicatable in real world usage. Adding to backlog.

Based on https://github.com/Automattic/jetpack/pull/15228/files
Background https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html
Using package https://github.com/hjylewis/trashable

TODO:
Add test coverage to simulate unmount and confirm promise gets cancelled without errors

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -12,6 +12,7 @@ import {
 } from '@woocommerce/base-context';
 import { useEffect, useRef, useCallback } from '@wordpress/element';
 import { useStoreNotices } from '@woocommerce/base-hooks';
+import makeTrashable from 'trashable';
 
 /**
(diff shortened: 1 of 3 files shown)


senadir:
Interesting! however, as far as I understand this cancels promises, we're looking at aborting the fetch, for that we can use `AbortController`, however, I'm not sure if canceling promises cancels fetch.

see:
https://davidwalsh.name/cancel-fetch
https://stackoverflow.com/questions/30233302/promise-is-it-possible-to-force-cancel-a-promise
https://twitter.com/kentcdodds/status/1246219272562421760

haszari:
Unassigned myself as reviewer for now, happy to come back when we're ready to focus on this.

mikejolley:
Closing this stale request. Issue is still open for future attempts.

PR #2133, opened 2020-04-06
Don't show "select" labels for shipping or payment method steps if there is only one option

### LINKED ISSUE #2088: Polish: Don't show "select" labels for shipping or payment method steps if there is only one option.
There are "select..." labels in the shipping method and payment method steps that are useful if there are multiple options to select from, but seem redundant if there is only one option (and thus no ability to "select"):

We should make the display of these labels conditional on whether there is more than one option to select from.

mikejolley:
If there is only 1 shipping method, or payment method, hide the "select a X" text. This PR also hides the containing div if the description that gets passed is empty.

Fixes #2088

### Screenshots

Screenshot 2020-04-06 at 17 19 01
Screenshot 2020-04-06 at 17 18 52

1. Only enable 1 shipping method/payment method. View checkout and confirm text is hidden.
2. Enable > 1 shipping and payment method. View checkout and confirm text is displayed.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -336,7 +336,7 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 								'woo-gutenberg-products-block'
 							) }
 							description={
-								paymentMethods && paymentMethods.length > 1
+								Object.keys( paymentMethods ).length > 1
 									? __(
 											'Select a payment method below.',
 											'woo-gutenberg-products-block'
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -25,7 +25,7 @@ import {
 	useBillingDataContext,
 	useValidationContext,
 } from '@woocommerce/base-context';
-import { useStoreCart } from '@woocommerce/base-hooks';
+import { useStoreCart, usePaymentMethods } from '@woocommerce/base-hooks';
 import {
 	ExpressCheckoutFormControl,
 	PaymentMethods,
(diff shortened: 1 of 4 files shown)


assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -288,10 +290,14 @@ const Checkout = ( {
 									'Shipping options',
 									'woo-gutenberg-products-block'
 								) }
-								description={ __(
-									'Select a shipping method below.',
-									'woo-gutenberg-products-block'
-								) }
+								description={
+									shippingRates.length > 1
```
Aljullu, line 294: Correct me if I'm wrong, but `shippingRates` is an array of packages. Each package has a `shipping_rates` property which contains the list of options.

So in this case, I think we should check whether there is at least one package with more than one option. Currently, the description is hidden if there is one single package, even if it has more than one shipping option:
imatge

mikejolley:
@Aljullu should be resolved. I added some small utils to help with this.

assets/js/blocks/cart-checkout/checkout/utils.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Get the number of packages in a shippingRates array.
+ *
+ * @param {Array} shippingRates Shipping rates and packages array.
+ */
+export const getShippingRatesPackageCount = ( shippingRates ) => {
+	return Object.keys( shippingRates ).length;
```
Aljullu, line 7: I just merged it, but shouldn't this be:
```suggestion
	return shippingRates.length;
```
`shippingRates` is an array, not an object.

I will create a follow-up PR.

Aljullu: I tested and `Object.keys()` works for arrays too. In any case, I removed it in #2153.

PR #2134, opened 2020-04-06
Get preview data from Editor Context /

### LINKED ISSUE #2114: Don't include `@woocommerce/previews` in frontend builds.
In #2103 it was noticed that the frontend build for the all products block increased in size significantly. After some sleuthing by the squad, we noticed that `useStoreCart` imports `@woocommerce/previews` which include some weighty preview images.

A quick solution here would be to simply ensure that frontend builds alias `@woocommerce/previews` to a basic skeleton object instead of the data utilized in the editor. This should result in significant size improvements.

Alternatively (**and preferably**), we should pass preview data via the EditorContextProvider. The provider could receive preview data as a prop so that only editor builds will import the preview data. Then we'd just need to do a search for `@woocommerce/previews` imports and make sure any code importing that switches to implement `useEditorContext` hook for grabbing preview data instead.

Long term, we should probably also look to decreasing the size of the previews in the backend as well.

senadir:
This PRs does three things, the first (which ended up being useless for this issue but still a better refactor) simply moves all previewData to be exposed from the `EditorContextProvider` instead of getting directly from `@woocommerce/resources-previews`.
So now we import `@woocommerce/resources-previews` inside the editor context and pass it to the provider and other hooks consume that context, however this didn't solve anything since we're still importing `useEditorContext` which lives with the other imports, so I did three solutions (I reverted the second and third in favor of the fourth)

Second solution:
- I aliased resource previews to `null-loader`, however this pattern won't be supported in webpack 5, and I can't seem to get the example in webpack docs working (aliasing it to null).
this solved the issue.

Third Solution:
- I separated context definition from the provider creation, so now when I import `useContextProvider` I don't have to import preview data with it, this solved the issue but I'm not happy with the naming or config (any ideas to refactor are welcome).

Forth Solution (suggested by @nerrad):
- do things as step one, but instead of passing the preview on the provider definition, pass it as a prop in the Cart/Edit or Checkout/Edit, much cleaner solution.

Fixes #2114

Cart block didn't use any previews, Checkout did
1. Test Checkout and Cart in the editor and see nothing is broken.
2. Test them in frontend.

for this to work, we will see a reduced bundle size in the bot comment here, I tested locally and preview is no longer imported.


assets/js/base/context/editor/index.js
```diff
@@ -23,11 +25,16 @@ export const useEditorContext = () => {
...
+ * @param {Object}   props                 Incoming props for the provider.
+ * @param {*}        props.children        The children being wrapped.
+ * @param {CartData} [props.previewCart]   The preview data for editor.
+ * @param {number}   [props.currentPostId] The post being edited.
  */
-export const EditorProvider = ( { children, currentPostId = 0 } ) => {
+export const EditorProvider = ( {
+	children,
+	currentPostId = 0,
+	previewCart,
```
mikejolley, line 36: Should this have a default since it's optional?

assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -52,15 +53,9 @@ const Block = ( props ) => (
...
 
-const Checkout = ( {
-	attributes,
-	cartCoupons = [],
-	cartItems = [],
-	cartTotals = {},
-	scrollToTop,
-	shippingRates = [],
-} ) => {
+const Checkout = ( { attributes, cartCoupons = [], scrollToTop } ) => {
```
mikejolley, line 56: May as well give `cartCoupons` the same treatment here and useStoreCart

assets/js/base/context/editor/index.js
senadir: what kind of default are we hoping for? an empty object or the cart skeleton?

senadir: I went with a basic skeleton {} so that we can catch any future issue (of a component reaching for preview data when he shouldn’t).

PR #2138, opened 2020-04-07
Show "save payment information" checkbox only if customer ID > 0

### LINKED ISSUE #2118: Save payment information checkbox should only be available for logged in users.
I noticed while testing in an incognito window that we currently show the save payment information checkbox for stripe even if the user is not logged in. Obviously this isn't good (unless there's logic in place to automatically create a user - which there isn't?).

I'm not sure if woo still saves payment info for anonymous users or not (subscription management?).  I'm guessing saved payment info always requires a user account.

mikejolley:
Adds customerId to the checkout API response and checkout context so that the Stripe gateway can show the "save payment information" checkbox conditionally.

Fixes #2118

1. With stripe enabled, view the checkout whilst logged in. "save payment information" checkbox is visible.
2. repeat whilst logged out. "save payment information" is not shown.

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
@@ -13,6 +13,7 @@ import { InlineCard, CardElements } from './elements';
 import { Elements, useElements, useStripe } from '@stripe/react-stripe-js';
 import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
+import { useCheckoutContext } from '@woocommerce/base-context';
 
 /**
  * @typedef {import('../stripe-utils/type-defs').Stripe} Stripe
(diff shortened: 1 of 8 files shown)


senadir:
This is unrelated to this PR but that data is not being persisted anywhere, should it be within Billing data?

mikejolley:
@senadir What data? Are you talking about saved payment methods? If so it's within the scope of stripe gateway, not checkout.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
```diff
@@ -13,6 +13,7 @@ import { InlineCard, CardElements } from './elements';
 import { Elements, useElements, useStripe } from '@stripe/react-stripe-js';
 import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
+import { useCheckoutContext } from '@woocommerce/base-context';
```
nerrad, line 16: Don't use hooks directly in any of the payment method integrations.  All data is exposed via `use-payment-method-interface`. So the `customerId` will need to be passed through on `usePaymentMethodInterface` (and related type-defs etc will need updated).

mikejolley: Can you check again @nerrad?

We should review all data on this interface before making public just to ensure keys etc make sense.

PR #2144, opened 2020-04-07
Fix cached shipping fees on consecutive orders

### LINKED ISSUE #2139: Fix cached shipping fees on consecutive orders.
I noticed while testing the work I'm doing in #2127 that it appears shipping fees are caching between consecutive orders.

To reproduce:

- ~I don't know if this is only happening for logged in users or not~, Happens for me with both logged in and non-logged in users (incognito mode).
- Complete a purchase with credit card or cheque for a product that requires shipping.
- Make note of the selected shipping and the order-received on finishing checkout.
- Add an item to the cart that doesn't require shipping. 
- In checkout, take note the totals etc that show shipping is not added to the total for this order.
- Complete checkout for the order using credit card or cheque.
- 🐛note that the order received page shows the selected shipping rate from the _previous_ order in the just completed order.

I haven't tried reproducing this on master yet, but thought I'd get the issue up while I'm finishing off the express payment pull to see if anyone else can reproduce.

mikejolley:
Fixes shipping method saving on new orders, due to missing shipping calculation when shipping is not needed.

At the same time, I found a bug where the customer data was also not being saved to the order which is fixed in this PR also.

Fixes #2139

I also took this opportunity to sync some of our code to core - see https://github.com/woocommerce/woocommerce/pull/26101 I've updated todos to match.

1. Follow instructions on https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2139
2. Checkout and ensure billing/shipping fields have values from your customer when logged in.

### DIFF
--- a/src/RestApi/StoreApi/Utilities/CartController.php
+++ b/src/RestApi/StoreApi/Utilities/CartController.php
@@ -272,7 +272,7 @@ class CartController {
 	 * @return int
 	 */
 	protected function get_remaining_stock_for_product( $product ) {
-		// @todo Remove once min support for WC reaches 4.0.0.
+		// @todo Remove once min support for WC reaches 4.1.0.
 		if ( \class_exists( '\Automattic\WooCommerce\Checkout\Helpers\ReserveStock' ) ) {
 			$reserve_stock_controller = new \Automattic\WooCommerce\Checkout\Helpers\ReserveStock();
 		} else {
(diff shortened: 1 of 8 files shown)


src/RestApi/StoreApi/Utilities/CartController.php
```diff
@@ -364,25 +364,29 @@ public function get_cart_coupons( $callback = null ) {
...
+			if ( ! isset( $packages[ $key ]['package_id'] ) ) {
+				$packages[ $key ]['package_id'] = $key;
 			}
 		}
 
+		// See if we need to calculate anything.
+		if ( ! WC()->cart->needs_shipping() ) {
+			return [];
+		}
+
```
Aljullu, line 389: Could this if clause be moved to the top of the function so we return early if it doesn't need shipping?

PR #2146, opened 2020-04-07
Refactor some of the shipping hooks/context usage

mikejolley:
This is a follow on from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2133 so some code is duplicated until that is merged and this is rebased.

- Removes `useShippingRates` which had very limited value after other refactors
- Ensure when using `useShippingAddress` it comes from `useShippingDataContext`, otherwise data is not in sync
- Moved some data from `useShippingRates` to `useStoreCart`
- Made shipping calculator use context directly, and shipping location use `useStoreCart`

Context should be used for _current, possibly not persisted, values_. That is, if you're updating an address, use context. This will then update the address and return new rates behind the scenes.

`useStoreCart` contains retrieved/persisted values. So if you are displaying rates and an address, this is what you want.

1. Test shipping calculator in cart
2. Update address in checkout - view sidebar
3. Smoke test anything shipping related in cart/checkout


mikejolley:
@senadir yup probably lots of improvements can be made beyond this. My goal with this PR however was to clear up where the cart calculator and other components get data as it wasn't consistent.

PR #2153, opened 2020-04-08
Use new util functions in ShippingRatesControl

Aljullu:
#2133 introduced two new util functions that can be used in the `ShippingRatesControl` component to remove some logic.

1. Preview the _Cart_ and _Checkout_ blocks and verify there are no regressions.

### DIFF
--- a/assets/js/base/components/cart-checkout/shipping-rates-control/index.js
+++ b/assets/js/base/components/cart-checkout/shipping-rates-control/index.js
@@ -11,6 +11,10 @@ import LoadingMask from '@woocommerce/base-components/loading-mask';
  * Internal dependencies
  */
 import Packages from './packages';
+import {
+	getShippingRatesPackageCount,
+	getShippingRatesRateCount,
+} from '../../../utils/shipping-rates';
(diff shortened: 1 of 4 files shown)


assets/js/base/components/cart-checkout/shipping-rates-control/index.js
```diff
@@ -11,6 +11,10 @@ import LoadingMask from '@woocommerce/base-components/loading-mask';
  * Internal dependencies
  */
 import Packages from './packages';
+import {
+	getShippingRatesPackageCount,
```
senadir, line 15: I don’t know, I feel the variable name, and the extra hassle of importing it and assigning it and remembering what it does and where is much larger than the actual abstraction by itself, what do you think?

Aljullu: Yeah, I think it's a way to abstract the fact that the variable is named `shippingRates` but in fact it's an array of packages. I don't have any strong opinion on this but would like to hear @mikejolley's thoughts.

PR #2156, opened 2020-04-08
Handle errors in processOrder

### LINKED ISSUE #2059: CheckoutProcessor - Update success/error response handling
Follow up from #2044

When the `CheckoutProcessor` component has made it's requests, we need to handle the success/error and either ensure a redirect occurs, or errors are surfaced correcrly.

Aljullu:
Fixes #2059.

### Screenshots
imatge

There are several errors that might occur when processing an order, for each one of them, we need to test that the user is not redirected to the success page and that the appropiate error appears on top of the _Checkout_ block.

**Error 1**
Path not available for some reason. You can trigger it modifying [L149](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/context/cart-checkout/checkout/processor/index.js#L149) from `context/cart-checkout/checkout/processor/index.js`:
```diff
-path: '/wc/store/checkout',
+path: '/wc/store/wrong-path',
```

**Error 2**
There was an error while parsing the JSON returned by the server. You can trigger it adding these lines in [L167](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/context/cart-checkout/checkout/processor/index.js#L167) from `context/cart-checkout/checkout/processor/index.js`:
```JS
fetchResponse = {
	json: Promise.reject(),
};
```

**Error 3**
The response was not ok. You can trigger it adding these lines in [L170](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/context/cart-checkout/checkout/processor/index.js#L170) from `context/cart-checkout/checkout/processor/index.js`:
```JS
fetchResponse = {
	ok: false,
};
```

For each of these errors:
1. Add the code that will trigger the error.
2. Make a purchase.
3. Verify the error appears and you were not redirected to the success page.

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -176,7 +176,7 @@ const CheckoutProcessor = () => {
 						} else {
 							addErrorNotice(
 								__(
-									'Something went wrong. Please check your payment details and try again.',
+									'Something went wrong. Please contact us to get assistance.',
 									'woo-gutenberg-products-block'
 								),
 								{
@@ -199,7 +199,7 @@ const CheckoutProcessor = () => {
 				const message =
 					error.message ||
 					__(
-						'Something went wrong. Please contact the site administrator to get assistance.',
+						'Something went wrong. Please contact us to get assistance.',
 						'woo-gutenberg-products-block'
 					);
 				addErrorNotice( message, {
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -10,7 +10,7 @@ import {
 	usePaymentMethodDataContext,
 	useValidationContext,
 } from '@woocommerce/base-context';
-import { useEffect, useRef, useCallback } from '@wordpress/element';
+import { useEffect, useRef, useCallback, useState } from '@wordpress/element';
 import { useStoreNotices } from '@woocommerce/base-hooks';
 
 /**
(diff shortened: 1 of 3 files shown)


assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -182,23 +185,33 @@ const CheckoutProcessor = () => {
...
 					dispatchActions.setComplete();
+					setIsProcessingOrder( false );
 				} );
 			} )
 			.catch( ( error ) => {
-				addErrorNotice( error.message, {
+				const message =
+					error.message ||
+					__(
+						'Something went wrong. Please contact the site administrator to get assistance.',
```
Aljullu, line 202: Here, I rewrote the text to suggest contacting the site administrator instead of "try again" (as suggested in an error some lines above :point_up: ). If I'm understanding it correctly, the payment has already been done but the purchase could not be completed, so we shouldn't suggest users to purchase it again, otherwise they would pay twice. Does that make sense?

mikejolley: Since this is customer facing we should say "Please contact us to get assistance" rather than "Please contact the site administrator to get assistance" which is pretty unfriendly.

Aljullu: Good point. I will change it. :+1: 

What about this one:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/context/cart-checkout/checkout/processor/index.js#L176
would make it sense changing it too?

mikejolley: Depends if it's fixable by the user :p

Aljullu: Even if the user can fix it, when `processOrder` runs, the payment has already been done, no? In that case I think we shouldn't suggest the user to try again to avoid the payment being processed again.

Aljullu: I changed the text of the errors in: 49980c291c8ff30f2537dc9a3c138692cb54d4b5.

PR #2160, opened 2020-04-08
Update previews and editor styles for cart and checkout blocks

### LINKED ISSUE #1487: Show sample cart content in editor and block inserter preview
The goal of this issue is that the cart block shows a realistic representation of a full cart:

- In the block picker preview pane
- In the block "full cart" mode when editing

This is somewhat underway, the full cart currently shows sample data from `previews/cart-items.js`. Before launch we'll need to ensure the preview looks good and may want to tweak the sample data.

This has some prerequisites as noticed in the dependencies.

~~The goal here is that if the store has an existing product, then use that for the cart preview in the editor. Otherwise use a sample product to power the preview.~~ **Update: we're using sample data, see discussion in comments**

There may be some extra considerations here:

- Should we only display preview for certain product types?
- Should we narrow the scope for the initial release and just use a sample product for now (cc @pmcpinto , @garymurray )? Based on feedback for the corresponding [checkout block issue](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1479) I'm guessing we'll want to just do sample content for this initial iteration too? **See below - yes, limiting scope to sample data for now**

mikejolley:
This PR updates cart and checkout block previews.

- Fixes some display issues in the editor due to inheriting default styles
- Splits up the cart block into a block.js file to follow the patterns in checkout block and others
- Renamed `empty-cart` to `empty-cart-edit` so it's clear this is the editor side only
- Adds `isPreview` attribute to cart and checkout blocks so preview can be rendered
- Adds new SVGs for the 2 new previews
- Updates cart sample data to include correct images and totals data.

Fixes #1487

### Screenshots

Cart preview:
Screenshot 2020-04-08 at 14 17 46

Checkout preview:
Screenshot 2020-04-08 at 14 17 58

Cart data:
Screenshot 2020-04-08 at 14 21 37

1. Check the reviews by going to add a cart/checkout block via the editor
2. Smoke test cart editor and frontend cart to ensure the block still functions


assets/js/base/components/cart-checkout/totals/totals-coupon-code-input/index.js
```diff
@@ -21,7 +21,7 @@ import './style.scss';
 const TotalsCouponCodeInput = ( {
 	instanceId,
 	isLoading = false,
-	initialOpen,
+	initialOpen = false,
```
senadir, line 24: why are we closing this by default?

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -164,6 +165,10 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
  * Component to handle edit mode of "Cart Block".
  */
 const CartEditor = ( { className, attributes, setAttributes } ) => {
+	if ( attributes?.isPreview ) {
```
senadir, line 168: optional chaining shouldn’t be needed here, attributes should always be available (as an empty object) you can also give it a default value `{}`.

assets/js/blocks/cart-checkout/checkout/edit.js
```diff
@@ -268,6 +271,11 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 
 const CheckoutEditor = ( { attributes, setAttributes } ) => {
 	const { className } = attributes;
+
+	if ( attributes?.isPreview ) {
```
senadir, line 275: you can destruct `isPreview` from attributes as well, it will return undefined if it's not set.

assets/js/base/components/cart-checkout/totals/totals-coupon-code-input/index.js
mikejolley: It should be closed by default like on frontend. In the editor it was visible.

PR #2169, opened 2020-04-08
Use cart data to provide country and currency_code for payment method registration config

nerrad:
Fixes #2016.

Since cart data is already available on the settings object and it is populated with values even when the cart is empty - we can pass that data along to payment methods `canMakePayment` config option so it can use the data as a part of the logic for determining whether a payment method can make payment or not.  So in this pull:

- refactored payment method api registration so that `canMakePayment` is expected to be a function. The function receives a `cartData` object and `usePaymentMethodRegistration` will handle feeding cart data and also wrapping the return value of `canMakePayment` in a promise so payment methods no longer have to return a promise (super handy for offline payment methods like Cheque).
- Stripe payment request express payment method now uses this `cartData` object to obtain shippingAddress.country and cartTotals.currency_code to create the stripe `paymentRequest` object used for determining whether payment can be made or not.

## To Test

This impacts all current implemented payment methods so testing needs to validate they still complete purchases without issue. It's likely that googlepay payment method should show up better for team members not in Canada (what things were hardcoded to before). @senadir interested in knowing if this improves things for your testing.


senadir:
Still not working but this time it is more helpful than before, I got this error:
```
Uncaught (in promise) Error: Problem with payment method initializationInvalid value for paymentRequest(): country should be one of the following strings: AE, AT, AU, BE, BR, CA, CH, CZ, DE, DK, EE, ES, FI, FR, GB, GR, HK, IE, IN, IT, JP, LT, LU, LV, MX, MY, NL, NO, NZ, PH, PL, PT, RO, SE, SG, SI, SK, US. You specified: DZ.
```

Google pay does support Algeria for things like online payments

```
With participating websites and apps, you can use this Google Pay feature in Albania, Algeria, American Samoa, Angola, Antigua and Barbuda, Argentina, Australia, Austria, Azerbaijan, Bahrain, Belarus, Belgium, Brazil, Bulgaria, Canada, Chile, Colombia, Croatia, Czechia, Denmark, Dominican Republic, Egypt, Estonia, Finland, France, Germany, Greece, Hong Kong, Hungary, India, Indonesia, Ireland, Israel, Italy, Japan, Jordan, Kazakhstan, Kenya, Kuwait, Lebanon, Lithuania, Luxembourg, Malaysia, Mexico, Netherlands, New Zealand, Norway, Oman, Pakistan, Panama, Peru, Philippines, Poland, Portugal, Qatar, Romania, Russian Federation, Saudi Arabia, Singapore, Slovakia, South Africa, South Africa, Spain, Sri Lanka, Sweden, Switzerland, Taiwan, Thailand, Turkey, Ukraine, United Arab Emirates, United Kingdom of Great Britain and Northern Ireland, United States of America, Uruguay, and Vietnam.
```

So my thinking so far, we should pass the merchant country, not the customer, when I changed my shipping country to US for example, it still won't work but with more helpful message (I need to enable HTTPS, even on localhost)

```
You may test your Stripe.js integration over HTTP. However, live Stripe.js integrations must use HTTPS.
```

so I will test on a live website and see.

senadir:
so I went ahead and tested on a live site and google pay still didn't work, Mike confirmed the same issue as well
https://ephemeral-senadir-20200408.atomicsites.blog/checkout-block/

for comparison, we also tested on https://google-pay-demo.glitch.me/ to confirm that google pay works for us, and it does on that platform.

nerrad:
> So my thinking so far, we should pass the merchant country, not the customer, when I changed my shipping country to US for example, it still won't work but with more helpful message (I need to enable HTTPS, even on localhost).

I currently have it using the same as the current payment extension: 

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/99318bbbdbe101a8a4b6b9d4a0d5613ee0960665/assets/js/stripe-payment-request.js#L336-L346

So the country will be whatever the default country is set for customers in your site settings. I don't want to diverge from what the current payment extension is doing because I'm assuming that team has worked out what it should be.

nerrad:
Just noting that @mikejolley was able to confirm GooglePay worked on my test site: http://ephemeral-nerrad-20200302.atomicsites.blog/

Issue #2168, opened 2020-04-08
Handle express payment `active payment` toggle via checkout instead of express payment method.


nerrad:
Currently express payment methods need to take care of setting active payment method when their payment request button is clicked and also restore active payment method when the payment request modal is closed. Ideally this logic should be moved to be handled by checkout so that the express payment method doesn't need to handle it.

Also, for this issue I'd also refactor the current implementation of deriving server side payment id for express payment methods from the `paymentMethodData` (see [this comment thread](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2127#discussion_r404891414). With moving setting active payment method to checkout, this should be trivial to revert and improve.

PR #2170, opened 2020-04-08
Refactor logic for handling active payment method with express payment methods via checkout

nerrad:
Fixes: #2168 

This pull addresses a couple things:

## Handling setting status and active payment method via checkout for express payment methods.

While this improves things so express payment methods don't have to worry about state changes etc in checkout, they still will need to implement the provided `onClick` and `onClose` handlers provided via checkout. 

## Explicitly provide a `paymentMethodId` that is separate from `activePaymentMethod`

The configuration object for registered payment methods now includes a `paymentMethodId` property and the existing `id` property was renamed to `name`.  This new property allows for payment methods to indicate a value to use in the request as the payment method identifier when it's different than the `name` property (which is what it will default to if `paymentMethodId` is not provided).  

I also improved some of the stripe server side things so that we only modify ordermeta for stripe payment_request packages.

## To test

This impacts all payment method processing so testing needs to cover cheque, stripe cc, and stripe payment request (express payment) purchases.  Ensure the flow works as expected and there are no console errors. Also ensure the order details show the correct expected values.

### DIFF
--- a/src/Library.php
+++ b/src/Library.php
@@ -31,7 +31,7 @@ class Library {
 		add_filter( 'woocommerce_register_shop_order_post_statuses', array( __CLASS__, 'register_draft_order_post_status' ) );
 		add_filter( 'woocommerce_valid_order_statuses_for_payment', array( __CLASS__, 'append_draft_order_post_status' ) );
 		add_action( 'woocommerce_cleanup_draft_orders', array( __CLASS__, 'delete_expired_draft_orders' ) );
-		add_action( 'woocommerce_rest_checkout_process_payment_with_context', array( __CLASS__, 'process_legacy_payment' ), 10, 2 );
+		add_action( 'woocommerce_rest_checkout_process_payment_with_context', array( __CLASS__, 'process_legacy_payment' ), 999, 2 );
 	}
 
 	/**
@@ -216,7 +216,7 @@ class Library {
 	 * @param PaymentContext $context Holds context for the payment.
 	 * @param PaymentResult  $result  Result of the payment.
 	 */
-	public static function process_legacy_payment( PaymentContext $context, PaymentResult &$result ) {
+	public static function process_legacy_payment( PaymentContext $context, PaymentResult $result ) {
 		if ( $result->status ) {
 			return;
 		}
--- a/src/Library.php
+++ b/src/Library.php
@@ -216,7 +216,7 @@ class Library {
 	 * @param PaymentContext $context Holds context for the payment.
 	 * @param PaymentResult  $result  Result of the payment.
 	 */
-	public static function process_legacy_payment( PaymentContext $context, PaymentResult $result ) {
+	public static function process_legacy_payment( PaymentContext $context, PaymentResult &$result ) {
 		if ( $result->status ) {
 			return;
 		}
--- a/src/Payments/Integrations/Stripe.php
+++ b/src/Payments/Integrations/Stripe.php
@@ -194,7 +194,6 @@ final class Stripe extends AbstractPaymentMethodType {
 			// phpcs:ignore WordPress.Security.NonceVerification
 			$post_data = $_POST;
 			$_POST     = $context->payment_data;
-			$context->set_payment_method( 'stripe' );
 			WC_Stripe_Payment_Request::add_order_meta( $context->order->id, $context->payment_data );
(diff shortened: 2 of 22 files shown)


src/Payments/Integrations/Stripe.php
```diff
@@ -189,11 +189,14 @@ private function get_button_locale() {
...
 	 */
 	public function add_payment_request_order_meta( PaymentContext $context, PaymentResult $result ) {
-		// phpcs:ignore WordPress.Security.NonceVerification
-		$post_data = $_POST;
-		$_POST     = $context->payment_data;
-
-		WC_Stripe_Payment_Request::add_order_meta( $context->order->id, $context->payment_data );
-		$_POST = $post_data;
+		$data = $context->payment_data;
+		if ( $data['payment_request_type'] && 'stripe' === $context->payment_method ) {
```
mikejolley, line 193: is `payment_request_type` used anymore?

src/Payments/Integrations/Stripe.php
```diff
@@ -189,11 +189,14 @@ private function get_button_locale() {
...
+			// phpcs:ignore WordPress.Security.NonceVerification
+			$post_data = $_POST;
+			$_POST     = $context->payment_data;
+			$context->set_payment_method( 'stripe' );
```
mikejolley, line 197: You already checked this in the parent conditional - it was already `stripe`

src/Library.php
```diff
@@ -216,7 +216,7 @@ public static function delete_expired_draft_orders() {
 	 * @param PaymentContext $context Holds context for the payment.
 	 * @param PaymentResult  $result  Result of the payment.
 	 */
-	public static function process_legacy_payment( PaymentContext $context, PaymentResult &$result ) {
+	public static function process_legacy_payment( PaymentContext $context, PaymentResult $result ) {
```
mikejolley, line 219: Just checking; did you test that legacy payment methods could still append things to $result and that was seen by the caller?

src/Library.php
```diff
@@ -31,7 +31,7 @@ public static function init() {
 		add_filter( 'woocommerce_register_shop_order_post_statuses', array( __CLASS__, 'register_draft_order_post_status' ) );
 		add_filter( 'woocommerce_valid_order_statuses_for_payment', array( __CLASS__, 'append_draft_order_post_status' ) );
 		add_action( 'woocommerce_cleanup_draft_orders', array( __CLASS__, 'delete_expired_draft_orders' ) );
-		add_action( 'woocommerce_rest_checkout_process_payment_with_context', array( __CLASS__, 'process_legacy_payment' ), 10, 2 );
+		add_action( 'woocommerce_rest_checkout_process_payment_with_context', array( __CLASS__, 'process_legacy_payment' ), 999, 2 );
```
mikejolley, line 34: Good call 👍 

Reminded me of https://www.reddit.com/r/ProgrammerHumor/comments/58zwnu/the_only_way_to_deal_with_zindex/

src/Payments/Integrations/Stripe.php
nerrad: Yes, it's passed through via the `$_POST` switcheroo to the stripe extension. I just check if it's present here first to determine whether the switcheroo is needed.

src/Payments/Integrations/Stripe.php
nerrad: oops, leftover from a previous iteration of this pull where I was trying to just detect if `payment_request` was the `payment_method` value and then switch to `stripe` (which didn't work because payment_method is validated earlier than the hook (for good reason).

src/Library.php
nerrad: woops another artifact from initial iterations inthis pull, I put back.

src/Payments/Integrations/Stripe.php
mikejolley: In that case, check if `! empty` or it will show a notice for other gateways.

mikejolley:
```suggestion
		if ( ! empty( $data['payment_request_type'] ) && 'stripe' === $context->payment_method ) {
```

nerrad: Oh ya, missed that, will add.

PR #2175, opened 2020-04-09
Update cart after conflicts

### LINKED ISSUE #2131: Trying to update or delete a non-existing cart Item will leave you with stale data.
**Describe the bug**
When you try to delete or update the quantity of a cart item that is no longer in your cart (delete somewhere else), no error will be shown, and the item will still be in your cart, the total will also reflect that, however if this errors happens, chances are your Cart is already stale and should be updated, a refresh fixes this, but the server should return the cart data nevertheless in those cases, and possibly with the error in it as well?

endpoints:
```
cart/update-item
cart/delete-item
``` 

**To Reproduce**

1. Open a Cart block with some products in it.
2. Open the same block or classic cart in another tab.
3. Delete an element there.
4. try to delete it in the original tab.
5. it will fade but it will back to normal and won't delete.
6. no errors are shown, the total is also not updated.

**Expected behavior**
The item should be updated anyway and the total should be reflected.

mikejolley:
I realise there is no standard recommending this, however, it seems like an acceptable solution to our problem. If you try to update something in cart which is gone/no longer exists, this can be due to conflicting local state within the client; therefore one could argue a 409 response is suitable:

> The HTTP 409 Conflict response status code indicates a request conflict with current state of the server.

With this 409 response, as part of the error data, I'm appending the _current_ state, which allows the cart to update and reconcile any differences.

Fixes #2131

1. Open up a full cart in two tabs.
2. Remove an item from one tab.
3. Remove the same item from the other tab.

In both instances the cart should now update and the item be removed. If you test the endpoint directly and try to remove a non-existing item, you'll get a response like this:

```json
{
  "code": "woocommerce_rest_cart_invalid_key",
  "message": "Cart item no longer exists or is invalid.",
  "data": {
    "status": 409,
    "cart": {
      "coupons": [],
      "shipping_rates": []...
```

### DIFF
--- a/src/RestApi/StoreApi/Routes/CartRemoveCoupon.php
+++ b/src/RestApi/StoreApi/Routes/CartRemoveCoupon.php
@@ -60,9 +60,14 @@ class CartRemoveCoupon extends AbstractCartRoute {
 		$controller  = new CartController();
 		$cart        = $controller->get_cart_instance();
 		$coupon_code = wc_format_coupon_code( $request['code'] );
+		$coupon      = new \WC_Coupon( $coupon_code );
+
+		if ( $coupon->get_code() !== $coupon_code || ! $coupon->is_valid() ) {
+			throw new RouteException( 'woocommerce_rest_cart_coupon_error', __( 'Invalid coupon code.', 'woo-gutenberg-products-block' ), 403 );
+		}
 
 		if ( ! $controller->has_coupon( $coupon_code ) ) {
-			throw new RouteException( 'woocommerce_rest_cart_coupon_invalid_code', __( 'Coupon no longer exists or is invalid.', 'woo-gutenberg-products-block' ), 409 );
+			throw new RouteException( 'woocommerce_rest_cart_coupon_invalid_code', __( 'Coupon is not applied to the cart.', 'woo-gutenberg-products-block' ), 409 );
 		}
 
 		$cart = $controller->get_cart_instance();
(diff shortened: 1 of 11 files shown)


src/RestApi/StoreApi/Routes/CartRemoveCoupon.php
```diff
@@ -69,9 +60,14 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
...
 		$coupon_code = wc_format_coupon_code( $request['code'] );
+		$coupon      = new \WC_Coupon( $coupon_code );
+
+		if ( $coupon->get_code() !== $coupon_code || ! $coupon->is_valid() ) {
+			throw new RouteException( 'woocommerce_rest_cart_coupon_error', __( 'Invalid coupon code.', 'woo-gutenberg-products-block' ), 403 );
+		}
 
 		if ( ! $controller->has_coupon( $coupon_code ) ) {
-			throw new RouteException( 'woocommerce_rest_cart_coupon_invalid_code', __( 'Coupon does not exist in the cart.', 'woo-gutenberg-products-block' ), 404 );
+			throw new RouteException( 'woocommerce_rest_cart_coupon_invalid_code', __( 'Coupon is not applied to the cart.', 'woo-gutenberg-products-block' ), 409 );
```
nerrad, line 70: The wording is better than what it was but it still feels kinda clunky. Maybe `Coupon did not apply to the cart.`?

mikejolley: I will reword for clarity but this is the remove endpoint, so it's not that it did not apply, but that it cannot be removed because it was not already applied.

senadir:
In theory, this PR should also have fixed https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2213 right?

PR #2179, opened 2020-04-09
Fix entities in coupon validation error messages

### LINKED ISSUE #2164: Coupon codes with apostrophes don't work
**To Reproduce**
1. Create a coupon code with an apostrophe, for example `don'tpayshipping`.
2. In the _Cart_ block, try using that coupon.
3. An error appears: `Coupon "don\&#039;tpayshipping" does not exist!`.

**Expected behavior**
The coupon should be correctly applied, like it is in the _Cart_ shortcode.

**Screenshots**
imatge

mikejolley:
Fixes the entities in validation error messages.

The coupon issue with ' characters didn't break for me, but it could have been missing unslash. Paging @Aljullu to retest with this branch.

Fixes #2164

1. Create coupon called `won't apply`
2. Apply it. Should work.
3. Apply `won't apply2`. Error message should be correctly formatted.

### DIFF
--- a/assets/js/base/components/validation/validation-input-error.js
+++ b/assets/js/base/components/validation/validation-input-error.js
@@ -3,6 +3,7 @@
  */
 import { useValidationContext } from '@woocommerce/base-context';
 import PropTypes from 'prop-types';
+import { decodeEntities } from '@wordpress/html-entities';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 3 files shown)


assets/js/base/components/validation/validation-input-error.js
```diff
@@ -26,7 +27,9 @@ export const ValidationInputError = ( {
 
 	return (
 		<div className="wc-block-form-input-validation-error" role="alert">
-			<p id={ getValidationErrorId( elementId ) }>{ errorMessage }</p>
+			<p id={ getValidationErrorId( elementId ) }>
+				{ decodeEntities( errorMessage ) }
```
Aljullu, line 31: What would you think on decoding it in `useStoreCartCoupons`:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/hooks/cart/use-store-cart-coupons.js#L62

This way we already store the error message decoded.

PR #2185, opened 2020-04-09
Fix currency format in On Sale badge

### LINKED ISSUE #2155: Cart block: `Save %d` doesn't respect currency format
**To Reproduce**
1. In your currency settings, set the decimal separator to comma `,`.
2. Go to the _Cart_ block with a product on sale.
2. Notice the `Save 1.82€!` shows the value separated with a dot instead of a comma.

**Expected behavior**
The value in the save badge should use the correct price format.

**Screenshots**
imatge

Aljullu:
Fixes #2155.

1. In your currency settings, set the decimal separator to comma `,`.
2. Go to the _Cart_ block with a product on sale.
2. Verify the `Save 1,23€!` badge shows the value separated with a comma instead of a dot.

### DIFF
--- a/assets/js/base/components/cart-checkout/product-sale-badge/index.js
+++ b/assets/js/base/components/cart-checkout/product-sale-badge/index.js
@@ -1,8 +1,9 @@
 /**
  * External dependencies
  */
-import { __, sprintf } from '@wordpress/i18n';
-import { formatPrice } from '@woocommerce/base-utils';
+import { __experimentalCreateInterpolateElement } from 'wordpress-element';
+import { __ } from '@wordpress/i18n';
+import FormattedMonetaryAmount from '@woocommerce/base-components/formatted-monetary-amount';
 import PropTypes from 'prop-types';
 
 /**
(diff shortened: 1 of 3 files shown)


assets/js/base/components/cart-checkout/product-sale-badge/index.js
```diff
@@ -25,10 +26,16 @@ const ProductSaleBadge = ( { currency, saleAmount } ) => {
 	}
 	return (
 		<div className="wc-block-sale-badge">
-			{ sprintf(
-				/* translators: %s discount amount */
-				__( 'Save %s!', 'woo-gutenberg-products-block' ),
-				formatPrice( saleAmount, currency )
+			{ __experimentalCreateInterpolateElement(
+				__( 'Save <price></price>!', 'woo-gutenberg-products-block' ),
```
nerrad, line 30: You should be able to just do a self-closing `<price />` here.  Also might be good to add a translators comment just to make it clear that `<price/>` should not be translated.

PR #1949, opened 2020-03-11
CSS editor fixes for Cart and Checkout blocks

Aljullu:
#### Cart
_Before:_
imatge

_After:_
imatge

#### Checkout
_Before:_
imatge

_After:_
imatge

### How to test the changes in this Pull Request:

1. Open the _Cart_ block in the editor and verify buttons have the proper background color and text is properly aligned.
2. Open the _Checkout_ block in the editor and verify there isn't a white box on top of the _First name_ field border.

PR #2186, opened 2020-04-09
Show feedback on buttons/checkout when processing

### LINKED ISSUE #2172: Show feedback (e.g. progress or spinner) when submitting a valid order
When the order is submitted by clicking checkout `Place order` button, there's no feedback. We should disable the submit button and form, and show some kind of loading feedback.

mikejolley:
- Allows buttons to display a custom spinner when loading. I did not use `<spinner>` here because I wanted a different animation that would work better on buttons.
- Updates "Proceed to checkout", "Apply coupon", and "Place order" buttons to use spinners
- "Place order" button reflects checkout state and shows a checkmark when done before the redirect.
- Fields on checkout are disabled whilst processing is active.

Fixes #2172

1. Go to checkout from the cart - button should show spinner once clicked
2. Apply a coupon. Button shows a spinner whilst loading
3. Place order. Watch button matches above gifs.


assets/js/base/components/text-input/style.scss
```diff
@@ -75,3 +75,7 @@
 		margin-bottom: 0;
 	}
 }
+
+fieldset:disabled {
```
Aljullu, line 79: Could this selector target `.wc-block-checkout-step` instead of the more broad `fieldset`? It could then be added to `<FormStep>`'s style. I think we should avoid tag selectors as much as possible, to prevent affecting styles not related to our blocks.

assets/js/blocks/cart-checkout/checkout/style.scss
```diff
@@ -125,6 +125,15 @@
 		border-radius: 3px;
 		height: auto;
 		margin-left: auto;
+
+		.wc-block-components-button__text {
+			line-height: 24px;
+
+			svg {
```
Aljullu, line 132: Again, I'm a bit wary of using tag selectors. :slightly_smiling_face:  Could we use a child combinator here? This way, if some day we want to add an inner component which has other `svg` elements, they would not be affected by this rule.
```suggestion
			> svg {
```

haszari:
Awesome, spinner looks ace!

Just a thought about components. A button with spinner state seems like a very reusable component - so this is an important precedent :) Definitely worth considering from a design point of view and also if we can share this for reuse elsewhere in Woo (i.e. a woo component) or if there is an exisiting component we can use. In that sense, moving to `base/components` seems appropriate to tm.

mikejolley:
> Just a thought about components. A button with spinner state seems like a very reusable component - so this is an important precedent :) Definitely worth considering from a design point of view and also if we can share this for reuse elsewhere in Woo (i.e. a woo component) or if there is an exisiting component we can use. In that sense, moving to base/components seems appropriate to tm.

I think we should wait for use cases to appear tbh, so we know whats needed for generalisation :)

Thanks for the love.

mikejolley:
> I noticed you moved the <Button> component back to base/components. In #1949 (comment) we considered moving it to cart-checkout. I don't have a strong opinion on this because I think it's too early to know if it's going to be used in other blocks or not, I just want to be wary that we don't start moving it around from one directory to the other and back to the first one.

@Aljullu Thanks for feedback. I moved it up a level because it's shared by a few buttons now. Ill move it back and merge - we can revisit when needed.

PR #2188, opened 2020-04-09
Use icons directly from stripe

### LINKED ISSUE #2064: Stripe: Pull cc images from gateway.
Currently I'm using a (large) svg for the stripe payment gateway accepted cc card image. Instead this should be pulled from the payment extension (see the [`get_icon` method](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/14bb72f0028c65dd64343253ab56764882d3cb39/includes/class-wc-gateway-stripe.php#L195-L217))

senadir:
Fixes #2064

This uses icons directly from stipe gateway, however they don't go over the same filter as the classic checkout, so we might want to account for adding filter in the future.

I went with an array of objects instead of an object since this will have arbitrary keys (name of cc).

1. In checkout see that we're using individual icons instead of a single svg.
2. weight drop.


senadir:
it's misleading how the bundle action calculates the percentage from the new size, should be from the old one to the new, so 91% reduced.

assets/js/base/components/payment-methods/style.scss
```diff
@@ -150,6 +150,17 @@
...
+
+	.stripe-icon {
+		height: 18px;
+		width: auto;
+		margin-right: $gap-small;
+
+		&:last-child {
+			margin-right: 0;
+		}
+	}
```
nerrad, line 163: The name of the class is too specific to payment method. Anything in our components should be generic so that it can be used by other payment methods.  So instead of `.stripe-icon` maybe call it `.wc-blocks-credit-cart-icon` (and then if payment methods want to get the styles for free, they use that class)?

src/Payments/Integrations/Stripe.php
```diff
@@ -181,6 +182,53 @@ private function get_button_locale() {
...
+	 *
+	 * @return array Array containing arrays of icons metadata.
+	 */
+	private function get_icons() {
+		$icons_src = [
+			[
+				'id'  => 'visa',
+				'url' => WC_STRIPE_PLUGIN_URL . '/assets/images/visa.svg',
+				'alt' => __( 'Visa', 'woo-gutenberg-products-block' ),
+			],
```
nerrad, line 196: I'm not sure why this has to be an array of arrays. I think it'd be better to do:

```js
$icons_src = [
    'visa' => [ 'url' => WC_STRIPE_PLUGIN_URL . '/assets/images/visa.svg', 'alt' => __( 'Visa', 'woo-gutenberg-products-block' ) ]
];
```

I know you give the reason:

> I went with an array of objects instead of an object since this will have arbitrary keys (name of cc).

...but I'm not sure why that matters.  Having the id as the index:

- ensures there's only ever ONE icon for each company. 
- makes it easier for extensions to filter and replace images for a given icon (if/when we add filtering in the future).

I may not understand your rationale so I'm open to more details there, but based on what I see I'd like to see the shape changed.

assets/js/base/components/payment-methods/style.scss
senadir: I went with the same classes that are used by that the gateway to ensure compatibility but I can change that.

src/Payments/Integrations/Stripe.php
senadir: this was the first case I went with and. I had the same reasons in mind, then I went on to write the types for it and I kinda felt that a key should probably be a constant name related to that data, not the part of the actual data, however, I don’t have any strong feelings about this and I'm happy to revert it.

assets/js/base/components/payment-methods/style.scss
nerrad: > I went with the same classes that are used by that the gateway

That makes sense, but the reason I'm asking for the change is:

- we're intentionally trying to keep the gateway specific stuff out of the blocks plugin and limited to the payment method builds. This makes it easier to do an extraction when we move the payment method to the relevant extension.
- we want to avoid gateway specific styles and incorporate styles that _could_ be implemented by any gateway to align with the overall style of the blocks. So using more generic naming helps signal that.
- If a gateway/payment method needs to add it's own styling then that css should be built with the payment method file (but so far I haven't seen anything that suggests that needs to happen).

senadir: This feels logical to me.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
```diff
@@ -76,14 +76,16 @@ const CreditCardComponent = ( { billing, eventRegistration, components } ) => {
...
 			<div className="wc-blocks-credit-card-images">
-				{ getStripeServerData().icons.map( ( { id, url, alt } ) => (
-					<img
-						src={ url }
-						alt={ alt }
-						key={ id }
-						className={ `.wc-blocks-credit-${ id }-icon wc-blocks-credit-cart-icon` }
-					/>
-				) ) }
+				{ Object.entries( getStripeServerData().icons ).map(
```
nerrad, line 79: wasn't there a problem with using `Object.entries` with IE11?  If so you can probably just map through the keys and then reference each value using the provided keys:

```js
const icons = getStripeServerData().icons;
Object.keys( icons ).map( ( id ) => {
    const { url, alt } = icons[ id ];
    // ...
} );
```

assets/js/payment-method-extensions/payment-methods/stripe/stripe-utils/type-defs.js
```diff
@@ -267,24 +267,28 @@
  * @typedef {Object} Stripe Stripe api object.
  */
 
+/**
+ * @typedef {Object} CCIcon
```
nerrad, line 271: I'd prefer a full name `CreditCardIcon` here instead of `CCIcon`. `CCI` can be ambiguous as to it's meaning.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
senadir: It was [`Object.fromEntries`](https://caniuse.com/#feat=mdn-javascript_builtins_object_fromentries) that caused issues (not being supported in edge as well), granted after a google search it seems IE11 doesn’t support [`Object.entries`](https://caniuse.com/#feat=object-entries) nor does it support [`Object.values`](https://caniuse.com/#feat=object-values) as well, to my full surprise, good catch, will change it.

nerrad: Ahh, well I _know_ we have polyfills for Object.values (and we likely do for `Object.entries`) too in which case this will be okay.

assets/js/payment-method-extensions/payment-methods/stripe/stripe-utils/type-defs.js
```diff
@@ -277,18 +277,23 @@
...
+ * @property {string}                      button.type      The type of button.
+ * @property {string}                      button.theme     The theme for the button.
+ * @property {string}                      button.height    The height (in pixels) for
+ *                                                          the button.
+ * @property {string}                      button.locale    The locale to use for stripe
+ *                                                          elements.
+ * @property {boolean}                     inline_cc_form   Whether stripe cc should use
+ *                                                          inline cc
+ *                                                          form or separate inputs.
+ * @property {{[k:string]:CreditCardIcon}} icons            Contains supported cc icons.
```
nerrad, line 296: Thanks for doing this, and taking care of the spacing too. I appreciate the effort 👏

senadir: I know we usually wrap at 80, but I remember you increased the limit to 100?

nerrad: > but I remember you increased the limit to 100?

Well it's not a hard increase (comment blocks aren't currently linted), I just added another column guide in my IDE for 100 so for comment blocks I could use that as a guide if 80 was too short.  Doing it on 90 here was fine :)

Issue #2066, opened 2020-03-30
Wire up Stripe intents.


nerrad:
According to p1585580390054500-slack-CNXMMQBDW (slack conversation), Stripe CC requires intents to be added because:

> It is required to use payment intents, especially for European Economic Area, as they might need 3DS v2 verification to cover some regulations.

So, this issue is for ensuring that gets done.

nerrad:
An update, I started work on this, but as I did, I discovered that intents isn't even implemented in the stripe extension on the checkout page yet.  It's only implemented for the "Pay for Order" page (which our block doesn't handle currently):

https://github.com/woocommerce/woocommerce-gateway-stripe/blob/master/includes/class-wc-gateway-stripe.php#L878

So pending the result of the convo I started in slack (p1586461249145000-slack-CNXMMQBDW), I'm putting this on hold.

nerrad:
Moved back, from the convo I _do_ need to implement. I missed a hook where payment intent info is returned on the payment success redirect url from checkout.  This might need some creative work on how we deal with payment methods that add this kind of info in the response.  Likely what we need to do is have a `checkoutBeforeRedirect` (to accompany a `CHECKOUT_BEFORE_REDIRECT` status)  event emitter that payment methods can hook into to capture the response of the checkout and act on it before the checkout actually redirects. This will provide an opportunity for the redirect url to be modified or changed if necessary by the payment method (or even do a redirect server side - like stripe does for payment intents).

PR #2189, opened 2020-04-09
Refactor checkout status and event emitters to support stripe intents and more complex payment methods.

nerrad:
Fixes: #2066

This pull contains numerous fixes and improvements to payment method processing for stripe and for implementing stripe intents. It covers the following areas:

## Changes to Checkout Status and Event Emitters

A short time ago I introduced a `PROCESSING_COMPLETE` status to checkout to help with express payment functionality and having a clear way to indicate when payment processing is done. However the status naming in checkout has been a bit confusing and this was compounded when I realized some payment methods will need a way to interact with the _response from the server processing_. So as a part of this pull I implemented the following changes to Checkout status.

### Statuses

| Status | Event Emitters | Description |
|------- | ------------ | ---------- |
| PRISTINE | None | On initial load this is the status. |
| IDLE | None | On any _initial_ change in the checkout form, the checkout status is changed to IDLE. When a retry is possible after processing with an error, checkout is set to this status.
| BEFORE_PROCESSING | `onCheckoutBeforeProcessing` | This is triggered when the processor order is clicked (or checkout is submitted |
| PROCESSING | None | This is triggered after all observers on the `BEFORE_PROCESSING` have completed without  error.  Payment status and emitters flow start when this status is set. |
| AFTER_PROCESSING | `onCheckoutAfterProcessingWithSuccess`<br>`onCheckoutAfterProcessingWithError` | This is triggered after the checkout has been processed server side.
| COMPLETE | none | When processing is complete this is set. The checkout processor will redirect on change to this status. |

### Event Emitters

The following emitters and their behaviour were adjusted/added in this pull:

| Emitter | Triggered on status | Observers Abort? | Provided Data to Observer | Description |
|--------|---------------------|-------------------|----------------------------|-------------|
| `onCheckoutBeforeProcessing` | checkout `BEFORE_PROCESSING` | No | `{}` | Currently validation checks are observing this emitted event. Responses are merged. |
| `onCheckoutAfterProcessingWithSuccess` | checkout `AFTER_PROCESSING` and checkout does not `hasError` | Yes | `{ redirectUrl, orderId, customerId, customerNote, processingResponse }` | Typically this is what payment methods register an observer for to handle the response from the checkout server. Any response that `!== true` will abort subsequent observers from executing. |
| `onCheckoutAfterProcessingWithError` | checkout `AFTER_PROCESSING` and checkout `hasError` | Yes | `{ redirectUrl, orderId, customerId, customerNote, processingResponse }` | Similar to the `onCheckoutAfterProcessingWithSuccess` except this is triggered when there is a checkout error. |
| `onPaymentProcessing` | payment `PROCESSING` (which in turn is triggered by checkout `PROCESSING` | Yes | `{}` | Payment methods do their processing (prior to checkout being submitted to the server) during this event.

Note: `onPaymentSuccess`, `onPaymentFail`, and `onPaymentError` were all removed because I made the decision that payment status in checkout is only necessary for _before_ sending things to the server (i.e. client side payment processing errors). Payment methods should hook into `onCheckoutAfterProcessingWithSuccess` and `onCheckoutAfterProcessingWithError` to determine what to do with the server response and essentially this affects what state the checkout is in after that (which will reset payment status if needed).  I think this simplifies the logic. If we determine there's need for the payment status emitters in the future they can be added back in, but currently payment methods don't need them.

## Checkout flow changes

With the status changes the flow becomes a bit more simplified. Checkout status will be more reliable and _only_ change at specific points in the flow. The `COMPLETE` status is more stable and has one meaning: "Everything has processed and is done so redirect to whatever is set as the redirect url."

Here's an updated flow diagram:

Image 2020-04-12 at 4 55 40 PM

## Implementing support for Stripe Payment Intents

With the improvements in the flow, I was able to more easily implement Stripe Payment Intent support. For Stripe payment intents, the payment method needs to react to the checkout server response which _may_ have intent information provided (signalling the need for further authorization from the user).  Thus, the Stripe CC payment method sets an observer on the `onCheckoutAfterProcessingWithSuccess` to check for intents and react accordingly.

Image 2020-04-11 at 5 03 33 PM

## Handling for payment method errors.

With the improvements in flow, I also incorporated handling for payment method errors (and general checkout errors from the server). This also included adding a notice provider to the payment method step (the context is `wc/payment-area`.  I also added a new hook `useEmitResponse` that normalizes expectations around what can be in observer responses for various events along with exposing a set of constants to payment methods for payment status types (aligned with what is set in `PaymentResult` server side for payment statuses) and for notice contexts (so payment methods can indicate where their error message should be displayed).

Image 2020-04-12 at 7 46 04 AM
Image 2020-04-12 at 9 50 28 AM

## To test

This pull impacts all payment processing. So testing will need to cover all current payment methods. For testing purposes, you can use the various card numbers [listed here](https://stripe.com/docs/testing#regulatory-cards) to trigger various payment types (for example trigger stripe intents using `4000000000003220`). 

Generally speaking, the checkout block _should_ handle nearly all error conditions just fine right now. However, there _may_ still be some remaining glitches.  Because this pull has already got fairly large, as long as it doesn't block _successful_ payment processing, we should just make note of the remaining bugs uncovered by testing this and follow-up after the pull is merged to master.


nerrad:
As a part of this work the following are things I'm initially looking at implementing:

### Introduce new checkout statuses and rework existing.

#### `BEFORE_PROCESSING`

This is the current `PROCESSING` status. It is anything that runs before kicking off the processing status. 

This has an event emitter.

#### `PROCESSING` 

This is the current `PROCESSING_COMPLETE` status.  It is where payment processing is kicked off by the payment data context.

This does _not_ have an event emitter.

#### `AFTER_PROCESSING`

This is a new status that is set by the checkout processor after it receives the response.

It will emit two events:

`onAfterProcessingWithSuccess`

This fires if the status is `AFTER_PROCESSING` and there is no checkout error. This is an emitWithAbort emitter so any observer can abort by returning an error package.

Observers for this event will receive the redirectUrl, and checkout data as arguments (including `orderId`, `orderStatus`, `customerNote`, `customerId`, and `paymentResult` (which are the details they payment method provided in the response from checkout processing).

Observers have three response options:

**Return boolean true**:

This means that the observer did everything it needed to do successfully and checkout can continue.

**Return a success package**:

This is an object including:

```js
const successExample = {
    type: 'success',
    // checkout will redirect to this url.
    redirectUrl: 'https://someurl.com',
};
```

**Return a fail package**:

This is an object including:

```js
const failExample = {
    type: 'fail',
    // An error message that will be displayed to the user.
    message: 'An error message to be displayed',
    // if true, then checkout will display the message but reset checkout status to IDLE, which allows the 
    // user to reattempt (possibly making changes mentioned in the error message).
    // if false, then it means that the failure is irrecoverable and checkout cannot be retried.
    retry: true,
    // This is a url to redirect to.
    redirectUrl: 'https://someurl.com'
};
```

mikejolley:
@nerrad I still need to code review and test, but just giving early feedback that the statuses look good, and I prefer this other the original. I am going to leave https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2186 unmerged until this gets merged since it should simplify my PR also.

assets/js/base/context/cart-checkout/checkout-state/actions.js
```diff
@@ -5,8 +5,11 @@ import { TYPES } from './constants';
 
 const {
```
mikejolley, line 6: Can we get inline docs here matching your PR description?

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -47,12 +47,12 @@ const preparePaymentData = ( paymentData ) => {
...
 	const {
 		hasError: checkoutHasError,
-		onCheckoutProcessing,
-		onCheckoutCompleteSuccess,
+		onCheckoutBeforeProcessing,
 		dispatchActions,
 		redirectUrl,
 		isProcessing: checkoutIsProcessing,
-		isProcessingComplete: checkoutIsProcessingComplete,
+		isBeforeProcessing: checkoutIsBeforeProcessing,
```
mikejolley, line 54: I don't think we need to rename these anymore? I don't see any conflicts.

src/Library.php
```diff
@@ -250,9 +249,14 @@ public static function process_legacy_payment( PaymentContext $context, PaymentR
 		// Restore $_POST data.
 		$_POST = $post_data;
 
+		// Clear notices so they don't show up in the block.
+		wc_clear_notices();
```
mikejolley, line 253: Note. I have a PR somewhere I need to raise to capture and surface these notices instead of discarding them.

src/RestApi/StoreApi/Schemas/CheckoutSchema.php
```diff
@@ -166,9 +166,31 @@ protected function get_checkout_response( \WC_Order $order, PaymentResult $payme
...
+	/**
+	 * This prepares the payment details for the response so it's following the
+	 * schema where it's an array of objects.
+	 *
+	 * @param array $payment_details An array of payment details from the processed payment.
+	 *
+	 * @return array An array of objects where each object has the key and value
+	 *               as distinct properties.
+	 */
+	protected function prepare_payment_details_for_response( array $payment_details ) {
```
mikejolley, line 184: 👍

assets/js/base/context/cart-checkout/checkout-state/actions.js
nerrad: This set of things is all the dispatch actions constants. If you're referring to the status descriptions from the pull description, this is documented with the `CheckoutStatusConstants` type-def in `@woocommerce/type-defs/checkout.js`.  Also in the same type-def file is the `CheckoutDispatchActions` type-def which documents the _publicly exposed_ dispatch actions (publicly disposed actions on the context do not include `setPristine` so I can't document `@type {CheckoutDispatchActions}` as an inline doc here.  All the documentation exists for the publicly exposed api via the context itself (this file is internal to the context only).

assets/js/base/context/cart-checkout/checkout/processor/index.js
nerrad: True. I think the aliases make it a bit clearer in the implemented code though (so less to avoid conflicts, more to make things readable where implementation occurs - i.e. _what_ is processing?).  Eventually, this could be refactored so logic is extracted into smaller testable bits outside of this component and at that point aliasing might have less value.

src/Library.php
nerrad: So does this need an issue created so it's not missed or is that something I can just trust you'll follow-up on?

I had considered just capturing the notice instead of hooking in to grab _specific_ payment method errors (as done in this pull) but then I realized that the Woo notice api has no way to differentiate between payment related notices and other type of notices (only whether it's error, success etc). So even _if_ we capture notices, I'm concerned the utility for our use-case might be limited because we can't pipe notices to the proper contexts (payment, shipping, coupons, general checkout etc).

nerrad:
> When you leave card field blank, you get both a notice and inline errors on the form. Is this desirable?

The inline notice should appear before submitting the form, if it doesn't we should probably try to hook things up so that it's not submitted if left blank?

> Notices don't clear if more notices are added. Leave card blank and submit. Then enter card number and set expiry to date in past and submit again.

Ya, submitting should automatically clear all notices in the checkout. It's likely it's only clearing notices within the checkout context (and not the new `wc/payment-area` context).

src/Library.php
mikejolley: I'll create a new PR - I've lost the code 🤦

PR #2203, opened 2020-04-13
create a notice instead of throwing an errors if there is an error for shipping update

### LINKED ISSUE #2199: Update shipping errors should be recoverable.
**Describe the bug**
Currently, we have two errors types `woocommerce_rest_cart_shipping_rates_missing_country` 
`woocommerce_rest_cart_shipping_rates_invalid_country` that are not recoverable, so the Cart and Checkout blocks break when one of them is encountered, previously the UI didn't have a direct one to get to them (they're triggered if you don't pass a country or pass a country that is not within the "ship to" field), however the express payment has no such restrictions, it has all available countries and I'm not sure if it forces you to select a country by default or not (if it does, the first error can be skipped, but I still think it should be recoverable). 

**To Reproduce**
I don't have any express payment working for me, so this issue and reproduce steps are based on assumptions

1. in WooCommerce -> Settings -> General -> Shipping location(s) -> Ship to specific countries, define some countries.
2. In Express payment shipping section, select a country that is not within those countries.
3. the whole block will break.

How did I test this?
I hardcoded a country that is not available in that list, the block breaks when I try to update shipping.

**Expected behavior**
An error or notice should be available, depending on how we handle Express payment validation (a top-level notice maybe?)

senadir:
Fixes #2199

I'm looking for feedback into this, I'm not sure how should we handle this, would the current implementation be enough, I think we should signal to the express payment method that this failed, I believe this is already done (we just make it recoverable now).

This also needs testing, I tested by hardcoding a country value in the action dispatcher, triggering this from normal UI can't be done, it can only be done in express checkout, so someone should test there:

1. in WooCommerce -> Settings -> General -> Shipping location(s) -> Ship to specific countries, define some countries.
2. In Express payment shipping section, select a country that is not within those countries.
3. The block should only return a notice.

### DIFF
--- a/assets/js/base/hooks/shipping/use-shipping-address.js
+++ b/assets/js/base/hooks/shipping/use-shipping-address.js
@@ -12,7 +12,7 @@ import { decodeEntities } from '@wordpress/html-entities';
  * Internal dependencies
  */
 import { useStoreCart } from '../cart/use-store-cart';
-import { useThrowError } from '../use-throw-error';
+import { useStoreNotices } from '../use-store-notices';
 import { pluckAddress } from '../../utils';
 
 const shouldUpdateStore = ( oldAddress, newAddress ) =>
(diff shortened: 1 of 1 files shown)


senadir:
@nerrad I created a follow up for the second point
https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2240

I didn't create a follow up for the first one since we don't need it 400&cid=C8X6Q7XQU-slack-C8X6Q7XQU

However, handling the error one is still confusing for me and I couldn't find examples of catching errors from the hook and dispatching it.

nerrad:
> However, handling the error one is still confusing for me and I couldn't find examples of catching errors from the hook and dispatching it.

No worries, thanks for creating the followup issue. You can merge this pull because it's needed anyways.

PR #2208, opened 2020-04-14
Fix build styles incosistencies

### LINKED ISSUE #2191: Shipping Radio buttons are overlapping text in Cart block
**Describe the bug**
In Cart block, it seems radio buttons are overlapping the text
image

This only happens in Cart, checkout seems to be fine.
image

### How to reproduce:

1. Have at least two shipping options on the same zone to trigger a select.
2. do a production build `npm run build`
3. on Cart, notice the overlap.

Aljullu:
Fixes #2191 and some other minor CSS issues here and there (see test instructions).

To review this PR, it might be easier to go commit-by-commit because they are quite small and self-contained.

1. Run `npm run build`.
2. Verify _Cart_ radio buttons are correctly aligned in the shipping options.
imatge
3. Verify label and value don't break in two lines in the `<Select>` component:
imatge
4. Verify there are no regressions in the sidebar panels:
imatge
5. Verify there are no regressions in the `change address` button:
imatge

### DIFF
--- a/assets/js/base/components/sidebar-layout/style.scss
+++ b/assets/js/base/components/sidebar-layout/style.scss
@@ -29,28 +29,28 @@
 				margin-right: 0;
 			}
 		}
+	}
 
-		> .components-panel__body-title,
-		.components-panel__body-toggle {
-			&,
-			&:hover,
-			&:focus,
-			&:active {
-				background-color: transparent;
-				color: inherit;
-			}
+	.components-panel__body-title,
+	.components-panel__body-toggle {
+		&,
+		&:hover,
+		&:focus,
+		&:active {
+			background-color: transparent;
+			color: inherit;
 		}
+	}
 
-		.components-panel__body-toggle {
-			font-weight: normal;
-			font-size: inherit;
-			padding-left: 0;
-			padding-right: 36px;
+	.components-panel__body-toggle {
+		font-weight: normal;
+		font-size: inherit;
+		padding-left: 0;
+		padding-right: 36px;
 
-			&.components-button,
-			&.components-button:focus:not(:disabled):not([aria-disabled="true"]) {
-				color: inherit;
-			}
+		&.components-button,
+		&.components-button:focus:not(:disabled):not([aria-disabled="true"]) {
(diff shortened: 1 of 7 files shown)


assets/js/base/components/sidebar-layout/style.scss
```diff
@@ -29,28 +29,28 @@
...
+		font-size: inherit;
+		padding-left: 0;
+		padding-right: 36px;
 
-			&.components-button,
-			&.components-button:focus:not(:disabled):not([aria-disabled="true"]) {
-				color: inherit;
-			}
+		&.components-button,
+		&.components-button:focus:not(:disabled):not([aria-disabled="true"]) {
```
senadir, line 52: not related, but this is a wild selector, good luck overriding it 😆

bin/webpack-helpers.js
```diff
@@ -165,8 +165,8 @@ const stableFrontEndEntry = {
 };
 
 const experimentalFrontEndEntry = {
-	checkout: './assets/js/blocks/cart-checkout/checkout/frontend.js',
 	cart: './assets/js/blocks/cart-checkout/cart/frontend.js',
+	checkout: './assets/js/blocks/cart-checkout/checkout/frontend.js',
```
senadir, line 169: seems weird why would this happen in production and not dev, but good catch

Aljullu:
Thanks for the review @senadir!

> Change Address seems to be lower than the text preceding it.

Good catch. Should be fixed in f82dc1e65f9f50174b1053c9eb6324d1f7bbf530.

> Radio buttons are indeed fixed, however, maybe we can have better alignment there or better parity with checkout in terms of spacing, or is the design defining them like this?

Yes, the extra spacing in _Checkout_'s shipping options is defined in the designs, so I think we should keep it.

assets/js/base/components/sidebar-layout/style.scss
Aljullu: Agree, but this selector is indeed to override a selector from Gutenberg. :sweat:

senadir:
:shipit:

PR #2214, opened 2020-04-14
Lazy load stripe.js and preserve payment method registration order.

nerrad:
This pull implements the following:

- `stripe.js` can be lazy loaded by the `@stripe/stripe-js` library we are using which means we don't have to enqueue and register `stripe.js` server side.  Advantages with lazy loading is `stripe.js` will be loaded on demand and thus editor instances will not load it if the checkout block is not in the post content.
- Switching the above caused Stripe cc to always appear as the last tab in the payment methods (and same for any express payment buttons). To preserve order, I refactored payment method initialization so the order of payment method registration is preserved. The logic added will eventually make it simpler to support following the same payment method order that is currently set by merchants on the payment settings page (which can be done in a follow-up eventually...I think there's already an issue for it).
- I fixed an issue where adding the checkout block initially to a page will not show registered payment methods because they are only added as a dependency on frontend script. Now they are added as a dependency on the editor script too so that when you add the checkout block to content the first time, registered payment methods are available.

## To test

This affects stripe payment method registration. So verify that:

- stripe payment methods show up on the frontend (and no errors in the console).
- The checkout block already in post content loads in the editor with no errors and with stripe payment method elements loading.
- When inserting the checkout block added to a brand new page, payment methods show up in the block without any errors in the console.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/actions.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/actions.js
@@ -11,8 +11,8 @@ const {
 	ERROR,
 	FAILED,
 	SUCCESS,
-	SET_REGISTERED_PAYMENT_METHOD,
-	SET_REGISTERED_EXPRESS_PAYMENT_METHOD,
+	SET_REGISTERED_PAYMENT_METHODS,
+	SET_REGISTERED_EXPRESS_PAYMENT_METHODS,
 } = ACTION_TYPES;
 
 /**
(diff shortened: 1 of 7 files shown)


nerrad:
> I noticed that this includes changes pluralising the payment methods APIs/actions etc – I figure this is related to the refactoring to support using the configured order?

Yes, instead of dispatching the initialized payment method configuration one at a time, this dispatches the sorted initialized payment methods once (which also will result in less render calcs).

> Regarding the lazy load of stripe.min.js I didn't see any change in behaviour from master:

`stripe.min.js` isn't what gets loaded. It's `stripe.js` from `https://js.stripe.com/v3/` that gets loaded. I just noticed that while it _is_ lazy loaded, it still is loaded in the editor without the checkout block in the post content because of how the component is initialized as a part of the registration process which happens regardless of whether the block is inserted or not. 

However, these changes are still an improvement over what we had before so I still think it's worthwhile to merge.

PR #2228, opened 2020-04-15
Handle errors when loading payment methods

### LINKED ISSUE #2065: Handle errors on `StripePromise` for the stripe payment method `canMakePayment` logic.
Currently `StripePromise` silently discards any errors caught by it's exception handling. The primary reason for doing this is because the only time it was throwing errors for me in my testing is when things were loaded in the editor and the stripe.js is currently _not_ loaded in the editor context.

However, we should still have error handling and maybe do something different in the editor context. Some ideas:

- if in editor context don't throw error (except if #2056 implements loading stripe.js we can't do this).
- also need to detect what type of error (i.e. if because invalid api key or api key isn't available then need to show an appropriate error on the frontend?).
- context aware error? If in frontend and not logged in as admin, show shopper friendly error. Otherwise show merchant friendly error to help with debugging?

### LINKED ISSUE #2220: Stripe external files failing to load will disable other payment methods
**Describe the bug**
So I first saw this when I had a bit slow internet, I confirmed by blocking stripe.com domain in the devtools network panel, other payments methods won't load.
image

**To Reproduce**

1. Have two payment methods (checks and Stripe)
2. in the devtools, block js.stripe.com 
image

3. Relaod the page with devtools open
4. Watch that both payment options are not working
image

**Expected behavior**
Only stripe should fail.

**Screenshots**

* Core version: [e.g. 5.0.3]
* Gutenberg version (if installed): [e.g. 4.8.0]
* WooCommerce version: [e.g. 3.5.3]
* WooCommerce Blocks version: [e.g. 1.3.1]
* Site language:

* OS: [e.g. iOS]
* Version [e.g. 22]

**Smartphone (please complete the following information):**

* Device: [e.g. iPhone6]
* OS: [e.g. iOS8.1]
* Version [e.g. 22]

**Additional context**

Aljullu:
Fixes #2065.
Fixes #2220.

### Screenshots
_Frontend:_
imatge

_Editor:_
imatge

1. To force Stripe to fail you can:
	* Go into the Stripe plugin settings and uncheck `Enable Test Mode` without adding any live key.
	* Follow the steps in #2220.
	* Or replace [L15](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/payment-method-extensions/payment-methods/stripe/stripe-utils/load-stripe.js#L15) of `load-stripe.js` to force an error.
```diff
-stripe = loadStripe( getApiKey() );
+stripe = undefinedFunctionThatFails( getApiKey() );
```
2.  Verify this is true:
	* Stripe silently fails in the frontend for normal users. Making other payment methods available.
	* For admins, it does the same but it also shows an error in the console.
	* In the editor, it shows a notice.

### DIFF
--- a/assets/js/base/components/payment-methods/payment-method-error-boundary.js
+++ b/assets/js/base/components/payment-methods/payment-method-error-boundary.js
@@ -4,18 +4,24 @@
 import { __ } from '@wordpress/i18n';
 import { Component } from 'react';
 import { Notice } from 'wordpress-components';
+import PropTypes from 'prop-types';
+import { CURRENT_USER_IS_ADMIN } from '@woocommerce/block-settings';
 
 class PaymentMethodErrorBoundary extends Component {
 	state = { errorMessage: '' };
 
 	static getDerivedStateFromError( error ) {
+		const { isEditor } = this.props;
+
+		const errorMessage =
+			error.message && ( isEditor || CURRENT_USER_IS_ADMIN )
+				? error.message
+				: __(
+						'There was an error loading the payment method.',
+						'woo-gutenberg-products-block'
+				  );
 		return {
-			errorMessage:
-				error.message ||
-				__(
-					'There was an error loading the payment method.',
-					'woo-gutenberg-products-block'
-				),
+			errorMessage,
 		};
 	}
 
@@ -33,5 +39,13 @@ class PaymentMethodErrorBoundary extends Component {
 		return this.props.children;
 	}
 }
+
+PaymentMethodErrorBoundary.propTypes = {
+	isEditor: PropTypes.bool,
+};
+
+PaymentMethodErrorBoundary.defaultProps = {
+	isEditor: false,
+};
 
 export default PaymentMethodErrorBoundary;
--- a/assets/js/base/components/payment-methods/payment-method-error-boundary.js
+++ b/assets/js/base/components/payment-methods/payment-method-error-boundary.js
@@ -5,11 +5,6 @@ import { __ } from '@wordpress/i18n';
 import { Component } from 'react';
 import { Notice } from 'wordpress-components';
 
-/**
- * Internal dependencies
- */
-//import './editor.scss';
-
 class PaymentMethodErrorBoundary extends Component {
 	state = { errorMessage: '' };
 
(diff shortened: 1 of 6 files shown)


senadir:
I know that errors are provided by Stripe.js, but they feel more like a developer error, could we alias them to a custom error text like "go to stripe settings and add your API key" ?

Aljullu:
> I know that errors are provided by Stripe.js, but they feel more like a developer error, could we alias them to a custom error text like "go to stripe settings and add your API key" ?

That would be nice but I wonder if it would increase the scope of these issues too much and if it's really required for the MVP. In the screenshot above I'm showing the API error, but the error returned by Stripe could be anything.

Currently, this PR is just displaying the error provided by Stripe. If we wanted to add custom error messages we would need to parse either the Stripe plugin configuration or the errors returned by it and show our custom error texts, which doesn't seem to be straight-forward to me. Or do you had something else in mind?

assets/js/base/components/payment-methods/payment-method-error-boundary.js
```diff
@@ -0,0 +1,49 @@
...
+		const { hasError, errorMessage } = this.state;
+		const { isEditor } = this.props;
+
+		if ( hasError ) {
+			return (
+				<Notice isDismissible={ false } status="error">
+					{ errorMessage && ( isEditor || CURRENT_USER_IS_ADMIN )
+						? errorMessage
+						: __(
+								'There was an error loading the payment method.',
```
nerrad, line 30: I think this notice appearing to shoppers could be improved a bit, maybe something like this?:

> This site is experiencing difficulties with this payment method. Please contact the owner of the site for assistance.

At some point this may be something we'd want to introduce as a configurable text string.

Aljullu:
So yeah, my idea was that:
* Shoppers would not see any error in the frontend if the payment method fails to load. The payment method would silently not appear but no errors will be shown. (There is still the chance that the error boundary might catch an unhandled exception, but that would be rare because payment methods are expected to take care of handling errors internally and display them with validation errors or error notices on top).
* Admins in the frontend will not directly see any error either. But errors will be shown in the console.
* In the editor, errors will be shown as a notice.

Does that make sense? I don't mind changing it, but I think it's desirable that a bad setup or broken plugin doesn't show up in the frontend for normal shoppers.

assets/js/base/components/payment-methods/payment-method-error-boundary.js
Aljullu: Added in 0814de191360992cf2af7dbc485d097c8392a8d4.

nerrad:
> Does that make sense? 

Works for me, it's something we can iterate on if we find out otherwise once the blocks get wider usage.

PR #2234, opened 2020-04-15
Improve  "sold individually" validation

### LINKED ISSUE #2215: Cart is not respecting Sold individually option
**Describe the bug**
When you have a product that you only sell one in an order via the `Sold individually: Enable this to only allow one of this item to be bought in a single order` inventory option, the normal cart will not allow you add more than one, or change quantity, Cart Block will allow you to change quantity and checkout with more that one.

image
image
image

it also breaks old cart total, even if that cart won't allow you to change quantity (and will say a hardcoded 1, even if subtotal is not)
image

**To Reproduce**

1. make a product be sold individually
2. add it to cart
3. change quantity
4. checkout like nothing happened.

**Expected behavior**
You shouldn't be able to change the quantity above 1.

mikejolley:
- Adds checks for "sold individually" to the cart update endpoint.
- Adds checks for "sold individually" to the cart validation function (for checkout).
- Enforces a maximum of 1 if sold individually on the client side.
- Ensures add to cart events, if fired, are not never-ending by using the promise.

Fixes #2215

1. Edit an item and make it "sold individually"
2. Ensure only 1 can be added from store product pages.
3. Ensure maximum of 1 is allowed on cart page.
4. Add an item to cart which is not "sold individually". Set qty to 10. Edit and make "sold individually" true. Now view cart page and see errors.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
@@ -28,11 +28,21 @@ import Dinero from 'dinero.js';
  * @param {boolean}     backOrdersAllowed Whether to allow backorders or not.
  * @param {number|null} lowStockAmount    If present the number of stock
  *                                        remaining.
+ * @param {boolean}     soldIndividually  Whether an item is sold individually or not.
  *
  * @return {number} The maximum number value for the quantity input.
  */
-const getMaximumQuantity = ( backOrdersAllowed, lowStockAmount ) => {
+const getMaximumQuantity = (
+	backOrdersAllowed,
+	lowStockAmount,
+	soldIndividually
+) => {
+	if ( soldIndividually ) {
+		return 1;
+	}
+
 	const maxQuantityLimit = getSetting( 'quantitySelectLimit', 99 );
+
 	if ( backOrdersAllowed || ! lowStockAmount ) {
 		return maxQuantityLimit;
 	}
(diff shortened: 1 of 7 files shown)


assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -133,7 +144,8 @@ const CartLineItemRow = ( { lineItem } ) => {
 					quantity={ quantity }
 					maximum={ getMaximumQuantity(
 						backOrdersAllowed,
-						lowStockRemaining
+						lowStockRemaining,
+						soldIndividually
```
senadir, line 148: should we possibly emit out the whole quantity input if it’s sold individually?

mikejolley: Core does but I prefer consistentcy personaly.

nerrad: I agree with MIke here. It keeps the ui consistent.

mikejolley:
@nerrad I'm working on coupons where it removes the coupon and shows the error, so retry would function. Strange though, I thought we had a back to cart link.

> All Products block will still allow the user to click the "Add to Cart" button for the limited product with no feedback as to why the quantity doesn't increment (can be done in followup).

This isn't new - before this PR in master and on 2.5 it errors instead (console error). I don't think we have an error notice system implemented in all products so it's not an easy fix and should be a follow up.

PR #2237, opened 2020-04-16
rename docker-related npm scripts so it's clear they are e2e infrastructure

haszari:
I _think_ (correct me if I'm wrong) our docker-compose setup is purely for e2e testing, and not intended for general development (or other purposes). If anyone is using this docker compose setup for non-e2e purposes let me know!

With that in mind, I propose we rename the relevant npm scripts - use `test:e2e` prefix instead of docker. The idea is to name the command on what it should achieve (rather than how it is implemented) - e.g. `npm run test:e2e:up` will bring up the e2e test environment :)

It's becoming more common for repositories to have a docker-compose developer setup built in, so I think using `e2e` in the name is clearer and reduces the need to document things.

Note also I have removed `npm run docker:clear-all` as this force deletes all images (not just ones related to blocks e2e), [even from running containers](https://docs.docker.com/engine/reference/commandline/rmi/), so is a dangerous convenience. If this is still important to workflows perhaps we can rename it so it's clear it's a force delete 💣 

1. On `master`, if you are running existing e2e setup, take it down before switching branch e.g. `npm run docker:down`.
2. Switch to this branch.
3. Launch e2e environment, e.g. `npm run test:e2e:up`.
4. Wait a moment while e2e gets set up - ☕ 
5. Run e2e tests and confirm everything's working the way it should `npm run test:e2e`.

Please also review / test other e2e workflows that you know about and use.


package.json
```diff
@@ -54,6 +48,11 @@
 		"storybook": "start-storybook  -c ./storybook -p 6006 --ci",
 		"storybook:build": "build-storybook  -c ./storybook -o ./storybook/dist",
 		"test": "wp-scripts test-unit-js --config tests/js/jest.config.json",
+		"test:e2e:down": "docker-compose down",
+		"test:e2e:down:reset": "docker-compose down -v",
+		"test:e2e:ssh": "docker exec -it woo-blocks_wordpress-www_1 /bin/bash",
+		"test:e2e:up": "docker-compose up -d",
+		"test:e2e:up:build": "docker-compose up --build -d",
 		"test:e2e": "./tests/bin/e2e-test-integration.js",
```
mikejolley, line 56: When I mentioned chaining I meant here; we can run the docker up commands too so you don't need to manually use the commands above.

senadir: depending on if you're testing things, you would run `test:e2e:up` once but would run  `test:e2e` many times.

haszari: Yep exactly @senadir. Also there may need to be a long wait between `up` and running the first test; and the user needs to know to `down` when they are finished as well. 

Having said that I agree with @mikejolley we should make this as simple as possible to run – let's keep iterating on it and polishing the developer experience :)

PR #2241, opened 2020-04-16
Update payment method data context to use noticeContexts.

nerrad:
While [working on documentation](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1476) I realized that I forgot to update the payment method data context to use the `noticeContexts` "constants" from the new `useEmitResponse` hook.  The work in this pull adds that and also allows for the possibility of `messageContext` being returned from a registered observer that targets a specific notice area.

## To Test

This impacts payment method processing and where a notice is displayed, you can verify nothing is broken by using a card number for the CC (eg `4000000000000002`) that will result in a declined error notice.

Note: while testing you may experience the following issues that are already known (and have issues):

- checkout will scroll to the top on failed checkout processing of the cc (but the declined notice should still be in the payment methods step).
- If you try immediately submitting a valid cc number, checkout will process but the `declined` notice won't disappear when clicking the checkout submit button.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -99,6 +99,7 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 		isSuccessResponse,
 		isErrorResponse,
 		isFailResponse,
+		noticeContexts,
 	} = useEmitResponse();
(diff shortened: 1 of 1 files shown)


nerrad:
> We could check if `hasValidationErrors` is true before scrolling to the top. Would that work?

I don't know for sure but I imagine it would? Is the scrollToTop functionality _only_ for validation errors or any notice?

I thought there was already an issue for this but I can't find it. Would you mind taking care of creating an issue @Aljullu ?

Issue #2213, opened 2020-04-14
Previously applied coupons that no longer meet criteria are still visible in cart (even if they do nothing)


senadir:
**Describe the bug**
If you have a coupon that requires a certain condition (min amount spent or certain products) and you meet that condition, you can apply it, however, if you remove that condition after it (reduce your total, remove certain items), the coupon will not change the total however it will stay visible and you can't delete it.

note: the invalid coupon code is the error I get when I try to delete it.
![image](https://user-images.githubusercontent.com/6165348/79249481-3abd9d80-7e75-11ea-8a14-059024bbdb85.png)

**To Reproduce**

1. setup a coupon that only works with a certain condition (min amount or certain product).
2. meet that condition. 
3. add the coupon
4. unmeet it
5. the coupon is still there, doing nothing, yet you cannot remove it.

**Expected behavior**
the coupon should be gone once you unmeet the condition, if possible, a notice should be printed.

see old cart
![image](https://user-images.githubusercontent.com/6165348/79249688-8e2feb80-7e75-11ea-9543-f1c5b5efb069.png)

PR #2245, opened 2020-04-17
Coupon and Cart Item Validation Fixes

### LINKED ISSUE #2212: Coupons with restrictions are showing raw errors (HTML) when restrictions are not met.
**Describe the bug**
1- have a coupon with a restriction (min or max spend).
2- try to apply it when you don't meet that restriction.
3- the error will have HTML in it

image

I think a good solution (instead of remapping all old errors to new ones) is to filter out HTML from them.

### LINKED ISSUE #2218: Ensure our Add to Cart has parity with Core Add to Cart in term of filters
**Describe the bug**
A recent bug (#2215) raised this point, however that bug is core check, not a filter check, so to make sure other plugins that add restrictions work fine, we need to ensure filter parity between Core "Add to Cart" and our API "Add To Cart", see @mikejolley comment p1586883607249300-slack-C8X6Q7XQU

As a part of the slack discussion, it was noted that merchants can use plugins like [WooCommerce Product Dependencies ](https://en-gb.wordpress.org/plugins/woocommerce-product-dependencies/) to restrict buying some products, however, in our case you can add products via the API directly, so we can verify filter parity by making sure no restrictions set by this plugin can be bypassed by our API.

### LINKED ISSUE #2235: No validation when you created an email restricted coupon
1. Create a new coupon 
2. In usage restriction, set allowed emails to `*@automattic.com` this should allow any a12s to place an order.
3. go to Checkout Block and place that coupon.
4. checkout with any other email (Gmail or something)
4. notice that there is no validation whatsoever and your order passes normally.
image

image

**Expected behavior**
It should not allow me to finish the order, classic checkout gives this error and removes the coupon.

```
Sorry, it seems the coupon "a12s" is not yours - it has now been removed from your order.
```
image

mikejolley:
This PR has several fixes related to validation on the cart and checkout. It resolves the following bug reports:

_No validation when you created an email restricted coupon #2235_

Fixed by running late validation before payment. Payment is prevented, and restricted coupons are removed. This updates cart totals so the payment can be attempted again.

Fixes #2235

_Ensure our Add to Cart has parity with Core Add to Cart in term of filters #2218_

We were missing 2 general purpose filters from woo core that plugins use to run custom validation:

- `woocommerce_add_to_cart_validation`
- `woocommerce_check_cart_items`

I've added them to the API and included our Notice->Exception logic to avoid notices being added in the background.

Fixes #2218

_Coupons with restrictions are showing raw errors (HTML) when restrictions are not met. #2212_

All we get from core is a message; this is resolved by stripping HTML tags before providing the API response.

Fixes #2212

_Previously applied coupons that no longer meet criteria are still visible in cart (even if they do nothing) #2213_

We have logic in place to gather errors for cart items, so I added similar logic for coupons. The coupon is removed, and a notice is surfaced. To keep notices visible when doing lots of cart updates, I made it dismissable and prevented all notice being cleared after cart updates.

#2213

**Test case 1:**
1. Create a coupon with a min spend
2. Apply min spend coupon without meeting requirements. Check error message is correctly formatted.
3. Meet requirements by adding products to cart. Add coupon again.
4. Reduce cart qty until requires are not met again. Check notice is shown and the coupon is removed.

**Test case 2:**
1. Create a coupon with a usage restriction of `*@automattic.com`
2. Apply it.
3. Go to place an order without setting the billing email to an automattic address.
4. See error notice and cart removal. Cart totals should be correct.

**Test case 3:**
1. Create a product with a dependency using https://en-gb.wordpress.org/plugins/woocommerce-product-dependencies/
2. Try to add it to cart. Add to cart will fail. There is no notice because of https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2242

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/checkout-order-error/index.js
+++ b/assets/js/blocks/cart-checkout/checkout/checkout-order-error/index.js
@@ -15,7 +15,16 @@ import {
 	PRODUCT_NOT_PURCHASABLE,
 	PRODUCT_NOT_ENOUGH_STOCK,
 	PRODUCT_SOLD_INDIVIDUALLY,
+	GENERIC_CART_ITEM_ERROR,
 } from './constants';
+
+const cartItemErrorCodes = [
+	PRODUCT_OUT_OF_STOCK,
+	PRODUCT_NOT_PURCHASABLE,
+	PRODUCT_NOT_ENOUGH_STOCK,
+	PRODUCT_SOLD_INDIVIDUALLY,
+	GENERIC_CART_ITEM_ERROR,
+];
 
 /**
  * When an order was not created for the checkout, for example, when an item
@@ -59,12 +68,7 @@ const CheckoutError = () => {
 const ErrorTitle = ( { errorData } ) => {
 	let heading = __( 'Checkout error', 'woo-gutenberg-products-block' );
 
-	if (
-		errorData.code === PRODUCT_NOT_ENOUGH_STOCK ||
-		errorData.code === PRODUCT_NOT_PURCHASABLE ||
-		errorData.code === PRODUCT_OUT_OF_STOCK ||
-		errorData.code === PRODUCT_SOLD_INDIVIDUALLY
-	) {
+	if ( cartItemErrorCodes.includes( errorData.code ) ) {
 		heading = __(
 			'There is a problem with your cart',
 			'woo-gutenberg-products-block'
@@ -84,12 +88,7 @@ const ErrorTitle = ( { errorData } ) => {
 const ErrorMessage = ( { errorData } ) => {
 	let message = errorData.message;
 
-	if (
-		errorData.code === PRODUCT_NOT_ENOUGH_STOCK ||
-		errorData.code === PRODUCT_NOT_PURCHASABLE ||
-		errorData.code === PRODUCT_OUT_OF_STOCK ||
-		errorData.code === PRODUCT_SOLD_INDIVIDUALLY
-	) {
+	if ( cartItemErrorCodes.includes( errorData.code ) ) {
 		message =
 			message +
 			' ' +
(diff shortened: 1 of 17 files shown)


senadir:
> Coupons with restrictions are showing raw errors (HTML) when restrictions are not met.

I can confirm that this is working.
image

> Previously applied coupons that no longer meet criteria are still visible in cart (even if they do nothing) #2213

I can confirm that this is working.
image

assets/js/blocks/cart-checkout/checkout/checkout-order-error/index.js
```diff
@@ -84,12 +88,7 @@ const ErrorTitle = ( { errorData } ) => {
 const ErrorMessage = ( { errorData } ) => {
 	let message = errorData.message;
 
-	if (
-		errorData.code === PRODUCT_NOT_ENOUGH_STOCK ||
-		errorData.code === PRODUCT_NOT_PURCHASABLE ||
-		errorData.code === PRODUCT_OUT_OF_STOCK ||
-		errorData.code === PRODUCT_SOLD_INDIVIDUALLY
-	) {
+	if ( cartItemErrorCodes.includes( errorData.code ) ) {
```
senadir, line 91: I like this, much cleaner.

src/RestApi/StoreApi/Utilities/OrderController.php
```diff
@@ -97,6 +97,100 @@ public function sync_customer_data_with_order( \WC_Order $order ) {
...
+				$this->validate_coupon_email_restriction( $coupon, $order );
+			} catch ( \Exception $error ) {
+				$coupon_errors[ $coupon_code ] = $error->getMessage();
+			}
+			try {
+				$this->validate_coupon_usage_limit( $coupon, $order );
+			} catch ( \Exception $error ) {
+				$coupon_errors[ $coupon_code ] = $error->getMessage();
+			}
+		}
```
senadir, line 127: I see that we're validating case by case here, why aren’t we using Core validation?

mikejolley: The core functions make use of global `POST` so it's not suitable for us.

senadir: so in a follow up next cycle, we should see how those extensions p2-pbIy4N-7d are doing it, ideally, it would be a couple of filters or actions here and there.

PR #2247, opened 2020-04-17
Do not scroll to top when payment methods have an error

### LINKED ISSUE #2243: Payment method errors shouldn't scroll the page to the top
Coming from #2241.

**To Reproduce**
1. In the _Checkout_ block with Stripe introduce a wrong card number: `4000000000000002`.
2. Click on `Place Order`.
3. Notice it scrolls to the top of the block, even though the error notice was displayed in the payment method step.

Aljullu:
Fixes #2243.

### Screenshots
Screencast

1. Submit the _Checkout_ form without filling any input field and verify the errors appear and the page scrolled to the top.
2. Fill the form but leaving the card data empty and submit again. Verify the payment method errors appear but the page didn't scroll to the top.

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -61,7 +61,6 @@ const CheckoutProcessor = () => {
 	const {
 		activePaymentMethod,
 		currentStatus: currentPaymentStatus,
-		errorMessage,
 		paymentMethodData,
 		expressPaymentMethods,
 		paymentMethods,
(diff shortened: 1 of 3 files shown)


assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -113,16 +112,6 @@ const CheckoutProcessor = () => {
...
 
-	useEffect( () => {
-		if ( errorMessage ) {
-			addErrorNotice( errorMessage, {
-				id: 'payment-method-error',
-			} );
-		} else {
-			removeNotice( 'payment-method-error' );
-		}
-	}, [ errorMessage ] );
```
Aljullu, line 124: This is no longer needed if payment methods use their own notice context.

PR #2255, opened 2020-04-20
Fix snackbar being too high and covering coupon section

### LINKED ISSUE #2204: snackbars are too high in desktop
try applying a coupon, you will get a snackbar, however, it is too high

image

tested in Chrome, storefront 2.5.5 on master.

senadir:
Fixes #2204

I reverted those values to same values Gutenberg set for notices, also, it applied 100% width, I removed that so it doesn't cover the coupon or button section.
image

1. Make sure the snackbar looks good on mobile and desktop.
2. Make sure it doesn't cover the inputs to the right.
3. Make sure no regressions are created.

### DIFF
--- a/assets/js/base/components/store-notices-container/style.scss
+++ b/assets/js/base/components/store-notices-container/style.scss
@@ -27,8 +27,9 @@
 
 .components-snackbar-list {
 	position: fixed;
-	bottom: 100px;
-	left: 100px;
+	bottom: 40px;
+	left: 16px;
+	width: auto;
 
 	@include breakpoint( "<600px" ) {
 		position: fixed;
--- a/assets/js/base/components/store-notices-container/style.scss
+++ b/assets/js/base/components/store-notices-container/style.scss
@@ -27,20 +27,19 @@
 
 .components-snackbar-list {
 	position: fixed;
-	bottom: 40px;
+	bottom: 20px;
 	left: 16px;
 	width: auto;
 
-	@include breakpoint( "<600px" ) {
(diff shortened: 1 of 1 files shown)


assets/js/base/components/store-notices-container/style.scss
```diff
@@ -27,8 +27,9 @@
 
 .components-snackbar-list {
 	position: fixed;
-	bottom: 100px;
-	left: 100px;
+	bottom: 40px;
+	left: 16px;
+	width: auto;
 
 	@include breakpoint( "<600px" ) {
```
Aljullu, line 34: I know that's not introduced in this PR so it can be fixed as a follow-up, but just leaving it here in case you want to fix it as well. I think we should use the same breakpoints we use for the _Cart_ and _Checkout_ blocks, which is `782px` instead of `600px`.

assets/js/base/components/store-notices-container/style.scss
```diff
@@ -27,8 +27,9 @@
 
 .components-snackbar-list {
 	position: fixed;
-	bottom: 100px;
-	left: 100px;
+	bottom: 40px;
+	left: 16px;
+	width: auto;
```
Aljullu, line 32: One more comment: I think after adding `width: auto` here, it should be safe to remove L39 `width: fit-content;`.

senadir:
> Which values did you use as a reference? For me the snackbars in the editor are 20px from the bottom and 16px from the left:

weird, mine is 40px (running master)
image

PR #2256, opened 2020-04-20
Add quantity limit at product level

### LINKED ISSUE #2251: Trying to increase an item quantity to above stock level should reduce it back to stock level.
**Describe the bug**
have an item with stock managed and a stock quantity to say 5 and a threshold below that.
In your cart, you can increase that value to how much you want.
you will receive an error saying that the stock quantity is limited to 5, however your cart quantity will still be the amount you set, the total will still reflect that, continuing to checkout will give an error about the stock.

**Expected behavior**
Your quantity should be reduced to the stock amount.

This is mainly to prevent people from shooting themselves in the foot.

mikejolley:
Moves the quantity selector limit from global to product level so that we can enforce a max quantity if stock level is lower than the limit. This prevents stock errors due to customers adding too much to the cart.

Fixes #2251

1. Give a product stock level 10, no backorders
2. Give a product stock level 10, with backorders
3. Make a product sold individually

Add each product to the cart and see if appropriate maximum is used.


mikejolley:
@senadir separate bug. Pushed now - with backorders on, stock checks are not needed. Confirm now?

PR #2264, opened 2020-04-20
Checkout: Rename address2 attribute

### LINKED ISSUE #2262: Checkout: 'Apartment, suite, etc.' attribute is not applied in the frontend
**To Reproduce**
1. Create a page with the _Checkout_ block and toggle `Apartment, suite, etc.` so its disabled.
2. Check the page in the frontend.
3. Notice the `Apartment, suite, etc.` is still visible.

mikejolley:
For some odd reason when `data-` attributes are converted to camelCase, numbers are handled differently. `data-show-address-2-field` becomes `showAddress-2Field` rather than the expected showAddress2Field`.

To fix this, renamed the attribute to showApartmentField`. _Note, this does invalidate blocks._

Fixes #2262

1. Create checkout block.
2. Toggle apartment field off.
3. Ensure it's not shown on the frontend.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/attributes.js
+++ b/assets/js/blocks/cart-checkout/checkout/attributes.js
@@ -12,7 +12,7 @@ const blockAttributes = {
 		type: 'boolean',
 		default: false,
 	},
-	showAddress2Field: {
+	showApartmentField: {
 		type: 'boolean',
 		default: true,
 	},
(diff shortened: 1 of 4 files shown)


senadir:
> For some odd reason when data- attributes are converted to camelCase, numbers are handled differently. data-show-address-2-field becomes showAddress-2Field rather than the expected showAddress2Field`.

I also think we should upstream this later, but I doubt they will fix it now since it might break previous blocks.

Aljullu:
> I also think we should upstream this later,

Unfortunately, I don't think we can upstream anything here, this behavior depends on browsers:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset#Name_conversion

PR #2267, opened 2020-04-21
Update payment method config to include icons

### LINKED ISSUE #1472: Cart: Create api for payment methods to expose card icons.
The cart block has the following area where payment method icons are displayed:

Image 2020-01-03 at 9 28 50 AM

* Payment methods should register there component for this area using the registration api introduced in #1349

### LINKED ISSUE #2324: Cart block: displays payment gateways logos even if I only have cheque active
**Describe the bug**
In the cart page see the payment gateways logos even if I only have cheque active

**To Reproduce**

1. Make sure that Cheque is the only payment turned on
2. Add the cart block to a page
3. Visit the cart page and see the payments' logos.

**Expected behavior**
Not seeing the payment gateways logos if the store only accepts cheque.

**Screenshots**
https://d.pr/i/VOh5Ml

mikejolley:
This might need to wait for final approval from @nerrad, but here goes.

This exposes an `icons` property in payment method configs which can either be `null` or a React element containing payment icons.

Proposed changes:

- Loads registered payment methods on the CART page, so we can pull icons
- Payment config: Added `icons` attribute
- Payment config: Updated Stripe config to return it's credit card icons
- Payment config: Updated Check config to return null icons (no icons)
- Payment config: Name should be element or string, we shouldn't force use of `strong`. Updated validation.
- Payment methods component: Split this out of the "proceed to checkout" button component, moved them up in the tree, and removed the payment icons that existed there before since Payment Methods will be returning these instead.
- Payment methods component: Added styling to keep icons small and spaced correctly.

Fixes #1472

The biggest ❓ from me is, do we want to allow payment methods to return a component with their icons, or should we instead do something more format? cc @nerrad for feedback.

### Screenshots

Screenshot 2020-04-21 at 14 38 57

With stripe enabled, confirm icons are displayed on the cart page.

fixes #2324


assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
```diff
@@ -82,16 +82,7 @@ const CreditCardComponent = ( {
 				/>
 			) }
 			<div className="wc-blocks-credit-card-images">
```
senadir, line 84: might as well moved the container div since the icons style relies on it.

mikejolley:
Hmmm duplication might be an issue unless users can control display, or there is some kind of standard. Will wait for @nerrad.

mikejolley:
@nerrad Looking at how Stripe currently gets it's icons (`getStripeServerData().icons`) - how can gateways consolidate that or know which of the icons are "common"? Can gateways replace icons, are they forced to use common icons, or can they do what they want anyway by using unique ids? Does the merchant have any control over this at all?

I'm wondering if we're going about this wrong by leaving it in the hands of gateways. I can definitely imagine a case where a gateway could support lots and lots of obscure gateways but the merchant only wants to surface the common ones (e.g. mastercard, visa). Or limit the number of icons shown at once.

senadir:
I believe the best choice here is to not let the payment gateway register anything, and leave that area as an inner block for merchants to manually insert what they like in, or could be restricted to what type of input.

nerrad:
> how can gateways consolidate that or know which of the icons are "common"? Can gateways replace icons, are they forced to use common icons, or can they do what they want anyway by using unique ids? Does the merchant have any control over this at all?

This is where I think we need to shift things. It'll be up to Woo (currently via blocks, but eventually probably in core) to determine what "common" cc icons are and the base source urls for those. Likely something that would have to be filterable. An advantage of this approach is that eventually we could expose a ui/ux for merchants to swap out what they want to use for a specific cc icon.

The idea here is that we document common cc icon "ids" that gateways can use to indicate what they support (`visa`, `mastercard`, `diner`, `discover`, `american_express` etc) and gateways use those to indicate support.

> I'm wondering if we're going about this wrong by leaving it in the hands of gateways.

If you're referencing what's currently in this pull, I agree. I think what we have currently will lead to more problems (duplication being a major one). This is why I think we need to centralize _common_ cc icons (and I know defining "common" could be tricky...) and expose those for gateways to indicate support for (various gateways will have their own criteria for support as well - like stripe does geographically). Then we leave less common or gateway specific icons (like paypal) up to the payment method to provide.

> I can definitely imagine a case where a gateway could support lots and lots of obscure cards but the merchant only wants to surface the common ones (e.g. mastercard, visa). Or limit the number of icons shown at once.

Right, by using the api approach I'm suggesting here, we can surface a ui/ux in the block controls eventually that basically shows a list of all the icons and allows for finer control by the merchant.

> If we do leave to gateways, what about making this ID based only, rendering elements using that ID, and handling icon display using CSS (if possible)?

Ya that's a good idea too (mostly just for the common cc icons?). 

> I believe the best choice here is to not let the payment gateway register anything, and leave that area as an inner block for merchants to manually insert what they like in, or could be restricted to what type of input.

That might be the way we expose this in a future iteration for merchants to adjust (and the inner blocks would be pre-populated by icons supported by registered payment methods and the geographical area the shop supports). However, I don't think we should ship i1 with this as it needs more discussion I think.

nerrad:
Also I forgot to mention, the icons in the editor context for the cart are kind of wonky:

Image 2020-04-27 at 9 33 56 AM

assets/js/base/components/cart-checkout/payment-method-icons/index.js
```diff
@@ -0,0 +1,41 @@
...
+		return null;
+	}
+
+	return (
+		<ul className="wc-block-cart__payment-method-icons">
+			{ iconConfigs.map( ( icon ) => {
+				const iconProps = {
+					...icon,
+					...getCommonIconProps( icon.id ),
+				};
```
nerrad, line 29: nice!

assets/js/base/components/cart-checkout/payment-method-icons/utils.js
```diff
@@ -0,0 +1,32 @@
...
+			icon = {
+				id: raw,
+				alt: raw,
+				src: null,
+			};
+		} else {
+			icon = {
+				id: raw.id,
+				alt: raw.alt,
+				src: raw.src,
```
nerrad, line 22: There's no validation that `id`, `alt`, and `src` exist here right? probably should only set this if the icon is a valid object.

assets/js/payment-method-extensions/payment-methods/paypal/index.js
```diff
@@ -39,6 +40,13 @@ const paypalPaymentMethod = {
 	),
 	content: <Content />,
 	edit: <Content />,
+	icons: [
+		{
+			id: 'paypal',
+			alt: __( 'PayPal', 'woo-gutenberg-products-block' ),
+			src: `${ WC_BLOCKS_ASSET_URL }img/paypal-mark.jpg`,
+		},
+	],
```
nerrad, line 49: so feasibly, this payment method could just do something like this right?

```js
icons: [ 'paypal' ]
```

or:

```js
icons: [ { id: 'paypal' } ],
```

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
```diff
@@ -12,6 +12,7 @@ import { InlineCard, CardElements } from './elements';
 import { Elements, useElements, useStripe } from '@stripe/react-stripe-js';
 import { useState } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
+import { PaymentMethodIcons } from '@woocommerce/base-components/cart-checkout';
```
nerrad, line 15: Unfortunately we can't use components directly in the payment method like this (because this will be extracted to the payment method extension). Instead, any components we want payment methods to have access to should be passed through as props via the `usePaymentMethodInterface`hook implementation in our payment method components.  So that means you'd probably add the `PaymentMethodIcons` component to [this property](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/d63689a83a28f8868c5c4bcbcbfd455f59dae8aa/assets/js/base/hooks/payment-methods/use-payment-method-interface.js#L187-L190) and then consume the prop in the `StripeCreditCard` component.

tsconfig.json
```diff
@@ -27,7 +27,7 @@
 			"@woocommerce/block-components/*": [ "assets/js/components/*" ],
 			"@woocommerce/block-data": [ "assets/js/data" ],
 			"@woocommerce/block-hocs": [ "assets/js/hocs" ],
-			"@woocommerce/blocks-registry/*": [ "assets/js/blocks-registry/*" ],
+			"@woocommerce/blocks-registry": [ "assets/js/blocks-registry" ],
```
nerrad, line 30: 😮 good catch.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
nerrad: Incidentally, this is likely also why we're seeing stripe css getting built (which is unnecessary).

assets/js/payment-method-extensions/payment-methods/paypal/index.js
```diff
@@ -4,6 +4,7 @@
 import { registerPaymentMethod } from '@woocommerce/blocks-registry';
 import { __ } from '@wordpress/i18n';
 import { getSetting, WC_ASSET_URL } from '@woocommerce/settings';
+import { WC_BLOCKS_ASSET_URL } from '@woocommerce/block-settings';
```
nerrad, line 7: If this resulted in 1.34kb being added to the bundle build, that's a bit alarming and we should avoid it.

assets/js/payment-method-extensions/payment-methods/paypal/index.js
mikejolley: Only if we include `paypal` in the common methods; should I add this as a common method too?

assets/js/payment-method-extensions/payment-methods/paypal/index.js
nerrad: It looks like `@woocommerce/block-settings` is not [defined as an external here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/869741051e4d52d9f05ff63f5f0b62e6fea09f33/bin/webpack-helpers.js#L25-L29), which might also impact build sizes for all our bundles!

assets/js/payment-method-extensions/payment-methods/paypal/index.js
nerrad: If this is the mark for paypal standard? Maybe make the icon a bit more specific `paypal-standard` in that case?

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
mikejolley: @nerrad this might actually be redundent - I'm not sure why Strike outputs the credit card icons below the description. Was this in the design? Is it needed?

We discussed using card icons in the tab header which would also make this a duplication.

nerrad: True, I think the icons were moved there as a result of the same conversation that removed them from the tab label... so ya, we can probably remove them from here if we put them back in the tab label.

assets/js/payment-method-extensions/payment-methods/paypal/index.js
mikejolley: PayPal has no specific marks. PayPal does also support credit cards but not clear which. Do you think we should leave out the marks for PayPal entirely?

nerrad: sure, I think it was only added here since it might get used with the cart icon output right?But per the recent slack convo (C8X6Q7XQU-slack-C8X6Q7XQU) maybe we don't want paypal output with the cart icons? I'm okay with rolling without paypal either as a common icon or implemented for cart.

mikejolley: Let's kill it with 🔥

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
mikejolley: I'm happy to do that in this PR - we'll still need this Component in the tab label I think.

nerrad: Hmm so if we put this for the tab label, we might need to have some way of signaling to our payment method component that the `label` prop contains icons for the label. Alternatively, if `label` is a react node, we can `cloneElement` it when generating the tab label and feed `<PaymentMethodIcons>` as a prop to the element so the payment method can consume it as a part of it's label element.

nerrad: I'll do a gist for you.

nerrad: Here's what I'd do: https://gist.github.com/nerrad/3953a14a5b314f2b401a93ce5fad5397

Note, this does mean we might want to go back requiring label be a react node (otherwise we'll have only do `cloneElement` if `label` is not a string.

mikejolley:
Improved now @nerrad 
Screenshot 2020-04-28 at 17 57 21

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
```diff
@@ -32,12 +32,24 @@ const StripeComponent = ( props ) => {
 	return <StripeCreditCard stripe={ stripePromise } { ...props } />;
 };
 
+const StripeLabel = ( { components = {} } ) => {
+	const { PaymentMethodIcons } = components;
+
+	if ( ! PaymentMethodIcons || cardIcons.length === 0 ) {
+		return __( 'Credit/Debit Card', 'woo-gutenberg-products-block' );
+	}
```
nerrad, line 40: good call!

PR #2270, opened 2020-04-21
Abstract Page selector and fix some issues in sidebar.

### LINKED ISSUE #2192: Show page slug as well when multiple pages have the same.
Currently, when you want to select the page that a block should lead to, you will be presented with pages in your website, however, there is no telling which page is which if they have the same name, a suggestion would be to append the page slug if other instances exist.

image

### LINKED ISSUE #2193: Hide "if you want to use this as your default block" notice if that’s already the case.
Currently, we show a notice reminding people of how to make this page the default Cart or Checkout page, however that notice is still there if that's the case, so it should be removed.

image

senadir:
Page selector will now show the page slug if the page title is not set, and will show the page title - page slug if other pages has the same name.

Moves the page selector component in Cart and Checkout blocks to its own component.
Fixes a bug that kept the "set this page as your default" notice even if it was set (context always retuned 0 for current post).
image
Fixing that unveiled another bug in the page selector condition logic, so it was removed:
The condition would not show the proceed to checkout button (or cart) if:
- The current page is the default Cart page (or checkout) and the select value is 0 (the default value), this was not spotted because the current page ID always retuned 0 because it was being called outside the context.

Some styles didn't apply to the checkout options because they were wrapped in a `.editor-styles-wrapper` which only applies to the editor content, not the sidebar.

Fixes #2192 
Fixes #2193

1. Make sure the select page still shows.
2. Make sure they save correctly.
3.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/edit.js
+++ b/assets/js/blocks/cart-checkout/cart/edit.js
@@ -21,6 +21,7 @@ import {
 	CartProvider,
 } from '@woocommerce/base-context';
 import { __experimentalCreateInterpolateElement } from 'wordpress-element';
+import { useRef } from '@wordpress/element';
 import { getAdminLink } from '@woocommerce/settings';
 import { previewCart, cartBlockPreview } from '@woocommerce/resource-previews';
 
@@ -38,6 +39,7 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 		checkoutPageId,
 	} = attributes;
 	const { currentPostId } = useEditorContext();
+	const { current: savedCheckoutPageId } = useRef( checkoutPageId );
 	return (
 		<InspectorControls>
 			{ currentPostId !== CART_PAGE_ID && (
@@ -107,7 +109,9 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 					/>
 				</PanelBody>
 			) }
-			{ ! ( currentPostId === CART_PAGE_ID && checkoutPageId === 0 ) && (
+			{ ! (
+				currentPostId === CART_PAGE_ID && savedCheckoutPageId === 0
+			) && (
 				<PageSelector
 					pageId={ checkoutPageId }
 					setPageId={ ( id ) =>
--- a/assets/js/blocks/cart-checkout/cart/edit.js
+++ b/assets/js/blocks/cart-checkout/cart/edit.js
@@ -107,21 +107,24 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 					/>
 				</PanelBody>
 			) }
-			<PageSelector
-				pageId={ checkoutPageId }
-				setPageId={ ( id ) => setAttributes( { checkoutPageId: id } ) }
-				defaultPageId={ CART_PAGE_ID }
-				labels={ {
-					title: __(
-						'Proceed to Checkout button',
-						'woo-gutenberg-products-block'
-					),
-					default: __(
-						'WooCommerce Checkout Page',
-						'woo-gutenberg-products-block'
-					),
-				} }
-			/>
+			{ ! ( currentPostId === CART_PAGE_ID && checkoutPageId === 0 ) && (
+				<PageSelector
+					pageId={ checkoutPageId }
+					setPageId={ ( id ) =>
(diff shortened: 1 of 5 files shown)


assets/js/components/utils/index.js
```diff
@@ -177,3 +177,22 @@ export const getTerms = ( attribute ) => {
...
+ * @param  {string} page.title.raw Page title.
+ * @param  {string} page.slug      Page slug.
+ * @param  {Array}  pages          Array of all pages.
+ * @return {string}                Formatted page title to display.
+ */
+export const formatTitle = ( page, pages ) => {
+	if ( ! page.title.raw ) {
+		return page.slug;
+	}
+	const hasSimilar =
```
mikejolley, line 195: From what I'm seeing, this isn't checking for "similar" results but duplicates..maybe this should be called `unique`, and is true when no duplicates exist?

assets/js/components/page-selector/index.js
```diff
@@ -0,0 +1,52 @@
...
+			return select( 'core' ).getEntityRecords( 'postType', 'page', {
+				status: 'publish',
+				orderby: 'title',
+				order: 'asc',
+				per_page: 100,
+			} );
+		}, [] ) || null;
+	// We should not render this if this page is the default one page
+	// and the link to option is set to default.
+	if ( ! ( currentPostId === defaultPageId && pageId === 0 ) && pages ) {
```
mikejolley, line 26: I think having the above logic _here_ limits reuse of the page selector. This use case is going to be specific to cart and checkout blocks. Can we move it to the block's edit.js instead? Including the use of editor context.

assets/js/components/utils/index.js
Aljullu: Nit: instead of `unique`, what about `isUnique`? I think it makes it easier to understand the function returns a boolean.

senadir:
> Also a bug, if you change the select to use the default page, the option dissapears instantly. Should it remain until page reload/save?

This will only happen if this page is already set to the default page and your previous choice was not the default, I'm not sure there is a direct way to do this, I can hack it with a useState or a useRef.

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -8,20 +8,20 @@ import {
...
 import { SHIPPING_ENABLED, CART_PAGE_ID } from '@woocommerce/block-settings';
 import BlockErrorBoundary from '@woocommerce/base-components/block-error-boundary';
 import {
 	EditorProvider,
 	useEditorContext,
 	CartProvider,
 } from '@woocommerce/base-context';
-import { useSelect } from '@wordpress/data';
 import { __experimentalCreateInterpolateElement } from 'wordpress-element';
+import { useRef } from '@wordpress/element';
```
senadir, line 24: To avoid any future conflicts, I opted to load this from `@wordpress/element`

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -38,17 +38,8 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
...
-			return select( 'core' ).getEntityRecords( 'postType', 'page', {
-				status: 'publish',
-				orderby: 'title',
-				order: 'asc',
-				per_page: 100,
-			} );
-		}, [] ) || null;
 	const { currentPostId } = useEditorContext();
-
+	const { current: savedCheckoutPageId } = useRef( checkoutPageId );
```
senadir, line 42: We save the supplied (saved) value in a ref and use it, this will ensure it will get updated when changed, another method could have been
```
const [ savedCheckoutPageId ] = useState( checkoutPageId );
```
But this pattern is more readable and known.

assets/js/components/page-selector/index.js
```diff
@@ -0,0 +1,50 @@
...
+	const pages =
+		useSelect( ( select ) => {
+			return select( 'core' ).getEntityRecords( 'postType', 'page', {
+				status: 'publish',
+				orderby: 'title',
+				order: 'asc',
+				per_page: 100,
+			} );
+		}, [] ) || null;
+	// We should not render this if this page is the default one page
```
mikejolley, line 22: Comment is no longer needed here.

PR #2273, opened 2020-04-22
Inject block attributes rather than save them

### LINKED ISSUE #1997: Inject Block `data-` attributes to prevent future invalidation
See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1995#issuecomment-601111747

mikejolley:
Currently, we pass saved attributes to our frontend code using HTML `data-` attributes which get added to the saved block markup via `save`. 

See: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/blocks/cart-checkout/cart/index.js#L46-L67

The big problem with this is that when block attributes change, get added or removed, or deprecated, we either **cause block invalidation errors** or need to write lengthly backwards compatibility code to handle the migration. 

This PR contains an alternative solution; rather than save attributes to the block, it _dynamically injects attributes via the PHP render function_. That means our frontend blocks still have access to the same `data-` attributes, we just add them during render automagically.

Closes #1997 

**Note, this invalidates current cart/checkout blocks. I have no implemented this for the All Products block which is in production.**

1. Create cart/checkout blocks.
2. Set attributes.
3. Check attributes persist on frontend.


src/BlockTypes/AbstractBlock.php
```diff
@@ -94,6 +94,36 @@ public function enqueue_editor_assets() {
...
+	 * @return string Rendered HTML attributes.
+	 */
+	protected function get_html_data_attributes( array $attributes ) {
+		$data = [];
+
+		foreach ( $attributes as $key => $value ) {
+			if ( is_bool( $value ) ) {
+				$value = $value ? 'true' : 'false';
+			}
+			$data[] = 'data-' . esc_attr( strtolower( preg_replace( '/(?<!\ )[A-Z]/', '-$0', $key ) ) ) . '="' . esc_attr( $value ) . '"';
```
nerrad, line 121: Why is the `preg_replace` needed here? I'm assuming you want to handle something like this, `a KeyHere` so it becomes, `a-key-here` (where `esc_attr` takes care of removing the spaces)?

mikejolley:
So @senadir and @nerrad, counter points to those suggestions:

> maybe we can leverage them (at least it ensures we don't end up with stale implementation)

> As hinted by Nadir, we might be able to avoid a lot of additional code here if we just use this parser client side.

This shouldn't be something which goes stale - we have specific requirements for the `data-` attributes in terms of allowed characters and case. The suggested parser would add overhead (particularly on the frontend?), whereas this approach takes advantage of native features (`dataset`) without the need to implement  a parsing library.

> Why is the preg_replace needed here? I'm assuming you want to handle something like this, a KeyHere so it becomes, a-key-here (where esc_attr takes care of removing the spaces)?

Yes this. The preg_replace should handle converting a prop to the correct `data-` attribute, and so that `dataset` converts it back to the correct prop on the frontend. The $key will be in camelCase.

If you feel using the parser would be better, feel free to followup, but let's get this merged before we cause too many invalidation errors for those testing beta.

PR #2276, opened 2020-04-22
Support itemized taxes in cart/checkout

### LINKED ISSUE #2211: Cart and Checkout not respecting Display tax totals: itemized
**Describe the bug**
If you have multiple taxes, and choose the setting Display tax totals: itemized, Cart and Checkout blocks will not show each tax alone, it will still shown them as itemized, as compared to classical Cart and Checkout in which each tax is shown alone.

Blocks:
image

Shortcodes:
image

**To Reproduce**

1. Setup two taxes.
2. In tax settings, set Display tax totals to itemized
3. view Cart and Cart block and compare.

mikejolley:
Adds support for the itemized tax setting in core.

Fixes #2211

### Screenshots

Screenshot 2020-04-22 at 14 30 26

1. In WC > Settings > Taxes, change the tax totals setting to show itemized taxes instead of a single total. You might want to add some extra taxes so rows would be displayed.
2. View cart page.
3. See tax rows/names.

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/totals-taxes-item/index.js
+++ b/assets/js/base/components/cart-checkout/totals/totals-taxes-item/index.js
@@ -3,7 +3,10 @@
  */
 import { __ } from '@wordpress/i18n';
 import PropTypes from 'prop-types';
-import { TAXES_ENABLED } from '@woocommerce/block-settings';
+import {
+	TAXES_ENABLED,
+	DISPLAY_ITEMIZED_TAXES,
+} from '@woocommerce/block-settings';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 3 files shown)


assets/js/base/components/cart-checkout/totals/totals-taxes-item/index.js
```diff
@@ -3,26 +3,44 @@
...
 
 const TotalsTaxesItem = ( { currency, values } ) => {
-	const { total_tax: totalTax } = values;
+	const { total_tax: totalTax, tax_lines: taxLines } = values;
 
 	if ( ! TAXES_ENABLED ) {
 		return null;
 	}
 
+	if ( ! DISPLAY_ITEMIZED_TAXES ) {
```
senadir, line 23: Do we need this option? I think it we can always return `tax_lines` as an array (even if it has one) since this logic already lives in the backend

senadir: On a second look, it doesn’t look like it’s included, separate lines will always be returned, we can probably change `get_tax_lines` to return a total in that case or return all items, and skip `total_tax` in the API, but I'm not sure  I like this as well.

As a note, this is just a suggestion, not a blocker.

mikejolley: I prefer using the totals field as that is what the field is intended for, and to note, it might have a different "label" than the tax lines field.

PR #2278, opened 2020-04-23
Add notice errors to add to cart button and All products

### LINKED ISSUE #2242: All Products - No notices when add to cart fails
See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2234#pullrequestreview-394744614

When you add to cart, but it fails, there is no notice informing the user. We could implement snackbars here, and likely need notices context.

To reproduce, have a sold individually product, then try to add to cart multiple times. The first add works, the rest do not.

senadir:
Adds notices to All products, shows an error if `addToCart` returns an error, rethrows the error from `*addToCart` action, clears previous notices if adding was successful (At first I cleared notices before adding, this caused a jump if this add also has an error).

Fixes #2242

### Screenshots
Note: I still have this issue https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2206 present, not sure how to get it fixed (installing Woo assets didn't fix it).
image

1. Have a product that is sold individually or is limited in quantity.
2. try adding above the allowed amount in All Products.
3. you should see a notice.
4. try adding a normal product, notice should be gone and the error will be added.

### Changelog

> An error notice will be shown in All Product if the customer is trying to add a product above stock or sold individually.

### DIFF
--- a/assets/js/base/hooks/use-store-add-to-cart.js
+++ b/assets/js/base/hooks/use-store-add-to-cart.js
@@ -61,6 +61,7 @@ export const useStoreAddToCart = ( productId ) => {
 				addErrorNotice( error.message, {
 					context: 'wc/all-products',
 					id: 'add-to-cart',
+					isDismissible: true,
 				} );
 			} )
 			.finally( () => {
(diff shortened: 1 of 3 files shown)


senadir:
> build/all-products-frontend.js | 71.5 kB | +53.7 kB (75%)

Not fun, investigating.

senadir:
The increase seems to be related to `Notice` now pulling in the `wordpress-component`, with many dependencies, I've filled in the issue https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2280, there isn't much we can do here, we might see if we can move away from it.

senadir:
> Should we decode the message from the API?
 done in b33da95

I was thinking we should probably decode at the notice implementation level, but decided not to later since we can end up over decoding.

Issue #1752, opened 2020-02-17
Migrate CountryInput and StateInput to ComboboxControl


Aljullu:
Follow-up of #1574 and #1575.

In the legacy country and state controls, you could write the country name in an input field inside the select so it filtered the available options:

![imatge](https://user-images.githubusercontent.com/3616980/72445014-57343600-37b1-11ea-9195-ed852dc6951a.png)

We would like the same for the CountryInput and StateInput components. In order to do that, we should migrate from the `CustomSelectControl` to the `ComboboxControl` that might be introduced in Gutenberg: https://github.com/WordPress/gutenberg/pull/19657.

haszari:
[`ComboBoxControl` has shipped / merged](https://developer.wordpress.org/block-editor/components/combobox-control/), but may not be available in the package yet: https://github.com/WordPress/gutenberg/issues/24443

nerrad:
@Aljullu should the description or the title of this issue be changed to reflect the work you submitted to Gutenberg around related needs we have (I think it was a different component you ended up working on)?

Aljullu:
> I think it was a different component you ended up working on

@nerrad Correct, I worked on `FormTokenField`, the PR can be found here: https://github.com/WordPress/gutenberg/pull/29110. The idea was to mimic the behavior of our `DropdownSelector` component (used in the Attribute Filter block) in Gutenberg's `FormTokenField`. That would allow us to eventually replace `DropdownSelector` with `FormTokenField` (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3762).

Having said that, I'm not sure `FormTokenField` is the best fit for `CountryInput` and `StateInput`. `FormTokenField` works best when multiple values are allowed, while `ComboboxControl` seems to be better suited when only a single option can be selected like in this case.

PR #2281, opened 2020-04-23
Migrate toward a native Select and update WordPress Components to the latest version.

senadir:
This PR updates `@wordpress/components` to the latest version 9.5, it also replaces `CustomSelectControl` to a native solution:

Firefox:
image

Chrome:
image

The case for a native select is better accessibility, better autocomplete without hacks, and less external dependencies, `CustomSelectControl` was built for when you need a custom-styled option, the docs suggests that you should use a native select if that's not the case.

Fully migrating to 9.5 still needs some work (we need to handle dependencies not handled by WP 5.3 that are extracted anyway).

At best, we can discuss this PR regarding the select component, revert the wordpress-component upgrade, merge this and handle WordPress components in a separate PR.


senadir:
> Size Change: +20.3 kB (1%)
Well, one thing we can conclude from updating is that it won't fix the bundle size issues, it increases them.

senadir:
> An issue I'm facing is that when running npm i && npm start in this branch, the blocks stop working completely. 😕 Can you reproduce it as well? (It might be something weird in my environment)

yes, this is expected since I didn't yet resolve the dependency issue, one solution would be to install Gutenberg to test this.

> After using autocomplete, the arrow is not visible.

autocomplete seems to add a background, the arrow is using a background, I think we can migrate to a hardcoded svg like the checkbox.

> Also, if you look at Spain, you will notice it's slightly cut-off at the bottom.

I did notice it but couldn't fix it yet

Aljullu:
@senadir one more question: could the `remove dashicon` commit be split in another PR or it's tied to this one? I see the bundle size wins are quite huge :tada: so it might make sense to review it separately so both changes don't block each other. But I'm not sure if that's possible, so asking. :slightly_smiling_face:

senadir:
> @senadir one more question: could the remove dashicon commit be split in another PR 

yes I have it in another branch that I didn't push yet.

nerrad:
> One of the reasons to use the CustomSelectControl was that we were planning to migrate to the ComboboxControl at some point because it would allow searching through the options (see #1752). Has this been discarded for now? If it has, I agree there is no need to use CustomSelectControl over a native select. I would like to know @nerrad's opinion on that.

It looks like the biggest performance gains at this point will be eliminating the dashicons import which could be done in a separate pull.

Since we're already close to release and we've ironed out a lot of the bugs related to `CustomSelectControl`, I'd prefer to keep that in place and focus our efforts on a `ComboboxControl` (#1752) replacement at some point (but not for the initial release of the blocks).

If there were larger gains from converting to native select then I'd be more supportive of making the change.

senadir:
@nerrad @Aljullu I'd appreciate if we can give this PR another shot.

nerrad:
@senadir there's various discussion in this pull already. Is there something we are doing right now that increases the importance of surfacing this? Have you considered the last comment I made in this PR, what feedback do you have regarding it?

senadir:
Yes I did consider that comment and here's my take form it:

- We fixed dashicon but still see some bundlesize for things like downshift (which is used here and only here).
- We stopped working on this because we were close to releasing the blocks, we should be in a comfortable position now to iterate on it (in a cooldown period maybe).
- a combobox still causes the same performance issues without any obvious benefits (except typing, which I feel is a wrong use for that component).

Ultimately the goal is to have a lighter native component when a custom one doesn't offer much value.

nerrad:
> We fixed dashicon but still see some bundlesize for things like downshift (which is used here and only here)

How much impact does downshift add on the size?

> a combobox still causes the same performance issues without any obvious benefits (except typing, which I feel is a wrong use for that component).

Help me understand better why it's the wrong use for the component.

I've added the cooldown label to this - I think it's worthwhile to at least see:

a. How using native selects impacts bundlesize.
b. What the support is for the styling on native selects across browsers.
c. Have a concrete list of what we gain and what we lose (and then decide whether that's an acceptable trade-off).

nerrad:
@senadir is this still something you feel is worth investigating (see [my comment here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2281#issuecomment-734786575))?

PR #2282, opened 2020-04-23
Migrate REST API usage to the Store API

mikejolley:
Upon reviewing the REST API overrides in blocks, to see what needed to be ported to the REST API itself, I found that most endpoints simply needed to expand access capabilities to authors/non-admin users. 

As well as expanding access rights, some of the endpoints were then needing to shape data to hide private things, or return data in a specific opinionated format. This didn't seem like a good solution to me, so I worked through each endpoint to see if the **Store API** could be used instead. Store API is public, so there should be no auth issues in doing this.

I had success with the existing Store endpoints (products, attributes) so decided to port over some of the other endpoints too, creating new Store API endpoints for tags, categories, reviews all following similar practices and only allowing READ access to public data.

This PR contains that work :)

- Introduces endpoints for tags, categories, reviews
- Expands some endpoints with missing data and params
- Removes all REST API overrides
- Refactors some HOCS which used Rest API
- Switches all APIFetches to Store API

1. Test a variety of blocks including reviews by x, featured product, products by attribute/tag/category.
2. Confirm all products block works
3. Double check cart/checkout is still functional. I didn't touch too much there but did move some code around.


senadir:
Conflict on initial commit 😆

mikejolley:
@senadir Not marked for review yet, fixing tests.

mikejolley:
@senadir I'm going to do some more testing tomorrow and I had to comment out the tests for with-attributes - I redid the HOC with hooks and it isn't compatible with existing tests. Marking for review though.

assets/js/components/utils/index.js
```diff
@@ -24,23 +20,21 @@ const getProductsRequests = ( {
 	queryArgs = [],
 } ) => {
 	const defaultArgs = {
-		per_page: IS_LARGE_CATALOG ? 100 : -1,
+		per_page: IS_LARGE_CATALOG ? 100 : 0,
```
senadir, line 23: Wouldn’t 0 return no product?

src/StoreApi/Routes/ProductCategories.php
```diff
@@ -0,0 +1,52 @@
...
+ */
+
+namespace Automattic\WooCommerce\Blocks\StoreApi\Routes;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * ProductCategories class.
+ */
+class ProductCategories extends AbstractTermsRoute {
```
senadir, line 16: This class is not implementing the count property right.

in Master:
Screen Shot 2020-04-24 at 12 53 14 PM

In this branch:
Screen Shot 2020-04-24 at 12 32 33 PM

src/StoreApi/Utilities/ProductQuery.php
```diff
@@ -31,17 +31,36 @@ public function prepare_objects_query( $request ) {
 			'paged'               => $request['page'],
 			'post__in'            => $request['include'],
 			'post__not_in'        => $request['exclude'],
-			'posts_per_page'      => $request['per_page'],
+			'posts_per_page'      => $request['per_page'] ? $request['per_page'] : -1,
```
senadir, line 34: I know that if `per_page` is set to 0, this will evaluate to -1, but why does that? wouldn't make sense to let the client use the same logic as WP?

assets/js/components/utils/index.js
mikejolley: Nope. I noticed some functions treat 0 as "unlimited" and adopted that instead of -1 which also makes validation easier.

src/StoreApi/Utilities/ProductQuery.php
mikejolley: it's easier to check ! per_page then set to -1 than handle 0 and -1 separately. The schema validates from 0 and has a message than leaving blank, or using 0, maps to unlimited.

Some functions in WP use 0 instead of -1 - I think the terms query object.

mikejolley:
Good catch @senadir I am using the count from WP, but some counts enhance to include children terms. Will revise.

PR #2286, opened 2020-04-24
Avoid loading Assets API during REST requests

### LINKED ISSUE #2043: Performance: avoid unnecessary bootstrap during API requests.

It appears that the following code can be skipped for REST API requests:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/src/Domain/Bootstrap.php#L64-L93

Don't bother registering asset scripts during API requests where they certainly won't be enqueued.

**Additional context**

In testing in my local environment, commenting out the highlighted lines during a REST API request resulted in `7727` fewer function calls and a `0.27s` faster response time. The biggest savings comes from skipping several different file system operations.

mikejolley:
This is another attempt at https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2248. That PR was reverted because it interfered with saving the post in admin when using blocks, I think due to not loading some assets in admin.

This PR supercedes https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2260 which is not in sync with master.

- Updated bootstrap to register classes in the container, but only init assets on non-API requests.
- Updated bootstrap to skip file operations/build checks on rest requests.
- Moved all Payments code to Payments/API for better organisation.
  - Added hooks to inject script dependencies
  - Added hooks to inject script data
- Made RestApi class non static so it can be part of the containers.
- Removed `remove_core_blocks` since we support WC > 4

Closes #2043

**To test:**
- Smoke test adding some blocks and saving in the editor
- Check cart/checkout blocks are functioning to confirm API is working 
- Confirm payment methods still display on checkout


mikejolley:
Thanks, going to merge this and rebase my other PRs which touch same code :)

PR #2287, opened 2020-04-25
Table creation validation for install routine

### LINKED ISSUE #2279: Cart block: Removing item from cart gives js error
**Describe the bug**
When I have a list of items in the cart(built using the Cart block) and I click on "Remove item", the item fades and freezes. There is a JS error in the console. Though this does not happen for all items, I could reproduce it many times. I haven't been able to find the pattern, though it happened definitely for certain products on my local installation.

**To Reproduce**

1. Add a few different products to the cart.
2. View cart
3. Click on the "Remove item" link of some of the products
4. See error in JS Console, though not for all products. The item fades, but freezes.
`Uncaught (in promise) SyntaxError: Unexpected token < in JSON at position 0`

**Expected behavior**
No error should be seen and the product should be removed from the cart without any issues.
**Screenshots**
https://d.pr/i/sHTZ7b

* Core version: 5.4
* WooCommerce version: 3.7; 4.0.1(tried on both)
* WooCommerce Blocks version: 2.6.0-beta.1 
* Site language: English
* Any other plugins installed: Stripe, WC Subscriptions

* OS: Mac
* Browser Chrome
* Version Version 81.0.4044.11

mikejolley:
Improves the install routine to protect against missing table errors in stock reservation class;

- Includes create table method which returns if the table exists or not
- New db_schema_version to track schema updates
- If table creation fails, an admin notice is added
- If table creation fails, db_schema_version is not incremented
- Reserve stock class only runs logic if db_schema_version is satisfied

Fixes #2279

The admin notice informs the user which table could not be created on which database, by which user:

Dashboard ‹ Local Testing — WordPress 2020-04-25 08-32-47

1. Delete the wc_reserve_stock table prior to running this PR
2. Checkout branch
3. Table will be created, stock functionality continues to work

### DIFF
--- a/src/Library.php
+++ b/src/Library.php
@@ -58,9 +58,10 @@ class Library {
 			return;
 		}
 
-		$table_name = $wpdb->prefix . 'wc_reserved_stock';
-		$collate    = $wpdb->has_cap( 'collation' ) ? $wpdb->get_charset_collate() : '';
-		$exists     = self::maybe_create_table(
+		$show_errors = $wpdb->hide_errors();
+		$table_name  = $wpdb->prefix . 'wc_reserved_stock';
+		$collate     = $wpdb->has_cap( 'collation' ) ? $wpdb->get_charset_collate() : '';
+		$exists      = self::maybe_create_table(
 			$wpdb->prefix . 'wc_reserved_stock',
 			"
 			CREATE TABLE {$wpdb->prefix}wc_reserved_stock (
@@ -73,6 +74,10 @@ class Library {
 			) $collate;
 			"
 		);
+
+		if ( $show_errors ) {
+			$wpdb->show_errors();
+		}
 
 		if ( ! $exists ) {
 			return self::add_create_table_notice( $table_name );
@@ -99,7 +104,6 @@ class Library {
 			return true;
 		}
 
-		$wpdb->hide_errors();
 		$wpdb->query( $create_sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
 
 		return in_array( $table_name, $wpdb->get_col( 'SHOW TABLES', 0 ), true );
--- a/src/Library.php
+++ b/src/Library.php
@@ -54,7 +54,7 @@ class Library {
 		$schema_version    = 260;
 		$db_schema_version = (int) get_option( 'wc_blocks_db_schema_version', 0 );
 
-		if ( $db_schema_version < $schema_version ) {
+		if ( $db_schema_version > $schema_version ) {
 			return;
 		}
 
--- a/src/Library.php
+++ b/src/Library.php
@@ -49,24 +49,19 @@ class Library {
 	 * Set up the database tables which the plugin needs to function.
 	 */
 	public static function maybe_create_tables() {
-		$db_version = get_option( 'wc_blocks_db_version', 0 );
+		global $wpdb;
 
-		if ( version_compare( $db_version, \Automattic\WooCommerce\Blocks\Package::get_version(), '>=' ) ) {
+		$schema_version    = 260;
+		$db_schema_version = (int) get_option( 'wc_blocks_db_schema_version', 0 );
+
+		if ( $db_schema_version < $schema_version ) {
 			return;
 		}
 
-		global $wpdb;
-
-		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
-
-		$wpdb->hide_errors();
-		$collate = '';
-
-		if ( $wpdb->has_cap( 'collation' ) ) {
-			$collate = $wpdb->get_charset_collate();
-		}
-
-		dbDelta(
+		$table_name = $wpdb->prefix . 'wc_reserved_stock';
+		$collate    = $wpdb->has_cap( 'collation' ) ? $wpdb->get_charset_collate() : '';
+		$exists     = self::maybe_create_table(
+			$wpdb->prefix . 'wc_reserved_stock',
 			"
 			CREATE TABLE {$wpdb->prefix}wc_reserved_stock (
 				`order_id` bigint(20) NOT NULL,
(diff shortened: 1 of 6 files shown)


mikejolley:
Tests are failing on older installs for some reason. Will check it out monday.

mikejolley:
Tests are fixed - it was our test suite at fault.

src/Library.php
```diff
@@ -49,24 +49,19 @@ public static function define_tables() {
...
-		$db_version = get_option( 'wc_blocks_db_version', 0 );
-
-		if ( version_compare( $db_version, \Automattic\WooCommerce\Blocks\Package::get_version(), '>=' ) ) {
-			return;
-		}
-
 		global $wpdb;
 
-		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+		$schema_version    = 260;
```
nerrad, line 54: Is this a hardcoded number that will get changed anytime we have a schema update? Should we document this somewhere in the repo docs? Would there be value in having this as a part of the `Package` configuration (so we have one class that's responsible for all version information - including db version)?

src/Library.php
```diff
@@ -79,7 +74,57 @@ public static function maybe_create_tables() {
...
+
+		if ( in_array( $table_name, $wpdb->get_col( 'SHOW TABLES', 0 ), true ) ) {
+			return true;
+		}
+
+		$wpdb->hide_errors();
+		$wpdb->query( $create_sql ); // phpcs:ignore WordPress.DB.PreparedSQL.NotPrepared
+
+		return in_array( $table_name, $wpdb->get_col( 'SHOW TABLES', 0 ), true );
+	}
```
nerrad, line 106: I'm curious, why not just use `maybe_create_table` directly? If it's because `maybe_create_table` is a pluggable function I don't think we should worry about that, there may be legit cases where it's been overridden (and we're not really responsible for those cases imo).  If it's because we're hiding errors here, we can still call `$wpdb->hide_errors` before calling `maybe_create_table` (and return to previous value of `$wpdb->show_errors` after calling).

src/Library.php
```diff
@@ -79,7 +74,57 @@ public static function maybe_create_tables() {
...
+	 * @return bool False on error, true if already exists or success.
+	 */
+	protected static function maybe_create_table( $table_name, $create_sql ) {
+		global $wpdb;
+
+		if ( in_array( $table_name, $wpdb->get_col( 'SHOW TABLES', 0 ), true ) ) {
+			return true;
+		}
+
+		$wpdb->hide_errors();
```
nerrad, line 102: This will hide errors for this and all subsequent queries in the same request. We probably only want to hide the errors for this query only right? The function returns the _current_ value of `$wpdb->show_errors` before changing it so we can use that to reset back after executing the query.

src/StoreApi/Utilities/ReserveStock.php
```diff
@@ -24,6 +38,10 @@ final class ReserveStock {
 	public function get_reserved_stock( \WC_Product $product, $exclude_order_id = 0 ) {
 		global $wpdb;
 
+		if ( $this->db_version < 260 ) {
```
nerrad, line 41: This is used in a lot of places, I wonder if extracting to a method would make the condition a bit more readable. Example:

```php
if ( $this->has_valid_db_version() )
```

src/Library.php
mikejolley: I didn't want to tie this to version so it would run less frequently. Since DB related things live here, maybe best to keep the schema version here too?

It should be bumped when making a change to the dB or creating new tables.

src/Library.php
mikejolley: Because I did some searching and found examples of some people removing the file where this function is found...after installing WordPress. The require seemed risky. 

Since it's pluggable also, and fairly small, I thought it best to include it to minimise chance of conflict.

src/Library.php
nerrad: > I didn't want to tie this to version so it would run less frequently. 

Right I agree you don't want to tie to the plugin version, but I still think it's valuable to keep anything to do with versioning in a single class/file so there's a canonical place to go for that information. Alternatively, could have a designated DB service class that contains the schema and version info (but that might be overkill for the plugin). I'm just wary of `Library` being the place where we throw in all plugin initialization type things.

mikejolley:
> Should there maybe be a delay on future attempts?

What do you mean? Once it's ran and succeeded, no install code is ran any more. It has to be on all requests to catch requests after updates reliably.  That hasn't changed in this PR.

nerrad:
> What do you mean? Once it's ran and succeeded, no install code is ran any more. It has to be on all requests to catch requests after updates reliably. That hasn't changed in this PR.

I mean if the update _fails_, there will continue to be re-attempts to create the db on every request right? The only time there won't be re-attempts is when the table was created successfully.

src/Library.php
mikejolley: Wherever it goes it should be kept with the schema so it's not forgotten. It's all consumed here and has little to do with the Package. 

I'm also not sure how permanent this code is; when this table and class is merged to core, we no longer need it (once minimum support is bumped). This and the ReserveStock class can be removed. 

We could alternatively just track the install of this table specifically and remove the version system with that in mind?

mikejolley:
@nerrad that's correct. Not easy to work around that though. We want to make sure the routine retries once the admin sees it, and fixes the issue. Building a system to delay, add notices and allow retry for this might add too much scope?

src/Library.php
nerrad: Let's just keep it here for now then but keep in mind:

- I think we should keep the version system, it's unknown if we'll have other scenarios where a new table is needed and versioning is the easiest to work with for those cases (less code churn).
- I'm still wary about dumping things like this into Library. I agree that we'll be removing this once the table is in core and minimum required version is bumped so it's likely not too much of a worry here (as long as we are pro-active about removing redundant code).

src/Library.php
nerrad: It looks like `maybe_create_table` is also defined in `wp-admin/includes/upgrade.php` which is probably a more reliable file to require?

nerrad: nah scratch that, it means loading a bunch of other files on every request too (which could be problematic). Let's just roll with what you have (but still address the hidden errors comment I made).

nerrad:
> Building a system to delay, add notices and allow retry for this might add too much scope?

It depends on how problematic it is that this happens on every request after the fail (as in every frontend and every backend - rest api as well?). That sounds problematic to me, especially on a potentially heavy trafficked site. Would there be any concern with only doing table creation attempts on `wp-admin` requests only?

mikejolley:
We can try on admin init certainly. It just means sites that auto update (if that happens?) Will temporarily not reserve stock for orders. For this table it's probably not a big issue.

mikejolley:
This is ready to re-review.

nerrad:
I didn't merge because I'm not sure about the failing travis tests. Doesn't look like they are related and might be something happening because of a change elsewhere?

mikejolley:
@nerrad Ill check again - they pass locally. Once passing Ill merge.

PR #2289, opened 2020-04-26
Pin WooCommerce version

senadir:
Pins WooCommere version to a branch release/4.1, this will prevent package-lock each time we run an npm install and will not install Woo again each time.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29547,8 +29547,8 @@
       }
     },
     "woocommerce": {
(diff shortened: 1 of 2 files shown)


senadir:
The WooCommerce is meant for e2e test utilities, I think I would prefer to pin this to tag instead of a branch (branches could be deleted or updated), keeping this updated is not a concern, for now, if it breaks, our e2e will break, pinning this so it doesn't install a new version every time we run npm install.
As for why a github repo and not an npm package, it's because the package is not yet published, the github repo is the only place to link it.
Pinning this also avoids having package-lock.json changing on every install, we discussed today in slack if we want to remove it or not and decided that we should keep it.

PR #1337, opened 2019-12-06
Update NPM dependencies and script dependency handling

mikejolley:
Updates `package.json` to use latest NPM packages, and latest WordPress dependencies. Also uses latest `serialize-javascript` to fix a reported vulnerability.

With the update to `@wordpress/scripts`, the way in which dependencies are generated has changed from `json` files to `php` files with arrays of data. I've updated our asset handlers to match.

### How to test the changes in this Pull Request:

1. Successful `npm install` and `npm run build`
2. Successful `npm run test`
3. Spot check existing blocks for errors.

PR #2291, opened 2020-04-27
Make all font sizes accessible + add docs about font-size() mixin

### LINKED ISSUE #2284: Add docs about the font-size mixin and when to use it
Coming from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2283#discussion_r414127119.

**Describe the bug**
Add a section in `coding-guidelines.md` specifying we prefer `@include font-size()` instead of plain `font-size` in CSS for a11y reasons.

Aljullu:
Fixes #2284.

Follow-up of #2283.

* Add docs about preferring the `font-size()` mixin instead of px font-sizes.
* Switch all remaining px font-sizes to the mixin.
* Fixes the alignment of input box labels.

### Screenshots

_Before:_

_After:_

Most of the font-sizes changed in this PR affect the editor, where Gutenberg and many themes reset the font size, so it's hard to test. I opened an issue for Storefront, see #1337.

Something to test in the frontend:
1. Create a page with the _Checkout_ block.
2. Change your browser default font size and visit that page in the frontend.
3. Verify input fields are accessible and properly aligned.

### DIFF
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -118,3 +118,7 @@
 	// IE11 doesn't support overflow-wrap neither word-break: break-word, so we fallback to -ms-work-break: break-all.
 	-ms-word-break: break-all;
 }
+
+@function rem($size, $base: 16px ) {
+	@return $size / $base * 1rem;
+}
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -1,4 +1,4 @@
-// Rem output with px fallback
+// Converts a font-size in px to rem and optionally sets a relative line-height.
 @mixin font-size($sizeValue: 16, $lineHeight: false ) {
 	font-size: $sizeValue + px;
 	font-size: ($sizeValue / 16) + rem;
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -3,7 +3,7 @@
 	font-size: $sizeValue + px;
 	font-size: ($sizeValue / 16) + rem;
 	@if ($lineHeight) {
-		line-height: $lineHeight;
+		line-height: $lineHeight / $sizeValue;
 	}
 }
 
--- a/assets/js/base/components/payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/style.scss
@@ -130,7 +130,7 @@
 
 	.wc-block-gateway-input.focused.empty,
 	.wc-block-gateway-input:not(.empty) {
-		padding: $gap-large $gap $gap-smallest;
+		padding: calc(1.25em + #{$gap-smallest}) $gap 0.25em; // With a font-size of 16px, this will be `24px 16px 4px`.
 		& + label {
 			transform: translateY(#{$gap-smallest}) scale(0.75);
 		}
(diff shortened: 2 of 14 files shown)


Aljullu:
> First up is the largest font size, all fields seem to be okay except stripe payment method fields (which is likely outside of our control because this is their component):

Yes... styling those components is a bit tricky because they are inside an iframe and get some styling by default from Stripe. I took another round at this and I think I could get it to display nicely: 75df25063fbc7b2bfd52fcc0524fb550cfaa01f4.

> The following two are at the smallest font-size:

Ah, good catch, smaller font-sizes were not working fine yet (that was happening in `master` as well, so not introduced in this PR). I think the commit linked above and this one: 9cc12b3a26e200d6220ce200095bc4a1d924ae79 should fix most issues with small font-sizes.

> hmm these values seem contrived (I know they likely aren't). Is it something else that can be exposed via variable or mixin?
> Is there a formula you follow to arrive at these em values?

Yeah, completely agree here. I copied the mixin @senadir added in #2320 and updated the `font-size` mixin so it uses relative line-heights, this way all calculations are explicit.

assets/js/base/components/payment-methods/style.scss
```diff
@@ -92,13 +92,13 @@
 	}
 
 	label {
+		@include font-size(16);
 		position: absolute;
-		transform: translateY(#{$gap-small});
+		transform: translateY(0.75em);
```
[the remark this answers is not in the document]
Aljullu: Good question. Currently, this is only used for the text input and select control components, so I think we can wait before creating a variable until we see whether it's a value we use in other parts of the layout or if it's only used in this case.

assets/css/abstracts/_mixins.scss
```diff
@@ -118,3 +118,7 @@
 	// IE11 doesn't support overflow-wrap neither word-break: break-word, so we fallback to -ms-work-break: break-all.
 	-ms-word-break: break-all;
 }
+
+@function rem($size, $base: 16px ) {
```
Aljullu, line 122: I noticed `font-size()` and `rem()` mixins have different signatures (`font-size` requires unitless variables while `rem` requires `px` variables). I think it would be great to unify them, but I didn't want to cause conflicts with #2320, instead I think we can deal with this in a follow-up.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-element-options.js
```diff
@@ -7,6 +7,27 @@ import { useState, useEffect, useCallback } from '@wordpress/element';
...
+const getDocumentStyle = ( property, defaultValue ) => {
+	let documentStyle = {};
+	if (
+		typeof window === 'object' &&
+		typeof window.getComputedStyle === 'function'
+	) {
+		documentStyle = window.getComputedStyle( document.documentElement );
+	}
+	return documentStyle[ property ] || defaultValue;
+};
```
nerrad, line 29: ahh nifty 👏

Aljullu:
Ah, my bad, I broke it in the last commit and didn't notice it. :man_facepalming: Mixins need to be interpolated inside CSS functions. Should be fixed in: 75d6344dd2f298447d1bea3fb21579858e87c305.

assets/css/abstracts/_mixins.scss
senadir: passing units or not shouldn’t be an issue as long as the base is in the same unit-less world, using px made sense to me since we're converting from px.

Aljullu:
Thanks for all the reviews @nerrad. I did a small tweak in f887c7c56ea0d1624d88d14d7361628bfb4799a7 and it should be looking good now.

senadir:
did we ever address or see why All products increased by 50kb? its size increased 4 folds (from 17kb).

Issue #1483, opened 2020-01-03
Order Notes Field and options in the Checkout


nerrad:
The merchant should have the ability to add a order notes field to the checkout block.  The following work will need completed.

* An order notes field input component (which consists of both the order notes input and the checkbox for toggling showing/hiding that input).
* The merchant can enable/disable whether the order notes field shows. It is enabled by default.
* Order notes are always _optional_ for shoppers.  So when the merchant has enabled this feature, the shopper sees a checkbox that they can check to add order notes (selecting shows the order notes input).

nerrad:
Reducing scope by removing from current cycle. Can be added in a later iteration.

haszari:
Is this included in v1 @garymurray @mikejolley ? 

Currently the checkbox is visible and does nothing, so perhaps we can hide it for now. (Spotted when testing #2283)

mikejolley:
We can move this in soon I think. Does not appear to need much effort.

Aljullu:
@LevinMedia according to the designs, the _Add order notes?_ checkbox is displayed inside the _Shipping options_ section:

![imatge](https://user-images.githubusercontent.com/3616980/87442691-e3a2a280-c5f4-11ea-90e1-742dadcc1be4.png)

However, according to the way the shortcode works, the checkbox should be visible even when the order doesn't require shipping. Where should it be shown in those cases?

- Inside the _Billing address_ section?
- Inside the _Payment methods_ section? (Notice this section will not exist if the order price is 0)
- In its own section with another title?
- Or it shouldn't be shown at all and that's a difference with the current shortcode?

![imatge](https://user-images.githubusercontent.com/3616980/87443406-bd313700-c5f5-11ea-9d9c-f4f7b1cb5ced.png)

LevinMedia:
@Aljullu let's add it as a new section below the payment method, I don't think we need a new title. 

<img width="744" alt="Screen Shot 2020-07-15 at 7 59 14 PM" src="https://user-images.githubusercontent.com/4500952/87621497-c2fe4980-c6d5-11ea-8af0-f10a3a18dd66.png">

<img width="704" alt="Screen Shot 2020-07-15 at 7 59 20 PM" src="https://user-images.githubusercontent.com/4500952/87621516-ce517500-c6d5-11ea-9614-1427b12b9eca.png">

PR #2318, opened 2020-04-27
Remove order note

### LINKED ISSUE #2308: Checkout: Order notes component not usable
**Describe the bug**
The order notes component shows on the checkout page, it is unusable currently and you cannot remove it via the block settings:

Screenshot 2020-04-27 at 12 46 04

It should be removed if we are not going to enable it for the initial release.

senadir:
The checkbox for order note was there in we did the skeleton for the checkout, we never implemented but it’s still visible.
We will tackle this later, for now, I'm deleting it so people don't experience a broken feature.

Fixes #2308

image

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -347,13 +347,6 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 										}
 									/>
 								) }
-								{ /*@todo This is not implemented*/ }
-								<CheckboxControl
-									className="wc-block-checkout__add-note"
-									label="Add order notes?"
-									checked={ false }
-									onChange={ () => null }
-								/>
 							</FormStep>
 						) }
 						{ cartNeedsPayment && (


assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -347,13 +347,6 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 										}
 									/>
 								) }
-								{ /*@todo This is not implemented*/ }
-								<CheckboxControl
-									className="wc-block-checkout__add-note"
```
Aljullu, line 352: Should we remove the [CSS selector](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/blocks/cart-checkout/checkout/style.scss#L5) targeting this checkbox?

mikejolley: We can keep it; I think I'll give adding this back a shot during cooldown once all other issues are resolved :) Ref #1483

PR #2320, opened 2020-04-27
Use SVG for checkbox

senadir:
I updated the checkbox to use an svg rather than an icon to better align with the design, the svg is 2 lines of code so no impact there, I also slightly updated the checkbox to well align with the design.

Now:
image

Before:
image


senadir:
80a4be1 adds rems instead of px.

assets/css/abstracts/_mixins.scss
```diff
@@ -118,3 +118,7 @@
 	// IE11 doesn't support overflow-wrap neither word-break: break-word, so we fallback to -ms-work-break: break-all.
 	-ms-word-break: break-all;
 }
+
+@function rem($size, $base: 16px ) {
```
Aljullu, line 122: Nice!

Aljullu, line 122: Nice!

assets/js/base/components/checkbox-control/index.js
```diff
@@ -35,6 +35,15 @@ const CheckboxControl = ( {
 				onChange={ ( event ) => onChange( event.target.checked ) }
 				{ ...rest }
 			/>
+			<svg
+				className="wc-block-checkbox__mark"
+				aria-hidden="true"
+				xmlns="http://www.w3.org/2000/svg"
+				viewBox="0 0 24 24"
+			>
+				<path d="M0 0h24v24H0z" fill="none" />
```
Aljullu, line 44: Could this line be removed?
```suggestion
```

Everything seems to look good for me after removing it:
imatge

senadir: you mean it fixes the IE11 issue?

Aljullu: Ah no, it's not related. I think that `<path>` can be removed because it has `fill="none"`, I couldn't see any difference after removing it.

senadir:
the IE11 issues come from the fact that we're not hiding the input and the IE11 doesn't have support for appearance https://caniuse.com/#feat=css-appearance
I will hide and construct another box.

Aljullu:
> the IE11 issues come from the fact that we're not hiding the input and the IE11 doesn't have support for appearance https://caniuse.com/#feat=css-appearance
> I will hide and construct another box.

Yeah, IE11... :man_facepalming: Another thing you could try is to add a background to `.wc-block-checkbox__mark` so it covers the native checkmark. Not sure how it will work, but sharing the idea in case it helps.

senadir:
I went on and hid the checkout (`opacity: 0; position: absolute`) this should take it out of the flow and hide it without breaking anything hopefully.

Aljullu:
After the last commit I'm seeing the IE11 issue again:
imatge

Can you reproduce @senadir?

senadir:
can we detect IE11 and simply hide the svg?

Aljullu:
It's a hack, but I guess that something like this would work?
```CSS
@media screen and (-ms-high-contrast: active), (-ms-high-contrast: none) {
	.wc-block-checkbox__mark {
		display: none;
	}
}
```

Taken from: https://stackoverflow.com/questions/11173106/apply-style-only-on-ie#36448860.

nerrad:
@senadir this pull hasn't seen any action since April. Are you still intending on finishing this off?

senadir:
yes, it just had very low priority, can we keep it on open until I get back to it and rebase it?

senadir:
@Aljullu I updated this PR, can you give it another look or a check :shipit:

senadir:
I fixed the misalign somewhere else and forget to port it here, fixed now.

assets/js/base/components/checkbox-control/style.scss
```diff
@@ -7,37 +7,45 @@
...
 	}
 
-	.wc-block-components-checkbox__input[type="checkbox"] + .wc-block-components-checkbox__label {
+	.wc-block-components-checkbox__mark {
+		fill: #fff;
+		position: absolute;
+		left: rem(1px);
+		top: rem(5px);
+		width: 16px;
+		height: 16px;
```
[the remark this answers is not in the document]
Aljullu: It looks like this wasn't solved. `width` and `height` should be `em()`. Otherwise this is how it looks like if I set my browser default font size to 32px:
imatge

senadir:
Ah, I mistaken your comment for the two lines above them (that used rem).

haszari:
When testing #2981 I saw a minor issue with checkbox in Twenty Twenty in Safari. It's a bit offset to the right (though I think this looks cool in its own way..). Not sure if that PR includes changes from this PR, mentioning in case this is new bug due to SVG implementation.

Screen Shot 2020-08-14 at 1 41 32 PM

PR #2330, opened 2020-04-28
Import from `@woocommerce/settings` in `@woocommerce/block-settings`

### LINKED ISSUE #2329: Fix `@woocommerce/block-settings` alias not importing from `@woocommerce/settings` external.
While reviewing #2267, in an early iteration, @mikejolley imported from `@woocommerce/block-settings` into the paypal payment method integration and that resulted in a bundle jump of 1.43 kb in size. This was a bit odd to me because for the most part block-settings _should_ be referencing the `@woocommerce/settings` external so the affected size should be minimal.

However, I discovered that block-settings is [actually importing directly](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/869741051e4d52d9f05ff63f5f0b62e6fea09f33/assets/js/settings/blocks/constants.js#L4) from the `@woocommerce/settings` files hence it's bundling that api directly instead of using what's exposed on the external in the build.

nerrad:
Fixes #2329 

See #2329 for details. This pull should shave some bytes off of built bundles.

## To test

This impacts anything that utilizes server values passed through as block settings (so anything that imports data from `@woocommerce/block-settings`. It should be immediately clear if something is awry in testing involving components that use that alias (which include All Products, Checkout, Cart blocks etc).

### DIFF
--- a/assets/js/settings/blocks/constants.js
+++ b/assets/js/settings/blocks/constants.js
@@ -1,7 +1,7 @@
 /**
- * Internal dependencies
+ * External dependencies
  */
-import { getSetting } from '../shared';
+import { getSetting } from '@woocommerce/settings';
 
 export const CURRENT_USER_IS_ADMIN = getSetting( 'currentUserIsAdmin', false );
 export const REVIEW_RATINGS_ENABLED = getSetting(
--- a/assets/js/settings/blocks/store-api-nonce.js
+++ b/assets/js/settings/blocks/store-api-nonce.js
@@ -2,11 +2,7 @@
  * External dependencies
  */
 import apiFetch from '@wordpress/api-fetch';
-
-/**
- * Internal dependencies
- */
-import { getSetting } from '../shared';
+import { getSetting } from '@woocommerce/settings';
 
 // Cache for the initial nonce initialized from hydration.
 let nonce = getSetting( 'storeApiNonce' );


nerrad:
So this did give us some bytes back, not quite the number I expected but still better than nothing!

mikejolley:
> Is the problem that we were bundling code/js related to woo settings instead of/as well as pulling it from a global?

@haszari we have some externals and one of those is `@woocomerce/settings` which is enqueued separately. Webpack maps any instance of `@woocomerce/settings` it finds to the wcSettings global.

By importing from '/shared', webpack isn't mapping anything, it's importing those files directl;y, hence the larger file sizes before the change.

Did I explain that correct :) ¯\_(ツ)_/¯

nerrad:
Mike explained well, to add to what he wrote, `@woocommerce/block-settings` is _not_ built to a separate bundle but is just aliased for builds because it essentially is just convenience wrappers over block settings that are available on the `wc.wcSettings` global (or the `@woocommerce/settings` external. It can be confusing but in this case:

- `@woocommerce/settings` is an alias mapped to external global which is built to it's own bundle.
- `@woocommerce/block-settings` is an alias mapped to an internal path which is imported directly in builds.

So the question might be _why_ do we use the aliases instead of referencing `wc.wcSettings` directly, or in the case of block-settings `assets/js/settings/blocks`? 

- using package format aliases means less code churn if/when in the future those could become actual published packages - the code is more reusable in various contexts.
- related to the above, it's easier for build tools to alias those to different things if necessary.
- it's easier to configure jest to map those to something different or mock if necessary.
- the recent introduction of typescript allows us to map those to specific paths for better code completion.

PR #2348, opened 2020-04-29
Fix low stock badge not breaking in a new line

### LINKED ISSUE #2342: Cart - low stock badge has no margin
I saw this in editor, not sure if it's on front end also (I'm guessing it affects front end also).

Aljullu:
Fixes #2342.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Open the _Cart_ block with a product that is low in stock.
2. Verify the low stock badge renders in its own line (instead of next to the product name).

### DIFF
--- a/assets/js/base/components/cart-checkout/product-name/style.scss
+++ b/assets/js/base/components/cart-checkout/product-name/style.scss
@@ -1,4 +1,6 @@
 .wc-block-product-name {
 	@include font-size(16);
 	@include wrapBreakWord();
+	display: block;
+	width: max-content;
 }


assets/js/base/components/cart-checkout/product-name/style.scss
```diff
@@ -1,4 +1,6 @@
 .wc-block-product-name {
 	@include font-size(16);
 	@include wrapBreakWord();
+	display: block;
+	width: max-content;
```
Aljullu, line 5: This CSS property doesn't work on IE11, but that's only to make the clickable area or the link smaller, so I think it can be comprised inside graceful degradation for old browsers.

PR #2352, opened 2020-04-29
Fix payment method error notices issues

### LINKED ISSUE #2217: Checkout Block: Clear notices in all notice context areas when checkout is submitted.
First reported by @mikejolley  from testing in #2216:

> Notices don't clear if more notices are added. Leave card blank and submit. Then enter card number and set expiry to date in past and submit again.

I agree that submitting should automatically clear all notices in the checkout. It's likely that the current action is only clearing notices within the main `wc/checkout` context area and to fix we need to make sure _all_ notices are cleared on submit.

### LINKED ISSUE #2327: Payment method errors appear twice
**To Reproduce**
1. Go to the _Checkout_ page fill the form but leave the credit card details empty.
2. Click on _Place order_.
3. Notice the same error notice appears twice.

**Screenshots**
Peek 2020-04-28 17-03

Aljullu:
Fixes #2217.
Fixes #2327.

### Screenshots
_Before:_
imatge

_After:_
Peek 2020-04-29 11-43

1. Go to the _Checkout_ page fill the form but leave the credit card details empty.
2. Click on _Place order_.
3. Verify only one error appears.
4. Fill the credit card details with an expired card (for example: `4242424242424242`, `11/11`, `111`).
5. Verify the previous error disappeared and it was replaced by another one.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -140,7 +140,10 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 				id: 'wc-express-payment-error',
 			} );
 		} else {
-			removeNotice( 'wc-express-payment-error' );
+			removeNotice(
+				'wc-express-payment-error',
+				'wc/express-payment-area'
+			);
 		}
 	};
 	// ensure observers are always current.
@@ -274,7 +277,7 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 		// allows for other observers that return true for continuing through
 		// to the next observer (or bailing if there's a problem).
 		if ( currentStatus.isProcessing ) {
-			removeNotice( 'wc-payment-error' );
+			removeNotice( 'wc-payment-error', noticeContexts.PAYMENTS );
 			emitEventWithAbort(
 				currentObservers.current,
 				EMIT_TYPES.PAYMENT_PROCESSING,
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -312,7 +312,7 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 				}
 			} );
 		}
-	}, [ currentStatus, setValidationErrors, setPaymentStatus ] );
+	}, [ currentStatus.isProcessing, setValidationErrors, setPaymentStatus ] );
 
 	/**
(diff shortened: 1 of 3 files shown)


assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -140,7 +140,10 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 				id: 'wc-express-payment-error',
 			} );
 		} else {
-			removeNotice( 'wc-express-payment-error' );
+			removeNotice(
+				'wc-express-payment-error',
+				'wc/express-payment-area'
```
haszari, line 145: Curious about how this `removeNotice` API works - does the context parameter clear out all notices in that context (i.e. clear out all express payment notices), or is it more like a namespace, so `wc-express-payment-error` is only removed if it's in `wc/express-payment-area` context?

If it's the former, can we simplify the API so it clears a whole notice context - i.e. `clearAllNotices( 'wc/express-payment-area' )`?

Related, are there multiple contexts in play on the checkout, and do we want to clear them all on every submit? E.g. `clearAllNotices()` ?

Aljullu: You can see the `removeNotice()` signature here:
https://github.com/WordPress/gutenberg/blob/master/packages/notices/src/store/actions.js#L147

Basically, you pass the id of the notice to remove and (optionally) its context.

In `payment-method-data-context.js `, we are adding notices to different contexts, so when removing them we need to pass the context as well.

> Related, are there multiple contexts in play on the checkout, and do we want to clear them all on every submit? E.g. `clearAllNotices()` ?

We have a system in place to remove all notices from the current context:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/hooks/use-store-notices.js#L40-L47

Used here, for example:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/context/cart-checkout/checkout-state/index.js#L182

There isn't an easy way to remove all notices from another context than the current one. And in fact, I'm not sure we would want to do that even if it was possible. I imagine, for example, a `Subscribe to our newsletter` block in the sidebar that used the notices API as well. It might show an error notice when the wrong email is introduced, but we don't want that notice to disappear when submitting _Checkout_ for example, because they have no relation between.

Aljullu:
Thanks for the reviews @mikejolley and @haszari.

> My only comment; does it make sense to keep these notices dismissable? Side-quest, the dismiss button wraps which looks real odd.

Good point, I created an issue: #2365.

PR #2362, opened 2020-04-29
Payment methods `placeOrderButtonLabel` config

### LINKED ISSUE #2326: Allow payment methods to control Place Order Button text
A new config parameter for payment methods is needed to control the Place Order button text during checkout, when the payment method is active.

Use case, PayPal should rename this text to "Proceed to PayPal".

mikejolley:
Removes the `submitLabel` property from checkout context (which wasn't used by cart nor checkout, despite being implemented) and replaces it with text controllable by Payment Methods/

Since this needs various contexts to work (payments being the main one), I added this to the `useCheckoutSubmit` hook, and then implemented `useCheckoutSubmit` in the Place Order Button component.

For gateways, they just need to include `placeOrderButtonLabel` in the config. When active, this value will be used. Otherwise, the default "Place Order" will be used instead.

Fixes #2326

1. Go to checkout, ensure place order button reads "Place Order"
2. Select PayPal. Order button should update.

### DIFF
--- a/assets/js/base/hooks/checkout/use-checkout-submit.js
+++ b/assets/js/base/hooks/checkout/use-checkout-submit.js
@@ -1,12 +1,39 @@
 /**
  * External dependencies
  */
-import { useCheckoutContext } from '@woocommerce/base-context';
+import { __ } from '@wordpress/i18n';
+import {
+	useCheckoutContext,
+	usePaymentMethodDataContext,
+} from '@woocommerce/base-context';
+import { usePaymentMethods } from '@woocommerce/base-hooks';
 
 /**
- * Returns the submitLabel and onSubmit interface from the checkout context
+ * Returns the submitButtonText, onSubmit interface from the checkout context,
+ * and an indication of submission status.
  */
 export const useCheckoutSubmit = () => {
-	const { submitLabel, onSubmit } = useCheckoutContext();
-	return { submitLabel, onSubmit };
+	const {
+		onSubmit,
+		isCalculating,
+		isBeforeProcessing,
+		isProcessing,
+		isAfterProcessing,
+		isComplete,
+		hasError,
+	} = useCheckoutContext();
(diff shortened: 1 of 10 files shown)


assets/js/base/hooks/checkout/use-checkout-submit.js
```diff
@@ -1,12 +1,39 @@
...
+		isCalculating,
+		isBeforeProcessing,
+		isProcessing,
+		isAfterProcessing,
+		isComplete,
+		hasError,
+	} = useCheckoutContext();
+	const { paymentMethods } = usePaymentMethods();
+	const { activePaymentMethod } = usePaymentMethodDataContext();
+	const paymentMethod = paymentMethods[ activePaymentMethod ] || {};
```
nerrad, line 27: No need to change in this pull, but just got thinking I wonder if it'd be useful to expose the `activePaymentMethodObject` via `usePaymentMethods` (which would internally consume `usePaymentMethodDataContext` to expose it)? It could reduce some code in place likely.

assets/js/base/hooks/checkout/use-checkout-submit.js
```diff
@@ -1,12 +1,39 @@
...
+	return {
+		submitButtonText:
+			paymentMethod?.placeOrderButtonLabel ||
+			__( 'Place Order', 'woo-gutenberg-products-block' ),
+		onSubmit,
+		isCalculating,
+		waitingForProcessing:
+			isProcessing || isAfterProcessing || isBeforeProcessing,
+		waitingForRedirect: isComplete && ! hasError,
+	};
```
nerrad, line 38: Great call on expanding what is exposed from this hook 👏 .

assets/js/base/hooks/checkout/use-checkout-submit.js
mikejolley: yes! I was looking for this but it didn't exist.

PR #2367, opened 2020-04-30
Rename the discount section when discount total is zero

### LINKED ISSUE #2302: Cart: Free shipping coupon - strange UX
**Describe the bug**
When I enter a coupon that allows free shipping, it does "discount" the order by the shipping amount and selects the free shipping option, but it says that the discout is $0 - whereas it should reflect the discounted amount I would think?

**To Reproduce**
Create a coupon that allows free shipping - and view the UX once you enter the coupon amount - the discount shows as $0.

**Expected behavior**
The discounted amount would reflect the value of the shipping amount discounted.

**Screenshots**

Screenshot 2020-04-27 at 11 52 38

mikejolley:
This is awaiting feedback from @garymurray to see if it satisfies the issue or not.

If the Discount total is 0, this renames the total row to "Coupons" which keeps coupon display in the same place, but avoids inferring there is 0 discount.

Fixes #2302

This is instead of splitting out shipping coupons as suggested in the issue, because:

- We don’t have a way in the API right now to return if a coupon grants free shipping.
- Some free shipping coupons might also give a $ discount, so we’d have to also consider that
- We’d have to render 2 coupon sections and feed two separate lists of coupons

Additionally, I have not touched shipping rate ordering since this is user controlled (drag and drop) in WP admin. They can move free shipping first in the list to do what was suggested in #2302

### Screenshots

screenshot_2020-04-30_at_12 38 48
screenshot_2020-04-30_at_12 38 26

1. Apply a free shipping coupon
2. Confirm matches screenshots

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/totals-discount-item/index.js
+++ b/assets/js/base/components/cart-checkout/totals/totals-discount-item/index.js
@@ -30,6 +30,9 @@ const TotalsDiscountItem = ( {
 	}
 
 	const discountTaxValue = parseInt( totalDiscountTax, 10 );
+	const displayDiscountValue = DISPLAY_CART_PRICES_INCLUDING_TAX
+		? discountValue + discountTaxValue
+		: discountValue;
 
 	return (
 		<TotalsItem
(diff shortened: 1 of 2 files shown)


garymurray:
> This is awaiting feedback from @garymurray to see if it satisfies the issue or not.

I am happy with this - we can move forward pending @LevinMedia's design input.

> Additionally, I have not touched shipping rate ordering since this is user controlled (drag and drop) in WP admin. They can move free shipping first in the list to do what was suggested in #2302

Because it seems trying to do this would touch Woo core I agree we'll have to leave it for now.

LevinMedia:
@garymurray @mikejolley Works for me

PR #2370, opened 2020-04-30
Use container queries to load Cart and Checkout responsive styles

### LINKED ISSUE #2274: Block styling should use "mobile" styles if used on narrow themes or within narrow containers
When cart and checkout blocks are used within narrow containers, we still use the full desktop view. It causes odd wrapping.

Since we don't know what container widths various themes use, we should probably toggle on mobile styles if we detect a narrow container.

### LINKED ISSUE #2357: Responsive layout not responding in Gutenberg previews
While testing with Gutenberg 7.9.1 and WooCommerce 4.1.0-rc.1 the cart and checkout preview is not adapting well when using the tablet mobile previews: 

Screenshot 2020-04-29 at 15 27 04

Screenshot 2020-04-29 at 15 14 09

Aljullu:
Fixes #2274.
Fixes #2357.

This PR introduces a new hook: `useContainerQueries()` which allows setting a class name dynamically based on the container width. This will allow our blocks to be responsive based on the container width instead of the viewport width.

### Advantages
* Users might add the _Cart_ or _Checkout_ block in narrow containers (like in #2274, imagine the _Cart_ and _Checkout_ blocks inside a column block). In those cases, blocks will display the mobile styles so they don't break.

| Before | After |
| imatge | imatge |

* Some themes main content area is very narrow and the _Cart_ and _Checkout_ blocks were not playing well. For example, TwentyTwenty.

* With the last Gutenberg release, which allows previewing for `tablet` and `mobile`, our blocks will adapt to the display width so they give a more accurate preview :tada: :

| Before | After |
| imatge | imatge |

* Knowing the block width in the JS side could potentially simplify our markup/styles. If we know a component is only rendered in desktop, we will no longer need to always render it and hide it on mobile with `display: none`.

### A drawback
We lose the opportunity to have responsive 'skeletons'. Skeletons are added from the PHP side so there is no way to know the container width in advance.
Peek 2020-04-30 15-06

Some ideas:
* We could completely remove the skeleton from the PHP side and instead rely on JS loading states. We lose the advantage of the skeleton being loaded immediately, but that would fix the issue completely and we would be able to remove some code.
* Always default to one of the sizes, for example, always showing the `large` layout even on mobile (this is how it's currently done, see screenshot above). That will make the skeleton not to look good in some sizes.
* Use native CSS mediaqueries for the skeleton. That would mean we would need to duplicate a lot of CSS (that could be done with a mixin, but it would still add complexity to the codebase). In addition, it will not fix the issue completely, because mediaqueries will still differ from container queries.

Personally, I would go for the first option. But wanted to get some feedback from the theme before proceeding. cc @mikejolley for feedback as well because if I'm not wrong, you worked on those skeletons. :point_up_2: 

### Next steps

All in all, I think the advantages overpass the drawback. But happy to discuss. :slightly_smiling_face: 

If this PR is accepted, I think we could open a follow-up in order to update all other blocks to use container queries instead of media queries when possible.

### Testing steps

1. Test the _Cart_ and _Checkout_ blocks in different viewport sizes.
2. Verify they still have responsive styles and those are applied according to the container width.

### DIFF
--- a/assets/js/base/hooks/use-container-queries.js
+++ b/assets/js/base/hooks/use-container-queries.js
@@ -1,0 +1,45 @@
+/**
+ * External dependencies
+ */
+import { useResizeObserver } from 'wordpress-compose';
+
+/**
+ * Returns a resizeListener element and a class name based on its width.
+ * Class names are based on the smaller of the breakpoints:
+ * https://github.com/WordPress/gutenberg/tree/master/packages/viewport#usage
+ * _Note: `useContainerQueries` will return an empty class name `` until after
+ * first render_
+ *
+ * @return {Array} An array of {Element} `resizeListener` and {string} `className`.
+ *
+ * @example
+ *
+ * ```js
+ * const App = () => {
+ * 	const [ resizeListener, containerQueryClassName ] = useContainerQueries();
+ *
+ * 	return (
+ * 		<div className={ containerQueryClassName }>
+ * 			{ resizeListener }
+ * 			Your content here
+ * 		</div>
+ * 	);
+ * };
+ * ```
+ */
+export const useContainerQueries = () => {
+	const [ resizeListener, { width } ] = useResizeObserver();
+
+	let className = '';
+	if ( width > 782 ) {
+		className = 'is-large';
+	} else if ( width > 600 ) {
+		className = 'is-medium';
+	} else if ( width > 480 ) {
+		className = 'is-small';
+	} else if ( width ) {
+		className = 'is-mobile';
+	}
+
+	return [ resizeListener, className ];
+};
--- a/assets/js/base/hooks/use-container-queries.js
+++ b/assets/js/base/hooks/use-container-queries.js
@@ -7,6 +7,8 @@ import { useResizeObserver } from 'wordpress-compose';
  * Returns a resizeListener element and a class name based on its width.
  * Class names are based on the smaller of the breakpoints:
  * https://github.com/WordPress/gutenberg/tree/master/packages/viewport#usage
+ * Values are also based on those breakpoints minus ~80px which is aproximately
+ * the left + right margin in Storefront with a font-size of 16px.
  * _Note: `useContainerQueries` will return an empty class name `` until after
  * first render_
  *
@@ -31,11 +33,11 @@ export const useContainerQueries = () => {
 	const [ resizeListener, { width } ] = useResizeObserver();
 
 	let className = '';
-	if ( width > 782 ) {
+	if ( width > 700 ) {
 		className = 'is-large';
-	} else if ( width > 600 ) {
+	} else if ( width > 520 ) {
 		className = 'is-medium';
-	} else if ( width > 480 ) {
+	} else if ( width > 400 ) {
 		className = 'is-small';
 	} else if ( width ) {
 		className = 'is-mobile';
--- a/assets/css/abstracts/_breakpoints.scss
+++ b/assets/css/abstracts/_breakpoints.scss
@@ -5,7 +5,7 @@
 
 // Think very carefully before adding a new breakpoint.
 // The list below is based on wp-admin's main breakpoints
-// See https://github.com/WordPress/gutenberg/tree/master/packages/viewport#breakpoints
+// See https://github.com/WordPress/gutenberg/tree/master/packages/viewport#usage
 $breakpoints: 480px, 600px, 782px, 960px, 1280px, 1440px;
 
 @mixin breakpoint( $sizes... ) {
--- a/assets/js/base/components/cart-checkout/form-step/style.scss
+++ b/assets/js/base/components/cart-checkout/form-step/style.scss
@@ -12,8 +12,8 @@ $line-offset-from-circle-size: 8px;
 	background: none;
 	margin: 0;
 
-	@include breakpoint( ">782px" ) {
-		padding-right: $gap-larger;
+	.is-large & {
+		padding-right: $gap-large;
 	}
 }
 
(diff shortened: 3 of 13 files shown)


senadir:
I forget to comment on skeletons, I think we can skip them since we're hydrating the block from the server anyway so it's shown for a much shorter period.

Aljullu:
Thank you for all the feedback. :ok_hand: I will try to answer it below.

> The mobile style seems to keep in a little early which is most apparent on checkout.

Yes... I agree. I tried to use the same [breakpoints](https://github.com/WordPress/gutenberg/tree/master/packages/viewport#usage) from `@wordpress/viewport`, but probably it doesn't make a lot of sense considering those are media query breakpoints and we are doing container queries.

I tried to ease this issue with three small commits: 572140bdcf532436ceb199ea077d60d8b83028df, 29e95dfd5d42bc45804450429c2279588e97b541 and 14e67a5abd05cba6001ecfaf92d2a0b2ae2c7f7e. Let me know if they help @mikejolley.

> Is it possible to toggle the order summary off by default on mobile view? And possibly hide product summaries? The checkout form was not visible on my instead because it was pushed down. This felt like undesirable behaviour.

That's probably more a design question for @LevinMedia?

> Why didn't we go with the native container query approach?

@senadir What do you mean with the native approach? :thinking: 

> Regarding skeleton. Ideally having it from the server allows it to show up quicker for the user (as Mike pointed out), however I understand the difficulty with media views in that case. Could we serve two skeletons from the server and what is shown depends on the media query? So we'd have some vanilla js that executes before react is loaded that shows the appropriate skeleton on page load for the media query?

@nerrad I'm not sure to understand you here. You mean using media queries for skeletons in contrast of container queries, which will be used for the rendered block? Or you proposed writing a vanilla JS snippet that would run before React is loaded that would add the container query class?

The first approach has the issue that there is no reliable way to map media queries with container queries, so that would not really fix the bug in all cases.

The second approach would work and sounds like a good solution to me. I imagine we will want to load it inline so it's executed immediately, correct? Should that JS snippet be inlined on the block render function?

nerrad:
> Or you proposed writing a vanilla JS snippet that would run before React is loaded that would add the container query class?

That :). Whether or not you inline it in the block `frontendRender` function depends on whether it can be abstract enough to be a reusable pattern for all blocks with skeletons (likely).

senadir:
> What do you mean with the native approach?

I think the word native was a bit misplaced there, currently, there is no approved spec for element queries and one has implemented them natively, it seems the discussion died since 2017, the most promising spec out there is by [Tommy Hodgins](https://tomhodgins.github.io/element-queries-spec/element-queries.html).
Another hub for resources and implementation is in https://elementqueries.com/ and https://github.com/eqcss/eqcss

eqcss uses script tags, that's one syntax but can [still work inside CSS](https://github.com/eqcss/eqcss/wiki/Using-EQCSS-with-CSS-Preprocessors), another one is CSS-Element-Queries which uses native CSS syntax and a slightly similar approach as yours currently 
https://marcj.github.io/css-element-queries/
Another approach that I would have preferred is a postcss loader that extracts the queries from our CSS, see https://github.com/ZeeCoder/container-query

So in general, a "native" approach is any approach in which we directly include the actual query inside our CSS, instead of hoping for breakpoint classes, both are fine, but was wondering which solution will provide us the most value, currently the approach you went with the most liberal as it only gives you a class and you're free to do any scss with it.

assets/js/base/hooks/use-container-queries.js
```diff
@@ -0,0 +1,47 @@
...
+ */
+export const useContainerQueries = () => {
+	const [ resizeListener, { width } ] = useResizeObserver();
+
+	let className = '';
+	if ( width > 700 ) {
+		className = 'is-large';
+	} else if ( width > 520 ) {
+		className = 'is-medium';
+	} else if ( width > 400 ) {
```
haszari, line 40: Longer term might be nice to have a way for client code to parameterise/customise the breakpoints. This is great though :)

Aljullu: Yes, that's a great idea. For the _Cart_ and _Checkout_ block, this breakpoints were working fine, but if we see this hook being used in other blocks that require other breakpoints, we could let the consumer pass the breakpoints/classnames as a parameter.

Aljullu:
> That :). Whether or not you inline it in the block `frontendRender` function depends on whether it can be abstract enough to be a reusable pattern for all blocks with skeletons (likely).

I created a follow-up so we don't block this PR: #2402.

> So in general, a "native" approach is any approach in which we directly include the actual query inside our CSS, instead of hoping for breakpoint classes, both are fine, but was wondering which solution will provide us the most value, currently the approach you went with the most liberal as it only gives you a class and you're free to do any scss with it.

Yeah, I agree, but there are two reasons that make me prefer the approach in this PR:
- It uses a `@wordpress/*` dependency instead of a third-party one. That would ease upstreaming changes, keeping the bundle size under control, etc.
- Container queries don't have a standard syntax yet, there are several proposals but the way queries will be written is not set in stone. Before adding new [at-rules](https://developer.mozilla.org/en-US/docs/Web/CSS/At-rule) that might make our CSS invalid, I felt that using class names was a simpler approach and didn't require any learning.

---

I will proceed merging this PR now given that it was already approved. We can iterate/refactor it if needed in follow-ups.

LevinMedia:
>Is it possible to toggle the order summary off by default on mobile view? And possibly hide product summaries? The checkout form was not visible on my instead because it was pushed down. This felt like undesirable behaviour.

@mikejolley @Aljullu I think we can go ahead and collapse the order summary by default in this case. I think if we do that, we don't have to hide the products descriptions, either.

PR #2371, opened 2020-04-30
Use Reakit in payment tabs

### LINKED ISSUE #1474: Improve or replace Tabs Component.

senadir:
This PR add reakit to our tabs, which is a more accessible solution. thanks for @diegohaz for the help in debugging this.

Fixes #1474

#### Accessibility

- [ ] I've tested using a screen reader (I didn't).
~~- [ ] All animations respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)~~ no animation was added.
~~- [ ] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)~~ no color changed.

### Screenshots

ezgif com-video-to-gif (7)

1. In checkout try changing the payment method, or tabbing between them.
2. try 1 or 2 or 3 methods.


assets/js/base/components/payment-methods/payment-methods.js
```diff
@@ -70,7 +70,7 @@ const PaymentMethods = () => {
 		currentPaymentMethodInterface.current = paymentMethodInterface;
 	}, [ paymentMethods, paymentMethodInterface, activePaymentMethod ] );
 
-	const getRenderedTab = useCallback(
+	const getRenderedTab = useMemo(
 		() => ( selectedTab ) => {
```
diegohaz, line 74: I think it's the same as:
```suggestion
	const getRenderedTab = useCallback(
		( selectedTab ) => {
```

assets/js/base/components/tabs/index.js
```diff
@@ -1,84 +1,82 @@
...
+					tabId={ `${ instanceId }-${ name }` }
 					className="wc-block-components-tabs__content"
-					tabIndex={ 0 }
 				>
-					{ children( selected ) }
-				</div>
-			) }
+					{ children( name ) }
+				</TabPanel>
+			) ) }
```
nerrad, line 79: If we're mapping anyways, I think we should just pass the content as an object from the PaymentMethods component and just have our `<Tabs>` component receive the content as a part of the `tabs` object (where panel content would be on the `content` prop.  It'll be less confusing.

The only reason I used a render prop for children originally was for lazy rendering of the tab content on demand, but the current implementation renders all tabs at once (which might be better for screen readers maybe).

assets/js/base/components/tabs/index.js
```diff
@@ -1,93 +1,79 @@
...
-	);
-	// Maybe set selected tab if it's empty but we have tabs or an initial tab
-	// This covers instances when the selected value isn't initialized on mount.
-	useEffect( () => {
-		if ( ! selected && tabs.length > 0 ) {
-			setSelected(
-				initialTabName || ( tabs.length > 0 ? tabs[ 0 ].name : '' )
-			);
-		}
-	}, [ selected, tabs, initialTabName ] );
```
senadir, line 37: reakit should cover this by default.

assets/js/base/components/tabs/index.js
```diff
@@ -1,93 +1,79 @@
...
 		return null;
 	}
-	const handleClick = ( tabKey ) => {
-		setSelected( tabKey );
-		onSelect( tabKey );
-	};
-	const selectedTab = tabs.find( ( tab ) => tab.name === selected );
-	if ( ! selectedTab ) {
-		throw new Error( 'There is no available tab for the selected item' );
-	}
```
senadir, line 48: tabs and content are returned on the same object, this felt redundant and caused us to guard against empty Tabs from `paymentMethod`
is there a scenario in which this will trigger?

PR #2374, opened 2020-04-30
Store billing address to local state and persist globally to prevent loss of data

### LINKED ISSUE #2361: "Use same address for billing" does not persist
I submitted an order with different info in the shipping and billing sections. When following up with a new order the shipping info was retrieved but the checkbox for "Use same address for billing" was checked and the billing address use in the previous order was not retrieved:

Checkout-new – My WordPress Site

Checkout-new – My WordPress Site 2

Screenshot 2020-04-29 at 17 12 27

mikejolley:
Finally done. Took longer than I wanted but I managed to fix some other issues at the same time.

1. This PR adds a local state for address data in the form of `useCheckoutAddress`. This hook persists data globally, but also stores local values so billing address is never lost.
2. If logged in, if you have a billing address separate from shipping address, the "use shipping as billing" checkbox is automatically unticked
3. I found an issue that triggered validation on all fields on change. I fixed this by using shallow equal for state/country data which are objects such as `{key: "NY", name: "New York"}`. Now validation runs on the current field only. Performance++
4. I found the issue that was causing validation to be stale once fields were unmounted. We were missing a dependency in the `useEffect` call doing the unmount 🤦 cc @Aljullu 

Fixes #2361

Smoke testing checkout, but also covering these cases:

1. Guest user. On checkout you should have the usual behaviour where "use shipping as billing" is ticked and billing is blank.
2. Logged in user. Edit your profile and use a different billing address. On checkout, "use shipping as billing" is unchecked, and both addresses are populated in the form.
3. Test inputting address and confirming validation is working and values are persisted.

### DIFF
--- a/assets/js/base/components/select/validated.js
+++ b/assets/js/base/components/select/validated.js
@@ -4,6 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { useEffect } from 'react';
 import { useValidationContext } from '@woocommerce/base-context';
+import { useShallowEqual } from '@woocommerce/base-hooks';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { withInstanceId } from '@woocommerce/base-hocs/with-instance-id';
@@ -30,13 +31,18 @@ const ValidatedSelect = ( {
 } ) => {
 	const selectId = id || 'select-' + instanceId;
 	errorId = errorId || selectId;
+
+	// Prevents re-renders when value is an object, e.g. {key: "NY", name: "New York"}
+	const currentValue = useShallowEqual( value );
+
 	const {
 		getValidationError,
 		setValidationErrors,
 		clearValidationError,
 	} = useValidationContext();
+
 	const validateSelect = () => {
-		if ( ! required || value ) {
+		if ( ! required || currentValue ) {
 			clearValidationError( errorId );
 		} else {
 			setValidationErrors( {
@@ -50,7 +56,7 @@ const ValidatedSelect = ( {
 
 	useEffect( () => {
 		validateSelect();
-	}, [ value ] );
+	}, [ currentValue ] );
 
 	// Remove validation errors when unmounted.
 	useEffect( () => {
@@ -68,7 +74,7 @@ const ValidatedSelect = ( {
 				'has-error': error.message && ! error.hidden,
 			} ) }
 			feedback={ <ValidationInputError propertyName={ errorId } /> }
-			value={ value }
+			value={ currentValue }
 			{ ...rest }
 		/>
 	);
--- a/assets/js/base/components/select/validated.js
+++ b/assets/js/base/components/select/validated.js
@@ -63,7 +63,7 @@ const ValidatedSelect = ( {
 		return () => {
 			clearValidationError( errorId );
 		};
-	}, [] );
+	}, [ errorId ] );
 
 	const error = getValidationError( errorId ) || {};
 
--- a/assets/js/base/components/select/validated.js
+++ b/assets/js/base/components/select/validated.js
@@ -63,7 +63,7 @@ const ValidatedSelect = ( {
 		return () => {
 			clearValidationError( errorId );
 		};
-	}, [] );
+	}, [ clearValidationError ] );
 
 	const error = getValidationError( errorId ) || {};
 
--- a/assets/js/base/components/select/validated.js
+++ b/assets/js/base/components/select/validated.js
@@ -63,7 +63,7 @@ const ValidatedSelect = ( {
 		return () => {
 			clearValidationError( errorId );
 		};
-	}, [ clearValidationError ] );
+	}, [] );
 
 	const error = getValidationError( errorId ) || {};
 
--- a/assets/js/base/components/state-input/state-input.js
+++ b/assets/js/base/components/state-input/state-input.js
@@ -4,7 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import PropTypes from 'prop-types';
 import { decodeEntities } from '@wordpress/html-entities';
-import { useCallback, useEffect } from '@wordpress/element';
+import { useCallback } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -30,15 +30,7 @@ const StateInput = ( {
 				name: decodeEntities( countryStates[ key ] ),
 		  } ) )
 		: [];
-	// @todo: remove this code block when issue https://github.com/woocommerce/woocommerce/issues/25854 is merged
-	// Defaults to the first state when selecting a country with states, this is here
-	// until a bug in Woo core is fixed.
-	// see: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1919
-	useEffect( () => {
-		if ( ! value && options.length ) {
-			onChangeState( options[ 0 ].key );
-		}
-	}, [ country ] );
+
 	/**
 	 * Handles state selection onChange events. Finds a matching state by key or value.
 	 *
--- a/assets/js/base/components/text-input/validated.js
+++ b/assets/js/base/components/text-input/validated.js
@@ -37,6 +37,7 @@ const ValidatedTextInput = ( {
 
 	const textInputId = id || 'textinput-' + instanceId;
 	errorId = errorId || textInputId;
+
 	const validateInput = ( errorsHidden = true ) => {
 		if ( inputRef.current.checkValidity() ) {
 			clearValidationError( errorId );
@@ -63,7 +64,7 @@ const ValidatedTextInput = ( {
 		return () => {
 			clearValidationError( errorId );
 		};
-	}, [] );
+	}, [ clearValidationError ] );
 
 	const errorMessage = getValidationError( errorId ) || {};
 	const hasError = errorMessage.message && ! errorMessage.hidden;
--- a/assets/js/base/components/text-input/validated.js
+++ b/assets/js/base/components/text-input/validated.js
@@ -64,7 +64,7 @@ const ValidatedTextInput = ( {
 		return () => {
 			clearValidationError( errorId );
 		};
-	}, [] );
+	}, [ errorId ] );
 
 	const errorMessage = getValidationError( errorId ) || {};
 	const hasError = errorMessage.message && ! errorMessage.hidden;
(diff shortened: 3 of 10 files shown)


assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -0,0 +1,100 @@
...
+ *
+ * @param {Object} address1 First address.
+ * @param {Object} address2 Second address.
+ */
+const isSameAddress = ( address1, address2 ) => {
+	const diff = Object.keys( defaultAddressFields ).filter( ( field ) => {
+		return address1[ field ] !== address2[ field ];
+	} );
+	return diff.length === 0;
+};
```
senadir, line 23: this could be simplified with `Array.every`

```suggestion
const isSameAddress = ( address1, address2 ) => {
	return Object.keys( defaultAddressFields ).every( ( field ) => address1[ field ] === address2[ field ] );
};
```

assets/js/base/components/select/validated.js
```diff
@@ -50,14 +56,14 @@ const ValidatedSelect = ( {
...
-	}, [ value ] );
+	}, [ currentValue ] );
 
 	// Remove validation errors when unmounted.
 	useEffect( () => {
 		return () => {
 			clearValidationError( errorId );
 		};
-	}, [] );
+	}, [ clearValidationError ] );
```
nerrad, line 66: 👏

assets/js/base/components/state-input/state-input.js
```diff
@@ -30,15 +30,7 @@ const StateInput = ( {
...
 		: [];
-	// @todo: remove this code block when issue https://github.com/woocommerce/woocommerce/issues/25854 is merged
-	// Defaults to the first state when selecting a country with states, this is here
-	// until a bug in Woo core is fixed.
-	// see: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1919
-	useEffect( () => {
-		if ( ! value && options.length ) {
-			onChangeState( options[ 0 ].key );
-		}
-	}, [ country ] );
```
nerrad, line 41: One question about removing this, are we going to require WC 4.0.1+ for the blocks? Right now we have 4.0 (does that include patch versions automatically)?

assets/js/base/components/text-input/validated.js
```diff
@@ -63,7 +64,7 @@ const ValidatedTextInput = ( {
 		return () => {
 			clearValidationError( errorId );
 		};
-	}, [] );
+	}, [ clearValidationError ] );
```
nerrad, line 67: 👏

assets/js/base/context/cart-checkout/validation/index.js
```diff
@@ -132,9 +142,14 @@ export const ValidationContextProvider = ( { children } ) => {
...
-		} );
+		if (
+			validationErrors[ property ] &&
+			validationErrors[ property ].hidden !== false
+		) {
+			updateValidationError( property, {
+				hidden: false,
+			} );
+		}
 	};
```
nerrad, line 153: `updateValidationError` already does a check for `property` on `validationErrors`. I'm guessing you added this logic to handle only updating if hidden is something different. I think we should move this logic check into `updateValidationError` (do a `shallowEqual` on the values to determine whether to update the state?) so that we can trust state is only going to be updated if necessary (avoiding this kind of gotcha anywhere else `updateValidationError` might be used).  Note, I do realize this is partial updates that are merged with the existing state, however, I still think we should do shallowEqual comparison on the object partials.

assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -0,0 +1,100 @@
...
+	const [ billingFields, setBillingFields ] = useState( billingData );
+
+	// This tracks the state of the "shipping as billing" address checkbox. It's
+	// initial value is true (if shipping is needed), however, if the user is
+	// logged in and they have a different billing address, we can toggle this off.
+	const [ shippingAsBilling, setShippingAsBilling ] = useState(
+		() =>
+			needsShipping &&
+			( ! customerId || isSameAddress( shippingAddress, billingData ) )
+	);
```
nerrad, line 51: What if user is logged in and they've never completed a purchase before (so there's no billing address associated with the customer)? Should we also check for an empty state on the billingData?

assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -0,0 +1,100 @@
...
+
+	// Pushes to global state when changes are made locally.
+	useEffect( () => {
+		setShippingAddress( shippingFields );
+
+		if ( shippingAsBilling ) {
+			setBillingData( shippingFields );
+		}
+	}, [ shippingFields ] );
```
nerrad, line 60: `shippingAsBilling` needs set here as a dependency.

assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -0,0 +1,100 @@
...
+		setBillingData( billingFields );
+	}, [ billingFields ] );
+
+	useEffect( () => {
+		if ( shippingAsBilling ) {
+			setBillingData( shippingFields );
+		} else {
+			setBillingData( billingFields );
+		}
+	}, [ shippingAsBilling ] );
```
nerrad, line 72: `shippingFields` and `billingFields` should be dependencies here.

assets/js/base/components/state-input/state-input.js
mikejolley: The bug was an edge case, I don't think we need to force a version requirement even though it's fixed in a core fix release.

assets/js/base/hooks/checkout/use-checkout-address.js
mikejolley: In this case there would be no addresses, so the same address condition would pass and the box would be checked (checked default is what we want). I tested by passing 2 empty addresses to this function.

mikejolley:
> email address field value can't be changed (either logged in or incognito).

Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2374/commits/b8910a852c1cfe303f41621a992044ec84e112bd

mikejolley:
> If I clear some required billing fields while logged in and click to submit the order, the order processes instead of warning about missing fields and halting the processing. While logged out (via incognito), the email field is blank and processing is halted and scrolls to the first invalid field but it doesn't show any invalid error messaging or highlights. I also don't know if it was halted because address fields were missing or the email field is blank (which I can't enter a value for).

@nerrad I unchecked the box and as soon as I clear a required field I get a validation notice which blocks checkout. Wasn't able to proceed as you described.

assets/js/base/context/cart-checkout/validation/index.js
mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2374/commits/e227e5e22c5f2d1db13b37ace1c0b1f35e6693ad

nerrad:
> I clear a required field I get a validation notice which blocks checkout

Correct, I get that too, but it doesn't block clicking the button to proceed for me and also if I click the button it just finished the checkout for me. I'll try again when I review again and see what happens (and grab a screen recording if it does).

nerrad:
Here's some behaviour I'm seeing (good news is I can't seem to reproduce the submitting with invalid fields issue I got earlier - might have been a fluke). Note this is testing against a `npm run build` build.

If I clear the email address (which I can do now 👍 for logged in user, haven't tested logged out yet), then submit, I get a notice and the form scrolls to the email address field, however there is no validation error message in the field.

Screen Shot 2020-05-04 at 3 08 15 PM

If I uncheck the checkbox, and all the fields are empty - then submit the form, the form will scroll to the first empty required field, but again there is no inline validation error message (nor in this case any "main" error notice).

Screen Recording 2020-05-04 at 03 09 PM


To verify this isn't something also occurring on master, here's screenshots from the same scenario against a `npm run build` of current master:

Image 2020-05-04 at 3 15 19 PM

Screen Recording 2020-05-04 at 03 17 PM

mikejolley:
This is rebased and fixed now - ready for rereview.

assets/js/type-defs/contexts.js
```diff
@@ -43,7 +43,7 @@
  *                                                        selected.
  * @property {CartShippingAddress}  shippingAddress       The current set
  *                                                        address for shipping.
- * @property {function()}           setShippingAddress    A function for setting
+ * @property {Function}             setShippingAddress    A function for setting
```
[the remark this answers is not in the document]
mikejolley: Yeah this was to fix warnings in my ide.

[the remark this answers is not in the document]
mikejolley: There are other `@property {Function}` comments in that file. In the review you may have only seen the function() one. For example:

```
 @property {Function}             setSelectedRates
```

and 

```
@property {Function}                setBillingData
```

So yes, I will ignore for now and they can all be updated if necessary :)

assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -0,0 +1,107 @@
...
+			setShippingAddress( shippingFields );
+		}
+
+		const newBillingData = shippingAsBilling
+			? shippingFields
+			: billingFields;
+
+		if ( ! isShallowEqual( newBillingData, billingData ) ) {
+			setBillingData( newBillingData );
+		}
```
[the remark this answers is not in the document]
mikejolley: _State_ might not be; it can be an object. This should be switched to lodash isEqual perhaps.

assets/js/type-defs/contexts.js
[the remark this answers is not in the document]
nerrad: Hah, you're right I didn't see them (or remember them). It's very likely I switched to `Function` too when I encountered the same errors and before I knew about the `function(string) => Object` syntax :)

assets/js/base/hooks/checkout/use-checkout-address.js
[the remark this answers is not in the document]
nerrad: `isEqual` is slow, so if we can avoid it, that would be good because this is a fairly hot code path. Still, you could try it and see how it impacts things.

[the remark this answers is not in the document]
mikejolley: @nerrad I was able to avoid it for the validation class in the end. But the address one I did use it, because of state and country and possibly other data where we need to compare the entire address object, not just a new/specific property.

Issue #2325, opened 2020-04-27
Checkout block: displays login message when the store has that option turned off


pmcpinto:
**Describe the bug**
The checkout page displays a message inviting the user to login during checkout, when that option is turned off in settings.

**To Reproduce**
1. WooCommerce - Settings - Accounts & Privacy
2. Make sure that "Allow customers to place orders without an account" is turned on
3. Make sure that "Allow customers to log into an existing account during checkout" is turned off
4. See the message "Already have an account? Log in." in the checkout frontend. 

**Expected behavior**
Not seeing a message suggesting to login when a store doesn't have that option turned on.
**Screenshots**
https://d.pr/i/5aC7oS

PR #2375, opened 2020-05-01
Don't show checkout block for logged out users unless store is configured to allow guest checkout

### LINKED ISSUE #2252: Checkout Block does not respect "Guest checkout" option
**Describe the bug**
Currently, our block will always let you checkout as a guest or a user, even if Guest checkout option is set to limit that.

**To Reproduce**

1. disable guest checkout in WooCommerce settings.
2. Try to checkout from an incognito window.
3. You will still be able to checkout.

**Expected behavior**
You should not be able to checkout.

### LINKED ISSUE #2349: Log in button at checkout block displayed for logged in user.
**Describe the bug**

Saved Stripe cards are shown at the same time as `Log in` button at checkout block.

**To Reproduce**

1. Log in as admin.
2. Open new tab with the shop.
3. Go to `My account` page and add a payment method with Stripe. E.g. 4242 4242 4242 4242 (any date and cvc).
4. Add some products to the cart and go to checkout page.
5. Saved payment method is shown as well as `Log in` link at the top of the block.

**Expected behavior**
Expect either `Log in` link is not shown or saved cards not displayed.

**Screenshots**
image
image

* Core version: 5.3.2 
* WooCommerce version: 4.0.1
* WooCommerce Blocks version: 2.6.0-beta.1
* Site language: en_US

WooCommerce PayPal Checkout Gateway |   | by WooCommerce – 1.6.20 – 1.6.21 is available – Not tested with the active version of WooCommerce
-- | -- | --
WooCommerce Stripe Gateway |   | by WooCommerce – 4.3.3
WooCommerce Blocks |   | by Automattic – 2.6.0-beta.1
WooCommerce Subscriptions |   | by WooCommerce – 3.0.3
WooCommerce |   | by Automattic – 4.0.1
WordPress Importer |   | by wordpressdotorg – 0.7

* OS: OS X 10.15.4
* Browser Safari
* Version  13.1 

**Additional context**

haszari:
Fixes #2252

Fixes #2349

There are a few existing options affecting checkout when user is logged out - this PR handles `Allow customers to place orders without an account`. If this is not enabled, then the checkout is only displayed when users are logged in.

This PR also has a fix for #2349 - easy to fix here since it also needs `CURRENT_USER_LOGGED_IN`. 

See also related issue #2325.

### Screenshots

Screen Shot 2020-05-04 at 11 21 35 AM

#### Test #2252 - guest checkout store option works correctly for checkout block
1. Set up a page with checkout block. 
2. Go to `WooCommerce > Settings > Accounts & Privacy` and toggle `Allow customers to place orders without an account`.
3. View checkout page when logged in and logged out / anon (use incognito window or other browser. Ensure that checkout is not available when logged out as per your store settings.

Compare with existing checkout. Note that the checkout shortcode allows user to sign up as part of checkout - so that behaviour is different (see [comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2252#issuecomment-622208649) and new issue #2377).

#### Test #2349 - checkout should only show login prompt when appropriate and necessary
1. Go to `WooCommerce > Settings > Accounts & Privacy` and enable `Allow customers to log into an existing account during checkout`.
2. Add products to cart & view checkout when logged in and anon. Should show login prompt as appropriate.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -48,6 +48,7 @@ import {
 import { getSetting } from '@woocommerce/settings';
 import withScrollToTop from '@woocommerce/base-hocs/with-scroll-to-top';
 import {
+	CURRENT_USER_LOGGED_IN,
 	CHECKOUT_SHOW_LOGIN_REMINDER,
 	CHECKOUT_ALLOWS_GUEST,
 	CHECKOUT_ALLOWS_SIGNUP,
@@ -161,6 +162,17 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 
 	if ( ! isEditor && ! hasOrder ) {
 		return <CheckoutOrderError />;
+	}
+
+	if (
+		! CURRENT_USER_LOGGED_IN &&
+		! CHECKOUT_ALLOWS_GUEST &&
+		! CHECKOUT_ALLOWS_SIGNUP
+	) {
+		return __(
+			'You must be logged in to checkout.',
+			'woo-gutenberg-products-block'
+		);
 	}
 
 	const loginPrompt = () =>
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -164,9 +164,19 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 	}
 
 	if ( ! CURRENT_USER_LOGGED_IN && ! CHECKOUT_ALLOWS_GUEST ) {
-		return __(
-			'You must be logged in to checkout.',
-			'woo-gutenberg-products-block'
+		return (
+			<>
+				{ __(
+					'You must be logged in to checkout. ',
+					'woo-gutenberg-products-block'
+				) }
+				<a href="/wp-login.php">
+					{ __(
+						'Click here to log in.',
+						'woo-gutenberg-products-block'
+					) }
+				</a>
+			</>
 		);
 	}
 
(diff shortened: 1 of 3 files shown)


mikejolley:
I also found 2349 and fixed it in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2374, but I didn't need to introduce global settings because we already expose the customerId via the CheckoutProvider.

nerrad:
I appreciate picking up restricting guest checkout in this pull (to respect the checkout setting), however I think in those cases we should include a link to login when the checkout is restricted.

haszari:
> I also found 2349 and fixed it in #2374, but I didn't need to introduce global settings because we already expose the customerId via the CheckoutProvider.

Aha - nice @mikejolley. I've reworked this PR to use that same approach for now (so this PR is self-consistent), I'm happy to rebase or remove this if needed :)

> I appreciate picking up restricting guest checkout in this pull (to respect the checkout setting), however I think in those cases we should include a link to login when the checkout is restricted.

Good point @nerrad  – I've added this now, see new screenshot in PR description.

nerrad:
Just a note, we should probably set the `redirect_to` parameter on the url you use for the link so that the shopper is returned to the checkout after logging in. We should probably do the same for the `loginPrompt` variable in `checkout/block.js` as well.

haszari:
>  Just a note, we should probably set the redirect_to parameter on the url you use for the link so that the shopper is returned to the checkout after logging in. We should probably do the same for the loginPrompt variable in checkout/block.js as well.

That's good feedback @nerrad – I think I'll get that sorted in this PR (for both) rather than add a follow up.

PR #2378, opened 2020-05-01
Change summary word count via API parameters

### LINKED ISSUE #2293: Cart: Shorten product description in demo data
The product description we are using in the demo data is to long. I think we can shorten it to and make it clearer what it would actually be. I would suggest using: `Product short description` instead.

mikejolley:
This PR allows the summary word count returned from the API (for products and cart items) to be changed based on a new param. Example usage:

```curl
GET https://local.wordpress.test/wp-json/wc/store/products/32?summary_max_words=10
```

The default is 15, and this is what cart will be using right now.

Fixes #2293

What has changed in this PR:

- Added `summary_max_words` param to cart and product APIs.
- API classes have a `get_item_response` method. This was changed thoughout to accept the request parameter.
- Added `get_item_responses` method to handle collections more easily.
- In thinking we needed to support this in `withProduct` HOC for the Featured Product block, I refactored that as a functional component and updated tests. Yay!...
- ...however, after doing this I found out the Featured Product block was using `summary` when it should be using `short_description` like the PHP frontend 🤦 In changed that to the correct description but left the updated HOC in place because it's an improvement.
- Made sure All Products block requests 150 word limit like before. In a future iteration we could perhaps make this configurable.

1. Give a product a long short description and add it to your cart.
2. Check cart and checkout page limits it's length
3. Check all products block (with a summary inner block) shows more of the description. 150 words like before.
4. Test the endpoint (example above) adjusts the summary.

Note: The summary logic returns the first paragraph if the content contains several paragraphs of text. This is not word limited and already has tests.

### DIFF
--- a/assets/js/base/components/product-list/index.js
+++ b/assets/js/base/components/product-list/index.js
@@ -55,6 +55,7 @@ const generateQuery = ( { sortValue, currentPage, attributes } ) => {
 		catalog_visibility: 'catalog',
 		per_page: columns * rows,
 		page: currentPage,
+		summary_max_words: 150,
 	};
 };
 
--- a/assets/js/blocks/featured-product/block.js
+++ b/assets/js/blocks/featured-product/block.js
@@ -316,7 +316,9 @@ const FeaturedProduct = ( {
 						<div
 							className="wc-block-featured-product__description"
 							dangerouslySetInnerHTML={ {
-								__html: product.summary,
+								__html: !! product.short_description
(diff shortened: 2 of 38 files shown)


mikejolley:
Removing this from review/qa - I want to try a different solution.

mikejolley:
Closing in favour of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2387

PR #2384, opened 2020-05-02
Add Todo Probot config

nerrad:
I installed the [`@todo` probot github app to our repo](https://github.com/JasonEtco/todo) for managing todos. The idea here is that we'll ensure any pulls merged with @todo comment blocks will have issues automatically created for them. This ensures follow-ups are documented.

I edited a todo in a file to see how this will work. Note:

- When `@todo` comment (case-insensitive) is in a commit pushed to a pull, the pull will just receive a comment. After the pull is merged, any remaining incomplete todos will have issues created for them. This ensures that you can have _in progress_ work in a pull with todos that get handled before merge.
- If you include `@body` after the todo, the text following that tag will get added as additional text in the comment (or issue description).

### DIFF
--- a/.github/config.yml
+++ b/.github/config.yml
@@ -1,0 +1,3 @@
+todo:
+  blobLines: 10
+  label: false
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -40,7 +40,9 @@ const preparePaymentData = ( paymentData ) => {
 };
 
 /**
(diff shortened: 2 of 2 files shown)


nerrad:
Ugh, it cc's the entire `woocommerce` group!

senadir:
just set `autoAssign` to false, we don't need CC in issues since we triage and assign when needed
https://github.com/JasonEtco/todo#available-options

nerrad:
Hmm ya, I thought it defaulted to false but it doesn't - yup will do that.

PR #2387, opened 2020-05-03
Remove API summaries in favour of client side code

### LINKED ISSUE #2293: Cart: Shorten product description in demo data

mikejolley:
In 2378 I found that we were not using the Product Summary PHP class on the frontend; only in API responses consumed by the client side code.

Due to this, I think it's better to move this logic to the client side, and use the short/long description instead. This will give us more control over description rendering without additional calls to the API to get an updated summary.

This is an alternative solution to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2378 which Fixes #2293.

1. Test atomic product summary block within All Products. Summary word limit is 150.
2. Check items in cart/checkout. Summary word limit is 15.


mikejolley:
We should probably get this in before the next beta if possible.

assets/js/base/components/cart-checkout/product-summary/index.js
```diff
@@ -0,0 +1,30 @@
...
+	if ( ! source ) {
+		return null;
+	}
+
+	const countType = getSetting( 'wordCountType', 'words' );
+
+	return (
+		<Summary source={ source } maxLength={ 15 } countType={ countType } />
+	);
+};
```
nerrad, line 23: maybe avoid variable declaration here and just invoke `getSetting` directly for the `countType` prop?

```suggestion
	return (
		<Summary source={ source } maxLength={ 15 } countType={ getSetting( 'wordCountType', 'words' ) } />
	);
};
```

assets/js/base/components/summary/index.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Summary component.
+ *
+ * @param {Object} props Component props.
+ * @param {string} props.source Source text.
+ * @param {number} props.maxLength Max length of the summary, using countType.
+ * @param {string} props.countType One of words, characters_excluding_spaces, or characters_including_spaces.
+ * @param {string} props.className Class name for rendered component.
+ */
```
nerrad, line 19: 👍 to documenting props here.

assets/js/base/components/summary/utils.js
```diff
@@ -0,0 +1,131 @@
...
+/**
+ * Remove HTML tags from a string.
+ *
+ * @param {string} htmlString String to remove tags from.
+ * @return {string} Plain text string.
+ */
+const removeTags = ( htmlString ) => {
+	const tagsRegExp = /<\/?[a-z][^>]*?>/gi;
+	return htmlString.replace( tagsRegExp, '' );
+};
```
nerrad, line 70: We should include some tests to verify this does what is expected (so have test strings in the existing test suite that have tags that should be removed). Would also be good to include strings with exploit attempts in them (i.e. `<script="alert('something');">` in the string to ensure they get removed.

mikejolley: To be 100% clear, this is not here for security. The strings coming from the API are currently 'trusted' and output raw. The tags are removed here so the tags themselves are not included in the character and word counts, or cut off and left unclosed.

nerrad: Makes sense. But are we sure that someone in the future won't grab this to strip tags from "untrusted" output? Obviously if they did without double-checking the code, that's a problem in itself, but still - either we assume that might happen, or we make it clear what the limits of the function are.

mikejolley: Maybe, or they might do what I did and steal some code from Gutenberg :p This was a regex I found in the wordcount package. I think if we added checks for script tags/security hardening, it might give the wrong impression that this is secure. They are likely other cases that would need to be considered if this was for security e.g. encoded tags. We don't need to take care of this here in this component.

PR #2391, opened 2020-05-04
Fix long product names regression

### LINKED ISSUE #2376: Long product names upset layout in checkout order summary
### To reproduce
- Edit the name of a product so it's super long with long words (think part numbers at a store that only sells truck/automotive parts) - e.g. `Long Semicircular Widget Flange, brown LBS_WF_5389u90ri3905i930ti609i509kmklm-b`
- Add that product to your cart, go to checkout (block of course)

The product name takes up more width than it should, doesn't wrap correctly, and the price is nudged out of the box. 

It looks like the `break-word` style is still applied, but the container width is too large and/or it's `flex-grow`ing bigger than we want.

Aljullu:
Fixes #2376.

Fixes a regression introduced in #2348, which was making long product names to overflow in the _Cart_ and _Checkout_ blocks.

### Screenshots
imatge

1. Edit a product so it has a name with a very long word and add it to your cart.
2. Go to the _Cart_ and _Checkout_ blocks and make sure it doesn't break the layout.

### DIFF
--- a/assets/js/base/components/cart-checkout/product-name/style.scss
+++ b/assets/js/base/components/cart-checkout/product-name/style.scss
@@ -2,5 +2,5 @@
 	@include font-size(16);
 	@include wrapBreakWord();
 	display: block;
-	width: max-content;
+	max-width: max-content;
 }
--- a/assets/js/blocks/cart-checkout/checkout/style.scss
+++ b/assets/js/blocks/cart-checkout/checkout/style.scss
@@ -117,6 +117,7 @@
 	.wc-block-order-summary-item__header {
 		display: flex;
 		flex-wrap: wrap;
+		justify-content: space-between;
 	}
 
 	.wc-block-product-name {


assets/js/base/components/cart-checkout/product-name/style.scss
```diff
@@ -2,5 +2,5 @@
 	@include font-size(16);
 	@include wrapBreakWord();
 	display: block;
-	width: max-content;
+	max-width: max-content;
```
senadir, line 5: not defining a width could also be an alternative

Aljullu: Yeah... for reference: that declaration makes the link clickable area smaller so there are no accidental clicks.

haszari:
That's a nice looking `BeanieBeanieBeanieBeanieBeanieBeanieBeanieBeanieBeanieBeanieBeanieBeanie with Logo`!

Issue #2394, opened 2020-05-04
[Single Product Block] Implement Inner Blocks Editing Interface


mikejolley:
Implement the Inner Blocks editing interface. This will also be the _preview_ state so only edit and layout views are needed in the editor.

See pbIy4N-ds-p2

Issue #2393, opened 2020-05-04
Single Product Atomic Inner Blocks


mikejolley:
The single product block requires additional atomic blocks and components to show product data and the full add to cart form experience.

garymurray:
> Product description/short description. The current mocks appear to show a short description. We need to decide how full descriptions are rendered. "additional information" is out of scope for i1, however, description is mentioned. So this needs clarification from @garymurray

The intent was that this would use what has been set as the `Product short description`.

> The designs are missing such as "in stock" text. Does this need revision?

Can you provide a list of what core currently shows there so we can decide how much of that we can/should show? I do believe for instance 'In stock` is something we would want to show.

mikejolley:
> Can you provide a list of what core currently shows there so we can decide how much of that we can/should show? I do believe for instance 'In stock` is something we would want to show.

Wasn't this explored by design team already? 

- Sale badge
- Product title
- Product price (for variable products this is a range)
- Stock text i.e. "In Stock" or "X left in stock"
- Cart qty/button
- SKU
- Terms list (categories and tags)
- Main image
- Image gallery
- Then the full description/attributes/reviews section
- Then related products and upsells

garymurray:
Thanks for those. @LevinMedia this should be able to inform any design work on the single block.

> Wasn't this explored by design team already?

TBH, I am not sure - I had assumed most of this had been taken into account for already in the designs that had been worked on - but now that we taking a closer look at the designs it seems lots might have just been overlooked.

LevinMedia:
@mikejolley @garymurray 

the scope from https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2379 looks correct, I have a few comments to add:

>Product description/short description. The current mocks appear to show a short description. We need to decide how full descriptions are rendered. "additional information" is out of scope for i1, however, description is mentioned.

This should use the short description - the full description is out of scope, and would be built into another block that would be part of a larger product page template, along with reviews etc)

>Core feature parity - The designs are missing such as "in stock" text. Does this need revision?

I think this was overlooked, as I never saw mention of it in the original posts. If it's not too much extra scope, I think it would be nice to have. Likewise, the products SKU, categories and tags would be nice to haves as well. I'm not sure why they weren't included in the original scope. 

If we're able to include all of the above, the only thing we'd be missing for full feature parity with core is gallery controls.

LevinMedia:
[...]

PR #2399, opened 2020-05-04
Single Product Block

### LINKED ISSUE #2392: [Single Product Block] Block Creation
Create and register the barebones block with the following title and description:

> **Single Product**
> Display a single product with its details and quantity.

- [x] Check ^ description with @garymurray. Doesn't seem great (quantity has no context).
- [x] Add new icon for block
- [x] Implement product picker using existing components (see featured product)
- [x] Sidebar settings (others from mockup will be part of atomic blocks)
  - [x] Product picker 
- [x] Create placeholder for product details/frontend/inner block view for now. This will be completed in a follow up.

### LINKED ISSUE #2535: Investigate Inner Block rendering without the use of LayoutConfig
LayoutConfig was created to render atomic blocks whilst passing a product object to each child.

With the inclusion of context, we may be able to use native gutenberg rendering instead and have Atomic blocks look at the context instead.

Using native methods of rendering Inner Blocks should make our implementation more future proof, and dropping LayoutConfig removes extra complexity and points in the code which can break.

mikejolley:
Creates the basic structure for the Single Product Block (behind the experimental flag).

Closes #2392
Closes #2535

- Creates the block, with icon and description
- Implements the product picker in edit mode and the inspector
- Moves the inner block layout functionality to shared atomic utils (used by all products and single product blocks)
  - utils also include block->component maps and anything else related to atomic components
- ~~Introduces a new hook named `useSyncedLayoutConfig ` which syncs inner block changes to block attributes as a LayoutConfig array. This is only being used by Single Products Block currently. The All Products block does not update instantly and so does not need this (yet)~~
- **Introduces a new method of rendering inner blocks without layout configs.** (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2542)
- Removes dedicated atomic/components dir so that the components can share attribute data with atomic/blocks when rendering on the frontend.
- Combines `useProductLayoutContext` and `useInnerBlockConfigurationContext` since both are used at the same time in all contexts. This allows the parent block to be passed, as well as a prefix for the CSS classnames.
- Introduces `ProductDataContext` to share real product data with atomic blocks.
   - For all products block, this is where _preview data_ is passed. Prior to this change, the atomic blocks themselves looked up preview data.
   - For the single product block, it passes real product data via this context.
   - Atomic blocks feed this data to the atomic components
- Updates atomic blocks to have proper loading states and consume `ProductDataContext`.
   - Prior to this some of the atomic blocks had loading states, but others did not. I've improved this so anytime _product_ is null, it's considered in loading state. This is useful if the product data is being pulled from the API.
- Shared contexts so Atomic blocks are functional (see https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2399#issuecomment-631560641) This ensures contexts work when wrapping atomic blocks in the editor and on frontend.

_Known follow ups:_
- Style duplication will be resolved in #2527. Right now there are duplicate styles across all products and single products which could be improved.
- The preview content will be added in a follow up when the design is finalised.
- The inner block editor will be further developed as part of #2394
- There is a basic atomic block structure in this PR as an example, but it won't be completed until #2393 is resolved.
- There is a basic implementation of columns support as a POC but this will be looked at some more in #2526

**Screenshots**

Edit interface:
Screenshot 2020-05-06 at 10 14 14
Screenshot 2020-05-06 at 10 15 26

Loading state:
placeholder

Inner Block state:
content

**How to test**

1. Confirm _All Products Block_ remains functional in editor/frontend
2. In development mode, confirm the block is built and shown in the editor
3. Insert the block and select a product. Ensure it persists on save.
4. Try toggling between edit and layout mode.
5. Confirm placeholder details are shown on frontend.


nerrad:
@mikejolley how would you feel about putting the bare minimum in place for this block and merging in behind a feature flag? That way we can avoid a long running development branch and keep pull requests relatively lean?

mikejolley:
@nerrad given the amount of focuses this cycle I want to keep PR reviews down; so one review/test of the block itself. Supporting changes (e.g. API) I will raise separately and rebase. 

There isn't much point in putting an incomplete block behind a flag imo. I think those have more value for beta testing?

nerrad:
> There isn't much point in putting an incomplete block behind a flag imo. I think those have more value for beta testing?

There's a few reasons why I prefer the feature flag approach:

- Merging is less of a problem. If you have a long running separate pull, with various branches off of it for in progress development work, not only do you have to potentially deal with conflicts between those branches but also potential conflicts coming from master into the main development branch. This juggling between branches increases the potential for bad merge artifacts or merging in the incorrect order.
- We have tooling created to help gauge impact of changes. In most cases, this tooling (particularly bundle size) only works at a single branch level off of master. We lose the insight of that tooling if we use a base branch for the feature (only seeing the impact of a pull when it's merged back into the feature branch). This is particularly problematic when these tools require on a clean history (via rebasing) which is harder to do with the feature branch approach.
- The flag isn't really for testing. It's created so we have a way to have in progress work in master without worry of releasing it prematurely to users either in the wp.org channel or in the woo core. It only impacts _release_ builds so our development happens without issues.
- Having multiple focuses actually is a reason why I'd want this merged sooner because otherwise there's less visibility into the potential issues new components or changes in the feature branch might have with other in progress work.
- Creating new components or improving existing ones can be done with less of a hassle because while you can still do that off master and then merge up into the feature branch - if you have other branches _off of the feature brach_ there's an extra step involved. 

About the only benefit I see for having a feature branch is to keep that work isolated from the rest of the code and _prevent prematurely exposing_ to users. I think the latter is solved nicely by the feature flag and with the former I don't think that's something we need to worry about (and actually I'd sooner know about impacts to other existing code _earlier_ in the development cycle rather than later after we merge the whole caboodle sometime down the road.

I'm definitely _all ears_ 👂 if you have some reasons for keeping things in a feature branch that I haven't considered?

mikejolley:
> Merging is less of a problem. If you have a long running separate pull, with various branches off of it for in progress development work

> We have tooling created to help gauge impact of changes. In most cases, this tooling (particularly bundle size) only works at a single branch level off of master

> Having multiple focuses actually is a reason why I'd want this merged sooner because otherwise there's less visibility into the potential issues new components or changes in the feature branch might have with other in progress work.

I'm not doing that. I'm using this branch to get the block working and to a testable, reviewable state. I'm happy to put it behind a flag at that point, but not before. I mentioned the focuses because I know everyone is busy, so I'd prefer to have testable, full featured PRs rather than smaller, but harder to test without context, PRs that potentially block further work.

> Creating new components or improving existing ones can be done with less of a hassle because while you can still do that off master and then merge up into the feature branch - if you have other branches off of the feature brach there's an extra step involved.
Regardless, any API changes or non-block specific components will be raised on master separately, and then rebased here once merged. I am not creating additional branches for review off of this branch.

I imagine I'll change this to a proper PR and request review when it satisfies https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2392 I mainly created a draft to track progress (this PR is marked draft/in-progress).

> The flag isn't really for testing. It's created so we have a way to have in progress work in master without worry of releasing it prematurely to users either in the wp.org channel or in the woo core.

I don't think we need to discuss this here and now, but we should really have 3 flags IMO. development, feature plugin, and core fwiw. If current flag usage is documented somewhere let me know so I can read up on it's proper purpose.



Already stated this was just tracking the creation of a block, not other components or touching other code.

Obviously creating this draft has caused some comfusion so I'm just going to close it and raise again when I'm ready.  I didn't intend to kick off a large workflow debate.

nerrad:
My apologies, I'm pretty sure I misunderstood the purpose of this pull then :) I thought you were intending on using this for a long running feature branch (and I also missed the important indicator that this was a _draft_ post) 🤦

nerrad:
The latest commit I pushed (202e5a2) fixes an issue where the context was not being seen by atomic components within the atomic blocks for the single product block. Essentially the problem was that the single product block was using an instance of `InnerBlockConfigurationProvider` that was _different_ than the instance built in the `all-products` bundle (which is where the atomic blocks were registered). Thus the components being rendered were from the atomic blocks in the `all-products` bundle and their usage of `useInnerBlockConfigurationContext` was assuming the same instance of the context defined in the `all-products` bundle.

The fix, is to push the `InnerBlockConfigurationProvider` and hook into a shared library. That way we're ensuring both blocks are using the correct instance.

Going forward, we should probably take some time to figure out how we want to approach this because there are implications either way.

## Build a bundle for shared context.

Anytime we share context across bundles that implement the context wrapping _atomic blocks_, we should put that context in the new shared context build I did in the commit. 

Pros:

- there's a clear export for shared context.

Cons:

- we really have to watch what we put in this shared context build because there's risk of circular dependencies if it needs things like hooks or components (basically, everything in this bundle _must_ be self-contained).
- This bundle could grow over time, and it would basically mean anything using this bundle will get all contexts loaded via the external (instead of just what is needed in the build). This mostly has impact on the frontend. We could get around that by just having frontend builds import directly from the context folder since the frontend doesn't register blocks.

## Other options?

I thought, maybe we could have atomic blocks built to their own bundle. If we went this way, it still won't work because we'd run into the same problem we have now with the context defined in the atomic blocks bundle being different than that used by the context provider defined in the various parent block bundles.

The only way around this would be to do some fancy build alias swapping so editor builds pull the context from the atomic blocks external and frontend builds pull from the context from the main file.

## What should we do?

Since any approach I can think of results in differing builds between frontend and backend, I think we should just go with the approach I implemented here for now _but_ only implement usage of the external for editor builds. On the frontend we can alias `@woocommerce/shared-context` to direct imports.

Eventually, we're going to have to revisit this when/if we implement InnerBlocks in cart and checkout. That will be _tricky_ because all the hooks used in the cart and checkout contexts create some circular dependencies when the contexts are built to a separate build.

mikejolley:
> I'm not allowed to add inner blocks specific to the product (in the screenshot, I was trying to add back the Product Summary block).

This is due to using columns blocks which have their own Inner Block configs. We'll need to look at this as part of https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2526, either fixing columns or adding a custom column block.

mikejolley:
> Product Image: The Sale badge doesn't seem to work for me:

There is a bug here, but also it could be when a product is not on sale. This seems to be something we'll need to work out of the design side since we're working with live data cc @LevinMedia

assets/js/atomic/blocks/product/button/block.js
```diff
@@ -0,0 +1,150 @@
+/**
+ * External dependencies
```
mikejolley, line 2: Moved contents of Atomic Components to a block.js per atomic block. This is so we can reuse the same block for edit.js and frontend.js, and also allows frontend.js to see the defined blockAttributes so it can map dataset to block attributes correctly.

assets/js/atomic/blocks/product/image/attributes.js
```diff
@@ -0,0 +1,19 @@
+/**
+ * Internal dependencies
+ */
+export const blockAttributes = {
```
mikejolley, line 4: Defining these separately so the frontend.js can validate attributes against them.

assets/js/atomic/blocks/product/save.js
```diff
@@ -0,0 +1,12 @@
+/**
+ * External dependencies
+ */
+import classnames from 'classnames';
+
+const Save = ( { attributes } ) => {
+	return (
+		<div className={ classnames( 'is-loading', attributes.className ) } />
```
mikejolley, line 8: This is new; saves some markup for the InnerBlock so we can render from this on the frontend instead of using layoutconfig.

assets/js/atomic/utils/create-blocks-from-template.js
```diff
@@ -0,0 +1,18 @@
...
+ * External dependencies
+ */
+import { createBlock } from '@wordpress/blocks';
+
+/**
+ * Creates blocks for a given inner blocks Template.
+ *
+ * @param {Array} template Inner Blocks Template.
+ */
+export const createBlocksFromTemplate = ( template ) => {
```
mikejolley, line 11: This utility will create blocks recursively from a defined default template. Used currently to reset the layout back to default.

assets/js/atomic/utils/get-block-map.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * External dependencies
+ */
+import { getRegisteredInnerBlocks } from '@woocommerce/blocks-registry';
+
+/**
+ * Internal dependencies
+ */
+import ProductButton from '../blocks/product/button/frontend';
```
mikejolley, line 9: Rather than using atomic/components, this is mapping to the frontend.js file of each defined Atomic Block. The frontend.js file is responsible for converting dataset attributes to block attributes, then rendering the block's component.

assets/js/atomic/utils/get-layout-config.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * Converts inner blocks to a list of layout configs.
+ *
+ * @param {Object[]} innerBlocks Inner block components.
+ */
+export const getLayoutConfig = ( innerBlocks ) => {
```
mikejolley, line 6: Moved from base-utils. This might be removable in a follow up when layout configs are deprecated/removed.

assets/js/atomic/utils/render-product-layout.js
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Maps a layout config into atomic components.
+ *
+ * @param {string} blockName Name of the parent block. Used to get extension children.
+ * @param {Object} product Product object to pass to atomic components.
+ * @param {Object[]} layoutConfig Object with component data.
+ * @param {number} componentId Parent component ID needed for key generation.
+ */
+export const renderProductLayout = (
```
mikejolley, line 14: Moved from product-list-item/utils.js

assets/js/atomic/utils/render-product-layout.js
```diff
@@ -0,0 +1,58 @@
...
+				return null;
+			}
+
+			const productID = product.id || 0;
+			const keyParts = [ 'layout', name, index, componentId, productID ];
+
+			return (
+				<LayoutComponent
+					key={ keyParts.join( '_' ) }
+					attributes={ attributes }
```
mikejolley, line 51: This was changed to maintain compatibility with the atomic blocks - before we just passed all props.

assets/js/base/utils/render-frontend.js
```diff
@@ -0,0 +1,148 @@
+/**
+ * External dependencies
```
mikejolley, line 2: These utils were moved form assets/js/utils since they are shared and used throughout.

assets/js/base/utils/render-frontend.js
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * Render Inner Block Components from markup.
+ *
+ * @param {Object}   props           Render props.
+ * @param {Array} props.children  Child nodes.
+ * @param {string}   props.keyPrefix Prefix for keys.
+ * @param {Object}   props.blockMap  Child blocks will be mapped to components.
+ */
+const renderChildren = ( { children, blockMap, keyPrefix = '' } ) => {
```
mikejolley, line 56: This is where the new magic happens; this converts inner blocks into usable components.

assets/js/blocks/active-filters/frontend.js
```diff
@@ -19,8 +19,8 @@ const getProps = ( el ) => {
 	};
 };
 
-renderFrontend(
-	'.wp-block-woocommerce-active-filters',
-	withRestApiHydration( Block ),
-	getProps
-);
+renderFrontend( {
```
mikejolley, line 22: Passing an object instead of individual arguments since the renderFrontend method keeps growing.

assets/js/blocks/single-product/constants.js
```diff
@@ -0,0 +1,49 @@
...
+					[ 'woocommerce/product-summary' ],
+					[ 'woocommerce/product-button' ],
+				],
+			],
+		],
+	],
+];
+
+export const ALLOWED_INNER_BLOCKS = [
+	'core/columns',
```
mikejolley, line 46: This might be temporary but is included for now; we need to find a way to limit blocks within columns in a follow up. An issue exists.

assets/js/blocks/single-product/style.scss
```diff
@@ -0,0 +1,227 @@
+// Waiting for render.
```
mikejolley, line 1: There is a followup to combine product styles. This CSS is temporary.

mikejolley:
The other reported issues should be resolved.

LevinMedia:
>There is a bug here, but also it could be when a product is not on sale. This seems to be something we'll need to work out of the design side since we're working with live data 

@mikejolley I'll be looking into innerblock settings this week, and I'll be sure to give that one some thought.

PR #2401, opened 2020-05-05
clarify various checkout block editor sidebar wording

### LINKED ISSUE #2341: Checkout footer / link options heading & help text could be improved
The heading and the help text are a bit abstract – would be great if we could make this clearer or highlight why these options are useful.

Since all this stuff is in the footer/bottom perhaps it's `Footer options`, or since it's all "link" related perhaps could use `Links`. Or maybe this is related to `No distraction checkout`?

Similar with `Form options` - these are all extra address fields, so maybe we could narrow this down as `Address options`.

I realise it's tricky getting this text right and a lot of thought has gone into these. If we can make these options even clearer it'd be great IMO :)

FYI @garymurray @LevinMedia  - you're the experts here

haszari:
Fixes #2341

This PR makes a few changes to Checkout block sidebar wording based on #2341. I've aimed for simplicity and clarity. The goal is to make it super clear what the options do, and keep the wording to a minimum. A related goal is to use categories (panels) in the sidebar to provide a more context, and to keep the number of sidebar panels to a minimum.

### Screenshots
New wording:
Screen Shot 2020-05-05 at 4 21 59 PM

Previous wording:
Screen Shot 2020-05-05 at 4 23 17 PM

What's changed?

- Block description: `Display a checkout form so your customers can submit orders.`  ~~`Display the checkout experience for customers.`~~
- Address options panel: 
  - `Address options` ~~`Form options`~~
  -  `Include additional address fields in the checkout form.` ~~`Choose whether your checkout form requires extra information from customers`~~
- Navigation / links panel: 
  - `Navigation options` ~~`Content`~~ 
  - Removed the redundant description ~~`Choose additional content to display.`~~
  - Streamlined policy link description `Shows links to your "terms and conditions" and "privacy policy" pages.` ~~`Shows a list of links to your "terms and conditions" and "privacy policy" pages.`~~

Also, I think `Return to cart` (the editor UI and the link itself) should probably be sentence case? (Currently title case.)

1. Go to block editor, add or edit a Checkout block.
2. Read through the sidebar options – is it clear what they do and why they might be useful for particular store use cases?

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/edit.js
+++ b/assets/js/blocks/cart-checkout/checkout/edit.js
@@ -229,7 +229,7 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 						}
 						labels={ {
 							title: __(
-								'Return to cart button',
+								'Return to Cart button',
 								'woo-gutenberg-products-block'
 							),
 							default: __(
--- a/assets/js/blocks/cart-checkout/checkout/edit.js
+++ b/assets/js/blocks/cart-checkout/checkout/edit.js
@@ -76,11 +76,14 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 				</Notice>
 			) }
 			<PanelBody
-				title={ __( 'Form options', 'woo-gutenberg-products-block' ) }
+				title={ __(
+					'Address options',
+					'woo-gutenberg-products-block'
+				) }
 			>
 				<p className="wc-block-checkout__controls-text">
 					{ __(
-						'Choose whether your checkout form requires extra information from customers.',
+						'Include additional address fields in the checkout form.',
 						'woo-gutenberg-products-block'
 					) }
 				</p>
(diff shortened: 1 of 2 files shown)


haszari:
@LevinMedia @garymurray I've started us off with some wording tweaks – keen for your feedback on my new wording. And also any tweaks/suggestions that come up when you reviewing the editor / sidebar wording for cart & checkout as a whole.

assets/js/blocks/cart-checkout/checkout/edit.js
```diff
@@ -146,21 +149,18 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
...
-					) }
-				</p>
 				<ToggleControl
 					label={ __(
 						'Show links to policies',
 						'woo-gutenberg-products-block'
 					) }
 					help={ __(
-						'Shows a list of links to your "terms and conditions" and "privacy policy" pages.',
+						'Shows links to your "terms and conditions" and "privacy policy" pages.',
```
Aljullu, line 163: Should the links be capitalized?
```suggestion
						'Shows links to your "Terms and conditions" and "Privacy policy" pages.',
```

garymurray:
Changes look good to me.

Just not sold on the use of the term `Navigation options` but can't think of anything better right now.

assets/js/blocks/cart-checkout/checkout/edit.js
haszari: Maybe ... previously they weren't, and personally I try to avoid capitalisation if not useful (since these are in quotes too), so I'll skip this for now. Could be added later of course.

haszari:
> Just not sold on the use of the term Navigation options but can't think of anything better right now.

Agreed, let's try and come up with something even better as we get these wrapped up. 

I'll merge this now since it's an improvement, and we can tweak / tidy more in future PRs, perhaps after a wider design review.

PR #2417, opened 2020-05-06
Update Stripe/payment payment notices

### LINKED ISSUE #2216: Checkout Block: Fix potentially unnecessary payment area notice when inline error already present for stripe cc.
This was reported by @mikejolley in #2189:

> When you leave card field blank, you get both a notice and inline errors on the form. Is this desirable?

My response there was that the inline notice should appear before submitting the form, if it doesn't we should probably try to hook things up so that the form is not submitted if any of the payment fields for the active payment method are empty.

### LINKED ISSUE #2365: Checkout error notices should not be dismissable
Raise by @mikejolley in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2352#pullrequestreview-402749074.

**To Reproduce**

1. Go to the _Checkout_ page fill the form but leave the credit card details empty.
2. Click on _Place order_.
3. Notice the error that appear has a close button.

**Expected behavior**
It should not be possible to close the notice.

**Screenshots**
imatge

mikejolley:
This PR updates the behaviour of the error notices in the Stripe integration.

- Validation notices (which display inline) will no longer also be shown in the payment-area notices section. Instead, it returns an empty string. I've updated the `PaymentMethodDataProvider` to only show messages when there is content. Closes #2216
- When notices are added to the payment section they are no longer dismissable. Notices will be removed when re-attempting payment, or switching payment methods via the tabs. Closes #2365 
- I noticed a bug when focussing on CC fields where the placeholder was not hidden. This was solved by tracking `isEmpty` for each credit card field, rather than a single state value for all 3 fields.

1. Smoke test Stripe; cause validation issue and see what notices are shown
2. Ensure you can complete payment with a valid card

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-checkout-subscriptions.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-checkout-subscriptions.js
@@ -62,7 +62,9 @@ export const useCheckoutSubscriptions = (
 			const type = event.error.type;
 			const code = event.error.code || '';
 			let message = getErrorMessageForTypeAndCode( type, code );
-			message = message || event.error.message;
+			if ( message === null ) {
+				message = event.error.message;
+			}
 			setError( error );
 			return message;
 		};
--- a/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-checkout-subscriptions.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-checkout-subscriptions.js
@@ -61,10 +61,9 @@ export const useCheckoutSubscriptions = (
 		onStripeError.current = ( event ) => {
 			const type = event.error.type;
 			const code = event.error.code || '';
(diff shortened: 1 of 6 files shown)


assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-checkout-subscriptions.js
```diff
@@ -62,7 +62,9 @@ export const useCheckoutSubscriptions = (
 			const type = event.error.type;
 			const code = event.error.code || '';
 			let message = getErrorMessageForTypeAndCode( type, code );
-			message = message || event.error.message;
+			if ( message === null ) {
+				message = event.error.message;
+			}
```
nerrad: This could probably be simplified by using the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator). I've implemented in 89e40e0 so this pull can be merged.

PR #2420, opened 2020-05-06
Add an inline skeleton sizing script

### LINKED ISSUE #2402: Cart and Checkout: apply responsive class names to skeletons
In #2370 we updated the _Cart_ and _Checkout_ blocks so they use container queries instead of responsive queries. One issue we identified was that skeletons would not react to those breakpoints, since that logic is loaded with React.

This issue is about writing a vanilla JS code snippet that executes immediately and adds the `is-large`, `is-medium`, etc. classes to the skeleton wrapper based on their width. It wouldn't need to subscribe to viewport resizes because the skeleton will only be visible for a short period of time.

mikejolley:
Similar to our script to `renderFrontend`, this PR adds an inline script to add a classname to the skeleton as soon as possible, based on container size, as soon as the skeleton is rendered.

Fixes #2402

1. Resize window so you are seeing medium/cart sized cart or checkout.
2. Inspect source.
3. Check the skeleton markup has the correct class, e.g. `is-small`.

I tested this by console logging also.

### DIFF
--- a/src/BlockTypes/AbstractBlock.php
+++ b/src/BlockTypes/AbstractBlock.php
@@ -155,8 +155,8 @@ abstract class AbstractBlock {
 
 			if ( containers.length ) {
 				Array.prototype.forEach.call( containers, function( el, i ) {
-					var w = el.offsetWidth;
-					var classname = '';
+					const w = el.offsetWidth;
+					let classname = '';
 
 					if ( w > 700 )
 						classname = 'is-large';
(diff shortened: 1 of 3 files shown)


src/BlockTypes/AbstractBlock.php
```diff
@@ -143,4 +143,37 @@ protected function enqueue_data( array $attributes = [] ) {
...
 	}
+
+	/**
+	 * Script to append the correct sizing class to a block skeleton.
+	 *
+	 * @return string
+	 */
+	protected function get_skeleton_inline_script() {
+		return "<script>
+			const containers = document.querySelectorAll( 'div.wc-block-skeleton' );
```
Aljullu, line 154: Nit: `containers` is a const while `w` and `classname`, defined some lines below, are vars. :point_down: 

I think it would be nice to keep it consistent and either always use var or always const/let. I don't have any preference, it looks like [const](https://caniuse.com/#feat=const) and [let](https://caniuse.com/#feat=let) support in all browsers including IE11 is good enough.

Issue #2350, opened 2020-04-29
Checkout - Payment fails for saved Stripe card (or other saved payment methods)


vbelolapotkov:
**Describe the bug**
Payment fails when trying to pay with saved stripe card.
The same card can be used as new card.

**To Reproduce**

1. Enable Stripe gateway.
2. Log in and go to my account.
3. Add at least one payment method. E.g. 4242 4242 4242 4242 (any date and cvc). 
4. Add product to the card and go to checkout page.
5. Use saved payment method to place the order.
6. Error is shown: `Something went wrong. Please contact us to get assistance.`

**Expected behavior**
Order is placed successfully and used payment method is chardged.

**Screenshots**
![image](https://user-images.githubusercontent.com/3139099/80584380-41721600-8a1a-11ea-9ddc-f0415c0f5c82.png)
![image](https://user-images.githubusercontent.com/3139099/80584410-4cc54180-8a1a-11ea-98e9-67fe992cd952.png)

**WordPress (please complete the following information):**

* Core version: 5.3.2 
* Gutenberg version (if installed): n/a
* WooCommerce version: 4.0.1
* WooCommerce Blocks version: 2.6.0-beta.1
* Site language: en_US
* Any other plugins installed: 

WooCommerce PayPal Checkout Gateway |   | by WooCommerce – 1.6.20 – 1.6.21 is available – Not tested with the active version of WooCommerce
-- | -- | --
WooCommerce Stripe Gateway |   | by WooCommerce – 4.3.3
WooCommerce Blocks |   | by Automattic – 2.6.0-beta.1
WooCommerce Subscriptions |   | by WooCommerce – 3.0.3
WooCommerce |   | by Automattic – 4.0.1
WordPress Importer |   | by wordpressdotorg – 0.7

**Desktop (please complete the following information):**

* OS: OS X 10.15.4
* Browser Safari
* Version  13.1

nerrad:
Hmm, I wonder if we forgot to wire up saved payment method processing...

haszari:
Spotted another potential issue when testing this. When I save the payment method from checkout, it didn't show up as an option on next checkout.

This is what I did:

- Add a saved card via my account (same steps as in issue description above).
- Complete a purchase, with a different card number, and checked `Save payment information to my account for future purchases.` checkbox.
- Add something to new cart, view checkout - I expected to see both my saved cards, or the most recently added one.

[...]

Issue #2424, opened 2020-05-06
Fix multiple hits to the stripe source endpoint.


nerrad:
While testing #2417 I noticed that there quite a bit of Stripe source requests happening when submitting a payment for Stripe CC. I counted 22 in master and 17 in the branch for #2417 (but I think the number fluctuates).

<img width="290" alt="Image 2020-05-06 at 2 08 08 PM" src="https://user-images.githubusercontent.com/1429108/81214055-44827e80-8fa5-11ea-8929-1380c674307a.png">

While the request itself happens really quickly, the concern I have here is that Stripe rate-limiting could start kicking in for sites if this is not reined in.

I think this is related a bit to some things @mikejolley brought to my attention in slack about `useCheckoutSubscriptions` executing some effects more frequently than we'd like.

I've created this issue to investigate and resolve.

Issue #2411, opened 2020-05-06
Checkout: "Save payment information to my account" doesn't save the payment method


haszari:
In checkout there's an option to save the card/payment method to your account, so you can check out faster & easier next time.

![Screen Shot 2020-05-06 at 1.50.37 PM.png](https://images.zenhubusercontent.com/5ddb3cb39529f10001e618e5/72139072-bae0-44ee-9a91-a83f2f276346)

This isn't hooked up yet - we need to get this working :)

See also related issue #2350 (previously saved payment methods don't work for payment)

PR #2425, opened 2020-05-06
Fix saved payment method selection not working for purchases.

nerrad:
Fixes: #2350

Originally I was just going to investigate and report findings to @haszari for #2350 so he could finish it off, but once I figured out the problem, the fixes were fairly easy and I figured just adding it and pushing for review would save time.

This pull fixes the issue where a selected saved payment method token would not work for completing the purchase. The following is done in this pull:

- Fixed incorrect arguments used for changing the payment status to success for saved payment method selection (I missed this todo with some refactoring that was done).
- Fixed the payment status getting set to `PRISTINE` too liberally. We only want that to happen if the payment status is not already successful. Otherwise the payment method token that is in the state would get wiped.
- Added saved payment token handling to the server (this is agnostic to payment method so _any_ saved payment method processing should "just work" - but I didn't test non supported payment methods for now).

## To test

- You have to setup a saved payment method via the shortcode cart for now because of #2411 
- Try making a purchase with that saved payment method, it should complete without issue.
- Try making a purchase with Stripe CC, it should complete without issue.
- Try making a purchase with Cheque, it should complete without issue.
- Try making a purchase with Chrome Pay or Apple Pay, either should complete without issue.
- Try breaking things (example, in checkout switch between saved payment method and selecting a payment method and then back again).
- Try making purchases in incognito mode. Should complete without issue.

I tested all of the above except the last one because I hit stripe rate limiting (likely because of #2424).

### DIFF
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -106,9 +106,7 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 	const updateToken = useCallback(
 		( token ) => {
 			if ( token !== '0' ) {
-				// @todo this will need updated when the signature changes to
-				// allow no billing data included.
-				setPaymentStatus().success( null, {
+				setPaymentStatus().success( {
 					payment_method: activePaymentMethod,
 					savedPaymentMethodToken: token,
 				} );
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -139,7 +139,12 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 				name: `wc-saved-payment-method-token-new`,
 			} );
 		}
-	}, [ customerPaymentMethods, selectedToken ] );
+	}, [
+		customerPaymentMethods,
+		selectedToken,
+		setActivePaymentMethod,
+		setPaymentStatus,
+	] );
 	const updateToken = useCallback(
 		( token ) => {
 			if ( token === '0' ) {
@@ -148,7 +153,7 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 			setSelectedToken( token );
 			onSelect( token );
 		},
-		[ setSelectedToken, setPaymentStatus, onSelect, activePaymentMethod ]
+		[ setSelectedToken, setPaymentStatus, onSelect ]
 	);
 	useEffect( () => {
 		if ( selectedToken && currentOptions.current.length > 0 ) {
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -105,23 +105,26 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 				if ( paymentMethods.length > 0 ) {
 					options = options.concat(
 						paymentMethods.map( ( paymentMethod ) => {
+							const option =
+								type === 'cc' || type === 'echeck'
+									? getCcOrEcheckPaymentMethodOption(
+											paymentMethod,
+											setActivePaymentMethod,
+											setPaymentStatus
+									  )
+									: getDefaultPaymentMethodOptions(
+											paymentMethod,
+											setActivePaymentMethod,
+											setPaymentStatus
+									  );
 							if (
 								paymentMethod.is_default &&
 								selectedToken === ''
 							) {
 								setSelectedToken( paymentMethod.tokenId + '' );
+								option.onChange( paymentMethod.tokenId );
 							}
-							return type === 'cc' || type === 'echeck'
-								? getCcOrEcheckPaymentMethodOption(
-										paymentMethod,
-										setActivePaymentMethod,
-										setPaymentStatus
-								  )
-								: getDefaultPaymentMethodOptions(
-										paymentMethod,
-										setActivePaymentMethod,
-										setPaymentStatus
-								  );
+							return option;
 						} )
 					);
 				}
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -57,6 +57,7 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 	const {
 		setPaymentStatus,
 		customerPaymentMethods,
+		setActivePaymentMethod,
 		activePaymentMethod,
 	} = usePaymentMethodDataContext();
 	const [ selectedToken, setSelectedToken ] = useState( '' );
@@ -106,6 +107,7 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 	const updateToken = useCallback(
 		( token ) => {
 			if ( token !== '0' ) {
+				setActivePaymentMethod( activePaymentMethod );
 				setPaymentStatus().success( {
 					payment_method: activePaymentMethod,
 					savedPaymentMethodToken: token,
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -9,16 +9,24 @@ import {
 } from '@woocommerce/base-context';
 import RadioControl from '@woocommerce/base-components/radio-control';
 
-/** @typedef {import('@woocommerce/type-defs/contexts').CustomerPaymentMethod} CustomerPaymentMethod */
+/**
+ * @typedef {import('@woocommerce/type-defs/contexts').CustomerPaymentMethod} CustomerPaymentMethod
+ * @typedef {import('@woocommerce/type-defs/contexts').PaymentStatusDispatch} PaymentStatusDispatch
+ */
 
 /**
  * Returns the option object for a cc or echeck saved payment method token.
  *
  * @param {CustomerPaymentMethod} savedPaymentMethod
- *
+ * @param {function(string):void} setActivePaymentMethod
+ * @param {PaymentStatusDispatch} setPaymentStatus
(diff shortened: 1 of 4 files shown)


src/Payments/Api.php
```diff
@@ -135,6 +135,15 @@ public function process_legacy_payment( PaymentContext $context, PaymentResult &
 			return;
 		}
 
+		// if we have savedPaymentMethodToken in the request, then convert into
+		// the format expected by the post.
+		// phpcs:ignore WordPress.Security.NonceVerification
```
haszari, line 140: Would be good to have  a comment explaining the `phpcs:ignore`. I'm guessing this is because we already trust everything in `POST` because the nonce is checked earlier in the API request handling?

src/Payments/Api.php
```diff
@@ -135,6 +135,15 @@ public function process_legacy_payment( PaymentContext $context, PaymentResult &
...
+		if ( isset( $_POST['savedpaymentmethodtoken'] ) ) {
+			$token_key = 'wc-' . $context->payment_method . '-payment-token';
+			// phpcs:ignore WordPress.Security.NonceVerification, WordPress.Security.ValidatedSanitizedInput.MissingUnslash, WordPress . Security . ValidatedSanitizedInput . InputNotSanitized
+			$_POST[ $token_key ] = wc_clean( $_POST['savedpaymentmethodtoken'] );
```
haszari, line 144: Does this have to happen on the server – I'm wondering if we can convert it to `wc-stripe-payment-token` on the client. That way the API is consistent, there's not two different POST params related to this. i.e. tighten it up

assets/js/base/components/payment-methods/saved-payment-method-options.js
```diff
@@ -108,9 +108,7 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 	const updateToken = useCallback(
 		( token ) => {
 			if ( token !== '0' ) {
-				// @todo this will need updated when the signature changes to
-				// allow no billing data included.
-				setPaymentStatus().success( null, {
+				setPaymentStatus().success( {
```
haszari, line 111: From my fiddling yesterday I think we might need to also `setActivePaymentMethod()` here. That should fix the issue with selecting `Cheque payments` tab then selecting a saved stripe card.

haszari: Also wondered about switching this argument list to an object, named-parameters style, to protect against "argument in wrong position" regressions. Not sure if this is risky / better but maybe worth considering.

e.g. something like this

```js
setPaymentStatus().success( {
  paymentMethodData: {
     payment_method: activePaymentMethod,
     savedPaymentMethodToken: token,
  }
} );
```

haszari:
I think I've fixed those issues with bfe10d537fb2a478a9876c8bafec35f67af8a15c

- switching from cheque to saved card uses cheque - now fixed, by ensuring we `setActivePaymentMethod()`
- if multiple saved cards, always uses first card - I can't reproduce this anymore, so either I was mistaken (🤪) or `setActivePaymentMethod()` fixes that too 🙂

Tagging in some fellow @woocommerce/merch-squad for review / test.

haszari:
@nerrad Note my review still has some code comments that might be worth addressing, but if this reviews/tests well I think it's mergeable.

haszari:
Note I haven't tested with Chrome & Apple Pay.

nerrad:
> Side note - the term payment method

I use the term payment method specifically for the "method used by the shopper to make a payment". Most _shopper_ facing functionality is referenced as "payment method" (or some variation). 

The term payment _gateway_ refers to the service processing the payments. 

In some cases, one gateway might service _multiple_ payment methods, or it might just service one. Let's look at Stripe for instance. The _gateway_ is `stripe` (and that's the slug used to represent it). It services multiple payment methods (`cc`, `paymentRequest`...and potentially in the future _sepa_ etc). In other cases the gateway may have just one payment method, an example here would be cheque.

I'm _definitely_ all for consistency and open to addressing it with blocks, but another challenge is that there exists equally inconsistent term usage across WooCommerce core and extension.

So as an action will you help with the following Rua?

- create an issue that we'll use to settle on the terms we'll use and how they're used (i.e. what they represent).
- we can use that as a basis to audit and fix any code in the blocks repo that is inconsistent.
- if this ends up involving code churn in various places (especially potentially things like `PaymentMethodRegistry` that are public apis then we'll want to fix before release. But if it ends up just being a few minor internal variable naming we can probably punt until after release.

assets/js/base/components/payment-methods/saved-payment-method-options.js
nerrad: > From my fiddling yesterday I think we might need to also setActivePaymentMethod() here.

The reason this worked is due to happy circumstance with an existing bug (`activePaymentMethod` should be a dependency). It's actually the incorrect fix here because activePaymentMethod will be whatever is selected in the tab, instead we need to update the activePaymentMethod via the selected payment option. I've got a fix underway for that.

> Also wondered about switching this argument list to an object, named-parameters style, to protect against "argument in wrong position" regressions. Not sure if this is risky / better but maybe worth considering.

What places in the code are dependent on the order of the properties? The argument **is** [`paymentMethodData`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/81b4633ae78d8903cb866f9579e95590337bbcab/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js#L252-L256).

src/Payments/Api.php
nerrad: ya good catch, definitely could do it on the client and it would remove the need for this code 👍

nerrad:
Alright @haszari this is ready for another review and test. Note the changes in the latest commits (and their notes).

assets/js/base/components/payment-methods/saved-payment-method-options.js
haszari: > What places in the code are dependent on the order of the properties? The argument is paymentMethodData.

The original issue looked like an arg-order issue to me - we were passing `paymentMethodData` as arg 2, but it was arg 1 (fix was to remove the `null` placeholder for arg 1). If we made the func take an object, i.e. named parameters, we could reduce risk of that kind of bug in future. No biggie, not a blocker.

assets/js/base/components/payment-methods/saved-payment-method-options.js
```diff
@@ -31,24 +39,46 @@ const getCcOrEcheckPaymentMethodOption = ( { method, expires, tokenId } ) => {
...
+	setPaymentStatus
+) => {
 	return {
 		value: tokenId,
 		label: sprintf(
 			__( 'Saved token for %s', 'woo-gutenberg-products-block' ),
 			method.gateway
 		),
 		name: `wc-saved-payment-method-token-${ tokenId }`,
+		onChange: ( token ) => {
```
haszari, line 74: Feels like this little chunk is a reusable func we could factor out - probably fine as is though. e.g. something like 

```js
function applySavedPaymentMethod( gatewaySlug, token ) { … }

// and then each radio option could use it like
onChange: () => applySavedPaymentMethod( method.gateway, token )
```

Not sure if that's better – just mentioning as it took me a moment to understand.

nerrad: I was wondering if you'd mention that :). It's a good spot, but I usually follow the ["rule of three"](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)) before I refactor chunks out.

nerrad:
> If there's a saved method, it's selected by default, but it's not persisted in the store as "payment success with token x" because that happens onChange. Then the submit order request fails because payment_data is [].
> 
> I'm guessing we need to get that logic to fire when setting up the initial / default payment method.

OOO good catch! Ya the code definitely needs to handle that! I thought I had tested this but I clearly didn't.

assets/js/base/components/payment-methods/saved-payment-method-options.js
nerrad: Oooohhh, sorry I misunderstood, I thought you were referring to something with the _property_ order (and missed the important usage of the word _argument_ in your comment). Hmm, ya I can see your point, but beyond that, there isn't much compelling reason to me for the resulting code churn at the moment to do that.

haszari:
> payment method 🦺

Follow up issue here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2442

As always, feel free to tweak / reword :)

haszari:
@nerrad I pushed a fix to ensure the default saved payment option is used if shopper doesn't do anything - 87c6d82.

With this fix I tested a few things and I think this is good to go - as before, I think we should get an independent review & test from someone else before merge. Tagging @mikejolley  :)

Here's what I tested:

- Default selected saved payment option works correctly - i.e. when user doesn't touch payment options :)
- Saved payment options preview works correctly in editor - shows a fake saved card, and "new cc" option is selected so card form is visible in preview.
- Correct saved card is used when there are two (or more), and shopper selects via radio button.

Note I hit rate limits so didn't get to test:

- Using a new CC to pay.
- With 2 or more saved payment options, set the second one to the default in my account, ensure that card is used for payment.

While I was looking at it I had a few ideas for making that code clearer, but probably dangerous to pursue right now. Noting here for transparency.

- The two funcs used for building an radiobutton "option" object differ only in how the label is rendered; might be clearer to have two separate funcs for generating the labels, and move the rest of the code inline to the useEffect / paymentMethods.map.
- Feels a little weird triggering the `onChange` manually to trigger the default, especially next to `setSelectedToken`. Curious if there's a way we can have a single bottleneck/callback for "apply this payment method", so `onChange` and applying the initial default call the same func.

nerrad:
The fix you added @haszari looked to be the best available for now. 

I went through all the tests again after rebasing against master and adding a commit that also fixes the useEffect dependencies and things are looking good for me.

nerrad:
> (minor) The only issue I noticed is styling related - this wraps unnecessary due to a flex basis rule

There's no styling changes in this pull so that pre-exists in master? Should we create an issue?

Issue #1916, opened 2020-03-09
The Rating section HTML inconsistency in the products grid


dashkevych:
Here is how the rating html looks in the Best Selling Products, On Sale Products, Hand-picked Products, Top Rated Products and Newest Products blocks:

![HTML markup](https://i.snipboard.io/60vm45.jpg)

Here is how the rating html looks in the All Products block:

![HTML markup](https://i.snipboard.io/JE67od.jpg)

--

WordPress: 5.4-RC1-47424
WooCommerce: 3.9.3

nerrad:
@Aljullu or @mikejolley what are your thoughts on this one?  

It appears the `ProductRating` component isn't following the same structure as what is exposed by the server side rendered product grid blocks.  Should we align the two?

Aljullu:
> Should we align the two?

Absolutely, we should also align it with the _Reviews_ blocks:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/components/review-list-item/index.js#L126-L149

gigitux:
After the investigation, the HTML of the first screenshot is generated by [WC Core](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f935e4190c3313b0ec25ba01dd1980af6fdb8932/src/BlockTypes/AbstractProductGrid.php#L510), and the other one is generated by [WC Blocks ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f935e4190c3313b0ec25ba01dd1980af6fdb8932/assets/js/atomic/blocks/product-elements/rating/block.js#L44)

Issue #1967, opened 2020-03-13
Merge the markup from atomic/blocks/product-elements/ with base/components/cart-checkout/


Aljullu:
Initially discussed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1959#discussion_r391842881.

We have very similar components in `atomic/blocks/product-elements/` (used for the _All Products_ and _Single Product_ blocks) and `base/components/cart-checkout/` (used for the _Cart_ and _Checkout_ blocks). This issue is about moving the markup from Atomic blocks to independent components as much as possible. Probably that would mean moving some components out of the `cart-checkout` subfolder.

Aljullu:
I'm going to add this to the Block Styling epic. If we want to have a set of class names that themes can rely on, we should consolidate our blocks on a set of components.

In WC Core themes can style all prices with something like:
```SCSS
.price {
	ins {
		font-weight: 400;
	}
}
```

In Blocks, that would not be possible because we are rendering prices in different components with different class names ([/base/components/cart-checkout/product-price/index.js](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/components/cart-checkout/product-price/index.js) and [/atomic/components/product/price/index.js](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/atomic/components/product/price/index.js), for example).

PR #2428, opened 2020-05-07
Fix product grid blocks markup & design inconsistencies

### LINKED ISSUE #2107: Product "On sale":  HTML structure inconsistency in the products grid
**Describe the bug**
There is HTML inconsistency of displaying the "On sale" container in the products grid

**To Reproduce**

1. Add the "Hand-picked Products" block or the "Newest Products" block or the "Best Selling Products" block to the content area. Make sure it contains the product on sale.
2. Add the "All Products" block to the content area. Make sure it contains the product on sale.
3. Compare the HTML structure of these blocks.

**Expected behavior**

The products grid has the same structure. One block displays this section in the `wc-block-grid__product-image` container while others display it outside of it. Also, one block is missing accessibility improvements in the "On sale" section.

**Screenshots**

The "Hand-picked Products" block or the "Newest Products" block or the "Best Selling Products" block.

Best Selling Products HTML

The "All Products" block.

All Products HTML

**WordPress:**

* Core version: [5.5-alpha-47534]
* WooCommerce version: [4.0.1]

### LINKED ISSUE #2231: Incosistent display of small product images in product grids
I'm opening this issue here even though it might end up requiring changes in Storefront and Core.

## To Reproduce
1. Create a product with a small image (50x50 px, for example).
2. See how it looks in the _Shop_ page, the _Handpicked products_ block and the _All Products_ block.
3. Notice sometimes it's scaled-up and sometimes it's not. Sometimes it's centered and sometimes it's left-aligned.

#### Shop
imatge

### All Products
imatge

### Handpicked Products
imatge

#### Shop
imatge

### All Products
imatge

### Handpicked Products
imatge

This pull in Storefront is related even though not completely fixing this: #1297.

### LINKED ISSUE #2414: Fix markup discrepancies between product grid blocks and All Products block
The markup of the _All Products_ block and the legacy product grid blocks is slightly different and that makes styles to render differently between them. This issue is about aligning both markups so styling is easier and we reduce the number of inconsistencies.

Aljullu:
Part of #1916.
Fixes #2107.
Fixes #2231.
Fixes #2414.

Changes included in this PR:
- Images in product grid blocks now occupy all the available width.
- _All Products_ prices have the same style as the 'legacy' product grid blocks.
- `Sale!` has been renamed to `Sale` and now it's a `div` instead of a `span` (shouldn't affect layout).
- There are some other small CSS changes.

### Screenshots

_Before (Handpicked products on top, All Products below):_
imatge

_After (Handpicked products on top, All Products below):_
imatge

cc @LevinMedia

1. Create a page with the _Handpicked products_ block and the _All Products_ block.
2. Both, in the editor and the frontend, verify they look almost the same. There might still be some differences like the _Sale_ badge position, but that can be changed by the user with the inner blocks. Product name colors are different per https://github.com/woocommerce/storefront/issues/1366.
3. Test the CSS snippets in the release notes section below and make sure they undo the changes introduced in this PR.

### Release notes
Images in product grid blocks now expand to occupy all the available horizontal space if they are small. You can undo this change with this CSS snippet:
```CSS
.wc-block-grid__products .wc-block-grid__product-image img {
	width: auto;
}
```

Prices in the _All Products_ block follow the same layout as the other product grid blocks. You can undo this change with:
```CSS
.wc-block-grid__product-price .wc-block-grid__product-price__regular {
	font-size: 0.8em;
	line-height: 1;
	color: #555;
	margin-top: -0.25em;
	display: block;
}
.wc-block-grid__product-price .wc-block-grid__product-price__value {
	letter-spacing: -1px;
	font-weight: 600;
	display: block;
	font-size: 1.25em;
	line-height: 1.25;
	color: #000;
	margin-left: 0;
}
.wc-block-grid__product-price .wc-block-grid__product-price__value span {
	white-space: nowrap;
}
```

### Changelog

> The All Products block and the other product grid blocks now share more styles and the markup is more similar (see release post or docs to learn how to undo this change).


mikejolley:
@Aljullu before I review this I just want to check something. 

Note the weight and size of the price on the current All Products block. These were design tweaks made intentionally - shouldn't we be updating the old grid blocks to match the _new_ styles, rather than reverting all products block styles?

Aljullu:
> Note the weight and size of the price on the current All Products block. These were design tweaks made intentionally - shouldn't we be updating the old grid blocks to match the new styles, rather than reverting all products block styles?

Yeah, good point, I should have specified that. I spoke about that with @LevinMedia in Slack. It was decided that:
- The prices will be displayed in one line, as they are in the 'legacy' product grid blocks.
- Smaller images must fill the available space, as they do in the _All Products_ block.

LevinMedia:
>Note the weight and size of the price on the current All Products block. These were design tweaks made intentionally - shouldn't we be updating the old grid blocks to match the new styles, rather than reverting all products block styles?

@mikejolley can you point me to where that decision was made? I don't want to steam roll something if I shouldn't I figured it was better to roll it back to be consistent with the legacy product blocks, since changing those to look like the all products block would impact significantly more stores.

mikejolley:
> @mikejolley can you point me to where that decision was made?  

I cannot sorry, but remember the designs for the All Products Block (and layouts) all came from design team and Josh. Maybe in figma somewhere, or a research post? I'm not sure where to look.

src/BlockTypes/AbstractProductGrid.php
```diff
@@ -357,7 +355,7 @@ public function render_product( $id ) {
 	 * @return string
 	 */
 	protected function get_image_html( $product ) {
-		return '<div class="wc-block-grid__product-image">' . $product->get_image( 'woocommerce_thumbnail' ) . '</div>';
+		return '<div class="wc-block-grid__product-image"><a href="' . esc_url( $product->get_permalink() ) . '">' . $product->get_image( 'woocommerce_thumbnail' ) . '</a></div>';
```
mikejolley, line 358: This would be a breaking change on the `woocommerce_blocks_product_grid_item_html` filter. The link was previously filterable.

nerrad:
> Blocks is an opportunity to change markup, classnames, styles, without worrying too much about the legacy markup found in woo core. In my mind, blocks is supposed to supersede and replace shortcodes, so I'm confused why we need to match them stylewise.

For what it's worth, I strongly agree with Mike here. If I understand correctly, there's competing approaches outlined in this pull:

## Approach One: Preserve styling history between Woo core and Blocks

As far as I can tell, the primary impetus for this approach is the goal of having fairly indistinguishable style between similar views in core and the blocks (i.e. all product grids look "the same" and inherit the same things from themes etc.).

### Pros
Assuming this is pulled off without as planned...

- merchants that have had their stores customized should have the blocks just fit right in.
- themes that have customized layouts etc should just have the blocks fit in.
- functionally there is no difference between how something like the block product grid looks on a site and the "legacy" product grids look on the site.

### Cons

- There's not much incentive for merchants to switch to the new blocks beyond potential additional functionality they have as from an appearance point of view they are just the same output.
- Maintaining this consistency is not easy because block skeleton's are different than "legacy" skeletons. So I think we'll likely have a uphill battle trying to preserve this consistency long term.
- Custom JS scripts that sites may have customized to hook in on existing css selectors on legacy views will also potentially hook in on the blocks which could have downsides when the skeleton isn't an exact match.
- We effectively make the _base_ design of all blocks the same as what has been in woo for a number of years _unless_ we decide to update those style values - which means we also potentially update _all_ sites.

## Approach 2: Blocks have their own selectors and implement new base style and design patterns. 

### Pros

- We have the opportunity to implement a _fresh_ default for the base design of various things.
- We can spot places to improve the selector usage and standardize around a convention to take us into the future.
- Less risk of conflict with existing themes and customizations out there which might not work well with the block skeleton.
- Less risk of custom JS scripts hooking in and having unexpected behaviour.
- There's a clear demarcation between the fresh new designs and behaviour introduced by blocks vs legacy views. This potentially introduces additional incentive to use the new blocks over legacy views.

### Cons

- We don't have blocks covering or replacing all views yet. This means that there is increased potential for inconsistent views on a site implementing both blocks and legacy views.
- There will be merchants, builders, or developers who _want_ to keep everything consistent in design across blocks and legacy views and this will require customizing the styles for the blocks along with their other site customizations.

## Wrap up

Out of the two approaches, I'm definitely leaning more to Approach 2, primarily because I think it's going to be harder to maintain consistency between shortcode or legacy content and block content going forward into the future of WordPress theming and implementation of block content. I'm concerned that there will be a continual selector battle between default theme/shortcode styling and more up-to-date design and styling defaults we have with blocks.

With a broader view, I also think of the upcoming _new_ blocks we introduce (such as Cart & Checkout) which _intentionally_ diverge from the "legacy" implementations. What kind of potential clash will exist between the blocks we've styled to match existing legacy views and blocks we've _intentionally_ styled to be _different_ from existing legacy views? I think the better approach is to decide on the default design language we want to use across *all* woo blocks consistently so there's no confusion why some blocks look shortcode output, but others have a more fresher look. 

From the standpoint of themers and those developing for sites, it also creates the potential for one day getting rid of old styles from the site and just keeping those specific to the blocks (so a gradual deprecation over time).

Aljullu:
Thanks for the feedback @nerrad and @mikejolley. I might not have been clear enough because I think there has been some misunderstanding here. This PR has no relation at all neither with shortcodes neither with the _Shop_ page template. Instead, it tries to resolve some markup differences between PHP product grid blocks (like _Hand-picked Products_ block, _Top Rated Products_ block, etc.) and the _All Products_ block.

So I completely agree with your concerns and with approach 2 presented by @nerrad.

But we have received some feedback (see #1916 and #2107, for example), about our blocks using different markup and class names whether they were PHP-based or JS-based. This PR was trying to solve that.

> Just would like clarification from design because IMO this is the wrong direction to take—aside from the earlier comments on the evolution of the design, this ties us back to legacy class names (`star-rating`, `price`, etc) which conflict much more easily in the wild due to no prefixes. 

If the issue is that I added the `star-rating` and `price` classes to the _All Products_ block, I'm happy to remove them. Indeed, I wasn't 100% sure about adding them in the first place, but given that we will not be able to remove them from PHP-based blocks in the short term, it made sense to me to keep consistency with the _All Products_ block. But I don't have a strong opinion on this, I see benefits from both approaches, so I'm fine removing them if there is an agreement. :+1: 

> I remember specifically a figma file with explorations into price range appearance and this was the chosen option.

Whether we should make the price style consistent between PHP and JS-based blocks and which one it should be, is something that I don't have a strong opinion either and I think it's a design decision more than a technical one. I think it all depends on whether we want blocks to be consistent or we consider them to serve different use-cases so it's fine to have different styles.

src/BlockTypes/AbstractProductGrid.php
Aljullu: Ah, good catch! I will work on a solution if the discussion below is addressed.

nerrad:
> Whether we should make the price style consistent between PHP and JS-based blocks and which one it should be, is something that I don't have a strong opinion either and I think it's a design decision more than a technical one. I think it all depends on whether we want blocks to be consistent or we consider them to serve different use-cases so it's fine to have different styles.

So it sounds to me that what we all are in agreement on is having consistent styles among our blocks and _less_ concerned with preserving consistency the old shortcode/template views andthe blocks.

We have a tricky situation here where folks may have _already styled_ the php based blocks (because they've been out longer) and that means we have to be careful about what we remove or add to those blocks because of how it may affect styling.

So here's a suggested proposal:

- we consider our base for consistency the new blocks. So whatever standard/convention we are using for classnames and/or styling, we use that.
- This means we align the old blocks with new classes and try to keep html structure the same too where possible. 
- We announce these changes accompanying a release, why we made them, and how folks can customize/fix their designs if they are affected.
- We make sure Storefront (and child themes) account for the changes.

While I know the main intention behind this pull is aligning the inconsistency between blocks, it does seem that at least in part, the older blocks were styled to be more consistent with existing shortcode/template output. So I do think we need to make sure the communication piece when this is released is taken care of (particularly the "Why").

Thoughts? (cc @LevinMedia as well).

garymurray:
Regarding the style changes proposed in this PR to the weight and size of the discounted price - I spent some time trying to find where this change was made/proposed by going through Figma/P2/Github/meetings and I cannot locate it.

In the absence of knowing why the change was made - i.e. if there was research to show it was better - I do feel we should change this back to the style proposed in the PR - not because it aligns with the existing shortcodes - but more so that it appears to be more consistent with the style used by online retailers. 

Currently the All Products block is also used by less stores than the others blocks - so changing it back would have far less of an impact than changing all the other blocks to align to the style as is in the current All Products block.

Aljullu:
> * we consider our base for consistency the new blocks. So whatever standard/convention we are using for classnames and/or styling, we use that.
> * This means we align the old blocks with new classes and try to keep html structure the same too where possible.

That's a tricky one because we don't have a direct control of the entire markup of PHP-based blocks, instead we render the templates returned by WC Core functions. For example, see https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/src/BlockTypes/AbstractProductGrid.php#L393.

If we want to customize the class names, we would need to update those functions in Core so they dynamically accept a custom class name. Considering we can achieve the same result via CSS selectors on our end, I'm not sure it makes sense updating several core functions just for that. Thoughts?

> We announce these changes accompanying a release, why we made them, and how folks can customize/fix their designs if they are affected.

That's a good point, sounds good to me. The changes this PR adds to PHP product grid blocks are:

1. Small images are now expanded to cover the entire width.
2. Product names are displayed as links instead of titles.

It shouldn't be difficult to add a CSS snippet to revert those changes. Should we add it to the `readme.txt` or leave it for the release post?

We can also document the changes to the _All Products_ block and offer CSS snippets to undo them.

> We make sure Storefront (and child themes) account for the changes.

For Storefront, this PR was already merged https://github.com/woocommerce/storefront/pull/1344 which fixes an issue with the _All Products_ block and incidentally it will make styles work if this PR is merged.

For child themes I took a quick look at some of them and couldn't find any adding specific styles to product grid blocks. Do you know if there is an easy way to search the code of all of them at once? Otherwise, would the release notes be enough?

Aljullu:
I rebased this PR so:
* It's up to date with latest master.
* I split up the fix for star ratings (see #2507).
* In cad1668fb12bebb1a7e97efb167b8e4666ae53d7 I undid some selector changes that might cause issues with themes.

This way this PR is smaller and completely focused towards the block inconsistencies.

nerrad:
> Should we add it to the readme.txt or leave it for the release post?

Both?

> For child themes I took a quick look at some of them and couldn't find any adding specific styles to product grid blocks. Do you know if there is an easy way to search the code of all of them at once? Otherwise, would the release notes be enough?

Honestly, I don't know 😄!  I'm less concerned about what our themes do out of the box and what custom changes folks might have made for their site. What's the risk of breakage there? I'm assuming in those cases any custom css on other's sites should still apply? Regardless, yes having something in release notes would be good.

@mikejolley can you give a final review/test of this so we can get merged?

Aljullu:
> What's the risk of breakage there? I'm assuming in those cases any custom css on other's sites should still apply?

Yeah, we aren't removing any class name, instead adding some new ones, so I think breakage with themes should be minimal.

However, after thinking about it, I'm a bit concerned about scaling up images and would like your thoughts on that. If a store has very narrow images, the layout might break:

`master`:
imatge

This branch:
imatge

Notice we already have this issue in _All Products_ in `master`. But with this branch we would introduce it to the 'legacy' product grid blocks. Should we instead unify the designs so small images are not scaled-up? cc @LevinMedia 

> @mikejolley can you give a final review/test of this so we can get merged?

I still need to take a look at this: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2428#discussion_r423617769, and would like to write the CSS snippets & release texts as part of this PR, so moving this PR back to In progress for now.

LevinMedia:
@Aljullu I feel like the vast majority of product images are going to be square or rectangular in aspect ratio, and that doing it the way you've got it in the PR is going work best. 👍

Aljullu:
@mikejolley this is ready for another look. Some comments:
- I updated the PR description with new testing steps and new screenshots.
- I added a release notes section with a proposed text that could go in the release post. Wondering if we should create a GitHub label to track PRs that have this kind of notes? cc @nerrad 
- @mikejolley I kept the `price` and `star-rating` classes, this way theme styles are inherited. The reason for this is that I don't think we can rely on themes switching to the new class names until #1967 is fixed.
- Instead of adding a note in `readme.txt`, I added it in the `docs` folder (you can read them [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fix/2414-product-grid-inconsistencies/docs/theming/product-grid-270.md)).

nerrad:
> I added a release notes section with a proposed text that could go in the release post. Wondering if we should create a GitHub label to track PRs that have this kind of notes? cc @nerrad

Maybe a `status:needs-dev-note` label? Then as a part of the release pull request any issues/pulls having that status indicates needing surfaced explicitly in release notes (as opposed to just a bullet item in the changelog).

assets/js/atomic/components/product/price/index.js
```diff
@@ -20,7 +20,7 @@ const ProductPrice = ( { className, product } ) => {
 			<div
 				className={ classnames(
 					className,
-					`${ layoutStyleClassPrefix }__product-price`
+					`${ layoutStyleClassPrefix }__product-price price`
```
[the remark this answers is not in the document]
Aljullu: I don't know! :smile: Should be fixed in the merge commit.

PR #2429, opened 2020-05-07
Fix X-WC-Store-API-Nonce updates 

### LINKED ISSUE #2406: Fix invalid X-WC-Store-API-Nonce Error on All Products block while logged out.
**Describe the bug**

At first I thought this was just an artifact of how I was testing but it's become consistent enough that I'm writing an issue to both track this and also have someone else confirm if they are getting it. If reproducible, I think it potentially blocks releasing 2.6.0 (hence the labels).

**To Reproduce**

1. Setup All Products block on a page (I have all filter blocks setup too, not sure if that matters, but if you can't reproduce with just the All Product block, maybe try that too).
2. Load the page in an icognito browser instance.
3. Click to add a product to the cart. Should be no error. Click to add a second item to the cart. This is where I'd get the nonce error and am unable to add any product to the cart until...
4. The weird thing here is that if I reload the page hosting the block, I won't get that error anymore so the nonce is clearly being refreshed okay.

So the weird thing is that I can _consistently_ reproduce this on every _initial_ load of an incognito browser instance.

**Expected behavior**
I expect to _never_ see the invalid `X-WC-Store-API-Nonce` error

* Core version: 5.4
* Gutenberg version (if installed): Not active
* WooCommerce version: 4.1
* WooCommerce Blocks version: Master
* Site language: English
* Any other plugins installed: Yes, but don't think they apply here.

* OS: Mac OSX
* Browser Chrome
* Version 81.0.4044.129 (Official Build) (64-bit)

The response object from the failed request:

```json
{"code":"woocommerce_rest_invalid_nonce","message":"X-WC-Store-API-Nonce is invalid.","data":{"status":403}}
```

mikejolley:
The `X-WC-Store-API-Nonce` refreshes after API requests if the customer session changes - this is more apparent for guests in incognito with no cart.

This process works fine, however, the `nonce` variable cache was not updating correctly because the `store-api-nonce.js` file was imported several times on the same page.

To fix this, and to ensure there is only ever one nonce stored, I've moved this into a separate script with it's own handle that is a dependency of the `data-store` script which uses needs this for `API_FETCH_WITH_HEADERS`.

Fixes #2406

The instructions in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2406 are correct; ensure you are not running PayPal express extension and have nothing in your cart!

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
 	},
 	globals: {
 		wcSettings: 'readonly',
+		wcStoreApiNonce: 'readonly',
 		page: true,
 		browser: true,
 		context: true,
--- a/assets/js/middleware/index.js
+++ b/assets/js/middleware/index.js
@@ -1,4 +1,4 @@
 /**
  * Internal dependencies
  */
-export * from './constants';
+import './store-api-nonce';
--- a/assets/js/middleware/store-api-nonce.js
+++ b/assets/js/middleware/store-api-nonce.js
@@ -2,10 +2,10 @@
  * External dependencies
  */
 import apiFetch from '@wordpress/api-fetch';
-import { getSetting } from '@woocommerce/settings';
 
+// @ts-ignore wcStoreApiNonce is window global
 // Cache for the initial nonce initialized from hydration.
-let nonce = getSetting( 'storeApiNonce' );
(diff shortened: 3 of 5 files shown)


nerrad:
Tested and confirmed fixed 🎉 I'll get this merged as soon as travis is 👍

PR #2432, opened 2020-05-07
Introduce ContainerQueryContextProvider - hide order summary on mobile

### LINKED ISSUE #2423: Checkout block: Order summary should be closed by default on mobile
See discussion in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2370#issuecomment-624107676. We want the _Order summary_ panel to be collapsed by default on mobile.

imatge

mikejolley:
Introduces a new context provider named `ContainerQueryContextProvider`. This consumes `useContainerQueries` so that all children in the tree can see the container size, for example, to know if isMobile is true or not.

This is used on the checkout order summary area so the default state of that is collapsed unless we're on a larger width device/container.

Fixes #2423

1. On desktop, view the checkout. Order summary is open.
2. Switch to mobile view. View checkout. Order summary is closed.
3. Check is-x classes remain on the checkout and cart wrappers.


assets/js/base/context/container-query-context.js
```diff
@@ -0,0 +1,53 @@
...
+import { useContainerQueries } from '@woocommerce/base-hooks';
+import classNames from 'classnames';
+
+const ContainerQueryContext = createContext( {
+	containerQueryClassName: '',
+	isMobile: false,
+	isSmall: false,
+	isMedium: false,
+	isLarge: false,
+	isLoading: true,
```
nerrad, line 15: `isLoading` could conflict with the prop exposed from `LoadingMask` right?  Or were you thinking consumers would just alias it if they need to?

assets/js/base/context/container-query-context.js
```diff
@@ -0,0 +1,53 @@
...
+import { createContext, useContext } from '@wordpress/element';
+import { useContainerQueries } from '@woocommerce/base-hooks';
+import classNames from 'classnames';
+
+const ContainerQueryContext = createContext( {
+	containerQueryClassName: '',
+	isMobile: false,
+	isSmall: false,
+	isMedium: false,
+	isLarge: false,
```
nerrad, line 14: If all of these are false by default, how will a consumer that implements `useContainerQueryContext` outside of the context tree behave? I think at least one of these values should be true by default.

assets/js/base/context/container-query-context.js
nerrad: In thinking about it, we could just leave this name as is. Consumers can alias.

nerrad: Or maybe `isCalculating` would be more accurate?

assets/js/base/context/container-query-context.js
Aljullu: Before the first render we don't know the container size, so I think it makes sense that all of them are `false` by default. Consumers can check `isLoading` value to know if the component was rendered or not. Actually that's how it's done in this PR in `sidebar/order-summary.js`.

assets/js/base/context/container-query-context.js
Aljullu: Yeah... I can't find a better name so they all work for me. But my feeling is that there is nothing really _loading_ or _calculating_ so that might be misleading. Would something like `isPristine` or `isUnrendered` sound good? I'm not sure about them either, though. :confused:

Aljullu:
Tested and working fine. :ok_hand: 

> - I wonder about the name of the provider. `ContainerQueryContext` seems a bit ambiguous about its purpose. I realize that I also approved `useContainerQueries` which is what the context name is related to. I'm not sure of what would be a better name (for the hook _and_ the context), maybe `ContainerWidthCalculatorContext`?  I'm not going to be sticky on this, if you and @Aljullu are happy with the name we can roll with it.

I don't have a strong opinion on this. I named it like that to match the [CSS standard proposal](https://github.com/WICG/container-queries/). But considering our implementation is JS-based, I wouldn't argue about finding another name.

`ContainerWidthCalculatorContext` sounds good to me. Just wondering if the word `Calculator` is needed? Would `ContainerWidthContext` work? I will let @mikejolley decide whether it should be renamed.

assets/js/base/context/container-query-context.js
nerrad: > Actually that's how it's done in this PR in sidebar/order-summary.js.
Ahh right!

assets/js/base/components/sidebar-layout/sidebar-layout.js
```diff
@@ -3,27 +3,20 @@
...
-	const [ resizeListener, containerQueryClassName ] = useContainerQueries();
-
 	return (
-		<div
-			className={ classNames(
-				'wc-block-sidebar-layout',
-				className,
-				containerQueryClassName
-			) }
+		<ContainerQueryContextProvider
```
senadir, line 15: This is mostly a question to @Aljullu that I forget to mention in the initial PR, why is this wrapping the sidebar only? wouldn't we need it in the main view?

Aljullu: Actually it's wrapping everything, the structure is like this:
```HTML
<div class="wc-block-sidebar-layout">
	<div class="wc-block-main">...</div>
	<div class="wc-block-sidebar">...</div>
</div>
```
I know the class name `wc-block-sidebar-layout` for the wrapper might be a bit misleading, but couldn't find a better alternative. I would be happy to rename it if somebody has a better idea. :slightly_smiling_face:

assets/js/base/context/container-query-context.js
nerrad: `isRendering`?

mikejolley:
I've gone ahead and renamed those items. Once tests pass this just needs final approval.

assets/js/type-defs/contexts.js
```diff
@@ -94,51 +74,43 @@
...
- *                                might contain)
+ * @property {Object}  method     The payment method object (varies on what it might contain)
  * @property {string}  expires    Short form of expiry for payment method.
- * @property {boolean} is_default Whether it is the default payment method of
- *                                the customer or not.
+ * @property {boolean} is_default Whether it is the default payment method of the customer or not.
  * @property {number}  tokenId    The id of the saved payment method.
- * @property {Object}  actions    Varies, actions that can be done to interact
- *                                with the payment method.
+ * @property {Object}  actions    Varies, actions that can be done to interact  with the payment
```
Aljullu, line 99: Not blocking, but a double-space was introduced here.
```suggestion
 * @property {Object}  actions    Varies, actions that can be done to interact with the payment
```

assets/js/type-defs/contexts.js
```diff
@@ -273,12 +245,10 @@
...
  * @typedef {Object} EditorDataContext
  *
- * @property {boolean}                 isEditor           Indicates whether in
- *                                                        the editor context
- *                                                        (true) or not (false).
+ * @property {boolean}                 isEditor           Indicates whether in the editor context.
  * @property {number}                  currentPostId      The post ID being edited.
- * @property {Object}                  previewData        Object containing preview
- *                                                        data for the editor.
+ * @property {Object}                  previewData        Object containing preview  data for the
```
Aljullu, line 250: And here. :slightly_smiling_face: 
```suggestion
 * @property {Object}                  previewData        Object containing preview data for the
```

PR #2436, opened 2020-05-07
Add eslint-plugin-react-hooks

nerrad:
This pull adds the [eslint-plugin-react-hooks](https://github.com/facebook/react/tree/master/packages/eslint-plugin-react-hooks) package so that we can get warned when there may be missing dependencies for hooks.

I've noticed that a frequent conversation that has come up in pull reviews and in slack for the team is around the usage of dependencies and it's also something that I (and others) have missed in reviews. Adding this eslint config will help with alerting us when we're not adding necessary dependencies (which can lead to subtle bugs that can turn up over time).

Note:

- This is set to `warn` not `error`. There _may_ be the occasional _rare_ case where dependencies are not included but that should be documented with the code.
- Currently there are ~60 warnings in the code with linting running after adding this. I think the best plan here would be to fix those warnings as we work in the related code (or tackle them in individual pulls) otherwise we have a lot of testing to do to cover all the impacted things (vs doing it in smaller chunks that will already be tested).

## Testing

Since I didn't make any actual code changes with this pull, there's no impact to existing logic.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -3,6 +3,7 @@ module.exports = {
 		'plugin:@wordpress/eslint-plugin/recommended',
 		'prettier',
 		'plugin:jest/recommended',
+		'plugin:react-hooks/recommended',
 	],
 	env: {
 		'jest/globals': true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -6007,6 +6007,12 @@
             "@typescript-eslint/experimental-utils": "^1.13.0"
           }
         },
+        "eslint-plugin-react-hooks": {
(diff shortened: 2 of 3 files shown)


nerrad:
> I saw a bunch of warnings that we should follow up!

Yup that's why I wrote this in the description:

> Currently there are ~60 warnings in the code with linting running after adding this. I think the best plan here would be to fix those warnings as we work in the related code (or tackle them in individual pulls) otherwise we have a lot of testing to do to cover all the impacted things (vs doing it in smaller chunks that will already be tested).

Issue #2426, opened 2020-05-07
Allow merchant to disable payment via stripe saved cards - respect Stripe "Enable Payment via Saved Cards" option


haszari:
When testing #2425 I saw that Stripe has a setting to disable saved cards:

![Screen Shot 2020-05-07 at 2.14.27 PM.png](https://images.zenhubusercontent.com/5ddb3cb39529f10001e618e5/1d11d9cf-1602-42b9-90af-6c1df8e73803)

If this is not checked, I'm thinking we shouldn't show the saved cards option in checkout (and possibly the option to save cards). Here's what the old shortcode does (note no saved cards, and no `Save my card` checkbox.)

![Screen Shot 2020-05-07 at 2.15.57 PM.png](https://images.zenhubusercontent.com/5ddb3cb39529f10001e618e5/a8bf00c6-dba6-4d27-822f-1fe7fd4c44f8)

Closely related to #2411 (see also #2350).

haszari:
cc @nerrad I've added this to current sprint

PR #2453, opened 2020-05-11
Allow shopper to save Stripe payment method in user account for subsequent purchases

### LINKED ISSUE #2411: Checkout: "Save payment information to my account" doesn't save the payment method

haszari:
Fixes #2411

This PR allows shoppers to save payment details to their account so they can checkout future purchases quicker (by using a saved payment method).

The API had generic support for this, as `paymentMethodData` is passed through to the Stripe payment processing handler server side. This code looks for `wc-stripe-new-payment-method` in `POST` and if specified the payment method is saved to the user account (via [`$customer->add_source()`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/master/includes/abstracts/abstract-wc-stripe-payment-gateway.php#L622).

So in this PR the fix is to pass `shouldSavePayment` using this existing key.

Also in this PR I've set the checkbox to be unchecked by default, consistent with previous (shortcode checkout) behaviour. I'm not sure if this is intentional/by design, going with opt-in as the safest option. We can switch this to enabled by default if needed.

Note there's a commented out section when (I _think_) saved card is used. This is untested (other PR in flight for using saved card to pay - #2425). If this is a saved method, then it's not appropriate to send `wc-stripe-new-payment-method`/`shouldSaveCard` - as it's not a new payment method, is already saved.

See also related PRs/issues: #2425 #2426 

And #2450 (merged)

1. Add something to cart, proceed to checkout.
2. Select new stripe card payment option and enter [test card details](https://stripe.com/docs/testing#cards).
3. Check `Save payment information to my account` checkbox.
6. Submit purchase.

Confirm that the card is saved - should see it in `My account` and on subsequent purchases should be available as a saved payment option (checkout block and shortcode). Should work as a payment option in shortcode checkout.

Note that #2425 is in progress - saved payment options don't currently work in checkout block.

### DIFF
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -16,10 +16,12 @@ import {
 } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import {
+	useCheckoutContext,
 	useEditorContext,
 	usePaymentMethodDataContext,
 } from '@woocommerce/base-context';
 import { PaymentMethodIcons } from '@woocommerce/base-components/cart-checkout';
+import CheckboxControl from '@woocommerce/base-components/checkbox-control';
 
 /**
  * Internal dependencies
@@ -67,6 +69,8 @@ const PaymentMethods = () => {
 	const [ selectedToken, setSelectedToken ] = useState( '0' );
 	const { noticeContexts } = useEmitResponse();
 	const { removeNotice } = useStoreNotices();
+	const { customerId } = useCheckoutContext();
+	const [ shouldSavePayment, setShouldSavePayment ] = useState( false );
 
 	// update ref on change.
 	useEffect( () => {
@@ -81,16 +85,38 @@ const PaymentMethods = () => {
 				currentPaymentMethods.current,
 				isEditor
 			);
+			const { options } = currentPaymentMethods.current[
+				activePaymentMethod
+			];
 			return paymentMethod ? (
 				<PaymentMethodErrorBoundary isEditor={ isEditor }>
 					{ cloneElement( paymentMethod, {
 						activePaymentMethod,
 						...currentPaymentMethodInterface.current,
 					} ) }
+					{ customerId > 0 && options.allowSavePaymentToken && (
+						<CheckboxControl
+							className="wc-block-checkout__save-card-info"
+							label={ __(
+								'Save payment information to my account for future purchases.',
+								'woo-gutenberg-products-block'
+							) }
+							checked={ shouldSavePayment }
+							onChange={ () =>
+								setShouldSavePayment( ! shouldSavePayment )
+							}
+						/>
+					) }
 				</PaymentMethodErrorBoundary>
 			) : null;
 		},
-		[ isEditor, activePaymentMethod ]
+		[
+			isEditor,
+			activePaymentMethod,
+			shouldSavePayment,
+			setShouldSavePayment,
+			customerId,
+		]
 	);
 
 	if (
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -86,7 +86,7 @@ const PaymentMethods = () => {
 				currentPaymentMethods.current,
 				isEditor
 			);
-			const { options } = currentPaymentMethods.current[
+			const { supports } = currentPaymentMethods.current[
 				activePaymentMethod
 			];
 			return paymentMethod ? (
@@ -95,7 +95,7 @@ const PaymentMethods = () => {
 						activePaymentMethod,
 						...currentPaymentMethodInterface.current,
 					} ) }
-					{ customerId > 0 && options.allowSavePaymentInfo && (
+					{ customerId > 0 && supports.savePaymentInfo && (
 						<CheckboxControl
 							className="wc-block-checkout__save-card-info"
 							label={ __(
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -58,6 +58,8 @@ const PaymentMethods = () => {
 	const {
 		customerPaymentMethods = {},
 		setActivePaymentMethod,
+		shouldSavePayment,
+		setShouldSavePayment,
 	} = usePaymentMethodDataContext();
 	const { isInitialized, paymentMethods } = usePaymentMethods();
 	const currentPaymentMethods = useRef( paymentMethods );
@@ -70,7 +72,6 @@ const PaymentMethods = () => {
 	const { noticeContexts } = useEmitResponse();
 	const { removeNotice } = useStoreNotices();
 	const { customerId } = useCheckoutContext();
-	const [ shouldSavePayment, setShouldSavePayment ] = useState( false );
 
 	// update ref on change.
 	useEffect( () => {
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -95,7 +95,7 @@ const PaymentMethods = () => {
 						activePaymentMethod,
 						...currentPaymentMethodInterface.current,
 					} ) }
-					{ customerId > 0 && options.allowSavePaymentToken && (
+					{ customerId > 0 && options.allowSavePaymentInfo && (
 						<CheckboxControl
 							className="wc-block-checkout__save-card-info"
 							label={ __(
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -28,15 +28,23 @@ import { useStoreCart, useStoreNotices } from '@woocommerce/base-hooks';
  *
  * @param {Object} paymentData Arbitrary payment data provided by the payment
  *                             method.
+ * @param {boolean} shouldSave Whether to save the payment method info to
+ *                             user account.
  *
  * @return {PaymentDataItem[]} Returns the payment data as an array of
- *                                 PaymentDataItem objects.
+ *                             PaymentDataItem objects.
  */
-const preparePaymentData = ( paymentData ) => {
-	return Object.keys( paymentData ).map( ( property ) => {
+const preparePaymentData = ( paymentData, shouldSave ) => {
+	const apiData = Object.keys( paymentData ).map( ( property ) => {
 		const value = paymentData[ property ];
 		return { key: property, value };
 	}, [] );
+	const savePaymentMethodKey = `wc-${ paymentData.paymentMethod }-new-payment-method`;
+	apiData.push( {
+		key: savePaymentMethodKey,
+		value: shouldSave,
+	} );
+	return apiData;
 };
 
 /**
@@ -66,6 +74,7 @@ const CheckoutProcessor = () => {
 		paymentMethodData,
 		expressPaymentMethods,
 		paymentMethods,
+		shouldSavePayment,
 	} = usePaymentMethodDataContext();
 	const { addErrorNotice, removeNotice } = useStoreNotices();
 	const currentBillingData = useRef( billingData );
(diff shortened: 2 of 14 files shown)


haszari:
FYI @nerrad - keen for your thoughts/feedback on this too. Feel free to grab the PR and run with it if it's close.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-payment-processing.js
```diff
@@ -92,7 +92,9 @@ export const usePaymentProcessing = (
 								paymentMethod: PAYMENT_METHOD_NAME,
 								paymentRequestType: 'cc',
 								stripe_source: sourceId,
-								shouldSavePayment,
+								// I think this is a saved token, so `shouldSavePayment`
+								// is redundant.
+								// 'wc-stripe-new-payment-method': shouldSavePayment,
```
haszari, line 97: I _think_ we can delete all this, not relevant when paying with a saved payment method token.

haszari:
Looks like e2e tests are spectacularly broken, I don't think this is related to this branch?!

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
```diff
@@ -34,9 +34,9 @@ const CreditCardComponent = ( {
 	const { customerId } = billing;
 	const [ sourceId, setSourceId ] = useState( '' );
 	const stripe = useStripe();
-	const [ shouldSavePayment, setShouldSavePayment ] = useState(
-		customerId ? true : false
-	);
+	// Saving payment method is always opt-in (based on previous/shortcode
+	// behaviour); so default to false.
+	const [ shouldSavePayment, setShouldSavePayment ] = useState( false );
```
nerrad, line 39: The design called for checked by default, but I do think this should be opt-in so I'm fine with this change. 

cc @garymurray just for a gut check here. Gary, I pinged you to get your feedback on whether the "Save payment information to my account for future purchase." checkbox should default to checked or default to unchecked. Currently (according to Rua) the shortcode checkout has this checkbox default to _unchecked_.  Original implementation prior to this pull was to default to checked in keeping with the design specs but its unclear whether the design was communicating the desired default or not.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
```diff
@@ -34,9 +34,9 @@ const CreditCardComponent = ( {
 	const { customerId } = billing;
 	const [ sourceId, setSourceId ] = useState( '' );
 	const stripe = useStripe();
-	const [ shouldSavePayment, setShouldSavePayment ] = useState(
-		customerId ? true : false
-	);
+	// Saving payment method is always opt-in (based on previous/shortcode
+	// behaviour); so default to false.
```
nerrad, line 38: I don't think we need this comment block. We don't need rationale in the code for every default decision. If you are leaving it to help guide reviews, one way I deal with that is I'll leave my _own_ review comments on a pull for reviewers to see (rather than comment directly in the code).

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-payment-processing.js
nerrad: Two things:

- comment blocks are unnecessary - let's remove them.
- we'll still want to include the `wc-stripe-new-payment-method` property and value pair for this object. The reason why this block is here is in case there was an error needing handled during payment processing, the payment method keeps the already created source_id to avoid creating it again.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-payment-processing.js
```diff
@@ -131,14 +133,15 @@ export const usePaymentProcessing = (
 					);
 				}
 				setSourceId( response.source.id );
+				const savePaymentMethodKey = `wc-${ PAYMENT_METHOD_NAME }-new-payment-method`;
```
nerrad, line 136: 👍 nice that this is all that is needed. However it does make me wonder since the savePaymentMethodKey can be dynamically arrived, we should probably explore shifting the saved payment method checkbox component _outside_ of payment methods all-together and have checkout handle it. That will remove one more thing from payment methods to have to worry about and gives us tighter control over the behaviour of this checkbox.

haszari:
> I have a few comments, but from the code you've written, it confirms my suspicions that we could probably remove the save payment method checkbox from the payment methods all together and just have checkout handle it (as noted in one of my review comments). I'd like us to try that approach instead because it will simplify what payment methods have to worry about.

That would potentially be a design/UX change - right now that checkbox is displayed as part of the stripe CC tab, and is closely associated with it. @nerrad can we clarify the design/expectations here - since it's a change from the legacy model (and I think the block UI design).

If we make the code/API side of it generic, do we change the UI to match that model? For example, show the "save my payment method" checkbox outside the payment method tabs vs. require that each payment method provides a checkbox somewhere and hooks it up to the `wc-XXX-new-payment-method` POST field.

nerrad:
> That would potentially be a design/UX change - right now that checkbox is displayed as part of the stripe CC tab, and is closely associated with it. @nerrad can we clarify the design/expectations here - since it's a change from the legacy model (and I think the block UI design).

To be **clear**, from the perspective of design, it would behave the same as it is now, checkout would only output the checkbox in the _same_ area as the payment method and only for the payment methods indicating they support saved payment methods. So visually, there will be _no change_ from what we have now.

nerrad:
> If we make the code/API side of it generic, do we change the UI to match that model? For example, show the "save my payment method" checkbox outside the payment method tabs vs. require that each payment method provides a checkbox somewhere and hooks it up to the wc-XXX-new-payment-method POST field.

Effectively, checkout would just add the checkbox in the rendered tab below the cloned payment method element (but only for payment methods indicating support for saving payment methods):

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1f2af79cb7a0cd93da4e0b5cd1ad73c54c4dbe94/assets/js/base/components/payment-methods/payment-methods.js#L77-L94

assets/js/base/context/cart-checkout/payment-methods/constants.js
```diff
@@ -17,14 +17,18 @@ export const ACTION_TYPES = {
 	SET_REGISTERED_PAYMENT_METHODS: 'set_registered_payment_methods',
 	SET_REGISTERED_EXPRESS_PAYMENT_METHODS:
 		'set_registered_express_payment_methods',
+	SET_SHOULD_SAVE_PAYMENT_METHOD: 'set_should_save_payment_method',
```
haszari, line 20: I managed to get a long way without defining my new action constant here – I think I was switching on `undefined` in the reducer (🙀). Any ideas how we can make this more bulletproof?

Maybe in other files we import `ACTION_TYPES` and then use full `ACTION_TYPES.bla`, or perhaps we have a module just for action types, with each exported individually.

assets/js/base/context/cart-checkout/payment-methods/constants.js
```diff
@@ -17,14 +17,18 @@ export const ACTION_TYPES = {
...
 };
 
 /**
  * @todo do typedefs for the payment event state.
  */
 
+// Note - if fields are added/shape is changed, you may want to update
+// PRISTINE reducer clause to preserve your new field.
```
haszari, line 28: cc @nerrad  - added this comment as it bit me, the checkbox state kept getting reset. Not sure if it helps much but thought this might bite someone else in future.

haszari:
This is ready for another follow up review @nerrad :)

Now the checkbox and API support is handled automatically. Payment methods only need to opt in and handle saving on the server side.

assets/js/base/context/cart-checkout/payment-methods/constants.js
nerrad: > Maybe in other files we import ACTION_TYPES and then use full ACTION_TYPES.bla, or perhaps we have a module just for action types, with each exported individually.

The reducer _does_ import `ACTION_TYPES` and then destructure to use. I'm not sure how what you are describing would solve the problem you experienced? One reason for setting things up the way they are now is to clarify the dependencies via imports. I'm open to ideas but I'm not sure how you can make that more "bulletproof" against incorrect implementation?

assets/js/blocks-registry/payment-methods/payment-method-config.js
```diff
@@ -20,6 +20,10 @@ export default class PaymentMethodConfig {
 		this.edit = config.edit;
 		this.canMakePayment = config.canMakePayment;
 		this.paymentMethodId = config.paymentMethodId || this.name;
+		this.options = {
+			allowSavePaymentInfo: false,
+			...config.options,
+		};
```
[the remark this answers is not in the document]
nerrad: I added changes for this in 5186a2a

nerrad:
Alright, this has the updated changes and tests well. Just need a confidence check on my commits.

assets/js/base/components/payment-methods/payment-methods.js
```diff
@@ -81,16 +86,38 @@ const PaymentMethods = () => {
...
+					{ customerId > 0 && supports.savePaymentInfo && (
+						<CheckboxControl
+							className="wc-block-checkout__save-card-info"
+							label={ __(
+								'Save payment information to my account for future purchases.',
+								'woo-gutenberg-products-block'
+							) }
+							checked={ shouldSavePayment }
+							onChange={ () =>
+								setShouldSavePayment( ! shouldSavePayment )
```
mikejolley, line 107: This is minor, but this seems risky to me since it toggles the boolean without considering the state of `checked`. I know in this context we know the initial state but I'm wondering what would happen if something modified this firing events.

nerrad:
Good feedback Mike. I think the focus issue might be fixed with the work done in #2371 but at a minimum testing in there should pick this up hopefully.

assets/js/base/components/payment-methods/payment-methods.js
nerrad: > I'm wondering what would happen if something modified this firing events.

Do you have some examples in mind where that might occur? Are you talking about something hooking directly into the dom event?

mikejolley: Regular javascript, possibly browser autocomplete (if that doesn't fire change events?). Basically this code is trusting the initial state without checking the state onChange.

nerrad: 🤔 I think what we're doing here is basically saying the value of the checkbox is connected to the current state. So thinking through what _could_ happen:

- If external javascript changes the checkbox, it's not going to do anything (because it depends on the app state). IMO that's okay, this is a controlled component - we don't _want_ external javascript affecting the app state.
- If autocomplete causes state to change... what autocomplete would that be? This isn't a standard autocomplete field?

By setting state this way we're essentially saying that the _state_ of the checkbox is controlled by the component and _reacts_ to user changes on the checkbox.

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -28,15 +28,23 @@ import { useStoreCart, useStoreNotices } from '@woocommerce/base-hooks';
...
+	const apiData = Object.keys( paymentData ).map( ( property ) => {
 		const value = paymentData[ property ];
 		return { key: property, value };
 	}, [] );
+	const savePaymentMethodKey = `wc-${ paymentData.paymentMethod }-new-payment-method`;
+	apiData.push( {
+		key: savePaymentMethodKey,
+		value: shouldSave,
+	} );
+	return apiData;
```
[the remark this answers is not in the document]
haszari: Aha, that did feel just a little too "convenient" - thanks for making this more robust :)

Issue #2407, opened 2020-05-05
Broken chip styles


LevinMedia:
**Describe the bug**
Chips in the editor interface are broken

**To Reproduce**
View a block such as products by category - note the chips have broken styles. 

**Expected behavior**
Compare the style of the chips in the blocks to the chips in the analytics filters. They should match. 

**Screenshots**

Expected:

<img width="351" alt="Screen Shot 2020-05-05 at 1 27 07 PM" src="https://user-images.githubusercontent.com/4500952/81113816-a83d7680-8ed5-11ea-9aa1-81901fc6afb6.png">

Observed:

<img width="777" alt="Screen Shot 2020-05-05 at 1 41 48 PM" src="https://user-images.githubusercontent.com/4500952/81114095-2f8aea00-8ed6-11ea-80e3-e755d2c84391.png">

<img width="274" alt="Screen Shot 2020-05-05 at 1 40 07 PM" src="https://user-images.githubusercontent.com/4500952/81113977-f488b680-8ed5-11ea-9c3a-f1e2d82fb2ed.png">

Gutenberg version: 8.0.0
WordPress Version: 5.4.1
WooCommerce Version: 4.1.0-rc.2

Aljullu:
This was fixed in https://github.com/woocommerce/woocommerce-admin/pull/3993, but there wasn't a release of `@woocommerce/components` yet, so we are using a version without this fix.

Fixing #2419 will probably solve this.

nerrad:
Since #2419 is in Current Sprint, I'm going to move this there too. It can be bumped if #2419 does _not_ fix this.

Aljullu:
Now that it has been decided that we will not load components from the external, this is not blocked by #2419. Instead, we need to update `@woocommerce/components` to version 5.1.0 or later, but it hasn't been released yet.

Related issue in WC Admin: https://github.com/woocommerce/woocommerce-admin/issues/5006.

senadir:
Notes from: C8X6Q7XQU-slack-p1600163121223700
- @Aljullu tried to update to 5.1.0 but failed.
- Blocks weren't registered, this is mostly due to a dependency issue missing.
[...]

PR #2456, opened 2020-05-11
Update placeholder styles

### LINKED ISSUE #2408: Fix padding, alignment, and various UI inconsistencies in the block editor interfaces
There are various padding and alignment issues in block editor UI - these bugs are in the preview state of the blocks, as well as in the sidebar controls

Preview state:

Screen Shot 2020-05-05 at 1 52 01 PM

Sidebar:

Screen Shot 2020-05-05 at 1 54 01 PM

Additionally some preview states have a grey background, while others have a white background:

Screen Shot 2020-05-04 at 9 46 47 PM

Screen Shot 2020-05-04 at 9 44 56 PM

Gutenberg version: 8.0.0
WordPress Version: 5.4.1
WooCommerce Version: 4.1.0-rc.2

Aljullu:
Fixes #2408.

### Screenshots
_Before:_
imatge

_After:_
imatge

_(notice chip styles are broken, that's being tracked in: #2407)_

1. Activate the latest version of Gutenberg.
2. Add a _Products by Category_ and a _Featured Product_ block in a post.
3. Verify the placeholder background is white, buttons are aligned, there is a border around the list, etc.
3. Deactivate Gutenberg and verify styles still look good and nothing is broken.

### Changelog

> Updated the placeholder styles to match the core block styles.

### DIFF
--- a/assets/css/editor.scss
+++ b/assets/css/editor.scss
@@ -55,3 +55,12 @@
 .components-placeholder .woocommerce-search-list__list {
 	border: 1px solid $core-grey-light-500;
 }
+
+// Work-around to make the SearchList component work fine with the last versions of @wordpress/components.
+// Ideally it should be possible to remove this once this issue is fixed in WC Admin:
+// https://github.com/woocommerce/woocommerce-admin/issues/4349
+.components-placeholder .woocommerce-search-list__list .woocommerce-search-list__item {
+	border-radius: 0;
+	height: auto;
+	text-align: left;
+}
--- a/assets/css/editor.scss
+++ b/assets/css/editor.scss
@@ -50,3 +50,8 @@
 		margin: -$gap 0 $gap-small;
 	}
 }
+
+// Adds border to placeholder lists, so they can be distinguished from the white background.
+.components-placeholder .woocommerce-search-list__list {
+	border: 1px solid $core-grey-light-500;
+}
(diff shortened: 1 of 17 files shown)


Aljullu:
> I'm running Gutenberg 7.9, the styles are not aligned in the middle

I'm unable to reproduce it with Gutenberg 8.0 neither `master`. :confused: Could you try if you can reproduce it with the latest version of Gutenberg from .org or `master`?

> the hover/focus background contrast is low (the background color changed, so I guess that was the reason).
> The focus border is still shifted as well.

Yeah, I noticed that as well and was planning to take a deeper look later. This seems to be an issue between WC Admin's `<SearchListControl />` and the last versions of `@wordpress/components`, so we should probably fill an issue in wc-admin's repo and work on it there.

senadir:
I should have noted that I only tested in Chrome, after updating the issue was still there, I tested in Firefox and it seems to work fine.

Aljullu:
> I should have noted that I only tested in Chrome, after updating the issue was still there, I tested in Firefox and it seems to work fine, so only chrome for now is buggy.

Ah, good catch, I should have tested Chrome as well. This also seems to be an issue with `SearchListControl` styles.

I opened an issue in WC Admin so we can track it: https://github.com/woocommerce/woocommerce-admin/issues/4349, but given that those changes will need to be done in the other repo, I think they shouldn't block this PR.

nerrad:
@senadir in light of Albert's most recent comment, are you in agreement this pull is ready for merge?

Aljullu:
> .woocommerce-search-list__list .woocommerce-search-list__item needs a height: auto. This prevents the button being too small which is why the text is off center.

Thanks for the tip! I really feel the fix belongs to WC Admin. That's why I created https://github.com/woocommerce/woocommerce-admin/issues/4349, but updating `@wordpress/components` there is a bit tricky (needs changes in several components), so it might make sense adding a work-around on our end for now. Done in ec48c1ca65519a468de26d8dca634d1cc8ae05e1.

PR #2461, opened 2020-05-11
Use display table instead of flex in radio control group

mikejolley:
Fixes label wrapping (see https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2425#pullrequestreview-409165842) in radio components.

Before: 
81566459-d2d67600-9392-11ea-8678-b72bbdf07e4f

After:
Screenshot 2020-05-11 at 15 14 45
Screenshot 2020-05-11 at 15 13 54

### DIFF
--- a/assets/js/base/components/radio-control/_mixin.scss
+++ b/assets/js/base/components/radio-control/_mixin.scss
@@ -7,10 +7,9 @@
 	}
 
 	.wc-block-radio-control__option-layout {
+		display: table;
+		width: 100%;
 		border-bottom: 1px solid $core-grey-light-600;
-		display: flex;
-		flex-wrap: wrap;
-		justify-content: space-between;
 		padding-bottom: $gap-small;
 	}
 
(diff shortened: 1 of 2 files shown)


assets/js/base/components/radio-control/_mixin.scss
```diff
@@ -24,27 +23,35 @@
...
+		display: table-row;
+
+		span {
+			display: table-cell;
+		}
+
+		.wc-block-radio-control__secondary-label,
+		.wc-block-radio-control__secondary-description {
+			text-align: right;
+			min-width: 50%;
```
Aljullu, line 37: Would it work removing `min-width` from here? For me, it seems to cause the shipping name to break earlier than necessary:

_with `min-width`:_
imatge

_without `min-width`:_
imatge

mikejolley: It's just spans within spans, easy fix :)

mikejolley:
All fixed @Aljullu - just a selector issue.

mikejolley:
@Aljullu I'd prefer a follow up if that is actually an issue. I have a feeling it's there so that when you have secondary labels or descriptions, it forces the secondary one from being pushed too small if the primary description is long.

Thanks!

PR #2464, opened 2020-05-12
Avoid const and let in inline JS

### LINKED ISSUE #2439: JS error when add both Cart & Checkout blocks to a post
### To reproduce
- Edit a post or page.
- Add the `Cart` block.
- Add the `Checkout` block.
- Publish, view on front end.

Expected: works (?) or prevent adding both to same page. See also related issue for post vs. page: #2440

Actual: JavaScript error in console: `Uncaught SyntaxError: Identifier 'containers' has already been declared`

Maybe related to loading skeleton stuff, error is on this line:

```js
const containers = document.querySelectorAll( 'div.wc-block-skeleton' );
```

mikejolley:
Since these inline scripts can be loaded in multiple times, avoid `const` and `let`.

Fixes #2439

1. Add cart and checkout blocks to a page
2. View console errors on frontend.

### DIFF
--- a/src/BlockTypes/AbstractBlock.php
+++ b/src/BlockTypes/AbstractBlock.php
@@ -151,12 +151,12 @@ abstract class AbstractBlock {
 	 */
 	protected function get_skeleton_inline_script() {
 		return "<script>
-			const containers = document.querySelectorAll( 'div.wc-block-skeleton' );
+			var containers = document.querySelectorAll( 'div.wc-block-skeleton' );
 
 			if ( containers.length ) {
 				Array.prototype.forEach.call( containers, function( el, i ) {
-					const w = el.offsetWidth;
-					let classname = '';
+					var w = el.offsetWidth;
+					var classname = '';
 
 					if ( w > 700 )
 						classname = 'is-large';


src/BlockTypes/AbstractBlock.php
```diff
@@ -151,12 +151,12 @@ protected function enqueue_scripts( array $attributes = [] ) {
...
 		return "<script>
-			const containers = document.querySelectorAll( 'div.wc-block-skeleton' );
+			var containers = document.querySelectorAll( 'div.wc-block-skeleton' );
 
 			if ( containers.length ) {
 				Array.prototype.forEach.call( containers, function( el, i ) {
-					const w = el.offsetWidth;
-					let classname = '';
+					var w = el.offsetWidth;
+					var classname = '';
```
Aljullu, line 159: Actually here it would be fine using `const` and `let` because they are scoped inside the inner function. But having it as `var` makes sense as well to keep it consistent.

senadir:
those code blocks could also be scoped in a `{}` to fix this, this also works.

src/BlockTypes/AbstractBlock.php
mikejolley: Yeah that's why I changed it. I didn't want to mix const/let with old vars.

PR #2465, opened 2020-05-12
Remove Shipping totals editor.scss file

Aljullu:
Coming from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2448#pullrequestreview-408415164. Most of the styles in `totals-shipping-item/editor.scss` could be either removed or moved to `style.scss` just doing some small tweaks to the specificity of selectors. This PR handles that and also renames some CSS class names of those components.

### Screenshots
imatge
_(there shouldn't be any visual difference)_

Just do some smoke testing of the _Cart_ block in the editor and the frontend to verify there are no visual regressions.


senadir:
I noticed a weird googly eye in your screenshot, I couldn't reproduce it in both Firefox and chrome, editor or frontend, so I'm sure why it existed.

image

Aljullu:
Good catch. I debugged it in the past and I'm pretty sure it's not a bug in our end. I couldn't find steps to reproduce it, it just randomly happens sometimes. But I think it must be a bug in Firefox; I'm using WebRender, which is still beta in Linux.

PR #2466, opened 2020-05-12
Update Stripe Method to display generic icon and text in label, with cards below

### LINKED ISSUE #2438: No text title on credit card payment method tab
Ideally the payment method tab should also include a text title, e.g. `Credit Card` alongside some card icons.

Here's a mockup from @garymurray - tagging in @LevinMedia for next steps:

Summary of considerations / ideas:

- Include text title and icons on the tab - icons help make it recognisable as payment option.
- Currently lots of card icons, do we need them all on the tab?
   - Could use "generic" card icon so we don't need to have an icon for everything, e.g. show `generic visa mastercard` (focus on popular brands in tab).
   - In future may want to allow merchant to control which card icons are displayed (e.g. different brands are dominant in different regions/industries).
- Needs to work well on mobile.
- Need to explore what happens when there are many gateways - e.g. 3 may be common. `And more…` tab?
- Do we allow for tabs to wrap / have multiple lines?

And a related consideration: "Branding" the tab.

- The PayPal tab has a logo image in the mockup, should CC tab show that it is powered by Stripe?
- Maybe need some guidance around logos on tabs, so all gateways appear of similar prominence. A big bright logo might confuse users / make it harder to select a gateway.

Some specific suggestions/ideas from slack thread (ref below). 

> So we could potentially try include the title and icons on the tab?

> We could limit the number of displayed card icons, and just say “credit” rather the credit/debit to save on space

> I think it could wrap - but also aware it could look funky alongside PayPal which won’t wrap

> some merchants wanting control over what icons display

p1588823147160600-slack-woo-merchandising

mikejolley:
This PR makes some changes to the Stripe payment method labelling, with some supporting changes.
 
- Adds `align` support to `PaymentMethodIcons`
- Changes Stripe label to include generic CC icon (desktop only)
- Moves full list of payment icons to Stripe content
- Updates desktop and mobile styles

Fixes #2438

### Screenshots

Mobile:
Screenshot 2020-05-12 at 12 55 05

Desktop:
Screenshot 2020-05-12 at 12 55 52

1. With Stripe enabled, view the cart - check icon appearance
2. View checkout, check icon appearance
3. Check mobile and desktop views

### DIFF
--- a/assets/js/base/components/cart-checkout/payment-method-icons/style.scss
+++ b/assets/js/base/components/cart-checkout/payment-method-icons/style.scss
@@ -1,14 +1,45 @@
 .wc-block-cart__payment-method-icons {
 	display: block;
 	text-align: center;
-	margin: $gap 0 0;
+	margin: 0 0 #{$gap - 2px};
 
 	.wc-blocks-payment-method-icon {
 		display: inline-block;
-		margin: 1px 2px;
+		margin: 0 4px 2px;
 		padding: 0;
 		width: auto;
-		height: 26px;
+		height: 24px;
 		vertical-align: middle;
+	}
+
+	&--align-left {
+		text-align: left;
+
+		.wc-blocks-payment-method-icon {
+			margin-left: 0;
+			margin-right: 8px;
+		}
+	}
+
+	&--align-right {
+		text-align: right;
+
+		.wc-blocks-payment-method-icon {
+			margin-right: 0;
+			margin-left: 8px;
+		}
+	}
+
+	&:last-child {
+		margin-bottom: 0;
+	}
+}
+
+.is-mobile,
+.is-small {
+	.wc-block-cart__payment-method-icons {
+		.wc-blocks-payment-method-icon {
+			height: 16px;
+		}
 	}
 }
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -143,7 +143,7 @@ const PaymentMethods = () => {
 							? label
 							: cloneElement( label, {
 									activePaymentMethod,
-									...currentPaymentMethodInterface.current,
+									components: { ...currentPaymentMethodInterface.current.components },
 							  } ),
 					ariaLabel,
 				};
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -142,8 +142,10 @@ const PaymentMethods = () => {
 						typeof label === 'string'
 							? label
 							: cloneElement( label, {
-									activePaymentMethod,
-									components: { ...currentPaymentMethodInterface.current.components },
+									components: {
+										...currentPaymentMethodInterface.current
+											.components,
+									},
 							  } ),
 					ariaLabel,
 				};
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -20,7 +20,6 @@ import {
 	useEditorContext,
 	usePaymentMethodDataContext,
 } from '@woocommerce/base-context';
-import { PaymentMethodIcons } from '@woocommerce/base-components/cart-checkout';
 import CheckboxControl from '@woocommerce/base-components/checkbox-control';
 
 /**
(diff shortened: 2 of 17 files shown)


assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
```diff
@@ -32,18 +34,16 @@ const StripeComponent = ( props ) => {
...
-	const { PaymentMethodIcons } = components;
-
-	if ( ! PaymentMethodIcons || cardIcons.length === 0 ) {
-		return __( 'Credit/Debit Card', 'woo-gutenberg-products-block' );
-	}
-
-	return <PaymentMethodIcons icons={ cardIcons } />;
+const StripeLabel = () => {
+	return (
+		<span className="payment-method-label-with-icon">
```
mikejolley, line 39: We need to style these differently than tabs with just icons, no text. I added some CSS rules for this so there is correct spacing.

assets/js/blocks/cart-checkout/cart/checkout-button/style.scss
```diff
@@ -1,3 +1,8 @@
 .wc-block-cart__submit-button {
 	width: 100%;
+	margin: 0 0 $gap;
```
mikejolley, line 3: This is so there is space between the buttons and the icons, unless there are no icons in which case no margins are needed.

assets/js/base/components/cart-checkout/payment-method-icons/style.scss
```diff
@@ -1,14 +1,45 @@
...
-		margin: 1px 2px;
+		margin: 0 4px 2px;
 		padding: 0;
 		width: auto;
-		height: 26px;
+		height: 24px;
 		vertical-align: middle;
 	}
+
+	&--align-left {
```
mikejolley, line 15: BEM style modifiers FYI

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
```diff
@@ -3,6 +3,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { useEffect, useState } from '@wordpress/element';
+import { Icon, card } from '@woocommerce/icons';
```
nerrad, line 6: The problem with this is `@woocommerce/icons` is just an alias to internal (non published) components. So we'll have to publish the package and/or export it as an external for when stripe is eventually moved to the payment method extension.

Also wcPay won't be able to use this for it's integration.

I wonder if instead, we should expose a `<PaymentMethodLabel/>` component on the `components` prop provided to the label config and that way Payment methods can use that along with a blessed set of icons to select from?

So something _roughly_ like the following could be passed in via the `components` property:

```jsx
import { card, Icon } from '@woocommerce/icons';

const PaymentMethodLabel = ( { iconType = 'card', text } ) => {
  iconWhitelist = { card };
  if ( iconWhitelist[ iconType ] ) {
    return <span className="payment-method-label-with-icon">
			<Icon srcElement={ iconWhitelist[ iconType ] } />
			{ __( 'Credit / Debit Card', 'woo-gutenberg-products-block' ) }
		</span>
  }
  return text;
}
```

mikejolley: Hmm @nerrad this feels like over engineering... Maybe we just use an SVG directly for this until we have suitable public components? Or just expose a common cc icon component only. It's not really clear if other icons would even be useful in this context.

nerrad: Well we _could_ just expose a `CreditCardLabel` component that is very specific to this design and just receives the text content. Then at least any payment methods offering credit cards could just implement this component (with optionally their custom text)?

nerrad: The main thing I'm concerned about here is removing the dependency on the aliased package so we don't have to deal with this when the stripe integration is moved into it's extension.

mikejolley:
Used your suggestion of a label component @nerrad - can you take another peek? I added an icon for check too using the same component:

Block Checkout – Local Testing 2020-05-13 15-06-05

assets/js/base/components/payment-methods/payment-methods.js
```diff
@@ -143,7 +142,8 @@ const PaymentMethods = () => {
 						typeof label === 'string'
 							? label
 							: cloneElement( label, {
-									components: { PaymentMethodIcons },
+									activePaymentMethod,
+									...currentPaymentMethodInterface.current,
```
nerrad, line 146: I'm not sure we want to expose the entire interface to the label component. We should just expose components:

```suggestion
									components: { ...currentPaymentMethodInterface.current.components },
```

PR #2467, opened 2020-05-12
Refactor `usePaymentMethodRegistration` to ensure payment methods are populated once initialization is true

### LINKED ISSUE #2449: NoPaymentMethods shows up before paymentMethods finishes loading
**Describe the bug**
Found this while working on #2371, we had a flash of `<NoPaymentMethods />` before actual methods show up.
The fix was to wait for `isInitialized` to be true, check `paymentMethods` if it’s still 0, then show a no payment notice, doing that still didn't' fix it, this is because `isInitialized` even if `paymentMethods` is still empty.

**To Reproduce**
This is hard to test since it depends on how fast payment method load.

To better see, we should use console.log, in https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b4e63ffb61b785148e01a6d2710b74e189d77825/assets/js/base/components/payment-methods/payment-methods.js#L96 log both `isInitialized` and `Object.keys( currentPaymentMethods.current ).length` you will see that `isInitialized` even if currentPaymentMethod is 0.

mikejolley:
This is a refactor of `usePaymentMethodRegistration` to change how initialization of payment methods occurs.

Before, there were a series of `useEffect` calls which initialized methods, then dispatched those methods once initialization was complete (after all promises resolved). Since `useEffect` runs after renders, there were 1 or 2 renders where `isInitialized` was `true`, but `paymentMethods` was an empty `{}`.

I've changed the logic slightly so `isInitialized` is only set to true at the same time as the call to the `dispatcher`.

Fixes #2449

1. Smoke test payments - ensure methods are displayed and rendered correctly
2. Console log as explained in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2449 - with this PR you should see the list of paymentMethods populated at the same time as isInitialized being true.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -6,7 +6,7 @@ import {
 	getPaymentMethods,
 	getExpressPaymentMethods,
 } from '@woocommerce/blocks-registry';
-import { useState, useEffect, useRef } from '@wordpress/element';
+import { useState, useEffect, useRef, useCallback } from '@wordpress/element';
 import {
 	useEditorContext,
 	useShippingDataContext,
@@ -53,15 +53,6 @@ const usePaymentMethodRegistration = (
 	registeredPaymentMethods
 ) => {
 	const [ isInitialized, setIsInitialized ] = useState( false );
-
-	/**
-	 * @type {Object} initializedMethodsDefault Holds payment methods that have been initialized.
-	 */
-	const [
-		initializedPaymentMethods,
-		setInitializedPaymentMethods,
-	] = useState( {} );
-
 	const { isEditor } = useEditorContext();
 	const { shippingAddress } = useShippingDataContext();
 	const { cartTotals, cartNeedsShipping } = useStoreCart();
@@ -71,22 +62,6 @@ const usePaymentMethodRegistration = (
 		shippingAddress,
 	} );
 
-	/**
-	 * Update initialized payment methods in local state.
-	 *
-	 * @param {Object} paymentMethod Payment method object that has been initialized.
-	 * @param {boolean} enabled True if the method can be used, or false if payment is not possible via this method.
-	 */
-	const setInitializedPaymentMethod = ( paymentMethod, enabled = true ) => {
-		setInitializedPaymentMethods( ( paymentMethods ) => ( {
-			...paymentMethods,
-			[ paymentMethod.name ]: {
-				enabled,
-				paymentMethod,
-			},
-		} ) );
-	};
-
 	useEffect( () => {
 		canPayArgument.current = {
 			cartTotals,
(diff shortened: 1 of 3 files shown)


assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -32,32 +52,38 @@ const usePaymentMethodRegistration = (
...
+	 * @param {boolean} enabled True if the method can be used, or false if payment is not possible via this method.
+	 */
+	const setInitializedPaymentMethod = ( paymentMethod, enabled = true ) => {
 		setInitializedPaymentMethods( ( paymentMethods ) => ( {
 			...paymentMethods,
-			[ paymentMethod.name ]: paymentMethod,
+			[ paymentMethod.name ]: {
+				enabled,
+				paymentMethod,
+			},
```
mikejolley, line 86: I made this change so we can track initialization of methods even is canPay resolves as false. The list of methods is filtered later before dispatching.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -69,80 +95,57 @@ const usePaymentMethodRegistration = (
...
+				.then( ( canPay ) => {
+					if ( canPay.error ) {
+						throw new Error( canPay.error.message );
+					}
+					setInitializedPaymentMethod( current, canPay );
+				} )
+				.catch( handleRegistrationError );
 		}
-	}, [ isInitialized, isEditor ] );
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
mikejolley, line 119: There were comments specifically in place with regards to registeredPaymentMethods not being a dependency. I left this in place.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -69,80 +95,57 @@ const usePaymentMethodRegistration = (
...
-				const newSet = {};
-				for ( const paymentMethodName in registeredPaymentMethods ) {
-					if ( initializedPaymentMethods[ paymentMethodName ] ) {
-						newSet[ paymentMethodName ] =
-							initializedPaymentMethods[ paymentMethodName ];
-					}
+		if (
+			! isInitialized &&
+			Object.keys( initializedPaymentMethods ).length ===
+				Object.keys( registeredPaymentMethods ).length
```
mikejolley, line 130: This replaces the count ref that was there previously. `initializedPaymentMethods` contains the payment methods regardless of the result from the promise, so we can rely on its size to know if we're done or not.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -69,80 +95,57 @@ const usePaymentMethodRegistration = (
...
+		) {
+			const newSet = {};
+			for ( const paymentMethodName in registeredPaymentMethods ) {
+				if (
+					initializedPaymentMethods[ paymentMethodName ] &&
+					initializedPaymentMethods[ paymentMethodName ].enabled
```
mikejolley, line 136: enabled if canPay was true!

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -69,80 +95,57 @@ const usePaymentMethodRegistration = (
...
+						initializedPaymentMethods[
+							paymentMethodName
+						].paymentMethod;
 				}
-				return newSet;
-			};
-			dispatcher( reSortByRegisteredOrder() );
+			}
+			dispatcher( newSet );
+			setIsInitialized( true );
```
mikejolley, line 145: This is the most important change; setIsInitialized is only called once dispatcher has also been called, so there are no renders with stale data.

haszari:
I think this PR broke payment methods with anon/incognito user/not logged in:

At 6ed4ed4a423159101cbb63bb070df049823a9143 (PR merge commit) I see this in incognito:

Screen Shot 2020-05-13 at 11 06 18 AM

Previous commit b0977750ba2d66634756f4588458374640cde762:

Screen Shot 2020-05-13 at 11 07 51 AM

nerrad:
🤔 yup I can reproduce

PR #2469, opened 2020-05-12
Account for middlewares as effectful code

### LINKED ISSUE #2468: X-WC-Store-API-Nonce header missing in production builds
**Describe the bug**
Checkout and Cart are currently broken on production

**To Reproduce**
1. create a production build `npm run build`
2. in Checkout, trigger a server request (change shipping, update info).
3. Block will break with the error `Error: Missing the X-WC-Store-API-Nonce header. This endpoint requires a valid nonce.`

senadir:
In recent change, the code responsible for adding the nonce to future request has been moved and was no longer considered effectful, this fixes it.

Fixes #2468

1. run a production build `npm run build`
2. make sure changing shipping, info, and address in Checkout works without breaking the block.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,8 @@
 		"*.scss",
 		"./assets/js/atomic/blocks/product/**",
 		"./assets/js/filters/**",
-		"./assets/js/settings/blocks/**"
+		"./assets/js/settings/blocks/**",
+		"./assets/js/middleware/**"
 	],
 	"repository": {
 		"type": "git",


senadir:
is it reproducible in master? I didn't notice it

haszari:
> is it reproducible in master? I didn't notice it

Still investigating, don't worry about it - I'll log a follow up if needed.

haszari:
I bet it's this: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2467

That PR was merged after this branch, and I can't reproduce in master.

So .. I'm merging this, we're good :)

PR #2473, opened 2020-05-13
Fix regression for payment methods in incognito mode.

nerrad:
This fixes a regression introduced by #2467 (see @haszari's [comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2467#issuecomment-627647336)).  After investigation, it's more that after simplifying and reducing the number of unnecessary renders, it exposed some incorrect assumptions being made elsewhere for the tab rendering.

While Rua reported it only happens for incognito mode, it's actually unrelated to incognito and more related to whether there are saved payment methods (incognito just never has them so it's more visible there).

## To test

- Ensure you have no saved payment methods (best way to test is using incognito).
- Verify that any configured payment method options appear
- Verify you can complete a purchase successfully.

**Note**: while testing this pull I discovered a new issue that I can't find the source for and is puzzling because when I git checkout commits in the history where I know it worked before, it's still not working 😕 .  The issue is:

- while logged out
- submit a payment with cheque or stripe (happens with either) with all valid fields filled out.
- the server is returning a 400 error code for me with the following:

Image 2020-05-12 at 9 11 35 PM

And the actual error message in the network request is:

```
{"code":"rest_invalid_param","message":"Invalid parameter(s): billing_address","data":{"status":400,"params":{"billing_address":"Invalid email address."}}}
```

There definitely is an email address.

What's odder is while logged in it works (I suspect because it's using a customer that already has an email address server side).

### DIFF
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -80,16 +80,18 @@ const PaymentMethods = () => {
 	}, [ paymentMethods, paymentMethodInterface, activePaymentMethod ] );
 
 	const getRenderedTab = useCallback(
-		() => ( selectedTab ) => {
+		( selectedTab ) => {
 			const paymentMethod = getPaymentMethod(
 				selectedTab,
 				currentPaymentMethods.current,
 				isEditor
 			);
-			const { supports } = currentPaymentMethods.current[
-				activePaymentMethod
-			];
-			return paymentMethod ? (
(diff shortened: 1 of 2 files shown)


haszari:
I can reproduce the missing billing email issue on b0977750ba2d66634756f4588458374640cde762 - master prior to merge of #2467.

Logged a separate issue for clarity: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2474

PR #2477, opened 2020-05-13
Ensure email and phone are persisted through `setBillingData`

### LINKED ISSUE #2474: billing_address.email field missing when order submitted
**Describe the bug**
When logged out, orders fail with a missing email address. 

This issue was discovered by @nerrad when testing https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2473. 

Note that this issue can not be reproduced on current master. The regression from #2467 (fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2473) masks the issue - if no payment methods are available (when logged out), then you can't submit an order (and see the bug).

To reproduce, check out b0977750ba2d66634756f4588458374640cde762 (prior to merge of #2467) or @nerrad fix branch (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2473).

Screen Shot 2020-05-13 at 1 51 51 PM

**To Reproduce**

0. Clone & build b0977750ba2d66634756f4588458374640cde762 or #2473
1. Open incognito browser or log out. Ensure store allows anon orders (Woo Accounts setting).
2. Add stuff to cart, proceed to checkout. Complete all the form fields, especially email.
99. Submit a payment with cheque or stripe (happens with either) with all valid fields filled out.

Expected: `billing_address.email` field is sent to server, purchase success

Actual: fail with error, and `billing_address.email` field is empty in request

mikejolley:
This should resolve 2474 by ensuring email and phone fields are passed to billingData at all times. This is needed because shippingFields do not contain this data.

Fixes #2474

I'm not sure how to test this fully. I was console logging what was persisted to setBillingData and verifying email was included in various cases.

### DIFF
--- a/assets/js/base/context/cart-checkout/billing/index.js
+++ b/assets/js/base/context/cart-checkout/billing/index.js
@@ -29,15 +29,12 @@ export const useBillingDataContext = () => {
 export const BillingDataProvider = ( { children } ) => {
 	const [ billingData, setBillingDataState ] = useState( DEFAULT_STATE );
 
-	const setBillingData = useCallback(
-		( newData ) => {
-			setBillingDataState( ( prevState ) => ( {
-				...prevState,
-				...newData,
-			} ) );
-		},
-		[ setBillingDataState ]
-	);
+	const setBillingData = useCallback( ( newData ) => {
+		setBillingDataState( ( prevState ) => ( {
+			...prevState,
+			...newData,
+		} ) );
+	}, [] );
 
 	/**
 	 * @type {BillingDataContext}
--- a/assets/js/base/hooks/checkout/use-checkout-address.js
+++ b/assets/js/base/hooks/checkout/use-checkout-address.js
@@ -58,9 +58,13 @@ export const useCheckoutAddress = () => {
 			setShippingAddress( shippingFields );
 		}
 
-		const newBillingData = shippingAsBilling
-			? shippingFields
-			: billingFields;
+		// Uses shipping or billing fields depending on shippingAsBilling checkbox, but ensures
+		// billing only fields are also included.
+		const newBillingData = {
+			...( shippingAsBilling ? shippingFields : billingFields ),
+			email: billingFields.email,
+			phone: billingFields.phone,
+		};
 
 		if ( ! isEqual( newBillingData, billingData ) ) {
 			setBillingData( newBillingData );
--- a/assets/js/base/hooks/checkout/use-checkout-address.js
+++ b/assets/js/base/hooks/checkout/use-checkout-address.js
@@ -53,7 +53,6 @@ export const useCheckoutAddress = () => {
 	);
 
 	// Pushes to global state when changes are made locally.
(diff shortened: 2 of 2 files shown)


assets/js/base/context/cart-checkout/billing/index.js
```diff
@@ -29,15 +29,12 @@ export const useBillingDataContext = () => {
...
-			} ) );
-		},
-		[ setBillingDataState ]
-	);
+	const setBillingData = useCallback( ( newData ) => {
+		setBillingDataState( ( prevState ) => ( {
+			...prevState,
+			...newData,
+		} ) );
+	}, [] );
```
mikejolley, line 37: This dependency wasn't needed; useState provides a safe function. https://reactjs.org/docs/hooks-reference.html#usestate

assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -53,19 +53,31 @@ export const useCheckoutAddress = () => {
...
+			...( shippingAsBilling ? shippingFields : billingFields ),
+			email: billingFields.email,
+			phone: billingFields.phone,
+		};
 
 		if ( ! isEqual( newBillingData, billingData ) ) {
 			setBillingData( newBillingData );
 		}
-	}, [ shippingFields, billingFields, shippingAsBilling ] );
+		// eslint-disable-next-line react-hooks/exhaustive-deps
```
mikejolley, line 73: I needed to not include setBillingData and setShippingAddress here. Doing so prevents input.

assets/js/base/context/cart-checkout/billing/index.js
nerrad: Correct 👍 (and linting will not warn for it). But it also does no harm to include it (which is useful for just having a habit of adding dependencies). I'm not opinionated on this though so if folks remove or add the dependency I'm okay with either.

assets/js/base/hooks/checkout/use-checkout-address.js
nerrad: Hmm that smells 🐟  to me. If `setBillingData` and `setShippingAddress` are memoized, then adding them as dependencies should have no effect. If it does, then there's a bug somewhere else.

The good news is I added them back in as dependencies and it tested just fine. When you say "Doing so prevents input", do you mean you are unable to change the value of any inputs? Because if so, I had no problems with that.

mikejolley: Let me test again and get back to you. When I added all suggested dependencies earlier (and when I first worked on this) it preventing text input on billing fields like email.

haszari: I had that issue yesterday too when I was experimenting with this and added all the deps prompted by linter - I couldn't type in any checkout fields. This was in a branch prior to #2467, so I'm guessing there were other fixes upstream that fixed the underlying issue 🎉

Issue #2180, opened 2020-04-09
Fix stripe payment request payment timing out when selecting same shipping address as what's already set in checkout.


nerrad:
This bug affects the Stripe Payment Request Express checkout (GooglePay, ApplePay etc). To reproduce:

- add an item requiring shipping to the cart.
- click the express button to make purchase (from the checkout block).
- select the same shipping address in the payment request window as what is currently set for the checkout.
- Notice the loading indicator for payment request will keep going until a timeout.

The issue here is that if the selected address is the same, then the `onShippingRatesChange` hook won't fire and thus the payment method observer won't ever update the rates as expected.

PR #2482, opened 2020-05-13
Fix all shipping address changes in stripe payment request timing out with no rate update.

nerrad:
So this was a tricky one to debug, but essentially while working on #2180 I discovered that at some point there was a change made that broke _all_ shipping address update in Stripe payment request payment methods. At first I thought it was my branch but I was able to reproduce on master.

To reproduce:

- Pay with chrome pay or ApplePay (most easily reproduced with Chrome Pay).
- Select any shipping address in the payment modal.
- Notice rates don't update and there's usually a timeout error. 

The bug got exposed by #2477 because that hook listens for changes in `shippingAddress` and overrides with `shippingFields`. See here:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a57b865d119993da845651aca9e658236823678f/assets/js/base/hooks/checkout/use-checkout-address.js#L57-L59

Even though the linked lines were not changes in #2477, the dependencies for that effect _were_ so they _now run anytime shippingAddress changes_.

There's actually no problem with the changes made in that pull because the dependencies _should_ be there. It was just a happy accident that what was a bug before enabled express payment methods to work.

However, the issue still remains and I realized that there's really no need for shippingAddress (and the equivalent setter) to be mirrored in local state within this hook. The only need for the hook is to update and cache `billingData`.

So in this pull I removed the shipping fields state and just re-implemented shipping address and the setter directly (aliasing so that the implementation in the checkout form can stay the same).

This fixes the issue with express payment method shipping changes.

## To test

- First verify that reported issue above is changed.

Next **very important** we need to verify that the original issue `useCheckoutAddress` is intended to fix is not regressed. The pull for that issue is #2374 and you can find testing steps in the pull description there.

### DIFF
--- a/assets/js/base/hooks/checkout/use-checkout-address.js
+++ b/assets/js/base/hooks/checkout/use-checkout-address.js
@@ -51,8 +51,8 @@ export const useCheckoutAddress = () => {
 		// billing only fields are also included.
 		const newBillingData = {
 			...( shippingAsBilling ? shippingAddress : billingFields ),
-			email: billingFields.email,
-			phone: billingFields.phone,
+			email: billingFields.email || billingData.email,
+			phone: billingFields.phone || billingData.phone,
 		};
 
 		if ( ! isEqual( newBillingData, billingData ) ) {
--- a/assets/js/base/hooks/checkout/use-checkout-address.js
+++ b/assets/js/base/hooks/checkout/use-checkout-address.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import defaultAddressFields from '@woocommerce/base-components/cart-checkout/address-form/default-address-fields';
-import { useState, useEffect } from '@wordpress/element';
+import { useState, useEffect, useCallback } from '@wordpress/element';
 import {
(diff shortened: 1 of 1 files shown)


nerrad:
Thanks for the review and testing Rua!

> Default saved payment method doesn't work - handle as follow up :)

Just to clarify, that wasn't introduced in this pull, it happened in master too (why I created a separate issue).

> In Firefox, with a "customer" account (e.g. sign up on my account page, then purchase), I didn't see the Use same address for billing checkbox at all! 

😮 Yikes! Does this happen in master too (I'm checking now before merging).

nerrad:
Hmm I couldn't reproduce the firefox issue you noted in either master or this branch 🤔 . I'll go ahead and merge this.

haszari:
> Hmm I couldn't reproduce the firefox issue you noted in either master or this branch 🤔 . I'll go ahead and merge this.

Yep - it's not an issue - that order had only virtual products, no shipping address is needed. This is correct behaviour :)

PR #2484, opened 2020-05-14
Fix shipping rate and address handling in Stripe payment request payment method.

### LINKED ISSUE #2180: Fix stripe payment request payment timing out when selecting same shipping address as what's already set in checkout.

### LINKED ISSUE #2240: Expose shippingRates errors to express payment methods.
Taken from #2203

> Expose the error to express payment methods. Likely could be saved to the shipping context state and surfaced back through the related shipping emitters so that anything hooked into the emitters can react to the errors.

We need to catch and surface errors from `useShippingAddress` to `ShippingDataProvider` and emit them from there so that express payment methods that are listening can see the errors and react accordingly.

nerrad:
Fixes #2240 
Fixes #2180 

You can review the above listed issues for description of the problems being solved. In this pull:

- address an issue where if the same shipping address as what is already set in the checkout is selected during the payment request, then the modal would time out and rates would be unavailable. This is because the checkout would consider the addresses the same when the attempt is made to change the state and not trigger an update request (which in turn would not trigger a rates update). So the fix is to have the payment method integration compare addresses before deciding to send the state update and only do so if necessary. Note: ideally it'd be nice to just pre-select the shipping address on payment request trigger, but that's not doable with the api.
- address an issue where if a payment request method (notably Apple Pay) allows for the submission of an address that does not have rates, we send trigger a fail for that payment method. Currently we don't have control over what error appears for the payment method but it will usually show something like "There are no available rates for this address". We're somewhat limited by what we can do here.
- address an issue where if a payment request method (notably Apple Pay) allows for the submission of an invalid address (state, address or postcode), the shipping context provider detects that, sets the shipping error status to invalid address, and triggers the appropriate shipping fail handle. With Apple Pay the result will be something like this:

Image 2020-05-14 at 9 56 16 AM

- Along with that, this pull adds a new state and setter to the store notices context provider to allow for suppressing displayed notices. With this checkout can suppress notices while the express payment modal is open and then restore them when it is closed. It prevents things like this:

Image 2020-05-14 at 8 29 40 AM

- finally one other things I noticed and fixed in this pull while testing (and debugging) express payment methods is that if you opened the payment request modal and then closed it, then opened it again, the listeners (I called them "handlers" in the code) registered to  `paymentRequest.on` would get _re-registered_ which of course meant when the event fired, the listeners would fire for every time the modal was closed and re-opened! With the fix, handlers are only registered once and are removed from the event when the modal is closed (or unmounted).

Also in this pull I refactored the stripe payment request integration so various logic is split out into more discrete files. This makes it easier to debug and improves readability. 

- `use-checkout-subscriptions.js`: this is mirrored off of the same thing I did for the Stripe CC integration and contains all the subscriptions to checkout events.
- `use-initialization.js`: This contains all the payment request initialization and setting of event handlers on Stripe PaymentRequest API events.
- `use-event-handlers.js`: This contains all the logic for managing the event handler cache for the handlers registered to the Stripe PaymentRequest API events.

Also in this pull is some fixing of dependencies in the payment request integration and in the shipping data context.

## To test

This impacts:

- deriving shipping rates from the server
- shipping address fields (and billing data fields)
- purchasing "regular" payment methods (Stripe CC, cheque, paypal) - but you only have to verify one of those. Basically your just verifying that making a purchase with either of these has the shipping data and totals appropriately applied.
- purchasing via express payment methods (ChromePay or ApplePay). To assist with testing I've updated my [ephemeral site with the production build of this branch](https://ephemeral-nerrad-20200302.atomicsites.blog). ApplePay is enabled on this site in test mode, so you can visit in safari and you should see the ApplePay button if you are in a supported region.

Special instructions:

- Both ChromePay and ApplePay need tested.
- Both logged in and logged out (or incognito) contexts must be tested.
- Besides checking that expected behaviour for normal purchases isn't affected, you also need to test behaviour when shipping rates aren't returned. I don't know if you can configure to have only rates for a specific country etc in Woo but what I did was just changed [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/85f8f1b11890e0dc0544c8bf2fdc3496b8afcc1d/src/StoreApi/Schemas/CartSchema.php#L277) to return an empty array for rates (after opening the ChromePay modal and before selecting an address). You should see the modal return a message about there being no rates for the selected address.  
- You can also test invalid state entered using ApplePay which will test the error handling. When you do this, also try cancelling the modal in the error condition, and you should see the error notice appear in checkout after the modal is closed.
- Also need to verify that when logged in (and only logged in, doesn't apply for non-logged in or incognito), and there's already a shipping address filled in. If you select the same shipping address in the express payment method, that it selects immediately and shows you the rates for that address (without any wait).

**Note:** Keep in mind that for Payment Request purchases, it is expected that when you select a new address in the payment request modal, it's only sending Country, state, Postal Code and City to checkout (even if the address has street). So it may appear it's broken if you notice changes in the underlying checkout form as things react to the express payment, but it's not. That's a privacy feature of the payment request api to not provide the full address on initial selection. When the `Pay` button is clicked in the modal, the API exposes the full billing info and shipping data - so the integration is able to send that back to checkout. The thing to watch for is that you can complete purchase in varying conditions and the final order has the correct data.

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
@@ -33,7 +33,7 @@ import {
  *                                                                       registering observers to
  *                                                                       events.
  * @param {Object}                 props.paymentRequestEventHandlers     Cached handlers registered
- * \                                                                     for paymentRequest events.
+ *                                                                       for paymentRequest events.
  * @param {function(string):void}  props.clearPaymentRequestEventHandler Clears the cached payment
  *                                                                       request event handler.
  * @param {BillingDataProps}       props.billing
--- a/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
@@ -93,14 +93,12 @@ export const useCheckoutSubscriptions = ( {
 
 	// subscribe to events.
 	useEffect( () => {
-		const onShippingRatesEvent = ( forSuccess = true ) => (
-			shippingRates
-		) => {
+		const onShippingRatesEvent = ( shippingRates ) => {
 			const handlers = eventHandlers.current;
 			const billingData = currentBilling.current;
 			if ( handlers.shippingAddressChange && isProcessing ) {
 				handlers.shippingAddressChange.updateWith( {
-					status: forSuccess ? 'success' : 'fail',
+					status: 'success',
 					shippingOptions: normalizeShippingOptions( shippingRates ),
 					total: getTotalPaymentItem( billingData.cartTotal ),
 					displayItems: normalizeLineItems(
@@ -109,6 +107,18 @@ export const useCheckoutSubscriptions = ( {
 				} );
 				clearPaymentRequestEventHandler( 'shippingAddressChange' );
 			}
+		};
+		const onShippingRatesEventFail = ( currentErrorStatus ) => {
+			const handlers = eventHandlers.current;
+			if ( handlers.shippingAddressChange && isProcessing ) {
+				handlers.shippingAddressChange.updateWith( {
+					status: currentErrorStatus.hasInvalidAddress
+						? 'invalid_shipping_address'
+						: 'fail',
+					shippingOptions: [],
+				} );
+			}
+			clearPaymentRequestEventHandler( 'shippingAddressChange' );
 		};
 		const onShippingSelectedRate = ( forSuccess = true ) => () => {
 			const handlers = eventHandlers.current;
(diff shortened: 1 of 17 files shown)


assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
```diff
@@ -0,0 +1,241 @@
...
+ *
+ * @param {boolean}                props.canMakePayment                  Whether the payment request
+ *                                                                       can make payment or not.
+ * @param {boolean}                props.isProcessing                    Whether the express payment
+ *                                                                       method is processing or not.
+ * @param {EventRegistrationProps} props.eventRegistration               Various functions for
+ *                                                                       registering observers to
+ *                                                                       events.
+ * @param {Object}                 props.paymentRequestEventHandlers     Cached handlers registered
+ * \                                                                     for paymentRequest events.
```
haszari, line 36: typo \ :)

PR #2485, opened 2020-05-14
Only set to pristine when checkout is idle, not when idle status changes

### LINKED ISSUE #2483: Fix default saved payment method selection not working in specific circumstances.
Spotted this bug while doing more testing. To reproduce.

- Test while logged in (because it affects saved payment methods).
- Setup saved payment methods so you have multiple saved payment methods from stripe.
- Start a checkout but don't select any other payment methods and just leave the default checked.
- When you submit the checkout, you'll get an error. 

If you look at the request, you'll see that the package has `false` for the saved payment method.

mikejolley:
This appears to be have been caused by an overzealous useEffect hook. Rather than firing when the checkout is idle, it was firing on all changes to `checkoutIsIdle` and `currentStatus.isSuccessful`.

From what I can tell, saved payment methods are always `currentStatus.isSuccessful`, meaning this would run on first mount and reset saved methods to pristine state.

Fixes #2483

See #2483

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -268,7 +268,7 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 	// When checkout is returned to idle, set payment status to pristine
 	// but only if payment status is already not finished.
 	useEffect( () => {
-		if ( ! currentStatus.isSuccessful ) {
+		if ( checkoutIsIdle && ! currentStatus.isSuccessful ) {
 			dispatch( statusOnly( PRISTINE ) );
 		}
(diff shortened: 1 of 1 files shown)


assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -268,7 +268,7 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 	// When checkout is returned to idle, set payment status to pristine
 	// but only if payment status is already not finished.
 	useEffect( () => {
-		if ( ! currentStatus.isSuccessful ) {
```
nerrad, line 271: oooo yup, good catch, I'm pretty sure I had it that way in #2425 but I must have resolved a conflict wrong when I was rebasing (dependency seems to support that :) )

PR #2495, opened 2020-05-15
Fix Editor performance issues with checkout block.

### LINKED ISSUE #2493: Checkout block: 'Maximum update depth exceeded' in editor
**To Reproduce**
1. `git co fix/2491-set-is-suppressed-editor` so you don't encounter #2491.
2. Add a _Checkout_ block to a post.
3. Notice performance of the editor is degraded.
4. Notice an error appears in the console: `Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.`.

nerrad:
Fixes #2493

See #2493 for a description of the bug. Tracing the stack trace I derived that the cause of the react error was an excessive state setting of billing data happening here:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/9516ad289be4e5510b38c3307f9fb0ed0f914b65/assets/js/base/hooks/checkout/use-checkout-address.js#L58-L60

Verified by commenting out that setter.

Since the setter is only triggered when there is not deep equal quality between existing billing data and new billing data, I traced the cause of _that_ to `shippingAddress` always having less properties than `billingData`. Discovered that, `shippingAddress` is _initialized_ with less properties here:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/9516ad289be4e5510b38c3307f9fb0ed0f914b65/assets/js/base/hooks/cart/use-store-cart.js#L79-L84

But only in the **editor**, which is why it surfaced. There was a commit where I removed unnecessary `shippingFields` state in the `useCheckoutAddress` hook that I think was originally masking the problem because the default for `shippingFields` had all the properties.

## Testing

- First verify this fixes the performance issue in the editor (should be no errors when loading the block in the editor).
- Next verify cart loads fine on frontend and backend.
- Next verify that checkout works as expected (in particular the behaviour for the shipping fields and billing fields when switching the "use same address for billing" checkbox.
- Verify that you can make a purchase using cc on the frontend.
- Verify that you can make a purchase using payment request button (either chrome pay or apple pay). 

I tested all of the above and it worked well for me.


assets/js/payment-method-extensions/payment-methods/stripe/stripe-utils/type-defs.js
```diff
@@ -294,6 +294,8 @@
  *                                                          inline cc
  *                                                          form or separate inputs.
  * @property {{[k:string]:CreditCardIcon}} icons            Contains supported cc icons.
+ * @property {boolean}                     allowSavedCards  Used to indicate whether saved cards
+ *                                                          can be used.
```
nerrad, line 298: Not related to this pull but I discovered it when in the process of debugging so added the missing type-def property declaration.

PR #2497, opened 2020-05-15
Introduce Variation Picker Component

### LINKED ISSUE #2398: [Atomic Block] Variation Picker Component
A new variation picker component is needed to support variable product add to cart forms. These pickers display available variation attributes, and hide/show options based on validity:

This component also needs to support default selections so it can be wired up to block settings:

**API Requirements**

Currently the products API does give us child variation IDs, but doesn't provide attribute data (from the parent). Whilst this could be determined by looping over child variation IDs, that won't be efficient enough for this use case.

We'll need to surface from the API at a minimum:

- A list of product attributes (from the parent) which will form the values in the select boxes
- A collection of variations with _their_ attributes, and data such as ID, price, etc.
- (possibly) A way to match against a variations in the collection based on attributes. I think some collections support this already, this just needs checking with variations only.

Woo core has 2 approaches to these forms. 

1. For variations with few attributes and SKUs, the matching is all done client side using child variation attribute data to find a valid match.
2. For variations with many attributes and SKUs, all attributes are chosen, and then a request is made via AJAX to return a matching variation. This is to prevent performance issues from loading too much variation data at once.

What should we do here? I think there would be value of approaching this in a single way, but I am mindful of 'variation abuse' and performance issues if there are hundreds or even thousands of variation SKUs under a single product. cc @nerrad 

**UX**

The select boxes have designs, but nothing interaction wise has had design. Whilst we can mirror core, this probably needs design review before this issue can be resolved. 

- How we handle invalid combinations of attributes. Are the invalid options greyed out of hidden? Is there a way to reset the form once selections have been made? See core variations with multiple attributes for reference.
- After selecting attributes (and a variation ID is matched), what is shown? Core shows variation price and description/sku. 
- Similarly, if a match is not found, what is displayed? Core I believe shows an error stating the combination is not available.

Please also note, the 'greying out' or 'disabling' of invalid combinations in core is also impacted by the number of variations a product has. See the API section above; finding matching variations on the client side allows for this, whearas the ajax requests do not.

cc @garymurray

mikejolley:
This is the variation picker component (#2398) for the single products block.

This adds new API data (variations and attributes) and finds matches on the client side via a new Variation Attribute Picker Form. This has been added to the Add to Cart Form Atomic Block. It will show attributes and match a variation ID, but is not yet hooked up to the form.

Known follow ups:

- Hooking up to the cart form
- Supporting default attributes

Closes #2398

### How to test

1. Add a single product block for a variable product
2. Confirm the picker is shown in admin and on the frontend, and attributes are selectable on the frontend form.
3. Selecting all attributes should match a variation (this will be shown below the form)

### Screenshots

Single Product – local wordpress test 2020-06-17 17-34-49


assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
```diff
@@ -0,0 +1,113 @@
...
+					selected={ selectedAttributes[ attributeName ] }
+					onChange={ ( selected ) => {
+						setSelectedAttributes( {
+							...selectedAttributes,
+							[ attributeName ]: selected,
+						} );
+					} }
+				/>
+			) ) }
+			<p>Matched variation ID: { variationId }</p>
```
mikejolley, line 108: This is temporary for testing purposes.

assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/attribute-select-control.js
```diff
@@ -0,0 +1,36 @@
...
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { decodeEntities } from '@wordpress/html-entities';
+import { SelectControl } from '@wordpress/components';
+
+// Default option for select boxes.
+const selectAnOption = {
+	value: '',
+	label: __( 'Select an option', 'woo-gutenberg-products-block' ),
```
Aljullu, line 11: Should this option be [disabled](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/select-control#options) so it can't be manually selected by customers?

assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
```diff
@@ -0,0 +1,113 @@
...
+	}, [
+		selectedAttributes,
+		variationAttributes,
+		attributeNames,
+		hasSelectedAllAttributes,
+	] );
+
+	// @todo Hook up Variation Picker with Cart Form.
+	return (
+		<div className="wc-block-components-product-add-to-cart-attribute-picker">
```
Aljullu, line 93: Using BEM, this class name should be `wc-block-components-product-add-to-cart__attribute-picker`, right?

assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/attribute-select-control.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+const AttributeSelectControl = ( {
+	attributeName,
+	options = [],
+	selected = '',
+	onChange = () => {},
+} ) => {
+	return (
+		<SelectControl
+			className="wc-block-components-product-add-to-cart-attribute-picker-select"
```
Aljullu, line 27: And this one: `wc-block-components-product-add-to-cart__attribute-picker-select`.

assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/utils.js
```diff
@@ -0,0 +1,112 @@
...
+ */
+export const getVariationsMatchingSelectedAttributes = ( {
+	selectedAttributes,
+	variationAttributes,
+	attributeNames,
+	exclude = '',
+} ) => {
+	return Object.keys( variationAttributes ).filter( ( variationId ) =>
+		attributeNames
+			.filter( ( attributeName ) => attributeName !== exclude )
```
Aljullu, line 93: Instead of filtering the attribute names here, why not just passing to this function the `attributeNames` array already filtered? This would avoid looping through the array inside another loop (`filter` inside a `filter`) and would simplify the function signature with one less argument.

src/StoreApi/Schemas/ProductSchema.php
```diff
@@ -424,6 +519,146 @@ protected function get_product_quantity_limit( \WC_Product $product ) {
...
+	 * @param \WC_Product $product Product instance.
+	 * @return array
+	 */
+	protected function get_attributes( \WC_Product $product ) {
+		$attributes = array_filter( $product->get_attributes(), [ $this, 'filter_valid_attribute' ] );
+		$return     = [];
+
+		foreach ( $attributes as $attribute_slug => $attribute ) {
+			// Only visible and variation attributes will be exposed by this API.
+			if ( ! $attribute->get_visible() && ! $attribute->get_variation() ) {
```
Aljullu, line 621: Should this be an _or_ instead of an _and_? If I'm understanding this clause correctly, currently it only skips attributes that are not visible **and** are not variation attributes. But we would like to skip all attribute that meet one of these two conditions, right?

assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/attribute-select-control.js
mikejolley: User should be able to remove selections because other selects are narrowed down by available options.

mikejolley:
> This is looking great! I noticed that when trying to add to the Cart, I get error 400 and { code: 'woocommerce_rest_variation_id_from_variation_data', message: 'No matching variation found.' }. Is that expected?

@Aljullu Yes, the selected variation and attributes is not wired up to the form yet. There is a follow up.

assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
mikejolley: Actually no, I'm not sure. This is a separate component right? It doesn't need to be nested. The select one below should change though.

assets/js/atomic/blocks/product/add-to-cart/style.scss
```diff
@@ -4,12 +4,31 @@
...
+					appearance: menulist;
+					min-width: 60%;
+					border: em($gap-smallest) solid $white;
+					box-shadow: 0 0 0 1px #ccc;
+					margin: 1px;
+					border-radius: 3px;
+					padding: 0;
+					background: $white;
+					color: #43454b;
+				}
```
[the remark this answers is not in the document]
mikejolley: This is why I wanted to use our checkout selects ;( Design seem to be using native select, I just bumped up the padding. Ill remove the styles and worry about it later.

mikejolley:
@Aljullu made some changes. The follow ups are todods and can be seen in comments above.

assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/style.scss
```diff
@@ -0,0 +1,23 @@
...
+		margin: 0 0 em($gap-small) 0;
+
+		select {
+			@include font-size(regular);
+			font-family: inherit;
+			min-width: 60%;
+			min-height: 1.75em;
+			border: 1px solid $input-border-gray;
+			border-radius: 4px;
+			margin: 0;
```
Aljullu, line 20: I'm still not sure about styling the native select. :thinking:  Setting a custom border and border-radius makes the select look completely different from how it would look natively. And it makes it stand out compared to the select from the All Products and Reviews blocks.

| | Firefox | Chrome | Web (Webkit engine) |
| --- | --- | --- | --- |
| Product block | imatge | imatge | imatge |
| All Products block | imatge | imatge | imatge |

Also, it was not clear to me why we need to set a font size here? Can't we just inherit the one from the theme? Looking at the WC Core product page, it looks like the only style properties the select has are `max-width` and `vertical-align`. Could we follow that approach here too and only write CSS for the layout, but not the specific styles?

PR #2498, opened 2020-05-15
Remove some unique font-size

Aljullu:
This PR removes all instances of `11px` or `13px` font-sizes as a first step to reduce the number of different font-sizes used in our blocks.

* fb9c78adbb89dcfc2b6db25826ef3807b00a0f0e: removes a couple of them that were already inherited from the placeholder styles, so there was no reason to declare them again.
* 9c92227f4d0e658fa25fb9b878ae562f099fadf0: removes the font-sizes of the _Product Search_ and _Product Categories_ buttons. They had no text inside them (only a SVG), so removing it doesn't make any difference.
imatge
* 142a0fa7e30a8dccea785c410d258ac4e00611b1: increases the font-size of the _Order summary_ quantity circles from 11px to 12px. Tagging @LevinMedia for feedback/awareness: in the designs that font was 11px but making it 12px aligns it with several other texts in the _Checkout_ block and the visual difference is minimal (I could barely see any difference in the screenshots below).

_Before:_
imatge

_After:_
imatge

1. Add the _Product Search_ and _Product Categories_ (select the _Dropdown_ display) and verify there are no regressions with the button.
2. Make sure there are no regression either with the _Checkout_ block placeholders when there are no shipping or payment methods.

### DIFF
--- a/assets/js/base/components/payment-methods/no-payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/no-payment-methods/style.scss
@@ -14,7 +14,6 @@
 		}
 
 		.wc-block-checkout__no-payment-methods-placeholder-description {
-			@include font-size(13px);
 			display: block;
 			margin: 0.25em 0 1em 0;
 		}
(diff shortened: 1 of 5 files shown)


LevinMedia:
That works for me, @Aljullu thanks! 👍

PR #2502, opened 2020-05-18
Add notices docs

Aljullu:
Add docs to the frontend notices system.

### DIFF
--- a/docs/block-client-apis/notices.md
+++ b/docs/block-client-apis/notices.md
@@ -143,7 +143,7 @@ Whether notices are suppressed. If true, it will hide the notices from the front
 
 ### Statuses
 
-All notices must have one of the following status: `default`, `error`, `success`, `info`, `warning`.
+All notices must have one of the following statuses: `default`, `error`, `success`, `info`, `warning`.
 
 ## Example usage
 
--- a/docs/block-client-apis/notices.md
+++ b/docs/block-client-apis/notices.md
@@ -1,0 +1,172 @@
+# useStoreNotices()
+
+The `useStoreNotices()` hook allows reading and manipulating notices in the frontend.
+
+## API
+
+### `addDefaultNotice( text = '', noticeProps = {} )`
+
+Create a new notice with `default` status. If the context is not specified in the `options` object, the current context is used.
+
+| Argument      | Type   | Description                                                                                                                                                                              |
+| ------------- | ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `text`        | string | Text to be displayed in the notice.                                                                                                                                                      |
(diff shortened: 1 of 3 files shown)


docs/block-client-apis/notices.md
```diff
@@ -0,0 +1,174 @@
...
+Create a new notice with `default` status. If the context is not specified in the `options` object, the current context is used.
+
+| Argument      | Type   | Description                                                                                                                                                                              |
+| ------------- | ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `text`        | string | Text to be displayed in the notice.                                                                                                                                                      |
+| `noticeProps` | Object | Object with the notice props. Refer to the [Gutenberg docs](https://github.com/WordPress/gutenberg/blob/master/packages/notices/src/store/actions.js#L46) to know the available options. |
+
+#### `addErrorNotice( text = '', noticeProps = {} )`
+
+Create a new error notice. If the context is not specified in the `options` object, the current context is used.
```
senadir, line 20: > If the context is not specified in the `options` object, the current context is used.

Preferably this should say `noticeProps` object, hopefully with an example, and could possibly only mentioned once instead of each section?

docs/block-client-apis/notices.md
```diff
@@ -0,0 +1,174 @@
...
+#### `removeNotice( id, ctx )`
+
+Remove an existing notice. If the context is not specified, the current context is used.
+
+| Argument | Type   | Description                                                                                                 |
+| -------- | ------ | ----------------------------------------------------------------------------------------------------------- |
+| `id`     | string | Id of the notice to remove.                                                                                 |
+| `ctx`    | string | Context where the notice to remove is stored. If the context is not specified, the current context is used. |
+
+#### `removeNotices( status = null )`
```
senadir, line 84: I know this is just a doc and this is not related but I feel like the API is all over the place when it comes to context, sometimes in an object, sometimes a string param and sometimes it's implied and you can't even set it.

docs/block-client-apis/notices.md
Aljullu: Yeah, good point. Changed in: fea3e386a63271a466401d8f2bd86d7c6905aee2.

docs/block-client-apis/notices.md
Aljullu: Yup, in general it uses the current context so it's implied, but some specific methods allow specifying a different context. That's useful for payment method processing, for example, so they can add errors either to the express payment methods or the regular ones. I think we could iterate to make the API more clear after the first release.

nerrad:
As an aside, this might be a good candidate for a `@woocommerce/notices` package (to move to the wc-admin repository at some point)?

PR #2504, opened 2020-05-18
Todo action from woocommerce/automations test.

nerrad:
I've created a new todos automation (see [woocommerce/automations](https://github.com/woocommerce/automations)) that should help us with ensuring follow-ups for todos are created when added in pulls.

This will not apply retroactively.

If the test passes then I'll close this pull instead of merging, the workflow is already setup in master.

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -28,6 +28,8 @@ import './filters/get-block-attributes';
 //
 // Let's see if this one works okay.
 
+// @todo Another test.
+
 setCategories( [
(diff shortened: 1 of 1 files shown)


nerrad:
There's still an issue with the `cc @` done in the todo comment. It's actually behaving as expected but what's not expected is that the payload value that is being used for that would be the _organization_ user instead of the commit author. 

I know what to do for a fix, I'll update this todo when it's implemented.

nerrad:
Alright, all looks good, going to close this and delete the branch.

PR #2507, opened 2020-05-19
Fix decimal ratings not correctly displayed

Aljullu:
I split this off from #2428 so it's not blocked by the other changes in that PR.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Make sure you have at least one product with an average rating with decimals, ie 4.5.
2. View the _All Products_ block or a legacy product grid block in the frontend.
3. Verify the number of stars is correct.

### Changelog

> Fix star ratings in product grid blocks so it correctly displays non-integer ratings.

### DIFF
--- a/assets/js/atomic/components/product/rating/index.js
+++ b/assets/js/atomic/components/product/rating/index.js
@@ -19,7 +19,7 @@ const ProductRating = ( { className, product } ) => {
 	};
 
 	const ratingText = sprintf(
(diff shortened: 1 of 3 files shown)


Aljullu:
I missed that, good catch @nerrad! Fixed in 42577c3012981ed6603a7b5ec39874703747184b. I updated the string of the _Reviews_ blocks as well, even though their ratings will be integers, this way they share the same translate string.

Issue #2418, opened 2020-05-06
Styles found in Woo Core need consolidating into Blocks plugin


mikejolley:
There are a number of theme specific styles for blocks in core which ideally should be moved to blocks so both plugins are not fighting for specificity. 

Example: 
https://github.com/woocommerce/woocommerce/blob/master/assets/css/twenty-twenty.scss#L2103-L2140

From a quick scan, it only appears to affect Twenty Twenty. These particular styles are causing issues on the All Products Block, so they will need updating when merged back to blocks:

![Screenshot 2020-05-06 at 16 01 27](https://user-images.githubusercontent.com/90977/81194761-f1b0c300-8fb4-11ea-89e7-94c8803ee8b3.png)

Aljullu:
I created a PR in Core (https://github.com/woocommerce/woocommerce/pull/26516) that should fix the glitches caused by these styles. So this issue can be exclusively about moving those styles to WC Blocks without any changes.

From the file @mikejolley [linked](https://github.com/woocommerce/woocommerce/blob/master/assets/css/twenty-twenty.scss), we would need to move all selectors that start with `.wc-block`.

In `product-list/style.scss` we also have some [styles targeting specific 20xx themes](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/components/product-list/style.scss#L353-L376).

As we build new blocks and new WP core themes are released, theme specific styles might keep increasing. So I think we should consider building a system to only load theme-specific styles if that theme is enabled, like they have in core. Does that sound reasonable @mikejolley?

mikejolley:
@Aljullu Possibly yes, if the volume of styles causes an increase in file size we're not happy with.

Aljullu:
After #2513 was merged, I think we can close this issue. If in the future we see a system to dynamically import CSS for specific themes is needed, we can open an issue for that.

PR #2513, opened 2020-05-20
Import TwentyTwenty block styles from Core

Aljullu:
Part of #2418.

1. With WooCommerce Core in `master`, verify there are no visual regressions.
2. With WooCommerce Core in https://github.com/woocommerce/woocommerce/pull/26516 branch (`fix/twenty-twenty-block-styles`), verify:
  2.1 In 'legacy' product grid blocks the `Sale!` badge is aligned to the top right.
  2.2 In 'legacy' product grid blocks the link underline is black.
  2.3 In _All Products_ block, prices are displayed with the correct font-family and font-sizes look the same as 'legacy' product grid blocks.
  2.4 In _All Products_ block, when there is a product with a long description, the layout is not broken.

### Screenshots
_Hand-Picked Products:_
| Before | After |
|---|---|
| imatge | imatge |

_All Products:_
| Before | After |
|---|---|
| imatge | imatge |

### Changelog

> Improvements to the display of product grid blocks in Twenty Twenty.

### DIFF
--- a/assets/css/editor.scss
+++ b/assets/css/editor.scss
@@ -20,10 +20,6 @@
 			list-style: none;
 			margin: 0 (-$gap/2) $gap;
 
-			.wp-block-button__link {
-				color: inherit;
-			}
-
 			.wc-block-grid__product {
 				margin: 0 0 $gap-large 0;
 			}
--- a/assets/js/base/components/product-list/style.scss
+++ b/assets/js/base/components/product-list/style.scss
@@ -217,7 +217,6 @@
 	}
 }
 
-.editor-styles-wrapper .wc-block-grid__products .wc-block-grid__product .wc-block-grid__product-onsale,
 .wc-block-grid__product-onsale {
 	border: 1px solid #43454b;
 	color: #43454b;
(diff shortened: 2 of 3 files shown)


Aljullu:
> On the frontend, in the handpicked product block (chrome) the spacing around the sale badge doesn't match your screenshot. it is flush to the top, but has spacing on the right. The all products block is correct.

Ah, good catch. That happened with WooCommerce Core in `master`: the 'old' styles were competing with the new ones. In 38276913c1c6e77a19a57990b513d87f6b31307f, I increased the selector specificity so our styles override the styles from Core.

imatge

Ideally we will be able to reduce that selector specificity once the original styles are removed from WC Core.

> In the editor, this styling is completely missing and the sale badge is white and inlined like with Storefront.

Yeah... that was not introduced in this PR but good point.

The styles are not being added because we lack the `twentytwenty` class in the editor. I think the more future-proof way to handle that would be to load those styles dynamically depending on the selected theme, this way we don't need to scope them inside the `.twentytwenty` class. Does that sound like a good plan or you had something else in mind @mikejolley?

mikejolley:
@Aljullu We could try using `admin_body_class`?

Aljullu:
Done! Willing to take another look @mikejolley?

**Known issues:**
- Font size is different in admin and frontend for the _All Products_ block. I'm pretty sure that will be fixed once we migrate to `em` instead of `rem`.
- _All Products_ links don't have the correct color in the editor. That seems to affect links from other blocks as well (try adding a paragraph block with a link) so I think it's an issue with TwentyTwenty/Gutenberg.

PR #2514, opened 2020-05-20
Exclude three more files from the build ZIP

Aljullu:
When releasing RC1 I noticed we were including three unnecessary files in the generated ZIP: `blocks.ini`, `docker-compose.yml` and `tsconfig.json`. This PR excludes them in the build scripts.

1. Run `npm run package-plugin`.
2. Verify the generated zip files doesn't include those files.

### DIFF
--- a/bin/build-plugin-zip.sh
+++ b/bin/build-plugin-zip.sh
@@ -83,6 +83,9 @@ copy_dest_files() {
 		--exclude=package.json \
 		--exclude=package-lock.json \
 		--exclude=none \
+		--exclude=blocks.ini \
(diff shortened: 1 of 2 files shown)


nerrad:
> Do we need experimental features in the feature plugin? I would think perhaps we do?

This will depend on the changes @senadir is making. Any comments here Nadir?

senadir:
block.ini should not be in the final zip, it is git ignored and is only meant for local dev

> This will depend on the changes @senadir is making. Any comments here Nadir?

The new system will still need it to communicate between dev, but it won't live in the final zip, so we can safely skip it.

PR #2517, opened 2020-05-20
Align Product Categories List select and button

### LINKED ISSUE #2515: Product Categories block is not vertically aligned
**To Reproduce**
1. Add a _Product Categories List_ block and select the Dropdown display style.
2. View it in the frontend.
3. Notice the select and the button are not properly aligned.

**Screenshots**
imatge

Aljullu:
Fixes #2515.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Add a _Product Categories List_ block and select the Dropdown display style.
2. View it in the frontend.
3. Verify the select and the button are aligned.

### Changelog

> Fixed alignment between select and button in Product Cateogories List block with Dropdown style.

### DIFF
--- a/assets/js/blocks/product-categories/style.scss
+++ b/assets/js/blocks/product-categories/style.scss
@@ -67,7 +67,7 @@
 	border: none;
 	cursor: pointer;
 	background: none;
-	padding: rem(8, 13);
+	padding: 0 rem(8, 13);
 	color: #555d66;
 	position: relative;
 	overflow: hidden;


Aljullu:
> Minor nitpick - the arrow icon seems off-centre (too high), might look slightly better closer to the bottom, maybe a little smaller. But that's not a blocker at all :)

I can't reproduce it, for me it looks vertically centered, might it be a visual effect because on hover there is a bottom shadow? :thinking: 

About making it smaller, currently it has the same size as the _Product Search_ button. I'm not sure about changing its size without changing _Product Search_ button as well.

I will merge this PR for now and we can open follow-ups if we want to make more changes to the button.

PR #2530, opened 2020-05-21
Checkout steps improvements

Aljullu:
Small fixes to the Checkout steps:
- Don't render step heading content if it's empty (eba6f611ccf3fb88df0f9a7ed3f8ca4dc4322cdb).
- Make checkout step headings `<h2>` (2ae2a78250ac7364ebb94987da130c63f88667af). Even though they are `aria-hidden`, I still think it's better to follow the proper heading hierarchy.
- Add empty alt value to form step counter (cd13fe2c6fafc372ce03097d78c14fe2fca506da). The number was a bit confusing when it was read by screen readers. Notice alt text in generated content is [not supported by all browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/content#Browser_compatibility) so we need to keep the `content` declaration with no alt text as a fallback.

There is not much to test. Basically go to the _Checkout_ block and verify there are no regressions with the steps.

### DIFF
--- a/assets/js/base/components/cart-checkout/form-step/index.js
+++ b/assets/js/base/components/cart-checkout/form-step/index.js
@@ -14,9 +14,11 @@ const StepHeading = ( { title, stepHeadingContent } ) => (
 		<h4 aria-hidden="true" className="wc-block-checkout-step__title">
 			{ title }
 		</h4>
-		<span className="wc-block-checkout-step__heading-content">
-			{ stepHeadingContent }
-		</span>
+		{ !! stepHeadingContent && (
+			<span className="wc-block-checkout-step__heading-content">
+				{ stepHeadingContent }
+			</span>
+		) }
 	</div>
 );
 
--- a/assets/js/base/components/cart-checkout/form-step/index.js
+++ b/assets/js/base/components/cart-checkout/form-step/index.js
@@ -11,9 +11,9 @@ import './style.scss';
 
 const StepHeading = ( { title, stepHeadingContent } ) => (
 	<div className="wc-block-checkout-step__heading">
-		<h4 aria-hidden="true" className="wc-block-checkout-step__title">
+		<h2 aria-hidden="true" className="wc-block-checkout-step__title">
 			{ title }
-		</h4>
+		</h2>
 		{ !! stepHeadingContent && (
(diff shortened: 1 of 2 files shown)


assets/js/base/components/cart-checkout/form-step/index.js
```diff
@@ -11,12 +11,14 @@ import './style.scss';
 
 const StepHeading = ( { title, stepHeadingContent } ) => (
 	<div className="wc-block-checkout-step__heading">
-		<h4 aria-hidden="true" className="wc-block-checkout-step__title">
+		<h2 aria-hidden="true" className="wc-block-checkout-step__title">
```
nerrad, line 14: Should these use the new `Title` component introduced in #2486 ?

Aljullu: Related to: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2486#discussion_r429275237

I think `Title` should be an opinionated block only used when we want it to appear with those specific styles. For headings with other styles, I don't think there is any benefit on using `Title` over native `h1`, `h2`... tags.

nerrad: The comment I [left here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2486#discussion_r429296378), applies here as well. 

So from your comment, I think you are suggesting here that you don't think the step headings should be linked to the style tree inherited from `Title`?

Aljullu: Actually, after seeing the new designs of the _Cart_ and _Checkout_ blocks, it looks like the _Checkout_ step titles and the _Cart_ title will have the same styles, so contradicting my previous comment, I do think this is a good use-case to use `Title` here.

Will update this PR accordingly after #2486 is merged.

Aljullu: I have been experimenting with this and I'm not sure the same styles will work, I will discuss this with @LevinMedia, but I would like to decouple it from this PR.

I created a follow-up issue to investigate if the `Title` component can be used in the _Checkout_ block (#2562) so this PR is not blocked by this.

Issue #2529, opened 2020-05-21
Checkout block: pressing Enter when the card inputs have focus reloads the page (Firefox)


Aljullu:
I'm not sure if that's an issue we can fix or it's a bug with Stripe's library. But filling it for awareness.

**To Reproduce**
1. With Firefox, go to the _Checkout_ block.
2. Fill the form including the card number inputs.
3. Still having focused one of the card number inputs, press <kbd>Enter</kbd>.
4. Notice the page reloads.

**Expected behavior**
When pressing <kbd>Enter</kbd> it should either submit the _Checkout_ form or do nothing. In other browsers, it seems to do nothing. But reloading the page is confusing.

**Screenshots**
![Peek 2020-05-21 12-40](https://user-images.githubusercontent.com/3616980/82551321-42512000-9b60-11ea-925b-ee84ab9acab1.gif)

nerrad:
I think this is a blocker to release because it would be jarring for shoppers to have this happen (and is also an a11y issue).

PR #2532, opened 2020-05-21
Include `onSubmit` handler for CheckoutForm component.

nerrad:
Fixes: #2529.

See #2529 for background.

Currently the checkout form does not have a submit handler defined for the form itself. It appears that Firefox assumes a default event handler for the submit action if it's not explicitly defined.

Besides the fact that this caused unexpected reloads of the page in Firefox, there are also some potential a11y implications here for pressing the enter key (which is an alternative way for submitting a form) doing nothing.

The resolution for this is to include receiving an `onSubmit` event handler in the `CheckoutForm` component and preventing default behaviour in favor of the passed in event handler if it exists. Then I just ensured to pass through the `onSubmit` callback from the Checkout context state.

Note, while this fixes the reported Firefox bug, and _also_ adds submit handling when the Enter key is pressed in while a stripe field has focus, I was a bit surprised that if any other fields in the form have focus, enter key submission is not respected. I suspect this is because it is mapped for other uses by the fields. I don't think that should hold up this pull getting merged though (and am uncertain whether it's something we need to be concerned about for initial release).

## To test

- See the reproducing steps in #2529 and verify while checked out on this branch that the issue reported there is fixed.
- You should also notice that any validation will run.
- Verify clicking the submit button doesn't cause any issues.
- Verify using enter in the coupon code field doesn't cause any issues.

### DIFF
--- a/assets/js/base/components/cart-checkout/form/index.js
+++ b/assets/js/base/components/cart-checkout/form/index.js
@@ -9,9 +9,20 @@ import PropTypes from 'prop-types';
  */
 import './style.scss';
 
-const CheckoutForm = ( { className, children } ) => {
+const CheckoutForm = ( {
+	className,
+	children,
+	onSubmit = ( event ) => void event,
+} ) => {
+	const formOnSubmit = ( event ) => {
+		event.preventDefault();
+		onSubmit( event );
+	};
 	return (
-		<form className={ classnames( 'wc-block-checkout-form', className ) }>
+		<form
+			className={ classnames( 'wc-block-checkout-form', className ) }
+			onSubmit={ formOnSubmit }
+		>
 			{ children }
 		</form>
 	);
@@ -20,6 +31,7 @@ const CheckoutForm = ( { className, children } ) => {
 CheckoutForm.propTypes = {
 	className: PropTypes.string,
 	children: PropTypes.node,
+	onSubmit: PropTypes.func,
 };
 
 export default CheckoutForm;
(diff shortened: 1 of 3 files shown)


assets/js/base/components/cart-checkout/form/index.js
```diff
@@ -9,9 +9,20 @@ import PropTypes from 'prop-types';
  */
 import './style.scss';
 
-const CheckoutForm = ( { className, children } ) => {
+const CheckoutForm = ( {
+	className,
+	children,
+	onSubmit = ( event ) => void event,
```
haszari, line 15: Do we want to expose the event to the client code? We probably don't need to, so we could consider keeping that internal to `CheckoutForm`, and just expect a generic callback with no params.

Not a blocker - I could imagine it might be useful to expose this to client in some circumstances.

assets/js/base/context/cart-checkout/checkout-state/index.js
```diff
@@ -305,7 +305,8 @@ export const CheckoutStateProvider = ( {
 		isSuccessResponse,
 	] );
 
-	const onSubmit = useCallback( () => {
+	const onSubmit = useCallback( ( event ) => {
+		event.preventDefault();
```
haszari, line 309: Are we calling `preventDefault` twice? Here and inside `CheckoutForm` `onFormSubmit`.

assets/js/base/components/cart-checkout/form/index.js
nerrad: It's common to pass along the event.

assets/js/base/context/cart-checkout/checkout-state/index.js
nerrad: Ooops this was from an early iteration. I forgot to remove this, good catch!

Issue #2458, opened 2020-05-11
Migrate all font-size declarations to use the font-size() mixin


Aljullu:
Migrating all our `font-size` and `line-height` declarations to use the `font-size()` SCSS mixin will make it easier to ensure we use a consistent set of sizes and will ease the transition to global styles (see #2403).

Aljullu:
Given that #2533 was merged and we have a follow-up for filter blocks (#2536), I think we can mark this issue as closed.

PR #2533, opened 2020-05-22
Use named font sizes instead of values in px

Aljullu:
Part of #2458.

This PR updates the `font-size` mixin to use named font sizes instead of px values.

The font size names are `smaller`, `small`, `regular`, `large` and `larger`, I chose this ones so it's consistent with the `$gap-xyz` [variables](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/css/abstracts/_variables.scss#L1-L7) we have for margins/paddings.

Searching & replacing seemed to work quite well in most cases. :ok_hand: But I did some manual changes as well, I tried to comment all of them inline with the code. I didn't modify yet the _All Products_ and filter blocks until we agree on how to map their font sizes to the named ones (#2536).

1. Open the _Cart_ and _Checkout_ blocks and verify they look the same as they did.
2. It would be good to smoke test some other blocks as well to verify there are no regressions (no need to test the _All Products_ and filter blocks, they will be handled in another pull).

### DIFF
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -121,7 +121,7 @@ $font-sizes: (
 	-ms-word-break: break-all;
 }
 
-// Converts a px unit to rem.
-@function rem($size, $base: 16px) {
-	@return $size / $base * 1rem;
+// Converts a px unit to em.
+@function em($size, $base: 16px) {
+	@return $size / $base * 1em;
 }
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -1,10 +1,14 @@
-// Converts a font-size in px to rem and optionally sets a relative line-height.
-@mixin font-size($sizeValue: 16px, $lineHeight: false) {
-	font-size: ($sizeValue / 16px) * 1rem;
+$font-sizes: (
+	"smaller": 0.75,
+	"small": 0.875,
+	"regular": 1,
+	"large": 1.25,
+	"larger": 2,
+);
 
-	@if ($lineHeight) {
-		line-height: $lineHeight / $sizeValue;
-	}
+// Maps a named font-size to its em equivalent.
+@mixin font-size($sizeName) {
+	font-size: map-get($font-sizes, $sizeName) * 1em;
 }
 
 @keyframes loading-fade {
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -1,4 +1,4 @@
-$font-sizes: (
+$fontSizes: (
 	"smaller": 0.75,
 	"small": 0.875,
 	"regular": 1,
@@ -8,7 +8,7 @@ $font-sizes: (
 
 // Maps a named font-size to its em equivalent.
 @mixin font-size($sizeName) {
-	font-size: map-get($font-sizes, $sizeName) * 1em;
+	font-size: map-get($fontSizes, $sizeName) * 1em;
 }
 
 @keyframes loading-fade {
--- a/assets/js/base/components/cart-checkout/button/style.scss
+++ b/assets/js/base/components/cart-checkout/button/style.scss
@@ -5,7 +5,7 @@
 	color: $white;
 	display: flex;
 	font-weight: bold;
-	min-height: rem(48px); // 16px * 3 = 48px, same height as text-input
+	min-height: 3em;
 	justify-content: center;
 	line-height: 1;
 	padding: 0 $gap-small;
--- a/assets/js/base/components/payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/style.scss
@@ -69,7 +69,8 @@
 	}
 
 	.wc-block-gateway-input {
-		@include font-size(16px, 22px);
+		@include font-size(regular);
+		line-height: 1.375;
 		background-color: #fff;
 		padding: $gap-small $gap;
 		border-radius: 4px;
@@ -92,7 +93,8 @@
 
 	label {
 		@include reset-typography();
-		@include font-size(16px, 22px);
+		@include font-size(regular);
+		line-height: 1.375;
 		position: absolute;
 		transform: translateY(0.75em);
 		left: 0;
--- a/assets/js/base/components/payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/style.scss
@@ -70,7 +70,7 @@
 
 	.wc-block-gateway-input {
 		@include font-size(regular);
-		line-height: 1.375;
+		line-height: 1.375; // =22px when font-size is 16px.
 		background-color: #fff;
 		padding: em($gap-small) $gap;
 		border-radius: 4px;
@@ -94,7 +94,7 @@
 	label {
 		@include reset-typography();
 		@include font-size(regular);
-		line-height: 1.375;
+		line-height: 1.375; // =22px when font-size is 16px.
 		position: absolute;
 		transform: translateY(0.75em);
 		left: 0;
--- a/assets/js/base/components/payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/style.scss
@@ -72,7 +72,7 @@
 		@include font-size(regular);
 		line-height: 1.375;
 		background-color: #fff;
-		padding: $gap-small $gap;
+		padding: em($gap-small) $gap;
 		border-radius: 4px;
 		border: 1px solid $input-border-gray;
 		width: 100%;
@@ -131,7 +131,7 @@
 
 	.wc-block-gateway-input.focused.empty,
 	.wc-block-gateway-input:not(.empty) {
-		padding: rem($gap-large) $gap rem($gap-smallest);
+		padding: em($gap-large) $gap em($gap-smallest);
 		& + label {
 			transform: translateY(#{$gap-smallest}) scale(0.75);
 		}
--- a/assets/js/base/components/select/style.scss
+++ b/assets/js/base/components/select/style.scss
@@ -6,7 +6,7 @@
 	label {
 		@include reset-typography();
 		@include font-size(regular);
-		line-height: 1.375;
+		line-height: 1.375; // =22px when font-size is 16px.
 		position: absolute;
 		transform: translateY(0.75em);
 		transform-origin: top left;
--- a/assets/js/base/components/select/style.scss
+++ b/assets/js/base/components/select/style.scss
@@ -1,5 +1,5 @@
 .wc-block-select {
-	height: rem(48px);
+	height: 3em;
 	position: relative;
 	margin-bottom: rem($gap-large);
 
@@ -45,11 +45,11 @@
 		color: $input-text-active;
 		font-family: inherit;
 		font-weight: normal;
-		height: rem(48px);
+		height: 3em;
 		letter-spacing: inherit;
 		line-height: 1;
 		overflow: hidden;
-		padding: rem($gap-large) $gap rem($gap-smallest);
+		padding: em($gap-large) $gap em($gap-smallest);
 		text-overflow: ellipsis;
 		text-transform: none;
 		white-space: nowrap;
--- a/assets/js/base/components/select/style.scss
+++ b/assets/js/base/components/select/style.scss
@@ -5,7 +5,8 @@
 
 	label {
 		@include reset-typography();
-		@include font-size(16px, 22px);
+		@include font-size(regular);
+		line-height: 1.375;
 		position: absolute;
 		transform: translateY(0.75em);
 		transform-origin: top left;
@@ -38,7 +39,7 @@
 	}
 
 	.components-custom-select-control__button {
-		@include font-size(16px);
+		@include font-size(regular);
 		background-color: #fff;
 		box-shadow: none;
 		color: $input-text-active;
@@ -79,7 +80,7 @@
 	}
 
 	.components-custom-select-control__item {
-		@include font-size(16px);
+		@include font-size(regular);
 		margin-left: 0;
 		padding-left: $gap;
 	}
--- a/assets/js/base/components/text-input/style.scss
+++ b/assets/js/base/components/text-input/style.scss
@@ -1,6 +1,6 @@
 .wc-block-text-input {
 	position: relative;
-	margin-bottom: rem($gap-large);
+	margin-bottom: em($gap-large);
 	white-space: nowrap;
 
 	label {
@@ -38,7 +38,7 @@
 	input[type="email"] {
 		@include font-size(regular);
 		background-color: #fff;
-		padding: rem($gap-small) $gap;
+		padding: em($gap-small) $gap;
 		border-radius: 4px;
 		border: 1px solid $input-border-gray;
 		width: 100%;
@@ -46,7 +46,7 @@
 		font-family: inherit;
 		margin: 0;
 		box-sizing: border-box;
-		height: rem(48px);
+		height: 3em;
 		min-height: 0;
 		color: $input-text-active;
 
@@ -59,7 +59,7 @@
 	&.is-active input[type="url"],
 	&.is-active input[type="text"],
 	&.is-active input[type="email"] {
-		padding: rem($gap-large) 0 rem($gap-smallest) $gap;
+		padding: em($gap-large) 0 em($gap-smallest) $gap;
 	}
 
 	&.has-error input {
--- a/assets/js/base/components/text-input/style.scss
+++ b/assets/js/base/components/text-input/style.scss
@@ -5,7 +5,7 @@
 
 	label {
 		@include reset-typography();
-		@include font-size(16px);
+		@include font-size(regular);
 		position: absolute;
 		transform: translateY(0.75em);
 		left: 0;
@@ -36,7 +36,7 @@
 	input[type="url"],
 	input[type="text"],
 	input[type="email"] {
-		@include font-size(16px);
+		@include font-size(regular);
 		background-color: #fff;
 		padding: rem($gap-small) $gap;
 		border-radius: 4px;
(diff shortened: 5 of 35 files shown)


assets/js/blocks/attribute-filter/edit.js
```diff
@@ -237,7 +237,7 @@ const Edit = ( { attributes, setAttributes, debouncedSpeak } ) => {
 				) }
 			</p>
 			<Button
-				className="wc-block-attribute-filter__add_attribute_button"
+				className="wc-block-attribute-filter__add-attribute-button"
```
Aljullu, line 240: Not related to this pull, but I took the chance to change this class name so it follows BEM convention.

assets/js/blocks/attribute-filter/editor.scss
```diff
@@ -11,18 +11,13 @@
 	}
 	.components-placeholder__fieldset {
 		display: block; /* Disable flex box */
-
-		p {
-			@include font-size(14px);
```
Aljullu, line 16: Regarding this font size and the one below, they were making the placeholder text and button smaller.

_How it looked:_
Screenshot from 2020-05-21 15-58-43

_How it looks after removing those declarations:_
Screenshot from 2020-05-21 15-58-13

I took a look at the designs, they use the old placeholder style, but I don't see a reason why that text should be smaller than the rest:
imatge

assets/js/blocks/cart-checkout/checkout/style.scss
```diff
@@ -106,11 +106,11 @@
 		padding-left: $gap-large;
 		padding-top: $gap;
 		padding-bottom: $gap;
-		line-height: $gap-large;
+		line-height: 1.375;
 
 		p,
 		.wc-block-product-metadata {
-			line-height: $gap;
+			line-height: 1.375;
```
Aljullu, line 113: These line heights needed to be relative.

assets/js/blocks/product-search/style.scss
```diff
@@ -15,11 +15,11 @@
 		border: none;
 		cursor: pointer;
 		background: none;
-		padding: 8px;
+		padding: 0 0.5em;
```
Aljullu, line 18: Like with in the _Product Categories List_ block, this button takes all the available vertical space so there is no need to have top and bottom padding. This way both buttons share the same styles (see #2517).

assets/js/base/components/cart-checkout/button/style.scss
```diff
@@ -5,7 +5,7 @@
 	display: flex;
 	font-size: inherit;
 	font-weight: bold;
-	min-height: rem(48px); // 16px * 3 = 48px, same height as text-input
+	min-height: 3em;
```
nerrad, line 8: Why not using the new `em` function (`em(48px)`) (assuming there's good reason to switch from `rem` to `em`)?

assets/js/base/components/payment-methods/style.scss
```diff
@@ -92,7 +93,8 @@
 
 	label {
 		@include reset-typography();
-		@include font-size(16px, 22px);
+		@include font-size(regular);
+		line-height: 1.375; // =22px when font-size is 16px.
```
nerrad, line 97: We originally had `line-height` included as an argument of the `font-size` mixin, why did you choose to extract it back out?

assets/js/base/components/select/style.scss
```diff
@@ -38,17 +39,17 @@
...
 	.components-custom-select-control__button {
-		@include font-size(16px);
+		@include font-size(regular);
 		background-color: #fff;
 		box-shadow: none;
 		color: $input-text-active;
 		font-family: inherit;
 		font-weight: normal;
-		height: rem(48px);
+		height: 3em;
```
nerrad, line 48: why not `em(48px)` here?

assets/js/base/components/text-input/style.scss
```diff
@@ -36,17 +36,17 @@
...
 		border-radius: 4px;
 		border: 1px solid $input-border-gray;
 		width: 100%;
-		line-height: 1.375;
+		line-height: 1.375; // =22px when font-size is 16px.
 		font-family: inherit;
 		margin: 0;
 		box-sizing: border-box;
-		height: rem(48px);
+		height: 3em;
```
nerrad, line 49: `em( 48px )`?

assets/js/blocks/attribute-filter/edit.js
nerrad: This is released code though, so we might want to make sure it's noted in the dev notes/changelog for the next release.

assets/js/blocks/cart-checkout/checkout/style.scss
nerrad: It feels like these should have their own mixin helper.

assets/js/blocks/product-categories/style.scss
```diff
@@ -67,11 +67,11 @@
 	border: none;
 	cursor: pointer;
 	background: none;
-	padding: 0 rem(8, 13);
+	padding: 0 0.5em;
```
nerrad, line 70: curious why you're not using `em()` for all these values?

assets/js/blocks/attribute-filter/edit.js
Aljullu: Those classes (`wc-block-attribute-filter__add-attribute-button` and `wc-block-price-slider__add-product-button`) are only used in the editor, they refer to the button in this placeholder:
Screenshot from 2020-05-21 15-58-13

In that case, when an element is only displayed in the editor, do we need dev notes too?

assets/css/abstracts/_mixins.scss
```diff
@@ -1,10 +1,14 @@
...
+	"small": 0.875,
+	"regular": 1,
+	"large": 1.25,
+	"larger": 2,
+);
+
+// Maps a named font-size to its em equivalent.
+@mixin font-size($sizeName) {
+	font-size: map-get($fontSizes, $sizeName) * 1em;
 }
```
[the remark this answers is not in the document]
Aljullu: We saw `rem` units were not playing well with many themes because they set `<html>` font size to something very small and then make it bigger in the `<body>`. One example of this is Twenty Twenty:
https://github.com/WordPress/twentytwenty/blob/master/style.css#L117

In pb0Spc-Cv-p2 I proposed switching to `em` to fix this. Considering `rem` can give so different results depending on the theme, I think we should avoid it when possible.

assets/js/base/components/cart-checkout/button/style.scss
Aljullu: In pb0Spc-Cv-p2's comments, @mikejolley raised the concern that it might be confusing for developers to write px units but see they are `em`'s in the resulting CSS. In this concrete case, if you are debugging a theme with a font size of 14px and want to know how high that element will be, I think it's easier to rationale "`3em = 14*3 = 42`" than thinking "`48 / 16 = 3, 14 * 3 = 42`".

I still think the `em()` mixin makes sense to convert px variables to `em` values. But in other cases it might be more intuitive to use `em` than `px`. Does that make sense?

assets/js/blocks/product-categories/style.scss
Aljullu: Like before, I think using `em()` (or `rem()` as it was until now) makes this code harder to understand.

In fact, there is a curious bug here. That element had 13px font-size, and we needed a padding of 8px, so it was `rem(value: 8, base: 13)`. But in #2498 I removed the font-size 13px but forgot to update this mixin. :man_facepalming: So instead of 8px now the padding is 9.8px (=8/13*16).

To avoid the need to do all this kind of calculations, I think we should get used to use `em` values when possible.

assets/js/base/components/payment-methods/style.scss
Aljullu: The reason is similar to why I removed several calls to the `rem()` mixin. You can see [this comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2533#discussion_r429988538) for an example on how things might start getting difficult if we use px values when they are in fact `em` under-the-hood.

For the line-height, my reasoning was a bit the same. If the font sizes are named, does it make sense the line-height being in px? What if we change the `large` for size from `1.25` to `1.5` at some point? If line height are specified in px, we would need to update all of them while if they are unitless, they will automatically adapt.

I agree `1.375` looks like a magic number, here. That's why I added a comment next to it explaining where it comes from. Another option would be to create SCSS variables for different line heights so we can reuse them and we know where they come from. Do you think that would make it easier?

Aljullu:
Thanks for the review @nerrad! There were some comments which were very similar so I tried to only answer one of each to focus the discussion there.

The goal of this PR is to make writing and reading styles easier for us. I tried to explain the rationale for the changes, but none of them is a hard opinion. I acknowledge what I see as being harder/easier might differ from somebody else, so I'm open to discuss all changes individually to make it easier for the team to modify styles.

assets/css/abstracts/_mixins.scss
```diff
@@ -1,10 +1,14 @@
...
-
-	@if ($lineHeight) {
-		line-height: $lineHeight / $sizeValue;
-	}
+$fontSizes: (
+	"smaller": 0.75,
+	"small": 0.875,
+	"regular": 1,
+	"large": 1.25,
+	"larger": 2,
```
senadir, line 6: I think `smaller` `small`... would fit well with sizing but not much with font sizes, as they provide no obvious context, I would possibly suggest something like `heading` `caption` `body` and so on, they provide more context.

see examples in:
- [Polaris](https://github.com/Shopify/polaris-react/blob/master/src/styles/foundation/_typography.scss#L17-L56)
- [Material Design](https://github.com/material-components/material-components-web/blob/master/packages/mdc-typography/_variables.scss#L166-L178)

However, there’s no size fit all, other opinions on this:
- [Font size naming convention in designer news](https://www.designernews.co/stories/105780-font-size-naming-convention).
- [Naming typography patterns in CSS](https://keenanpayne.com/blog/naming-typography-patterns-css.html) 
- [Typography in Design Systems](https://superfriendlydesign.systems/articles/typography-in-design-systems/)
- [CSS utility classes for sizing and naming conventions](https://catalin.red/css-utility-classes-naming-conventions/)

### Why do I care so much about such a small thing?

Because those names will in the future be a public API for our blocks, both in global styles, extending and our style guide, getting them right is somehow a bit important.

This is by no means a blocker, we can address this later, we can merge this, or we can address it now, your call.

Aljullu: Loads of good ideas here, thanks for the feedback @senadir. So if I understand you correctly, you are proposing to define a `heading` named font size, for example, so if we ever get a `--wp-gs-heading-font-size` custom property (or whichever system global styles ends up using), we could easily update the font sizes of all headings to use the style from global styles. Is that correct?

While I really like this idea, I have a couple of thoughts about this:
* Currently, if we have a heading or a button that should inherit the theme font size, I think it shouldn't have any font size defined in its CSS and instead rely on the inherited value. Said in other words: all the font sizes introduced in this PR are variations of the body text font size. If an element must honor the theme styles (for example, an `<h2>` needs to render as all other `<h2>` are rendered in that theme), it should just inherit the font size. At least, until global styles land.
* In parallel to this, I think it's a bit too early to know how the final implementation of global styles will be and which specific values we will be able to use. In the examples I have seen, we would be able to read the heading, body and button font sizes, for example. But what about caption, for example? There are dozens of HTML elements and I don't think global styles will expose specific styles for all of them, so my position with this would be to wait a bit more to see how global styles evolve before embracing this.

Let me know how that sounds! In any case, I agree we shouldn't block this PR because of that since it can be tackled in a follow-up.

senadir: Well, I'm not suggesting 1:1 with GS, global styles could be an example but it’s still early for it, my point is that naming font sizes large, small... provides no context to us as developers or to third-party developers, font size depends on context, you will probably never have a heading that is 10px, having semantic names will make sure we're using correct tags and correct placement.
Global styles could be another layer of abstraction above this if we don't want to expose everything, I don't disagree on the values or how they're used, mainly the naming and how intuitive it can get.

assets/css/abstracts/_mixins.scss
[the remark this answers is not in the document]
mikejolley: 👍 EM is relative to the current content too so it should fit better within themes. Not all themes use the html/body class for the base size.

assets/css/abstracts/_mixins.scss
```diff
@@ -117,7 +121,7 @@
 	-ms-word-break: break-all;
 }
 
-// Converts a px unit to rem.
-@function rem($size, $base: 16px) {
-	@return $size / $base * 1rem;
+// Converts a px unit to em.
+@function em($size, $base: 16px) {
+	@return $size / $base * 1em;
 }
```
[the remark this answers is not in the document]
mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2533#discussion_r429430535

assets/css/abstracts/_mixins.scss
mikejolley: > I think smaller small... would fit well with sizing but not much with font sizes, as they provide no obvious context, I would possibly suggest something like heading caption body and so on, they provide more context.

Isn't the point here to have a flexible system where the content is unknown? Is we based it on selectors we could run into code smell where the design uses one size, and we're forced to use a named size that doesn't match our usage.

docs/contributors/coding-guidelines.md
```diff
@@ -85,11 +85,9 @@ Styles placed in a `style.scss` file will be built into `build/style.css`, to lo
 
 ### Accessible font sizes
 
-We want our font sizes to be declared with rem or em units instead of hardcoded px. That's important for accessibility because it allows users to make the text bigger and easier to read.
+Font sizes must be defined using the `font-size()` mixin, it takes a size name (`smaller`, `small`, `regular`, `large`. `larger`) and returns a font-size declaration in em units. This allows having a consistent set of font sizes we use across blocks and using em means they are accessible and fit better in different themes.
```
mikejolley, line 88:
```suggestion
Font sizes must be defined using the `font-size()` mixin, it takes a named size (`smaller`, `small`, `regular`, `large`. `larger`) and returns a font-size declaration in `em` units. This provides a consistent set of font sizes to be used across blocks, and by using `em` it increases the likelihood that blocks are accessible and fit better within different themes.
```

assets/css/abstracts/_mixins.scss
```diff
@@ -1,10 +1,14 @@
...
+	"smaller": 0.75,
+	"small": 0.875,
+	"regular": 1,
+	"large": 1.25,
+	"larger": 2,
+);
+
+// Maps a named font-size to its em equivalent.
+@mixin font-size($sizeName) {
+	font-size: map-get($fontSizes, $sizeName) * 1em;
```
mikejolley, line 11: Just a small question on mixin usage. Are there any advantages to the mixin over variables?

i.e. 
```
@include font-size(larger);
```
vs
```
font-size: $larger;
```

Aljullu: A mixin gives us more control, so the idea is that it will allow us to easily migrate all font sizes to support global styles in the future. There is an example here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2403#issuecomment-624056776.

That example is a bit old (it uses px→rem instead of named→em) and global styles are still in development so its shape might be different from the one in that example, but in general I think having all font sizes in a mixin will eventually make it easier to embrace global styles.

However, I acknowledge that with search & replace we can easily switch between `@include font-size(larger);` ←→ `font-size: $larger;`, so I don't have a strong preference on any of them other than in the future we might be forced to use mixins.

docs/contributors/coding-guidelines.md
Aljullu: Thanks for the suggestion. Merged! :+1:

assets/css/abstracts/_mixins.scss
senadir: > Isn't the point here to have a flexible system where the content is unknown

A design system still needs to follow certain constraints 

> If we based it on selectors we could run into code smell where the design uses one size, and we're forced to use a named size that doesn't match our usage.

We're not basing it on selector but on context, a body-sized text could be any element, not strictly applied to the body tag, by having those rules, a designer or a developer will ask the question "Is this a heading? if not, why is the size the same as other headings within this design", if we're going to communicate and emphasize importance and context via fonts sizes, variables should be named accordingly.

assets/css/abstracts/_mixins.scss
mikejolley: Can we support both? Readability I prefer standard font-size. In no means a blocker though :)

assets/css/abstracts/_mixins.scss
mikejolley: @senadir Not all headers are equal. There are also times where a heading may be styled differently right? I think the defined sizes work better for our use case. We have a base and everything else is just a slight variation of that. Perhaps we just need inline docs to identify which are most commonly used for headings?

assets/js/blocks/attribute-filter/edit.js
mikejolley: We shouldn't do - this is internal.

assets/css/abstracts/_mixins.scss
Aljullu: If by headings we are referring to `<h1>`, `<h2>`, etc., I don't think we need to establish a relationship between them and specific font sizes, in several occasions we need to display headings as if they were regular text.

In the other hand, for headings that need to honor the font size defined by the theme (I'm thinking on the Filter block titles, for example) I think not setting any font size on our end and letting styles cascade is the best option.
imatge

assets/css/abstracts/_mixins.scss
Aljullu: Hmm, I'm a bit on the edge with this. While I don't have any strong preference between both systems, I would really prefer that we stick with one. It really simplifies things when searching for code.

I will merge this as-is for now, and we can migrate in a follow.

Aljullu:
Thanks for all the reviews!

I created a follow-up to explore ways to simplify line-height property: #2584.

PR #2542, opened 2020-05-22
[Experiment] Alternative layout rendering

mikejolley:
This is a POC for single product block. I am aware it breaks All Products currently.


mikejolley:
I'm happy with this approach; I've spotted no issues from switching to this implementation, and All Products is working now. We'll want to add back compat for switching layout config -> inner blocks in a follow up so the older code can be pruned. I'm merging this into single products now so I can refine that other PR for final review.

PR #2544, opened 2020-05-22
Refactor E2E setup routine

senadir:
This PR does some things:

1- Remove unnecessary setup routine from jest-setup (delete posts and delete products), our tests don't affect those things so there is no need to run that, I ported them when I was setting things from Gutenberg.
2- It checks if we already have products before importing, this means we won't delete and import product on each test, reduce the time since we will cover 28 block.
3- Removes `clearLocalStorage` because it won't work unless I delete pages? didn't make sense to me, it was throwing `LocalStorage access denied` we don't manipulate local storage in our e2e tests, so this should be fine.
4- I tried to use [pptr-testing-library](https://github.com/testing-library/pptr-testing-library) didn't work well, seems that one is not fully stable yet for prep tests, causing ` Error: Execution context was destroyed, most likely because of a navigation.` when used.

Note, this test will also be deleted once we import Products via JSON, writing that json is tedious and I don't want it to block me right now while I ensure other blocks work fine.

No testing instructions required, this is safe as long as travis is green.

### DIFF
--- a/tests/e2e-tests/config/jest.setup.js
+++ b/tests/e2e-tests/config/jest.setup.js
@@ -55,7 +55,7 @@ async function setupBrowser() {
  */
 async function importSampleProducts() {
 	await switchUserToAdmin();
-	// Visit `/wp-admin/edit.php?post_type=product` so we can see a list of products and delete them.
+	// Visit `/wp-admin/edit.php?post_type=product` so we can see a list of products and decide if we should import them or not.
 	await visitAdminPage( 'edit.php', 'post_type=product' );
 	const emptyState = await page.evaluate( () =>
 		window.find( 'Ready to start selling something awesome' )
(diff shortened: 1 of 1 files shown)


tests/e2e-tests/config/jest.setup.js
```diff
@@ -46,97 +45,48 @@ const OBSERVED_CONSOLE_MESSAGE_TYPES = {
...
+		await page.select(
+			'.wc-importer-mapping-table tr:nth-child(29) select',
+			''
+		);
+		await page.click( '.wc-actions .button-next' );
+		await page.waitForXPath(
+			"//*[@class='woocommerce-importer-done' and contains(., 'Import complete! ')]"
+		);
+		await switchUserToTest();
+	}
```
nerrad, line 89: Is there not any concern there that we pollute data between test runs if we're not resetting the known state between tests runs in the db?

senadir: this is a valid concern, but for our case, we're not changing or manipulating products in any way in our tests, we're just querying them, so the initial state of our products should be the same when we finish the tests, and between each test, if a test must change the data shape, that test should be responsible for resetting it.

PR #2545, opened 2020-05-22
Ensure default selected tab is set for payment methods selection ui.

### LINKED ISSUE #2531: Checkout: Fix payment method step not having a default payment method selected.
**Describe the bug**
If you load the checkout with no saved payment methods, there should be a default payment method selected (the first tab). However, there is none selected.

**To Reproduce**

1. Easiest to observe in incognito mode for your browser.
2. Add some products to the cart.
3. Visit the checkout (rendered by the block).
4. Observe there is no default payment method selected:

**Expected behavior**
I expect to see a default payment method (Credit Card in this case) selected and showing content:

I reproduced this in Chrome and Firefox (4.2.RC1 of Woo Core, 2.5.7RC1 of Storefront, and Woo Blocks master).

nerrad:
Fixes #2531 (see for details).

The solution here was to just use the built-in functionality Reakit has for setting the default tab.

## To test

- In incognito mode, make sure that the payment methods tabs has the default selected tab to the first tab.
- Next, while logged in as a users with saved payment methods, make sure that when you select the "Use a new payment method option", the payment methods tabs show with the first tab selected.

### DIFF
--- a/assets/js/base/components/tabs/index.js
+++ b/assets/js/base/components/tabs/index.js
@@ -20,12 +20,10 @@ const Tabs = ( {
 	instanceId,
 	id,
 } ) => {
-	const initialTab =
-		( initialTabName || !! tabs.length ) &&
-		`${ instanceId }-${ initialTabName ? initialTabName : tabs[ 0 ].name }`;
-	const tabState = useTabState( {
-		selectedId: initialTab,
-	} );
+	const initialTab = initialTabName
+		? { selectedId: `${ instanceId }-${ initialTabName }` }
+		: undefined;
+	const tabState = useTabState( initialTab );
 	if ( tabs.length === 0 ) {
 		return null;
 	}


assets/js/base/components/tabs/index.js
```diff
@@ -20,12 +20,10 @@ const Tabs = ( {
...
-	const initialTab =
-		( initialTabName || !! tabs.length ) &&
-		`${ instanceId }-${ initialTabName ? initialTabName : tabs[ 0 ].name }`;
-	const tabState = useTabState( {
-		selectedId: initialTab,
-	} );
+	const initialTab = initialTabName
+		? { selectedId: `${ instanceId }-${ initialTabName }` }
+		: undefined;
+	const tabState = useTabState( initialTab );
```
senadir, line 26: weird, I assumed this will always default to the preselected value or the first tab, reakit handles this logic internally `tabs[ 0 ].name`.

PR #2548, opened 2020-05-23
Add fixtures data for all entities.

senadir:
This PR depends (on builds upon https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2544) I set it against master so that travis can run.

This was a huge blocker for me, glad I get it done (almost).

ِAdds fixture data for settings, taxes, coupons, shipping, payment, reviews, and a single product (needed for reviews).

those items are loaded from WC REST API, so this is done very quickly.
the code is run only once before all tests start, after finishing tests, all created items will be deleted, for reference, only the created ones will get deleted.
If you happen to create something in a test, make sure to delete it yourself or append it to the delete queue.

```js
beforeAll( async () => {
// you created a product here
// single item
global.wc.products.push( productId );
// many items
global.wc.products = [ ...productIds, global.wc.products ];
});
```
Same applies for taxes, coupons, and shipping zones, each one is exposes in an array in:

```js
global.wc {
    coupons,
    products,
    taxes,
    shippingMethods,
}
```

* Start the docker instance `npm run test:e2e:up:build`
* build the blocks to the latest version `npm run build`
* run some e2e tests `npm run test:e2e`
* When tests end, visit http://localhost:8084
* Make sure there are no coupons, products,  shipping zones, or taxes.
* Make sure fixtures actually loaded, this is done by checking the store address, it will be `60 29th Street #343`.
* Stop the docker, by running `npm run test:e2e:down`

What will follow is more tests and more fixtures for products, for now, we will still use the standard routine we had from before.

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
 		"@types/react": "16.9.35",
 		"@types/wordpress__data": "4.6.7",
 		"@types/wordpress__element": "2.4.1",
+		"@woocommerce/woocommerce-rest-api": "^1.0.1",
 		"@wordpress/babel-preset-default": "4.10.0",
 		"@wordpress/base-styles": "1.4.0",
 		"@wordpress/blocks": "6.12.0",
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
 		"@types/react": "16.9.35",
 		"@types/wordpress__data": "4.6.7",
 		"@types/wordpress__element": "2.4.1",
-		"@woocommerce/woocommerce-rest-api": "^1.0.1",
+		"@woocommerce/woocommerce-rest-api": "1.0.1",
 		"@wordpress/babel-preset-default": "4.10.0",
 		"@wordpress/base-styles": "1.4.0",
 		"@wordpress/blocks": "6.12.0",
--- a/tests/e2e-tests/config/fixture-loaders.js
+++ b/tests/e2e-tests/config/fixture-loaders.js
@@ -1,0 +1,291 @@
+/* eslint-disable no-unused-vars */
+/* eslint-disable no-console */
+/**
+ * External dependencies
+ */
+const WooCommerceRestApi = require( '@woocommerce/woocommerce-rest-api' )
+	.default;
+
+const WooCommerce = new WooCommerceRestApi( {
+	url: 'http://localhost:8084/', // Your store URL
+	consumerKey: 'consumer_key', // Your consumer key
+	consumerSecret: 'consumer_secret', // Your consumer secret
+	version: 'wc/v3', // WooCommerce WP REST API version
+	axiosConfig: {
+		auth: {
+			username: 'admin',
+			password: 'password',
+		},
+	},
+} );
+
+const setupSettings = () => {
+	WooCommerce.post( 'settings/general/batch', {
+		update: [
+			{
+				id: 'woocommerce_store_address',
+				value: '60 29th Street #343',
+			},
+			{
+				id: 'woocommerce_store_city',
+				value: 'San Francisco',
+			},
+			{
+				id: 'woocommerce_store_country',
+				value: 'US:CA',
+			},
+			{
+				id: 'woocommerce_store_postcode',
+				value: '94110',
+			},
+			{
+				id: 'woocommerce_allowed_countries',
+				value: 'specific',
+			},
+			{
+				id: 'woocommerce_specific_allowed_countries',
+				value: [ 'DZ', 'CA', 'NZ', 'ES', 'GB', 'US' ],
+			},
+			{
+				id: 'woocommerce_ship_to_countries',
+				value: 'specific',
+			},
+			{
+				id: 'woocommerce_specific_ship_to_countries',
+				value: [ 'DZ', 'CA', 'NZ', 'ES', 'GB', 'US' ],
+			},
+			{
+				id: 'woocommerce_enable_coupons',
+				value: 'yes',
+			},
+			{
+				id: 'woocommerce_calc_taxes',
+				value: 'yes',
+			},
+			{
+				id: 'woocommerce_currency',
+				value: 'USD',
+			},
+		],
+	} ).catch( ( error ) => console.log( error ) );
+};
+
+const createTaxes = () => {
+	return WooCommerce.post( 'taxes/batch', {
+		create: [
+			{
+				country: 'US',
+				rate: '5.0000',
+				name: 'State Tax',
+				shipping: false,
+				priority: 1,
+			},
+			{
+				country: 'US',
+				rate: '10.000',
+				name: 'Sale Tax',
+				shipping: false,
+				priority: 2,
+			},
+			{
+				country: 'UK',
+				rate: '20.000',
+				name: 'VAT',
+				shipping: false,
+			},
+		],
+	} )
+		.then( ( response ) =>
+			response.data.create.map( ( taxes ) => taxes.id )
+		)
+		.catch( ( error ) => console.log( error ) );
+};
+
+const deleteTaxes = ( ids ) => {
+	WooCommerce.post( 'taxes/batch', {
+		delete: ids,
+	} ).catch( ( error ) => console.log( error ) );
+};
+const createCoupons = () => {
+	const coupons = [
+		{
+			code: 'coupon',
+			discount_type: 'fixed_cart',
+			amount: '5',
+		},
+		{
+			code: 'oldcoupon',
+			discount_type: 'fixed_cart',
+			amount: '5',
+			date_expires: '2020-01-01',
+		},
+		{
+			code: 'below100',
+			discount_type: 'percent',
+			amount: '20',
+			maximum_amount: '100.00',
+		},
+		{
+			code: 'above50',
+			discount_type: 'percent',
+			amount: '20',
+			minimum_amount: '50.00',
+		},
+		{
+			code: 'a12s',
+			discount_type: 'percent',
+			amount: '100',
+			individual_use: true,
+			email_restrictions: '*@automattic.com%2C *@a8c.com',
+		},
+		{
+			code: 'freeshipping',
+			discount_type: 'percent',
+			amount: '0',
+			free_shipping: true,
+		},
+	];
+	return WooCommerce.post( 'coupons/batch', { create: coupons } )
+		.then( ( response ) =>
+			response.data.create.map( ( coupon ) => coupon.id )
+		)
+		.catch( ( e ) => console.log( e.response.data.message ) );
+};
+const deleteCoupons = ( ids ) => {
+	return WooCommerce.post( 'coupons/batch', {
+		delete: ids,
+	} ).catch( ( error ) => console.log( error ) );
+};
+// currently this only creates a single product for the sake of review
+// more products will be added later
+const createProducts = () => {
+	return WooCommerce.post( 'products/batch', {
+		create: [
+			{
+				name: 'Woo Single #1',
+				type: 'simple',
+				regular_price: '21.99',
+				virtual: true,
+				downloadable: true,
+				downloads: [
+					{
+						name: 'Woo Single',
+						file:
+							'http://demo.woothemes.com/woocommerce/wp-content/uploads/sites/56/2013/06/cd_4_angle.jpg',
+					},
+				],
+				images: [
+					{
+						src:
+							'http://demo.woothemes.com/woocommerce/wp-content/uploads/sites/56/2013/06/cd_4_angle.jpg',
+					},
+				],
+			},
+		],
+	} )
+		.then( ( products ) => {
+			createReviews( products.data.create[ 0 ].id );
+			return products.data.create.map( ( product ) => product.id );
+		} )
+		.catch( ( e ) => console.log( e.response.data.message ) );
+};
+// deleting products will also delete review.
+const deleteProducts = ( ids ) => {
+	WooCommerce.post( 'products/batch', {
+		delete: ids,
+	} ).catch( ( error ) => console.log( error ) );
+};
+const createReviews = ( id ) => {
+	return WooCommerce.post( 'products/reviews/batch', {
+		create: [
+			{
+				product_id: id,
+				review: 'Looks fine',
+				reviewer: 'John Doe',
+				reviewer_email: 'john.doe@example.com',
+				rating: 4,
+			},
+			{
+				product_id: id,
+				review: 'I love this album',
+				reviewer: 'John Doe',
+				reviewer_email: 'john.doe@example.com',
+				rating: 5,
+			},
+			{
+				product_id: id,
+				review: 'a fine review',
+				reviewer: "John Doe' niece",
+				reviewer_email: 'john.doe@example.com',
+				rating: 5,
+			},
+		],
+	} ).catch( ( e ) => console.log( e.response.data.message ) );
+};
+const enableCheque = () => {
+	return WooCommerce.post( 'payment_gateways/cheque', {
+		enabled: true,
+	} ).catch( ( error ) => console.log( error ) );
+};
+const enablePaypal = () => {
+	return WooCommerce.post( 'payment_gateways/paypal', {
+		enabled: true,
+	} ).catch( ( error ) => console.log( error ) );
+};
+
+const enablePaymentGateways = () => {
+	Promise.all( [ enableCheque(), enablePaypal() ] );
+};
+
+const createShippingMethods = () => {
+	return WooCommerce.post( 'shipping/zones', { name: 'UK' } )
+		.then( ( response ) => {
+			return response.data.id;
+		} )
+		.then( ( zoneId ) => {
+			WooCommerce.put( `shipping/zones/${ zoneId }/locations`, {
+				code: 'UK',
+			} );
+			return zoneId;
+		} )
+		.then( ( zoneId ) => {
+			WooCommerce.post( `shipping/zones/${ zoneId }/methods`, {
+				method_id: 'flat_rate',
+				settings: {
+					title: 'Normal Shipping',
+					cost: '20.00',
+				},
+			} );
+			return zoneId;
+		} )
+		.then( ( zoneId ) => {
+			WooCommerce.post( `shipping/zones/${ zoneId }/methods`, {
+				method_id: 'free_shipping',
+				settings: {
+					title: 'Free Shipping',
+					cost: '00.00',
+					requires: 'coupon',
+				},
+			} );
+			return zoneId;
+		} )
+		.catch( ( e ) => console.log( e.response.data.message ) );
+};
+// deleting a zone will also delete methods and locations.
+const deleteShippingZone = ( id ) => {
+	WooCommerce.delete( `shipping/zones/${ id }`, {
+		force: true,
+	} ).catch( ( error ) => console.log( error ) );
+};
+module.exports = {
+	setupSettings,
+	createTaxes,
+	deleteTaxes,
+	createCoupons,
+	deleteCoupons,
+	createProducts,
+	deleteProducts,
+	enablePaymentGateways,
+	createShippingMethods,
+	deleteShippingZone,
+};
--- a/tests/e2e-tests/config/fixture-loaders.js
+++ b/tests/e2e-tests/config/fixture-loaders.js
@@ -238,44 +238,49 @@ const enablePaymentGateways = () => {
 };
 
 const createShippingMethods = () => {
-	return WooCommerce.post( 'shipping/zones', { name: 'UK' } )
-		.then( ( response ) => {
-			return response.data.id;
-		} )
-		.then( ( zoneId ) => {
-			WooCommerce.put( `shipping/zones/${ zoneId }/locations`, {
-				code: 'UK',
-			} );
-			return zoneId;
-		} )
-		.then( ( zoneId ) => {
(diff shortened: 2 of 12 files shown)


senadir:
cc @ObliviousHarmony while we get the tool done, I went with this approach, migration to that tool should simple as long as we agree on a basic standard, for this code, the interface used will always:
- Return a promise that resolves to the created entities IDs.
- A promise that resolves to nothing for things that we can't destroy (settings).
- Destroy something returns nothing as well.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
+/* eslint-disable no-console */
+/**
+ * External dependencies
+ */
+const WooCommerceRestApi = require( '@woocommerce/woocommerce-rest-api' )
+	.default;
+
+require( 'dotenv' ).config();
```
senadir, line 8: Environment variables are not available this early so we load them.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+	url: `${ process.env.WORDPRESS_BASE_URL }:${ process.env.WORDPRESS_PORT }/`, // Your store URL
+	consumerKey: 'consumer_key', // Your consumer key
+	consumerSecret: 'consumer_secret', // Your consumer secret
+	version: 'wc/v3', // WooCommerce WP REST API version
+	axiosConfig: {
+		auth: {
+			username: process.env.WORDPRESS_LOGIN,
+			password: process.env.WORDPRESS_PASSWORD,
+		},
+	},
```
senadir, line 20: consumer key and secret are not used currently (only placeholder to avoid errors) because we can't create them dynamically yet, for that we use Basic auth with user and password.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+} );
+
+const setupSettings = () => {
+	WooCommerce.post( 'settings/general/batch', {
```
senadir, line 24: fixtures that have no use later don't return a promise, the function will just return undefined.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+			{
+				country: 'UK',
+				rate: '20.000',
+				name: 'VAT',
+				shipping: false,
+			},
+		],
+	} )
+		.then( ( response ) =>
+			response.data.create.map( ( taxes ) => taxes.id )
```
senadir, line 100: fixtures that are needed later will resolve to an array of newly created things.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+		)
+		.catch( ( error ) => console.log( error ) );
+};
+
+const deleteTaxes = ( ids ) => {
```
senadir, line 105: that array is later passed to the delete (teardown) function that removes them.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+						file:
+							'http://demo.woothemes.com/woocommerce/wp-content/uploads/sites/56/2013/06/cd_4_angle.jpg',
+					},
+				],
+				images: [
+					{
+						src:
+							'http://demo.woothemes.com/woocommerce/wp-content/uploads/sites/56/2013/06/cd_4_angle.jpg',
+					},
+				],
```
senadir, line 183: This is pulled straight from the docs, later we will create more products.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+			},
+		],
+	} )
+		.then( ( products ) => {
+			createReviews( products.data.create[ 0 ].id );
```
senadir, line 188: since reviews depend on products, we create them inside of the product.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+const enableCheque = () => {
+	return WooCommerce.post( 'payment_gateways/cheque', {
+		enabled: true,
+	} ).catch( ( error ) => console.log( error ) );
+};
+const enablePaypal = () => {
+	return WooCommerce.post( 'payment_gateways/paypal', {
+		enabled: true,
+	} ).catch( ( error ) => console.log( error ) );
+};
```
senadir, line 235: we only enable blocks supported payment methods.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+						title: 'Free Shipping',
+						cost: '00.00',
+						requires: 'coupon',
+					},
+				} );
+				return zoneId;
+			} )
+			.catch( ( e ) => console.log( e.response.data.message ) );
+
+	return Promise.all( [ UKShipping() ] );
```
senadir, line 276: you can only create one shipping zone at a time, we can be refactored so that the create part is dynamic, for now, you assign the create promise to a variable `UKShipping` or anything and pass it to promise.all.

tests/e2e-tests/config/jest.config.js
```diff
@@ -14,7 +14,8 @@ module.exports = {
 
 	// Where to look for test files
 	roots: [ '<rootDir>/tests/e2e-tests/specs' ],
-
+	globalSetup: '<rootDir>/tests/e2e-tests/config/setup.js',
+	globalTeardown: '<rootDir>/tests/e2e-tests/config/teardown.js',
```
senadir, line 18: global setup and teardown only run once, before all tests.

tests/e2e-tests/config/setup.js
```diff
@@ -0,0 +1,32 @@
...
+	createTaxes,
+	createCoupons,
+	createProducts,
+	createShippingMethods,
+	enablePaymentGateways,
+} = require( './fixture-loaders' );
+const { setup: setupPuppeteer } = require( 'jest-environment-puppeteer' );
+
+module.exports = async ( globalConfig ) => {
+	await setupPuppeteer( globalConfig );
```
senadir, line 15: However, puppeteer also hooks into globalSetup, so we have to call it here.

tests/e2e-tests/config/setup.js
```diff
@@ -0,0 +1,32 @@
...
+	Promise.all( [
+		setupSettings(),
+		createTaxes(),
+		createCoupons(),
+		createProducts(),
+		createShippingMethods(),
+		enablePaymentGateways(),
+	] ).then( ( results ) => {
+		const [ , taxes, coupons, products, shippingMethods ] = results;
+		global.wc = {
```
senadir, line 25: `global` is preserved and can be used inside tests and in the teardown phase.

tests/e2e-tests/config/setup.js
```diff
@@ -0,0 +1,32 @@
...
+	await setupPuppeteer( globalConfig );
+	Promise.all( [
+		setupSettings(),
+		createTaxes(),
+		createCoupons(),
+		createProducts(),
+		createShippingMethods(),
+		enablePaymentGateways(),
+	] ).then( ( results ) => {
+		const [ , taxes, coupons, products, shippingMethods ] = results;
```
senadir, line 24:
``` 
const [ , taxes, coupons, products, shippingMethods ] = results;
```
we start destructuring with a comma, this means we want to ignore the first value in the array, that is `setupSettings` return value `undefined`

senadir:
this seems to work fine for me locally but isn't working on travis, seems entrypoint not running maybe? or not being updated.

senadir:
had a typo 🤦🏻‍♂️

tests/e2e-tests/config/jest.setup.js
```diff
@@ -46,97 +45,48 @@ const OBSERVED_CONSOLE_MESSAGE_TYPES = {
...
-	// local path for sample data that is included with woo.
-	await page.keyboard.type(
-		'wp-content/plugins/woocommerce/sample-data/sample_products.csv'
-	);
-	await page.click( '.wc-actions .button-next' );
-	await page.waitForSelector( '.wc-importer-mapping-table' );
-	await page.select(
-		'.wc-importer-mapping-table tr:nth-child(29) select',
-		''
+	// Visit `/wp-admin/edit.php?post_type=product` so we can see a list of products and delete them.
```
haszari, line 58: Is this comment still accurate?

tests/e2e-tests/config/jest.setup.js
```diff
@@ -46,97 +45,48 @@ const OBSERVED_CONSOLE_MESSAGE_TYPES = {
...
+		await page.waitForSelector( '.wc-importer-mapping-table' );
+		await page.select(
+			'.wc-importer-mapping-table tr:nth-child(29) select',
+			''
+		);
+		await page.click( '.wc-actions .button-next' );
+		await page.waitForXPath(
+			"//*[@class='woocommerce-importer-done' and contains(., 'Import complete! ')]"
+		);
+		await switchUserToTest();
```
haszari, line 88: This code seems to be running the csv import – do we still need this? Is this temporary, until we add more products using the API?

tests/e2e-tests/config/setup.js
haszari: Not sure on the naming of this, could be clearer - `global` is very generic and `wc` is pretty generic too :)

Is this something like `testDataIds` ?

haszari: Ah, just realised this is node `global`.

tests/e2e-tests/config/teardown.js
```diff
@@ -0,0 +1,20 @@
...
+const {
+	deleteTaxes,
+	deleteCoupons,
+	deleteProducts,
+	deleteShippingZones,
+} = require( './fixture-loaders' );
+const { teardown: teardownPuppeteer } = require( 'jest-environment-puppeteer' );
+module.exports = async ( globalConfig ) => {
+	await teardownPuppeteer( globalConfig );
+	const { taxes, coupons, products, shippingMethods } = global.wc;
```
haszari, line 13: Might be worth a comment or two to make it clear that [node `global`](https://nodejs.org/api/all.html#all_global) is used to communicate data between setup and teardown.

Or perhaps it's possible to do this some other way - e.g. use a single module for setup/teardown with module-local state.

tests/e2e-tests/config/setup.js
senadir: TIL,  I assumed it was provided by Jest and not node, but yeah `global` is something we don't control. wc is just a placeholder, there is no risk of this data touching anything else, this global in particular (this execution context) is not shared with tests or other things.

senadir:
> Note - here are the steps I used for testing – the steps in the PR were a bit light.

Good point, I will update the PR.

tests/e2e-tests/config/fixture-loaders.js
senadir: I think we can, I was testing using node directly, which doesn’t have import, jest have babel support, but I'm not sure if it is enabled on globalSetup and globalTeardown, will test.

tests/e2e-tests/config/fixture-loaders.js
senadir: good point.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+			free_shipping: true,
+		},
+	];
+	return WooCommerce.post( 'coupons/batch', { create: coupons } )
+		.then( ( response ) =>
+			response.data.create.map( ( coupon ) => coupon.id )
+		)
+		.catch( ( e ) => console.log( e.response.data.message ) );
+};
+const deleteCoupons = ( ids ) => {
```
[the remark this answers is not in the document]
senadir: yes, linter didn't seem to bother I think, I kept 1 empty line between unrelated functions.

tests/e2e-tests/config/fixture-loaders.js
```diff
@@ -0,0 +1,297 @@
...
+		delete: ids,
+	} ).catch( ( error ) => console.log( error ) );
+};
+// currently this only creates a single product for the sake of review
+// more products will be added later
+const createProducts = () => {
+	return WooCommerce.post( 'products/batch', {
+		create: [
+			{
+				name: 'Woo Single #1',
```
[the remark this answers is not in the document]
senadir: That's a good idea, might move them to preview.

tests/e2e-tests/config/fixture-loaders.js
senadir: the problem is that for your test to work right, you need to know what product has the reviews exactly, all the data is static but the IDs are not static, so we will still need a way to get the id from the product and pass it to reviews, it could be like this in the setup.js file

```
const productsIds = await createProducts();
await createReviews( productIds[ 0 ] );
```

This coupling will get more apparent once we put Categories and Tags in it since you can only assign a category by its ID in the product.

another approach is to query all products inside the review, filter to the product you want, get its ID and pass it to reviews, a similar thing can happen inside Product and Categories.
This will be an extra request but that isn’t an issue when it comes to tests, however, we need to ensure the order of execution, so we will move away from Promise.all to async-await.

tests/e2e-tests/config/jest.setup.js
senadir: No.

Another point, changes in `tests/e2e-tests/config/jest.setup.js` are not done in this PR, they come from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2544, this PR is based on it (I forget to mention that, fixed now), but I'm comparing this PR against master and not that PR (so that travis can run).

tests/e2e-tests/config/jest.setup.js
senadir: Yes, we still need it, it is temporary, read the full explanation here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2544

tests/e2e-tests/config/teardown.js
senadir: > Or perhaps it's possible to do this some other way - e.g. use a single module for setup/teardown with the module-local state.

I'm not sure I fully understand this, setup and teardown must be two functions, global is the only way to share scope between them, any code I declare at the top of the file (outside the function) will be ignored.

senadir:
Okay so I addressed most or all of the comment:

- Added JSDocs to all functions and some documentation here and there 1be9d46
- ~~Used `import` syntax here 7293e23~~ I reverted them to commonJS require in ac1c62eeb, see below why.
- Refactored fixtures into its own file, refactored shipping zone factory to be more dynamic e8cf97f
- Fixture data are a function that returns an array or object of the data, why a function? some data is dynamic and depends on a variable, reviews are an example of that, by doing that, we ensure our data is correct.


Why not use `import`:
- part of developing and writing those functions means testing them locally to see if they work, outside of their normal flow, for that I use `node tests/e2e-tests/fixtures/fixture-loaders.js`, sadly the version of node we use (latest LTS) doesn't support import.

tests/e2e-tests/config/setup.js
haszari: Just to be clear, the name `wc` is still a bit too generic in my opinion - I'd prefer it has a more specific, clearer name. For example `fixtureData` or `testDataIds` (`= { products: array of ids, coupons:  etc }`). Not a blocker.

tests/e2e-tests/config/jest.setup.js
haszari: Ah, that makes it hard to review the code in this PR. Might be a better idea to get #2544 reviewed and merged first so this PR's changes are clear.

So .. is that comment getting deleted?

tests/e2e-tests/config/jest.setup.js
haszari: I took a look at 2544 – I can see there are some changes getting made there, but I didn't understand the intention/goal. I'm guessing this will be clearer when the 2544 is merged :)

tests/e2e-tests/config/teardown.js
haszari: The main feedback here is that we should avoid using global if possible. Not a big deal, probably not a blocker.

By "module" I mean js source file. In node you can have [module state](https://nodejs.org/api/all.html#globals_global) - local variables in a module can be accessed by all functions in the file, but not outside the module.

Right now the setup and teardown functions are each contained within their own modules, I'm not sure if it's practical to move them to a shared module. I see the jest config references them by filename rather than function object. In jest docs it looks like maybe they could maybe be defined in the same file - they are [passed as callbacks to APIs like `beforeAll`](https://jestjs.io/docs/en/setup-teardown#one-time-setup).

There are other ways to share state in JS without using global/window - e.g. using a closure.  For example, define the setup and teardown and their state in a closure:

```js
export function fixtureDataSetup() {
   var fixtureDataIds = {};

   function setup() {
     // make products etc & save ids
    fixtureDataIds.products = ...;
  }
  
  function teardown() {
     // delete stuff
     removeProducts( fixtureDataIds.products );
  } 

  return { 
     setup,
     teardown,
  };
}

// And then out in the main / config you'd tell Jest to run `setup` and `teardown` as global setup/teardown func.
```

This might be overkill - just exploring options. I could imagine a world where the test data ids would be useful to tests - e.g. a test that submits a review for a particular product id. In that case, the closure variable could be exposed (directly, or via getter funcs etc).

If we stick with the global I think it's a good idea to add a big comment explaining your intentions for it – global variables need big comments :)

package.json
```diff
@@ -83,6 +83,7 @@
 		"@types/react": "16.9.35",
 		"@types/wordpress__data": "4.6.7",
 		"@types/wordpress__element": "2.4.1",
+		"@woocommerce/woocommerce-rest-api": "^1.0.1",
```
haszari, line 86: Probably want to pin this version (remove `^`) - RenovateBot will pin it anyway, save a PR.

tests/e2e-tests/fixtures/fixture-data.js
```diff
@@ -0,0 +1,192 @@
...
+	},
+	{
+		code: 'freeshipping',
+		discount_type: 'percent',
+		amount: '0',
+		free_shipping: true,
+	},
+];
+
+const Reviews = ( id ) => [
```
haszari, line 40: If this is dynamic, let's name it like a regular function - `addReviewsToProduct` or something like that.

tests/e2e-tests/fixtures/fixture-data.js
```diff
@@ -0,0 +1,192 @@
+const Coupons = () => [
```
haszari, line 1: Might be worth documenting here that the shape of all this data is dictated by the relevant APIs.

Not sure if its worthwhile making them all functions – most are not dynamic. (Not a blocker just a thought)

haszari:
Looking good @senadir ! I've added a couple more comments with some ideas, but I'm happy for this to merge if you're confident. Note with a lot of this discussion I'm mostly trying to better understand your intentions/plans for how we wrangle our e2e test infrastructure.

If you prefer I'm happy to give this another review after #2544 - or let me know if you want me to review #2544 directly :)

tests/e2e-tests/fixtures/fixture-data.js
senadir: However we don't know when or how this data might get dynamic, some coupons might need certain products IDs or Categories, shipping might need a class that is created in Products, Products will need reviews, there's always overlap, so I wanted to keep the interface the same, so all fixtures are a function, this should make using them more predictable.

tests/e2e-tests/fixtures/fixture-data.js
senadir: it's the one responsible for adding reviews, the ones in `fixture-loaders` are already named like that, however `ReviewsInProduct` or `ReviewsToProduct` would make sense.

tests/e2e-tests/config/teardown.js
senadir: jest expects setup and teardown to be the default export on that file, there's a lot of care of why they shouldn't context, I can see your example and I think it might work but I do believe its a bit overkill for this exact scenario (and would make code readability a bit jumpy between three files).

In the context of Jest, teardown and setup run on a separate context compared to actual test and code, there is no sharing, I can't access the same global nor can I access the same `.wc` variable, so in that small context, I think it's really safe.
Using global is the recommended way to handle shared data in Jest documentation when it comes to those two functions.

senadir: >  test that submits a review for a particular product id.

This is not really possible with how jest work since this data is not accessible to it, I was thinking of ways we can circulate around this, in the rare cases we need such a thing, we can initiate our fixture from within the test, so, inside a `beforeAll`, this will require us to change the way our loaders work (by receiving fixtures as a prop).

senadir:
Once travis pass in this PR, I will rebase it against #2544 so you can review it better, I still think you should review #2544 first.

senadir:
so in the last update, I added:

- Docs to all fixtures data 57e632f
- changed fixture loaders to accept fixture data as a param, you can reuse the same functions inside individual tests to create custom data for that test dd51a2f
- renamed `global.wc` to `global.fixtureData` and added a comment in the `setup.js` file. dd0a7d2

senadir:
seems like I have merged this by mistake, still up for review @nerrad

PR #2549, opened 2020-05-24
Run e2e tests serially in travis to increase speed

senadir:
Based on [this piece of docs in Jest](https://jestjs.io/docs/en/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server) and[ jest-puppeteer](https://github.com/smooth-code/jest-puppeteer#running-puppeteer-in-ci-environments) running tests one at a team will actually increase the speed and reduce the timeout since e2e servers are different from your local SSD machine.

steps to test:
- Travis should pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ jobs:
               - npm run test:e2e:up
               - composer install
               - npm run build:e2e-test
-              - npm run test:e2e
+              - npm run test:e2e:ci
           env:
               - WOOCOMMERCE_BLOCKS_PHASE=experimental
         - stage: deploy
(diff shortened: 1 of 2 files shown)


senadir:
It seems we already limit the number of workers in our tests https://github.com/woocommerce/woocommerce-gutenberg-products-block//blob/857b32a57a06995b50f8a62af8115165399b99d2/tests/bin/e2e-test-integration.js#L27 so this is not needed.

Maybe we can remove the worker limit and only have them in E2E to increase the speed locally?
closing anyway for now, feel free to reopen if you have other thoughts.

PR #2551, opened 2020-05-25
Fix Products by Category JSON response error

Aljullu:
I'm not sure what are the exact steps to reproduce this issue, I found it in a random test site I have but everything was working fine in all others. After adding the _Products by Category_ block, the list of categories doesn't appear and instead it shows an error (`The response is not a valid JSON response.`):
imatge

Inspecting the JSON response, it's like this:
```JSON
Warning: intval() expects at most 2 parameters, 3 given in .../wp-includes/rest-api/class-wp-rest-request.php on line 804
[{"id":21,"name":"Accessories","slug":"accessories","description":"","parent":18,"count":5,"image":null,"review_count":0, ...}, ...]
```

After this PR, everything works fine again:
imatge

Tagging this as `blocker` because I was unable to reproduce it in 2.5.15, so it must be a regression in 2.6.

1. Add a _Products by Category_ block and verify there are no errors and you can select a category.

### DIFF
--- a/src/StoreApi/Routes/AbstractTermsRoute.php
+++ b/src/StoreApi/Routes/AbstractTermsRoute.php
@@ -163,7 +163,7 @@ abstract class AbstractTermsRoute extends AbstractRoute {
 	 * @param mixed $value Value to convert to an int.
 	 * @return int
 	 */
-	protected function get_intval( $value ) {
+	public function get_intval( $value ) {
 		return intval( $value, 10 );
 	}
 }
--- a/src/StoreApi/Routes/AbstractTermsRoute.php
+++ b/src/StoreApi/Routes/AbstractTermsRoute.php
@@ -41,7 +41,7 @@ abstract class AbstractTermsRoute extends AbstractRoute {
 			'type'              => 'integer',
 			'minimum'           => 0,
 			'maximum'           => 100,
-			'sanitize_callback' => [$this, 'get_intval' ],
+			'sanitize_callback' => [ $this, 'get_intval' ],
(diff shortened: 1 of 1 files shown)


mikejolley:
@Aljullu If this is never set to -1 for any reason we could/should use `absint` to avoid rolling our own methods https://developer.wordpress.org/reference/functions/absint/

Aljullu:
@mikejolley the reason that we opted for our own wrapping method instead of `absint` was that we weren't sure if -1 was ever used so we wanted to play it save to avoid breaking any current usage.

Looking at the code, I can't spot any obvious place were `per_page: -1` is set, but might be worth double-checking by somebody else more used to the PHP-side of the codebase. :slightly_smiling_face: 

But even if -1 is not currently used in our codebase, does it make sense to restrict it? What are the downsides of using our own methods for validation?

mikejolley:
@Aljullu 

> even if -1 is not currently used in our codebase, does it make sense to restrict it? 

yes it most certainly does make sense to restrict to valid values. The terms API uses 0 for unlimited results rather than -1, and I recall making all endpoints follow this rather than allowing for -1 values. The min is also set to 0 in the validation schema.

nerrad:
Sorry Albert, I should have thought to look at the schema, I agree with Mike here, regardless of what happens in the models, the schema for the API has 0 as the minimum value so it looks like `absint` is the way to go here (WP uses it internally on the `number` value too for the Term Query).

Aljullu:
Makes sense, thanks for the input @mikejolley @nerrad. I will push a PR fixing this today. :+1:

PR #2553, opened 2020-05-25
Improve readme.txt with note about nature of feature plugin

nerrad:
The readme.txt file parsed for `wp.org` does not have any clear indication that it is a Feature plugin. I think it's important we highlight:

- It is a feature plugin where we experiment/iterate on new blocks and updates to existing blocks before they land in WooCommerce core.
- Only folks comfortable with using the bleeding edge of blocks should use this plugin.
- The _stable_ blocks are bundled with WooCommerce core

I also took the opportunity to update the README.md content in here to align with what is in readm.txt.

### DIFF
--- a/readme.txt
+++ b/readme.txt
@@ -12,7 +12,9 @@ License URI: https://www.gnu.org/licenses/gpl-3.0.html
 
 WooCommerce Blocks are the easiest, most flexible way to display your products on posts and pages! This
 
-**Note: This is the feature plugin for WooCommerce Blocks which has the bleeding edge new blocks and updates to existing blocks for wider usage and testing before we release them in WooCommerce Core.**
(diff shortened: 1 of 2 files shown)


readme.txt
```diff
@@ -1,16 +1,20 @@
...
+Tested up to: 5.4
 Requires PHP: 5.6
 Stable tag: 2.6.0-rc1
 License: GPLv3
 License URI: https://www.gnu.org/licenses/gpl-3.0.html
 
 == Description ==
 
-WooCommerce Blocks are the easiest, most flexible way to display your products on posts and pages!
+WooCommerce Blocks are the easiest, most flexible way to display your products on posts and pages! This
```
Aljullu, line 13:
```suggestion
WooCommerce Blocks are the easiest, most flexible way to display your products on posts and pages!
```

PR #2563, opened 2020-05-26
Use absint for per-page param validation

Aljullu:
Second take of #2551.


nerrad:
I'll cherry pick this into 2.6.0

PR #2566, opened 2020-05-26
Updates for the abstract block class

mikejolley:
_[#2399](_https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/) is a mammoth PR so I'm breaking out some fixes and changes to review separately._

This updates the Abstract Block class to:

1. Allow the block name to be added via the constructor. I will need this to register Atomic Blocks later on to allow class reuse, otherwise I'd need a class per atomic block.
2. Fixes a white space issue in the attributes replacement.
3. Fixes array/object handling - they need converting to JSON before appending to the HTML.

These are under the hood changes for future use, so nothing specific to test against so long as tests pass and no obvious regressions occur.

### DIFF
--- a/src/BlockTypes/AbstractBlock.php
+++ b/src/BlockTypes/AbstractBlock.php
@@ -37,8 +37,13 @@ abstract class AbstractBlock {
 
 	/**
 	 * Constructor
+	 *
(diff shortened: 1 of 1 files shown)


senadir:
Providing dynamic block names make sense to me.

PR #2568, opened 2020-05-26
Create shared contexts handle for inner blocks

mikejolley:

Background behind these changes can be seen in this comment from @nerrad https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2399#issuecomment-631560641

Changes:

- Shared contexts so Atomic blocks are functional (see #2399 (comment)) This ensures contexts work when wrapping atomic blocks in the editor and on frontend.
- Combines useProductLayoutContext and useInnerBlockConfigurationContext since both are used at the same time in all contexts. This allows the parent block to be passed, as well as a prefix for the CSS classnames.
- Introduces ProductDataContext to share product data with atomic blocks. This is not in use yet; it is for the single product block.

This isn't in use by multiple blocks yet, so to test please just Smoke Test the All Products block and ensure there are no regressions in the layout (front and backend).


mikejolley:
... @haszari that error is identical to the one you posted in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2441 so it seems clear to me it's unrelated to my changes. 

> I'd really appreciate a bit more rationale/detail about the intention in this PR description. I realise that this may be covered in the linked comment but it's easy to get rabbit holed

I only linked to a single comment, not a rabbit hole. If the linked comment wasn't clear I would have pasted the key bits but I'm not sure thats the case here. I;m not sure what more explanation I can offer on top of the comment which is pretty clear.

> This is a pretty mammoth change on its own!

¯\_(ツ)_/¯

haszari:
> Sounds like you were just derailed by the error message?

Yep totally – apologies, I guess this is a risk when we have known issues like this. 

Thanks for highlighting the "context" (pun intended!) - still not fully clear on the details but I'll do a general smoke test of All Products & friends. I see you have @nerrad tagged to review too - I'm sure he'll give this a good code review.

mikejolley:
> I'm not familiar with InnerBlockConfigurationProvider, so need to research to understand what it provides

I think we'll need to document these more after Inner Blocks work is concluded. There is a readme file for each context however. This one provides a classname for contained inner block components (so they all get the same prefix) and a parent block name.

> we ended up with two contexts for different kinds of block.

Yup :)

> Is there a pattern/convention we could use to prevent/detect issues like this happening in future? How do you spot this kind of problem in a review, or how would it show up as bugs in e2e or manual testing?

It's difficult - I spotted this when my context values were not picked up by the nested components. This occurs because each frontend block has it's own script file.

PR #2574, opened 2020-05-27
Delete orders using delete method in batches of 20

### LINKED ISSUE #2570: Expose deleted draft order ids to plugins that might need them.
I experienced a bug when I visited the Woo Admin order list table route (`?page=wc-admin&path=%2Fanalytics%2Forders`) to see how our new draft status orders (introduced with the Checkout block) appear there. You can read up on my report of this bug and how to reproduce  in the [issue I created in the WC-Admin repo](https://github.com/woocommerce/woocommerce-admin/issues/4447).  Please _do read the report there_ because it provides not only the bug that was happening (a fatal that borks the view) but also some context around _why_ it was happening that is also background for this issue.

I've also created an issue in our repo because this is a blocker to releasing since even if WC-Admin adds defensive checking, it's just a bandaid over the issue and won't resolve cases where Woo Blocks 2.6.0 is used in earlier versions of WooCommerce core that have WC-Admin active or a site has the WC-Admin feature plugin active.

Also, it exposes a flaw that could surface with _any_ extension that syncs any sort of custom table with the order table (or order meta etc).

The flaw is that those extensions (or WC-Admin) have no awareness when draft status orders are _deleted_ from the database and thus are not able to update their own tables accordingly.

I haven't taken a dive into WooCommerce code yet (because I discovered this way past my work hours and I'm tired...just wanted to get this documented :) ) but I'm assuming there's an action hook that could be fired when an order is deleted that we could surface in our draft status cleanup logic (and we just make sure we grab the deleted ids before deleting to surface via the hook).

So there's a couple things we need to do before we can release 2.6.0:

## Prevent fatals happening in earlier versions of WooCommerce core (or installs with WC-Admin plugin active for Woo Core 4.0 environments).

There's a variety of options here:

- in our code we manually delete from the `wc_order_stats` table as a part of our cleanup (if it exists).
- It looks like woo-admin [hooks into `delete_post` to capture when orders are deleted](https://github.com/woocommerce/woocommerce-admin/blob/714660f67ee03ba94bc0febf30d2acbfe22bca7d/src/API/Reports/Orders/Stats/DataStore.php#L110-L112). So we could just fire off that hook for each order that was deleted (advantage here is that in theory that should work for any other extension hooking into that hook for the same reasons).

## Add in a bulk action for extensions (including wc-admin too hook into) for doing bulk deletes after this cleanup.

I'm not sure how feasible this is or whether we should just fire the `delete_post` hook here?

mikejolley:
Removes the custom SQL query and replaces it with a batch delete of 20 orders per run. Schedules more events if the number of delete orders matches the batch limit.

Fixes #2570

1. Create draft orders
2. Run job manually in Tools > Scheduled actions. 
3. Change order dates, re-run to ensure orders are deleted.

I've tested with a handful of orders. Currently testing with > 20 (will follow up with results).

### DIFF
--- a/src/Library.php
+++ b/src/Library.php
@@ -126,22 +126,30 @@ class Library {
 	}
 
 	/**
-	 * Delete draft orders older than a day.
+	 * Delete draft orders older than a day in batches of 20.
 	 *
 	 * Ran on a daily cron schedule.
 	 */
 	public static function delete_expired_draft_orders() {
-		global $wpdb;
-
-		$wpdb->query(
-			"
-			DELETE posts, term_relationships, postmeta
-			FROM $wpdb->posts posts
-			LEFT JOIN $wpdb->term_relationships term_relationships ON ( posts.ID = term_relationships.object_id )
-			LEFT JOIN $wpdb->postmeta postmeta ON ( posts.ID = postmeta.post_id )
-			WHERE posts.post_status = 'wc-checkout-draft'
-			AND posts.post_modified <= ( NOW() - INTERVAL 1 DAY )
-			"
+		$orders = wc_get_orders(
+			[
+				'date_modified' => '<' . strtotime( '-1 DAY' ),
+				'limit'         => 20,
+				'status'        => 'wc-checkout-draft',
+				'type'          => 'shop_order',
+			]
 		);
+		$count  = 0;
+
+		if ( $orders ) {
+			foreach ( $orders as $order ) {
+				$order->delete( false );
+				$count ++;
+			}
+		}
+
+		if ( 20 === $count && function_exists( 'as_enqueue_async_action' ) ) {
+			as_enqueue_async_action( 'woocommerce_cleanup_draft_orders' );
+		}
 	}
 }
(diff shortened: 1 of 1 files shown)


src/Library.php
```diff
@@ -126,22 +126,30 @@ public static function append_draft_order_post_status( $statuses ) {
...
+
+		if ( $orders ) {
+			foreach ( $orders as $order ) {
+				$order->delete( false );
+				$count ++;
+			}
+		}
+
+		if ( 20 === $count && function_exists( 'as_enqueue_async_action' ) ) {
+			as_enqueue_async_action( 'woocommerce_cleanup_draft_orders' );
```
nerrad, line 152: Do we want this running as soon as possible, or should we stagger it a bit?

mikejolley: This is equivalent to using using `time()` or `time() + seconds`. If I recall, AS will manage memory and resource limits but I'll double check.

nerrad: Mike shared this with me in slack which is a thing I learned about AS, so we should be okay here:

>  If there are actions to be processed, Action Scheduler will stake a unique claim for a batch of 25 actions and begin processing that batch. The PHP process spawned to run the batch will then continue processing batches of 25 actions until it uses 90% of available memory or has been processing for 30 seconds.

PR #2577, opened 2020-05-27
Single Product Block Edit Mode

mikejolley:

This PR adds the Edit View to the Single Product Block (the product selection part). _This does not include the inner block editor._

While I was adding the block preview, I cleaned up some other resource images to reduce bundle size.

### Screenshots

Edit Page ‹ Local Testing — WordPress 2020-05-27 16-47-47

1. Use the inserted and find single product block
2. Ensure there is a skeleton like block preview
3. Insert the block. Test the product selection and click done. The block should go blank (expected at this point).
4. Check the inspector. When you're in the selected state, you should see a "Product" Section. This should contain an instance of the product selector.


assets/js/previews/single-product-block.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * External dependencies
+ */
+import { WC_BLOCKS_ASSET_URL } from '@woocommerce/block-settings';
+
+export const singleProductBlockPreview = (
+	<img
+		src={ WC_BLOCKS_ASSET_URL + 'img/single-product-preview.svg' }
+		alt="Single Product Preview"
```
Aljullu, line 9: Should this be localized? I'm not sure if it's possible to navigate to this image with assistive technology, if not, I think it would be just ok to leave `alt` blank.

PD: I noticed the alt text for the `Checkout` and `Product grid` previews is not localizable either, so we might handle all of them in a follow-up if it's easier.

assets/js/blocks/single-product/edit/editor.scss
```diff
@@ -0,0 +1,17 @@
+.wc-block-single-product {
+	&.components-placeholder {
+		// Reset the background for the placeholders.
+		background-color: rgba(139, 139, 150, 0.1);
```
Aljullu, line 4: Why setting the background here? Couldn't it just use the inherited one?

assets/js/blocks/single-product/edit/editor.scss
```diff
@@ -0,0 +1,17 @@
...
+	}
+
+	.components-resizable-box__handle {
```
Aljullu, line 7: Where is this element rendered? I couldn't find it.

assets/js/blocks/single-product/edit/editor.scss
```diff
@@ -0,0 +1,17 @@
...
+		z-index: 10;
+	}
+}
+.wc-block-single-product__message {
```
Aljullu, line 11: And I couldn't find this one either. :confused:

mikejolley: I copy pasted from another block - not relevant here.

PR #2585, opened 2020-05-28
Add "edit product" card to single product block sidebar

### LINKED ISSUE #2519: Show a link to edit product details in the inspector for single products
This should be displayed only after a selection has been made:

It should open a new window.

mikejolley:
Adds a card which links to the edit product page to the sidebar of the Single Product Block.

Fixes #2519

Some slight deviations from the design (known):

- Shown above the advanced tab due to no available slot fill below it
- Link text has underline for admin consistency
- Link icon is different due to us already having this icon in our blocks

### Screenshots

Edit Page ‹ Local Testing — WordPress 2020-05-28 13-46-11

1. Insert single product block.
2. Select a product.
3. View the sidebar; you should see the new component.
4. Check the link goes to the correct place in a new tab

### DIFF
--- a/assets/js/blocks/single-product/edit/edit-product-link.js
+++ b/assets/js/blocks/single-product/edit/edit-product-link.js
@@ -4,36 +4,33 @@
 import { __ } from '@wordpress/i18n';
 import { Icon, external } from '@woocommerce/icons';
 import { ADMIN_URL } from '@woocommerce/settings';
-import { Card, CardBody } from '@wordpress/components';
 
 /**
  * Component to render an edit product link in the sidebar.
  */
 const EditProductLink = ( { productId } ) => {
 	return (
-		<Card className="wc-block-single-product-edit-card">
-			<CardBody>
-				<div className="wc-block-single-product-edit-card-title">
-					<a
-						href={ `${ ADMIN_URL }post.php?post=${ productId }&action=edit` }
-						target="_blank"
-						rel="noopener noreferrer"
-					>
-						{ __(
-							"Edit this product's details",
-							'woo-gutenberg-products-block'
-						) }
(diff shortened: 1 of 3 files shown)


assets/js/blocks/single-product/edit/edit-product-link.js
```diff
@@ -0,0 +1,40 @@
...
+import { Icon, external } from '@woocommerce/icons';
+import { ADMIN_URL } from '@woocommerce/settings';
+import { Card, CardBody } from '@wordpress/components';
+
+/**
+ * Component to render an edit product link in the sidebar.
+ */
+const EditProductLink = ( { productId } ) => {
+	return (
+		<Card className="wc-block-single-product-edit-card">
```
Aljullu, line 14: Should the class name be `wc-block-single-product__edit-card` instead? This also applies to the two class names below. :point_down:

Aljullu: Also, I'm a bit confused. What are the benefits on using a `Card` over a normal `div`? The feedback prompt, which is a component quite similar, is using a `div`.

mikejolley: I saw the Block Name + Description in the inspector uses cards so I followed suit. Should I not?

Aljullu: You mean the `.block-editor-block-card`?

imatge

Even though it's named `card`, I don't think it uses the `<Card>` component internally. At least, if I search for `.components-card` in the inspector I can only see one result (the one introduced in this PR).

mikejolley: @Aljullu Thanks I see now. I just pushed the changes to use DIV.

PR #2586, opened 2020-05-28
Add layout editor to the single product block

mikejolley:
Part of #2394 

Adds the Inner Block Editor to the Single Products Block. Includes:

- The inner block editor itself
- The default block list
- The reset layout button in the sidebar
- New atomic utils to support the new functionality

Known issue follow ups:

- There is no limit on what blocks can be inserted into the columns
- No frontend rendering
- Only sample data is shown for inner blocks. Live data will be supported after implementing context for inner blocks.

### Screenshots

Edit Page ‹ Local Testing — WordPress 2020-05-28 14-39-58

1. Insert the Single Product Block
2. Check inner blocks are rendered
3. Add a block
4. Use then 'reset layout' button in the sidebar and confirm it resets back to default

### DIFF
--- a/assets/js/atomic/utils/create-blocks-from-template.js
+++ b/assets/js/atomic/utils/create-blocks-from-template.js
@@ -1,0 +1,18 @@
+/**
+ * External dependencies
+ */
+import { createBlock } from '@wordpress/blocks';
+
+/**
+ * Creates blocks for a given inner blocks Template.
+ *
+ * @param {Array} template Inner Blocks Template.
+ */
+export const createBlocksFromTemplate = ( template ) => {
+	return template.map( ( [ name, atts = {}, innerBlocks = [] ] ) => {
+		const children = innerBlocks
+			? createBlocksFromTemplate( innerBlocks )
+			: [];
+		return createBlock( name, atts, children );
+	} );
+};
--- a/assets/js/atomic/utils/get-block-map.js
+++ b/assets/js/atomic/utils/get-block-map.js
@@ -1,0 +1,33 @@
+/**
+ * External dependencies
+ */
+import { getRegisteredInnerBlocks } from '@woocommerce/blocks-registry';
+
+/**
+ * Internal dependencies
+ */
+import {
+	ProductTitle,
+	ProductPrice,
+	ProductButton,
+	ProductImage,
+	ProductRating,
+	ProductSummary,
+	ProductSaleBadge,
+} from '../components/product';
+
+/**
+ * Map blocks names to components.
+ *
+ * @param {string} blockName Name of the parent block. Used to get extension children.
+ */
+export const getBlockMap = ( blockName ) => ( {
+	'woocommerce/product-price': ProductPrice,
+	'woocommerce/product-image': ProductImage,
+	'woocommerce/product-title': ProductTitle,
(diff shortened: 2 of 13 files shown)


assets/js/atomic/utils/create-blocks-from-template.js
```diff
@@ -0,0 +1,18 @@
...
+ * Creates blocks for a given inner blocks Template.
+ *
+ * @param {Array} template Inner Blocks Template.
+ */
+export const createBlocksFromTemplate = ( template ) => {
+	return template.map( ( [ name, atts = {}, innerBlocks = [] ] ) => {
+		const children = innerBlocks
+			? createBlocksFromTemplate( innerBlocks )
+			: [];
+		return createBlock( name, atts, children );
```
haszari, line 16: Cool! Were and @nerrad discussing this in slack the other day? 

I'm curious about whether this is standard/expected use of gutenberg APIs or if we have special needs. Might be worth adding a bit more info about this and our expectations for it. Is this something we would prefer to import from Gutenberg? How stable is `createBlock` API?

haszari: Could also add JSDoc typedefs for the expected param shape and the return value.

assets/js/atomic/utils/get-block-map.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+export const getBlockMap = ( blockName ) => ( {
+	'woocommerce/product-price': ProductPrice,
+	'woocommerce/product-image': ProductImage,
+	'woocommerce/product-title': ProductTitle,
+	'woocommerce/product-rating': ProductRating,
+	'woocommerce/product-button': ProductButton,
+	'woocommerce/product-summary': ProductSummary,
+	'woocommerce/product-sale-badge': ProductSaleBadge,
+	...getRegisteredInnerBlocks( blockName ),
```
haszari, line 32: I see this is not new .. I was not aware of this. Do we need a hint/reminder/process to update this if we ever change those blocks? I guess the main component and the block name are both quite stable. This feels like it could potentially be generated from block config. 

None of this is a blocker, just the first time I've seen this woo innerblocks magic :)

Longer term, some of this might be appropriate for Gutenberg APIs - i.e. other blocks plugins might want to allow innerblock editing in a similar manner to Woo Blocks. (I'm guessing this already has been thought about & discussed somewhat!)

mikejolley:
Yup those are known; I already made a list in the main PR instructions.

Thanks

assets/js/atomic/utils/get-block-map.js
mikejolley: Extensibility is a follow up. R/e the map, there is no API, config, or method in Gutenberg to map a block to a component, that's why this has to exist. We're saying if you see this block name, render this component. This is for frontend usage where we use dynamic components rather than saved content like regular gutenberg blocks.

assets/js/atomic/utils/create-blocks-from-template.js
mikejolley: I think createBlock is stable, this code isn't actually new and is in the current production build - I've just moved it so reuse is easier. The difference is that this works recursively rather than over a flat list of blocks.

PR #2587, opened 2020-05-28
Update atomic block structure

mikejolley:

This PR moves some code around and restructures the Atomic Blocks. Why? Because in order to render Inner Blocks without a LayoutConfig, we need some extra data shared between the component and the block (mainly attributes). The PR contains the following changes:

- Moves components from atomic/components/<component>/index to atomic/blocks/<block>/block.js. This also more closely matches other non-atomic blocks
- Adds an alias for `@woocommerce/atomic-blocks` which is used to import the blocks themselves.
- Adds a map of block components so we can map block->component when rendering layouts still
- Splits the atomic block files up into edit.js, attributes.js, from the existing index.js. The code is unchanged, just moved.

I have not yet added context to this components— that will be done in a follow up.

Since this changes structure only, testing should be focussed around All Products block. Ensure existing blocks continue to work (edit the layout), and render correctly on the frontend.


assets/js/atomic/blocks/index.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Internal dependencies
```
mikejolley, line 2: This file is imported by all products block to register the atomic blocks. These imports ensure all atomic blocks are loaded.

assets/js/atomic/blocks/product/block-components.js
```diff
@@ -0,0 +1,7 @@
+export { default as ProductButton } from './button/block';
+export { default as ProductImage } from './image/block';
+export { default as ProductPrice } from './price/block';
+export { default as ProductRating } from './rating/block';
+export { default as ProductSaleBadge } from './sale-badge/block';
+export { default as ProductSummary } from './summary/block';
+export { default as ProductTitle } from './title/block';
```
mikejolley, line 7: This file is used in the block->component map. We map block name to the block component itself.

assets/js/atomic/blocks/product/button/edit.js
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Internal dependencies
+ */
+import Block from './block';
+
+export default ( { attributes } ) => {
+	return (
+		<Disabled>
+			<Block { ...attributes } />
```
mikejolley, line 14: Passing of attributes here is temporary; when we implement context there will be no need to pass attribute data.

mikejolley:
@Aljullu it was the lightBlockWrapper, I've removed it.

PR #2591, opened 2020-05-29
Expand Feature flags

### LINKED ISSUE #2451: Tweak Experimental Flag feature gating to allow for holding back from WC core.
While we do currently have feature gating in place for features in the blocks plugin, it's just a binary flag. This means that anything we choose to expose on releases of the plugin, will also get release if that tag is included in WC core. 

However, I can see the use case where we may want things _only_ available in the plugin and not in what gets bundled with WC core (for instance giving the Cart and Checkout blocks some extended time in the plugin before bundling with WC core).

So this issue is for implementing improvements to the feature flag system so that we can select what "stage" a feature is (i.e. only dev, plugin only, everywhere).

senadir:
closes #2451 

This PR changes and expand our feature flags to make them more dynamic and to work with the core/plugin/dev model, so it does the following:

- It changes our flags names from `stable` and `experimental` to numbers, this is a requirement so we can have collapsable flags, so we have 3 flags:
    - `experimental` is now `3`.
    - `plugin` is now `2`.
    - `core` is now `1`.
 for most cases you always compare up, this means if you want something to work only in dev you check if the flag is bigger than 2 `WOOCOMMERCE_BLOCKS_PHASE > 2`, if you want something to be in the feature plugin you compare for bigger than 1 `WOOCOMMERCE_BLOCKS_PHASE > 1`.

- It updates our build process slightly, core and stable files and blocks should live `stableMainEntry` and `stableFrontEndEntry`, this means things like Cart, Checkout, and other blocks, new unstable blocks like Single Product should live in `experimentalMainEntry` or `experimentalFrontEndEntry`, those files are only built if you run `npm run build` and `npm start`, they won't be built if you run `npm run deploy`, you can force not building them locally by passing a different flag so `WOOCOMMERCE_BLOCKS_PHASE=2 npm run build`.

- There's now a constant named `WC_BLOCKS_IS_FEATURE_PLUGIN`, it should be defined if we're in running feature plugin and won't be defined when running from core, in most cases, you won't use this constant.

- in PHP you will still use `WOOCOMMERCE_BLOCKS_PHASE` to define your features, same as PHP with same exact flags.

- In JS, you won’t be using `process.env.WOOCOMMERCE_BLOCKS_PHASE` anymore, you will have to use `WOOCOMMERCE_BLOCKS_PHASE` a constant exported from blocks settings

```js
import { WOOCOMMERCE_BLOCKS_PHASE } from '@woocommerce/block-settings';

if ( WOOCOMMERCE_BLOCKS_PHASE > 1 ) {
	registerBlockType( 'woocommerce/checkout', settings );
}
if ( WOOCOMMERCE_BLOCKS_PHASE > 2 ) {
	registerBlockType( 'woocommerce/single-product', settings );
}
```

- `woocommerce/feature-flag` eslint rule is disabled until we merge this and update it.

- Our deployment command `npm run deploy` will now append `WOOCOMMERCE_BLOCKS_PHASE=2` this means when you deploy, it will work on the context of feature plugin, but WooCommrce core will consume it on the context of Core automatically.

## How should you gate your code?

In most cases, you will have code that is:
- safe to run in WooCommerce core, do nothing.
- safe to run in the feature plugin only, it will be either:
    -  A file, wrap the enqueue of that file in a flag.
    - A code block, wrap it in `WOOCOMMERCE_BLOCKS_PHASE`.
- A code that only runs in development or `experimental`, it will be either:
    - A file, wrap the enqueue, this is the most important part.
    - Skip building the file (to reduce the final zip file) by moving it to `experimentalMainEntry` or `experimentalFrontEndEntry`.
   - A code block, wrap it in `WOOCOMMERCE_BLOCKS_PHASE`.
   - A JS code block, if it’s heavy and will affect the final bundle size, we can see into reintroducing dead code elimination (I didn't add this because I wanted to reduce complexity unless we have a use case for it).

## Test instructions:

- Build the plugin normally `npm run build`, you should be able to add any block, including Cart and Checkout and Single Product.

- Build the plugin in feature plugin mode `WOOCOMMERCE_BLOCKS_PHASE=2 npm run build`, you should be able to add any block, including Cart and Checkout but **not Single Product.**

- Build plugin in Core mode `WOOCOMMERCE_BLOCKS_PHASE=1 npm run build`, you should be able to add our core blocks, and **not Cart, Checkout, or Single Product.**

- If you want to further test, you can try to create a zip `WOOCOMMERCE_BLOCKS_PHASE=2 npm run package-plugin` to generate a zip that you can upload to other websites, it would simulate a feature plugin.


src/Package.php
```diff
@@ -44,9 +44,25 @@ protected static function get_package() {
...
+	 *
+	 * We don't need to do any loading or logic here, this will simply
+	 * define the phase as core:1, if we're running a feature plugin, this
+	 * will get redefined to stable:2, later down the execution cycle we will
+	 * check for blocks.ini and either set the flag to this value or the one
+	 * in the feature flag.
+	 *
+	 * @since 2.7.0
+	 */
+	protected static function initiate_feature_flag() {
```
mikejolley, line 62: Since this gets used in define_feature_flag and this is the default, I think we can get rid of this and assume its 1 unless otherwise declared.

woocommerce-gutenberg-products-block.php
```diff
@@ -110,7 +110,20 @@ function() {
...
 
+/**
+ * Reinitialize the feature flag.
+ *
+ * We want to redefine the lower feature flag to 2 if we're running from the feature plugin.
+ * This file (and code) will only run in the plugin and not WooCommerce core.
+ *
+ * @since 2.7.0
+ */
+function woocommerce_blocks_reinitialize_feature_flag() {
```
mikejolley, line 123: You could actually set a constant here instead of using plugins_loaded hooks - we know that if this file is loaded we're in **feature plugin context**. So set the constant to 2 and use that later.

woocommerce-gutenberg-products-block.php
```diff
@@ -110,7 +110,20 @@ function() {
...
+	global $woocommerce_blocks_phase;
+	$woocommerce_blocks_phase = '2';
```
mikejolley, line 125: Incidentally, I understand we're using numeric values here but I think it would be much much more clear to consumers if we used named contexts e.g. WOOCOMMERCE_BLOCKS_CONTEXT = feature_plugin or package?

src/Domain/Bootstrap.php
```diff
@@ -130,17 +130,21 @@ function() {
...
-			if ( file_exists( __DIR__ . '/../../blocks.ini' ) ) {
-				$woo_options = parse_ini_file( __DIR__ . '/../../blocks.ini' );
-				$flag        = is_array( $woo_options ) && 'experimental' === $woo_options['woocommerce_blocks_phase'] ? 'experimental' : 'stable';
-			} else {
-				$flag = 'stable';
-			}
-		}
-		define( 'WOOCOMMERCE_BLOCKS_PHASE', $flag );
+		add_action(
+			'plugins_loaded',
```
mikejolley, line 134: What are we waiting for exactly? I don't think we need to wait - we just need to see if a global/constant is defined by this point and derive our context.

src/Domain/Bootstrap.php
```diff
@@ -130,17 +130,21 @@ function() {
...
+			function() {
+				global $woocommerce_blocks_phase;
+				$allowed_flags = [ '1', '2', '3' ];
+				if ( file_exists( __DIR__ . '/../../blocks.ini' ) ) {
```
mikejolley, line 138: This always runs; so we can keep this here and then set phase based on context; package or feature plugin.

src/Domain/Bootstrap.php
senadir: we need to wait for the code `woocommerce-gutenberg-products-block.php` to run, since this file runs before it, alternatively we can the code in `woocommerce-gutenberg-products-block.php` at priority 9 and run this without a hook

senadir: or even further avoid hooks all together as you said, took me some time to understand it 😅

assets/js/blocks/cart-checkout/cart/index.js
```diff
@@ -51,4 +51,6 @@ const settings = {
 	},
 };
 
-registerBlockType( 'woocommerce/cart', settings );
+if ( WOOCOMMERCE_BLOCKS_PHASE > 1 ) {
```
mikejolley, line 54: Do you think it would be useful to add a util function for this that supports named properties? e.g. `is_build( 'experimental' )`

assets/js/blocks/cart-checkout/cart/index.js
```diff
@@ -51,4 +51,6 @@ const settings = {
...
+	registerBlockType( 'woocommerce/cart', settings );
```
mikejolley, line 55: We could instead add a wrapper around `registerBlockType ` that supports a phase argument. It might make registration easier and improve DRY, and remove the need to import the settings everywhere.

bin/webpack-helpers.js
```diff
@@ -600,13 +587,6 @@ const getPaymentMethodsExtensionConfig = ( options = {} ) => {
...
-			new DefinePlugin( {
-				// Inject the `WOOCOMMERCE_BLOCKS_PHASE` global, used for feature flagging.
-				'process.env.WOOCOMMERCE_BLOCKS_PHASE': JSON.stringify(
-					// eslint-disable-next-line woocommerce/feature-flag
-					process.env.WOOCOMMERCE_BLOCKS_PHASE || 'experimental'
-				),
-			} ),
 		],
 		resolve,
 	};
```
mikejolley, line 592: Might want to get someone else to review webpack who knows more :)

src/Assets.php
```diff
@@ -81,12 +81,15 @@ public static function register_assets() {
 		self::register_script( 'wc-price-filter', plugins_url( self::get_block_asset_build_path( 'price-filter' ), __DIR__ ), $block_dependencies );
 		self::register_script( 'wc-attribute-filter', plugins_url( self::get_block_asset_build_path( 'attribute-filter' ), __DIR__ ), $block_dependencies );
 		self::register_script( 'wc-active-filters', plugins_url( self::get_block_asset_build_path( 'active-filters' ), __DIR__ ), $block_dependencies );
-		self::register_script( 'wc-checkout-block', plugins_url( self::get_block_asset_build_path( 'checkout' ), __DIR__ ), $block_dependencies );
-		self::register_script( 'wc-cart-block', plugins_url( self::get_block_asset_build_path( 'cart' ), __DIR__ ), $block_dependencies );
 
-		if ( 'experimental' === WOOCOMMERCE_BLOCKS_PHASE ) {
+		if ( WOOCOMMERCE_BLOCKS_PHASE > 2 ) {
```
mikejolley, line 85: Again a corresponding PHP utility function might be useful. It's also clearer to read code if the phases are named - but you don't need to remove the numbered phases from behind the scenes, they are fine.

src/Assets.php
```diff
@@ -81,12 +81,15 @@ public static function register_assets() {
...
 			self::register_script( 'wc-single-product', plugins_url( self::get_block_asset_build_path( 'single-product' ), __DIR__ ), $block_dependencies );
 		}
+
+		if ( WOOCOMMERCE_BLOCKS_PHASE > 1 ) {
+			self::register_script( 'wc-checkout-block', plugins_url( self::get_block_asset_build_path( 'checkout' ), __DIR__ ), $block_dependencies );
+			self::register_script( 'wc-cart-block', plugins_url( self::get_block_asset_build_path( 'cart' ), __DIR__ ), $block_dependencies );
+		}
```
mikejolley, line 92: At some point we should move script registration to the block classes - I always forget to add them here :)

src/Domain/Bootstrap.php
```diff
@@ -130,17 +130,22 @@ function() {
...
-				$flag = 'stable';
-			}
+		if ( defined( 'WC_BLOCKS_IS_FEATURE_PLUGIN' ) ) {
+			$default_flag = '2';
+		} else {
+			$default_flag = '1';
 		}
-		define( 'WOOCOMMERCE_BLOCKS_PHASE', $flag );
+
+		$allowed_flags = [ '1', '2', '3' ];
```
mikejolley, line 139: Is there a reason we use strings here instead of numbers?

src/Domain/Bootstrap.php
```diff
@@ -130,17 +130,22 @@ function() {
...
-				$woo_options = parse_ini_file( __DIR__ . '/../../blocks.ini' );
-				$flag        = is_array( $woo_options ) && 'experimental' === $woo_options['woocommerce_blocks_phase'] ? 'experimental' : 'stable';
-			} else {
-				$flag = 'stable';
-			}
+		if ( defined( 'WC_BLOCKS_IS_FEATURE_PLUGIN' ) ) {
+			$default_flag = '2';
+		} else {
+			$default_flag = '1';
 		}
```
mikejolley, line 137:
```suggestion
		$default_flag = defined( 'WC_BLOCKS_IS_FEATURE_PLUGIN' ) ? '2' : '1';
```

woocommerce-gutenberg-products-block.php
```diff
@@ -20,6 +20,7 @@
 
 $minimum_wp_version = '5.2';
 
+define( 'WC_BLOCKS_IS_FEATURE_PLUGIN', true );
```
mikejolley, line 23: Maybe check if this is defined first just in case.

assets/js/blocks/cart-checkout/cart/index.js
senadir: yeah now that we don’t have dead code elimination this could be useful, and it could help enforce tighter guidelines like throwing an error if the flag you provided doesn’t exist.

src/Domain/Bootstrap.php
senadir: Environment variables and ini values are treated as strings, to avoid casting and recasting things all over the place, I assume everything is a string and when I define the flag I turn it to an int.

assets/js/blocks/cart-checkout/cart/index.js
nerrad: I think if we're going to add convenience utility functions, it'd be nicer to do something like `isExperimentalBuild()` or `isStableBuild()`.

assets/js/blocks/cart-checkout/cart/index.js
nerrad: Love this! Maybe `registerGatedBlockType`?

bin/webpack-helpers.js
```diff
@@ -150,16 +149,17 @@ const stableMainEntry = {
...
-	cart: './assets/js/blocks/cart-checkout/cart/index.js',
-	checkout: './assets/js/blocks/cart-checkout/checkout/index.js',
+	'single-product': './assets/js/blocks/single-product/index.js',
 };
 
 const mainEntry =
-	process.env.WOOCOMMERCE_BLOCKS_PHASE === 'stable'
+	// env variables are strings, so we compare against a string, so we need to parse it.
+	parseInt( process.env.WOOCOMMERCE_BLOCKS_PHASE, 10 ) < 3
 		? stableMainEntry
```
nerrad, line 163: So this means that we still have dead code removal on _releases_ for anything marked experimental right? Might be a good idea to also check that `process.env.WOOCOMMERCE_BLOCKS_PHASE` is defined before using here? If it's undefined won't this cause a ReferenceError (especially since it looks like we no longer define a default for `process.env.WOOCOMMERCE_BLOCKS_PHASE`)?

src/Assets.php
nerrad: Agreed, there's a bunch of things I'd like to do around that eventually (including incorporating using `block.json` files which can contain all the things around the blocks and thus make the initialization code more dynamic).

woocommerce-gutenberg-products-block.php
nerrad: Agreed! That would also allow for overriding in `wp-config.php` if necessary.

bin/webpack-helpers.js
senadir: It shouldn’t cause a Reference Error because `process.env` is always defined, `WOOCOMMERCE_BLOCKS_PHASE` not being defined will result in experimental files being generated which is safer than them not being generated right?

assets/js/blocks/cart-checkout/cart/index.js
senadir: where should those methods live 🤔

bin/webpack-helpers.js
nerrad: I am okay with experimental files being built if `WOOCOMMERCE_BLOCKS_PHASE` is not defined, that's not what I'm calling out here. I thought we might have an issue if `parseInt` returned `NaN` but it looks like we won't:

Image 2020-06-02 at 3 32 39 PM

So all is okay here.

assets/js/blocks/cart-checkout/cart/index.js
nerrad: > where should those methods live

Good question! Maybe block settings (since that's where `WOOCOMMERCE_BLOCKS_PHASE` comes from)?

senadir:
This is ready for another review now.

senadir:
The last commits adds:

- a check for `WC_BLOCKS_IS_FEATURE_PLUGIN` before defining it.
- some JS helpers like
```js
registerExperimentalBlockType( name, settings ) // returns the block or undefined
registerFeaturePluginBlockType( name, settings ) // returns the block or undefined
isExperimentalBuild() // returns Boolean
isFeaturePluginBuild() // returns Boolean
```

it also adds helpers in PHP

```php
Package::is_experimental_build() // returns Boolean
Package::is_feature_plugin_build() // returns Boolean
```

assets/js/settings/blocks/feature-flags.js
```diff
@@ -0,0 +1,57 @@
+/**
```
nerrad, line 1: 👍 on putting these in their own file (and the name is great)!

assets/js/settings/blocks/feature-flags.js
```diff
@@ -0,0 +1,57 @@
...
+ * External dependencies
+ */
+import { registerBlockType } from '@wordpress/blocks';
```
nerrad, line 4: One concern I have here is that we are now loading this api on the frontend for anything using block-settings right? is that problematic? I think I might have been the one who suggested putting this api in block-settings so that could be my fault here.

It doesn't look like it's affected build sizes significantly, so maybe this would be okay (since `@woocommerce/block-settings` is essentially just an alias for the path and thus it should be tree-shakeable).

assets/js/settings/blocks/feature-flags.js
```diff
@@ -0,0 +1,57 @@
...
+ * Registers a new experimental block provided a unique name and an object defining its
+ * behavior. Once registered, the block is made available as an option to any
+ * editor interface where blocks are implemented.
+ *
+ * @param {string} name     Block name.
+ * @param {Object} settings Block settings.
+ *
+ * @return {?Object} The block, if it has been successfully registered;
+ *                    otherwise `undefined`.
+ */
```
nerrad, line 21: Thanks for being so thorough with the comment blocks here 👏 !

src/Package.php
```diff
@@ -65,6 +65,23 @@ public static function get_path() {
...
+	}
+
+	/**
+	 * Checks if we're executing the code in an feature plugin or experimental build mode.
+	 *
+	 * @return boolean
+	 */
+	public static function is_feature_plugin_build() {
+		return WOOCOMMERCE_BLOCKS_PHASE > 1;
+	}
```
nerrad, line 84: Please add the code in `src/Domain/Package.php` and then this file will just wrap the calls to the methods there (similar to other methods in this class).

It may seem odd but there's a transitory refactor in process that's not complete yet. `src/Package.php` is required because of the current way things are implemented in WooCommerce core (and that might never change), but we're transitioning to using DI in the blocks code (code debt!) where `src/Package` will eventually _only_ be used for the package registration (and kept mostly for back-compat) in WooCommerce core and `src/Domain/Package` will be injected wherever it's methods are needed.

assets/js/settings/blocks/feature-flags.js
senadir: I had similar conerns, but I'm assuming this is tree shakeable, bundle bot didn’t report any extra weight so It’s safe.

PR #2592, opened 2020-05-29
Add context support to Atomic Blocks

mikejolley:
#2399 is a mammoth PR so I'm breaking out some fixes and changes to review separately.

This PR makes our Atomic Blocks context ready. The context support is backwards compatible with the old way of passing a product as a prop - that takes priority if used.

Changes:

- Refactored Atomic Blocks to not break when no product is given - this is important whilst context is loaded.
- Renamed `useProductDataContextContext` to `useProductDataContext`
- Renamed `useInnerBlockConfigurationContext` to `useInnerBlockLayoutContext` to better reflect its use.
- Added context handling to each Atomic Block
- Removed default "preview" data and instead injected it via the new context!

1. Using All Products Block, check there are no regressions in the layout edit view, in particular, check the preview data is shown correctly since this is now coming from context.
2. Check frontend view of All Products for regressions. Add to cart buttons etc.

That should cover it. Nothing else is using context except the layout edit preview data, so just testing all works as before is all thats needed.

### DIFF
--- a/assets/js/atomic/blocks/product/image/block.js
+++ b/assets/js/atomic/blocks/product/image/block.js
@@ -34,10 +34,12 @@ const ProductImage = ( {
 	saleBadgeAlign = 'right',
 	...props
 } ) => {
-	const productDataContext = { ...useProductDataContext(), ...props };
+	const productDataContext = useProductDataContext();
+	const { product } = productDataContext || props;
+
 	const { layoutStyleClassPrefix } = useInnerBlockLayoutContext();
-	const { product } = productDataContext;
 	const componentClass = `${ layoutStyleClassPrefix }__product-image`;
+
 	const [ imageLoaded, setImageLoaded ] = useState( false );
 
 	if ( ! product ) {
--- a/assets/js/atomic/blocks/product/image/block.js
+++ b/assets/js/atomic/blocks/product/image/block.js
@@ -35,7 +35,7 @@ const ProductImage = ( {
 	...props
 } ) => {
 	const productDataContext = useProductDataContext();
-	const { product } = productDataContext || props;
+	const product = props.product || productDataContext.product;
 
 	const { layoutStyleClassPrefix } = useInnerBlockLayoutContext();
 	const componentClass = `${ layoutStyleClassPrefix }__product-image`;
--- a/assets/js/atomic/blocks/product/image/block.js
+++ b/assets/js/atomic/blocks/product/image/block.js
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types';
 import { Fragment, useState } from '@wordpress/element';
 import classnames from 'classnames';
 import { PLACEHOLDER_IMG_SRC } from '@woocommerce/block-settings';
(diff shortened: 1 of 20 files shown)


assets/js/blocks/products/all-products/edit.js
```diff
@@ -203,7 +207,11 @@ class Editor extends Component {
 					<div className="wc-block-grid has-1-columns">
 						<ul className="wc-block-grid__products">
 							<li className="wc-block-grid__product">
-								<InnerBlocks { ...InnerBlockProps } />
+								<ProductDataContextProvider
```
mikejolley, line 210: Preview data is passed to the context here.

assets/js/atomic/blocks/product/title/block.js
```diff
@@ -3,30 +3,57 @@
...
 	headingLevel = 2,
 	productLink = true,
+	...props
 } ) => {
-	const { layoutStyleClassPrefix } = useInnerBlockConfigurationContext();
-	if ( ! product.name ) {
-		return null;
+	const productDataContext = { ...useProductDataContext(), ...props };
+	const { layoutStyleClassPrefix } = useInnerBlockLayoutContext();
+	const { product } = productDataContext;
```
Aljullu, line 31: Personal opinion here, but these two lines were confusing me a bit:
```JS
const productDataContext = { ...useProductDataContext(), ...props };
const { product } = productDataContext;
```
I think this can be rewritten so the spread syntax is not necessary:
```JS
const productDataContext = useProductDataContext();
const { product } = productDataContext || props;
```
I know it's just personal taste, but proposing it in case you like it (it's affecting all other inner blocks). :slightly_smiling_face:

assets/js/atomic/blocks/product/image/block.js
```diff
@@ -2,101 +2,129 @@
...
-		<div
-			className={ classnames(
-				className,
-				`${ layoutStyleClassPrefix }__product-image`
-			) }
-		>
+		<div className={ classnames( className, componentClass ) }>
 			{ productLink ? (
 				<a href={ product.permalink } rel="nofollow">
-					{ renderedSalesAndImage }
```
Aljullu, line 88: Just curious, why not keeping `renderedSalesAndImage`? It saved some duplicated lines of code.

mikejolley: I found it a little hard to follow the logic before, where now it's in two clear blocks.

assets/js/atomic/blocks/product/title/block.js
mikejolley: @Aljullu This breaks :) Context has value - we just want to take the product from it, or props, with props taking priority.

mikejolley: This works

```
const productDataContext = useProductDataContext();
const product = props.product || productDataContext.product;
```

mikejolley:
> Code looks good and is testing well, but I get a JS error when editing the All Products block in the editor (pressing the pencil icon):

Not happening here @Aljullu 

Made the other changes requested :)

Issue #2459, opened 2020-05-11
Replace all hardcoded font-color declarations with alternatives in the Cart and Checkout blocks


Aljullu:
Hardcoded font-color make our blocks not to play well with different background colors: a black text will not be visible if the store has a black background. We should evaluate and implement alternatives depending on each case:
* Some times, two different font-weight values work as well as two different colors.
* Some other times, opacity is good enough to simulate different shades of the same color.

Some of this will need design approval/recommendations, so better discuss each change with designers.

Aljullu:
Let's rename this issue so it's only about fixing this in the Cart and Checkout blocks. We have another issue to identify other blocks affected by this (#2539).

PR #2597, opened 2020-06-01
Update Cart & Checkout heading styles

### LINKED ISSUE #2141: Evaluate if we can move away from Card component
Currently, we're using Card component from Gutenberg, it comes with its own baggage, mainly `@emotion` which adds around 30kb gzip of extra weight, we get no value from it, Card component is simple and we should consider building our own.

### LINKED ISSUE #2142: Evaluate if we can move away from PanelBody and PanelRow
Currently, we use PanelBody and PanelRow in our frontend code, however, it comes with its baggage, mainly dashicon, which adds a significant amount to our bundle (around 50kb gzip), the component is fairly simple to implement.

### LINKED ISSUE #2562: Analyze if we can reuse the Title component in the Checkout block
In #2486 we introduced a `Title` component that can be reused in titles for other blocks. This issue is about investigating if it can be reused in the Checkout block titles. That will depend on the styles and how similar they are.

Aljullu:
Fixes #2562.
Closes #2141.
Closes #2142.
Part of #2459.
Part of #2605.

This PR:
* Removes all hard-coded colors, font weight, font family, etc. from headings. So they are inherited from the theme.
* I replaced the `<Panel>` component from Gutenberg with a `<DisclosureWidget>`. `Panel` had some issues:
  * It was forcing the button to be an `<h2>`, that was not semantically correct because in some cases it needed to be an `<h3>` (shipping packages, for example).
  * Its styles were really influenced by their presence in the editor and it had some hard-coded colors. Resetting them in the frontend was getting harder and harder.
* I removed the `<Card>` component following to the designs in: pbIy4N-dY-p2.
#### Accessibility

### Screenshots

_Cart:_
imatge

_Checkout:_
imatge

1. In Storefront, go to _Appearance_ > _Customize_ and change the _Heading_ color.
2. Visit the _Cart_ and _Checkout_ blocks and verify the heading color is applied to the headings.
3. Verify there are no regressions related to the `Panel` → `DisclosureWidget` change. Click on the `Order Summary` title, the `Coupon?`, etc.
4. Verify the _Cart_ and _Checkout_ blocks don't have `Card`s in the sidebar.
5. Feel free to try with other themes and verify the heading styles are correctly applied to the blocks.

### Changelog

> The Cart and Checkout blocks now use the heading styles provided by the theme.

### DIFF
--- a/assets/js/base/components/disclosure-widget/index.js
+++ b/assets/js/base/components/disclosure-widget/index.js
@@ -29,6 +29,7 @@ const DisclosureWidget = ( {
 		>
 			<TitleTag>
 				<button
+					aria-expanded={ isOpen }
 					className="wc-blocks-components-disclosure-widget__button"
 					onClick={ () => setIsOpen( ! isOpen ) }
 				>
--- a/assets/js/base/components/disclosure-widget/index.js
+++ b/assets/js/base/components/disclosure-widget/index.js
@@ -1,0 +1,61 @@
+/**
+ * External dependencies
+ */
+import { useState } from '@wordpress/element';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import { Icon, chevronUp, chevronDown } from '@woocommerce/icons';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+const DisclosureWidget = ( {
+	buttonContent,
+	buttonWrapperTag = 'div',
+	children,
+	className,
+	initialOpen = false,
+} ) => {
+	const [ isOpen, setIsOpen ] = useState( initialOpen );
+
+	const TagName = `${ buttonWrapperTag }`;
+	return (
+		<div
+			className={ classNames(
+				className,
+				'wc-blocks-components-disclosure-widget'
+			) }
+		>
+			<TagName>
+				<button
+					className="wc-blocks-components-disclosure-widget__button"
+					onClick={ () => setIsOpen( ! isOpen ) }
+				>
+					<Icon
+						aria-hidden="true"
+						className="wc-blocks-components-disclosure-widget__button-icon"
+						srcElement={ isOpen ? chevronUp : chevronDown }
+					/>
+					{ buttonContent }
+				</button>
+			</TagName>
+			<div
+				className="wc-blocks-components-disclosure-widget__content"
+				hidden={ ! isOpen }
+			>
+				{ children }
+			</div>
+		</div>
+	);
+};
+
+DisclosureWidget.propTypes = {
+	buttonContent: PropTypes.element,
+	buttonWrapperTag: PropTypes.string,
+	className: PropTypes.string,
+	initialOpen: PropTypes.bool,
+};
+
+export default DisclosureWidget;
--- a/assets/js/base/components/disclosure-widget/index.js
+++ b/assets/js/base/components/disclosure-widget/index.js
@@ -12,15 +12,15 @@ import { Icon, chevronUp, chevronDown } from '@woocommerce/icons';
 import './style.scss';
 
 const DisclosureWidget = ( {
-	buttonContent,
-	buttonWrapperTag = 'div',
 	children,
 	className,
 	initialOpen = false,
+	title,
+	titleTag = 'div',
 } ) => {
 	const [ isOpen, setIsOpen ] = useState( initialOpen );
 
-	const TagName = `${ buttonWrapperTag }`;
+	const TagName = `${ titleTag }`;
 	return (
 		<div
 			className={ classNames(
@@ -38,7 +38,7 @@ const DisclosureWidget = ( {
 						className="wc-blocks-components-disclosure-widget__button-icon"
 						srcElement={ isOpen ? chevronUp : chevronDown }
 					/>
-					{ buttonContent }
+					{ title }
 				</button>
 			</TagName>
 			<div
@@ -52,10 +52,10 @@ const DisclosureWidget = ( {
 };
 
 DisclosureWidget.propTypes = {
-	buttonContent: PropTypes.element,
-	buttonWrapperTag: PropTypes.string,
 	className: PropTypes.string,
 	initialOpen: PropTypes.bool,
+	title: PropTypes.element,
+	titleTag: PropTypes.string,
 };
 
 export default DisclosureWidget;
--- a/assets/js/base/components/disclosure-widget/index.js
+++ b/assets/js/base/components/disclosure-widget/index.js
@@ -16,11 +16,10 @@ const DisclosureWidget = ( {
 	className,
 	initialOpen = false,
 	title,
-	titleTag = 'div',
+	titleTag: TitleTag = 'div',
 } ) => {
 	const [ isOpen, setIsOpen ] = useState( initialOpen );
 
-	const TagName = `${ titleTag }`;
 	return (
 		<div
 			className={ classNames(
@@ -28,7 +27,7 @@ const DisclosureWidget = ( {
 				'wc-blocks-components-disclosure-widget'
 			) }
 		>
-			<TagName>
+			<TitleTag>
 				<button
 					className="wc-blocks-components-disclosure-widget__button"
 					onClick={ () => setIsOpen( ! isOpen ) }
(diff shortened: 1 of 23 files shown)


senadir:
Nice seeing some weight improvements, also closes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2142

assets/js/base/components/disclosure-widget/index.js
```diff
@@ -0,0 +1,61 @@
...
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import { Icon, chevronUp, chevronDown } from '@woocommerce/icons';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+const DisclosureWidget = ( {
```
nerrad, line 14: I'm not sure on the name of this component. _Disclosure_ connotes something privacy or terms and conditions related, and _Widget_ suggests a generic action tool.

Maybe `InformationPanel` or `DescriptionPanel`? Or even just `Panel` (while realizing the potential confusion with the WordPress component, I think if we're dropping usage of that anyways, this shouldn't be that much of an issue).

Aljullu: Sounds good. Renamed to `Panel` in 92be5de5f52c4bc6453c893b00faa141b47a227c.

Aljullu:

> In the above screenshot, the focus border on the coupon panel is overlapped by the contents of the panel when opened. Not sure if this is intended or not, but looks weird.

Good catch! I made the button clickable area a bit smaller in 48ace8565a0200aab744cabef8073ddebd31ae49. I think the target area is still large enough so a11y is not affected and visually it looks better:

Peek 2020-06-03 12-31

> In the following two screenshots (one desktop, and one mobile) I'm intending to illustrate the size of the titles (eg "Shopping Cart" and "2 items"). The weight of these titles doesn't seem to be that distinctive from the other content in the cart/checkout. Is that intentional?

I think that's what we want per-design (tagging @LevinMedia in case he has some more feedback). The goal is to inherit all heading styles and only override the font-size. That means that in Storefront the titles have a light font weight.

If necessary, we could add some styles in Storefront targeting specially these titles so they have a bolder font weight. In parallel, we could add some docs for 3rd-party themes that want to do the same.

> Also, even though they are h2's they don't seem to be inheriting from h2 styles in the theme (again, intentional?)

They should inherit everything except for the size. Is there any other style they are not inheriting?

nerrad:
> They should inherit everything except for the size.

Ahh, I missed that, in that case it appears to be working as expected. Let's wait for feedback from @LevinMedia on the other questions (in case there are any changes) and then I'll do the final review/run through.

Aljullu:
Thanks for testing this too @senadir. However, I'm unable to reproduce. This is how it looks for me with 2020:

imatge

senadir:
It works fine in small width but is broken on larger width

Aljullu:
I'm still unable to reproduce. What do you mean by larger width? I tried the _Full Width Template_ and it looks fine:
imatge

nerrad:
Thanks for checking more themes Nadir! I also notice some issues with Twenty Twenty (Chrome on Mac):

The following appears on both the default template width and the full-width template:

Image 2020-06-03 at 9 45 27 AM

The following appears in the editor:

Image 2020-06-03 at 9 46 00 AM

Also, even with full-width template selected for twenty-twenty, the checkout is constrained to the mobile view in the editor (might have to do with the container width in the editor? Or potentially a bug with twenty-twenty).

I tested with the latest twenty-twenty theme.

nerrad:
The latest issues make me wonder if we should settle on a common set of themes we test styles against as a part of this work to ensure we're not introducing potentially worse behaviour with changes? I fully realize getting things working well across all themes is going to be hard (and in some cases, simply impossible) but if we have a common sampling of various theme variations that can be used for the testing, it could help improve for _most_ themes.

Also, we should consider looking into automating some visual regression testing against whatever theme sampling we decide on (might be good for a cooldown). Since so much of what we are stepping is affected by themes and frontend styles, I think it will be helpful for us to automate this as a part of our transitory work between old styles and eventual Global Styles/block themes.

senadir:
Okay so this is an edge case as far as I see, if I insert the block alone directly, it won’t show, if I have the block inside a  column (my initial setup), it will look like this

image


tested on master and it also existed there, so not blocking

senadir:
> Also, we should consider looking into automating some visual regression testing against whatever theme sampling we decide on (might be good for a cooldown). Since so much of what we are stepping is affected by themes and frontend styles, I think it will be helpful for us to automate this as a part of our transitory work between old styles and eventual Global Styles/block themes.

on an individual basis, our components works fine for the most time, the issues we run into is other themes styles leaking into ours, we can look into automating this in storefront and see how things work out.

Aljullu:
> The following appears on both the default template width and the full-width template:

Oh, right. Express shipping methods don't appear by default in my local install and I missed it. :facepalm: 

> The following appears in the editor:

I think this was also happening in master, but this PR is a good opportunity to fix it: 3995214047b61ff964095bfe4f0fc61ee12aaf36.

> Okay so this is an edge case as far as I see, if I insert the block alone directly, it won’t show, if I have the block inside a column (my initial setup), it will look like this

Ah, good catch! Should be fixed in the same commit. Twenty Twenty seems to add selectors with a very high specificity and that was not playing well with our blocks.

> Also, even with full-width template selected for twenty-twenty, the checkout is constrained to the mobile view in the editor (might have to do with the container width in the editor? Or potentially a bug with twenty-twenty).

My guess is that this is an issue with TwentyTwenty. Ideally, the editor available space should change depending on the select template (this is how it works with Storefront). I created an issue in Trac: https://core.trac.wordpress.org/ticket/50311.

> Also, we should consider looking into automating some visual regression testing against whatever theme sampling we decide on (might be good for a cooldown). Since so much of what we are stepping is affected by themes and frontend styles, I think it will be helpful for us to automate this as a part of our transitory work between old styles and eventual Global Styles/block themes.

:100: 

> on an individual basis, our components works fine for the most time, the issues we run into is other themes styles leaking into ours, we can look into automating this in storefront and see how things work out.

I think @nerrad's point (or at least what I understood :smile: ) was to have an automated tool to report changes in a set of themes. Most of us are usually in Storefront so it's easy to spot regressions there, but a regression in another theme might pass inadvertently (like the `Express Checkout` title being mispositioned in this PR).

nerrad:
> I think @nerrad's point (or at least what I understood 😄 ) was to have an automated tool to report changes in a set of themes.
That 👍

LevinMedia:
>Also, even with full-width template selected for twenty-twenty, the checkout is constrained to the mobile view in the editor (might have to do with the container width in the editor? Or potentially a bug with twenty-twenty).

@nerrad I've noticed this behavior on a number of themes... and even on themes that do allow for the full width of the cart and checkout on desktop, sometimes they get positioned off to the side in a strange way. 

It's been discussed previously in slack with @senadir and @haszari - but I feel like this is a great case for introducing some kind of control that breaks the blocks out of the theme template, and forces them into a width and pattern of responsive behavior that we have complete control over.

nerrad:
> All that said - @Aljullu can you bump up the font size from default to large on all headings i think that's going to help without being overwhelming. This would also have to be updated in the cart to match.

I think even that bump will help significantly as it will help distinguish it a bit more from the rest of the content imo.

Aljullu:
> Interesting to note the right column headings didn't pick up the styles correctly :monocle_face: 

Ah, good catch. We weren't resetting the italics added by the `<button>` element, that has been fixed in ef99c02bae5e6b887047f0c04d237f8a05c6eb57. In the other hand, Galleria seems to force the font weight to light with an `!important` rule, so I don't think that's something we want to fix on our end. Should we open an issue in their repo?

> All that said - @Aljullu can you bump up the font size from `default` to `large` on all headings i think that's going to help without being overwhelming. This would also have to be updated in the cart to match.

Done in 11c856b04312baef6ff3f64bbf971c541ac13a04. It includes some small tweaks that were needed to make sure everything still aligned with the new heading size.

Lastly, I also pushed a commit adding padding to the address form and Shipping options in the _Cart_ block, so they align with the rest of the column:
| Before | After |
| --- | --- |
| imatge | imatge |

PR #2615, opened 2020-06-03
Unify Cart headers

### LINKED ISSUE #2614: Cart: merge the left and right titles
Currently, the _Cart_ block has two titles, one left aligned and the other one right aligned. We want to merge them into a single one:
imatge
↓
imatge

Aljullu:
Fixes #2614.

#### Accessibility

### Screenshots
_Before:_
imatge

_After:_
imatge

Open the _Cart_ block and verify you see a heading in the form of `X products - X items` and that the numbers are correct.

### Changelog

> The Cart block titles have been merged into one.


nerrad:
Before reviewing/merging, I'd like some feedback from @LevinMedia on [this comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2614#issuecomment-638138380) by @mikejolley. 

I think there's value in having a header that summarizes the contents of the cart, but at first glance I'm confused by the current phrasing (count of products and count of "items"). At least with the prior heading Items was just a total count of everything in the cart/checkout.

**If we're going to have a heading summary**, I think we should think through a bit more what we include here. Some options:

- "You have x items in the cart"
- "You have x products in the cart"
- "You have x products (total of y items)"

Another potential issue with having _any_ wording here, is that some stores may want to customize the word "products" (subscriptions, bookings, events etc). So until that is possible, it might be better to just go with a more generic "items" for the summary header (assuming we keep it)?

haszari:
Agree we should review the wording before making this change. Listing the number of items and rows/products could be extra complexity - do we need all two different numbers at the top? Do they help convert?

Also previously, the count was right-aligned, which felt like a typical table layout.

I realise this was discussed elsewhere (slack) but there's not a lot of rationale/justification for the change in this PR or issue #2614.

Aljullu:
Updated with the feedback from https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2614#issuecomment-638980303. This is how it looks now:
imatge

PR #2619, opened 2020-06-03
Update item count badge styles

### LINKED ISSUE #2617: Cart & Checkout: update item count badge so it works in light and dark backgrounds
We want to update the designs of the item count badge:
screen-shot-2020-05-22-at-9 19 30-am

Aljullu:
Fixes #2617.

Small PR updating the item count badge styles in the _Cart_ block.

### Screenshots
| Before | After |
| --- | --- |
| imatge | imatge |

Open the _Cart_ block and verify the item count badge matches the new styles.

### Changelog

> The item count badges of the Checkout block have been updated so it looks better in light & dark backgrounds.

### DIFF
--- a/assets/js/base/components/cart-checkout/index.js
+++ b/assets/js/base/components/cart-checkout/index.js
@@ -2,6 +2,7 @@ export { default as AddressForm } from './address-form';
 export { default as Button } from './button';
 export { default as CheckoutForm } from './form';
 export { default as FormStep } from './form-step';
+export { default as OrderSummary } from './order-summary';
 export { default as PlaceOrderButton } from './place-order-button';
 export { default as Policies } from './policies';
 export { default as ProductImage } from './product-image';
--- a/assets/js/base/components/cart-checkout/order-summary/index.js
+++ b/assets/js/base/components/cart-checkout/order-summary/index.js
@@ -9,9 +9,9 @@ import Panel from '@woocommerce/base-components/panel';
 /**
  * Internal dependencies
  */
-import CheckoutOrderSummaryItem from './order-summary-item.js';
+import OrderSummaryItem from './order-summary-item.js';
 
-const CheckoutOrderSummary = ( { cartItems = [] } ) => {
+const OrderSummary = ( { cartItems = [] } ) => {
 	const { isLarge, hasContainerWidth } = useContainerWidthContext();
 
 	if ( ! hasContainerWidth ) {
@@ -20,19 +20,19 @@ const CheckoutOrderSummary = ( { cartItems = [] } ) => {
 
 	return (
 		<Panel
-			className="wc-block-order-summary"
+			className="wc-block-components-order-summary"
 			initialOpen={ isLarge }
 			title={
-				<span className="wc-block-order-summary__button-text">
+				<span className="wc-block-components-order-summary__button-text">
 					{ __( 'Order summary', 'woo-gutenberg-products-block' ) }
 				</span>
 			}
 			titleTag="h2"
 		>
-			<div className="wc-block-order-summary__content">
+			<div className="wc-block-order-components-summary__content">
 				{ cartItems.map( ( cartItem ) => {
 					return (
-						<CheckoutOrderSummaryItem
+						<OrderSummaryItem
 							key={ cartItem.key }
 							cartItem={ cartItem }
 						/>
@@ -43,10 +43,10 @@ const CheckoutOrderSummary = ( { cartItems = [] } ) => {
 	);
 };
 
-CheckoutOrderSummary.propTypes = {
+OrderSummary.propTypes = {
 	cartItems: PropTypes.arrayOf(
 		PropTypes.shape( { key: PropTypes.string.isRequired } )
 	),
 };
 
-export default CheckoutOrderSummary;
+export default OrderSummary;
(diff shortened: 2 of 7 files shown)


nerrad:
Twenty Twenty styling looks a little bit awkward (padding is off a bit), so you may want to see how it looks with larger numbers?

Image 2020-06-03 at 10 36 56 AM

senadir:
Totally subjective but if we’re hardcoding colors why not a white circle and black text?

Aljullu:
> Twenty Twenty styling looks a little bit awkward (padding is off a bit)

I'm unable to reproduce, this is how it looks for me in 2020: :confused: 
imatge

In your screenshot, the Panel arrow looks misaligned too. Maybe you switched branches and styles were not correctly generated?

> so you may want to see how it looks with larger numbers?

I initially tested with `99`, which is the largest number of items I could add with the _Cart_ block and it was working fine. In 801181ee2361b1909bd2f4e583c93127a5ab7257 I just added `white-space: nowrap;` to make sure it never breaks in two lines whatever the number is:
imatge

> I must confess I like the original style better (completely subjective) but I understand the reasoning behind this change. It just feels so heavy compared with the other elements on the page (for storefront at least).

I see what you mean, even though in the previous design I thought the other way around, it felt the count badges were easy to miss.

One option might be to try to retrieve the background color via JS and then make the count badge match the theme colors. But I'm a bit wary of this because background colors might have transparencies, be gradients, images, etc. So it would not be bullet-proof. @nerrad what do you think?

Another option would be to leave this PR as-is, and then expect themes to customize these colors. We could take care of updating Storefront and we could add docs for 3rd-party themes.

PD: that's an example of an issue that we would not have with global styles. :slightly_smiling_face: 

> Totally subjective but if we’re hardcoding colors why not a white circle and black text?

For reference, this is how it would look:
imatge

cc'ing @LevinMedia in case he has some feedback on this.

LevinMedia:
@Aljullu @senadir 

>Totally subjective but if we’re hardcoding colors why not a white circle and black text?

Sorry I missed this option when I was doing the original designs, but yeah, that looks much better. Let's go with that.

Aljullu:
Summary of the last commits:
* 6b0ab17e4f22da6b731c1f0419eda730b336d99a switches black and white colors following the feedback above. :point_up_2: 
* b926fd89efa61b9e9b58f58d346d0d4cd35c6da5 adds documentation for themes on how to change the colors of the count badge.
* 6a047126c8ac2a53576f295412d76c87e817c021 moves _Order summary_ CSS selectors out of the parent class `.wc-block-checkout__sidebar`. This shouldn't make any visual difference but will make it easier for themes to style it because specificity is lower so a single class is enough to modify the styles.

nerrad:
I definitely like the badge designs better. In testing I discovered some more things and I realize you've merged and closed this pull (which is okay) but if there aren't known issues for these will you see that issues get created? I also know it's likely at least two of these issues weren't existing from the work in this pull, but I just discovered them so thought I'd make note of them 😄 

This screenshot has inline comments (comments apply to Storefront and Twenty Twenty):

Image 2020-06-05 at 11 03 32 AM

The padding and margins on the validation for credit card fields seems off:

Image 2020-06-05 at 11 07 50 AM

I still see some off center padding for the numbers in the quantity bubbles for Twenty Twenty (both frontend and editor/mobile):

Image 2020-06-05 at 11 10 32 AM

For the above, I noticed you mentioned you couldn't reproduce, what's odd is that the font in your screenshot is much different than what's in my screenshot for the Twenty Twenty Theme. Since I don't see a place in the theme settings for customizing font, I'm wondering if this is a case of your system font being different than the system font displayed on my OS (and Twenty Twenty is defaulting to system font for this style)?

Aljullu:
Ah, sorry! I went ahead merging because it was already approved and there wasn't any update. I should have asked for another review instead. :sweat_smile: 

> This screenshot has inline comments (comments apply to Storefront and Twenty Twenty):

1a. Express checkout left/right padding: I'm unable to reproduce. Is that with Storefront `master` and WC Blocks `master`?
1b. Express checkout vertical align: right, good eye with that! I fixed it in: #2652.
2. The title vertical centering off will be fixed with #2649.

> The padding and margins on the validation for credit card fields seems off:

Created an issue: #2650. I added it to 2.7.0 milestone because I expect it to be a quick fix, but might move to 2.8.0 if I see it's harder to fix, I don't think it should be blocking.

> I still see some off center padding for the numbers in the quantity bubbles for Twenty Twenty (both frontend and editor/mobile):

Yeah, still unable to reproduce it here. I created an issue to investigate it: #2651. This one targeting 2.8.0.

nerrad:
> Ah, sorry! I went ahead merging because it was already approved and there wasn't any update. I should have asked for another review instead.

No worries 😄 ! I think it was fine to merge, we can handle the others in followups.

> 1a. Express checkout left/right padding: I'm unable to reproduce. Is that with Storefront master and WC Blocks master?

Storefront 2.5.7 and WC Blocks this branch.

Aljullu:
> > 1a. Express checkout left/right padding: I'm unable to reproduce. Is that with Storefront master and WC Blocks master?
> 
> Storefront 2.5.7 (released) and WC Blocks this branch.

I was still unable to reproduce this issue. But #2704 refactors the Express checkout CSS, so it might be a good chance to test if it's fixed now.

nerrad:
Still not seeing much padding on #2704 branch:

Image 2020-06-12 at 11 27 04 AM

(this is a production build)

PR #2645, opened 2020-06-04
Add Atomic SKU block

### LINKED ISSUE #2521: [Atomic Block] SKU
Create an SKU atomic block and component for inclusion in Product Inner Blocks.

mikejolley:
Note this is branched off of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2607 so https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2607 needs to be merged first.

Adds an atomic block to display the product SKU.

Fixes #2521

### Screenshots

Edit Page ‹ Local Testing — WordPress 2020-06-04 17-10-10
Edit Page ‹ Local Testing — WordPress 2020-06-04 17-10-00

1. Add a Single Product Block
2. SKU should be in the layout
3. Ensure it can be interacted with and shows the link to edit the product in the inspector
4. Ensure it displays on the frontend


assets/js/atomic/blocks/product/sku/block.js
```diff
@@ -0,0 +1,55 @@
...
+	if ( ! sku ) {
+		return null;
+	}
+
+	return (
+		<div
+			className={ classnames(
+				className,
+				'wc-block-components-product-sku',
+				`${ parentClassName }__product-sku`
```
haszari, line 41: I'm coming in late on the class names discussion .. keen to understand why we need both classes here. Not a blocker, just something I need to get up to speed on.

If having both classes with parent prefix here is a common pattern, perhaps we can wrap `classnames` and make our own little utility for atomic blocks. Then we can document the expectations/rationale and ensure all atomic blocks do it in a consistent way. (Not a blocker just thinking out loud.)

assets/js/components/edit-product-link/index.js
```diff
@@ -0,0 +1,40 @@
...
+ */
+import { __ } from '@wordpress/i18n';
+import { Icon, external } from '@woocommerce/icons';
+import { ADMIN_URL } from '@woocommerce/settings';
+import { InspectorControls } from '@wordpress/block-editor';
+
+/**
+ * Component to render an edit product link in the sidebar.
+ */
+const EditProductLink = ( { productId } ) => {
```
haszari, line 12: I see this is not new on this branch - I'm assuming it's been reviewed elsewhere and there's nothing new/changed in this branch. Code looks fine though :)

mikejolley:
> There was one potential issue - after removing the SKU from Single Product block, I couldn't add it back.

I've not enabled this yet. I will once all atomic blocks are finished.

> If there's no sku, the block is invisible. We might want a subtle placeholder?

I've had similar questions about the sale badge which isn't shown unless a product is on sale. I'm not sure if there is a solution yet, but this will need design input cc @LevinMedia 

> Some stores are likely to want more control/flexibility with the SKU: prefix text (and other presentation/layout concerns). 

David has a document with presentation features to be added later. I'd err on not offering too much here until in the hands of users.

assets/js/atomic/blocks/product/sku/block.js
```diff
@@ -0,0 +1,55 @@
...
+ * Internal dependencies
+ */
+import './style.scss';
+
+/**
+ * Product SKU Block Component.
+ *
+ * @param {Object} props             Incoming props.
+ * @param {string} [props.className] CSS Class name for the component.
+ * @param {Object} [props.product]   Optional product object. Product from context will be used if
```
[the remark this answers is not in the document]
mikejolley: We support both so All Products block doesn't need to be updated whilst single products block is in development.

assets/js/atomic/blocks/product/sku/block.js
mikejolley: We did want to drop the prefixed ones but it's kept to avoid breaking compatibility with all products block. Now it's just here for consistency with the other atomic blocks.

Some of the suffixes differ though. We can maybe unify once we get all atomic blocks added.

assets/js/components/edit-product-link/index.js
mikejolley: Yeah just moved so it can be consumed by multiple blocks.

PR #2648, opened 2020-06-05
Rename wc-single-product script name

### LINKED ISSUE #2646: product page image not visible (styling/single product regression?)
**Describe the bug**
With WooCommerce Blocks installed and active (current master - 6b1eea33056071c28bb796514d7c1972ea0fdd47), the main product image is not showing on product pages. 

Screen Shot 2020-06-05 at 1 10 09 PM

With WooCommerce Blocks deactivated, the image shows up. This affects all browsers I tried - Chrome, Safari, Firefox.

Looking at the CSS I don't immediately see why it's not showing - 🤷🏻‍♂️

I spotted this when testing Storefront (https://github.com/woocommerce/storefront/pull/1368). (Confirmed the storefront PR branch is not relevant to this - issue reproducible on Storefront master.)

Apologies if this is already known/reported or in progress on another branch - I couldn't find an issue so logging one.

**To Reproduce**
1. Ensure your store has products with images.
2. View a product page, e.g. `/product/sunglasses/`

With WooCommerce Blocks active on master, you should not see a product image.

If you deactivate WooCommerce Blocks, the image is visible.

Aljullu:
Fixes #2646.

WC Core has a script named `wc-single-product` that is used for the Single Product page image. We were naming the Single Product block script with the same name, so there was a conflict. This PR renames our script to `wc-single-product-block`.

1. Go to a product's page.
2. Verify the images load as usual.
3. Verify there are no regressions in the single product block.

### DIFF
--- a/src/Assets.php
+++ b/src/Assets.php
@@ -85,7 +85,7 @@ class Assets {
 		self::register_script( 'wc-cart-block', plugins_url( self::get_block_asset_build_path( 'cart' ), __DIR__ ), $block_dependencies );
 
 		if ( 'experimental' === WOOCOMMERCE_BLOCKS_PHASE ) {
-			self::register_script( 'wc-single-product', plugins_url( self::get_block_asset_build_path( 'single-product' ), __DIR__ ), $block_dependencies );
+			self::register_script( 'wc-single-product-block', plugins_url( self::get_block_asset_build_path( 'single-product' ), __DIR__ ), $block_dependencies );
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


haszari:
Should we be prefixing all our script handles to protect against regressions like this? e.g `wc-blocks-*` Keen for your thoughts on this @nerrad.

nerrad:
> Should we be prefixing all our script handles to protect against regressions like this? 

Good question! Ideally, yes. I don't know what the historical context is for the script handle naming but all the "legacy" blocks have no prefix/suffix  and we really can't change the handle now because of how it might be referenced as a dependency in the wild. 

Still it'd be a good idea to try and use a consistent naming schema (namespaced better) for new blocks going forward. I've created #2660 for moving forward with this.

Thanks Rua!

PR #2649, opened 2020-06-05
Update Checkout step progress indicator design

Aljullu:
Fixes #2618.

### Screenshots
_Storefront:_
imatge

_Twenty Twenty:_
imatge

_Bistro:_
imatge

### How to test the changes in this Pull Request:

1. Load the _Checkout_ block and verify step title numbers have the new styles.
2. Check with different themes to verify it looks good in all of them.
3. Verify the skeleton displayed while loading looks ok as well.

### Changelog

> Checkout step progress indicator design has been updated to match the theme headings style.

PR #2652, opened 2020-06-05
Fix styling issues introduced in #2597 and #2619

Aljullu:
I introduced some style regressions in two recent PRs #2597 and #2619.

This PR fixes them. A commit-by-commit explanation:

* 2f6c91d128e0ef78ad72cbe215b5632d98131316 moves the _Order summary_ styles to its own stylesheet. After creating the component in #2619, that was the natural move. The commit doesn't introduce any change, it's just moving code around so it shouldn't produce any visual difference.

* 4c87654c69bd1f2d4c597302146d6a9496367692 fixes a wrong class name I introduced while moving the _Order summary_ component in #2619. Visually, it fixes an issue in Twenty Twenty (and maybe other themes) where the _Order summary_ table was not full-width:

| Before | After |
| --- | --- |
| Screenshot | Screenshot |

* fe28ea9298972560d46b9d7e071063bee44b1796 moves the express checkout title a bit higher, so visually it's aligned with the border:

imatge

* fab69f3ecb3f97212b7c036a3db62e73525e938e fixes the shipping options not correctly aligned in the _Cart_ sidebar. In parallel, I needed to add some right padding to the panel button to avoid the text overlapping the arrow icon (960804b264292836e8e47a3d0001789d137aedce).

| Before | After |
| --- | --- |
| imatge | imatge |

Check that the _Cart_ and _Checkout_ blocks look like the screenshots above and that there are no new regressions.


Aljullu:
Thanks for the review @haszari!

> 1 - Checkout form step circled numbers & headings are not aligned. Seems to affect all themes the same, slightly misaligned.

That's fixed in #2649. :slightly_smiling_face: 

> 2 - 2x widths of separator lines in sidebar. In both the cart and checkout there are wide separators for the main divisions, and some nested more narrow separators (between products in cart, between shipping options in checkout). I wonder if it's a bit heavy having two levels of separator here - are the narrow separators needed? Maybe whitespace is enough, and we save separators for the major divisions.

That's per-design, smaller separators are for elements inside panels while wider separators are for panel elements themselves. I opened an issue (#2663) about removing the border around the _Order summary_ panel, that might help alleviate this.

> In Twenty Nineteen I saw order summary product separators full width (looks like mobile styles applying for my "small" browser window). When they aren't indented, the hierarchy is not quite right maybe? In Twenty Twenty it's using mobile layout all the time due to narrow content width.

Correct, separators are always the same width in mobile. I think this issue might be alleviated by #2663 as well. We can open an issue in the future if that's still an issue after #2663 is fixed.

Issue #2638, opened 2020-06-04
[GlobalStep]Credit/Debit card details i.e., Card Number, Expire date and CVV/CVC fails to refresh.


gglobalstep:
**Environment:**
Windows 10: Chrome(v83.0.4103.61)
MAC Book Air: Safari(v13.0.5 (15608.5.11))

**Steps to Reproduce:** 
1. Login to the wp-admin dashboard using valid credentials.
2. Create and publish Cart and Checkout page.
3. Go to site.
4. Add a product to cart.
5. Proceed to checkout.
6. Enter Card detail as "4000 0000 0000 0002"
7. Click on place order.
8. Note that "The card was declined" error message is displayed.
9. Also note that credit card details fails to clear.
10. Now enter valid credit card details i.e., "4242 4242 4242 4242".
11. Note that "Sorry, we are unable to process your payment at this time. Please retry later" error message is displayed.
12. Observe that user fails to place order unless and unit user refresh the page.

**Actual Result:**
Credit/Debit card details i.e., Card Number, Expire date and CVV/CVC fails to refresh.

**Expected Result:**
Credit card details should get clear and user should be able to place order with another credit card details.

[#2638.zip](https://github.com/woocommerce/woocommerce-gutenberg-products-block/files/4730978/2638.zip)

nerrad:
I'm able to reproduce this.

PR #2655, opened 2020-06-05
Fix retrying after failed payment processed server side.

nerrad:
Fixes: #2638

For reference, see the original report in #2638. Essentially what was happening:

-  The payment status was not updating for a selected payment method (when it's not a saved token selected) after the server processing returns an error.
-  This in turn meant the `onPaymentProcessing` event would not fire again on an attempt of a new credit card number right away.
-  This in turn meant that, in the case of Stripe, a new source was not created (from the new CC details) and when submitted, the server was processing the same _previous_ credit card entered.

The fix involved:

- Add a new state property in the payment method data context for the payment method data slice that tracks whether a saved token is selected (true) or not (false).
- Update the `SavedPaymentMethodOptions` component so that it will set this state when a saved payment method token is selected.
- Add an effect hook to the payment method data context to update the payment status to pristine given the right condition of the checkout status, the payment status, and whether a saved payment token was selected or not.

## To Test

This impacts:

- saved payment method token payments
- payment methods
- Express payments (not directly).

So to test, you need to verify that all payment methods type purchases work. Also as an extra verification, here are various test steps to try:

### Scenario One: initial report fixed

Follow the steps reported in #2638 and ensure that it is fixed.

### Scenario Two: Trying to break via validation errors.

You'll need to be logged in with a user that has saved payment methods.

1. On checkout, select CC payment method and then select a saved payment method again.
2. Leave one of the required fields empty.
3. Submit the checkout which should cause a validation error on the field.
4. Fix the validation error
5. Submit again and checkout should complete using the selected saved payment method.

### Scenario Three: Variation of trying to break via validation errors.

1. On checkout, select CC payment method.
2. Fill out cc number that will trigger declined card (`4000 0000 0000 0002`).
3. After server response (with error), clear a required field.
4. Select saved payment method.
5. Submit the checkout -> this should produce a validation error.
6. Fix the field.
7. Submit the checkout and this should result in the purchase completing successfully.

### Scenario Four: Payment with Cheque after failed CC.

1. On Checkout, select CC payment method.
2. Fill out cc number that will trigger declined card (`4000 0000 0000 0002`).
3. After server response (with error), select cheque payment method.
4. Submit the checkout and this should result in the purchase completing successfully for the cheque payment method.

### Changelog

> bug: Fix bug in Checkout block preventing a retry of credit card payment when first credit card used fails and a new one is tried.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -276,6 +276,23 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 		}
 	}, [ checkoutIsIdle, currentStatus.isSuccessful ] );
 
+	// if checkout has an error and payment is not being made with a saved token
+	// and payment status is success, then let's sync payment status back to
+	// pristine.
+	useEffect( () => {
+		if (
+			checkoutHasError &&
+			currentStatus.isSuccessful &&
+			! paymentData.hasSavedToken
+		) {
+			dispatch( statusOnly( PRISTINE ) );
+		}
+	}, [
+		checkoutHasError,
+		currentStatus.isSuccessful,
+		paymentData.hasSavedToken,
+	] );
+
 	// set initial active payment method if it's undefined.
 	useEffect( () => {
 		const paymentMethodKeys = Object.keys( paymentData.paymentMethods );
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -45,6 +45,7 @@ const getCcOrEcheckPaymentMethodOption = (
 			setPaymentStatus().success( {
 				payment_method: method.gateway,
 				[ savedTokenKey ]: token,
+				isSavedToken: true,
 			} );
 		},
 	};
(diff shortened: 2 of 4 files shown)


nerrad:
@Aljullu I set this as a blocker for 2.7.0 because the inability for new cc to be tried (error in number etc) could be problematic for stores, so it'd be good to get this fix out asap.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -276,6 +276,23 @@ export const PaymentMethodDataProvider = ( { children } ) => {
...
 	}, [ checkoutIsIdle, currentStatus.isSuccessful ] );
 
+	// if checkout has an error and payment is not being made with a saved token
+	// and payment status is success, then let's sync payment status back to
+	// pristine.
+	useEffect( () => {
+		if (
+			checkoutHasError &&
+			currentStatus.isSuccessful &&
+			! paymentData.hasSavedToken
```
Aljullu, line 286: I'm curious, why is this last condition necessary? What would happen if a saved credit card has expired and the user tries to pay with it so the checkout fails? Wouldn't that also trigger the bug?

nerrad: When a saved payment method token is used, the payment status is set to ‘success’ because all the processing happens server side (there is no client side processing). So in the case that a saved payment method was expired and caused an error, there would be an error message returned from the server and the user could then select another payment method. So the bug here doesn’t affect saved payment methods tokens because it has a slightly different flow.

The payment method status is used primarily for client side payment method processing.

PR #2662, opened 2020-06-08
Fix credit card validation errors alignment

### LINKED ISSUE #2650: Checkout: credit card validation errors are not properly aligned
Notice by @nerrad in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2619#issuecomment-639559750:
> The padding and margins on the validation for credit card fields seems off:
Image 2020-06-05 at 11 07 50 AM

Aljullu:
Fixes #2650.

I also took the chance to change the text input cursor to `text`, so it follows better the behavior of native input fields.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. In the _Checkout_ block, verify the cursor in input fields is `text` (see the `CVV/CVC` field in the screenshots above to see the difference).
2. Introduce some invalid numbers in the credit card input fields and verify the error messages are properly aligned.

### Changelog

> Improved alignment of credit card validation error messages.

### DIFF
--- a/assets/js/base/components/payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/style.scss
@@ -163,6 +163,16 @@
 	}
 }
 
+.is-mobile,
+.is-small {
+	.wc-card-expiry-element,
+	.wc-card-cvc-element {
+		.wc-block-form-input-validation-error > p {
+			min-height: 28px;
+		}
+	}
+}
+
 .wc-blocks-credit-card-images {
 	padding-top: $gap-small;
 	display: flex;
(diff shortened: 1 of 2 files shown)


Aljullu:
Fixed in e455d50cdb271b43ce2e751cb9426ec5a3b40355:
imatge

Notice this issue is not specific to the credit card inputs but it's reproducible in all other inputs where the validation might error break in two lines:
imatge

The reason for that is that we didn't want the content to jump up and down when errors appear/disappear, so we were constrained to show them in the available space between inputs. Fortunately, in the credit card inputs we can increase the height a bit more without worrying to overflow an input below.

Issue #2280, opened 2020-04-23
Evaluate if we can move away from Notice component


senadir:
**Describe the bug**
Investigating the increased bundle size in `All Product` it seems it was caused by us pulling in `Notice` component, that pulled in other dependencies like Dashicon, this could be fixed by https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2140 or by building our custom component.

Notice in itself also pulls dashicons and some other dependencies.

senadir:
@nerrad I would love if you can help me evaluate this and reach a decision here, notices is hitting us hard now (the bulk of the weight introduced in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2399 is by notices) all products is the same, we can assume that notices will be added to every interactive block we will develop (all products, Cart and Checkout, single product and any future product).

We should evaluate if we should move it to an independent bundle or build something our own, the notice implementation is simple and straightforward, the snackbar is slightly more complicated but still very doable.

nerrad:
With the bulk of the weight coming from Dashicons, I'd prefer to do what's needed to eliminate that instead of building our own replacement for what we're using from GB core.

nerrad:
I just realized that I hadn't prioritized this. I think we need to consider this a blocker to releasing 2.7.0 because that's the tag that will be merged to WC Core and thus introduce the significant weight increases in the bundle-sizes for various existing built files (All products, and filter blocks).

I'd like to avoid duplicating work in creating our own notice components in lieue of the WordPress package. @senadir , I think you were working on a way to address this?

Aljullu:
Even after the improvements from #2664, the size of `all-products-frontend.js` still increased from 14,85KB (WC Core 5.2) to 38,41KB (WC Blocks 2.7). Should we reopen this issue to investigate if we can improve this further?

senadir:
Removing notice will potentially remove other dependencies, but if we ever found ourself doing animations we can reuse some of those things, you can create an issue, but I'm certain the extra weight is from the effect-ful imports in `@wordpress/components` that we can’t get rid of unless we remove `@wordpress/components` (or try something similar to #2664).

nerrad:
Keep in mind we use the snackbar notices which in turn uses react-spring. I think at some point we're going to have to look investigate using lazy loading more. That way some of the extra package dependencies could be pulled in only when needed.

[...]

PR #2664, opened 2020-06-08
Ignore dashicons imports

nerrad:
Fixes: #2280

As a part of using Notice components from `@wordpress/components` directly for our new notices system, the `dashicon` package ended up getting pulled in for anything using notices. This resulted in an increase of **~85kb** (unminified) for all builds!

While dashicons are removed in later iterations of `@wordpress/components`, we're [unable to bump up our dependencies yet](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2140) because of issues with non-available packages in the versions of WP we support.

While we do have some components using dashicons (from the package version we use) the solution is to use the `webpack.NormalModuleReplacementPlugin` to replace any `dashicons` imports with a custom version that eliminates the excessive size.

## To test

This impacts anything loading our notices components. While this includes the new Product Block, because this is blocking 2.7.0, testing should primarily focus on:

- All Products
- Cart and Checkout.

Essentially testing involves:

- ensuring the frontend and editor views load without console errors.
- ensuring notices work as expected. I tested using coupons and also submitting the checkout with a empty required field (to trigger the error notice). Make sure the "dismiss x" icon shows up on the notice
- ensure the little down arrow on select dropdowns (in the checkout) show up correctly

### DIFF
--- a/assets/js/icons/library/arrow-down-alt2.js
+++ b/assets/js/icons/library/arrow-down-alt2.js
@@ -2,11 +2,14 @@
  * External dependencies
  */
 import { SVG } from 'wordpress-components';
+import classnames from 'classnames';
 
 const arrowDownAlt2 = ( { className, size, ...extraProps } ) => {
-	const iconClass = [ 'dashicon', 'dashicons-arrow-down-alt2', className ]
-		.filter( Boolean )
-		.join( ' ' );
+	const iconClass = classnames(
+		'dashicon',
+		'dashicons-arrow-down-alt2',
+		className
+	);
 	return (
 		<SVG
 			xmlns="http://www.w3.org/2000/svg"
--- a/assets/js/icons/library/arrow-down-alt2.js
+++ b/assets/js/icons/library/arrow-down-alt2.js
@@ -1,0 +1,24 @@
+/**
+ * External dependencies
+ */
+import { SVG } from 'wordpress-components';
+
+const arrowDownAlt2 = ( { className, size, ...extraProps } ) => {
+	const iconClass = [ 'dashicon', 'dashicons-arrow-down-alt2', className ]
+		.filter( Boolean )
+		.join( ' ' );
+	return (
+		<SVG
+			xmlns="http://www.w3.org/2000/svg"
+			viewBox="0 0 20 20"
+			className={ iconClass }
+			width={ size }
+			height={ size }
+			{ ...extraProps }
+		>
+			<path d="M5 6l5 5 5-5 2 1-7 7-7-7z" />
+		</SVG>
+	);
+};
+
+export default arrowDownAlt2;
--- a/bin/webpack-helpers.js
+++ b/bin/webpack-helpers.js
@@ -12,7 +12,7 @@ const chalk = require( 'chalk' );
 const { omit } = require( 'lodash' );
 const NODE_ENV = process.env.NODE_ENV || 'development';
 
-const dashIconReplacmentModule = path.resolve(
+const dashIconReplacementModule = path.resolve(
 	__dirname,
 	'../assets/js/module_replacements/dashicon.js'
 );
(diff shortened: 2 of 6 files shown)


assets/js/icons/library/arrow-down-alt2.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * External dependencies
+ */
+import { SVG } from 'wordpress-components';
+
+const arrowDownAlt2 = ( { className, size, ...extraProps } ) => {
+	const iconClass = [ 'dashicon', 'dashicons-arrow-down-alt2', className ]
+		.filter( Boolean )
+		.join( ' ' );
```
senadir, line 9: why not use classnames, a dependency we already use (that is quite lightweight).

assets/js/icons/library/arrow-down-alt2.js
```diff
@@ -0,0 +1,24 @@
...
+	return (
+		<SVG
+			xmlns="http://www.w3.org/2000/svg"
+			viewBox="0 0 20 20"
```
senadir, line 13: not sure how will the icon render, but if we want to keep with Gutenberg new grid system, this should be changed to `"-2 -2 24 24"`, but if the icon renders fine and sharp, we should be okay.

bin/webpack-helpers.js
```diff
@@ -6,11 +6,17 @@ const MergeExtractFilesPlugin = require( './merge-extract-files-webpack-plugin'
 const MiniCssExtractPlugin = require( 'mini-css-extract-plugin' );
 const ProgressBarPlugin = require( 'progress-bar-webpack-plugin' );
 const DependencyExtractionWebpackPlugin = require( '@wordpress/dependency-extraction-webpack-plugin' );
+const { NormalModuleReplacementPlugin } = require( 'webpack' );
 const WebpackRTLPlugin = require( 'webpack-rtl-plugin' );
 const chalk = require( 'chalk' );
 const { omit } = require( 'lodash' );
 const NODE_ENV = process.env.NODE_ENV || 'development';
 
+const dashIconReplacmentModule = path.resolve(
```
Aljullu, line 15: Typo:
```suggestion
const dashIconReplacementModule = path.resolve(
```

assets/js/icons/library/arrow-down-alt2.js
mikejolley: +1 I have same qu

assets/js/icons/library/arrow-down-alt2.js
```diff
@@ -0,0 +1,24 @@
...
+ * External dependencies
+ */
+import { SVG } from 'wordpress-components';
+
+const arrowDownAlt2 = ( { className, size, ...extraProps } ) => {
+	const iconClass = [ 'dashicon', 'dashicons-arrow-down-alt2', className ]
+		.filter( Boolean )
+		.join( ' ' );
+	return (
+		<SVG
```
mikejolley, line 11: Not sure if you did or not, but I tested and some very small savings could be made if running the SVG though https://jakearchibald.github.io/svgomg/ - I do this on all icon imports I introduce.

assets/js/icons/library/arrow-down-alt2.js
nerrad: Because I copy/pasted from gb for this blindly :).  I'll add.

assets/js/icons/library/arrow-down-alt2.js
nerrad: Icon seems to be rendering fine, this was copied and pasted directly from the existing dashicons implementation. Since this is mostly temporary (we can replace once we can bump component versions), not sure if we need to change.

assets/js/icons/library/arrow-down-alt2.js
nerrad: I just copied and pasted directly from the existing dashicons... are the savings enough to warrant implementing?

nerrad: Ya the difference is super miniscule, not worth the change imo.

mikejolley: 168 bytes → 138 bytes for the X

nerrad: Really? it was only 1B when I used the tool, maybe I'm using it wrong.

PR #2665, opened 2020-06-08
Fix release PR template link in release docs

Aljullu:
The link to the `release pull request template` was going to the [generic pull request template](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/.github/pull_request_template.md), instead of the [specific one for releases](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/.github/release_pull_request_template.md).

### DIFF
--- a/docs/releases/readme.md
+++ b/docs/releases/readme.md
@@ -47,7 +47,7 @@ _Outcome_: **Team is aware of release and in agreement about what fixes & featur
 
 #### Create a release pull request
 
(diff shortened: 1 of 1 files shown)


haszari:
Looking at that info about the release PR, I'm wondering if we need to update the release doc a little. Perhaps some of the steps could be moved to the PR, or maybe we can split out some of the info in the new section into specific steps to follow or advice/guidance.

Aljullu:
Thanks for the review @haszari! Yup, I'm pretty sure we will keep iterating on those docs (I just created another PR today :slightly_smiling_face: ). Merging this one for now.

PR #2683, opened 2020-06-10
Pass WP version to docker-compose as ENV var

senadir:
In preparation for https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2396 this changes our setup so we can pass a WordPress version from the env values to docker.

Tests should pass.


senadir:
build passed fine when we passed 5.4 as a value
image

I will try to change it to 5.3 and see if it actually downloads 5.3, I have doubt that this didn’t work and docker defaulted to the latest version.

senadir:
as observed, the var is used correctly.
https://travis-ci.com/github/woocommerce/woocommerce-gutenberg-products-block/jobs/347052119#L2534

Reverting back to 5.4.

PR #2686, opened 2020-06-10
Expand travis matrix

### LINKED ISSUE #2396: Expand our WordPress test matrix to cover all versions we support
After getting https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2395 it would make more sense to test our blocks in different WordPress versions, currently, we test in the latest WordPress version, we should probably pin that version.

Recently we discussed that we need to increase our support from WP 5.0 to 5.2, we also have blocks that only work in 5.3, and we should cover them all in 5.4 (along with if the new blocks will only work in 5.4 or 5.3).

So the matrix is getting complicated and the range of testing is getting complicated.
We should possibly use an ENV variable to decide what tests on run directly inside our e2e, we will see how that goes.

senadir:
TIL:
This PR changes our E2E test setup to run on several versions of WP, mainly: WP 5.4, WP 5.3, WP 5.2, WP latest with Gutenberg installed, it also tweaks some tests to run on appropriate versions only (basically version gating them, also possible to feature gate them).

It does this by introducing a new variable `GUTENBERG_LATEST` and `WP_VERSION` (defined in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2683), based on those variables, it decide what version of WordPress it should install, and if it should install Gutenberg or not.
This will also allow us to selectively decide what WooCommerce version we want to install if we ever wanted to (not implemented in this PR).
In our E2E, we skip certain block tests if they don’t meet the requirements, like Cart and Checkout, and Single Product, and All Products.
These changes will allow us to run our E2E tests with certain feature gates (not implemented in this PR).

---
Fixes #2396

### Testing:

Travis should pass.

- Locally, define the wp version you want to test with, this has to be defined when you start the docker instance, not when you run the tests, you can do that by running `WP_VERSION=5.3 npm run test:e2e:up:build`
- Run tests, confirm they pass.
- To change the version, run `npm run test:e2e:down:reset` then `WP_VERSION=5.4 npm run test:e2e:up:build`.
- If you don't pass a version, the version defined in .env will be used, that is 5.4 or latest.
- To run with Gutenberg, build the docker instance with `GUTENBERG_LATEST=true npm run test:e2e:up:build`.

### Follow ups:

- [ ] Create a helper function to skip whole tests based on WP version and Feature flag, or extend `describe` and `test` to include a check param.
- [ ] Document how to update the WP version in our tests and common issues and how we can solve them.


tests/e2e-tests/specs/backend/all-products.test.js
```diff
@@ -15,6 +15,13 @@ const block = {
 	class: '.wc-block-all-products',
 };
 
+/**
+ * @todo: write helpers to simplify version and feature gating tests.
+ */
+if ( process.env.WP_VERSION < 5.3 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( 'skipping all other things', () => {} );
```
mikejolley, line 23: This skipping message should probably include a reason so it's obvious when viewing output.

senadir: I plan to abstract those later in a follow up PR with better messages.

PR #2688, opened 2020-06-11
`registerInnerBlock` ->`registerBlockComponent`

mikejolley:
This PR repurposes `registerInnerBlock` (which was unused and not fully fit for purpose) into a new registration function named `registerBlockComponent`. For 2 reasons:

- Atomic blocks (based on recent conversations) may not be inner blocks.
- Atomic blocks may not have parents. `registerBlockComponent` parent is optional.

Additionally, `registerInnerBlock` wasn't actually registering blocks. It was registering components to map the blocks. I feel the new naming is clearer.

As well as updating the function and tests, this makes our Atomic blocks register themselves using `registerBlockComponent` in a components-init file.

1. After building, check for regressions in All Products and Single Products block.
2. Check components still render on the frontend. If so, registration is working.


mikejolley:
@nerrad hmm was the old API public? Do I need to leave the old registration in place too? I'm not sure how public it was or not.

nerrad:
> @nerrad hmm was the old API public? Do I need to leave the old registration in place too? I'm not sure how public it was or not.

It was public via `wp.BlockRegistry`. I recall a few conversations I've seen where folks were trying to use it to extend the allowed inner blocks. So ya, the likelihood of it being used is there. Are you able to leave it in place wrapping the new method? If so, you could also use the [`@wordpress/deprecated`](https://github.com/WordPress/gutenberg/tree/master/packages/deprecated) package to deprecate the old function as a signal to use the new one.

Aljullu:
@mikejolley in case it's useful, there is an example extension using that API here:

https://github.com/Aljullu/woocommerce-blocks-extension/tree/inner-blocks-extension

assets/js/blocks-registry/block-components/register-block-component.js
```diff
@@ -24,20 +24,24 @@ const assertOption = ( options, optionName, expectedType ) => {
  *
  * @export
  * @param {Object}   options           Options to use when registering the block.
- * @param {string}   options.main      Name of the parent block.
+ * @param {string}   [options.parent]  Name of the parent block, or blank to register globally.
```
haszari, line 27: Are we using `parent` anywhere for real? (I see it in the tests.) Is this needed for something in a follow up, i.e. are we planning to use it?

assets/js/atomic/blocks/component-init.js
```diff
@@ -0,0 +1,74 @@
...
+
+registerBlockComponent( {
+	blockName: 'woocommerce/product-tag-list',
+	component: ProductTagList,
+} );
+
+registerBlockComponent( {
+	blockName: 'woocommerce/product-stock-indicator',
+	component: ProductStockIndicator,
+} );
```
haszari, line 74: It seems clearer to me to have these registering themselves vs. the previous map/generic approach 👍

assets/js/blocks-registry/block-components/register-block-component.js
mikejolley: It could be used if you wanted a different component to handle a specific parent block, plus it was from the old API so kept it around.

mikejolley:
Thanks for the feedback. I've improved the docblocks and added tests for the old deprecated methods which are now still in place.

assets/js/blocks-registry/block-components/register-block-component.js
```diff
@@ -1,8 +1,47 @@
...
 import { registeredBlockComponents } from './registered-block-components-init';
 
+/**
+ * Register a Block Component.
+ *
+ * WooCommerce Blocks allows React Components to be used on the frontend of the store in place of
+ * Blocks instead of just serving static content.
+ *
+ * Registering a Block Component allows you to define which React Component should be used in place
+ * of a registered Block. The Component, when rendered, will be passed all Block Attributes.
```
haszari, line 18: These new docs are 👌

PR #2694, opened 2020-06-11
Fix Cart and Checkout sidebar styling issues

### LINKED ISSUE #2503: Cart: double border when there is only one shipping option and no taxes
**To Reproduce**

1. Disable taxes in your store.
2. Disable all shipping methods but one.
3. Go to the _Cart_.
4. Notice there is a double border between the shipping method and the _Coupon Code?_. It happens both in mobile and desktop.

**Screenshots**
imatge

### LINKED ISSUE #2658: Checkout (Order summary): there should be some padding between the product name and price
The product name and price might be too close in the _Order summary_ panel (in the sidebar). We should leave some padding in one of the two elements to make sure they don't collide:
imatge

### LINKED ISSUE #2663: Checkout: remove borders around Order summary panel
According to the designs, the _Order summary_ panel shouldn't have top and bottom border:

| `master` | Design |
| --- | --- |
| imatge | imatge |

### LINKED ISSUE #2692: Checkout order summary: too much space between product description and variations in editor
In the _Order summary_ panel of the Checkout block sidebar, there is too much space between the product description and the variations when in the editor view.

_Editor:_
imatge

_Frontend:_
imatge

We should remove the description bottom margin in the editor. It seems to be a specificity issue between two selectors.

### LINKED ISSUE #2693: Cart shipping options: no margin below the 'no shipping options' notice
**To Reproduce**
1. Disable all shipping options from your store.
2. Go to the Cart block.
3. Notice there is no margin below the 'no shipping options' notice.

**Expected behavior**
There should be some space between the 'no shipping options' notice and the _Coupon Code?_ border.

**Screenshots**
imatge

Aljullu:
Fixes #2503.
Fixes #2658.
Fixes #2663.
Fixes #2692.
Fixes #2693.

#2503

1. Disable all shipping methods but one.
2. Go to the Cart page and verify there is no double-border between the shipping method and the _Coupon Code_ panel (the border should be 1px instead of 2px as it was before).

imatge

#2658 

1. Create a product with a long name.
2. Go to the Checkout page and resize the window to trigger different sizes.
3. Verify there is always spacing between the product name and the price in the _Order summary_ panel.
imatge

#2663 

1. Still in the Checkout page, verify the _Order summary_ panel doesn't have top and bottom borders.

imatge

Fixing this, made the Checkout headers not to be aligned when there are no express payment methods. This has been fixed in 185610e2d9d8bee06169e8a25d78800e3c43b0f4, but would be worth testing it too.

1. Make sure you don't have any express payment method enabled.
2. Go to the Checkout page.
3. Verify the step 1 title and the sidebar title are aligned.

imatge

#2692

1. Add the Checkout block to a page or post and, in the editor, verify there is no spacing between the product description and the product variations in the _Order summary_.

imatge

#2693

1. Disable all shipping options from your store.
2. Go to the Cart block.
3. Verify there is margin below the 'no shipping options' notice.

imatge

### Changelog

> Several style enhancements to the Cart and Checkout block sidebar.

### DIFF
--- a/assets/js/base/components/cart-checkout/shipping-rates-control/style.scss
+++ b/assets/js/base/components/cart-checkout/shipping-rates-control/style.scss
@@ -34,31 +34,8 @@
 	margin-bottom: 0;
 }
 
-// Resets when it's inside a panel.
 .wc-block-shipping-rates-control {
-	.wc-block-shipping-rates-control__package.components-panel__body {
-		border-bottom: none;
-
-		&,
-		&.is-opened {
-			padding-bottom: 0;
-		}
-
-		.wc-block-shipping-rates-control__package-items {
-			margin: 0;
-		}
-
-		.wc-block-radio-control {
-			width: 100%;
-		}
-
-		.wc-block-radio-control__option {
-			margin-right: 0;
-			max-width: none;
-
-			&:first-child {
-				border-top: 1px solid $core-grey-light-600;
-			}
-		}
+	.wc-blocks-components-panel__content {
+		padding-bottom: 0;
 	}
 }
--- a/assets/js/base/components/cart-checkout/form-step/style.scss
+++ b/assets/js/base/components/cart-checkout/form-step/style.scss
@@ -19,7 +19,7 @@
 }
 
 .wc-block-checkout-step__content {
-	padding-bottom: $gap-larger;
+	padding-bottom: em($gap-large);
 }
 
 .wc-block-checkout-form fieldset.wc-block-checkout-step:disabled {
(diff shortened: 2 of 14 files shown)


assets/js/base/components/cart-checkout/shipping-rates-control/style.scss
```diff
@@ -34,31 +34,8 @@
 	margin-bottom: 0;
 }
 
-// Resets when it's inside a panel.
 .wc-block-shipping-rates-control {
-	.wc-block-shipping-rates-control__package.components-panel__body {
```
Aljullu, line 39: The class `.components-panel__body` was used by the `Panel` component from `@wordpress/components`. Since we migrated to our own `Panel` component which has different classes, all the code below was dead code and can be safely removed.

Aljullu:
Thanks for reviewing @senadir.

Good point with the titles. I think we want to inherit the line-height from the theme, so I wouldn't undo that. But taking a look at the new designs, it looks like the margins should be a bit bigger now than they were before. I updated them in 17f629c83fb12e663e1a8047fa4a4dfe1cd202c9 and this is how it looks now:

imatge

senadir:
yeah, that should fix it.

PR #2700, opened 2020-06-12
Fix Product Categories List align attribute

### LINKED ISSUE #2696: [GlobalStep] Fix “Product Category List” displaying error on changing alignment setting.
**Environment:**
Windows 10: Chrome(v83.0.4103.61)(64-bit), Firefox(v77.0.1)(64-bit)

**Steps to Reproduce:**
1. Open a test site.
2. Login with valid credentials.
3. From dashboard go to Pages.
4. Add new page.
5. Add Product Category List block.
6. Now in editor change alignment from Wide Width to Full Width and vice versa.
7. Observe that randomly “Error loading block invalid parameter attribute…” is displayed on Editor.

**Actual Result:**
On “Product Category List” block upon changing block alignment error stating “Error loading block invalid parameter attribute…” is displayed on Editor.

**Expected Result:**
Content inside the block should align according to the settings i.e., Wide Width or Full width and no error message should be displayed.

**Refer Artifact:**
[#2696.zip](https://github.com/woocommerce/woocommerce-gutenberg-products-block/files/4765478/2696.zip)

Aljullu:
Fixes #2696.

I added this PR to the 2.8.0 milestone, given that it's not a regression and we are doing biweekly releases, I don't think we need to do a point release for it. cc @nerrad 

### Screenshots

Peek 2020-06-12 11-45

1. Add a Product Categories List block to a page.
2. Switch to _Full Width_ align.
3. Verify the block doesn't show an error.
4. If you are using Storefront or another theme with sidebar, make sure the page has the _Full Width_ template.
4. Open the page in the frontend and verify the Product Categories List block is aligned as a full width block.

### Changelog

> Fix an error appearing in the Product Categories List block with _Full Width_ align.

### DIFF
--- a/assets/js/blocks/product-categories/index.js
+++ b/assets/js/blocks/product-categories/index.js
@@ -36,6 +36,13 @@ registerBlockType( 'woocommerce/product-categories', {
 	},
 	attributes: {
 		/**
+		 * Alignment of the block.
+		 */
+		align: {
+			type: 'string',
+		},
+
+		/**
 		 * Whether to show the product count in each category.
 		 */
 		hasCount: {
--- a/src/BlockTypes/ProductCategories.php
+++ b/src/BlockTypes/ProductCategories.php
@@ -43,6 +43,7 @@ class ProductCategories extends AbstractDynamicBlock {
 		return array_merge(
 			parent::get_attributes(),
 			array(
+				'align'          => $this->get_schema_align(),
 				'className'      => $this->get_schema_string(),
 				'hasCount'       => $this->get_schema_boolean( true ),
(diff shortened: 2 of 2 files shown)


Aljullu:
> However, it looks like setting alignment has no impact on the actual html attributes injected for the frontend rendering of the block (likely because we don't handle this attribute server-side).

Ah, right, I missed that!

In defec5da4c4fa61ec15f5ee502dc3183e0249544, I refactored the render function of the Categories List block so it prints the `align` class name. The code is based on the product grid blocks.

Tested with WP 5.2, 5.3 and 5.4 and I couldn't spot anything weird, so this is ready for another look.

PR #2704, opened 2020-06-12
Remove background color from Express Checkout title

### LINKED ISSUE #2659: Express checkout: title background is white even in themes with other background colors
imatge
_(Screenshot with Bistro)_

The background behind the express checkout titles should be transparent. Currently it's white so it hides the border behind. We should find another way to achieve this layout without relying on a background color.

Aljullu:
Fixes #2659.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Change to a theme with a background color different from white or in the customizer change Storefront's background color to another one.
2. Visit the Checkout page and verify that:
  * The `Express checkout` title doesn't have a background color different from the rest of the page and has left and right padding (so it doesn't collide with the border).
  * The `Express checkout` and `Order summary` titles are aligned.
  * The `Express checkout` box border is 1px wide, like in the [new designs](https://user-images.githubusercontent.com/3616980/83534129-c0161380-a4f0-11ea-985f-851b40d2e92b.png).

### Changelog

> Remove background color from Express Checkout title.

### DIFF
--- a/assets/js/base/components/payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/style.scss
@@ -20,6 +20,7 @@ $border-width: 1px;
 			border-radius: 5px 0 0 0;
 			content: "";
 			display: block;
+			margin-right: $gap-small;
 			pointer-events: none;
 			width: #{$gap-larger - $gap-small - $border-width * 2};
 		}
@@ -31,6 +32,7 @@ $border-width: 1px;
 			border-radius: 0 5px 0 0;
 			content: "";
 			display: block;
+			margin-left: $gap-small;
 			pointer-events: none;
 			flex-grow: 1;
 		}
@@ -38,8 +40,6 @@ $border-width: 1px;
 
 	.wc-block-components-express-checkout__title {
 		flex-grow: 0;
-		padding-left: $gap-small;
-		padding-right: $gap-small;
 		transform: translateY(-50%);
 	}
 
--- a/assets/js/base/components/payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/style.scss
@@ -13,6 +13,7 @@ $border-width: 1px;
 		top: 0;
 		vertical-align: middle;
 
+		// Pseudo-elements used to show the border before and after the title.
 		&::before {
 			border: $border-width solid transparent;
 			border-left-color: currentColor;
--- a/assets/js/base/components/payment-methods/style.scss
+++ b/assets/js/base/components/payment-methods/style.scss
@@ -1,25 +1,58 @@
+$border-width: 1px;
+
 .wc-block-components-express-checkout {
 	margin: auto;
-	border: 2px solid $black;
-	border-radius: 5px;
-	padding: 8px;
 	position: relative;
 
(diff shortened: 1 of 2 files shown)


assets/js/base/components/payment-methods/style.scss
```diff
@@ -31,15 +32,14 @@ $border-width: 1px;
...
+			margin-left: $gap-small;
 			pointer-events: none;
 			flex-grow: 1;
 		}
 	}
 
 	.wc-block-components-express-checkout__title {
 		flex-grow: 0;
-		padding-left: $gap-small;
-		padding-right: $gap-small;
```
nerrad: I can verify this change fixes the gap on the express checkout title:

Image 2020-06-12 at 3 42 43 PM

assets/js/base/components/payment-methods/style.scss
```diff
@@ -1,25 +1,58 @@
...
+			border-radius: 5px 0 0 0;
+			content: "";
+			display: block;
+			margin-right: $gap-small;
+			pointer-events: none;
+			width: #{$gap-larger - $gap-small - $border-width * 2};
+		}
+
+		&::after {
+			border: $border-width solid transparent;
```
haszari, line 29: Are these pseudo elements used to create the "padding" around the title (i.e. overlay the other express payment border? Might be worth a little comment explaining the intention/approach, or maybe package as a mixin (or class) if that's practical.

assets/js/base/components/payment-methods/style.scss
Aljullu: We had two selectors with the same specificity competing. The issue happened because depending on the build mode, stylesheets were loaded in a different order, so the styles applied where different.

> If this is a style loading issue, should we document this in an issue for further investigation?

Created an issue: #2713.

assets/js/base/components/payment-methods/style.scss
Aljullu: Good point. Added a comment: 2b6f524a1f7d8a0fcc67b4fc5f638de98a8271ac. I don't think we need to create a mixin for it yet considering we are only using it here. We can extract it in the future if we see there are other instances where we use it.

Aljullu:
> It's tricky to get control of the whitespace around the heading when there is the express payment box behind! I'm keen to better understand the technique you've used :)

@haszari Yeah, I wasn't even sure it was doable at all, but then I found a way. Basically the border is composed by the border of three elements:
- The container: has left, bottom and right border (in the screenshot, the blue border). It can't have top border because it would appear behind the `Express checkout` title.
- The title `::before` pseudo-element: has left and top border (in the screenshot, the red border).
- The title `::after` pseudo-element: has right and top border (in the screenshot, the orange border).
imatge

Hope this explanation makes sense. :slightly_smiling_face:  The CSS solution was a bit tricky, but I don't think it will make it more difficult for themes, because border color is inherited from the text color, so it will change all at once.

assets/js/base/components/payment-methods/style.scss
haszari: Thanks for the visual explanation below, ingenious fix!

PR #2707, opened 2020-06-12
Generate block pages for tests.

senadir:
**TIL:** This PR changes the way we insert and create blocks in our tests, instead of inserting the block each time and comparing it with a snapshot, a very strict process, overlay strict, that causes issues, it will attempt to open a page with that block already there and confirm that it loads fine, so it delegates the validation to Gutenberg, the thing we can about in the first place.

### What is introduced:

- a new alias package `@woocommerce/blocks-test-utils` that will hold all of our test utils, a  growing collection (in other branches at least).
- an E2E utility function `visitBlockPage`, that takes a string, in the shape of `BLOCK_NAME Block`, it will attempt to:
    - Search for that page.
    - If it exists, it will visit it.
    - If it doesn’t exist, it will:
        - Create a new page.
        - Inserts that block.
        - Get the page content.
        - Generate a new `block-name.fixture.json` that contains the title and content.
        - Save it on `__fixtures__` folder on the same level as the test that caused it.
        - Block Fixtures are treated like snapshots and should committed.
- Expand our fixture loaders with two functions `createBlockPages` and `deleteBlockPages`.
    - `createBlockPages` will loop our `spec` folder (the place we hold our E2E tests in), searching for `*.fixture.json` files, and generating pages preloaded with that content.
    - `deleteBlockPages` will take an array of pages IDs and delete them (to keep tests clean).
- Refactor our existing E2E tests in some ways  (that I hope to abstract further):
    - All tests have a `block` object containing the block title, slug, and classname, those are helpful to copy paste code between tests.
    - All tests follow the pattern of:
        1. Calling  `visitBlockPage` beforeAll.
        2. Ensure the block renders fine (no Gutenberg crash, no ErrorBoundary crash, no validation crash, and the block actually rendering), I will later abstract those to a jest matcher, so basically `await expect( page ).toRenderBlock( block );`
        3. Ensure test can be inserted more than once or only once.
    - Entering the block page happens once per block, previously it was once per test, **this is important**, as your test will break if another test change something, take `product-search`, one test will hide the title, the next test will try to edit the title text but it will break because the title is hidden, **you should always remember to reset the changes your test does.**
I went with this because it drastically reduces our test duration, each test takes less than 500ms to be done, if we go all the process, it will take up to 5 seconds.

with `beforeEach`
```
  All Products Block
    ✓ can only be inserted once (4054ms)
    ✓ renders without crashing (2432ms)
```

with `beforeAll`
```
 PASS  tests/e2e-tests/specs/backend/all-products.test.js (7.433s)
  All Products Block
    ✓ can only be inserted once (374ms)
    ✓ renders without crashing (172ms)
```

Now it might not seem like much, we still have some manual product import, and few tests, but with two other branches coming, we will test 28 block, with 5 or 6 tests per block, spanning 4 WordPress configurations.

### How to test

- Tests should pass in travis.
- Locally you could try running `npm run test:e2e` after starting docker with `npm run test:e2e:up:build` give it some time to be fully up.
- You can try to delete the `__fixtures__` and running the tests to see if it was generated.
- You can see a test visually by running `npm run test:e2e-dev PATH_TO_TEST`.

### Follows ups:

- Create a jest matcher to abstract some tests. #2709
- Document the usage of those patterns, including how to write E2E tests and reasoning behind them.
- Write a P2 of how we write those tests.
- Document the Fixture pattern.


tests/e2e-tests/specs/backend/product-search.test.js
```diff
@@ -2,37 +2,67 @@
...
 
-		expect( await getEditedPostContent() ).toMatchSnapshot();
+		await expect( page ).toMatchElement( block.class );
 	} );
 
 	it( 'can toggle field label', async () => {
-		await insertBlock( 'Product Search' );
+		await openDocumentSettingsSidebar();
+		// we focus on the block
+		await page.click( block.class );
```
senadir, line 54: In future versions of GB, the settings sidebar is closed by default, see open it anyway.
also when visiting a page, no block is focused by default, so we click on our block to see the settings.

tests/e2e-tests/specs/backend/single-product.test.js
```diff
@@ -0,0 +1,58 @@
...
+		await switchUserToAdmin();
+		await visitBlockPage( `${ block.name } Block` );
+	} );
+
+	it( 'can be inserted more than once', async () => {
+		await insertBlock( block.name );
+		expect( await getAllBlocks() ).toHaveLength( 2 );
+		await page.keyboard.press( 'Tab' );
+		await page.keyboard.press( 'Delete' );
+		expect( await getAllBlocks() ).toHaveLength( 1 );
```
senadir, line 33: To not create conflict with other tests, we delete the newly created second block, I'd argue this test is not needed because all blocks can be inserted more than once by default.

tests/utils/visit-block-page.js
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	if ( isFullscreenMode ) {
+		await page.evaluate( () =>
+			wp.data
+				.dispatch( 'core/edit-post' )
+				.toggleFeature( 'fullscreenMode' )
+		);
+	}
+}
```
senadir, line 36: discovering I need this took me a lot of time, you can’t click the inserter without toggling off the welcome guide, same for fullscreenMode.
When creating a new post, this is done directly in the `createNewPost` helper, since we navigate to this post, we need to do it.
I didn’t test without fullscreen, so I might try deleting it and seeing if the test still pass.

tests/utils/visit-block-page.js
```diff
@@ -0,0 +1,75 @@
...
+export async function visitBlockPage( title ) {
+	let link = '';
+	// Visit Import Products page.
+	await visitAdminPage( 'edit.php', 'post_type=page' );
+	// search for the page.
+	if ( await page.$( '#post-search-input' ) ) {
```
senadir, line 42: when there are no pages whatsoever, `#post-search-input` is not rendered.

tests/utils/visit-block-page.js
```diff
@@ -0,0 +1,75 @@
...
+		await page.type( '#post-search-input', title );
+		await page.click( '#search-submit', { waitUntil: 'domcontentloaded' } );
```
senadir, line 44: when navigating to another page, we always wait for dom to load before continuing.

tests/utils/visit-block-page.js
```diff
@@ -0,0 +1,75 @@
...
+		const pageLink = await page.$x( `//a[contains(text(), '${ title }')]` );
+		if ( ( await pageLink.length ) > 0 ) {
+			link = await page.evaluate(
+				( a ) => a.getAttribute( 'href' ),
+				pageLink[ 0 ]
+			);
```
senadir, line 50: clicking the link directly seems to cause racing issues, so I opted for getting the link and passing it to `page.goto` as it handles waiting better.

tests/utils/visit-block-page.js
```diff
@@ -0,0 +1,75 @@
...
+		}
+	}
+	if ( link ) {
+		await visitPage( link );
+	} else {
```
senadir, line 55: if the page already exists, visit it, if not, we go to the fixture creating part.

tests/utils/visit-block-page.js
```diff
@@ -0,0 +1,75 @@
...
+		await createNewPost( { postType: 'page', title } );
+		await insertBlock( title.replace( /block/i, '' ).trim() );
+		const pageContent = await getEditedPostContent();
+		await outputFile(
+			`${ dirname(
+				module.parent.parent.filename ||
+					module.parent.filename ||
+					module.filename
```
senadir, line 63: Depending on how you call the test `npm run test:e2e` or `npm run test:e2e test/path`, we need to reach to that test path to generate the file on the same level, generating on the same level is simply a cosmetic choice to keep code together.

tests/utils/visit-block-page.js
senadir: Fullscreen was not needed, deleted in 9d6ee8e

tests/e2e-tests/fixtures/fixture-loaders.js
```diff
@@ -249,6 +253,51 @@ const deleteShippingZones = ( ids ) => {
...
 };
 
+const createBlockPages = () => {
+	return glob( `${ dirname( __filename ) }/../specs/**/*.fixture.json` )
+		.then( ( files ) => {
+			return Promise.all( files.map( ( file ) => readJson( file ) ) );
+		} )
+		.then( ( files ) => {
+			return Promise.all(
+				files.map( ( { title, pageContent: content } ) => {
```
nerrad, line 263: Seems like there might be some unnecessary additional looping here. Could you make the initial map an async call and thus only have to loop through the files once? Something like this:

```js
return Promise.all( files.map( async( file ) => { const file = await readJson( file ); /* rest of code here */ }
```

tests/e2e-tests/specs/backend/all-products.test.js
```diff
@@ -3,20 +3,53 @@
...
+			( await page.content() ).match(
+				/There was an error whilst rendering/gi
+			)
+		).toBeNull();
+		// Validation Error
+		expect(
+			( await page.content() ).match(
+				/This block contains unexpected or invalid content/gi
+			)
+		).toBeNull();
```
nerrad, line 51: looks like this is something that could be extracted to a helper at some point (not necessarily in this pull)

tests/e2e-tests/specs/backend/cart.test.js
```diff
@@ -3,28 +3,53 @@
...
-describe( 'Cart Block', () => {
-	beforeEach( async () => {
-		await switchUserToAdmin();
-		await createNewPost();
-	} );
+import { visitBlockPage } from '@woocommerce/blocks-test-utils';
+
+/**
+ * Internal dependencies
+ */
```
nerrad, line 14: looks like this isn't needed because there are no internal dependencies? This applies to other specs as well, I suspect it's cruft from when you switched to using the `@woocommerce/blocks-test-utils` alias?

tests/e2e-tests/specs/backend/single-product.test.js
nerrad: Ya the test is probably not needed.

tests/utils/visit-block-page.js
nerrad: These pull request comments were super helpful, but I think it'd be nice to put them inline in the code for our future selves (or future maintainers).

tests/e2e-tests/fixtures/fixture-loaders.js
senadir: I was trying to avoid adding async/await to this file so it can run while testing or directly via node or in other instances without needing to transpile it.
So basically there are two flat loops here (not nested loops), this is not much different from doing async as it has to wait for each file.

nerrad: I think you may be missing my point here. I'm not saying `createBlockPages` has to return a promise. I never pointed out nested loops I'm pointing out that your looping through the set of `files` _twice_. Once to get the JSON, and then again on all the JSON objects to do the request based on them. That's unnecessary imo I it looks to me like you could do this in **one** loop.

PR #2730, opened 2020-06-17
Switch to use `wp-env` as the development/test environment

nerrad:
This pull is an attempt to switch away from the custom docker environment to use the [`@wordpress/env`](https://github.com/WordPress/gutenberg/tree/master/packages/env) instead.

Some advantages with using `wp-env` are:

- We utilize the work of multiple developers that are constantly improving this package.
- The availability of both a *test* environment **and** a developer environment when executing `npm run wp-env start`. The test environment is on a different port than the dev environment and uses a different db. Basically out of the box developers can get up and running with a development environment for blocks.
- The devx for wp-env (imo) is much nicer (especially when perusing builds on travis because there isn't a whole slew of output from the build).
- Environment setup appears to be slightly faster.
- Configuration is much more straightforward (imo).

## Issues encountered.

While I still prefer `@wordpress/env` there were some issues encountered that needed resolved in order for this to work on Travis builds:

- Although `wp-env` supports phpunit (not officially in their docs yet), I couldn't get it to work either locally or in travis. We can't use it out of the box anyways because of our requirement for WooCommerce phpunit test utils. For now I left the existing docker configuration for phpunit in place (Note: it was never actually running on travis!)
- There are permissions issues with using a non default WP version for `wp-env` builds (see [this comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2730#issuecomment-645469466). As a workaround for now, I've mapped a `.htaccess` file from the bin folder to the wp-env so that it picks up our rewrite rules change as a part of the setup. This can be revisited once the related Gutenberg Issue is addressed (and means we can't do any tests involving uploading files).


nerrad:
There's still some things I'm working on here that I'll resume tomorrow:

* [ ] switch `phpunit` to use `wp-env` (in progress locally)
* [ ] The WP version selector for wp-env is still not correct (which is likely why the snapshot is failing) (in progress locally)
* [ ] remove all traces of old docker environment.

nerrad:
After spending some time troubleshooting things, it looks like the issue here is that when the installed version of WordPress is changed from the default, (via the `core` config property in wp-env config), then there are permissions issues with the files. This in turn seems to affect any wp-rewrite changes (presumably because `.htaccess` isn't updated) and the ability to install Gutenberg plugin after the container is spun up (because `plugins` folder permissions have changed).

The Gutenberg issue could likely be resolved by modifying the configuration on spinning up the environment, but the rewrite will be more difficult (assuming this is the cause, it's the only thing that makes sense atm).

nerrad:
related: https://github.com/WordPress/gutenberg/issues/22515

So I think one way to work around this will be to manually map an .htaccess file on initialization. I'll try that out when I resume this.

.wp-env.json
```diff
@@ -0,0 +1,13 @@
...
+		"https://downloads.wordpress.org/plugin/woocommerce.latest-stable.zip",
+		"https://github.com/WP-API/Basic-Auth/archive/master.zip",
+		"."
+	],
+	"themes": [
+		"https://downloads.wordpress.org/theme/storefront.latest-stable.zip"
+	],
+	"mappings": {
+		".htaccess": "./bin/.htaccess"
+	}
```
nerrad, line 12: This is a workaround for the permissions issue with non default WP versions in builds. In those builds the `.htaccess` file isn't created because of broken permissions in the wp-cli container. By mapping an already prepared .htaccess file we avoid the need to write the file during the setup phase.

bin/wp-env-config.sh
```diff
@@ -0,0 +1,19 @@
...
+npm run wp-env run tests-cli wp user get customer 2> /dev/null
+
+## if 0 is the exit code then we can leave otherwise we'll try creating the user
+if [ $? -eq 0 ]
+then
+	EXIT_CODE=0
+else
+  npm run wp-env run tests-cli wp user create customer customer@woocommercecoree2etestsuite.com --user_pass=password --role=customer
+  EXIT_CODE=$?
+fi
```
nerrad, line 13: This captures exit codes from wp commands so we don't create a fail condition in travis (unless there definitely is something that goes wrong).

tests/e2e-tests/specs/backend/__fixtures__/product-search.fixture.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"title":"Product Search Block","pageContent":"<!-- wp:woocommerce/product-search {\"formId\":\"wc-block-product-search-5\"} -->\n<div class=\"wp-block-woocommerce-product-search\"><div class=\"wc-block-product-search\"><form role=\"search\" method=\"get\" action=\"http://localhost:8889/\"><label for=\"wc-block-product-search-5\" class=\"wc-block-product-search__label\">Search</label><div class=\"wc-block-product-search__fields\"><input type=\"search\" id=\"wc-block-product-search-5\" class=\"wc-block-product-search__field\" placeholder=\"Search products…\" name=\"s\"/><input type=\"hidden\" name=\"post_type\" value=\"product\"/><button type=\"submit\" class=\"wc-block-product-search__button\" label=\"Search\"><svg aria-hidden=\"true\" role=\"img\" focusable=\"false\" class=\"dashicon dashicons-arrow-right-alt2\" xmlns=\"http://www.w3.org/2000/svg\" width=\"20\" height=\"20\" viewbox=\"0 0 20 20\"><path d=\"M6 15l5-5-5-5 1-2 7 7-7 7z\"></path></svg></button></div></form></div></div>\n<!-- /wp:woocommerce/product-search -->"}
```
nerrad: for some reason the `wp-env` context changed this fixture (and the accompanying snapshot). Fortunately it looks like things still passed on travis with this change (across all WP versions). So at least we're good there.

.env
```diff
@@ -5,8 +5,8 @@ WORDPRESS_DB_USER=wordpress
 WORDPRESS_DB_PASSWORD=wordpress
 
 # WordPress CLI environment
-WORDPRESS_PORT=8084
-WORDPRESS_BASE_URL=http://localhost
+WORDPRESS_PORT=8889
+WORDPRESS_BASE_URL=http://localhost:8889
```
senadir, line 9: any specific reason for this port?

.travis.yml
```diff
@@ -78,44 +107,34 @@ jobs:
...
               - WP_VERSION=5.4
+              - E2E_TESTS=1
               - WOOCOMMERCE_BLOCKS_PHASE=3
         - name: E2E Tests (WP 5.4 with Gutenberg plugin)
           script:
-              - npm ci
-              - npm run test:e2e:up
-              - composer install
-              - npm run build:e2e-test
+              - npm run wp-env run tests-cli wp plugin install gutenberg --activate
```
senadir: should this be `npm run wp-env run tests-cli "wp plugin install gutenberg --activate"`?

bin/wp-env-config.sh
```diff
@@ -0,0 +1,19 @@
...
+then
+	EXIT_CODE=0
+else
+  wp user create customer customer@woocommercecoree2etestsuite.com --user_pass=password --role=customer
+  EXIT_CODE=$?
+fi
+
+## set permalinks for easier wp-json
+wp rewrite structure '/%postname%/'
+wp plugin list
```
senadir, line 17: this should be deleted.

package.json
```diff
@@ -49,17 +49,14 @@
 		"storybook": "start-storybook  -c ./storybook -p 6006 --ci",
 		"storybook:build": "build-storybook  -c ./storybook -o ./storybook/dist",
 		"test": "wp-scripts test-unit-js --config tests/js/jest.config.json",
-		"test:e2e:down": "docker-compose down",
-		"test:e2e:down:reset": "docker-compose down -v",
-		"test:e2e:ssh": "docker exec -it woo-blocks_wordpress-www_1 /bin/bash",
-		"test:e2e:up": "docker-compose up -d",
-		"test:e2e:up:build": "docker-compose up --build -d --quiet-pull",
```
senadir, line 56: We should probably document later how can you start and stop the docker instance
`npm run wp-env start` `npm run wp-env stop`

tests/e2e-tests/fixtures/fixture-loaders.js
```diff
@@ -103,13 +105,14 @@ const deleteCoupons = ( ids ) =>
 /**
  * Create Products and call createReviews.
  *
+ * currently this only creates a single product for the sake of reviews.
+ *
+ * @todo  add more products to e2e fixtures data.
```
senadir, line 110: moving this todo created a new todo.

tests/e2e-tests/specs/backend/__fixtures__/product-search.fixture.json
senadir, line 1: Not sure why those updated to 5 btw, in a pristine instance of the editor, they should be 2, could be that we're not deleting local storage (probably not).

.env
nerrad: It's the default set for the test environment by `wp-env`

.travis.yml
nerrad: yup updated :)

bin/wp-env-config.sh
nerrad: why? Permalinks is necessary and `wp plugin list` helps verify the plugins that were installed and activated.

package.json
nerrad: Agreed, and likely we can just point to the package docs (so folks can see all the other stuff they can do beyond `start` and `stop`)

tests/e2e-tests/fixtures/fixture-loaders.js
nerrad: it created a comment, but it shouldn't create a new issue on merge (because the title will match) _or_ reopen the existing issue if it was closed. At least that's how it _should_ work (we'll see!)

tests/e2e-tests/specs/backend/__fixtures__/product-search.fixture.json
nerrad: The very first time I ran `wp-env` locally the update surfaced. Honestly it's a bit of a mystery to me why the numbers change too. This definitely needs a closer look in a follow-up.

bin/wp-env-config.sh
senadir: I was referring to `wp plugin list`, we only used to test if the plugin was installed, if you think we need it for future reference, we can keep it.

timnolte:
@nerrad have you figured out a fix for running PHP Unit under wp-env? I'm trying to get this setup for a plugin myself right now.

PR #2742, opened 2020-06-17
Auto-merge renovate dependency updates (except major versions)

### LINKED ISSUE #2625: Standardise and streamline RenovateBot dependency updates
We are handling RenovateBot dependency update PRs ad hoc. Some of these we may be able to automerge (which will save us time and get us updated faster). We may still need a process to smoke test or otherwise check this - could be a porter or weekly responsibility, or part of release. For bigger updates (major), we might want some process to ensure we keep on top of them, possibly tied to releases or cooldown/sprint cycle.

This issue is triggered from @nerrad comment in #2609:

> > What's our Renovate config - is it worth setting up automerge for these PRs, and QAing them as part of regular work and release process?

> Ya we probably could setup auto-merge for patch bumps. Our renovate config could probably use some fine-tuning for not only auto-merge on patch bumps but also investigate creating pulls for bumping major versions for a few of our dependencies. I think our current strategy only bumps within the pinned version (except for jetpack-autoloader) which means we're potentially getting way behind in some updates to packages (which can make for more painful updates later on if we need to update).

- Team process for how we handle dependency updates/upgrades.
- Enable automerge for Renovate for low-risk updates.

haszari:
Fixes #2625

Enables automerge for all updates except major, using the approach from [renovate docs](https://docs.renovatebot.com/configuration-options/#automerge).

I also removed our [schedule config](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/598), as there's no need to defer automerges to a particular part of the week. We can roll this back if people prefer a schedule. It's possible that dependency updates could cause issues when we're in the middle of a release. cc @Aljullu @mikejolley for thoughts on this.

This is the expected behaviour for this config, I think we have to merge to confirm this (!):

- PRs will be generated for major updates, as before. We'll need to manually test, review and merge these.
- All other updates will be automerged immediately.

Regarding process, here's what I'm thinking:

- For major updates, let's stick with the current auto-round-robin review approach. The assigned reviewer can reassign or ping others if needed.
- I've added a new step to the release process to note any updates and ensure they've been tested/reviewed.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,6 +2,7 @@
 	"extends": [ "config:base" ],
 	"lockFileMaintenance": { "enabled": true },
 	"ignoreDeps": [ "husky" ],
+	"schedule": [ "before 3am on wednesday" ],
 	"composer": {
 		"enabled": false
 	},
--- a/renovate.json
+++ b/renovate.json
@@ -6,11 +6,15 @@
 	"composer": {
 		"enabled": false
 	},
-  "labels": ["type: dependencies", "skip-changelog"],
+	"labels": ["type: dependencies", "skip-changelog"],
 	"packageRules": [
 		{
 			"packageNames": [ "automattic/jetpack-autoloader" ],
 			"rangeStrategy": "bump"
-		}
-	]
+		},
+    {
+      "updateTypes": ["minor", "patch", "pin", "digest"],
+      "automerge": true
+    }
+  ]	]
 }
--- a/renovate.json
+++ b/renovate.json
@@ -2,7 +2,6 @@
 	"extends": [ "config:base" ],
 	"lockFileMaintenance": { "enabled": true },
 	"ignoreDeps": [ "husky" ],
-	"schedule": [ "before 3am on wednesday" ],
 	"composer": {
 		"enabled": false
 	},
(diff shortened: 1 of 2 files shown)


haszari:
@nerrad I tweaked the heading levels in the releases doc, can you check these? The `Create a release pull request` has sub-sections for different scenarios, and they were H3, now nested correctly as H4. But I could be reading this wrong :)

haszari:
Note to reviewers - if you think this is good to go please merge, no need to wait for me :)

renovate.json
```diff
@@ -2,15 +2,18 @@
...
 			"packageNames": [ "automattic/jetpack-autoloader" ],
 			"rangeStrategy": "bump"
-		}
-	]
+		},
+    {
+      "updateTypes": ["minor", "patch", "pin", "digest"],
+      "automerge": true
+    }
+  ]	]
```
nerrad, line 18: There is an errant extra bracket here and the indenting is off.

haszari: Woah, good spotting. I think I need to change my json colour scheme - it seems to hide anything erroneous like this. 🤦 

Now fixed :)

PR #2745, opened 2020-06-18
Remove text color from Cart and Checkout blocks

### LINKED ISSUE #2459: Replace all hardcoded font-color declarations with alternatives in the Cart and Checkout blocks

### LINKED ISSUE #2540: Cart block: discounted prices should be wrapped in a <del> tag
In WC Core and some other WC Blocks, when a product is on sale the regular price is wrapped in a `<del>` tag and the discounted price is inside an `<ins>` tag.
imatge

However, in the Cart block they are all wrapped in `<span>`:
imatge

We still want to preserve the screen reader text (`Previous price:` and `Discounted price:`, because most screen readers don't announce `<del>` tags, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del)).

Aljullu:
Fixes #2459.
Fixes #2540.

A couple of notes:
* According to the designs, borders are not fully opaque. There wasn't an easy way to use `currentColor` for borders and at the same time reduce its opacity without reducing the children elements opacity too. The best solution I could find was to render a pseudoelement responsible for displaying the borders and reduce the opacity there. I created a SCSS mixin to avoid duplicating the code in several elements.
* Removing colors meant that discounted prices were no longer light gray. To solve that, I wrapped the ProductPrice component with the legacy `price` class so we would inherit theme styles.

### Screenshots

_Storefront (default):_
imatge

_Storefront with different colors:_
imatge

_Bistro:_
imatge

1. With Storefront, go to Appearance > Customize and change the typography color.
2. Verify the color is applied to the Cart and Checkout text and borders.
3. Test other themes to verify there are no regressions.

### Changelog

> The Cart and Checkout blocks now use the font colors provided by the theme.

### DIFF
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -122,6 +122,27 @@ $fontSizes: (
 	-ms-word-break: break-all;
 }
 
+// Shows a border with the current color and a custom opacity. That can't be achieved
+// with normal border because `currentColor` doesn't allow tweaking the opacity, and
+// setting the opacity of the entire element would change the children's opacity too.
+@mixin with-translucent-border($border-width: 1px, $opacity: 0.3) {
+	position: relative;
+
+	&::after {
+		border-style: solid;
+		border-width: $border-width;
+		bottom: 0;
+		content: "";
+		display: block;
+		left: 0;
+		opacity: $opacity;
+		pointer-events: none;
+		position: absolute;
+		right: 0;
+		top: 0;
+	}
+}
+
 // Converts a px unit to em.
 @function em($size, $base: 16px) {
 	@return $size / $base * 1em;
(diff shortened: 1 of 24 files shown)


haszari:
Saw a symptom of this when testing Storefront - the cart uses black/grey, checkout uses customizer heading colours.

Screen Shot 2020-06-19 at 1 06 33 PM

Screen Shot 2020-06-19 at 1 07 14 PM

Adding the screenshots so we can check this off - 🤞 fixed in this PR :)

assets/css/abstracts/_mixins.scss
```diff
@@ -122,6 +122,27 @@ $fontSizes: (
...
+		content: "";
+		display: block;
+		left: 0;
+		opacity: $opacity;
+		pointer-events: none;
+		position: absolute;
+		right: 0;
+		top: 0;
+	}
+}
```
senadir, line 144: Still fail to understand how this work, at what point are we inheriting the color and where's getting assigned to?

PR #2748, opened 2020-06-18
Cart & Checkout: show shipping method prices with taxes when required

### LINKED ISSUE #2626: Shipping costs in cart not respecting tax settings
When shipping costs are set to include tax in WC settings TAX tab, they are erroneously shown without tax in the listing of shipping options in the cart summary.

Aljullu:
Fixes #2626.

1. Go to _WooCommerce_ > _Settings_ > _Tax_ > _Tax options_ and set _Display prices during cart and checkout_ to _Including tax_:
imatge
2. Set a flat rate shipping method with cost 5:
imatge
3. Set default tax rates to 10%:
imatge
4. Now, as a customer, add a product that needs shipping to your cart and visit the _Cart_ page (with the block).
5. Verify the shipping method options show the prices with taxes.

| `main` | This branch |
| --- | --- |
| imatge | imatge |

6. Go to the Checkout page and verify shipping method prices also appear with taxes.

### Changelog

> Cart and Checkout blocks display shipping methods with tax rates if that's how it's set in the settings.

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/totals-shipping-item/shipping-rate-selector.js
+++ b/assets/js/base/components/cart-checkout/totals/totals-shipping-item/shipping-rate-selector.js
@@ -6,23 +6,31 @@ import FormattedMonetaryAmount from '@woocommerce/base-components/formatted-mone
 import { decodeEntities } from '@wordpress/html-entities';
 import { getCurrencyFromPriceResponse } from '@woocommerce/base-utils';
 import { ShippingRatesControl } from '@woocommerce/base-components/cart-checkout';
+import { DISPLAY_CART_PRICES_INCLUDING_TAX } from '@woocommerce/block-settings';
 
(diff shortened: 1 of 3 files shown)


Aljullu:
Thanks for the review @haszari.

> One thought, perhaps we can share more of the logic for `renderShippingRatesControlOption`? Looks like the version in `checkout/` is mostly the same but doesn't have the `delivery time` stuff. Could be an option (prop) on a shared component, or maybe they should be consistent. If you think this is worthwhile perhaps we can add a follow up issue.

Yeah... There are some slight differences in the way options are rendered between the Cart and Checkout. The `renderShippingRatesControlOption` adds a lot of flexibility but might be overcomplicating things unnecessarily (we could instead just pass a `compact` prop and render one layout or the other depending on it?).

I will leave it for now, so we can get this merged into 2.8, but we can improve it in the future if necessary.

PR #2753, opened 2020-06-19
optionally render product description in best-sellers block

### LINKED ISSUE #1286: Allow for toggling product description field in Product blocks.
It would be nice if the products of the product blocks can also show their product description.

An option for enabling/disabling product descriptions.

**Describe alternatives you've considered**
There aren't really alternatives.

haszari:
Fixes #1286

This adds a new option to the Best Sellers block for displaying the short description text. This defaults to hidden, consistent with previous behaviour.

This is somewhat similar to the forthcoming behaviour for customising the content displayed in the [(single) Product block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2379). In that block, there are element blocks that can be selected and moved around for things like price, title, description, and these are implement as react components and rendered on the front end. The existing "product grid" blocks are implemented using PHP and rendered on the server. 

In this PR, I've added support to the PHP server-side render (SSR) code for the description field (just for `Best Sellers` to start). I'm keen for feedback about whether we should continue and implement this across all these SSR grid blocks, or whether we should hold off in anticipation of using react/front-end render for these blocks too. cc @nerrad @mikejolley for thoughts on this. 

#### Accessibility

### Screenshots

Screen Shot 2020-06-19 at 3 53 46 PM

Screen Shot 2020-06-19 at 3 54 13 PM

1.
2.
3.

### Changelog

### DIFF
--- a/assets/js/components/grid-content-control/index.js
+++ b/assets/js/components/grid-content-control/index.js
@@ -11,6 +11,8 @@ import { ToggleControl } from '@wordpress/components';
  */
 const GridContentControl = ( { onChange, settings } ) => {
 	const { button, price, rating, title } = settings;
+	const description = settings?.description ?? false;
+
 	return (
 		<Fragment>
 			<ToggleControl
@@ -63,6 +65,27 @@ const GridContentControl = ( { onChange, settings } ) => {
 			/>
 			<ToggleControl
 				label={ __(
+					'Short description',
+					'woo-gutenberg-products-block'
+				) }
+				help={
+					description
+						? __(
+								'Product description is visible.',
+								'woo-gutenberg-products-block'
+						  )
+						: __(
+								'Product description is hidden.',
+								'woo-gutenberg-products-block'
+						  )
+				}
+				checked={ description }
+				onChange={ () =>
+					onChange( { ...settings, description: ! description } )
+				}
+			/>
+			<ToggleControl
+				label={ __(
 					'Add to Cart button',
 					'woo-gutenberg-products-block'
 				) }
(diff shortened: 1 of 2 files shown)


haszari:
Added to QA/review pipeline for feedback about approach (PHP SSR vs react element blocks).

mikejolley:
Are we not hoping to move away from these blocks in the future? If we are, I don't think we should be adding features to the SSR blocks because it causes potential breaking changes in the future as we try to migrate users across.

If we do introduce this, 1) it needs to be off by default and 2) we shouldn't confuse "description" with "short description" in the code behind the scenes.

Additionally, if it is off by default, how many users are likely to benefit or want this? Is it achievable with a snippet right now instead?

haszari:
Closing this PR. 

Our focus is adding features & flexibility to the blocks using inner blocks approach – e.g. `All Products` and in-progress `Single Product` block.

In future we may look revamping the best sellers and other related server-side render blocks to use inner blocks.

helgatheviking:
For what it's worth, I also needed short descriptions in the hand-picked products block over the weekend and was just about to submit a PR for this (I think I changed the same two files) but I found this one first.

>Are we not hoping to move away from these blocks in the future? 

Are you? If so, is that public somewhere so folks will know whether to dedicate time to extending it or not. And... where is this under development so we can take a look?

I was able to add the short description via a [PHP snippet](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1286#issuecomment-688483230) but there's no way to add a UI for it like proposed in this PR.

PR #2755, opened 2020-06-19
Use WooCommerce thumbnail images in Blocks

### LINKED ISSUE #2719: Cart, Checkout and Review blocks should use product thumbnail image instead of full-size image
Reported as an idea in:
https://ideas.woocommerce.com/forums/133476-woocommerce/suggestions/40644121-review-image-used-is-currently-the-large-image-it

**To Reproduce**
1. Add a reviews block (any of them).
2. In the sidebar options, select _Product_ as the _Review image_ attribute.
3. Visit the page in the frontend and inspect the image.
4. Notice the image has the original width, even though we are rendering it in a 48x48px size.

**Screenshots**
imatge
imatge

### LINKED ISSUE #2728: Several of our blocks don't honor 'Thumbnail cropping' setting
**To Reproduce**
1. Create a product with a big vertical image.
2. Go to Appearance > Customize > WooCommerce > Product Images
3. Select the option to crop the images as 1:1.
4. Go to a page with some of our blocks and notice most of them don't honor the setting:
  * All Products show the image without cropping:
imatge
  * Same for the Cart and Checkout blocks:
imatge
imatge
  * Review blocks product images are squared in the frontend but, if you select a custom cropping size in the settings, they don't honor it. So they seem to be affected by this bug as well. Also, even with 1:1 cropping they don't look squared in the editor:
    * _Editor_:
imatge:
    * _Frontend_:
imatge

PHP-based product grid blocks correctly show the images cropped.

**Screenshots**
For reference, this is how it looks in the Shop page:
imatge

**Expected behavior**
All blocks should honor the settings and display images cropped if that's the selected setting.

**Related issue:** #2719.

Aljullu:
Fixes #2719.
Fixes #2728.

This PR updates several blocks (Cart, Checkout and all Reviews blocks), so they use the WooCommerce thumbnail instead of the full-size image. It's smaller, so it should be more performant, and it honors the cropping settings set in the customizer.

It also adds an _Image sizing_ attribute to the Product image atomic block. So All Products and, in the future, Product block users will be able to choose between the full size image and the cropped one.

cc @LevinMedia I added this copy to the new attribute, let me know if it makes sense or we should use something else:
Screenshot

### Screenshots

_Cart block with images cropped to 1:2_
Screenshot

1. Go to Appearance > Customize > WooCommerce > Product images and change the cropping options.
2. Test the Cart, Checkout and Review blocks (for Review blocks, you might need to change its attributes so it shows the product image instead of the customer image) and verify they show the cropped image.
3. Edit an old All Products block and verify the block didn't invalidate.
4. Edit it and select the Product image inner block. There, toggle the _Image sizing_ attribute.
5. Verify when _Cropped_ is selected, the cropped image is displayed.
6. Repeat the process with the Product block.

### Changelog

> Blocks now respect the product image cropping settings. For the All Products block, the user can switch between the cropped thumbnail and the full size image.


Aljullu:
When editing the All Products block, it's not possible to see the changes produced by the 'Image sizing' attribute:
Peek 2020-06-19 15-41

That's because we don't have a preview image that updates according to the customizer cropping settings. This will not be a problem when we migrate the All Products block to the new editing experience because it won't rely on preview data, so I don't think it should be blocking. But posting it here for awareness/feedback.

senadir:
@Aljullu why are some files deleted from this PR?

PR #2759, opened 2020-06-19
Experiment - Products API Fields Support

mikejolley:
Some changes from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2750 plus fields support for withProduct.


assets/js/shared/context/product-data-query.js
```diff
@@ -0,0 +1,53 @@
...
+		}
+		const fieldsToFetch = difference( fields, retrievedFields.current );
+
+		if ( fieldsToFetch.length === 0 ) {
+			return;
+		}
+
+		apiFetch( {
+			path: addQueryArgs( `/wc/store/products/${ productId }`, {
+				_fields: [ ...fieldsToFetch, 'id' ],
```
nerrad, line 40: So this means that _at a minimum_ this will return just the id. I wonder if we should default to the entire product response if no fields are provided (`undefined`)?  Otherwise consuming code has to be aware of all the fields available on the endpoint if all fields are desired?

assets/js/shared/context/product-data-query.js
```diff
@@ -0,0 +1,53 @@
...
+			} ),
+		} ).then( ( theProduct ) => {
+			setProductData( ( prevState ) => {
+				return { ...prevState, ...theProduct };
```
nerrad, line 44: So I'm wondering if there's potential danger here if in between requests old fields in the state might have changed on the server, but if they weren't requested in the latest request then there's a mix of old and new data?

assets/js/shared/context/product-data-query.js
```diff
@@ -0,0 +1,53 @@
...
+			} );
+			retrievedFields.current = concat( retrievedFields.current, fields );
+		} );
```
nerrad, line 47: Should add a catch here for any error handling.

assets/js/shared/context/product-data-query.js
mikejolley: There is some degree of risk if requests are spread apart, but I believe in our use case these requests are likely to be close. e.g. in testing I had the add to cart form do a second product request once the first request had determined product type.

Given the API responses are cached this could already be an issue right?

assets/js/shared/context/product-data-query.js
mikejolley: This ProductDataQuery component has a pretty specific use-case, so I don't think it's unreasonable to make consuming code handle this, remembering the consumer could also pass in `[]` if desired. Atomic blocks are the only consumers.

assets/js/shared/context/product-data-query.js
mikejolley: What should error handling look like? Throw an error to be caught higher up?

mikejolley:
@nerrad regarding all products block, I'm going to need some help here or in a follow up to support these changes since it's more complex (brought this up on p2). I think given these changes we should look at doing that before merging, just in case there are roadblocks with this approach. 

Can look at adding more tests once that is resolved.

mikejolley:
In retrospect, I don't think this change is worth the effort. It increases complexity for dubious gain. We'd be better served long term by looking at something like GraphQL if we want conditional queries. Thats not something we should do short term. We'd see more performance gains from improving core data structures for the things being queried.

PR #2763, opened 2020-06-23
Call add to cart in click handler

mikejolley:
Fixes the click handler in the atomic `button` component to make add to cart work again.
 
1. Add all to products block
2. Add some simple products to the cart
3. Spinner should show and cart should update

### DIFF
--- a/assets/js/atomic/blocks/product/button/block.js
+++ b/assets/js/atomic/blocks/product/button/block.js
@@ -101,7 +101,9 @@ const AddToCartButton = ( { product } ) => {
 		buttonProps.href = permalink;
 		buttonProps.rel = 'nofollow';
 	} else {
-		buttonProps.onClick = addToCart;
+		buttonProps.onClick = () => {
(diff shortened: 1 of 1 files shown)


assets/js/atomic/blocks/product/button/block.js
```diff
@@ -101,7 +101,9 @@ const AddToCartButton = ( { product } ) => {
 		buttonProps.href = permalink;
 		buttonProps.rel = 'nofollow';
 	} else {
-		buttonProps.onClick = addToCart;
+		buttonProps.onClick = () => {
+			addToCart();
+		};
```
senadir, line 106: Took me some time to understand why this was an issue when assigning `addToCart` directly to `onClick` we're also passing the args from `onClick` to `addToCart`, `addToCart` expects a quantity `int` as the first variables (defaults to 10), `onClick` passes an `event` object, causing this issue.
By calling `addToCart` on the callback, we avoid this.

PR #2775, opened 2020-06-24
Update dependency rules to enforce order and `WooCommerce dependencies`

### LINKED ISSUE #2052: Review usage of imports and be consistent for external vs internal dependency sequence.
In one of my recent pulls, @mikejolley noted that I had been defining internal dependencies before external dependencies. I didn't even realize I was doing it wrong and it should be external dependencies defined before internal dependencies.

This issue is to make sure a review is done to fix the inconsistency (and maybe a good opportunity to add a lint rule?)

mikejolley:
This PR probably needs some debate due to the `WooCommerce` rule, but giving you the option :) 

- Uses codeframe formatter to improve output of linting errors
- Adds a linting rule to ensure the order of dependencies is External, WooCommerce, Internal.
- Adds a linting rule to enforce `@woocommerce` dependencies under their own section

I've updated one file to check the rules pass.

GitHub Desktop 2020-06-24 11-22-52

Closes #2052

Note: WooCommerce admin uses this structure in it's dependencies too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
 		"lint:ci": "npm run lint:js && npm run lint:css",
 		"lint:css": "stylelint 'assets/**/*.scss'",
 		"lint:css-fix": "stylelint 'assets/**/*.scss' --fix",
-		"lint:js": "wp-scripts lint-js",
+		"lint:js": "wp-scripts lint-js -f codeframe",
 		"lint:js-fix": "eslint assets/js --ext=js,jsx --fix",
 		"lint:php": "composer run-script phpcs ./src",
 		"package-plugin": "./bin/build-plugin-zip.sh",
@@ -186,11 +186,11 @@
 	},
 	"lint-staged": {
 		"*.scss": [
-			"npm run lint:css"
+			"npm run lint:css -s"
 		],
 		"*.js": [
 			"prettier --write",
-			"npm run lint:js"
+			"npm run lint:js -s"
 		],
 		"*.php": [
 			"php -d display_errors=1 -l",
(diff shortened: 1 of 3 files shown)


nerrad:
I still think external/woocommerce/internal is too many unnecessary groupings. But I care more about consistency across projects. Personally, I don't think we should make this decision isolated from other teams and it'd be worth a P2 to settle the question.

package.json
```diff
@@ -186,11 +186,11 @@
 	},
 	"lint-staged": {
 		"*.scss": [
-			"npm run lint:css"
+			"npm run lint:css -s"
 		],
 		"*.js": [
 			"prettier --write",
-			"npm run lint:js"
+			"npm run lint:js -s"
```
nerrad, line 193: 🤔 I don't think I've seen the `-s` flag before (nor see it defined [in the cli args list](https://eslint.org/docs/2.13.1/user-guide/command-line-interface))

mikejolley: Odd it worked... silent mode for `npm` to avoid all the error text added after the lint.

nerrad: ahhh, so an npm flag (not eslint) 👍

nerrad:
I started an internal discussion on creating a `@woocommerce/eslint-plugin` package that will among other things import a variation of this custom rule (See p6q6po-aLM-p2).

We'll keep this pull open in the meantime as I will use some of the code here in the custom plugin (and we can repurpose the pull to pull from the package when it's published).

nerrad:
Just a heads up that I [created a pull for adding a `@woocommerce/eslint` package](https://github.com/woocommerce/woocommerce-admin/pull/4714) that will supersede much of the work in here.

mikejolley:
Unassigning myself. This can be closed when that other change is made I presume.

senadir:
We can close this now that https://github.com/woocommerce/woocommerce-admin/pull/4714 is merged.

PR #2778, opened 2020-06-25
Tweak some logic in our cart Integration test.

senadir:
As part of writing a guide on how to use React Testing Library, I found some issues in our test, some I managed to fix, and once I left, for now, I left comments on the PR to explain what I changed and the motives behind it.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/test/block.js
+++ b/assets/js/blocks/cart-checkout/cart/test/block.js
@@ -39,6 +39,9 @@ describe( 'Testing cart', () => {
 		expect(
 			await screen.getByText( /Proceed to Checkout/i )
 		).toBeInTheDocument();
+		/* We have an unfixed bug in our test in which the full cart triggers a POST
+		 * request to `wc/store/cart/update-item` causing the fetch to be called twice.
+		 */
 		expect( fetchMock ).toHaveBeenCalledTimes( 2 );
 	} );
 	it( 'renders empty cart if there are no items in the cart', async () => {
--- a/assets/js/blocks/cart-checkout/cart/test/block.js
+++ b/assets/js/blocks/cart-checkout/cart/test/block.js
@@ -13,6 +13,11 @@ import { defaultCartState } from '../../../../data/default-states';
 
 describe( 'Testing cart', () => {
 	beforeEach( async () => {
+		fetchMock.mockResponse( ( req ) => {
+			if ( req.url.match( /wc\/store\/cart/ ) ) {
+				return Promise.resolve( JSON.stringify( previewCart ) );
+			}
+		} );
 		// need to clear the store resolution state between tests.
 		await dispatch( storeKey ).invalidateResolutionForStore();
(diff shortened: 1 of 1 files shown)


assets/js/blocks/cart-checkout/cart/test/block.js
```diff
@@ -13,16 +13,19 @@ import { defaultCartState } from '../../../../data/default-states';
 
 describe( 'Testing cart', () => {
 	beforeEach( async () => {
-		fetchMock.mockResponseOnce( ( req ) => {
-			if ( req.url.match( /wc\/store\/cart/ ) ) {
-				return Promise.resolve( JSON.stringify( previewCart ) );
-			}
-		} );
```
senadir, line 20: this was moved from `beforeEach` to the `'renders cart if there are items in the cart'` test because it caused an issue with the empty cart test, as it triggered to calls, causing `toHaveBeenCalledTimes` to always have two (one for the inline in that test, and one for this).

assets/js/blocks/cart-checkout/cart/test/block.js
```diff
@@ -32,11 +35,14 @@ describe( 'Testing cart', () => {
...
-			).toBeInTheDocument();
-		} );
+		await waitFor( () => expect( fetchMock ).toHaveBeenCalled() );
+		expect(
+			await screen.getByText( /Proceed to Checkout/i )
+		).toBeInTheDocument();
+		/* We have an unfixed bug in our test in which the full cart triggers a POST
+		 * request to `wc/store/cart/update-item` causing the fetch to be called twice.
+		 */
+		expect( fetchMock ).toHaveBeenCalledTimes( 2 );
```
senadir, line 45: As noted in the comment, this is a known issue for now, the block is triggering a post request for `wc/store/cart/update-item` with this body `{key: 1, quantity: 1}`, item with key 1 has a quantity of 2, I'm not sure why is that value is being reduced, but if we change the previewData quantity to 1, this is fixed, if we change the second item quantity to more than 1, it will also trigger a request for it to be reduced, I choose not to change the quantity so that we don't mask this bug.

assets/js/blocks/cart-checkout/cart/test/block.js
```diff
@@ -13,7 +13,7 @@ import { defaultCartState } from '../../../../data/default-states';
 
 describe( 'Testing cart', () => {
 	beforeEach( async () => {
-		fetchMock.mockResponseOnce( ( req ) => {
+		fetchMock.mockResponse( ( req ) => {
```
senadir, line 16: we need this so we don't end up mocking each request (causing multiple requests in the empty test for example).

assets/js/blocks/cart-checkout/cart/test/block.js
nerrad: Ya this is odd. It sounds like something specific to the test environment but we should definitely investigate. Could you please make this comment a todo block so that a follow-up issue will automatically be created to handle this (or just create the issue manually if you want)?

PR #2779, opened 2020-06-25
Sort states and countries correctly using `deep_sort_with_accents`

### LINKED ISSUE #1902: States are not correctly sorted in Cart and Checkout blocks
Core sorts countries alphabetically, but doesn't do the same for country states, so in some locales the _State_ dropdown might be unordered.

imatge

See #1903 for an in-progress PR trying to fix this.

mikejolley:
Fixes state sorting by removing accents before performing the sort.

Alternative fix for https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1903

Closes #1903 
Fixes #1902

1. Switch to Catalan.
2. Dashboard > Updates > Download translation packs
3. Go to Block Checkout
4. Select Espanya
5. View state dropdown. Verify Illes Balears is correctly sorted after Huelva.

### Changelog

> Correctly sort states after translation on Block Checkout.

### DIFF
--- a/src/BlockTypes/Cart.php
+++ b/src/BlockTypes/Cart.php
@@ -78,15 +78,12 @@ class Cart extends AbstractBlock {
 			AssetDataRegistry::class
 		);
 
-		$block_data = [
-			'shippingCountries' => [ WC()->countries, 'get_shipping_countries' ],
-			'shippingStates'    => [ WC()->countries, 'get_shipping_country_states' ],
-		];
+		if ( ! $data_registry->exists( 'shippingCountries' ) ) {
+			$data_registry->add( 'shippingCountries', $this->deep_sort_with_accents( WC()->countries->get_shipping_countries() ) );
+		}
 
-		foreach ( $block_data as $key => $callback ) {
-			if ( ! $data_registry->exists( $key ) ) {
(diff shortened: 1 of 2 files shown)


src/BlockTypes/Cart.php
```diff
@@ -103,6 +100,27 @@ protected function enqueue_data( array $attributes = [] ) {
 		do_action( 'woocommerce_blocks_cart_enqueue_data' );
 	}
 
+	/**
+	 * Removes accents from an array of values, sorts by the values, then returns the original array values sorted.
+	 *
+	 * @param array $array Array of values to sort.
+	 * @return array Sorted array.
+	 */
+	protected function deep_sort_with_accents( $array ) {
```
mikejolley, line 109: This could become a utility but there is no obvious home for it so let's keep it across both files for this iteration and move if it becomes useful in other contexts.

haszari:
While writing [testing instructions for 2.9.0](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2816/files) I noticed that this issue affects the `[checkout]` shortcode as well, so perhaps we should log the issue with core and fix that too. Thoughts?

Screen Shot 2020-07-06 at 3 23 11 PM

PR #2780, opened 2020-06-25
Add custom jest e2e matcher

### LINKED ISSUE #2709: Introduce Jest matcher to facilitate E2E block validation.
In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2707 I introduced some reforms to how we write tests, this will result in some repetitive pattern, that consists of four expect calls

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/67f59ac46114a50efae143b912ce7bc7938878de/tests/e2e-tests/specs/backend/all-products.test.js#L35-L53

This could be abstracted to 

```
await expect( page ).toRenderBlock( block );
```

as it would allow for cleaner tests.

senadir:
This PR adds a custom matcher to jest for us to use in E2E tests `toRenderBlock` it checks if the page has any broken blocks in it, either due to not being registered, validation issues, or js error (caught by our block boundary), it also checks if the blocks exist, since the rest can all pass with no block.

The matcher is async so you will need to append `await` for it to work.

closes #2709 

### How to use

```
await expect( page ).toRenderBlock( {
	name: 'Single Product',
	class: '.wc-block-single-product'
} );
```
Or if you follow the new process of writing tests, it should be like this

```
await expect( page ).toRenderBlock( block );
```

### DIFF
--- a/tests/e2e-tests/specs/backend/all-products.test.js
+++ b/tests/e2e-tests/specs/backend/all-products.test.js
@@ -34,25 +34,6 @@ describe( `${ block.name } Block`, () => {
 	} );
 
 	it( 'renders without crashing', async () => {
-		// Gutenberg error
-		expect(
-			( await page.content() ).match(
-				/Your site doesn’t include support for/gi
-			)
-		).toBeNull();
-		// Our ErrorBoundary
-		expect(
-			( await page.content() ).match(
-				/There was an error whilst rendering/gi
-			)
-		).toBeNull();
-		// Validation Error
-		expect(
-			( await page.content() ).match(
-				/This block contains unexpected or invalid content/gi
-			)
-		).toBeNull();
-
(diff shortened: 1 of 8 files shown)


tests/e2e-tests/specs/backend/all-products.test.js
```diff
@@ -34,25 +34,6 @@ describe( `${ block.name } Block`, () => {
...
-		).toBeNull();
-		// Validation Error
-		expect(
-			( await page.content() ).match(
-				/This block contains unexpected or invalid content/gi
-			)
-		).toBeNull();
-
-		await expect( page ).toMatchElement( block.class );
+		await expect( page ).toRenderBlock( block );
```
nerrad, line 37: This is such a nice reduction of code 👍

PR #2787, opened 2020-06-25
Add storybook story for icon library

nerrad:
This pull adds a storybook story for the `@woocommerce/icons` library.

Image 2020-06-25 at 5 36 45 PM

This is pretty much a direct take from the equivalent Gutenberg implementation except I modified the styling a bit so more icons display on the screen. The story is setup so that any new icons added to the library will automatically display with the story.

In the process of doing this pull the story revealed the following:

- There was direct use of `wp` global for filters we were using instead of importing from the package (external)
- The dashIcon replacements we were injecting in lieu of the full dashicon library were being exported as react elements instead of nodes from the icon library which was inconsistent with other icons in the library. I switched that to export node, and then updated the dashicon module replacement to use `cloneElement` instead of `createElement`.
- The woo icon used for our block category was also being exported as a react element instead of node. So I fixed that too. It also needed updated to correctly receive `width`, `height` and other props (so cloning to pass those values through would work). I updated the implementation where the category is registered so that it's implemented using the `Icon` component pattern used for all icons in the library.
- The `arrowBack` icon from the library didn't have `viewBox` set, so that meant it wasn't reflecting changes in size properly.

## To Test

The story can be tested by running `npm run storybook` and viewing the story for the icons. When this pull is merged to master they should show up in the generated storybook as well.

Due to the changes in other files, there should be some smoke testing of the impacted code:

### wp global to package import

So the interesting thing here is that this code is related to the work don in [this pull](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1613). However, I discovered while testing, that although the grid blocks are initializing with the expected defaults, all attributes _are not_ getting saved with the blocks as expected. This is also happening on master too. So I'm not sure if there was another change that impacted this (maybe filters loading too late?) but if this behaviour is unexpected we'll likely need to investigate via a follow-up issue (cc @mikejolley).

### Dashicon replacements.

The affected icons are the chevron on our selects (in checkout):

Image 2020-06-25 at 5 11 26 PM

...and the "x" dismiss icon in notifications:

Image 2020-06-25 at 5 50 26 PM

* [ ] Verify that they icons appear and there are no errors in the console in both the frontend and backend.

### Woo Icon

Verify that the woo icon displays correctly for the WooCommerce Blocks category in the Block picker.

### `arrowBack` icon.

This icon is used with the "Return to Cart" item in the checkout block:

Image 2020-06-25 at 5 13 32 PM

* [ ] Verify that the `arrowBack` icon displays as expected.


assets/js/icons/stories/index.js
```diff
@@ -0,0 +1,89 @@
...
+				// eslint-disable-next-line no-restricted-syntax
+				id="filter-icons"
+				type="search"
+				value={ filter }
+				placeholder="Icon name"
+				onChange={ ( event ) => setFilter( event.target.value ) }
+			/>
+			<div
+				style={ {
+					display: 'flex',
```
Aljullu, line 40: What's the benefit of having inline styles instead of a `style.scss` file? The `QuantitySelector` component has a [`stories/style.scss` file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/base/components/quantity-selector/stories/style.scss), so it might be good to decide what's the best approach and be consistent with it.

assets/js/icons/stories/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+export default {
+	title: 'WooCommerce Blocks/@woocommerce/icons',
+	component: Icon,
+};
+
+const LibraryExample = () => {
+	const [ filter, setFilter ] = useState( '' );
+	const filteredIcons = omitBy( availableIcons, ( _icon, name ) => {
+		return name.indexOf( filter ) === -1;
```
Aljullu, line 23: You can use `includes` if you want:
```suggestion
		return ! name.includes( filter );
```

assets/js/icons/stories/index.js
nerrad: This is just a story and not a component we use so I'm just using inline styles for the purpose of the story :)

nerrad: >  The QuantitySelector component has a stories/style.scss file, so it might be good to decide what's the best approach and be consistent with it.

In that case the _css class_ is being modified for the selector on the component.  Personally I think we should avoid styles on actual components as much as possible in stories (favoring what the _default_ look for the component is). That will help us catch potential issues with the default not being usable.

nerrad:
> Should we open an issue to investigate that?

Yup, see #2789.

PR #2788, opened 2020-06-26
Don't load frontend scripts in the editor

### LINKED ISSUE #1910: frontend.js files are loaded in the editor
**To Reproduce**
1. Create a post with the _All Products_ block and filters blocks, for example, even though any JS-rendered block would work.
2. Open your browser devtools and switch to the network tab. There, filter by files that contain the string `frontend` in their name.
3. Reload the tab.
4. Notice the `-frontend.js` file of each block was loaded, even though that file is only used in the site frontend.

**Expected behavior**
All files named `frontend.js` are only used in the frontend, so we should avoid loading them in the editor.

Aljullu:
Fixes #1910 

This PR:
* Moves `register_block_script` and `get_block_asset_build_path` to the AssetApi class (134d212b5f376e0eca3565ad3e0818bedb790548).
* Updates `register_block_script` so it doesn't register block scripts in admin (f61fcc95da5299b29bab5d75752abbc78b70b367).
* Fixes the version comparison so 5.2.x point releases get the `legacy` files (e0dc45e8a37c52630cb045b8d4677f79ab64d271).

1. Create a post with some JS-based blocks (_All Products_ block and filters blocks, for example).
2. Open your browser devtools and switch to the network tab. There, filter by files that contain the string `frontend` in their name.
3. Reload the tab.
4. Verify no file ending with `-frontend.js` was loaded.
5. With an WP version between 5.2.0 and 5.3.0 (ie: 5.2.5) and verify legacy scripts are still being enqueued.

### Changelog

> Scripts only relevant to the frontend side of blocks are no longer loaded in the editor.

### DIFF
--- a/src/Assets/Api.php
+++ b/src/Assets/Api.php
@@ -146,7 +146,7 @@ class Api {
 	 */
 	public function get_block_asset_build_path( $filename, $type = 'js' ) {
 		global $wp_version;
-		$suffix = version_compare( $wp_version, '5.2', '>' )
+		$suffix = version_compare( $wp_version, '5.3', '>=' )
 			? ''
 			: '-legacy';
 		return "build/$filename$suffix.$type";
--- a/src/Assets/Api.php
+++ b/src/Assets/Api.php
@@ -40,7 +40,7 @@ class Api {
 	 * @return string The cache buster value to use for the given file.
 	 */
 	protected function get_file_version( $file ) {
-		if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
+		if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG && file_exists( $this->package->get_path() . $file ) ) {
 			return filemtime( $this->package->get_path( trim( $file, '/' ) ) );
 		}
 		return $this->package->get_version();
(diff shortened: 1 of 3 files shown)


src/Assets/Api.php
```diff
@@ -40,7 +40,7 @@ public function __construct( Package $package ) {
 	 * @return string The cache buster value to use for the given file.
 	 */
 	protected function get_file_version( $file ) {
-		if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ) {
+		if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG && file_exists( $this->package->get_path() . $file ) ) {
```
nerrad, line 43: good additional change here 👍 .  Although I wonder if we should still log instances when the file doesn't exist? Should there be at least a log trail of filemtime failing because the file doesn't exist for potential troubleshooting?

src/Library.php
```diff
@@ -69,7 +69,7 @@ public static function register_blocks() {
 		];
 		// Note: as a part of refactoring dynamic block registration, this will be moved
 		// to block level config.
-		if ( version_compare( $wp_version, '5.2', '>' ) ) {
+		if ( version_compare( $wp_version, '5.3', '>=' ) ) {
```
nerrad, line 72: I'm guessing you made the change here because it covers scenarios where the version might be `5.2.1` etc right? (in which case 💯 ).

src/Assets/Api.php
Aljullu: I'm not sure to understand you. You mean throwing an error if the file doesn't exist?

src/Library.php
Aljullu: Yes, exactly!

Aljullu:

> I tested Single Product and Checkout blocks since those are the ones I was concerned about. Single Product works okay, but it looks like there is a problem with the payment method script registration for the checkout block on the frontend (it's not loading). Will you look into that please?

Honestly, I'm not sure what was going on, but in 70e54dd9db050e27491d160f8e33e9d15faac897 I refactored `register_assets()` so it uses the `register_scripts()` method from the AssetApi class, and it seems to have fixed the issue.

src/Assets/Api.php
```diff
@@ -88,26 +88,31 @@ public function register_script( $handle, $relative_src, $dependencies = [], $ha
 			$version = $this->get_file_version( $relative_src );
 		}
 
-		wp_register_script( $handle, $src, $dependencies, $version, true );
+		wp_register_script( $handle, $src, apply_filters( 'woocommerce_blocks_register_script_dependencies', $dependencies, $handle ), $version, true );
```
Aljullu, line 91: I took this filter from `src/Assets.php`, we were inconsistently applying it in Assets but not in AssetsApi. Now that we have a unique way to register scripts, I think it makes sense to keep the filter.

nerrad: ahh that would explain it. The Payments API uses that to register the handle for payment method scripts as dependencies on certain handles (so various registration happens before the blocks load).

src/Assets/Api.php
nerrad: No I mean `error_log` if it doesn't exist. Previously there'd be a php warning if `filemtime` was used on a file that doesn't exist and that'd be a signal that something went wrong with the build. So my comment was birthed out of wondering if we're okay with losing that signal that something isn't right with the build process.

I'm on the fence here, and I'd be okay with leaving this code as is.

Issue #2718, opened 2020-06-15
Document how to write E2E tests using the new par


senadir:
With https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2707 merged, I changed the way we approach E2E tests for blocks, this needs documentation so other team members can approach this easily.

nerrad:
Looks like #2800 got the ball rolling on this issue so I think we can close it. Improving and iterating on docs is an ongoing thing.

PR #2800, opened 2020-07-01
add basic docs for running e2e tests

haszari:
Related #2718

Super basic docs for how to run our e2e tests, and clarifying the two kinds of "JavaScript tests" that we have.

- Read through the doc changes.
- Is it helpful, does it make sense?
- If you were a totally overwhelmed newbie in the team, would this help you run the e2e tests?

### DIFF
--- a/docs/contributors/javascript-testing.md
+++ b/docs/contributors/javascript-testing.md
@@ -51,3 +51,4 @@ When you're done, you may want to shut down the test environment:
 
 - `npm run wp-env stop` to stop the test environment
 
+**Note:** There are a number of other `wp-env` commands that would be useful to know. You can see them [here](https://github.com/WordPress/gutenberg/blob/master/packages/env/README.md).
--- a/docs/contributors/javascript-testing.md
+++ b/docs/contributors/javascript-testing.md
@@ -51,4 +51,4 @@ When you're done, you may want to shut down the test environment:
 
 - `npm run wp-env stop` to stop the test environment
 
-**Note:** There are a number of other useful `wp-env` commands. You find out more [in wp-env docs](https://github.com/WordPress/gutenberg/blob/master/packages/env/README.md).
+**Note:** There are a number of other useful `wp-env` commands. You can find out more in the [wp-env docs](https://github.com/WordPress/gutenberg/blob/master/packages/env/README.md).
--- a/docs/contributors/javascript-testing.md
+++ b/docs/contributors/javascript-testing.md
@@ -4,9 +4,18 @@ Tests for JavaScript in the Blocks plugin are powered by [Jest](https://jestjs.i
 
 The Blocks plugin follows the same patterns as Gutenberg, therefore for instructions on writing tests you can [refer to this page in the Gutenberg Handbook](https://developer.wordpress.org/block-editor/contributors/develop/testing-overview/).
 
-## Running Tests
+We have two kinds of JavaScript tests:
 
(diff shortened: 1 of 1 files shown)


docs/contributors/javascript-testing.md
```diff
@@ -4,9 +4,18 @@ Tests for JavaScript in the Blocks plugin are powered by [Jest](https://jestjs.i
...
+- JavaScript unit tests - test APIs, hooks, library functionality that we use to build blocks or expose to plugin authors.
+- End-to-end (e2e) tests - test blocks from the user interface. 
+
+These tests are all run automatically on open PRs by Travis CI.
+
+## How to run JavaScript unit tests
+
+Unit tests are implemented near the code they test, in `*.test.js` files.
+
+Assuming you've already followed the [Getting Started Guide](getting-started.md) on setting up node and other dependencies, unit tests are ran from the command line using the following command:
```
Aljullu, line 18: I noticed `composer install` is required for e2e tests too, so I think it would be good to move this line outside the unit tests section. Probably above so it's clear dependencies are needed for both, unit and e2e tests?

docs/contributors/javascript-testing.md
```diff
@@ -18,3 +27,25 @@ Additionally,
...
+To set up to run e2e tests:
+
+- `npm run build:e2e-test` to build container images used for e2e
+- `npm run wp-env start` to start the test environment
+
+Then, to run the tests:
+
+- `npm run test:e2e` 
+
+When you're iterating on a new test you'll run this repeatedly until your test is just right.
```
Aljullu, line 46: Do we know why this is required? I wonder if it might be valuable to add an explanation.

nerrad: I'm assuming this is just a nod to the fact that while tests are in development they might need to be re-run to pick up changes as the test is being developed.

Also, a couple notes (that don't affect this pull, but to keep in mind in future iterations of our test suite):

- ideally we should switch to use `wp-scripts test:e2e` for our tests.
- doing the above will get us access to a number of e2e helpers including: watching changed files and automatically running e2e tests on them, a special debug mode for debugging tests.

docs/contributors/javascript-testing.md
```diff
@@ -18,3 +27,25 @@ Additionally,
...
+
+When you're done, you may want to shut down the test environment:
+
+- `npm run wp-env stop` to stop the test environment
```
nerrad, line 50: Might be good to link to the [`wp-env` docs] for all the other commands that can be executed?
```suggestion
- `npm run wp-env stop` to stop the test environment

**Note:** There are a number of other `wp-env` commands that would be useful to know. You can see them [here](https://github.com/WordPress/gutenberg/blob/master/packages/env/README.md).
```

docs/contributors/javascript-testing.md
nerrad: Agreed.

docs/contributors/javascript-testing.md
haszari: Clarified this a bit in 6be54268f8404a9c1b610f73d198532fe6824f3b - yes this is just referring to code => test iterate loop :)

docs/contributors/javascript-testing.md
haszari: Good point, shifted this and tweaked in 6be54268f8404a9c1b610f73d198532fe6824f3b

haszari:
Thanks for the feedback - merging :)

Feel free to tweak this further as we go.

> ideally we should switch to use wp-scripts test:e2e for our tests

I agree, should we open an issue to track this @nerrad ?

Issue #2395, opened 2020-05-04
Cover basic block creation for all of our blocks in E2E.


senadir:
As we decided to do a biweekly release, we will be shipping more code more frequently, previously we tested all of legacy and old block catalog manually, across WordPress versions, this is a tedious and lengthy process, we should aim to extend our current basic E2E tests (create and save a block) to all of our blocks (old and new).

#### Blocks that already have tests
- [x] woocommerce/all-products - test already in main
- [x] woocommerce/single-product - test already in main, block is still in development
- [x] woocommerce/checkout - test already in main
- [x] woocommerce/cart - test already in main
- [x] woocommerce/product-search - test already in main

🙌 @senadir for getting these up and running and working reliably in Travis against multiple versions of WordPress!

#### Reviews blocks
- [x] woocommerce/all-reviews - ~~WIP on #2756~~ now in #2799 @haszari 
- [x] woocommerce/reviews-by-product @mikejolley 
- [x] woocommerce/reviews-by-category @mikejolley 

#### `All Products` filter blocks
- [ ] woocommerce/price-filter - WIP on #2756 
- [ ] woocommerce/attribute-filter - WIP on #2756 
- [ ] woocommerce/active-filters - WIP on #2756 

#### Product grid blocks (legacy/server-side-render) https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4566
- [ ] woocommerce/featured-category
- [ ] woocommerce/featured-product
- [ ] woocommerce/handpicked-products
- [ ] woocommerce/product-best-sellers
- [ ] woocommerce/product-categories
- [x] woocommerce/product-category
- [ ] woocommerce/product-new
- [x] woocommerce/product-on-sale
- [x] woocommerce/products-by-attribute
- [ ] woocommerce/product-top-rated

haszari:
I see there is a bunch of progress toward this in #2756 🎉

I'm going to grab this issue and that PR. Here's my plan:

- Find out where we're at - what block create tests are already running successfully on master.
- Map out the remaining blocks we need to test (maybe categorised into related chunks) - will add a checklist to this issue.
- Add new tests in new PRs (will borrow heavily from #2756) and keep this issue updated with progress.

senadir:
This was never fully done, but I still hope we can revisit this in the future.

Aljullu:
[...]

PR #2808, opened 2020-07-02
Reviews by product/category e2e tests

mikejolley:
Part of #2395 

Adds e2e tests for the Reviews by Category and Reviews by Product blocks.

Includes extra fixture data so we have 2 products, and those products have a product category. Since the API requires a category ID, this is converted during product creation. There may be a better way to do this, but since the API requires and ID and doesn't support names it's difficult to work around.

1. `npm run wp-env clean`
2. `npm run test:e2e`


tests/e2e-tests/fixtures/fixture-loaders.js
```diff
@@ -102,24 +102,66 @@ const deleteCoupons = ( ids ) =>
...
+
+	return WooCommerce.post( 'products/categories/batch', {
+		delete: ids,
+	} );
+};
+
 /**
  * Create Products.
  *
  * currently this only creates a single product for the sake of reviews.
```
haszari, line 138: I think we can delete this stale comment (you've improved how this works) :)

tests/e2e-tests/specs/backend/reviews-by-category.test.js
```diff
@@ -0,0 +1,48 @@
...
+		// Selected.
+		await page.waitForSelector(
+			'.wc-block-review-list .wc-block-review-list-item__item:not(.is-loading)'
+		);
+		expect(
+			await page.$$eval(
+				'.wc-block-review-list .wc-block-review-list-item__item',
+				( reviews ) => reviews.length
+			)
+		).toBeGreaterThanOrEqual( 6 ); // Fixture data has three reviews per product, and there are multiple products.
```
haszari, line 46: Right now the fixture data has 6 total reviews all within `Music` category, so this might not catch if the category filtering broke. Not a big deal, could switch one of the reviews to a non-`Music` product to make the test more robust :)

tests/e2e-tests/specs/backend/reviews-by-category.test.js
```diff
@@ -0,0 +1,48 @@
...
+		await expect( page ).toRenderBlock( block );
+	} );
+
+	it( 'shows category selector', async () => {
+		await expect( page ).toMatchElement(
+			`${ block.class } .woocommerce-search-list`
+		);
+	} );
+
+	it( 'can select a category and show reviews', async () => {
```
haszari, line 29: I couldn't get this test to pass (tried running twice):

```
 FAIL  tests/e2e-tests/specs/backend/reviews-by-product.test.js (35.76s)
  Reviews by Product Block
    ✓ renders without crashing (190ms)
    ✓ shows product selector (6ms)
    ✕ can select a product and show reviews (30019ms)

  ● Reviews by Category Block › can select a category and show reviews

    : Timeout - Async callback was not invoked within the 30000ms timeout specified by jest.setTimeout.Timeout - Async callback was not invoked within the 30000ms timeout specified by jest.setTimeout.Error:
```

Ran in dev mode (`npm run test:e2e-dev`) and I think I saw the problem - no reviews on the selected product:

Screen Shot 2020-07-03 at 10 42 44 AM

haszari: D'oh! I needed to `wp-env clean` to get the new fixture data. We can probably make this clearer in the docs. 

Over time it'll become clearer if the fixture data is slower-moving than the tests themselves – if they are closely coupled maybe we can change the move the data setup to the start of the test command, so `npm run test:e2e` re-applies the fixture data and runs tests.

haszari: So this is good to go :)

PR #2814, opened 2020-07-03
Document Cart & Checkout button styles

### LINKED ISSUE #2761: Cart & Checkout: style button depending on theme colors
According to the last designs, the _Proceed to Checkout_ and _Place Order_ button styles should depend on theme colors:

imatge

In #2494 we removed the `button` class to avoid style conflicts in the editor, so we don't have an easy way to inherit theme styles (at least, until global styles are an option).

My proposal would be to leave it as-is in WC Blocks and instead style it directly in Storefront. Something like:

```SCSS
.wc-block-components-button {
  background-color: $button_alt_background_color;
  color: $button_alt_text_color;
}
```

This way, we ensure old `.button` styles from themes are not applied to the button, so we minimize the possibility of style conflicts.

The biggest con of this approach is that buttons will show 'default' colors for all themes not Storefront-based. I think it's fair enough to expect themes to update to support this new component. We could document it.

Tagging @mikejolley and @haszari for feedback. If this approach makes sense, I would add the new styles to Storefront and write some docs for WC Blocks repo.

Aljullu:
Fixes #2761.

This PR adds docs for themes to style the `Button` component and the Cart block submit container element.

It also has a small tweak in `.wc-block-cart__submit-container`: it makes it so the box shadow is added by a pseudo element. This allows using the `currentColor` value for the shadow color. This way, themes can change the box shadow color without the need to rewrite the entire `box-shadow` property.

### Screenshots

_(there shouldn't be any visual change)_

1. Do some smoke testing in the Cart & Checkout blocks and verify there are no regressions.

### DIFF
--- a/docs/theming/cart-and-checkout.md
+++ b/docs/theming/cart-and-checkout.md
@@ -47,7 +47,7 @@ alternatively, themes can override the `box-shadow` property completely:
 
 ```CSS
 .wc-block-cart__submit-container::before {
-	box-shadow: 0 -10px 20px 10px rgba(214,209,203,0.5);
+	box-shadow: 0 -10px 20px 10px rgba(214, 209, 203, 0.5);
 }
 ```
 
--- a/docs/theming/cart-and-checkout.md
+++ b/docs/theming/cart-and-checkout.md
@@ -10,7 +10,7 @@ By default, it uses a combination of black and white borders and shadows so it h
 
 ```CSS
 .wc-block-components-order-summary-item__quantity {
-	background: #f9f4ee;
+	background-color: #f9f4ee;
 	border-color: #4b3918;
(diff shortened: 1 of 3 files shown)


docs/theming/cart-and-checkout.md
```diff
@@ -1,5 +1,58 @@
...
+
+Take into consideration the container has a top box shadow that might not play well with some dark background colors. If needed, it can be modified directly setting the `color` property (internally, shadow color uses `currentColor`, so it honors the `color` property):
+
+```CSS
+.wc-block-cart__submit-container::before {
+	color: rgba(214, 209, 203, 0.5);
+}
+```
+
+alternatively, themes can override the `box-shadow` property completely:
```
haszari, line 46: `Alternatively`

PR #2818, opened 2020-07-06
Make stylesheets order consistent between build and dev

### LINKED ISSUE #2713: Investigate why stylesheets are bundled in different order depending on the build mode
See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2704#discussion_r439610736 for an example.

Some times we face style bugs that are only reproducible on `npm run build`, but not on `npm run start`. That's because stylesheets are bundled in a different order, so selectors competing with the same specificity might be applied or not depending on the order they were added to the bundle.

This issue is about investigating why these differences happen and what we could do to avoid them.

Aljullu:
Fixes #2713.

First, in the main branch, let's write some styles to make the issue visible:
* In [`assets/js/base/components/payment-methods/style.scss`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/base/components/payment-methods/style.scss), add these styles:
```CSS
body {
	border: 4px solid red;
}
```
* In [`assets/js/blocks/cart-checkout/cart/style.scss`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/blocks/cart-checkout/cart/style.scss), add these styles:
```CSS
body {
	border: 4px solid blue;
}
```
1. Run `npm run start` and see the body has a blue border.
2. Run `npm run build` and notice the body border is now red. Styles are different between start and build! :scream: 

Now, checkout this branch (`fix/2713-stylesheets-build-order`):
1. Run `npm run start` and see the body border is still blue.
2. Run `npm run build` and notice the body border is now blue. Styles are the same between start and build! :tada: 

In addition to that, it would be good to test that there are no regressions when running `npm run package plugin` and `npm run storybook`.

#### How the CSS is currently generated:

Even though we have several [webpack configs](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-configs.js#L537-L540) (`core`, `main`, `front` and `payments`), all CSS files are generated from the `main` config.

In most cases, that worked fine: `main` has all blocks' `index.js` as the entry point, those files were in charge of importing `style.scss`, `editor.scss` and the components in use, so all required styles were caught by webpack and later bundled in `style.css`, `editor.css`.

However, there were some cases where this config failed: for example, legacy builds didn't have the All Products block, so they were missing the product list styles. To fix that, the entry point [was added manually](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-entries.js#L26-L31) (see #2000).

Something similar happened with the BlockErrorBoundary styles: that component was only used in frontend scripts, so was not caught by `main` entries. Because of that, it had to be [added manually too](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-entries.js#L58-L59). Some time later, the BlockErrorBoundary has been imported from scripts that are loaded in the `main` config, so it's no longer necessary to explicitly declare that SCSS file in the entries list. However, we didn't notice it so it's still there.

The current approach, has another issue, the one that is causing #2713. Since we started using tree shaking, styles order stopped being something reliable and it depended on the order components and blocks were imported. That's why `npm run start` and `npm run build` generate the CSS with a different order. We were basically facing this issue: https://github.com/webpack/webpack/issues/7094.

#### How the CSS is generated with this pull:

In order to fix the issues above and (hopefully) simplify the way our CSS is generated, I split the CSS generation config from `main` into its own config (`styling`). It declares the CSS files as entries, this guarantees styles are always included in the same order.

#### Pros of this approach:
* CSS order is consistent between dev and prod.
* All CSS files under `assets` are included, so there is no need to manually add `scss` files that are not imported from `main` config.
* Some editor styles that leaked into `style.css` with the previous approach, are no longer there, producing a 6% decrease in file size for `style.css`.

#### Cons of this approach:

* With this approach, there is no way to know which component styles should be bundled or not. If a component is only used in an experimental block, its styles will still be bundled in the final `style.css`. Same for legacy builds: they have the same component styles as regular builds. You will notice that means `legacy` stylesheets size increased a lot.

I know this is not a perfect solution. It solves some issues while introducing some cons. I would like to get some feedback whether the team feels this is worth merging or somebody has another idea. :slightly_smiling_face:


assets/js/base/components/panel/style.scss
```diff
@@ -7,24 +7,25 @@
...
-		height: auto;
-		line-height: 1;
-		margin-bottom: em(6px);
-		margin-top: em(6px);
-		padding-bottom: em($gap-small - 6px);
-		padding-right: #{24px + $gap-smaller};
-		padding-top: em($gap-small - 6px);
-		position: relative;
-		text-align: left;
-		width: 100%;
```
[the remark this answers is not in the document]
Aljullu: Removed. :+1:

assets/js/base/components/quantity-selector/stories/index.js
```diff
@@ -19,11 +18,13 @@ export const Default = () => {
 	const [ quantity, changeQuantity ] = useState();
 
 	return (
-		<QuantitySelector
-			disabled={ boolean( 'Disabled', false ) }
-			quantity={ quantity }
-			onChange={ changeQuantity }
-			itemName="widgets"
-		/>
+		<div style={ { width: 100 } }>
```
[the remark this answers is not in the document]
Aljullu: This is moving the styles from [`/quantity-selector/stories/style.scss`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2818/files#diff-32004167aea03f11226ddf9e1c86df70) so they are inline. I thought it was discussed that we wanted to avoid CSS files for stories, but I might be wrong.

bin/webpack-configs.js
```diff
@@ -29,8 +29,27 @@ const dashIconReplacementModule = path.resolve(
...
+		customSummary: ( time ) => {
+			console.log(
+				chalk.green.bold(
+					`${ progressBarPrefix }${ name } assets build completed (${ time })`
+				)
+			);
+		},
+	};
+};
+
```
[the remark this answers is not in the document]
Aljullu: Done! See: #2875.

assets/js/base/components/quantity-selector/stories/index.js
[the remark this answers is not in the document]
nerrad: Oops ya sorry, that's right, I missed that this was in a story file.

bin/webpack-entries.js
```diff
@@ -2,19 +2,139 @@
...
+	checkout: {
+		dir: 'cart-checkout/checkout',
+		frontend: 'cart-checkout/checkout/frontend.js',
+	},
+	'single-product': {
+		dir: 'single-product',
+		frontend: 'single-product/frontend.js',
+		isExperimental: true,
+	},
+};
```
nerrad, line 82: Is there someway we could automate building this map?

bin/webpack-entries.js
```diff
@@ -2,19 +2,139 @@
...
-		// Shared blocks code
-		blocks: './assets/js/index.js',
+	'product-best-sellers': {
+		dir: 'product-best-sellers',
+	},
+	'product-category': {
+		dir: 'product-category',
+	},
+	'product-categories': {
+		dir: 'product-categories',
```
nerrad, line 18: Is there some way we could automate this map?

senadir:
I'm mainly concerned about the lack of tree-shaking and feature gating, does this PR mean we will bundle all of our styles in the same file? why not individual files? and the issue with feature gating is that we can introduce large or breaking CSS changes that we don't want to land right now until we fix or announce, I think we can implement some sort of feature gating for styles now? like you did for blocks?

bin/webpack-entries.js
Aljullu: I did a small refactor in b4d96edcb5e193b790fa7f4620d5328beb483303 so we are less explicit about directories and instead they are inferred from the block name.

Aljullu:
Thanks for the reviews @mikejolley and @senadir!

> I do wonder about automation though. Can webpack scan directories? If it were to scan the blocks directory it could derive the name of the majority of the blocks, the exceptions being cart-checkout and reviews which have subdirs.

That might be worth experimenting but I would leave it out from this PR (I don't want to keep adding things to this PR so it keeps growing). For `cart-checkout` and `reviews` we could just simply use an `index.js` that includes the `index.js` from the subdirectories. I guess it would work.

> why not individual files?

Notice that hasn't changed in this PR. We were bundling all blocks styles in the same CSS file and with this PR this doesn't change. I think both approaches have their pros and cons, but having everything in the same file makes it easy to share component styles accross blocks. Many components are used in several blocks, so if we wanted to create a specific CSS file for each block, we should either create a `components.css` file that all blocks could import, or include the component styles with each block CSS file, which would increase assets size if there are many blocks in the same page and will affect caching negatively.

> and the issue with feature gating is that we can introduce large or breaking CSS changes that we don't want to land right now until we fix or announce, I think we can implement some sort of feature gating for styles now? like you did for blocks?

Yes, that's the main con I found with this approach. We no longer have a system to know which components are used by experimental/stable/legacy blocks. That's why legacy blocks CSS size increased a lot. I can't find any way to change this without being affected again by https://github.com/webpack/webpack/issues/7094.

senadir:
> Yes, that's the main con I found with this approach. We no longer have a system to know which components are used by experimental/stable/legacy blocks. That's why legacy blocks CSS size increased a lot. I can't find any way to change this without being affected again by webpack/webpack#7094.

Why not have two objects? stable styles and experimental styles, and merge them if we're on an experimental build?

Aljullu:
> Why not have two objects? stable styles and experimental styles, and merge them if we're on an experimental build?

That's more or less how it works for blocks. It's a single object, but we filter out blocks with `isExperimental` if we are building the styles for non-experimental environments:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2818/files#diff-eaffccd6b2ce22b246ec8decc3e9c671R14

The issue is with components, we don't have an easy way to know which components are used by experimental blocks and which are used by stable blocks. We would need to manually flag each component whether it's experimental or not. That would solve the problem but I see that adding some burden to us every time we create a new component. I don't have a strong opinion, so whatever the team feels about this. :slightly_smiling_face:

bin/webpack-configs.js
```diff
@@ -439,6 +348,123 @@ const getPaymentsConfig = ( options = {} ) => {
...
+		},
+		optimization: {
+			splitChunks: {
+				minSize: 0,
+				cacheGroups: {
+					editor: {
+						// Capture all `editor` stylesheets and the components stylesheets.
+						test: ( module = {} ) =>
+							module.constructor.name === 'CssModule' &&
+							( findModuleMatch( module, /editor\.scss$/ ) ||
```
haszari, line 413: This is pretty cool, automatically building `editor.css` from `editor.scss` sources. Should we document this somewhere? Not sure if this is new, but either way it's a convention that we must follow when implementing new blocks. (if I understand correctly)

bin/webpack-entries.js
```diff
@@ -2,19 +2,84 @@
...
-		wcSharedContext: './assets/js/shared/context/index.js',
+// List of blocks that should be used as webpack entry points. They are expected
+// to be in `/assets/js/blocks/[BLOCK_NAME]`. If they are not, their relative
+// path should be defined in the `customDir` property. The scripts below will
+// take care of looking for `index.js`, `frontend.js` and `*.scss` files in each
+// block directory.
+// If a block is experimental, it should be marked with the `isExperimental`
+// property.
+const blocks = {
+	'handpicked-products': {},
```
haszari, line 15: I ❤️ how clear and explicit this is, and the doc comment :)

It would be great to document maybe in markdown how the build system works and what the expectations/conventions are – there's a lot of complex decisions and tradeoffs here, might be worth writing a little doc.

bin/webpack-configs.js
Aljullu: It's not really new but I think it was not documented anywhere. Added some docs in #2953.

Issue #2442, opened 2020-05-08
clarify wording/APIs relating to payment method (used by shoppers) and payment gateway (implemented by extensions, installed by merchants)


haszari:
Follow up issue from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2425#issuecomment-625402179

In our code and the new payment-related APIs, we may be using `payment method` to refer to different things. For example:

1. An instance of a payment method, used by a shopper, to pay for an order - e.g. `stripe credit card ending with 1234` or `paypal account with email xyz@123`. These can be saved to user account - i.e. `pay with my saved card` feature.
2. A service, gateway or mechanism that merchant installs/configures so shoppers can pay - e.g. `Stripe Credit Card` `Paypal` `Cheque` `Cash on Delivery`. Aka a payment provider.

Note - payment extensions often bundle multiple gateways/mechanisms.

The goal of this issue is to ensure we are aligned and clear on these terms and any related wording. This helps to ensure the APIs, and potentially the UI too, are consistent and easy for payment providers to implement, and for merchants to configure. 

Note this is not specifically a blocks issue – these terms are in Woo core too. However, since we are implementing new payment related APIs for Cart & Checkout, it's a good time for us to review. Then we can fix any low-hanging fruit, or note any inconsistencies for follow up later.

mikejolley:
Any customer facing wording should definitely avoid the term `gateway` and instead use `method`.

haszari:
Related consideration - some gateways/extensions support multiple different payment methods. For example, `Stripe` extension includes `Stripe - Credit Card` and `Stripe - Payment Request`.

In the API we should aim to use consistent naming so this distinction is clear. In the new JS API it looks like we're using `name` for the extension, e.g. `stripe` and `id` for a payment method within an extension, e.g. `cc` or `payment-request`. I'm not sure what terms are used for these concepts in the Woo core gateway / extensibility API.

nerrad:
The motivation behind this issue is sound. However I think the actionable item here is to watch for inconsistencies in code reviews and fix as we come across them.

PR #948, opened 2019-09-06
Add version string replacement script for deploy

nerrad:
Closes: #901 

This pull adds a couple features for our deploy process:

- Removes the need to manually change the version string in `readme.txt`, `package.json`, `woocommerce-gutenberg-products-block.php` and `src/Package.php`.  It will automatically update the version strings using whatever is provided via the `github-deploy.sh` script process.
- Will replace any instances of `$VID:$` in any `*.php` files within the `src` directory.  This comes in really handy when adding `@since $VID:$` comment blocks in the code.  However, also handy any place else you want to have updated with the correct version for a release.  **Note:** $VID:$ will not be updated on a pre-release.

## Rationale

- Automating as much as possible helps eliminate human error and bloopers.  There's a number of files that need updated with version changes, this removes the need to remember to do that.
- I'm doing some refactoring of php files which will result in a number of `@since` updates, so this makes it easier to ensure the correct version is updated for those whenever that work is released.
- Related to the above, no more having to think about what version to use for `@since` :)

## To test

You can test the version script in isolation by doing the following:

- in some php file add a `$VID:$` string anywhere.  Then in the root folder of the plugin in a bash shell, execute this:

```bash
VERSION=5.6 ./bin/version-changes.sh
```

The above should change:

- wherever there is `$VID:$` to `5.6`
- update `package.json` to have the given value for it's version item (in this case `5.6`)
- update `readme.txt` to have the given value for it's `Stable Tag:` value
- update `woocommerce-gutenberg-blocks.php` file to have the given value for the version.
- update `src/Package.php` version constant to be the given value.

Nothing else should be changed.  You can discard the changes once you've verified it work.

We'll also verify this on the next run of github deploy (I'm going to test this in a fork before I merge).

**Update:** I tested running `github-deploy.sh` on a fork and other than having to fix the path to the `version-script.sh`, it worked.  So after review this is ready to go,.

PR #2831, opened 2020-07-08
Support cash on delivery payment method in checkout block

### LINKED ISSUE #2623: No cash on delivery in checkout
When selecting Cash On Delivery as the payment method. It doesn’t show as a payment method in the checkout page.

If it is the only payment method selected you get a warning saying no payment methods are available.

To test it wasn’t a plugin conflict all plugins were de activated expect woocommerce and woocommerce blocks

haszari:
~~**Note - this PR branch is based off #2840.** So you'll see all the diffs from #2840 here too. #2840 implements API changes that this PR needs - allow payment methods to show/hide dynamically (based on shipping method in this case) - #2838.~~ Now #2838 is merged, and this has been rebased to `main` 🧶 

Fixes #2623

Adds support for the [core `Cash on Delivery` payment method (gateway)](https://docs.woocommerce.com/document/cash-on-delivery/) in the cart block.

### Screenshots

Currently the tab just appears/disappears. We might want to add a spinner or other visual finesse for this - cc @LevinMedia 

Screen Capture on 2020-07-10 at 14-32-42

We can handle that as follow up, as part of improving UI for many (> 3) payment gateways.

1. Go to `WooCommerce > Settings > Payments` and configure `Cash on Delivery` payment gateway.
2. Set up cart & checkout pages with blocks.
3. Add stuff to cart, proceed to checkout.
4. Complete purchase with COD and confirm everything's working as expected.
2. COD has options so merchant can limit it to shippable products and / or specific shipping methods. Experiment with these settings and confirm that COD works correctly, and is only available when appropriate.
5. Confirm that all other payment gateways still work correctly.

Bonus points - set up more complicated shipping (multiple packages, different carriers) & orders and confirm that the correct payment methods are available dependent on selected shipping carrier.

### Changelog

> Support `Cash on Delivery` core payment gateway in the Checkout block. #2831


assets/js/payment-method-extensions/payment-methods/cheque/index.js
```diff
@@ -12,7 +12,7 @@ import { decodeEntities } from '@wordpress/html-entities';
 import { PAYMENT_METHOD_NAME } from './constants';
 
 const settings = getSetting( 'cheque_data', {} );
-const defaultLabel = __( 'Check Payment', 'woo-gutenberg-products-block' );
+const defaultLabel = __( 'Check payment', 'woo-gutenberg-products-block' );
```
haszari, line 15: I _think_ we should be using sentence case for all our labels, including payment methods. So I've fixed this one - happy to change to title case if needed.

assets/js/payment-method-extensions/payment-methods/cod/index.js
```diff
@@ -0,0 +1,71 @@
...
+	return <PaymentMethodLabel text={ label } />;
+};
+
+/**
+ * Determine whether COD is available for this cart/order.
+ *
+ * @param boolean cartNeedsShipping True if the cart contains any physical/shippable products.
+ * @return boolean True if COD payment method should be displayed as a payment option.
+ */
+const canMakePayment = ( { cartNeedsShipping } ) => {
```
haszari, line 48: `Cash on delivery` can be selectively enabled for specific shipping methods only: https://docs.woocommerce.com/document/cash-on-delivery/

Screen Capture on 2020-07-08 at 16-48-11

I think we'll need to extend this interface and pass the selected shipping option (alongside cart info), and call it whenever user changes shipping. Will keep working on this tomorrow, wanted to mention this now in case we have a plan for how this would work.

cc @nerrad @Aljullu

assets/js/payment-method-extensions/payment-methods/cheque/index.js
Aljullu: Absolutely! Both Gutenberg and Calypso are converging to sentence case, so we should too.

assets/js/payment-method-extensions/payment-methods/cod/index.js
Aljullu: Good catch. I don't think anything has been discussed about how to implement that, so your proposed solution of passing the shipping option to the payment method sounds good to me. :+1:

nerrad: This is a good example of a scenario I was thinking of and [why various `cartData` is passed through the `canMakePayment` method](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/5e668c083f5c9492cfd0d4a5f9dd56a463f5b1d5/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js#L59-L92). So 👍 for taking this approach (and noticing that it applied to cash on delivery).

assets/js/payment-method-extensions/payment-methods/cod/index.js
```diff
@@ -0,0 +1,71 @@
...
+	// If store allows COD for virtual orders, good to go.
+	if ( settings.enableForVirtual ) {
+		return true;
+	}
```
nerrad, line 52: Doesn't this need to account for virtual orders in the cart as well?

nerrad:
Rua and I chatted about this in slack (p1594243362275900-slack-C8X6Q7XQU), particularly around the issue he identified with payment methods possibly needing other conditions to determine whether they are shown or not (as is the typical case with COD and Shipping that requires COD or not).

While this can be somewhat addressed by utilizing the `canMakePayment` interface on the paymentMethods API. One additional problem he identified is that `canMakePayment` is currently only designed to run once on initialization. This effectively means that if shipping selection changes by the user (or shipping options appear after user has entered their address) whether or not a payment method is available or not won't update.

To summarize my thoughts on this (related specifically to this pull):

1. This pul lis isolated to adding a COD payment method.
2. Rua has identified the typical (and I think valid) use-case existing now where COD being available to make payment is tied to a selected shipping method.
3. The current approach in this pull only works for the initial state of the chipping method in the checkout when payment methods are initialized, but will not change if the shipping method changes.
4. This then means, if address info has not been entered yet for a purchase, COD could potentially not appear if shipping info depends on the address. Alternatively if a shopper switches between a selected shipping method tied to the COD payment method visibility and back, the COD option will not update.

In light of the above, I think we need to pause on this pull and avoid merging it until we are certain that the impact to existing expected behaviour for COD is not affected. I do not want a scenario where we potentially release the Checkout block with COD functionally only partially implemented where merchants set certain settings and not have it work as expected (potentially without the merchant even knowing).  Yes, we've communicated that C&C blocks are a preview, but based on the number of folks I've seen asking for COD support, I'm very cautious about releasing an unfinished implementation.

I'd like to do the following as some next steps:

- Create an issue summarizing what Rua discovered.
- Work out the API we will expose to payment methods (it may mean having `canMakePayment` fire multiple times or it may mean we introduce a new interface on the payment methods api) and then create a pull adding that API (along with needed documentation). 
- Ensure the api changes does not break any currently payment methods supporting the checkout block (and we can test it against this pull to make sure it meets needs in a separate merged branch for testing purposes).
- Merge that pull to main.
- Bring the changes in to this branch and finish this pull off.

assets/js/payment-method-extensions/payment-methods/cod/index.js
haszari: Follow up issue: #2838

haszari:
Added `blocked` label; this is blocked on the related payment method API changes - #2838

assets/js/base/context/cart-checkout/payment-methods/reducer.js
```diff
@@ -107,10 +107,7 @@ const reducer = (
 		case SET_REGISTERED_PAYMENT_METHODS:
 			return {
 				...state,
-				paymentMethods: {
-					...state.paymentMethods,
-					...paymentMethods,
-				},
+				paymentMethods,
```
haszari, line 110: Now we're replacing/resetting the `paymentMethods` array whenever the available options (gateways) change.

Previously this action was appending - so this is a change in behaviour. I'm hoping this is fine, since the action name is `SET_` not `ADD_` - this implementation seems consistent. 

Without this change, COD will appear when you select a COD-compatible shipping option, but **won't disappear** when you select a shipping option that doesn't support it :)

@nerrad  - We could change this and track the "availability" of each method using a boolean flag in state. That's a bigger change but might be more appropriate.

haszari:
Removing `blocked` as #2840 is available now - feel free to re-add tag if #2840 needs changes or follow up.

docs/extensibility/payment-method-integration.md
```diff
@@ -63,7 +63,7 @@ Here's some more details on the configuration options:
...
+-   `canMakePayment` (required): A callback to determine whether the payment method should be available as an option for the shopper. The function will be passed an object containing data about the current order. Return a boolean value - true if payment method is available for use. If your gateway needs to perform async initialisation to determine availability, you can return a promise (resolving to boolean). This allows a payment method to be hidden based on the cart, e.g. if the cart has physical/shippable products (example: `Cash on delivery`). The function Keep in mind this function is only invoked once in the mounting process of the checkout block. Also can be used to disable a payment method based on environment, browser or geolocation restrictions (example: `Apple Pay`).
```
haszari, line 66: We could make this optional - default to `true`. Also, could also support payment methods just returning boolean directly; i.e. detect if it's a callback or promise and handle accordingly.

nerrad:
@haszari I changed the base branch for this pull to the `fix-2838...` branch so that the diff is clearer for reviewing. When 2838 gets merged into main, you can just change the base back to the main branch.

docs/extensibility/payment-method-integration.md
nerrad: > Also, could also support payment methods just returning boolean directly; i.e. detect if it's a callback or promise and handle accordingly.

This already happens:

> https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0421af44603090cebe9df79cee04c62bbf1959c1/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js#L91-L93

The value returned is wrapped in a Promise and resolved immediately. So it handles whether it's a promise or not already. The only thing it doesn't handle is if it's a straight boolean or not which we can add.

nerrad: > We could make this optional - default to true. 

I'm on the fence with this. I think it's useful to signal to payment methods that this property should be provided. I realize there are a few payment methods that will always be `true` here.

src/Payments/Integrations/CashOnDelivery.php
```diff
@@ -0,0 +1,103 @@
+<?php
+/**
+ * Cash on Delivery (COD) (core) gateway implementation.
+ *
+ * @package WooCommerce/Blocks
+ * @since 2.10.0
```
nerrad, line 6: You can use `@since $VID:$` (here and everywhere you've hardcoded the version in this pull) and it will automatically get replaced with the correct version at the time of release by our version bump script. That allows us to not worry about what the version may or may not be when this is actually released.

src/Payments/Integrations/CashOnDelivery.php
```diff
@@ -0,0 +1,103 @@
...
+		$this->settings = get_option( 'woocommerce_cod_settings', [] );
+	}
+
+	/**
+	 * Returns if this payment method should be active. If false, the scripts will not be enqueued.
+	 *
+	 * @return boolean
+	 */
+	public function is_active() {
+		return ! empty( $this->settings['enabled'] ) && 'yes' === $this->settings['enabled'];
```
nerrad, line 55: `filter_var( $this->settings['enabled'], FILTER_VALIDATE_BOOLEAN )`  (here and elsewhere) is a nice way to handle this kind of thing as it explicitly allows for various boolean type indicators ('yes', 'no', etc). It'd also be future proof if Woo ever changed the settings values to actual booleans (or 1/0).

assets/js/base/context/cart-checkout/payment-methods/reducer.js
haszari: Moved across to  #2840  :)

docs/extensibility/payment-method-integration.md
haszari: > The only thing it doesn't handle is if it's a straight boolean or not which we can add.

Sounds good - I guess we can do that as follow up? Not particularly important for this PR. If we allow plain boolean then there's no urgent need for a default (trivial methods just pass `true`).

src/Payments/Integrations/CashOnDelivery.php
haszari: Cool! What does the substitution - is this something new? Added these in 44f5eb9b4d591cecf3f41b1648a9123f16511f89

src/Payments/Integrations/CashOnDelivery.php
haszari: Good point - fixed in all the core payment methods :) e504ba352568eecaebd1705842cbce63aa05fcb3

assets/js/payment-method-extensions/payment-methods/cod/index.js
```diff
@@ -0,0 +1,85 @@
...
+
+	// Look for an not-supported shipping method in the user's selected
+	// shipping methods. If one is found, then COD is not allowed.
+	const selectedNotSupported = Object.values( selectedShippingMethods ).find(
+		( shippingMethodId ) => {
+			return ! settings.enableForShippingMethods.includes(
+				shippingMethodId
+			);
+		}
+	);
```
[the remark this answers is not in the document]
haszari: Switched here - I agree this is clearer and more appropriate. 914b978aa9004b58cc13a4263d76ae639d7a56c9

haszari:
Thanks for the feedback @nerrad - this is ready for review again, all the feedback's been addressed. Still waiting on #2838 which should be merged soon :)

src/Payments/Integrations/CashOnDelivery.php
Aljullu: It was added in #948. :slightly_smiling_face:

LevinMedia:
>Currently the tab just appears/disappears. We might want to add a spinner or other visual finesse for this - cc @LevinMedia

@haszari lets re-address this once we merge the changes that better support two or more gateways (the list with radio buttons) I think visual change will be less abrupt if they're in a list. If it's still too much, we can add something to smooth it out. Perhaps a spinner as you suggested.

assets/js/payment-method-extensions/payment-methods/cod/index.js
```diff
@@ -0,0 +1,85 @@
...
+	return ! selectedNotSupported;
+};
+
+/**
+ * Cash on Delivery (COD) payment method config object.
+ */
+const cashOnDeliveryPaymentMethod = {
+	name: PAYMENT_METHOD_NAME,
```
haszari, line 76: While working on this I realised that in core/PHP this is referred to as `id` not `name` (it's a "slug" id IMO, e.g. `stripe` `cod` `bacs`). If we can safely name this part of the API/implementations consistent with core I think we should - i.e. make the prop `id`.

Thoughts @nerrad ? This is a pretty minor detail, but could make things clearer for implementors/devs out there.

haszari:
@nerrad / @Aljullu This is ready for final review & test.

Are there issues open for any of the following potential follow up? Keen to link em in and add info discovered as part of testing this PR/issue.

- Icon for `Cash on Delivery` payment method (and `Bank Transfer`) - I don't think this is a blocker. Though I am curious whether we want to add icons for all core payment methods, or what our expectations/guidelines are. 
- UI improvements for more (> 3) payment methods - e.g. show as accordion/radio buttons rather than tabs. 

cc @LevinMedia as these are all in the design/flow realm :)

docs/extensibility/payment-method-integration.md
nerrad: > Sounds good - I guess we can do that as follow up?

We could, I'm not sure whether it needs to be high priority though (or even done this cycle). I see it more as a "nice to have" than "must have".

assets/js/payment-method-extensions/payment-methods/cod/index.js
nerrad: We do [have a `paymentMethodId` prop](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7f6233dae5667beed59b7123c3d52cb9759634a1/assets/js/blocks-registry/payment-methods/payment-method-config.js#L22) and it is [passed through to the server](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7f6233dae5667beed59b7123c3d52cb9759634a1/assets/js/base/context/cart-checkout/checkout/processor/index.js#L194).

We have a `name` vs `id` because the server processing of a payment method might have the same gateway but a different payment method in that gateway. For instance Stripe gateway classes are used to process both Stripe CC and Stripe payment_request. So `paymentMethodId` is used to refer to the payment gateway on server requests, and name is used internally client side to differentiate payment methods in the registration for constructing the dynamic react nodes displaying the payment methods (css `id`, css classnames etc).

You've already raised [an issue around naming consistency](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2442) and this is something that maybe could be visited there, but for now this is a bit tricky because there's already some inconsistency around how terms are interchangeably used within Woo Core and gateways as well. `name` and `paymentMethodId` were chosen as something we could standardize around in the client side registration. I'm open to revisiting, but if we do this it should be discussed and worked out separately. 

I will say that I disagree with this statement:

> This is a pretty minor detail, but could make things clearer for implementors/devs out there.

I'm all for making things clearer, but I don't think this is a minor detail - in this case changing `name` to `id` won't make anything clearer compared to what it currently is imo.

FWIW, some background on when `paymentMethodId` was introduced is in #2170

nerrad:
> Icon for Cash on Delivery payment method (and Bank Transfer) - I don't think this is a blocker. Though I am curious whether we want to add icons for all core payment methods, or what our expectations/guidelines are.

I think we should have icons for all payment methods assuming that's a pattern we're establishing for the selections. I'll defer to @LevinMedia on this. It can be done as a followup (I don't know if there are issues for it or not).

assets/js/payment-method-extensions/payment-methods/cod/index.js
haszari: That makes sense - I [added a comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2442#issuecomment-658450682) to #2442 about this, I can understand this is tricky to get right and be consistent with Woo core. 

No urgency with this, will no doubt come up as we integrate more gateways and get feedback on the API.

haszari:
> However, I do see a problem when a shipping method is selected for which the COD is not available and it switches to the first payment method. It looks like the tab content for the new active payment method tab is not being selected (I'm guessing active payment method isn't being updated correctly).

Good catch @nerrad - I agree, the tab should select properly and show the appropriate payment UI. Fixed in ef3d3d1a09a0cab04fe14fff6f7a165be03820ff.

Note the behaviour is that it "restores" the previously-selected gateway (rather than always defaulting to the first one). I don't quite understand why, but this seems like a nice convenience. 

Ready for review again :)

LevinMedia:
@haszari the updated designs for the checkout section remove the icons, so we don't have to worry about it  :) 

https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2628

Generally speaking though I don't think the icons add much. Perhaps with the exception of the credit card, the icons won't make it any easier to scan the page and understand what you're looking at - in fact, I think they have the opposite effect, by adding to the clutter.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -293,13 +293,18 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 		paymentData.hasSavedToken,
 	] );
 
-	// set initial active payment method if it's undefined.
+	// Set active (selected) payment method as needed.
 	useEffect( () => {
 		const paymentMethodKeys = Object.keys( paymentData.paymentMethods );
+		if ( ! paymentMethodsInitialized || ! paymentMethodKeys.length ) {
+			return;
```
nerrad, line 300: Good early abort here 👏

PR #2832, opened 2020-07-08
lots of release process doc tweaks

haszari:
Have added this stuff:

- add outcomes for all H4s (we may be able to remove some similar material from "body" content)
- mention that svn is needed
- clarify relationship to PR and PR checklist in a few places
  - we have 2 docs to follow now, which could be confusing
- clarify release metadata step - this is metadata, not version number
- clarify relationship to woo core
- add dev blog post as final happy 🎉 step

#### Other ideas for improving this doc

One other thing that we might want to do - move the cherry picking / branch stuff to an appendix, it's really big and unwieldy, and necessary to skip over 2/3rds of it each release.

This bit, these three H5 sections: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2a6a0c05be5608e0fa51a7c92705fe93df94475a/docs/releases/handling-releases.md#patch-releases-against-latest-main-branch

Overall this doc is getting quite big, especially alongside the checklist in PR. Maybe we can make a smaller numbered list version and move some of the detail out to appendix section. 

Also would be good to add info about the release schedule (every 2 weeks, release by end of wednesday), and the typical timeframe for doing a release. E.g. small releases should take no more than half day, bigger releases you might want to get started ASAP on the monday. (?)

### DIFF
--- a/docs/releases/handling-releases.md
+++ b/docs/releases/handling-releases.md
@@ -1,11 +1,12 @@
 # Handling Releases
 
-This document outlines the process of releasing new versions of the blocks plugin.
+This document (and the [related PR template](../../.github/release_pull_request_template.md)) outlines the process of releasing new versions of the blocks plugin.
 
 ## Prerequisites - what you need to release WooCommerce Blocks
 
 -   You should be set up for development - for more info see [this doc](../contributors/getting-started.md).
 -   Install & set up [GitHub hub](https://hub.github.com) tools.
+-   You will need `svn` installed for pushing to WPORG.
 -   Configure a GitHub token for release scripts to use.
     -   https://github.com/settings/tokens
     -   Select the following permissions:
@@ -46,9 +47,13 @@ _Outcome_: **Team is aware of release and in agreement about what fixes & featur
     -   For _major_ and _minor_ releases, create branch: `release/X.X`.
     -   For _patch_ releases, the branch should already exist.
 
+_Outcome_: __There is a branch on origin repo for the release.__ (Note - it may not have all changes on it yet - that's next.)
+
 #### Create a release pull request
 
-Using the [release pull request template](../../.github/release_pull_request_template.md), create a pull request for the release branch you just created. This pull request will have changes merged to the min branch, but might not be a straight merge if it contains cherry-picked commits. The pull request also contains the checklist to go over as a part of the release along with being a place to contain all planning/communication around the release. The checklist should be completed and the pull request has an approved review from at least one team member before you do the Github deploy or release the plugin to WordPress.org.
+Using the [release pull request template](../../.github/release_pull_request_template.md), create a pull request for the release branch. This contains a checklist to go over, and serves as a place to contain all planning/communication around the release. The release process can span a few days, so it's this PR is an essential way for the team to keep track.
+
+You'll work through the checklist over the rest of the process below. The PR will remain open and unmerged until the release is live - you'll merge it much later (see `After release` below).
 
 ##### Patch releases against latest main branch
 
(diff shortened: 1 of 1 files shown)


docs/releases/handling-releases.md
```diff
@@ -46,9 +47,13 @@ _Outcome_: **Team is aware of release and in agreement about what fixes & featur
...
 
+_Outcome_: __There is a branch on origin repo for the release.__ (Note - it may not have all changes on it yet - that's next.)
+
 #### Create a release pull request
 
-Using the [release pull request template](../../.github/release_pull_request_template.md), create a pull request for the release branch you just created. This pull request will have changes merged to the min branch, but might not be a straight merge if it contains cherry-picked commits. The pull request also contains the checklist to go over as a part of the release along with being a place to contain all planning/communication around the release. The checklist should be completed and the pull request has an approved review from at least one team member before you do the Github deploy or release the plugin to WordPress.org.
+Using the [release pull request template](../../.github/release_pull_request_template.md), create a pull request for the release branch. This contains a checklist to go over, and serves as a place to contain all planning/communication around the release. The release process can span a few days, so it's this PR is an essential way for the team to keep track.
```
nerrad, line 54:
```suggestion
Using the [release pull request template](../../.github/release_pull_request_template.md), create a pull request for the release branch. This contains a checklist to go over, and serves as a place to contain all planning/communication around the release. The release process can span a few days, so this PR is an essential way for the team to keep track.
```

docs/releases/handling-releases.md
```diff
@@ -146,27 +159,46 @@ _Outcome_: **Customers can install/update via WPORG; WPORG plugin page is up to
...
 -   Close any epics that are completed as of this release and remove any unfinished issues in that from the epic.
 
+_Outcome:_ __The release milestone is archived and is an accurate record of what was included.__
+
+_Outcome:_ __There is a new release milestone ready for the next iteration.__
+
 #### Create pull request for updating the package in WooCommerce core.
 
-If the tagged release should be updated in WooCommerce core, do this immediately following our release.
+All releases (except RCs, betas etc) should be included in WooCommerce core. We do this by adding a PR on WooCommerce Core repo immediately after our release is completed.
```
nerrad, line 184:
```suggestion
All releases (except pre-releases) should be included in WooCommerce core. We do this by adding a PR on WooCommerce Core repo immediately after our release is completed.
```

haszari: Leaving this as is – `all releases except pre-releases` is less clear in my opinion.

haszari:
> I agree with your other notes about future improvements but I'd like to hold off implementing any of that until the automation is done. Much of this will be simplified and be redundant once the automation lands and we can probably end up with a much more concise document that just has a checklist.

Sounds good! Where are we tracking the automation goals/project/todos by the way? Perhaps I can help with that.

docs/releases/handling-releases.md
nerrad: > is less clear in my opinion.

How so? Aren't RCs, betas pre-releases? The concern I have is the `etc` you tacked on there. What is `etc`?

nerrad:
> Where are we tracking the automation goals/project/todos by the way?

[here](https://github.com/woocommerce/automations/issues/6). I've currently got an in progress pull locally that I haven't pushed yet, as soon as I have the basics (creating a pull request) I'll push it. Then other items from the issue can be done in individual pulls.

docs/releases/handling-releases.md
haszari: We can take off the etc - let's revisit this as we iterate.

PR #2833, opened 2020-07-08
Cart block: show products on backorder

### LINKED ISSUE #2825: Cart block not showing when product is on backorder
If you allow backorders for an out-of-stock product, an alert is shown on the core WooCommerce Cart page, but not in the WC Blocks Cart page.

Core WC cart page:

Markup on 2020-07-07 at 10:55:41

Cart block:

Markup on 2020-07-07 at 10:56:00

**Expected behavior**
I would expect a notice in the cart block similar to the stock WC cart page.

From: https://wordpress.org/support/topic/backorder-message-not-visible-on-cart-page/

Aljullu:
Fixes #2825.

### Screenshots

_Cart:_
imatge

_Checkout:_
imatge

1. Edit a product and under Inventory set:
  1.1 Manage stock? Checked.
  1.2 Stock quantity: 0.
  1.3 Allow backorders? Allow, but notify customer.
2. Update the product and go to the frontend of the store.
3. Add that product to the cart.
4. Go to the Cart page and verify the Cart block shows the product is on backorder.
5. Proceed to the Checkout page and verify the backorder notification appears in the sidebar.
6. Would be good to verify there are no style regressions with the 'low stock' and 'product sale' badges.

### Changelog

> Cart and Checkout blocks show a notification for products on backorder.


assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -5,6 +5,7 @@ import { __, sprintf } from '@wordpress/i18n';
 import { getCurrency } from '@woocommerce/base-utils';
 import Label from '@woocommerce/base-components/label';
 import {
+	ProductBackorderNotification,
```
haszari, line 8: `Notification` doesn't seem quite right for this – to me that suggests this is a notice or snackbar notification. I'd go with `badge` for consistency, e.g. `ProductBackorderBadge`. Minor nitpick - your call :)

assets/js/base/components/cart-checkout/product-badge/index.js
```diff
@@ -0,0 +1,29 @@
...
+ */
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+
+const ProductBadge = ( { children, className } ) => {
```
haszari, line 12: Nice - abstract out this component 🚀

assets/js/base/components/cart-checkout/product-variation-data/index.js
```diff
@@ -9,7 +9,7 @@ import classNames from 'classnames';
  * Returns a formatted element containing variation details.
  */
 const ProductVariationData = ( { className, variation = [] } ) => {
-	if ( ! variation ) {
+	if ( ! variation || variation.length === 0 ) {
```
haszari, line 12: Why this change? Just curious.

src/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -45,80 +45,86 @@ public function __construct( ImageAttachmentSchema $image_attachment_schema ) {
...
 			],
-			'backorders_allowed'  => [
+			'backorders_allowed'          => [
 				'description' => __( 'True if backorders are allowed past stock availability.', 'woo-gutenberg-products-block' ),
 				'type'        => [ 'boolean' ],
 				'context'     => [ 'view', 'edit' ],
 				'readonly'    => true,
 			],
-			'sold_individually'   => [
+			'show_backorder_notification' => [
```
haszari, line 108: If this can be renamed safely (badge vs notification) then I'd vote for that. Though I can see that `notification` is used for this in core api / backend.

src/StoreApi/docs/cart-items.md
```diff
@@ -35,6 +35,7 @@ curl "https://example-store.com/wp-json/wc/store/cart/items"
     "sku": "woo-beanie",
     "low_stock_remaining": null,
     "backorders_allowed": false,
+    "show_backorder_notification": false,
```
haszari, line 38: Thanks for keeping the API docs updated 🎖️

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -112,6 +114,9 @@ const CartLineItemRow = ( { lineItem = {} } ) => {
 					disabled={ isPendingDelete }
 				/>
 				<ProductLowStockBadge lowStockRemaining={ lowStockRemaining } />
+				{ showBackorderNotification && (
+					<ProductBackorderNotification />
+				) }
```
haszari, line 119: I'm thinking maybe we need to combine the badge logic into a single component. I _think_ we only ever want to show one badge at a time - they aren't independent. 

In my testing I got `low stock` and `on backorder` showing by bumping up the quantity of a low-stock, allows-backorder product:

Screen Shot 2020-07-13 at 10 41 13 AM

I think if the product is on backorder, we show the backorder badge only – it supercedes the low-stock badge. So `low stock` means shopper 'has the last one' and should order fast, `backorder` means that one or more items is on backorder, so shipping might be a while off.

assets/js/base/components/cart-checkout/product-variation-data/index.js
Aljullu: Ah, I should have explained that. It's a small change to avoid an empty `div` to appear.

Before this change, when `variation` was an empty array, this component was rendering `<div class="wc-block-components-product-variation-dat"></div>`. With the check added in this PR, we verify the array has some items before rendering the `div`.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
Aljullu: Good point. Changed in 4c26498c0f063763e0592ccb698cf08a2280eea1.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
Aljullu: Good catch. In 4c26498c0f063763e0592ccb698cf08a2280eea1 I made it so the stock badge doesn't appear if the backorder badge is shown.

src/StoreApi/Schemas/CartItemSchema.php
Aljullu: Fixed in 4c26498c0f063763e0592ccb698cf08a2280eea1 too.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -60,7 +62,12 @@ const OrderSummaryItem = ( { cartItem } ) => {
 						value={ linePrice }
 					/>
 				</div>
-				<ProductLowStockBadge lowStockRemaining={ lowStockRemaining } />
+				{ ! showBackorderBadge && (
+					<ProductLowStockBadge
+						lowStockRemaining={ lowStockRemaining }
+					/>
+				) }
+				{ showBackorderBadge && <ProductBackorderBadge /> }
```
haszari, line 70: I see from this logic that the two badges have different behaviour/interface:

- low stock badge auto hides itself when appropriate
- backorder badge always renders something, so the client component needs to control render directly

I can see why this is the case - `lowStockRemaining == 0`, but it might be clearer to use ternary logic here, e.g.

```
showBackorderBadge ? <ProductBackorderBadge /> : lowStockRemaining && <ProductLowStockBadge lowStockRemaining={ lowStockRemaining } />
```

🤔 

Or maybe that's more confusing! Not a blocker just sharing an observation - totally happy with the code as is.

Aljullu: I like it! Prettier breaks it into several lines, so I don't think it's more confusing at all. Changed in: e635ecfa207030cdbd103fd2ec0e58249c8a2a7f.

PR #2840, opened 2020-07-09
allow payment methods to disable based on shipping (API change)

### LINKED ISSUE #2838: Support for payment methods that change availability based on user actions/data in checkout - e.g. COD supported for specific shipping methods only
`Cash on Delivery` (COD) payment method can be enabled for merchant-selected shipping options only. For example, a store uses one particular shipping service that allows COD, and others that don't support it. 

In the checkout block, this means that the COD payment method needs to be made available only for valid shipping methods. And if the shopper changes shipping method, or address (etc) the COD payment method may need to disable/enable dynamically. See related PR in progress for COD - #2831

Currently, payment methods return boolean from the [`canMakePayment`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/docs/extensibility/payment-method-integration.md#client-side-integration) callback to indicate if the payment method can be used. This callback is currently called once on initialisation (note: some payment methods might call an API from `canMakePayment`).

Extend the payment methods JS API to allow for dynamically-available payment methods. 

One way to do this would be to pass `canMakePayment` all relevant data (e.g. selected shipping methods, cart items), and retrigger the callback when anything changes. For example, when a different shipping option is selected, `canMakePayment` is called again. If we went this route, we'd need to carefully document this so payment methods that run expensive or async (e.g. third-party API) code know to cache it in some way. Potentially could provide a different API/callback for expensive initialisation.

Another option could allow payment methods to register for events/callbacks when specific data changes (e.g. different shipping method selected).

Note - payment methods might depend on other changing state, so we should consider that when implementing any JS API changes. For example, COD also potentially depends on the cart items (if merchant disables `Accept COD if the order is virtual`). If we implement #2583 then this will affect COD - the new API will have to support this case also. I.e. user removes physical products from order (on checkout page), COD may no longer be available.

haszari:
Fixes #2838

This PR changes the payment methods registration hook so that it calls `canMakePayment` on each payment method whenever shopper changes shipping. 

The goal here is to better support payment methods that need to disable (hide/show) dependent on the selected shipping rate, or potentially other factors in future. See #2838 and #2831 for more info - `Cash on Delivery` needs this capability.

This is quite different from the previous usage/expectations of the API, so will need careful testing.

The change to the API can't be directly tested, as none of the current payment methods use the feature.

So please test all supported payment methods and flows. I've tested all but `Stripe Payment Request` (Apple / Chrome Pay) as it's not available for me.

To test the new feature, you can hack an existing payment method `canMakePayment` callback and add the new `selectedShippingMethods` parameter. 

- `canMakePayment` should be called whenever shipping method changes.
- The selected shipping method(s) should be passed to `canMakePayment`.

### Some other context / notes
As far as I understand, `canMakePayment` has two purposes:

1. Determine if a payment method should be available (i.e. presented to user in checkout).
2. Initialise (set up) the payment method.

In my view, (1) is the primary purpose, and (2) could be considered a side effect. Stripe (CC and Payment Request) are the only methods that use the "initialisation" feature right now. In any case, this is how the API works currently so we need to take care making this change.

I hope that the expensive, async init for Stripe in `canMakePayment` is effectively cached, so on subsequent calls the promise returns immediately. If not, we may need to add memoization or other caching.

Other thoughts we might want to follow up:

- We could make `canMakePayment` interface simpler:
   - default to true if payment method does nothing
   - allow a boolean value with no callback
- There's some code in `usePaymentMethodRegistration()` which catches initialization errors and shows a notice to admins. I think this is primarily there for Stripe integration, and is not really general purpose. We could consider moving this out so each payment method extension handles its own init, if needed, so that `canMakePayment` is focused specifically on availability (which could depend on successful init where appropriate!). 

### Changelog

> Add suggested changelog entry here. TBC TBC

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -11,7 +11,7 @@ import {
 	useEditorContext,
 	useShippingDataContext,
 } from '@woocommerce/base-context';
-import { useStoreCart } from '@woocommerce/base-hooks';
+import { useStoreCart, useShallowEqual } from '@woocommerce/base-hooks';
 import { CURRENT_USER_IS_ADMIN } from '@woocommerce/block-settings';
 
 /**
@@ -54,10 +54,8 @@ const usePaymentMethodRegistration = (
 ) => {
 	const [ isInitialized, setIsInitialized ] = useState( false );
 	const { isEditor } = useEditorContext();
-	const {
-		selectedRates: selectedShippingMethods,
-		shippingAddress,
-	} = useShippingDataContext();
+	const { selectedRates, shippingAddress } = useShippingDataContext();
+	const selectedShippingMethods = useShallowEqual( selectedRates );
 	const { cartTotals, cartNeedsShipping } = useStoreCart();
 	const canPayArgument = useRef( {
 		cartTotals,
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -54,12 +54,16 @@ const usePaymentMethodRegistration = (
 ) => {
 	const [ isInitialized, setIsInitialized ] = useState( false );
 	const { isEditor } = useEditorContext();
-	const { shippingAddress } = useShippingDataContext();
+	const {
+		selectedRates: selectedShippingMethods,
+		shippingAddress,
+	} = useShippingDataContext();
 	const { cartTotals, cartNeedsShipping } = useStoreCart();
 	const canPayArgument = useRef( {
 		cartTotals,
 		cartNeedsShipping,
 		shippingAddress,
+		selectedShippingMethods,
 	} );
 
 	useEffect( () => {
@@ -67,51 +71,64 @@ const usePaymentMethodRegistration = (
 			cartTotals,
 			cartNeedsShipping,
 			shippingAddress,
+			selectedShippingMethods,
 		};
-	}, [ cartTotals, cartNeedsShipping, shippingAddress ] );
+	}, [
+		cartTotals,
+		cartNeedsShipping,
+		shippingAddress,
+		selectedShippingMethods,
+	] );
 
-	const resolveCanMakePayments = useCallback( async () => {
-		let initializedPaymentMethods = {},
-			canPay;
-		const setInitializedPaymentMethods = ( paymentMethod ) => {
-			initializedPaymentMethods = {
-				...initializedPaymentMethods,
+	const refreshCanMakePayments = useCallback( async () => {
+		let availablePaymentMethods = {};
+		const addAvailablePaymentMethod = ( paymentMethod ) => {
+			availablePaymentMethods = {
+				...availablePaymentMethods,
 				[ paymentMethod.name ]: paymentMethod,
 			};
 		};
 		for ( const paymentMethodName in registeredPaymentMethods ) {
-			const current = registeredPaymentMethods[ paymentMethodName ];
+			const paymentMethod = registeredPaymentMethods[ paymentMethodName ];
 
+			// In editor, shortcut so all payment methods show as available.
 			if ( isEditor ) {
-				setInitializedPaymentMethods( current );
+				addAvailablePaymentMethod( paymentMethod );
 				continue;
 			}
 
+			// In front end, ask payment method if it should be available.
 			try {
-				canPay = await Promise.resolve(
-					current.canMakePayment( canPayArgument.current )
+				const canPay = await Promise.resolve(
+					paymentMethod.canMakePayment( canPayArgument.current )
 				);
 				if ( canPay ) {
 					if ( canPay.error ) {
 						throw new Error( canPay.error.message );
 					}
-					setInitializedPaymentMethods( current );
+					addAvailablePaymentMethod( paymentMethod );
 				}
 			} catch ( e ) {
+				// If user is admin, show payment `canMakePayment` errors as a notice.
 				handleRegistrationError( e );
 			}
 		}
-		// all payment methods have been initialized so dispatch and set
-		dispatcher( initializedPaymentMethods );
+
+		// Re-dispatch available payment methods to store.
+		dispatcher( availablePaymentMethods );
+
+		// Note: some payment methods use the `canMakePayment` callback to initialize / setup.
+		// Example: Stripe CC, Stripe Payment Request.
+		// That's why we track "is initialised" state here.
 		setIsInitialized( true );
 	}, [ dispatcher, isEditor, registeredPaymentMethods ] );
 
-	// if not initialized invoke the callback to kick off resolving the payments.
+	// Determine which payment methods are available initially and whenever
+	// shipping methods change.
+	// Some payment methods (e.g. COD) can be disabled for specific shipping methods.
 	useEffect( () => {
-		if ( ! isInitialized ) {
-			resolveCanMakePayments();
-		}
-	}, [ resolveCanMakePayments, isInitialized ] );
+		refreshCanMakePayments();
+	}, [ refreshCanMakePayments, selectedShippingMethods ] );
 
 	return isInitialized;
 };
--- a/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
@@ -16,30 +16,47 @@ const ApplePayPreview = () => <img src={ applePayImage } alt="" />;
 const canPayStripePromise = loadStripe();
 const componentStripePromise = loadStripe();
 
+let apiInitialized = false,
+	canPay = false;
+
+// Initialise stripe API client and determine if payment method can be used
+// in current environment (e.g. geo + shopper has payment settings configured).
+function paymentRequestAvailable( currencyCode ) {
+	// If we've already initialised, return the cached results.
+	if ( apiInitialized ) {
+		return canPay;
+	}
+
+	return canPayStripePromise.then( ( stripe ) => {
+		if ( stripe === null ) {
+			return false;
+		}
+		// Do a test payment to confirm if payment method is available.
+		const paymentRequest = stripe.paymentRequest( {
+			total: {
+				label: 'Test total',
+				amount: 1000,
+			},
+			country: getSetting( 'baseLocation', {} )?.country,
+			currency: currencyCode,
+		} );
+		return paymentRequest.canMakePayment().then( ( result ) => {
+			canPay = !! result;
+			apiInitialized = true;
+			return canPay;
+		} );
+	} );
+}
+
 const PaymentRequestPaymentMethod = {
 	name: PAYMENT_METHOD_NAME,
 	content: <PaymentRequestExpress stripe={ componentStripePromise } />,
 	edit: <ApplePayPreview />,
 	canMakePayment: ( cartData ) =>
-		canPayStripePromise.then( ( stripe ) => {
-			if ( stripe === null ) {
-				return false;
-			}
-			// do a test payment request to check if payment request payment can be
-			// done.
-			const paymentRequest = stripe.paymentRequest( {
-				total: {
-					label: 'Test total',
-					amount: 1000,
-				},
-				country: getSetting( 'baseLocation', {} )?.country,
-				// eslint-disable-next-line camelcase
-				currency: cartData?.cartTotals?.currency_code?.toLowerCase(),
-			} );
-			return paymentRequest
-				.canMakePayment()
-				.then( ( result ) => !! result );
-		} ),
+		paymentRequestAvailable(
+			// eslint-disable-next-line camelcase
+			cartData?.cartTotals?.currency_code?.toLowerCase()
+		),
 	paymentMethodId: 'stripe',
 };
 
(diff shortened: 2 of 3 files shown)


assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -54,64 +54,81 @@ const usePaymentMethodRegistration = (
...
 				if ( canPay ) {
 					if ( canPay.error ) {
 						throw new Error( canPay.error.message );
 					}
-					setInitializedPaymentMethods( current );
+					addAvailablePaymentMethod( paymentMethod );
 				}
 			} catch ( e ) {
+				// If user is admin, show payment `canMakePayment` errors as a notice.
 				handleRegistrationError( e );
```
haszari, line 113: This notice feels tightly coupled to Stripe integration - though maybe this could be useful for other methods in future. 

I'm thinking we could decouple the initialisation of the payment methods from `canMakePayment`. Not sure if we need to be involved in the init explicitly - maybe payment methods can manage this themselves. 

Related, is it essential that we wrap `canMakePayment` in try/catch? Feels heavy-handed, though I could understand that might be useful if this was an "initialise yourself" callback.

haszari:
If this works out, we'll need to update the API docs. Made some [tweaks in the COD PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2831/files), but this is a bigger change.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -54,64 +54,81 @@ const usePaymentMethodRegistration = (
...
+	// Determine which payment methods are available initially and whenever
+	// shipping methods change.
+	// Some payment methods (e.g. COD) can be disabled for specific shipping methods.
 	useEffect( () => {
-		if ( ! isInitialized ) {
-			resolveCanMakePayments();
-		}
-	}, [ resolveCanMakePayments, isInitialized ] );
+		refreshCanMakePayments();
+	}, [ refreshCanMakePayments, selectedShippingMethods ] );
```
Aljullu, line 131: This `useEffect` is running some times when it's not needed. That's because `selectedShippingMethods` is an object, so when added as an effect dependency, the effect is running again every time because the object reference is always different.

That could be fixed with [`useShallowEqual`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/base/hooks/use-shallow-equal.js). Something like:

```JSX
const newSelectedShippingMethods = useShallowEqual(
	selectedShippingMethods
);
```

Then, you could replace `selectedShippingMethods` instances with `newSelectedShippingMethods` and it shouldn't trigger the effect so often.

nerrad: Yes this is a problem. This also changes the logic because `isInitialized` is removed as a dependency.

Here's what I think we should do instead that could also help address some of the point Rua made:

> I'm thinking we could decouple the initialisation of the payment methods from canMakePayment. Not sure if we need to be involved in the init explicitly - maybe payment methods can manage this themselves

There _is_ clear need for payment methods to differentiate between when `canMakePayment` is called as a part of the payment method setup/initialization in checkout and when it's called for other data changes that might cause a change in the availability.  So what I think we should do is:

Introduce a new property on the `canPayArgument.current` of 'isInitialized', and it would be the value of the initialized state.

Keep the _existing_ initialization logic the same (including the `useEffect` here) so that we are preserving the initialization process. So `use-payment-method-registration` hook is explicitly for initialization and only sets the `canPayArgument.current.isInitialized` to true when initialization completes. Payment methods can then use that property to derive when in checkout is in payment method initialization state.

Introduce a _new_ hook that is explicitly for calling `canMakePayment` when various other data (shipping etc) is changed and you feed the hook the `isInitialized` value. So something like this for a signature (name of hook subject to change :) ):

```js
useCanMakePaymentUpdates( isInitialized, paymentMethods, expressPaymentMethods  )
```

Then this hook is used for calling `canMakePayment` on payment methods with the updated data (but will always call with the argument property `isInitialized` true (and only _starts_ calling on data changes when that is true).

Then the hook can be implemented in the same place as `usePaymentMethods` and `useExpressPaymentMethods` are called (`payment-method-data-context.js`) where it receives the `isInitializeValue` returned from those, along with the initialized payment methods from the `paymentData` state in the context. 

A few things will be accomplished by doing this:

- separate initialization logic, from data change/update logic (easier to test and maintain).
- makes the purpose of each hook explicitly clear.
- Payment methods can detect in their `canMakePayment` callback whether they are being initialized by checkout or just being called because of a data update. So it makes it possible for them to decide what needs to happen for the payment method in those different states.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
nerrad: > This notice feels tightly coupled to Stripe integration - though maybe this could be useful for other methods in future.

How so? It will only happen if the payment method throws an error from it's `canMakePayment` callback. There's no reference to stripe in this logic.

> I'm thinking we could decouple the initialisation of the payment methods from canMakePayment. Not sure if we need to be involved in the init explicitly - maybe payment methods can manage this themselves.

I agree with splitting up logic more, however I disagree with your assessment. Checkout IS responsible for initializing payment methods, otherwise payment methods have no idea _when_ to initialize themselves. To be clear, in this existing logic (not with your changes) the purpose of this hook is to:

- grab all the registered payment methods.
- run through everyone of them and fire the `canMakePayment` hook to determine whether checkout should include that payment method in it's "ready to make payment" state (which essentially determines whether the payment displays or not).
- this is a checkout initialization routine for payment methods to determine whether they display on _initial_ checkout load. Whether or not payment methods do any of their initialization in `canMakePayment` is something checkout doesn't care about. It only cares that at this point, payment methods are saying - "yup you can show me to the customer".

I have some other thoughts on this in a later comment but I do think we need to keep this _checkout initialization state for payments_ distinct from where `canMakePayment` is called AFTER this for any potential updates. The reason is that with these changes, payment methods will need to know this state because `canPayment` is no longer called only once. So while checkout doesn't care if payment methods do initialization when `canMakePayment` is called, payment methods typically _might_ need to do their initialization during this call to determine the answer to "Can I make payment?" and they might only want to do that once. So basically we need to provide data to the payment method to make that decision.

> Related, is it essential that we wrap canMakePayment in try/catch? Feels heavy-handed, though I could understand that might be useful if this was an "initialise yourself" callback.

What about this feels heavy-handed? It's unclear how that term applies to using try/catch? 

This is essential because of the possibility third party payment methods might cause an error here and we want to catch that early and handle as opposed to having the third party payment method break checkout. The reason why there is a `CURRENT_USER_IS_ADMIN` check inside `handleRegistrationError` is so that we only surface the payment method related error to the store admin when they go to figure out why the payment method isn't showing up for a shopper (which is critical because we have no idea what the thrown error might contain from the third party payment method and I don't think we want to risk exposing sensitive information).

haszari: > How so? It will only happen if the payment method throws an error from it's canMakePayment callback. There's no reference to stripe in this logic.

Stripe is the only payment method right now that's doing anything initialisation in `canMakePayment`. But it's quite possible this pattern will work well for other payment gateways too.

> I agree with splitting up logic more, however I disagree with your assessment. Checkout IS responsible for initializing payment methods, otherwise payment methods have no idea when to initialize themselves.

For sure, I understand how it currently works. I'm wondering if we need to do this inside `canMakePayment` or if we can have a separate hook for initialisation. Coupling init to `canMakePayment` makes the interface more complicated – would be nice if we could avoid that.

What do you think about the possibility of making the init callback a separate property, and focus `canMakePayment` just on availability? Interested to hear your thoughts on that approach. 

Anyway I'll take another pass on this and revise it in line with your feedback :)

haszari: > So basically we need to provide data to the payment method to make that decision.

One thought on this - do we need to manage that boolean for the gateways? They can easily detect when they haven't initialised themselves. Just thinking out loud here, looking for ways to simplify the API where we can.

nerrad: > One thought on this - do we need to manage that boolean for the gateways?

It's a signal that checkout has done the initial loading of payment methods. You're right payment methods could likely manage their own internal needs here. I'm happy with you trying my suggested approach without passing this through to the payment method. But I still think the new hook will need the value.

nerrad: > Stripe is the only payment method right now that's doing anything initialisation in canMakePayment. But it's quite possible this pattern will work well for other payment gateways too.

To be clear, this is _explicitly_ defensive code to catch ANY errors from payment methods borking something in their `canMakePayment` callback. For critical flows like checkout I think it's crucial we have this kind of protection. So yes, Stripe is the only payment method right now that's doing any initialization, but that is _definitely_ not going to remain that way.

nerrad: > For sure, I understand how it currently works. I'm wondering if we need to do this inside canMakePayment or if we can have a separate hook for initialisation. Coupling init to canMakePayment makes the interface more complicated – would be nice if we could avoid that.

Let's think of this from the perspective of checkout and the payment method.

Checkout:

- only cares if the payment method reports it's ready to make payments.
- It doesn't care how the payment method initializes.
- It doesn't care how the payment method figures out it's ready to make payments.
- It just cares that if `canMakePayment` returns true, then the registered payment method is included for display to the shopper in the app.

Payment method

- doesn't know when checkout is going to fire the callback it provides on `canMakePayment` and really doesn't care. 
- it just cares that the values it's provided by checkout are enough for it to report whether it can make payment or not.

Given that information, I'm not sure we need to provide a separate interface for initializing  when essentially that's totally up to the payment method and how it chooses/needs to initialize. In writing this, I realize this further supports [your comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2840#discussion_r452491572) that in fact we don't need to send along the `isInitialized` boolean to the payment method provided callback.

Essentially the major change here from the perspective of payment methods is that `canMakePayment` can be invoked more than once and so payment methods need to account for that if there's anything expensive they do in the callback.

haszari: That works! I'll implement the optimisation suggested by @Aljullu and then rebase #2831 to make sure we've solved the problem for COD :)

haszari:
@nerrad  this is ready for another review - I think - the only change is the use of `useShallowEqual`. Let me know if I've missed anything else. One thing to double-check/consider - make sure that Stripe is only doing its expensive initialisation once. In my testing I think I confirmed this is working correctly.

FYI @Aljullu  I couldn't see a difference in number of calls to `refreshCanMakePayments` after using `useShallowEqual`. I get two calls on load and 2 calls when I change shipping option. Possibly there's other dependencies/interaction causing this, let me know if you think we should follow this up now. E.g. optimise to ensure that selecting a shipping option only calls back once.

I'll rebase #2831 on to this branch and get the dynamic COD availability working, so you can review this in context & test with the initial use case :)

haszari:
Note - this PR has been rebased on to master (to get the latest).

haszari:
Note - I haven't made any related API docs changes for this (or tests!) in this PR. I'm keen to do this as follow up - feel free to spin off issues for these @Aljullu @nerrad if you like.

Aljullu:
> FYI @Aljullu  I couldn't see a difference in number of calls to `refreshCanMakePayments` after using `useShallowEqual`. I get two calls on load and 2 calls when I change shipping option.

For me, it made a difference when modifying the shipping address (without changing the shipping method).

haszari:
Thanks for catching that issue with Payment Request (Apple/Chrome pay) @nerrad. I've made the fix in 05e63de - the stripe API is now initialised once on the first call, then we return the cached `canPay` value.

In my testing this is looking good but I can't test payment request (tried adding my card to Chrome but looks like it's not supported - geo or card related) - can you give this another test / review please?

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
```diff
@@ -16,30 +16,47 @@ const ApplePayPreview = () => <img src={ applePayImage } alt="" />;
...
+// in current environment (e.g. geo + shopper has payment settings configured).
+function paymentRequestAvailable( currencyCode ) {
+	// If we've already initialised, return the cached results.
+	if ( apiInitialized ) {
+		return canPay;
+	}
+
+	return canPayStripePromise.then( ( stripe ) => {
+		if ( stripe === null ) {
+			return false;
```
nerrad, line 32: Hmm we should probably set the cache here and return (because if stripe isn't in the callback then there's something wrong and non-recoverable, so no sense in repeated calls to the canMakePayment callback triggering additional checks).

As in my original example:

```js
if ( stripe === null ) {
    isInitialized = true;
    return canPay;
}
```

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
```diff
@@ -16,30 +16,47 @@ const ApplePayPreview = () => <img src={ applePayImage } alt="" />;
 const canPayStripePromise = loadStripe();
 const componentStripePromise = loadStripe();
 
+let apiInitialized = false,
```
nerrad, line 19: It's convention to use the `is` prefix on variables holding boolean type values (especially when they are in conditional expressions) as it reads more naturally. I think `isInitialized` is sufficient because the scope is already limited to this file but if you want to include api in there then `isApiInitialized` or even `isStripeInitialized` if you want to be super specific.

Issue #2836, opened 2020-07-08
All Products block doesn't work when added to the empty cart


Aljullu:
**To Reproduce**
1. Add the Cart block and switch to the Empty Cart edit mode.
2. Replace the Newest products block with the All Products block.
3. View it in the frontend.
4. Notice the All Products block only shows the placeholders:

![imatge](https://user-images.githubusercontent.com/3616980/86940083-a779c880-c142-11ea-8e86-7ebf51e3169c.png)

PR #2842, opened 2020-07-09
Deregister core cart/checkout scripts and styles when rendering the blocks

### LINKED ISSUE #2239: remove WooCommerce core enqueued styles and scripts when a block page is the default Cart and Checkout page.
**Describe the bug**
Currently Woo Core loads some styles and scripts in the Cart and Checkout page, if our block is the default page, those styles would still load anyway (since they use `is_checkout()` and `is_cart()`) functions, therefore adding extra weights.
Checkout loads 44kb of gzipped JS and CSS.
Cart loads 38kb of gzipped JS and CSS.

I did a small modification in [this line 363](https://github.com/woocommerce/woocommerce//blob/4ae1e1f5fc6e9da4c4d0348d1590cd09b51fbbce/includes/class-wc-frontend-scripts.php#L363) and 366 and 375 by adding another condition

```php
if ( is_cart() && ! has_block( 'woocommerce/cart' ) )
```

That condition reduced bundle size in Checkout 13kb, and Cart to 9kb (most notably it removed WooSelect), other wins are still possible (some scripts that are loaded on all pages, not just those pages, can also be removed from those pages).

Before:
image

After:
image

**Wins**: 29kb gzipped from Woo alone, and 14kb gzipped from other places.

Before:
image

After:
image

**Wins**: 33kb gzipped from Woo alone.

**To Reproduce**
0. set the Cart and Checkout pages as the default Cart and Checkout pages.
1. Open devtools in the network tab.
2. Open one of the pages.
3. let the page download, now filter with `plugins/woocommerce/` it should leave the dependencies that woo is loading directly (this exclude other dependencies caused by woo that are loaded from other places, say wp-includes).
4. See the scripts that are not needed.

Not fully sure how to solve this from within our render function, since it's called way late, so we might need to add an action that runs after scripts finished registering to remove them.

Aljullu:
Fixes #2239 

This PR deregisters the scripts and styles used by cart and checkout shortcodes when rendering the blocks. I took the list of scripts/styles to deregister from these lines in Core:

https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-frontend-scripts.php#L363-L377

It looks like we were relying on some logic from `cart.js` to redirect the empty cart to the full cart when a product was added. I needed to rewrite that logic in the block. In order to do that, I needed to subscribe to a jQuery event since that's what's used in Core. I added a translation layer to convert jQuery events to JS native events. While not ideal, that's the best solution I could find given the current situation where core uses jQuery.

**Test Empty cart redirects to the Full cart when a product is added:**

1. Go to the cart page without having any product in the cart. Add one from the Block below and verify you are redirected to the `full cart` view.
1. Repeat the step above but before doing that, edit the empty cart template and replace the Newest products block with a shortcode (ie: `[products limit="3" columns="3" visibility="featured" ]`).

Note: It's not possible to test this flow with the All Products block because of #2836.

**Test there are no regressions in the purchase flow:**

1. Do a purchase from start to end with the blocks and verify everything works and the confirmation page appears after payment.
1. Do a purchase with the shortcodes and verify there are no regressions: functionality should work as usual and styles should be loaded.

### Changelog

> Don't load shortcode Cart and Checkout scripts when using the blocks.

### DIFF
--- a/assets/js/base/utils/legacy-events.js
+++ b/assets/js/base/utils/legacy-events.js
@@ -1,19 +1,51 @@
 const Event = window.Event || null;
 
-// This is a hack to trigger cart updates till we migrate to block based cart
-// that relies on the store, see
-// https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1247
-export const triggerFragmentRefresh = () => {
+/**
+ * Wrapper function to dispatch an event so it's compatible with IE11.
+ *
+ * @param {string}  name       Name of the event to dispatch.
+ * @param {boolean} bubbles    Whether the event bubbles.
+ * @param {boolean} cancelable Whether the event is cancelable.
+ */
+const dispatchEvent = ( name, bubbles = false, cancelable = false ) => {
 	// In IE, Event is an object and can't be instantiated with `new Event()`.
 	if ( typeof Event === 'function' ) {
-		const event = new Event( 'wc_fragment_refresh', {
-			bubbles: true,
-			cancelable: true,
+		const event = new Event( name, {
+			bubbles,
+			cancelable,
 		} );
 		document.body.dispatchEvent( event );
 	} else {
 		const event = document.createEvent( 'Event' );
-		event.initEvent( 'wc_fragment_refresh', true, true );
+		event.initEvent( name, bubbles, cancelable );
 		document.body.dispatchEvent( event );
 	}
+};
+
+// This is a hack to trigger cart updates till we migrate to block based cart
+// that relies on the store, see
+// https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1247
+export const triggerFragmentRefresh = () => {
+	dispatchEvent( 'wc_fragment_refresh', true, true );
+};
+
+/**
+ * Function that listens to a jQuery event and dispatches a native JS event.
+ * Useful to convert WC Core events into events that can be read by blocks.
+ *
+ * @param {string}  jQueryEventName Name of the jQuery event to listen to.
+ * @param {string}  nativeEventName Name of the native event to dispatch.
+ * @param {boolean} bubbles         Whether the event bubbles.
+ * @param {boolean} cancelable      Whether the event is cancelable.
+ */
+export const translateJQueryEventToNative = (
+	jQueryEventName,
+	nativeEventName,
+	bubbles = false,
+	cancelable = false
+) => {
+	// @ts-ignore jQuery is window global
+	jQuery( document ).on( jQueryEventName, () => {
+		dispatchEvent( nativeEventName, bubbles, cancelable );
+	} );
 };
--- a/assets/js/base/utils/legacy-events.js
+++ b/assets/js/base/utils/legacy-events.js
@@ -45,6 +45,10 @@ export const translateJQueryEventToNative = (
 	cancelable = false
 ) => {
 	// @ts-ignore -- jQuery is window global
+	if ( typeof jQuery !== 'function' ) {
+		return;
+	}
+	// @ts-ignore -- jQuery is window global
 	jQuery( document ).on( jQueryEventName, () => {
 		dispatchEvent( nativeEventName, bubbles, cancelable );
 	} );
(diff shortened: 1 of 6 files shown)


assets/js/base/utils/legacy-events.js
```diff
@@ -1,19 +1,51 @@
...
+ * @param {boolean} bubbles         Whether the event bubbles.
+ * @param {boolean} cancelable      Whether the event is cancelable.
+ */
+export const translateJQueryEventToNative = (
+	jQueryEventName,
+	nativeEventName,
+	bubbles = false,
+	cancelable = false
+) => {
+	// @ts-ignore jQuery is window global
```
nerrad, line 47: Typical pattern for comments on linting blocks is something like this:

```suggestion
	// @ts-ignore -- jQuery is window global
```

assets/js/blocks/cart-checkout/cart/block.js
```diff
@@ -1,22 +1,51 @@
...
+const Block = ( { emptyCart, attributes, scrollToTop } ) => {
 	const { cartItems, cartIsLoading } = useStoreCart();
 
+	useEffect( () => {
+		const invalidateCartData = () => {
+			if ( cartItems.length === 0 ) {
+				dispatch( storeKey ).invalidateResolutionForStore();
+				scrollToTop();
+			}
+		};
```
nerrad, line 33: Nifty! While this works great for cartItems removed from the block, I wonder if we should also add an event listener for when an item is removed from the cart js _widget_ and show the empty cart placeholder when the items in the cart are empty? If you agree, definitely can be done in another issue (I just thought it would be a good use of the jQuery translation layer as well)

assets/js/base/utils/legacy-events.js
```diff
@@ -1,19 +1,51 @@
...
+ * @param {string}  nativeEventName Name of the native event to dispatch.
+ * @param {boolean} bubbles         Whether the event bubbles.
+ * @param {boolean} cancelable      Whether the event is cancelable.
+ */
+export const translateJQueryEventToNative = (
+	jQueryEventName,
+	nativeEventName,
+	bubbles = false,
+	cancelable = false
+) => {
```
nerrad, line 46: Nice 👏

assets/js/base/utils/legacy-events.js
```diff
@@ -1,19 +1,51 @@
...
+	// @ts-ignore jQuery is window global
+	jQuery( document ).on( jQueryEventName, () => {
+		dispatchEvent( nativeEventName, bubbles, cancelable );
+	} );
```
nerrad, line 50: I think we'll still want to add some handling if `jQuery` is NOT present in the window. There will be effort to eliminate jQuery usage so we'll want to gracefully handle when it's not present. Can you add a check here for that?

Aljullu: Makes sense. Added the check in 35f7a60854a13fc1635e30d8a36af87738e920d5.

assets/js/blocks/cart-checkout/cart/block.js
Aljullu: Good point, done in 4cfb838be992f6a4b25bcaf8d80129e408097c95.

The Cart widget was not working for me in the Cart page, so to test I needed to create a post with the Cart block and the Cart widget in the sidebar:

Peek 2020-07-10 12-07

assets/js/base/utils/legacy-events.js
```diff
@@ -44,7 +44,11 @@ export const translateJQueryEventToNative = (
...
 ) => {
-	// @ts-ignore jQuery is window global
+	// @ts-ignore -- jQuery is window global
+	if ( typeof jQuery !== 'function' ) {
+		return;
+	}
+	// @ts-ignore -- jQuery is window global
 	jQuery( document ).on( jQueryEventName, () => {
 		dispatchEvent( nativeEventName, bubbles, cancelable );
 	} );
```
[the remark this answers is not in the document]
Aljullu: That's something that was bothering me too. I was relying on the fact that the jQuery event would be cleared out when navigating to another page. But I think it makes more sense registering and unregistering it with component mount/unmount.

I refactored it in c708db70d76a834b2ce6a4d46c5dee24905ad9f6, so the `useEffect` takes care of adding and removing the jQuery → native event subscriptions when it's mounted and unmounted.

PR #2851, opened 2020-07-13
Support "create account" option in checkout block

### LINKED ISSUE #2377: Checkout Block: Add support for - `Allow customers to create an account during checkout` option
There's a related option that we don't yet support - `Allow customers to create an account during checkout`.

Screen Shot 2020-05-01 at 1 43 21 PM

_Originally posted by @haszari in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2252#issuecomment-622208649_

haszari:
Fixes #2377

This PR adds support for creating an account during checkout. 

This is controlled by a new block option `Allow shopper to sign up for a user account during checkout`. If enabled, the user will be prompted to create an account at checkout (if they are not logged in already).

Screen Shot 2020-08-06 at 11 05 08 AM

This is similar in principle to an existing store setting `Allow customers to create an account during checkout`. We ignore the store setting (in preference of the new block option), though the block defaults to the current value of the store setting. 

The `Allow customers to place orders without an account` option (`woocommerce_enable_guest_checkout`) is also relevant. If enabled, the `Create an account` checkbox will be shown; otherwise, an account is required, so the checkbox is not offered to shopper at checkout. This is [consistent with current woo behaviour](https://github.com/woocommerce/woocommerce/blob/02cf0dfaed5923513de0c88add597d1560c2cfd2/includes/class-wc-checkout.php#L1006) - if an account is required, user isn't shown the checkbox, as they don't have a choice.

### Feature gating – dev build only
This feature is currently gated to the dev build only. It should not be present or visible in the feature plugin (release build) or when blocks is included as a package in WooCommerce core.

Most of the code is live and included in all builds, but won't be triggered due to feature gating:

- All public methods of `CreateAccount` service are protected and will silently return if the feature is disabled.
- In the editor, the sidebar option `Allow create account` is only displayed when appropriate.
- In the front end, the `Create an account?` checkbox is not rendered if feature is disabled.

Note that various business logic and schema support for the feature is included in all builds, for example:

- Setting in redux store.
- Block attribute.
- Field in checkout REST API schema.

##### Why feature gate to dev plugin? 
The sign-up flow is essentially a core platform feature. We want to improve the flow to ensure we have a seamless easy checkout experience following best practices.

The goal with this PR was to implement the improved signup flow as part of blocks plugin, for eventual inclusion in Woo core. This of course should not break or affect any core signup flows/modules.

This was not practical – in particular, we needed a new `Set password` screen in core which is not possible to add from an extension. 

So this PR currently uses a temporary, non-store-themed set password screen from WordPress core. This allows us to test the flow but is not the experience we want to ship to stores.

In parallel we'll implement a new `Set password` screen in Woo core (https://github.com/woocommerce/woocommerce/issues/27754). When this has shipped in core we can update the checkout signup flow to use it and ship in the feature plugin (and ultimately in woo core package).

#### Merchant experience

New block option for allowing signup during checkout. Currently in a group by itself.

Screen Shot 2020-08-06 at 11 27 31 AM

Some aspects of the email can be customised by the merchant at `WooCommerce > Settings > Emails`. The options for `New account` email should work the same with this new email and the (legacy) core email.

Screen Shot 2020-08-11 at 3 46 37 PM

#### Shopper experience

A new checkbox is added to the first checkout step (when appropriate):

Screen Shot 2020-07-23 at 3 01 49 PM

Welcome email with set password link:

Screen Shot 2020-08-11 at 3 48 58 PM

Screen Shot 2020-08-11 at 3 46 15 PM

Set password form:

Screen Shot 2020-08-05 at 4 00 54 PM

#### Test feature is not present in production build
- Build a release version of the plugin, e.g. `npm run build:deploy`.
- Edit checkout block – ensure no `Account options` sidebar option is present.
- View checkout block on front end as anon/guest/incognito – ensure no `Create an account` checkbox is displayed.
- Complete various test purchases and ensure no accounts are created.

#### Test basic create account via checkout flow
0. Rebuild dev version of plugin, e.g. `npm start`.
1. Edit checkout page. Add checkout block if needed. Edit the block and enable signup option via checkbox in sidebar. Publish the page.
2. Switch to a browser/session where you are not logged in to your woo store - e.g. incognito or different browser.
3. Add stuff to cart, proceed to checkout.
4. Should see `Create an account` checkbox. 
5. Check checkbox and complete checkout process (complete email and other essential details).
6. Confirm that an account was created correctly and is associated with the order.

**Note** to fully test the end to end experience you will need to access emails sent to your new customer; e.g. set up MailHog or some other solution giving you full access to emails sent from your site.

#### Test other scenarios
Repeat similar tests similar to above:

- With `Allow shopper to sign up` block setting enabled/disabled.
- With different Woo accounts options (e.g. guest checkout disabled) and confirm account is created when appropriate.
- With various account creation errors, e.g. use an email address already "owned" by an existing account.
- Ensure `Create an account` checkbox is only shown when appropriate.
- Ensure account is only created if appropriate.

#### Test customisation / theming / hooks / extensibility
- Test with a theme that overrides email templates. Is it possible to theme this new email too? 
- Do relevant existing hooks function as expected with this new email?

#### Run PHPunit tests
There's a [new test class including some PHPunit tests of `CreateAccount` service class](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fix/2377-create-account-on-checkout/tests/php/Domain/Services/CreateAccount.php). These are WordPress-style integration tests, not true unit tests. By that I mean that they set up WP / Woo state and then run our service and check the appropriate state changes in the site (e.g. user account is created).

You can run these tests locally: `npm run phpunit -- ./tests/php/Domain/Services/CreateAccount.php`

### How this is implemented
The goal here was to retain as much of the existing behaviour and customisation hooks as possible, while still allowing us to replace the email. Long term, this will need to merge into core - and we need a plan for how we'll do that.

#### New `CreateAccount` service
There is a new `CreateAccount` service , triggered from the checkout REST API via a new action hook `woocommerceblocks_create_account_for_order_request`, passed the order and request as parameters.

The main entry point is `from_order_request()` which first determines if an account should be created. If so, this calls `create_customer_account()` method, which serves the same purpose as `wc_create_new_customer()` from core. 

The intention is that this fires all the same hooks as woo core and works the same, apart from the email notification.

When triggering the `woocommerce_created_customer` hook, an extra flag `is_checkout_block_customer_signup` is added to the user data to signal to the email handler that this is a new, block-based signup (and we want new improved behaviour).

#### Fork of `New Account` email (from WooCommerce Core)
`CreateAccount` also hooks various WooCommerce core behaviour to effectively replace the "new account" email sent to the shopper after signup.

The reason for this is so the email can include a password reset link; the core email includes a generated password in the email body, which is not ideal.

The core email handler (`woocommerce_created_customer_notification` hook) is first removed, and then a custom handler is registered. In this handler, there's logic to detect `is_checkout_block_customer_signup` - if it's present, we trigger our new email, otherwise we manually invoke the core email. 

To implement the email template, there's essentially a fork of `WC_Email_Customer_New_Account` and related templates. This removes the password render code and replaces it with a magic link to set an initial password. 

The class also forces a default path (into blocks plugin) when loading templates, to ensure our custom templates are used. I _think_ this should allow themes to override the templates similar to overriding WC core templates (let's test that!).

### Changelog

I've left this changelog here for now. 

> Support creating a new user account as part of checkout flow (in Checkout block) (Feature gated to dev builds). #2851


haszari:
In progress on this - @Aljullu @nerrad if you have a moment to give this a quick review, I'd appreciate it.

I'm adding a new provider/context to store the create account user input, and will pass it to API via `CheckoutProcessor::processCheckout`. And then .. implement create account handling in the API backend in `Routes::Checkout::get_route_post_response`. Will also need appropriate error handling, and tidy up the password input - make it consistent with other inputs & implement "reveal password" (eye icon) consistent with other places in core.

Keen for any advice, in case I'm on the wrong track or there's more to consider!

Aljullu:
@haszari I took a quick look and it looks promising! The log in flow is completely new to me, so I'm not sure I can be of much help with this, but looking at the code everything made sense to me. I saw some styling issues, but I think they will be easy to fix (basically some margins that need to be adjusted and the [TextInput](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/components/text-input/style.scss#L39) component needs styles for `type=password`).

> implement "reveal password" (eye icon) consistent with other places in core

I think this can be left for a follow-up. It would be a nice enhancement but I don't consider this to be a blocker for this PR.

src/StoreApi/Routes/Checkout.php
```diff
@@ -164,6 +164,31 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
...
+				$password,
+				array(
+					'first_name' => $order_object->get_billing_first_name(),
+					'last_name'  => $order_object->get_billing_last_name(),
+				)
+			);
+
+			if ( is_wp_error( $customer_id ) ) {
+				return $this->get_route_error_response(
+					400,
```
haszari, line 183: I see we have a few deliberate `500` errors triggered in here – is that by design? I would expect 500 errors only in exceptional, non-expected circumstances, and all other errors should use a value in the 4xx range. I've gone with 400 here for now, but I'm keen to understand our approach for API error codes in blocks.

haszari:
Still a work-in-progress, but I'm keen for early feedback on all aspects!

@mikejolley - Any thoughts/guidance on the API changes and where that code should live? Lots of polish needed here. Also keen to explore sharing the "create account" logic with woo core or move to a controller class maybe.

@nerrad - I'm using the big checkout context to store the `createAccount` boolean. Keen for feedback on this or ideas for alternatives.

Anyone - General testing of the interaction between the various checkout signup options – I think there's a few combinations I need to support.

What's left to do?

- Proper error handling in the API, especially for "email already in use" case. There may also be missing hooks/woo integration for the create account flow that we should consider.
- Handle all combinations of `CHECKOUT_ALLOWS_GUEST` (require account) `CHECKOUT_ALLOWS_SIGNUP` (etc .. maybe other relevant options / filters).

src/StoreApi/Routes/Checkout.php
```diff
@@ -164,6 +164,32 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
 		// Check order is still valid.
 		$order_controller->validate_order_before_payment( $order_object );
 
+		// Create a new user account as necessary.
+		if ( ! is_user_logged_in() && ( WC()->checkout()->is_registration_required() || ! empty( $request['create_account'] ) ) ) {
```
mikejolley, line 168: I'm wondering if we should avoid using properties from the core checkout class due to it being so deeply coupled to the shortcode and checkout page. We could look this up from woo core settings so this API is not dependent on core checkout.

src/StoreApi/Routes/Checkout.php
```diff
@@ -164,6 +164,32 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
...
+			$username    = '';
```
mikejolley, line 169: Why are these variables needed?

assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -239,6 +242,19 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 								onChange={ setEmail }
 								required={ true }
 							/>
+							{ CHECKOUT_ALLOWS_SIGNUP && (
```
mikejolley, line 245: Just pointing out this does not cover the use case when signup is a requirement. In which case, no input is needed, but the checkout API should create an account if the requester is logged out.

src/StoreApi/Routes/Checkout.php
mikejolley: Look up standard HTTP error codes. 4xx are client errors. 5xx are server errors. If the request fails, try to derive why from the WP error.

src/StoreApi/Routes/Checkout.php
haszari: Sounds good - will make those changes :)

src/StoreApi/Routes/Checkout.php
haszari: Work in progress - these will be gone, thanks for highlighting!

assets/js/blocks/cart-checkout/checkout/block.js
haszari: Agreed, I still need to handle that case. Thanks for reviewing, I appreciate the early feedback :)

src/StoreApi/Routes/Checkout.php
```diff
@@ -317,6 +330,75 @@ protected function create_or_update_draft_order() {
...
+				'last_name'  => $last_name,
+			)
+		);
+
+		if ( is_wp_error( $customer_id ) ) {
+			switch ( $customer_id->get_error_code() ) {
+				case 'registration-error-email-exists':
+					throw new RouteException(
+						'woocommerce_rest_checkout_create_account_failure',
+						__( 'An account is already registered with your email address. Please log in.', 'woo-gutenberg-products-block' ),
```
haszari, line 386: I'd love to show a login link when this error happens. I don't think we currently support this in the [notices system](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/docs/block-client-apis/notices.md), so if we were to implement that it would need to be follow up. 

Keen for others thoughts on this, doesn't feel like an urgent priority, but could be a nice enhancement to the notices system and make the checkout flow smoother for that case :)

haszari: Looks like the woo core version of this error message has a link handled by javascript: https://github.com/woocommerce/woocommerce/blob/e736ac4145d9aceb3574922d01824ede546a07f3/includes/wc-user-functions.php#L47

src/StoreApi/Routes/Checkout.php
haszari: Wrapping up this thread: I've [whitelisted specific error codes](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2851/files#diff-e82bdb511fe4e8a787289e47fef92f88R366) returned from `wc_create_new_customer()`, and pass the error message down for the error notice. I'm using `400` for all these errors, because they all represent client errors from the user's input, e.g. using an email that already has an account.

nerrad:
Just noting this has been on my list of pulls to get to but I haven't had a chance to yet. I am aware of specific areas you've requested my feedback on and I will take a look soon, but for now I just want to comment on a few things around the general approach being taken:

I realize there has been some discussion in [the related issue](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2377) about the flow of this but I have some more thoughts on this:

- For _new_ WooCommerce merchants (and likely new users to the blocks), I definitely agree we should default to what is recognized as the best default for account creation (remove friction and just have it happen automatically given known information).
- However for existing merchants that switch to using the blocks, they may already have flows that they want to retain for account creation. We could still have the blocks default to the behaviour as begun here, but I do think we need to recognize the potential disconnect between how they may be used to the settings configuring the account creation flow in the shortcode checkout vs what happens in the checkout block flow.
- With that in mind, I think we should _embrace_ that this is a _new_ account creation flow and separate it completely from the existing settings requirements (with all the rationale we can give for this choice).

This means:

- We don't use existing WC core logic for creating the account, or we submit pulls to WooCommerce core so it's abstracted more so we _can_ reuse pieces (similar to the approach taken for the new reserve stock class/table) which is I think one of the options you suggested Rua. I'd be in favor of that option even though it potentially creates a Woo Core requirement...which we may have to progressively enhance on. We also may be able to creatively use the `get_option` filter to filter the options and then remove the filter after as a workaround until any code submissions to woo core land.
- We leave the existing settings as something that controls account creation behaviour for the _legacy_ flow. 
- We release following the behaviour established in this pull (and the above) with an eye to seeing what folks come back with regarding how they want to customize the new account creation flow and make decisions based on that feedback. I would argue that any changes should involve extensibility points to make it easier for folks to customize, but that the default behaviour should _not_ be controlled via options.

Regarding creation of password. I do _not_ think the created password should be included in any outgoing email. That has for a while now been established as a bad practice and I think we should use this opportunity to move away from that. The common pattern now is to include a link in the email that goes back to the site for creating a password to use.

Another thing to keep in mind is whether or not the shopper is automatically logged in after completing their order. But that might be something that could be controlled server side (and not something we have to worry about for now).

(cc @LevinMedia for awareness)

src/StoreApi/Routes/Checkout.php
haszari: Related: #2854

haszari:
@nerrad I'm pretty much aligned 100% with your thoughts on the "legacy" create account options and how they interact with new checkout block. 

One thing I'd like to clarify before we dig further - does this issue affect (block) our plans for launching checkout block? I think the answer to this is nuanced (i.e. it depends!). To me this is a platform issue that is affecting our checkout block, so I think we need wider discussion & participation to decide how we address it in WooCommerce core - cc @peterfabian for awareness for now.

Having said that I think you've covered the salient points in your analysis @nerrad. I've summarised below as issues (in Woo core vs blocks codebase) that we could log / fix as part of a solution - I think these include most of the ideas in the discussion so far.

##### WooCommerce Core
1. Don't (never) show a user's password in email; always send a password reset link instead. This will help our flow and is a general security enhancement.
2. Tweak wording/UX of existing options to clarify that they are legacy and only affect shortcode (not blocks). 
3. Implement a new create account routine/API better suited to our purposes (and potentially refactor the existing function to use the new implementation). This might not include various action/filter hook points as the current routine, i.e. deliberately limit extensibility for now.

Of these, I think (1) is the least controversial, and also helps unblock us. I'd be happy to implement this if there's agreement :)

The other two issues are more of a proposal that needs wider discussion (user experience, developer experience, backward compatibility).

I am in progress on a Woo Core PR which is a "lite" version of (3), will link here when the PR is up ([branch link](https://github.com/woocommerce/woocommerce/tree/fix/signup-auto-username-password-api-options)). 

##### WooCommerce Blocks
5. Don't use the `Allow customers to create an account during checkout` option for checkout block. Instead, add a block setting, and clarify/document that the Woo option is a legacy option, affecting shortcode checkout only (see Woo core item 2 above).
7. Hook `get_option` to prevent error notices due to checkout block not supporting user-supplied username/password.

I can implement both of these in this PR, so will get started on that. If they work out, then this could help unblock the checkout block; if not, we have more information about the best solution :)

haszari:
Progress update on this PR:

- Added an [editor option for `Create account` checkbox in checkout block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2851/files#diff-d7dac78f12e1a99f2210b015cac8d268R159). 
- Replaced the create account logic with a [new function based on core `register_new_user`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2851/files#diff-e82bdb511fe4e8a787289e47fef92f88R419).

This working end-to-end now but is not shippable yet, there are a few things to work through (either in this PR or follow up). I _think_ these are all somewhat related - things we get "for free" by using the Woo create account API, that we need to either hook up or reimplement.

#### No store branding/site theme on welcome email and set initial password screen
We're using a fairly low-level API - `register_new_user()`. This generates a plain-text email, and links to a wp-admin form for setting the account password:

Screen Shot 2020-08-05 at 3 59 59 PM

Screen Shot 2020-08-05 at 4 00 54 PM

It might be possible to filter this email to theme it - `wp_new_user_notification_email`. 

Similarly, we might be able to hook the render of the `wp-admin` form to theme it, and/or override the URL, and implement a new "set initial password" screen using similar template approach as other Woo account screens (to reduce the burden on themes adding support). 

Need to investigate more, but I'm thinking we should handle these two items as follow up issues.

#### Reinstate essential Woo extensibility/integration hooks as needed
The new API (`Checkout::create_customer_account()`) doesn't fire all the same hooks as the existing woo core API ([`wc_create_new_customer `](https://github.com/woocommerce/woocommerce/blob/e736ac4145d9aceb3574922d01824ede546a07f3/includes/wc-user-functions.php#L41)). 

I'm not certain whether these hooks should fire in the new API - most of them appear to be legacy "render extra stuff" or "handle the extra fields you rendered" hooks. 

In my testing so far I've found one issue that's related to this - the new user has role `Subscriber` (should be `Customer`). This is a blocker for this PR IMO, so I'm going to look at that next. 

Screen Shot 2020-08-05 at 4 10 51 PM

haszari:
> In my testing so far I've found one issue that's related to this - the new user has role Subscriber (should be Customer). This is a blocker for this PR IMO, so I'm going to look at that next.

This was an easy fix - I've added code to set the role manually. 
2bb344b

In the woo core routine, [this is handled by `woocommerce_new_customer_data` filter](https://github.com/woocommerce/woocommerce/blob/e736ac4145d9aceb3574922d01824ede546a07f3/includes/wc-user-functions.php#L41) - I missed that initially.

src/Assets.php
```diff
@@ -182,8 +182,8 @@ public static function get_wc_block_data( $settings ) {
 					'privacy'  => self::format_page_resource( $page_ids['privacy'] ),
 					'terms'    => self::format_page_resource( $page_ids['terms'] ),
 				],
-				'checkoutAllowsGuest'           => 'yes' === get_option( 'woocommerce_enable_guest_checkout' ),
-				'checkoutAllowsSignup'          => 'yes' === get_option( 'woocommerce_enable_signup_and_login_from_checkout' ),
+				'checkoutAllowsGuest'           => filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN ),
+				'checkoutAllowsSignup'          => filter_var( get_option( 'woocommerce_enable_signup_and_login_from_checkout' ), FILTER_VALIDATE_BOOLEAN ),
```
haszari, line 186: `checkoutAllowsSignup` (store setting) is no longer needed by the block - could remove this, or use it for the default value of the block option.

haszari: Now using the store setting for the default value of the block option. 5a3c1f99c16566dc913ee440e2131b068e0f2e13

haszari:
This is ready for review - pinging @Aljullu (DRI for cart/checkout wrap) and @nerrad (because of the new API, and related considerations re email/theming/extensibility).

src/StoreApi/Routes/Checkout.php
```diff
@@ -320,6 +333,121 @@ protected function create_or_update_draft_order() {
...
+			// WordPress core error codes.
+			case 'empty_username':
+			case 'invalid_username':
+			case 'empty_email':
+			case 'invalid_email':
+			case 'email_exists':
+			case 'registerfail':
+				return new RouteException(
+					'woocommerce_rest_checkout_create_account_failure',
+					wp_strip_all_tags( $error->get_error_message() ),
```
haszari, line 380: We're stripping all HTML so any error messages are plain text, for our notice API. 

This was originally added because [WooCommerce Core generates errors with log in links](https://github.com/woocommerce/woocommerce/blob/e736ac4145d9aceb3574922d01824ede546a07f3/includes/wc-user-functions.php#L47) and we don't have a way to render those safely in blocks `Notice` system (yet). 

This strip might not be needed (I haven't checked all the core errors), but it's safer to leave it in place.

Regarding the `Please log in.` link - this would be a nice improvement to a specific error case, for example:

- Shopper is already signed up, but not logged in. (Maybe they forgot they had an account.)
- Shopper completes a purchase and requests an account via checkout.
- Shopper sees error notice "You already have an account - please log in".

In Gutenberg [`Notice` there is support for an `notice actions`](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/notice), which we could use to add `Log in` CTA/button/link to notices. We'd need to support it in our Notice API, so I've left this as follow up for now - keen for feedback. (I can make an issue if we think this is worth pursuing cc @nerrad.)

assets/js/blocks/cart-checkout/checkout/edit.js
```diff
@@ -149,6 +150,25 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
...
+						'woo-gutenberg-products-block'
+					) }
+					checked={ allowCreateAccount }
+					onChange={ () =>
+						setAttributes( {
+							allowCreateAccount: ! allowCreateAccount,
+						} )
+					}
+				/>
+			</PanelBody>
```
Aljullu, line 171: Not an issue with this specific PR, but I found it quite confusing that if you have `Allow customers to place orders without an account` unchecked in the settings, this toggle doesn't produce any visual difference in the editor and in the frontend it changes whether the form is visible to guests or not.

That's seems quite counter-intuitive with the text associated with the toggle (`Allow shopper to sign up for a user account during checkout`). I find this very confusing in WC Core as well, but wondering if we should take the chance to try to make the text a bit more clear.

LevinMedia:
@haszari was testing this out today - when clicking the "already have an account: log in" link, can we route the user through the /my-account screen, rather than the main wp-admin login? 

Screen Shot 2020-08-06 at 8 34 27 AM

I attempted an incognito checkout using a new email address and the create an account checkbox, but I never got the follow up email to reset / create the password. Could you take a look at that with me when you're online next?

haszari:
@LevinMedia thanks for testing!

> when clicking the "already have an account: log in" link, can we route the user through the /my-account screen, rather than the main wp-admin login?

This is a separate issue, reported here - #2918 . Totally agree this could be improved. Currently in `Backlog` pipeline.

> I attempted an incognito checkout using a new email address and the create an account checkbox, but I never got the follow up email to reset / create the password. Could you take a look at that with me when you're online next?

Email is tricky on dev sites! Before we go further, can you confirm that you get other emails from your site (e.g. merchant "new order" email or use regular register flow)? In my dev setup I had to set up Mailhog to get access to emails sent by Woo/WordPress (I might post some docs about this setup); on other setups the mail might be getting directed to spam (e.g. on Atomic Ephemeral platform, if site email is `@automattic.com`).

src/StoreApi/Routes/Checkout.php
haszari: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2985

assets/js/base/context/cart-checkout/checkout-state/actions.js
```diff
@@ -17,6 +17,7 @@ const {
 	INCREMENT_CALCULATING,
 	DECREMENT_CALCULATING,
 	SET_ORDER_ID,
+	SET_CREATE_ACCOUNT,
 	SET_ORDER_NOTES,
```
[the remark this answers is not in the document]
haszari: Going with `shouldCreateAccount` - this is a tricky one to name. It's `userAccountRequested` or `shouldCreateCustomerAccountAfterCheckout`. It's not a typical "is" boolean; the value of the boolean is whether the user requested an account. 

Note that this should not be 100% consistent with the name of the block attribute, as they mean different things.

- Block attribute - `allowCreateAccount` - whether the shopper should be given the option to create an account, i.e. should there be a checkbox rendered.
- API / reducer state - `shouldCreateAccount` - whether the shopper requested a new account.

haszari:
@nerrad @Aljullu - This is ready for at least an initial review. I haven't fully tested yet, and there's various todos and code tidies, but I would appreciate an early code review to confirm the approach and start thinking about implications (for themers, translation, core merge).

senadir:
sidenote: is shopper the term we want to use publicly? isn't an internal term only and we should use customer.

src/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,233 @@
...
+				// Remove core "new account" handler; we are going to replace it.
+				remove_action( 'woocommerce_created_customer_notification', array( $wc_emails_instance, 'customer_new_account' ), 10, 3 );
+
+				// Add custom "new account" handler.
+				add_action(
+					'woocommerce_created_customer_notification',
+					function( $customer_id, $new_customer_data = array(), $password_generated = false ) use ( $wc_emails_instance ) {
+						// If this is a block-based signup, send a new email
+						// with password reset link (no password in email).
+						if ( isset( $new_customer_data['is_checkout_block_customer_signup'] ) ) {
```
nerrad, line 52: I don't see this key (`is_checkout_block_customer_signup`) being implemented anywhere?

src/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,233 @@
...
+		if ( is_user_logged_in() ) {
+			return false;
+		}
+
+		$checkout_requires_account = false === filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN );
+		if ( $checkout_requires_account ) {
+			return true;
+		}
+
+		$user_requested_account = ! empty( $request['should_create_account'] ) && true === filter_var( $request['should_create_account'], FILTER_VALIDATE_BOOLEAN );
```
nerrad, line 133: You could probably simplify by doing something like this:

```php
protected function should_create_customer_account( $request ) {
	if ( is_user_logged_in() || empty( $request['should_create_account'] ) ) {
		return false;
	}

	if ( 
		false === filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN ) ||
		true === filter_var( $request['should_create_account'], FILTER_VALIDATE_BOOLEAN )
	) {
		return true;
	}

	return false;
}
```

src/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,233 @@
...
+
+	/**
+	 * Check request options and store (shop) config to determine if a user account
+	 * should be created as part of order processing.
+	 *
+	 * @param \WP_REST_Request $request The current request object being handled.
+	 *
+	 * @return boolean True if a new user account should be created.
+	 */
+	protected function should_create_customer_account( $request ) {
```
nerrad, line 123: You should typehint for `WP_REST_Request` here.

src/Domain/class-wc-email-customer-new-account-blocks.php
```diff
@@ -0,0 +1,190 @@
+<?php
+/**
+ * Class WC_Email_Customer_New_Account_Blocks file.
+ *
+ * @package  WooCommerce/Blocks
+ */
+
+use WC_Email;
```
nerrad, line 8: convention is to include `\WC_Email` in imports (root namespace)

nerrad: Some feedback on this class:

- Don't use the old loading process used in Core. Include proper namespacing, and CamelCase for the class name. For now, let's include this `src/Domain/Services/Email` - so the namespace would be `Automattic\WooCommerce\Blocks\Domain\Services\Email` and the fully qualified class name would be `Automattic\WooCommerce\Blocks\Domain\Services\Email\CustomerNewAccount`.
- Instead of the _file_ instantiating the class (when included), you can import this class in `CreateAccount` and then just instantiate it when assigning to the `new_account_email` property.
- With the above you can also drop the `class_exists` check.

src/Domain/class-wc-email-customer-new-account-blocks.php
```diff
@@ -0,0 +1,190 @@
...
+			// Note - we're using the same ID as the real email.
+			// This ensures that any merchant tweaks (Settings > Emails)
+			// apply to this email (consistent with the core email).
+			$this->id                    = 'customer_new_account';
+			$this->customer_email        = true;
+			$this->title                 = __( 'New account', 'woo-gutenberg-products-block' );
+			$this->description           = __( 'Customer "new account" emails are sent to the customer when a customer signs up via checkout or account blocks.', 'woo-gutenberg-products-block' );
+			$this->template_html         = 'emails/customer-new-account-blocks.php';
+			$this->template_plain        = 'emails/plain/customer-new-account-blocks.php';
+			$this->default_template_path = plugin_dir_path( __FILE__ ) . 'templates/';
```
nerrad, line 69: Let's chat about this, I think we'll need to use some dependency injection here but there's bunch of ways of doing it so we should hash it out a bit before committing. Ideally ALL paths will come from `Automattic\WooCommerce\Blocks\Domain\Package`.  The `templates` folder should also be at `src/templates` I think (or even `/templates`).

src/Domain/class-wc-email-customer-new-account-blocks.php
```diff
@@ -0,0 +1,190 @@
...
+		 * @return string
+		 */
+		public function get_content_html() {
+			return wc_get_template_html(
+				$this->template_html,
+				array(
+					'email_heading'      => $this->get_heading(),
+					'additional_content' => $this->get_additional_content(),
+					'user_login'         => $this->user_login,
+					'user_pass'          => $this->user_pass,
```
nerrad, line 141: Should we exclude `user_pass` from template args here because that's not provided (same applies to `get_content_plain`)?

src/Domain/templates/emails/customer-new-account-blocks.php
```diff
@@ -0,0 +1,32 @@
...
+
+do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
+
+<?php /* translators: %s: Customer username */ ?>
+<p><?php printf( esc_html__( 'Hello %s,', 'woo-gutenberg-products-block' ), esc_html( $user_login ) ); ?></p>
+<?php /* translators: %1$s: Site title, %2$s: Username, %3$s: My account link */ ?>
+<p><?php printf( esc_html__( 'Thanks for creating an account on %1$s. Your username is %2$s. You can access your account area to view orders, change your password, and more at: %3$s', 'woo-gutenberg-products-block' ), esc_html( $blogname ), '<strong>' . esc_html( $user_login ) . '</strong>', make_clickable( esc_url( wc_get_page_permalink( 'myaccount' ) ) ) ); // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped ?></p>
+<?php if ( $set_password_url ) : ?>
+	<p><a href="<?php echo esc_attr( $set_password_url ); ?>"><?php printf( esc_html__( 'Click here to set your new password.', 'woo-gutenberg-products-block' ) ); ?></a></p>
+<?php endif; ?>
```
nerrad, line 22: I realize this is just the first iteration, but I think we'll want to improve on the wording a bit. It's entirely possible the user won't be able to access their account if the login session has expired. So we probably shouldn't direct them to the account page in this new initial email but instead just direct them to set their new password (same applies to the plain email template).

src/Domain/Services/CreateAccount.php
haszari: It's set in Checkout API - https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2851/files#diff-fcddc6e9639fa3e5ae3508597cb81054R210

haszari: (if it wasn't, I would have discovered during testing!)

src/Domain/templates/emails/customer-new-account-blocks.php
haszari: Agree - this really needs design. I'm currently using a hybrid of the existing woo new account templates and the wording from core for the set initial password section (i.e. combo of all prior art).

Note that the generic "my account" link should be useful since some users may use the email as a way to get back to their account; the reset password link will expire, but my account will always show login form, reset password options so user can get back in. I'm guessing this is part of the reason it's included in these emails.

src/Domain/Services/CreateAccount.php
nerrad: doh, I missed that!

src/Domain/Services/CreateAccount.php
haszari: There's some complexity to this - depending on store options, we may make an account irrespective of what user requests. I'll revamp so it's clear and simplify as much as possible 😄

haszari: Updated and commented this here: e95bf7f9b86ab8033ee2a8174e8fb3175ea09087

src/Domain/class-wc-email-customer-new-account-blocks.php
haszari: Good spotting! 4219194f8294b9fe1b10d25728b7a88962bbcbdc

src/Domain/class-wc-email-customer-new-account-blocks.php
haszari: ✨ 41824a1e268fd8599b0c90c696d58da56be2fe37

src/Domain/class-wc-email-customer-new-account-blocks.php
haszari: Took a stab at this - 648a572a236ad8801b7e9088e9f07f4031757fe0 

Hopefully a good start, there's a tricky dependency on `WC_Email` from woo core. `Package::get_path` was definitely what I was looking for 🏞️

src/StoreApi/Routes/Checkout.php
```diff
@@ -168,16 +169,11 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
...
-			);
-
-			// Log the customer in and associate with the order.
-			wc_set_customer_auth_cookie( $customer_id );
-			$order_object->set_customer_id( get_current_user_id() );
+		try {
+			// TODO full docs for new hook, including throw.
+			do_action( 'woocommerceblocks_create_account_for_order_request', $order_object, $request );
+		} catch ( Exception $error ) {
+			$this->handle_create_account_error( $error );
```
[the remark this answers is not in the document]
haszari: Now calling `CreateAccount` directly, no hook: ad3c19d34bf4976cd1694e54414d43107423978b

haszari:
Thanks for the review @nerrad - I think I've addressed all the feedback, except the unit test. Keen for re-review to confirm how I've handled it and any follow ups. Also cc @Aljullu  - you had early review feedback which may have been lost. Once the dust settles on this approach I'll circle back to make sure we've addressed all review feedback.

> Also, let's add some phpunit tests for the new service class please.

Agreed, though I'm ignoring unit tests (for today at least) until this has had more testing, and team review to confirm this direction in general. Also maybe review the wider flow from design point of view; I think that's important before merging this, even if we're ok with the tech/architecture POV.

haszari:
> sidenote: is shopper the term we want to use publicly? isn't an internal term only and we should use customer.

Good question @senadir . I don't think there's anything particularly internal about shopper, though it's helpful for us specifically because merchants are customers of woo :)

I agree though, this wording (e.g. the block option) should be reviewed from a design point of view, and we should use consistent terms for roles/personas. So customer might be a better fit here.

src/Domain/Services/CreateAccount.php
```diff
@@ -120,18 +133,26 @@ public function from_order_request( \WC_Order $order, \WP_REST_Request $request
...
-		$checkout_requires_account = false === filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN );
-		if ( $checkout_requires_account ) {
+		// From here we know that the shopper is not logged in.
+
+		if ( false === filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN ) ) {
+			// Store requires an account for all checkouts (purchases).
+			// Create an account independent of shopper option in $request.
+			// Note - checkbox is not displayed to shopper in this case.
 			return true;
 		}
```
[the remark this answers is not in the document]
nerrad: Incidentally, this is why I think there's value in writing phpunit tests for this service class. Tests help describe expectations for the things you are testing and also help catch changes in those expectations when refactoring happens over the course of a review. I feel we've established a baseline of the approach we're taking where there's not much risk of effort spent on tests being wasted.

src/Domain/Services/Email/CustomerNewAccout.php
```diff
@@ -0,0 +1,188 @@
...
+
+namespace Automattic\WooCommerce\Blocks\Domain\Services\Email;
+
+use \WP_User;
+use \WC_Email;
+use Automattic\WooCommerce\Blocks\Domain\Package;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
```
nerrad, line 16: Do we actually need this? The file doesn't execute anything if loaded directly and at most will just fatal error because `\WC_Email` won't be available? I know this is a common pattern in WP but I've always been puzzled by it's blanket use.

src/Domain/Services/Email/CustomerNewAccout.php
```diff
@@ -0,0 +1,188 @@
...
+		// Note - we're using the same ID as the real email.
+		// This ensures that any merchant tweaks (Settings > Emails)
+		// apply to this email (consistent with the core email).
+		$this->id                    = 'customer_new_account';
+		$this->customer_email        = true;
+		$this->title                 = __( 'New account', 'woo-gutenberg-products-block' );
+		$this->description           = __( 'Customer "new account" emails are sent to the customer when a customer signs up via checkout or account blocks.', 'woo-gutenberg-products-block' );
+		$this->template_html         = 'emails/customer-new-account-blocks.php';
+		$this->template_plain        = 'emails/plain/customer-new-account-blocks.php';
+		$this->default_template_path = $package->get_path( '/templates/' );
```
nerrad, line 73: Noice 😄

src/StoreApi/Routes/Checkout.php
```diff
@@ -170,8 +172,9 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
 
 		// Create a new user account as necessary.
 		try {
-			// TODO full docs for new hook, including throw.
-			do_action( 'woocommerceblocks_create_account_for_order_request', $order_object, $request );
+			$create_account = Package::container()->get( CreateAccount::class );
+			$create_account->from_order_request( $order_object, $request );
+
```
nerrad, line 177: 👍

src/StoreApi/Routes/Checkout.php
```diff
@@ -170,8 +172,9 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
...
 		} catch ( Exception $error ) {
 			$this->handle_create_account_error( $error );
```
nerrad, line 179: Rather than being _specifically_ for handling create account errors, I wonder if this should just be `handle_error`? That way it's not tightly coupled to errors related to create account (and feasibly other error codes could be handled by the method in the future).

LevinMedia:
@haszari this is looking great! Generally all the wording / looks good to me. The only thing I noticed was the checkbox for create an account was smashed up against the email field: 

Screen Shot 2020-08-13 at 11 34 20 AM

src/Domain/Services/Email/CustomerNewAccout.php
haszari: Good point! This is a remnant from paste-adapt of woo core code. Fixed here: f5cd2d5f6ecaf08e77882ab9e0b2d16df50df67c

src/StoreApi/Routes/Checkout.php
haszari: I can work with that, made the change here: 65a72403f7dda3b9eb84494f3abf545a179288db

Incidentally, I went with the specific name as it's private – can easily rename later when generalising - and to more clearly indicate the intention (since it's not intended to handle any error right now).

src/Domain/Services/CreateAccount.php
```diff
@@ -19,26 +21,37 @@ class CreateAccount {
...
 		// Override core email handlers to add our new improved "new account" email.
 		add_action(
 			'woocommerce_email',
 			function ( $wc_emails_instance ) {
-				$this->new_account_email = require_once dirname( __DIR__ ) . '/class-wc-email-customer-new-account-blocks.php';
+				// Late instantiate CustomerNewAccount because it depends on WC_Email,
+				// which is not available at `init()` time.
+				if ( ! $this->new_account_email ) {
+					$this->new_account_email = new CustomerNewAccount( $this->package );
+				}
```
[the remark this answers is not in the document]
haszari: Good point! Maybe this was more relevant in a previous iteration ... ? Made this change here: 89042da92decd0edbba6d836b79e7fa296987301

In some previous iteration, I hit `WC_Email undefined` errors because this was initialised too early, but I don't recall whether I actually needed it in a member 🤔

src/Domain/Services/CreateAccount.php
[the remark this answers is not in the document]
haszari: I agree - I'll get some test coverage in place.

Regarding the relationship to guest account option – I haven't considered not implementing that. It might be fine/possible to simplify it, so I'll give this a try. 

If we shipped it like this it would be different behaviour from existing checkout; though perhaps that's a useful improvement (as you suggest). We might need to document this and/or tweak the UX to avoid surprises/bug reports. FYI @LevinMedia for awareness, I'll follow up when I've investigated further (and/or prototyped it).

Pureandwhole:
Good day,
I see @nerrad removed this fix from the 3.4.0 milestone.
Is there anyone that can perhaps give us an rough estimate on when we can expect this fix to be pushed to main?
Thank you

nerrad:
Hi @PureandWholeStore, 

It was removed from 3.4.0 because there's still a few questions around how we're going to roll this out. Once those are resolved the pull will be merged and we'll assign it a milestone.

PureandWholeStore:
Hi @nerrad 
Thank you for your reply.

I know this is perhaps not the place to ask this question but ive searched far and wide and im still unable to properly install the dev version of **woocommerce-gutenberg-products-block/fix/2377-create-account-on-checkout**

Ive tried to upload the the files manually to the website as well as to pull it via **WP Pusher** but still to no avail.

Im still getting a **composer install** error:

> **Your installation of the WooCommerce Blocks feature plugin is incomplete. Please run composer install within the wp-content/plugins/woo-gutenberg-products-block directory.**

How do i do this?

Could you please direct me to a page explaining how to properly install a dev version of a plugin?

> Installing the development version
> 
> 1.     Make sure you have WordPress 5.3+ and WooCommerce 4.0+
> 2.     Get a copy of this plugin using the green "Clone or download" button on the right.
> 3.     **npm install to install the dependencies.**
> 4.     composer install to install core dependencies.
> 5.     npm run build (build once) or npm start (keep watching for changes) to compile the code.
> 6.     Activate the plugin.
> 
> The source code is in the assets/ folder and the compiled code is built into build/.

From where should **npm install to install the dependencies.** be run?
We are on shared hosting with cpanel

Thank you very much

nerrad:
Hi @PureandWholeStore, 

The instructions for developing WooCommerce blocks and contributing to the codebase are intended for a target audience of developers familiar with node, npm and composer usage, not end users of the plugin. Providing support for learning those technologies is outside the scope of what we can provide here.  You could try googling for things like ["beginner's guide to using npm"](https://www.google.com/search?rlz=1C5CHFA_enCA859CA859&sxsrf=ALeKk02n8bKRrA1_kQTumNBwx3A3o9AOsg%3A1600277744228&ei=8ExiX5SnDdqC9PwP2r-TgAw&q=beginners+guide+to+npm+install&oq=beginners+guide+to+npm+install&gs_lcp=CgZwc3ktYWIQAzIICCEQFhAdEB46BAgAEEc6BAgjECc6BQgAEJECOgUIABCxAzoICAAQsQMQgwE6BAgAEEM6BwgAELEDEEM6CggAELEDEIMBEEM6AggAOgQIABAKOgwIABCxAxBDEEYQ-QE6BggAEBYQHlCnDliFNGCzNmgCcAF4AIABowGIAesYkgEFMjEuMTGYAQCgAQGqAQdnd3Mtd2l6yAEIwAEB&sclient=psy-ab&ved=0ahUKEwiU3_GTm-7rAhVaAZ0JHdrfBMAQ4dUDCA0&uact=5).

haszari:
Hi @PureandWholeStore  - thanks for your interest in this feature! 

Just a reminder, it's not recommended to run development versions of this plugin (such as this branch) on your production store.

haszari:
Updating this PR after discussion elsewhere. This is the plan:

1. Wrap up & release this PR in Woo blocks plugin:
  - Fix the issue with the non-branded `Set password` screen. We'll use the existing [`Lost password` template as a model](https://github.com/woocommerce/woocommerce/blob/master/templates/myaccount/form-lost-password.php). The URL and title will be updated as appropriate (e.g. `Set password).
  - Implement unit tests for new service classes.
  - Review, test & merge the PR.
  - Release in WooCommerce Blocks plugin (version tbc).
2. Implement the new improved create account flow in WooCommerce core. This will be a parallel implementation, using some of the same code.
  - Open a PR implementing create account with set initial password flow.
  - Use this as appropriate for existing create account store settings, and phase out/remove flows that send a password in email. 
  - Note: the existing create account settings (generate password, generate username) will not be changed; for now we'll keep these but their behaviour may change. E.g. generate password will require user to set a password via flow (no password sent in email!).
  - Review, test and merge PR,
  - Release in WooCommerce core (version tbc).

This allows us to use this new flow for the Cart & Checkout blocks in the feature plugin soon – as soon as step 1 is complete.

Long term, after step 2 is complete and available in min supported version of Woo core (L-2 policy), we can remove the implementation in WooCommerce Blocks.

haszari:
Update - pushed a commit using [`lost-password` form](https://github.com/woocommerce/woocommerce/blob/master/templates/myaccount/form-lost-password.php) for resetting the password. 

This functions, but has pros & cons:

- Pros
  - Uses Woo / store theming, look & feel.
  - Allows customer to set a password.
- Cons
  - The title is `Lost password` - we need to hook/override to `Set password`.
  - The form doesn't automatically log the user in.

I'm still investigating. It may be possible to address those issues and still use Woo core functionality as is (or with some hooks/overrides). Or worst case we may need to fork and add a new my-account form / screen based on this with the features we need (page title + auto login).

haszari:
The `lost-password` form is a registered my-account endpoint, [set up in `WC_Query`](https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-query.php#L80).

Our goal is to show a set password form with an appropriate url (`set-password`) & title if the user is coming from the new block-based flow, and leave the existing lost-password form intact for the lost password flow.

I think what we need to do is either override it the endpoint when appropriate - i.e. checkout block signup flow only. OR add a new endpoint for our needs. Investigating this; adding a new endpoint seems like the safest approach, though could be tricky from a plugin.

haszari:
Update – I think the cleanest path to adding a new `my-account/set-password` handler is to add it in core directly.

Here's what I've tried/discovered, still investigating.

- There's a [documented mechanism](https://github.com/woocommerce/woocommerce/wiki/Customising-account-page-tabs#manipulating-the-menu) for adding new `my-account` endpoints, similar to `downloads`, `orders` etc. TLDR: add a rewrite rule, add new item to my-account menu, render content in a hook.
- The main use case here is for adding new UI/widgets in the logged-in account screen. 
- For the new `set-password` handler, this is not a good fit:
  - We don't want it to be included in the my-account sidebar/menu.
  - It should only be shown for when using a reset password token (aka logged out).

The current password reset UX is implemented as part of the my-account shortcode: see [WC_Shortcode_My_Account::lost_password()](https://github.com/woocommerce/woocommerce/blob/02cf0dfaed5923513de0c88add597d1560c2cfd2/includes/shortcodes/class-wc-shortcode-my-account.php#L228).

- This method fires if the user is logged out, and path/query includes `lost-password`.
- If user has a valid password reset token (in cookie) the reset password form is rendered - user can enter new password.
- Otherwise the lost password form is shown so user can enter email/username to kick off the reset process.

What we really need to do is hook this flow somewhere so we can use the set password form within our flow, without affecting the existing reset password flow.  As far as I can tell there aren't suitable hook points anywhere in the existing flow.

haszari:
Update (decisions made in discussions with @nerrad ). Here's the plan:

- The branded `set-password` screen will be implemented in a separate PR in Woo core (decouple from this PR).
- Wrap up this PR:
   - Roll back this PR to using the WP core (unbranded) login screen. 
   - Gate this checkout signup feature behind experimental flag, so we can merge the PR without shipping this in Woo core or feature plugin. This will allow us to test the flow and iterate in dev builds.
   - Merge 🎉 

Then in future once the core PR merges we can link to the new `set-password` screen and get this shipped in a future version of blocks feature plugin.

assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -180,7 +182,12 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 		window.location.href
 	) }`;
 
-	if ( ! isEditor && ! customerId && ! CHECKOUT_ALLOWS_GUEST ) {
+	if (
+		! isEditor &&
+		! customerId &&
+		! CHECKOUT_ALLOWS_GUEST &&
+		! attributes.allowCreateAccount
+	) {
```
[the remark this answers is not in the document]
haszari: > So what happens here if checkout is configured to not allow guest, the block is configured to allow creations of accounts, and the shopper does not create an account? As is, it appears that checkout will complete just fine.

The behaviour here is based on existing core behaviour. Here's my understanding of it:

- The site is configured that an account is required (`woocommerce_enable_guest_checkout==false`).
- The site is configured that accounts can be created as part of checkout processing (`woocommerce_enable_signup_and_login_from_checkout=true`).
- So when a guest makes a purchase, Woo will always sign user up – they don't have a choice, and aren't presented with the option.

My approach here was to use the existing design as a reference for the new behaviour, but it's possible we'd want to something different in block-based checkout. Ultimately this is a design decision – maybe @nerrad + @LevinMedia discuss & come up with a plan.

haszari:
@nerrad This PR is ready for another review - it's now feature gated to dev builds only. The description has been updated with info about feature gating and the plan for when this will go in feature plugin etc.

Note – we previously discussed adding unit tests for `CreateAccount` service. I haven't done this yet – I'm happy to add one in this PR or as follow up. Let me know in your review 😸

nerrad:
As per [the plan](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2851#issuecomment-694627042) and my earlier comments in review, yes please add the unit tests. History has shown us that we aren't really consistent at adding tests in follow-ups and I'd like to get us in the habit of covering key functionality in tests before merging as much as possible. So imo, not having the tests is a blocker to merge.

Once you've done that I'll review the pull at the earliest opportunity. Thanks!

package.json
```diff
@@ -45,6 +45,7 @@
 		"package-plugin:zip-only": "./bin/build-plugin-zip.sh -z",
 		"package-plugin:deploy": "npm run build:deploy && npm run package-plugin:zip-only",
 		"phpunit": "docker-compose up -d db && docker-compose up -d --build wordpress-unit-tests && docker exec -it --workdir /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block wordpress_test php ./vendor/bin/phpunit",
+		"phpunit:quick": "docker exec -it --workdir /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block wordpress_test php ./vendor/bin/phpunit",
```
haszari, line 48: I added this because it seems wasteful / slow to `docker up --build` every time we run tests. 

This should be moved to a separate PR, and maybe rationalise the PHPunit commands with other test commands, e.g.

- `npm run test:php:up`
- `npm run test:php`

src/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,257 @@
...
+	 *
+	 * @throws \Exception If an error is encountered when creating the user account.
+	 *
+	 * @param string $user_email The email address to use for the new account.
+	 * @param string $first_name The first name to use for the new account.
+	 * @param string $last_name  The last name to use for the new account.
+	 *
+	 * @return int User id if successful
+	 */
+	public function create_customer_account( $user_email, $first_name, $last_name ) {
```
haszari, line 205: I made this public purely so I could test it .. which is not ideal, would rather keep it protected/private. 

Keen for suggestions on how to wrangle this.

assets/js/blocks/cart-checkout/checkout/block.js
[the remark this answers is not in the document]
haszari: Note that I've added PHPunit tests for this behaviour now.

haszari:
@nerrad We now have a good suite of tests for the `CreateAccount` service! 🎉 So this is ready for a final review.

A couple of notes:

- TIL that WordPress PHPUnit tests are often [closer to the "integration test" end of the spectrum](https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/). Realising this really helped me get the tests doing something useful (this logic / service is tightly coupled to WP/Woo state, and that's ok).
- The behaviour with "allow guest account" option is consistent with existing Woo core, and the tests test this behaviour. We could iterate on this but I believe it needs to be considered design / flow point of view before we propose any changes. I'm sticking with the existing logic to reduce the scope of things we're iterating on; in future we may change this as part of other checkout => signup improvements (e.g. signup after purchase completed).

assets/js/blocks/cart-checkout/checkout/block.js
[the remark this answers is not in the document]
haszari: Here's the relevant code in core btw: https://github.com/woocommerce/woocommerce/blob/02cf0dfaed5923513de0c88add597d1560c2cfd2/includes/class-wc-checkout.php#L1006

haszari:
Looks like I have some work to do to get the tests running in Travis x different WP versions (and some other travis issues). Will look at those tomorrow. In the meantime, still keen for reviews / testing / feedback on the PR :)

haszari:
I've investigated the Travis failures. These ones I'm ignoring as they are outside of scope for this PR:

- `6176.5 Javascript/CSS Lint check` - this is failing because of various `useEffect` dependencies in existing code
- `6176.7 E2E Tests (WP 5.4)` - unreliable, intermittently fails (I may try re-running when I get everything else sorted out)
- `6176.9 E2E Tests (WP 5.5 with Gutenberg plugin)` - currently failing on master, explicitly allowed to fail

With the PHPUnit failures, I think that's because the feature is disabled (via feature flag). Running an experiment on that now .. https://travis-ci.com/github/woocommerce/woocommerce-gutenberg-products-block/jobs/392250275

If this is the case, we might need to tweak the PHPUnit jobs so there's one that has experimental flag (`WOOCOMMERCE_BLOCKS_PHASE=3`). 

I noticed with the two existing jobs they are intended to test against different PHP versions – 5.6 and 7.1. Looking at the output it seems they are both running on `Runtime:       PHP 7.4.10`. Maybe the env var is not picked up in the container (or some layer on the way).

haszari:
Success! (ish) https://travis-ci.com/github/woocommerce/woocommerce-gutenberg-products-block/jobs/392250275

The tests pass if I hard-code to enable the feature. So the tests are failing because the experimental feature flag is not set in PHPUnit travis job.

Here's what I'm going to do:

- Put back the feature flag (undo my hack-test).
- Add skip/early return logic to my tests so they succeed if run without experimental flag (`WOOCOMMERCE_BLOCKS_PHASE=3`).

Then that unblocks this PR, with one important consideration: - the `CreateAccount` tests are not run as part of our suite. I'm thinking we can/should handle this as follow up. 

cc @nerrad @senadir - keen for thoughts on the PHPUnit setup and how to ensure these tests get run as part of the suite. Here's an idea:

- Add a travis job that runs our PHPUnit suite with `WOOCOMMERCE_BLOCKS_PHASE=3`; or fix the current jobs if that's the intention.
- Look at the PHP version issue to ensure that the tests are running (in the container) with specified PHP version. (Maybe as follow up.)

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,236 @@
...
+				'should_create_account' => true,
+				'enable_guest_checkout' => true,
+			],
+		);
+	}
+
+    /**
+     * Test exception is thrown if email is invalid or malformed.
+     *
+     * @dataProvider invalid_email_data
```
nerrad, line 177: 👍 nice use of `@dataProvider` here.

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,236 @@
...
+		return [
+			'user_id' => $user_id,
+			'order' => $test_order,
+		];
+	}
+
+    /**
+     * Test successful user signup cases.
+     *
+     * @dataProvider create_customer_data
```
nerrad, line 72: Nice use of `@dataProvider` 👍

nerrad (changes requested):
Thanks Rua for taking care of adding the feature gating and the PHPUnit tests. I haven't tested the functionality again as I've already tested it in previous reviews, so this is just a code review and providing answers to your questions. Assume everything I've left a comment on is something blocking my acceptance unless I've indicated it can be done in a follow-up or is optional (you can still reply back  with some rationale if you disagree with what I've written though).

Besides the inline review comments, I've responded to some of your earlier posts/questions below 👇 

> 6176.5 Javascript/CSS Lint check - this is failing because of various useEffect dependencies in existing code

It's not due to the `useEffect` warnings (which are just _warnings_, they won't cause the linting check to fail). The fail is because of these linting errors:

```
/home/travis/build/woocommerce/woocommerce-gutenberg-products-block/assets/js/base/context/cart-checkout/checkout-state/reducer.js
  63:1  error  Missing JSDoc @param "action.shouldCreateAccount" declaration  jsdoc/require-param
  67:0  error  Missing @param "action.shouldCreateAccount"                    jsdoc/check-param-names
```

> Add a travis job that runs our PHPUnit suite with WOOCOMMERCE_BLOCKS_PHASE=3; or fix the current jobs if that's the intention.

I think all PHPUnit tests should run against the full code base (including experimental), so we should just make sure the constant is defined for the tests. While the environment variable is set for the phpunit tests, I think that's only used by the webpack build process to determine how to build the `blocks.ini` file (which is [read by PHP to set the feature phase](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/bea1d852b5bbca435c4141a4b5bf1f7bedbc73ee/src/Domain/Bootstrap.php#L129-L141)). I don't think we should to add time to the jobs by running `npm run build` so what you could try is adding this to the `script` config block right before the phpunit script run:

```bash
- echo 'woocommerce_blocks_phase = 3' > blocks.ini
```

I'd like to have phpunit tests running and passing before this is merged in travis as it shouldn't be time consuming to get that fixed and removes the need for a (more time consuming imo) follow-up.

> I noticed with the two existing jobs they are intended to test against different PHP versions – 5.6 and 7.1. Looking at the output it seems they are both running on Runtime: PHP 7.4.10. Maybe the env var is not picked up in the container (or some layer on the way).

Ya that looks to be the case, phpunit tests were switched to use docker and I think you're right that the travis env variable is not picking up the version. We _can_ fix this in a follow-up. WooCommerce core bumped the minimum required php version to 7.0 with the release of WC 4.1 and we _currently_ require WC 4.2 so I've had Nadir bump up requirements for the `3.5.0` release of the blocks. In the follow-up we should also consider testing against 7.0 and the _latest_ php version (we are only testing against 7.1 which is old).

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,236 @@
...
+
+		$result = $this->execute_create_customer_from_order(
+    		'maryjones@testperson.net',
+    		'Mary',
+    		'Jones',
+			[
+				'should_create_account' => true,
+				'enable_guest_checkout' => true,
+			],
+		);
```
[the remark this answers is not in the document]
haszari: Zoinks! Fixed here: 7a12383d25e6d207ba7aa8b1bb2cfe500045fa50

I see we exclude tests folder from phpcs linter, `<exclude-pattern>*/tests/*</exclude-pattern>`. I wonder if we can re-enable this to catch these automatically? I'm guessing there's a good reason. (I had a quick play and couldn't get the linter to find whitespace issues in tests files, maybe excluded in Woo or WP core rules?)

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,236 @@
...
+	 * and calling from_order_request. Used for all tests.
+	 *
+	 * @return assoc array with keys [ 'user_id', 'order' ] if successful.
+	 */
+	private function execute_create_customer_from_order( $email, $first_name, $last_name, $options = [] ) {
+		/// -- Test-specific setup start.
+
+		// Can't log out the user in a unit test ("headers already sent" error).
+		// So - we are assuming the tests run in an environment where the user is logged out.
+		// wp_logout();
```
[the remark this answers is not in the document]
haszari: Aha! I've added this here: ff17fedd5a022c9fc9517e7fc75f9c681410f1b4 - should make the tests that little bit more robust and explicit.

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -0,0 +1,236 @@
...
+ * - Confirm correct output &/or changes to site state.
+ * https://make.wordpress.org/core/handbook/testing/automated-testing/writing-phpunit-tests/
+ *
+ * @since $VID:$
+ */
+class CreateAccount extends WP_UnitTestCase {
+
+	private function get_test_instance() {
+		// Would be good to find a way to use a mock Package, e.g. to control experimental flag.
+		return new TestedCreateAccount( new Package( 'test', './' ) );
```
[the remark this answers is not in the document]
haszari: Sounds good to me! #3211

tests/php/Domain/Services/CreateAccount.php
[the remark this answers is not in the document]
nerrad: 🤔 I'm guessing it's to avoid some of the extra documentation around parameter definition, function comment blocks etc that is typically not needed for test files because the functions are typically more descriptive. We can look at expanding the lint coverage on test files for php too but that will surface a lot more linting fails needing handled, so probably good doing in separate issue.

assets/js/settings/blocks/constants.js
```diff
@@ -105,8 +110,6 @@ export const CART_PAGE_ID = storePages.cart.id;
 export const CART_URL = storePages.cart.permalink;
 
 export const CHECKOUT_ALLOWS_GUEST = getSetting( 'checkoutAllowsGuest', false );
-export const CHECKOUT_ALLOWS_SIGNUP = getSetting(
-	'checkoutAllowsSignup',
-	false
-);
+export const CHECKOUT_ALLOWS_SIGNUP =
+	isExperimentalBuild() && getSetting( 'checkoutAllowsSignup', false );
```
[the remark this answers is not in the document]
haszari: Good catch – I agree, adjusting/masking the store setting value is risky. 6c63837f3df3a720a71aa80013a33b9049a663e1

I didn't actually need to gate the flag - this is used as the default value for the block attribute, which is not feature gated. The attribute and value are included in all builds, but the UI and behaviour are gated to feature plugin only.

src/Domain/Services/CreateAccount.php
```diff
@@ -26,10 +26,28 @@ public function __construct( Package $package ) {
...
+	 * @return True if Checkout sign-up feature should be made available.
+	 */
+	public static function is_feature_enabled() {
+		// This new checkout signup flow is gated to dev builds for now.
+		// The main reason for this is that we are waiting on an new
+		// set-password endpoint/form in WooCommerce Core.
+		// When that's available we can review this and include in feature
+		// plugin alongside checkout block.
+		return Package::is_experimental_build();
+	}
```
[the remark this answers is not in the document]
haszari: I've rejigged the test class so it requires/assumes experimental build here: 428c6bd0d3a1a3b650cf51322dd4f62de1f4ef69

That also includes the `blocks.ini` config for `woocommerce_blocks_phase=3` so 🤞 for next travis run 😄

haszari:
Ready for another review @nerrad – tests are now running and passing in travis  🎉

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -98,130 +88,112 @@ public function test_create_customer_from_order( $email, $first_name, $last_name
...
-    public function create_customer_data()
-    {
-        return [
-        	// User requested an account.
-        	[
-        		'maryjones@testperson.net',
-        		'Mary',
-        		'Jones',
+	public function create_customer_data()
+	{
```
nerrad, line 92: nitpick: why is the bracket on a separate line? I know it's a PSR standard but we don't follow PSR (unfortunately).

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -36,9 +33,8 @@ private function get_test_instance() {
 	private function execute_create_customer_from_order( $email, $first_name, $last_name, $options = [] ) {
 		/// -- Test-specific setup start.
 
-		// Can't log out the user in a unit test ("headers already sent" error).
-		// So - we are assuming the tests run in an environment where the user is logged out.
-		// wp_logout();
+		// Simulate (force) logged out user – if the user is logged in, no account will be created.
+		wp_set_current_user( 0 );
```
nerrad, line 37: Apologies, but it looks like there's no need to do this because the [WordPress test suite handles this in `tearDown`](https://github.com/WordPress/wordpress-develop/blob/19513ff6a0fc27b2ff7e9debeb6e94b7e1b4d658/tests/phpunit/includes/abstract-testcase.php#L169).  So effectively, the current user will already be not set for the test (which is why it worked fine without it).

We could probably remove.

nerrad (approved):
Thanks for addressing the feedback Rua! It's all looking good and nice to see travis is green with the changes 🎉 

I have a couple comments that should be addressed but doing so doesn't require another review so I'm pre-approving. After you make the changes and have verified it tests well I'm good with it :shipit: Woohoo!

tests/php/Domain/Services/CreateAccount.php
haszari: Right, and we can safely assume that the tests run with no user logged in (even if we were the first test, before any `tearDown`!). Removed: 81b3a0eea114e78379c59916f0cf13f1fc226590

tests/php/Domain/Services/CreateAccount.php
haszari: Aha! This was because I was pasting from PHPUnit docs!! Fixed: 9edaf7aff87f5814de2bbb374a7b62716c8b166e

haszari:
I'm going to give this a final test pass on a real (ephemeral) site to confirm there's no hiccups with the feature flag.

- [ ] set up store so emails are delivered somewhere (change store email to non-a8c address)
- [ ] set up the store settings so user can choose an account
- [ ] test with shortcode checkout, no blocks plugin
- [ ] install the current release version of blocks
- [ ] switch checkout to block
- [ ] build release version of this branch, install
- [ ] confirm checkout works and no account is created, i.e. everything is hidden behind feature flag
- [ ] confirm editing checkout page works, no validation errors
- [ ] build debug/dev version of plugin, install in site
- [ ] test signup feature works

haszari:
Finally merging! Had to kick travis a few times to get one of the e2e tests to pass.

haszari:
Note – this has merged but won't be available in a release of the plugin until we have the set password flow working well. 

The next step for that is to implement it in WooCommerce core: https://github.com/woocommerce/woocommerce/issues/27754

cc @nerrad – I can get started on this tomorrow or next week, I'm assuming we want to keep moving and get this wrapped up so we can really ship it.

nerrad:
> cc @nerrad – I can get started on this tomorrow or next week, I'm assuming we want to keep moving and get this wrapped up so we can really ship it.

Yes you can get the start of the pull up for WooCommerce Core.

Issue #2852, opened 2020-07-14
⚛️ ‎‎‏‏‎ Product Title - Update icon, inspector, and toolbar controls


mikejolley:
The Product Title Element Block requires:

- [ ] A new icon
- [ ] Typography controls
- [ ] Color controls
- [ ] Heading level toolbar controls
- [ ] Alignment toolbar controls

Alignment needs specifying in the default template of the single product block so that:

- In the product block, the headline should default to h2, and left aligned.
- In grid based blocks, the headline should default to h2 and center aligned.

See https://woogrowp2.wordpress.com/2020/07/01/product-title-block/

senadir:
Leaving this out `and left-aligned.` as setting this value explicitly would mean it's always on the left even in RTL, the title is left aligned on LTR and right-aligned on RTL unless you explicitly set an absolute value for it.

PR #2859, opened 2020-07-14
rename HeadingToolbar to HeadingSwitcher

senadir:
This PR is needed for #2852

`HeadingToolbar` was named toolbar, even though it's not a toolbar component, but a switcher used in the sidebar, this PR renames it to `HeadingSwitcher`, subsequent PR will add a `HeadingToolbar` that adds a button to the block toolbar, with `HeadingSwitcher` inside of it.


senadir:
this might not be needed, `HeadingToolbar` does work like a toolbar indeed.

Issue #2872, opened 2020-07-15
Move draft order cleanup behind the feature flag.


nerrad:
Related to #2844, one thing we should be doing is moving the draft order deletion to behind a feature flag when the plugin is included in WC Core as a package. The checkout block is not exposed to WC core users yet so there's no need to be cleaning up `wc-checkout-draft` status orders yet.

PR #2874, opened 2020-07-15
Move Draft order logic behind feature flag.

nerrad:
Fixes: #2872 

Since much of the draft order functionality was spread throughout the codebase, in this pull I refactored all the draft order logic into a service class. This made it easier to put all it's logic behind a single feature flag conditional and gives us one class to go to for draft order implementation. If/when this moves to core, it also provides a single point of reference for what needs moved into WC Core.

As noted in #2872, there's not much sense in setting up a scheduled action for deleting draft orders if the environment won't have any. So this pull not only prevents adding the schedule if in the Woo Core package environment, but also unschedules any existing schedules that might be present (being schedules might be running in that environment unnecessarily if folks already upgraded to WC 4.3).

While technically, the unscheduling is not _necessary_. This is a good preventative measure for folks that have already installed WC 4.3 which has added the schedules unnecessarily.

## To test

The same steps can be followed as noted in #2574 (which I've included here to make it easier):

1. Create draft orders
2. Run job manually in Tools > Scheduled actions.
3. Change order dates, re-run to ensure orders are deleted.

Besides the above, it'd be good to simulate behaviour when in a non-feature plugin environment. This can be done by setting `woocommerce_blocks_phase = 1` in the `blocks.ini` file in your local environment and checking to see that the scheduled actions have disappeared.

~**Note:** Besides running phpunit, I have _not_ tested this yet because I ran out of time before a meeting and the end of my day. I thought I'd still throw this up for review and possible testing anyways, but I do hope to get to testing this later.~

I have tested the above steps and simulating behaviour when in a non-feature plugin environment.

### DIFF
--- a/src/Domain/Services/DraftOrders.php
+++ b/src/Domain/Services/DraftOrders.php
@@ -167,8 +167,7 @@ class DraftOrders {
 				as_enqueue_async_action( 'woocommerce_cleanup_draft_orders' );
 			}
 		} catch ( Exception $error ) {
-			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
-			error_log( $error->getMessage() );
+			wc_caught_exception( $error, __METHOD__ );
 		}
 	}
 
@@ -187,21 +186,20 @@ class DraftOrders {
 			return;
 		}
 
-		$prefix = __CLASS__ . '::delete_expired_draft_orders ';
 		$suffix = ' This is an indicator that something is filtering WooCommerce or WordPress queries and modifying the query parameters.';
 
 		// if count is greater than our expected batch size, then that's a problem.
 		if ( count( $order_results ) > 20 ) {
-			throw new Exception( $prefix . ' is getting an unexpected number of results returned from the query.' . $suffix );
+			throw new Exception( 'There are an unexpected number of results returned from the query.' . $suffix );
 		}
 
 		// if any of the returned orders are not draft (or not a WC_Order), then that's a problem.
 		foreach ( $order_results as $order ) {
 			if ( ! ( $order instanceof WC_Order ) ) {
-				throw new Exception( $prefix . ' returned results containing a value that is not a WC_Order.' . $suffix );
+				throw new Exception( 'The returned results contain a value that is not a WC_Order.' . $suffix );
 			}
 			if ( ! $order->has_status( self::STATUS ) ) {
-				throw new Exception( $prefix . ' has an order that is not a `wc-checkout-draft` status in the results.' . $suffix );
+				throw new Exception( 'The results contain an order that is not a `wc-checkout-draft` status in the results.' . $suffix );
 			}
 		}
 	}
--- a/src/Domain/Services/DraftOrders.php
+++ b/src/Domain/Services/DraftOrders.php
@@ -8,11 +8,16 @@
 namespace Automattic\WooCommerce\Blocks\Domain\Services;
 
 use Automattic\WooCommerce\Blocks\Domain\Package;
+use Exception;
+use WC_Order;
 
 /**
  * Service class for adding DraftOrder functionality to WooCommerce core.
  */
 class DraftOrders {
+
+	const DB_STATUS = 'wc-checkout-draft';
+	const STATUS    = 'checkout-draft';
 
 	/**
 	 * Holds the Package instance
@@ -35,35 +40,33 @@ class DraftOrders {
 	 */
 	public function init() {
 		if ( $this->package->is_feature_plugin_build() ) {
-			add_filter( 'wc_order_statuses', [ __CLASS__, 'register_draft_order_status' ] );
-			add_filter( 'woocommerce_register_shop_order_post_statuses', [ __CLASS__, 'register_draft_order_post_status' ] );
-			add_filter( 'woocommerce_valid_order_statuses_for_payment', [ __CLASS__, 'append_draft_order_post_status' ] );
-			add_action( 'woocommerce_cleanup_draft_orders', [ __CLASS__, 'delete_expired_draft_orders' ] );
+			add_filter( 'wc_order_statuses', [ $this, 'register_draft_order_status' ] );
+			add_filter( 'woocommerce_register_shop_order_post_statuses', [ $this, 'register_draft_order_post_status' ] );
+			add_filter( 'woocommerce_valid_order_statuses_for_payment', [ $this, 'append_draft_order_post_status' ] );
+			add_action( 'woocommerce_cleanup_draft_orders', [ $this, 'delete_expired_draft_orders' ] );
 			add_action( 'admin_init', [ $this, 'install' ] );
 		} else {
 			// Maybe remove existing cronjob if present because it shouldn't be needed in the environment.
-			$this->maybe_remove_cronjobs();
+			add_action( 'admin_init', [ $this, 'uninstall' ] );
 		}
 	}
 
 	/**
 	 * Installation related logic for Draft order functionality.
+	 *
+	 * @internal
 	 */
 	public function install() {
 		$this->maybe_create_cronjobs();
 	}
 
-
 	/**
 	 * Remove cronjobs if they exist (but only from admin).
+	 *
+	 * @internal
 	 */
-	protected function maybe_remove_cronjobs() {
-		if ( ! is_admin() ) {
-			return;
-		}
-		if ( function_exists( 'as_next_scheduled_action' ) && true === as_next_scheduled_action( 'woocommerce_cleanup_draft_orders' ) ) {
-			as_unschedule_all_actions( 'woocommerce_cleanup_draft_orders' );
-		}
+	protected function uninstall() {
+		$this->maybe_remove_cronjobs();
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


src/Domain/Services/DraftOrders.php
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * Append draft status to a list of statuses.
+	 *
+	 * @param array $statuses Array of statuses.
+	 * @return array
+	 */
+	public static function append_draft_order_post_status( $statuses ) {
+		$statuses[] = 'checkout-draft';
```
nerrad, line 118: This was a straight copy/paste from pre-existing code, but I just noticed and wondered, should this be `wc-checkout-draft` @mikejolley?

src/Library.php
```diff
@@ -114,76 +110,4 @@ protected static function register_atomic_blocks() {
...
-
-	/**
-	 * Register custom order status for orders created via the API during checkout.
-	 *
-	 * Draft order status is used before payment is attempted, during checkout, when a cart is converted to an order.
-	 *
-	 * @param array $statuses Array of statuses.
-	 * @return array
-	 */
-	public static function register_draft_order_status( array $statuses ) {
```
nerrad, line 126: I doubt anyone would be calling these functions directly, but do you think I should leave them as wrappers around the new function calls?

src/Domain/Services/DraftOrders.php
```diff
@@ -0,0 +1,151 @@
...
+	public function __construct( Package $package ) {
+		$this->package = $package;
+	}
+
+	/**
+	 * Set all hooks related to adding Checkout Draft order functionality to Woo Core.
+	 */
+	public function init() {
+		if ( $this->package->is_feature_plugin_build() ) {
+			add_filter( 'wc_order_statuses', [ __CLASS__, 'register_draft_order_status' ] );
```
mikejolley, line 38: Now moved, these methods no longer need to be static, or referenced with `__CLASS__`

src/Domain/Services/DraftOrders.php
```diff
@@ -0,0 +1,151 @@
...
+			add_filter( 'woocommerce_register_shop_order_post_statuses', [ __CLASS__, 'register_draft_order_post_status' ] );
+			add_filter( 'woocommerce_valid_order_statuses_for_payment', [ __CLASS__, 'append_draft_order_post_status' ] );
+			add_action( 'woocommerce_cleanup_draft_orders', [ __CLASS__, 'delete_expired_draft_orders' ] );
+			add_action( 'admin_init', [ $this, 'install' ] );
+		} else {
+			// Maybe remove existing cronjob if present because it shouldn't be needed in the environment.
+			$this->maybe_remove_cronjobs();
```
mikejolley, line 45: Any reason why install waits for admin_init and this one does not?

src/Domain/Services/DraftOrders.php
```diff
@@ -0,0 +1,151 @@
...
+	/**
+	 * Maybe create cron events.
+	 */
+	protected function maybe_create_cronjobs() {
+		if ( function_exists( 'as_next_scheduled_action' ) && false === as_next_scheduled_action( 'woocommerce_cleanup_draft_orders' ) ) {
+			as_schedule_recurring_action( strtotime( 'midnight tonight' ), DAY_IN_SECONDS, 'woocommerce_cleanup_draft_orders' );
+		}
+	}
+
+
```
mikejolley, line 78: Extra line breaks between these if my view is accurate

src/Domain/Services/DraftOrders.php
mikejolley: Core expects unprefixed names in this filter.

src/Domain/Services/DraftOrders.php
```diff
@@ -0,0 +1,151 @@
...
+				'date_modified' => '<=' . strtotime( '-1 DAY' ),
+				'limit'         => $batch_size,
+				'status'        => 'wc-checkout-draft',
+				'type'          => 'shop_order',
+			]
+		);
+
+		if ( $orders ) {
+			foreach ( $orders as $order ) {
+				$order->delete( true );
```
mikejolley, line 141: I was going to do this in a follow up but since we're moving all these methods anyway I think it should be here. We should wrap the delete in a `has_status` check to guard, and to be extra cautious I'd actually bail on the entire query if anything comes back unexpected. I'll think about this and suggest something.

src/Library.php
mikejolley: I don't see the utility in these for public use. Ideally they would have been private, but filters must be public scoped. Do you think we should add `@internal` or even `@private` docblocks to these methods, even though we're technically prevented from doing this explicitly?

tests/php/Domain/Services/DeleteDraftOrders.php
```diff
@@ -66,9 +81,12 @@ public function test_delete_expired_draft_orders() {
...
 
 		// Run delete query.
-		Library::delete_expired_draft_orders();
+		DraftOrders::delete_expired_draft_orders();
 
 		// Only 1 should remain.
 		$this->assertEquals( 1, (int) $wpdb->get_var( "SELECT COUNT(ID) from $wpdb->posts posts WHERE posts.post_status = 'wc-checkout-draft'" ) );
+
+		// The non-draft order should still be present
+		$this->assertEquals( 1, (int) $wpdb->get_var( "SELECT COUNT(ID) from $wpdb->posts posts WHERE posts.post_status = 'wc-on-hold'" ) );
```
mikejolley, line 90: 👍

src/Domain/Services/DraftOrders.php
```diff
@@ -0,0 +1,151 @@
...
+				$count ++;
```
mikejolley, line 142:
```suggestion
				// Explicitly checks the order status again here to ensure we have a draft. Guards against incorrect filter usage. 
				if ( $order->has_status( 'checkout-draft' ) ) {
					$order->delete( true );
					$count ++;
				}
```

mikejolley: If someone does filter, then count won't increment and subsequent jobs won't be queued. We could look into logging of some sort, but could also just wait for a core deletion method to become available thats hook-less (option 4).

src/Domain/Services/DraftOrders.php
nerrad: hmm ya it probably should wait for admin_init too.

src/Domain/Services/DraftOrders.php
```diff
@@ -0,0 +1,208 @@
...
+					$order->delete( true );
+					$count ++;
+				}
+			}
+			if ( $batch_size === $count && function_exists( 'as_enqueue_async_action' ) ) {
+				as_enqueue_async_action( 'woocommerce_cleanup_draft_orders' );
+			}
+		} catch ( Exception $error ) {
+			// phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
+			error_log( $error->getMessage() );
```
mikejolley, line 171: This should be wrapped with a check IMO `if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {`

mikejolley: Or we could use https://github.com/woocommerce/woocommerce/blob/dd49f89e77acaaa097251fe0a5dd69320ded48c6/includes/wc-deprecated-functions.php#L92

src/Domain/Services/DraftOrders.php
```diff
@@ -0,0 +1,208 @@
...
+	 * Asserts whether incoming order results are expected given the query
+	 * this service class executes.
+	 *
+	 * @param WC_Order[] $order_results The order results being asserted.
+	 * @param int        $expected_batch_size The expected batch size for the results.
+	 * @throws Exception If any assertions fail, an exception is thrown.
+	 */
+	private function assert_order_results( $order_results, $expected_batch_size ) {
+		// if not an array, then just return because it won't get handled
+		// anyways.
```
mikejolley, line 185: Nit: why does this comment wrap but the ones below do not?

nerrad: Habit, I wrap comments in JavaScript all the time, but sometimes remember that with PHP there's no hard rule on wrapping (so it's just a flip/flop here).

src/Domain/Services/DraftOrders.php
nerrad: I normally agree on logging only on non-production, but this would be triggered in an exceptional case where the logs can serve as a troubleshooting tool for why potentially no draft orders were cleaned up on this schedule.

tests/php/Domain/Services/DeleteDraftOrders.php
```diff
@@ -61,14 +86,18 @@ public function setUp() {
 	 */
 	public function test_delete_expired_draft_orders() {
 		global $wpdb;
+		$status = DraftOrders::DB_STATUS;
 
 		// Check there are 3 draft orders from our setup before running tests.
-		$this->assertEquals( 3, (int) $wpdb->get_var( "SELECT COUNT(ID) from $wpdb->posts posts WHERE posts.post_status = 'wc-checkout-draft'" ) );
+		$this->assertEquals( 3, (int) $wpdb->get_var( "SELECT COUNT(ID) from $wpdb->posts posts WHERE posts.post_status = '$status'" ) );
```
mikejolley, line 92: It would be better to use $wpdb->prepare here even though it's in the test suite (for status).

PR #2878, opened 2020-07-16
Fix undefined index 'enabled' for multiple payment method registrations

### LINKED ISSUE #2873: Fix undefined index 'enabled' for multiple payment method registrations.
Full notice being printed when running phpunit tests (example for one of the payment methods):

```
Notice: Undefined index: enabled in /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block/src/Payments/Integrations/BankTransfer.php on line 55
[15-Jul-2020 19:25:03 UTC] PHP Notice:  Undefined index: enabled in /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block/src/Payments/Integrations/CashOnDelivery.php on line 55
```

It looks like even with filter_var, we still need to verify that the index exists before attempting to use it.

mikejolley:
Prevents notices by checking first if values are set. Adds a helper to get a setting from the array of settings, with optional fallback.

Fixes #2873

On a clean install (without first saving gateway settings) check no notices are shown in the backend with debug mode enabled.

### DIFF
--- a/src/Payments/Integrations/AbstractPaymentMethodType.php
+++ b/src/Payments/Integrations/AbstractPaymentMethodType.php
@@ -25,6 +25,24 @@ abstract class AbstractPaymentMethodType implements PaymentMethodTypeInterface {
 	protected $name = '';
 
 	/**
+	 * Settings from the WP options table
+	 *
+	 * @var array
+	 */
+	protected $settings = [];
+
+	/**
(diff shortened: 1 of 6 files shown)


mikejolley:
@nerrad That could be a follow up, but Stripe would need updating too as it uses $settings directly. I did not want to mess with that class too much in this PR :)

PR #2884, opened 2020-07-17
Add autocapitalize attribute to Address Form fields

### LINKED ISSUE #2564: Set autocapitalization attribute for Cart & Checkout inputs
When introducing my name, the city name, etc. I would expect the mobile keyboard to autocapitalize the first letter. This can be accomplished with the [`autocapitalize` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize), we should evaluate which inputs need it and implement it.

Some more docs:
https://developers.google.com/web/updates/2015/04/autocapitalize#when_should_i_use_this

Screenshot

_(if I'm writing my name, the keyboard should show capital letters when the input field is empty)_

Aljullu:
Fixes #2564.

1. In order to test this, you will need to visit the Checkout page with a handheld device with Chrome for Android or iOS Safari.
2. Navigate through the input fields in the address form.
3. Verify when a field is focused, the keyboard shows capital letters by default according to the settings.

I took the [suggested capitalization settings](https://developers.google.com/web/updates/2015/04/autocapitalize#when_should_i_use_this) from this article.

### Changelog

> Cart and Checkout form fields show autocapitalized keyboard on mobile depending on the expected value.

### DIFF
--- a/assets/js/base/components/cart-checkout/address-form/default-address-fields.js
+++ b/assets/js/base/components/cart-checkout/address-form/default-address-fields.js
@@ -26,7 +26,7 @@ const AddressFields = {
 			'woo-gutenberg-products-block'
 		),
 		autocomplete: 'given-name',
-		autocapitalize: 'words',
+		autocapitalize: 'sentences',
 		required: true,
 		hidden: false,
 		index: 1,
@@ -38,7 +38,7 @@ const AddressFields = {
 			'woo-gutenberg-products-block'
 		),
 		autocomplete: 'family-name',
-		autocapitalize: 'words',
+		autocapitalize: 'sentences',
 		required: true,
 		hidden: false,
 		index: 2,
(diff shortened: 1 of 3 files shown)


assets/js/base/components/cart-checkout/address-form/default-address-fields.js
```diff
@@ -70,6 +74,7 @@ const AddressFields = {
 			'woo-gutenberg-products-block'
 		),
 		autocomplete: 'address-line2',
+		autocapitalize: 'words',
```
haszari, line 77: Possibly this could be `sentences`, but I don't think it's a big deal either way. E.g. `Second floor apartment 2`

haszari: Hmm, now that I've tested, I think we should use `sentences` everywhere :)

Aljullu:
Hmm, that's unfortunate. I replaced `autocomplete="words"` with `autocomplete="sentences"`. Not perfect, but it's still better than no capitalization at all.

PR #2903, opened 2020-07-22
Add to cart context provider

### LINKED ISSUE #2737: Surface add to cart errors in the single product block.
Surface add to cart errors in the single product block.

If the addToCart function within useStoreAddToCart fails, a notice should be shown on the product page.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2c9c8743c315a9bceb208ed3f7550dbaa87c11ad/assets/js/base/context/add-to-cart-form-context.js#L94-L112
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in 2c9c8743c315a9bceb208ed3f7550dbaa87c11ad when #2705 was merged. cc @mikejolley

### LINKED ISSUE #2738: Add Event Callbacks to the Add to Cart Form.
Add Event Callbacks to the Add to Cart Form.

- onChange should trigger when a form element changes, so for example, a variation picker could indicate that it&#x27;s ready.
- onSuccess should trigger after a successful add to cart. This could be used to reset form elements, do a redirect, or show something to the user.
- onFail should trigger when adding to cart fails. Form elements might show extra notices or reset. A fallback might be to redirect to the core product page in case of incompatibilities.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2c9c8743c315a9bceb208ed3f7550dbaa87c11ad/assets/js/base/context/add-to-cart-form-context.js#L102-L125
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in 2c9c8743c315a9bceb208ed3f7550dbaa87c11ad when #2705 was merged. cc @mikejolley

### LINKED ISSUE #2757: Hook up Variation Picker with Cart Form.
Hook up Variation Picker with Cart Form.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/97634ec194b239bb699d042037fb7dbf38a924d9/assets/js/atomic/blocks/product/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js#L93-L105
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in 97634ec194b239bb699d042037fb7dbf38a924d9 when #2497 was merged. cc @mikejolley

### LINKED ISSUE #2856: ⚛️ ‎‎‏‏‎ Add to Cart - Cart context provider API
The cart context provider (used in the product add to cart element block) needs to allow form elements to communicate so that a product can be added to the cart with additional data if needed, or to show/update certain product data based on its state. For example, the price may be changed after selecting a variation.

We need to design and create an API to facilitate this so that important functionality is exposed.

This includes, but may not be limited to:

- Providing product data
- Sharing common cart data (qty, other form element values)
- Exposing events (with event subscribers) so form elements can validate or perform actions such as surfacing error notices, or adding supplemental data to requests

See the Checkout Payment Gateway API for inspiration (subscribers).

This is closely linked to #2736 and #2738

mikejolley:
This PR implements the Add To Cart Context provider and hooks it up to the variation picker, quantity input, and add to cart button.

Closes #2856 - implemented updated context
Closes #2757 - made variation picker work with new context provider
Closes #2738 - implements events
Closes #2737 - notices are surfaced using store notices context

This is all heavily modelled upon the Checkout State Provider. It implements form state (IDLE, PRISTINE, DISABLED, PROCESSING) and exposes dispatch actions to set cart request data for the API to work. We can expand these APIs as we look further at extension needs and other product types, as well as the form-element registration system we've yet to implement, but that's the next step.

## AddToCartFormContextProvider

This provider wraps children with Add to Cart Form state context, validation context and includes a `FormSubmit` component. This is similar to the Checkout Processing Component and listens for events to submit the request to the API.

## Add To Cart Form Context

The context exposes data about the add to cart form, as well as current status of the form submission. Status includes:

- PRISTINE: Form is in it's initialized state.
- IDLE: When form state has changed but there is no activity happening.
- DISABLED: If the form cannot be submitted due to missing constraints, this status is assigned.
- BEFORE_PROCESSING: This is the state before form processing begins after the add to cart button has been pressed. Validation occurs at point point.
- PROCESSING: After BEFORE_PROCESSING status emitters have finished successfully.
- AFTER_PROCESSING: After server side processing is completed  this status is set.

Anything added to `requestParams` is sent along with the API request to add to cart; this is what the variation form is hooking into to set the variation ID and variation data.

```
/**
 * @typedef {Object} AddToCartFormContext
 *
 * @property {boolean}                        showFormElements   True if showing a full add to cart form.
 * @property {Object}                         product            The product object to add to the cart.
 * @property {number}                         quantity           Stores the quantity being added to the cart.
 * @property {number}                         minQuantity        Min quantity that can be added to the cart.
 * @property {number}                         maxQuantity        Max quantity than can be added to the cart.
 * @property {Object}                         requestParams      List of params to send to the API.
 * @property {boolean}                        isIdle             True when the form state has changed and has no activity.
 * @property {boolean}                        isDisabled         True when the form cannot be submitted.
 * @property {boolean}                        isProcessing       True when the form has been submitted and is being processed.
 * @property {boolean}                        isBeforeProcessing True during any observers executing logic before form processing (eg. validation).
 * @property {boolean}                        isAfterProcessing  True when form status is AFTER_PROCESSING.
 * @property {boolean}                        hasError           True when the form is in an error state. Whatever caused the error (validation/payment method) will likely have triggered a notice.
 * @property {AddToCartFormEventRegistration} eventRegistration  Event emitters that can be subscribed to.
 * @property {AddToCartFormDispatchActions}   dispatchActions    Various actions that can be dispatched for the add to cart form context data.
 */
```

Event emitters exist for the stages of form submission:

- `onAddToCartAfterProcessingWithSuccess` - Used to register a callback that will fire after form has been processed and there are no errors.
- `onAddToCartAfterProcessingWithError` - Used to register a callback that will fire when the form has been processed and has an error.
- `onAddToCartBeforeProcessing` - Used to register a callback that will fire when the form has been submitted before being sent off to the server.

Available dispatch actions include:

- `resetForm` - Dispatches an action that resets the form to a pristine state.
- `submitForm` - Dispatches an action that tells the form to submit.
- `setQuantity` - Dispatches an action that sets the quantity to the given value.
- `setRequestParams` - Dispatches an action that sets params for the add to cart request (key value pairs).
- `setHasError` - Dispatches an action that sets the status to having an error.
- `setAfterProcessing` - Dispatches an action that sets the status to after processing and also sets the response data accordingly.

**To test:**

- Create a single product block for a simple product. Check you can add various quantities to the cart without issue.
- Repeat the test with a variable product. The variation attribute picker should function and validate if missing attributes.
- Test the checkout flow - I moved some shared files (event emitters and validation)

I tested by adding all product types as single product blocks to a page with and without showFormElements enabled.


assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-select-control.js
```diff
@@ -19,17 +23,67 @@ const selectAnOption = {
...
+				onChange={ onChange }
+				required={ true }
+				className={ classnames(
+					'wc-block-components-product-add-to-cart-attribute-picker__select',
+					{
+						'has-error': error.message && ! error.hidden,
+					}
+				) }
+			/>
+			<ValidationInputError
```
mikejolley, line 82: It was not possible to reuse the checkout selects because they use the CustomSelect component and our designs called for native controls.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/utils.js
```diff
@@ -69,23 +69,71 @@ export const getSelectControlOptions = (
...
+/**
+ * Given a list of terms, filter them and return active options for the select boxes. This factors in
+ * which options should be hidden due to current selections.
+ *
+ * @param {Object} attributes List of attribute names and terms.
+ * @param {Object} variationAttributes Attributes for each variation keyed by variation ID.
+ * @param {Object} selectedAttributes Attribute Name Value pairs of current selections by the user.
+ * @return {Array} Select box options.
+ */
+export const getActiveSelectControlOptions = (
```
mikejolley, line 81: This is moved from the attribute-picker file to utils for readability.

assets/js/base/context/add-to-cart-form/form/submit/index.js
```diff
@@ -0,0 +1,141 @@
...
+	// Triggers form submission to the API.
+	const submitFormCallback = useCallback( () => {
+		setIsSubmitting( true );
+		removeNotice( 'add-to-cart' );
+
+		const fetchData = {
+			id: product.id || 0,
+			quantity,
+			...requestParams,
+		};
```
mikejolley, line 68: This is potential to extend this and the API further as we look at extensions needs, but this should be flexible enough to deal with it.

assets/js/blocks/single-product/block.js
```diff
@@ -17,6 +18,7 @@ import { BLOCK_NAME } from './constants';
  */
 const Block = ( { isLoading, product, children } ) => {
 	const className = 'wc-block-single-product wc-block-layout';
+	const noticeContext = `woocommerce/single-product/${ product?.id || 0 }`;
```
mikejolley, line 21: This keeps any notices within the single product block which threw them.

assets/js/base/context/add-to-cart-form/form-state/event-emit.js
```diff
@@ -0,0 +1,53 @@
...
+	ADD_TO_CART_AFTER_PROCESSING_WITH_SUCCESS:
+		'add_to_cart_after_processing_with_success',
+	ADD_TO_CART_AFTER_PROCESSING_WITH_ERROR:
+		'add_to_cart_after_processing_with_error',
+};
+
+/**
+ * Receives a reducer dispatcher and returns an object with the
+ * callback registration function for the checkout emit
+ * events.
```
nerrad, line 22: Slight change needed for description here.

assets/js/base/context/add-to-cart-form/form-state/index.js
```diff
@@ -0,0 +1,316 @@
...
+		[]
+	);
+
+	/**
+	 * This Effect is responsible for disabling or enabling the form based on the provided product.
+	 */
+	useEffect( () => {
+		const status = addToCartFormState.status;
+		const willBeDisabled =
+			! product.id || ! productIsPurchasable( product );
```
nerrad, line 150: `willBeDisabled` could probably be calculated outside of this effect since `product` is unlikely to change often for this provider. Which means we could avoid unnecessary hits to `productIsPurchaseable`.  Then you'd just have `willBeDisabled` as a dependency.

assets/js/base/context/add-to-cart-form/form-state/reducer.js
```diff
@@ -0,0 +1,151 @@
...
+ */
+export const reducer = ( state = DEFAULT_STATE, { quantity, type, data } ) => {
+	let newState;
+	switch ( type ) {
+		case SET_PRISTINE:
+			newState = DEFAULT_STATE;
+			break;
+		case SET_IDLE:
+			newState =
+				state.state !== IDLE
```
nerrad, line 43: This should be `state.status !== IDLE`.  I had a look over the `checkout-state` reducer and I see the same bug!  Fixing could reduce the number of unnecessary renders.
```suggestion
				state.status !== IDLE
```

assets/js/base/context/add-to-cart-form/form-state/reducer.js
```diff
@@ -0,0 +1,151 @@
...
+					? {
+							...state,
+							status: IDLE,
+					  }
+					: state;
+			break;
+		case SET_DISABLED:
+			newState =
+				state.state !== DISABLED
```
nerrad, line 52: Same here, this should be `state.status`.

```suggestion
				state.status !== DISABLED
```

assets/js/base/context/add-to-cart-form/form-state/reducer.js
```diff
@@ -0,0 +1,151 @@
...
+					: state;
+			break;
+		case SET_REQUEST_PARAMS:
+			newState = {
+				...state,
+				requestParams: {
+					...state.requestParams,
+					...data,
+				},
+			};
```
nerrad, line 75: Is there a way to determine whether the new requestParams need to be be added or not, or is it intentional that incoming `requestParams` are always merged and override existing state?

Also, is there any validation needed for what `data` contains here?

assets/js/base/context/add-to-cart-form/form/index.js
```diff
@@ -0,0 +1,38 @@
...
+	children,
+	product,
+	showFormElements,
+} ) => {
+	return (
+		<AddToCartFormStateContextProvider
+			product={ product }
+			showFormElements={ showFormElements }
+		>
+			<ValidationContextProvider>
```
nerrad, line 32: This needs to be the parent of `AddToCartFormStateContextProvider` because the form state provider consumes the validation context.

assets/js/base/context/add-to-cart-form/form/submit/index.js
```diff
@@ -0,0 +1,141 @@
...
+		hasError,
+		isProcessing,
+		requestParams,
+	} = useAddToCartFormContext();
+	const {
+		hasValidationErrors,
+		showAllValidationErrors,
+	} = useValidationContext();
+	const { addErrorNotice, removeNotice } = useStoreNotices();
+	const { receiveCart } = useStoreCart();
```
nerrad, line 34: This would be something that would be good to lift up to the `AddToCartFormContextProvider` (anything Cart specific). We could do this kind of state lifts (I think I have one other comment mentioning this) in follow-ups. This is also an area where there may be shared functionality used by the cart provider (another reason to evaluate/work on separately)

assets/js/base/context/add-to-cart-form/index.js
```diff
@@ -0,0 +1,3 @@
+export * from './form';
+export * from './form-state';
+//export * from './form-elements';
```
nerrad, line 3: maybe just remove the comment block?

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
```diff
@@ -19,86 +20,72 @@ import {
...
-			variationAttributes,
-			attributeNames,
+	// Set requests params as variation ID and data changes.
+	useEffect( () => {
+		dispatchActions.setRequestParams( {
+			id: variationId,
+			variation: Object.keys( selectedAttributes ).map(
+				( attributeName ) => {
+					return {
+						attribute: attributeName,
```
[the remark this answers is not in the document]
mikejolley: This is passed to the API. The API doesn't camel case props and only needs key value pairs.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
```diff
@@ -19,86 +20,72 @@ import {
...
+	const [ selectedAttributes, setSelectedAttributes ] = useState( {} );
+	const { dispatchActions } = useAddToCartFormContext();
 
 	const attributeNames = Object.keys( attributes );
-	const hasSelectedAttributes =
-		Object.values( selectedAttributes ).filter( Boolean ).length > 0;
 	const hasSelectedAllAttributes =
 		Object.values( selectedAttributes ).filter(
 			( selected ) => selected !== ''
 		).length === attributeNames.length;
```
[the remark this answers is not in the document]
mikejolley: Slightly confused about this comment. The one below questions the need for memo, but if this was momorized it would have similar dependencies. What is your criteria for requiring memo when reviewing?

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-select-control.js
mikejolley: The idea is to not custom style the selects. This was a design decision. Personally I wanted to use checkout style selects.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/utils.js
```diff
@@ -69,23 +69,71 @@ export const getSelectControlOptions = (
...
- * @param {Object} props.attributeNames List of all possible attribute names.
+ * @param {Object} attributes List of attribute names and terms.
+ * @param {Object} variationAttributes Attributes for each variation keyed by variation ID.
+ * @param {Object} selectedAttributes Attribute Name Value pairs of current selections by the user.
  * @return {Array} List of matching variation IDs.
  */
-export const getVariationsMatchingSelectedAttributes = ( {
-	selectedAttributes,
+export const getVariationsMatchingSelectedAttributes = (
+	attributes,
```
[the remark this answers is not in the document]
mikejolley: It could, it did before. I just wanted each of these functions to accept the same args to make them easier to work with.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/utils.js
```diff
@@ -105,3 +153,24 @@ export const getVariationsMatchingSelectedAttributes = ( {
 		} )
 	);
 };
+
+/**
+ * Given a list of variations and a list of attribute values, return the best match.
```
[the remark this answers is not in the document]
mikejolley: First is returned as you say. Admin decides the order in which variations are returned based on the order they are defined/saved.

assets/js/atomic/blocks/product-elements/add-to-cart/shared/add-to-cart-button.js
```diff
@@ -14,20 +15,37 @@ const AddToCartButton = () => {
...
-		formSubmitting,
-		onSubmit,
+		isDisabled,
+		isProcessing,
+		eventRegistration,
+		hasError,
+		dispatchActions,
 	} = useAddToCartFormContext();
+	const { cartQuantity } = useStoreAddToCart( product.id || 0 );
+	const [ addedToCart, setAddedToCart ] = useState( false );
```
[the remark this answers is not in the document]
mikejolley: Cart qty could but the other is specific to a button instance. We only want the check to appear on the button that was used here.

assets/js/base/context/add-to-cart-form/form-state/reducer.js
mikejolley: In this implementation, it should override. We could add replace or add actions if this is deemed to be useful.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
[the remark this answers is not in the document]
nerrad: Sorry for the confusion 😄 .  The question here isn't necessarily connected to memoization, but rather if this could be a potentially expensive operation. If the answer is yes, then I think we might want to take a closer look at how we can reduce how often this is calculated (of which memoization might be one option, depending on how often depencies might change.

On first glance, the dependencies here don't look like they'd change as much as some of the dependencies in the below code that is already memoized.

> What is your criteria for requiring memo when reviewing?

Typically I consider a few things:

- memoizing has a cost of it's own with additional overhead. If dependencies change relatively frequently, then any gains from memoization are negated and performance could actually be worse than no memoization.
- does the calculated value get used as a prop in child components or in updating state? If yes, then it's possible that memoization will reduce the amount of times that state update/prop causes component re-renders.
- does the calculated value get used as a dependency for hooks? If yes, then it's possible memoization can reduce the impact on those hooks.
- does the calculated value resolve to a primitive? If so, then memoization is _likely_ not needed (unless that calculation itself is expensive). Otherwise, things that have reference changes (Objects, Arrays, functions) on every call to the component typically do get memoized to ensure child components implementing them don't re-render unnecessarily.

In the case I commented on below, I reviewed the source of `attributes` being fed into the `AttributePicker` component and it looks like they ultimately come from `getAttributes` (`variation-attributes/utils`) and as far as I can tell, that function returns either a new Array or Object (incidentally, I just noticed that, shouldn't it return a consistent type?) everytime it is called. So for the purpose of `attributes` being a dependency on `useMemo` the memoization is not accomplishing anything because the value is rememoized everytime.  So you're actually hurting performance by the use of `useMemo` in this case.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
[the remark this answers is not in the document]
mikejolley: Would you prefer we had a setX for each key in the API schema? e.g. setRequestID setRequestVariation and so on.

assets/js/base/context/add-to-cart-form/form-state/reducer.js
mikejolley: Should I take care of checkout too or are you doing it?

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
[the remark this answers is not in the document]
nerrad: I don't think we need `setX` keys, typically schema's just describe the property and the http method describes what is done with those properties. `setX` feels more like RPC endpoint behaviour.

My comment here is less about the REST API schema shape and more about how the property is stored in state. It sounds like you might be intending on mapping the state 1:1 with the REST API schema, which has value as long as it's consistently applied. Otherwise, I don't see anything wrong with having a layer that takes the data in state and maps that to the shape the REST endpoint requires when prepping for the request.

assets/js/base/context/add-to-cart-form/form-state/reducer.js
nerrad: > Should I take care of checkout too or are you doing it?

I can do it.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
```diff
@@ -19,86 +20,72 @@ import {
...
-	// Gets valid attribute options for the picker taking current selections into account.
+	// Get options for each attribute picker.
 	const filteredAttributeOptions = useMemo( () => {
-		const options = [];
+		return getActiveSelectControlOptions(
+			attributes,
+			variationAttributes,
+			selectedAttributes
+		);
+	}, [ attributes, variationAttributes, selectedAttributes ] );
```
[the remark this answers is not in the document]
mikejolley: The calculation logic does loop a lot and does seem computational heavy. Perhaps it would be better to use a ref to track attributes because they don't change often.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/utils.js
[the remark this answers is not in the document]
mikejolley: I've added tests for all utils in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2903/commits/f721fddc0c8a813293a0915feacb8e4dc7e66491

assets/js/base/context/add-to-cart-form/form/submit/index.js
mikejolley: I think once we work on form-elements, what needs moving to context will become clearer, so I'm happy to have this in a follow up.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/utils.js
```diff
@@ -69,23 +69,71 @@ export const getSelectControlOptions = (
...
+	attributes,
+	variationAttributes,
+	selectedAttributes
+) => {
+	const options = [];
+	const attributeNames = Object.keys( attributes );
+	const hasSelectedAttributes =
+		Object.values( selectedAttributes ).filter( Boolean ).length > 0;
+
+	attributeNames.forEach( ( attributeName ) => {
```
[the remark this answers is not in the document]
mikejolley: Yes there is lots of looping (same on the PHP side!) but I don't think I can do much optimisation. Instead I've added unit tests and inline comments to explain *why* it does the loops. It needs to do quite a bit of comparison to built the list of valid variations taking selections into account.

mikejolley:
@nerrad I think this could have another pass now. I've added tests for the attribute picker utils and responded to most feedback; the one remaining is https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2903#discussion_r459476716 which Ill reply inline.

R/e typedefs they exist for the new context provider.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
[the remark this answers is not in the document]
mikejolley: I did indeed intend to map 1:1 with the REST API schema since this schema should be documented and be easiest to work with. I'm also assuming we may allow extensions to add to that schema in the future which is why a 1:1 might make sense—it would be weird to rename some props from state and not others?

I think if the issue is with the _schema_ naming we should tackle in a follow up, although personally I'm happy with the key value pairs of Attribute e.g. _Color_ and Value e.g. _Blue_.

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/variable/variation-attributes/attribute-picker.js
```diff
@@ -1,104 +1,104 @@
...
+	const attributesRef = useRef( attributes );
+	const variationAttributesRef = useRef( variationAttributes );
 	const [ variationId, setVariationId ] = useState( 0 );
+	const [ selectedAttributes, setSelectedAttributes ] = useState( {} );
 
-	// @todo Support default selected attributes in Variation Picker.
-	const [ selectedAttributes, setSelectedAttributes ] = useState( [] );
+	// Keeps refs in sync.
+	useEffect( () => {
+		if ( ! isShallowEqual( attributesRef.current, attributes ) ) {
```
nerrad, line 33: We have a [custom `useShallowEqual` hook](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/base/hooks/use-shallow-equal.js) that could be used (could condense some of this code). Same for the following effect for `variationAttributesRef`.

mikejolley:
> For a simple product, I can add to the cart, but the (legacy) cart widget doesn't update the value. Albert recently added a helper that handles updating the cart widget with the Cart block, and it'd be good to implement that here too.

That was me :) 

> Variations aren't working for me.

@nerrad can you just double check this is purchasable in the actual store because if it's missing price or is marked non-purchasable you'll see the message you shared.

nerrad:
> That was me :)

woops

> @nerrad can you just double check this is purchasable in the actual store because if it's missing price or is marked non-purchasable you'll see the message you shared.

Ahh, found out what the problem was. The product I was testing with was indeed not purchaseable 🤦‍♂️ .  The other variable product I tried didn't seem to be working as expected because the Add to Cart _button_ block was being used instead of just the Add to Cart block. Once I resolved that things worked.

I noticed an issue with encoded attribute labels being displayed as options in the select (and let Mike know in slack).

mikejolley:
Ah I see 👍  Ill leave that out for now because outside cart events won't affect the forms, only the other way.

PR #2904, opened 2020-07-22
Fix JS-rendered blocks inside Empty Cart

### LINKED ISSUE #2836: All Products block doesn't work when added to the empty cart

Aljullu:
Fixes #2836.

This PR refactors how `renderFrontend` works so blocks that might contain JS-rendered inner blocks can trigger a `wc-blocks_render_blocks_frontend` JS event to render the inner blocks.

### Screenshots

Peek 2020-08-03 20-10

1. Add the Cart block and switch to the Empty Cart edit mode.
2. Replace the Newest products block with the All Products block. Feel free to add other JS-rendered blocks like a filter or a Reviews block.
3. View the page in the frontend.
4. Verify all blocks render correctly when the cart is empty. Try adding a product to the cart (so it switches to the Full Cart view) and removing it (so it switches back to the Empty Cart view). Verify the blocks still render correctly.

### Changelog

> Fixed an issue that was making some blocks not to render correctly in the Empty cart template.


assets/js/base/utils/render-frontend.js
```diff
@@ -4,22 +4,41 @@
 import { render, Suspense } from '@wordpress/element';
 import BlockErrorBoundary from '@woocommerce/base-components/block-error-boundary';
 
+// Some blocks take care of rendering their inner blocks automatically. For example,
+// the empty cart. In those cases, we don't want to trigger the render function of
+// inner components on load. Instead, the wrapper block can trigger the event
+// `wc-blocks_render_blocks_frontend` to render its inner blocks.
+const selectorsToSkipOnLoad = [ '.wp-block-woocommerce-cart' ];
```
Aljullu, line 11: For now, I'm hard-coding the Cart class name here, but if we see this is needed in other blocks, we could try to dynamically fill this array when blocks are rendered with the Data Registry.

assets/js/base/utils/legacy-events.js
```diff
@@ -3,30 +3,40 @@ const Event = window.Event || null;
...
+ * @param {HTMLNode}  [options.element]    Element that dispatches the event. By
+ *                                         default, the body.
  */
-const dispatchEvent = ( name, bubbles = false, cancelable = false ) => {
+export const dispatchEvent = (
+	name,
+	{ bubbles = false, cancelable = false, element }
+) => {
+	if ( ! element ) {
+		element = document.body;
```
senadir, line 19: why not have this in the destructuring declaration?

assets/js/base/utils/render-frontend.js
```diff
@@ -53,4 +79,27 @@ export const renderFrontend = ( {
...
+	} );
+	// Render wrappers inner blocks when the event `wc-blocks_render_blocks_frontend`
+	// is triggered.
+	Array.prototype.forEach.call( wrappersToSkipOnLoad, ( wrapper ) => {
+		wrapper.addEventListener( 'wc-blocks_render_blocks_frontend', ( e ) => {
+			renderBlockFrontend( { ...props, wrapper: e.target } );
+		} );
+	} );
+};
+
```
senadir, line 104: While I get the higher-level logic, I'm having a hard time understanding the logic happening here, what exactly are we doing? waiting for an event to render again?

assets/js/base/utils/legacy-events.js
Aljullu: Ah, I don't know why. I thought it was breaking e2e tests but now it seems to work fine. Changed it in 04ae943c675b94e599a1d1700075a79ef8be5004.

assets/js/base/utils/render-frontend.js
Aljullu: `renderFrontend` does three things:
* Gets the wrapper elements that should be skipped on page load. That means that inner blocks of those wrappers should not be rendered on page load but instead when an event on the wrapper is triggered.
* Renders the block frontend for all blocks which are not inside those wrappers.
* For each of those wrappers, adds an event listener so, when the `wc-blocks_render_blocks_frontend` event is triggered, inner blocks are rendered.

In a343fc08c6d02ac9476ba6131eb8f8a3e03fb7c5, I spent some time renaming functions, variables and re-organizing code trying to make code a bit simpler. Hopefully it's a bit easier to understand now. :slightly_smiling_face:

nerrad:
Looks like there's some JS tests failing that need addressed here before merge.

Aljullu:
It seemed to be caused by 04ae943c675b94e599a1d1700075a79ef8be5004, I reverted it and tests are passing again.

assets/js/base/utils/legacy-events.js
Aljullu: Had to revert this. Taking a closer inspection what happened was:
* `document.body.querySelector( ... )` returned `null` if it didn't find the element, so `element` was null.
* when destructuring a nullish object property, the default value doesn't seem to be applied, so `element` was not getting assigned to `document.body`, instead it kept the `null` value.

TIL
```JS
const wizard = { name: 'Harry', middleName: null, lastName: 'Potter'};
const { middleName = 'James' } = wizard;
console.log( middleName ); // null
```

nerrad: 👍 if `middleName` was `undefined` however, then `James` would be the value. `null` is considered a value which is why the default doesn't get assigned.

PR #2921, opened 2020-07-28
add hover style to ensure consistent margin/padding for order summary so it doesn't jump on mouseover

### LINKED ISSUE #2914: Checkout block - extra margin-top on summary heading (button) on hover
The checkout summary sidebar jumps down on mouseover:

It looks like this is coming from `.wc-blocks-components-panel__button:hover` css rule, likely we just need to override the hover state in `.wc-block-components-order-summary .wc-blocks-components-panel__button`. 

Saw this in current `main`, haven't checked release version.

haszari:
Fixes #2914

Our order summary has `margin-top: 0`/`padding-top: 0` so box aligns with top of checkout left pane (form/steps). This rule is applied to a button component, and there's a hover style rule with a different (default) margin. So we need to override the `:hover` rule too so the order summary doesn't bounce on mouse-over.

1. Set up checkout block, add stuff to cart, proceed to checkout.
2. On desktop, mouseover order summary. 
3. Shouldn't bounce / jump.

### DIFF
--- a/assets/js/base/components/cart-checkout/order-summary/style.scss
+++ b/assets/js/base/components/cart-checkout/order-summary/style.scss
@@ -1,6 +1,5 @@
 .wc-block-components-order-summary {
-	.wc-blocks-components-panel__button,
-	.wc-blocks-components-panel__button:hover {
+	.wc-blocks-components-panel__button {
 		margin-top: 0;
 		padding-top: 0;
 	}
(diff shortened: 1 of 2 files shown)


haszari:
> I also wonder if, instead of fixing it in Order summary CSS, we could decrease the specificity of the panel button styles. I remember #2818 originally changed that (see #2818 (comment)) but I ended up forgetting why that change was done so I removed it from the PR. 😅

That's a good tip @Aljullu - I agree, would be better to simplify the parent styles to avoid the need for specific overrides. I've reinstated your fix here e735dd32bfbb3f3432b6966377bfe71db32ac4d5, and it's testing well for me in Safari and Firefox. Can you give it a retest / re-review to be sure? Feel free to merge if you're happy.

PR #2926, opened 2020-07-28
Fix  missing permissions_callback arg in StoreApi route definitions

### LINKED ISSUE #2908: [WordPress 5.5 beta] Fix REST API notices in endpoints
WordPress core had this changeset committed fairly recently: https://core.trac.wordpress.org/changeset/48526

> REST API: Issue a _doing_it_wrong when registering a route without a permission callback.

It looks like this will cause some notices with the endpoints in this plugin because we don't have that property defined in our route registrations that I can tell (likely will break things using the routes too with `WP_DEBUG_DISPLAY` set to true).

nerrad:
Fixes #2908 

See #2908 for details. WordPress 5.5 introduces tighter requirements for the REST API endpoint registration which includes the requirement that `permission_callback` argument is included with the configuration object for each route.

Our StoreApi is all publicly accessible routes, so this property was excluded. The fix was just to be explicit about these routes being public and thus the property is included with the WordPress `__return_true` callback provided as the value.

## To test

Unfortunately, this does impact all routes (but should be minimal impact, it's just explicitly defining behaviour that was already present). There's two primary things to watch for while testing:

- The PHP notice for missing `permission_callback` is no longer logged.
- There is no breakage to existing behaviour.

Some REST API test coverage for this does exist in the PHP unit tests.

Basically you can smoke test all behaviour for:

* [ ] Reviews block
* [ ] All Products and filter blocks
* [ ] Cart Block
* [ ] Checkout Block

Watch for any console/network errors and unexpected (and unreported) behaviour.

### DIFF
--- a/src/StoreApi/Routes/Cart.php
+++ b/src/StoreApi/Routes/Cart.php
@@ -32,9 +32,10 @@ class Cart extends AbstractCartRoute {
 	public function get_args() {
 		return [
 			[
-				'methods'  => \WP_REST_Server::READABLE,
-				'callback' => [ $this, 'get_response' ],
-				'args'     => [
+				'methods'             => \WP_REST_Server::READABLE,
+				'callback'            => [ $this, 'get_response' ],
+				'permission_callback' => '__return_true',
+				'args'                => [
 					'context' => $this->get_context_param( [ 'default' => 'view' ] ),
 				],
 			],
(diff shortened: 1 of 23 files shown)


Aljullu:
I opened an issue for the PHP tests failing: #2936.

PR #2934, opened 2020-07-29
Show payment gateways in correct (merchant-configured) order

### LINKED ISSUE #2157: Checkout Block: Re-order registered payment methods initially by the order they are configured in WooCommerce Payment settings
WooCommerce payment settings allows for setting the display order in the frontend for payment methods. We don't account for any of that currently in the new checkout block but we could account for that by doing some sorting based on those settings in [this code](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js).

Likely, in iterations of the block, the ux of the block could allow for the merchant to change the order via block settings.

haszari:
Fixes #2157

Merchants can configure the order that payment methods / gateways are displayed, by reordering them in `WooCommerce > Settings > Payments`. This PR uses this configured order in the checkout block.

This is achieved by:

- Adding a new bit of script data `wcSettings.paymentGatewaySortOrder`. This is an array of payment method names.
- In `usePaymentMethodRegistration`, we're now iterating over `paymentGatewaySortOrder` when determining whether each payment method should be presented (`refreshCanMakePayments`).

The result is that the payment method tabs should be in the same order as the store settings, and consistent with the order of the radiobuttons in classic / shortcode checkout.

##### Note - COD is not always displayed in merchant-configured order

Note there is one exception to the ordering. COD can be configured to require a specific shipping method. When the page is rendered, the current shipping option affects `wcSettings.paymentGatewaySortOrder` - i.e. COD may be excluded. If it is excluded from the list, then it will be added to the end if user selects a relevant shipping option in checkout. This might be different if COD is ordered first in config.

This is an edge case and shouldn't cause any major issues. However, we may want to reconsider (in future) and implement our own `wcSettings.paymentGatewaySortOrder`, or potentially allow reordering as a block setting, edited wysiwyg in Gutenberg.

### Screenshots

Screen Shot 2020-07-29 at 4 22 55 PM

1. Set up a few different payment methods. Also set up checkout block.
2. Reorder your payment methods how you like in `WooCommerce > Settings > Payments`, then click `Save changes` to persist.
3. On front end, add stuff to cart and proceed to checkout. 

Payment methods should display using the order you chose. 

Complete a few test purchases to confirm that payment methods are still working and there are no regressions in checkout.

Also test payment methods that may be hidden based on checkout state, e.g. make COD only available with specific shipping methods.

### Changelog

> Checkout block: Payment gateways are shown in the correct order as configured in store settings.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -12,7 +12,10 @@ import {
 	useShippingDataContext,
 } from '@woocommerce/base-context';
 import { useStoreCart, useShallowEqual } from '@woocommerce/base-hooks';
-import { CURRENT_USER_IS_ADMIN } from '@woocommerce/block-settings';
+import {
+	CURRENT_USER_IS_ADMIN,
+	PAYMENT_GATEWAY_ORDER,
+} from '@woocommerce/block-settings';
 
 /**
  * If there was an error registering a payment method, alert the admin.
@@ -86,8 +89,13 @@ const usePaymentMethodRegistration = (
 				[ paymentMethod.name ]: paymentMethod,
 			};
 		};
-		for ( const paymentMethodName in registeredPaymentMethods ) {
+
+		for ( let i=0; i<PAYMENT_GATEWAY_ORDER.length; i++  ) {
+			const paymentMethodName = PAYMENT_GATEWAY_ORDER[i];
 			const paymentMethod = registeredPaymentMethods[ paymentMethodName ];
+			if ( ! paymentMethod ) {
+				continue;
+			}
 
 			// In editor, shortcut so all payment methods show as available.
 			if ( isEditor ) {
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -90,8 +90,8 @@ const usePaymentMethodRegistration = (
 			};
 		};
 
-		for ( let i=0; i<PAYMENT_GATEWAY_SORT_ORDER.length; i++  ) {
-			const paymentMethodName = PAYMENT_GATEWAY_SORT_ORDER[i];
+		for ( let i = 0; i < PAYMENT_GATEWAY_SORT_ORDER.length; i++ ) {
+			const paymentMethodName = PAYMENT_GATEWAY_SORT_ORDER[ i ];
 			const paymentMethod = registeredPaymentMethods[ paymentMethodName ];
 			if ( ! paymentMethod ) {
 				continue;
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -14,7 +14,7 @@ import {
 import { useStoreCart, useShallowEqual } from '@woocommerce/base-hooks';
 import {
 	CURRENT_USER_IS_ADMIN,
-	PAYMENT_GATEWAY_ORDER,
+	PAYMENT_GATEWAY_SORT_ORDER,
 } from '@woocommerce/block-settings';
 
 /**
@@ -90,8 +90,8 @@ const usePaymentMethodRegistration = (
 			};
 		};
 
-		for ( let i=0; i<PAYMENT_GATEWAY_ORDER.length; i++  ) {
-			const paymentMethodName = PAYMENT_GATEWAY_ORDER[i];
+		for ( let i=0; i<PAYMENT_GATEWAY_SORT_ORDER.length; i++  ) {
+			const paymentMethodName = PAYMENT_GATEWAY_SORT_ORDER[i];
 			const paymentMethod = registeredPaymentMethods[ paymentMethodName ];
 			if ( ! paymentMethod ) {
 				continue;
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -6,7 +6,13 @@ import {
 	getPaymentMethods,
 	getExpressPaymentMethods,
 } from '@woocommerce/blocks-registry';
-import { useState, useEffect, useRef, useCallback, useMemo } from '@wordpress/element';
+import {
+	useState,
+	useEffect,
+	useRef,
+	useCallback,
+	useMemo,
+} from '@wordpress/element';
 import {
 	useEditorContext,
 	useShippingDataContext,
@@ -149,6 +155,6 @@ export const usePaymentMethods = ( dispatcher ) => {
 		return ordered;
 	}, [ paymentMethods, PAYMENT_GATEWAY_SORT_ORDER ] );
 	return usePaymentMethodRegistration( dispatcher, orderedPaymentMethods );
-}
+};
 export const useExpressPaymentMethods = ( dispatcher ) =>
 	usePaymentMethodRegistration( dispatcher, getExpressPaymentMethods() );
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -7,12 +7,7 @@ import {
 	getPaymentMethods,
 	getExpressPaymentMethods,
 } from '@woocommerce/blocks-registry';
-import {
-	useState,
-	useEffect,
-	useRef,
-	useCallback,
-} from '@wordpress/element';
+import { useState, useEffect, useRef, useCallback } from '@wordpress/element';
 import {
 	useEditorContext,
 	useShippingDataContext,
@@ -137,7 +132,12 @@ const usePaymentMethodRegistration = (
 		// Example: Stripe CC, Stripe Payment Request.
 		// That's why we track "is initialised" state here.
 		setIsInitialized( true );
-	}, [ dispatcher, isEditor, registeredPaymentMethods, paymentMethodsSortOrder ] );
+	}, [
+		dispatcher,
+		isEditor,
+		registeredPaymentMethods,
+		paymentMethodsSortOrder,
+	] );
 
 	// Determine which payment methods are available initially and whenever
 	// shipping methods change.
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -6,7 +6,7 @@ import {
 	getPaymentMethods,
 	getExpressPaymentMethods,
 } from '@woocommerce/blocks-registry';
-import { useState, useEffect, useRef, useCallback } from '@wordpress/element';
+import { useState, useEffect, useRef, useCallback, useMemo } from '@wordpress/element';
 import {
 	useEditorContext,
 	useShippingDataContext,
@@ -90,12 +90,8 @@ const usePaymentMethodRegistration = (
 			};
 		};
 
-		for ( let i = 0; i < PAYMENT_GATEWAY_SORT_ORDER.length; i++ ) {
-			const paymentMethodName = PAYMENT_GATEWAY_SORT_ORDER[ i ];
+		for ( const paymentMethodName in registeredPaymentMethods ) {
 			const paymentMethod = registeredPaymentMethods[ paymentMethodName ];
-			if ( ! paymentMethod ) {
-				continue;
-			}
 
 			// In editor, shortcut so all payment methods show as available.
 			if ( isEditor ) {
@@ -139,7 +135,20 @@ const usePaymentMethodRegistration = (
 	return isInitialized;
 };
 
-export const usePaymentMethods = ( dispatcher ) =>
-	usePaymentMethodRegistration( dispatcher, getPaymentMethods() );
+export const usePaymentMethods = ( dispatcher ) => {
+	const paymentMethods = getPaymentMethods();
+	const orderedPaymentMethods = useMemo( () => {
+		const ordered = [];
+		for ( let i = 0; i < PAYMENT_GATEWAY_SORT_ORDER.length; i++ ) {
+			const paymentMethodName = PAYMENT_GATEWAY_SORT_ORDER[ i ];
+			const paymentMethod = paymentMethods[ paymentMethodName ];
+			if ( paymentMethod ) {
+				ordered.push( paymentMethods[ paymentMethodName ] );
+			}
+		}
+		return ordered;
+	}, [ paymentMethods, PAYMENT_GATEWAY_SORT_ORDER ] );
+	return usePaymentMethodRegistration( dispatcher, orderedPaymentMethods );
+}
 export const useExpressPaymentMethods = ( dispatcher ) =>
 	usePaymentMethodRegistration( dispatcher, getExpressPaymentMethods() );
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies
  */
+import { union } from 'lodash';
 import { __, sprintf } from '@wordpress/i18n';
 import {
 	getPaymentMethods,
@@ -11,7 +12,6 @@ import {
 	useEffect,
 	useRef,
 	useCallback,
-	useMemo,
 } from '@wordpress/element';
 import {
 	useEditorContext,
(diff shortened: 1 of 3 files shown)


src/Payments/Api.php
```diff
@@ -80,8 +80,12 @@ public function add_payment_method_script_dependencies( $dependencies, $handle )
 	 * Add payment method data to Asset Registry.
 	 */
 	public function add_payment_method_script_data() {
-		$script_data = $this->payment_method_registry->get_all_registered_script_data();
+		// Enqueue the order of enabled gateways as `paymentGatewayOrder`.
+		$available_gateways = WC()->payment_gateways->get_available_payment_gateways();
+		$this->asset_registry->add( 'paymentGatewayOrder', array_keys( $available_gateways ) );
 
+		// Enqueue all registered gateway data (settings/config etc).
+		$script_data = $this->payment_method_registry->get_all_registered_script_data();
```
haszari, line 88: We could potentially tweak the structure of the data to combine order and config. So each gateway's script data is would be nested under the gateway id, in a single key on wcSettings. For example:

```js
wcSettings [ 
   bacs: {
      accountName: 'Nike.com bank account', 
      ...
   },
   stripe: {
   } 
]
```

In this PR I've just added a new key for the order to minimise code impact :)

mikejolley:
Made a comment here https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2157#issuecomment-665606856 but reviewing now :)

assets/js/settings/blocks/constants.js
```diff
@@ -64,6 +64,8 @@ export const SHIPPING_METHODS_EXIST = getSetting(
 	false
 );
 
+export const PAYMENT_GATEWAY_ORDER = getSetting( 'paymentGatewayOrder', [] );
```
mikejolley, line 67: Can we avoid using "order"? It confuses with "orders". We should try using "sorting" where possible because it's clearer of intent.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -86,8 +89,13 @@ const usePaymentMethodRegistration = (
 				[ paymentMethod.name ]: paymentMethod,
 			};
 		};
-		for ( const paymentMethodName in registeredPaymentMethods ) {
+
+		for ( let i=0; i<PAYMENT_GATEWAY_ORDER.length; i++  ) {
```
mikejolley, line 93: If we managed to get the sorting value in the registered payment methods config we could pre-sort these using something like https://lodash.com/docs/4.17.15#sortBy and avoid changing these loops.

assets/js/settings/blocks/constants.js
haszari: I went with `paymentGatewaySortOrder`. Totally unfortunate that we have the clash with `order`!

`paymentGatewaySorting` didn't seem quite right to me, since it's not a sort type (e.g. alpha, etc), it's a manual reordering. 

If we do move this into the config then maybe we could represent as `weight` or `displayPriority` param on each gateway data blob, and we wouldn't have this naming issue :)

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -86,8 +89,13 @@ const usePaymentMethodRegistration = (
 				[ paymentMethod.name ]: paymentMethod,
 			};
 		};
-		for ( const paymentMethodName in registeredPaymentMethods ) {
+
+		for ( let i=0; i<PAYMENT_GATEWAY_SORT_ORDER.length; i++  ) {
+			const paymentMethodName = PAYMENT_GATEWAY_SORT_ORDER[i];
```
nerrad, line 94: I liked Mike's idea of trying to preserve order by registration client side (via `registerPaymentMethod`) and I experimented a bit but the logic ends up being more complicated than what we have here (for little gain) so I think doing the loop here is sufficient. The only concern I have here though is that effectively whether a payment method shows up or not is if it's name is present in this `PAYMENT_GATEWAY_SORT_ORDER` setting. Are we sure that this setting is reliable enough? From the usage, I think it's fair to assume it is?

Also, there's some formatting issues with these lines:

```suggestion
		for ( let i = 0; i < PAYMENT_GATEWAY_SORT_ORDER.length; i++ ) {
			const paymentMethodName = PAYMENT_GATEWAY_SORT_ORDER[ i ];
```

nerrad, line 94: There's a couple other things to consider here:

1. Express payment methods go through this logic. 

With the current loop, I think this means that some of them might not get registered. As an example, I don't think chrome pay and applePay will show up because `payment_request` is the registered name for that payment method and that identifier won't be in the `PAYMENT_GATEWAY_SORT_ORDER` array.  Currently, I'm also not aware of any ordering mechanism for Express Payment Methods in Woo Settings.

2. I think it might be good to extract the ordering out into the `usePaymentMethods` hook exported in this file.

This way the ordering only happens once based on the `PAYMENT_GATEWAY_SORT_ORDER` (you could memoize it). Then you can leave the api the way it was originally in `usePaymentMethodRegistration` and it takes care of the current problem for express payment methods.

You may need to cast the registered payment methods object (for both express payment methods and payment methods) to a Map in the process (to ensure you are preserving order). 

For example:

```js
export const usePaymentMethods = ( dispatcher ) => {
	const paymentMethods = getPaymentMethods();
	const orderedPaymentMethods = useMemo( () => {
		const ordered = new Map();
		for ( let i = 0; i < PAYMENT_GATEWAY_SORT_ORDER.length; i++ ) {
			const paymentMethodName = PAYMENT_GATEWAY_SORT_ORDER[ i ];
			const paymentMethod = paymentMethods[ paymentMethodName ];
			if ( paymentMethod ) {
				ordered.set( paymentMethodName, paymentMethods[ paymentMethodName ] );
			}
		}
	}, [ paymentMethods ] );
	return usePaymentMethodRegistration( dispatcher, orderedPaymentMethods );
}
```

Then `usePaymentMethodRegistration` would have to handle looping through the Map (which is also why you might need to convert Express payment methods to a map too). 

Or you could just convert to an array of Payment Methods as a part of the ordering methods and pass that through to `usePaymentMethodRegistration` (and just pass through `Object.values( getExpressPaymentMethods()` as the argument from `useExpressPaymentMethods`). The array would probably be simpler to work with and still would preserve order.

haszari: Thanks for the formatting fix, I'm not sure why this isn't working on commit for me. The hook seems to be running after commit, or it's using the wrong config – I see formatting-only diffs after commit! Will figure that out separately.

> The only concern I have here though is that effectively whether a payment method shows up or not is if it's name is present in this PAYMENT_GATEWAY_SORT_ORDER setting. Are we sure that this setting is reliable enough? From the usage, I think it's fair to assume it is?

Good point - yes we are relying on that. I think that's a safe assumption, it's powered by the same code that used in traditional checkout (`WC()->payment_gateways->get_available_payment_gateways()`). If for some reason a method is lost or the value is empty/missing, it will cause a clear UI bug (no/missing payment methods!) so we should catch that. 

We could put in protection for this, but I don't think that's worthwhile.

haszari: Right, that seems more robust, and also good to separate out from the registration loop (which is quite a bit of code). Thanks - will get on that :)

haszari: > With the current loop, I think this means that some of them might not get registered. As an example, I don't think chrome pay and applePay will show up

Just tested - good catch, this broke Apple Pay for me 🤕

haszari: Started implementing this in db32784e405cafe9d05205331da0d634f7d5a5cb, using an array (rather than map) ... then realised that I need to change the interface to `usePaymentMethodRegistration` to take array, and fix up express methods too (as you noted). 

It's now infinite-looping – warning for anyone who checks out this branch!. I'm guessing because of the dispatch of available payment methods in `refreshCanMakePayments` - this ultimately changes `registeredPaymentMethods` (I'm guessing here). So probably need to move the reordering code further out, or do it "late" deep in `refreshCanMakePayments` as before.

haszari: OK got it sorted. I've changed `usePaymentMethodRegistration` so it also takes an order param, and I'm passing this in for express and standard.

There's an extra complexity - if COD is dependent on shipping, and a non-COD-compatible shipping method is selected on page load, then COD is not present in the `PAYMENT_GATEWAY_SORT_ORDER` setting. So to handle that, I'm merging the order and all keys using lodash `union`. So you were right @nerrad - we do need to handle methods not being present in that array. 

So with that all in mind, this definitely needs a new review and thorough testing – it would be great if we can find a more elegant way of handling this. 

We may also want to fix `PAYMENT_GATEWAY_SORT_ORDER` setting server side so it always returns all methods. The fact that it's hiding COD for us is a sign that it's not the right fit for our needs.

haszari:
Thanks all for the thorough reviews. I've got this working correctly for both express and standard payment methods (i.e. I fixed the issue that broke express payment methods) and also should work correctly for dynamically-available methods (I think COD is the only example of this).

Please refer to [review comments](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2934#discussion_r462719271) above for more info. Since the initial round of reviews I introduced an infinite dependency loop so we need to take care here.

When reviewing, please feel free to suggest alternative approaches – the current code is hopefully robust but perhaps there's a more elegant fix that won't have dependency loop :)

haszari:
Note: there's a little ordering issue with COD: 

- COD is in correct position if COD-compatible shipping is selected on page load
- COD is shown at end if COD-incompatible shipping is selected on page load

That's not a major problem, but it might be a sign we want to do this differently.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -131,7 +149,22 @@ const usePaymentMethodRegistration = (
...
-export const usePaymentMethods = ( dispatcher ) =>
-	usePaymentMethodRegistration( dispatcher, getPaymentMethods() );
-export const useExpressPaymentMethods = ( dispatcher ) =>
-	usePaymentMethodRegistration( dispatcher, getExpressPaymentMethods() );
+export const usePaymentMethods = ( dispatcher ) => {
+	const standardMethods = getPaymentMethods();
+	// Ensure all methods are present in order.
+	// Some payment methods may not be present in PAYMENT_GATEWAY_SORT_ORDER if they
+	// depend on state, e.g. COD can depend on shipping method.
+	const displayOrder = union(
```
Aljullu, line 157: I didn't test it, but just leaving the suggestion here in case you think it makes sense: if you want to avoid using lodash, you can use a native [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) here. It will ensure items are not duplicated and they are in order.

```JSX
const displayOrder = new Set( [
	...PAYMENT_GATEWAY_SORT_ORDER,
	...Object.keys( standardMethods )
] );
```

Then, you will need to replace the [line above](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2934/files#diff-71ebb9a05002d896e445bb08e1294645R98) (`for ( let i = 0; i < paymentMethodsSortOrder.length; i++ ) { ... }`) with `paymentMethodsSortOrder.forEach( ... )`.

As I said, not a blocker at all. :slightly_smiling_face:

haszari:
Adding `blocked` label for now. In it's current state this branch can cause an infinite loop - `refreshCanMakePayments` gets repeatedly called, so the new reordering code is triggering some hook dependencies somewhere up the chain. For me this is happening no matter what on every full-cart page load, but it might depend on payment methods or shipping config.

@nerrad Any ideas/suggestions on what might be causing this, and how to tweak the logic to avoid this risk?

I'm thinking this _might_ be because we're dispatching payment methods to redux store in this callback, and that's causing the callback to fire again. We might be able to persist `canMakePayment` and/or order differently in the store (e.g. arrays or props) to make it more robust, and/or use debounce or different equivalence checks to prevent unnecessary firing of `refreshCanMakePayments`.

Aljullu:
@haszari I think that's because `paymentMethodsSortOrder` is an array that is created on every render so, when evaluating the effect dependencies, it's a different instance.

A quick solution would be to wrap it in `useShallowEqual`. This way, if the previous and the new array have the same values in the same order, it will use the previous instance and because of that the `useEffect()` would not be triggered.
```JSX
const currentPaymentMethodsSortOrder = useShallowEqual(
	paymentMethodsSortOrder
);
```

haszari:
Thanks all  - this is now working well. I've applied all your suggestions @Aljullu and I haven't been able to trigger an infinite render loop. So ready for one last review. Still open to wider discussion about the approach and how it fits in to payment method initialisation.

@nerrad you might want a look at this before merge. Specifically the fact that `wcSettings.paymentGatewaySortOrder` doesn't include COD if non-COD shipping method is selected, and the effects of that (COD not always shown in correct position). FYI :)

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
```diff
@@ -86,8 +94,13 @@ const usePaymentMethodRegistration = (
 				[ paymentMethod.name ]: paymentMethod,
 			};
 		};
-		for ( const paymentMethodName in registeredPaymentMethods ) {
+
+		for ( let i = 0; i < paymentMethodsOrder.length; i++ ) {
```
Aljullu, line 98: Nit: you can use `forEach` if you want:
```JS
await paymentMethodsOrder.forEach( async ( paymentMethodName ) => {
  ...
} );
```

Below, you would need to change the `continue` with `return`. Not blocking, just a suggestion in case you think it would make code cleaner. :slightly_smiling_face:

Aljullu, line 98: Nit: you can use `forEach` if you want:
```JS
await paymentMethodsOrder.forEach( async ( paymentMethodName ) => {
  ...
} );
```

Below, you would need to change the `continue` with `return`. Not blocking, just a suggestion in case you think it would make code cleaner. :slightly_smiling_face:

src/Payments/Api.php
```diff
@@ -80,8 +80,12 @@ public function add_payment_method_script_dependencies( $dependencies, $handle )
 	 * Add payment method data to Asset Registry.
 	 */
 	public function add_payment_method_script_data() {
-		$script_data = $this->payment_method_registry->get_all_registered_script_data();
+		// Enqueue the order of enabled gateways as `paymentGatewaySortOrder`.
+		$available_gateways = WC()->payment_gateways->get_available_payment_gateways();
```
Aljullu, line 84: > Note - COD is not always displayed in merchant-configured order

Could that be fixed if replacing `get_available_payment_gateways()` with `payment_gateways()`? The latter returns all payment methods, not only the available ones, so COD would be in the array even if not available when loading the block.

src/Payments/Api.php
```diff
@@ -80,8 +80,12 @@ public function add_payment_method_script_dependencies( $dependencies, $handle )
...
+		$this->asset_registry->add( 'paymentGatewaySortOrder', array_keys( $available_gateways ) );
```
nerrad, line 85: This should be wrapped in a check for whether this is already registered:

```suggestion
		if ( ! $this->asset_registry->exists( 'paymentGatewaySortOrder' ) ) {
			$this->asset_registry->add( 'paymentGatewaySortOrder', array_keys( $available_gateways ) );
		}
```

Otherwise you can get a thrown exception if the key already exists (which happened to me while trying to test).

haszari: Interesting, I'll add the check. Curious why this code would be firing multiple times though - is this expected?

haszari: Ah I see the other assets are protected by a check, missed that. Added check here: 16139feb4f37a8cd30ed164c38bbf3129b2c7ea8

src/Payments/Api.php
haszari: Good thinking! I've added that here 88b35a4cb85b108680915937eab3284a68e2ac20 and it seems to work well. 

One risk with this is that it adds every gateway to the list, even if they are disabled by the merchant:

Screen Shot 2020-08-05 at 11 04 31 AM

Screen Shot 2020-08-05 at 11 03 31 AM

That's probably fine, just a little inefficient. I can't imagine a merchant would want to protect this information (what gateway extensions are installed) but if so this could be considered a leak :)

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
haszari: This breaks! (Just tried it.) The [`foreach` doesn't wait for the async callback](https://advancedweb.hu/how-to-use-async-functions-with-array-foreach-in-javascript/), so nothing gets registered. 

I could switch it to `map` and `Promise.all`, not sure if that's clearer than the traditional indexed for loop. Thoughts @nerrad  @Aljullu ? Happy to change to `map` if that's more appropriate or more clear.

nerrad: Interesting! I'm fine with leaving the code as is. The for loop is clearer to me than using Array.map with Promise.all would be here.

src/Payments/Api.php
nerrad: Good observation, but I don't think this is an information leak. What actionable things can be done with this that would be harmful to merchants?

haszari: Off hand I can't think of any :)

haszari:
Merging :)

Added typedefs to protect against issues like 7653232. @nerrad how can we run type check locally, is this included in `lint:js`?

PR #2940, opened 2020-07-29
Fix reviews blocks in WP 5.2

Aljullu:
While testing 3.1.0, I noticed reviews blocks were not working on WP 5.2. That seems to be because `renderFrontend()` relies on `Suspense`, which is not available in the React version of WP 5.2.

This PR replaces Suspense with a noop component if it's not available.

1. In a WP 5.2 install, create a page with the All Reviews block, view it in the frontend and verify the block shows correctly.
2. In dev build of `main`, verify there are no regressions with the All Products and Single Product block.

This regression was introduced after 3.0 was released, so I don't think we need to add anything to the changelog.

### DIFF
--- a/assets/js/base/utils/render-frontend.js
+++ b/assets/js/base/utils/render-frontend.js
@@ -22,6 +22,7 @@ export const renderFrontend = ( {
 	const containers = document.querySelectorAll( selector );
 
 	if ( containers.length ) {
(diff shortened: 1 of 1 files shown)


nerrad:
I forgot to also mention that All Products should be included for testing notes as innerblocks are lazy loaded too.

PR #2944, opened 2020-07-30
Add inspector notice for unsupported product types

### LINKED ISSUE #2857: ⚛️ ‎‎‏‏‎ Add to Cart - create fallback or support for non-core product types
Non-core product types—create method to detect if unsupported and show a fallback (possibly a link to the woo core product page, or investigate something more fancy (SSR?).

mikejolley:
This PR adds a notice element if a product type is not supported by the add to cart form.

Products which support form elements have the toggle in the inspector like this:

Edit Page ‹ one wordpress test — WordPress 2020-07-30 14-38-25

Products which are unsupported (external, and currently grouped) instead show a notice:

Edit Page ‹ one wordpress test — WordPress 2020-07-30 14-39-00

I've added an inline todo for the follow up which is to introduce a system of registering support for a product type.

Closes #2857

In #2857 I proposed an experiment/idea to use SSR for fallback forms if the product type is unsupported. This did not work unfortunately. This is what I tried:

**1) SSR to load the cart form from the API**

SSR did work in rendering the form, eventually, but I faced many issues:

- Templates are not available on REST endpoints, so I had to manually include woo core template hooks + files
- Whilst the forms did render, scripts and other assets were missing
- Because SSR is an ajaxified request, when the form does eventually render, any scripts that handle add to cart functionality have already fired on `jquery.ready()`, so the forms are non-functional.

We did talk about perhaps deferring the load of the assets, but we realised 3rd party product types would be completely out of our control.

**2) Rendering the template directly, or using a shortcode**

I tired rendering the template to the page directly instead, in the hopes that the jquery events would still fire. But this was also problematic:

- The events were inconsistent. Depending on where the render occured the form was non-functional.
- Getting the form in the correct position meant rendering it within a specific atomic block. These blocks had no product ID context and so would need it saved to attributes. Even then, it didn't render in the correct position.

Since these methods failed, I've gone with a notice instead. Non supported types will link through to the product pages.

1. Add the single product block to a page
2. Select a supported type (simple or variable) - check inspector controls
3. Support a non-supported type (external or grouped) - check notice in inspector controls
4. Repeat for a 3rd party product type such as bookings. It should show a link and the notice in the inspector.

### DIFF
--- a/assets/js/atomic/blocks/product-elements/add-to-cart/shared/add-to-cart-button.js
+++ b/assets/js/atomic/blocks/product-elements/add-to-cart/shared/add-to-cart-button.js
@@ -14,6 +14,8 @@ import { useStoreAddToCart } from '@woocommerce/base-hooks';
 const AddToCartButton = () => {
 	const {
 		showFormElements,
+		productIsPurchasable,
+		productHasOptions,
 		product,
 		isDisabled,
 		isProcessing,
@@ -23,8 +25,6 @@ const AddToCartButton = () => {
 	} = useAddToCartFormContext();
 	const { cartQuantity } = useStoreAddToCart( product.id || 0 );
 	const [ addedToCart, setAddedToCart ] = useState( false );
-	const isPurchasable = product.is_purchasable;
-	const hasOptions = product.has_options;
 	const addToCartButtonData = product.add_to_cart || {
 		url: '',
 		text: '',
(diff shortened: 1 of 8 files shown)


assets/js/atomic/blocks/product-elements/add-to-cart/shared/add-to-cart-button.js
```diff
@@ -47,23 +48,30 @@ const AddToCartButton = () => {
...
-	}
+	/**
+	 * We can show a real button if we are:
+	 *
+	 *  	a) Showing a full add to cart form.
+	 * 		b) The product doesn't have options and can therefore be added directly to the cart.
+	 * 		c) The product is purchasable.
+	 *
+	 * Otherwise we show a link instead.
+	 */
```
haszari, line 59: ❤️ big clear intentional comments like this :)

assets/js/base/utils/product-data.js
```diff
@@ -0,0 +1,31 @@
...
+	 * When introducing the form-element registration system, include a method of defining if a
+	 * product type has support.
+	 *
+	 * If, as an example, we went with an inner block system for the add to cart form, we could allow
+	 * a type to be registered along with it's default Block template. Registered types would then be
+	 * picked up here, as well as the core types which would be defined elsewhere.
+	 */
+	const supportedTypes = [ 'simple', 'variable' ];
+
+	return supportedTypes.includes( product.type || 'simple' );
```
haszari, line 30: Should this default to `simple` if `type` is broken/missing? I think this should probably be more paranoid / explicit - if type is present and one of the supported values then add to cart is available, otherwise default to no support.

PR #2949, opened 2020-07-31
Fixes related to Jetpack Autoloader 2.0 package bump

### LINKED ISSUE #2945: Fix jetpack autoloader configuration and setup and deploy process
With Jetpack Autoloader bump to 2.0 there's some unintended side-effects which if not addressed means the blocks package included in Woo Core will always override the active feature plugin. 

Here's a shortlist of things that need to be done:

- on publishing a release, `composer.json` needs to have an explicit version added to it, and then `composer dump-autoload` to regenerate the autoloader with the correct classmap version strings. This ensures active feature plugin overrides lower version classmaps in Woo Core.
- `composer.json` should _not_ have a version constant defined otherwise, to ensure that the classmaps are generated with at `dev-` prefix indicating they are a developer version.
- developers need to define the `JETPACK_AUTOLOAD_DEV` constant in their `wp-config.php` file for local development of a checked out copy of the plugin so that development always overrides what's in core (this may be addressed upstream, but for now this is a big change for local development). We could add a new check in the plugin main file to determine if we're in a development environment and the constant isn't set to show a notice with a warning (similar to how we do for non-built files on a checkout). Any related developer documentation around this requirement needs updated.

nerrad:
Fixes #2945 

See #2945 for details and rationale.

### Testing version bump changes.

In this pull the `./bin/version-changes.sh` script was modified to add an `"version"` entry to `composer.json` for the version being bumped to.

To verify things work as expected, you can run this script locally:

```bash
VERSION=3.0.0 sh ./bin/version-changes.sh
```

After running that you should note the following:

- `composer.json`, `package.json`, `readme.txt`, `src/Package.php`, `woocommerce-gutenberg-products-block.php` should be changed.
- The above files should only have a version bump to the specified version (in this case `3.0.0`)
- `composer.json`, should have the `version` property added by the script (otherwise it exists without a `version` property in development).

### Testing missing development constant notice

This will require a bit additional setup:

1. Make sure you are running WooCommerce 4.4 beta 1 (or checked out the built release branch for that version - it must be that version, not beta2).
2. Make sure you are running the latest WP 5.5 release candidate.
3. Without doing anything in this branch, you should see the following notice:

Image 2020-07-31 at 12 09 51 PM

4. Go to the frontend of your site and load a page that has one of our blocks that pulls from the REST API. You should see a bunch of deprecation notices related to our endpoints. This is an indicator that the package from Woo Core is being run rather than the classes from this branch (which have the REST API fixes).
5. Add `define( 'JETPACK_AUTOLOAD_DEV', true )` to your `wp-config.php` file (you'll want to leave it defined there when developing Woo Blocks going forward). **Note:** You _may_ get a fatal error because your checked out version of Woo Core has the constant defined, in that case just comment out the constant definition in Woo Core (Woo core 4.4. will not be released with the defined constant as it's being removed).
6. Verify that the notice disappears in your wp-admin dashboard. Also the deprecated REST API notices should disappear as well.

### Testing deploy

This is a bit trickier because we don't actually want to deploy. What I do is just add an exit right after this line in `bin/github-deploy.sh`

```
output 2 "Committing version change..."
echo
```
Essentially make sure you either comment out all `git` related commands in the script or do an exit before they run.

Then I'll do `sh ./bin/github-deploy.sh` and after it executes, take a look at the `vendor/composer/jetpack_autoload_classmap.php` file and look for any of the plugin classes defined in there. There should be a `'version' => '3.0.0.0',` string in the map that matches whatever version you entered during the deploy script (the extra `.0` appended is expected).

### DIFF
--- a/bin/github-deploy.sh
+++ b/bin/github-deploy.sh
@@ -107,8 +107,10 @@ if [ "$(echo "${PROCEED:-n}" | tr "[:upper:]" "[:lower:]")" != "y" ]; then
 fi
 
 # Version changes
-output 2 "Updating version numbers in files (note pre-releases will not have the readme.txt stable tag updated)..."
+output 2 "Updating version numbers in files and regenerating php autoload classmap (note pre-releases will not have the readme.txt stable tag updated)..."
 source $RELEASER_PATH/bin/version-changes.sh
+
+composer dump-autoload
 
 output 2 "Committing version change..."
 echo
--- a/bin/version-changes.sh
+++ b/bin/version-changes.sh
@@ -17,3 +17,6 @@ perl -i -pe 's/"version":*.+/"version": "'${VERSION}'",/' package.json
 
 # Update version in src/Package.php
 perl -i -pe "s/version \= '*.+';/version = '${VERSION}';/" src/Package.php
+
+# Add version to composer.json
+perl -i -pe 's/"type":*.+/"type":"wordpress-plugin",\n\t"version": "'${VERSION}'",/' composer.json
--- a/woocommerce-gutenberg-products-block.php
+++ b/woocommerce-gutenberg-products-block.php
@@ -10,7 +10,7 @@
  * Requires at least: 5.2
  * Requires PHP: 5.6
  * WC requires at least: 4.0
- * WC tested up to: 4.2
+ * WC tested up to: 4.4
  *
  * @package WooCommerce\Blocks
  * @internal This file is only used when running as a feature plugin.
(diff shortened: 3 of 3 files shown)


nerrad:
I don't think we need to do a patch release for this for the following reasons:

- If Woo Core 4.4 is released before this feature plugin and someone has 3.1.0 installed and active, what's in Woo Core 4.4 is identical so it's okay that Woo Core classes are loaded.
- For our local development we can develop against Woo Core _now_ (even without this branch) by just adding the constant to our `wp-config.php`.
- Releasing this as a part of `3.2.0` will allow the feature plugin to override what's in Woo Core (given Woo Core has an older version of the package installed).

senadir:
skipping this in the changelog.

PR #2952, opened 2020-08-03
Simplify reviews render method

Aljullu:
I originally thought I needed to do a deprecation of the Reviews blocks in #2904, so ended up cleaning a bit the render logic to make my life easier. In the end, there is no need for the deprecation, but I still think the refactor is worth merging, so I split it into this PR. The changes include:
1. The logic to know which class name to use (`wc-block-all-reviews`, `wc-block-reviews-by-product`, `wc-block-reviews-by-category`) is now in `getBlockClassName`. Previously, those classes were explicitly defined in the editor but inferred from `attributes` in the frontend which I found a bit confusing. With the changes in this PR they are always inferred from `attributes` inside `getBlockClassName`, so there is a single source of truth.
2. I moved the logic to calculate the `data` attributes to its own util function (`getDataAttrs`) so it's easier to reuse if we ever write deprecations and makes code more organized.

1. Still in `main`, add the three reviews blocks (All Reviews, Reviews by Product, Reviews by Category) to a page.
2. Checkout this branch.
3. Verify the previously added blocks didn't invalidate.
4. Verify the blocks still work fine in the frontend.

### DIFF
--- a/assets/js/blocks/reviews/all-reviews/edit.js
+++ b/assets/js/blocks/reviews/all-reviews/edit.js
@@ -62,7 +62,6 @@ const AllReviewsEditor = ( { attributes, setAttributes } ) => {
 			{ getInspectorControls() }
 			<EditorContainerBlock
 				attributes={ attributes }
-				className="wc-block-all-reviews"
 				icon={
 					<Icon
 						icon={ discussion }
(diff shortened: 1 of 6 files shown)


Aljullu:
> One tiny comment - does this need a changelog entry? Since there's no issue/impact that this fixes, and it's a low risk refactor, I'd say doesn't need one, but maybe safer to mention it. (Happy either way, keen to get aligned on how we use release changelog.)

Probably not, I just removed it!

PR #2962, opened 2020-08-04
Fix undefined variable in ProductSchema.php closure 

### LINKED ISSUE #2896: Fix undefined variable `$product` in ProductSchema.php
I saw these notices happening in my php logs:

```
 PHP Notice:  Undefined variable: product in /srv/www/wordpress-one/public_html/wp-content/plugins/woocommerce-gutenberg-products-block/src/StoreApi/Schemas/ProductSchema.php on line 561
```

It looks like the closure needs to pass through the `$product` variable via the `use` statement.

senadir:
Fixes the issue by defining `use ( $product )` in the closure, as far as my googling went, this syntax is supported from PHP 5.3, so we're safe (WC requires 7.0).

Fixes #2896

1. You shouldn't see the notice anymore

### How to test
0- Make sure you have WP_DEBUG set to true.
1- Load a page that already contains a product data, so single product or All Products, either in the editor or frontend.
2- Make sure no notices are printed to the page, you can check the source code or at the top of the page.

### Changelog

> Fix an undefined variable PHP notice related to Product REST API.

### DIFF
--- a/src/StoreApi/Schemas/ProductSchema.php
+++ b/src/StoreApi/Schemas/ProductSchema.php
@@ -555,7 +555,7 @@ class ProductSchema extends AbstractSchema {
 		$attributes                  = array_filter( $product->get_attributes(), [ $this, 'filter_variation_attribute' ] );
 		$default_variation_meta_data = array_reduce(
 			$attributes,
-			function( $defaults, $attribute ) {
+			function( $defaults, $attribute ) use ( $product ) {
 				$meta_key              = wc_variation_attribute_name( $attribute->get_name() );
 				$defaults[ $meta_key ] = [
 					'name'  => wc_attribute_label( $attribute->get_name(), $product ),


assets/js/blocks/products/utils.js
```diff
@@ -35,7 +35,7 @@ export const renderNoProductsPlaceholder = ( blockTitle, blockIcon ) => (
 			className="wc-block-products__add-product-button"
 			isDefault
 			isLarge
-			href={ adminUrl + 'post-new.php?post_type=product' }
+			href={ ADMIN_URL + 'post-new.php?post_type=product' }
```
nerrad, line 38: wowzer! good catch here!

senadir: this was from another PR, not sure how it got here 😅, props to Albert.

senadir: this was from another PR, not sure how it got here 😅, props to Albert.

nerrad:
Tested okay, I also verified that there was no impact to variation meta data returned as a result of this change.

Issue #2628, opened 2020-06-04
Checkout: update design of the payment methods


Aljullu:
Related p2: pbIy4N-g9-p2

In order to offer better support for dark themes and to handle better the case when there are >2 payment methods, we want to redesign the payment methods section.

One payment method:
<img width="512" alt="screen-shot-2020-05-27-at-12 25 46-pm" src="https://user-images.githubusercontent.com/3616980/83740907-c672cf80-a657-11ea-8267-e241e3e1d623.png">

Two payment methods:
<img width="512" alt="screen-shot-2020-05-27-at-12 28 13-pm" src="https://user-images.githubusercontent.com/3616980/83740755-90355000-a657-11ea-8597-11f289ad2f5f.png">

Three or more payment methods:
<img width="505" alt="screen-shot-2020-05-27-at-12 36 12-pm" src="https://user-images.githubusercontent.com/3616980/83740775-94fa0400-a657-11ea-8297-7b36b50d956e.png">

haszari:
Linking in #2157 - using merchant-defined order for payment methods, as it could be closely related.

haszari:
Started looking at this …

I see we're using and customising/styling [Reakit Tabs component](https://reakit.io/docs/tab/) for these tabs. With the new design, we're only using "Tabs" for the 1 or 2 tab case, otherwise the component is more like an accordion. Reakit doesn't have an accordion. We'll likely want to share lots of styles between each UI mode, and behaviour too (e.g. keyboard selection etc).

I'm wondering if it's best to continue using Reakit tabs vs making a custom control that handles both modes. 

Also I'm curious how much our current Tabs component is built specifically for payment methods. I don't think it's used for anything else yet, and we've customised the styling quite a bit.

Aljullu:
I'm not sure how the `Tabs` component works and how easy it would be to customize it so it looks like an accordion. It would be great to always use the same component no matter how many payment methods are rendered, but I acknowledge that might be too complex and it might make no sense to over-complicate a component to make it look like something else.

In case we need to create an accordion component from scratch, it might be a good idea if it internally used the [`Panel` component](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/base/components/panel/index.js). It allows having a title and an expandable area underneath, which seems quite similar to what we need for the accordion. We would need to make sure only one of the panels is open at the same time, but besides that it might be a good fit?

haszari:
[...]

PR #2968, opened 2020-08-05
Remove generic payment method icons (Stripe CC + Cheque)

### LINKED ISSUE #2669: Fix payment icon layout issue with Twenty Twenty theme (non-mobile layout)
While testing 2.7.0, I noticed that with Twenty Twenty, the payment method icons for payment methods are incorrectly positioned:

They should be on the same line as the text content.  Since they are at least left aligned, this isn't a huge issue but I think it would still be good to be consistent across themes for this.

haszari:
Fixes #2669 

This fixes a couple of issues by removing the generic icons for some payment methods.

- Based on previous discussion, we're [moving away from generic payment method icons (e.g. a card icon)](https://github.com/Automattic/woocommerce-payments/pull/723#discussion_r455190773); icons/imagery should be used for recognisable payment brands only (e.g. PayPal). cc @LevinMedia – can you confirm this is the current recommendation?
- #2669 is a specific wrap issue with Twenty Twenty theme. This is not an issue when the icons are not present.

The `PaymentMethodLabel` component now no longer needs support for icons - so I've removed this. We might simplify this further - remove the component and allow payment methods to provide a string or custom component for their label. cc @nerrad - this is an extensibility API change

### Screenshots

Screen Shot 2020-08-05 at 4 41 29 PM

1. Set up Stripe CC (Stripe extension) and Check (core) payment methods.
2. Set up a page with checkout block, add stuff to cart.
3. View checkout and confirm payment method tabs render and function correctly.

### Changelog

> Removed generic icons for Check and Stripe Credit Card to reduce visual clutter in Checkout block.

May need a dev note - this changes the API/docs for implementing a payment method - specifically the `PaymentMethodLabel` interface has changed.

### DIFF
--- a/assets/js/base/components/cart-checkout/payment-method-label/index.js
+++ b/assets/js/base/components/cart-checkout/payment-method-label/index.js
@@ -1,18 +1,44 @@
 /**
+ * External dependencies
+ */
+import classnames from 'classnames';
+import { Icon, bank, bill, card, checkPayment } from '@woocommerce/icons';
+
+/**
  * Internal dependencies
  */
 import './style.scss';
+
+const namedIcons = {
+	bank,
+	bill,
+	card,
+	checkPayment,
+};
 
 /**
  * Exposed to payment methods for the label shown on checkout. Allows icons to be added as well as
  * text.
  *
  * @param {Object} props Component props.
+ * @param {*} props.icon Show an icon beside the text if provided. Can be a string to use a named
+ *                       icon, or an SVG element.
  * @param {string} props.text Text shown next to icon.
  */
-export const PaymentMethodLabel = ( { text = '' } ) => {
-	const className = 'wc-block-components-payment-method-label';
-	return <span className={ className }>{ text }</span>;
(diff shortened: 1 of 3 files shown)


nerrad:
Chatted with @LevinMedia in slack (p1597154267257400-slack-C8X6Q7XQU) and here's a summary:

- confirm that generic icons get removed
- please **keep** the icon attribute as it will still be needed for things like paypal etc that will get applied in future pulls around payment method select structure.

haszari:
Summary of slack discussion:

- Although generic icons are no longer recommended for payment methods, well-known payment brand icons (e.g. PayPal) will be supported.
- The layout and UI for multiple payment methods will change in a future PR (#2628).
- As part of this change, brand icons will need to be explicitly supported, so that the blocks plugin has control of how they are displayed (e.g. where on tab/radio button etc).
- So we are keeping `PaymentMethodLabel.icon` prop as it may be part of this solution.

senadir:
We no longer need a dev note for this.

PR #2974, opened 2020-08-06
Fix focus styles of error states for the text input and select components

### LINKED ISSUE #2182: Checkout block with theme TwentyTwenty: focused invalid fields should have a red border
**To Reproduce**
1. Switch to 2020 theme and open the _Checkout_ block.
2. Make some input invalid (for example, making it blank) and unfocus + focus it.
3. Notice the border of the input field is black.

_Firefox:_
image

_Chrome:_
imatge

Screenshots from @senadir.

**Expected behavior**
The border should be red, as it is when unfocused.

### LINKED ISSUE #2973: Twenty Twenty: Error form elements have different styles across browsers when focused
**To Reproduce**
1. Switch your theme to Twenty Twenty.
2. Go to the Checkout block and press 'Place order` without filling any form detail.
3. Tab through the input fields under 'Shipping address' and look at the outline that appears on focus.

_Chrome:_
imatge

_Firefox:_
imatge

Notice that 'City' fields are focused and have different outline color between Chrome and Firefox. That seems to be because outline color behaves inconsistently across browsers. As an example, the [`outline-style` MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/outline-style#Values) looks different in Chrome and Firefox:
_Chrome:_
imatge
_Firefox:_
imatge
That's because with `outline-style: auto`, Firefox doesn't honor the `outline-color` property (and it seems to be [in purpose](https://bugzilla.mozilla.org/show_bug.cgi?id=1031664#c13)). For us, that means we can't just change the `outline-color` property and assume it will render red, because if the theme didn't specify an `outline-style` property, the color will be ignored.

Aljullu:
To solve the issues above, this PR includes this changes:
1. Increase the specificity of the `border-color` declaration (fixes #2182).
2. Define the `outline-style` property for focus states (fixes #2973).

#### Storefront

_Before (Chrome):_
imatge

_After (Chrome):_
imatge

_Before (Firefox):_
imatge

_After (Firefox):_
imatge

#### Twenty Twenty
_Before (Chrome):_
imatge

_After (Chrome):_
imatge

_Before (Firefox):_
imatge

_After (Firefox):_
imatge

1. Go to the Checkout block and press 'Place order` without filling any form detail.
2. Tab through the input fields under 'Shipping address' and verify the outline that appears on focus is red in text inputs and selects.
3. Repeat the steps above at least with Storefront and Twenty Twenty and with Firefox, Chrome and Safari.

### Changelog

> Improved focus styles of error states on form elements.

### DIFF
--- a/assets/js/base/components/text-input/style.scss
+++ b/assets/js/base/components/text-input/style.scss
@@ -64,9 +64,15 @@
 	}
 
 	&.has-error input {
-		border-color: $error-red;
+		&,
+		&:hover,
+		&:focus,
+		&:active {
+			border-color: $error-red;
+		}
 		&:focus {
-			outline-color: $error-red;
+			outline: 1px dotted $error-red;
+			outline-offset: 2px;
 		}
 	}
 
--- a/assets/js/base/components/select/style.scss
+++ b/assets/js/base/components/select/style.scss
@@ -30,7 +30,16 @@
 
 	&.has-error {
 		.components-custom-select-control__button {
-			border-color: $error-red;
+			&,
+			&:hover,
+			&:focus,
+			&:active {
+				border-color: $error-red;
+			}
+			&:focus {
+				outline: 1px dotted $error-red;
+				outline-offset: 2px;
+			}
 		}
 	}
 


assets/js/base/components/text-input/style.scss
```diff
@@ -64,9 +64,15 @@
...
+		&,
+		&:hover,
+		&:focus,
+		&:active {
+			border-color: $error-red;
+		}
 		&:focus {
-			outline-color: $error-red;
+			outline: 1px dotted $error-red;
+			outline-offset: 2px;
```
haszari, line 75: _Maybe_ (soft maybe!) these could be bundled up in a mixin? Probably no need until they spread a bit further.

Aljullu:
Thanks for testing @haszari, I see there are some platform differences and that's a bit confusing.

I tested on BrowserStack with OS X Mojave and I think I could reproduce the issues, even though the screenshots I posted in the first comment of this PR look a bit different.

FF 80:

Peek 2020-08-10 13-17

Notice when focusing City the border becomes black (#2182) and the outline is blue (#2973).

In Chrome 84, you can also see that 'Country / Region' outline is blue while it should be red:

Peek 2020-08-10 13-28

---

> Screen Capture on 2020-08-07 at 09-33-15

I think you are reproducing #2182 here. When focusing 'City', the border changes to black. It's hard to notice because there is a red outline around it. I'm not sure why the outline looks red for you and blue for me, I tested it with BrowserStack.

> I can't reproduce the situation; no matter what I do, the select fields are always pre-populated with a value. So might be an issue, I just can't reproduce it.

Ah, I didn't think about this! If you go to WooCommerce > Settings > Default customer location, you should be able to set No location by default. In that case, the select is not prepopulated. You might need to open a private window, afterwards. Then, if you write something to an address field without having interacted with the select, the error will show up.

Aljullu:
> - The main bug is that error states sometimes had additional inconsistent non-red borders (lots of details, I got lost in the woods here palm_tree!)

Yes, actually there were two issues: border was not red in some themes (#2182) and outline was not red in any theme in FF (#2973).

> - This PR standardises styles for error states so they should always show:
>   - thin red border for field with error; and
>   - additional dotted red border when focused.

Exactly!

> One of the reasons I was confused is that my OS highlight colour setting was to pink, which is close ish to red. rofl Changed to green for today's testing.

Ah, right! Looking at your GIF above it looks pink instead of red. That was so confusing! :laughing:

PR #2975, opened 2020-08-06
Hide saved payment methods if their gateway is disabled

### LINKED ISSUE #2850: Checkout block shows saved Stripe card when Stripe payment method is disabled
**Describe the bug**
Stripe CC gateway allows the user to save their payment method (card) for next purchase. If the user has a Stripe-saved card, but Stripe CC gateway is disabled, the card is offered as a payment method in checkout block. In the traditional shortcode checkout, the saved card is not available unless Stripe CC is enabled.

**To Reproduce**
1. Set up checkout block, Stripe CC payment method, check `Enable Payment via Saved Cards`.
2. Complete a purchase with [Stripe test card](https://stripe.com/docs/testing) and check the `Save payment information to my account for future purchases.` checkbox on checkout.
7. Go to `WooCommerce > Settings > Payments` and disable Stripe CC payment method.
3. Add new stuff to cart, proceed to checkout.
4. Scroll down and see saved card, e.g. `Visa ending in 4242 (expires 02/22)`. 

Compare with shortcode checkout - no saved card offered. 

**Expected behavior**
Saved payment methods (e.g. `Visa ending in 4242` should only be offered if the gateway they are associated with (e.g. `Stripe – Credit Card`) is enabled. Or clarification that we're changing this behaviour.

Aljullu:
Fix #2850.

Steps from #2850.

1. Set up checkout block, Stripe CC payment method, check `Enable Payment via Saved Cards`.
2. Complete a purchase with [Stripe test card](https://stripe.com/docs/testing) and check the `Save payment information to my account for future purchases.` checkbox on checkout.
7. Go to `WooCommerce > Settings > Payments` and disable Stripe CC payment method.
3. Add new stuff to cart, proceed to checkout.
4. Scroll down and verify saved card (e.g. `Visa ending in 4242 (expires 02/22)`) is not there.

Also, test there are no regressions with the testing steps introduced in #2934.

### Changelog

> Saved payment methods are not shown to the user if their payment gateway has been disabled.

### DIFF
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -131,15 +131,17 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 					);
 				}
 			} );
-			currentOptions.current = options;
-			currentOptions.current.push( {
-				value: '0',
-				label: __(
-					'Use a new payment method',
-					'woo-gutenberg-product-blocks'
-				),
-				name: `wc-saved-payment-method-token-new`,
-			} );
+			if ( options.length > 0 ) {
+				currentOptions.current = options;
+				currentOptions.current.push( {
+					value: '0',
+					label: __(
+						'Use a new payment method',
+						'woo-gutenberg-product-blocks'
+					),
+					name: `wc-saved-payment-method-token-new`,
+				} );
+			}
 		}
 	}, [
 		customerPaymentMethods,
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -8,7 +8,6 @@ import {
 	usePaymentMethodDataContext,
 } from '@woocommerce/base-context';
 import RadioControl from '@woocommerce/base-components/radio-control';
-import { PAYMENT_GATEWAY_SORT_ORDER } from '@woocommerce/block-settings';
 
 /**
  * @typedef {import('@woocommerce/type-defs/contexts').CustomerPaymentMethod} CustomerPaymentMethod
@@ -105,15 +104,9 @@ const SavedPaymentMethodOptions = ( { onSelect } ) => {
 		if ( paymentMethodKeys.length > 0 ) {
 			paymentMethodKeys.forEach( ( type ) => {
 				const paymentMethods = customerPaymentMethods[ type ];
-				const enabledPaymentMethods = paymentMethods.filter(
-					( paymentMethod ) =>
-						PAYMENT_GATEWAY_SORT_ORDER.includes(
-							paymentMethod.method.gateway
-						)
-				);
-				if ( enabledPaymentMethods.length > 0 ) {
+				if ( paymentMethods.length > 0 ) {
 					options = options.concat(
-						enabledPaymentMethods.map( ( paymentMethod ) => {
+						paymentMethods.map( ( paymentMethod ) => {
 							const option =
 								type === 'cc' || type === 'echeck'
 									? getCcOrEcheckPaymentMethodOption(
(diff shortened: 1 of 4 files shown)


assets/js/base/components/payment-methods/saved-payment-method-options.js
```diff
@@ -8,6 +8,7 @@ import {
 	usePaymentMethodDataContext,
 } from '@woocommerce/base-context';
 import RadioControl from '@woocommerce/base-components/radio-control';
+import { PAYMENT_GATEWAY_SORT_ORDER } from '@woocommerce/block-settings';
```
nerrad, line 11: `SavedPaymentMethodOptions` is already consuming the Payment Method data context via `usePaymentMethodDataContext`. So you can grab the `paymentMethods` and `paymentMethodsInitialized` properties from there and use those to match against whether saved payment methods should display or not.

I even wonder if this should be handled upstream in the payment method data context and filter the available `customerPaymentMethod` by what is actually enabled in the `paymentMethods` object (which could be done via a effect in the context provider). That way this logic wouldn't need to exist in this component.

`PAYMENT_GATEWAY_SORT_ORDER` creates a coupling to server provided values that I think we should try to limit as much as possible in our components so down the road if this data is provided over REST or a transport other than via a WordPress host there's less refactoring involved.

src/Payments/Api.php
```diff
@@ -76,14 +76,25 @@ public function add_payment_method_script_dependencies( $dependencies, $handle )
...
 	 */
 	public function add_payment_method_script_data() {
 		// Enqueue the order of enabled gateways as `paymentGatewaySortOrder`.
 		if ( ! $this->asset_registry->exists( 'paymentGatewaySortOrder' ) ) {
-			$available_gateways = WC()->payment_gateways->payment_gateways();
-			$this->asset_registry->add( 'paymentGatewaySortOrder', array_keys( $available_gateways ) );
+			$payment_gateways = WC()->payment_gateways->payment_gateways();
+			$enabled_gateways = array_filter( $payment_gateways, array( $this, 'is_payment_gateway_enabled' ) );
+			$this->asset_registry->add( 'paymentGatewaySortOrder', array_keys( $enabled_gateways ) );
 		}
```
nerrad, line 98: Won't this break the logic that was implemented as a result of [this comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2934#discussion_r464924276)?

If you implement the changes I suggested in an earlier comment, there won't be need for this.

assets/js/base/components/payment-methods/saved-payment-method-options.js
Aljullu: :100: good idea. Should be fixed now.

src/Payments/Api.php
Aljullu: Not really: enabled and available are different concepts. For example, COD might be enabled but not available depending on the shipping method.

While this code is no longer needed for this PR, what do you think about keeping it? While I see you discussed that was not an information leak, at the same time I don't see any benefit on exposing disabled payment methods in the frontend, so I think we could keep the filter. Thoughts?

Aljullu:
Thanks for the review @nerrad. This is ready for another look. I see e2e tests are failing but I think it's not related to this pull.

src/Payments/Api.php
nerrad: > For example, COD might be enabled but not available depending on the shipping method.

Ahh, right that makes sense. I'm fine with keeping this but you should make sure testing steps for the work done in #2157 are done to cover any unexpected regressions (don't expect any but still...)

Aljullu: I consider testing steps in #2934 already cover any potential regression introduced in this PR. I updated the testing steps in this PR to also point to testing steps in #2934. Do you think we need to add anything else?

nerrad: Nope, if you're happy with that, it works for me!

PR #2976, opened 2020-08-06
Link cart/checkout merchant (editor) feedback prompt to special GitHub issue type

### LINKED ISSUE #2917: Redirect feedback for C&C blocks to a specific GitHub issue template
While feedback is coming in on the ideas board, it's a bit hard to remember to check it and then bring potential issues over to where we do most of our evaluation for future sprints/cycles. Discussion and conversation is also not really favorable on the ideas platform.

While in conversation with @garymurray, we thought it would be worthwhile to try switching the feedback link for C&C blocks to point to a specific template for a new issue here in our repository. The template could be used to head off potential issues that are more support related and/or already things on our radar (new payment method support for instance).

So for the purpose of this issue:

- Create a template specifically targeting Feedback for the Cart and Checkout blocks.
- Change the link in the C&C block feedback prompt to point to the new issue template.

haszari:
Fixes #2917 

This PR adds a new issue template for cart & checkout feedback, and links to it from the editor `Feedback` inspector panel.

Screen Shot 2020-08-06 at 3 11 26 PM

Note - the issue template is brief to encourage more feedback - don't want the template to feel like a heavy form "barrier". 

#### tidied issue templates
I've added `labels` to metadata, and also done this for our existing templates. This should automatically label new issues - I'll test to check this is working correctly after merge. Keen for feedback on which labels to use for each template (or if this is overkill).

Also I changed all the headings to use markdown heading format (e.g. `## Heading`) rather than bold. This is more semantically correct/clear, more compact, and easier to edit (bold has start/end markers; heading is simpler, just hashes on start of line). This also allows us to use heading hierarchy if that's useful (added to `Environment` section). A bit sneaky adding in this PR but I figure let's make these consistent while here - happy to revert this if team prefers other format.

#### `withFeedbackPrompt` HOC and existing component

As part of this I noticed that the `withFeedbackPrompt` HOC is no longer used, so I've deleted it. 

Also, the existing FeedbackPrompt links to the ideas board. I've retained this default behaviour, though it's not used - could simply remove it, focus specifically on cart/checkout (since that's the only use case for this component). 

- Read through the new [issue template](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/894062b53ca1d3f50fb95a75619f2ca752f524f5/.github/ISSUE_TEMPLATE/--cart-checkout-feedback.md). Does it make you want to provide clear useful feedback?
- Edit a page with cart or checkout block, confirm the inspector prompt is enticing and friendly (and links to the right place).


.github/ISSUE_TEMPLATE/---bug-report.md
```diff
@@ -32,7 +32,7 @@ If applicable, add screenshots to help explain your problem.
 
 **Desktop (please complete the following information):**
 
-* OS: [e.g. iOS]
+* OS: [e.g. macOS, Windows]
```
haszari, line 35: Fixed this while here - iOS is not a Desktop thing :)

nerrad:
Thanks for working on this Rua, I hope to get a closer look at this today (with confirmation) but before merging (*assuming the code review results in approval) I need to confirm one more time that we want to make this switch (I've got mixed messages needing clarified).

nerrad:
So just confirmed that we're going to do this as an experiment and see how this goes. I'll give a review later but a quick observation is that the template is on the long side. Directing folks specifically to an issue using the template should result in as little friction as possible. I know in the case of bugs, the more info there is the better, but for this specific case, I'd rather start with as little prompting as possible and we can iterate if needed. I think we need to position this template less as a "bug" or "feature" request and more as a questionnaire. I'll come back with some suggested questions sometime today.

nerrad:
So as a suggestion, I think we should start with changing this specific template to include only the following two questions and see how things go:

- What do you like about the blocks?
- What frustrates you or you think is missing?

haszari:
@nerrad That looks pretty good to me – keep it simple, especially since this is an experiment. I've made that change now, we can iterate as we go.

The one thing I'd like to tweak is to avoid the word `frustrate` if possible - thinking about how we can word that a little more openly/positively (don't want to prime frustration!). 

One other thing - is there a way we can automatically label these issues? I'll investigate, that could be helpful for triage etc.

haszari:
> One other thing - is there a way we can automatically label these issues? I'll investigate, that could be helpful for triage etc.

Yes we can :) 0a76d8a75c99683d0d1e94930cc47bd51d4f7e9f

https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository (see point 7)

.github/ISSUE_TEMPLATE/---bug-report.md
```diff
@@ -1,47 +1,45 @@
...
-
+labels: "type: bug" 
 ---
 
-**Describe the bug**
+## Describe the bug
 A clear and concise description of what the bug is.
 
-**To Reproduce**
+## To reproduce
```
haszari, line 10: Note - I've upgraded all our headings to real heading syntax while here. Just cleaner/simpler and more flexible IMO 😄 (I will admit this is a pet peeve of mine!)

.github/ISSUE_TEMPLATE/--cart-checkout-feedback.md
```diff
@@ -0,0 +1,15 @@
+---
+name: "\U0001F4AC Feedback – Cart & Checkout Blocks"
+about: Submit feedback about the new block-based Cart & Checkout.
+labels: "type: support", "◼️ block: cart", "◼️ block: checkout"
```
nerrad, line 4: I'd like creating a new label for this `type: feedback` and use it for this template. It'll make it easier to search for just feedback related things.

assets/js/components/feedback-prompt/index.js
```diff
@@ -13,7 +13,10 @@ import './style.scss';
 /**
  * Component to render a Feedback prompt in the sidebar.
  */
-const FeedbackPrompt = ( { text } ) => {
+const FeedbackPrompt = ( {
+	text,
+	url = 'https://ideas.woocommerce.com/forums/133476-woocommerce?category_id=384565',
+} ) => {
```
nerrad, line 19: Good call here with setting the default as the ideas board but allowing for overriding 👍

.github/ISSUE_TEMPLATE/--cart-checkout-feedback.md
haszari: Good idea - added here 65386c4f88e90a1eeb0c2cf0d66aefbec0fdddc4 - though I haven't added the label itself, we can handle that separately I guess?

Also in that commit update the `what do you think is missing` prompt wording, and added `Cart & Checkout` to reinforce that we are looking for feedback on those blocks specifically.

haszari:
Keen to get this merged so it doesn't hang around - it's had a review and feedback has been addressed. We can always iterate more of course 😄

PR #2980, opened 2020-08-07
detect failed/cancelled 3D secure payment auth and reset checkout state

### LINKED ISSUE #2743: 2x confusing error notices + broken checkout when user cancels 3D secure payment
**Describe the bug**
Stripe supports an extra ["3D Secure" payment verification modal](https://en.wikipedia.org/wiki/3-D_Secure) for some cards. There's a [test card number](https://stripe.com/docs/testing) you can use to trigger this modal.

If the user cancels the modal, the checkout is in a bad state and the error notices are unhelpful; the user cancelled, so the checkout should roll back to previous state.

**To Reproduce**
0. Set up your store with Stripe CC payment method, in test mode.
1. Set up page with checkout block (not shortcode).
2. Add stuff to cart, proceed to checkout.
3. Complete checkout form, and use this [test card number](https://stripe.com/docs/testing): `4000 0027 6000 3184`
4. Click `Place order`.
3. 3D secure / verification modal should pop up - click `CANCEL` in top right.
2. See error notices etc (see below).
1. Subsequent attempts to pay fail, even with different card number - I think the checkout is in a broken state. (Need to refresh page.)

**Expected behavior**
If the user cancels payment at any stage, the checkout should roll back to "ready to submit" state cleanly. Any error message should be consistent and reassuring, e.g. "payment cancelled by user" (Stripe maybe has recommended wording.)

There may be other modal-payment flows that we might need to test - e.g. [SCA](https://en.wikipedia.org/wiki/Strong_customer_authentication).

Also need to test other flows out of the 3DS modal, e.g. verification fail, verification succeed (etc).

**Screenshots**
Screen Shot 2020-06-17 at 4 01 02 PM

haszari:
Fixes #2743

#### Accessibility

1.
2.
3.

### Changelog

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-payment-intents.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/use-payment-intents.js
@@ -46,6 +46,14 @@ const openIntentModal = ( stripe, paymentDetails, emitResponse ) => {
 				return checkoutResponse;
 			}
 			checkoutResponse.redirectUrl = verificationUrl;
+			if ( response?.status === 'requires_payment_method' ) {
+				// This status indicates a 3D-secure auth has failed or been cancelled.
+				// Reset checkout so shopper can try again.
+				return {
+					type: 'error',
+					retry: true,
+				};
+			}
 			return checkoutResponse;
 		} )
 		.catch( function( error ) {


nerrad:
Are you still wanting to work on this @haszari? What's needed to keep this moving?

haszari:
> Are you still wanting to work on this @haszari? What's needed to keep this moving?

Thanks for the nudge @nerrad ! This PR doesn't really have anything of value, still figuring out the cause of the problem. I left a [comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2743#issuecomment-670317740) on the issue with where I was up to. Happy for anyone to take over the issue, I'm not actively working on it right now. I am keen to get this one fixed though!

nerrad:
This was fixed by #3272 so can be closed.

PR #2989, opened 2020-08-10
fix issue with no margins on phone input

### LINKED ISSUE #2979: Checkout block: phone field squished hard against other form elements
**Describe the bug**
Recently the `Phone` field layout has regressed in checkout, we've lost significant amount of gap, this is pretty serious:

Screen Shot 2020-08-07 at 9 42 49 AM

cc @senadir  who spotted this early 😄 

**To Reproduce**
Set up checkout block, add stuff to cart, view checkout block.

**Expected behavior**
Consistent margin & padding for all form elements.

senadir:
Fixes #2979 

This PR fixes the issue with overlapping margins, I simply switched to top margins, moved some styles, and added a new classname `.wc-block-components-state-input`.

### How to test

1- Verify the Checkout block for any possible changes, change to countries with three fields like Afghanistan and countries with predefined states like Spain.
2- Verify the Cart shipping calculator form.
3- Verify any weird spacing in other inputs in the Checkout form.

### Screenshots

image
image
image

### DIFF
--- a/assets/js/base/components/country-input/style.scss
+++ b/assets/js/base/components/country-input/style.scss
@@ -1,0 +1,5 @@
+.wc-block-components-country-input,
+.wc-block-components-state-input {
+	margin-top: em($gap-large);
+	position: relative;
+}
--- a/assets/js/base/components/country-input/style.scss
+++ b/assets/js/base/components/country-input/style.scss
@@ -1,5 +1,3 @@
-.wc-block-components-country-input,
-.wc-block-components-state-input {
+.wc-block-components-country-input {
 	margin-top: em($gap-large);
-	position: relative;
 }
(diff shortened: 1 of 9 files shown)


assets/js/base/components/country-input/style.scss
```diff
@@ -0,0 +1,5 @@
+.wc-block-components-country-input,
+.wc-block-components-state-input {
```
Aljullu, line 2: I'm not sure it's a good idea to have `state-input` styles in the `country-input` folder. Ideally each component should have its own styles in its folder. I would either:
1. duplicate this ruleset in `base/components/state-input/style.scss` or
2. move these styles to `base/components/select/style.scss`

How does it sound?

assets/js/base/components/country-input/style.scss
```diff
@@ -0,0 +1,5 @@
...
+	margin-top: em($gap-large);
+	position: relative;
```
Aljullu, line 4: Is `position: relative` needed? It looks like `.wc-block-components-select` already have it.

assets/js/base/components/country-input/style.scss
senadir: I wanted to avoid `base/components/select/style.scss` because that's a child component, will create a state input style.

PR #2990, opened 2020-08-11
Allow WP 5.4 + GB to fail until we fix it.

senadir:
The WP 5.4 + GB E2E test is currently failing on master and every branch, I'm going to fix it next week as part of the cooldown, currently, I just want to disable so that we get passing tests or see actual issues.

I added a todo to delete the failing part once we fix it, it should also serve a tracking issue for this.

### How to test

1- Travis should be green even if that test failed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -153,3 +153,17 @@ jobs:
               local_dir: storybook/dist
               on:
                   branch: main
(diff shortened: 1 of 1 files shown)


senadir:
looks good so far
image
image

senadir:
Tested locally and I see similar fails when running E2E on WP 5.4.2 + GB 8.7.1 + @wordpress/e2e-test-utils@4.11 (latest).

PR #2992, opened 2020-08-11
Use real previews for the Cart and Checkout blocks

senadir:
Part of #2981 

This PR uses real previews for the cart and checkout, something we can use because:
1- We have container queries so our blocks are responsive regardless of the screen size.
2- We don't have a select product step for those blocks, what you see is what you get.

### Screenshots
Before:
image
image

Screen Shot 2020-08-11 at 1 47 46 PM
Screen Shot 2020-08-11 at 1 48 43 PM

After:
image
image

Screen Shot 2020-08-11 at 2 40 32 PM
Screen Shot 2020-08-11 at 2 40 38 PM

1. Preview the cart and checkout inside the block inserter.
2. Make sure the block is not very long and overflowing.
3. Make sure the changes didn't leak to the actual block or the editor block.

### Changelog

> Cart and Checkout will show a live preview inside the block inserter and style selector.


assets/js/blocks/cart-checkout/cart/editor.scss
```diff
@@ -1,3 +1,9 @@
 .wc-block-cart__page-notice {
 	margin: 0;
 }
+
+.wp-block-woocommerce-cart.is-editor-preview {
+	max-height: 1000px;
```
senadir, line 6: This is a hack to limit the preview height, I filled an issue in GB 
https://github.com/WordPress/gutenberg/issues/24484

nerrad:
Just noting that we need to test this change across the various versions of WordPress Blocks support (currently WP 5.2+). I _think_ we could consider bumping the next release of Blocks to support WP 5.3+ too if needed and it would simplify any necessary progressive enhancement.

senadir:
Will test it in other versions, but this shouldn't impact anything because we're not touching anything new, preview attribute has been there since forever.

LevinMedia:
@senadir @nerrad while I appreciate the idea behind the preview for the styles, I don't think they're necessary in this case. The default, light inputs, are designed to work on all background colors - so this is really optional. Also, the nature of the toggle is such that activating it will provide instant feedback in the editor. If the look isn't desired, it can simply be toggled off. 

For this reason I think we should stick with a simple toggle switch:
Screen Shot 2020-08-11 at 12 14 58 PM

senadir:
@LevinMedia I agree, but I don't think that should be a blocker for this PR, but more for https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2981, I think we can merge this as a good enhancement to our blocks.
What I mean is, the ability to see the actual block on the inserter before inserting it, and how it will look with your theme.

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -143,12 +144,12 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
  *       in the frontend.
  */
 const CartEditor = ( { className, attributes, setAttributes } ) => {
-	if ( attributes.isPreview ) {
-		return cartBlockPreview;
```
Aljullu, line 147: It looks like `cartBlockPreview` is not used anywhere else, right? Can we remove `/assets/js/previews/cart-block.js` and `/assets/img/cart-preview.svg`? Same for Checkout preview.

assets/js/blocks/cart-checkout/cart/editor.scss
Aljullu: If we could display the desktop preview instead of the responsive mobile preview we would probably not have this problem, right? In addition, displaying the mobile preview in a desktop seems a bit weird.

One idea would be to extend the `EditorProvider` with an `isPreview` property. This way, inside `ContainerWidthContextProvider` we could check if it's a preview and, in that case, always apply the `isLarge` class. I don't think that should block this PR but if it's a solution that makes sense to you too, we could fill an issue.

assets/js/blocks/cart-checkout/cart/editor.scss
```diff
@@ -1,3 +1,9 @@
...
+	overflow: hidden;
+	position: relative;
```
Aljullu, line 8: I think `position: relative` can be removed, here?

assets/js/blocks/cart-checkout/checkout/editor.scss
```diff
@@ -22,3 +22,9 @@
 .components-panel__body-title .components-button {
 	opacity: 1;
 }
+
+.wp-block-woocommerce-checkout.is-editor-preview {
+	max-height: 1000px;
+	overflow: hidden;
+	position: relative;
```
Aljullu, line 29: And here. :slightly_smiling_face:

PR #3010, opened 2020-08-14
Deprecate wc.wcSettings.setSetting function.

### LINKED ISSUE #1838: Remove `setSetting` and usage from `@woocommerce/settings` alias.
`setSetting`  was originally exposed as an interface to mutating the `wcSettings` global which exposed data from the server. However, this is a bad pattern to expose because in the context of react apps, mutated global state can never be reacted to and it's a "hidden state change" that makes it far to easy for client code using this to create bugs in apps reliant on these values.

As such, the `@woocommerce/settings` interface should be provided as a way to explicitly expose _immutable_ settings from the server.  While we can't _enforce_ immutability, we can at least remove any hints that there is an explicit way to override these values globally (and use module scoping to limit the impact of changes).  If settings from the server need to be changed, then it is preferable to use them to hydrate some other state in the app (preferably a custom wp.data store or something like it). 

Thus, this issue is to remove the `setSetting` interface that is currently exposed on the `wc.wcSettings` global (and `@woocommerce/settings` alias).

There is effort already underway in wc-admin to remove this dependency so co-ordination will have to be made there with how/when this lands.

Some suggested steps:

- soft-deprecate (add a warning if it's used) (potentially use `@wordpress/deprecated` package)
- eventually (when wc-admin is ready) remove the logic from `setSetting` so it does nothing and warning is updated to console.error (so calls to `setSetting` won't error, but also won't update any values in the global).
- At some point (determined by wc-admin usage) remove the function all-together.

nerrad:
Fixes #1838

The `wc.wcSettings.setSetting` function should never have been added (I added it!) as global configuration settings coming from the server should be immutable. Allowing these values to be mutated means that code relying on these values can't trust them. Any code wanting to work with server side setting values that could be mutated over the life of a session should be using ajax or REST API endpoints to update and get current values.

Unfortunately, since this has been in the wild for a bit, we can't just completely remove this function. Instead this pull begins with a deprecation notice (powered by `@wordpress/deprecated`) and sets the version of `3.8.0` for when the function will be complete removed. This allows for 6 versions between when this is released (3.2.0) in the feature plugin and when it will be removed. It also allows for 3 releases of WooCommerce Core before removal (where the plugin is included as a package).

As far as _known_ impacts, WooCommerce Admin _was_ using this function a while ago (2019) but has long since stopped using it so I don't anticipate any issues there (cc @psealock and @timmyc for visibility)

## To Test

There's nothing user facing for this.

- with `WP_SCRIPT_DEBUG` set to true in your `wp-config.php` file.
- load any route that has a Woo block on it powered by REST API.
- type this in the browser console: `wc.wcSettings.setSetting( 'someValue', 10 )`
* [ ] Verify that you see the following message in the console:

> setSetting is deprecated and will be removed from WooCommerce Blocks in version 3.8.0. Please use a locally scoped value instead instead. Note: wc.wcSettings is a global settings configuration object that should not be mutated during a session. Hence the removal of this function.

### DIFF
--- a/assets/js/settings/shared/set-setting.js
+++ b/assets/js/settings/shared/set-setting.js
@@ -28,7 +28,7 @@ import deprecated from '@wordpress/deprecated';
 export function setSetting( name, value, filter = ( val ) => val ) {
 	deprecated( 'setSetting', {
 		version: '3.8.0',
-		alternative: `a locally scoped value for "${ name }" instead`,
+		alternative: `a locally scoped value for "${ name }"`,
 		plugin: 'WooCommerce Blocks',
 		hint:
 			'wc.wcSettings is a global settings configuration object that should not be mutated during a session. Hence the removal of this function.',
--- a/assets/js/settings/shared/set-setting.js
+++ b/assets/js/settings/shared/set-setting.js
@@ -28,7 +28,7 @@ import deprecated from '@wordpress/deprecated';
 export function setSetting( name, value, filter = ( val ) => val ) {
 	deprecated( 'setSetting', {
 		version: '3.8.0',
-		alternative: 'a locally scoped value instead',
+		alternative: `a locally scoped value for "${ name }" instead`,
 		plugin: 'WooCommerce Blocks',
 		hint:
 			'wc.wcSettings is a global settings configuration object that should not be mutated during a session. Hence the removal of this function.',
--- a/assets/js/settings/shared/set-setting.js
+++ b/assets/js/settings/shared/set-setting.js
@@ -19,7 +19,7 @@ import deprecated from '@wordpress/deprecated';
  *                                               to sanitize the setting (eg.
  *                                               ensure it's a number)
  *
- * @todo  Remove setSetting function from `@woocommerce/settings`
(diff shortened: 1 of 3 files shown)


assets/js/settings/shared/set-setting.js
```diff
@@ -3,17 +3,35 @@
...
+ * @todo  Remove setSetting function from `@woocommerce/settings`.
+ *
+ *  The `wc.wcSettings.setSetting` function was deprecated beginning with WooCommerce Blocks 3.2.0
+ *  and can be removed completely with WooCommerce Blocks 3.8.0 (that gives 6 versions of the
+ *  feature plugin for warning and 3 versions of WooCommerce core).
  */
 export function setSetting( name, value, filter = ( val ) => val ) {
+	deprecated( 'setSetting', {
+		version: '3.8.0',
+		alternative: `a locally scoped value for "${ name }" instead`,
```
senadir, line 31:
```suggestion
		alternative: `a locally scoped value for "${ name }"`,
```
this causes a double instead.

PR #3013, opened 2020-08-17
update minimum required version of WordPress core to 5.3

### LINKED ISSUE #2995: Update our min required WP version to WordPress 5.3 and update .travis.yml config for tests.
[WordPress 5.5](https://wordpress.org/news/2020/08/eckstine/) has shipped. With our L-2 support policy, that means the next version of WooCommerce Blocks will require a minimum of WordPress 5.3 (instead of 5.2). The following needs updated:

- README.md?
- readme.txt
- woocommerce-gutenberg-products-block.php
- .travis.yml (bump what versions of WP are tested against).
- drop legacy builds (and any explicit WP version checks).

Re legacy builds and WP version checks, it's probably good to leave the logic in place but just don't build the legacy builds, that way we can resume legacy builds if needed for progressive enhancement.

haszari:
Fixes #2995

With our L-2 policy for minimum WordPress version, with WooCommerce Blocks 3.2.0 our new min WP is 5.3 (up from 5.2).

This PR updates this everywhere needed (see diff for details).

Also, the "legacy" webpack builds are all disabled - these are only needed for WordPress 5.2 and earlier. The config is still available, just not enabled. (If we need to reinstate these builds we can.)

The release process doc has also been updated, with a new step for checking/updating user-facing docs. The main reason for this is to update the minimum requirements mentioned on WCCOM / docs site when these change. Also it's a handy reminder to ensure we have nice docs for new features!

1. Run release build `npm run build` and smoke test to ensure things are all working correctly.
2. Test dev build and ensure there are no problems.
3. Review the diff/codebase and ensure that WP version has been updated everywhere it should be!

### Changelog

> WooCommerce Blocks now requires WordPress 5.3 or greater (recommend latest).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,13 +104,6 @@ jobs:
               - npm run lint:ci
           env:
               - WOOCOMMERCE_BLOCKS_PHASE=3
-        - name: E2E Tests (WP 5.2)
-          script:
-              - npm run test:e2e
-          env:
-              - WP_VERSION=5.2
-              - E2E_TESTS=1
-              - WOOCOMMERCE_BLOCKS_PHASE=3
         - name: E2E Tests (WP 5.3)
           script:
               - npm run test:e2e
--- a/.travis.yml
+++ b/.travis.yml
@@ -118,14 +118,21 @@ jobs:
               - WP_VERSION=5.4
               - E2E_TESTS=1
               - WOOCOMMERCE_BLOCKS_PHASE=3
-        - name: E2E Tests (WP 5.4 with Gutenberg plugin)
+        - name: E2E Tests (WP 5.5)
+          script:
+              - npm run test:e2e
+          env:
+              - WP_VERSION=5.5
+              - E2E_TESTS=1
+              - WOOCOMMERCE_BLOCKS_PHASE=3
+        - name: E2E Tests (WP 5.5 with Gutenberg plugin)
           script:
               - npm run wp-env run tests-cli "wp plugin install gutenberg --activate"
               - npm install @wordpress/e2e-test-utils@latest
               - chmod -R 767 ./
               - npm run test:e2e
           env:
-              - WP_VERSION=5.4
+              - WP_VERSION=5.5
               - E2E_TESTS=1
               - GUTENBERG_LATEST=true
               - WOOCOMMERCE_BLOCKS_PHASE=3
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -104,13 +104,6 @@ jobs:
               - npm run lint:ci
           env:
               - WOOCOMMERCE_BLOCKS_PHASE=3
-        - name: E2E Tests (WP 5.2)
-          script:
-              - npm run test:e2e
-          env:
-              - WP_VERSION=5.2
-              - E2E_TESTS=1
-              - WOOCOMMERCE_BLOCKS_PHASE=3
```
senadir, line 113: Deleting this is only half of it, previously we supported `5.2`, `5.3`, `5.4`, and `5.5 + GB` now we should add `5.5` and `5.5 + GB`, which means you should leave everything as is and just update the `WP_VERSION` var and name in each suite.

docs/releases/handling-releases.md
```diff
@@ -199,6 +199,23 @@ Use previous posts for inspiration. If the release contains new features, or API
...
 _Outcome:_ __There's a public release announcement, with clear info about what's new, roughly within a day of actual release.__
 
+#### Update user-facing documentation as needed
+
+When the plugin is released, ensure user-facing documentation is kept up to date with new blocks and compatibility information. The dev team should update documents in collaboration with support team and WooCommerce docs guild.
+
+In particular, please review and update as needed:
+
+- Are there any new blocks in this release? Ensure they have adequate user documentation.
+- Ensure any major improvements or changes are documented.
```
senadir, line 209: where should this information live?

.travis.yml
haszari: Thanks @senadir , that makes sense 😄

docs/releases/handling-releases.md
haszari: By "this information" do you mean:

1. The hint/process that the docs should be updated as part of the release.
2. The information about the actual improvements / changes to the plugin.

Either way, a good question – maybe add this to team async meeting thread (and/or sprint retro) for wider discussion.

Here's my thoughts on the two above items:

1. I think there should be a reminder to update the docs in the release process, especially for the version/compatibility info. However I don't think this should be the only place we consider this – should be keeping it in mind over sprint process.
2. New blocks and features should documented on http://docs.woocommerce.com.

All that said, the release process doc is getting quite long. Maybe we could move some of this material to team handbook (P2) and keep this doc as a lean checklist.

PR #3018, opened 2020-08-17
Add Compatibility notice guide

### LINKED ISSUE #3017: Implement a compatibility notice when a user views or inserts Cart or Checkout for the first time.
We need to implement a compatibility notice guide that is shown to merchants the first time they insert or view the Cart or Checkout block, it should be shown only once.

P2: pbIy4N-CR-p2

We should use `Guide` component.

Designs: https://www.figma.com/file/n90hX6hG3wwQbYwsYsrJnj/Gutenberg-based-cart-%26-checkout?node-id=2278%3A45486

Screen Shot 2020-08-17 at 10 43 45 AM

Screen Shot 2020-08-17 at 10 43 50 AM

senadir:
WIP 
Fixes #3017

Has some known issues:

Will be visible for a second before disappearing when inserting Cart or Checkout.
You can better view it when you open a page that already has those blocks.
You can't fully dismiss it as it will show up again on next reload.

Will fill the metadata once I finish this.

#### Accessibility

1.
2.
3.

### Changelog


senadir:
We're still going to revisit this soon.

PR #3024, opened 2020-08-19
load `@woocommerce/components` from global, remove package dependency

### LINKED ISSUE #2419: Load @woocommerce/components from an external
Given that our minimum WC required version is 4.0, we can now load `@woocommerce/components` from the external. This will decrease our bundle size and ensure we use the same version as WooCommerce admin screens.

@senadir assigning it to you per the Slack chat we had yesterday. :slightly_smiling_face:

haszari:
Fixes #2419

WooCommerce Blocks has required Woo 4.0+ for a while now. This includes WooCommerce Admin, so we have the opportunity to load `@woocommerce/components` from an external (global) instead of pulling in the npm package.

This PR loads `@woocommerce/components`  from `wc.components` JS global and removes the dependency on the package.

We'll need to take care with this – any components we need will need to be available in the min Woo (wc-admin) version we support; and/or we'll need defensive checks to only use them if available. Also, if we start iterating on `@woocommerce/components` – e.g. adding or improving components – then we'll need to wait until they are released (and in min supported version?).

Alternatively, if we need to stay on the latest `@woocommerce/components`, we can abandon this PR and use the package. Note there's an upstream issue blocking us from using 5.0.0 of the package: https://github.com/woocommerce/woocommerce-admin/issues/5006

1. Clean build: `npm ci`, `composer install` then build (dev or production)
2. Smoke test all blocks, focus on blocks that use `@woocommerce/components` - some info below.
3. Edit / add new blocks, change settings etc – everything should work correctly.
4. Front end behaviour should be unaffected.

I believe we use woo components in editor only (`SearchListControl` `SearchListItem`); so best to focus testing on editing blocks.

Repeat tests with older versions of Woo, in particular, oldest supported version.

##### All Products family
- `Filter by Attribute`

##### Reviews
- `Reviews by Category`
- `Reviews by Product`

##### Blocks Components
These components are used when configuring blocks, e.g. selecting product(s) or product attributes. Ensure you've tested each dependent component at least once 🥇 

- `ProductsControl`
- `ProductTagControl`
- `ProductControl`
- `ProductCategoryControl`
- `ProductAttributeTermControl`

### Changelog

> Dev: Load `@woocommerce/components` from JS global (was previously loaded from npm package dependency).

### DIFF
--- a/assets/css/editor.scss
+++ b/assets/css/editor.scss
@@ -1,6 +1,3 @@
-// Import the woocommerce components stylesheet
-@import "~@woocommerce/components/build-style/style.css";
-
 // Hack to hide preview overflow.
 .editor-block-preview__content {
 	overflow: hidden;
--- a/bin/webpack-helpers.js
+++ b/bin/webpack-helpers.js
@@ -8,6 +8,7 @@ const FORCE_MAP = process.env.FORCE_MAP || false;
 
 const wcDepMap = {
 	'@woocommerce/blocks-registry': [ 'wc', 'wcBlocksRegistry' ],
+	'@woocommerce/components': [ 'wc', 'components' ],
 	'@woocommerce/settings': [ 'wc', 'wcSettings' ],
 	'@woocommerce/block-data': [ 'wc', 'wcBlocksData' ],
 	'@woocommerce/shared-context': [ 'wc', 'wcSharedContext' ],
@@ -16,6 +17,7 @@ const wcDepMap = {
 
 const wcHandleMap = {
 	'@woocommerce/blocks-registry': 'wc-blocks-registry',
+	'@woocommerce/components': 'wc-components',
 	'@woocommerce/settings': 'wc-settings',
 	'@woocommerce/block-settings': 'wc-settings',
 	'@woocommerce/block-data': 'wc-blocks-data-store',
--- a/package-lock.json
+++ b/package-lock.json
@@ -1427,15 +1427,6 @@
 				"regenerator-runtime": "^0.13.4"
 			}
 		},
-		"@babel/runtime-corejs2": {
-			"version": "7.7.4",
-			"resolved": "https://registry.npmjs.org/@babel/runtime-corejs2/-/runtime-corejs2-7.7.4.tgz",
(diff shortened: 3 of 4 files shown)


haszari:
I haven't tested this fully yet - keen to see what Travis / e2e says 😁 

Keen for testing/reviews either way.

nerrad:
Would be good to also measure what shifting to the global means in total asset load for routes using `@woocommerce/components`. While it loads via the external, that external file will still have it's own weight (plus any dependencies the script handle for woo components loads).

assets/css/editor.scss
```diff
@@ -1,6 +1,3 @@
-// Import the woocommerce components stylesheet
-@import "~@woocommerce/components/build-style/style.css";
-
```
senadir, line 3: I'm seeing some styling issues in that component.

image
image

I think, due to our special styling behavior, we might want to leave this file or move it somewhere in our webpack build cc @Aljullu 

Also, from my point of view, if we're only using this component alone, I'd say leave the package as is (load from node_module) and work on tree shaking the library, we're now loading the whole package, and we only need a fraction of it.

on main
image

senadir (changes requested):

assets/css/editor.scss
nerrad: > Also, from my point of view, if we're only using this component alone, I'd say leave the package as is (load from node_module) and work on tree shaking the library, we're now loading the whole package, and we only need a fraction of it.

I agree.

senadir: thinks work fine when I leave the stylesheet in place
image

haszari:
> Also, from my point of view, if we're only using this component alone, I'd say leave the package as is (load from node_module) and work on tree shaking the library, we're now loading the whole package, and we only need a fraction of it.

> I agree.

That makes sense to me @senadir @nerrad  – let's keep it simple for now. We're only using one or two components so loading as an npm package is more efficient. Also this has the benefit that we can rely on a known version; the external/global could have different versions of components depending on Woo core or wc-admin version on the site.

So with that in mind, I'm going to close this PR 🎉

PR #3025, opened 2020-08-19
clarify intentions / purpose of different component folders in docs

### LINKED ISSUE #2954: Consider renaming /assets/js/components/ to /assets/js/editor-components/
All components inside `/assets/js/components/` are exclusively used inside the editor, but the folder naming is a bit conflicting, especially with folders like `/assets/js/base/components/`, we should consider renaming it to something more relevant to the editor.

haszari:
Relates to #2954

~~In progress but I welcome all comments & feedback!~~ Fully ready for feedback now. This PR is focused solely on improving the docs around our components.

~~I haven't renamed the actual folder/alias package yet. There will be a moderate amount of code impact for that change, so I'll do that in one hit when we're aligned on the naming/meaning of the different folders (packages) of components.~~ **Update**: This PR is just a docs change, to help illustrate/clarify the different component collections. We can rename the folders (if appropriate) in a future PR.

This PR adds new docs (readmes and readme content) to flesh out our high-level docs for components.

- The `Storybook` doc is now a top-level doc about components - Storybook is a part of this.
- In the main doc I've attempted to capture the essential purpose of each folder. Definitely keen for feedback / buy-in on this! Trying to keep it really high-level but accurate.
- Both components folders have a readme now.
- `js/components` (woo editor components) readme is trimmed down. It previously had docs for most of the sub-folders, but we aren't keeping this up to date, so I've removed it. If we want to keep that level of detail it should be in JSDoc comments, Storybook, or individual component readme files (IMO). Again, keen for feedback on this – the idea is to [prefer "live" docs](https://www.sourceallies.com/2013/05/agile-manifesto-working-software-over-comprehensive-documentation/) and the code over readmes that we might not keep updated.

1. Review the docs changes – use `Rich diff` for best results.
2. Consider the naming of the folders/packages – let me know if you have any ideas here.
3. Consider the life cycle of components in each folder - are they likely to graduate to WooCommerce Admin, Gutenberg, or elsewhere? Bonus points for feedback/ideas about this.


docs/contributors/storybook.md
```diff
@@ -1,11 +1,26 @@
...
+We have components in a few folders, for different contexts.
+
+- [`assets/js/base/components`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/main/assets/js/base/components)
+- [`assets/js/editor-components`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/main/assets/js/components)
+- [`assets/js/icons`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/main/assets/js/icons)
+
+__`base/components`__ are used in front-end code (as well as editor & admin). These components are typically general purpose (and the largest collection). These components help us build consistent interfaces across the front end (shopper) experience and elsewhere. In time, some of these components may move to a general purpose lib, such as [`@wordpress/components` in Gutenberg](https://github.com/WordPress/gutenberg/tree/master/packages/components).
+
+__`base/editor-components`__ are used in the editor UI for our blocks. These are much more specific to WooCommerce Blocks. They allow us to build a consistent and powerful UI for merchants for authoring content relating to Woo data - e.g. selecting products or product attributes.
```
Aljullu, line 18: I'm a bit wary on using `editor-components` before that folder exists. I would just list `assets/js/components` for now until we rename the folder.

Related to that, if we are renaming `assets/components` to `assets/editor-components`, shouldn't we also rename `assets/hocs` to `assets/editor-hocs`? :thinking:

nerrad: > I'm a bit wary on using editor-components before that folder exists. I would just list assets/js/components for now until we rename the folder.

I agree with Albert here.

> Related to that, if we are renaming assets/components to assets/editor-components, shouldn't we also rename assets/hocs to assets/editor-hocs

Also agree, we'll have to make sure only editor specific hocs are in this folder though.

haszari: Good spotting - I renamed it in the docs first to get feedback before making the change for real (as noted in PR description).

haszari:
Thanks for the feedback @nerrad @Aljullu . Other than the (potential) name change for `js/base/components`, do you have any other feedback or thoughts? For example, do the readmes align with your model/ideas for the two collections? 

Sounds like renaming to `editor-components` is more controversial, so I'm now thinking we hold off on that. I'll update that in the docs (to align with the current naming). We can still rename, but let's decouple that from this (docs/readme) PR.

Then we can iterate on the meaning/naming in the docs. In future we can consider renaming the folders if/when we have better name ideas – coming up with the right name is challenging, and getting it slightly wrong would be worse than the current "generic" naming IMO.

haszari:
Ready for another review!

docs/contributors/components.md
```diff
@@ -0,0 +1,40 @@
...
+
+__`assets/js/base/components`__ are used in front-end code (as well as editor & admin). 
+These components are typically general purpose (and the largest collection). 
+These components help us build consistent interfaces across the front end (shopper) experience and elsewhere. 
+In time, some of these components may move to a general purpose lib, such as [`@wordpress/components` in Gutenberg](https://github.com/WordPress/gutenberg/tree/master/packages/components).
+
+__`assets/js/components`__ are used in the editor UI for our blocks. 
+These components are typically more specialised to WooCommerce Blocks. 
+They allow us to build a consistent and powerful UI for merchants for authoring content relating to Woo data - e.g. selecting products or product attributes.
+
```
Aljullu, line 24: I'm not 100% sure about the distinction saying `assets/js/base/components` have a more general purpose while `assets/js/components` are more specialized to WC Blocks.

I think some components under `assets/js/components` might be useful in other projects outside WC Blocks (ProductsControl, PageSelector, etc.). In the other hand, some components under `assets/js/base/components` are clearly unique to WC Blocks like some C&C components.

In my opinion, the unique distinction between those two directories is that one is used for editor components and the other want for components that can be used either in the editor or the frontend.

Aljullu (commented):
Overall, it looks good to me. I'm just not sure about the distinction between `assets/js/base/components` and `assets/js/components`. I left a comment inline. :point_down:

docs/contributors/components.md
haszari: That's true - it's not easy to characterise the different groups, other than editor/non-editor (though there are probably some counterexamples in `base/components`).

By the way, when I say "general purpose" I mean things like radio buttons, checkboxes etc - not specific to woo at all, and even potentially useful outside of a WordPress context. If components are closer to general purpose, they're much more likely to graduate to another repo. 

Many of our components are at least partially specialised - to woo blocks, GB editor, or woocommerce data. I think it's quite tricky/subtle to find good candidates for reuse here. 

Let's see what feedback we get from the rest of the team on the meaning / intention of the two folders, then update the wording.

haszari: Interesting that we're coming back around to `editor-components`!

haszari:
Thanks for the feedback @Aljullu  @nerrad ! I've tweaked the main readme and readmes in each component folder to align with discussions.

This is ready for another review now. Added @senadir to review list for awareness, but I'll merge as soon as I see a positive review.

When reviewing please check:

- `components.md` adequately explains the expectations/conventions for each collection.
- The individual collection readme files are consistent with this. 
- The docs are general enough that they won't go stale as we iterate (in particular the readmes).

#### #2954 Renaming ~~`js/components`~~ `js/editor-components`
Based on the P2 discussion and the resulting doc changes, I think it's a good idea to rename the smaller collection `js/editor-components`. I'll do that in a follow up PR - if you have feedback / ideas about the rename please comment on the issue :)

docs/contributors/components.md
```diff
@@ -0,0 +1,43 @@
...
+__`assets/js/base/components`__ are used in front-end code, as well as editor & admin. 
+These components help us build consistent interfaces across the front end (shopper) experience and elsewhere.
+Because they can be used in the front end and editor, components in this folder should:
+
+-  Perform efficiently - i.e. not adversely affect page performance/experience.
+-  Have lean dependencies - i.e. not bloat the payload unnecessarily.
+-  Look consistent in common themes; ideally should allow themes to adjust appearance as necessary.
+
+__`assets/js/components`__ are used in the editor UI for our blocks. 
+These components are typically more specialised to WooCommerce Blocks.
```
nerrad, line 25: Given the comments, I'd remove this line. I would read this and despite the usage of "typically", I'd still discard this folder containing any components that can be re-used in other projects (which is incorrect). I think for this iteration of the docs let's stick to what is super clear and that's the fact that this specific folder is for editor only components.

haszari: Works for me!

haszari: 96a97da

Issue #3028, opened 2020-08-20
Update Storybook deps, switch to new Storybook api/format, consider using `args` in Storybook stories


haszari:
Looking at the Storybook docs, I think there's a new way to declare your components props, which will automatically generate docs and knobs (now called `Controls`):

https://storybook.js.org/docs/react/writing-stories/args

https://storybook.js.org/docs/react/writing-stories/introduction#using-args

Maybe we can use this to make it easier to write stories & document our components.

We might need to upgrade for this – and we probably should anyway. We're on `"@storybook/react": "5.3.18",`, latest is 6+:

<img width="273" alt="Screen Shot 2020-08-20 at 3 47 18 PM" src="https://user-images.githubusercontent.com/4167300/90714774-86d58e80-e2fc-11ea-99ba-2d0c6c865add.png">

(Side note: why is RenovateBot not suggesting this update?!)

haszari:
Might split this into two separate issues – update vs adopt `args`.

haszari:
Tried updating all storybook deps to latest, storybook build fails – I think this means we need `core/js` or to hook it up to our babel. Getting lots of errors like this:

```
ERROR in ./assets/js/base/components/block-error-boundary/index.js
Module not found: Error: Can't resolve 'core-js/modules/es.array.concat' in '/Users/haszari/_automattic/merch-repos/woocommerce-blocks/assets/js/base/components/block-error-boundary'
 @ ./assets/js/base/components/block-error-boundary/index.js 1:0-41
 @ ./assets/js/base/utils/render-frontend.js
 @ ./assets/js/base/utils/index.js
 @ ./assets/js/base/context/add-to-cart-form/form-state/index.js
 @ ./assets/js/base/context/add-to-cart-form/index.js
 @ ./assets/js/base/context/index.js
 @ ./assets/js/base/components/cart-checkout/totals/totals-coupon-code-input/stories/index.js
 @ ./assets/js sync ^\.(?:(?:^|\/|(?:(?:(?!(?:^|\/)\.).)*?)\/)stories\/(?!\.)(?=.)[^/]*?\.js)$
 @ ./storybook/generated-stories-entry.js
[...]

PR #3029, opened 2020-08-20
Add storybook demo for Chip

haszari:
Part of #3031 

#### Accessibility

- [x] ~~All animations respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)~~

1. Execute `npm run storybook` and visit the URL (mine is http://localhost:6006/)
2. Navigate to `@base-components` > `Chip` and apply the following testing steps to both `Chip` and `RemovableChip`
3. Change the knobs at the bottom of the storybook panel and ensure the expected changes occur to the component.
4. Test with screen reader on and by using only the keyboard.
5. Check the accessibility tab of Storybook and verify that the contrast of the component is ok.

### DIFF
--- a/assets/js/base/components/chip/stories/index.js
+++ b/assets/js/base/components/chip/stories/index.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { text, select } from '@storybook/addon-knobs';
+import { text, select, boolean } from '@storybook/addon-knobs';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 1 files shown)


haszari:
While working on this I noticed that our Storybook is out of date, and the new Storybook seems to have quite a different API. 

Would be good to get updated soon and use the new API/approach for new stories.

Logged here: #3028

haszari:
FYI @woocommerce/rubik - if you're writing new stories, this branch has some (`Chip` `CheckboxControl` so far).

senadir (approved):

PR #3030, opened 2020-08-20
Add storybook demo for CheckboxControl

haszari:
Part of #3031 

#### Accessibility

- [ ] ~~All animations respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)~~

1. Execute `npm run storybook`
2. Visit the storybook URL (for me it is http://localhost:6006/) and then go to `@base-components` > `CheckboxControl` > `CheckboxControl`
3. Test the component to ensure the box becomes checked when clicking the label or checkbox.
4. Test with screen reader on and by using only the keyboard.
5. Check the accessibility tab of Storybook and verify that the contrast of the component is ok.

### DIFF
--- a/assets/js/base/components/checkbox-control/stories/index.js
+++ b/assets/js/base/components/checkbox-control/stories/index.js
@@ -1,0 +1,27 @@
+/**
+ * External dependencies
+ */
+import { text } from '@storybook/addon-knobs';
+import { useState } from 'react';
+
+/**
+ * Internal dependencies
+ */
+import Component from '../';
+
+export default {
+	title: 'WooCommerce Blocks/@base-components/CheckboxControl',
+	component: CheckboxControl,
+};
+
+export const CheckboxControl = () => {
+	const [ checked, setChecked ] = useState( false );
+
+	return (
+		<Component
+			label={ text( 'Label', 'Yes please' ) }
+			checked={ checked }
+			onChange={ ( value ) => setChecked( value ) }
+		/>
+	);
+};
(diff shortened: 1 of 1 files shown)


Aljullu:
@opr is this ready for review? I see the diff is quite big (+28k, -19k) and I'm not sure what files need to be reviewed, any clues? I think maybe undoing the merge and instead rebasing this branch with trunk would simplify the diff? Otherwise, what about cherry picking the relevant commits into a new branch based on trunk? I'm a bit wary of merging a PR as big as this one, specially considering github-actions reports some [assets increase](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3030#issuecomment-677520689).

assets/js/base/components/checkbox-control/stories/index.js
```diff
@@ -0,0 +1,27 @@
+/**
+ * External dependencies
+ */
+import { text } from '@storybook/addon-knobs';
+import { useState } from 'react';
+
+/**
+ * Internal dependencies
+ */
+import Component from '../';
```
Aljullu: It seems like in other stories we are importing the components with their name (in this case, it would be `CheckboxControl`), and the story has a generic name (in some that I checked, [it was named `Default`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/components/read-more/stories/index.js#L11)). Do you think we should keep it consistent here?

Aljullu (approved):
Thanks for updating the PR, @opr! The story works great and code looks good. I just added a minor comment about naming the component, but not a blocker.

assets/js/base/components/checkbox-control/stories/index.js
opr: Yes I agree! I'll make this change

PR #3043, opened 2020-08-24
Add Webpack config docs

Aljullu:
Add documentation about how Webpack is configured in WC Blocks and the different configs we have.

I tried to keep sections short and link directly to the relevant parts of the code, this reduces the burden to update the documentation when code changes and it makes it easy for the reader to find out where in the code each section can be found.

I excluded feature gating from these doc page, I think it requires a doc page on its own, but I don't mind adding it here if you think it should be part of it.

Read the new/modified docs and:

1. Verify everything is correct and text is clear enough.
2. Think of anything I might have forgotten or something you would add to those docs.

### DIFF
--- a/docs/contributors/css-build-system.md
+++ b/docs/contributors/css-build-system.md
@@ -26,6 +26,6 @@ All files described above are generated in a LTR version and a RTL version. The 
 
 Webpack config is split between several files, some relevant ones for the CSS build system are:
 
--   [`webpack.config.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/webpack.config.js): file that exports the configs.
--   [`bin/webpack-configs.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-configs.js): definition of the different configs used by Webpack.
--   [`bin/webpack-entries.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-entries.js): entries used by Webpack definitions.
+-   [`webpack.config.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/webpack.config.js): Top level webpack config. Includes support for legacy and main build.
+    -   [`bin/webpack-configs.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-configs.js): Code for generating each build config. This most closely resembles a classic webpack config - if you're looking for something, start here.
+        -   [`bin/webpack-entries.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-entries.js): Code for generating [webpack `entry` definitions](https://webpack.js.org/concepts/entry-points/) and mapping source files to entry points. If you're adding a new block or module to the build, start here.
--- a/docs/contributors/css-build-system.md
+++ b/docs/contributors/css-build-system.md
@@ -10,11 +10,13 @@ CSS files are built with Webpack, which gathers all SCSS files in the app and pr
 
 ## Legacy builds
 
-When building WC Blocks, a special build targeting old versions of WordPress is generated. Some blocks might not be available in that legacy build. For example, All Products and Filter blocks not being available on WP 5.2.
+In the past, when building WC Blocks, special builds targeting old versions of WordPress were generated. Those builds were named 'legacy builds' and might have a smaller set of blocks available. For example, All Products and Filter blocks were not available on WP 5.2.
 
-Legacy builds use their own CSS files with the suffix `-legacy`, this allows saving some bytes because non-legacy block styles are not included.
+Currently, those builds are no longer generated since WC Blocks doesn't support WP 5.2 anymore, but the built system is still in place in case we need legacy builds in the future.
 
-> Note: All components' styles are included in legacy CSS files no matter if they are only used by non-legacy blocks. See [#2818](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2818) for more details.
+Legacy builds used their own CSS files with the suffix `-legacy`, this allowed saving some bytes because non-legacy block styles were not included.
+
+> Note: All components' styles were included in legacy CSS files no matter if they were only used by non-legacy blocks. See [#2818](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2818) for more details.
 
 ## Right-to-left
 
--- a/docs/contributors/js-build-system.md
+++ b/docs/contributors/js-build-system.md
@@ -33,6 +33,8 @@ There are several aliases for internal imports which make importing files across
 +import { useStoreCartCoupons } from '@woocommerce/base-hooks';
 ```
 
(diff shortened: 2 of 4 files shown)


webpack.config.js
```diff
@@ -53,8 +53,7 @@ const FrontendConfig = {
 };
 
 /**
- * This is a temporary config for building the payment methods integration script until it can be
- * moved into the payment extension(s).
+ * Config for building the payment methods integration scripts.
```
haszari, line 56: The "temporary" comment still applies here I think - so maybe worth adding it back. E.g. something like:

`Eventually payment method code will move to relevant extensions.`

cc @nerrad might have good ideas for how to word this.

docs/contributors/css-build-system.md
```diff
@@ -10,11 +10,13 @@ CSS files are built with Webpack, which gathers all SCSS files in the app and pr
...
 
-Legacy builds use their own CSS files with the suffix `-legacy`, this allows saving some bytes because non-legacy block styles are not included.
+Currently, those builds are no longer generated since WC Blocks doesn't support WP 5.2 anymore, but the built system is still in place in case we need legacy builds in the future.
 
-> Note: All components' styles are included in legacy CSS files no matter if they are only used by non-legacy blocks. See [#2818](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2818) for more details.
+Legacy builds used their own CSS files with the suffix `-legacy`, this allowed saving some bytes because non-legacy block styles were not included.
+
+> Note: All components' styles were included in legacy CSS files no matter if they were only used by non-legacy blocks. See [#2818](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2818) for more details.
```
haszari, line 19: Thanks for clarifying the status / rationale for the legacy build - useful info 🧡

docs/contributors/js-build-system.md
```diff
@@ -0,0 +1,55 @@
+# JS build system
+
+WooCommerce Blocks uses Webpack to build the files that will be consumed by browsers. There are several different Webpack configs in order to build files for different contexts of the plugin. They can all be found in [`webpack.config.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/webpack.config.js#L148-L160), but this is a quick summary:
+
+-   `CoreConfig`: config for shared libraries like settings, blocks data or some HOCs and context.
+-   `MainConfig`: config that builds the JS files used by blocks in the editor and responsible for registering the blocks in Gutenberg.
```
haszari, line 6: Typo - `the JS files used by blocks in the editor and is responsible for registering the blocks in Gutenberg`.

docs/contributors/js-build-system.md
```diff
@@ -0,0 +1,55 @@
...
+
+Some of the Babel plugins we use can be found in [`webpack-configs.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-configs.js).
+
+## External scripts
+
+Several scripts are loaded as externals. That means that they are imported in WC Blocks as any other package, but instead of being bundled in our built files, they are loaded from an external file. We use [`@wordpress/dependency-extraction-webpack-plugin`](https://developer.wordpress.org/block-editor/packages/packages-dependency-extraction-webpack-plugin/) to automatize dependency extraction for common WP scripts.
+
+## Aliases
+
+There are several aliases for internal imports which make importing files across the file tree easier. Instead of having to write a long relative path, they allow writing an alias:
```
haszari, line 29: These aliases also make refactoring easier 🎉

docs/contributors/js-build-system.md
```diff
@@ -0,0 +1,55 @@
...
+There were legacy builds of the `MainConfig`, `FrontendConfig` and `StylingConfig`. Currently, those builds are no longer generated since WC Blocks doesn't support WP 5.2 anymore, but the built system is still in place in case legacy builds are needed in the future.
+
+## Relevant files
+
+Webpack config is split between several files:
+
+-   [`webpack.config.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/webpack.config.js): file that exports the configs.
+-   [`bin/webpack-configs.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-configs.js): definition of the different configs used by Webpack.
+-   [`bin/webpack-entries.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-entries.js): entries used by Webpack definitions.
+-   [`bin/webpack-helpers.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-helpers.js): includes several utils to load externals.
```
haszari, line 55: Would love to clarify these different files a bit more. One thing that might help is to indent to show the dependency between the files (since it's not simple/linear). Here's an idea, though I am not an expert on each file - just sharing my understanding of how it works 😄 


- [`webpack.config.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/webpack.config.js): Top level webpack config. Includes support for legacy and main build.
    - [`bin/webpack-configs.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-configs.js): Code for generating each build config. This most closely resembles a classic webpack config - if you're looking for something, start here.
        -   [`bin/webpack-entries.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-entries.js): Code for generating [webpack `entry` definitions](https://webpack.js.org/concepts/entry-points/) and mapping source files to entry points. If you're adding a new block or module to the build, start here.
    -   [`bin/webpack-helpers.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/bin/webpack-helpers.js): Includes utils to load external code at run time, e.g. some dependencies from Woo and WordPress core.

webpack.config.js
haszari: Hmm, maybe not temporary – the core payment methods (COD, cheque etc) might stay here forever?

haszari (approved):
Nice work @Aljullu ! This is looking great, and really useful to docs about how our webpack config is structured.

I've left some comments but nothing blocking - approved 👌🏻

webpack.config.js
Aljullu: Ah, yes. I wanted to comment this line but later forgot about it. :facepalm: 

> Hmm, maybe not temporary – the core payment methods (COD, cheque etc) might stay here forever?

Yes, that was my thinking too. Stripe integration might be temporary, but support for WC Core methods will stay, correct?

docs/contributors/js-build-system.md
Aljullu: Fixed eb027e76d81193d65b21e8dfb14c7b982cb3aae7.

docs/contributors/js-build-system.md
Aljullu: Added a line about it: c3c23b3b7298a068bace03393dc7de261405f10c. :+1:

docs/contributors/js-build-system.md
Aljullu: I love this! Updated the docs with your proposal: 87c97a11a5d872a499f6c5042b80172f4d5d1666.

PR #3044, opened 2020-08-24
Fix CartItems unit tests

Aljullu:
PHP unit tests started to fail recently, the issue is that we were adding a variable product to the cart via `WC()->cart->add_to_cart()` with the wrong attributes. The issue only became visible after this PR in Core: https://github.com/woocommerce/woocommerce/pull/27115. It sets some stricter validation rules that our tests were not complying with. After updating the attributes, tests pass again.

1. Run `npm run phpunit` and verify all tests pass.
2. Verify PHP tests pass in Travis too.


Aljullu:
@nerrad correct, the issue was in the way our tests were written: the attributes we were defining in the test file were not valid and, with the new validation rules, the `add_to_cart()` action was failing. There is no need to do any change in production code.

PR #3057, opened 2020-08-26
Tweak insertOnce tests and fix reviews tests.

### LINKED ISSUE #2993: investigate why is WP 5.4 + GB E2E test is failing and remove the skip
In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2990, we added an `allow_fail` for WP 5.4 + GB E2E test because it kept failing for us on main, the reasons are not clear yet, but it could be an API change.

senadir:
This PR does a couple of things:

- Fixes some fragile Reviews tests that relied on all products having a review, now we will select products or categories that actually have reviews (using fixture data), it also introduces a new helper function `getElementByText` that gets an element by class name and content (static or regex).

- Refactor our `can only be inserted once` tests, we had problems with `insertBlock` helper function because it waited for the block to actually insert before moving forward, causing a timeout, we now search for the block and check if the button is disabled or not.

### New utilities

- `async closeInserter()` is now a global utility that you can use, it will close the inserter after failed inserts.
- `async getElementByText( selector<String>, pattern<String> )` that will return an element using a selector and a pattern that can be a regular expression.

fixes #2993, I removed the allow fail declaration. 
### How to confirm changes
Travis should pass for this test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -155,17 +155,4 @@ jobs:
               local_dir: storybook/dist
               on:
                   branch: trunk
-    # @TODO: Investigate why is WP 5.4 + GB e2e test failing and don't allow it to fail.
-    allow_failures:
-      - name: E2E Tests (WP 5.5 with Gutenberg plugin)
-        script:
-          - npm run wp-env run tests-cli "wp plugin install gutenberg --activate"
-          - npm install @wordpress/e2e-test-utils@latest
-          - chmod -R 767 ./
-          - npm run test:e2e
-        env:
-          - WP_VERSION=5.5
-          - E2E_TESTS=1
-          - GUTENBERG_LATEST=true
-          - WOOCOMMERCE_BLOCKS_PHASE=3
 
--- a/tests/e2e/specs/backend/all-products.test.js
+++ b/tests/e2e/specs/backend/all-products.test.js
@@ -1,11 +1,7 @@
 /**
  * External dependencies
  */
-import {
-	insertBlock,
-	getAllBlocks,
-	switchUserToAdmin,
-} from '@wordpress/e2e-test-utils';
+import { switchUserToAdmin } from '@wordpress/e2e-test-utils';
 
 import { visitBlockPage } from '@woocommerce/blocks-test-utils';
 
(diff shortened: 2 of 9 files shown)


tests/e2e-tests/specs/backend/reviews-by-category.test.js
```diff
@@ -32,7 +40,11 @@ describe( `${ block.name } Block`, () => {
 		await page.waitForSelector(
 			`${ block.class } .woocommerce-search-list__item`
 		);
-		await page.click( `${ block.class } .woocommerce-search-list__item` );
+		const categoryWithReviews = await findElementWithText(
+			`.woocommerce-search-list__item`,
+			fixtureProducts()[ 0 ].categories[ 0 ]
+		);
+		await categoryWithReviews.click();
```
Aljullu, line 47: I don't know why this test is failing for me here. I tried debugging it with `npm run test:e2e-dev` and I saw only the 'Uncategorized' item is shown in the selector:

imatge

I can keep investigating what's going on, but if somebody has any idea, that would be appreciated. :slightly_smiling_face:

Aljullu (commented):
One of the tests is failing for me (it might be something wrong with my local install, considering that it's passing in Travis), so I'm unable to give an approval to the PR, but I took a look and overall code looks good.

tests/e2e-tests/specs/backend/reviews-by-category.test.js
senadir: It seems you don't have a category named `Music` that should be automatically created as part of the fixtures, I'm not really sure why your instance fails to insert it.
https://github.com/woocommerce/woocommerce-gutenberg-products-block//blob/116e962aa4b1cd63252a18988506d911a5da2814/tests/e2e-tests/config/setup.js#L36

tests/e2e-tests/specs/backend/cart.test.js
```diff
@@ -35,9 +32,15 @@ describe( `${ block.name } Block`, () => {
...
 	it( 'can only be inserted once', async () => {
-		await insertBlock( block.name );
+		await searchForBlock( block.name );
+		const disabledInsertButton = await findElementWithText(
+			'button.editor-block-list-item-woocommerce-cart',
+			block.name
+		);
+		expect(
+			await disabledInsertButton.evaluate( ( button ) => button.disabled )
+		).toBe( true );
```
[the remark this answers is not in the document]
senadir: Yeah I thought of that.

nerrad:
FYI: I ran the e2e tests in this branch and they all passed for me locally.

nerrad:
@senadir this pull has been open for a while, besides a rebase, what else is needed to see this through to merge? Will it potentially improve the intermittent test fails we've been seeing?

senadir:
After several attempts to rebase this, I'm going to simply close and start over.

PR #3062, opened 2020-08-27
Split Checkout block component into smaller files

Aljullu:
The component for the Checkout block kept growing and growing as we added features, it was over 440 lines of code and it was becoming difficult to work with. This PR splits it into smaller files.

Even though the line diff of this PR is quite big, it doesn't add or remove any lines other than moving files/code around, renaming some CSS classes and updating the imports according to the new file structure.

Do some smoke testing of the Checkout block: add it in the editor, change some attributes, verify it works in the frontend, try making a purchase, etc.


haszari:
Proactive refactoring, I like! 🏅

assets/js/base/components/form/index.js
```diff
@@ -18,12 +13,10 @@ const CheckoutForm = ( {
...
 		onSubmit( event );
 	};
+
 	return (
 		<form
-			className={ classnames(
-				'wc-block-components-checkout-form',
-				className
-			) }
+			className={ classnames( 'wc-block-components-form', className ) }
```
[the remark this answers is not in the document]
haszari: e.g. is this `NumberedForm` - where are the checkout-specific styles/behaviour (vs generic)?

assets/js/blocks/cart-checkout/checkout/form/index.js
```diff
@@ -0,0 +1,323 @@
...
+	loginToCheckoutUrl: PropTypes.string.isRequired,
+	requireCompanyField: PropTypes.bool.isRequired,
+	requirePhoneField: PropTypes.bool.isRequired,
+	showApartmentField: PropTypes.bool.isRequired,
+	showCompanyField: PropTypes.bool.isRequired,
+	showOrderNotes: PropTypes.bool.isRequired,
+	showPhoneField: PropTypes.bool.isRequired,
+};
+
+export default CheckoutForm;
```
haszari, line 323: I see this is the business end, still a big component. Good to start breaking it down, and we can iterate more in future 😄

assets/js/blocks/cart-checkout/checkout/block.js
```diff
@@ -220,204 +119,15 @@ const Checkout = ( { attributes, scrollToTop } ) => {
...
-													'woo-gutenberg-products-block'
-											  )
-									}
-									value={ orderNotes }
-								/>
-							</FormStep>
-						) }
-					</CheckoutForm>
+					<CheckoutForm
+						loginToCheckoutUrl={ loginToCheckoutUrl }
```
haszari, line 123: Do we need to pass the login URL through all the layers? Could be local to the component that needs it, or refactored out to utils/settings.

docs/theming/class-names-update-330.md
```diff
@@ -45,4 +45,8 @@ WC Blocks 3.3.0, introduced express payment methods in the Cart block. In order
 		<td><code>wc-block-components-express-checkout-payment-event-buttons</code></td>
 		<td><code>wc-block-components-express-payment__event-buttons</code></td>
 	</tr>
+	<tr>
+		<td><code>wc-block-components-checkout-form</code></td>
+		<td><code>wc-block-checkout__form</code></td>
```
haszari, line 50: Ah I see this change is documented! 👌

haszari (approved):
Thanks for streamlining this @Aljullu  - great to see our code getting tidier and easier to wrangle as it grows.

I did some quick smoke testing in Chrome, Safari and Firefox, with Storefront. Didn't see any issues.

I did get confused by a quirk with `Phone` field (unrelated to this PR) - if the order is virtual, there's no `Shipping address` field, and no `Phone` field. I understand why this is the case – phone is a shipping field (not included in billing addr). But the block options aren't clear, so this could be confusing. `Phone` and `Require phone number` only apply to shippable orders. I haven't added a follow up issue since this might be expected behaviour for most merchants – do you think this is worth adding an issue for?

Screen Shot 2020-08-28 at 9 21 17 AM

haszari:
Spotted a visual issue in editor when reviewing this: #3068

assets/js/base/components/form/index.js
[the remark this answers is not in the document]
Aljullu: I assumed this is superseded by https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3062#discussion_r478695479, correct? :slightly_smiling_face:

assets/js/blocks/cart-checkout/checkout/form/index.js
Aljullu: Yeah, at some point it might make sense to keep splitting this one into smaller chunks. But I feel now it's much more self-contained (it's only the form) than before (it was the entire block inside a single component).

assets/js/blocks/cart-checkout/checkout/block.js
Aljullu: Ah, good point. Moved it to a `utils.js` file in bc03b8906bcb8990600a195e7cce1c98997f24b3.

docs/theming/class-names-update-330.md
Aljullu: Yes, and moved it to a specific file for 3.4.0 in 0710e73a2e856f84384667451904f1e8e736f08d.

Aljullu:
Thanks for the review @haszari!

> I did get confused by a quirk with `Phone` field (unrelated to this PR) - if the order is virtual, there's no `Shipping address` field, and no `Phone` field. I understand why this is the case – phone is a shipping field (not included in billing addr). But the block options aren't clear, so this could be confusing. `Phone` and `Require phone number` only apply to shippable orders. I haven't added a follow up issue since this might be expected behaviour for most merchants – do you think this is worth adding an issue for?

Good catch! I don't have a strong opinion on this, it's true it might be confusing for stores that only sell digital products, but I think that can be considered the expected behavior?

haszari:
> Good catch! I don't have a strong opinion on this, it's true it might be confusing for stores that only sell digital products, but I think that can be considered the expected behavior?

Yep - agreed. Just noting the potentially confusing UX/behaviour when I see it 😄

assets/js/base/components/form/index.js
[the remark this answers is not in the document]
haszari: > > Is this form component abstract or is it somewhat coupled to checkout?
> > e.g. is this NumberedForm - where are the checkout-specific styles/behaviour (vs generic)?

> I assumed this is superseded by #3062 (comment), correct? 🙂

Yes ish - though I am still curious how generic this `Form` component is, might need more work/tweaks when used elsewhere (not in Checkout). We can iterate when that happens, if it is an issue.

PR #3086, opened 2020-08-31
Fix JS console error when COD is enabled and no shipping method is available.

### LINKED ISSUE #2987: JS error when selecting an address with no available shipping options
## To reproduce
1. Add products to your cart that need shipping.
2. Go to the Checkout block and select an address without any shipping option available.
3. Look at the console and you will notice there is a JS error:
```
Uncaught (in promise) Error: Problem with payment method initialization: can't access property "includes", selectedMethod is undefined
```

## Expected behavior
There shouldn't be a JS error.

## Screenshots
imatge

senadir:
Only return selected shipping methods from useSelectShippingRate, this fixes an issue with COD in which we try to see if user can make a payment based on the selected shipping method, the hook returned a broken object `{ 0: undefined }` but it should have returned an empty object instead.

Fixes #2987

1. Enable COD and limit it to a certain shipping method type.
2. Add that method to a certain address.
3. Select another address on the Checkout block, you shouldn't see an error with `selectedMethod is undefined` on the console.

### Changelog

> Fix JS console error when COD is enabled and no shipping method is available.

### DIFF
--- a/assets/js/base/hooks/shipping/use-select-shipping-rate.js
+++ b/assets/js/base/hooks/shipping/use-select-shipping-rate.js
@@ -34,7 +34,9 @@ export const useSelectShippingRate = ( shippingRates ) => {
 				)
 				// A fromEntries ponyfill, creates an object from an array of arrays.
 				.reduce( ( obj, [ key, val ] ) => {
-					obj[ key ] = val;
+					if ( val ) {
+						obj[ key ] = val;
+					}
 					return obj;
 				}, {} ),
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
This solution looks good to me. :+1: 

If I understand it correctly, that means `selectedShippingRates` will only contain packages that have a selected shipping rate, correct? Do you think we could add this info to the doc comment above `useSelectShippingRate`? I feel this area might be confusing so maybe it's better to explicitly explain it.

PR #3088, opened 2020-08-31
Fix an issue with COD not showing when first enabled

### LINKED ISSUE #3087: COD is not visible unless you interact with its settings
## Describe the bug
I found this bug a couple of weeks ago and was about to fix it the next day but then it fell off, COD is not visible when you first set it up.

## To reproduce

1. You can either test this on a new website, or delete `woocommerce_cod_settings` option from `wp_options` table.
2. Enable Cash on Delivery, don't set it up, leave everything as is, it should work with all shipping methods.
3. Create an order and enter a valid address that has a shipping method with it.
4. You won't see COD as a payment option.
5. Go to COD settings, select a value for "Enable for shipping methods".
6. Save the settings.
7. Delete what select to return the field to its initial state.
8. Visit checkout again, you will see that COD is now working.

## Expected behavior
COD should work for the first time without touching any option.

senadir:
This PR fixes 3087 by defaulting to an array if `enable_for_methods` is not set.

The issue here is that WooCommerce will set `enable_for_methods` as an empty string when interacting with the field, but if you don't touch it, it won't be set, and it would default to an empty array.
This also changes the logic of COD `canMakePayment` to account for arrays, previously it assumed that `""` was the only empty value, but `[]` could also mean we don't have any values, doing `! settings.enableForShippingMethods` is not enough because `[]` is a truthy value, you can check this for examples about truthy values (#2024).

The original steams from WooCommerce and how it changes value types between states (switching from an array to an empty string).

Fixes #3087

1. You can either test this on a new website, or delete `woocommerce_cod_settings` option from `wp_options` table.
2. Enable Cash on Delivery, don't set it up, leave everything as is, it should work with all shipping methods.
3. Create an order and enter a valid address that has a shipping method with it.
4. You will see COD as a payment option.
5. Go to COD settings, select a value for "Enable for shipping methods".
6. Save the settings.
7. Delete what select to return the field to its initial state.
8. Visit checkout again, COD is still working.


src/Payments/Integrations/CashOnDelivery.php
```diff
@@ -70,7 +70,11 @@ private function get_enable_for_virtual() {
 	 * @return array Array of shipping methods (string ids) that allow COD. (If empty, all support COD.)
 	 */
 	private function get_enable_for_methods() {
-		return $this->get_setting( 'enable_for_methods', [] );
+		$enable_for_methods = $this->get_setting( 'enable_for_methods', [] );
+		if ( '' === $enable_for_methods ) {
```
haszari, line 74: Nice - I wonder if there is a standard way to enforce array return value? 

For example, we might want to coerce/cast or call a validate function to ensure that the option value is ALWAYS converted to `array`. Similar to `filter_var( ..., FILTER_VALIDATE_BOOLEAN )`, but for array type (note - `filter_var` doesn't support this!).

The empty string check works fine for this case, I'm just wondering if we can make it more robust.

haszari (approved):
Tested and confirmed this fixes the issue nicely. Thanks for logging and fixing this @senadir – I think I've seen this but didn't know how to reproduce it 🎖️ 

I'm pre-approving this - totally happy for it to merge as is. If you come up with a more robust way to coerce to `array` feel free to add that before merge – up to you.

PR #3089, opened 2020-08-31
Fix an issue in which sale price didn't have a strikethrough on mobile and correct shade on desktop.

### LINKED ISSUE #2991: "Price Strikethrough Text" and "SAVE $100" for on sale items not showing in Mobile Cart Page in Portrait Orientation"
## Describe the bug

In the "Cart" page, on-sale Items display both the the regular price and the sale price.  On Desktop, it shows the "Regular Price" with a strikethrough and a box that says "SAVE $100" as well as the new price which is working fine.

In mobile, it only shows the "Regular Price" and "Sale Price" side by side with no strikethrough and no box that says "SAVE $100"  It would be confusing for clients seeing two prices on the cart page.

It however, shows when the phone is rotated to landscape.  It just can't be seen in portrait

## To reproduce

1. Add an "On-sale" product to cart
2. Press "View Cart
## Expected behavior
The cart will display the product with its regular price with strikethrough formatting and the discount price next to it

## Screenshots
IMG_8694
______________

IMG_8696

* Core version: 
* Gutenberg version (if installed): 
* WooCommerce version: 4.3.2
* WooCommerce Blocks version: 3.1.0
* Site language:
* Device: iPhone XR
* OS: iOS 13.3
* Browser:

 Safari
* Version

### LINKED ISSUE #3080: Cart block: original (undiscounted) prices render without strikethrough on Seedlet theme
## Describe the bug
In the Cart block, if products are discounted, the original price is shown with strikethrough formatting. On Seedlet theme, it's rendering as normal. This could be confusing 😕 

Screen Shot 2020-08-31 at 5 15 07 PM

Note that original prices are displaying correctly in other places, e.g. All Products block.

Observed while testing 3.3.0 release.

## To reproduce
1. Add some products to store. Ensure some products have sale prices.
2. Install & activate Seedlet theme.
1. Set up a page with cart block.
1. On front end, add discounted stuff to cart and view cart. 
3. Notice original prices are not strikethrough.

## Expected behavior
Original prices for discounted products should always be displayed with strikethrough, so it's clear what price shopper will pay.

senadir:
Fixes #2991
Fixes #3080 (edit - @haszari)

This fixes two issues with sale price:

- It was the same color as `normal price` on desktop (it should have been a lesser shade).
image
This issue did not exist on mobile, because `del` and its display value.
- It didn't have a strikethrough on mobile, but it did have on desktop, strikethrough is provided by `del`
image

Both issues stem from `<del>` and what you assign for a display value for its children.

WooCommerce core provides `opacity: 0.5` for `.sale del`, this works if `<del>` children's display value is an inline value (inline, inline-block, inline-table, inline-grid, inline-flex), but not `block` or any block level display value.
`<del>` natively styles inner text with `strikethrough` but that will not work if children's display value is either `inline-table` or `inline-block`, but all other values, block or inline, will work.
Clearly this is too much logic to juggle, so I reached for `del` and any accessibility issues we might face if we delete it, turns out it provides no value whatsoever over `text-decoration` both will not be announced, sources:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/del
- http://pauljadam.com/demos/css-line-through-del-ins-accessibility.html

So I deleted the tag.

### Screenshots
image
image
image

1. Add a product with sale to your cart.
2. Check the product, on both mobile and desktop, the sale price should be gray and has a strikethrough.
3. Check All Product, the same should be applied there.


assets/js/base/components/cart-checkout/product-price/style.scss
```diff
@@ -3,3 +3,7 @@
 		margin-left: 0.5em;
 	}
 }
+.wc-block-components-product-price__regular {
+	text-decoration: line-through;
+	opacity: 0.6;
```
senadir, line 8: I tested some shades and 0.6 was the closest to the designs.

nerrad:
Including @Aljullu on this issue for review. I will try to get to it at some point, but with a busy week I don't want merged blocked by my lack of action here.

haszari:
Looks like this fixes #3080 as well 🎉 ! I've added to description, please keep in mind when testing / merging.

Aljullu (commented):
Confirming it fixes the issues, but I think we should try to keep the `<del>` tags in prices. They are used by some themes to style the regular price.

For example, these screenshots are with [OceanWP](https://oceanwp.org/):

| `main` branch | This branch |
| --- | --- |
| imatge | imatge |

Ideally, we should just write the markup and let themes add the styling on top of it.

senadir:
I added it back, can you confirm that it works? it has no effect on desktop because the tag is inline and its children are block.

haszari:
Saw this issue when testing 3.3.0, seems to be exactly the same issue as in original PR description:

Screen Shot 2020-09-02 at 12 21 07 PM

Screen Shot 2020-09-02 at 12 21 01 PM

assets/js/base/components/cart-checkout/product-price/style.scss
Aljullu: I don't think we should force a specific opacity, that should be handled by the theme. Also, notice adding `opacity: 0.6` here, makes text not accessible in Storefront with Firefox (`#d2d2d2` over `#ffffff`).

imatge

---

I see the original issue (`main` branch in Chrome or Safari) is because `opacity` is not inherited from the theme because there is a `<span>` inside a `<del>` and it has `display: block`. I wonder if something like [this](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3065/files#diff-c8f144109dd5ca6eef80b585bbd79a6eR62-R76) would work, so classes are applied directly to the `<del>` tag:

```JSX
<FormattedMonetaryAmount
	renderText={ ( val ) => (
		<del
			className={ classNames(
				'wc-block-components-product-price__regular',
				className
			) }
		>
			{ val }
		</del>
	) }
	currency={ currency }
	value={ regularValue }
/>
```

Doing the change above, I think we can also remove the `text-decoration: line-through;` declaration.

Actually, taking a closer look, all issues fixed in this PR seem to be fixed in #3065 as well. (Too bad I didn't notice when creating the PR :facepalm: ). So we can either:
1. Update this PR as explained above, merge it and rebase #3065.
2. Get #3065 reviewed and if it does fixes the issues, close this PR in favor of that one.

Aljullu:
@senadir #3065 has been merged and it seems to fix the same issues as this PR. Do you think it makes sense closing this PR now?

PR #3090, opened 2020-08-31
Use wp_login_url instead of hardcoding login path

### LINKED ISSUE #3037: Give developers access to change the login link when user is not logged in directly into the checkout block

senadir:
We had `wp-login.php` hardcoded in our code, this prevented some security plugins from changing its value, plugins like https://wordpress.org/plugins/change-wp-admin-login/ for example, in this pr we use [`wp_login_url`](wp_login_url) which has a `login_url` filter.

Fixes #3037

1. In an incognito window and add some products and then go to Checkout.
2. See the login URL, it should take you to `wp-login.php?redirect_to=[CURRENT_URL]`
3. Enable checkout for users only from WooCommerce -> Settings -> Accounts -> uncheck `Allow customers to place orders without an account`.
4. Visit Checkout again, the link should do the same.
5. Install and activate https://wordpress.org/plugins/change-wp-admin-login/.
6. Visit Settings -> Permalinks and change the login URL to something else.
7. Do 1,2,3 and 4 again, confirm that your value is respected.


Aljullu (approved):
Works like a charm. :ok_hand:

Issue #2319, opened 2020-04-27
Input on checkout lags on Firefox mobile


peterfabian:
**Describe the bug**
It's a bit hard to describe because it's also based on a feeling of increased latency between the finger and the output on the screen, but the input fields on the checkout block lag quite a bit and miss some letters when typing/deleting quickly. This might be more pronounced on older devices. 

From what I saw, it doesn't happen on the old checkout screen with the same browser, so it seems like there is some additional processing happening that slows it down? Also, Chrome does not suffer from the same behavior, IIRC. 

Please see the vid below:
https://youtu.be/-c_HOuG_5xs

**To Reproduce**

1. In mobile Firefox, open 100 tabs
2. Navigate to the WC store, add item to cart
3. Go to checkout
4. Try to fill in any field in the Shipping options

**Expected behavior**
More snappy reactions.

Core version: 5.4
WooCommerce version: 4.0.1
WooCommerce Blocks version: 2.6.0-beta.1
Site language: en_US
Any other plugins installed:

```
Active plugins (8)
--
Companion Plugin |   | by Osk – 1.11
Facebook for WooCommerce |   | by Facebook – 1.11.0
Jetpack by WordPress.com |   | by Automattic – 8.4.2
Mailchimp for WooCommerce |   | by Mailchimp – 2.3.6
WooCommerce Stripe Gateway |   | by WooCommerce – 4.3.3
WooCommerce Blocks |   | by Automattic – 2.6.0-beta.1
WooCommerce - ShipStation Integration |   | by WooCommerce – 4.1.35
WooCommerce |   | by Automattic – 4.0.1
Smartphone (please complete the following information):
```

Device: Galaxy S8+
OS: Android 9
Browser Firefox 68.7.0

nerrad:
This does sound like a potential performance issue.

PR #3106, opened 2020-09-02
Split Checkout form component into smaller files (II)

Aljullu:
This is a follow-up of #3062 and part of the fixes I will add for #2319.

Do some smoke testing of the Checkout block:
* create a new page and add the Checkout block, change some attributes in the sidebar and verify changes take place
* verify the Checkout block works in the frontend: try changing your address, picking a different shipping and payment method, etc.
* verify you can make a purchase

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/form/address-step.js
+++ b/assets/js/blocks/cart-checkout/checkout/form/address-step.js
@@ -1,0 +1,96 @@
+/**
+ * External dependencies
+ */
+import { useMemo } from '@wordpress/element';
+import PropTypes from 'prop-types';
+import { useCheckoutAddress } from '@woocommerce/base-hooks';
+import { useShippingDataContext } from '@woocommerce/base-context';
+
+/**
+ * Internal dependencies
+ */
+import BillingFieldsStep from './billing-fields-step';
+import ContactFieldsStep from './contact-fields-step';
+import ShippingFieldsStep from './shipping-fields-step';
+import './style.scss';
+
+const AddressStep = ( {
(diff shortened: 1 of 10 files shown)


assets/js/blocks/cart-checkout/checkout/form/address-step.js
```diff
@@ -0,0 +1,96 @@
...
+					addressFieldsConfig={ addressFieldsConfig }
+					billingFields={ billingFields }
+					defaultAddressFields={ defaultAddressFields }
+					requirePhoneField={ requirePhoneField }
+					setPhone={ setPhone }
+					setShippingAsBilling={ setShippingAsBilling }
+					setShippingFields={ setShippingFields }
+					shippingAsBilling={ shippingAsBilling }
+					shippingFields={ shippingFields }
+					showPhoneField={ showPhoneField }
```
senadir, line 73: I feel like we're passing too many props to each component, maybe we can move the hook to inside `ShippingFieldsStep` and `BillingFieldsStep`, this is unrelated to this PR, just an idea.

senadir (approved):

assets/js/blocks/cart-checkout/checkout/form/address-step.js
Aljullu: I found some weird behavior when calling `useCheckoutAddress()` from `ShippingFieldsStep` and `BillingFieldsStep` separately, that's why I moved it one level up so the hook is shared by both components. I will proceed merging as-is, we can investigate it further in the future.

PR #3107, opened 2020-09-02
Wrap complex computations in Country/State inputs with useMemo

### LINKED ISSUE #2319: Input on checkout lags on Firefox mobile

Aljullu:
While investigating #2319, I noticed `CountryInput` and `StateInput` components were taking quite a bit of time to render. I think we can improve that wrapping the options parse function in a `useMemo()` hook.

### Screenshots & performance info

Using Chrome devtools with CPU 6x slowdown ([see how](https://calibreapp.com/blog/react-performance-profiling-optimization)), I could see a small performance improvement. Notice the screenshots below are just an example, performance really varies each time the component is rendered.

_Before:_
imatge

_After:_
imatge

Also, notice this PR by itself doesn't fix 2319, but I still think it's worth merging.

Go to the Checkout block and verify you can select a custom country and state and no JS error is shown in the console when changing it.

### DIFF
--- a/assets/js/base/components/country-input/country-input.js
+++ b/assets/js/base/components/country-input/country-input.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies
  */
+import { useMemo } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import PropTypes from 'prop-types';
 import { decodeEntities } from '@wordpress/html-entities';
@@ -27,10 +28,14 @@ const CountryInput = ( {
 		'woo-gutenberg-products-block'
 	),
 } ) => {
-	const options = Object.keys( countries ).map( ( key ) => ( {
-		key,
-		name: decodeEntities( countries[ key ] ),
-	} ) );
+	const options = useMemo(
+		() =>
+			Object.keys( countries ).map( ( key ) => ( {
+				key,
+				name: decodeEntities( countries[ key ] ),
+			} ) ),
+		[ countries ]
+	);
 
 	return (
 		<div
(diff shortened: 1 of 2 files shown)


haszari (approved):
Nice surgical optimisation @Aljullu !

In my testing (thanks for linking to perf tools post, very helpful) I saw `CountryInput` go from 120-160 ms (on `main`) down to < 20 ms on this branch. So that's a nice improvement 🔥

PR #3108, opened 2020-09-02
Create DebouncedValidatedTextInput component

### LINKED ISSUE #2319: Input on checkout lags on Firefox mobile

Aljullu:
Fixes #2319.
Based on #3106.

On each input change in the Checkout component we were updating the shipping/billing address in the context. That was an expensive operation because there are many components subscribed to those contexts, so many components were re-rendering when typing.

This PR adds a `DebouncedValidatedTextInput` component, which is a wrapper around `ValidatedTextInput` but with a debounced `onChange` prop, this avoids updating the contexts on each key pressed.

### Screenshots

While performance difference is quite subtle and difficult to notice in a video, you can see how deleting characters is much faster in the second video:

_Before:_
https://cldup.com/zk5ZOiMG00.mp4

_After:_
https://cldup.com/0q-bXKASIu.mp4

1. With a lower-end device or Chrome devtools with CPU: 6x slowdown ([see more info](https://calibreapp.com/blog/react-performance-profiling-optimization)), interact with the Checkout text input fields.
2. Verify they are much more responsive than before.
3. Smoke test the Checkout block:
	* create a new page and add the Checkout block, change some attributes in the sidebar and verify changes take place
	* verify the Checkout block works in the frontend: try changing your address, picking a different shipping and payment method, etc.
	* verify you can make a purchase

### Changelog

> Performance improvements when interacting with the Cart and Checkout input fields.

### DIFF
--- a/assets/js/base/components/cart-checkout/address-form/index.js
+++ b/assets/js/base/components/cart-checkout/address-form/index.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
-import { ValidatedTextInput } from '@woocommerce/base-components/text-input';
+import { DebouncedValidatedTextInput } from '@woocommerce/base-components/text-input';
 import {
 	BillingCountryInput,
 	ShippingCountryInput,
@@ -172,7 +172,7 @@ const AddressForm = ( {
 				}
 
 				return (
-					<ValidatedTextInput
+					<DebouncedValidatedTextInput
 						key={ field.key }
 						id={ `${ id }-${ field.key }` }
 						className={ `wc-block-components-address-form__${ field.key }` }
--- a/assets/js/base/components/country-input/country-input.js
+++ b/assets/js/base/components/country-input/country-input.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies
  */
+import { useMemo } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import PropTypes from 'prop-types';
 import { decodeEntities } from '@wordpress/html-entities';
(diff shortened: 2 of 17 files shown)


senadir (approved):
I didn't review the code, but I can confirm that this fixes the performance issues, thank you Albert.

PR #3109, opened 2020-09-03
Fix product reviews schema date fields to use new (WP 5.5) `date-time` format 

### LINKED ISSUE #3005: Fix incorrect schema for `ProductReviewSchema` on `date-time` types
WordPress 5.5 introduces a [doing_it_wrong notice for schema types that are non-standard](https://make.wordpress.org/core/2020/07/16/rest-api-parameter-json-schema-changes-in-wordpress-5-5/) (i.e `date-time`) types. Currently this doesn't affect us significantly because the notice only applies to put/post requests. However we do have [one instance of this in our REST endpoints](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a404e5b24814240f15e50aa4f983b787bb9b36f1/src/StoreApi/Schemas/ProductReviewSchema.php#L52-L69). The correct implementation would be to have `{ type: 'string', format: 'date-time' }`. 

This isn't an urgent fix needed, but still trivial and should be done sooner vs later.

haszari:
Fixes #3005

This PR fixes the schema for product reviews API endpoint. 

In WordPress 5.5, the `type` [must be one of the primitive types](https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#primitive-types); previously we had `date-time`. 

Schemas can [declare the intended format of `type: string`](https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#format), and for our dates we should use `date-time` aka [RFC3339](https://tools.ietf.org/html/rfc3339#section-5.8).

So, this PR uses `type: string, format: date-time` for the two review date fields. 

I did a quick search to see if there were other API dates that should use this format - looks like this is the only API affected.

To test this I added some reviews and an `All Reviews` block and checked the response. 

### Screenshots

Screen Shot 2020-09-03 at 3 57 51 PM

1. Add some reviews.
2. Add an all reviews block.
3. View on front end and inspect API response, compare with schema.

I only tested with WordPress 5.5 - should we test with 5.3 or 5.4 as well? 

Also keen to confirm this fixes the `doing it wrong` error - is there a way to reproduce that (on `main`/without these changes)? Or is this only validated when POSTing?

### Changelog

> dev: Corrected schema for date fields in product reviews API.


haszari:
I notice that the dates are both the same - one maybe should have a timezone offset. Not sure if this is a bug, or if so what impact it would have…

Aljullu (approved):
> I only tested with WordPress 5.5 - should we test with 5.3 or 5.4 as well?

I tested with 5.4 and everything looked fine. :ok_hand:

haszari:
> That should depend on the store timezone (in Settings > General > Timezone) that was selected when the review was written. Maybe you had UTC+0?

Yep that was it 👍

Issue #3110, opened 2020-09-03
Storefront child themes issues with Cart and Checkout blocks


Aljullu:
I tested all themes listed in the [Storefront Themes category in WooCommerce.com](https://woocommerce.com/product-category/themes/storefront-child-theme-themes/) with the Cart and Checkout blocks. Below is a list of issues I detected. Notice some of this issues might be in WC Blocks, some others in Storefront and some others in the child theme. However, I'm listing all of them here so we can investigate it further and open issues in the relevant repos.

I think, some of the themes listed in that page are not child themes of Storefront. So in the list below I split them between the ones that are Storefront child themes (they have some minor issues) and other themes (the experience is mostly broken).

### Arcade
_(I couldn't find any issue)_

### Bistro
_(I couldn't find any issue)_

### Bookshop
* Panel titles and selects have a green background.
![imatge](https://user-images.githubusercontent.com/3616980/92102380-d2993500-edde-11ea-9978-8d2e671b22ca.png)

### Boutique
_(I couldn't find any issue)_

### Deli
* Order summary quantity badge is not readable.
![imatge](https://user-images.githubusercontent.com/3616980/92101305-39b5ea00-eddd-11ea-9fa4-00bf895c6f32.png)
* Text input and button don't have the same height. (Similar issue is affecting Block Shop)
![imatge](https://user-images.githubusercontent.com/3616980/92101373-50f4d780-eddd-11ea-951a-16ac1837ff3f.png)
* Proceed to Checkout button in Cart has a weird brown background. (Similar issue is affecting Arcade, Block Shop and ToyShop)
![imatge](https://user-images.githubusercontent.com/3616980/92101446-6bc74c00-eddd-11ea-8599-6112327e3bc9.png)

### Galleria
* Panel titles have a border on hover that causes the content to jump.
![Peek 2020-09-03 12-04](https://user-images.githubusercontent.com/3616980/92101590-a204cb80-eddd-11ea-88bf-2709f07434d3.gif)

### Homestore
* Quantity input overflows in mobile.
![imatge](https://user-images.githubusercontent.com/3616980/92101769-e1331c80-eddd-11ea-818a-1176b9d879bb.png)

### Hotel
* Proceed to Checkout button has no background/border.
![imatge](https://user-images.githubusercontent.com/3616980/92124300-e738f580-edfd-11ea-9fd1-b4f1fdc4003f.png)
* Cart: there is no padding between table contents and border.
[...]

PR #3111, opened 2020-09-03
Fix Cart & Checkout sidebar layout broken in some themes

Aljullu:
Cart and Checkout blocks were not playing well with themes that didn't define `box-sizing: border-box;` for all elements. I discovered that while investigating #3110, but it has been for some time in my todo list, so I decided to create a PR directly.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Install and activate [Artisan](https://woocommerce.com/products/artisan/) or [Threads](https://woocommerce.com/products/threads/) theme.
2. Go to the Cart or Checkout blocks and verify the sidebar is shown on the right instead of below.
3. Switch back to Storefront or Twenty Twenty and verify there are no regressions.

### Changelog

> Fixed a style conflict that was affecting some themes and was breaking their sidebar layout in the Cart and Checkout blocks.

### DIFF
--- a/assets/js/base/components/sidebar-layout/style.scss
+++ b/assets/js/base/components/sidebar-layout/style.scss
@@ -5,6 +5,7 @@
 	position: relative;
 
 	.wc-block-components-main {
+		box-sizing: border-box;
 		margin: 0;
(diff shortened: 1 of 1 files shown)


senadir (approved):
Can confirm that it fixes that issue.

PR #3115, opened 2020-09-06
Add @woocommerce/eslint-plugin dependency

### LINKED ISSUE #2957: Implement new `@woocommerce/eslint-plugin` package.
The `@woocommerce/eslint-plugin` package has been released and can be used to configure eslinting for Woo projects.

I'm assigning this to myself because I created the package and would like to fix any bugs that might popup as we implement.

This package can replace all our customized eslint configuration and helps standardize Woo projects around a common configuration for JS linting.

nerrad:
Fixes #2957

WooCommerce now has a standard eslint configuration (that embraces and extends off of the `@wordpress/eslint-plugin` package) that we can adopt and use for convention enforcement going forward. There isn't significant differences from what we are doing already but there are a few things worth nothing with this package install:

- A number of corollary package imports are no longer needed (prettier, various `@wordpress/*` imports etc). Except there's a known issue in `@wordpress/eslint-plugin` being pulled in as a dependency that causes a lint script error if a project doesn't have a `.prettierrc` config file (or config object) defined. So for now, our `.prettierrc` config needs to be kept intact.
- `@wordpress/scripts` is updated to `12.0.0` and along with it comes a more up-to-date `@wordpress/eslint-plugin` package. The update in the eslint-plugin package brings updates numerous more eslint rules including jsdoc rulesets and i18n related rulesets.
- updated linting for react testing library.
- remove obsolete custom dependency-grouping plugin in our repo because that's been replaced by the custom rule in the `@woocommerce/eslint-plugin` package.
- fixed some syntax changes for eslint plugins in our custom feature-flag plugin.
- updating `@wordpress/scripts` broke our test configurations. I had to fix both unit tests and e2e tests as a result.
- with the e2e script switch to use `wp-scripts` for the test commands, a side benefit is that puppeteer is now configured to only install chromium on demand, so `npm install` should go a bit faster for folks not worried about running e2e tests locally.

Due to the above and to avoid having a huge changeset in one pull. This pull primarily handles the changes in the eslint configuration and a run of prettier fix across all files impacted by the prettier update. New rule violations have been temporarily added to the `rules` config disabling the rules so they can be handled in individual pulls separate from this one.

**For Reviewing:**

I've split up things into separate commits so you can review by commit for a smaller changeset. The prettier fix commit was an auto-fix via script.

## Testing

Do note, that you should probably do `npm ci` on this branch so that you get a fresh install of packages and if you're using VSCode, you'll want to close and re-open it. I think there's some caches that need reset in the IDE.  I'm not sure what's needed for other IDEs here.

- Anything related to builds and tests, and linting should be verified by travis.
- You can do a smoke test of a build locally to ensure there are no javascript errors in the console. However, I'm confident this doesn't impact builds being it's mostly a eslint configuration and test environment related changes.


haszari (approved):
Nice to streamline this config and align more closely with Woo/WordPress core linter rules. Didn't see anything alarming in the code changes - all looking good. I did a clean build locally and smoke tested a few blocks (All Products + Price Filter, Featured Category, New Products) and everything worked well.

nerrad:
Thanks Rua, I hadn't put this up for review yet because I'm still cleaning up some failures on travis (hope to push something soon).

package.json
```diff
@@ -116,43 +111,32 @@
 		"core-js": "3.6.5",
 		"create-file-webpack": "1.0.2",
 		"cross-env": "6.0.3",
-		"css-loader": "3.6.0",
```
Aljullu, line 119: I'm not sure to understand why some dependencies were removed.

Aljullu (approved):
Everything is looking good to me. I left some comments/questions below about some things I'm not sure to understand, but approving from my part. :ok_hand:

tests/e2e/config/setup.js
```diff
@@ -24,13 +24,17 @@ module.exports = async ( globalConfig ) => {
 	await setupPuppeteer( globalConfig );
 
 	try {
+		// do setupSettings separately which hopefully gives a chance for WooCommerce
+		// to be configured before the others are executed.
+		await setupSettings();
+		const pages = await createBlockPages();
```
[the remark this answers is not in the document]
nerrad: oops yup, it also turns out that this has nothing to do with what was needed so I'm moving them back into the promise array.

package.json
nerrad: They are now included in the version of `wp-scripts` we're shipping with. It makes it easier for managing these dependencies because we just need to update `wp-scripts`.

senadir (approved):
I did a whole look up, looks good to me, as long as tests are running and passing, this can be shipped.

package.json
nerrad: I should add, there may be more dependencies we can remove as well, I didn't want to get too carried away in this pull though.

tests/e2e/config/setup.js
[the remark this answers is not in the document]
nerrad: Actually it looks like this _was_ needed for a fix (this has been a "fun" pull to work through).  But the duplicate call to createBlockPages can definitely be removed.

PR #3131, opened 2020-09-09
Ensure shopper saved card is used as default payment method (default was being overwritten in some circumstances)

### LINKED ISSUE #3120: Checkout block: Wrong payment method used (e.g. BACS) when shopper selects saved payment method (e.g. saved Stripe CC)
## Describe the bug
When submitting an order using a previously-saved payment method, the order is completed using a different payment method. For example, user selects `Visa ending in 4242` (saved Stripe CC), but the order is processed using BACS.

Apologies if this is tricky to reproduce – I've seen it a few times today, on different stores (ephemeral / production build, local dev), so I'm confident there's a real issue. But it may require specific setup of payment methods (or other config).

## To reproduce
1. Set up checkout page using block.
2. Enable BACS and Stripe payment methods (only, though may reproduce with other combinations).
3. Complete a purchase using a new payment method, Stripe CC, check `Save payment information to my account for future purchases`.
4. Add something to cart, proceed to checkout.
5. **Leave the default selected payment method - i.e. the first saved payment method**. Submit checkout.

Expected: Stripe CC used for order.

Actual: BACS used for order.

### Screenshots
Here's a gif of this happening (this was on ephemeral with production versions of everything):

Screen Capture on 2020-09-08 at 16-34-21

Here's a pic of my setup in case something's relevant - the order may be important.

Screen Shot 2020-09-08 at 4 46 01 PM

### Environment
WooCommerce Blocks plugin version: 3.2 (release), 3.3.0 (release), 3.4.0-dev

haszari:
Fixes #3120

The issue here is that when a user has saved payment method(s) - e.g. a Stripe saved credit card (token), the logic that sets an initial default payment method can overwrite the saved payment method. 

~~This PR fixes this by adding a guard to the default / init effect. If there's a saved payment method, it now exits early; previously was overwriting the saved payment method, causing #3120.~~

Update: this PR now uses the [functional form of `setState`](https://reactjs.org/docs/faq-state.html#how-do-i-update-state-with-values-that-depend-on-the-current-state) to apply the default payment method. This ensures that the default is only set if there is no payment method active (selected); previously this decision was made using a stale copy of state. Hat tip @Aljullu for the suggestion 🎉 

- Set up checkout page using block.
- Enable BACS and Stripe payment methods (only, though may reproduce with other combinations). Note BACS needs to be earlier in the order than Stripe (drag handle in `WooCommerce > Settings > Payments`).
- Complete a purchase using a new payment method, Stripe CC, check `Save payment information to my account for future purchases`.
- Add something to cart, proceed to checkout.
- Leave the default selected payment method - i.e. the first saved payment method. Submit checkout.

Should always complete the purchase with the correct payment method - the one visibly selected when user clicks submit, i.e. the saved card.

Please also test a variety of other payment method scenarios and ordering - with / without saved cards (are there any other gateways that support saved payment methods?), with more/less gateways available, in different orders, and with dynamically-available gateways (COD can depend on shipping option). In all cases confirm the following:

- There is a payment method tab selected by default. 
- Submitting checkout without touching payment section uses the correct payment method (i.e. saved card, payment method tab).
- Selecting a non-default saved card (from the default) works correctly, uses correct card/token (may need to check network requests to check token).
- Selecting a non-default payment method uses correct method.
- Saving a payment method (card) works and is available for use next checkout with that account.
- Payment tab order should match the configured order.

### Changelog

> Fix: Ensure that the checkout correctly uses shopper's previously-saved payment method when selected by default (e.g. Stripe saved card). In some circumstances, a different payment method was used (e.g. BACS).

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -341,6 +341,11 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 			return;
 		}
 
+		// Customer has saved card/payment methods - no need to set a default.
+		if ( customerPaymentMethods ) {
+			return;
+		}
+
 		// If there's no active payment method, or the active payment method has
 		// been removed (e.g. COD vs shipping methods), set one as active.
 		if (
@@ -356,6 +361,7 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 		paymentMethodsInitialized,
 		paymentData.paymentMethods,
 		setActivePaymentMethod,
+		customerPaymentMethods,
 	] );
 
 	// emit events.
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -353,12 +353,7 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 			}
 			return currentActivePaymentMethod;
 		} );
-	}, [
-		paymentMethodsInitialized,
-		paymentData.paymentMethods,
-		setActive,
-		customerPaymentMethods,
-	] );
+	}, [ paymentMethodsInitialized, paymentData.paymentMethods, setActive ] );
 
 	// emit events.
 	useEffect( () => {
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -341,26 +341,22 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 			return;
 		}
 
-		// Customer has saved card/payment methods - no need to set a default.
-		if ( customerPaymentMethods ) {
-			return;
-		}
-
-		// If there's no active payment method, or the active payment method has
-		// been removed (e.g. COD vs shipping methods), set one as active.
-		if (
-			! activePaymentMethod ||
-			! paymentMethodKeys.includes( activePaymentMethod )
-		) {
-			setActivePaymentMethod(
-				Object.keys( paymentData.paymentMethods )[ 0 ]
-			);
-		}
+		setActive( ( currentActivePaymentMethod ) => {
+			// If there's no active payment method, or the active payment method has
+			// been removed (e.g. COD vs shipping methods), set one as active.
+			if (
+				! currentActivePaymentMethod ||
+				! paymentMethodKeys.includes( currentActivePaymentMethod )
+			) {
+				dispatch( statusOnly( PRISTINE ) );
+				return Object.keys( paymentData.paymentMethods )[ 0 ];
+			}
+			return currentActivePaymentMethod;
+		} );
 	}, [
-		activePaymentMethod,
 		paymentMethodsInitialized,
 		paymentData.paymentMethods,
-		setActivePaymentMethod,
+		setActive,
 		customerPaymentMethods,
 	] );
 


assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -341,6 +341,11 @@ export const PaymentMethodDataProvider = ( { children } ) => {
...
 		}
 
+		// Customer has saved card/payment methods - no need to set a default.
+		if ( customerPaymentMethods ) {
+			return;
+		}
+
 		// If there's no active payment method, or the active payment method has
 		// been removed (e.g. COD vs shipping methods), set one as active.
 		if (
```
Aljullu, line 351: There is an alternative approach that doesn't require the `customerPaymentMethods` dependency. It would involve using `useState` functional updates. Something like:

```JSX
setActive( ( currentActivePaymentMethod ) => {
	// If there's no active payment method, or the active payment method has
	// been removed (e.g. COD vs shipping methods), set one as active.
	if (
		! currentActivePaymentMethod ||
		! paymentMethodKeys.includes( currentActivePaymentMethod )
	) {
		return Object.keys( paymentData.paymentMethods )[ 0 ];
	}
	return currentActivePaymentMethod;
} );
```

This would ensure `currentActivePaymentMethod` is evaluated right at the moment when the state is updated, so the payment method activated in `SavedPaymentMethodOptions` would already be active.

I didn't do a lot of testing of this approach, so feel free to ignore, but wanted to propose it in case you think it might be a good alternative.

Aljullu (approved):
Thanks for investigating this issue and creating a fix. If I understand what's going on is that in `SavedPaymentMethodOptions` we are [setting the saved payment method as the active payment option](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/base/components/payment-methods/saved-payment-method-options.js#L127), but that was [overridden in the `PaymentMethodDataProvider` useEffect](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js#L350). With this PR, we bail out soon if there are saved customer payment methods so the issue is fixed, is that correct?

Below I suggested an alternative approach that I think would work as well and is more generic because it doesn't handle the customer payment methods case specially. But I might have missed some special cases, so feel free to ignore if you don't think it will work.

> Please also test a variety of other payment method scenarios and ordering - with / without saved cards (are there any other gateways that support saved payment methods?), with more/less gateways available, in different orders, and with dynamically-available gateways (COD can depend on shipping option).

I did a variety of tests and everything seemed to work alright. Seeing there are so many different scenarios is a good proof that it would be nice adding e2e tests at some point! :slightly_smiling_face:

Aljullu (changes requested):
Actually I think I found an issue. If I open the Checkout page in a private window (incognito), the first tab is not selected. That happens because `customerPaymentMethods` is an (empty) object and evaluates to true so the effect returns too early.

imatge

Can you reproduce that too?

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
haszari: Smart! I like this, and of course since it's more general it avoids the issue with my truthy test below. Thanks for picking up on this, much better fix 🚀 

The crux of the issue is that we were previously using a stale copy of `currentActivePaymentMethod`; so it makes sense that we'd fix this by using the functional form of `setState`. The principle to keep in mind here (just reminding myself!) - [if a state update depends on the previous value, use functional form](https://reactjs.org/docs/faq-state.html#how-do-i-update-state-with-values-that-depend-on-the-current-state). 

I've made the change and I'll do some more testing of a few scenarios. Note the other change from the sketch above is the reset of checkout status to `PRISTINE` when applying the default.

haszari:
I've tested the new approach (functional `setState` to use fresh version of active payment method state) and didn't find any issues. Tested checkouts with and without saved cards, saving new cards, reordering payment methods in admin, and submitting checkout with default-selected method or manually selecting. 

@Aljullu ready for another review!

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -341,21 +341,23 @@ export const PaymentMethodDataProvider = ( { children } ) => {
...
+			}
+			return currentActivePaymentMethod;
+		} );
 	}, [
-		activePaymentMethod,
 		paymentMethodsInitialized,
 		paymentData.paymentMethods,
-		setActivePaymentMethod,
+		setActive,
+		customerPaymentMethods,
```
Aljullu, line 360: Is this dependency needed?

Aljullu (approved):
Thanks for applying the feedback. I did another round of testing and everything looked good. I just added a small comment about an effect dependency that I guess can be removed, but pre-approving. :+1: 

Travis tests are failing, but I don't think it's related to this PR. I will spend some time today investigating what's going on with them.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
haszari: No, and it looks like it hasn't been needed for a while! Removed it and retested (just in case) 👍

haszari:
Squash-merging this as I believe the test failure is unrelated - see #3139

PR #3135, opened 2020-09-09
Remove selectedToken state from PaymentMethods

Aljullu:
Small refactor: there is no need to keep `selectedToken` in the state of PaymentMethods if we can directly read it from `paymentMethodData` in the context.

1. Without any previous saved payment method, open the Checkout block and verify the payment method tabs are displayed normally.
2. With a user with previous saved payment methods<sup>*</sup>, open the Checkout block and verify the payment method tabs are hidden by default until you select `Use a new payment method`.

<sup>*</sup> In order to save a payment method with a user. Enable the WooCommerce Stripe plugin, set the keys and make a purchase with a user selecting the `Save payment information to my account for future purchases.` option. Next time you visit the Checkout with that user, the saved payment method will show up.

### DIFF
--- a/assets/js/base/components/payment-methods/payment-methods.js
+++ b/assets/js/base/components/payment-methods/payment-methods.js
@@ -11,7 +11,6 @@ import {
 	cloneElement,
 	useRef,
 	useEffect,
-	useState,
 	useCallback,
 } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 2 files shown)


nerrad (approved):
Looks good to me Albert! Including your tests, I also tested various payment methods and starting then cancelling Express payments and everything seemed to work well. Let's :shipit: 👍

PR #3139, opened 2020-09-10
Update package-lock.json (fix e2e tests)

Aljullu:
It looks like Travis tests were failing because they couldn't find puppeteer. Adding it back to package.json fixes the issue.

cc @nerrad since puppeteer was removed in #3115. I'm not sure why tests were passing in that PR but suddenly stopped passing in `main`. Let me know if I'm missing something or you have a better idea of what might be going on. :slightly_smiling_face:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3064,20 +3064,20 @@
 			}
 		},
 		"@octokit/endpoint": {
-			"version": "6.0.5",
-			"resolved": "https://registry.npmjs.org/@octokit/endpoint/-/endpoint-6.0.5.tgz",
-			"integrity": "sha512-70K5u6zd45ItOny6aHQAsea8HHQjlQq85yqOMe+Aj8dkhN2qSJ9T+Q3YjUjEYfPRBcuUWNgMn62DQnP/4LAIiQ==",
+			"version": "6.0.6",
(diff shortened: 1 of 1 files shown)


nerrad:
In the pull I did the changes in, the package-lock.json has `puppeteer` defined as a top-level dependency. Currently in main, it is not a top-level dependency so a subsequent commit has modified the package-lock.json.

Travis does `npm ci` so it installs explicitly and only from package-lock.json.

I'm not sure why or how this changed, but if we need to add puppeteer as an explicit top-level dependency (which we shouldn't give how `wp-scripts` has it as a dependency, we should do this instead:

```
"puppeteer": "npm:puppeteer-core@3.0.0",
```

I wonder if the problem was created by a renovate commit? 

Can you try this Albert? With a local checkout of the main branch:

- delete `package-lock.json` and `node_modules`
- run `npm install`.

With the above, does the `package-lock.json` have `puppeteer` as a top-level dependency?

Aljullu:
Thanks for the explanations @nerrad! Yes, you're right, the issue was the missing puppetter in the root level of dependencies in `package-lock.json`. As you said, the issue started appearing after I merged a renovate PR.

I updated this PR with a newly generated `package-lock.json` after a fresh install and tests seem to pass again.

If we merge this, next step I will rebase other renovate PRs and see if tests pass there or the same issue is back.

nerrad (approved):

PR #3143, opened 2020-09-14
Update .gitattributes to reduce what is pulled into projects via composer.

### LINKED ISSUE #3140: Remove unnecessary files from release tag.
Currently in our release tags, things like `storybook` and various configuration files in the root directory are included in the tag when it's unnecessary (and means that composer pulls in these files to the package bundled with WooCommerce core releases).

We still need to include the `assets` folder because of WordPress.org releases, but I'd like to investigate doing the following:

- review what files/folders are _not_ needing included in any _release_ tags.
- make any necessary changes to the build scripts so that release tags are built without the extraneous files/folder
- retain those files in the `-dev` tag created for each release (so we can checkout the full development builds if we need to for some reason in the future).

This is assigned to the 3.4.0 milestone because 3.4.0 will be going into the next release of WooCommerce core and removing these extraneous files make it much cleaner for WC Core (especially given the various context WooCommerce core is used in).

Since I'm doing the next release of Woo Blocks, if needed, I can take care of this as a part of the release process.

nerrad:
Fixes #3140

This pull updates the `.gitattributes` file to cover all folders and files we want excluded when this project is pulled in as a dependency via composer. 

## Testing

If you run `git archive --format=zip -o test.zip HEAD` while checked out on this branch, and then check the contents of test.zip, it should only have the following files:

Image 2020-09-14 at 11 43 25 AM

Note that `vendor` and `build` folders are not explicitly excluded but are just not showing up in this test because `git archive` only includes what's being tracked by the branch. Our releases add `vendor` and `build` folders so composer will pull those in too.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,17 @@
+# Ignore all hidden files and directories
 /.* export-ignore
+
+# Ignore all markdown files
 /*.md export-ignore
+
+# Ignore following directories and contents
 /node_modules* export-ignore
 /tests export-ignore
 /bin export-ignore
+/docs export-ignore
+/storybook export-ignore
+
+# Ignore following configuration files
 /phpcs.xml export-ignore
 /phpunit.* export-ignore
 /composer.lock export-ignore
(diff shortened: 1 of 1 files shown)


nerrad:
I'm going to merge this in because testing is pretty clear/obvious and it doesn't affect wp.org deploys. If there is an issue with the package bump in core I'm willing to address that when it happens as a part of the release process.

PR #3147, opened 2020-09-15
Fix State label for Spain

Aljullu:
Small PR fixing the _State_ label for Spain. It should be `Province` instead of `State`. You can see WC Core for reference: https://github.com/woocommerce/woocommerce/blob/master/includes/class-wc-countries.php#L1250

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Go to the Checkout block and change the country to Spain.
2. Verify the field below the country changes its label to `Province` instead of `State`.

### Changelog

> In address forms, the 'State' label is now correctly replaced with 'Province' when the country is Spain.

### DIFF
--- a/assets/js/base/components/cart-checkout/address-form/country-address-fields.js
+++ b/assets/js/base/components/cart-checkout/address-form/country-address-fields.js
@@ -198,9 +198,9 @@ const countryAddressFields = {
 	ES: {
 		...postcodeBeforeCity,
 		state: {
-			label: __( 'State', 'woo-gutenberg-products-block' ),
+			label: __( 'Province', 'woo-gutenberg-products-block' ),
 			optionalLabel: __(
-				'State (optional)',
+				'Province (optional)',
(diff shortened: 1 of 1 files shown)


haszari (approved):
Gave this a quick test - looks good to me! 

I wonder if there are any other differences between our `country-address-fields.js` and core's `class-wc-countries.php`?

Aljullu:
> I wonder if there are any other differences between our `country-address-fields.js` and core's `class-wc-countries.php`?

I didn't do a full review, I just changed this one which stood out to me and checked a couple more to make sure it was not a general issue affecting other countries, but couldn't find any other mistakes.

senadir:
Mike exported that file several months ago when updating the components. We now have two sources of truth, which can cause more issues like this in the future.
We might want to either:
- Consolidate that file so that we always use the PHP one.
- Create an automated build process that generates `country-address-fields.js` from that file.

PR #3161, opened 2020-09-18
Use noticeContexts from useEmitResponse instead of hardcoded values

Aljullu:
Inspired by https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3134#discussion_r490623807, this PR replaces all instances of `wc/payment-area` and `wc/express-payment-area` with their values from `useEmpitResponse`'s  `noticeContexts`.

I think changes are quite safe, but a good way to test would be forcing some errors to payment methods and ensuring they appear in the correct places.

1. Install & activate WooCommerce Stripe. 
2. Enable Stripe CC payment method - don't add an api key (or delete the option).
3. Add checkout block to checkout page.
4. On front end, add something to cart and proceed to checkout with an admin user.
5. Verify an error appears in the express payment methods section and in the payment methods step.

checkout-errors

### DIFF
--- a/assets/js/base/components/payment-methods/express-payment/cart-express-payment.js
+++ b/assets/js/base/components/payment-methods/express-payment/cart-express-payment.js
@@ -2,7 +2,10 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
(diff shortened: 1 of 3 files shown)


nerrad (approved):
LGTM! Let's :shipit:

Issue #2122, opened 2020-04-06
Make sure we build in a way for gateway extensions to unregister the integration bundled with blocks.


nerrad:
The goal will be to eventually move block compatibility for gateway extensions to the extensions themselves. This means we'll need a clean way for gateways to override what the blocks register internally for compatibility to provide an easier transition between the blocks and gateways.  This will have to be done before launching the blocks or we could make this harder to implement later in a painless way for merchants (version comparisons etc).

So I've created this issue just to make sure we don't forget this (will flesh out exactly how to do this - or what needs tweaked later).

senadir:
if you have time later can you flesh out the details because this still feels fuzzy to me.

nerrad:
Server side, could likely just remove the actions/hooks that the `Automattic\WooCommerce\Blocks\Payments\Integrations\Stripe.php` class is hooked into. Client side we may need an `unregisterPaymentMethod` and `unregisterExpressPaymentMethod` function as companions to the `registerPaymentMethod` and `registerExpressPaymentMethod` apis.  The tricky part is unregistration would have to be called after registration.

nerrad:
In #3165 an experimental implementation of this was created. So this issue can now be used to determine whether we remove it from experimental or not.

senadir:
Is this what we did for the recent Stripe integration? or would `__experimentalDeRegisterPaymentMethod` be enough here?

nerrad:
We can close this 👍

PR #3165, opened 2020-09-19
Fix broken Express Payment Method use in the Checkout block for logged out or incognito users.

### LINKED ISSUE #3163: Fix: Broken Express Payment methods in checkout block.
## Describe the bug

With the express payment method for a new shopper (non-logged in) it is expected that all they have to do when hitting the checkout block is to click the express payment button, pick their payment details, and once it submits the order will complete.

Currently on all releases of blocks from 3.0.0 to current main branch, this is broken as user information (email address, first and last name) are not updated in the store and the form submit validation kicks in:

Image 2020-09-18 at 4 16 21 PM

## To reproduce

Make sure you are either logged out or using an incognito mode browser instance.

1. Add a product to cart.
2. Load the page with the checkout block.
3. Click express payment (Chrome Pay if using Chrome, Apple Pay if using Safari).
4. Choose account details in the express payment modal and submit. See error.

If you try the above steps using a checked out copy of Woo Blocks `2.9.0` everything will work as expected, submitting the express payment modal will complete the transaction.

## Additional Notes

It looks like this has been out in the wild since the release of 3.0.0, so I don't think we'd need to rush a patch release. However, I still am going to make this a blocker for 3.5.0 because it's a pretty significant bug for shoppers using express payments.

nerrad:
Fixes #3163 

Details about the issue can be found in #3163. The bug was introduced with the work done in #2831. 

In this pull:

- I modified the offending condition causing the bug (See details inline on the changeset).
- I added some tests that both replicate the bug (without the fix) and demonstrate things work after applying the fix (prevents against regressions). The test also verifies that closing express payment method modal will result in the correct active payment state change.
- Needed for the tests was a way to de-register registered payment methods. This is something queued for development (see #2122) anyways. I added the apis as "experimental" for now, since it's only used for the tests I added and this may need some more thought before making it more public.

## To Test

Make sure you are either logged out or using an incognito mode browser instance.

* Add a product to cart.
* Load the page with the checkout block.
* Click express payment (Chrome Pay if using Chrome, Apple Pay if using Safari).
* [ ]  Choose account details in the express payment modal and submit. Verify that the checkout processes correctly.

### Other scenarios

Since any change in the checkout flow can have an unexpected impact (especially since automated test coverage is still small), there's other scenarios that should be tested as a part of this work:

- Start paying with Express payment method and then close the modal and attempt to pay via another payment method. That should work. (note, closing and immediately paying by the already selected _saved payment method_ is currently expected to not work due to [bug outlined here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3165#issuecomment-695348865))
- Try switching between various payment methods before making a payment and completing the order. There should be no errors and the correct payment method is attached to the order.
- Verify COD payment method works as expected (especially when changes in address result in it not being available).
- Try flipping between saved payment methods and new payment methods and verify completing an order works as expected.
- Try the above in incognito and logged in and verify things work as expected.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -344,10 +344,10 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 		setActive( ( currentActivePaymentMethod ) => {
 			// If there's no active payment method, or the active payment method has
 			// been removed (e.g. COD vs shipping methods), set one as active.
-			if (
-				! currentActivePaymentMethod ||
-				! paymentMethodKeys.includes( currentActivePaymentMethod )
-			) {
+			// Note: It's possible that the active payment method might be an
+			// express payment method. So we only reset this if there's no
+			// current active payment method.
+			if ( ! currentActivePaymentMethod ) {
 				dispatch( statusOnly( PRISTINE ) );
 				return Object.keys( paymentData.paymentMethods )[ 0 ];
 			}
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -337,6 +337,10 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 	// Set active (selected) payment method as needed.
 	useEffect( () => {
 		const paymentMethodKeys = Object.keys( paymentData.paymentMethods );
+		const allPaymentMethodKeys = [
+			...paymentMethodKeys,
+			...Object.keys( paymentData.expressPaymentMethods ),
+		];
 		if ( ! paymentMethodsInitialized || ! paymentMethodKeys.length ) {
 			return;
 		}
@@ -345,15 +349,23 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 			// If there's no active payment method, or the active payment method has
 			// been removed (e.g. COD vs shipping methods), set one as active.
 			// Note: It's possible that the active payment method might be an
-			// express payment method. So we only reset this if there's no
-			// current active payment method.
-			if ( ! currentActivePaymentMethod ) {
+			// express payment method. So registered express payment methods are
+			// included in the check here.
+			if (
+				! currentActivePaymentMethod ||
+				! allPaymentMethodKeys.includes( currentActivePaymentMethod )
+			) {
 				dispatch( statusOnly( PRISTINE ) );
 				return Object.keys( paymentData.paymentMethods )[ 0 ];
 			}
 			return currentActivePaymentMethod;
 		} );
-	}, [ paymentMethodsInitialized, paymentData.paymentMethods, setActive ] );
+	}, [
+		paymentMethodsInitialized,
+		paymentData.paymentMethods,
+		paymentData.expressPaymentMethods,
+		setActive,
+	] );
 
 	// emit events.
 	useEffect( () => {
--- a/assets/js/base/context/cart-checkout/payment-methods/test/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/test/payment-method-data-context.js
@@ -1,0 +1,138 @@
+/**
+ * External dependencies
+ */
+import { render, screen, fireEvent, waitFor } from '@testing-library/react';
+import { previewCart } from '@woocommerce/resource-previews';
+import { dispatch } from '@wordpress/data';
+import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
+import {
+	registerPaymentMethod,
+	registerExpressPaymentMethod,
+	__experimentalDeRegisterPaymentMethod,
+	__experimentalDeRegisterExpressPaymentMethod,
+} from '@woocommerce/blocks-registry';
+import { CheckoutExpressPayment } from '@woocommerce/base-components/payment-methods';
+/**
+ * Internal dependencies
+ */
+import {
+	usePaymentMethodDataContext,
+	PaymentMethodDataProvider,
+} from '../payment-method-data-context';
+import { defaultCartState } from '../../../../../data/default-states';
+
+const registerMockPaymentMethods = () => {
+	[ 'cheque', 'bacs' ].forEach( ( name ) => {
+		registerPaymentMethod(
+			( Config ) =>
+				new Config( {
+					name,
+					label: name,
+					content: <div>A payment method</div>,
+					edit: <div>A payment method</div>,
+					icons: null,
+					canMakePayment: () => true,
+					ariaLabel: name,
+				} )
+		);
+	} );
+	[ 'express-payment' ].forEach( ( name ) => {
+		registerExpressPaymentMethod( ( Config ) => {
+			const Content = ( {
+				onClose = () => void null,
+				onClick = () => void null,
+			} ) => {
+				return (
+					<>
+						<button onClick={ onClick }>
+							{ name + ' express payment method' }
+						</button>
+						<button onClick={ onClose }>
+							{ name + ' express payment method close' }
+						</button>
+					</>
+				);
+			};
+			return new Config( {
+				name,
+				content: <Content />,
+				edit: <div>An express payment method</div>,
+				canMakePayment: () => true,
+				paymentMethodId: name,
+			} );
+		} );
+	} );
+};
+
+const resetMockPaymentMethods = () => {
+	[ 'cheque', 'bacs' ].forEach( ( name ) => {
+		__experimentalDeRegisterPaymentMethod( name );
+	} );
+	[ 'express-payment' ].forEach( ( name ) => {
+		__experimentalDeRegisterExpressPaymentMethod( name );
+	} );
+};
+
+describe( 'Testing Payment Method Data Context Provider', () => {
+	beforeEach( async () => {
+		registerMockPaymentMethods();
+		fetchMock.mockResponse( ( req ) => {
+			if ( req.url.match( /wc\/store\/cart/ ) ) {
+				return Promise.resolve( JSON.stringify( previewCart ) );
+			}
+		} );
+		// need to clear the store resolution state between tests.
+		await dispatch( storeKey ).invalidateResolutionForStore();
+		await dispatch( storeKey ).receiveCart( defaultCartState );
+	} );
+	afterEach( async () => {
+		resetMockPaymentMethods();
+		fetchMock.resetMocks();
+	} );
+	it( 'toggles active payment method correctly for express payment activation and close', async () => {
+		const TriggerActiveExpressPaymentMethod = () => {
+			const { activePaymentMethod } = usePaymentMethodDataContext();
+			return (
+				<>
+					<CheckoutExpressPayment />
+					{ 'Active Payment Method: ' + activePaymentMethod }
+				</>
+			);
+		};
+		const TestComponent = () => {
+			return (
+				<PaymentMethodDataProvider>
+					<TriggerActiveExpressPaymentMethod />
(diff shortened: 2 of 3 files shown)


assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -337,23 +337,35 @@ export const PaymentMethodDataProvider = ( { children } ) => {
...
 			// If there's no active payment method, or the active payment method has
 			// been removed (e.g. COD vs shipping methods), set one as active.
+			// Note: It's possible that the active payment method might be an
+			// express payment method. So registered express payment methods are
+			// included in the check here.
 			if (
 				! currentActivePaymentMethod ||
-				! paymentMethodKeys.includes( currentActivePaymentMethod )
+				! allPaymentMethodKeys.includes( currentActivePaymentMethod )
 			) {
```
nerrad, line 357: The offending condition was `! paymentMethodKeys.includes( currentActivePaymentMethod )`  for the following reasons:

- This effect callback is primarily for ensuring that there is an active payment method set when none is set.
- The value for `activePaymentMethod` could represent either a 'regular' payment method or an express payment method.
- The condition was only checking registered _payment methods_ for whether the key existed (not express payment methods).

So the fix was to ensure express payment methods are included in this conditional check as well.

nerrad:
While testing, I'm currently experiencing an error for the following scenario:

- Logged in user with saved payment methods.
- On visiting checkout click express payment method to initiate express payment and then immediately close the modal.
- Click to submit order. Payment fails with "Something went wrong. Please contact us to get assistance." message.

Network request has:

```json
{
  "billing_address": {
    "first_name": "Darren",
    "last_name": "Ethier",
    "company": "",
    "address_1": "123 Anytown St.",
    "address_2": "",
    "city": "Anytown",
    "state": "ON",
    "postcode": "n2t2y4",
    "country": "CA",
    "email": "darren@roughsmootheng.in",
    "phone": "5188808893"
  },
  "shipping_address": {
    "first_name": "Darren",
    "last_name": "Ethier",
    "company": "",
    "address_1": "123 Anytown St.",
    "address_2": "",
    "city": "Anytown",
    "state": "ON",
    "postcode": "n2t2y4",
    "country": "CA"
  },
  "customer_note": "",
  "payment_method": "stripe",
  "payment_data": [{ "key": "wc-stripe-new-payment-method", "value": false }]
}
```

assets/js/base/context/cart-checkout/payment-methods/test/payment-method-data-context.js
```diff
@@ -0,0 +1,138 @@
...
+				);
+			};
+			return new Config( {
+				name,
+				content: <Content />,
+				edit: <div>An express payment method</div>,
+				canMakePayment: () => true,
+				paymentMethodId: name,
+			} );
+		} );
```
haszari, line 63: Awesome to have a test for this, nice work!

I'm a little confused about the specifics of the interface we're testing, and the expectations on the context (i.e. the behaviour of the API). The part I'm not clear about – how does an express payment method trigger `active` status? I think that's a key part of the behaviour being tested here, I don't understand how this happens. Note I haven't worked on the express payments API at all (i.e. this is all quite new to me).

I think this is what the test is verifying:

1. By default, the first non-express payment method is set as active.
2. While the mock express payment modal is up, the active payment method is set (to the express payment method!).
3. When the mock express payment method is closed, the active payment method reverts to a non-express method (specifically, the first available).

(2) above is something I'm keen to better understand. It sounds like express payment methods are expected to show a modal, and while this is happening, the checkout will recognise that the express payment method is active. Can we document this in the [express payment API readme](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/docs/extensibility/payment-method-integration.md#express-payment-methods---registerexpresspaymentmethod-paymentmethodcreator-)? Again, I don't have experience of express payment methods so I'm keen to get a better understanding of the broad strokes of the express payment flow/UX, and how that's implemented in blocks, and how express payment methods are expected to hook into that.

haszari (commented):
Good spotting @nerrad !

I haven't been able to fully review/test this as I don't have any express payment methods set up and working (I don't think any are supported for me, don't know how to "fake" it for testing).

I'm happy to do another test pass if that's useful, e.g. regression testing regular payment methods – just ping me for another review.

### unit test works well
The test looks really impressive and is inspiring, I can see how this is a really powerful way to demonstrate the expectations of our API / payment layer and verify that this doesn't break as we iterate. Trying to really understand this test made it clear that I've got a lot more to learn about how our express payment methods work.

I ran the test successfully, and also broke the context to confirm the test catches it, so this aspect is looking great.

- `npm run test -- assets/js/base/context/cart-checkout/payment-methods/test/payment-method-data-context.js`

Screen Shot 2020-09-21 at 10 48 24 AM

### I'm confused about what "active" payment method means
One thing that I'm confused about is the `active` payment method (in context/state). I think I'd previously assumed this was a regular payment method that the user has selected, i.e. a radio button/tab. I had never considered that this might be set to an express payment method (though express methods are outside my experience). 

I wonder if we should represent the UI state (which radio button/tab is selected) separate from the actual method currently being used for payment (i.e. a regular or express payment method). What happens with the payment tabs/radio buttons when an express payment modal is displayed - e.g. is there no radio button selected while the modal is displayed?

assets/js/base/context/cart-checkout/payment-methods/test/payment-method-data-context.js
nerrad: Would you be open to me going over this in a sync call sometime? I'll be happy to document a bit more about this too but keep in mind the API is expressly (pun not intended) focusing on the interface exposed to payment method integrations (which really don't care how checkout does things, only that the interface works as expected). So I'm not sure the express payment API readme is the best place for this additional documentation.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -337,23 +337,35 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 	// Set active (selected) payment method as needed.
 	useEffect( () => {
 		const paymentMethodKeys = Object.keys( paymentData.paymentMethods );
+		const allPaymentMethodKeys = [
+			...paymentMethodKeys,
+			...Object.keys( paymentData.expressPaymentMethods ),
+		];
```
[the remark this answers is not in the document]
nerrad: Not really. The first condition still only cares about paymentMethodKeys because we only _update_ `activePaymentMethod` _from_ paymentMethods (not express payment method) when it's empty. So the first condition bailing early when there are no registered payment methods is still valid.

assets/js/base/context/cart-checkout/payment-methods/test/payment-method-data-context.js
nerrad: > how does an express payment method trigger active status?

When shopper clicks the express payment button, that triggers the `activePaymentMethod` to be set to the express payment method.

nerrad: You can see the actual logic here:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f76c7e46ce63d88059d8ce8b05d9409b78988e5f/assets/js/base/components/payment-methods/express-payment-methods.js#L31-L41

nerrad:
Thanks for the review Rua! Regarding this:

> I'm happy to do another test pass if that's useful, e.g. regression testing regular payment methods – just ping me for another review.

Did you test some of the other scenarios that didn't require express payment methods (eg COD testing)? In slack I suggested trying to test Chrome Pay via connecting through a VPN (to hopefully get by geographical restrictions) but if that doesn't work, if you could just focus your testing on things that don't include express payment methods and I'll take responsibility for the other tests (or maybe @senadir can do verification there if he's able to).

nerrad:
> I wonder if we should represent the UI state (which radio button/tab is selected) separate from the actual method currently being used for payment (i.e. a regular or express payment method). What happens with the payment tabs/radio buttons when an express payment modal is displayed - e.g. is there no radio button selected while the modal is displayed?

You can think of it this way:

- `activePaymentMethod` is state explicitly used to indicate what payment method is used for completing the transaction when the order is submitted. 
- `activePaymentMethod` _might not_ directly correlate to selected "regular" payment methods (saved tokens or otherwise) because in the case of Express payments, those effectively happen in a layer parallel to the main checkout ui. 

When a shopper is in the express payment modal context, all the shipping selections, and billing details are contained within the express payment modal. Behind the ui, the payment method is communicating with checkout for getting shipping pricing updates etc via the payment method interface. This is also updating checkout state. When the shopper completes the Express payment method purchase, that triggers the same flow as clicking the Process order button. Effectively, the shopper doesn't go back to the checkout at all _unless they choose to cancel the express payment flow_.

assets/js/base/context/cart-checkout/payment-methods/test/payment-method-data-context.js
haszari: > Would you be open to me going over this in a sync call sometime?

Sounds like a great idea :)

Thanks for the code pointers, will have a closer look today.

haszari:
@nerrad I'll give this another test today. I'll likely focus on regression testing of regular payment methods, but will also look at using a VPN to test region-dependent express payment methods (not sure how far I can get with this, i.e. whether I'd need real/valid payment details set up).

Might be a good idea to get a review from someone with express payment fully set up.

> I think we definitely could consider potential improvements and changes to the logic here, but rather than discussing hypotheticals - if you're interested, I'd prefer if you take some time to learn what's happening and maybe come up with some proposed improvements with some rationale for how it improves things?

I'm definitely interested to better understand the express payments layer and the expectations on both sides of the extensibility API. I think we could improve our docs for devs/users in this area too :) Most of my ponderings here are focused on this goal - I explore by asking and hypothesising.

nerrad:
> I'm definitely interested to better understand the express payments layer and the expectations on both sides of the extensibility API. I think we could improve our docs for devs/users in this area too :) Most of my ponderings here are focused on this goal - I explore by asking and hypothesising.

Sounds good. To be clear though, I don't think this should block this pull getting merged. I also want to weigh carefully the risk of creating extensive documentation for contributors to this plugin too early considering we're still iterating.

assets/js/base/context/cart-checkout/payment-methods/test/payment-method-data-context.js
haszari: There are some docs buried in the typedefs! So we have docs around the place, just I'm not familiar with it all. We could add some of this to the markdown docs too. 

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/assets/js/type-defs/registered-payment-method-props.js#L210

The typedef docs are useful here because they explains that express payment components should implement `onClick` and `onClose`, and expectations related to that.

haszari: (Note I agree this is all non blocking for this PR - but at the same time I want to understand it better to review effectively.)

haszari (approved):
I did a bunch of regression testing of regular payment methods with local dev build. 

I tested in Chrome (admin user, logged in, saved cards) and Safari (anon/guest shopper). Tested various scenarios with COD enable/disable based on shipping, using default payment method/saved card, selecting different payment methods. Didn't see any issues 👍 

I'm approving the PR - though @nerrad you may want to get another review/test with express payments setup & working. 

Note - I didn't attempt testing express via VPN, not sure if this is practical, e.g. if there are other hurdles/blockers other than geo, e.g. support. Still keen to investigate this but didn't want to get bogged down experimenting.

nerrad:
Thanks for the testing and reviews Rua! I'm going to go ahead and merge this, I've tested the express payments fairly thoroughly myself.

PR #3167, opened 2020-09-20
Re-enable WooCommerce dependency group rule for linting and fix errors.

nerrad:
Part of #3118 

In this pull the `@woocommerce/dependency-group` eslint rule is re-enabled and all errors are fixed.

There is no impact to user facing code.


senadir (approved):
Our build process and tests should catch any missing changes, so this is good to go!

PR #3168, opened 2020-09-20
Enable and fix all jsdoc rule violations

nerrad:
As a part of #3118, this pull enables all the jsdoc eslint rules and fixes the violations. Note I did each rule fix in it's own commit if that's easier for the reviewer.

This pull request does not impact any user facing features (only affects jsdocs).

Over time, there's more we can do to improve our jsdocs (custom typedefs etc) but for now I focused on the bare minimum to get the rule enabled so going forward we are following the WP standard.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -38,7 +38,6 @@ module.exports = {
 		// - @worpdress/no-unused-vars-before-return
 		// - testing-library/no-await-sync-query
 		// - @woocommerce/dependency-group
-		'jsdoc/require-param': 'off',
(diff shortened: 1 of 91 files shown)


haszari (approved):
Approving this based on a quick eyeball of the diff. (I haven't tested anything.)

Great work getting this tidied up @nerrad . We'll have to keep an eye out for these in reviews and keep iterating - e.g. parameter descriptions etc. 

Fantastic to have a baseline of all violations fixed 🙌 – and eslint will of course keep us on our toes from now on.

PR #3190, opened 2020-09-24
link to myaccount page from checkout so signup page has store theme

### LINKED ISSUE #2918: Checkout block links to wp-admin login page, not themed like store front end
**Describe the bug**
If guest checkout option is disabled, the Checkout block has a link prompting the shopper to log in. 

This links to a wp-admin login form, which is not themed, and has no store branding.

Originally reported by `Silver Surfer` in a comment on Woo Blocks release post [here](https://woocommerce.wordpress.com/2020/07/22/woocommerce-blocks-3-0-release-notes/#comment-7151).

**To Reproduce**

1. Go to `WooCommerce > Settings > Accounts`.
2. Require login for checkout by disabling these options: 
   - `Allow customers to place orders without an account` => no
   - `Allow customers to create an account during checkout` => no
3. Save settings. Ensure you have checkout block published on a page.
4. On front end, as anon user (incognito or other browser), add stuff to cart and view checkout block.
5. See login prompt and click it.

**Expected behavior**
Store themed login / signup page.

**Screenshots**

haszari:
Fixes #2918

This PR changes the link used for checkout login prompt so it links to `myaccount` page. Previously this linked to an unbranded wp-admin login page.

##### Known issue - successful login doesn't redirect back to checkout
When using `my-account`, the user is not redirected back to the checkout after login. The wp-admin page handles this correctly. So that's a serious issue, as we don't want to lose potential customers to the login flow.

cc @nerrad do you think we should address that in this PR or handle as follow up?

I'm thinking we may need to implement that behaviour in Woo core, e.g. add support for the [`redirect_to` url param](https://github.com/WordPress/WordPress/blob/aad1fa48ea68b05329eec92e1cca8b178b1641f7/wp-login.php#L586). I [searched WooCommerce code](https://github.com/woocommerce/woocommerce/search?q=redirect_to) and couldn't find anywhere that implements it (though there's support in some admin screens).

Or maybe there's a different woo endpoint that shows a login form and handles redirect – I couldn't find one. The shortcode checkout has an inline login form, so it doesn't have the same needs as the block.

#### Accessibility

1.
2.
3.

### Changelog

### DIFF
--- a/src/Assets.php
+++ b/src/Assets.php
@@ -182,7 +182,7 @@ class Assets {
 				'baseLocation'                  => wc_get_base_location(),
 				'woocommerceBlocksPhase'        => WOOCOMMERCE_BLOCKS_PHASE,
 				'hasDarkEditorStyleSupport'     => current_theme_supports( 'dark-editor-style' ),
-				'loginUrl'                      => wp_login_url(),
+				'loginUrl'                      => wc_get_page_permalink( 'myaccount' ),
 
 				/*
 				 * translators: If your word count is based on single characters (e.g. East Asian characters),


senadir:
> do you think we should address that in this PR or handle as follow up?

I know you tagged Darren, but I don't think we should handle this in a follow up, we will be creating a regression.
`my-account` has a redirect feature but it's only available on `POST` and not `GET`, we might open a PR in core to add support for `GET` as well.
C0E1AV8T0-slack-p1600357460037500
See this issue as well (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3137)

nerrad:
I agree with Nadir here, it’ll be a regression if there’s no redirect. For checkout flow this is critical and could lead to loss of conversions (which works against the goals of checkout).

If you’re unable to implement this without a pull in core, we may need to surface this for more discussion. The _ideal_ solution here is to have login happen in context (no page redirects anywhere). But that increases the scope so we’d have to make a decision on prioritizing that.

senadir:
I did some searching to see how feasible it is to "have login happen in context", it's not straightforward as I hoped:

1- Create a new login endpoint that we need to maintain and secure.
2- Change the position of `LoginPrompt` to be on the same level as `email`, and when you click on it, it shows an extra password field.
3- Submitting the login should not submit the checkout. If the logic is successful, the customer shipping and billing information should be returned, so our `login` endpoint will not just log someone, but return the cart as well.
4- We should handle any errors or validation errors inline.

So it's doable, but not straightforward.

nerrad:
Thanks for the investigation Nadir, that's along the lines of what I was thinking as well 👍

nerrad:
I'm going to close this pr because it's out of date. I'll leave the issue open.  The PR can still be used as a reference and the branch re-opened if necessary.

Issue #3141, opened 2020-09-11
Investigate why e2e tests fail in renovate PRs


Aljullu:
Travis e2e tests are failing in all Renovate PRs:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pulls/app%2Frenovate

The issue seems to be that renovate modifies the file `package-lock.json` and removes puppeteer from the top level dependencies, so Travis doesn't find it when trying to run the tests. See #3139 for more details.

What's weird is that running `rm -rf node_modules/ && rm package-lock.json && npm i` locally in those branches sets puppeteer as a top-level dependency, so for some reason the `package-lock.json` generated by renovate is not the same as the one generated locally.

Edited: as specified in the comments below, :point_down: rebasing the branches doesn't fix the issue.

haszari:
Note that Renovate has some options for rebasing PRs. 

- Can tick a box on specific PR to request an automated rebase.

<img width="713" alt="Screen Shot 2020-09-14 at 8 36 02 AM" src="https://user-images.githubusercontent.com/4167300/93031633-912b4400-f680-11ea-98db-41fa8b69fb26.png">

- Can [configure renovate to rebase PRs whenever target branch (e.g. `main`) changes](https://docs.renovatebot.com/presets-default/#rebasestaleprs).

Note - I don't think either will completely fix the "different package-lock.json on renovate" issue, just mentioning as it might be part of a solution.

As an experiment I kicked off a rebase for #3022 👀

haszari:
Linking in other issues related to improving our Renovate setup: #2811 #2806

Aljullu:
> Note - I don't think either will completely fix the "different package-lock.json on renovate" issue, just mentioning as it might be part of a solution.

Right! I did try rebasing them with no luck so, while I agree that rebasing them is good, there must be something else. :confused:

haszari:
I see there's a separate renovate PR for bumping the lock file: #3078 🤷

Aljullu:
> I see there's a separate renovate PR for bumping the lock file: #3078 shrug

Currently, it doesn't seem to make any changes to `puppeteer`, but it might be interesting to see how that PR is updated when rebasing it with a failing `main` branch.

nerrad:
We need to figure this out in order to unblock renovate updates, so I'm pushing it to a higher priority in the Current sprint pipeline. If no one else grabs it this week, I'll try to jump on it.

nerrad:
[...]

PR #3192, opened 2020-09-24
Update @wordpress/scripts to latest version

nerrad:
This updates the `@wordpress/scripts` package to the latest version.

I started this as a part of working on #3141 thinking it might help with the dependency issue for puppeteer but I don't think it makes much difference. The update is still useful in keeping the package up to date.

Since this mostly just impacts the scripts we use for running e2e and jest tests, the travis build should be sufficient coverage for whether this gets merged or not.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,12 +73,12 @@
 			},
 			"dependencies": {
 				"debug": {
-					"version": "4.1.1",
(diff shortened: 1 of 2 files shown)


nerrad:
I'm going to go ahead and merge this since travis passes.

Issue #2487, opened 2020-05-14
Explore ways so form elements in blocks don't look disabled in the editor


Aljullu:
When adding the _Cart_ or _Checkout_ blocks in the editor, all form elements look disabled, which makes them look different from the frontend. This issue is about exploring a way to keep them disabled but look like if they weren't.

See previous discussion in #2437 

### Screenshots
_Editor:_
![imatge](https://user-images.githubusercontent.com/3616980/81926959-24753f80-95e3-11ea-8cd4-6374ff3870ce.png)

_Frontend:_
![imatge](https://user-images.githubusercontent.com/3616980/81926524-5df97b00-95e2-11ea-88a0-1cefeed1c044.png)

senadir:
we stop using `<Disabled>` and create our own `<Disabled>` component, in it, we can disable remove user events using `user-events: none` this will handle mouse things, focus will still be an issue, we can fix that by using `inert` an HTML property that removes a block from the page flow without removing it, like disabled but without adding disabled, it's not supported, for now, there're some polyfills, this is one https://github.com/WICG/inert/
I suggest we just copy the code and it to our custom `<Disabled>` component.

ralucaStan:
Another issue I found was that in other blocks is that this disabled style is not normalized.
All the inputs in the image below are disabled, but they have different styling -> grey or white background.
<img width="633" alt="Screenshot 2021-02-18 at 13 49 11" src="https://user-images.githubusercontent.com/1628454/108370933-82cbf000-71fd-11eb-9d6d-4b56dfd40264.png">

Another thing I found is that the disabled product buttons aren't consistent as well in the editor
<img width="731" alt="Screenshot 2021-02-18 at 13 59 28" src="https://user-images.githubusercontent.com/1628454/108371009-98411a00-71fd-11eb-903f-6b58cb9ae668.png">

PR #3219, opened 2020-09-30
Load WC Blocks CSS after editor CSS

### LINKED ISSUE #3068: Checkout Country input is not wysiwyg styling in editor
## Describe the bug
The `Country` input control looks a bit different in the editor. 

Screen Shot 2020-08-28 at 9 11 03 AM

- Square corners.
- Semi-transparent background.
- Height bug (too short). 

Compare with front end:

Screen Shot 2020-08-28 at 9 31 26 AM

I'm assuming this is a regression that snuck in.

## To reproduce

1. Edit a page with checkout block. 
2. Scroll down to country input.

## Expected behavior
Consistent appearance in editor and front end.

Aljullu:
Fixes #3068.

The issue was caused because editor CSS was leaking into the block. It could be fixed ensuring our CSS is loaded after `wp-edit-post-css`. Notice this fix might no longer be needed if https://github.com/WordPress/gutenberg/issues/20797 gets merged into Gutenberg.

### Screenshots

_Before:_
imatge

_After:_
imatge

1. Edit a page with checkout block.
2. Scroll down to country input.
3. Verify it has rounded corners and the correct height.

Notice the semi-transparent background is not fixed since it's a different issue (#2487).

### Changelog

> Fix wrong styling of Checkout's country input in the editor.

### DIFF
--- a/src/Assets.php
+++ b/src/Assets.php
@@ -35,7 +35,17 @@ class Assets {
 	 */
 	public static function register_assets() {
 		$asset_api = Package::container()->get( AssetApi::class );
-		self::register_style( 'wc-block-vendors-style', plugins_url( $asset_api->get_block_asset_build_path( 'vendors-style', 'css' ), __DIR__ ), [] );
+
+		// In the editor, we want our stylesheets to load after editor CSS (see #3068).
+		// This might be no longer necessary when https://github.com/WordPress/gutenberg/issues/20797
+		// is merged.
+		if ( is_admin() ) {
+			$block_style_dependencies = array( 'wp-edit-post' );
+		} else {
+			$block_style_dependencies = array();
+		}
+
+		self::register_style( 'wc-block-vendors-style', plugins_url( $asset_api->get_block_asset_build_path( 'vendors-style', 'css' ), __DIR__ ), $block_style_dependencies );
 		self::register_style( 'wc-block-editor', plugins_url( $asset_api->get_block_asset_build_path( 'editor', 'css' ), __DIR__ ), array( 'wp-edit-blocks' ) );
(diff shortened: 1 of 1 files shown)


src/Assets.php
```diff
@@ -35,7 +35,17 @@ public static function init() {
 	 */
 	public static function register_assets() {
 		$asset_api = Package::container()->get( AssetApi::class );
-		self::register_style( 'wc-block-vendors-style', plugins_url( $asset_api->get_block_asset_build_path( 'vendors-style', 'css' ), __DIR__ ), [] );
+
+		// In the editor, we want our stylesheets to load after editor CSS (see #3068).
+		// This might be no longer necessary when https://github.com/WordPress/gutenberg/issues/20797
+		// is merged.
```
nerrad, line 41: Should we create an issue for the backlog for removing this extra code when the fix in Gutenberg lands and we can safely remove because the minimum version of WordPress we support has the fix (if so, you could convert this comment to a todo)?

nerrad (approved):
Nice solution here Albert 👍 (and great sleuthing).

I confirmed the select is fixed as described. I also reviewed our other blocks and ensured interfaces weren't affected by this change.

Let's 🚢 🎉

chvillanuevap:
Could this also be causing the Full Site Editing page in Gutenberg to be completely blank as reported in WordPress/gutenberg#27938? I'm using WooCommerce 4.8.0. I don't have the WooCommerce Blocks plugin installed.

Aljullu:
Hi @chvillanuevap, thanks for raising this! I created a new issue so we can investigate it: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3626.

Issue #3166, opened 2020-09-19
Fix starting express payment method and then switching to pay by saved payment method.


nerrad:
This bug surfaces when a shopper that is logged in and has a saved payment method does the following steps when checking out using the checkout block:

1. Click Express Payment method (chrome pay or apple pay).
2. Abandon paying by the express payment method (cancelling or closing the modal).
3. Leaving the saved payment method option selected without choosing any other payment method.
4. Submitting the order.

The order processing fails with the notice: "Something went wrong. Please contact us to get assistance."

Network request for the checkout processing has the following as the payload:

```json
{
  "billing_address": {
    "first_name": "Darren",
    "last_name": "Ethier",
    "company": "",
    "address_1": "123 Anytown St.",
    "address_2": "",
    "city": "Anytown",
    "state": "ON",
    "postcode": "N465T2",
    "country": "CA",
    "email": "dudedandy@gmail.com",
    "phone": "1231231234"
  },
  "shipping_address": {
    "first_name": "Darren",
    "last_name": "Ethier",
    "company": "",
    "address_1": "123 Anytown St.",
    "address_2": "",
    "city": "Anytown",
    "state": "ON",
    "postcode": "N465T2",
    "country": "CA"
  },
  "customer_note": "",
  "payment_method": "stripe",
  "payment_data": [{ "key": "wc-stripe-new-payment-method", "value": false }]
}
```

Saved payment methods are expected to have something like this as the network request:

```json
{
  "billing_address": {
    "first_name": "Darren",
    "last_name": "Ethier",
    "company": "",
    "address_1": "123 Anytown St.",
    "address_2": "",
    "city": "Anytown",
    "state": "ON",
    "postcode": "N465T2",
    "country": "CA",
    "email": "dudedandy@gmail.com",
    "phone": "1231231234"
  },
  "shipping_address": {
    "first_name": "Darren",
    "last_name": "Ethier",
    "company": "",
    "address_1": "123 Anytown St.",
    "address_2": "",
    "city": "Anytown",
    "state": "ON",
    "postcode": "N465T2",
    "country": "CA"
  },
  "customer_note": "",
  "payment_method": "stripe",
  "payment_data": [
    {
      "key": "payment_method",
      "value": "stripe"
    },
    {
      "key": "wc-stripe-payment-token",
      "value": 1
    },
    {
      "key": "isSavedToken",
      "value": true
    },
    {
      "key": "wc-stripe-new-payment-method",
[...]

PR #3226, opened 2020-10-01
Refactor payment methods components

Aljullu:
This PR includes several refactors I did as part of fixing #3166 (I will open follow-up PRs with actual fixes tomorrow). I wanted to extract refactors to a separate PR so they are easier to review and don't get blocked by other changes.

Basically, the rationale of the changes in this PR is:
* Make components smaller so they are easier to work with.
* Remove some `useRef()` usages that didn't seem to be necessary anymore.
* Simplify some of the logic along the way.

This PR doesn't include any behavior change, so everything should keep working as before.
1. Go to the Checkout block and interact with the payment method tabs. Verify you can toggle them and tab contents always appear.
2. Finish a purchase with one of the tabs and verify the order was done with the correct payment method.

It would be good to test that there are no performance regressions with `trunk`:
* With the devtools Network tab, verify there are no more calls to Stripe API than there were in `trunk`.
* Feel free to add some `console.log()` in some components to verify there are not excessive re-renders compared to `trunk`.

### DIFF
--- a/assets/js/base/components/payment-methods/new-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/new-payment-method-options.js
@@ -1,0 +1,136 @@
+/**
+ * External dependencies
+ */
+import {
+	usePaymentMethods,
+	usePaymentMethodInterface,
+	useStoreNotices,
+	useEmitResponse,
+} from '@woocommerce/base-hooks';
+import { cloneElement, useRef, useCallback } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import {
+	useCheckoutContext,
+	useEditorContext,
+	usePaymentMethodDataContext,
+} from '@woocommerce/base-context';
+import CheckboxControl from '@woocommerce/base-components/checkbox-control';
+
+/**
+ * Internal dependencies
+ */
+import Tabs from '../tabs';
+import PaymentMethodErrorBoundary from './payment-method-error-boundary';
+
+/**
+ * Returns a payment method for the given context.
+ *
+ * @param {string} name The payment method slug to return.
+ * @param {Object} paymentMethods The current registered payment methods
+ * @param {boolean} isEditor Whether in the editor context (true) or not (false).
+ *
+ * @return {Object} The payment method matching the name for the given context.
+ */
+const getPaymentMethod = ( name, paymentMethods, isEditor ) => {
+	let paymentMethod = paymentMethods[ name ] || null;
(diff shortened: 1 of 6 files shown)


assets/js/base/components/payment-methods/new-payment-method-options.js
```diff
@@ -0,0 +1,82 @@
...
+	usePaymentMethodDataContext,
+} from '@woocommerce/base-context';
+
+/**
+ * Internal dependencies
+ */
+import Tabs from '../tabs';
+import NewPaymentMethodTab from './new-payment-method-tab';
+
+const NewPaymentMethodOptions = () => {
```
nerrad, line 23: Please add some jsdocs here.

assets/js/base/components/payment-methods/new-payment-method-tab.js
```diff
@@ -0,0 +1,45 @@
...
+	usePaymentMethodDataContext,
+} from '@woocommerce/base-context';
+import CheckboxControl from '@woocommerce/base-components/checkbox-control';
+
+/**
+ * Internal dependencies
+ */
+import PaymentMethodErrorBoundary from './payment-method-error-boundary';
+
+const NewPaymentMethodTab = ( { children, allowsSaving } ) => {
```
nerrad, line 17: Nice splitting this out into its own component.

Please add some jsdocs here.

nerrad (approved):
Really great work here @Aljullu 👏🏻 - this cleans up the code nicely and reduces non-needed refs. I have a couple comments about jsdocs.

Also why the `new-` and `New` prefix for the new components? I'd be fine with just dropping it?

I tested this fairly thoroughly including testing Apple Pay on my [test site](https://ephemeral-nerrad-20201002.atomicsites.blog) with this branch.

I also tested for possible regressions with bugs I've encountered before but it's all looking good.

Pre-approving, however I would like to see my comments addressed re dropping the `new` prefix and adding jsdocs.

Aljullu:

> Also why the `new-` and `New` prefix for the new components? I'd be fine with just dropping it?

The idea was to differentiate `New` from `Saved` payment methods. But dropping `new` makes sense since that's the default for all payment methods. Done in 600dd0cdae652036d5732c7164fef9ff50bfe592.

> jsdocs

Added in 54f4f919621bac2757efde36d9f8edc2b67e4989. I also added propTypes when necessary in 122083398463dce1443911a9185b26bc94ac0eea.

PR #3227, opened 2020-10-01
Don't load contents of payment method hidden tabs

Aljullu:
(Part of fixing #3166)

Currently, the contents of all payment methods tabs are loaded in the tree even if they are not in the selected tab.

imatge

This has the issue that payment methods that add required form elements break the purchase flow because required fields are loaded but invisible to the user. For Stripe, we added [a check](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js#L88) so its elements are not rendered if it's not active, but I don't think that's something payment methods need to check in their side if we can do it for them in Blocks.

### Breaking change?

I don't consider this to be a breaking change because I can't think of any legit case where an inactive payment method might need its contents to be loaded in the page before it's selected. So adding it to the changelog is enough in my opinion and no dev note is required.

This PR doesn't include any behavior change, so everything should keep working as before.
1. Go to the Checkout block and interact with the payment method tabs. Verify you can toggle them and tab contents always appear.
2. Finish a purchase with one of the tabs and verify the order was done with the correct payment method.

### Changelog

> Developers: payment method components are no longer loaded in tree if they are not selected in the UI.

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
@@ -7,7 +7,6 @@ import { useState } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import { PAYMENT_METHOD_NAME } from './constants';
 import { getStripeServerData } from '../stripe-utils';
 import { useCheckoutSubscriptions } from './use-checkout-subscriptions';
 import { InlineCard, CardElements } from './elements';
@@ -83,11 +82,11 @@ const CreditCardComponent = ( {
 
 export const StripeCreditCard = ( props ) => {
 	const { locale } = getStripeServerData().button;
-	const { activePaymentMethod, stripe } = props;
+	const { stripe } = props;
 
-	return activePaymentMethod === PAYMENT_METHOD_NAME ? (
+	return (
 		<Elements stripe={ stripe } locale={ locale }>
 			<CreditCardComponent { ...props } />
 		</Elements>
-	) : null;
+	);
 };
--- a/assets/js/base/components/tabs/index.js
+++ b/assets/js/base/components/tabs/index.js
@@ -69,7 +69,8 @@ const Tabs = ( {
 					tabId={ `${ instanceId }-${ name }` }
 					className="wc-block-components-tabs__content"
 				>
-					{ content }
+					{ tabState.selectedId === `${ instanceId }-${ name }` &&
+						content }
 				</TabPanel>
 			) ) }
 		</div>


haszari (commented):
This is a nice little enhancement to the API 👍 I like that this reduces the expectations on each payment method component, and clarifies expectations (payment component DOM will be rendered when payment method is active/selected).

The code looks solid, no issues there.

I noticed one tiny quirk (not a blocker) when testing - the container elements are still rendered, with `display: none`. This seems redundant, or not serving a useful purpose. We could consider not rendering the wrapper, or not adding `display: none` since they don't render any content.

Screen Shot 2020-10-02 at 10 41 53 AM

I tested payments with a couple of payment gateways and things worked well. 

However I found one potential blocker – if the shopper enters data into a payment form (e.g. CC number) then switches tab (possibly by accident) their input is lost. Previously I assume we were getting this feature for free by keeping the DOM elements around all the time.

Screen Capture on 2020-10-02 at 10-47-50

That's a bit unfortunate; maybe we could ship like this but the safest/friendliest UI would retain that input. To do that we'd need to adjust the API expectations – i.e. require that methods manage this state themselves, and/or provide some standard way to store the input (as part of the API).

Aljullu:
> However I found one potential blocker – if the shopper enters data into a payment form (e.g. CC number) then switches tab (possibly by accident) their input is lost. Previously I assume we were getting this feature for free by keeping the DOM elements around all the time.

Good point. I originally tested Stripe in `trunk` and values weren't saved. That's because elements were hidden anyway but that logic was in the payment method side instead of Blocks.

GIF from `trunk`:
Peek 2020-10-02 11-08

Maybe that's different in other payment methods? What is the `Credit Card` tab in your screenshot? Is that PayPal? :thinking:

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/payment-method.js
```diff
@@ -83,11 +82,11 @@ const CreditCardComponent = ( {
 
 export const StripeCreditCard = ( props ) => {
 	const { locale } = getStripeServerData().button;
-	const { activePaymentMethod, stripe } = props;
+	const { stripe } = props;
 
-	return activePaymentMethod === PAYMENT_METHOD_NAME ? (
```
nerrad, line 88: Nice being able to remove this!

nerrad (approved):
This looks okay and seems to test fine, but while I was I noticed a bug which can be reproduced on `trunk` as well (so I created #3234)

Good to :shipit:! Thanks Albert.

haszari:
> > However I found one potential blocker – if the shopper enters data into a payment form (e.g. CC number) then switches tab (possibly by accident) their input is lost. Previously I assume we were getting this feature for free by keeping the DOM elements around all the time.

> Good point. I originally tested Stripe in trunk and values weren't saved. That's because elements were hidden anyway but that logic was in the payment method side instead of Blocks.

> Maybe that's different in other payment methods? What is the Credit Card tab in your screenshot? Is that PayPal? 🤔

It's [WC Pay](https://github.com/Automattic/woocommerce-payments). I just tested on trunk and I see that it's probably the only affected payment method (right now), so maybe this isn't a big deal? Perhaps we can add a follow up issue for it.

I'm curious to hear @nerrad's thoughts on whether the user input should be preserved when switching between tabs, i.e. priority / severity of that issue. Also whether this is something we should consider as part of the payment method API contract – i.e. is it fully up to payment methods, OR do we provide support/help for them in some way (e.g. keep the DOM elements around, or provide a place to store user input, e.g. reducer store).

Happy for this PR to merge and we can look at the user input issue as a follow up at appropriate priority.

nerrad:
> I'm curious to hear @nerrad's thoughts on whether the user input should be preserved when switching between tabs, i.e. priority / severity of that issue. Also whether this is something we should consider as part of the payment method API contract – i.e. is it fully up to payment methods, OR do we provide support/help for them in some way (e.g. keep the DOM elements around, or provide a place to store user input, e.g. reducer store).

The system has been designed so that payment methods are responsible for their own input management and Checkout is agnostic to how they manage their component display. I think if Checkout goes beyond that, it creates additional public api surface areas to be maintained.

With that said, from the shopper perspective, there's two ways of looking at this:

1. The inputs clearing on switching a tab could be viewed as a confidence booster (they don't save my CC details until actually needed).
2. The inputs clearing on switching a tab could be viewed as an annoyance (ugh, where did my credit card details go).

I'm also wary of the limited data we have on whether this would be an actual issue in the usual purchase flow. For example, how often would shoppers switch among different payment methods _after_ they've already picked how they are going to pay?

So from my perspective, I think while the surfaced issue is valid, I consider it low priority - and I'd like to await more feedback from either payment method extension developers or users of the block before we address anything here.

Aljullu:
Given the discussion above, I will proceed merging this PR as-is. We can create a follow-up issue if we want the user input to persist when switching tabs.

PR #3236, opened 2020-10-04
Use store-themed set password form for checkout signup

### LINKED ISSUE #3237: Signup during checkout should use store themed page for setting initial account password
## Describe the bug
Follow up from #2851

When users sign up during checkout, they set their new account password via a set-password form linked from a new account email. Currently this is an unbranded wp-admin themed page, because we need to add support for a set password flow in core: https://github.com/woocommerce/woocommerce/issues/27754

Screen Shot 2020-08-11 at 3 46 15 PM

When that flow is in place, we should adopt it in blocks (if available) so the set password screen is consistent with the look & feel of the store.

haszari:
Fixes #3237

This PR is a companion with https://github.com/woocommerce/woocommerce/pull/27898 in WooCommerce core - recommend testing them together.

This PR uses `my-account/lost-password` endpoint for the set-password flow for checkout signup. The endpoint uses store theme, so that improves the experience and makes setting a password more reassuring. (The flow currently uses a wp-admin themed password set form.)

The lost password form provides the functionality we need – securely set an account password. However, it is not ideal from a UX point of view as the page title (`Lost password`) doesn't match the user scenario - the user is setting their password for the first time, they have not lost anything!

To improve that part of the experience, there are changes in Woo Core to allow customizing the endpoint page title based on the `action` url param (https://github.com/woocommerce/woocommerce/pull/27898). Once this is available (aiming for Woo 4.7), we'll use `action=newaccount` to trigger the new page title (`Set password`).

### Screenshots

Existing behaviour (`master`, dev build of blocks plugin):

Screen Shot 2020-08-05 at 4 00 54 PM

Using `lost-password` endpoint, no core changes (this PR, any supported Woo version):

Screen Shot 2020-10-08 at 10 01 53 AM

Using `lost-password` endpoint, new title in Woo core (this PR + https://github.com/woocommerce/woocommerce/pull/27898):

Screen Shot 2020-10-08 at 10 01 22 AM

Follow the detailed instructions in https://github.com/woocommerce/woocommerce/pull/27898.

Also repeat the test with different versions of Woo core - e.g. oldest supported version - to ensure everything works as expected.

### Changelog

> Enhancement: Use a store-themed set password page when creating an account from Checkout block. #3236

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -27,7 +27,7 @@ import {
 	CHECKOUT_ALLOWS_GUEST,
 	isExperimentalBuild,
 } from '@woocommerce/block-settings';
-import { compareVersions, getSetting } from '@woocommerce/settings';
+import { compareWithWooVersion, getSetting } from '@woocommerce/settings';
 
 /**
  * Internal dependencies
@@ -91,7 +91,7 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 	const allowCreateAccount =
 		attributes.allowCreateAccount &&
 		isExperimentalBuild() &&
-		compareVersions( getSetting( 'wcVersion' ), '4.7', '>=' );
+		compareWithWooVersion( '4.7.0', '<=' );
 
 	useEffect( () => {
 		if ( hasErrorsToDisplay ) {
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -22,9 +22,12 @@ import {
 	SidebarLayout,
 	Main,
 } from '@woocommerce/base-components/sidebar-layout';
-import { getSetting } from '@woocommerce/settings';
 import withScrollToTop from '@woocommerce/base-hocs/with-scroll-to-top';
-import { CHECKOUT_ALLOWS_GUEST } from '@woocommerce/block-settings';
+import {
+	CHECKOUT_ALLOWS_GUEST,
+	isExperimentalBuild,
+} from '@woocommerce/block-settings';
+import { compareVersions, getSetting } from '@woocommerce/settings';
 
 /**
  * Internal dependencies
@@ -81,6 +84,15 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 		checkoutHasError &&
 		( hasValidationErrors || hasNoticesOfType( 'default' ) );
 
+	// Checkout signup is feature gated to WooCommerce 4.7 and newer;
+	// uses updated my-account/lost-password screen from 4.7+ for
+	// setting initial password.
+	// Also currently gated to dev builds only.
+	const allowCreateAccount =
+		attributes.allowCreateAccount &&
+		isExperimentalBuild() &&
+		compareVersions( getSetting( 'wcVersion' ), '4.7', '>=' );
+
 	useEffect( () => {
 		if ( hasErrorsToDisplay ) {
 			showAllValidationErrors();
@@ -96,7 +108,7 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 		! isEditor &&
 		! customerId &&
 		! CHECKOUT_ALLOWS_GUEST &&
-		! attributes.allowCreateAccount
+		! allowCreateAccount
 	) {
 		return (
 			<>
@@ -128,7 +140,7 @@ const Checkout = ( { attributes, scrollToTop } ) => {
 						showPhoneField={ attributes.showPhoneField }
 						requireCompanyField={ attributes.requireCompanyField }
 						requirePhoneField={ attributes.requirePhoneField }
-						allowCreateAccount={ attributes.allowCreateAccount }
+						allowCreateAccount={ allowCreateAccount }
 					/>
 					<div className="wc-block-checkout__actions">
 						{ attributes.showReturnToCart && (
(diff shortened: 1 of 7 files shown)


Aljullu (commented):
Really nice to see how this is taking shape. I tested with WC 4.6 and with branch `add/allow-customise-reset-password-page-title` and everything worked great, but there was one thing that I couldn't get to work as I expected.

When I submitted an order selecting the `Create an account?` checkbox, I was automatically logged in. That meant, that when accessing the `/my-account/lost-password/` link from the email, I was redirected to the My Account page (even though the URL and title was the one of the `Set password` page). Any ideas of what might be going on?

imatge

haszari:
> When I submitted an order selecting the Create an account? checkbox, I was automatically logged in. That meant, that when accessing the /my-account/lost-password/ link from the email, I was redirected to the My Account page (even though the URL and title was the one of the Set password page). Any ideas of what might be going on?

Dang – good spotting @Aljullu . In my testing I was carefully logging out first or testing in anon window missed this. I'll take a look, we really need the set password form to work whether user is logged in or anon.

Note the `Create an account` option [logs the customer in as part of checkout](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Domain/Services/CreateAccount.php#L125), so that part of this is by design.

haszari:
> When I submitted an order selecting the Create an account? checkbox, I was automatically logged in. That meant, that when accessing the /my-account/lost-password/ link from the email, I was redirected to the My Account page (even though the URL and title was the one of the Set password page). Any ideas of what might be going on?

I've now fixed this - the change is in the Woo core side, and will need review. @Aljullu  can you also review/test the core PR and leave feedback there (as Rubik representative). Feel free to ping someone else for another review, keen for some thorough reviews on this.

The existing `lost-password` render logic is in an `if ! logged in` conditional. In our use case, the customer could be logged in our anon – it depends when and where they click the set password link in the email. So I've extended the if condition to show `lost-password` form if applicable (based on endpoint/query param). This is the change: 2099ae8fc24080a0b235641ceb48216e52d5bb82

Aljullu (approved):
Looking good now! My only thought is that the issue I reported above will still be present in WC<=4.6, correct? Do you think we should disable the new account feature for old WC installs?

haszari:
> Looking good now! My only thought is that the issue I reported above will still be present in WC<=4.6, correct? Do you think we should disable the new account feature for old WC installs?

Good point! Feature gating to 4.7+ would be one option. Another option would be to use the wp-admin set password screen for older versions of Woo. 

@nerrad any thoughts on which approach would be best here? I'm leaning towards option 2 - Checkout signup in blocks is only supported for WooCommerce 4.7 or newer. That seems cleanest, is a nice simple story, and encouragement to keep WooCommerce updated to the latest.

haszari:
> > Looking good now! My only thought is that the issue I reported above will still be present in WC<=4.6, correct? Do you think we should disable the new account feature for old WC installs?

> Good point! Feature gating to 4.7+ would be one option. Another option would be to use the wp-admin set password screen for older versions of Woo.

@Aljullu I've updated the PR so the full feature (`Allow customer to create an account in checkout block`) is only enabled for WooCommerce 4.7+ (see recent commits). This means we have a consistent, store-themed signup process from checkout block in 4.7+, and avoids a buggy (set password doesn't work when logged in) or inconsistent (use of wp-admin theme) UX <4.7.

I propose that we remove the feature flag when this PR is merged, and the corresponding changes in WooCommerce core are merged and slated for release (in 4.7 🤞). Then checkout signup will be available in the blocks feature plugin alongside checkout block (i.e. remove dev-only status).

@nerrad Are you on board with this approach? Let me know with a review or a comment on this PR – if you have feedback let me know and I'll adapt the plan.

Here's next steps (assuming no issues with plan above):

- [Get another Proton review of core change](https://github.com/woocommerce/woocommerce/pull/27898) – this has changed since original @vedanshujain review. 
- Confirm / finalise removal of dev-only feature flag.
- Merge this PR.
- Merge [core PR](https://github.com/woocommerce/woocommerce/pull/27898) and finalise what Woo version will include it (aiming for 4.7).
- New PR in blocks to remove dev-only feature flag (and tweak Woo version flag if needed).

Aljullu (approved):
Approving again the new changes. The plan outlined looks good to me, let's see what @nerrad thinks about it.

> Merge this PR.
> Merge core PR and finalise what Woo version will include it (aiming for 4.7).

Maybe I would change the order of these steps? That would prevent having to update the WC version in Blocks' code if Core's PR is delayed to a future release (4.8 or later).

nerrad:
> Maybe I would change the order of these steps? That would prevent having to update the WC version in Blocks' code if Core's PR is delayed to a future release (4.8 or later).

I agree with this. There's real risk of having a release of blocks out with the wrong Woo Core version gating this.

haszari:
Update – as discussed, will wrap up the core PR before merging this. The plan is to release this feature (checkout signup) in the blocks feature plugin, i.e. remove the dev-only feature flag 🎉 

- [Wrap up core PR and get it merged](https://github.com/woocommerce/woocommerce/pull/27898). This is in the final stages, 🤞 merge in next day or so.
- Finalise target Woo version (aiming for 4.7) and update this PR if needed.
- Merge, release 🚀

haszari:
Merging!

- The core PR has now merged, and will ship in WooCommerce 4.7.

haszari:
I've logged a follow up issue for exposing this feature in a blocks plugin release, targeting 3.7.0.  #3286

PR #3238, opened 2020-10-05
Add __experimental_woocommerce_blocks_checkout_order_processed action

senadir:
This is the first half of two PRs that should close this [issue](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3233)

It adds `__experimental_woocommerce_blocks_checkout_order_processed ` action that mimics `woocommerce_checkout_order_processed` but with a slightly different signutare.

```php
// blocks action
do_action( "__experimental_woocommerce_blocks_checkout_order_processed", $order_id, $order_object );
// classic action
do_action( "woocommerce_checkout_order_processed", $order_id, $posted_data, $order_object );
```

We omit $posted_data because we don't have any, and simply pass an empty array so we don't break any type check.

I decided to create `woocommerce_blocks_checkout_order_processed` instead of firing `woocommerce_checkout_order_processed` directly because of the missing post data issue.

This action was created for WooCommerce subscriptions, which thankfully doesn't use any `$_POST` data.

It's hard to test this right now, you will need to check out this PR in WooCommerce Subscriptions and follow the testing instructions there.

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -182,6 +182,10 @@ class Checkout extends AbstractRoute {
 		// Persist customer address data to account.
 		$order_controller->sync_customer_data_with_order( $order_object );
 
+		// Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
+		// But we're opting for a new action because the original ones attaches POST data.
+		do_action( 'woocommerce_blocks_checkout_order_processed', $order_object->get_id(), [] );
+
 		if ( ! $order_object->needs_payment() ) {
 			$payment_result = $this->process_without_payment( $order_object, $request );
 		} else {
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -184,7 +184,7 @@ class Checkout extends AbstractRoute {
 
 		// Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
 		// But we're opting for a new action because the original ones attaches POST data.
-		do_action( 'woocommerce_blocks_checkout_order_processed', $order_object->get_id(), [] );
+		do_action( 'woocommerce_blocks_checkout_order_processed', $order_object->get_id(), [], $order_object );
 
 		if ( ! $order_object->needs_payment() ) {
 			$payment_result = $this->process_without_payment( $order_object, $request );
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -188,7 +188,7 @@ class Checkout extends AbstractRoute {
 		* NOTE: this hook is still experimental, and might change or get removed.
 		* @todo: Document and stabilize __experimental_woocommerce_blocks_checkout_order_processed
 		*/
-		do_action( '__experimental_woocommerce_blocks_checkout_order_processed', $order_object->get_id(), $order_object );
+		do_action( '__experimental_woocommerce_blocks_checkout_order_processed', $order_object );
 
 		if ( ! $order_object->needs_payment() ) {
 			$payment_result = $this->process_without_payment( $order_object, $request );
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -182,9 +182,13 @@ class Checkout extends AbstractRoute {
 		// Persist customer address data to account.
 		$order_controller->sync_customer_data_with_order( $order_object );
 
-		// Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
-		// But we're opting for a new action because the original ones attaches POST data.
-		do_action( 'woocommerce_blocks_checkout_order_processed', $order_object->get_id(), $order_object );
+		/*
+		* Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
+		* But we're opting for a new action because the original ones attaches POST data.
+		* NOTE: this hook is still experimental, and might change or get removed.
+		* @todo: Document and stabilize __experimental_woocommerce_blocks_checkout_order_processed
+		*/
+		do_action( '__experimental_woocommerce_blocks_checkout_order_processed', $order_object->get_id(), $order_object );
 
 		if ( ! $order_object->needs_payment() ) {
 			$payment_result = $this->process_without_payment( $order_object, $request );
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -184,7 +184,7 @@ class Checkout extends AbstractRoute {
 
 		// Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
 		// But we're opting for a new action because the original ones attaches POST data.
-		do_action( 'woocommerce_blocks_checkout_order_processed', $order_object->get_id(), [], $order_object );
+		do_action( 'woocommerce_blocks_checkout_order_processed', $order_object->get_id(), $order_object );
 
 		if ( ! $order_object->needs_payment() ) {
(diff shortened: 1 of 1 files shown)


senadir:
For reference, this is the place that `woocommerce_checkout_order_processed`, right before processing the order.
https://github.com/woocommerce/woocommerce//blob/d48f1d4e2efcfab8206799df2085f021ec273722/includes/class-wc-checkout.php#L1156

src/StoreApi/Routes/Checkout.php
```diff
@@ -177,6 +177,10 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
 		// Persist customer address data to account.
 		$order_controller->sync_customer_data_with_order( $order_object );
 
+		// Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
+		// But we're opting for a new action because the original ones attaches POST data.
+		do_action( 'woocommerce_blocks_checkout_order_processed', $order_object->get_id(), [], $order_object );
```
haszari: If we're making a new hook, perhaps it's better to remove the empty / unused `$posted_data` argument. 

This might mean more refactoring on the Subscriptions end, but it more closely matches our intention and what will be passed to extensions implementing the hook.

If we end up needing to pass extra params/options in future we can add an `$options` as the last arg, with specific/known keys.

haszari (approved):
I tested this alongside https://github.com/woocommerce/woocommerce-subscriptions/pull/3825 and it's working pretty well!

The only feedback I have regards the `$posted_data` argument. If we're not planning on using it, and this is a brand new hook, let's not add a placeholder. Also more generally, it looks like in the old hook this would just provide all of the `$_POST` request params – personally I'd prefer we expose specific options/params as needed, rather than provide full access in case it's needed – then we can avoid issues like this in the future.

I realise that might not be practical, as we'll need more refactoring in the Subs plugin. Essentially the hard-coded empty array would be a local concern with Subscriptions code base, rather than provided by the API. If we can do it cleanly I think we should. (Not a blocker, keen for input from Rubik & subs teams.)

By the way, in the Subscriptions PR I see a lot more discussion of the new hook and related considerations - @senadir can you move/copy that info to this PR, as it's most relevant here, and should be visible to Blocks plugin users/developers.

Note – I spotted two new Subs-related integration issues while testing:

- [Hide unsupported payment methods when order contains subscription](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3252)
- [Handle anon/guest checkouts when order contains subscription](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3253)

senadir:
> By the way, in the Subscriptions PR I see a lot more discussion of the new hook and related considerations - @senadir can you move/copy that info to this PR, as it's most relevant here, and should be visible to Blocks plugin users/developers.

The discussion is not relevant to Blocks because it touches a hook provided by Subscriptions to its third party developers, and nothing from our end.
Here's what I wrote
<blockquote>
It hooks <code>process_checkout</code> to a new hook <code>woocommerce_blocks_checkout_order_processed</code> that works almost exactly like <code>woocommerce_checkout_order_processed</code>.

The different is that <code>woocommerce_blocks_checkout_order_processed</code> doesn't pass $posted_data (passes an empty array).

As far as I've searched, Subscription doesn't use $posted_data at all, and just pass it along to woocommerce_checkout_create_subscription, which is not used at all inside the codebase.

This means any third-party apps that assume $posted_data is set will face an issue with Checkout Block orders.

A list of codebases I found on github that uses that action
https://github.com/search?q=woocommerce_checkout_create_subscription&type=code

Some ways around this, but I'm keen to hear what you have:

Create a new action instead of woocommerce_checkout_create_subscription that doesn't pass POST data (I don't like this solution).
Annonce that publicly, third-party plugins and developers who wish to use Subscriptions with Checkout block will need to update the code around using POST data.
</blockquote>

senadir:
I've updated the code to remove the posted data, it makes sense and I can walk around it in Subs.

src/StoreApi/Routes/Checkout.php
```diff
@@ -177,6 +177,10 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
 		// Persist customer address data to account.
 		$order_controller->sync_customer_data_with_order( $order_object );
 
+		// Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
+		// But we're opting for a new action because the original ones attaches POST data.
+		do_action( 'woocommerce_blocks_checkout_order_processed', $order_object->get_id(), $order_object );
```
nerrad: I wonder if we should add an `__experimental` prefix on these new actions/filters for now as a signal that it _could_ change. I'm wary about merging in new actions and filters as public apis while we're still building the extension integrations. 

I realize this could complicate extensions we're building the integrations in (i.e. having to make sure that the right action slug is used when it eventually goes stable), and I don't think we'll want to keep this experimental for long. But while the extensibility is in progress I think it still is useful.

haszari:
> > By the way, in the Subscriptions PR I see a lot more discussion of the new hook and related considerations - @senadir can you move/copy that info to this PR, as it's most relevant here, and should be visible to Blocks plugin users/developers.
> The discussion is not relevant to Blocks because it touches a hook provided by Subscriptions to its third party developers, and nothing from our end.

Thanks for adding the info here too @senadir . To be clear, I'm referring to the new blocks hook – `woocommerce_blocks_checkout_order_processed`. This is a new extensibility hook introduced by WooCommerce Blocks (not Subscriptions), so should be documented here (in this repo, this PR introduces it). Subscriptions is the first client of the hook, but it's owned by blocks 😃  

We may also need dev note and update our docs - how do we document our hooks? Perhaps we can use PHPdoc (like core) or add a markdown doc in https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/docs/extensibility

src/StoreApi/Routes/Checkout.php
haszari: Good idea. We can perhaps review `__experimental` hooks (etc) each release/cycle (I think you suggested the idea the other day @nerrad 😄).

Alongside marking them experimental, I think it's a good idea to document them upfront (i.e. don't wait til they are stable) - we could use PHPdoc or add a markdown doc.

nerrad: PHP doc preferably for now (while also indicating it's experimental in the doc too), and then bring out to our `/docs/extensibility` folder once it moves from being experimental.

src/StoreApi/Routes/Checkout.php
```diff
@@ -177,6 +177,14 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
 		// Persist customer address data to account.
 		$order_controller->sync_customer_data_with_order( $order_object );
 
+		/*
+		* Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
+		* But we're opting for a new action because the original ones attaches POST data.
+		* NOTE: this hook is still experimental, and might change or get removed.
+		* @todo: Document and stabilize __experimental_woocommerce_blocks_checkout_order_processed
+		*/
+		do_action( '__experimental_woocommerce_blocks_checkout_order_processed', $order_object->get_id(), $order_object );
```
haszari: Do we need to pass the order ID and the order object? Surely the ID is redundant if we are passing the whole object.

senadir:
Throwing this here for when we initially built this function and decided to not trigger this action
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2011#discussion_r398763600

nerrad:
Just a heads up that with the addition of #3348, the implementation of the `__experimental` action in this pull should be documented in the new document.

src/StoreApi/Routes/Checkout.php
senadir: That's a good idea, I removed the ID.
Initially, I was trying to keep parity with the original hook.
Using the object only would require some refactoring in subscriptions to create a wrapper function.

senadir: One other risk is passing outdated data or data that is yet to be saved, but we should have saved everything before this point.

PR #3254, opened 2020-10-08
Allow switching to/from blocks <=> shortcodes for cart and checkout from admin settings

### LINKED ISSUE #3187: Implement setting for switching between Cart and Checkout blocks and Shortcodes for checkout flow.
This still needs some design work, once that's done the following outcomes are aimed for.

- Merchant would access the setting on the WooCommerce -> Advanced settings page.
- Using the select input, they'd be able to switch between Blocks and Shortcode for checkout flow and code will take care of swapping out block or shortcode in page content on switching.
- There'd be a link to the relevant pages (possibly a prompt) for editing blocks when they are switched to.
- Add tracks events so we can get data on how many stores are using shortcodes vs. blocks.

Questions:

- What if people want to just implement one or the other block and not both (i.e. Cart block but not checkout block). I've seen this behaviour from folks who _can_ use the Cart block but not the Checkout block yet because Checkout block doesn't support an extension they use.
- If the above happens, and people manually change the pages - should that be reflected in the toggle in the settings somehow?

haszari:
Fixes #3187

In progress!

Screen Shot 2020-10-08 at 4 20 51 PM

#### Accessibility

1.
2.
3.

### Changelog


haszari:
Here's what's left to do to get this feature shipshape:

- [x] Fix linter warnings.
- [x] Get the detection logic working correctly with all forms of the block (e.g. customised props) and shortcodes.
- [x] Add some automated tests for `sniff_page_format` and other functionality in the service class.
- [x] Get the switcher working well for shortcode => block (partially implemented).
- [x] Get the switcher working well for block => shortcode (not yet implemented).
- [x] Include a good default empty cart state (inner blocks) when switching to block-based cart.
- [ ] Correctly handle when user selects a different cart/checkout page AND changes format in the same request. 
- [ ] Show a warning when switching format will lose content, e.g. when switching from format `custom` to block or shortcode only.
- [ ] Any tweaks to warning message (or other UX) based on store revisions options. I.e. if revisions are disabled, extra warnings because there's no easy way to roll back (or.. if revisions are enabled, less need for a scary warning).
- [ ] Record tracks events when switching format so we can monitor of usage of the feature.

Note - there are various `TODO`s in the code. Most of these are not intended as follow ups and should be fixed in this PR.

haszari:
> Correctly handle when user selects a different cart/checkout page AND changes format in the same request. 

Just discovered a potential issue / UX problem with current approach. 

This settings page allows users two change two things:

- Which pages are selected for cart/checkout.
- The content of the current cart/checkout pages.

If the user selects different page(s) and clicks `Save`, then any request to change page format refers to a different (previous) page. So we should probably ignore these, and may want to tweak the UI so the user doesn't get confused.

cc @LevinMedia 

Here's a screenshot to illustrate:

Screen Shot 2020-10-20 at 2 53 59 PM

- If user selects different checkout pages in (1),
  - we should not make any changes to the page content (2).

src/Domain/Services/CartCheckoutPageFormat.php
```diff
@@ -0,0 +1,365 @@
...
+
+	/**
+	 * Replace the contents of specified page with a cart block.
+	 * Work in progress!
+	 *
+	 * @param Object $cart_page WP page object for cart page, e.g. return value of `get_post()`.
+	 */
+	private static function set_cart_page_format_block( $cart_page ) {
+		// TODO this is hard-coded sample content. Should we store somewhere else?
+		// Should we add a hook to allow customise?
```
Aljullu: > Should we add a hook to allow customise?

I don't think so. The Cart block markdown can't be modified without causing the block invalidation, so we don't want extensions to hook in here? But agree it would be great to store it as a constant on top of the file or somewhere else.

src/Domain/Services/CartCheckoutPageFormat.php
```diff
@@ -0,0 +1,365 @@
...
+		}
+
+		$info['cart_page_id']     = $cart_page_id;
+		$info['checkout_page_id'] = $checkout_page_id;
+
+		$cart_page     = get_post( $cart_page_id );
+		$checkout_page = get_post( $checkout_page_id );
+
+		if ( ! $cart_page || ! $checkout_page ) {
+			return $info;
```
Aljullu: Notice this early return makes it so that if one of the pages is not defined, the format of the other would not be loaded correctly.
imatge

Would refactoring this to something like the code below work?
```PHP
if ( $cart_page ) {
  // do all Cart page logic here
}
if ( $checkout_page ) {
  // do all Checkout page logic here
}
return $info;
```

src/Domain/Services/CartCheckoutPageFormat.php
```diff
@@ -0,0 +1,365 @@
...
+			'title' => __( 'Cart and checkout format', 'woo-gutenberg-products-block' ),
+			'type'  => 'title',
+
+			'desc'  => sprintf(
+				// Translators: %s Documentation url.
+				__( 'Choose between block or shortcode formats. <a href="%s">Learn more.</a>', 'woo-gutenberg-products-block' ),
+				'https://docs.woocommerce.com/document/cart-checkout-blocks-support-status/'
+			),
+
+			__( 'Choose between block or shortcode formats. Learn more.', 'woo-gutenberg-products-block' ),
```
Aljullu: Is this line required?

src/Domain/Services/CartCheckoutPageFormat.php
```diff
@@ -0,0 +1,365 @@
...
+			'id'    => 'cart_checkout_format_options_title',
+		];
+
+		$description = 'Cart page is not set – select a page in the section above.';
+		if ( $block_status['cart_page_id'] ) {
```
Aljullu: Notice if the page is not defined, its id is `-1`, so this evaluates to true.

src/Domain/Services/CartCheckoutPageFormat.php
```diff
@@ -0,0 +1,365 @@
...
+				get_edit_post_link( $block_status['cart_page_id'] )
+			);
+		}
+		$items[]     = [
+			'title'   => __( 'Cart format', 'woo-gutenberg-products-block' ),
+			'desc'    => $description,
+			'type'    => 'select',
+			'id'      => 'cart_checkout_format_options_cart_format',
+			'value'   => $block_status['cart_page_format'],
+			'options' => $block_shortcode_options,
```
Aljullu, line 165: What do you think about making it disabled when the page is not defined?
```PHP
			'custom_attributes' => [ 'disabled' => true ],
```

Aljullu (commented):
This looks very promising! It's a really complex feature and the need to use regexp makes it even more complex, but I like how you dealt with it. I'm sure there will be many edge cases we will need to test. For now, I found a couple of weird states when one of the pages is undefined. I left some inline comments below. :point_down:

haszari:
Thanks for the review @Aljullu  - lots of good feedback in there, will add fixes when I get back to this.

Edit: I hadn't pushed all the latest changes, including a bunch of functionality (my bad). Though all your review feedback was still relevant. Now all addressed, so thanks for the assist 👍

src/Domain/Services/CartCheckoutPageFormat.php
haszari: Good idea. Moved to class constants here: 89ac2fd73df81efd84a310a200c339d32716828d

(Note I still have a couple of todo comments related to the hard-coded content to consider before merge.)

src/Domain/Services/CartCheckoutPageFormat.php
haszari: Yep - good catch - fixed here: 556ec1e6cb07e8640033a36c32976916df415ba5 😄

src/Domain/Services/CartCheckoutPageFormat.php
haszari: No .. 556ec1e6cb07e8640033a36c32976916df415ba5 😅

src/Domain/Services/CartCheckoutPageFormat.php
haszari: Good spotting, I think I meant to use `cart_page` (etc) here, as it's false if missing (or a page object if available). Fixed up here: 9782c5314d8864dec69a8393a30414d277401a61

src/Domain/Services/CartCheckoutPageFormat.php
haszari: Disabling (on page load) when the page is unset will definitely help. We still have an issue if the user changes the cart page dynamically, but that's more of an edge case. 94b41db063b44b07d23280953c006f5e3efb082c

Thanks for the tip about `custom_attributes`, I was not aware of that 😃

haszari:
I'm pausing work on this briefly – we're revisiting the design. There are potential issues having the UI to select a page, and the UI to change the content of those pages in the same screen (settings tab). 

TLDR the UI for changing pages to shortcode/block will work better if it's standalone, so we're looking for where best to locate it.

nerrad:
I'm going to close this pr for now but leave the related issue open. If/when we revisit this we can reopen the branch and use as reference for any resumed work.

PR #3256, opened 2020-10-08
Simplify emitEventWithAbort logic

Aljullu:
Small PR that refactors `emitEventWithAbort`, it should work equivalently but with fewer lines of code.

Fortunately, event emitters have good test coverage, so if the refactor produced any output difference, it will probably be caught by tests.

Steps to test it manually:
1. Place an order using the Stripe payment gateway with the card number `4000000000000101` (it will fail with an invalid CVC error).
2. Verify the order isn't processed and an error is displayed in the frontend.


haszari (commented):
Logic looks good, I tested with `4000000000000101` card number, saw the appropriate error, then also tested with `4242` card and confirmed payments still work.

haszari (approved):

PR #3260, opened 2020-10-08
Save order after creating an account

senadir:
I found this bug while investigating https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3253
The issue was that subscriptions would attempt to create a new order via `new WC_Order( $order_id )`, that order didn't have the customer id in it.

After some digging and asking around, it seems that we need to call `$order->save` after altering the order.

We can use this option (in this PR) by calling `save` after creating the account.

This means we're calling save twice: once in `update_order_from_request` and in here.

I'm not fully sure if save is an expensive method to call, it updates the database directly so it feels expensive, but I don't have any data to prove it.

If it proves to be expensive, we can attempt to create the account before syncing the data.

This means, instead of fetching the email and name from `$order`, we will fetch them from `$request`.

I prefer the second solution, but I'm not sure if it would prove to be problematic, I have a  feeling it won't be problematic because the request will always contain all of the data needed.

### How to test
- It's better to test this in combo with https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3238 and https://github.com/woocommerce/woocommerce-subscriptions/pull/3825 to see the subscription creating the account, but that's not straightforward, so we will ensure we have no regression.

- On incognito, checkout and create an account.
- Make sure the account contains the correct data.

### DIFF
--- a/src/Domain/Services/CreateAccount.php
+++ b/src/Domain/Services/CreateAccount.php
@@ -115,15 +115,13 @@ class CreateAccount {
 		}
 
 		$customer_id = $this->create_customer_account(
-			$order->get_billing_email(),
-			$order->get_billing_first_name(),
-			$order->get_billing_last_name()
+			$request['billing_address']['email'],
+			$request['billing_address']['first_name'],
+			$request['billing_address']['last_name']
 		);
-
 		// Log the customer in and associate with the order.
 		wc_set_customer_auth_cookie( $customer_id );
 		$order->set_customer_id( get_current_user_id() );
-		$order->save();
 
 		return $customer_id;
 	}
--- a/src/Domain/Services/CreateAccount.php
+++ b/src/Domain/Services/CreateAccount.php
@@ -98,18 +98,16 @@ class CreateAccount {
 	}
 
 	/**
-	 * Create a user account for specified order and request (if necessary).
+	 * Create a user account for specified request (if necessary).
 	 * If a new account is created:
-	 * - The order is associated with the account.
 	 * - The user is logged in.
 	 *
-	 * @param \WC_Order        $order   The order currently being processed.
 	 * @param \WP_REST_Request $request The current request object being handled.
 	 *
 	 * @throws Exception On error.
 	 * @return int The new user id, or 0 if no user was created.
 	 */
-	public function from_order_request( \WC_Order $order, \WP_REST_Request $request ) {
+	public function from_order_request( \WP_REST_Request $request ) {
 		if ( ! self::is_feature_enabled() || ! $this->should_create_customer_account( $request ) ) {
 			return 0;
 		}
@@ -121,7 +119,6 @@ class CreateAccount {
 		);
 		// Log the customer in and associate with the order.
 		wc_set_customer_auth_cookie( $customer_id );
-		$order->set_customer_id( get_current_user_id() );
 
 		return $customer_id;
 	}
--- a/src/Domain/Services/CreateAccount.php
+++ b/src/Domain/Services/CreateAccount.php
@@ -123,6 +123,7 @@ class CreateAccount {
 		// Log the customer in and associate with the order.
 		wc_set_customer_auth_cookie( $customer_id );
 		$order->set_customer_id( get_current_user_id() );
+		$order->save();
 
 		return $customer_id;
 	}
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -158,12 +158,6 @@ class Checkout extends AbstractRoute {
 		// Ensure order still matches cart.
 		$order_controller->update_order_from_cart( $order_object );
 
-		// If any form fields were posted, update the order.
-		$this->update_order_from_request( $order_object, $request );
-
-		// Check order is still valid.
-		$order_controller->validate_order_before_payment( $order_object );
-
 		// Create a new user account as necessary.
 		// Note - CreateAccount class includes feature gating logic (i.e. this
 		// may not create an account depending on build).
@@ -178,6 +172,11 @@ class Checkout extends AbstractRoute {
 				$this->handle_error( $error );
 			}
 		}
+		// If any form fields were posted, update the order.
+		$this->update_order_from_request( $order_object, $request );
+
+		// Check order is still valid.
+		$order_controller->validate_order_before_payment( $order_object );
 
 		// Persist customer address data to account.
 		$order_controller->sync_customer_data_with_order( $order_object );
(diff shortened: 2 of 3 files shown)


haszari (approved):
@senadir did you see an issue/bug in normal use, or does this only affect Subscriptions integration?

When I test `trunk` branch and select `Create an account` option, the order is associated with the new customer account:

Screen Shot 2020-10-09 at 10 42 00 AM

So that suggests to me that the order is getting saved somewhere (probably later in checkout API). 

Adding an extra `save()` call after assigning the user id shouldn't cause any issues, but if this is only affecting subscriptions, maybe we need a fix somewhere else - the `save` could mask other issues. If we do need to add an explicit `save()`, perhaps we should do this as late as possible in the API handler - e.g. at the end of [`get_route_post_response()`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/StoreApi/Routes/Checkout.php#L146).

haszari:
Woops - that review was meant to be a comment, not an approval! Don't know how to roll back the approval … anyway @senadir let me know your thoughts.

haszari (changes requested):
**Update**: Looking at where the [new `order_processed` hook is fired](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fbd18ad3f20ff11519d2481b5880cfe3587ab5f6/src/StoreApi/Routes/Checkout.php#L182) (in #3238) I can see a couple of possibilities/issues:

- We're passing `$order_object` to the hook, and to subscriptions, but it's not using it. So the bug could be that it's pulling a stale copy of the order from the DB instead of using the up-to-date `$order_object` arg?
- If we wanted to add an explicit `save()`, it's probably better to do that later in `get_route_post_response` after everything that touches order, e.g. after `sync_customer_data_with_order` or maybe even `process_payment`.

Based on that, I think this fix/issue is specifically related to the Subs integration work in progress (#3238), so maybe best to fix within that PR. 

Though @senadir  if you know of an issue caused by the lack of explicit `$order->save()` (on `trunk`) please let me know!

senadir:
> We're passing $order_object to the hook, and to subscriptions, but it's not using it. So the bug could be that it's pulling a stale copy of the order from the DB instead of using the up-to-date $order_object arg?
This is not related, we're only passing the object to keep parity, Subscriptions uses the id to initiate the object.

> If we need to add an explicit save(), it's probably better to do that later in get_route_post_response after everything that touches order, e.g. after sync_customer_data_with_order or maybe even process_payment.

Processing payment already saves the order, but deletes the cart in doing so. We need to fire the action after everything is done but just before the payment is processed.

> Based on that, I think this fix/issue is specifically related to the Subs integration work in progress (#3238), so maybe best to fix within that PR.

Calling Save is not for Subscriptions only, but for any other existing code that wants to use this.
Refactoring Subscriptions and all places in it that depend on this is severely out of scope, I will try this lightly and update the process_checkout function in Subscriptions and see if it works.

senadir:
I still think we should create the account before saving, currently in `trunk`, we save in `update_order_from_request`.
We save again once we process the order. The evidence is that if I return just before processing, no account would be created
image

I chatted a bit with Ron Rennick and Vedanshu Jain from Proton
According to Ron:
> If you modify an object in a function, you should always save it before exiting the function.

In our situation, we save, modify, run the action, and save again.

> save(), perhaps we should do this as late as possible in the API handler - e.g., at the end of get_route_post_response().

This is the current behavior, as we already save when processing payment, what's left in the function after that is simply preparing the response.

However, we process after running the action. 

## Conclusion
 
1- **We need to save the order after creating an account.**
2- **We don't want to save 2 times in a row redundantly**.
3- **We should create the account from the request data and save it after that.**

- Our hook won't work.
- No code hooking into it will get the account data.
- It's not related to subscriptions. We simply didn't notice it because we didn't run any code between creating the account and processing the payment.

## Workaround
- I can fix this by passing the order object directly to Subscriptions instead (tested and it worked), but that's not a solution.
If any third party code that uses the order ID assumes he will get an updated order, but the data only lives in the object at that moment and not in the DB, so it's best if we save.

senadir:
I've gone ahead and updated the PR to remove the extra save I added, replaced the order in which we save and create an account, so now:

- We create an account using the `$request` object.
- We save the order.
- We trigger the action.
- We process payment (or process without payment), in both cases, we save after that.

I tested with:

- A Subscription.
- No Subscription.
- Logged out
- Logged in.

I feel like this is the way we should continue.

senadir:
I investigated a bit more, it seems WooCommerce Core also uses posted data (akin to `$request`) to create a user

https://github.com/woocommerce/woocommerce//blob/d48f1d4e2efcfab8206799df2085f021ec273722/includes/class-wc-checkout.php#L1007-L1017

src/Domain/Services/CreateAccount.php
```diff
@@ -116,11 +116,10 @@ public function from_order_request( \WC_Order $order, \WP_REST_Request $request
 		}
 
 		$customer_id = $this->create_customer_account(
-			$order->get_billing_email(),
-			$order->get_billing_first_name(),
-			$order->get_billing_last_name()
+			$request['billing_address']['email'],
+			$request['billing_address']['first_name'],
+			$request['billing_address']['last_name']
```
haszari, line 121: The `$order` param is now only used once in `from_order_request()` to set the user (`$order->set_customer_id`). You could factor that out to `get_route_post_response()` and then we wouldn't need to pass $order to `from_order_request`, that might simplify this more.

haszari:
Thanks for digging further @senadir  - reordering the logic seems reasonable from my point of view. I based the previous code on the equivalent logic in Woo core, but as far as I'm aware reordering it shouldn't cause any issues (🤞 !!). 

I'm not clear on what issues this causes, other than with the new hook you're adding for Subscriptions. Are you testing this branch/PR in combination with other new code for Subscriptions- e.g. the new hook in #3238? Perhaps it's better to make this change in that PR directly so the need/benefit is clear. 

It's hard for me to review this change in isolation without seeing a specific issue caused by the existing code (especially when we spent so long testing and reviewing create account!!).

src/Domain/Services/CreateAccount.php
haszari: (Not a blocker, just a thought)

haszari (approved):
Sorry for all the back & forth and long-winded review comments @senadir! I was having trouble getting my head around whether this change was useful on its own, or if it's required as part of of the new hook (#3238).

I've tested the following checkout signup flows and didn't find any issues. In each case the account was set up and associated with the order (if appropriate).

- Anon user, create an account with checkout.
- Logged-in user, no account is offered no created as part of checkout.
- With guest checkout store option disabled, and create account enabled in block, the account is created automatically (and there is no checkbox shown to shopper).

Looking at the code, I don't see any obvious problem with reordering the logic, and changing the service so it's using data from `$request` instead of `$order`. I did notice that perhaps we could refactor a little more so the service isn't passed an order at all - just an idea, not a blocker.

Assuming the new hook is merged, or we plan to add other PHP hooks in a similar place in the flow, I'm happy for this to merge (though I'll need to rebase my PR #3236 !).

PR #3261, opened 2020-10-09
Make 'retry' property on errors from checkoutAfterProcessingWithSuccess/Error observers default to true if it's undefined

### LINKED ISSUE #3221: The submit button of the checkout block remains disabled after a failed payment
## Describe the bug

The bug is already described in the description of [this PR](https://github.com/Automattic/woocommerce-payments/pull/723) (the "Erroneous state after confirmation" section).

## To reproduce

1. Checkout & setup WooCommerce Payments in test mode.
2. Use the checkout block and an SCA-enabled test card (ex. `4000002500003155`)
3. In the payment authentication (SCA) modal, decline the payment.
4. Check the submit button. It will be disabled and contain a check mark.

## Expected behavior

The submit button should go back to it's original and usable state.

## Screenshots

Erroneous state

cc @nerrad as you confirmed the issue [here](https://github.com/Automattic/woocommerce-payments/pull/723#pullrequestreview-434849473):

> "Even though the confirmation failed, the submit button displays a seemingly successful check-mark." Ya this looks buggy, and I'll need to investigate.

Aljullu:
Fixes #3221.

Errors in WC Payments were not recoverable because the `retry` property was not specified in the error response, so it was considered `false`. This behavior was a bit inconsistent in our codebase, because `retry` [needed to be truthy in `CHECKOUT_AFTER_PROCESSING_WITH_SUCCESS`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/cart-checkout/checkout-state/index.js#L281), but [was allowed to be undefined in `CHECKOUT_AFTER_PROCESSING_WITH_ERROR`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/cart-checkout/checkout-state/index.js#L240-L243).

This PR updates the Checkout Provider so if retry is undefined, it treats it as if it was `true`.

* Install WC Payments and set it up.
* Submit an order using WC Payments with the card number `4000 0027 6000 3184`, but in the validation popup, press on `Fail authentication` or `Cancel`.
* Verify the `Place Order` button doesn't stay disabled and you can resubmit the order.

### Changelog

> Dev: errors produced on `checkoutAfterProcessingWithSuccess/Error` observers no longer need `retry: true` to allow the user to retry the payment. You can set `retry: false` if you want to force disabling submitting the checkout form again. In addition, a new `shouldRetry` function is available from the `useEmitResponse` hook which enables setting more complex retry rules.

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -101,6 +101,7 @@ export const CheckoutStateProvider = ( {
 		isSuccessResponse,
 		isErrorResponse,
 		isFailResponse,
+		shouldRetry,
 	} = useEmitResponse();
 
 	// set observers on ref so it's always current.
@@ -237,10 +238,7 @@ export const CheckoutStateProvider = ( {
 							addErrorNotice( response.message, errorOptions );
 						}
 						// irrecoverable error so set complete
-						if (
-							typeof response.retry !== 'undefined' &&
-							response.retry !== true
-						) {
+						if ( ! shouldRetry( response ) ) {
 							dispatch( actions.setComplete( response ) );
 						} else {
 							dispatch( actions.setIdle() );
@@ -278,10 +276,7 @@ export const CheckoutStateProvider = ( {
 								: undefined;
 							addErrorNotice( response.message, errorOptions );
 						}
-						if (
-							typeof response.retry !== 'undefined' &&
-							response.retry !== true
-						) {
+						if ( ! shouldRetry( response ) ) {
 							dispatch( actions.setComplete( response ) );
 						} else {
 							// this will set an error which will end up
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -278,7 +278,10 @@ export const CheckoutStateProvider = ( {
 								: undefined;
 							addErrorNotice( response.message, errorOptions );
 						}
-						if ( ! response.retry ) {
+						if (
+							typeof response.retry !== 'undefined' &&
+							response.retry !== true
+						) {
 							dispatch( actions.setComplete( response ) );
 						} else {
 							// this will set an error which will end up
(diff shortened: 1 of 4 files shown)


senadir:
feels weird to treat undefined as true, why not make sure payment methods return the correct value?

What's the behavior we're aiming for here?

- Always retry unless specified
- Never retry unless specified

senadir:
Maybe, based on that, we can decide to rename this variable to make it obvious.

Aljullu:
> why not make sure payment methods return the correct value?

The idea is to provide good defaults, so there is less risk for payment methods to introduce bugs (like #3221).

I don't have any data supporting why the default should be true instead of false, so I'm happy to be proven wrong, but at least for Stripe and WC Payments we needed to support retrying after an error, and I can imagine that's the most common behavior among other payment methods.

In any case, `checkoutAfterProcessingWithSuccess` and `checkoutAfterProcessingWithError` are currently behaving differently, so whatever decision we end up taking, we need to make them work consistently.

> Always retry unless specified

Correct, that's the behavior we are trying to achieve: always retry unless `retry: false` (see discussion in 978-GH-woocommerce-payments).

Note: If we were to rename `retry` to something else, that would probably be a breaking change.

senadir (approved):
I think this makes sense in the broader terms, maybe we can change the wording from 'undefined' to saying it defaults to true. 

I'd probably also wait and see what Darren has to say about this.

nerrad (approved):
Looks good to me, I do wonder if we should add a `shouldRetry` function to the `useEmitResponse` hook to include this check and logic centralized there? That way the condition could just be:

```js
if ( shouldRetry( response ) )
```

Which is a bit clear and keeps the actual default coercion logic in the `useEmitResponse` hook?

I'm going to pre approve anyways - up to you Albert whether you modify for the above or not.

Aljullu:
Thanks for the reviews @senadir @nerrad!

I moved the logic of `shouldRetry` to `useEmitResponse` in 7fa0582fda1edb7de8f9b39ead3696f7de3f45c2 and updated the changelog notice of this PR accordingly.

Issue #3262, opened 2020-10-09
Integration Tracking (WC Subscriptions): Add Subscription Product data to store cart item endpoint


senadir:
In order to properly render the Subscription products in the cart and checkout block, we need to expose certain data, this data is three folds:

- Global subscription settings data (exposed via wcSettings).
- Subscription Product data (exposed via REST API).
- Synchronization global data (via wcSettings) and Product specific synchronization data (exposed via REST API)

Because synchronization is a complicated thing, I'm leaving it for i2 or something.

This issue would only track adding *Subscription Product data*, other issues would be responsible for tracking the rest.

Given the designs, I anticipate that we will need:
- Subscription interval (the number part of "every 3 months", / month), before we charge again.
- Subscription period (day, week, month, year).
- Subscription length (for how long before the subscription expires), this uses the same period as the subscription period.
- Trial length.
- Trial period (day, week, month, year).
- Fees (a price to be paid upfront).

We will also need some derived data that we can calculate ourself but might cause issues, this data is mostly around dates, we will derive all other data:
- First renewal, can be derived from (trial length * trial period) + (subscription period *  subscription interval) + current time.

senadir:
This is closed by 3839-gh-woocommerce/woocommerce-subscriptions

Issue #3244, opened 2020-10-06
Integration Tracking (WC Subscriptions): Add filters to our REST endpoints


senadir:
To expose certain data types in our interface, we need to have that data in the first place.

The first option is to call another API, but this will cause an extra API request.

The best option is to extend our store API, this can be done with extra filters.

For the initial work, we will need filters in:

- Cart items.
- Coupons.

Under consideration, but not final:
- Shipping methods.
- Product.

Those filters should be private (not publicly documented) until we're certain we want to ship them.

senadir:
After some discussions with @nerrad the approach we're taking here is this:
- instead of filtering the response directly `return apply_filter( "woocommerce_blocks_rest_cart_item", $data );` we would only append data, so our `$data` array would be like this:

```php
$data = [
// original data
"extensions" => apply_filter( "woocommerce_blocks_rest_cart_item", [] ),
];

// plugin code
add_filter( "woocommerce_blocks_rest_cart_item", function( $response ) {
    return $response[ $plugin_namespace ] = $data;
});
```

This ensures that our original data shape is never altered, but limits extensions ability to update its values from this filter. However, they can still hook into earlier places with existing filters in WooCommerce Core.

Plugins need to add data under a namespace to not compete for the same key.

We can go even fancier and introduce a wrapper thing that abstracts the filter.

```php
wc_blocks_extend_item_response( $plugin_namespace, $data );
```

The approach we use should be consistent between all of our extension endpoints.
Instead of trusting that extensions won't mess core data, we should only trust that they won't override each other data.

nerrad:
There's some good initial plans there, however I would like us to avoid adding the filter directly on the data response and introduce a registration mechanism (like you suggested) that allows for a few things:

- adding api requirements.
- validating what is registered.
- protecting extensions from clobbering other extensions (and thus increasing the risk of breaking other integrations).
- ensuring things are namespaced properly.

senadir:
After more digging, I found `add_rest_field`, which should work exactly as we want.

https://developer.wordpress.org/reference/functions/register_rest_field/

It checks all the things we need except the namespace one. We can implement something like this, or abstract this to our needs.

PR #3263, opened 2020-10-09
Add an interim filter for cart item and expose prepare_money_response

senadir:
### Add interim hook
In order to add subscription data to our endpoints, we need to provide some sort of a hook place, we discussed the approach we're going to take in this issue https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3244, but in order to not limit myself and overengineer it beforehand, I went ahead and created this interim filter that will be deleted once I implement what's discussed in that issue.

### Turn prepare_money_response to public
This function is used to format any monetary value, but it's protected, so third party code can't use it.
It's part of `AbstractSchema` and is only relevant when returning money in REST.
I temporarily turned it to public and will create an issue to decide if we want to keep it public or move it to a utility class.

## Why two changes in a single PR

Ideally, I would split those into two PRs, but to facilitate testing https://github.com/woocommerce/woocommerce-subscriptions/pull/3839 and because they're not permanent changes, I decided to keep them in a single PR. They're both needed for that PR.

### DIFF
--- a/src/StoreApi/Schemas/AbstractSchema.php
+++ b/src/StoreApi/Schemas/AbstractSchema.php
@@ -230,7 +230,7 @@ abstract class AbstractSchema {
 	 * @param int          $rounding_mode Defaults to the PHP_ROUND_HALF_UP constant.
 	 * @return string      The new amount.
 	 */
-	protected function prepare_money_response( $amount, $decimals = 2, $rounding_mode = PHP_ROUND_HALF_UP ) {
+	public function prepare_money_response( $amount, $decimals = 2, $rounding_mode = PHP_ROUND_HALF_UP ) {
 		return (string) intval(
 			round(
 				( (float) wc_format_decimal( $amount ) ) * ( 10 ** $decimals ),
--- a/src/StoreApi/Schemas/CartItemSchema.php
+++ b/src/StoreApi/Schemas/CartItemSchema.php
@@ -278,7 +278,7 @@ class CartItemSchema extends ProductSchema {
 	public function get_item_response( $cart_item ) {
 		$product = $cart_item['data'];
 
-		return [
+		$item_data = [
 			'key'                  => $cart_item['key'],
 			'id'                   => $product->get_id(),
 			'quantity'             => wc_stock_amount( $cart_item['quantity'] ),
@@ -305,6 +305,9 @@ class CartItemSchema extends ProductSchema {
 				]
 			),
 		];
+
+		// This filter will be delete an replaced by a proper register mechanism, don't use it.
+		return apply_filters( '__internal_woocommerce_blocks_cart_item', $item_data, $cart_item, $product, $this );
 	}
 
 	/**


senadir:
While this issue is linked to #3262, it should not close it. That issue should be closed by a PR in Subscriptions repo.

senadir:
Added relevant tracking issue for `prepare_money_response`
https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3264

nerrad:
> In order to add subscription data to our endpoints, we need to provide some sort of a hook place, we discussed the approach we're going to take in this issue #3244, but in order to not limit myself and overengineer it beforehand, I went ahead and created this interim filter that will be deleted once I implement what's discussed in that issue.

I appreciate this. However, I wonder if the same can be achieved by an early iteration of something like `__experimental_register_endpoint_data` (or some other better function name) or `register_rest_field` if it works (which I don't think it will because it looks like it's coupled to custom post types?) for enhancing our endpoints. I'm wary about doing what you're doing here (even with it explicitly being temporary) because we definitely know we won't be using filters long term. So let's iterate on an interface we _will_ use.

> I temporarily turned it to public and will create an issue to decide if we want to keep it public or move it to a utility class.

The problem with doing this is that it could land in a release as a public method which then restricts us from making it protected/private again. The other issue here is that there's no clear indication in _this pull alone_, why it's being made public. I understand that you are doing this because the object instance is being passed along through the temporary filter, but this to me is another indicator that we should be thinking a bit more carefully about the interface we want to build for registering different fields on the endpoints.

senadir:
> The problem with doing this is that it could land in a release as a public method which then restricts us from making it protected/private again. The other issue here is that there's no clear indication in this pull alone, why it's being made public. I understand that you are doing this because the object instance is being passed along through the temporary filter, but this to me is another indicator that we should be thinking a bit more carefully about the interface we want to build for registering different fields on the endpoints.

My basis for doing this is that if it's not announced publicly, it's not public. I understand that this is not a bullet-proof method.
This change is not final, but an interim solution while we [decide what to do.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3264)


What kind of change do you suggest for this PR? postpone the release of this method? or move it to an Utility class?

For the action, I do agree, I will handle it after our cooldown.

If both actions are to redo, I can close this PR and we can try again independently (register method in a PR and utility class in a PR).

nerrad:
> My basis for doing this is that if it's not announced publicly, it's not public. I understand that this is not a bullet-proof method.
> This change is not final, but an interim solution while we decide what to do.

In the current climate of people looking for ways to extend things, it's unlikely it won't get noticed. I know it's intended to be interim but that's all the more reason to avoid making public _in this class_.  Think of it this way:

- the only reason it's being made public is so it can be exposed to filter callbacks.
- the filter is a temporary implementation that we've already agreed is something we should avoid here.

Also think we should explore how the API we implement for extending the endpoint could automatically handle formatting money data so consumers don't do that.

senadir:
I'm going to close this PR and start over with a `__experimental_register_endpoint_data` prototype, this should be a globally available function @nerrad? something we never did in our plugin, so I don't know if this is the best approach.

**the question here is: should `__experimental_register_endpoint_data` be part of some class, or a regular function?**

For `prepare_money_response` I'm going to try and implement it as part of `__experimental_register_endpoint_data`

nerrad:
I think a _similar_ (but not exact) model for registration might be the [`register_post_meta` function](https://developer.wordpress.org/reference/functions/register_post_meta/) in WordPress core.

I wonder if this should be part of some sort of `ExtendRestApi` service class that could be used by extensions?

senadir:
> I wonder if this should be part of some sort of ExtendRestApi service class that could be used by extensions?

That would certainly solve the `prepare_money_response` but I'm not if it will complicate things more for developers.

I will start with that and see where it takes me.

PR #3269, opened 2020-10-12
Some changes/additions to the release checklist

Aljullu:
After releasing WC Blocks 3.7, I have to say the process worked like a charm. A big part of the tasks that were manual in the past are now automatic and the pull request format with the checklist works great. :clap: 

During the process, I found a couple of places where I think we could improve the checklist even a bit more.

### DIFF
--- a/.github/release-initial-checklist.md
+++ b/.github/release-initial-checklist.md
@@ -50,7 +50,7 @@ When creating testing notes, please write them from the perspective of a "user" 
   * [ ] Confirm [WooCommerce.com plugin page](https://woocommerce.com/products/woocommerce-gutenberg-products-block/) is updated.
   * [ ] Download zip and smoke test.
   * [ ] Test updating plugin from previous version.
-* [ ] Merge this pull request back into `trunk`. This may have merge conflicts needing resolved if there are any cherry-picked commits in the release branch. After merging, GitHub will automatically delete the branch, restore it.
+* [ ] Merge this pull request back into `trunk`. This may have merge conflicts needing resolved if there are any cherry-picked commits in the release branch.
 * [ ] Update version on the `trunk` branch to be for the next version of the plugin and include the `dev` suffix (e.g. something like [`2.6-dev`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/e27f053e7be0bf7c1d376f5bdb9d9999190ce158)) for the next version.
 * [ ] Update the schedules p2 with the shipped date for the release (Pca54o-1N-p2).
 * [ ] Clean up the release milestone and Zenhub.
--- a/.github/release-initial-checklist.md
+++ b/.github/release-initial-checklist.md
@@ -14,11 +14,11 @@ The release pull request has been created! This checklist is a guide to follow f
 
 When creating testing notes, please write them from the perspective of a "user" (merchant) familiar with WooCommerce. So you don't have to spell out exact steps for common setup scenarios (eg. "Create a product"), but do be specific about the thing being tested. Include screenshots demonstrating expectations where that will be helpful.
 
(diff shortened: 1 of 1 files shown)


.github/release-initial-checklist.md
```diff
@@ -14,11 +14,11 @@ The release pull request has been created! This checklist is a guide to follow f
...
 * [ ] Run `npm ci`
 * [ ] Run `npm run package-plugin:deploy`. This will create a zip of the current branch build locally.
-* [ ] Copy a link to the release zip into the testing notes you created earlier. To generate the link you can upload the zip as an attachment in a GitHub comment and then just copy the path (without publishing the comment).
+* [ ] Create testing notes for the release. You can usually go through the pull requests linked in the changelog and grab testing notes from each pull. Add the notes to `docs/testing/releases` and update the `docs/testing/releases/README.md` index.
+  * Note, make sure to differentiate between things in the testing notes that only apply to the feature plugin and things that apply when included in WooCommerce core as there may be variations there.
+* [ ] Copy a link to the release zip you created earlier into the testing notes. To generate the link you can upload the zip as an attachment in a GitHub comment and then just copy the path (without publishing the comment).
```
Aljullu, line 21: If you write the testing steps and then run `npm run package-plugin:deploy`, the command will complain that the directory is dirty because there are uncommitted changes. That can be avoided committing the testing steps (without the ZIP file yet) or staging the changes and restoring them after generating the ZIP. But I think it would be easier to generate the ZIP as the first step, and then write the testing steps once the ZIP file is ready.

.github/release-initial-checklist.md
```diff
@@ -50,11 +50,11 @@ When creating testing notes, please write them from the perspective of a "user"
   * [ ] Confirm [WooCommerce.com plugin page](https://woocommerce.com/products/woocommerce-gutenberg-products-block/) is updated.
   * [ ] Download zip and smoke test.
   * [ ] Test updating plugin from previous version.
-* [ ] Merge this pull request back into `trunk`. This may have merge conflicts needing resolved if there are any cherry-picked commits in the release branch.
+* [ ] Merge this pull request back into `trunk`. This may have merge conflicts needing resolved if there are any cherry-picked commits in the release branch. After merging, GitHub will automatically delete the branch, restore it.
```
Aljullu, line 53: Correct me if I'm wrong, but I think the intention is that `release/x.y` branches should not be deleted.

.github/release-initial-checklist.md
```diff
@@ -50,11 +50,11 @@ When creating testing notes, please write them from the perspective of a "user"
...
 * [ ] Update version on the `trunk` branch to be for the next version of the plugin and include the `dev` suffix (e.g. something like [`2.6-dev`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/e27f053e7be0bf7c1d376f5bdb9d9999190ce158)) for the next version.
 * [ ] Update the schedules p2 with the shipped date for the release (Pca54o-1N-p2).
 * [ ] Clean up the release milestone and Zenhub.
-  * [ ] Edit the GitHub milestone and add the current date as the due date (this is used to track ship date as well).
+  * [ ] Edit the [GitHub milestone](https://github.com/woocommerce/woocommerce-gutenberg-products-block/milestones) and add the current date as the due date (this is used to track ship date as well).
```
Aljullu, line 57: We can save some seconds to the release lead adding a link here.

senadir (approved):

.github/release-initial-checklist.md
nerrad: Restoring them will cause the pull request automation to run again, so for now we just leave deleted - thus intentionally this doesn't mention restoring. They can be restored as needed for hotfix down the road (there's an open issue in the pull request automation to fix as well).

Aljullu: Ah, that explains why there was a duplicate PR for 3.6.0 release (#3268). :smile: 

I reverted this change in 70a422bad5f4c33e29cd35bce8a2973e3bd35aa6.

PR #3272, opened 2020-10-13
Fix 3D secure payment errors

### LINKED ISSUE #2743: 2x confusing error notices + broken checkout when user cancels 3D secure payment

Aljullu:
Fixes #2743.

This PR fixes a couple of issues with 3D secure payment errors:

* After the error, it was not possible to make another payment without making a change in the card number. That was because we were not resetting the source id on error (5095d55ce25ca3361f384cafe94ea4eb70ada00c).
* Two errors were appearing instead of one. That's because the payment method error is created on `onCheckoutAfterProcessingSuccess`, when an error is created at that stage, it then triggers `onCheckoutAfterProcessingError` events. If no error was returned from `onCheckoutAfterProcessingError` events, we were adding a [default one](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/cart-checkout/checkout-state/index.js#L247-L257). That's why two errors were appearing: the one generated by the payment method in `onCheckoutAfterProcessingSuccess` and the default one generated by Blocks in `onCheckoutAfterProcessingError`.
~For now, I fixed it in Stripe's side saving the error message internally and returning it on `onCheckoutAfterProcessingError` (4a5c04eedcd9f63ab2c4cde2fef6cd95078a71a7), but I feel that's a bit complicated for payment methods. For example, the same issue occurs in WC Payments so we will need to fix it there too. That makes me wonder if we could make the process easier. Maybe removing the default error message that is added on `onCheckoutAfterProcessingError`? Another option would be not to trigger `onCheckoutAfterProcessingError` after an error occurred in `onCheckoutAfterProcessingSuccess`. Thoughts @nerrad?~
Update: we decided to add a check before showing the default error notice. So if there is already an error notice, the default one won't be added in `onCheckoutAfterProcessingError`.
* Fixed missing `useEffect()` dependencies (33280b5fd86272cd52ad810ac4e18539cf5d215b).
* Typo (d97a0fa3b94e464bd8d667e83c3e7f186a5269fc).

### Screenshots

Peek 2020-10-13 10-22

1. Install Stripe payment gateway and set it up.
2. Add products to the cart and go to a page with the Checkout block.
3. Place an order using Stripe with the card number `4000 0027 6000 3184` (it will open an authentication modal).
4. Click Cancel or fail the authentication.
5. Verify only one error is displayed, inside the payment methods step.
6. Without modifying the payment number, click on Place order again and verify the authentication modal appears again and you can place the order.

Now, let's pretend there is a payment gateway which doesn't add an error message when throwing an error. This way, we will ensure a default error notice is shown.

1. In [`use-payment-intents.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3272/files#diff-f1a85fd67379d04a2efe056f9ff6b51012e543cfd029212c8ccb7c657c17bd60R86-R91), make the following change:
```diff
			if (
				response.type === emitResponse.responseTypes.ERROR &&
				response.retry
			) {
				setSourceId( '0' );
+				return { type: response.type, retry: response.retry };
			}
```
2. Add products to the cart and go to a page with the Checkout block.
3. Place an order using Stripe with the card number `4000 0027 6000 3184` (it will open an authentication modal).
4. Click Cancel or fail the authentication.
5. Verify an error is displayed with a generic text, at the top of the form.

### Changelog

> Fix a couple of issues when a 3D secure payment was aborted: now only one error notice appears and the payment can be retried with the same card number.

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -280,7 +280,7 @@ export const CheckoutStateProvider = ( {
 							dispatch( actions.setComplete( response ) );
 						} else {
 							// this will set an error which will end up
-							// triggering the onCheckoutAfterProcessingWithErrors emitter.
+							// triggering the onCheckoutAfterProcessingWithError emitter.
 							// and then setting checkout to IDLE state.
 							dispatch( actions.setHasError( true ) );
 						}
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -292,6 +292,7 @@ export const CheckoutStateProvider = ( {
 		isErrorResponse,
 		isFailResponse,
 		isSuccessResponse,
+		shouldRetry,
 	] );
 
 	const onSubmit = useCallback( () => {
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -10,7 +10,6 @@ import {
 	useEffect,
 	useCallback,
 } from '@wordpress/element';
-import { __ } from '@wordpress/i18n';
 import { useStoreNotices, useEmitResponse } from '@woocommerce/base-hooks';
 
 /**
@@ -240,23 +239,11 @@ export const CheckoutStateProvider = ( {
 						// irrecoverable error so set complete
 						if ( ! shouldRetry( response ) ) {
 							dispatch( actions.setComplete( response ) );
-						} else {
-							dispatch( actions.setIdle() );
+							return;
 						}
-					} else {
-						// no error handling in place by anything so let's fall
-						// back to default
-						const message =
-							data.processingResponse?.message ||
-							__(
-								'Something went wrong. Please contact us to get assistance.',
-								'woo-gutenberg-products-block'
-							);
-						addErrorNotice( message, {
-							id: 'checkout',
-						} );
-						dispatch( actions.setIdle() );
 					}
+
+					dispatch( actions.setIdle() );
 				} );
 			} else {
 				emitEventWithAbort(
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -15,6 +15,7 @@ import {
 	useCheckoutNotices,
 	useStoreNotices,
 	useEmitResponse,
+	usePrevious,
 } from '@woocommerce/base-hooks';
 
 /**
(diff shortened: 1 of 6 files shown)


nerrad:
> For now, I fixed it in Stripe's side saving the error message internally and returning it on onCheckoutAfterProcessingError (4a5c04e), but I feel that's a bit complicated for payment methods. For example, the same issue occurs in WC Payments so we will need to fix it there too. That makes me wonder if we could make the process easier. Maybe removing the default error message that is added on onCheckoutAfterProcessingError? Another option would be not to trigger onCheckoutAfterProcessingError after an error occurred in onCheckoutAfterProcessingSuccess. Thoughts @nerrad?

Ya I see the issue here. The intention behind the design was to always ensure that other things hooked into the event emitters can react to any errors that might have happened (which is why an error from the success emitter will in turn trigger the `onCheckoutAfterProcessingWithErrors` emitter).

I do agree we need to try and make things fairly simple for payment methods instead of the approach you've done so far in this pull. For subscribers registered to the events, they should only care about reacting to the events, not having to worry about internals causing duplicate notices etc.

I think removing the default error message that is added on `onCheckoutAfterProcessingError` might be the way to go here, but could we do that only if there is an error notice already queued up (otherwise if there are no existing error notices, let's add the default)?

Aljullu:
> I think removing the default error message that is added on `onCheckoutAfterProcessingError` might be the way to go here, but could we do that only if there is an error notice already queued up (otherwise if there are no existing error notices, let's add the default)?

Make sense, I needed to create a `useCheckoutNotices()` hook so we could check notices in all contexts. I updated the PR description and testing steps accordingly.

This PR is ready for review again.

nerrad (commented):
The `useStoreNotices` hook exposes a `hasNoticesOfType` function. Would that suffice for doing the notice check and remove the need for the new `useCheckoutNotices` hook? That would reduce the new code added?

Aljullu:
`useStoreNotices()` only exposes notices in the current context, but in this case we want to count if there are error notices in any of these three contexts: `wc/checkout`, `wc/express-payment-area`, `wc/payment-area`. I don't think `useStoreNotices()` has any way to achieve that?

nerrad:
Ahh, right! I am a bit wary about that if we add other notice contexts to checkout, this won't automatically account for them. Still it might be premature optimization to try and figure that out right now.

budzanowski:
As far as I can understand the code it makes sense to me and it works.

budzanowski (approved):

PR #3284, opened 2020-10-14
Document how to update E2E WP versions.

senadir:
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2721

This adds some docs to how we update our E2E WP version numbers.

### DIFF
--- a/docs/contributors/javascript-testing.md
+++ b/docs/contributors/javascript-testing.md
@@ -7,7 +7,7 @@ The Blocks plugin follows the same patterns as Gutenberg, therefore for instruct
 We have two kinds of JavaScript tests:
 
 - JavaScript unit tests - test APIs, hooks, library functionality that we use to build blocks or expose to plugin authors.
-- End-to-end (e2e) tests - test blocks from the user interface. 
+- End-to-end (e2e) tests - test blocks from the user interface.
 
 These tests are all run automatically on open PRs by Travis CI.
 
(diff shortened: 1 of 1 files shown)


docs/contributors/javascript-testing.md
```diff
@@ -52,3 +52,46 @@ When you're done, you may want to shut down the test environment:
 - `npm run wp-env stop` to stop the test environment
 
 **Note:** There are a number of other useful `wp-env` commands. You can find out more in the [wp-env docs](https://github.com/WordPress/gutenberg/blob/master/packages/env/README.md).
+
+### How to update end-to-end tests suites
+
+We follow the same WordPress support policy as WooCommerce, this means we need to support the latest version, and the two previous one (L-2).
```
budzanowski, line 58: I think that it is `ones` instead of  `one`.

budzanowski:
Looks good. I wonder if we could just script it. WordPress has version API available: https://codex.wordpress.org/WordPress.org_API#Version_Check 
:shipit:

budzanowski (approved):
One small nitpick, but good to go otherwise.

senadir:
Automating it is a good option however I'm not sure of the investment (and maintenance) vs the benefit?
If we can automate an issue to be created once every release, it would hit a good balance of not too much work but enough awareness.

Issue #3283, opened 2020-10-14
Price slider max constraint doesn't match price filter input value


Aljullu:
## Describe the bug
It's possible to achieve a situation where price slider values differ from values shown in the input fields.

## To reproduce
If you are using WC test data, you can use the same values I will describe below, otherwise you might need to use different prices/attributes to achieve a similar result.

1. Create a page with Price Filter, Attribute Filter (chose _Color_ filter) and All Products blocks and visit it in the frontend.
2. Set the price slider to 30-70.
3. In the attributes filter, select _Blue_.
3. Notice the inputs say the price is filtered between 30-70, but the price slider represents it as being 30-50 (which is the maximum price of a product with the attribute Blue).

## Expected behavior
It should either show 30-50 in the input fields or the price slider should represent 30-70, so both elements are consistent.

## Screenshots
![Peek 2020-10-14 14-08](https://user-images.githubusercontent.com/3616980/95986830-e183f600-0e26-11eb-9b6a-2444f089095b.gif)

Aljullu:
Still reproducible.

danieldudzic:
I have just tested it now, and it seems to no longer be an issue:

![N626sHsn21](https://user-images.githubusercontent.com/905781/211021756-da08b181-2e4d-4564-aebd-badabac9fa67.gif)

Aljullu:
I can still reproduce:

![imatge](https://user-images.githubusercontent.com/3616980/211308870-c7ae4d6d-af30-4d17-92b9-b624891ee4fe.png)

See how the price range goes from 17 to 70, but the slider is represent it as if it was 17-20 with 20 being the maximum.

I think you need to select the price range _before_ setting the attribute filter.

Issue #3282, opened 2020-10-14
Price slider goes back to 0 after moving the min handle


Aljullu:
## Describe the bug
The minimum price handle of the price slider some times goes back to 0 after being moved. I'm not sure exactly why and when it happens. I wrote some testing steps below, but I could only reproduce it intermittently and only with Firefox, but it might affect other browsers.

## To reproduce

1. Go to a page with product filter blocks and All Products block.
2. Move the minimum price handle in the Filter products by price block.
3. Drop it right after it goes to a loading state.
4. Notice some times the handle moves back to 0.

## Screenshots
![Peek 2020-10-14 13-49](https://user-images.githubusercontent.com/3616980/95985216-9ec11e80-0e24-11eb-8b9e-dbc8c1db657c.gif)

nerrad:
I'm unable to reproduce after trying for a few minutes. It's _possible_ this issue got fixed recently.

Issue #3204, opened 2020-09-26
Make sure React Hook dependency warnings from ESLint rule are dealt with.


nerrad:
As of writing this issue there are **34** eslint lint rule warnings for the `react-hooks/exhaustive-deps` rule. Typically these are either due to missing dependencies in React hook usage, or unnecessary dependencies.

To date we've taken the approach that they should be addressed in any work touching existing hooks generating these warnings but I think it'd be a good idea to take care of them in one go during a cooldown. Hence this issue.

Missing dependencies are great code smell for either a non-detected bug, or a bug waiting to happen.

PR #3285, opened 2020-10-14
Fix React hook dependency warnings in filter an All Products blocks

Aljullu:
Part of #3204.

This PR handles the React hook dependency warnings produced by ESLint in the filter and All Products blocks. Some fixes were trivial while some others needed bigger refactors. Adding new dependencies to effects/memos/callbacks might have caused them to do unnecessary recalculations, when that was evident, I tried to prevent it either wrapping some dependencies with `useShallowEqual()` or refactoring it completely. However, I might have missed some and for sure there are many more optimizations to do.

The scope of changes in this PR is limited to the filter and All Products blocks, so those are the ones that need to be tested.

1. Create a page with the Filter products by price, Filter products by attribute, Active filters and All Products blocks.
2. Verify you can switch pages in the product list.
3. Verify you can change the sorting order and it sets the page back to 1.
4. Verify that when you activate a filter, the pagination goes back to 1.
5. Try setting several filters at the same time, removing them, clearing them all at once, etc.
6. Summarizing, try to break the blocks. :slightly_smiling_face: 
In the process I found a couple of bugs (#3282 and #3283), but I could reproduce them in `trunk` too, so I filled issues in GH.
9. Repeat the process above but first edit the filter blocks so they have the 'Filter button' visible.
imatge
10. Extra testing modifying other attributes of the blocks would be great.

### DIFF
--- a/assets/js/base/components/product-list/index.js
+++ b/assets/js/base/components/product-list/index.js
@@ -70,7 +70,7 @@ const generateQuery = ( { sortValue, currentPage, attributes } ) => {
 const extractPaginationAndSortAttributes = ( query ) => {
 	/* eslint-disable-next-line no-unused-vars, camelcase */
 	const { order, orderby, page, per_page, ...totalQuery } = query;
-	return totalQuery;
+	return totalQuery || {};
 };
 
 const announceLoadingCompletion = ( totalProducts ) => {
@@ -136,21 +136,25 @@ const ProductList = ( {
 		{ totalQuery, totalProducts },
 		areQueryTotalsDifferent
 	);
-	const isPreviousTotalQueryEqual =
-		typeof previousQueryTotals === 'object' &&
-		isEqual( totalQuery, previousQueryTotals.totalQuery );
-	useEffect( () => {
-		// If query state (excluding pagination/sorting attributes) changed,
-		// reset pagination to the first page.
-		if ( ! isPreviousTotalQueryEqual ) {
-			onPageChange( 1 );
 
-			// Make sure there was a previous query, so we don't announce it on page load.
-			if ( previousQueryTotals ) {
-				announceLoadingCompletion( totalProducts );
-			}
+	// If query state (excluding pagination/sorting attributes) changed,
+	// reset pagination to the first page.
+	useEffect( () => {
+		if ( isEqual( totalQuery, previousQueryTotals?.totalQuery ) ) {
+			return;
 		}
-	}, [ queryState ] );
+		onPageChange( 1 );
+
+		// Make sure there was a previous query, so we don't announce it on page load.
+		if ( previousQueryTotals?.totalQuery ) {
+			announceLoadingCompletion( totalProducts );
+		}
+	}, [
+		previousQueryTotals?.totalQuery,
+		totalProducts,
+		onPageChange,
+		totalQuery,
+	] );
 
 	const onPaginationChange = ( newPage ) => {
 		scrollToTop( { focusableSelector: 'a, button' } );
@@ -175,7 +179,9 @@ const ProductList = ( {
 	const { contentVisibility } = attributes;
 	const perPage = attributes.columns * attributes.rows;
 	const totalPages =
-		! Number.isFinite( totalProducts ) && isPreviousTotalQueryEqual
+		! Number.isFinite( totalProducts ) &&
+		Number.isFinite( previousQueryTotals?.totalProducts ) &&
+		isEqual( totalQuery, previousQueryTotals?.totalQuery )
 			? Math.ceil( previousQueryTotals.totalProducts / perPage )
 			: Math.ceil( totalProducts / perPage );
 	const listProducts = products.length
--- a/assets/js/base/components/product-list/index.js
+++ b/assets/js/base/components/product-list/index.js
@@ -96,6 +96,11 @@ const announceLoadingCompletion = ( totalProducts ) => {
 	}
 };
 
+const areQueryTotalsDifferent = (
+	{ totalQuery: nextQuery, totalProducts: nextProducts },
+	{ totalQuery: currentQuery } = {}
+) => ! isEqual( nextQuery, currentQuery ) && Number.isFinite( nextProducts );
+
 const ProductList = ( {
 	attributes,
 	currentPage,
@@ -129,12 +134,7 @@ const ProductList = ( {
 	// the total number of products is a finite number.
 	const previousQueryTotals = usePrevious(
 		{ totalQuery, totalProducts },
-		(
-			{ totalQuery: nextQuery, totalProducts: nextProducts },
-			{ totalQuery: currentQuery } = {}
-		) =>
-			! isEqual( nextQuery, currentQuery ) &&
-			Number.isFinite( nextProducts )
+		areQueryTotalsDifferent
 	);
 	const isPreviousTotalQueryEqual =
 		typeof previousQueryTotals === 'object' &&
--- a/assets/js/base/hooks/use-previous.js
+++ b/assets/js/base/hooks/use-previous.js
@@ -20,7 +20,7 @@ export const usePrevious = ( value, validation ) => {
 		) {
 			ref.current = value;
 		}
-	}, [ value, ref.current ] );
+	}, [ value, validation ] );
 
 	return ref.current;
 };
--- a/assets/js/base/hooks/use-query-state.js
+++ b/assets/js/base/hooks/use-query-state.js
@@ -5,6 +5,8 @@ import { QUERY_STATE_STORE_KEY as storeKey } from '@woocommerce/block-data';
 import { useSelect, useDispatch } from '@wordpress/data';
 import { useRef, useEffect, useCallback } from '@wordpress/element';
 import { useQueryStateContext } from '@woocommerce/base-context';
+import { usePrevious } from '@woocommerce/base-hooks';
+import isShallowEqual from '@wordpress/is-shallow-equal';
 import { assign } from 'lodash';
 
 /**
@@ -43,7 +45,7 @@ export const useQueryStateByContext = ( context ) => {
 		( value ) => {
 			setValueForQueryContext( context, value );
 		},
-		[ context ]
+		[ context, setValueForQueryContext ]
 	);
 
 	return [ queryState, setQueryState ];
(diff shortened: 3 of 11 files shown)


senadir:
Note that I'm aware of this PR and will get to it in the next couple of days unless it's blocking any other work or there's a risk it will get outdated quickly, in which case I can review it today?

Aljullu:
Sounds good @senadir. I don't think there is any hurry as long as it's dealt with before the end of cooldown. :slightly_smiling_face:

senadir (approved):
I'm actually a bit scared of merging this 😅
It's a lot of changes that are not obvious to me, I can see possible hidden regressions in the future.
I followed the testing steps and nothing broke, so we're comfortable with that.
I can't think of any other testing scenario.

So, I'm going to approve this and would be happy to merge it.
However, if you have any free time now or later, I would really appreciate some comments here and there to justify the changes, it would be helpful for future developers if we track a bug down to this PR.

Usually, I would suggest doing that in a follow-up issue, but the more we delay this, the greater chances that you forget why you did those changes.

assets/js/base/components/product-list/index.js
```diff
@@ -175,7 +179,9 @@ const ProductList = ( {
 	const { contentVisibility } = attributes;
 	const perPage = attributes.columns * attributes.rows;
 	const totalPages =
-		! Number.isFinite( totalProducts ) && isPreviousTotalQueryEqual
+		! Number.isFinite( totalProducts ) &&
+		Number.isFinite( previousQueryTotals?.totalProducts ) &&
+		isEqual( totalQuery, previousQueryTotals?.totalQuery )
```
Aljullu, line 184: The check that was named `isPreviousTotalQueryEqual`, has been moved here. Instead of checking that `previousQueryTotals` is an object, we directly check if it has the `totalProducts` property and it's a finite number.

assets/js/base/components/product-list/index.js
```diff
@@ -129,28 +134,27 @@ const ProductList = ( {
...
-		// If query state (excluding pagination/sorting attributes) changed,
-		// reset pagination to the first page.
-		if ( ! isPreviousTotalQueryEqual ) {
-			onPageChange( 1 );
-
-			// Make sure there was a previous query, so we don't announce it on page load.
-			if ( previousQueryTotals ) {
-				announceLoadingCompletion( totalProducts );
-			}
+		if ( isEqual( totalQuery, previousQueryTotals?.totalQuery ) ) {
```
Aljullu, line 143: Same as before.

assets/js/base/hooks/use-previous.js
```diff
@@ -20,7 +20,7 @@ export const usePrevious = ( value, validation ) => {
 		) {
 			ref.current = value;
 		}
-	}, [ value, ref.current ] );
+	}, [ value, validation ] );
```
Aljullu, line 23: Missing dep in a `useEffect()`. I needed to update calls to `usePrevious()` in other components to ensure `validation` doesn't change on each render (I commented it inline in those cases).

assets/js/base/hooks/use-query-state.js
```diff
@@ -43,7 +45,7 @@ export const useQueryStateByContext = ( context ) => {
 		( value ) => {
 			setValueForQueryContext( context, value );
 		},
-		[ context ]
+		[ context, setValueForQueryContext ]
```
Aljullu, line 48: Missing dep in a `useCallback()`. It should be a safe change.

assets/js/base/hooks/use-query-state.js
```diff
@@ -81,7 +83,7 @@ export const useQueryStateByKey = ( queryKey, defaultValue, context ) => {
 		( value ) => {
 			setQueryValue( context, queryKey, value );
 		},
-		[ context, queryKey ]
+		[ context, queryKey, setQueryValue ]
```
Aljullu, line 86: Missing dep in a `useCallback()`. It should be a safe change.

assets/js/base/hooks/use-query-state.js
```diff
@@ -117,15 +119,31 @@ export const useSynchronizedQueryState = ( synchronizedQuery, context ) => {
...
 	useEffect( () => {
-		setQueryState( assign( {}, queryState, currentSynchronizedQuery ) );
-		isInitialized.current = true;
-	}, [ currentSynchronizedQuery ] );
+		if (
+			! isShallowEqual(
+				previousSynchronizedQuery,
+				currentSynchronizedQuery
+			)
+		) {
```
Aljullu, line 135: We only need to run this effect when `currentSynchronizedQuery` changes, but since we added new dependencies in the deps array, we need to manually check that `currentSynchronizedQuery` is not equal to the previous one, so we don't run this effect in some unnecessary cases.

assets/js/blocks/active-filters/block.js
```diff
@@ -47,7 +47,13 @@ const ActiveFiltersBlock = ( {
...
 			displayStyle: blockAttributes.displayStyle,
 		} );
-	}, [ minPrice, maxPrice, formatPriceRange ] );
+	}, [
+		minPrice,
+		maxPrice,
+		blockAttributes.displayStyle,
+		setMinPrice,
+		setMaxPrice,
+	] );
```
Aljullu, line 56: Missing deps in a `useMemo()`, it should be safe.

assets/js/blocks/active-filters/block.js
```diff
@@ -64,7 +70,7 @@ const ActiveFiltersBlock = ( {
 				/>
 			);
 		} );
-	}, [ productAttributes ] );
+	}, [ productAttributes, blockAttributes.displayStyle ] );
```
Aljullu, line 73: Same. :slightly_smiling_face:

assets/js/blocks/attribute-filter/block.js
```diff
@@ -168,16 +169,11 @@ const AttributeFilterBlock = ( {
...
-		},
-		// There is no need to add blockAttributes.showFilterButton as a dependency.
-		// It will only change in the editor and there we don't need to call onSubmit in any case.
-		[ checked, onSubmit ]
-	);
+	useEffect( () => {
+		if ( ! blockAttributes.showFilterButton ) {
+			onSubmit( checked );
+		}
+	}, [ blockAttributes.showFilterButton, checked, onSubmit ] );
```
Aljullu, line 176: Missing dep in effect. `blockAttributes.showFilterButton` is not expected to change during the lifecycle of the component, so it should be safe.

assets/js/blocks/attribute-filter/block.js
```diff
@@ -188,17 +184,16 @@ const AttributeFilterBlock = ( {
...
-			if ( ! isShallowEqual( checked, checkedQuery ) ) {
-				setChecked( checkedQuery );
-			}
-		},
-		// We only want to apply this effect when the query changes, so we are intentionally leaving `checked` out of the dependencies.
-		[ currentCheckedQuery ]
-	);
+	useEffect( () => {
+		if (
+			! isShallowEqual( previousCheckedQuery, currentCheckedQuery ) && // checked query changed
```
Aljullu, line 191: We only want to run this effect when `currentCheckedQuery` changes, but since we added new deps, we need to ensure the `currentCheckedQuery` is not equal to the previous one.

assets/js/blocks/attribute-filter/block.js
```diff
@@ -215,19 +210,29 @@ const AttributeFilterBlock = ( {
 		[ attributeTerms ]
 	);
 
-	const onSubmit = () => {
-		if ( isEditor ) {
-			return;
-		}
+	const onSubmit = useCallback(
```
Aljullu, line 213: Since `onSubmit` is a dependency of some effects, I wrapped it in a `useCallback()`.

assets/js/blocks/attribute-filter/edit.js
```diff
@@ -258,45 +258,42 @@ const Edit = ( { attributes, setAttributes, debouncedSpeak } ) => {
 		</Placeholder>
 	);
 
-	const onDone = useCallback( () => {
+	const onDone = () => {
```
Aljullu, line 261: I don't think it made much sense wrapping this function in a `useCallback()` because it's not used inside any other hook.

assets/js/blocks/attribute-filter/edit.js
```diff
@@ -258,45 +258,42 @@ const Edit = ( { attributes, setAttributes, debouncedSpeak } ) => {
...
 		);
-	}, [] );
+	};
 
-	const onChange = useCallback(
-		( selected ) => {
-			if ( ! selected || ! selected.length ) {
-				return;
-			}
+	const onChange = ( selected ) => {
```
Aljullu, line 271: Same as before. :slightly_smiling_face:

assets/js/blocks/price-filter/block.js
```diff
@@ -26,10 +27,12 @@ import usePriceConstraints from './use-price-constraints.js';
  */
 const PriceFilterBlock = ( { attributes, isEditor = false } ) => {
 	const [ minPriceQuery, setMinPriceQuery ] = useQueryStateByKey(
```
Aljullu, line 29: `useQueryStateByKey()` sets the default value to an empty object, I added the `null` parameter so the default value is `null`, this avoids unnecessary re-calculations when using `minPriceQuery` and `maxPriceQuery` in hook dependencies.

assets/js/blocks/price-filter/block.js
```diff
@@ -73,29 +81,61 @@ const PriceFilterBlock = ( { attributes, isEditor = false } ) => {
 				setMaxPrice( prices[ 1 ] );
 			}
 		},
-		[ minConstraint, maxConstraint, minPrice, maxPrice ]
+		[ minPrice, maxPrice, setMinPrice, setMaxPrice ]
```
Aljullu, line 84: Replacing wrong deps in this callback, this change should be safe.

assets/js/blocks/price-filter/block.js
```diff
@@ -73,29 +81,61 @@ const PriceFilterBlock = ( { attributes, isEditor = false } ) => {
...
+		maxPrice,
+		minPriceQuery,
+		maxPriceQuery,
+		minConstraint,
+		maxConstraint,
+		previousMinConstraint,
+		previousMaxConstraint,
+		previousMinPriceQuery,
+		previousMaxPriceQuery,
+	] );
```
Aljullu, line 138: This is probably the most complex change in this PR, this effect was relying on missing deps to be run only when the min/max query price or the min/max constraint changed. However, in order to fix the linting warnings we needed to add several missing dependencies. That mean the effect run more often than before, that's why I needed to add extra checks to ensure the query price or the constraints had changed.

senadir:
Excellent explanation @Aljullu, Thank you for taking the time to write it.

PR #3290, opened 2020-10-16
Escape basename paths with spaces correctly while running E2E tests.

budzanowski:
The `sh` syntax used for basename extraction was causing the wrong path to be selected in the case when the path contained spaces.

For example, the default path for sites set up with [Local](https://localwp.com/) has "Local Sites" as the main folder. This caused the the `$(basename 'pwd')` to yield `Local` and not the plugin directory.

 Run e2e tests as usual

1. for a path that has no spaces
2. for a path that has spaces

The expected behaviour is that the tests are executed correctly. This change does not affect tests results is only addresses if the tests can be run at all on some setups.

### DIFF
--- a/bin/wp-env-pre-config.sh
+++ b/bin/wp-env-pre-config.sh
@@ -1,2 +1,3 @@
 #!/bin/sh
-npm run wp-env run tests-cli './wp-content/plugins/$(basename `pwd`)/bin/wp-env-config.sh'
+BASENAME=$(basename "`pwd`")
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Nice catch! LGTM

I tested in `trunk` with a path containing a space and tests were error-ing. After checking out this branch, they worked fine. I also tested this branch with a path without spaces and tests are still working as they used to. :ok_hand:

PR #3291, opened 2020-10-17
Respect Enable Taxes setting for checkout block taxe display.

### LINKED ISSUE #3195: When tax calculations is disabled in settings, total in cart and checkout shows 'includes £0 in taxes'
## Describe the bug
When the tax calculation setting is disabled, tax is still showing under the total as 'includes £0 in taxes', which is both incorrect and shouldn't be shown. (See screenshots)

## To reproduce
Make sure 'Enable taxes' is disabled in Woocommerce settings (under General Options), add items to basket, navigate to basket and then checkout and observe the total sum.

## Expected behaviour
Totals should not include any reference to taxes.

## Screenshots
Woocommerce Settings
Total in basket
Total in checkout

* Core version: 4.5.2
* WooCommerce version: 4.5.2
* WooCommerce Blocks version: 3.4.0
* Site language:

* OS: [e.g. macOS, Windows]
* Version [e.g. 22]

* Device: [e.g. iPhone6]
* OS: [e.g. iOS8.1]
* Version [e.g. 22]

budzanowski:
The `TotalsItem` component was only checking the `Display prices during cart and checkout` tax sub-setting and was ignoring the main Tax settings switch. This PR adds the missing check.

Fixes #3195 

### Screenshots
With the following settings:
image
image
The tax component is visible - correct behaviour.
image

When the `Enable taxes` settings is switched **off** without touching any of the settings in the `Tax` tab the taxes component is no longer visible ( previously it was still visible and that was the error ):
image

1. Enable taxes
2. Enable Display prices during cart and checkout ( Tax tab )
3. Add items to the cart.
4. View the checkout block -> tax component should be visible
5. Disable taxes ( don't change anything on the Tax tab )
6. View the checkout block -> tax component should be gone

### Changelog

Respect the main tax setting for the Checkout Block tax display.

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/totals-footer-item/index.js
+++ b/assets/js/base/components/cart-checkout/totals/totals-footer-item/index.js
@@ -16,6 +16,8 @@ import PropTypes from 'prop-types';
 import TotalsItem from '../totals-item';
 import './style.scss';
 
+const SHOW_TAXES = TAXES_ENABLED && DISPLAY_CART_PRICES_INCLUDING_TAX;
+
 const TotalsFooterItem = ( { currency, values } ) => {
 	const { total_price: totalPrice, total_tax: totalTax } = values;
 
@@ -26,8 +28,7 @@ const TotalsFooterItem = ( { currency, values } ) => {
 			label={ __( 'Total', 'woo-gutenberg-products-block' ) }
 			value={ parseInt( totalPrice, 10 ) }
 			description={
-				TAXES_ENABLED &&
-				DISPLAY_CART_PRICES_INCLUDING_TAX && (
+				SHOW_TAXES && (
 					<p className="wc-block-components-totals-footer-item-tax">
 						{ createInterpolateElement(
 							__(
--- a/assets/js/base/components/cart-checkout/totals/totals-footer-item/index.js
+++ b/assets/js/base/components/cart-checkout/totals/totals-footer-item/index.js
@@ -2,7 +2,10 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { DISPLAY_CART_PRICES_INCLUDING_TAX } from '@woocommerce/block-settings';
+import {
+	TAXES_ENABLED,
+	DISPLAY_CART_PRICES_INCLUDING_TAX,
+} from '@woocommerce/block-settings';
 import { createInterpolateElement } from 'wordpress-element';
 import FormattedMonetaryAmount from '@woocommerce/base-components/formatted-monetary-amount';
 import PropTypes from 'prop-types';
@@ -23,6 +26,7 @@ const TotalsFooterItem = ( { currency, values } ) => {
 			label={ __( 'Total', 'woo-gutenberg-products-block' ) }
 			value={ parseInt( totalPrice, 10 ) }
 			description={
+				TAXES_ENABLED &&
 				DISPLAY_CART_PRICES_INCLUDING_TAX && (
 					<p className="wc-block-components-totals-footer-item-tax">
 						{ createInterpolateElement(


haszari (approved):
Code looks solid - nice clean tweak to the logic. Thanks for fixing this @budzanowski !

I tested and confirmed the fix. During testing I also spotted a minor issue when switching taxes off/on. This doesn't block the PR - I think this issue is minor and shouldn't affect many stores in normal use. Logged a separate issue - #3296.

assets/js/base/components/cart-checkout/totals/totals-footer-item/index.js
```diff
@@ -23,6 +26,7 @@ const TotalsFooterItem = ( { currency, values } ) => {
 			label={ __( 'Total', 'woo-gutenberg-products-block' ) }
 			value={ parseInt( totalPrice, 10 ) }
 			description={
+				TAXES_ENABLED &&
```
senadir, line 29: I'm a bit worried about mixing conditions in the render return. I've run into this issue before in which having 3 conditions results in 9 implicit possibilities that you need to juggle in your mind.

In this case, `TAXES_ENABLED` and `DISPLAY_CART_PRICES_INCLUDING_TAX` are related, the second has no meaning if the first is disabled.

I have two options:

We add an extra condition outside the function called `DISPLAY_TAXES` that is false if taxes are disabled and assumes the value of `DISPLAY_CART_PRICES_INCLUDING_TAX` otherwise:

```js
const SHOW_TAXES = TAXES_ENABLED && DISPLAY_CART_PRICES_INCLUDING_TAX;
```

Another option is to disable `DISPLAY_CART_PRICES_INCLUDING_TAX` at the definition level, so that it always returns false if `TAXES_ENABLED` is false.
This option can have certain side effects that we can't see, but because of they are related (controlling the option of `DISPLAY_CART_PRICES_INCLUDING_TAX` is not possible if taxes are disabled) I think it's safe to use that option.

haszari: Hmm, interesting point @senadir . Could also factor out the tax summary to a mini-component and add the logic there, use early return to make it super simple. Something like:

```js
function TaxSummaryLine( { currency, totaltqx, } ) {
  if ( ! TAXES_ENABLED || ! DISPLAY_CART_PRICES_INCLUDING_TAX ) {
    return null;
  }

  return ( 
    <p className="wc-block-components-totals-footer-item-tax">
      { createInterpolateElement(
        __(
          'Including <TaxAmount/> in taxes',
        ),
        {
          TaxAmount: (
            <FormattedMonetaryAmount
              className="wc-block-components-totals-footer-item-tax-value"
              currency={ currency }
              displayType="text"
              value={ parseInt( totalTax, 10 ) }
            />
          ),
        }
      ) }
    </p>
  );
}
```

budzanowski: We can't change the definition of `DISPLAY_CART_PRICES_INCLUDING_TAX` to depend on `TAXES_ENABLED`. Although it sounds reasonable here this would make those two values behave differently to their serverside counterparts - that is an unnecessary complication, and it has the potential to generate bugs in the future.

I do like the idea of extracting that bit of logic outside into a `SHOW_TAXES` just outside the `return` statement. I would not want to put it further - Is that correct @senadir ? That extraction makes sense also because it gives a direct name to the outcome of the logical condition.

As for extraction into a separate component - I don't know much about the architecture of the project yet to even remotely know if that is a correct approach - it feels like overkill though. Especially the name makes things a bit confusing. This tax line would be just applicable to CART and CHECKOUT and not any other place.

senadir: > We can't change the definition of DISPLAY_CART_PRICES_INCLUDING_TAX to depend on TAXES_ENABLED. Although it sounds reasonable here this would make those two values behave differently to their serverside counterparts - that is an unnecessary complication, and it has the potential to generate bugs in the future.

In any other case, I would agree with this. But for this particular case, options like should reflect the default value when disabled, as for bugs, I'm okay with having them and generating them, there should be no case when in which `TAXES_ENABLED` is `false` and `DISPLAY_CART_PRICES_INCLUDING_TAX` because it's impossible to reach that case intentionally.

However, I'm not trying to block this PR, just to shed light on how complex mixing settings can get.

> I do like the idea of extracting that bit of logic outside into a SHOW_TAXES just outside the return statement. I would not want to put it further - Is that correct @senadir ?

Yes, generally we would move it to the top level of that file, outside any function, so that it's not regenerated on each render.

> As for extraction into a separate component - I don't know much about the architecture of the project yet to even remotely know if that is a correct approach - it feels like overkill though. Especially the name makes things a bit confusing. This tax line would be just applicable to CART and CHECKOUT and not any other place.

Having it as a separate component is completely okay, we have a couple of packages for cart and checkout components only.
Cart and Checkout are huge; if we leave them in a single file, they will reach thousands of lines.

But in this case, I'm between. It will provide more readability for this component.
You can abstract the component and leave it within that file.
The general goal is to break components to the lowest possible function, this is useful for several things, like extensibility, reusability and performance.
However, I don't feel strongly about this, if usually try not to abstract unless there's a problem, so don't feel like this should block you. You can merge it anyway.

budzanowski:
@senadir I have extracted the value. Abstracting into a component can happen when there will be a more obvious need for such move.

senadir (approved):

PR #3292, opened 2020-10-17
Add a php coding standards action

nerrad:
WordPress core and the Gutenberg projects are both moving away from travis for automated CI and I like the idea of doing the same for our repo for the following reasons:

- Unified interface (no need to go to another site/service)
- Inline annotations for for linting issues on pull requests.
- GitHub actions have a growing ecosystem and provides opportunities for more automations in the future.
- The future of the Travis service is a bit unknown since they were acquired a couple years ago (there's been little to no investment in new features).

This pull adds a Github workflow for linting php. It also updates the WooCommerce phpcs sniffs package to the latest version (so there should be some php compatibility checks up to PHP 7.4 now).

## To test

The new workflow should run in this pull and I'll experiment with adding deliberate linting issues to see what happens. Also if this works well, I'll remove the PHPCS linting in the travis build.

### DIFF
--- a/.github/workflows/php-coding-standards.yml
+++ b/.github/workflows/php-coding-standards.yml
@@ -44,7 +44,7 @@ jobs:
       - name: Set up PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: '8.0'
+          php-version: '7.4'
           coverage: none
           tools: composer, cs2pr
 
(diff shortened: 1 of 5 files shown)


nerrad:
Nice! Deliberate fail resulted in the annotation:

Image 2020-10-17 at 12 03 22 PM

Aljullu (approved):
LGTM :ok_hand:

PR #3294, opened 2020-10-17
Add workflow for linting js and css

nerrad:
WordPress core and the Gutenberg projects are both moving away from travis for automated CI and I like the idea of doing the same for our repo for the following reasons:

- Unified interface (no need to go to another site/service)
- Inline annotations for for linting issues on pull requests.
- GitHub actions have a growing ecosystem and provides opportunities for more automations in the future.
- The future of the Travis service is a bit unknown since they were acquired a couple years ago (there's been little to no investment in new features).

This pull adds a Github workflow for linting js and css (inline annotations not enabled yet on css linting for this workflow).

## To test

The new workflow should run in this pull and I'll experiment with adding deliberate linting issues to see what happens. Also if this works well I'll comment out the js&css linting in the travis build.

### DIFF
--- a/.github/workflows/js-css-linting.yml
+++ b/.github/workflows/js-css-linting.yml
@@ -30,11 +30,9 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: 12.x
-
-    - name: Npm install and build
+    - name: Npm install
       run: |
         npm install
-        npm run build
     - name: Lint JavaScript
       uses: bradennapier/eslint-plus-action@v3.4.2
     - name: Lint CSS
--- a/.github/workflows/js-css-linting.yml
+++ b/.github/workflows/js-css-linting.yml
@@ -36,10 +36,7 @@ jobs:
         npm install
         npm run build
     - name: Lint JavaScript
-      uses: tinovyatkin/action-eslint@v1
-      with:
-        repo-token: ${{secrets.GITHUB_TOKEN}}
-        check-name: All
+      uses: bradennapier/eslint-plus-action@v3.4.2
     - name: Lint CSS
       run: npm run lint:css
 
--- a/.github/workflows/js-css-linting.yml
+++ b/.github/workflows/js-css-linting.yml
@@ -1,0 +1,42 @@
+name: JavaScript and CSS Linting
+
+on:
+  pull_request:
+  push:
+    branches: [trunk]
+
+jobs:
+  check:
+    name: All
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Cache node modules
+      uses: actions/cache@v2
+      env:
+        cache-name: cache-node-modules
+      with:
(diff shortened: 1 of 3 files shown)


nerrad:
Nice, so the action I went with for eslinting will leave a comment as well as add annotations in the files tab. It's configurable so if we decide the comment is unnecessary we can remove it.

.github/workflows/js-css-linting.yml
```diff
@@ -0,0 +1,42 @@
...
+          ${{ runner.os }}-
+    - name: Use Node.js 12.x
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.x
+
+    - name: Npm install and build
+      run: |
+        npm install
+        npm run build
```
budzanowski, line 37: Do we need to build? I would think that we only care about the source files and not compilation output.

.github/workflows/js-css-linting.yml
```diff
@@ -0,0 +1,42 @@
...
+          ${{ runner.os }}-build-
+          ${{ runner.os }}-
+    - name: Use Node.js 12.x
+      uses: actions/setup-node@v1
+      with:
+        node-version: 12.x
+
+    - name: Npm install and build
+      run: |
+        npm install
```
budzanowski, line 36: If we don't need to do `build` then we can skip the `install` step here because `bradennapier/eslint-plus-action` triggers install automatically if `node_modules` folder is not found.

.github/workflows/js-css-linting.yml
nerrad: Good point, we don’t need to build.

.github/workflows/js-css-linting.yml
nerrad: I left `npm install` in place because it [looks like the action will use node 14](https://github.com/bradennapier/eslint-plus-action/blob/972d934f4b5b41945c4f489dc49717c80eb8be4c/Dockerfile#L1) and there could be some problems with that version (we currently are pinned to node 12 - LTS).

budzanowski: Nice. LGTM!

budzanowski (approved):

.github/workflows/js-css-linting.yml
senadir: But if we're not building, we're doing static linting, in which node version is not needed, I don't see why the version difference would play a role here.

senadir:
This is a great addition, but is the comment needed if it's successful? Can we turn it off for that case? If the linting fails, the PR cannot be merged anyways, so it's hard to miss.

nerrad:
> but is the comment needed if it's successful? Can we turn it off for that case?

Yup we can turn off. I'll do a config change in the workflow.

.github/workflows/js-css-linting.yml
nerrad: True. For now I'd prefer to be explicit about installing with the version used for the project. The action will not do a build if `node_modules` is found so it's not really a problem to have it this way.

PR #3298, opened 2020-10-19
Correctly process $0 checkouts

senadir:
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3229

We didn't add draft order to `woocommerce_valid_order_statuses_for_payment_complete` which is used when checking an order status before processing it, this one is invoked as part of `payment_complete` which we only call when processing empty orders.
When processing a paid order, we use `woocommerce_valid_order_statuses_for_payment` which we already integrated with.
This caused empty orders ($0) to never get processed and will always stay as a draft.

1. Add a $0 product to your cart and checkout.
2. Notice the order ID, check the order in the backend, it should be processing.
3. Checkout the product again, it should create a new order, not the previous one.

### DIFF
--- a/src/Domain/Services/DraftOrders.php
+++ b/src/Domain/Services/DraftOrders.php
@@ -41,6 +41,7 @@ class DraftOrders {
 			add_filter( 'wc_order_statuses', [ $this, 'register_draft_order_status' ] );
 			add_filter( 'woocommerce_register_shop_order_post_statuses', [ $this, 'register_draft_order_post_status' ] );
 			add_filter( 'woocommerce_valid_order_statuses_for_payment', [ $this, 'append_draft_order_post_status' ] );
+			add_filter( 'woocommerce_valid_order_statuses_for_payment_complete', [ $this, 'append_draft_order_post_status' ] );
 			add_action( 'woocommerce_cleanup_draft_orders', [ $this, 'delete_expired_draft_orders' ] );
 			add_action( 'admin_init', [ $this, 'install' ] );
 		} else {


senadir:
> This caused empty orders ($0) to never get processed and will always stay as a draft.

Extra context on why this happens is because `payment_complete` will check the current order status if it's within a set of predefined status and will abort if not.
That set is filterable via `woocommerce_valid_order_statuses_for_payment_complete` and we add `checkout-draft` to it.

haszari (approved):
Tested and confirmed the fix. Looking at the source for `woocommerce_valid_order_statuses_for_payment_complete` it seems this is the correct fix -as I understand it, we are declaring that our draft orders are eligible for post-payment processing/completed status. 

I also noticed when testing to reproduce the original issue that this is data loss - before this fix, subsequent free orders would overwrite the older free order. That feels serious to me, so we should add a good release note for this and possibly do some more extensive testing of draft orders.

I'm happy to approve this but it might be useful to get another review (perhaps from proton) to get some more feedback on the draft order and potential implications. Also would be good to get some docs added to core for those hooks to clarify expectations and confirm that we're using this as intended (I think we are, just would personally like more clarity around this). Maybe something to follow up as we work towards merging the new draft status into Woo core.

senadir:
It's a data loss, but a different kind of one (actually two-fold)

Draft orders will only override the current session, meaning you will override your previous order, and another user will override his.
That's because of the nature of how we handle orders, we reach to the session.
image

The data loss happening here is due to our 24h age limit for draft orders (before we delete them).

The original data loss is that your draft order would be lost by the next order you make, meaning any line items or order notes are deleted.

Customer data is still preserved in the Customer tab (I tested).  

I don't mind highlighting this in our next release post cc @nerrad for the approach announcing this without causing FUD. I feel like this impacts a subset of websites so we should be clear on the possible damage it might have caused.

> I'm happy to approve this but it might be useful to get another review (perhaps from proton) to get some more feedback on the draft order and potential implications. Also would be good to get some docs added to core for those hooks to clarify expectations and confirm that we're using this as intended (I think we are, just would personally like more clarity around this). Maybe something to follow up as we work towards merging the new draft status into Woo core.

I fail to see what kind of action we can take from here, adding docs is great, but I think that hook is self-documenting.

[The original function](https://github.com/woocommerce/woocommerce//blob/7485b9165f68c7986ea4c31809089b4a02502fe0/includes/class-wc-order.php#L99-L144) was written by Mike, as the draft order code and payment processing part, so we can assume we're on the good here and it was just something that fell off Mike's mind.

nerrad:
> I don't mind highlighting this in our next release post cc @nerrad for the approach announcing this without causing FUD. I feel like this impacts a subset of websites so we should be clear on the possible damage it might have caused.

I didn't fully understand what was being discussed in here (and was confused about where you folks were talking about what is the current behaviour in `trunk` vs  behaviour in this branch) so I went ahead and did some testing. Here's some notes (I realize this probably replicates information already posted in this pull, but I'm adding it in the interest of providing an alternate way of describing what's happening):

### Behaviour in `trunk`.

- Add a free product to cart (resulting in $0.00 for order total).
- Submit the order in the checkout with that product.
- The order is processed but the order status remains `checkout-draft`.
- With the same session, add another product to the cart and then complete the checkout (can be paid or free doesn't matter).
- The new purchase does _not_ create a new order, but re-uses the original order from above which means details on that order are lost.

Another way in which details on the original order would be lost is when the `checkout-draft` order is deleted as a part of the draft order cleanup wp-cron job.

### Behaviour in this branch.

- Add a free product to cart ( resulting in $0.00 for order total).
- Submit the order in the checkout with that product.
- The order is processed and the order status changes to `complete`.
- With the same session, add another product to the cart and then complete the checkout.
- The new purchase _creates a new order_. 

### What should be highlighted in release post.

This fix _does_ address two different types of data loss that could occur for stores that frequently offer products for $0.00 value and I do think it's worth highlighting, however I don't think we need to make a big deal about it. I would suggest something like this should suffice in the post:

> Any stores offering products that can result in free orders (where the order total equals 0) should update to this version of the WooCommerce Blocks plugin. There was a significant bug affecting order loss for orders made with the checkout block that totalled 0. 

Nadir, you can add this to your pull description as something that can be highlighted for the next release.

PR #3299, opened 2020-10-20
Change items organisation on smaller viewports.

### LINKED ISSUE #3257: Price overflow on smaller viewports
## Describe the bug
On small (iPhone 5/SE size) devices, if a price is larger than 3 digits, it overflows the container slightly.

## To reproduce

1. Add the WooCommerce Cart block to your Cart page
2. Resize the viewport down to 320ish pixels

## Expected behavior
I think maybe the price should drop to the next line or the quantity component would have the +/- buttons removed for small viewports?

## Screenshots
https://i.imgur.com/tk1bqtn.png

* Core version: 5.5.1
* WooCommerce version: 4.5.2
* WooCommerce Blocks version: 3.5.0
* Site language: English
* Any other plugins installed: Advanced Custom Fields PRO, Custom Post Type UI, Regenerate Thumbnails

* OS: macOS
* Browser Firefox
* Version 81.0.1

budzanowski:
On smaller viewports, the quantity selector and the line item cost were organized on the right side of the view. This caused price values to overflow on smaller viewports or prices with lots of digits:
image

To mitigate I have moved the quantity button on the left side and made the price span two last available grid cells.
This way there is a lot more place for the price line. Grid alignment was changed to `end` in order to provide a gap between rows.

I think that the design sits well with how the lines below are structured. If we would something other then more significant changes would be necessary.

Fixes #3257

### Screenshots

After the changes, the UI remains consistent when the grid CSS kicks in even for smaller widths then requested in the Issue:
image

Bigger viewports are consistent:
image

1. Add items to cart with a cost that has 3 leading digits.
2. Go to a page with a cart block.
3. Reduce the viewport and check how the price behaves below 320px widht.

### Changelog

Fix cart item price overflow on smaller viewports.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
@@ -200,6 +200,7 @@ table.wc-block-cart-items {
 				vertical-align: bottom;
 				padding-right: $gap;
 				align-self: end;
+				padding-top: $gap;
 
 				.wc-block-cart-item__remove-link {
 					display: none;
@@ -212,6 +213,8 @@ table.wc-block-cart-items {
 				grid-column: 2 / span 2;
 				grid-row-start: 2;
 				align-self: end;
+				justify-self: end;
+				width: fit-content;
 
 				.wc-block-components-formatted-money-amount {
 					display: inline-block;
--- a/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
@@ -214,7 +214,7 @@ table.wc-block-cart-items {
 				grid-row-start: 2;
 				align-self: end;
 				justify-self: end;
-				width: fit-content;
+				padding-bottom: 0.375em;
 
 				.wc-block-components-formatted-money-amount {
 					display: inline-block;
--- a/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
@@ -195,10 +195,11 @@ table.wc-block-cart-items {
 				padding-bottom: $gap;
 			}
 			.wc-block-cart-item__quantity {
-				grid-column-start: 2;
+				grid-column-start: 1;
 				grid-row-start: 2;
 				vertical-align: bottom;
 				padding-right: $gap;
+				align-self: end;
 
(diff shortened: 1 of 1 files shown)


budzanowski:
@LevinMedia The quantity selector was moved for the smaller viewports to make room for the big numbers. Is this OK? Please advise :)

assets/js/blocks/cart-checkout/cart/full-cart/style.scss
```diff
@@ -208,9 +210,11 @@ table.wc-block-cart-items {
 				}
 			}
 			.wc-block-cart-item__total {
-				grid-column-start: 3;
+				grid-column: 2 / span 2;
 				grid-row-start: 2;
-				align-self: center;
+				align-self: end;
+				justify-self: end;
+				width: fit-content;
```
Aljullu, line 217: Is the width needed here? I thought inheriting it from the grid would be enough.

Aljullu (approved):
Thanks for fixing this issue @budzanowski, the cart items CSS is a bit complex and it's nice seeing it being improved. Notice I'm just reviewing it code-wise, leaving the design review to @LevinMedia.

I confirm it fixes the issue and code looks good so approving. I have a suggestion to further simplify the CSS grid, but not blocking:

Before this PR, the grid had three columns, and that didn't change with this PR:
imatge

But with the new layout, I think it could be simplified so there are only two columns: `.wc-block-cart-item__product` and `.wc-block-cart-item__total` would need to end at column 3 instead of 4. `.wc-block-cart-items__row` columns width will probably need to change from `80px 132px` to `80px 1fr`.

imatge

I think that should work and make the CSS simpler (and we get rid of a magic number: `132px`).

_Tip:_ Firefox has a [CSS grid inspector](https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Examine_grid_layouts) which makes working on CSS grid much easier.

LevinMedia:
@budzanowski looks good to me, thank you! My only feed back would be aligning the price vertically to the middle so that it aligns with the copy in the size selector. 

Screen Shot 2020-10-26 at 4 11 56 PM

assets/js/blocks/cart-checkout/cart/full-cart/style.scss
budzanowski: Nope. You are right. Removed.
Chrome also has a grid inspector. Super useful tools, thanks for sharing.
image

PR #3300, opened 2020-10-20
Use @wordpress/base-styles and @automattic/color-studio as a base for our styles

### LINKED ISSUE #1331: Settling on the correct source for global styles.
It's a bit unclear currently in the project what should be the official source for styles (specifically colors at the moment but _could_ apply to other things too probably).  Currently we have the following potential and actual sources:

* `@woocommerce/components/build-style/style.css` (currently used in [`assets/css/editor.scss`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/css/editor.scss))
* `@automattic/color-studio` (wc-admin uses this)
* WP Editor Styles ([upcoming color palette for WordPress](https://make.wordpress.org/design/2019/11/26/proposal-a-new-color-palette-for-wordpress/))
* [Figma WordPress Foundation styles](https://www.figma.com/file/P6L7I2Iyyh2tcSw7jyELqfzV/WordPress-Foundation?node-id=812%3A337) (current WordPress styles and source for current mocks?)
* What we already have in the project ([`assets/css/abstracts`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/master/assets/css/abstracts))

For the [current cycle](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/975#workspaces/blocks-5a84601e79f97c2acc45cc58/reports/release?release=5de4fecab435b73314048e58) we're going to just go with what is in the mockups and implement some style variables for those things.

However, I think we need to start thinking about consolidating around _one_ source of truth for global styles and only override in the project as needed.  I think there's some overlap with the WooCommerce Admin project as well so @woocommerce/wc-admin I'd love some thoughts from you folks as well on this.

cc @garymurray who is going to try and pin down some things from the design side in terms of color styles.

Aljullu:
Fixes #1331.

This PR:
* Replaces several SCSS variables and hardcoded CSS values with variables from `@wordpress/base-styles` and `@automattic/color-studio` in order to align our colors with Gutenberg and WooCommerce brand. Notice that means this PR introduces some visual changes, but none should be too drastic since I tried to always use the closer match. We used to have many different shades of gray that are now reduced to 7, that's in line with what Gutenberg did to reduce the number of colors (https://github.com/WordPress/gutenberg/pull/23648).
* Decreases the specificity of price slider styles, so they are easier to theme + adds [documentation](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/update/styles-source/docs/theming/all-products-and-filters.md) on how to achieve it.
* Adds a SCSS mixin (`encode-color()`) so SCSS variables can be used in SVG data-urls.

Tip: it might be easier to review this PR with the `Hide whitespace changes` option on GH.

#### Accessibility

I tested color contrast in several blocks to ensure there were no regressions.

### Screenshots
Some screenshots:
imatge
imatge
imatge
imatge

Test all blocks in editor/frontend and verify there are no evident issues with its display (layout completely broken or colors that are not correctly assigned). You can copy & paste [this gist](https://gist.github.com/Aljullu/d9c76ed510ae6973bd41dcc4184e8eb2) into the editor to automatically add all blocks. You might need to paste it with <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd> so styles are removed and the editor can transform it into blocks.

### Changelog

> Updated several colors in blocks to align them with Gutenberg colors and WooCommerce brand.


budzanowski (approved):
Wow, great job! Lots of stuff here.

PR #3314, opened 2020-10-22
Fix React hook dependency warnings in Cart & Checkout blocks + withAttributes HOC

### LINKED ISSUE #2813: Investigate extra POST request on initial cart render.
Investigate extra POST request on initial cart render.

We have an unfixed bug in our test in which the full cart triggers a POSTrequest to &#x60;wc/store/cart/update-item&#x60; causing the fetch to be called twice.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/d8e23fb680a10d086c21d067fc80ae7b6a9f28e4/assets/js/blocks/cart-checkout/cart/test/block.js#L44-L55
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in d8e23fb680a10d086c21d067fc80ae7b6a9f28e4 when #2778 was merged. cc @senadir

Aljullu:
Part of #3204.
Related to #3285.
Fixes #2813.

#### Cart and Checkout blocks
Most of the changes are related to input field validation/displaying errors + the cart item quantity input, so these are the main flows that need to be tested.

1. In the Cart block, try increasing and decreasing the quantity of a product. Verify that while calculating the new totals, the _Proceed to Checkout_ button is disabled.
2. Verify you can remove an item from the cart.
3. Go to the Checkout block with a no-registered user and with the Stripe gateway set up.
4. Try submitting the form before interacting with any input item.
5. Verify validation errors appear because the input fields were empty.
6. Interact with the input fields and selects and verify when they are invalid (required field is empty, email field is not a valid email, etc.) a validation error is shown below it. And that error disappears when the input becomes valid.
7. Interact with Stripe input fields and verify validation errors are also displayed when they are invalid.
8. Try submitting the order when inputs are invalid and verify the order is not processed.
9. Fix the invalid inputs fields and submit the order. Verify the order is processed.

#### withAttributes HOC
1. Create a page and add a Products by Attribute block.
2. Verify attributes are listed.
3. Verify you can select an attribute and clicking Done shows the products that match that attribute.
4. Verify you can edit the block (pencil icon) and the previously selected attribute is selected by default.
5. Verify you can change to another attribute.

### DIFF
--- a/assets/js/base/components/text-input/validated.js
+++ b/assets/js/base/components/text-input/validated.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useRef, useEffect } from 'react';
+import { useCallback, useRef, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { useValidationContext } from '@woocommerce/base-context';
@@ -27,6 +27,7 @@ const ValidatedTextInput = ( {
 	showError = true,
 	...rest
 } ) => {
+	const [ isPristine, setIsPristine ] = useState( true );
 	const inputRef = useRef();
 	const {
 		getValidationError,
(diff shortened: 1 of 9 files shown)


assets/js/base/components/text-input/validated.js
```diff
@@ -39,39 +40,51 @@ const ValidatedTextInput = ( {
...
+			setIsPristine( false );
 		}
-	}, [ validateOnMount ] );
+	}, [
+		focusOnMount,
+		isPristine,
+		setIsPristine,
+		validateOnMount,
+		validateInput,
+	] );
```
senadir, line 80: Why did you merge two useEffects? [the general conscious is to separate useEffects ](https://reactjs.org/docs/hooks-effect.html#tip-use-multiple-effects-to-separate-concerns)unless you have a reason?

assets/js/blocks/cart-checkout/cart/test/block.js
```diff
@@ -40,13 +40,7 @@ describe( 'Testing cart', () => {
 			screen.getByText( /Proceed to Checkout/i )
 		).toBeInTheDocument();
 
-		/**
-		 * @todo Investigate extra POST request on initial cart render.
-		 *
-		 * We have an unfixed bug in our test in which the full cart triggers a POST
-		 * request to `wc/store/cart/update-item` causing the fetch to be called twice.
-		 */
-		expect( fetchMock ).toHaveBeenCalledTimes( 2 );
```
senadir, line 49: 💃 Finally fixed!

assets/js/hocs/with-attributes.js
```diff
@@ -3,7 +3,6 @@
  */
 import { useState, useEffect } from '@wordpress/element';
 import { getAttributes, getTerms } from '@woocommerce/editor-components/utils';
-import { find } from 'lodash';
```
senadir, line 6: 💯

assets/js/hocs/with-attributes.js
```diff
@@ -30,52 +29,54 @@ const getAttributeData = ( attributeId, attributeList, matchField = 'id' ) => {
  */
 const withAttributes = ( OriginalComponent ) => {
 	return ( props ) => {
-		const { selected } = props;
-		const [ attributes, setAttributes ] = useState( [] );
+		const { selected = [] } = props;
+		const selectedSlug = selected.length ? selected[ 0 ]?.attr_slug : null;
```
senadir, line 33: In my experience, when working with an expected data endpoint (like our own REST API), using optional chaining signals a code smell.
Disregarding line 32, do we expect `selected` to be an empty array?
Also if you defaulting to an empty array in line 32, `length` would evaluate to 0, which would skip directly to null, so no optional chaining required.

senadir (approved):
This is a great PR, unlike my comment here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3285#pullrequestreview-515665314 I feel more confident merging this PR, it makes sense and no change require documentation because the diff is self-documenting

assets/js/blocks/cart-checkout/cart/test/block.js
senadir: Should probably also close this
https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2813

assets/js/base/components/text-input/validated.js
Aljullu: Good point. I undid the merge in 14c4ae8978db24f48d7e4f59412ad75f2cab5015.

assets/js/blocks/cart-checkout/cart/test/block.js
Aljullu: Ah, yes. Thanks for pointing to the issue. I assigned it to me and linked it with this PR.

assets/js/hocs/with-attributes.js
Aljullu: Yeah... good point. I don't think the optional chaining is necessary at all, probably it's a left-over from an in-progress refactor I was doing. Fixed in bca57a5afe2af6d03834e0e3bd945abba1369482.

PR #3320, opened 2020-10-26
Make compatible to Composer v2

om4csaba:
Composer [version 2.0.0](https://github.com/composer/composer/releases/tag/2.0.0) is released and [version 1.9](https://github.com/composer/installers/releases/tag/v1.9.0) of the `composer/installers` package adds support to it.

There is no issue for it yet, but people will use it and stumbled the problem eventually.

#### Accessibility

Note: There is no Accessibility indication of this pull request.

1.
2.
3.

### Changelog

- (chore) Make compatible to Composer v2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
 	"prefer-stable": true,
 	"minimum-stability": "dev",
 	"require": {
-		"composer/installers": "1.7.0",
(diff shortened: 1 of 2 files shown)


nerrad:
Thanks for the pr @om4csaba !

I unpinned the version as different fix in #3323.

PR #3322, opened 2020-10-26
Use composer V1 for php code checks.

budzanowski:
Fixes #

#### Accessibility

1.
2.
3.

### Changelog


nerrad (approved):
Looks like this fixed :) Feel free to merge once checks are green and a brief pull description is added.

nerrad:
Hey @budzanowski I'm going to close this because we can leave the workflow as is (using composer 2) with the update in #3323

PR #3331, opened 2020-10-28
Allow shoppers to sign-up for an account from the Checkout block

### LINKED ISSUE #3286: Release checkout signup feature in blocks plugin (disable isExperimental feature gate)
Checkout signup (aka create account) is currently available in dev builds only via `isExperimentalBuild` feature flag.

This feature is now ready for release in the blocks plugin alongside the cart and checkout blocks.

To be clear, cart and checkout are currently gated to feature plugin only, and are not available in the WooCommerce Blocks package included in core. The goal of this issue is that the checkout signup feature is available whereever the checkout block is available - i.e. feature plugin.

Remove `isExperimentalBuild` feature gating in various places in the code.

haszari:
Fixes #3286

This PR removes the `isExperimental` feature flag for the checkout signup feature, so it can be launched to stores using WooCommerce Blocks feature plugin. 🎉 

The feature gating is implemented as follows:

- The Checkout block is currently feature gated to feature plugin only. 
   - The Checkout block editor option `allowCreateAccount` is only displayed if store is running WooCommerce 4.7 or newer.
- The service class implements back-end parts of the flow - creating the account, new account email, and a set password flow.
   - All functionality in this class is gated to feature plugin only AND WooCommerce 4.7 or newer.
- The Checkout API also will only invoke create account if store is running WooCommerce 4.7 or newer.

To summarise:

If a store is running WooCommerce Blocks plugin, and WooCommerce 4.7 or newer, all features will be available:

- Merchant can allow customers to sign up at checkout (block setting).
- Checkout API will create accounts when requested.
- New accounts created via checkout block will use the improved new-account email and set password flow.
  - Note: new accounts via other means will continue to use existing email & flow.

If the store is running an older version of WooCommerce, or the blocks plugin is not active, then no features should be available:

- No block option to allow signups.
- Checkout API should ignore any request to create account.
- Existing core new-account email & password handling.

### Screenshots
See related previous PRs for details of what's changed and how the feature works.

#2851

#3236

woocommerce/woocommerce#27898

- Test using a release build of the plugin. Previously this feature was only available in dev builds.
- Confirm that feature is still working correctly in dev build.
- Confirm that checkout signup is available and works correctly with blocks plugin + woo 4.7+.
- Confirm that checkout signup is not available and doesn't cause any issues with woo version older than 4.7.0.

Note: When we next **update the package included in WooCommerce Core, we'll need to confirm that checkout signup is not available** (unless recent blocks plugin is active). 

### Changelog

> Feature: Allow shoppers to sign-up for an account from the Checkout block. #2851 #3236 #3331

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/edit.js
+++ b/assets/js/blocks/cart-checkout/checkout/edit.js
@@ -57,8 +57,8 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 	// Checkout signup is feature gated to WooCommerce 4.7 and newer;
 	// uses updated my-account/lost-password screen from 4.7+ for
 	// setting initial password.
-	// Also implicitly gated to feature plugin, because Checkout 
-	// block is gated to plugin.
+	// Also implicitly gated to feature plugin, because Checkout
+	// block is gated to plugin
 	const showCreateAccountOption = compareWithWooVersion( '4.7.0', '<=' );
 	return (
 		<InspectorControls>
--- a/assets/js/blocks/cart-checkout/checkout/edit.js
+++ b/assets/js/blocks/cart-checkout/checkout/edit.js
@@ -17,7 +17,6 @@ import {
 	PRIVACY_URL,
 	TERMS_URL,
 	CHECKOUT_PAGE_ID,
-	isExperimentalBuild,
 } from '@woocommerce/block-settings';
 import { compareWithWooVersion, getAdminLink } from '@woocommerce/settings';
 import { createInterpolateElement } from 'wordpress-element';
@@ -58,9 +57,9 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 	// Checkout signup is feature gated to WooCommerce 4.7 and newer;
 	// uses updated my-account/lost-password screen from 4.7+ for
 	// setting initial password.
-	// Also currently gated to dev builds only.
-	const showCreateAccountOption =
-		isExperimentalBuild() && compareWithWooVersion( '4.7.0', '<=' );
+	// Also implicitly gated to feature plugin, because Checkout 
+	// block is gated to plugin.
+	const showCreateAccountOption = compareWithWooVersion( '4.7.0', '<=' );
 	return (
 		<InspectorControls>
 			{ currentPostId !== CHECKOUT_PAGE_ID && (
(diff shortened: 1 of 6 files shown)


assets/js/blocks/cart-checkout/checkout/edit.js
```diff
@@ -58,9 +57,9 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 	// Checkout signup is feature gated to WooCommerce 4.7 and newer;
 	// uses updated my-account/lost-password screen from 4.7+ for
 	// setting initial password.
-	// Also currently gated to dev builds only.
-	const showCreateAccountOption =
-		isExperimentalBuild() && compareWithWooVersion( '4.7.0', '<=' );
+	// Also implicitly gated to feature plugin, because Checkout
+	// block is gated to plugin
```
haszari, line 61: I wanted a full stop on this comment, but eslint action told me to remove it https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3331/commits/cc4f7ed0ef9bdd501fd1c7c99b422b326c508fe7!

Aljullu (approved):
I think we need to remove `isExperimentalBuild()` from these lines:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/cart-checkout/checkout/form/contact-fields-step.js#L34
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/cart-checkout/checkout/block.js#L93

Another detail (not blocking): I saw `AddressStep` doesn't have `allowCreateAccount` in the [propTypes](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/cart-checkout/checkout/form/address-step.js#L90), it would be nice to add it to ensure it's always a bool.

Besides that, everything was working and testing fine on my end, so pre-approving. :+1: 

> Feature: Allow shoppers to sign-up for an account from the Checkout block (WooCommerce Blocks plugin only).

Do we need to state that it's a WC Blocks plugin only feature? Currently the Checkout block is also only available in the feature plugin so it sounds redundant.

assets/js/blocks/cart-checkout/checkout/edit.js
nerrad: I wonder if the lint rule got confused by the `Checkout` sentence case at the end of the line. Might be a bug in the rule.

haszari:
> I think we need to remove isExperimentalBuild() from these lines…

Thanks for the review @Aljullu - good catch. I've also added to proptypes and tweaked the title/changelog since checkout block is plugin-only.

I'll do another round of testing with release build and Woo 4.6.x and 4.7 RC to double-check there aren't any holes in the logic, then merge & ship 3.7.0 🚢

PR #3335, opened 2020-10-28
Update composer.lock to v. 2

Aljullu:
This PR updates composer.lock to one generated by Composer 2.x. Dependencies have already been updated so they work with Composer 2 in a previous PR: #3323.

We are planning to update Composer in WC Core and WC Admin targeting WC 4.8, so we need to merge this PR in WC Blocks 3.8.0. I added the appropriate milestone and added the `blocker` label, but notice this shouldn't be merged until 3.7.0 branch is forked from `trunk`.

Similar to https://github.com/woocommerce/woocommerce-admin/pull/5462:
- Verify installation works on Composer 1.x:
- `rm -rf vendor && composer install`
- Check that blocks work (no PHP Fatals missing classes, etc)
- Verify installation works on Composer 2.x:
- `composer self-update --preview`
- `rm -rf vendor && composer install`
- Check that blocks work (no PHP Fatals missing classes, etc)

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -8,16 +8,16 @@
     "packages": [
         {
             "name": "automattic/jetpack-autoloader",
-            "version": "v2.4.0",
+            "version": "v2.5.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/Automattic/jetpack-autoloader.git",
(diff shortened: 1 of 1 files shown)


nerrad:
We'll also need to make sure the entire team knows to update to `composer 2` locally after this is merged?

nerrad (approved):

PR #3343, opened 2020-10-29
Move feature flag PHP logic to a service class

### LINKED ISSUE #3211: Move feature flag/gating system into a service class
We have a feature flag system for controlling what blocks and features are available in different builds – woo core package | feature plugin | dev build.

Our PHPUnit integration tests can't easily test different states of the feature flag right now – it's effectively hard-coded. It would be useful if we could control the state of the flag for our tests so we can test all variations of behaviour.

Refactor feature gating into mockable service class:

- Create a service class dedicated to feature gating and include that as a dependency injected in Automattic\WooCommerce\Blocks\Domain\Package. That way tests can mock the feature gating class as needed.
- The above would also keep all the logic contained for any changes we need to make when it comes to how we dynamically switch the phase on demand as needed.

Spun off from @nerrad comment here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2851#discussion_r496689445

We could potentially control this as part of the test environment setup. For example, run a test pass for each value of the flag. Note that this would require tests to be aware of the value of the flag, and skip / tweak assertions as necessary.

Aljullu:
Fixes #3211.

1. Run `npm run start` and go to the editor.
2. Verify you can add the Product element blocks (like Product Price) and the Cart and Checkout blocks.
3. Run `npm run build:deploy` and go back to the editor.
4. Verify you can't add the Product element blocks but you can still add the Cart and Checkout blocks.

### DIFF
--- a/src/Domain/Package.php
+++ b/src/Domain/Package.php
@@ -81,4 +81,22 @@ class Package {
 		$feature_gating = NewPackage::container()->get( FeatureGating::class );
 		return $feature_gating;
 	}
+
+	/**
+	 * Checks if we're executing the code in an experimental build mode.
+	 *
+	 * @return boolean
+	 */
+	public static function is_experimental_build() {
+		return self::feature()->is_experimental_build();
+	}
+
+	/**
+	 * Checks if we're executing the code in an feature plugin or experimental build mode.
+	 *
+	 * @return boolean
+	 */
+	public static function is_feature_plugin_build() {
+		return self::feature()->is_feature_plugin_build();
+	}
 }
--- a/src/Domain/Package.php
+++ b/src/Domain/Package.php
@@ -1,6 +1,9 @@
 <?php
 namespace Automattic\WooCommerce\Blocks\Domain;
 
+use Automattic\WooCommerce\Blocks\Package as NewPackage;
+use Automattic\WooCommerce\Blocks\Domain\Services\FeatureGating;
+
 /**
  * Main package class.
  *
@@ -70,20 +73,12 @@ class Package {
 	}
 
 	/**
-	 * Checks if we're executing the code in an experimental build mode.
+	 * Returns an instance of the the FeatureGating class.
 	 *
-	 * @return boolean
+	 * @return FeatureGating
 	 */
-	public static function is_experimental_build() {
-		return WOOCOMMERCE_BLOCKS_PHASE > 2;
-	}
-
-	/**
-	 * Checks if we're executing the code in an feature plugin or experimental build mode.
-	 *
-	 * @return boolean
-	 */
-	public static function is_feature_plugin_build() {
-		return WOOCOMMERCE_BLOCKS_PHASE > 1;
+	public function feature() {
+		$feature_gating = NewPackage::container()->get( FeatureGating::class );
+		return $feature_gating;
 	}
 }
--- a/src/Domain/Package.php
+++ b/src/Domain/Package.php
@@ -28,14 +28,23 @@ class Package {
 	private $path;
 
 	/**
+	 * Holds the feature gating class instance.
+	 *
+	 * @var FeatureGating
+	 */
+	private $feature_gating;
+
+	/**
 	 * Constructor
 	 *
-	 * @param string $version      Version of the plugin.
-	 * @param string $plugin_path  Path to the main plugin file.
+	 * @param string        $version        Version of the plugin.
+	 * @param string        $plugin_path    Path to the main plugin file.
+	 * @param FeatureGating $feature_gating Feature gating class instance.
 	 */
-	public function __construct( $version, $plugin_path ) {
-		$this->version = $version;
-		$this->path    = $plugin_path;
+	public function __construct( $version, $plugin_path, FeatureGating $feature_gating ) {
+		$this->version        = $version;
+		$this->path           = $plugin_path;
+		$this->feature_gating = $feature_gating;
 	}
 
 	/**
@@ -78,8 +87,7 @@ class Package {
 	 * @return FeatureGating
 	 */
 	public function feature() {
-		$feature_gating = NewPackage::container()->get( FeatureGating::class );
-		return $feature_gating;
+		return $this->feature_gating;
 	}
 
 	/**
--- a/src/Domain/Package.php
+++ b/src/Domain/Package.php
@@ -95,8 +95,8 @@ class Package {
 	 *
 	 * @return boolean
 	 */
-	public static function is_experimental_build() {
-		return self::feature()->is_experimental_build();
+	public function is_experimental_build() {
+		return $this->feature()->is_experimental_build();
 	}
 
 	/**
@@ -104,7 +104,7 @@ class Package {
 	 *
 	 * @return boolean
 	 */
-	public static function is_feature_plugin_build() {
-		return self::feature()->is_feature_plugin_build();
+	public function is_feature_plugin_build() {
+		return $this->feature()->is_feature_plugin_build();
 	}
 }
--- a/src/Domain/Services/FeatureGating.php
+++ b/src/Domain/Services/FeatureGating.php
@@ -1,0 +1,75 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\Domain\Services;
+
+/**
+ * Service class that handles the feature flags.
+ *
+ * @internal
+ */
+class FeatureGating {
+
+	/**
+	 * Current flag value.
+	 *
+	 * @var int
+	 */
+	private $flag;
+
+	const EXPERIMENTAL_FLAG   = 3;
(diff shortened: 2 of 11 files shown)


src/Domain/Services/FeatureGating.php
```diff
@@ -0,0 +1,62 @@
...
+	private static $flag;
+	const EXPERIMENTAL_FLAG   = 3;
+	const FEATURE_PLUGIN_FLAG = 2;
+	const CORE_FLAG           = 1;
+
+	/**
+	 * Set correct flag.
+	 */
+	public function init() {
+		$default_flag  = defined( 'WC_BLOCKS_IS_FEATURE_PLUGIN' ) ? 2 : 1;
```
budzanowski:
```suggestion
		$default_flag  = defined( 'WC_BLOCKS_IS_FEATURE_PLUGIN' ) ? self::FEATURE_PLUGIN_FLAG : self::CORE_FLAG;
```

budzanowski (approved):
I have left a small suggestion but otherwise, this looks great!

src/Domain/Services/FeatureGating.php
Aljullu: Good catch!

Aljullu:
Thanks for the review @budzanowski! I'm not merging it yet because I need to update the tests so they pass.

senadir (approved):
Should also update any docs we have (in P2 or in this repo).
Can be a follow-up.

Aljullu:
> Should also update any docs we have (in P2 or in this repo).

@senadir is there any specific document that you think that needs to be updated? I think all docs are still valid after the changes in this pull but I might be missing some.

src/Domain/Services/FeatureGating.php
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	/**
+	 * Returns the current flag value.
+	 *
+	 * @return int
+	 */
+	public static function get_flag() {
+		return self::$flag;
+	}
```
nerrad: With the class being instantiated (and shared), I'd prefer if these were non static methods. The current implementation means that the static method can be called by the class even without it being initialized.

nerrad (changes requested):
Thanks for the great start here Albert. However, I'd like to see a few changes (besides what was noted in the inline comment):

- Ideally the feature gating class should not be passed around or called directly but instead be a part of the `Automattic\WooCommerce\Blocks\Domain\Package` class.
- For the above, you can inject the FeatureGating service class instance into the `Package` class and then the `Automattic\WooCommerce\Blocks\Domain\Package` instance can return the `FeatureGating` instance via a getter (eg. `$package_instance->feature()->is_experimental_build()`) .
- `src/Package` would expose public thin wrappers around the `Automattic\WooCommerce\blocks\Domain\Package::feature` method for back-compat (ideally we need to move away from using `src/Package` anywhere but just what's needed for package inclusion in core - but that's for another issue).
- `FeatureGating` should have a constructor and its signature should be something like this:

```php
public function __construct( $flag = 0 ) {}
```

By default the `$flag` value isn't passed in and that should trigger the existing `init()` method to execute. However, this allows tests to override the flag for testing various feature flag conditions.

So essentially anywhere `Package` is implemented, the feature gating methods are available via `$package_instance->feature()`.

The idea here is that the `Package` service is the authority for any "meta" information about the plugin (or in the case when the plugin is included in WooCommerce core, the package - hence the name of the class). FeatureGating is extracted to a service for a couple reasons:

- It allows for mocking during tests (or testing various WooCommerce phase states).
- It allows for encapsulated changes to the feature gating process in the future.

Yet, `FeatureGating` is still passed into Package as DI as the methods exposed on it are in themselves "meta" utilities for the plugin/package. This also ensures any place `Package` is injected to a class via DI, feature gating utilities are available if/when needed without having to have the FeatureGating service injected separately.

senadir:
The only document I can find is this https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e1345b3693413ded6d6e5e16382d41c468f17f5c/docs/blocks/feature-flags-and-experimental-interfaces.md

Which doesn't go over *how* we gate certain things, so it seems we don't have that information in our repo.

Aljullu:
Thanks for the review @nerrad! I updated this PR with the changes proposed, I left a comment below about something that might be a code smell, but I couldn't figure out how to solve, so any feedback is welcome. :slightly_smiling_face:

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -19,9 +20,9 @@
  * @since $VID:$
  */
 class CreateAccount extends WP_UnitTestCase {
-
 	private function get_test_instance() {
-		return new TestedCreateAccount( new Package( 'test', './' ) );
+		Package::init();
```
Aljullu: I needed to manually init `Package` in order to get this test to pass, otherwise I was getting:

`Exception: Cannot construct an instance of Automattic\WooCommerce\Blocks\Domain\Services\FeatureGating because it has not been registered.`

I saw there was a discussion between @nerrad and @senadir earlier in Slack about a similar error, but I couldn't figure out how to solve it in this specific case. My idea was to inject `FeatureGating` into `Package` in order to avoid [getting it directly from the container](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3343/files#diff-deccd9a660832b41efc10cb868f4549a32152b0e3753555c179b84a4733ee7cbR81), but that doesn't seem to be possible because `Package` is constructed before `FeatureGating` is registered. Any clues?

src/Domain/Package.php
```diff
@@ -69,13 +72,23 @@ public function get_url( $relative_url = '' ) {
...
 
+	/**
+	 * Returns an instance of the the FeatureGating class.
+	 *
+	 * @return FeatureGating
+	 */
+	public function feature() {
+		$feature_gating = NewPackage::container()->get( FeatureGating::class );
+		return $feature_gating;
+	}
```
nerrad: We should be injecting `FeatureGating` into the package constructor and then assign to a property. Something like this:

```php
public function __construct( $version, $plugin_path, FeatureGating $feature_gating ) {
    $this->version = $version;
    $this->path = $plugin_path;
    $this->feature_gating = $feature_gating;
}
```

I realize this changes the signature of the class which is problematic for any extensions that might be using this class directly, but imo they shouldn't be using this class directly and should be retrieving via the container (which is the benefit of using a DIC). Likelihood of other extensions using this class directly is fairly slim.

Implementing this will mean a number of changes in our test suite where `Package` is being instantiated directly so that it uses the new signature (which incidentally is also how the feature gating can be mocked).

src/Domain/Package.php
```diff
@@ -84,6 +97,6 @@ public static function is_experimental_build() {
 	 * @return boolean
 	 */
 	public static function is_feature_plugin_build() {
-		return WOOCOMMERCE_BLOCKS_PHASE > 1;
+		return self::feature()->is_feature_plugin_build();
 	}
```
nerrad, line 109: `is_experimental_build` and `is_feature_plugin_build` should not be static methods (I realize this wasn't introduced in this pr).

tests/php/Domain/Services/CreateAccount.php
nerrad: Once you implement my request above to inject `FeaturePlugin` you'd be able to implement this via:

```
return new TestedCreateAccount( new Package( 'test', './', new FeaturePlugin( 3 ) );
```

That's one of the reasons for injecting dependencies because then mocking is much easier. As noted in the earlier comment, anywhere `Package` is instantiated directly will have to be updated to implement the new signature. Another benefit of this is that we are explicitly aware when testing what gating is in effect for the test.

nerrad (changes requested):
Refactor is coming along, thanks Albert, there's still some changes I think need made though (as noted inline).

src/Domain/Package.php
nerrad: As a confidence check, I did a [quick search](https://wpdirectory.net/search/01EPYG4PJVCSGETPG0JNE3HP4Y) to see if any plugins in wp.org are invoking `Automattic\WooCommerce\Blocks\Domain\Package` directly and don't see any results.

Aljullu:
Thanks for the review @nerrad! This is ready for another look. In fb5f160f17b55a6c1f41c670c08e2c34686f8176, I injected the FeatureGating class instance in the Package constructor, `new FeatureGating()` was the bit I was missing, I was trying to get it from the container and thus it didn't work. Thanks for your help! :slightly_smiling_face: 

In 41360466ee9ab725c65f957b09f3c13f345edbac, I made `is_experimental_build()` and `is_feature_plugin_build()` methods non static.

nerrad (approved):
Noice! LGTM - great work getting this across the finish line!

I haven't tested this but code wise things look good.

PR #3356, opened 2020-11-02
Introduce ExtendRestApi

### LINKED ISSUE #3244: Integration Tracking (WC Subscriptions): Add filters to our REST endpoints

senadir:
Part of https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3244

This PR introduces `ExtendRestApi` a class that 3PD should use to extend the StoreAPI response to add extra data.
The goal of using this instead of filters directly is:
- Protecting the integrity of the original checkout, so that people don't break existing flows.
- Protecting other 3PD from having their code being broken by a rogue plugin.
- Protecting core experience from breaking if a plugin malfunction.
- Ensuring any integration is well validated and has the correct data.

For this, we changed how AbstractSchema work to be aware of `ExtendRestApi`.

### Interface.
3PD should call a `register_endpoint_data` function, that adds correct schema and data to each endpoint.

### Signature 

```php
$ExtendRestApi->register_endpoint_data( string $namespace, string $endpoint, callable $schema_callback, callable $data_callback );
```

$schema_callback is a function that should return an array with metadata of new endpoint being registered:

```php
function mySchemaFunction() {
  return [
		'fee' => [
			'description' => __( 'X Product fees', 'my-plugin' ),
			'type'        => 'integer',
			'readonly'    => true,
		],
	];
}
```

$data_callback is a function that should return an array with actual data being registered:

```php
function myDataFunction( $cart_item ) {
	$data = []
	if ( $cart_item['type'] === 'x-product' ) {
		$myXProduct = new MyCustomProduct( $cart_item['item'] );
		$data = [
			'fee' => $myXProduct->get_fee();
		]
	}
  return $data;
} 
```

### Current endpoints that can be extended.
Currently, we only have `cart/items` that can be extended, it passes `$cart_item` to `$data_callback`.

### Code example

```php
use Automattic\WooCommerce\Blocks\StoreApi\Schemas\CartItemSchema;

add_action('init', function() {

		CartItemSchema::IDENTIFIER,
		"subscriptions",
		function () {
			return [
				'key' => [
					'type' => 'string',
				]
			];
		},
		function ( $cart_item ) {
			return [
				"key" => $cart_item['key']
			];
		},
	);
});
```

### How to test this change:

- Add the code example above in any plugin (even the main file of this plugin).
- In cart, inspect `store/cart` response.
- Under `items` you should see `extensions` and inside it `subscriptions` containing the data you added.
- Modify the code so that it causes an error, instead of `'key' => $cart_item['key']` try `'key' => new SomeClass`
- Visit cart again, `extensions` shouldn't include `subscriptions` anymore, an error would be logged.
- If you have debug enabled and logged in as an admin, you should an error being thrown.

### DIFF
--- a/src/Domain/Services/ExtendRestApi.php
+++ b/src/Domain/Services/ExtendRestApi.php
@@ -1,0 +1,131 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\Domain\Services;
+
+use Exception;
+use Automattic\WooCommerce\Blocks\Domain\Package;
+
+/**
+ * Service class to provide utility functions to extend REST API.
+ */
+class ExtendRestApi {
+	/**
+	 * Holds the Package instance
+	 *
+	 * @var Package
+	 */
+	private $package;
+
+	/**
+	 * Valid endpoints to extend
+	 *
+	 * @var array
+	 */
+	private $endpoints = [ 'item' ];
+
+	/**
+	 * Init - register handlers for internal hooks.
+	 */
+	public function init() {
+
+	}
+	/**
+	 * An endpoint that validates registration method call
+	 *
+	 * @param string   $endpoint The endpoint to extend.
+	 * @param string   $namespace Plugin namespace.
+	 * @param callable $schema_callback Callback executed to add schema data.
+	 * @param callable $data_callback Callback executed to add endpoint data.
+	 */
+	private function validate_endpoint_data( $endpoint, $namespace, $schema_callback, $data_callback ) {
+		if ( ! is_string( $namespace ) ) {
+			$this->throw_exception(
+				'You must provide a plugin namespace when extending a Store REST endpoint.'
+			);
+		}
+
+		if ( ! is_string( $endpoint ) || ! in_array( $endpoint, $this->endpoints, true ) ) {
+			$this->throw_exception(
+				'You must provide a valid Store REST endpoint to extend, valid endpoints are: ' . implode( ', ', $this->endpoints )
+			);
+		}
+
+		if ( ! is_callable( $schema_callback ) ) {
+			$this->throw_exception(
+				'$schema_callback must be a callable function.'
+			);
+		}
+
+		if ( ! is_callable( $data_callback ) ) {
+			$this->throw_exception(
+				'$data_callback must be a callable function.'
+			);
+		}
+	}
+
+	/**
+	 * Extends the items in store/cart endpoint
+	 *
+	 * @param string   $namespace Plugin namespace.
+	 * @param callable $schema_callback Callback executed to add schema data.
+	 * @param callable $data_callback Callback executed to add endpoint data, gets cart item data as param.
+	 */
+	public function register_cart_item_endpoint_data( $namespace, $schema_callback, $data_callback ) {
+		$this->validate_endpoint_data( 'item', $namespace, $schema_callback, $data_callback );
+
+		add_filter( '__internal_extend_cart_item_schema', function( $schema ) use ( $schema_callback, $namespace ) {
+			$schema_data = [];
+
+			try {
+				$schema_data = $schema_callback();
+			} catch ( Exception $e ) {
+				$this->throw_exception( $e );
+				return $schema;
+			}
+
+			if ( ! is_array( $schema_data ) ) {
+				$this->throw_exception(
+					'$schema_callback must return an array.'
+				);
+			}
+
+			return $schema[ $namespace ] = $schema_data;
+		}, 10, 1 );
+
+		add_filter( '__internal_extend_cart_item_data', function( $data, $cart_item ) use ( $data_callback, $namespace ) {
+			$endpoint_data = [];
+
+			try {
+				$endpoint_data = $data_callback( $cart_item );
+			} catch ( Exception $e ) {
+				$this->throw_exception( $e );
+				return $data;
+			}
+
+			if ( ! is_array( $endpoint_data ) ) {
+				$this->throw_exception( '$endpoint_data must return an array.' );
+			}
+
+			return $data[ $namespace ] = $endpoint_data;
+		},
+		10, 2 );
+	}
+
+	/**
+	 * Throws error or silently logs it.
+	 *
+	 * @param string|Exception $exception_or_error Error message or Exception.
+	 */
+	private function throw_exception( $exception_or_error ) {
+		if ( $exception_or_error instanceof Exception ) {
+			$exception = $exception_or_error;
+		} else {
+			$exception = Exception( $exception_or_error );
+		}
+
+		if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+			throw $exception;
+		} else {
+			wc_caught_exception( $exception_or_error );
+		}
+	}
+}
--- a/src/Domain/Services/ExtendRestApi.php
+++ b/src/Domain/Services/ExtendRestApi.php
@@ -71,9 +71,10 @@ class ExtendRestApi {
 				}
 
 				if ( ! is_array( $schema_data ) ) {
-					return $this->throw_exception(
+					$this->throw_exception(
 						'$schema_callback must return an array.'
 					);
+					return $schema;
 				}
 
 				$schema[ $namespace ] = $schema_data;
@@ -96,7 +97,8 @@ class ExtendRestApi {
 				}
 
 				if ( ! is_array( $endpoint_data ) ) {
-					return $this->throw_exception( '$endpoint_data must return an array.' );
+					$this->throw_exception( '$endpoint_data must return an array.' );
+					return $data;
 				}
 
 				$data[ $namespace ] = $endpoint_data;
--- a/src/Domain/Services/ExtendRestApi.php
+++ b/src/Domain/Services/ExtendRestApi.php
@@ -25,6 +25,7 @@ class ExtendRestApi {
 	public function __construct( Package $package ) {
 		$this->package = $package;
 	}
+
 	/**
 	 * Valid endpoints to extend
 	 *
@@ -52,28 +53,21 @@ class ExtendRestApi {
 	 */
 	public function register_endpoint_data( $endpoint, $namespace, $schema_callback, $data_callback ) {
 		if ( ! is_string( $namespace ) ) {
-			$this->throw_exception(
-				__( 'You must provide a plugin namespace when extending a Store REST endpoint.', 'woo-gutenberg-products-block' )
-			);
+			$this->throw_exception( 'You must provide a plugin namespace when extending a Store REST endpoint.' );
 		}
 
 		if ( ! is_string( $endpoint ) || ! in_array( $endpoint, $this->endpoints, true ) ) {
 			$this->throw_exception(
-				/* translators: 1: a list of endpoints. 2: endpoint provided by user. */
-				sprintf( __( 'You must provide a valid Store REST endpoint to extend, valid endpoints are: %1$s. You provided %2$s.', 'woo-gutenberg-products-block' ), implode( ', ', $this->endpoints ), $endpoint )
+				sprintf( 'You must provide a valid Store REST endpoint to extend, valid endpoints are: %1$s. You provided %2$s.', implode( ', ', $this->endpoints ), $endpoint )
 			);
 		}
 
 		if ( ! is_callable( $schema_callback ) ) {
-			$this->throw_exception(
-				__( '$schema_callback must be a callable function.', 'woo-gutenberg-products-block' )
-			);
+			$this->throw_exception( '$schema_callback must be a callable function.' );
 		}
 
 		if ( ! is_callable( $data_callback ) ) {
-			$this->throw_exception(
-				__( '$data_callback must be a callable function.', 'woo-gutenberg-products-block' )
-			);
+			$this->throw_exception( '$data_callback must be a callable function.' );
 		}
 
 		$this->extend_data[ $endpoint ][ $namespace ] = [
@@ -89,11 +83,11 @@ class ExtendRestApi {
 	 *
 	 * @param string $endpoint    A valid identifier.
 	 * @param array  $passed_args Passed arguments from the Schema class.
-	 * @return array Returns an array with registered endpoint data.
+	 * @return object Returns an casted object with registered endpoint data.
 	 * @throws Exception If a registered callback throws an error, or silently logs it.
 	 */
 	public function get_endpoint_data( $endpoint, array $passed_args = [] ) {
-		$registered_data = (object) [];
+		$registered_data = [];
 		if ( ! isset( $this->extend_data[ $endpoint ] ) ) {
 			return $registered_data;
 		}
@@ -102,19 +96,18 @@ class ExtendRestApi {
 
 			try {
 				$data = $callbacks['data_callback']( ...$passed_args );
+
+				if ( ! is_array( $data ) ) {
+					throw new Exception( '$data_callback must return an array.' );
+				}
 			} catch ( Throwable $e ) {
 				$this->throw_exception( $e );
 				continue;
 			}
 
-			if ( ! is_array( $data ) ) {
-				$this->throw_exception( __( '$data_callback must return an array.', 'woo-gutenberg-products-block' ) );
-				continue;
-			}
-
 			$registered_data[ $namespace ] = $data;
 		}
-		return $registered_data;
+		return (object) $registered_data;
 	}
 
 	/**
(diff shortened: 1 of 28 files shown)


src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Service class to provide utility functions to extend REST API.
+ */
+class ExtendRestApi {
+	/**
+	 * Valid endpoints to extend
+	 *
+	 * @var array
+	 */
+	private $endpoints = [ 'item' ];
```
senadir: Each time we add a new integration point, we would add it here.
Integration points should have a slug, this would keep them controlled and would allow us to easily depreciate them.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,129 @@
...
+				'$schema_callback must be a callable function.'
+			);
+		}
+
+		if ( ! is_callable( $data_callback ) ) {
+			$this->throw_exception(
+				'$data_callback must be a callable function.'
+			);
+		}
+	}
```
senadir: This function is going to be used internally by other extend methods, it does the namespace verification, the endpoint verification (not needed if we're going to create wrappers for each endpoint), and callback verification.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,129 @@
...
+
+	/**
+	 * Extends the items in store/cart endpoint
+	 *
+	 * @param string   $namespace Plugin namespace.
+	 * @param callable $schema_callback Callback executed to add schema data.
+	 * @param callable $data_callback Callback executed to add endpoint data, gets cart item data as param.
+	 */
+	public function register_cart_item_endpoint_data( $namespace, $schema_callback, $data_callback ) {
```
senadir: `register_cart_item_endpoint_data` is the first function we're going to implement, future ones are going to be things like `register_shipping_endpoint_data` `register_product_endpoint_data` `register_cart_endpoint_data` `register_checkout_endpoint_data`

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,129 @@
...
+		$this->validate_endpoint_data( 'item', $namespace, $schema_callback, $data_callback );
+
+		add_filter(
+			'__internal_extend_cart_item_schema',
+			function( $schema ) use ( $schema_callback, $namespace ) {
+				$schema_data = [];
+
+				try {
+					$schema_data = $schema_callback();
```
senadir: We execute the callback in try/catch to capture any possible errors from third-party integration code.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,129 @@
...
+				} catch ( Throwable $e ) {
+					$this->throw_exception( $e );
+					return $schema;
```
senadir: If we catch an error, we return the initial data, so ignoring this plugin completely.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,131 @@
...
+				}
+
+				if ( ! is_array( $schema_data ) ) {
+					$this->throw_exception(
+						'$schema_callback must return an array.'
+					);
+					return $schema;
+				}
```
senadir: Callbacks must return an array of data, so we check if that's the case or not.
This means we're forcing 3PD to always provide some sort of data and schema, we make adding schema or data skippable in the future like `extend_rest_route`

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,131 @@
...
+
+				$schema[ $namespace ] = $schema_data;
```
senadir: Each endpoint data is going to be added to that namespace.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,131 @@
...
+			1
+		);
+
+		add_filter(
+			'__internal_extend_cart_item_data',
+			function( $data, $cart_item ) use ( $data_callback, $namespace ) {
+				$endpoint_data = [];
+
+				try {
+					$endpoint_data = $data_callback( $cart_item );
```
senadir: We pass back any values we want from the filter back to 3PD.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,131 @@
...
+			$exception = Exception( $exception_or_error );
+		}
+
+		if ( defined( 'WP_DEBUG' ) && WP_DEBUG && current_user_can( 'manage_woocommerce' ) ) {
+			throw $exception;
+		} else {
+			wc_caught_exception( $exception_or_error );
+		}
+	}
+}
```
senadir: We should only throw an exception if debugging is enabled and the current user is an admin.

src/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -266,6 +266,13 @@ public function get_properties() {
 					]
 				),
 			],
+			'extensions'           => [
+				'description' => __( 'Extra data attached by extensions', 'woo-gutenberg-products-block' ),
+				'type'        => 'array',
+				'context'     => [ 'view' ],
+				'readonly'    => true,
+				'properties'  => apply_filters( '__internal_extend_cart_item_schema', [] ),
```
senadir: 3PD are also responsible for adding correct schema for each extension they have.

src/Domain/Services/ExtendRestApi.php
nerrad: What does `'item'` refer to here and how did you derive the name `item`?

src/Domain/Services/ExtendRestApi.php
nerrad: Instead of creating individual method names for each registration point, what about having something like:

```php
public function register_endpoint_data( $identifier, $namespace, $schema_callback, $data_callback ) { /**/ }
```

`$identifier` will be something like `cart_item`, `shipping` etc and that would be a part of the private `$valid_identifiers` property which can be used for validation:

```php
private $valid_identifiers = [ 'cart-items', 'shipping' ];
```

As far as identifiers, I would standardize around the ones used in the [`RoutesController`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/edda1fa24e4f0437737d8f738b9dce4b9505cc31/src/StoreApi/RoutesController.php#L69-L93) (similar to what is also used in the [`SchemaController`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/StoreApi/SchemaController.php).

As an aside, these identifiers should ideally be constants to reduce potential for human error and also extensions can use the constants. I would like to see the constants added in the concrete classes. So for example [`Automattic\WooCommerce\Blocks\StoreApi\Routes\CartItems`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/StoreApi/Routes/CartItems.php) would have this defined in the class:

```php
const IDENTIFIER = 'cart-items'
```

Then in [`RoutesController`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/edda1fa24e4f0437737d8f738b9dce4b9505cc31/src/StoreApi/RoutesController.php#L68-L94) you'd end up with something like this for each item in the `routes` array:

```php
$this->routes = [
     Routes\CartItems::IDENTIFIER => new Routes\CartItems( $this->schemas->get( Schemas\CartItemSchema::IDENTIFIER ) ),
];
```
(as you can tell you'd do something similar for schemas).

Then extensions can reference the identifiers without risk of human error (and IDE's will appropriately enhance references etc) by the class constant when registering their extra data via the new interface.

This class constant implementation could be done in a PR and merged separately from this pull request and brought in to it after merge (which would make the review easier).

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,131 @@
...
+	 *
+	 * @param string   $namespace Plugin namespace.
+	 * @param callable $schema_callback Callback executed to add schema data.
+	 * @param callable $data_callback Callback executed to add endpoint data, gets cart item data as param.
+	 */
+	public function register_cart_item_endpoint_data( $namespace, $schema_callback, $data_callback ) {
+		$this->validate_endpoint_data( 'item', $namespace, $schema_callback, $data_callback );
+
+		add_filter(
+			'__internal_extend_cart_item_schema',
```
nerrad: Instead of a filter, I would suggest `AbstractRoute` and `AbstractSchema` receive an instance of `ExtendRestApi` via dependency injection. Then you can expose all the registered endpoint data via a getter so the route/schema can be enhanced at runtime by what is registered. Eliminating the filter ensures that the only way for extensions to enhance the endpoints is via the API.

In going this approach you'd also have to register `ExtendRestApi` with the container in `Automattic\WooCommerce\Blocks\Domain\Bootstrap` so that extensions can get the single instance of it for registering with.

src/Domain/Services/ExtendRestApi.php
nerrad: When you implement what I suggested earlier about registering to an identifier, you would have something like:

```php
$schema[ $identifier ][ $namespace ] = $schema_data;
```

Then the getter that `AbstractRoute` or `AbstractSchema` would implement from `ExentdRestApi` should be able to retrieve all the registered extra data via the identifier for the concrete route/schema.

src/Domain/Services/ExtendRestApi.php
nerrad: Let's leave as is for now but this seems like something that would be useful as a helper for other conditions - @mikejolley do you recall if there is already a helper for this kind of thing in Woo Core (or in Blocks) for this kind of condition (specifically the condition for `WP_DEBUG` etc).

nerrad (dismissed):
Good start Nadir, I have some inline feedback that I think will help improve the implementation. I don't think we need to implement filters here (even if they are marked `_internal`) and we should avoid to force going through the new interface. I also would like to see us use class constants for various identifiers as explained in the inline comments.

src/Domain/Services/ExtendRestApi.php
senadir: It works like the suggested `$valid_identifiers`, I derived it from the endpoint name, but I think I can use the route name you pointed to.

src/Domain/Services/ExtendRestApi.php
senadir: > Instead of creating individual method names for each registration point

That was my original goal, and that's how `validate_endpoint_data` used to work (it was named `register_endpoint_data`), but because I was using filters, there's the limitation of passed arguments.
So Cart Item filter might pass 2 arguments, and shipping might pass just 1 or 3.
If you don't specify that value in the `add_filter` call, you would get a fatal error, so each endpoint wrapper was needed.
If we move away from filters, I will try implementing this.

src/Domain/Services/ExtendRestApi.php
mikejolley: No, core uses `wc_caught_exception` to log something but fails without a visible Exception regardless of user type.

I think regardless of how we handle the exception above, we should still log it via `wc_caught_exception` as well.

senadir:
@nerrad or @mikejolley this is ready for another code review. if we're okay on the new API, I will update the PR body with new examples.

senadir:
I merged Darren's PR into this after fixing linting issues.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,193 @@
...
+	private $package;
+
+	/**
+	 * Constructor
+	 *
+	 * @param Package $package An instance of the package class.
+	 */
+	public function __construct( Package $package ) {
+		$this->package = $package;
+	}
```
mikejolley, line 27: nit: linebreak

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,193 @@
...
+	public function register_endpoint_data( $endpoint, $namespace, $schema_callback, $data_callback ) {
+		if ( ! is_string( $namespace ) ) {
+			$this->throw_exception(
+				__( 'You must provide a plugin namespace when extending a Store REST endpoint.', 'woo-gutenberg-products-block' )
+			);
+		}
+
+		if ( ! is_string( $endpoint ) || ! in_array( $endpoint, $this->endpoints, true ) ) {
+			$this->throw_exception(
+				/* translators: 1: a list of endpoints. 2: endpoint provided by user. */
```
mikejolley: Is this correct syntax for translators comment? I see `/* Translators: %1$s ...` used in rest of codebase.

mikejolley: Should we be translating error messages? They are not customer facing. I see other examples of non-translated exceptions in the codebase.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,193 @@
...
+			$schema = [];
+
+			try {
+				$schema = $callbacks['schema_callback']( ...$passed_args );
+			} catch ( Throwable $e ) {
+				$this->throw_exception( $e );
+				continue;
+			}
+
+			if ( ! is_array( $schema ) ) {
```
mikejolley: I'm not sure if this will be cleaner so feel free to ignore, but you could do the array check within the try/catch block so throw_exception is called once in this method.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,193 @@
...
+	 * Format schema for an extension.
+	 *
+	 * @param string $namespace Error message or Exception.
+	 * @param array  $schema An error to throw if we have debug enabled and user is admin.
+	 *
+	 * @return array Formatted schema.
+	 */
+	private function format_extensions_properties( $namespace, $schema ) {
+		return [
+			/* translators: extension namespace */
```
mikejolley: Possibly missing %s?

src/StoreApi/Schemas/CartCouponSchema.php
```diff
@@ -2,6 +2,8 @@
 namespace Automattic\WooCommerce\Blocks\StoreApi\Schemas;
 
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\CartController;
+use Automattic\WooCommerce\Blocks\Domain\Services\ExtendRestApi;
```
mikejolley: Is this actually used within this class?

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,193 @@
...
+	/**
+	 * Returns the registered endpoint data
+	 *
+	 * @param string $endpoint    A valid identifier.
+	 * @param array  $passed_args Passed arguments from the Schema class.
+	 * @return array Returns an array with registered endpoint data.
+	 * @throws Exception If a registered callback throws an error, or silently logs it.
+	 */
+	public function get_endpoint_data( $endpoint, array $passed_args = [] ) {
+		$registered_data = (object) [];
```
mikejolley: This produces a fatal error in testing.

This should be left as an array until returned because on line 115 you are adding to the array.

So swap this to `return (object) $registered_data;` and update the docblock, assuming objects are expected where this method is called.

mikejolley (changes requested):
Reviewed the code but there is a bug in the object casting. Left a note.

src/Domain/Services/ExtendRestApi.php
senadir: I see this, the one you provided, and ones without any indication in our codebase, so I wasn't sure what pattern is this.

senadir:
Addressed feedback, ready for another review.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -0,0 +1,185 @@
...
+	 *
+	 * @param string $endpoint    A valid identifier.
+	 * @param array  $passed_args Passed arguments from the Schema class.
+	 * @return object Returns an casted object with registered endpoint data.
+	 * @throws Exception If a registered callback throws an error, or silently logs it.
+	 */
+	public function get_endpoint_data( $endpoint, array $passed_args = [] ) {
+		$registered_data = [];
+		if ( ! isset( $this->extend_data[ $endpoint ] ) ) {
+			return $registered_data;
```
mikejolley: Still missing an object cast on this line.

mikejolley (approved):
Looks better now. I have tested. I am preapproving but you are still missing an object cast (left a comment). Thanks.

src/Domain/Services/ExtendRestApi.php
senadir: yeah, I noticed that. Fixed.

senadir:
@nerrad giving that we went back and forth when implementing this and the PR you did (that we merged to this), I'm going to assume that you're okay with the current state of code so I'm going to dismiss from Github UI.

PR #3358, opened 2020-11-03
Sync shipping address with billing address when shipping address fields are disabled

### LINKED ISSUE #3341: Taxes won't react to changes from billing form
Reported here https://wordpress.org/support/topic/taxes-wont-update-when-province-changes

When you have different taxes for different addresses, the rates and values won't update when you change the address.
They would update if:

- You have shipping enabled.
- You have "Use shipping for billing" checked.

This means that any changes in the billing step are not accounted for, the shopper can either checkout with no taxes, or with outdated taxes.

### Steps to reproduce

- Disable shipping.
- Setup some taxes rates, different ones.
- On checkout block, observe your current taxes, it can be zero or an initial sum if you already have an address saved.
- Change your address to one with a different tax.
- Taxes won't change.
- Checkout, taxes are still the same and didn't update with your address.
- See the order, it has the new address you entered.

mikejolley:
### Background

Taxes in WC are calculated based on _shipping address_.

During checkout, we conditionally show shipping address fields based on a "cartNeedsShipping" constant, which is derived from products being sold and shipping settings. If shipping fields are not shown, we usually just show billing fields only.

Whilst WooCommerce core does something similar on checkout, it keeps the shipping address synced with billing address so that taxes still update as needed.

In the case of blocks, when shipping address fields are hidden, we're not syncing or updating billing address via the API, which results in incorrect taxes being displayed (#3341) and taxes not being recalculated when the order is placed.

In fact, if you test this using the instructions in #3341 you'll notice that if you go ahead and place the order, the billing address is saved, but the shipping address uses whatever was in the session previously, or is left blank if checking out as a guest. The taxes are shown based on the initial shipping address as you entered checkout.

### Solution

To fix this we need to:

- Keep shipping and billing addresses synced in the client and server
- Send addresses to the server and return updated carts

This PR introduces `CustomerDataProvider ` to store billing and shipping addresses and pass down via context.

To sync data to the server, a new hook named `useCustomerData` stores the addresses and pushes to the API via a new `update-customer` endpoint. This replaces the `update-shipping` endpoint. This is debounced to prevent multiple requests in quick sucession.

Fixes #3341

**Case 1**
1. Create 2 tax rates; one for UK and one for US
2. Delete all shipping methods and rates so shipping is disabled and no fields are shown during checkout
3. Go to checkout. Enter a UK address. See the tax rate.
4. Change country field. See the tax rates update after the API request completes.
5. Place order. Check the order has the same shipping and billing address saved.

**Case 2**
1. In general settings, in the option for where you ship to, select 'disable shipping calculations'
2. Go to checkout. Enter a UK address. See the tax rate.
3. Change country field. See the tax rates update after the API request completes.
4. Place order. Check the order has the same shipping and billing address saved.

**Case 3**
1. In tax settings, choose to base taxes on the billing address.
2. Go to checkout. Enter a different billing and shipping address.
3. Toggle the "ship to billing" checkbox and see if the rate changes after doing so reliably.

In all cases you should see a network request to `update-customer` a short while after entering any billing or shipping data, or toggling the 'use shipping address' checkbox on checkout.

### Changelog

> Sync shipping address with billing address when shipping address fields are disabled, and update billing address on the server using a new `update-customer` REST API endpoint.


senadir:
Instead of using the testing instructions above, I disabled shipping from `Settings -> General -> General Options -> Shipping locations` and I can't proceed, any change in Checkout would trigger an error `Shipping is disabled`

mikejolley:
@senadir Isn't that an unrelated issue? It's a bug but it's not what was logged in #3341

senadir:
image
image

When going over the testing instructions, I have two taxes for Alberta and British Columbia, switching between them doesn't update taxes, I can see the network request for `update-shipping` but taxes are still empty.

Using classic checkout, I can see the taxes being applied
image

senadir:
@mikejolley Well that's how I filled the original issue (disabled shipping), there are several ways to disable shipping, this PR only fixes one of them, but the original issue would stay the same.

senadir:
If you want to handle the backend in a follow-up, that's okay, but should #3341 should stay open (and possibly reformated).

senadir:
The above issue (taxes not updating) happens when I have taxes based on billing address
image

mikejolley:
3 in 1 then :) yay

mikejolley:
@senadir This still needs work but can you take a quick peek? I think this is the way we need to solve it; send the billing data to the server when fields change, because there are cases where it's used for taxes (not just shipping).

senadir:
Doing a functionality review before digging into code.

senadir:
So the summary is:
- When shipping is disabled, this works fine ✅
- When shipping is enabled but there are no shipping methods, this works fine ✅
- When shipping is enabled and there are shipping methods:
    - If you have taxes calculated based on the **shipping address:**
        - Updating shipping form will update taxes ✅
        - Updating billing form would produce an error `No address country code was given` 🔴
        - Having `Use same address for billing` would update taxes but still produce the above error 🔴.   
    - If you have taxes calculated based on the **billing address:**
        - Updating billing form would produce an error `No address country code was given` 🔴
        - Having `Use same address for billing` would **not** update taxes but still produce the above error 🔴.   

I hope that made sense?

senadir:
Well I see that you added a38ef62 that should solve them, testing again.

mikejolley:
Thanks @senadir I noticed those issues too and resolved.

Do you think we should update the stores conditionally to avoid more api requests? It would be looking at quite a few core settings to determine if billing needs to be pushed to the server or not.

assets/js/base/context/cart-checkout/billing/index.js
```diff
@@ -26,8 +46,15 @@ export const useBillingDataContext = () => {
...
 
+/**
+ * Billing Data context provider.
+ *
+ * @param {Object}  props          Incoming props for the provider.
+ * @param {Object}  props.children The children being wrapped.
+ */
 export const BillingDataProvider = ( { children } ) => {
-	const [ billingData, setBillingDataState ] = useState( DEFAULT_STATE );
+	const { billingAddress: initalBillingData } = useStoreCart();
```
mikejolley: Added to store cart and API for consistency with shipping_address

assets/js/base/context/cart-checkout/cart/index.js
```diff
@@ -22,6 +23,7 @@ export const CartProvider = ( { children, redirectUrl } ) => {
 			<ShippingDataProvider>
 				<PaymentMethodDataProvider>
 					{ children }
+					<CartProcessor />
```
mikejolley: Similar pattern to checkout in that it consumes contexts further up the tree. In this case it listens for shipping and billing address updates and dispatches updates to the API on change.

assets/js/base/context/cart-checkout/cart/processor/index.js
```diff
@@ -0,0 +1,56 @@
...
+	const { updateCustomerAddress } = useDispatch( storeKey );
+	const previousShippingAddress = useRef( shippingAddress );
+	const [ debouncedShippingAddress ] = useDebounce( shippingAddress, 400 );
+
+	// When the billing or shipping address changes we need to push the changes to the server to
+	// get an updated cart--things such as taxes may be affected. This will push both billing and
+	// shipping addresses to the server and get an updated cart in a single request.
+	useEffect( () => {
+		if (
+			shouldUpdateAddressStore(
```
mikejolley: If the address is changed, and has a `country`, we dispatch the update to the server. This is instead of the `useShippingAddress` hook which was only consumed by these contexts.

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -18,33 +19,13 @@ import {
...
- * Utility function for preparing payment data for the request.
- *
- * @param {Object}  paymentData          Arbitrary payment data provided by the payment method.
- * @param {boolean} shouldSave           Whether to save the payment method info to user account.
- * @param {Object}  activePaymentMethod  The current active payment method.
- *
- * @return {PaymentDataItem[]} Returns the payment data as an array of
- *                             PaymentDataItem objects.
- */
-const preparePaymentData = ( paymentData, shouldSave, activePaymentMethod ) => {
```
mikejolley, line 36: Moved to utils to keep this file smaller.

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -262,6 +206,94 @@ const CheckoutProcessor = () => {
...
+	] );
+
+	// Keep refs up to date.
+	useEffect( () => {
+		currentBillingData.current = billingData;
+		currentShippingAddress.current = shippingAddress;
+		currentRedirectUrl.current = redirectUrl;
+	}, [ billingData, shippingAddress, redirectUrl ] );
+
+	// When the billing or shipping address changes we need to push the changes to the server to
```
mikejolley: This monitors for address changes to both billing and shipping so they are updated together in a single request.

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -262,6 +206,94 @@ const CheckoutProcessor = () => {
...
+		) {
+			previousBillingData.current = currentBillingData.current;
+			previousShippingAddress.current = currentShippingAddress.current;
+
+			updateCustomerAddress( {
+				shipping_address: currentShippingAddress.current,
+				billing_address: {
+					...currentBillingData.current,
+					phone: currentBillingData.current.phone || null,
+					email: currentBillingData.current.email || null,
```
mikejolley: Using null prevents API validation for valid email addresses.

src/StoreApi/Routes/Checkout.php
```diff
@@ -123,8 +123,10 @@ protected function get_route_response( \WP_REST_Request $request ) {
 	 * @return \WP_REST_Response
```
mikejolley, line 123: The changes within this class ensure that totals reflect customer data posted via the form, so if the API updates don't run on the checkout itself, the correct taxes etc are still calculated.

nerrad (commented):
I haven't done an exhaustive review of this. In general I think you've identified a couple flaws with the current architectural implementation of Contexts in the code:

- Woo Core has a tight coupling between shipping and billing address with regards to representing _customer data_. 
- Woo Blocks was trying to keep these as discrete data sets that would get synced in key places.

What we've realized is that billing data and shipping data must _always_ be in sync regardless of the state of the UI/UX. 

We've already implemented a number of fixes due to this flawed architectural assumption (which is largely my fault) so I think we need to rethink this a bit.

Should we align around the new `CartUpdateAddress` route (and endpoint) for all address updates or is there still need for the separate endpoints? I'm wary about having more than one endpoint for updating discrete data (and there seems to be overlap between `CartUpdateAddress`, and `CartUpdateShipping`. `Routes\Checkout` also seems to have some overlap in that it updates addresses too.). It'd be nice to have a canonical endpoint for anything customer address related server side that all requests go through (which removes the guess work around what address endpoint to use).

Also, rather than an `AddressProcessor` what if we synchronize around this new endpoint for anything "customer address" related (which includes shipping and billing address data) and create a context specifically for _customer_data_? This would replace the Billing Data context and we'd then keep the shipping context just for handling the various updates related to shipping rates etc from shipping address updates.

> Note: I'm flexible on naming, `<CustomerAddressProvider />` would align more with the endpoint name primarily implemented in it (via `wp.data` store), but I'm also wondering if for extensibility it'd be good to make it a bit more generic and use it to represent anything related to customer data?

So the tree would be (for checkout):

```js
<CheckoutStateProvider>
    <CustomerDataProvider>
      <ShippingDataProvider>
        <PaymentMethodDataProvider>
          {children}
          <CheckoutProcessor/>
        </PaymentMethodDataProvider>
      </ShippingDataProvider>
    </CustomerDataProvider>
</CheckoutStateProvider>
```

The concern I have with the current approach is it appears to increase the complexity of the code (both in terms of number of endpoints and locations where address information is updated/retrieved) a bit more than I think is necessary and feels like another bandaid for the larger architectural problem (a clever bandaid at that 👍 ). I think what we've identified here is a need to keep customer data together (out of necessity with the WooCommerce data model), instead of splitting it into discrete data sets.

I realize, this would result in a more significant refactor. For payment methods at least, I think we'll still be able to preserve much of the public api exposed via `usePaymentMethodInterface` so that shouldn't be affected too much (beyond changing some of the data sources). However, I think we'll be able to simplify a lot of logic and flow as a result of the refactor too (and hopefully eliminate some code).

mikejolley:
This is ready for another review.

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -66,8 +44,8 @@ const CheckoutProcessor = () => {
 		shouldCreateAccount,
 	} = useCheckoutContext();
 	const { hasValidationErrors } = useValidationContext();
-	const { shippingAddress, shippingErrorStatus } = useShippingDataContext();
-	const { billingData } = useBillingDataContext();
+	const { shippingErrorStatus } = useShippingDataContext();
+	const { billingData, shippingAddress } = useCustomerDataContext();
```
nerrad, line 48: Gotta say, I like this import better. Seems to make more sense to have these values together.

assets/js/base/context/cart-checkout/checkout/processor/utils.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @typedef {import('@woocommerce/type-defs/payments').PaymentDataItem} PaymentDataItem
+ */
+
+/**
+ * Utility function for preparing payment data for the request.
+ *
+ * @param {Object}  paymentData          Arbitrary payment data provided by the payment method.
+ * @param {boolean} shouldSave           Whether to save the payment method info to user account.
+ * @param {Object}  activePaymentMethod  The current active payment method.
```
nerrad, line 10: Ideally we'd have `paymentData` and `activePaymentMethod` defined as typedefs so we are explicit about what is passed through here.

assets/js/base/hooks/cart/use-store-cart.js
```diff
@@ -9,6 +9,35 @@ import { useEditorContext } from '@woocommerce/base-context';
...
+	first_name: '',
+	last_name: '',
+	company: '',
+	address_1: '',
+	address_2: '',
+	city: '',
+	state: '',
+	postcode: '',
+	country: '',
+};
```
nerrad:
```suggestion
const defaultShippingAddress = {
	first_name: '',
	last_name: '',
	company: '',
	address_1: '',
	address_2: '',
	city: '',
	state: '',
	postcode: '',
	country: '',
};

const defaultBillingAddress = {
	...defaultShippingAddress,
	email: '',
	phone: '',
};
```

assets/js/base/hooks/customer/use-customer-data.js
```diff
@@ -0,0 +1,102 @@
...
+		billingAddress: initialBillingData,
+		shippingAddress: initialShippingAddress,
+	} = useStoreCart();
+
+	const [ customerData, setCustomerData ] = useState( {
+		billingData: initialBillingData,
+		shippingAddress: initialShippingAddress,
+	} );
+	const previousCustomerData = useRef( customerData );
+	const [ debouncedCustomerData ] = useDebounce( customerData, 400 );
```
nerrad: It looks like you can simplify things a bit by [passing in the equality function](https://github.com/xnimorz/use-debounce/blob/6f86d15fb4b112e9a9fb86ad90f29ca98f5332e2/src/useDebounce.ts#L11) to `useDebounce`. This would allow you to remove the comparison logic for `shouldUpdateBilling` and `shouldUpdateShipping` out of the below `useEffect` and also eliminate the extra `useRef` in the `useCustomerData` hook.

So something like:

```js
const [ debouncedCustomerData ] = useDebounce( 
  customerData,
  400,
  { 
    equalityFn: ( prevData, newData ) => {
      return shouldUpdateAddressStore( prevData.billingData, newData.billingData ) || shouldUpdateAddressStore( prevData.shippingAddress, newData.shippingAddress )
    }
  }
)
```

I realize the one downside to this approach is you won't know explicitly whether shipping address or billing address changed - but I'm not sure I understand why we care? Can't we just always send both the current debounced shipping address value and debounced billing address value in a request when there's a change in either (and let the server handle what addresses actually had a change)?

nerrad (commented):
In general looks great Mike! I like the refactor and I think it makes the logic more straightforward having Shipping and Billing data live together.

I do have some feedback for the `useCustomerData` hook that might be applicable.

I have not tested yet, but am happy to assist with that once the feedback is responded to/addressed. I'm mostly wanting to make sure express payment methods still work with the changes (they use the shipping address apis for getting rates etc). From what I've seen in the code I don't think our changes here should impact that at all but still good to verify.

assets/js/base/hooks/customer/use-customer-data.js
mikejolley: Yes but in the action that we dispatch we're setting some things to true when shipping rates are resolving. If we are only updating billing address that logic isn't needed. That could impact performance?

nerrad: Hmm, I see, and `shippingRatesAreResolving` will in turn mean the ui/ux for shipping rates will be showing a loading indicator everytime _billing address_ is changed right?

So it does look it's better to keep things as they are for now.

assets/js/base/context/cart-checkout/checkout/processor/utils.js
mikejolley: As this is unrelated to this PR I won't add it here, but I did notice this already has a todo: ```@todo do typedefs for the payment event state.``` in payment-methods/constants.js

mikejolley:
I think this is working now — tested with Stripe Chrome Pay. Would appreciate some help testing various cases :)

assets/js/base/hooks/customer/use-customer-data.js
mikejolley: I went ahead with this feedback and replaced the shipping rate resolving code with customer data resolving.

assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -46,55 +43,54 @@ export const useCheckoutAddress = () => {
...
-		}
-	}, [
-		billingFields,
-		shippingAsBilling,
-		billingData,
-		shippingAddress,
-		setBillingData,
-		setShippingAddress,
-	] );
+	const currentShippingAsBilling = useRef( shippingAsBilling );
```
nerrad, line 46: I'm not sure about this variable name. Is it really current as of the execution time? We typically use the `previous` prefix for `useRef` references (which is how it appears to be used in the logic here).

assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -46,55 +43,54 @@ export const useCheckoutAddress = () => {
...
+		if ( currentShippingAsBilling.current !== shippingAsBilling ) {
+			if ( shippingAsBilling ) {
+				previousBillingData.current = billingData;
+				setBillingData( shippingAddress );
+			} else {
+				setBillingData( {
+					...previousBillingData.current,
+					email: undefined,
+					phone: undefined,
+				} );
```
nerrad, line 76: Won't this clear out previous set email and phone number if the `shippingAsBilling` checkbox is unchecked? Do we want that behaviour?

assets/js/base/hooks/checkout/use-checkout-address.js
```diff
@@ -46,55 +43,54 @@ export const useCheckoutAddress = () => {
...
+			}
+		},
+		[ shippingAsBilling, setShippingAddress, setBillingData ]
 	);
 
-	const setEmail = ( value ) => void setBillingFields( { email: value } );
-	const setPhone = ( value ) => void setBillingFields( { phone: value } );
+	// This syncs billing data with shipping data if the checkbox is enabled.
+	useEffect( () => {
+		if ( currentShippingAsBilling.current !== shippingAsBilling ) {
```
nerrad, line 67: So just for clarity sake - I was originally wondering why `shippingAsBilling` was being tracked as a ref but it looks like why it's implemented this way is because:

- the `billingData` dependency could trigger this effect frequently (because it's an object that changes).
- tracking previous state of `shippingAsBilling` allows for escaping the execution of this logic and only doing so when the `shippingAsBilling` state changes.

The potential problem here is that we still want billing data to update when shipping data changes when `shippingAsBilling` is true but this problem is resolved  by the `setShippingFields` callback.

Is the above correct?

assets/js/base/hooks/customer/use-customer-data.js
```diff
@@ -19,73 +20,94 @@ export const useCustomerData = () => {
 	const { updateCustomerData } = useDispatch( storeKey );
 	const { addErrorNotice, removeNotice } = useStoreNotices();
 	const {
-		billingAddress: initialBillingData,
-		shippingAddress: initialShippingAddress,
+		billingAddress: cartBillingData,
```
nerrad, line 23: I'm getting a typedef warning here for property `billingAddress` not existing for type `StoreCart`.

assets/js/base/hooks/customer/use-customer-data.js
```diff
@@ -19,73 +20,94 @@ export const useCustomerData = () => {
...
+	useEffect( () => {
+		if (
+			! isShallowEqual(
+				currentShippingAddress.current,
+				cartShippingAddress
+			)
+		) {
+			currentShippingAddress.current = cartShippingAddress;
+		}
+	}, [ cartShippingAddress ] );
```
nerrad: Could the `currentShippingAddress` and `currentBillingAddress` still be in a single ref object and merge these two above effects so you'd end up with something like this?

```js
useEffect( () => {
  if (
    ! isShallowEqual(
      currentCustomerData.current.shippingAddress,
      cartShippingAddress
    )
  ) {
    currentCustomerData.current.shippingAddress = cartShippingAddress;
  }
  if (
    ! isShallowEqual(
      currentCustomerData.current.billingData,
      cartBillingData
    )
  ) {
    currentCustomerData.current.billingData = cartBillingData;
  }
}, [ cartShippingAddress, cartBillingData ] );
```

assets/js/base/hooks/customer/use-customer-data.js
```diff
@@ -19,73 +20,94 @@ export const useCustomerData = () => {
...
+			! (
+				shouldUpdateAddressStore(
+					currentBillingData.current,
+					debouncedCustomerData.billingData
+				) ||
+				shouldUpdateAddressStore(
+					currentShippingAddress.current,
+					debouncedCustomerData.shippingAddress
+				)
+			)
```
nerrad: Is this still needed if the `debouncedCustomerData` is already doing the equality check? Could you just rely on only updating customer data if `debouncedCustomerData` changes (which would be a dependency change)?

src/StoreApi/Routes/CartUpdateCustomer.php
```diff
@@ -135,24 +137,17 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
...
-			throw new RouteException(
-				'woocommerce_rest_cart_invalid_country',
-				sprintf(
-					/* translators: 1: valid country codes */
-					__( 'Address country is missing. Please provide one of the following: %s', 'woo-gutenberg-products-block' ),
-					implode( ', ', array_keys( $allowed_countries ) )
-				),
-				400
-			);
+			return [];
```
nerrad: Can you provide some more explanation on why this was changed (I see the phpdoc but it's still not clear to me why this was changed from an exception to returning an empty array).

nerrad (commented):
Changes are looking good, I still have some feedback though. Also I haven't tested yet because the branch is out of date with master (and there's some conflicts with rebasing). Ping me when it's rebased and ready for testing and I'll take a look.

assets/js/base/hooks/checkout/use-checkout-address.js
mikejolley: I dont think so--we merge billing data, so existing emails are preserved. I'll double check tommorow.

src/StoreApi/Routes/CartUpdateCustomer.php
mikejolley: Express payments don't always return a country code I found in testing. Additionally (and similarly) sometimes a billing email is provided but not the full address. This resulted in error 400.

At cart level, having a full complete address is not that important. We just want to track as much data as possible up until an order is created. This matches session behaviour in core. 

I'll update the doc block.

assets/js/base/hooks/checkout/use-checkout-address.js
mikejolley: This is current as in the context of `useCheckoutAddress` — if you see the return value from this hook it uses the `currentShippingAsBilling ` ref rather than `shippingAsBilling `. The reason I had to do it this way was because our billing address fields contain debounce code on the values, so if you return `shippingAsBilling `, those fields are rendered with the old values before this hook has had a chance to sync billing data.

I will add a comment to that effect if it helps.

assets/js/base/hooks/checkout/use-checkout-address.js
mikejolley: Confirmed this is a merge update.

assets/js/base/hooks/checkout/use-checkout-address.js
mikejolley: Right. This effect basically only handles the toggle event; when the checkbox is toggled we need to immediately update billing address to reflect this change. I will add a comment.

assets/js/base/hooks/customer/use-customer-data.js
mikejolley: I think you need to refresh your IDE; it's definitely defined.

assets/js/base/hooks/customer/use-customer-data.js
mikejolley: I'm going to apply this feedback, however, I'm not sure what is really best practice or more performant. The above useEffect will be triggered more often than just the billing or shipping one, so maybe `isShallowEqual` will run more often? On the otherhand, this would make it more consistent with other data in this hook so doing it 👍🏻

assets/js/base/hooks/customer/use-customer-data.js
nerrad: ahh okay, maybe it was due to me switching branches or something then.

assets/js/base/hooks/customer/use-customer-data.js
mikejolley: Yes this is needed. 

`debouncedCustomerData ` stores all customer data debounced, using shallow equals. 

`shouldUpdateAddressStore` on the other-hand only looks at critical address fields (country, state, postcode) which need pushing to the server. 

Removing this would result in more unnecessary server requests.

mikejolley:
Feedback is addressed. Waiting on automated tests.

src/StoreApi/Routes/CartUpdateCustomer.php
```diff
@@ -145,7 +145,8 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
 	 * @return array
 	 */
 	protected function prepare_address_fields( $address, $allowed_countries ) {
-		// Addresses require a country, otherwise are not valid--return an empty address.
+		// Addresses require a country, otherwise they are not valid. This return an empty address. There is no address
```
nerrad:
```suggestion
		// Addresses require a country, otherwise they are not valid. This returns an empty address. There is no address
```

nerrad (approved):
Thanks for addressing the feedback. I didn't get a chance to test but code looks good (aside from a minor typo that could be fixed). Pre-approving, your judgment call if you'd like someone else to do a confidence check on testing the changes.

PR #3361, opened 2020-11-03
introduce SlotFill for Sidebar

### LINKED ISSUE #3246: Integration Tracking (WC Subscriptions): Create SlotFill integration points for SidebarItems
In order to inject the recurring total [referenced in this issue](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3240), we will need certain slotFills in that place position, either as an extra item in the Cart total (like Subtotal and shipping) or a whole dedicated section (like Coupons and Total).

The same inject point should be available in Checkout.

Before we decide on this, we need to spec out some designs for it.

senadir:
This PR Implements a Slot that is rendered in both Cart and Checkout, just after `Totals` section.

This slot wrap each fill in an error boundary, so that if a plugin code breaks, it would not affect the block or any other extension hooking into that slot.

### Implmentation
```jsx
import { ExperimentalOrderMeta } from '@woocommerce/blocks-checkout';

const MyApp = () => {
  return (
    <ExperimentalOrderMeta>My stuff here</ExperimentalOrderMeta />
  );
}

registerPlugin( 'my-app', { render: MyApp } );
```

Closes #3246

### Testing instructions 
- The easiest way to test this is to checkout this branch 3940-gh-woocommerce/woocommerce-subscriptions
- In Cart or Checkout, you should see `This is my default export`
- Now to test error boundary, in the above plugin `assets/src/js/index.js` or anywhere in our plugin, add this code
```jsx
const myRender = () => {
	return (
		<ExperimentalOrderMeta>
			<div className="block-div-class">This is my other export</div>
		</ExperimentalOrderMeta>
	);
};

registerPlugin( 'woocommerce-two-three', {
	render: myRender,
} );
```
- Modify the original code so it throws a **runtime** error.
```jsx
const Bug = () => {
	throw Error( 'bug' );
};
const render = () => {
	const [ showBug, setShowBug ] = useState( false );
	return (
		<ExperimentalOrderMeta>
			<div
				className="block-div-class"
				onClick={ () => setShowBug( true ) }
			>
				This is my default export
			</div>
			{ showBug && <Byg /> }
		</ExperimentalOrderMeta>
	);
};
```
- On cart or Checkout, click on the text to trigger the error.
- Only that component is removed from the page, the rest should still be fine


assets/js/base/components/cart-checkout/order-review/index.js
```diff
@@ -0,0 +1,35 @@
...
+			className={ classnames(
+				className,
+				'wc-block-components-order-reviews'
+			) }
+		>
+			{ ( fills ) => {
+				fills.map( ( fill, i ) => (
+					<BlockErrorBoundary showErrorMessage={ false } key={ i }>
+						{ fill }
+					</BlockErrorBoundary>
```
senadir: The goal is that this would catch an error but disregard the message and simply not render it.
We can extend this to only show the error if the current user is an admin.

assets/js/base/components/cart-checkout/order-review/index.js
```diff
@@ -0,0 +1,35 @@
...
+	return <Fill>{ children }</Fill>;
+}
+
+function OrderReviewSlot( { className } ) {
+	return (
+		<Slot
+			bubblesVirtually
+			className={ classnames(
+				className,
+				'wc-block-components-order-reviews'
```
mikejolley: Did you mean 'wc-block-components-order-review'?

mikejolley (commented):
This implementation looks correct but I wanted to ask if we need to be mindful of slot placement with regards to the slots name. i.e. this one is called OrderReview, but where around the order review is it? Before? After? Within?

nerrad:
Besides Mike's feedback (which I agree with) should we make this an `ExperimentalOrderReview` component until we verify that it works in whatever implementation is exposed for plugins?

Also, should we add tests for this?

senadir:
I'm having second thoughts about this and I'm not sure yet if we're going to use this or not for the sidebar.
Taking a look at the latest designs that David shared with me, we would need to add 3-5 new slots in the sidebar alone.
This doesn't seem to be very scalable.
A possible approach here might be some kind of filter, that exposes abstracted core components while allowing 3PD to insert new components.

I have a lot of considerations for this approach, but the goal is to expose the least amount of implementation detail as possible while still giving the flexibility to reorder and insert new things.

```js
addFilter('woocommerce/blocks/sidebar', ( Sidebar ) => {
	return (
		<Sidebar>
			<MyView />
			<Sidebar.ShippingSummary />
			<UseShippingForAll />
			<Sidebar.ShippingSelector />
			<Sidebar.Coupon />
			<Sidebar.Taxes />
			<SubscriptionRecurringTotals />
			<Sidebar.Totals />
			<Sidebar.Button />
		</Sidebar>
	);
})
```

However, I'm not sure how will this solution cascade to several plugins, each doing its own ordering that relies on the fact that something comes after `Taxes` or `Totals`.

If not this, we might need to expose `preComponent` and `postComponent` for each component above (so like 12 slotFill).

This is just an exploration of the problem and I might post it in a P2 for more in-depth dusscestion.

nerrad:
Sounds good, I moved back to the `In Progress` pipeline.

senadir:
I'm revisiting this and addressing some feedback.
I'm going to hold on adding tests until we finalize if we want to continue with this approach or not.

senadir:
Updated this to be named `ExprimentalOrderMeta`, we can figure out other positions as we move.

assets/js/base/components/cart-checkout/order-meta/index.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * External dependencies
+ */
+import { createSlotFill } from 'wordpress-components';
```
senadir: Now that plugin area is merged, we can test this.
I suspect it won't work because of this custom `wordpress-components` therefore we end up with a different context.
I will try to do some testing.

senadir: So I tested this and I was right, it won't work unless we export `SlotFillProvider` from our repo.

src/Assets.php
```diff
@@ -95,7 +95,7 @@ public static function register_assets() {
 		}
 
 		if ( Package::feature()->is_feature_plugin_build() ) {
-			$asset_api->register_script( 'wc-blocks-checkout', 'build/wc-blocks-checkout.js', [], false );
+			$asset_api->register_script( 'wc-blocks-checkout', 'build/blocks-checkout.js', [], false );
```
mikejolley, line 98: wc-blocks-checkout/wc-checkout-block is going to be confusing. Can we be more specific in our naming here somehow?

mikejolley (approved):
Tested using your instructions and all is well. I am approving this and the subscriptions PR 👍🏻

src/Assets.php
senadir: we discussed [naming here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3613), we're trying to avoid conflict with core checkout script and our block checkout, if you have any idea of a good name we can implement it anytime because we still didn't publish this.

PR #3371, opened 2020-11-04
Fix account creation bugs

### LINKED ISSUE #3362: Store API checkout route should only create an account if store allows it
## Describe the bug
The checkout block, and the Store API support creating a user account when processing a checkout.

WooCommerce has a setting and a hook that can be used to disable signup at checkout. The checkout API should check these before creating accounts, and only do so if the store has checkout signup enabled.

- `woocommerce_enable_signup_and_login_from_checkout` option (checkbox in UI)
- `woocommerce_checkout_registration_enabled` hook

## To reproduce
1. In `WooCommerce > Settings > Accounts & Privacy`, uncheck `Allow customers to create an account during checkout` and save settings.
  - OR implement `woocommerce_checkout_registration_enabled` hook in a custom plugin and return false.
2. Add a checkout block to a published page with `Allow shoppers to sign up for a user account during checkout` option enabled.
3. On front end, logged out (or incognito) add something to cart and proceed to checkout.
4. Complete checkout form and select `Create an account`.
5. Submit.

## Expected behavior
The store setting or hook should probably prevent / disable checkout signup. Some combination of:

- When Store API is processing checkout, account should only be created if store allows it.
- The `Create an account` checkbox should only be shown if the store allows signup, and the editor UI for the block option should be dependent on the store setting (e.g. hide if disabled, or show a warning/info).

nerrad:
Fixes #3362

This PR ensures that the checkout block signup flow and API correctly checks store options/settings for whether checkout signup is allowed.

In WooCommerce there are two ways a merchant can enable or disable creating accounts as part of checkout processing:

- `woocommerce_enable_signup_and_login_from_checkout` option, aka `WooCommerce > Settings > Accounts & Privacy > Allow customers to create an account during checkout`
- `woocommerce_checkout_registration_enabled` hook

These are now consulted in both the checkout processing (Store API / `CreateAccount` service class) and the rendering of the checkout block. 

- If checkout signup is disabled, Store API will not create an account under any circumstances.
- If checkout signup is disabled, Checkout block will not render a `Create an account?` checkbox, i.e. won't offer this to shopper.

This PR also includes fixes for the relevant PHPUnit tests. These tests have been updated to use correct option values (`yes no` vs `true false`), alongside updating the tests to include `woocommerce_enable_signup_and_login_from_checkout` option.

All tests require:

- Checkout block published in woo checkout page.
- Items added to cart.
- Anon / incognito user session.

#### Test 1: Store API will only create an account if store options allow it
1. Render checkout block with `Create an account?` checkbox. We're going to use a stale copy of the page to submit a checkout API request with `should_create_account=true`.
   - Ensure `WooCommerce > Settings > Accounts & Privacy > Allow customers to create an account during checkout` is checked (temporarily enable).
   - Edit checkout page, ensure `Allow shoppers to sign up for a user account during checkout` is enabled, publish. 
2. In another browser/window, view checkout page (with items in cart, as anon user/logged out). Should see `Create an account` checkbox.
3. Complete checkout form and check `Create an account`. **Do not submit** (yet).
4. As admin, disable `WooCommerce > Settings > Accounts & Privacy > Allow customers to create an account during checkout` and save settings.
5. Submit checkout (with `should_create_account=true`). 
6. Ensure no account is created for the customer, and the order is processed correctly.

As an alternative to (1), could also hack the JS AJAX request to always send `should_create_account=true`, or use other approaches to call the API with that value set.

#### Test 2: Checkout block will only offer to create an account if store options allow it
1. As admin, disable `WooCommerce > Settings > Accounts & Privacy > Allow customers to create an account during checkout` and save settings.
2. View checkout page and ensure there is no `Create an account` checkbox.
3. Repeat the above test with different block settings and ensure no checkbox is displayed unless store option is enabled. E.g. allow checkout signup in settings, enable block `Allow shoppers to sign up for a user account during checkout` setting, publish, change store setting and refresh page.

##### Other possible tests
- Repeat similar tests with a custom plugin implementing `woocommerce_checkout_registration_enabled` hook to disable checkout signup.
- Test various combinations of store settings (e.g. guest checkout `Allow customers to place orders without an account`) and ensure that accounts are created when appropriate. See below for the various permutations of options & expected behaviour.

  - allow_guest=yes | allow_signup=yes
    - enable block option, checkout sign up works
    - disable block option, no checkout sign up available
  - allow_guest=yes | allow_signup=no
    - can't create an account by passing `should_create_account` API params
    - block option not available in editor
  - allow_guest=no  | allow_signup=no
    - block shows login prompt
    - can't purchase without logging in
  - allow_guest=no  | allow_signup=yes
    - block shows no option to sign up
    - successful orders sign up automatically (unless logged in already)
- Test with WooCommerce < v4.7 and ensure that checkout block signup feature is not available, and there is no way to sign up using checkout block.

### Changelog

> Checkout block: Signup now fully supports existing checkout signup store option and related hook. #3371

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/edit.js
+++ b/assets/js/blocks/cart-checkout/checkout/edit.js
@@ -17,6 +17,7 @@ import {
 	PRIVACY_URL,
 	TERMS_URL,
 	CHECKOUT_PAGE_ID,
+	CHECKOUT_ALLOWS_SIGNUP,
 } from '@woocommerce/block-settings';
 import { compareWithWooVersion, getAdminLink } from '@woocommerce/settings';
 import { createInterpolateElement } from 'wordpress-element';
@@ -59,7 +60,8 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 	// setting initial password.
 	// Also implicitly gated to feature plugin, because Checkout
 	// block is gated to plugin
-	const showCreateAccountOption = compareWithWooVersion( '4.7.0', '<=' );
+	const showCreateAccountOption =
+		CHECKOUT_ALLOWS_SIGNUP && compareWithWooVersion( '4.7.0', '<=' );
 	return (
 		<InspectorControls>
 			{ currentPostId !== CHECKOUT_PAGE_ID && (
--- a/src/Assets.php
+++ b/src/Assets.php
@@ -123,17 +123,6 @@ class Assets {
 	}
 
 	/**
-	 * Check whether store is configured to allow shoppers to sign up for
-	 * an account at checkout.
-	 * Inspired by WC_Checkout::is_registration_enabled().
-	 *
-	 * @return Boolean True if store is configured to allow signup.
-	 */
-	public static function is_checkout_signup_allowed() {
-		return apply_filters( 'woocommerce_checkout_registration_enabled', 'yes' === get_option( 'woocommerce_enable_signup_and_login_from_checkout' ) );
-	}
-
-	/**
 	 * Returns block-related data for enqueued wc-block-settings script.
 	 *
 	 * This is used to map site settings & data into JS-accessible variables.
@@ -153,6 +142,7 @@ class Assets {
 			'privacy'  => wc_privacy_policy_page_id(),
 			'terms'    => wc_terms_and_conditions_page_id(),
 		];
+		$checkout       = WC()->checkout();
 
 		// Global settings used in each block.
 		return array_merge(
@@ -198,8 +188,14 @@ class Assets {
 					'privacy'  => self::format_page_resource( $page_ids['privacy'] ),
 					'terms'    => self::format_page_resource( $page_ids['terms'] ),
 				],
-				'checkoutAllowsGuest'           => filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN ),
-				'checkoutAllowsSignup'          => self::is_checkout_signup_allowed(),
+				'checkoutAllowsGuest'           => $checkout instanceof \WC_Checkout && false === filter_var(
+					$checkout->is_registration_required(),
+					FILTER_VALIDATE_BOOLEAN
+				),
+				'checkoutAllowsSignup'          => $checkout instanceof \WC_Checkout && filter_var(
+					$checkout->is_registration_enabled(),
+					FILTER_VALIDATE_BOOLEAN
+				),
 				'baseLocation'                  => wc_get_base_location(),
 				'woocommerceBlocksPhase'        => WOOCOMMERCE_BLOCKS_PHASE,
 				'hasDarkEditorStyleSupport'     => current_theme_supports( 'dark-editor-style' ),
--- a/src/Assets.php
+++ b/src/Assets.php
@@ -123,6 +123,17 @@ class Assets {
 	}
 
 	/**
+	 * Check whether store is configured to allow shoppers to sign up for
+	 * an account at checkout.
+	 * Inspired by WC_Checkout::is_registration_enabled().
+	 *
+	 * @return Boolean True if store is configured to allow signup.
+	 */
+	public static function is_checkout_signup_allowed() {
+		return apply_filters( 'woocommerce_checkout_registration_enabled', 'yes' === get_option( 'woocommerce_enable_signup_and_login_from_checkout' ) );
+	}
+
+	/**
 	 * Returns block-related data for enqueued wc-block-settings script.
 	 *
 	 * This is used to map site settings & data into JS-accessible variables.
@@ -188,7 +199,7 @@ class Assets {
 					'terms'    => self::format_page_resource( $page_ids['terms'] ),
 				],
 				'checkoutAllowsGuest'           => filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN ),
-				'checkoutAllowsSignup'          => filter_var( get_option( 'woocommerce_enable_signup_and_login_from_checkout' ), FILTER_VALIDATE_BOOLEAN ),
+				'checkoutAllowsSignup'          => self::is_checkout_signup_allowed(),
 				'baseLocation'                  => wc_get_base_location(),
 				'woocommerceBlocksPhase'        => WOOCOMMERCE_BLOCKS_PHASE,
 				'hasDarkEditorStyleSupport'     => current_theme_supports( 'dark-editor-style' ),
--- a/src/Domain/Services/CreateAccount.php
+++ b/src/Domain/Services/CreateAccount.php
@@ -139,17 +139,23 @@ class CreateAccount {
 		}
 
 		// From here we know that the shopper is not logged in.
+		// check for whether account creation is enabled at the global level.
+		$checkout = WC()->checkout();
 
-		if ( false === filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN ) ) {
+		if ( $checkout instanceof \WC_Checkout && false === filter_var(
+			$checkout->is_registration_required(),
+			FILTER_VALIDATE_BOOLEAN
+		) ) {
 			// Store requires an account for all checkouts (purchases).
 			// Create an account independent of shopper option in $request.
 			// Note - checkbox is not displayed to shopper in this case.
 			return true;
 		}
 
-		// check for whether account creation is enabled at the global level.
-		$checkout = WC()->checkout();
-		if ( $checkout instanceof \WC_Checkout && false === filter_var( $checkout->is_registration_enabled(), FILTER_VALIDATE_BOOLEAN ) ) {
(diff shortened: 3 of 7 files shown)


src/Domain/Services/CreateAccount.php
```diff
@@ -147,6 +147,13 @@ protected function should_create_customer_account( \WP_REST_Request $request ) {
 			return true;
 		}
 
+		// check for whether account creation is enabled at the global level.
+		$checkout = WC()->checkout();
```
haszari: So far we've been avoiding using core classes like `WC_Checkout` when implementing Store API, because they are closely coupled to the rest of Woo (well, that's my understanding of our approach). 

I had assumed we'd reimplement something similar to `is_registration_enabled()` to factor that logic out from the legacy checkout class (and eventually share with core), to be consistent with that approach.

haszari: I've added our own version of this routine in f076ff062f1b63718c337111c2d705f7d76a50c0. It's in `Assets.php` temporarily, as that's where I needed it, but we should move somewhere else (service class?).

assets/js/blocks/cart-checkout/checkout/edit.js
```diff
@@ -59,7 +60,8 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 	// setting initial password.
 	// Also implicitly gated to feature plugin, because Checkout
 	// block is gated to plugin
-	const showCreateAccountOption = compareWithWooVersion( '4.7.0', '<=' );
+	const showCreateAccountOption =
+		CHECKOUT_ALLOWS_SIGNUP && compareWithWooVersion( '4.7.0', '<=' );
```
haszari, line 64: This option is now potentially confusing if the store has checkout disabled. For example, if store has checkout signup disabled, this option won't work. I don't think that's a blocker for this PR, but we might want to tweak the UI or add a warning message in future.

haszari (approved):
Pre-approving so you can see my comments, this is looking good (note @nerrad I have pushed one change). I haven't tested yet.

Here's the principles I have in mind for how this works, I've reviewed based on this and will test accordingly next.

- Store has a policy for whether signup is allowed in any checkout.
   - This is controlled by an option and a hook.
   - `woocommerce_enable_signup_and_login_from_checkout`
   - `woocommerce_checkout_registration_enabled`
- If store does not allow checkout signup:
   - The checkout API will not create an account even if requested.
   - The checkout block will not show a `Create an account?` checkbox.
- The checkout block has an option to show `Create an account` checkbox, this is visual only and is always available (but has no effect on front end if store doesn't allow checkout signup).

assets/js/blocks/cart-checkout/checkout/edit.js
nerrad: I agree, but in the absence of that UI treatment, I think it's preferable to _not_ show the option since it is ineffective if signup is globally disabled. So I think this is a safer starting point. I think it'd be actually more confusing if the option shows when signup is disabled globally.

src/Domain/Services/CreateAccount.php
nerrad: I think for now, let's just keep it as is since we're aligning with the existing Checkout flow implementation of this option (which is also filtered).

assets/js/blocks/cart-checkout/checkout/edit.js
haszari: Works for me :) In testing it seems pretty reasonable.

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -190,11 +194,23 @@ public function invalid_email_data() {
 	public function test_no_account_requested() {
```
haszari: Should rename this to `test_no_account_created` to cover the new use case.

src/Domain/Services/CreateAccount.php
```diff
@@ -139,14 +139,27 @@ protected function should_create_customer_account( \WP_REST_Request $request ) {
...
+			$checkout->is_registration_required(),
+			FILTER_VALIDATE_BOOLEAN
+		) ) {
 			// Store requires an account for all checkouts (purchases).
 			// Create an account independent of shopper option in $request.
 			// Note - checkbox is not displayed to shopper in this case.
 			return true;
 		}
 
+		if ( $checkout instanceof \WC_Checkout && false === filter_var(
```
haszari: Should we return early if `! $checkout instanceof \WC_Checkout` instead of multiple checks?

haszari: Yes we can 4cdb22e4ff8809e91897b97ce2ac697c7840e549

tests/php/Domain/Services/CreateAccount.php
haszari: Leaving this as is for now as the new test case isn't live yet (commented `test with explicitly turning off global registration`).

src/Assets.php
```diff
@@ -187,8 +188,14 @@ public static function get_wc_block_data( $settings ) {
...
-				'checkoutAllowsGuest'           => filter_var( get_option( 'woocommerce_enable_guest_checkout' ), FILTER_VALIDATE_BOOLEAN ),
-				'checkoutAllowsSignup'          => filter_var( get_option( 'woocommerce_enable_signup_and_login_from_checkout' ), FILTER_VALIDATE_BOOLEAN ),
+				'checkoutAllowsGuest'           => $checkout instanceof \WC_Checkout && false === filter_var(
+					$checkout->is_registration_required(),
+					FILTER_VALIDATE_BOOLEAN
+				),
+				'checkoutAllowsSignup'          => $checkout instanceof \WC_Checkout && filter_var(
+					$checkout->is_registration_enabled(),
+					FILTER_VALIDATE_BOOLEAN
+				),
```
haszari, line 198: Could consider renaming the block option const to more closely match the API. Probably no need to do this now, not a blocker.

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -15,6 +15,7 @@
  * that woocommerce_blocks_phase===3, i.e. dev build. Tests will fail
  * with other builds (release feature plugin, woo core package).
  * Related: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3211
+ * @group TESTS
```
nerrad: I forgot to remove this temporary group, it should be removed. Depending on who merges this pull to `trunk`, let's just make sure this is removed before merge.

tests/php/Domain/Services/CreateAccount.php
```diff
@@ -190,16 +195,28 @@ public function invalid_email_data() {
...
+		// 	'maryjones@testperson.net',
+		// 	'Mary',
+		// 	'Jones',
+		// 	[
+		// 		'can_register' => 'no',
+		// 		'should_create_account' => 'yes',
+		// 		'enable_guest_checkout' => 'yes',
+		// 	],
+		// );
+
```
nerrad: I forgot to remove the comments on this code.

tests/php/Domain/Services/CreateAccount.php
haszari: Test is back! So we should rename :)

haszari (approved):
Did another round of testing with WooCommerce 4.8-dev and 4.6.2, because checkout signup is disabled for Woo < v4.7. This was building Woo from relevant release branch (`master` vs `release/4.6`).

I didn't find any blocking issues, though there's one quirk with `allow_guest=no allow_checkout_signup=yes` that we might want to follow up later.

Here's recap of what I tested.

### WooCommerce 4.8.-dev
1. allow_guest=no  | allow_signup=no
    - block shows login prompt
    - can't purchase without logging in
1. allow_guest=yes | allow_signup=yes
    - enable block option, checkout sign up works 
    - enable block option, checkout sign up only does it if user requests
    - disable block option, no checkout sign up
1. allow_guest=yes | allow_signup=no
    - can't create an account by hacking `should_create_account` API param (or stale render of block with `Create an account` rendered)
    - block option not available in editor
1. allow_guest=no  | allow_signup=yes
    - block shows no option to sign up
    - successful orders sign up automatically (unless logged in already)
    - _quirk in editor_- can disable `Create an account` checkbox
       - if disabled, block prompts to log in

### WooCommerce 4.6.2
1. allow_guest=no  | allow_signup=no
    - block shows login prompt
    - can't purchase without logging in
1. allow_guest=yes | allow_signup=no
    - block option not available in editor (woo 4.6)
    - order doesn't create account
    - can't create an account by hacking `should_create_account` API param
1. allow_guest=yes | allow_signup=yes
    - block option not available in editor (woo 4.6)
    - order doesn't create account
    - can't create an account by hacking `should_create_account` API param
1. allow_guest=no  | allow_signup=yes
    - block option not available in editor (woo 4.6)
    - similar quirk to above - block prompts to log in, because block checkout signup is not available in Woo < 4.7

Regarding the quirk, the expected behaviour is debatable. Some might expect it to silently, automatically sign up (consistent with shortcode); others might expect the block option to inhibit or interact with the store settings (our current behaviour). Happy to leave this for follow up. cc @LevinMedia for some design consideration 🤔

PR #3376, opened 2020-11-05
Show the phone number field in the billing section when shipping is disabled in settings

### LINKED ISSUE #3231: Checkout Block: Phone field missing from checkout page when shipping is not configured
## Describe the bug
Phone field missing from checkout page when shipping is not configured

## To reproduce

1. Add the Checkout Block to your page
2. Make sure that Phone is checked
3. Save and check your page

## Expected behavior
To see the phone field in the public page. Currently, it is not displayed when you go to checkout.

* Core version: 5.5.1
* Gutenberg version: 9.1.0
* WooCommerce version: 4.5.2
* WooCommerce Blocks version: 3.5.0

* OS: macOS
* Browser: Chome and Firefox

### LINKED ISSUE #3346: Add phone input to billing form
As per our discussion in pca54o-sm-p2#comment-665 we decided to add the phone input to billing form as well.
The new flow would be:

- Show phone number in shipping form only.
- If shipping form is disabled (no shipping), show the phone number in the billing step.
- Phone number should still be saved on billing address in the backend.

We might need some refactoring to get this done.
We want to get phone number as an independent component that uses context and can be inserted anywhere.

mikejolley:
Moves the rendering of address fields to a single file so that PhoneNumber (new component) can be rendered in _either_ billing or shipping, depending on visibility, without duplication.

I was going to update context to handle this (hence why I cleaned up some missing docblocks) however, it made more sense to handle this within the address-step component and simply pass around the attributes as before.

Fixes #3231
Fixes #3346 

Ensure the phone number field is enabled in block settings, then test 2 cases:

1. With shipping enabled, ensure phone number displays beneath the shipping step
2. With shipping disabled (you can delete all rates and zones to edit general settings), ensure phone number displays beneath the billing step.

In both scenarios the values should change and be saved as billing information.

### Changelog

> Show the phone number field in the billing section when shipping is disabled in settings

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -66,6 +66,8 @@ const CheckoutContext = createContext( {
 	},
 	hasOrder: false,
 	isCart: false,
+	shouldCreateAccount: false,
+	setShouldCreateAccount: ( value ) => void value,
 } );
 
 /**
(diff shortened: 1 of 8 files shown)


budzanowski (approved):
This looks great. I was wondering why we only have one phone number field but the link to the thread in the Issue has clarified that.
Works nicely! :shipit:

PR #3377, opened 2020-11-06
Add WordPress.org assets folder and add latest assets.

kloon:
I was in the process of updating a bunch of our plugin WordPress.org assets and found that this plugin does not keep track of assets in github as many of our other plugins does. After speaking with @mikejolley he suggested I just make the update directly in SVN and then do a PR to introduce a new .wordpress-org folder with the assets that can become part of the release process in a cooldown.

Fixes #

#### Accessibility

1.
2.
3.

### Changelog


nerrad (approved):
One benefit of the way things are currently setup is wp.org assets can be updated separately via svn without having to update this repository (or do an official "release"). With the requirement of updating the assets through svn only it makes it pretty clear the path to doing so (at the expense of a more painful update process using svn).

With that said, I'm okay with making this change and I've reviewed our currently WordPress deploy action (which will supposedly handle the appropriate updates to the svn assets folder from `.wordpress-org`). I just thought it worthwhile to note what we'd lose by this change.

PR #3379, opened 2020-11-06
Exclude checkout-draft orders from WC Admin reports and My Account > Orders

### LINKED ISSUE #3301: Filter out the `checkout-draft` status from all order queries by default except where explicitly needed by Woo Blocks
As a part of the work on the Checkout Block we implemented a new order status for orders: `checkout-draft`.  While this status is feature gated, there are some implications around it's usage in how its surfaced in numerous places in the WooCommerce core ui/ux and potential confusion around its implementation (example totals for draft orders being included in sales total for WooCommerce reports).

While we are internally working on making this an official order status in WooCommerce core (see pca54o-ib-p2) , in the interim let's explore how we can exclude this new status from all order queries by default except where explicitly implemented by WooCommerce blocks.

Aljullu:
Fixes #3301.

This PR:
* Forces `checkout-draft` orders to be excluded when computing WooCommerce Admin reports data.
* Removes the option to enable/disable it from WooCommerce Admin settings, so the checkbox doesn't appear at all in Analytics > Settings.
* Hides draft orders from My Account > Orders.

1. Make sure your store has at least one completed order and one draft order. You can achieve that adding some products to the cart and navigating to the Cart and Checkout blocks but without completing the order.
2. Go to Analytics > Settings and verify there isn't a `Custom statuses` section or, if it exists, make sure none of the statuses is named `Draft`.
3. Go to the bottom of the page and click on `Delete Previously Imported Data`.
4. Wait until the process finishes and then import all data again.
5. Go to Analytics > Orders.
6. Verify the 'draft' order is not counted in the totals.
7. In the frontend, go to My Account > Orders with the user that made the draft order.
8. Verify the draft order is not listed there.

### Changelog

> Prevent draft orders from being counted in analytics reports and appearing in the My Account page.

### DIFF
--- a/assets/js/settings/shared/index.js
+++ b/assets/js/settings/shared/index.js
@@ -2,7 +2,6 @@
  * External dependencies
  */
 import compareVersions from 'compare-versions';
-import { addFilter } from '@wordpress/hooks';
 
 /**
  * Internal dependencies
@@ -11,6 +10,7 @@ import { getSetting } from './get-setting';
 
 export * from './default-constants';
 export { setSetting } from './set-setting';
+import './exclude-draft-status-from-analytics';
 
 /**
  * Note: this attempts to coerce the wpVersion to a semver for comparison
@@ -67,40 +67,3 @@ export { compareVersions, getSetting };
  * @return {string} Full admin URL.
  */
 export const getAdminLink = ( path ) => getSetting( 'adminUrl' ) + path;
-
-addFilter(
-	'woocommerce_admin_analytics_settings',
-	'woocommerce-admin',
-	( settings ) => {
-		const removeCheckoutDraft = ( optionsGroup ) => {
-			if ( optionsGroup.key === 'customStatuses' ) {
-				return {
-					...optionsGroup,
-					options: optionsGroup.options.filter(
-						( option ) => option.value !== 'checkout-draft'
-					),
-				};
-			}
-			return optionsGroup;
-		};
-
-		const actionableStatusesOptions = settings.woocommerce_actionable_order_statuses.options.map(
-			removeCheckoutDraft
-		);
-		const excludedStatusesOptions = settings.woocommerce_excluded_report_order_statuses.options.map(
-			removeCheckoutDraft
-		);
-
-		return {
-			...settings,
-			woocommerce_actionable_order_statuses: {
-				...settings.woocommerce_actionable_order_statuses,
-				options: actionableStatusesOptions,
-			},
-			woocommerce_excluded_report_order_statuses: {
-				...settings.woocommerce_excluded_report_order_statuses,
-				options: excludedStatusesOptions,
-			},
-		};
-	}
-);
--- a/assets/js/settings/shared/index.js
+++ b/assets/js/settings/shared/index.js
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import compareVersions from 'compare-versions';
+import { addFilter } from '@wordpress/hooks';
 
 /**
  * Internal dependencies
@@ -66,3 +67,40 @@ export { compareVersions, getSetting };
  * @return {string} Full admin URL.
  */
 export const getAdminLink = ( path ) => getSetting( 'adminUrl' ) + path;
+
+addFilter(
+	'woocommerce_admin_analytics_settings',
+	'woocommerce-admin',
+	( settings ) => {
+		const removeCheckoutDraft = ( optionsGroup ) => {
+			if ( optionsGroup.key === 'customStatuses' ) {
+				return {
+					...optionsGroup,
+					options: optionsGroup.options.filter(
+						( option ) => option.value !== 'checkout-draft'
+					),
+				};
+			}
+			return optionsGroup;
+		};
+
+		const actionableStatusesOptions = settings.woocommerce_actionable_order_statuses.options.map(
+			removeCheckoutDraft
+		);
+		const excludedStatusesOptions = settings.woocommerce_excluded_report_order_statuses.options.map(
+			removeCheckoutDraft
+		);
+
+		return {
+			...settings,
+			woocommerce_actionable_order_statuses: {
+				...settings.woocommerce_actionable_order_statuses,
+				options: actionableStatusesOptions,
+			},
+			woocommerce_excluded_report_order_statuses: {
+				...settings.woocommerce_excluded_report_order_statuses,
+				options: excludedStatusesOptions,
+			},
+		};
+	}
+);
--- a/assets/js/settings/shared/exclude-draft-status-from-analytics.js
+++ b/assets/js/settings/shared/exclude-draft-status-from-analytics.js
@@ -5,7 +5,7 @@ import { addFilter } from '@wordpress/hooks';
 
 addFilter(
 	'woocommerce_admin_analytics_settings',
-	'woocommerce-admin',
+	'woocommerce-blocks/exclude-draft-status-from-analytics',
 	( settings ) => {
 		const removeCheckoutDraft = ( optionsGroup ) => {
 			if ( optionsGroup.key === 'customStatuses' ) {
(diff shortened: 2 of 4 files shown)


nerrad:
As a last resort I'm fine with adding this exclusion specifically for wc-admin reports (I haven't reviewed code, just commenting on the approach). However, the goal with #3301 is to by default exclude draft orders from all order queries and then explicitly add it for queries Blocks needs. I made a suggestion [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3301#issuecomment-723162520) that on the surface seems feasible to me.

haszari (commented):
Tested and confirmed this PR works as intended 👍 

I don't have any concerns with the code, but I'm not sure about whether this is the best fix for the underlying issue. In general I agree with @nerrad that it would be good to try to filter these `checkout-draft` orders at source for all queries. If that approach works out, would we still need this PR? This is the main question/blocker from my point of view. 

Aside from that, it feels very heavy-handed to hook wc-admin to hide and exclude our order status. Wc-admin has nice transparent UI for excluding core and custom statii. So if we do need to customise the behaviour to ignore our status, would it be safer and simpler to simply change the default settings to exclude it (i.e. tick the `Exclude` box by default)? That way it's clear in the UI that it's excluded, and the store owner can add it back in if they choose.

If we do proceed with this PR, I think it would be a good idea to get a review from someone working on wc-admin as well.

One more comment (unrelated to this PR, but relevant to the wider discussion on `checkout-draft` orders). The name `Draft` used in the UI is not very clear, since the status is scoped to cart/checkout blocks only. I'm wondering if we should rename it to help make this clear - e.g. `Checkout Block Draft` (or something better).

Aljullu:
Thanks for the feedback @nerrad @haszari.

> As a last resort I'm fine with adding this exclusion specifically for wc-admin reports (I haven't reviewed code, just commenting on the approach). However, the goal with #3301 is to by default exclude draft orders from all order queries and then explicitly add it for queries Blocks needs. I made a suggestion [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3301#issuecomment-723162520) that on the surface seems feasible to me.

Makes sense. I applied your suggestion in 94cec6c83cf5b5ce05cd106daf217c0b08f13a03. Notice however WC Admin doesn't use `wc_get_orders()`, so we still need to hook into `woocommerce_analytics_excluded_order_statuses` and add the checkout draft status into the excluded list.

> Aside from that, it feels very heavy-handed to hook wc-admin to hide and exclude our order status. Wc-admin has nice transparent UI for excluding core and custom statii. So if we do need to customise the behaviour to ignore our status, would it be safer and simpler to simply change the default settings to exclude it (i.e. tick the `Exclude` box by default)? That way it's clear in the UI that it's excluded, and the store owner can add it back in if they choose.

Currently that's not possible, the default excluded statuses array can't be extended:

https://github.com/woocommerce/woocommerce-admin/blob/main/client/analytics/settings/config.js#L90

We would need to update WooCommerce Admin so it allows extensions to filter that array.

However, I feel like completely hiding the 'Draft' status from WooCommerce Admin is a better approach than allowing the end user to exclude/include it. Draft orders are deleted every day, so I see them more like the 'trashed' status: even though they exist internally, they don't have any meaning in the WC Analytics context. In line with WordPress Philosophy _[Decisions, not Options](https://wordpress.org/about/philosophy/#decisions)_, I don't think we should allow users to enable a status that is only used internally by WooCommerce, with the risk of breaking their store analytics.

In this direction, I created an issue in WC Admin repo in order to discuss whether we can hide statuses completely from WC Admin without the need to hook into each UI item: https://github.com/woocommerce/woocommerce-admin/issues/5600.

> The name `Draft` used in the UI is not very clear, since the status is scoped to cart/checkout blocks only. I'm wondering if we should rename it to help make this clear - e.g. `Checkout Block Draft` (or something better). 

I agree with your point that currently the name might be confusing for developers and merchants. Is it ok to break it into a different issue so we can discuss it outside of this PR?

haszari:
> > The name Draft used in the UI is not very clear, since the status is scoped to cart/checkout blocks only. I'm wondering if we should rename it to help make this clear - e.g. Checkout Block Draft (or something better).

> I agree with your point that currently the name might be confusing for developers and merchants. Is it ok to break it into a different issue so we can discuss it outside of this PR?

For sure, this is definitely a separate issue that shouldn't block this 👍 

Let me know if you want me to log it @Aljullu .

> In line with WordPress Philosophy Decisions, not Options, I don't think we should allow users to enable a status that is only used internally by WooCommerce, with the risk of breaking their store analytics.

Good point - raising this as feedback for wc-admin seems like a good approach here. Would be nice if your feature request led to a simpler UX for wc-admin users 😃

haszari (approved):
This is looking good, though the PR description and title are now stale. With the global hook for all order queries, this is a much more fundamental fix. It's important that the PR clearly explains the intention and impact of the change, especially in this case IMO. Can you update the title and description to make this clear before merge? Or if you like, split off a separate PR.

Your explanation of the approach to filtering checkout draft orders in wc-admin makes sense - works for me. At the same time I'm curious to see how that feedback plays out in wc-admin; maybe we will revisit this if wc-admin adds support that better suits our needs. Also possibly it's a good idea to get a review from wc-admin team before merge - up to you. 

I added a [comment](https://github.com/woocommerce/woocommerce-admin/issues/5600#issuecomment-725777028) to the wc-admin issue to raise visibility on some of these concerns.

Aljullu:
> This is looking good, though the PR description and title are now stale. With the global hook for all order queries, this is a much more fundamental fix. It's important that the PR clearly explains the intention and impact of the change, especially in this case IMO. Can you update the title and description to make this clear before merge? Or if you like, split off a separate PR.

Done, thanks for pointing this out. :+1: 

> Also possibly it's a good idea to get a review from wc-admin team before merge - up to you.

Sounds good. I will hold on merging this until getting we get more feedback.

Aljullu:
@haszari I tagged you for an extra review given that several things changed:
* I removed the hook from `wc_get_orders()` (see related Slack discussion p1605146329149000-slack-C8X6Q7XQU).
* I added a specific filter for orders displayed in My Account so drafts are not shown (7272d4b96cf247fc67da66bbb79cfea149dd898a).

haszari:
> @haszari I tagged you for an extra review

Thanks @Aljullu - will take another look today. Overnight I realised I need to test more extensively also, if we are filtering order query args .. though now I see that's reverted 😆

haszari (approved):
Code looks good, I'm on board with the new "targeted whack a mole" approach :)

Tested and confirmed this fixes draft order status showing up inappropriately in:

- My Account > Orders
- Analytics > Orders
- Analytics > Settings

Now that we've shifted the approach to exclude this status where needed (vs exclude by default in all queries), I think it's a good idea to tweak the original issue (change title or add a comment) to capture the outcome of that discussion. 

Also curious to hear wc-admin team feedback on this approach when they review. 

Overall this is looking good to go from my point of view 👍

Aljullu:
> Now that we've shifted the approach to exclude this status where needed (vs exclude by default in all queries), I think it's a good idea to tweak the original issue (change title or add a comment) to capture the outcome of that discussion.

Comment added: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3301#issuecomment-727230037

> Also curious to hear wc-admin team feedback on this approach when they review.

:+1: I will ping them next Monday.

assets/js/settings/shared/index.js
```diff
@@ -66,3 +67,40 @@ export { compareVersions, getSetting };
  * @return {string} Full admin URL.
  */
 export const getAdminLink = ( path ) => getSetting( 'adminUrl' ) + path;
+
+addFilter(
+	'woocommerce_admin_analytics_settings',
+	'woocommerce-admin',
```
jeffstieler: I believe this is meant to be the namespace of your plugin, not the plugin applying the filter ([`doAction()`](https://github.com/WordPress/gutenberg/tree/master/packages/hooks#api-usage) doesn't take a namespace argument).

PR #3380, opened 2020-11-06
Force node 12 for action

nerrad:
Currently our repository doesn't support node 14.x and this branch of node recently became LTS so GitHub actions now default to that version. This is an attempt to force a specific version of node for the action.

## To test

I _think_ this should be verified by the `bundlesize` action running in the pull request. But it's possible I might need to have the workflow in trunk first so we'll see on the first execution in this pr.


nerrad:
It looks like the only workflow needing updated is the one in this pr, since the action was successful I'll go ahead and merge.

PR #3381, opened 2020-11-06
Fix PHP8 Unsupported Types with Operators warnings for PHP8

nerrad:
Exakat scan audit of blocks caught 8 instances in the WooCommerce Blocks code base that triggered "PHP 8 Unsupported Types with Operators" rule failures. In this pull request I fixed 7 of them, the last one for [this code](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/src/StoreApi/Routes/AbstractTermsRoute.php#L134) is a false positive. I did tighten up the types in 476049f but that still wasn't enough to fix the rule fail for the calculation happening. I'm confident enough in the code that is present currently to ignore the false positive.

Image 2020-11-06 at 5 24 22 PM

## To Test

This impacts a few areas that will need tested:

**Commits d460aac and 7698570**
Impact: Various cart endpoints

- With the cart block, try increasing and decreasing quantity of products in the cart and verify that totals update etc. Verify there are no console errors or errors with the block.
- With the cart block try removing an item from the cart, verify totals update etc and there are no errors.
- In the All Products block, try adding products to the cart, verify there are no errors.

**Commit 2762798**
Impact: Checkout

- With the checkout block, loading the block and submitting a payment (doesn't matter what payment method) should give the expected results without any errors (in console or UI). Also verify that the order persists to the backend as expected after order is complete.

**Commit 476049f**
Impact: Attribute Filter Block

- Verify the attribute filter block returns expected terms for various settings and that there are no errors.
- Verify that pagination for results on filtered products works as expected.

**Commit a34611d**
Impact: Product Categories Block

- Verify that the block in the editor and in the frontend works as expected with no errors. Any problems should surface fairly quickly.

### DIFF
--- a/src/BlockTypes/ProductCategories.php
+++ b/src/BlockTypes/ProductCategories.php
@@ -129,7 +129,7 @@ class ProductCategories extends AbstractDynamicBlock {
 			]
 		);
 
-		if ( ! is_array( $terms ) || empty( $categories ) ) {
+		if ( ! is_array( $categories ) || empty( $categories ) ) {
 			return [];
 		}
 
--- a/src/BlockTypes/ProductCategories.php
+++ b/src/BlockTypes/ProductCategories.php
@@ -57,7 +57,7 @@ class ProductCategories extends AbstractDynamicBlock {
 		$uid        = uniqid( 'product-categories-' );
 		$categories = $this->get_categories( $attributes );
 
-		if ( ! $categories ) {
+		if ( empty( $categories ) ) {
 			return '';
 		}
 
@@ -129,7 +129,7 @@ class ProductCategories extends AbstractDynamicBlock {
 			]
 		);
 
-		if ( ! $categories ) {
+		if ( ! is_array( $terms ) || empty( $categories ) ) {
 			return [];
 		}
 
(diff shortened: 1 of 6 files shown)


src/BlockTypes/ProductCategories.php
```diff
@@ -129,7 +129,7 @@ protected function get_categories( $attributes ) {
 			]
 		);
 
-		if ( ! $categories ) {
+		if ( ! is_array( $terms ) || empty( $categories ) ) {
```
mikejolley: I assume the `is_array` check is to catch a returned wp error?

mikejolley (approved):
I fixed a small variable naming mistake and ran through the steps. Looks to be working now 👍🏻

src/BlockTypes/ProductCategories.php
nerrad: oh shoot, I had fixed this variable name mistake locally and forgot to push 🤦

nerrad: > I assume the is_array check is to catch a returned wp error?

Yes and basically to ensure it's of the expected type.

PR #3383, opened 2020-11-09
Cart and checkout should respect the global "Hide shipping costs until an address is entered" setting

### LINKED ISSUE #3148: Fix shipping options showing when "Hide shipping costs until address is entered" setting is checked and there is no address entered yet.
## Describe the bug
When logged out and before entering address in checkout (page is set to show default shipping cost), the default shipping cost is shown correctly under Shipping Options, but is shown as zero in the basket summary on the right hand side. See screenshot.

## To reproduce
1. Ensure you're logged out and there's a default shipping cost for the default zone
1. Add products to basket and proceed to checkout 
2. You will see the issue captured in the attached screenshot

## Expected behavior
The summary section should also show the default shipping cost. Not 0.

## Screenshots
Attached. 

* Core version: 5.5.1
* Gutenberg version (if installed):-
* WooCommerce version: 4.4.1
* WooCommerce Blocks version: 3.3.0
* Site language: English
* Any other plugins installed: many many others. If you ask about specific plugins I can let you know if they're installed.

* OS: macOS
* Browser: multiple
Screenshot 2020-09-15 at 11 04 35

mikejolley:
This PR fixes shipping discrepancies during cart and checkout caused by the global `woocommerce_shipping_cost_requires_address` setting. When `woocommerce_shipping_cost_requires_address ` is true, shipping total from the cart is **always** 0. This conflicts with the block level setting `Hide shipping costs until an address is entered` which has no control over this.

1. Removes the block level setting "Hide shipping costs until an address is entered" since there is no way to make the cart totals respect this and ignore the global setting.
2. The REST API will return `null` shipping totals if calculations have not been performed by core.
3. The REST API will return an empty array of shipping rates if calculations have not been performed by core.
4. Updated placeholders when shipping has not been returned in the above circumstances.
5. Fixed an issue where all fields were optional in the shipping calculator, since those fields are required by core on order to return rates.

Fixes #3148

### Screenshots

If no rates are returned from the API (customer has no input their full shipping address), the shipping section will show a message instead of the list of current rates:

Checkout – one wordpress test 2020-11-09 12-45-02

And in the summary area a similar message is shown:

Checkout – one wordpress test 2020-11-09 12-44-33

This is instead of showing a totals row with 0 cost which is not accurate.

1. Before running this PR, in Settings > Shipping > Shipping Options, turn on the option to require an address
2. Setup a single shipping rate
3. Edit the cart block and disable "Hide shipping costs until an address is entered"
4. In incognito, go to cart and see that the shipping rate is selected, but the shipping cost is 0. This is the issue reported in #3148.

Then apply this PR and confirm:

1. The messaging in cart and checkout states that shipping will be shown after providing an address.
2. After entering your address, rates are then displayed and the total cost is not 0
3. The "Hide shipping costs until an address is entered" block level option is no longer displayed when editing the block.

This is pending feedback from @nerrad since it removes the block level option and needs design approval.

### Changelog

> Ensures cart and checkout respect the global "Hide shipping costs until an address is entered" setting.


Aljullu (approved):
Pending feedback from @nerrad, approving from my side: code looks good and it tests as expected. :ok_hand: 

Only thing that I think might be worth doing before merging is to completely remove the `IS_SHIPPING_COST_HIDDEN` setting, which doesn't seem to be used anywhere else:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/settings/blocks/constants.js#L52

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Assets.php#L177

I also see e2e tests are failing. It doesn't seem to be related to this pull, but you could re-push it again so tests run again just to confirm.

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -192,13 +198,13 @@ public function get_properties() {
 						],
 						'total_shipping'     => [
 							'description' => __( 'Total price of shipping.', 'woo-gutenberg-products-block' ),
-							'type'        => 'string',
+							'type'        => [ 'string', 'null' ],
```
nerrad: Should we indicate what `null` means in the response here?

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -265,16 +271,22 @@ public function get_item_response( $cart ) {
 		// Get cart errors first so if recalculations are performed, it's reflected in the response.
 		$cart_errors = $this->get_cart_errors( $cart );
 
+		// The core cart class will not include shipping in the cart totals if `show_shipping()` returns false. This can
+		// happen if an address is required, or through the use of hooks. This tracks if shipping has actually been
+		// calculated so we can avoid returning costs and rates prematurely.
+		$has_calculated_shipping = $cart->show_shipping();
```
nerrad, line 277: This won't impact how shipping is shown in the checkout right?

nerrad (approved):
> I also see e2e tests are failing. It doesn't seem to be related to this pull, but you could re-push it again so tests run again just to confirm.

Those e2e fails look legit (and need fixed) because they are related to toggling shipping option.

Otherwise, I have a couple non-blocking comments in the code and it looks good. I have not tested (leaning on Albert and your testing there). Pre-approving subject to e2e fixes and addressing my feedback.

src/StoreApi/Schemas/CartSchema.php
mikejolley: Not this specifically but the `null` cases will (we show a message saying shipping will update once address info is complete).

src/StoreApi/Schemas/CartSchema.php
mikejolley: Added

PR #3399, opened 2020-11-13
Expose `discount_type` in Store API coupon endpoints

### LINKED ISSUE #3274: Integration Tracking (WC Subscriptions): Add coupon type to coupons endpoint
As mentioned in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3273, we need the coupon type value in our coupons endpoint, this is needed to display them differently in certain places.

This is a generic issue, but it is needed for WC Subscriptions.

We need to:
- Add the value to our endpoints.
- Update our docs (cart, checkout, coupons).
- Update our hooks to pass along that value.
- Possibly update any test fixtures.

haszari:
Fixes #3274

### Screenshots

Example after `POST`ing two coupons in cart, `POST /wc/store/cart/apply-coupon`

Screen Shot 2020-11-13 at 2 26 13 PM

1. Test cart & checkout with & without various coupons. Check network requests inspector in dev tools and confirm coupon type is present and correct. 

### Changelog

> Store API: Add `discount_type` field to coupon-related requests.

### DIFF
--- a/src/StoreApi/Schemas/CartCouponSchema.php
+++ b/src/StoreApi/Schemas/CartCouponSchema.php
@@ -24,8 +24,8 @@ class CartCouponSchema extends AbstractSchema {
 	 */
 	public function get_properties() {
 		return [
-			'code'   => [
-				'description' => __( 'The coupons unique code.', 'woo-gutenberg-products-block' ),
+			'code'          => [
+				'description' => __( 'The coupon\'s unique code.', 'woo-gutenberg-products-block' ),
 				'type'        => 'string',
 				'context'     => [ 'view', 'edit' ],
 				'arg_options' => [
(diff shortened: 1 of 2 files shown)


haszari:
@mikejolley I see in the schema for coupons that most fields have `'context' => [ 'view', 'edit' ]`. Does this mean that the schema allows POST (edit, write, mutate) of these fields?

For coupons, I'm thinking the only one that we allow POST for is the coupon ID. The rest (coupon type, discount etc) are all read-only. Should we change the schema to `view` only?

Or ... am I completely misunderstanding the meaning of `context` param 😸

senadir:
Context means the place this request is going to be consumed in, and it would return extra fields depending on that.
In our case, it means we're going to return this field when viewing content (storefront) and when editing content (editor, backend).
Different context requires certain authentications.
Your misunderstanding is a common one, you can see an [extended explanation here](https://make.xwp.co/2017/07/25/defining-context-in-the-wp-rest-api/).

src/StoreApi/Schemas/CartCouponSchema.php
```diff
@@ -8,6 +8,7 @@
  *
  * @internal This API is used internally by Blocks--it is still in flux and may be subject to revisions.
  * @since 2.5.0
+ * @since $VID:$ Coupon type (`discount_type`) added.
```
senadir, line 11: This API is not public yet so I don't think we need this.

senadir (approved):
I tested this and it works fine, it exposed WCS coupons types just fine, and the data is passed in `useStoreCartCoupons` and `useStoreCart` so no extra handling is required there.

src/StoreApi/Schemas/CartCouponSchema.php
haszari: Even though it's not public I still think it's useful to track when fields are added or schema changed – doesn't hurt right?

PR #3400, opened 2020-11-13
Fix incorrect usage of static method in Stripe payment method integration

nerrad:
While testing some checkout related issues, a fatal error got thrown in my testing environment (with PHP 8.0):

> PHP Fatal error:  Uncaught Error: Non-static method WC_Stripe_Payment_Request::add_order_meta() cannot be called statically in /srv/www/woo/public_html/wp-content/plugins/woo-blocks/src/Payments/Integrations/Stripe.php:249

It turns out, this was never a static method and just _happened_ to work okay in earlier php versions but will result in a fatal in PHP 8 (which is why it never got noticed until now). Obviously if this method was implementing any instance calls it would have surfaced earlier but because it's not, it didn't get surfaced.

I would have like to call the original method from it's instance but the `WC_Stripe_Payment_Request` class is instantiated directly in the Stripe add-on and not assigned to anything. On top of that, it registers filter and action hooks directly so it can't be re-instantiated anywhere.

Thus, in this pull I just reproduced the logic in the Stripe integration class.

I also updated our phpcs compat configuration to check for PHP 7+ because we don't support PHP 5.6 anymore.

## To Test

* [ ] Complete a purchase using any express payment method (Chrome pay or Apple Pay) and ensure the correct payment method used is recorded with the order meta.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -10,7 +10,7 @@
 
 	<!-- Configs -->
 	<config name="minimum_supported_wp_version" value="4.7" />
-	<config name="testVersion" value="5.6-" />
+	<config name="testVersion" value="7.0-" />
 
 	<!-- Rules -->
 	<rule ref="WooCommerce-Core" />
--- a/src/Payments/Integrations/Stripe.php
+++ b/src/Payments/Integrations/Stripe.php
@@ -246,7 +246,7 @@ final class Stripe extends AbstractPaymentMethodType {
 			// phpcs:ignore WordPress.Security.NonceVerification
 			$post_data = $_POST;
 			$_POST     = $context->payment_data;
-			WC_Stripe_Payment_Request::add_order_meta( $context->order->id, $context->payment_data );
+			$this->add_order_meta( $context->order, $data['payment_request_type'] );
 			$_POST = $post_data;
 		}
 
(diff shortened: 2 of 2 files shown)


budzanowski:
> I would have like to call the original method from it's instance but the WC_Stripe_Payment_Request class is instantiated directly in the Stripe add-on and not assigned to anything. On top of that, it registers filter and action hooks directly so it can't be re-instantiated anywhere.

You could use:
```PHP
public ReflectionClass::newInstanceWithoutConstructor ( ) : object
```

This will give you an instance of the class and at the same time it avoids the the problems introduced by the constructor.
Since this does not call the constructor this is a fast operation. This is an obvious hack, but at the same time we avoid maintaining and possiby versioning `add_order_meta` if it changes in future ( or maybe it is already different in some older versions ) of the Stripe plugin.

Aproving. Feel free to go with your approach if you feel that going ReflectioClass way is not a good choice.

budzanowski (approved):

nerrad:
Thanks for the review @budzanowski!

> You could use:
> `public ReflectionClass::newInstanceWithoutConstructor ( ) : object`

You're right, however in general I think it's a bad practice to use the Reflection API in production code. In this particular case it would introduce fragility because if the original class is refactored so that the method we're reflecting is using properties set in the constructor, it'd be a breaking change for our implementation.

A couple reasons for going the route I did in this pull request:

- The code from the method is minimal and fairly targeted in scope. Thus it's very unlikely any changes in the original method will impact this.
- This code in the blocks plugin will be moved back to the Stripe plugin (it's in our roadmap). The ideal solution here is that the Stripe extension is improved so doing something like this is not necessary (i.e. modifying the `WC_Stripe_Payment_Request` class so it integrates with both shortcode checkout flows and block checkout flows).

PR #3401, opened 2020-11-13
Print wordpress version in e2e tests.

budzanowski:
### Print WordPress core information when executing E2E tests.

This affects output inside Travis log. With this change we can check which version of WordPress was exactly used for tests.

This uses:
https://developer.wordpress.org/cli/commands/core/version/


senadir:
We already pass the version number in a variable WP_VERSION to each Travis task, do you need to know the precise version (patch version maybe)?

senadir (approved):
Happy to approve and merge if this is blocking any work.

PR #3403, opened 2020-11-13
Fix: orders not being placed when paying with an Express payment method from the Cart block

Aljullu:
In #3004 we introduced Express payment methods in the Cart block, but after the payment was processed the order was not placed. This PR fixes that.

In a device compatible with express payment methods and a site with Stripe payment gateway enabled: 
1. Verify Apple Pay or Google Pay buttons appear in the Cart block and clicking on them triggers the payment process.
2. Proceed with the payment and verify you are redirected to the success screen and the order appears in the admin (WooCommerce > Orders).
3. Verify express payment methods from the Checkout block are not broken.
4. Verify regular payment methods are not broken either.

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -147,14 +147,6 @@ class Checkout extends AbstractRoute {
 		$order_controller = new OrderController();
 		$order_object     = $this->get_draft_order_object( $this->get_draft_order_id() );
 
-		// In some cases, the order might not be created yet, that might happen
-		// when trying to checkout directly from the Cart block with an express
-		// payment method. In that case, try creating a draft order first.
-		if ( ! $order_object instanceof \WC_Order ) {
-			$order_object = $this->create_or_update_draft_order();
-		}
-
-		// If there is still no order object, throw an error.
 		if ( ! $order_object instanceof \WC_Order ) {
 			throw new RouteException(
 				'woocommerce_rest_checkout_invalid_order',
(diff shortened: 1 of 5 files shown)


src/StoreApi/Routes/Checkout.php
```diff
@@ -147,6 +147,14 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
 		$order_controller = new OrderController();
 		$order_object     = $this->get_draft_order_object( $this->get_draft_order_id() );
 
+		// In some cases, the order might not be created yet, that might happen
+		// when trying to checkout directly from the Cart block with an express
+		// payment method. In that case, try creating a draft order first.
+		if ( ! $order_object instanceof \WC_Order ) {
+			$order_object = $this->create_or_update_draft_order();
+		}
```
nerrad: I think this is okay but I wonder if we should just modify `get_draft_order_object` so that it returns a new one if there is no draft order id provided. From the doc blocks on the method, it seems like that's what it was intended to do. So basically in `get_draft_order_object` there'd be this:

```php
// old line
// $draft_order_object = $order_id ? wc_get_order( $order_id ) : false;
$draft_order_object = $order_id ? wc_get_order( $order_id ) : $this->create_or_update_draft_order();
```

nerrad (approved):
Thanks Albert for taking care of this. Changes look good here, however I do have a comment on where things could be slightly improved in the Checkout Store API side of things.

- I tested Chrome pay via Cart for both a logged in user and non-logged in user (incognito) and things worked okay.
- I tested Chrome pay via checkout for both logged in user and logged-out user and things worked okay.
- I tested saved payment method for logged in user and things worked okay.
- I tested credit card for logged out user and things worked okay.

I'm approving, but I think it'd be good to get a look see from @mikejolley on this as well given the StoreAPI change. On the surface I don't think it's problematic that we create an order on a checkout request if there isn't a draft order already created but think it'd be good to double-check.

Aljullu:

> I'm approving, but I think it'd be good to get a look see from @mikejolley on this as well given the StoreAPI change.

Agree. :+1:

src/StoreApi/Routes/Checkout.php
Aljullu: In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3403/commits/2577e4be8c3b41278b008e40f6816cc69078b5bf I applied a small refactor in line with your suggestion: making `get_draft_order_object()` return a new order object if no draft order existed. I needed to refactor `create_or_update_draft_order()` so it didn't depend on `get_draft_order_object()`. Also, I couldn't apply your suggestion directly because there is no guarantee the order ID will be of a draft order: the order ID stored in `store_api_draft_order` might be pointing to an order that is no longer a draft if the user has made a previous purchase.

mikejolley (approved):
Changes look good to me, since this is all session based I cannot think of any issues that should arise.

PR #3404, opened 2020-11-15
Support a plain js config argument to payment method registration APIs

### LINKED ISSUE #3064: Change registerPaymentMethod to accept a config object instead of a PaymentMethodConfig instance
In our payment method registration API, `registerPaymentMethod` takes a callback that passes `PaymentMethodConfig` class and expects you to return a new instance of it with your config.

```js
const cashOnDeliveryPaymentMethod = {
	name: PAYMENT_METHOD_NAME,
	label: <Label />,
	content: <Content />,
	edit: <Content />,
	icons: null,
	canMakePayment,
	ariaLabel: label,
};
// Config is PaymentMethodConfig
registerPaymentMethod(
	( Config ) => new Config( cashOnDeliveryPaymentMethod )
);
```

But upon further digging, @nerrad concluded that we don't need to expose `PaymentMethodConfig` to third party integrations, therefor exposing more implementation detail. We can accept a config object and handle the class initiation inside `registerPaymentMethod`.

```js
const cashOnDeliveryPaymentMethod = {
	name: PAYMENT_METHOD_NAME,
	label: <Label />,
	content: <Content />,
	edit: <Content />,
	icons: null,
	canMakePayment,
	ariaLabel: label,
};

registerPaymentMethod( cashOnDeliveryPaymentMethod );
```

So, to do that, we have several steps:

- [ ] Deprecate the old method with a notice that people should pass an object instead.
- [x] Provide a fallback inside `registerPaymentMethod` to accept either an object or a callback.
- [x] Update our internal payment methods to pass an object instead.
- [x] Update the integration docs for third-party developers.
- [x] Do the same for express payment methods.
- [ ] Include a dev-note for this when we ship those changes.

haszari:
Fixes #3064

This PR updates the two APIs for registering payment methods so they are simpler. They now both receive an `options` JS object. Previously was a callback which exposed internal details of how payment methods are handled.

The new interfaces are:

- `registerPaymentMethod( options )`
- `registerExpressPaymentMethod( options )`

The existing callback API is still supported if a function is passed.

All payment methods in this code base have been updated to the new simpler API. Docs have also been updated.

This PR also includes JSDoc typedefs for the `options` objects. This should help us catch any payment methods passing incorrect info, and when we support new fields to options.

#### Deprecating the old API - help!
In the issue it's suggested that we deprecate the old callback-style API. I don't know how we should do this – how do we deprecate JS APIs?

Perhaps we can add a `warn_deprecated_api()` routine that logs to console, possibly with an extra conditional (e.g. `SCRIPT_DEBUG` ?), and call this in the deprecated/legacy code path.

Update: using [`@wordpress/deprecated`](https://developer.wordpress.org/block-editor/packages/packages-deprecated/).

#### API docs
I've manually updated the markdown doc for now. Long term it might be a good idea to start documenting our JS APIs using code docs, for example using [JSDoc](https://jsdoc.app/) or some other tech. Then we can use this to automatically generate consistently formatted API docs, and also keep track of changes over different releases. 

@nerrad - keen for your thoughts on this. If there's interest I'm happy to take a look in the next cooldown.

I haven't looked to see what other projects are doing (e.g. Gutenberg, WooCommerce, wc-admin, Jetpack), maybe there's a convention we can adopt.

Update: using a JSDoc typedef now.

1. Check out this branch and build. 
2. Test all payment methods work correctly - e.g. complete an order with each payment method.

Note I haven't tested the following payment methods myself, would appreciate extra testing of the following. Please comment in your review which payment methods you tested – thanks!

- [ ] Stripe Payment Request - Chrome/Google Pay
- [ ] Stripe Payment Request - Apple Pay
- [ ] PayPal

Bonus points: 

- Test a 3rd-party payment method from another extension (e.g. WC Pay) - using new or old API.
- Hack a payment method in blocks code base to use callback-style registration and ensure fallback works correctly.

### Changelog

> Dev: Payment method registration APIs now receive a plain JavaScript object argument (previously expected a callback function).

##### Payment method registration APIs receive JS options arg (was callback)
In WooCommerce Blocks `{version}` the APIs for registering a payment method have changed.

The affected JavaScript functions are:

- `registerPaymentMethod( options )`
- `registerExpressPaymentMethod( options )`

Previously these APIs received a callback function, for example:

```js
const bankTransferPaymentMethod = {
	name: PAYMENT_METHOD_NAME,
	label: <Label />,
	content: <Content />,
	edit: <Content />,
	icons: null,
	canMakePayment: () => true,
	ariaLabel: label,
};
// OLD API
registerPaymentMethod( ( Config ) => new Config( bankTransferPaymentMethod ) );
```

Now you can simply pass the options object directly: 

```js
// New API
registerPaymentMethod( bankTransferPaymentMethod );
```

All payment methods (gateway extensions) should update to the new API as soon as possible. The previous callback-based API is supported as a fallback, and will be deprecated in a future release.

### DIFF
--- a/docs/extensibility/payment-method-integration.md
+++ b/docs/extensibility/payment-method-integration.md
@@ -8,7 +8,7 @@ The checkout block has an API interface for payment methods to integrate that co
 
 - [Client Side integration](#client-side-integration)
   - [Express payment methods - `registerExpressPaymentMethod( options )`](#express-payment-methods---registerexpresspaymentmethod-paymentmethodcreator-)
-  - [Payment Methods - `registerPaymentMethod( paymentMethodCreator )`](#payment-methods---registerpaymentmethod-paymentmethodcreator-)
+  - [Payment Methods - `registerPaymentMethod( options )`](#payment-methods---registerpaymentmethod-paymentmethodcreator-)
   - [Props Fed to Payment Method Nodes](#props-fed-to-payment-method-nodes)
 - [Server Side Integration](#server-side-integration)
   - [Processing Payment](#processing-payment)
@@ -20,7 +20,7 @@ The client side integration consists of an API for registering both _express_ pa
 
 In both cases, the client side integration is done using registration methods exposed on the `blocks-registry` API. You can access this via the `wc` global in a WooCommerce environment (`wc.wcBlocksRegistry`). You'll see that we also Webpack configured in the blocks repository to expose this API on `@woocommerce/blocks-registry` which you can implement in your own build process as well.
 
-### Express payment methods - `registerExpressPaymentMethod( paymentMethodCreator )`
+### Express payment methods - `registerExpressPaymentMethod( options )`
 
 ![Express Payment Area](https://user-images.githubusercontent.com/1429108/79565636-17fed500-807f-11ea-8e5d-9af32e43b71d.png)
 
@@ -38,12 +38,10 @@ _wc global_
 const { registerExpressPaymentMethod } = wc.wcBlocksRegistry;
 ```
 
-The registry function expects a function that will receive an `ExpressPaymentMethodConfig` creator as an argument and is expected to return a valid `ExpressPaymentMethodConfig` instance. As a very basic example, something like this:
+The registry function expects a javascript object with options specific to the payment method:
 
 ```js
-registerExpressPaymentMethod(
-	( ExpressPaymentMethodConfig ) => new ExpressPaymentMethodConfig( options )
-);
+registerExpressPaymentMethod( options );
 ```
 
 The options you feed the configuration instance should be an object in this shape:
--- a/docs/extensibility/payment-method-integration.md
+++ b/docs/extensibility/payment-method-integration.md
@@ -7,7 +7,7 @@ The checkout block has an API interface for payment methods to integrate that co
 ## Table of Contents <!-- omit in toc -->
 
 - [Client Side integration](#client-side-integration)
-  - [Express payment methods - `registerExpressPaymentMethod( paymentMethodCreator )`](#express-payment-methods---registerexpresspaymentmethod-paymentmethodcreator-)
+  - [Express payment methods - `registerExpressPaymentMethod( options )`](#express-payment-methods---registerexpresspaymentmethod-paymentmethodcreator-)
   - [Payment Methods - `registerPaymentMethod( paymentMethodCreator )`](#payment-methods---registerpaymentmethod-paymentmethodcreator-)
   - [Props Fed to Payment Method Nodes](#props-fed-to-payment-method-nodes)
 - [Server Side Integration](#server-side-integration)
@@ -66,7 +66,7 @@ Here's some more details on the configuration options:
 -   `canMakePayment` (required): A callback to determine whether the payment method should be available as an option for the shopper. The function will be passed an object containing data about the current order. Return a boolean value - true if payment method is available for use. If your gateway needs to perform async initialisation to determine availability, you can return a promise (resolving to boolean). This allows a payment method to be hidden based on the cart, e.g. if the cart has physical/shippable products (example: `Cash on delivery`); or for payment methods to control whether they are available depending on other conditions. Keep in mind this function could be invoked multiple times in the lifecycle of the checkout and thus any expensive logic in the callback provided on this property should be memoized.
 -   `paymentMethodId`: This is the only optional configuration object. The value of this property is what will accompany the checkout processing request to the server and used to identify what payment method gateway class to load for processing the payment (if the shopper selected the gateway). So for instance if this is `stripe`, then `WC_Gateway_Stripe::process_payment` will be invoked for processing the payment.
 
-### Payment Methods - `registerPaymentMethod( paymentMethodCreator )`
+### Payment Methods - `registerPaymentMethod( options )`
 
 ![Payment Method Area](https://user-images.githubusercontent.com/1429108/79565774-5d230700-807f-11ea-9335-0111ec306a47.png)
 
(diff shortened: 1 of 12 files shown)


senadir:
I'm going to do a functionality review for this soon.
However, I don't think we need to create any false sense of urgency. This is a low priority update that doesn't break anything, it's a simple code refactor and we still have total control on both cases because we're exposing the `Config` class and therefor control the implementation of it.

I'd also like us to follow up on [this comment](https://github.com/Automattic/woocommerce-payments/pull/723#issuecomment-727681433) to explain that this refactor is not shipped yet and implementing it will break the integration. 

Also shipping it now would require people to have both latest versions at the same time, while a phase period is better.

We should hold the urgency card until we actually need it (security, something blocking feature work, performance).

RadoslavGeorgiev:
How should the consumers of the `registerPaymentMethod` API know whether to pass a callback or a configuration object as a parameter?

Reading [this comment](https://github.com/Automattic/woocommerce-payments/pull/723#issuecomment-727681433) made me think I'd be able to use the new argument, _or_ the old one as fallback. The only way to do this right now would be to check `woocommerce-gutenberg-products-block`'s version in PHP and send data to JavaScript.

The good thing is that blocks support in WCPay is still an experimental feature, so the risks of updating it to the new API without a fallback are very low. Still, is this change really necessary? The only benefit I, as some who's not really involved in the repo, see is that the registration of the PM looks a bit cleaner.

nerrad:
@RadoslavGeorgiev I think Rua jumped the gun a bit in potentially attributing a sense of urgency to this that isn't necessary. While we will emphasize the new argument used for registration going forward, it's trivial for us to support both argument shapes for some time so WCPay can keep using the older signature until you are comfortably able to switch to the new signature (which may be what Rua meant by "as soon as practical" in his comment in the WC Pay repo).

haszari:
> attributing a sense of urgency to this that isn't necessary

Totally agree, I just wanted to give WCPay a heads up. No need to make the change any time soon, both signatures will be supported for the time being.

docs/extensibility/payment-method-integration.md
```diff
@@ -38,12 +38,10 @@ _wc global_
 const { registerExpressPaymentMethod } = wc.wcBlocksRegistry;
 ```
 
-The registry function expects a function that will receive an `ExpressPaymentMethodConfig` creator as an argument and is expected to return a valid `ExpressPaymentMethodConfig` instance. As a very basic example, something like this:
+The registry function expects a javascript object with options specific to the payment method:
```
nerrad:
```suggestion
The registry function expects a JavaScript object with options specific to the payment method:
```

docs/extensibility/payment-method-integration.md
```diff
@@ -84,19 +82,17 @@ _wc global_
 const { registerPaymentMethod } = wc.wcBlocksRegistry;
 ```
 
-The registry function expects a function that will receive an `PaymentMethodConfig` creator as an argument and is expected to return a valid `PaymentMethodConfig` instance. As a very basic example, something like this:
+The registry function expects a javascript object with options specific to the payment method:
```
nerrad:
```suggestion
The registry function expects a JavaScript object with options specific to the payment method:
```

nerrad (approved):
Looks great Rua! I've indicated a couple minor doc fixes. Also before merging I think it'd be good to add a typedef for the payment method config option (so the passed in object shape is described by the typedef).

I didn't test this - this is a code review only.

Pre-approving - the doc fixes should be applied before merge but the typedef is up to you.

senadir:
It seems I had a comment that got lost

> Deprecating the old API - help!
In the issue it's suggested that we deprecate the old callback-style API. I don't know how we should do this – how do we deprecate JS APIs?
> Perhaps we can add a warn_deprecated_api() routine that logs to console, possibly with an extra conditional (e.g. SCRIPT_DEBUG ?), and call this in the deprecated/legacy code path.

You can use `@wordpress/deprecated`, this is how we deprecated `setSetting` and I think you should use the same.
Just call the function when you detect that the user passed a callback.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2bcda21e3a03cb501e57287024f29a42d8914021/assets/js/settings/shared/set-setting.js#L27-L37

haszari:
Thanks for the reviews @nerrad  @senadir  - great tips!

This is ready to go (though feel free to review/feedback the changes) but **needs testing before merge**. I've tested all regular payment methods but still need a volunteer to test express payment (I can't test this).

> add a typedef for the payment method config option

Great idea, I've implemented this for both types, and also mentioned it in the docs. This will help us notice when we add new fields 🧑‍🍳. Note the type I used for the react elements is a bit fuzzy (`Object`) but is better than nothing, we can improve in future.

> [`@wordpress/deprecated`](https://developer.wordpress.org/block-editor/packages/packages-deprecated/)

TIL, thanks for this. I'm now calling it if the callback is passed. It's a slightly different use case - we're deprecating an arg type, not a whole function, but it does the job. Here's the console message you get if you pass a callback:

Screen Shot 2020-11-17 at 2 24 12 PM

haszari:
I've got Chrome Pay working on my ephemeral test site now. I've tested the "pay now" button shows up in cart and checkout, and clicking it triggers the modal. This test confirms that express payment method registration is working correctly.

Note I can't complete a successful payment with Chrome Pay - fails every time. I see this with the shortcode checkout, so I'm confident this isn't a bug with this PR :)

Screen Shot 2020-11-18 at 2 33 09 PM

We can merge this when someone else has given this a quick test, ideally with a successful express payment.

nerrad:
Tested chrome pay with incognito and it worked fine. I'm not seeing any errors related to the changes in this pr.

I _did_ reproduce what you are reporting with failed chrome pay under certain conditions for a _logged in user_ but was also able to reproduce on trunk so it's unrelated to those changes. It's the end of my day so I'll report the issue tomorrow.

haszari:
Thanks @nerrad! Merging 🚢 

> I did reproduce what you are reporting with failed chrome pay under certain conditions for a logged in user but was also able to reproduce on trunk so it's unrelated to those changes. It's the end of my day so I'll report the issue tomorrow.

I saw that with shortcode checkout too, so possibly unrelated to blocks.

PR #3406, opened 2020-11-16
Use latest ( tracked ) patch releases for E2E tests.

budzanowski:
Test tracked branches for E2E tests.

Using plane `5.X` tags is unpredictable. Using `5.X-branch` looks like a better option.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -97,21 +97,21 @@ jobs:
           script:
               - npm run test:e2e
           env:
-              - WP_VERSION=5.3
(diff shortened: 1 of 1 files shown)


budzanowski:
This fixes the annoying 5.3 vs 5.4-alpha issue. But it introduces alpha on the latest 5.5.x because we are currently in alpha for that release. I think that this may actually be OK - we preemptively check if the new version is not causing issues before the release.

budzanowski:
OK here we go.

Issue #3402, opened 2020-11-13
Express Payment Request API (Stripe) Errors not surfaced


mikejolley:
## Describe the bug
If an Express Payment method encounters an error (from the server side) nothing is exposed to the user in the cart/checkout.

## To reproduce

1. Create a product costing 0.01 and add to the cart. Ensure the total does not exceed 0.30. You may also need to reduce shipping costs.
2. Go to checkout and attempt to pay with Chrome Pay via Stripe.
3. Notice after attempting payment a generic processing error is shown. Network requests in the console will reveal an API error stating that the min order amount is 0.3.

## Expected behavior
There are 2 expectations here;

1. The error from the API needs to be exposed and output at the top of the checkout page.
2. Stripe Express Payment Methods should be disabled if the order total is less than 0.30.

nerrad:
Another case of an error not showing up can be reproduced by:

Configure your WooCommerce -> Settings -> Accounts & Privacy settings like so:

<img width="1157" alt="Image 2020-11-13 at 4 37 07 PM" src="https://user-images.githubusercontent.com/1429108/99123643-92b8b000-25ce-11eb-9f7a-6c7933081831.png">

- Checkout using express payment (eg chrome pay) with an email address matching a user account already on the site.
- The express modal will display an error after 30s and not close and you'll see the following error in the network response (using browser network tab): `registration-error-email-exists` (status code 400).

PR #3410, opened 2020-11-16
Show Express Payment Method Error Notices after Payment Failure

mikejolley:
This PR fixes the error handling side of Express Payment Methods so that API errors are surfaced in the UI. This also prevents multiple-similar error messages stacking if you try multiple times.

This is part of #3402 — leaving 3402 open so the specific Stripe case can be avoided with validation.

### Screenshots

Cart – one wordpress test 2020-11-16 16-54-55

There are 2 cases in #3402 you can try to get an error message to appear. For the min amount case:

1. Have a cart < 0.30 in total.
2. Try to pay via Chrome Pay (stripe)
3. After inputting card details, the paymentRequest will fail. A notice will be shown (see screenshots).
4. If you repeat this test without dismissing the error message, only one notice should be shown.

### Changelog

> Show Express Payment Method Error Notices after Payment Failure

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
@@ -166,7 +166,10 @@ export const useCheckoutSubscriptions = ( {
 			const handlers = eventHandlers.current;
 			let response = { type: responseTypes.SUCCESS };
 			if ( handlers.sourceEvent && isProcessing ) {
-				const { paymentStatus, paymentDetails } = checkoutResponse;
+				const {
+					paymentStatus,
+					paymentDetails,
+				} = checkoutResponse.processingResponse;
 				if ( paymentStatus === responseTypes.SUCCESS ) {
 					completePayment( handlers.sourceEvent );
 				}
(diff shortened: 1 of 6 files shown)


assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
```diff
@@ -166,21 +166,22 @@ export const useCheckoutSubscriptions = ( {
...
-					const paymentResponse = abortPayment(
-						handlers.sourceEvent,
-						paymentDetails?.errorMessage
-					);
+					abortPayment( handlers.sourceEvent );
 					response = {
 						type: responseTypes.ERROR,
-						message: paymentResponse.message,
+						message: paymentDetails?.errorMessage,
+						messageId: noticeContexts.EXPRESS_PAYMENTS, // This will allow only 1 express payment error to show at once.
```
Aljullu: Is this needed considering in `payment-method-data-context.js` you already remove all error notices from `noticeContexts.EXPRESS_PAYMENTS`?

Aljullu (approved):
It tests great and code looks good. I added a comment about a response `messageId` I'm not sure is needed, but approving. :ok_hand:

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-checkout-subscriptions.js
mikejolley: I can try removing *all* under the context instead, but these ones are raised elsewhere by one of the checkout emitters. The reason they don't clear is because no ID is provided, so it gets a random ID based on the value of context.

mikejolley: I just set the ID to context so the dev no longer needs to manually set a messageID. Same effect.

PR #3415, opened 2020-11-17
Stop hidden products from being linked in cart and checkout blocks 

### LINKED ISSUE #3370: Cart and Checkout blocks do not respect a product's catalog visibility status

If a product's catalog visibility is set as "hidden" then the product in the default cart does not link back to that product's page.

To reproduce..

1. Set a product's catalog visibility to "hidden"
image

2. Add that product to the cart... and view the cart block
The core cart will show the product title, but not link to the product's page. The cart block still shows a link:
image

When the product is not visible then there should be no link

opr:
Fixes #3370 

### Screenshots

Screenshot 2020-11-17 at 16 30 11

Screenshot 2020-11-17 at 16 32 14

The "T-shirt with Logo" product is hidden from the catalogue, and therefore its name or image does not link to the product page.

1. Create a page that contains the checkout and cart blocks.
2. Find a product in the dashboard and set its catalogue visibility to hidden.
3. Add that product to your cart.
4. Visit the page with the checkout and cart blocks and ensure the name of the hidden product does not link to the product page in the checkout sidebar and in the line items in the cart.
5. Repeat steps 2-4 with the catalogue visibility set to "Search results only".

### DIFF
--- a/assets/js/base/components/cart-checkout/product-name/index.js
+++ b/assets/js/base/components/cart-checkout/product-name/index.js
@@ -9,8 +9,13 @@ import { decodeEntities } from '@wordpress/html-entities';
  */
 import './style.scss';
 
-const ProductName = ( { name, permalink, disabled = false } ) => {
-	return (
+const ProductName = ( {
+	name,
+	permalink,
+	disabled = false,
+	hasLink = true,
+} ) => {
+	return hasLink ? (
 		// we use tabIndex -1 to prevent the link from being focused, pointer-events
 		// disabled click events, so we get an almost disabled link.
 		<a
@@ -20,6 +25,10 @@ const ProductName = ( { name, permalink, disabled = false } ) => {
 		>
 			{ decodeEntities( name ) }
 		</a>
+	) : (
+		<span className="wc-block-components-product-name">
+			{ decodeEntities( name ) }
+		</span>
 	);
 };
 
--- a/assets/js/base/components/cart-checkout/product-name/index.js
+++ b/assets/js/base/components/cart-checkout/product-name/index.js
@@ -9,26 +9,15 @@ import { decodeEntities } from '@wordpress/html-entities';
  */
 import './style.scss';
 
-const ProductName = ( {
-	name,
-	permalink,
-	disabled = false,
-	hasLink = true,
-} ) => {
-	return hasLink ? (
-		// we use tabIndex -1 to prevent the link from being focused, pointer-events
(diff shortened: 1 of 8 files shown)


assets/js/base/components/cart-checkout/product-name/index.js
```diff
@@ -9,8 +9,13 @@ import { decodeEntities } from '@wordpress/html-entities';
...
-	return (
+const ProductName = ( {
+	name,
+	permalink,
+	disabled = false,
+	hasLink = true,
+} ) => {
+	return hasLink ? (
 		// we use tabIndex -1 to prevent the link from being focused, pointer-events
 		// disabled click events, so we get an almost disabled link.
```
budzanowski: Would using `disabled` not be enough?

assets/js/base/components/cart-checkout/product-name/index.js
```diff
@@ -9,8 +9,13 @@ import { decodeEntities } from '@wordpress/html-entities';
  */
 import './style.scss';
 
-const ProductName = ( { name, permalink, disabled = false } ) => {
-	return (
+const ProductName = ( {
+	name,
+	permalink,
+	disabled = false,
+	hasLink = true,
```
budzanowski: ProductName.propTypes need an update after adding this.

budzanowski:
Left two small comments but it is good otherwise. LGTM.

budzanowski (approved):

assets/js/base/components/cart-checkout/product-name/index.js
opr: It seems `disabled` doesn't stop people clicking on the link, it stops people _tabbing_ to the link but not clicking it. I notice the comment 

```
// we use tabIndex -1 to prevent the link from being focused, pointer-events
// disabled click events, so we get an almost disabled link.
```

but notice also that `pointer-events: none` never gets applied to the `a` here. 

My counter question is: why do we keep the `a` here if it's supposed to be disabled?

I propose that we use the `disabled` prop, so no need to add `hasLink` but keep the logic that outputs a `span` if the link is disabled, or outputs an `a` if it is not.

assets/js/base/components/cart-checkout/product-name/index.js
opr: Won't be needed since we are now using `disabled`.

budzanowski (approved):

PR #3423, opened 2020-11-18
Prevent inappropriate checkout signup if order fails with coupon error

### LINKED ISSUE #3363: Checkout API can create a user account when order fails due to coupon restriction (or other failures)
## Describe the bug
~~This affects stores with no payment methods set up, with checkout block configured to allow sign-up on checkout. If the shopper checks `Create an account` and submits the order, the account is created, even though the order isn't processed because of no valid payment method.~~

There may be similar issues, e.g. if a payment fails.

**Update** Refocusing this bug on a specific case for now. When a restricted coupon is used, an account is created and associated with the draft order. 

## To reproduce
See [comment below for test setup & steps](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3363#issuecomment-728387054).

Previous "No payment methods" steps:

1. Add a checkout block to a published page with Allow shoppers to sign up for a user account during checkout option enabled.
2. In `WooCommerce > Settings > Payments` disable all payment methods.
4. On front end, logged out (or incognito) add something to cart and proceed to checkout.
4. Complete checkout form and select Create an account.
4. Submit.

## Expected behavior
~~No account created unless order succeeds. The order processing and any related logic/entities should only happen if the order succeeds (or preflight checks confirm that the order will succeed.~~

- Account should only be created when associated with a successful order. No account should be created for `checkout-draft` order status.
- Store API should behave predictably and consistently (TBD!) when creating accounts and progressing order state. There may be some differences with legacy / shortcode checkout processing.

haszari:
Fixes #3363

Orders can fail for various reasons, for example if a coupon is rejected. Shoppers can also request an account when submitting an order.

This PR reorders order processing so that accounts are only created if the order succeeds. This prevents accounts getting created for orders that are at `checkout-draft` status.

Specifically, this fixes a use case where a new shopper requests an account and uses an invalid coupon (e.g. after expiry date, or only allowed for specific email addresses). There are possibly other cases affected too.

1. Make a coupon with an email restriction e.g. `*@legit.com`.
2. In logged-out browser/user, add a product to cart, proceed to checkout.
3. Enter coupon code.
4. Enter an invalid email for coupon e.g. `try@sneaky.com`
5. Check `Create an account?` option and complete checkout fields.
6. Submit order.

expected: checkout fail, no account

actual: checkout fail, account signup

Bonus points: test other ways for the order to fail, e.g. coupon expiry date. Looks like `validate_order_before_payment()` is focused on coupons only.

### Changelog

> Fix – Checkout block: Prevent `Create an account` from creating up a user account if the order fails coupon validation.

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -158,6 +158,12 @@ class Checkout extends AbstractRoute {
 		// Ensure order still matches cart.
 		$order_controller->update_order_from_cart( $order_object );
 
+		// If any form fields were posted, update the order.
+		$this->update_order_from_request( $order_object, $request );
+
+		// Check order is valid, including validating any coupons.
+		$order_controller->validate_order_before_payment( $order_object );
+
 		// Create a new user account as necessary.
 		// Note - CreateAccount class includes feature gating logic (i.e. this
 		// may not create an account depending on build).
@@ -173,11 +179,6 @@ class Checkout extends AbstractRoute {
 				$this->handle_error( $error );
 			}
 		}
-		// If any form fields were posted, update the order.
-		$this->update_order_from_request( $order_object, $request );
-
-		// Check order is still valid.
-		$order_controller->validate_order_before_payment( $order_object );
 
 		// Persist customer address data to account.
 		$order_controller->sync_customer_data_with_order( $order_object );


senadir:
This PR reverts this https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3260 🤔

haszari:
>  This PR reverts this

@senadir  Can you expand on this comment? Do you mean this PR is not needed? Or this PR would re-introduce a previous issue?

Can you comment on the original issue - #3363. The order of processing in the API is tricky and there are lots of cases to consider (i.e. new and old bugs).

nerrad:
Tested chrome pay with incognito and it worked fine. I'm not seeing any errors related to the changes in this pr.

I _did_ reproduce what you are reporting with failed chrome pay under certain conditions for a _logged in user_ but was also able to reproduce on trunk so it's unrelated to those changes. It's the end of my day so I'll report the issue tomorrow.

opr:
#### ✅ Ineligible email
I've just tested this as well. It prevents account creation when an ineligible email is used. This works great.

#### ❌ Limited-use coupons
Tested with a single-use coupon being applied to cart A, but without completing checkout. Create a new incognito session and apply the coupon to cart B. Complete check out on cart B. Return to cart A and complete checkout. Checkout fails, but an account is created. 

#### ❌ Expiring coupons
I then tested with expired/expiring coupons and when a coupon expires **after** applying to the cart but **before** completing the checkout, the order completes without the discount being applied and the user is still created. Core behaviour is: Returns user to the checkout page with an error message, no order or account is created.

#### Edge cases (but should still work)

#### ❌ Coupons with minimum spend
If a coupon is modified to have a minimum spend value **after** it has been applied to a cart, then the checkout continues without applying the discount. The account and order are created. Core behaviour is: Returns user to the checkout page with an error message, no order or account is created.

#### ❌ Coupons with product restriction

If a coupon is modified to have a product restriction **after** it has been applied to a cart, then the checkout continues without applying the discount. The account and order created. Core behaviour is: Returns user to the checkout page with an error message, no order or account is created.

haszari:
Thanks for extensively testing coupon functionality @opr ! Good catch!

I think those are all new issues that we weren't aware of. It's probably worth logging them separately (maybe one big issue to start?). Can you log an issue for these coupon bugs? Thanks 🙇 

Let's keep this issue/PR focused on the order of checkout processing & account creation.

senadir:
> @senadir Can you expand on this comment? Do you mean this PR is not needed?

Meaning it does exactly the reverse of that PR, rendering it useless and might cause a regression, the order object will have unsaved data (customer ID) so code hooking into `__experimental_woocommerce_blocks_checkout_order_processed` will not have access to the latest order data.

Given the work https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3454, is this blocked right now? or obsolete?

senadir:
Note that I can't add products from All Products for guests (X-WC-Store-API-Nonce is invalid.)

haszari:
> Given the work #3454, is this blocked right now? or obsolete?

Good point @senadir! Yes, this PR is now superceded by #3454 - I'll close :)

PR #3424, opened 2020-11-18
Improve Stripe (chrome pay) express payment method availability

### LINKED ISSUE #3402: Express Payment Request API (Stripe) Errors not surfaced

mikejolley:
This PR will hide/show Stripe in the cart/checkout Express Payment area based on the value of the cart. Transactions lower than 0.30 cents are blocked.

This required a fix to the `SET_REGISTERED_EXPRESS_PAYMENT_METHODS` reducer becaused it merged, rather than replaced, available payment methods.

Fixes #3402

1. Enable Stripe + Chrome Pay
2. Add an item to the cart costing 0.10
3. No express payment method is visible
4. Increase quantity in cart to 3. Express payment method should then be shown.
5. Reduce the quantity again. The payment method will dissapear.

### Changelog

> Improve Stripe (chrome pay) express payment method availability so it's hidden when the transaction would be too low in value.

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
@@ -65,7 +65,7 @@ const PaymentRequestPaymentMethod = {
 	canMakePayment: ( cartData ) =>
 		paymentRequestAvailable( {
 			currencyCode: cartData?.cartTotals?.currency_code?.toLowerCase(),
-			totalPrice: parseInt( cartData?.cartTotals?.total_price, 10 ),
+			totalPrice: parseInt( cartData?.cartTotals?.total_price || 0, 10 ),
 		} ),
 	paymentMethodId: 'stripe',
 };
--- a/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
@@ -21,7 +21,12 @@ let isStripeInitialized = false,
 
 // Initialise stripe API client and determine if payment method can be used
 // in current environment (e.g. geo + shopper has payment settings configured).
-function paymentRequestAvailable( currencyCode ) {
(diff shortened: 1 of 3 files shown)


assets/js/payment-method-extensions/payment-methods/stripe/payment-request/index.js
```diff
@@ -57,10 +63,10 @@ const PaymentRequestPaymentMethod = {
 	content: <PaymentRequestExpress stripe={ componentStripePromise } />,
 	edit: <ApplePayPreview />,
 	canMakePayment: ( cartData ) =>
-		paymentRequestAvailable(
-			// eslint-disable-next-line camelcase
-			cartData?.cartTotals?.currency_code?.toLowerCase()
-		),
+		paymentRequestAvailable( {
+			currencyCode: cartData?.cartTotals?.currency_code?.toLowerCase(),
+			totalPrice: parseInt( cartData?.cartTotals?.total_price, 10 ),
```
budzanowski: Since we get `total_price` using the optional chaining operator we assume that we can get `undefined` as a result. Undefined passed to `parseInt` will get us `NaN`. When we plug the `NaN` to the test used on line 26  `if ( totalPrice < 30 )` we get `False` and continue but we should stop.

mikejolley: Good catch. I committed a fallback to `|| 0`.

budzanowski (approved):

Issue #2751, opened 2020-06-18
Display account login form inline with the Checkout Block


nerrad:
Currently the Checkout block has a Login link exposed if the store is configured to offer login for Customers. However, it’s behaviour is different than the shortcode checkout flow. The shortcode checkout flow allows for inline logging in without leaving the checkout.

I think it will be better if we allow for inline logging in as well.

This will require some design input (and prioritization) before tackling.

In the meantime, it is worth noting that while the current login flow does take the user to the native WordPress login form, it is configured to redirect back to the checkout after the shopper logs in.

haszari:
At face value, this feels like an opportunity for a new block. Either a login block, or a login/signup combo block.

I could imagine that this would be a useful block that merchants could use anywhere they want the user to log in, and then we'd use the same code for rendering a login form as a fallback for checkout when needed. Possibly could use a similar approach as empty cart's `InnerBlocks`.

Just mentioning that in case we have a login block or something similar later in the roadmap.

PR #3425, opened 2020-11-18
Prevent checkout step heading text overlapping actual heading on small viewports

### LINKED ISSUE #3082: 'Already have an account' text overflows 'Contact info' title
## To reproduce
1. Go to the Checkout block with a small viewport.
2. Notice how the 'Already have an account? Log in.' text overflows the 'Contact info' title.

## Expected behavior
Texts shouldn't collide.

## Screenshots
imatge

opr:
This change stops the `wc-block-components-checkout-step__heading-content` class being positioned absolutely, which caused it to overlap the checkout step title on smaller viewports. The heading content now drops below the heading when the text will overlap. It also adds the same grey line to the left of the heading content as on the checkout step description.

Fixes #3082

⚠️ Because Safari does not support the CSS property `gap` the text sits very close to the heading. There is not an easy way to set the margins on this text without specifically targeting Safari, however `gap` will be supported in the next version of Safari according to https://caniuse.com/flexbox-gap

#### Before
image

#### After
Screenshot 2020-11-18 at 11 56 56

#### In Safari
Screenshot 2020-11-18 at 12 05 09

1. Ensure the Allow customers to log into an existing account during checkout option is enabled in **WooCommerce -> Settings -> Accounts & Privacy**.
2. Log out of the site and add a product to your cart.
3. Go to a page with the checkout block on and reduce the viewport so far that the "Already have an account? Log in" text is about to overlap.
4. Ensure the text drops to below the header and does not overlap with the section heading.

### DIFF
--- a/assets/js/base/components/cart-checkout/form-step/style.scss
+++ b/assets/js/base/components/cart-checkout/form-step/style.scss
@@ -57,7 +57,6 @@
 
 .wc-block-components-checkout-step__heading-content {
 	@include font-size(smaller);
-	position: relative;
 
 	a {
 		font-weight: bold;
--- a/assets/js/base/components/cart-checkout/form-step/style.scss
+++ b/assets/js/base/components/cart-checkout/form-step/style.scss
@@ -12,6 +12,16 @@
 	.is-large & {
 		padding-right: $gap-large;
 	}
+
+	.wc-block-components-checkout-step__heading::after {
+		content: "";
+		border-left: 1px solid;
+		opacity: 0.3;
+		position: absolute;
+		left: -$gap-larger/2;
+		top: 2.5em;
+		bottom: em($gap) * -1;
+	}
 }
 
 .wc-block-components-checkout-step__container {
@@ -33,6 +43,8 @@
 	flex-wrap: wrap;
 	margin: em($gap-small) 0 em($gap);
 	position: relative;
+	align-items: center;
+	gap: em($gap);
 }
 
 .wc-block-components-checkout-step:first-child .wc-block-components-checkout-step__heading {
@@ -45,8 +57,7 @@
 
 .wc-block-components-checkout-step__heading-content {
 	@include font-size(smaller);
-	position: absolute;
-	right: 0;
+	position: relative;
 
 	a {
 		font-weight: bold;


assets/js/base/components/cart-checkout/form-step/style.scss
```diff
@@ -12,6 +12,16 @@
 	.is-large & {
 		padding-right: $gap-large;
 	}
+
+	.wc-block-components-checkout-step__heading::after {
```
mikejolley, line 16: I wasn't sure what this was for but I see it's for the border. This part doesn't work in Safari—expected?

assets/js/base/components/cart-checkout/form-step/style.scss
```diff
@@ -45,8 +57,7 @@
 
 .wc-block-components-checkout-step__heading-content {
 	@include font-size(smaller);
-	position: absolute;
-	right: 0;
+	position: relative;
```
mikejolley: Is position needed at all?

mikejolley (approved):
👍🏻  I tested this and it worked as described.

As you said, Safari doesn't support this so it looks like this:
Checkout – one wordpress test 2020-11-19 11-26-27

It's still acceptable and better than overlap through :)

Firefox also looked ok: 

Checkout – one wordpress test 2020-11-19 11-28-22

Curious as to why Chrome shows these fonts much larger than Safari and FF...

Anyhow, added a small note but seems good to merge.

As an aside (@nerrad) this link could be removed entirely if we prioritised https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2751-- the link text could go below email and trigger a password field to appear so it's all inline. Would be a better experience :)

nerrad:
> As an aside (@nerrad) this link could be removed entirely if we prioritised #2751 the link text could go below email and trigger a password field to appear so it's all inline. Would be a better experience :)

Agreed. It should surface in one of the weekly triages product and myself are doing.

assets/js/base/components/cart-checkout/form-step/style.scss
opr: This creates a short grey line at the bottom of the `step__heading` - it _is_ present in Safari, but because `gap` is not supported it does not line up well, so not immediately obvious that it is there. I think we opt to leave this in, because when Safari supports `gap` it'll line up automatically and have the grey border next to the "Already have an account" text as it does in Chrome and Firefox.

assets/js/base/components/cart-checkout/form-step/style.scss
opr: Good catch, no it isn't!

opr:
@mikejolley The font sizes are the same for me on Chrome and Firefox, although it looks like Firefox renders the text a little bit heavier. This is in TwentyTwenty theme.

### Chrome
chrome

### Firefox
ff

PR #3427, opened 2020-11-18
Ensure "Add a note to your order" section is styled correctly when disabled

### LINKED ISSUE #3426: Add a note to your order form field does not get disabled when submitting checkout form
## Describe the bug
When the checkout form is submitted all form fields get disabled apart from "Add a note to your order". This form field should also become disabled.

## To reproduce
1. Add an item to the cart
2. Go to a page with the checkout block
3. Fill in the form with dummy data
4. Submit the form and watch all fields become disabled, notice "Add a note to your order" does not.

## Expected behavior
The "Add a note to your order" field should become disabled when the form is submitted.

## Screenshots
Screenshot 2020-11-18 at 12 19 53

opr:
When the checkout form is submitted, the fields in the address fieldset are disabled. They have an opacity set to give a visual indication of this. The "Add a note to your order" section does not have this same visual indication. (Though the form fields within do get the disabled attribute)

Fixes #3426 

#### After
Screenshot 2020-11-18 at 15 04 12

1. Add a product to your cart and go to a page with the checkout block
2. Fill in the required form fields. Leave the "Add a note to your order" unticked.
3. Submit the form
4. Ensure the "Add a note to your order" section looks disabled
5. Restart the checkout process and fill in all required form fields.
5. Tick the box next to "Add a note to your order" and (optionally) fill in the textarea with something
6. Submit the form again and ensure the entire "Add a note to your order" section is disabled, including the textarea.

### DIFF
--- a/assets/js/base/components/cart-checkout/form-step/style.scss
+++ b/assets/js/base/components/cart-checkout/form-step/style.scss
@@ -14,6 +14,10 @@
 	}
 }
 
+.wc-block-components-checkout-step--disabled {
+	opacity: 0.6;
+}
+
 .wc-block-components-checkout-step__container {
 	position: relative;
 }
--- a/assets/js/base/components/cart-checkout/form-step/style.scss
+++ b/assets/js/base/components/cart-checkout/form-step/style.scss
@@ -26,10 +26,6 @@
 	padding-bottom: em($gap-large);
 }
 
-.wc-block-components-form .wc-block-components-checkout-step:disabled {
-	opacity: 0.6;
-}
-
 .wc-block-components-checkout-step__heading {
 	display: flex;
 	justify-content: space-between;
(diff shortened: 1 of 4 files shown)


assets/js/base/components/cart-checkout/form-step/style.scss
```diff
@@ -14,6 +14,10 @@
 	}
 }
 
+.wc-block-components-checkout-step--disabled {
```
Aljullu, line 17: With this new selector, do you think we can remove the one in [line 29](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fix/3426-disable-add-note-field/assets/js/base/components/cart-checkout/form-step/style.scss#L29)?

Aljullu (approved):
LGTM, good job! :rocket:  I added only one comment about removing a CSS selector that I think is now redundant, but everything else looks good so pre-approving.

assets/js/base/components/cart-checkout/form-step/style.scss
opr: You're right - it's no longer needed. Removed.

PR #3429, opened 2020-11-19
Handle change to logged-in status in checkout API AJAX requests

haszari:
Related #3363 #3423

The checkout API can sometimes create a new user account and log that user in – the `Create an account?` option.

If there are other errors that prevent checkout from succeeding (e.g. invalid coupon - see #3363), this can leave the checkout in a slightly broken state. The core of the issue is that the checkout block and state isn't updated to reflect the fact that the user is logged in. Also the API response includes a stale nonce, because the nonce is generated early in request handling, before account creation.

The goal of this PR is to fix these issues, and generally make the checkout UX robust to user login each time the checkout is submitted (i.e. after any AJAX request).

- Ensure the response includes a valid nonce.
- Trigger changes to state/props so the UX updates appropriately. For example:
  - Hide the `Create an account?` checkbox if user is now logged in.
  - Show a notice informing the user that their account was created, even though the order submission failed.

This PR is currently based off #3454. They are closely related; this PR depends on the updated nonce from #3454 when a new account is created. When #3454 is merged, this should be rebased to trunk (and retested).

1. Make a coupon with an email restriction e.g. `*@legit.com`.
1. In logged-out browser/user, add a product to cart, proceed to checkout.
1. Enter coupon code.
1. Enter an invalid email for coupon e.g. `try@sneaky.com`
1. Check `Create an account?` option and complete checkout fields.
1. Submit checkout.
1. Confirm that checkout shows appropriate feedback:
  - Invalid coupon has been removed.
  - New user account created - `Create an account` checkbox is gone + notice.
1. Submit checkout again - should succeed.

### "New account" notice

Note - I couldn't get a regular notice working, it seems to only display one at a time. Snackbar style works, and maybe is a good fit for this notice. @LevinMedia can you review this notice UX and wording and advise? I don't think the notice is essential (could be handled as an improvement later), so we could back out of this if needed.

Here are some ideas for wording for the notice:

  - `Congratulations! You have signed up for a user account.`
  - `Welcome! Your user account has been created.`
  - variations of the above with email / username

### Screenshot walkthrough

Screen Capture on 2020-11-27 at 14-05-21

### Changelog

> Fix – Checkout block: Correctly handle cases where the order fails with an error (e.g. invalid coupon) and a new user account is created.

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout-state/actions.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/actions.js
@@ -16,6 +16,7 @@ const {
 	SET_NO_ERROR,
 	INCREMENT_CALCULATING,
 	DECREMENT_CALCULATING,
+	SET_CUSTOMER_ID,
 	SET_ORDER_ID,
 	SET_SHOULD_CREATE_ACCOUNT,
 	SET_ORDER_NOTES,
@@ -61,6 +62,10 @@ export const actions = {
 	} ),
 	decrementCalculating: () => ( {
 		type: DECREMENT_CALCULATING,
+	} ),
+	setCustomerId: ( customerId ) => ( {
+		type: SET_CUSTOMER_ID,
+		customerId,
 	} ),
 	setOrderId: ( orderId ) => ( {
 		type: SET_ORDER_ID,
--- a/assets/js/base/context/cart-checkout/checkout-state/constants.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/constants.js
@@ -43,6 +43,7 @@ export const TYPES = {
 	SET_PROCESSING: 'set_checkout_is_processing',
 	SET_HAS_ERROR: 'set_checkout_has_error',
 	SET_NO_ERROR: 'set_checkout_no_error',
+	SET_CUSTOMER_ID: 'set_checkout_customer_id',
 	SET_ORDER_ID: 'set_checkout_order_id',
 	SET_ORDER_NOTES: 'set_checkout_order_notes',
 	INCREMENT_CALCULATING: 'increment_calculating',
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -61,6 +61,7 @@ const CheckoutContext = createContext( {
 		setAfterProcessing: ( response ) => void response,
 		incrementCalculating: () => void null,
 		decrementCalculating: () => void null,
+		setCustomerId: ( id ) => void id,
 		setOrderId: ( id ) => void id,
 		setOrderNotes: ( orderNotes ) => void orderNotes,
 	},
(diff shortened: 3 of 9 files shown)


haszari:
> Hide the `Create an account?` checkbox if user is now logged in.

Started looking at this and got stuck. Here's what's happening:

- Shopper adds an invalid coupon to checkout. 
- Shopper checks `Create an account?` = YES.
- Shopper submits order - `POST /checkout`.
  - `Checkout::get_route_post_response()`:
  - Server creates user account, logs user in. (`$create_account->from_order_request()`)
  - Coupon fails validation, error thrown, error response returned. (`OrderController::validate_order_before_payment()`)

(Note I haven't pushed any code related to this.)

I think what we need to do is return the new customer user ID in the checkout response, alongside the coupon error. But that seems problematic - it's a sign that the request has potential side effects, and this complicates response handling.

FYI @nerrad @mikejolley for thoughts - maybe we need to circle back to the issue and decide how best to handle this (my initial attempt doesn't feel like a good solution).

mikejolley:
> I think what we need to do is return the new customer user ID in the checkout response, alongside the coupon error. But that seems problematic - it's a sign that the request has potential side effects, and this complicates response handling.

I think this sounds like the thing to do; it would be handled kind of like the cart updates. Once the customer ID is added in state/context, the rest of the UI can react and hide things like login prompts etc.

haszari:
> > I think what we need to do is return the new customer user ID in the checkout response, alongside the coupon error. But that seems problematic - it's a sign that the request has potential side effects, and this complicates response handling.
> I think this sounds like the thing to do; it would be handled kind of like the cart updates. Once the customer ID is added in state/context, the rest of the UI can react and hide things like login prompts etc.

Thanks for the hint @mikejolley – I can see now that [cart API does something similar](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/StoreApi/Schemas/CartSchema.php#L319), and checkout supports [updating the cart alongside an error response](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/cart-checkout/checkout/processor/index.js#L244).

I've got a rough prototype working in this PR now, but there's a bit of work to do to get it polished. Feel free to give me an early code review if you like. 

The challenge for the checkout API (as I see it) is that the current logic allows anything to throw a `RouteException` to stop processing and return an error. Now we need to change things so there's one main processing path (e.g. `get_route_update_response()`), which checks for errors along the way, and controls whether subsequent processing is skipped. I've currently special-cased this for coupons, but if we like this approach I think we might need to generalise so all the other sub-routines return errors instead of throwing. Totally open to ideas for how best to organise this. The crux of it is that the state we need to return will be built up along the way in `get_route_update_response()`, and if an error happens, we need to return the state we have, and the error info.

haszari:
Rebased and retargeted this off #3454 for now, so this can use the customer ID and nonce returned with error responses.

mikejolley:
This PR introduced a snack bar notice and "errors" array in responses (unused?). I'm removing this code so as not to block it being merged. So long as we ensure nonces and user IDs are updated in the client, the checkout can react to these changes accordingly.

mikejolley (approved):
I've removed the code that wasn't needed and swapped the user_id in body for the new header. Tested with a coupon that failed after account creation and it showed the notice, updated the checkout view (hiding the account signup box), and updated nonces. Resubmission worked as normal.

PR #3430, opened 2020-11-19
Refresh PaymentRequest after cancelling payment to prevent addresses remaining populated on repeat attempts

### LINKED ISSUE #3414: Stripe/PaymentRequest pre populated address bug on repeat attempts
## Describe the bug
When using Stripe and Chrome Pay, if you attempt payment more than once you can circumvent shipping rate updates. This can lead to incorrect totals + shipping costs being displayed.

## To reproduce
You need to be using Stripe + Chrome Pay to see this issue, and you need multiple addresses setup in Chrome.

1. Go to the cart page with items in your cart.
2. Click on Pay Now (stripe Chrome pay)
3. The PaymentRequest modal is opened.
4. In the PaymentRequest window, select a delivery address. Choose the second address in the list, not the first address listed.
5. After choosing this address you'll see some updates on the page as a result of address selection.
6. _**Cancel the payment.**_
7. Click the Pay Now button again.
8. Notice the PaymentRequest modal is populated with a different delivery address and there was no API update triggered to update shipping address.

This can be a problem if you have addresses setup in different countries, and different shipping rates per country, because the totals are not updated.

## Expected behavior
The address should not be pre-populated when opening the PaymentRequest modal.

mikejolley:
Currently when using Stripe and Chrome Pay, a single PaymentRequest object is created when entering the cart and checkout page. From this point on the PaymentRequest is just updated as certain events happen.

The problem we see is that after closing and re-opening the PaymentRequest window, the address gets prepopulated on reattempts. Since the address is prepopulated without an address change event, we cannot capture this and thus show incorrect shipping prices.

To fix this, we need to destroy and recreate the PaymentRequest after the user clicks cancel.

Fixes #3414 

You need to be using Stripe + Chrome Pay to see this issue, and you need multiple addresses setup in Chrome.

- Go to the cart page with items in your cart.
- Click on Pay Now (stripe Chrome pay). The PaymentRequest window is opened.
- In the PaymentRequest window, select a delivery address.
- Cancel the payment.
- Click the Pay Now button again.
- The delivery address area should not have an address selected. This is the fix.

### Changelog

> Refresh PaymentRequest after cancelling payment to prevent addresses remaining populated on repeat attempts

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-initialization.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-initialization.js
@@ -43,11 +43,8 @@ export const useInitialization = ( {
 	const [ isFinished, setIsFinished ] = useState( false );
 	const [ isProcessing, setIsProcessing ] = useState( false );
 	const [ canMakePayment, setCanMakePayment ] = useState( false );
-
-	const currentPaymentRequest = useRef( paymentRequest );
-	const currentPaymentRequestType = useRef( '' );
+	const [ paymentRequestType, setPaymentRequestType ] = useState( '' );
 	const currentShipping = useRef( shippingData );
-
 	const {
 		paymentRequestEventHandlers,
 		clearPaymentRequestEventHandler,
@@ -56,41 +53,30 @@ export const useInitialization = ( {
 
 	// Update refs when any change.
 	useEffect( () => {
-		currentPaymentRequest.current = paymentRequest;
 		currentShipping.current = shippingData;
-	}, [ paymentRequest, shippingData ] );
+	}, [ shippingData ] );
 
-	// set paymentRequest.
+	// Create the initial paymentRequest object. Note, we can't do anything if stripe isn't available yet or we have zero total.
 	useEffect( () => {
-		// can't do anything if stripe isn't available yet or we have zero total.
 		if ( ! stripe || ! billing.cartTotal.value ) {
 			return;
 		}
-
-		// if payment request hasn't been set yet then set it.
-		if ( ! currentPaymentRequest.current && ! isFinished ) {
-			setPaymentRequest(
-				getPaymentRequest( {
-					total: billing.cartTotal,
-					currencyCode: billing.currency.code.toLowerCase(),
-					countryCode: getSetting( 'baseLocation', {} )?.country,
-					shippingRequired: shippingData.needsShipping,
-					cartTotalItems: billing.cartTotalItems,
-					stripe,
-				} )
-			);
+		if ( isFinished || isProcessing || paymentRequest ) {
+			return;
 		}
-		// otherwise we just update it (but only if payment processing hasn't
-		// already started).
-		if ( ! isProcessing && currentPaymentRequest.current && ! isFinished ) {
-			updatePaymentRequest( {
-				// @ts-ignore
-				paymentRequest: currentPaymentRequest.current,
-				total: billing.cartTotal,
-				currencyCode: billing.currency.code.toLowerCase(),
-				cartTotalItems: billing.cartTotalItems,
-			} );
-		}
+		const pr = getPaymentRequest( {
+			total: billing.cartTotal,
+			currencyCode: billing.currency.code.toLowerCase(),
+			countryCode: getSetting( 'baseLocation', {} )?.country,
+			shippingRequired: shippingData.needsShipping,
+			cartTotalItems: billing.cartTotalItems,
+			stripe,
+		} );
+		canDoPaymentRequest( pr ).then( ( result ) => {
+			setPaymentRequest( pr );
+			setPaymentRequestType( result.requestType || '' );
+			setCanMakePayment( result.canPay );
+		} );
 	}, [
 		billing.cartTotal,
 		billing.currency.code,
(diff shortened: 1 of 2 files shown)


assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-initialization.js
```diff
@@ -56,41 +53,30 @@ export const useInitialization = ( {
...
-		if ( ! isProcessing && currentPaymentRequest.current && ! isFinished ) {
-			updatePaymentRequest( {
-				// @ts-ignore
-				paymentRequest: currentPaymentRequest.current,
-				total: billing.cartTotal,
-				currencyCode: billing.currency.code.toLowerCase(),
-				cartTotalItems: billing.cartTotalItems,
-			} );
+		if ( isFinished || isProcessing || paymentRequest ) {
+			return;
```
nerrad: probably can just merge all these conditions with the previous ones?

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-initialization.js
```diff
@@ -134,10 +122,15 @@ export const useInitialization = ( {
...
-		const shippingAddressChangeHandler = ( event ) => {
+	const cancelHandler = useCallback( () => {
+		setIsFinished( false );
+		setIsProcessing( false );
+		setPaymentRequest( null );
+		onClose();
+	}, [ onClose ] );
+
+	const shippingAddressChangeHandler = useCallback(
+		( event ) => {
```
nerrad: Moving these functions outside of the effect and implementing as callbacks increases the overhead because instead of React tracking one cache it's tracking 5 (so impacts performance). Was this done to reduce the amount of code in the effect or was there some other reason?

assets/js/payment-method-extensions/payment-methods/stripe/stripe-utils/normalize.js
```diff
@@ -20,13 +20,17 @@
...
-	return cartTotalItems.map( ( cartTotalItem ) => {
-		return {
-			amount: cartTotalItem.value,
-			label: cartTotalItem.label,
-			pending,
-		};
-	} );
+	return cartTotalItems
+		.map( ( cartTotalItem ) => {
+			return cartTotalItem.value
```
haszari, line 25: Just checking - the intention here is to only keep line items that are > $0?

Is this related to the issue, or just a good fix to include alongside the bug fix?

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-initialization.js
```diff
@@ -56,41 +53,30 @@ export const useInitialization = ( {
...
-				// @ts-ignore
-				paymentRequest: currentPaymentRequest.current,
-				total: billing.cartTotal,
-				currencyCode: billing.currency.code.toLowerCase(),
-				cartTotalItems: billing.cartTotalItems,
-			} );
+		if ( isFinished || isProcessing || paymentRequest ) {
+			return;
 		}
+		const pr = getPaymentRequest( {
```
haszari, line 70: This new approach of recreating the request object seems much cleaner and simpler IMO, nice idea 👍

haszari (approved):
Code looks reasonable to me, I do prefer the simpler "recreate every time" tactic.

Tested on ephemeral, in Chrome. I tested both cart and checkout, switching between two different addresses within the modal and cancelling (multiple times) then eventually proceeding payment. Didn't see any problems – both orders worked correctly, and the shipping/totals correctly updated in the background when switching address via express modal. 

🚢 😃

nerrad (commented):
Nice cleanup of code here while working on this. 

Changes generally look okay except for the the one concern I noted around the extracting of functions declared in the `useEffect` to outside using `useCallback` instead.

I tested via express payment on the cart and this fixes the issue as described.  However, I do think testing notes should be expanded to cover:

- testing express payment method in the checkout
- testing Apple Pay. I recall there being some subtle differences between how Apple Pay works with Chrome that resulted in some logic changes to account for things (but not what logic that was or what the differences were unfortunately), so to be on the safe side, with this amount of change I think we should explicitly cover AP with the testing too.

I didn't test either of the above in this pull request but if you need help with AP I can try to do that. However it might be a few days before I can get to testing it.

assets/js/payment-method-extensions/payment-methods/stripe/payment-request/use-initialization.js
mikejolley: Some were inside and some were outside. I moved them outside to make that effect more readable but I can merge them back inside.

mikejolley:
@nerrad If you could test Apple Pay that would be great. Stipe docs indicate they should work in the same way (since the API is abstracted by Stipe) but we shall see :)

assets/js/payment-method-extensions/payment-methods/stripe/stripe-utils/normalize.js
mikejolley: This was a bug I found whilst implementing this fix - no shipping = NaN cost, which causes an error in Stripe.

nerrad:
Can confirm Apple Pay works 👍

PR #3434, opened 2020-11-20
update to npm 7

senadir:
This PR updates npm to 7 so we can use workspaces.

I also updated a couple of dependencies to bypass some errors.

I can't run e2e locally due to `nodegit` not being available, so I'm testing here.

Will add more testing instruction once I get e2e to work.

## What works

- [x] `npm install`.
- [x] `npm run build`.
- [x] `npm start`.
- [x] `npm run test`.
- [x] smoke testing.
- [ ] `npm run test:e2e`.
- [ ] `npm run wp-env`.


senadir:
It seems we also need to force our actions to use npm 7 cc @nerrad do you have a pointer for this?

nerrad:
For node we just have to remove this action in our workflows:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f775a622ec66bd3c401edbb5bb5893ed1940df97/.github/workflows/bundle-size.yml#L9-L12

For npm 7, since it's not lts, we'll probably just have to install it directly by adding the following to any workflows running npm install as a step:

```bash
npm install -g npm@7
```

senadir:
I updated that section to node 14.
Will add the npm one next until `setup-node` supports it
https://github.com/actions/setup-node/issues/213

If we remove it, won't break our workflow? why did it add at first?

nerrad:
> If we remove it, won't break our workflow? why did it add at first?

I added it because github actions by _default_ now install node 14 (because it's now LTS) and we weren't ready to update to node 14 yet :)

nerrad:
You can make changes to the workflows in this pull only.

PR #3436, opened 2020-11-20
Hide spinner on cart block's "Proceed to Checkout" link when page unloads

### LINKED ISSUE #3196: When using browser back button to return from checkout to cart, the loader remains visible on the Checkout CTA (Safari/Mac)
## Describe the bug
When using browser back button to return from checkout to cart, the loader remains visible on the Checkout CTA. It doesn't happen if you use the 'return to cart' CTA in the form.
This happens on Safari/Mac (desktop). The problem doesn't seem to exist on Firefox/Mac (desktop).

## To reproduce
1. Navigate from cart to checkout on Safari-Mac-desktop, 
2. then use the browser's back button (or swipe gesture) to navigate back to the cart
3. Observe the loader on the button, which doesn't go away (see screenshot)

## Expected behavior
The default state of the button.

## Screenshots
Screenshot 2020-09-25 at 15 02 35

* Core version: 4.5.2
* WooCommerce version: 4.5.2
* WooCommerce Blocks version: 3.4.0
* Site language:

* OS: macOS 10.14.6
* Browser safari
* Version 14.0

* Device: [e.g. iPhone6]
* OS: [e.g. iOS8.1]
* Version [e.g. 22]

opr:
This is required because of a feature of Safari where the page state is saved, including all class names, when a transition occurs. Navigating back to this page using the back button restores the page to that cached state, so in the case of the **Cart -> Checkout -> Back to cart** the spinner class remains on the "Proceed to Checkout" link on the cart block. Resetting the state so as to remove the class just before the page gets cached stops this from happening.

Fixes #3196

*Please test in Safari especially*
1. Add an item to the cart.
2. Visit a page with the cart block on.
3. Press the "Proceed to checkout" link.
4. A spinner should appear on the link.
5. When on the checkout page, use the browser's back button to get back to the page with the cart block.
6. When back on the cart block, ensure the "Proceed to checkout" link does not contain a spinner.

### Changelog

> Fix an issue where the proceed to cart button would show a spinner in Safari when navigating back to the page using the browser's back button.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/checkout-button/index.js
+++ b/assets/js/blocks/cart-checkout/cart/checkout-button/index.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useState } from '@wordpress/element';
+import { useState, useEffect } from '@wordpress/element';
 import { PaymentMethodIcons } from '@woocommerce/base-components/cart-checkout';
 import Button from '@woocommerce/base-components/button';
 import { CHECKOUT_URL } from '@woocommerce/block-settings';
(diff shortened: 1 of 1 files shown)


assets/js/blocks/cart-checkout/cart/checkout-button/index.js
```diff
@@ -41,6 +41,30 @@ const CheckoutButton = ( { link } ) => {
 	const [ showSpinner, setShowSpinner ] = useState( false );
 	const { paymentMethods } = usePaymentMethods();
 
+	useEffect( () => {
+		//add a listener for when the page is unloaded (specifically needed for Safari)
```
haszari: We typically use sentence case for our comments. Nitpick, not a blocker :)

haszari (approved):
Tested and confirmed this fixes the issue.

I agree with others that this seems like a lot of extra complexity to fix a caching issue in a specific browser. 

The unfortunate thing from my point of view is that I've never noticed the spinner on clicking `Proceed to Checkout`, and when testing for this PR, I only ever saw the spinner when clicking back (when testing on `trunk`). That is to say, I'm not sure how useful it is to have a spinner as feedback for navigating forward, especially if there's no harm in clicking the button twice. On the other hand, my machine and my site / network are fast … so maybe I'm not the target user. 

Are we considering adding this as a feature/fix to all our buttons? If the spinner (`.wc-block-components-button__spinner`) is used in other buttons then perhaps it's worth implementing this by default in our `Button`.

Note – I'm happy for this to merge (without changes) as think the above issues have been discussed & resolved by the team.

PR #3439, opened 2020-11-23
Update checkout payment methods design.

### LINKED ISSUE #2628: Checkout: update design of the payment methods

budzanowski:
This PR implements the updated checkout payments UX.

Fixes #2628 

Latest design available at pca54o-wS-p2 n90hX6hG3wwQbYwsYsrJnj-fi-4703%3A43711

### Screenshots

image

image

To test this one needs to use a combination of available and saved payment methods:

No saved payment methods.
One saved payment methods.
Multiple saved payment methods.

No payment methods.
One payment method.
Multiple saved payment methods.

The test should consist of checking the UI using combinations of saved and non-saved payment methods.
The other part of the test is ensuring that the payment methods still work.

### Changelog

> Update checkout block payment methods UI.


haszari:
Gave this some early testing, this is looking good and working well – I really like the new radio button UI. I found a few gaps / tweaks, I'm guessing most of these are already on your radar.

- [ ] Optimised / alternative simpler UI when there's one or two payment methods.
  - [x] No saved cards, one method - show method UI inline, no selector needed.
  - [x] No saved cards, two methods - show methods using tab UI (similar to current UI). 6b55b1011d7a9cf82becb466042d597afe06ee06
- [x] When the checkout loads there should be a payment method selected by default. In general, a payment method should always be selected. 5381cfc1828afb61a4b5fed5d61f8788863f98f1
   - Including if any payment methods disappear at runtime. This can happen with COD, which can be configured to only be available for specific shipping methods.
- [x] Full stop at end of payment method copy, e.g. `Use another payment method.`. 21d4532c2638b0e120669cb69fb28d93443d140a

I mostly tested in Safari and Firefox, with a range of payment methods, saved cards (or none) and window sizes (responsive modes seem to work ok). Tested saving payment methods to account, and as anon or logged-in user. Didn't see any payment issues, apart from being able to submit with no payment method selected 😄 

Will need to do another test pass on the styling but it's all looking very much like the designs and working well 🚀

assets/js/base/components/payment-methods/payment-methods.js
```diff
@@ -18,14 +21,24 @@ import SavedPaymentMethodOptions from './saved-payment-method-options';
  */
 const PaymentMethods = () => {
 	const { isInitialized, paymentMethods } = usePaymentMethods();
-	const [ showNewPaymentMethods, setShowNewPaymentMethods ] = useState(
-		true
-	);
-	const onChange = useCallback(
+	const [ savedMethodsToken, setSavedMethodsToken ] = useState( '' );
+	const [ paymentMethodsName, setPaymentMethodsName ] = useState( '' );
```
haszari: I don't think you'll need state for these. The payment method context should handle all this (and provide it to the component). Take a look at how `usePaymentMethodDataContext` and `usePaymentMethodInterface` hooks are used in `trunk` to get and change the active payment method.

assets/js/base/components/payment-methods/payment-method-options.js
```diff
@@ -42,50 +44,41 @@ const PaymentMethodOptions = () => {
...
-			} ) }
-			initialTabName={ activePaymentMethod }
-			ariaLabel={ __(
-				'Payment Methods',
-				'woo-gutenberg-products-block'
-			) }
-			id="wc-block-payment-methods"
+		<RadioControlAccordion
+			id={ 'wc-payment-method-options' }
+			selected={ selectedMethod }
```
haszari: You'll likely want to use `activePaymentMethod` here (ultimately powered by context - `PaymentMethodDataContext`). 

That gives you the defaulting behaviour (ensure a payment method is selected at all times). 

I hacked this change in my local dev and it fixes the missing default on load, and also handles if payment methods disappear at runtime (e.g. COD dependent on shipping). 

Screen Capture on 2020-11-25 at 15-22-23

haszari (commented):
@budzanowski I did a quick investigation on why the defaulting isn't working - added clues in the comments. This isn't a full review, apologies for not helping more (yet!). I'll check back in tomorrow and push some fixes (if needed) 👍

assets/js/base/components/payment-methods/payment-method-options.js
haszari: I've pushed a fix for this: 5381cfc1828afb61a4b5fed5d61f8788863f98f1

assets/js/base/components/payment-methods/payment-methods.js
haszari: @budzanowski I ended up removing lots of this when I moved saved payment method token (saved card) state into context. 

I don't think this caused problems, but there may be more tidying we can do here. There's some preexisting local state in `SavedPaymentMethodOptions` which I don't fully understand. We might be able to move that logic to context too, since the active payment method / token are closely related (i.e. same set of radio buttons).

haszari:
@budzanowski - Update! I've got a bunch more stuff working, and have also made moderate changes to local state (`useState`) along the way.  The one thing left to implement is showing no tab when there is a single payment method (and no saved card). Might just need some style tweaks in `Tabs` component.

Screen Shot 2020-11-26 at 1 48 45 PM

See also comments inline/above related to the changes I've pushed.

Once that's implemented, there may be more refactoring and tidying to do here - will be interesting to see what feedback we get from code review. In particular:

- Consider whether there is logic / behaviour implemented in some components that should be moved to context. (Might predates the PR, e.g. `SavedPaymentMethodOptions` token handling.)
- Review and tidy the `Tabs` and `RadioControlAccordion` components for consistent interface (maybe wrap them?). It feels weird to me that the `Tabs` block has so much more code than the else `RadioControlAccordion` block, maybe we can factor that out so `PaymentMethodOptions` is leaner.

And testing of course 🔨

haszari:
Just spotted an issue on this branch - might exist on trunk too (or possibly existing quirk with Woo core + Stripe), adding a note so I don't forget. 

- With stripe activated, save a payment card to user account.
- Deactivate stripe extension.
- Now view the checkout.
- Expected: no saved cards section, `Choose a payment method` and payment tabs (if 1-2 methods available).
- Actual: the payment methods UI uses radiobuttons mode, and the copy is incorrect (`Use another payment method`); it's behaving as if the saved card UI is displayed, but it's not (because Stripe is disabled).

Screen Shot 2020-11-27 at 3 28 23 PM

This might be an existing issue but it's more obvious on this branch because of the different modes (radiobuttons/tabs). Before this PR is merged the issue should be logged (or fixed) - @budzanowski can you take a look please? 🙇

haszari:
@budzanowski I pushed some initial styling tweaks for single-tab mode. This is just an experiment - feel free to roll it back. It might be better to not use a `Tab` component for the single-method mode.

Screen Shot 2020-11-27 at 4 01 26 PM

Screen Shot 2020-11-27 at 4 01 45 PM

The remaining style things to tweak (that I noticed):

- Remove "button" behaviours, or don't render a button (e.g. clickable, pointer cursor on hover).
- Add box around tabs/single payment method (consistent with radio mode).
- Polish the transition between 1-2 modes (e.g. COD disappear due to shipping). The tab highlight and rule/line transition is a tiny bit janky.

budzanowski:
@haszari regarding https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3439#issuecomment-734562598 : fixed https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3502

haszari:
> regarding #3439 (comment) : fixed #3502

Nice! I actually saw that issue / PR go through but didn't realise it was causing this customer issue!

Let's rebase this PR to latest `trunk` so we don't see that issue when testing (I'll leave that to you as PR owner).

assets/js/base/components/payment-methods/payment-method-options.js
```diff
@@ -42,52 +46,109 @@ const PaymentMethodOptions = () => {
...
+				</div>
+				<div className="wc-block-components-checkout-payment-methods__single-option-content">
+					{ option.content }
+				</div>
+			</div>
+		);
+	} else {
+		paymentMethodsSelectorUI = (
+			<Tabs
+				className="wc-block-components-checkout-payment-methods"
```
haszari: Should this class be tweaked now that it's not the only payment methods component, e.g. `wc-block-components-checkout-payment-methods__two-options-tabs`. Not a big deal but seems appropriate since we now have three variations here.

haszari: My main review comment got lost somehow .. rewrote it below (😭)

This is looking great. I think it should have one more review from someone else before merge (since we both worked on it).

I found a couple of styling issues. Functionality working well :)

- In two-tab mode, there's no box outline (figma has a box around the whole component, consistent with other modes).
- If `Order note` is disabled, the line overlaps the `Submit order` button.

Screen Shot 2020-12-07 at 4 15 51 PM

Probably a good idea to add all the block options that might be impacted to the test instructions – e.g. `Allow shopper to optionally add order notes`, `Show a "Return to cart" link` etc.

budzanowski: The latest trunk was merged back into this branch.

I thought about renaming the `wc-block-components-checkout-payment-methods` but I can't convince myself to this `wc-block-components-checkout-payment-methods__two-options-tabs.` The logic used for displaying payments using one of the 3 options is complicated and finding naming that would make sense especially for the radio buttons case would not be possible. So I will leave the single as the special cease and not change the others. I will revisit this decision if the reviewer will mark this.

Outline added:
image

The line issue is correceted:
image

budzanowski:
@haszari could you please take a quick last look. 
If we decide that we would like to clean up the Tabs component interface as you have pointed out I suggest doing this in a separate PR, as this is something beyond the scope here.

I will do a final look and do small tweaks myself and hopefully tomorrow this can go for a review.

budzanowski:
Oh yeah, the tests need fixing :P

assets/js/base/components/payment-methods/payment-method-options.js
haszari: > I thought about renaming the wc-block-components-checkout-payment-methods but I can't convince myself to this wc-block-components-checkout-payment-methods__two-options-tabs. 

Up to you - I don't think it's a blocker, will be interested to see what reviewers think. Naming it `__radio-buttons` is an option :)

The radiobuttons UI feels like the main/default option to me, the others feel like special cases/variations.

haszari (commented):
Retested and everything's working well. I tested a few different scenarios in Chrome and Safari. Great work @budzanowski 🙌 

I found one issue: user input is lost if UI changes due to shipping. Slight edge case but will annoy someone, hopefully we can preserve the input :)

Screen Capture on 2020-12-15 at 10-28-59

Other than that (and e2e) this is looking great, let's get a final review and get it in the next release!

budzanowski:
I have found additional problems with the saved payment methods. Addressing it here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3551

budzanowski:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3551 is in. Trunk was merged. Continuing.

budzanowski:
@haszari I can't reproduce the vanishing input problem. It just works. Could you verify again if you see the issue?

https://user-images.githubusercontent.com/17271089/102798758-d4006480-43b1-11eb-8db0-36cd0db1d847.mov

haszari:
@budzanowski Looks like the user input is still getting cleared in some circumstances - just reproduced this:

Screen Capture on 2020-12-22 at 08-34-44

To reproduce issues like this you'll need a payment method to dynamically appear/disappear. You can set this up by making COD only available for specific shipping methods. 

- Set up 2 shipping methods, e.g. `Flat rate` and `COD`.
- Enable COD and set `Enable for shipping methods` option to `COD`.

Screen Shot 2020-12-22 at 8 42 56 AM

Then test with different numbers of payment methods, and no saved card. With 3 enabled it will switch between tabs and radio and lose the input (gif above). With 2 methods, it will switch between tabs and "single" mode, similar issue.

budzanowski:
This is ready for review.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -445,6 +449,8 @@ export const PaymentMethodDataProvider = ( { children } ) => {
 		errorMessage: paymentData.errorMessage,
 		activePaymentMethod,
 		setActivePaymentMethod,
+		activeSavedToken,
+		setActiveSavedToken,
```
nerrad, line 453: You'll need to do a couple things here:

- update the typedefs in for the `PaymentMethodDataContext` in `type-defs/contexts`
- make sure `DEFAULT_PAYMENT_METHOD_DATA` in `base/context/cart-checkout/payment-methods/constants.js` is updated with the new properties (and an appropriate default value).

assets/js/base/components/payment-methods/payment-method-options.js
```diff
@@ -8,17 +8,17 @@ import {
...
 	useEditorContext,
 	usePaymentMethodDataContext,
 } from '@woocommerce/base-context';
+import classNames from 'classnames';
 
 /**
  * Internal dependencies
  */
-import Tabs from '../tabs';
 import PaymentMethodTab from './payment-method-tab';
```
nerrad: Should we change the name of this component from `PaymentMethodTab`?

assets/js/base/components/payment-methods/payment-method-options.js
```diff
@@ -42,50 +45,50 @@ const PaymentMethodOptions = () => {
...
+			content: (
+				<PaymentMethodTab allowsSaving={ supports.savePaymentInfo }>
+					{ cloneElement( component, {
+						activePaymentMethod,
+						...paymentMethodInterface,
+					} ) }
+				</PaymentMethodTab>
+			),
+		};
+	} );
```
nerrad: This might be something that could be optimized in a performance followup, although it's likely relatively quick to execute whenever there are render calcs.

assets/js/base/components/payment-methods/saved-payment-method-options.js
```diff
@@ -145,31 +145,23 @@ const SavedPaymentMethodOptions = ( { onChange } ) => {
...
 	return currentOptions.current.length > 0 ? (
 		<RadioControl
 			id={ 'wc-payment-method-saved-tokens' }
-			selected={ selectedOption }
-			onChange={ updateToken }
-			options={ [ ...currentOptions.current, newPaymentMethodOption ] }
+			selected={ activeSavedToken }
+			onChange={ ( ...args ) => {
+				updateToken( ...args );
+			} }
```
nerrad: Why is this wrapped in another function instead of just `onChange={ updateToken }`?

assets/js/base/components/payment-methods/test/payment-methods.js
```diff
@@ -61,51 +33,4 @@ describe( 'PaymentMethods', () => {
 			expect( noPaymentMethods.length ).toBeGreaterThanOrEqual( 1 );
 		} );
 	} );
-
-	test( 'should hide/show PaymentMethodOptions when a saved payment method is checked/unchecked', async () => {
-		registerMockPaymentMethods();
-		render(
```
nerrad: I appreciate why these tests are removed because behaviour has changed, but we should make sure the new behaviour is tested.

assets/js/base/components/radio-control-accordion/index.js
```diff
@@ -0,0 +1,73 @@
...
+ * External dependencies
+ */
+import classnames from 'classnames';
+import { withInstanceId } from '@woocommerce/base-hocs/with-instance-id';
+
+/**
+ * Internal dependencies
+ */
+import RadioControlOption from '../radio-control/option';
+import '../radio-control/style.scss';
```
nerrad: I'm curious why this style is imported here. Since we're building to the same bundle, the import in `radio-control/style.scss` should be sufficient. So I don't think you'll need this import.

assets/js/base/components/radio-control-accordion/index.js
```diff
@@ -0,0 +1,73 @@
...
+	return (
+		options.length && (
+			<div
+				className={ classnames(
+					'wc-block-components-radio-control',
+					className
+				) }
+			>
+				{ options.map( ( option ) => {
+					const hasOptionContent = 'content' in option;
```
nerrad: usage of `in` here assumes `option` is an object. Should we defensively verify that first?

nerrad (changes requested):
Coming along! There's some nice cleanup of code here particularly around how saved tokens are updated and exposed for components.

I didn't do exhaustive testing but I did notice the following (see notes in screenshot):

Image 2021-01-22 at 2 26 54 PM

I've made a few other comments inline with the code as well.

Also, please make sure accessibility testing is done on this change (keyboard switching through payment method options, correct a11y announcements etc).

budzanowski:
The section labels weight was taken directly from the design doc:
image
Should this be followed up with the design?

assets/js/base/components/payment-methods/saved-payment-method-options.js
budzanowski: Good suggestion. @haszari do you remember if there was a particular reason why this was done this way?

assets/js/base/components/radio-control-accordion/index.js
budzanowski: TIL. OK, I need to put webpack and bundling as the next item on my learning list.

budzanowski:
Payments label size fixed. Gosh, that was embarrassing. I did not have PayPal enabled the whole time 🤦🏻 
image

assets/js/base/components/radio-control-accordion/index.js
budzanowski: It feels a bit excessive but for sure it would not hurt. Sometimes I feel like things like that should just break - because this implies the wrong usage of the component.

budzanowski: Corrected.

assets/js/base/components/payment-methods/payment-method-options.js
budzanowski: Because we no longer have tabs? Yes, that makes sense. Change to `PaymentMethodCard`.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
budzanowski: Done.

assets/js/base/components/payment-methods/saved-payment-method-options.js
haszari: No idea, I'm always a fan of less functions, less indirection so this is good advice from @nerrad  IMO 😸. Maybe there used to be something else happening inside the inline func (which has since been removed).

assets/js/base/components/payment-methods/test/payment-methods.js
nerrad: Would still like to make sure we have test coverage on the new ui/ux - is this something planned for a followup?

nerrad (commented):
Thanks for addressing some of the feedback Bartek, testing seemed to go well.

Regarding section label weight: I see it matches designs, let's leave as is for now then.

There are still a few issues not addressed:

- My comment about test coverage.
- The regression in padding around the "Add a note to your order" checkbox.
- The inconsistent spacing between payment method option text and the radio.

Screenshots for the last two issues:

In this PR:

Image 2021-01-29 at 7 47 21 AM

In trunk:

Image 2021-01-29 at 7 45 29 AM

Also, can you please verify you've:

- tested accessibility of the ui.
- tested with mobile views (seemed to work okay in my testing other than the order note checkbox issue which persists here).

assets/js/base/components/payment-methods/test/payment-methods.js
budzanowski: I want to have it with this PR.

budzanowski:
I have noticed that Chrome has an accessibility helper in inspect tool:
image
I am not sure how correct the tool is so I have tested manually how it works and IMO it works correctly.

nerrad (approved):
Changes look good and test well. Designs in figma still have borders around the add notes checkbox:

Image 2021-02-01 at 11 15 23 AM

However, I'm fine with the current implementation - styles can be revisited in future work on things like the return to cart link and payment method button design mismatch.

Issue #3433, opened 2020-11-19
Compatibility Audit  for Twenty Twenty One Theme


nerrad:
This is the main issue for an audit of all compatibility for WooCommerce Blocks and the Twenty Twenty One theme.

> Before addressing the issues from this audit, make sure you are using the latest version of the [twenty twenty one theme](https://github.com/wordpress/twentytwentyone) in case any of these issues have been addressed.

**Note:** All my testing for the below was done in the Chrome browser. It's possible behaviour is different in different browsers so when completing the issues, please doublecheck how it looks in other browsers that we support and ensure any fixes don't impact other browser renderings either.

I've added this to the 3.8.1 milestone because we'll have to include this as a patch for WooCommerce Core 4.8 release so it's ready for WP 5.6.

So as we apply fixes we'll need to cherrypick for a 3.8.1 release (cc @budzanowski , as the release lead for the 3.8 branch you'll be called upon to do the 3.8.1 patch release when it's ready - I'll assist).

## Editor

These are the issues that show up in the editor:

### Text flow in Add to Cart buttons for Product Grid type blocks (affects frontend as well)

For all the product grid type blocks (including All Products) the text flow in the buttons is fairly wonky for the container restricting the size of the buttons in the block columns. Example screenshot for HandPicked Products:

<img width="773" alt="Image 2020-11-19 at 4 33 29 PM" src="https://user-images.githubusercontent.com/1429108/99726900-30a3f300-2a85-11eb-8660-97a7c48c8015.png">

### Jarring visual difference between "Add to Cart" button and "Read More" or "View Products" buttons in All Products Block.

Example:

<img width="871" alt="Image 2020-11-19 at 4 43 58 PM" src="https://user-images.githubusercontent.com/1429108/99727847-85943900-2a86-11eb-8d17-5c00c7358e8b.png">

### Cart and Checkout Blocks only display mobile view

I just wanted to note this as a part of the audit results but I'm not sure it's something we actually need to address as a part of this work.

### Price slider UI for Price Filter Block on frontend is wonky.

<img width="286" alt="Image 2020-11-19 at 4 56 37 PM" src="https://user-images.githubusercontent.com/1429108/99728987-423aca00-2a88-11eb-8fd5-9a601b40b91c.png">

### Wonky radio inputs for Cart and Checkout Blocks

[...]

PR #3443, opened 2020-11-23
Fix Twenty Twenty One Button and Placeholder Styling

mikejolley:
This is part of #3433

Fixes these points from the parent issue:

- Jarring visual difference between "Add to Cart" button and "Read More" or "View Products" buttons in All Products Block.
- Text flow in Add to Cart buttons for Product Grid type blocks (affects frontend as well)

Also:

- Fixes placeholder background color (grey on grey) in Twenty Twenty One
- Updates column based rules so buttons and content shrink as more columns are added
- Removed the checkmarks on cart buttons when more columns are shown

I smoke tested other themes to ensure these styles didn't break anything else.

### Screenshots

These are the updated placeholder styles:

Edit Page ‹ one wordpress test — WordPress 2020-11-23 13-52-08

And this shows the fixed buttons and content sizing:

1. View a product block in the editor. Buttons should be consistent. Loading placeholder should match theme.
2. Test various columns. Content should reduce in size to fit better horizonally.
3. Repeat on frontend.

cc @Aljullu

### DIFF
--- a/assets/css/abstracts/_colors.scss
+++ b/assets/css/abstracts/_colors.scss
@@ -18,6 +18,7 @@ $no-stock-color: $alert-red;
 $low-stock-color: $alert-yellow;
 $in-stock-color: $alert-green;
 
+$placeholder-color: var(--global--color-primary, $gray-200);
 $input-border-gray: #8d96a0;
 $input-border-dark: rgba(255, 255, 255, 0.4);
 $input-disabled-dark: rgba(255, 255, 255, 0.3);
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -46,38 +46,56 @@
 		margin-right: 0.5em;
 	}
 }
-.wc-block-grid__product-add-to-cart {
+.wc-block-grid__product-add-to-cart.wp-block-button {
 	word-break: break-word;
 	white-space: normal;
-	a,
-	button {
+	.wp-block-button__link {
 		word-break: break-word;
 		white-space: normal;
 		margin: 0 auto !important;
 		display: inline-flex;
 		justify-content: center;
+		text-align: center;
+		// Set button font size and padding so it inherits from parent.
+		padding: 0.5em;
+		font-size: 1em;
 
 		&.loading {
 			opacity: 0.25;
 		}
 
-		&::after {
-			margin-left: 0.5em;
-			display: inline-block;
-		}
-
 		&.added::after {
 			font-family: WooCommerce; /* stylelint-disable-line */
 			content: "\e017";
+			margin-left: 0.5em;
+			display: inline-block;
+			width: auto;
+			height: auto;
 		}
 
 		&.loading::after {
 			font-family: WooCommerce; /* stylelint-disable-line */
 			content: "\e031";
 			animation: spin 2s linear infinite;
+			margin-left: 0.5em;
+			display: inline-block;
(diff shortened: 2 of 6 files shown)


assets/js/base/components/product-list/style.scss
```diff
@@ -74,15 +74,25 @@
...
 		}
 	}
+	&.has-5-columns.alignfull,
+	&.has-4-columns:not(.alignwide):not(.alignfull) {
+		.wc-block-grid__product {
+			font-size: 0.75em;
+		}
+	}
+	&.has-3-columns:not(.alignwide):not(.alignfull) {
+		font-size: 0.92em;
```
Aljullu, line 94: I didn't test it, but do you think `0.875em` would work here? In that case we could use the [`font-size()` mixin](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/css/abstracts/_mixins.scss#L9-L12) with `small` size. And `0.75em` could be replaced with `font-size("smaller")`.

For `0.5em` and `0.58em`, we can either create new named font sizes in the mixin, or leave the hard-coded values here. I don't have a strong opinion.

assets/css/style.scss
```diff
@@ -46,38 +46,56 @@
...
+	.wp-block-button__link {
 		word-break: break-word;
 		white-space: normal;
 		margin: 0 auto !important;
 		display: inline-flex;
 		justify-content: center;
+		text-align: center;
+		// Set button font size and padding so it inherits from parent.
+		padding: 0.5em 1em;
+		font-size: 1em;
```
Aljullu, line 61: Can this be `font-size("regular")`? :slightly_smiling_face:

assets/css/abstracts/_colors.scss
```diff
@@ -18,6 +18,7 @@ $no-stock-color: $alert-red;
 $low-stock-color: $alert-yellow;
 $in-stock-color: $alert-green;
 
+$placeholder-color: var(--global--color-primary, $gray-200);
```
Aljullu, line 21: Correct me if I'm wrong, but `--global--color-primary` is not a variable we will get from Gutenberg but [something specific to the theme](https://github.com/WordPress/twentytwentyone/blob/trunk/assets/sass/01-settings/global.scss#L81), no?

In that case, I think we should only use that variable inside ` .theme-twentytwentyone { ... }` selectors.

assets/css/style.scss
```diff
@@ -46,38 +46,56 @@
 		margin-right: 0.5em;
 	}
 }
-.wc-block-grid__product-add-to-cart {
+.wc-block-grid__product-add-to-cart.wp-block-button {
```
Aljullu, line 49: Is `.wp-block-button` added for specificity? If that's the case, could we add a comment specifying that so it's not removed in the future?

Aljullu (approved):
This looks great, I like that most fixes are not TwentyTwentyOne-specific but instead will improve how buttons look in all themes.

I left some inline comments but I'm approving this PR since most of them are minor. There is only one that I think needs to be addressed (in this PR or in a follow-up): the one about the usage of `--global--color-primary`, but happy to discuss. :slightly_smiling_face:

nerrad:
Given the approval and need to include in pending release builds I'm going to merge this as is - please do address any follow-ups in subsequent prs.

PR #3444, opened 2020-11-23
Fix Twenty Twenty One Price filter, Active filters and radio control styling

Aljullu:
Part of #3433.

This PR fixes the Twenty Twenty One styling of:
* Filter products by price block.
* Active filters block.
* Radio control component (used in the shipping options of the Cart and Checkout blocks).

It also refactors the radio control styles so the file `_mixins.scss` is removed. It was there for legacy reasons, but in this PR everything has been moved to a `style.scss`.

#### Active filters
_Before:_
imatge

_After:_
imatge

#### Price filter
_Before:_
imatge

_After:_
imatge

_After (IE11):_
imatge

:point_up_2: Styling is not perfect in IE11, but everything should work fine. Twenty Twenty One has several other glitches on IE11, so I think it's ok not getting a pixel perfect design in IE11.

#### Radio control (in Checkout block)
_Before:_
imatge

_After:_
imatge

1. With the Twenty Twenty One theme installed, create a page with the Filter Products by Price, Active filters and All Products blocks. Create another page with the Checkout block (ensuring you have at least two shipping methods enabled).
2. Both, in the editor and the frontend verify all blocks look correct. (In the screenshots above you can see what has been changed in this PR).
3. It would be good to test that in some other themes (at least, Storefront and Twenty Twenty) that there are no regression.
4. The price slider component is quite complex, so it would be good to test it with different browsers as well.

### Known issues

Filter products by price block styles are not correctly applied in the editor:

imatge

That's because form styles are not applied in the editor in Twenty Twenty One. I opened an issue in its repo (https://github.com/WordPress/twentytwentyone/issues/862) which is now closed but I would like to follow-up later today or tomorrow.

cc @mikejolley


budzanowski:
In my test setup ( Chrome ) the radio buttons are not having a "perfect" circle shape and the inner dot is off center:
image


With a small adjustment ( disabling margin-top, min-heigh and min-width ) to the 
image

We get a much more pleasing shape and in the center dot:
image

The radio is a bit larger if that matters.

nerrad:
What browser is that @budzanowski? I'm not seeing that in Chrome:

Image 2020-11-23 at 10 42 25 AM

budzanowski:
Hmm,
image

budzanowski:
OK, I had the site's zoom set to 90%. With 100% it looks good. I assume that we don't care about anything else than 100%.

assets/js/base/components/radio-control/style.scss
```diff
@@ -1,4 +1,119 @@
...
+		.has-dark-controls & {
+			border-color: $controls-border-dark;
+
+			&:checked::before {
+				background: $input-text-dark;
+			}
+		}
+	}
+
+	@include breakpoint( ">782px" ) {
```
budzanowski, line 91: Is there a specific reason why we don't use the `.is-large`?

nerrad:
It looks like this issue (with styles on scaling) already existed in trunk (on Storefront too). I think any issues with scaling up would be more important, but non-blocking for this pr and we can address in a separate issue if there's need.

budzanowski:
Looks good. Tests good. I have left a small question but this is nonblocking IMO.

budzanowski (approved):

nerrad:
Just dropping in here that testing went well for me in Twenty Twenty One for the identified fixes. I also tested Storefront to make sure there was no impact to changes there. I only tested in Google Chrome.

assets/js/base/components/radio-control/style.scss
Aljullu: Hmmm, I thought the reason was that this component was built before we had the container-query classnames. But taking a second look but it seems like this media query makes the radio controls smaller. I think having bigger controls is intended to happen in handheld devices no matter the width of the block? So in this case it probably makes more sense using a CSS media query than container queries? :thinking:  I don't have a strong opinion, though.

I will merge this as-is, but I'm happy to change this in a follow-up if we agree on that.

Issue #3441, opened 2020-11-23
Integration Tracking (WC Subscriptions): Add future Subscriptions data to store cart endpoint


senadir:
Like https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3262 but this is for Subscriptions data, not Subscription Product.
When in Cart and Checkout, WCS will show you future subscriptions you're going to create, so we need to get all of that data and expose it.

Initial spike work would be to decide what kind of data we need to expose, this can be done by:
- looking at WCS code and see what's used to render that section.
- Seeing our new designs and what kind of data is there.
- Map it back to `wc/v1/subscriptions/` to use the same keys. Currently, this is what's exported from that package:

```json
{
  "id": 1546,
  "parent_id": 0,
  "status": "on-hold",
  "order_key": "wc_order_F2SkJCP08LSXS",
  "number": "1546",
  "currency": "DZD",
  "version": "4.7.0",
  "prices_include_tax": false,
  "date_created": "2020-10-25T13:12:36",
  "date_modified": "2020-10-29T09:40:30",
  "customer_id": 12,
  "discount_total": "0.00",
  "discount_tax": "0.00",
  "shipping_total": "0.00",
  "shipping_tax": "0.00",
  "cart_tax": "4703.55",
  "total": "28000.00",
  "total_tax": "4703.55",
  "billing": {
    "first_name": "User",
    "last_name": "Name",
    "address_1": "",
    "address_2": "",
    "city": "Oran",
    "state": "Oran",
    "postcode": "",
    "country": "DZ",
    "email": "user@automattic.com",
    "phone": ""
  },
  "shipping": {
    "first_name": "",
    "last_name": "",
    "company": "",
    "address_1": "",
    "address_2": "",
    "city": "",
    "state": "",
    "postcode": "",
    "country": ""
  },
  "payment_method": "",
  "payment_method_title": "",
  "transaction_id": "",
  "customer_ip_address": "",
  "customer_user_agent": "",
  "created_via": "importer",
  "customer_note": "",
  "date_completed": null,
  "date_paid": null,
  "cart_hash": "",
  "line_items": [
    {
      "id": 469,
      "name": "Product",
      "sku": "",
      "product_id": 894,
      "variation_id": 0,
      "quantity": 1,
      "tax_class": "",
      "price": "23296.45",
      "subtotal": "23296.45",
      "subtotal_tax": "0.00",
      "total": "23296.45",
[...]

PR #3445, opened 2020-11-23
Expose store/cart via ExtendRestApi to extensions

### LINKED ISSUE #3409: Integration Tracking (WC Subscriptions): Expose store/cart via ExtendRestApi
As explained in this post pca54o-z3-p2 we're changing our approach to integrate WCS.
Previously, the goal was to expose Subscription Products data on `cart/items` and using that data to construct the recurring totals view.
This is going hard and error-prone process, so we're switching to building the subscription data on the server (in WCS) and making it available in our API so that we can consume it and construct recurring totals section without much business logic involved.

To get to this point, we need to expose our `store/cart` endpoint so that WCS can hook into it and inject needed data.

Once the data is injected, we need to make available, either as part of `useStoreCart` or a new hook `useExtensionData`. 

If we're going to implement a new hook, we need to think about how to do that so that a single hook can cover several endpoints at once.

senadir:
Closes #3409

This PR exposes `store/cart` endpoint to be extensible, we need it so we can add Future Subscriptions data to it (a prerequisite for #3441).

### Testing steps
1. Add this code
```php
<?PHP
use Automattic\WooCommerce\Blocks\StoreApi\Schemas\CartSchema;

add_action(
	'init',
	function() {

			CartSchema::IDENTIFIER,
			'subscriptions',
			function () {
				return array(
					'key' => array(
						'type'        => 'string',
					),
				);
			},
			function () {
				return array( 'key' => 'hey' );
			}
		);
	}
);
```
2. Inspect the `cart` endpoint response (trigger a change in cart like quantity or shipping or coupons).
3. You should see `extensions => subscriptions => key => hey`.

### DIFF
--- a/src/Domain/Services/ExtendRestApi.php
+++ b/src/Domain/Services/ExtendRestApi.php
@@ -3,6 +3,7 @@ namespace Automattic\WooCommerce\Blocks\Domain\Services;
 
 use Automattic\WooCommerce\Blocks\Domain\Package;
 use Automattic\WooCommerce\Blocks\StoreApi\Schemas\CartItemSchema;
+use Automattic\WooCommerce\Blocks\StoreApi\Schemas\CartSchema;
 use Throwable;
 use Exception;
 
@@ -31,7 +32,7 @@ class ExtendRestApi {
 	 *
 	 * @var array
 	 */
-	private $endpoints = [ CartItemSchema::IDENTIFIER ];
+	private $endpoints = [ CartItemSchema::IDENTIFIER, CartSchema::IDENTIFIER ];
 
 	/**
 	 * Data to be extended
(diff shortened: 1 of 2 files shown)


senadir:
cc @mikejolley would it be wise to pass `$cart` to 3PD or are they expected to use `wc()->cart`, is there any difference here?

senadir:
Compared the outcome of $cart and wc()->cart and there's no difference, furthermore, $cart is simply an instance of `CartController::get_cart_instance` which simply check if $cart exists and throws an error.

```php
<?php
/**
 * Get main instance of cart class.
 *
 * @throws RouteException When cart cannot be loaded.
 * @return \WC_Cart
 */
public function get_cart_instance() {
	$cart = wc()->cart;

	if ( ! $cart || ! $cart instanceof \WC_Cart ) {
		throw new RouteException( 'woocommerce_rest_cart_error', __( 'Unable to retrieve cart.', 'woo-gutenberg-products-block' ), 500 );
	}

	return $cart;
}
```

Given this info, I don't think we need to pass $cart to that callback, 3PD can access it via `wc()->cart`.

mikejolley:
@senadir they should use the global for now I think.

budzanowski (approved):
Looks and tests well.

PR #3446, opened 2020-11-23
Fix radio controls and checkboxes in Twenty Twenty One dark theme

Aljullu:
Part of #3433.

I found one more issue between WC Blocks and Twenty Twenty One: when the dark mode was enabled, form controls were either not visible or looked weird. That happened because Twenty Twenty One adds its own styles to form elements which collide with ours.

If we see this issue happening with more themes in the future, we might need to evaluate whether it makes sense at all to style form controls on our end instead of relying on theme styles. For now, I only fixed the styles for Twenty Twenty One.

#### Radio control
_Before:_
imatge
(selected option is not visible)

_After:_
imatge

#### Checkbox
_Before:_
imatge

_After:_
imatge

1. Create a page with the Checkout block and enable the Twenty Twenty One theme. Make sure your store has at least two shipping options for the same zone.
2. In the customizer, go to Colors & Dark Mode and enable "Dark Mode support".
3. In the frontend, add products to your Cart and go to the page with the Checkout block.
4. Verify the radio boxes of the shipping methods section and the order notes checkbox look fine (see screenshots above).
5. Toggle the dark mode off and verify they look correct as well.

### DIFF
--- a/assets/js/base/components/checkbox-control/style.scss
+++ b/assets/js/base/components/checkbox-control/style.scss
@@ -61,3 +61,19 @@
 		display: none;
 	}
 }
+
+.theme-twentytwentyone {
+	.wc-block-components-checkbox__input[type="checkbox"] {
+		border-color: var(--form--border-color);
+		position: relative;
+	}
+
+	.wc-block-components-checkbox__input[type="checkbox"]:checked {
+		background-color: #fff;
+		border-color: var(--form--border-color);
+	}
+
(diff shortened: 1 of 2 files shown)


nerrad (approved):
Code changes look good, I had a little snafu because I tested with the block dark mode inputs setting turned on but that conflicts with the theme setting for dark mode. It works as expected leaving that block setting alone.

PR #3450, opened 2020-11-23
Fix checkbox and textarea styles in Twenty Twenty One when it has dark controls

Aljullu:
Checkbox and textarea styles were not correctly applied in Twenty Twenty One when the Checkout block used dark controls.

### Screenshots
_Before:_
imatge

_After:_
imatge

1. Create a page with the Checkout block and enable the Twenty Twenty One theme. Make sure your store has at least two shipping options for the same zone.
2. In the customizer, go to Colors & Dark Mode and chose a dark background color.
3. In the frontend, add products to your Cart and go to the page with the Checkout block.
4. Verify the order notes checkbox and text area look fine (see screenshots above).

### DIFF
--- a/assets/js/base/components/checkbox-control/style.scss
+++ b/assets/js/base/components/checkbox-control/style.scss
@@ -63,12 +63,15 @@
 }
 
 .theme-twentytwentyone {
-	.wc-block-components-checkbox__input[type="checkbox"] {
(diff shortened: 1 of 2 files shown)


nerrad (approved):
Looks good and works well - thanks!

PR #3451, opened 2020-11-23
Improve deploy script to properly account for releases not deployed to wp.org

nerrad:
I noticed when doing the 3.8.1 patch release that the `npm run deploy` script is not properly accounting for patch releases that are _not_ deployed to WordPress.org. In that scenario, we don't want to do a GitHub release because that triggers the `wp.org` workflow but we still want to push a git tag. 

This pull request fixes that (and was already tested in the `release/3.8.1` branch for the 3.8.1 release).


haszari (approved):
Looks good - I'm assuming `hub release create` creates a release AND a tag, so in the else (no github release) we just need to push a tag to memorialise the release in the repo 👍

PR #3454, opened 2020-11-24
refactor and reorder checkout processing

### LINKED ISSUE #3363: Checkout API can create a user account when order fails due to coupon restriction (or other failures)

haszari:
Fixes #3363

Related: #3260 #3423  https://github.com/woocommerce/woocommerce-subscriptions/pull/3825

This PR revamps `POST /checkout` processing in the Store API to address various issues and inconsistencies. 

The overall order of processing is as follows (more detail tbc):

1. Obtain Draft Order
1. Process Request
1. Process Customer
1. Validate Order
1. Process Payment

The following expectations apply to order status:

- Order should be `checkout-draft` status until payment is attempted.
- Order should only be `pending` once initial validation and processing has been attempted; it should be set to `pending` when it is _ready_ for payment processing.

The following expectations apply to checkout signup (aka create account):

- Customer accounts can be created as a side-effect of order processing; a checkout request can fail, yet still sign up the shopper and log them in.
- If an account is created:
  - Customer should be logged in immediately.
  - Customer id should be returned with response (even if it's an error) so client can update user-related UI. 
  - Any open / existing draft orders for the session should be updated with customer id.
  - Store API nonce should be refreshed and returned with response.
- If account creation fails:
  - Order should not proceed to payment; error should be returned, order should remain at `checkout-draft`. 

Note - not all the above account create items are implemented, though there is related work in #3429. (This might be handled as follow up once the key checkout processing is fully tested.)

##### Order date should match when order was "placed"

The order date is when draft order was created. If order sits around for a while, this could be misleading (e.g. up to a day out).

I'm wondering if we should reset the date when the order goes `pending`. We can probably log an issue and handle as follow up (not a blocker to this PR IMO). 

##### Returning customer id and updated nonce with error responses
We need to return the customer ID (and a new nonce) when customer account is created, in particular when there's an error (e.g. #3429). This PR implements this, and handling of the nonce client side. 

This is a basic fix for #3429 – this flow now works, but there's more UX polish to add (inform user that they have an account, remove `Create account` checkbox, etc). We can handle this as follow up in #3429.

##### Require an account to checkout
This is handled as before in `CreateAccount` with `$checkout->is_registration_required()` which checks well-known core options and filters.

Test all purchase / checkout flows, with and without user account, with various settings and errors, with all payment methods (!).

Please consider the following variations and test a good range of combinations:

- Order fail due to bad coupon.
- Order fail due to no stock.
- Order fail due to server-side payment error. (May be simplest to hack this in PHP, e.g. force a fail in BACS/COD.)
- Guest/anon orders or logged in.
- Various account settings: allow guest | require account, allow signup (x shopper `Create an account`=yes/no).

Since this is a revamp of checkout processing, I'd appreciate some extra care & time testing! Reviewers please leave a summary of the use-cases you tested. And please add ideas for variations that I've missed (and test 'em!).

### Changelog

> Fix: Fixed a bug in Checkout block (Store API) causing checkout to fail when using an invalid coupon and creating an account.
> Dev: Refactored and reordered Store API checkout processing to handle various edge cases and better support future extensibility.

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -216,10 +216,10 @@ const CheckoutProcessor = () => {
 					setIsProcessingOrder( false );
 				} );
 			} )
-			.catch( ( fetchResponse ) => {
-				fetchResponse.json().then( function ( response ) {
+			.catch( ( errorResponse ) => {
+				errorResponse.json().then( function ( response ) {
 					// Update nonce.
-					triggerFetch.setNonce( fetchResponse.headers );
+					triggerFetch.setNonce( errorResponse.headers );
 
 					// If updated cart state was returned, also update that.
 					if ( response.data?.cart ) {
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -216,8 +216,11 @@ const CheckoutProcessor = () => {
 					setIsProcessingOrder( false );
 				} );
 			} )
-			.catch( ( error ) => {
-				error.json().then( function ( response ) {
+			.catch( ( fetchResponse ) => {
+				fetchResponse.json().then( function ( response ) {
+					// Update nonce.
+					triggerFetch.setNonce( fetchResponse.headers );
+
 					// If updated cart state was returned, also update that.
 					if ( response.data?.cart ) {
 						receiveCart( response.data.cart );
--- a/src/StoreApi/Routes/AbstractRoute.php
+++ b/src/StoreApi/Routes/AbstractRoute.php
@@ -85,23 +85,15 @@ abstract class AbstractRoute implements RouteInterface {
 	}
 
 	/**
-	 * Converts an error to a response object.
-	 *
-	 * Based on WP_REST_Server (wp-includes/rest-api/class-wp-rest-server.php).
+	 * Converts an error to a response object. Based on \WP_REST_Server.
 	 *
 	 * @param WP_Error $error WP_Error instance.
 	 * @return WP_REST_Response List of associative arrays with code and message keys.
 	 */
 	protected function error_to_response( $error ) {
 		$error_data = $error->get_error_data();
-
-		if ( is_array( $error_data ) && isset( $error_data['status'] ) ) {
-			$status = $error_data['status'];
-		} else {
-			$status = 500;
-		}
-
-		$errors = array();
+		$status     = isset( $error_data, $error_data['status'] ) ? $error_data['status'] : 500;
+		$errors     = [];
 
 		foreach ( (array) $error->errors as $code => $messages ) {
 			foreach ( (array) $messages as $message ) {
@@ -113,10 +105,9 @@ abstract class AbstractRoute implements RouteInterface {
 			}
 		}
 
-		$data = $errors[0];
-		if ( count( $errors ) > 1 ) {
-			// Remove the primary error.
-			array_shift( $errors );
+		$data = array_shift( $errors );
+
+		if ( count( $errors ) ) {
 			$data['additional_errors'] = $errors;
 		}
 
--- a/src/StoreApi/Routes/AbstractRoute.php
+++ b/src/StoreApi/Routes/AbstractRoute.php
@@ -47,7 +47,7 @@ abstract class AbstractRoute implements RouteInterface {
 	 * Get the route response based on the type of request.
 	 *
 	 * @param \WP_REST_Request $request Request object.
-	 * @return \WP_Error|\WP_REST_Response
+	 * @return \WP_REST_Response
 	 */
 	public function get_response( \WP_REST_Request $request ) {
 		$response = null;
@@ -70,15 +70,67 @@ abstract class AbstractRoute implements RouteInterface {
 					$response = $this->get_route_response( $request );
 					break;
 			}
-			if ( 'GET' !== $request->get_method() && ! is_wp_error( $response ) ) {
-				$response->header( 'X-WC-Store-API-Nonce', wp_create_nonce( 'wc_store_api' ) );
-			}
 		} catch ( RouteException $error ) {
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode(), $error->getAdditionalData() );
 		} catch ( \Exception $error ) {
 			$response = $this->get_route_error_response( 'unknown_server_error', $error->getMessage(), 500 );
 		}
+
+		if ( is_wp_error( $response ) ) {
+			$response = $this->error_to_response( $response );
+		}
+
+		return $this->add_response_headers( $response );
+	}
+
+	/**
+	 * Add Store API headers to a response object.
+	 *
+	 * @param \WP_REST_Response $response Response object.
+	 * @return \WP_REST_Response
+	 */
+	protected function add_response_headers( $response ) {
+		$response->header( 'X-WC-Store-API-Nonce', wp_create_nonce( 'wc_store_api' ) );
 		return $response;
+	}
+
+	/**
+	 * Converts an error to a response object.
+	 *
+	 * Based on WP_REST_Server (wp-includes/rest-api/class-wp-rest-server.php).
+	 *
+	 * @param WP_Error $error WP_Error instance.
+	 * @return WP_REST_Response List of associative arrays with code and message keys.
+	 */
+	protected function error_to_response( $error ) {
+		$error_data = $error->get_error_data();
+
+		if ( is_array( $error_data ) && isset( $error_data['status'] ) ) {
+			$status = $error_data['status'];
+		} else {
+			$status = 500;
+		}
+
+		$errors = array();
+
+		foreach ( (array) $error->errors as $code => $messages ) {
+			foreach ( (array) $messages as $message ) {
+				$errors[] = array(
+					'code'    => $code,
+					'message' => $message,
+					'data'    => $error->get_error_data( $code ),
+				);
+			}
+		}
+
+		$data = $errors[0];
+		if ( count( $errors ) > 1 ) {
+			// Remove the primary error.
+			array_shift( $errors );
+			$data['additional_errors'] = $errors;
+		}
+
+		return new \WP_REST_Response( $data, $status );
 	}
 
 	/**
--- a/src/StoreApi/Routes/AbstractRoute.php
+++ b/src/StoreApi/Routes/AbstractRoute.php
@@ -80,16 +80,6 @@ abstract class AbstractRoute implements RouteInterface {
 			$response = $this->error_to_response( $response );
 		}
 
-		return $this->add_response_headers( $response );
-	}
-
-	/**
-	 * Add Store API headers to a response object.
-	 *
-	 * @param \WP_REST_Response $response Response object.
-	 * @return \WP_REST_Response
-	 */
-	protected function add_response_headers( $response ) {
 		$response->header( 'X-WC-Store-API-Nonce', wp_create_nonce( 'wc_store_api' ) );
 		return $response;
 	}
--- a/src/StoreApi/Routes/AbstractRoute.php
+++ b/src/StoreApi/Routes/AbstractRoute.php
@@ -81,6 +81,7 @@ abstract class AbstractRoute implements RouteInterface {
 		}
 
 		$response->header( 'X-WC-Store-API-Nonce', wp_create_nonce( 'wc_store_api' ) );
+		$response->header( 'X-WC-Store-API-User', get_current_user_id() );
 		return $response;
 	}
 
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -139,53 +139,45 @@ class Checkout extends AbstractRoute {
 	}
 
 	/**
-	 * Update and process payment for the order.
+	 * Update and process an order.
+	 *
+	 * 1. Obtain Draft Order
+	 * 2. Process Request
+	 * 3. Process Customer
+	 * 4. Validate Order
+	 * 5. Process Payment
 	 *
 	 * @throws RouteException On error.
 	 * @param \WP_REST_Request $request Request object.
 	 * @return \WP_REST_Response
 	 */
 	protected function get_route_post_response( \WP_REST_Request $request ) {
+		// --
+		// 1. Obtain Draft Order
+
 		// Update customer first since orders will be created using that data.
 		$this->update_customer_from_request( $request );
 
-		$order_controller = new OrderController();
-		$order_object     = $this->get_draft_order_object( $this->get_draft_order_id() );
+		// Get the order, including validation and reserving stock.
+		$order = $this->create_or_update_draft_order();
 
-		if ( ! $order_object instanceof \WC_Order ) {
-			throw new RouteException(
-				'woocommerce_rest_checkout_invalid_order',
-				__( 'This session has no orders pending payment.', 'woo-gutenberg-products-block' ),
-				500
-			);
+		// --
+		// 2. Process Request
+
+		// If any form fields were posted, update the order (e.g. order note, payment method).
+		$this->update_order_from_request( $order, $request );
+
+		// Validate payment method before proceeding.
+		if ( $order->needs_payment() ) {
+			// Will throw and return 400 if payment method missing or invalid.
+			$this->get_request_payment_method_id( $request );
 		}
 
-		// Ensure order still matches cart.
-		$order_controller->update_order_from_cart( $order_object );
+		// --
+		// 3. Process Customer
 
-		// Create a new user account as necessary.
-		// Note - CreateAccount class includes feature gating logic (i.e. this
-		// may not create an account depending on build).
-		if ( defined( 'WC_VERSION' ) && version_compare( WC_VERSION, '4.7', '>=' ) ) {
-			// Checkout signup is feature gated to WooCommerce 4.7 and newer;
-			// Because it requires updated my-account/lost-password screen in 4.7+
-			// for setting initial password.
-			try {
-				$create_account = Package::container()->get( CreateAccount::class );
-				$create_account->from_order_request( $request );
-				$order_object->set_customer_id( get_current_user_id() );
-			} catch ( Exception $error ) {
-				$this->handle_error( $error );
-			}
-		}
-		// If any form fields were posted, update the order.
-		$this->update_order_from_request( $order_object, $request );
-
-		// Check order is still valid.
-		$order_controller->validate_order_before_payment( $order_object );
-
-		// Persist customer address data to account.
-		$order_controller->sync_customer_data_with_order( $order_object );
+		// Update order with customer details, and sign up a user account as necessary.
+		$this->process_customer( $order, $request );
 
 		/*
 		* Fire woocommerce_blocks_checkout_order_processed, should work the same way as woocommerce_checkout_order_processed
@@ -193,17 +185,30 @@ class Checkout extends AbstractRoute {
 		* NOTE: this hook is still experimental, and might change or get removed.
 		* @todo: Document and stabilize __experimental_woocommerce_blocks_checkout_order_processed
 		*/
-		do_action( '__experimental_woocommerce_blocks_checkout_order_processed', $order_object );
+		do_action( '__experimental_woocommerce_blocks_checkout_order_processed', $order );
 
-		if ( ! $order_object->needs_payment() ) {
-			$payment_result = $this->process_without_payment( $order_object, $request );
+		// --
+		// 4. Validate Order
+
+		// Check order is still valid.
+		$order_controller = new OrderController();
+		$order_controller->validate_order_before_payment( $order );
+
+		// --
+		// 5. Process Payment
+
+		if ( ! $order->needs_payment() ) {
+			$payment_result = $this->process_without_payment( $order, $request );
 		} else {
-			$payment_result = $this->process_payment( $order_object, $request );
+			$payment_result = $this->process_payment( $order, $request );
 		}
+
+		// --
+		// Generate response & return.
 
 		$response = $this->prepare_item_for_response(
 			(object) [
-				'order'          => wc_get_order( $order_object ),
+				'order'          => wc_get_order( $order ),
 				'payment_result' => $payment_result,
 			],
 			$request
@@ -574,4 +579,38 @@ class Checkout extends AbstractRoute {
 
 		return $payment_data;
 	}
+
+	/**
+	 * Order processing relating to customer account.
+	 *
+	 * - Creates a customer account as needed (based on request & store settings).
+	 *   - If so, updates the order with the new customer ID.
+	 * - Updates the order with user details (e.g. address).
+	 *
+	 * @param \WC_Order        $order   Order object.
+	 * @param \WP_REST_Request $request Request object.
+	 */
+	protected function process_customer( \WC_Order $order, \WP_REST_Request $request ) {
+		$order_controller = new OrderController();
+
+		// Create a new user account as necessary.
+		// Note - CreateAccount class includes feature gating logic (i.e. this
+		// may not create an account depending on build).
+		if ( defined( 'WC_VERSION' ) && version_compare( WC_VERSION, '4.7', '>=' ) ) {
+			// Checkout signup is feature gated to WooCommerce 4.7 and newer;
+			// Because it requires updated my-account/lost-password screen in 4.7+
+			// for setting initial password.
+			try {
+				$create_account = Package::container()->get( CreateAccount::class );
+				$create_account->from_order_request( $request );
+				$order->set_customer_id( get_current_user_id() );
+			} catch ( Exception $error ) {
+				$this->handle_error( $error );
+			}
+		}
+
+		// Persist customer address data to account.
+		$order_controller->sync_customer_data_with_order( $order );
+	}
+
 }
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -2,6 +2,10 @@
 namespace Automattic\WooCommerce\Blocks\StoreApi\Routes;
 
 use \Exception;
+use \WP_Error;
+use \WP_REST_Server;
+use \WP_REST_Request;
+use \WP_REST_Response;
 use Automattic\WooCommerce\Blocks\Package;
 use Automattic\WooCommerce\Blocks\Domain\Services\CreateAccount;
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\CartController;
@@ -29,10 +33,10 @@ class Checkout extends AbstractRoute {
 	/**
 	 * Enforce nonces for all checkout endpoints.
 	 *
-	 * @param \WP_REST_Request $request Request object.
-	 * @return \WP_Error|\WP_REST_Response
+	 * @param WP_REST_Request $request Request object.
+	 * @return WP_Error| WP_REST_Response
 	 */
-	public function get_response( \WP_REST_Request $request ) {
+	public function get_response( WP_REST_Request $request ) {
 		$this->maybe_load_cart();
 		$response = null;
 		try {
@@ -40,7 +44,7 @@ class Checkout extends AbstractRoute {
 			$response = parent::get_response( $request );
 		} catch ( RouteException $error ) {
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode() );
-		} catch ( \Exception $error ) {
+		} catch ( Exception $error ) {
 			$response = $this->get_route_error_response( 'unknown_server_error', $error->getMessage(), 500 );
 		}
 		return $response;
@@ -54,7 +58,7 @@ class Checkout extends AbstractRoute {
 	public function get_args() {
 		return [
 			[
-				'methods'             => \WP_REST_Server::READABLE,
+				'methods'             => WP_REST_Server::READABLE,
 				'callback'            => [ $this, 'get_response' ],
 				'permission_callback' => '__return_true',
 				'args'                => [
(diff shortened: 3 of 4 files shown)


senadir:
> The new hook for Woo Subs use case (__experimental_woocommerce_blocks_checkout_order_processed) is still present for now, but will not be needed if Woo Subs switches to using standard Woo core filters (PR in progress - woocommerce/woocommerce-subscriptions#3883).

**This hook has nothing to do with account creation**, it is used for creating subscriptions from a cart that contains subscription products.

What woocommerce/woocommerce-subscriptions#3883 solves is mainly moving from using a custom complicated account forcing mechanism to just switching the value of `checkout->is_registration_required()` and letting Checkout handle how it can be forced.

**So we shouldn't delete that action if the above PR is merged.**

mikejolley:
r/e the `customer_id` response on fail, we _could_ do this for the `checkout` route like what is done in the abstract/cart — override `get_route_error_response` and return it there. Is that what you tried?

The other thing would just be to ensure `X-WC-Store-API-Nonce` is returned as a header when receiving an error so our middleware can catch it.

haszari:
> So we shouldn't delete that action if the above PR is merged.

Woops - thanks for pointing that out @senadir , I was confusing these new hooks! I'll update the comment to clarify the purpose, and keep this in mind when testing.

Do you think we should move this down into step 4, after validating the order (coupons etc), immediately before payment?

haszari:
> r/e the customer_id response on fail, we could do this for the checkout route like what is done in the abstract/cart — override get_route_error_response and return it there. Is that what you tried?

I didn't try that exactly @mikejolley - I did it in `AbstractRoute`, for all routes (!), using global get logged in user func to get the id.

I'll revisit this today, override `get_route_error_response` in checkout, track the newly created customer ID as a member, and return it when there's a new customer.

> The other thing would just be to ensure `X-WC-Store-API-Nonce` is returned as a header when receiving an error so our middleware can catch it.

Agreed, I'll use the approach in #3429 - refresh the nonce near the end of request processing.

Sounds like it's best to handle returning the ID and valid nonce as part of this PR, and then once this is merged, can rebase #3429 to fix up the logic on client.

haszari:
Updates…

#### Working well
- Now firing `__experimental_woocommerce_blocks_checkout_order_processed`. @senadir I've tweaked the comment for the hook to add more info, and tested buying a subscription - seems to work the same as current `trunk`. 🎉 
- Returning the `user_id` for all checkout error responses, via `Checkout::get_route_error_response()`. This seems simple and safe. Will need JS changes to make use of it (similar to #3429). 🎉 

#### Blocker
- ❌ Having issues refreshing the nonce for error requests. 

The problem here is that the way to set the nonce is to call a method on `WP_Rest_Response` - `$response->header( 'X-WC-Store-API-Nonce' … )`. When we're returning an error, e.g. coupon validation, or no stock, we may still need to return a new nonce (because customer account may have been created => login change). 

When the request is an error, we typically return `WP_Error` and WP converts that into a response for us. I.e. we don't have a response to call `->header()` on. 

@mikejolley  Do you know how we can solve this? 

One idea would be that we do our own error responses manually, instead of returning `WP_Error`, especially for common/expected errors like coupon validation. 

Or there may be another PHP mechanism for setting header that works well for this case. 

Note - this problem affects #3429 as well.

mikejolley:
@haszari I've not tested the full flow, but I've pushed a commit which exposes the nonce on all API responses, including Error responses. I had to borrow some code from WP core to make it happen but that should cater for our needs.

senadir:
Note that I can't add products from All Products for guests (X-WC-Store-API-Nonce is invalid.)

senadir:
I'm also going to copy the same comment I had in #3423, this change would cause regression and will undo #3260.
The order object will have unsaved data (customer ID) so code hooking into __experimental_woocommerce_blocks_checkout_order_processed will not have access to the latest order data.
image

#3260 fixed this issue by changing the order so that we save the Order after we finished creating an account but just before triggering the hook.

You can keep this PR as is (with minimal change) either by:
- Saving the order when you create a customer (therefore saving to DB twice in a row, which is not ideal).
- Not saving the order in `update_order_from_request` and doing one save before triggering the action.

Of course, WooCommerce Subscription and other plugins can simply use the object passed to them instead of reaching for `wc_get_order( $order->get_id() )` but we already discussed why this isn't ideal and would create bugs down the road for other plugins.

haszari:
> I've pushed a commit which exposes the nonce on all API responses, including Error responses. I had to borrow some code from WP core to make it happen but that should cater for our needs.

Awesome, thanks for that @mikejolley ! I've added a fix to catch this in the client (bd49ccca0050a5c6d505d8b75f16e55d03370db5) and now the "create account + invalid coupon" flow works correctly: submit => (account created) bad coupon message, removed from order => resubmit => success 🎉 

So that's that part is functional now. We can add more UX polish in #3429 as follow up. cc @opr who was excited about making this nicer with a "you have an account" notice :)

haszari:
> this change would cause regression

@senadir Just to clarify the issue here, I looked at the PR (no issue) and I think this is the bug - let me know if I've missed anything:

- Checkout block + WooCommerce Subscriptions enabled. Store should have a subscription product set up.
- Clone [this branch from subscriptions](https://github.com/woocommerce/woocommerce-subscriptions/pull/3825). This hooks up subscriptions order processing to create a subscription for blocks/Store API orders.
- As a guest / anon user, add a subscription product to cart.
- Ensure an account will be created - either store settings, or check `Create an account` in front end.
- Complete checkout and submit. Recommend using Stripe CC to pay (I think there are issues with other payment methods).
- Expected: new customer account, successful order, subscription created alongside order.
- Actual: order fails with `Invalid subscription customer_id.`

Screen Shot 2020-11-26 at 9 52 07 AM

Sounds like the simplest fix (at least for now) is to `save()` the order before firing the hook; it seems sensible to ensure that the passed `$order` matches the state in DB before calling any hook. Maybe we can optimise this later if needed, or require that extensions only use the order passed to the hook (I'm not sure of the tradeoffs with that approach).

I added this extra save (9933c2008603beda56e06a84b91036805e8eb365) and tested (with subs branch) and confirmed that use case (steps above) works correctly. 

> Note that I can't add products from All Products for guests (X-WC-Store-API-Nonce is invalid.)

@senadir Are you referring to [`All Products for WooCommerce Subscriptions`](https://woocommerce.com/products/all-products-for-woocommerce-subscriptions/?quid=46bb4b499eb4859ee19ae47471d5364b) or something else? Perhaps try again with the new nonce fix - it was sent, but not refreshed in the client-side store (bd49ccc).

If there is still an issue, perhaps we can work on this as follow up, since Subscriptions / extensions support is a work in progress right now.

haszari:
This is ready for review now! Would appreciate a thorough review (or 2) and lots of edge-case testing, since this is a moderate change to a critical API (checkout processing).

senadir:
> Are you referring to All Products for WooCommerce Subscriptions or something else? Perhaps try again with the new nonce fix - it was sent, but not refreshed in the client-side store (bd49ccc).

No, I was referring to our All Products block.
It works fine now.

senadir:
I tested the current implementation and it works fine (with subs).
So I went ahead and removed `$order->save()` from `update_order_from_request` so that we only save once, this still worked fine and produced no issues because we never reach for the DB in that function.

Now, the issue we're solving by introducing another `save` call could be introduced again if we decide to create another action in that flow, but I don't any reason to introduce a new action there and prefer to solve this issue once we decide to add a new action.

So I'd like to proceed with the above suggestion (deleting [this line]( https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/9933c2008603beda56e06a84b91036805e8eb365/src/StoreApi/Routes/Checkout.php#L467)) and do some testing on it as well.

haszari:
> removed $order->save() from `update_order_from_request` so that we only save once, this still worked fine and produced no issues because we never reach for the DB in that function.
> So I'd like to proceed with the above suggestion (deleting this line) and do some testing on it as well.

This makes sense - I can commit that change today and do some testing to be sure. It might be more robust to leave it in place, I'm not clear on the performance implications (keen for review feedback on that). @senadir If I don't get to this before I go AFK, please feel free to commit to this PR :)

assets/js/base/context/cart-checkout/checkout/processor/index.js
```diff
@@ -216,8 +216,11 @@ const CheckoutProcessor = () => {
 					setIsProcessingOrder( false );
 				} );
 			} )
-			.catch( ( error ) => {
-				error.json().then( function ( response ) {
+			.catch( ( fetchResponse ) => {
+				fetchResponse.json().then( function ( response ) {
```
nerrad: I'm curious why the change on argument name here? `error` was used because this is a catch callback (so it's just a nudge that this is handling an error response). If anything, should the variable name be changed I would suggest `errorResponse` instead.

haszari:
@senadir Regarding removing that extra `$order->save()`, I'm less confident that we should attempt to optimise now.

Tracing through the logic in `Checkout::get_route_post_response()` (and subroutines) I see multiple places where we modify `$order` or `$customer` object, and a multiple places where we (re)save them too. 

For example, `$customer` is saved in both [`Checkout::update_customer_from_request()`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fix/checkout-signup-order-processing-logic/src/StoreApi/Routes/Checkout.php#L449) (refactored on this branch) and  [`OrderController::sync_customer_data_with_order()`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fix/checkout-signup-order-processing-logic/src/StoreApi/Utilities/OrderController.php#L96).

I agree that it would be better to coordinate saving these entities to the database only once per POST request, but I'm not sure if that's practical, since we will likely have more extensibility hooks over time (and extensions might assume everything is persisted to the database when a hook is fired).

Or put another way - I don't think it should block this PR; we could follow up to ensure we are using a consistent approach, and ensure that approach isn't causing a performance issue (or an extensibility issue!). Interested to hear feedback from reviewers about this (cc @mikejolley).

**Update**: Just realised while working on #3429 - checkout processing can be aborted at any stage if anything throws an exception. So it's probably a good idea for us to persist objects early and often, so any changes are saved even if there is an error.

opr:
I've done some testing with various conditions, here are the results:

- ✅ Coupon on order with ineligible email - account created and order does not succeed
- ❌ Cannot proceed after this still because of the Invalid cookie nonce error
- ✅ Checkout without choosing to create account - succeeds and no account is created
- ✅ Checkout while opting to create account - order succeeds and account is created
- ✅ Stripe card declined - account created and order not created
- ⚠️ Out of stock product in order, good because the order doesn't succeed but small issue with how the error displays (&amp;quot; appears around the product name) Screenshot 2020-11-26 at 21 19 51
- ^ #3469 will take care of this

haszari:
Thanks for testing!

> Cannot proceed after this still because of the Invalid cookie nonce error

@opr Can you list exact steps for reproducing this? I've tested and confirmed this is fixed with latest changes, maybe I missed a use case. Make sure you're on the latest code and have rebuilt (this PR has been rebased recently).

opr:
> Can you list exact steps for reproducing this? I've tested and confirmed this is fixed with latest changes, maybe I missed a use case. Make sure you're on the latest code and have rebuilt (this PR has been rebased recently).

Sure, I did run `npm build` too, so I think everything is correct.

1. Create a coupon with an email restriction (I used _*@test.com_)
2. Ensure users can create accounts at checkout at block level and store level.
3. Go to the front-end and add two products, (I chose beanie and beanie with logo from the default products)
4. Go to a page with the checkout block, and apply the coupon from step 1.
4. Fill in all fields, and tick "Create account", ensuring your email is not valid for the coupon
5. Try to checkout, and get an error message.
6. Try to immediately check out again

See a video of the process here: https://www.loom.com/share/fe82275fada14f2cbc99ee34800b55ad

haszari:
Dang - I can't reproduce that @opr - maybe there's something hinky happening with a build. Here's a walkthrough from 21e5cd8703730d76563845ad68bf220f977dfa5f on this branch. 🤞🏻I'm hoping this was a stale JS build in your test environment…

https://d.pr/i/ooWTix (big gif, no embed)

nerrad:
I just tested to try and reproduce the invalid cookie nonce error as well and was unable to reproduce so it appears to be fixed from my testing.

src/StoreApi/Routes/AbstractRoute.php
```diff
@@ -70,17 +70,69 @@ public function get_response( \WP_REST_Request $request ) {
...
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode(), $error->getAdditionalData() );
 		} catch ( \Exception $error ) {
 			$response = $this->get_route_error_response( 'unknown_server_error', $error->getMessage(), 500 );
 		}
+
+		if ( is_wp_error( $response ) ) {
+			$response = $this->error_to_response( $response );
+		}
+
+		return $this->add_response_headers( $response );
```
nerrad: Are we sure we want to return a nonce header for _every_ error response?

src/StoreApi/Routes/AbstractRoute.php
```diff
@@ -70,17 +70,69 @@ public function get_response( \WP_REST_Request $request ) {
...
+	}
+
+	/**
+	 * Add Store API headers to a response object.
+	 *
+	 * @param \WP_REST_Response $response Response object.
+	 * @return \WP_REST_Response
+	 */
+	protected function add_response_headers( $response ) {
```
nerrad: I'm not sure on the need for this abstraction right now. Do we anticipate returning additional response headers here eventually or will there be cases where we only want a certain set of headers added to a response? Will this be called from a variety of different places? 

I would prefer we avoid adding this abstraction until we know it's necessary - not only that, but current implementation of the method is ambiguous - when seeing this method implemented, the first question I would have is, "what response headers are added"?

src/StoreApi/Routes/AbstractRoute.php
```diff
@@ -70,17 +70,69 @@ public function get_response( \WP_REST_Request $request ) {
...
-			}
 		} catch ( RouteException $error ) {
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode(), $error->getAdditionalData() );
 		} catch ( \Exception $error ) {
 			$response = $this->get_route_error_response( 'unknown_server_error', $error->getMessage(), 500 );
 		}
+
+		if ( is_wp_error( $response ) ) {
+			$response = $this->error_to_response( $response );
+		}
```
nerrad, line 81: `WP_Rest_Server` [already handles adding `wp_errors` to the response](https://github.com/WordPress/wordpress-develop/blob/69c3ab8a118679b3f560b7f332068e180d6d7310/src/wp-includes/rest-api/class-wp-rest-server.php#L439-L441) why do we need to do it here (and also create our own `error_to_response` method)?

Is it because we need to add the nonce header to the response after the conversion (instead of returning a wp_error)? If so, it looks like we could use the [`rest_post_dispatch` filter](https://github.com/WordPress/wordpress-develop/blob/69c3ab8a118679b3f560b7f332068e180d6d7310/src/wp-includes/rest-api/class-wp-rest-server.php#L463) instead to enhance with the nonce header after the wp_error conversion?

src/StoreApi/Routes/Checkout.php
```diff
@@ -139,71 +139,86 @@ protected function get_route_update_response( \WP_REST_Request $request ) {
...
-				$create_account = Package::container()->get( CreateAccount::class );
-				$create_account->from_order_request( $request );
-				$order_object->set_customer_id( get_current_user_id() );
-			} catch ( Exception $error ) {
-				$this->handle_error( $error );
-			}
+		// Validate payment method before proceeding.
+		if ( $order->needs_payment() ) {
+			// Will throw and return 400 if payment method missing or invalid.
+			$this->get_request_payment_method_id( $request );
```
nerrad: This method is being used to validate, when it explicitly returns a payment method for the given id. Instead, we should maybe modify `process_payment` so it receives an optional `$payment_method_id` as an argument, then you can retrieve the payment method here and feed it into `process_payment`. Then `process_payment` only needs to call `get_request_payment_method_id` if it's not present.

So `Checkout::process_payment` would have this as its signature:

```php
protected function process_payment( \WC_Order $order, \WP_REST_Request $request, string $payment_method_id  = '' ) { 
    /** ... **/
   $context->set_payment_method( $payment_method_id === '' ? $this->get_request_payment_method_id( $request ) : $payment_method_id );
    /** ... **/
}
```

src/StoreApi/Routes/Checkout.php
```diff
@@ -139,71 +139,86 @@ protected function get_route_update_response( \WP_REST_Request $request ) {
...
-		* But we're opting for a new action because the original ones attaches POST data.
-		* NOTE: this hook is still experimental, and might change or get removed.
-		* @todo: Document and stabilize __experimental_woocommerce_blocks_checkout_order_processed
+		* Action: woocommerce_blocks_checkout_order_processed (experimental).
+		* This hook informs extensions that $order has completed processing and is ready
+		* for payment.
+		*
+		* This is similar to existing core hook woocommerce_checkout_order_processed.
+		* We're using a new action:
+		* - To keep the interface focused (only pass $order, not passing request data).
```
nerrad: I think this is a good initial start. I do wonder however if some extensions might need the request data for any additional information that might have come through in the request. That's unknown at this point though and not something we expose as an option client side for the moment.

src/StoreApi/Routes/Checkout.php
```diff
@@ -139,71 +139,86 @@ protected function get_route_update_response( \WP_REST_Request $request ) {
...
+		* - This also explicitly indicates these orders are from checkout block/StoreAPI.
+		*
+		* Status: experimental; may change or be removed.
+		* Introduced: WooCommerce Blocks 3.8.0.
+		* https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3238
 		*/
-		do_action( '__experimental_woocommerce_blocks_checkout_order_processed', $order_object );
+		// Save the order so the param and database match (extensions may access DB).
+		$order->save();
+		do_action( '__experimental_woocommerce_blocks_checkout_order_processed', $order );
```
nerrad: Please follow the [WordPress documentation standards for action/filter hooks](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters). I know we haven't been consistent with this but I'd like us to fix that.

src/StoreApi/Routes/Checkout.php
```diff
@@ -139,71 +139,86 @@ protected function get_route_update_response( \WP_REST_Request $request ) {
...
+		// --
+		// 5. Process Payment
 
-		if ( ! $order_object->needs_payment() ) {
-			$payment_result = $this->process_without_payment( $order_object, $request );
+		if ( ! $order->needs_payment() ) {
+			$payment_result = $this->process_without_payment( $order, $request );
 		} else {
-			$payment_result = $this->process_payment( $order_object, $request );
+			$payment_result = $this->process_payment( $order, $request );
```
nerrad: This is where you'd feed in the `$payment_method_id` if you implement my suggested changes above.

src/StoreApi/Routes/Checkout.php
```diff
@@ -574,4 +603,38 @@ protected function get_request_payment_data( \WP_REST_Request $request ) {
...
+	 *
+	 * - Creates a customer account as needed (based on request & store settings).
+	 *   - If so, updates the order with the new customer ID.
+	 * - Updates the order with user details (e.g. address).
+	 *
+	 * @param \WC_Order        $order   Order object.
+	 * @param \WP_REST_Request $request Request object.
+	 */
+	protected function process_customer( \WC_Order $order, \WP_REST_Request $request ) {
+		$order_controller = new OrderController();
```
nerrad, line 554: I know this pattern wasn't established in this pr. But as a follow-up issue, I'd like to see us inject `OrderController` and `CartController` as dependencies for this class (I would make it a requirement before making the StoreAPI public).

nerrad (commented):
Starting to look pretty good! I'll defer to others regarding the order of processing changes made but I did a review of the implementation/syntax and have a few comments I'd like see addressed.

src/StoreApi/Routes/AbstractRoute.php
mikejolley: `rest_post_dispatch` potentially runs on every REST response from all APIs. Converting it ourselves gives us full, contained, control over error responses in the Store API without affecting non-store API endpoints.

Also pointing out that feedback above states:

> Are we sure we want to return a nonce header for every error response?

We would need a way to identify store API requests with this filter, and then further if we wanted to only affect *some* Store API requests. Just pointing this out because it's a downside to using the filter.

src/StoreApi/Routes/AbstractRoute.php
mikejolley: Yes; we should include the response in all Store API responses because:

a) It gives the client consistent headers to capture through a middleware
b) A login can occur outside of the Store API flow, so theoretically the nonce could refresh at any time (we may need a retry mechanism at some point)

src/StoreApi/Routes/Checkout.php
mikejolley: Why doesn't the linting catch this?

src/StoreApi/Routes/AbstractRoute.php
mikejolley: As an alternative, if the intention is to split up and group logic, but not abstract this, just make it private?

src/StoreApi/Routes/Checkout.php
senadir: Yes, this could happen, the original hook passes `$_POST` data.
However, passing $_POST was required because people who render anything in the checkout form and passing `POST`  would provide a way to catch it.
We're not planning on doing that, and since we do everything via REST, any passed back data should via an explicit field and via an explicit mechanism.
Any core data in `$request` that is not present in the order is data we might not want to expose, and exposing `$request` might be too much implementation detail.
So if there's something that we need from `$request` that is not present in order, we can expose it explicitly in the hook.

src/StoreApi/Routes/Checkout.php
senadir: TIL!

src/StoreApi/Routes/AbstractRoute.php
nerrad: If we hook into the filter only when this route is being processed it will end up only getting processed on this route right? An advantage of using the filter is that if we add it to a late priority (eg `999`) we also ensure the header is added at the last possible moment?

Loosely held opinion, but I like to avoid creating new code unnecessarily if possible here.

src/StoreApi/Routes/AbstractRoute.php
nerrad: Explanation makes sense. I just want to make sure we're not exposing an easy way for bad actors to grab nonces for malicious use  - I don't think they can because the request is still tied to either a cart session or an authenticated request.

src/StoreApi/Routes/AbstractRoute.php
nerrad: Sure, but the naming is still ambiguous, it's only adding the store api nonce header at the moment. Currently this method is only called in one place and it could easily be replaced with just adding the header to the response in that one place. It seems like premature abstraction for something we might not even use to add additional headers to the response for.

src/StoreApi/Routes/Checkout.php
nerrad: Right, I'm not suggesting we change this signature now. Just thought it was important to note that there may be need for passing along additional information from the request that might not be exposed on the order. It'd be good to think of how we could do that (but not for this PR).

src/StoreApi/Routes/Checkout.php
nerrad: > Why doesn't the linting catch this?

Ya I was surprised too.

src/StoreApi/Routes/Checkout.php
mikejolley: I think we can also move validation to the schema (enum property)

src/StoreApi/Routes/AbstractRoute.php
mikejolley: > An advantage of using the filter is that if we add it to a late priority (eg 999) we also ensure the header is added at the last possible moment?

This is not an advantage against inline header addition which is only done on Store API responses—late addition is of no concern with the current approach.

Another advantage of inline is that our `get_response` methods only ever return valid `WP_REST_Response` objects, rather than __maybe__ returning a `WP_Error` object, so it's more predictable. The code introduced is fairly small and does enable us to further customise the error responses if needed.

src/StoreApi/Routes/AbstractRoute.php
```diff
@@ -70,17 +70,51 @@ public function get_response( \WP_REST_Request $request ) {
...
+
+		$response->header( 'X-WC-Store-API-Nonce', wp_create_nonce( 'wc_store_api' ) );
+		$response->header( 'X-WC-Store-API-User', get_current_user_id() );
```
mikejolley, line 84: @haszari this is instead of user_id in responses. It's not consumed anywhere yet right?

nerrad: It looks like #3429 would need updated to get the user id.

src/StoreApi/Routes/Checkout.php
```diff
@@ -89,49 +101,79 @@ public function get_args() {
...
 	}
 
+	/**
+	 * Prepare a single item for response. Handles setting the status based on the payment result.
+	 *
+	 * @param mixed            $item Item to format to schema.
+	 * @param \WP_REST_Request $request Request object.
+	 * @return \WP_REST_Response $response Response data.
+	 */
+	public function prepare_item_for_response( $item, \WP_REST_Request $request ) {
```
nerrad: We shouldn't need the root `\` here with the import at the top of the file right?

src/StoreApi/Routes/Checkout.php
```diff
@@ -330,53 +333,44 @@ protected function is_valid_draft_order( $order_object ) {
...
-	 *
-	 * @return \WC_Order Order object.
 	 */
-	protected function create_or_update_draft_order() {
+	private function create_or_update_draft_order() {
 		$cart_controller  = new CartController();
 		$order_controller = new OrderController();
 		$reserve_stock    = \class_exists( '\Automattic\WooCommerce\Checkout\Helpers\ReserveStock' ) ? new \Automattic\WooCommerce\Checkout\Helpers\ReserveStock() : new ReserveStock();
-		$order_object     = $this->get_draft_order_id() ? wc_get_order( $this->get_draft_order_id() ) : null;
 		$created          = false;
```
nerrad: I know this variable wasn't introduced in this pr but I noticed in reviewing, what purpose does `$created` have in this method? It doesn't seem to be implemented anywhere (other than switched to `true` when creating the order)?

src/StoreApi/Routes/Checkout.php
```diff
@@ -141,105 +183,80 @@ protected function get_route_update_response( \WP_REST_Request $request ) {
...
 		// Validate payment method before proceeding.
-		if ( $order->needs_payment() ) {
-			// Will throw and return 400 if payment method missing or invalid.
-			$this->get_request_payment_method_id( $request );
+		if ( $this->order->needs_payment() && empty( $request['payment_method'] ) ) {
+			throw new RouteException(
+				'woocommerce_rest_checkout_missing_payment_method',
+				__( 'No payment method provided.', 'woo-gutenberg-products-block' ),
+				400
+			);
```
nerrad: `Checkout::update_order_from_request` already checks to see if `$request['payment_method']` exists. I think we could move this validation inside there to keep the logic flow in this method cleaner? Also it would help surface this error earlier in the stack because `update_order_from_request` already does some validation around whether a payment method exists for the payment method id passed in the request.

src/StoreApi/Routes/Checkout.php
```diff
@@ -627,14 +567,31 @@ protected function process_customer( \WC_Order $order, \WP_REST_Request $request
...
+							'registration-error-invalid-email',
+							__( 'Please provide a valid email address.', 'woo-gutenberg-products-block' ),
+							400
+						);
+					case 'registration-error-email-exists':
+						throw new RouteException(
+							'registration-error-email-exists',
+							apply_filters(
+								'woocommerce_registration_error_email_exists',
+								__( 'An account is already registered with your email address. Please log in.', 'woo-gutenberg-products-block' )
```
nerrad: I know this was moved here from `handle_error` method so was pre-existing to this pr, however we should probably implement WordPress coding standards for filters here. Might also be worth adding an inline dev note that any html will be escaped on the frontend when this is added to a notice in the blocks (because our notices system escapes html right?).

nerrad (approved):
Nice bit of refactoring here Mike in simplifying the logic flow, and doing a bit of code cleanup around scope and removing unnecessary methods. 

I have a few comments that are not blocking - you can address as you see fit so I'm pre-approving. However _I did not test_ as it requires more time than I can give at the moment for testing. So up to you whether you are happy with your testing of this pr or if you want to see if someone else can confirm testing.

src/StoreApi/Routes/Checkout.php
mikejolley: I think we should remove this filter (I'm going to)—I can see it was added to mirror core, however, the filtered strings are different and the core filter allows for HTML whereas this would be more limited. Since this is an error from the API it does not need to be filterable. The client can change the wording based on the error code cc @haszari

src/StoreApi/Routes/Checkout.php
mikejolley: That will need tweaking because it's required on POST but not on UPDATE right now. Ill see if that difference is still needed.

PR #3455, opened 2020-11-24
Ensure correct alignment of checkout notice's dismiss button

### LINKED ISSUE #3116: Checkout block error notice dismiss X is not right-aligned
## Describe the bug
The notices displayed in the Checkout block (in WooCommerce Blocks plugin) sometimes have a dismiss `x` button. 

In Storefront this is rendering next to the text; it should be right-aligned. It's working correctly in Twenty Twenty so I'm reporting as an issue against Storefront.

Screen Shot 2020-09-03 at 10 30 58 AM
Screen Shot 2020-09-03 at 10 32 02 AM

I think these notices are based in part on [`@wordpress/components` `Notice`](https://github.com/WordPress/gutenberg/tree/master/packages/components/src/notice), so there might be other ways to reproduce this.

## To Reproduce
1. Install and activate [WooCommerce Blocks](https://wordpress.org/plugins/woo-gutenberg-products-block/).
2. Edit Checkout page, delete shortcode, and add the Checkout block. Publish/save changes.
3. On front end, add something to cart and proceed to checkout.
4. In checkout, leave one or more required fields empty and submit checkout. Observe the error notice.

opr:
The change in this PR ensures the dismiss icon in notices will always be as far over to the right of the parent container as possible. Previously, it sat just alongside the text in the Storefront theme. Some themes override the position of the cross, so please test in Storefront where the original issue was, but bonus points for testing in other themes too.

Fixes #3116 

#### Before
Screenshot 2020-11-24 at 09 57 57

#### After
Screenshot 2020-11-24 at 09 58 35

1.  Enable Storefront theme
2. Add an item to the cart and go to the checkout block
3. Leave a required field on the checkout form blank and attempt checkout
4. Ensure the error notice's dismiss button is aligned to the right

### DIFF
--- a/assets/js/base/components/store-notices-container/style.scss
+++ b/assets/js/base/components/store-notices-container/style.scss
@@ -8,7 +8,7 @@
 		.components-notice__dismiss {
 			background: transparent none;
 			padding: 0;
-			margin: 0 0 0 0.5em;
+			margin: 0 0 0 auto;
 			border: 0;
 			outline: 0;
 			color: #fff;


Aljullu (approved):
Nice one-line fix. :shipit:

PR #3457, opened 2020-11-24
Update @wordpress/components to v. 11.1.1 and @wordpress/base-styles to v. 3.2.0

### LINKED ISSUE #2140: Evaluate if we can update to wordpress/components to or 9.0.0 9.2.0 or later
Currently, we use wordpress-components 8.5 in our project, it introduced a large increase in bundle size since it's not treeshaked, we should aim to upgrade it or remove it completely from our frontend bundle.

If we update to 9.0.0:
- We get treeshaking, it reduces the bundle size, however dashicon is still included, so we should move away from Panel and Panel body components.

If we upgrade to 9.2.0:
- We get treeshaking and removed dashicon, we still has to move away from card component since it includes extra dependencies like emotion that increases the bundle size (and bring no value to us).

### LINKED ISSUE #3146: JS warning when changing the value of the country or state select
## To reproduce
You need to have Gutenberg enabled in order to reproduce this issue.

1. Go to the Checkout block.
2. Change the value of the Country/Region select.
3. In the browser console, notice there is a warning: ``Warning: Cannot update a component (`ShippingDataProvider`) while rendering a different component (`CustomSelectControl`). To locate the bad setState() call inside `CustomSelectControl`, follow the stack trace as described in https://fb.me/setstate-in-render``

## Screenshots
Peek 2020-09-15 11-24
imatge

### LINKED ISSUE #3311: remove when `@woocommerce/base-styles` is updated.
remove when &#x60;@woocommerce/base-styles&#x60; is updated. Greys Use only when you truly need pure black. For UI, use $gray-900. Used for light gray backgrounds. Used sparingly for light borders. Used for most borders.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c8de44956cc0071565950bfbac6fb66352897159/assets/css/abstracts/_colors.scss#L4-L20
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in c8de44956cc0071565950bfbac6fb66352897159 when #3300 was merged. cc @Aljullu

### LINKED ISSUE #3313: replace with $grid-unit values when `@woocommerce/base-st...
replace with $grid-unit values when &#x60;@woocommerce/base-styles&#x60; is updated. github.com/WordPress/gutenberg/blob/master/packages/base-styles/_variables.scss#L23-L36 $grid-unit-50

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c8de44956cc0071565950bfbac6fb66352897159/assets/css/abstracts/_variables.scss#L3-L14
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in c8de44956cc0071565950bfbac6fb66352897159 when #3300 was merged. cc @Aljullu

Aljullu:
Fixes #2140.
Fixes #3146.
Fixes #3311.
Fixes #3313.

#### Regression testing

1. Verify you can build the project without problems: `npm ci && npm run build`.
1. Create a page with all WC Blocks (you can copy and paste —with <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd>— [this gist](https://gist.github.com/Aljullu/d9c76ed510ae6973bd41dcc4184e8eb2)) + add the Single Product block & some other Product element blocks.
2. In the editor, verify all of them render correctly, you can modify their attributes and there are no errors in the browser console.
3. In the frontend, also verify they render correctly, you can interact with them and there are no errors in the browser console.

#### Testing #3146

1. Go to the Checkout block.
2. Change the value of the Country/Region select.
3. Verify no errors appear in the browser console.


Aljullu:
Using the `explore` script (thanks @senadir for adding it!), I analyzed whether the bundle size increase was legit or not. For example, the Checkout frontend script before/after is as follows:

_Before:_
imatge

_After:_
imatge

Most of the size increase comes from `@wordpress/components`, which goes from 34.99KB to 56.11KB. `@wordpress/icons` and `@wordpress/primitives` seem to only add 2.33KB.

I investigated a little bit more why `@wordpress/components` increased so much (are we including components that we don't use?). But I couldn't spot anything off: increases seem legit. As an example, the Button component increased from 774B to 2.47KB, but that's just because more features have been added. I was able to reduce 1KB replacing a `for ... of` with a regular `for` loop, but this kind of optimizations would need to be done upstream, and they might make code less-readable, so I'm not sure if creating a PR in GB's repo makes sense at all.

**tl;dr:** file increases seem legit and there is not an easy way to avoid them other than making optimizations upstream so components are slimmer.

senadir:
note that bundle comment shows gzipped while the image you showed above is just minified.

Downshift, reakit, and spring-js are the big offenders here, but what changed between the before and after is Downshift (from 10 to 17).

I don't think there's much we can do here really. Our reliance on certain components causes this.

We can eliminate Downshift by using a regular select item instead of downshift and other stuff (see this PR https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2281).

I don't think we can do anything in this PR, but having the latest wp.components would allow us to provide fixes upstream.

assets/css/abstracts/_colors.scss
```diff
@@ -1,18 +1,6 @@
...
-// Greys
-$black: #000;			// Use only when you truly need pure black. For UI, use $gray-900.
-$gray-100: #f0f0f0;		// Used for light gray backgrounds.
-$gray-200: #e0e0e0;		// Used sparingly for light borders.
-$gray-300: #ddd;		// Used for most borders.
-$gray-400: #ccc;
-$gray-600: #949494;		// Meets 3:1 UI or large text contrast against white.
-$gray-700: #757575;		// Meets 4.6:1 text contrast against white.
-$gray-900: #1e1e1e;
-$white: #fff;
```
Aljullu, line 14: Those variables are defined in recent versions of `@wordpress/base-styles`, so there is no need to define them here too.

assets/css/abstracts/_variables.scss
```diff
@@ -1,14 +1,12 @@
...
-$gap-small: 12px; // $grid-unit-15
-$gap-smaller: 8px; // $grid-unit-10
-$gap-smallest: 4px; // $grid-unit-05
+$gap-largest: $grid-unit-50;
+$gap-larger: 4.5 * $grid-unit;
+$gap-large: $grid-unit-30;
+$gap: $grid-unit-20;
+$gap-small: $grid-unit-15;
+$gap-smaller: $grid-unit-10;
+$gap-smallest: $grid-unit-05;
```
Aljullu, line 9: Instead of hardcoding px values, let's use the `$grid-unit` variables from `@wordpress/components`.

assets/js/atomic/blocks/product-elements/title/frontend.js
```diff
@@ -10,4 +9,4 @@ import withFilteredAttributes from '@woocommerce/base-hocs/with-filtered-attribu
 import Block from './block';
 import attributes from './attributes';
 
-export default compose( withFilteredAttributes( attributes ) )( Block );
```
Aljullu, line 13: Since `@wordpress/compose` is no longer extracted, its dependency was generating a new frontend file specific for vendor files. Taking a closer look, it seems like `compose()` is not required here, so let's remove it.

assets/js/base/components/chip/test/__snapshots__/index.js.snap
```diff
@@ -106,9 +106,9 @@ exports[`RemovableChip should render custom aria label 1`] = `
     onKeyDown={[Function]}
   >
     <svg
-      aria-hidden="true"
+      aria-hidden={true}
       className="wc-block-components-chip__remove-icon"
-      focusable="false"
+      focusable={false}
```
Aljullu, line 111: This change was automatic when updating the snapshots and seems legit.

assets/js/base/components/payment-methods/test/payment-methods.js
```diff
@@ -53,10 +53,12 @@ describe( 'PaymentMethods', () => {
...
 
 		await waitFor( () => {
-			const noPaymentMethods = screen.queryByText(
+			const noPaymentMethods = screen.queryAllByText(
 				/no payment methods available/
 			);
-			expect( noPaymentMethods ).not.toBeNull();
+			// We might get more than one match because the `speak()` function
+			// creates an extra `div` with the notice contents used for a11y.
+			expect( noPaymentMethods.length ).toBeGreaterThanOrEqual( 1 );
```
Aljullu, line 61: Recent versions of the `Notice` component generate two elements with the notice text: one for the notice itself and another one used by the `speak()` a11y function. Our previous test was failing because it was expecting one and only one match.

assets/js/base/components/select/index.js
```diff
@@ -32,7 +32,7 @@ const Select = ( {
 					onChange( selectedItem.key );
 				} }
 				options={ options }
-				value={ value }
+				value={ value || null }
```
Aljullu, line 35: `value` might be undefined, but when it changes from undefined to a value, downshift shows a warning that the component is changing from being uncontrolled to controlled. That can be fixed setting `null` as the default value when non is selected.

senadir (approved):
Everything works fine but I'm seeing a small styling issue

image
image

If they're unrelated or are fixed, feel free to merge this.

package.json
```diff
@@ -165,7 +165,7 @@
 		"reakit": "1.3.0",
 		"trim-html": "0.1.9",
 		"use-debounce": "3.4.3",
-		"wordpress-components": "npm:@wordpress/components@8.5.0",
+		"wordpress-components": "npm:@wordpress/components@11.1.1",
```
haszari, line 168: I'm curious why we have this dep and the other one in `devDependencies`. What does this mean? Please excuse my ignorance - feel free to point me to docs 😄

haszari (approved):
Gave this a quick test - played with various blocks in editor and front end and didn't see any problems. FYI I was testing on WP 5.6-beta4.

Thanks for sorting this @Aljullu  - it's great to be on recent versions, and nice to see our base CSS variables pulling from Gutenberg/core directly. Nice work 🙌

package.json
Aljullu: Sure thing! Form the [docs](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file):
> * "dependencies": Packages required by your application in production.
> * "devDependencies": Packages that are only needed for local development and testing.

In our case:
- `wordpress-components` is bundled in frontend scripts, so it's a dependency of the production builds. Because of that, it's listed in `dependencies`.
- `@wordpress/components` is extracted by `DependencyExtractionWebpackPlugin`, so it's not bundled in the production builds, so it's listed in `devDependencies`.

For reference, dependencies were re-ordered in this #1663, and there was a similar discussion about `@wordpress/components`: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1633#discussion_r370339806.

Aljullu:
Thanks for the reviews @senadir & @haszari!

> Everything works fine but I'm seeing a small styling issue

Good catch! I could reproduce it too in WC Blocks `trunk` when using Gutenberg `master`, so it seems to be unrelated to this PR. I created an issue in WC Admin (https://github.com/woocommerce/woocommerce-admin/issues/5806).

PR #3463, opened 2020-11-25
Prevent exclude drafts filter being tree shaken

Aljullu:
In #3379 I introduced a filter so Draft order status didn't appear in WC Admin settings page. After review, [I moved the filter to its own file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3379/commits/bb7a5549723d617bf8457b9593b03473effd2486) thinking that was a low-risk change. But it was not! When Webpack tree-shakes the code on build, it will exclude that file since it's not marked as having side effects.

In order to solve this issue, there were two solutions:
1. Mark the file as having side effects in `package.json`.
2. Move the file to `/assets/js/filters/`, where we store other filters and which is already marked as having side effects.

This PR implements the second solution.

Make sure you build the plugin with `npm run build` instead of `npm run start` and then, repeat these two steps from #3379:

1. Make sure your store has at least one completed order and one draft order. You can achieve that adding some products to the cart and navigating to the Cart and Checkout blocks but without completing the order.
2. Go to Analytics > Settings and verify there isn't a `Custom statuses` section or, if it exists, make sure none of the statuses is named `Draft`.


mikejolley (approved):
I tested after building and this branch hid the status for me. Looks good to 🚢

PR #3464, opened 2020-11-25
Stop hidden/search-only product names being links in checkout sidebar

### LINKED ISSUE #3370: Cart and Checkout blocks do not respect a product's catalog visibility status

opr:
The checkout block was still displaying product names as links when the product was hidden from the catalogue. This was accidentally re-introduced when simplifying the logic used to determine whether the name should be a link or not.

This PR simplifies the logic, and passes the correct `disabled` prop to `ProductName`.

Picked up during testing by @nerrad here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3453

Fixes #3370

Original PR #3415

1. Create a page that contains the checkout and cart blocks.
2. Find a product in the dashboard and set its catalogue visibility to hidden.
3. Add that product to your cart.
4. Visit the page with the checkout and cart blocks and ensure the name of the hidden product does not link to the product page in the checkout sidebar and in the line items in the cart.
5. Repeat steps 2-4 with the catalogue visibility set to "Search results only".

### DIFF
--- a/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
+++ b/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
@@ -38,8 +38,8 @@ const OrderSummaryItem = ( { cartItem } ) => {
 		.multiply( quantity )
 		.convertPrecision( currency.minorUnit )
 		.getAmount();
-	const shouldLinkToProduct =
-		catalogVisibility !== 'hidden' && catalogVisibility !== 'search';
+	const isProductHiddenFromCatalog =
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
LGTM

PR #3465, opened 2020-11-25
Add the "Fix package lock" script

mikejolley:
Adds the script to fix renovate PRs (update package lock) to our npm scripts. Based on the script from @senadir, but updated to handle conflicts with trunk.

Run with:

```bash
npm run fix-package-lock branch/name
```

If there is a conflict with trunk, it's most likely `package-lock` because of other renovate merges. The script will try to resolve this automatically if it can, otherwise the rebase will abort.

I've tested this on several of the renovate PRs now (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3460).

### DIFF
--- a/bin/fix-package-lock.sh
+++ b/bin/fix-package-lock.sh
@@ -6,17 +6,17 @@ GREEN_BOLD='\033[1;32m';
 RED_BOLD='\033[1;31m';
 YELLOW_BOLD='\033[1;33m';
 error () {
-	echo -e "\n${RED_BOLD}$1${COLOR_RESET}\n";
+	echo "\n${RED_BOLD}$1${COLOR_RESET}\n";
 	exit 0;
 }
 status () {
-	echo -e "\n${BLUE_BOLD}$1${COLOR_RESET}\n"
+	echo "\n${BLUE_BOLD}$1${COLOR_RESET}\n"
 }
 success () {
-	echo -e "\n${GREEN_BOLD}$1${COLOR_RESET}\n"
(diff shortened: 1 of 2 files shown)


mikejolley:
I removed the rebase code—it wasn't reliable and so far it hasn't made a difference. You can rebase renovate PRs with the checkbox in the PR, then run the script, and then squash merge.

nerrad:
Maybe add a prompt to remind the person executing the script that a rebase should be done before running?

budzanowski:
Should we add `npm cache verify` for extra robustness?

bin/fix-package-lock.sh
```diff
@@ -0,0 +1,48 @@
...
+status "Removing package-lock.json...";
+rm package-lock.json
+
+status "Installing dependencies...";
+npm install
+
+status "Comitting updated package-lock.json...";
+git add package-lock.json
+git commit -m 'update package-lock.json'
+git push -f
```
haszari: [Suggest using `--force-with-lease` here for safety.](https://blog.developer.atlassian.com/force-with-lease/)

[Never force!](http://weiqingtoh.github.io/force-with-lease/)

haszari:
I added a comment about `--force-with-lease` .. but I think I'm happy just doing these manually. Adding a script seems like an over investment in a patching a temporary (🤞) problem with our renovate setup.

mikejolley:
Feedback addressed :)

> Adding a script seems like an over investment in a patching a temporary (🤞) problem with our renovate setup.

This is true, but it should help porters until then (and we don't know when it will be fixed).

budzanowski (approved):

PR #3468, opened 2020-11-26
Drop stock reservation when removing item from cart via the Store API

### LINKED ISSUE #2843: Drop stock reservation when removing item from cart
**Describe the bug**

I was exploring the behavior of reserving stock with Cart and Checkout blocks and did the following steps:

1. had some items in the cart as admin
2. switched the site to cart & checkout blocks
3. added new item to cart 
4. went to checkout, reserved stock record was created only for one item
5. went back to cart
6. removed all items from cart--the item that got reserved was still present in the wp_wc_reserved_stock--it might make sense to remove the reservation?

**Expected behavior**
Remove the stock reservation when item is removed from cart.

* Core version: 5.4.2
* WooCommerce version: 4.3.0
* WooCommerce Blocks version: 2.9.0

mikejolley:
If a user with a draft order removes an item from their cart, this PR will free up the stock held by the order manually. Usually it would be held for 10 mins.

Caveats:

- This can only capture events from the cart page/Store API. Other store events are not captured.
- Removing an item from the cart removes the hold on the stock for the entire draft order, until they checkout again. This is probably more heavy handed than ideal (?) but we only have access to methods to release stock for the entire order, not on a cart item basis.
- This cannot react to cart item changes i.e change in qty. Again, this would be heavy handed if we dropped stock on all change events.

This does fix the case in #2843 and would lesson the impact of held stock if a user abandons checkout and edits the cart, but it has limitations. Since stock is only held for 10 minutes anyway, I think this can be considered low importance/edge case.

Fixes #2843

Note, also in this PR I removed our local copy of the ReserveStock class. Our minimum supported version of WC is 4.3 so we no longer need this.

1. Add items that are stock managed to the cart.
2. Go to checkout.
3. View the contents of your wp_wc_reserved_stock table. I use Sequel Pro.
4. Go to the cart and remove an item from the cart.
5. Confirm the stock hold is deleted from wp_wc_reserved_stock.

### Changelog

> Remove held stock for a draft order if an item is removed from the cart.


budzanowski:
I can't really fathom the tradeoffs we are doing here. It looks like we are exchanging one bigger issue with a bunch of smaller issues. Since those are edge cases this should be mostly OK but, except when it happens it will be really hard to pinpoint the problem. Because our userbase is so big we will eventually get reports.

The code looks good. If you assess that this is an overall improvement, please merge.

budzanowski (approved):

PR #3469, opened 2020-11-26
Prevent checkout completing if any items are out of stock

### LINKED ISSUE #3412: Checkout block: pending order when product is out of stock, generic error message
## Describe the bug

## To reproduce

## Expected behavior

## Screenshots
Unhelpful error on front end: 

opr:
This PR will stop a checkout completing when an item on the order is out of stock. If this happens, an error will be shown to the user and the checkout will not complete.

Fixes #3412 

Related to PR #3454 

### Screenshots

Screenshot 2020-11-26 at 12 26 03

1. Add one (or more) products to the cart.
2. Ensure "Allow customers to create an account during checkout" is enabled in the WooCommerce settings and in the block specific settings too.
3. Load the checkout block and fill in all required fields. Tick create account too.
4. Make some/all (try with different combinations of which products are in and out of stock if possible) of the products out of stock.
* *Try with an item whose stock is managed at product level, and set the current stock level higher than one. Add more than one of the product to the cart, load the checkout block, then set the stock level to below the number in the cart.*
5. Try to complete the checkout process.
6. Ensure checkout does not complete and you see an appropriate error message.
7. Check that the customer's account was created and that the order did not complete.

### Automated tests
1. Please run `npm run phpunit` and verify all tests are passing. One test was added in this PR to ensure an exception is thrown during the checkout process.

### DIFF
--- a/src/StoreApi/Utilities/OrderController.php
+++ b/src/StoreApi/Utilities/OrderController.php
@@ -153,14 +153,20 @@ class OrderController {
 			);
 		}
 
-		// Ensure all items in cart are still in stock.
-		$cart_stock_check_result = wc()->cart->check_cart_item_stock();
-		if ( $cart_stock_check_result instanceof \WP_Error ) {
-			throw new RouteException(
-				'woocommerce_rest_checkout_items_out_of_stock_error',
-				implode( ', ', $cart_stock_check_result->get_error_messages() ),
-				400
-			);
+		// Ensure all items on draft order are still in stock.
+		foreach ( $order->get_items() as $item ) {
+			$product = wc_get_product( $item->get_data()['product_id'] );
+			if ( ! $product->is_in_stock() ) {
+				throw new RouteException(
+					'woocommerce_rest_checkout_item_out_of_stock_error',
+					sprintf(
+						// Translators: %s Out of stock item name.
+						__( 'Sorry, "%s" is not in stock. Please edit your basket and try again. We apologise for any inconvenience caused.', 'woo-gutenberg-products-block' ),
+						$product->get_name()
+					),
+					400
+				);
+			}
 		}
 	}
 
--- a/src/StoreApi/Utilities/OrderController.php
+++ b/src/StoreApi/Utilities/OrderController.php
@@ -152,6 +152,16 @@ class OrderController {
 				]
 			);
 		}
+
+		// Ensure all items in cart are still in stock.
+		$cart_stock_check_result = wc()->cart->check_cart_item_stock();
+		if ( $cart_stock_check_result instanceof \WP_Error ) {
+			throw new RouteException(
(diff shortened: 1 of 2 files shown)


src/StoreApi/Utilities/OrderController.php
```diff
@@ -152,6 +152,40 @@ public function validate_order_before_payment( \WC_Order $order ) {
...
+						sprintf(
+						// Translators: %s Not enough stock product.
+							__( 'Sorry, we do not have enough "%s" left in stock. Please edit your basket and try again. We apologise for any inconvenience caused.', 'woo-gutenberg-products-block' ),
+							$product->get_name()
+						),
+						400
+					);
+				}
+			}
+		}
```
haszari, line 188: Code looks reasonable, though I wonder if the `foreach product { check stock }` logic could be refactored so it's shared. Were we doing this elsewhere already, or are we calling through to logic in core? 

Either way, just checking if it's worthwhile sharing this code, and the error messages.

tests/php/StoreApi/Utilities/OrderController.php
```diff
@@ -0,0 +1,42 @@
...
+		$order = $class->create_order_from_cart();
+
+		$unmanaged_product->set_stock_status( 'outofstock' );
+		$unmanaged_product->save();
+
+		$managed_product->set_stock_quantity( 1 );
+		$managed_product->save();
+
+		$this->expectException( RouteException::class );
+		$class->validate_order_before_payment( $order );
```
haszari, line 39: Great to see a unit test for this included in the PR! 🙌

haszari (commented):
I did a quick code review, note I haven't tested this yet. Looks reasonable, though I recommend updating/rebasing this PR to latest `trunk` before testing (since order processing code has changed).

src/StoreApi/Utilities/OrderController.php
opr: This makes sense, and I agree. I've just seen [something you added here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/StoreApi/Routes/Checkout.php#L362) and now I'm wondering if making changes to core, in the [`reserve_stock_for_order`](https://github.com/woocommerce/woocommerce/blob/master/src/Checkout/Helpers/ReserveStock.php#L66-L125) method so that these checks are done there and more appropriate error messages are returned (as discussed in pca54o-DN-p2) would be better than doing the checks in our plugin?

haszari: Good question! That sounds like it's worth an investigation. There could be complexities caused by depending on this in Woo core (e.g. which version it's available vs what we support), but overall if we can work towards sharing this code and the error messages that's a good thing.

opr:
Closed in favour of #3656

PR #3472, opened 2020-11-27
Fix notice close button color in Twenty Twenty One dark mode

Aljullu:
The Notice close button was forced to be white, but that might not look right in themes that change the background of notices, like Twenty Twenty One in dark mode. This PR changes the CSS to use `currentColor`, so the close button always has the color from the text.

### Screenshots

_Before:_
imatge

_After:_
imatge

1. Install and activate Twenty Twenty One.
2. Go to Customize > Colors & Dark mode and check the Dark mode support checkbox.
3. In the frontend, make sure the dark mode is enabled and go to a page with the Checkout block.
4. Try submitting the form leaving one of the required inputs empty so the error notice appears.
5. Verify the close button is visible.

### Changelog

> Fix notice close button color in Twenty Twenty One dark mode.

### DIFF
--- a/assets/js/base/components/store-notices-container/style.scss
+++ b/assets/js/base/components/store-notices-container/style.scss
@@ -11,9 +11,9 @@
 			margin: 0 0 0 auto;
 			border: 0;
 			outline: 0;
-			color: #fff;
+			color: currentColor;
 			svg {
-				fill: #fff;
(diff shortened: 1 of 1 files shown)


nerrad (approved):
I didn't test but the change makes sense to me.

PR #3473, opened 2020-11-27
Create IE11 SCSS mixin

Aljullu:
While working on #3444 I thought it would be nice having an IE11 SCSS mixin so we don't need to repeat the hacky-media query all over the code base.

### Screenshots

_Filter products by price in IE11_

_Checkboxes in IE11_

1. With IE11, check that there are no visual regressions in the Filter products by price block and the checkboxes of the Checkout block (they should look like in the screenshots above).

### DIFF
--- a/assets/js/base/components/price-slider/style.scss
+++ b/assets/js/base/components/price-slider/style.scss
@@ -212,7 +212,7 @@
 	}
 }
 
-@mixin ie {
+@mixin ie-fixes() {
 	.wc-block-components-price-slider__range-input-wrapper {
 		background: transparent;
 		box-shadow: none;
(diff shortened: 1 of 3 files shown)


assets/js/base/components/price-slider/style.scss
```diff
@@ -263,27 +263,31 @@
...
 	}
 
-	&.is-loading,
-	&.is-disabled {
-		.wc-block-components-price-slider__range-input-wrapper {
-			@include placeholder();
-			box-shadow: none;
+	.wc-block-components-price-slider {
+		&.is-loading,
+		&.is-disabled {
```
Aljullu, line 268: While not related to this PR, I noticed these CSS selectors were missing the `.wc-block-components-price-slider` parent. It should be fixed now.

budzanowski (approved):

budzanowski:
The code looks, good I have approved it, although not tested. I don't have IE11 access. I should probably organize something.

PR #3478, opened 2020-11-27
Change register_endpoint_data to use an array of params instead of individual params. 

senadir:
This is a quick refactor that changes how `ExtendRestApi::register_endpoint_data` works.
Instead of passing parameters directly, we now pass an array of functions.

### Before:
```php
	array(
		CartSchema::IDENTIFIER,
		'subscriptions',
		 'my_schema_function',
		'my_data_function',
	)
);
```

### After:
```php
	array(
		'endpoint'        => CartSchema::IDENTIFIER,
		'namespace'       => 'subscriptions',
		'schema_callback' => 'my_schema_function',
		'data_callback'   => 'my_data_function',
	)
);
```

### Testing instructions 

- Add this code
```php
<?PHP
use Automattic\WooCommerce\Blocks\StoreApi\Schemas\CartSchema;

add_action(
	'init',
	function() {
			array(
				'endpoint'        => CartSchema::IDENTIFIER,
				'namespace'       => 'subscriptions',
				'schema_callback' => function () {
					return array(
						'key' => array(
							'type'        => 'string',
						),
					);
				},
				'data_callback'   => function () {
					return array( 'key' => 'hey' );
				},
			)
		);
	}
);
```

- On Cart block, open the network tab.
- Change the quantity of some item or shipping.
- inspect the response of the new API call.
- You should see a top-level `extensions -> subscriptions -> key -> hey`

### DIFF
--- a/src/Domain/Services/ExtendRestApi.php
+++ b/src/Domain/Services/ExtendRestApi.php
@@ -181,7 +181,7 @@ class ExtendRestApi {
 		return [
 			/* translators: %s: extension namespace */
 			'description' => sprintf( __( 'Extension data registered by %s', 'woo-gutenberg-products-block' ), $namespace ),
-			'type'        => 'object',
+			'type'        => [ 'object', 'null' ],
 			'context'     => [ 'view', 'edit' ],
 			'readonly'    => true,
 			'properties'  => $schema,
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Changes make sense to me and code looks and tests well. LGTM

PR #3488, opened 2020-12-01
CartItems must extend AbastractCartRoute otherwise the cart is not loaded from the session

ettoredn:
_edited by @senadir_ 

`store/cart/items` endpoint returns an error because cart is not loaded at that point.
This issue happens because `AbstractCartRoute` loads the cart before returning, `CartItems` didn't extend so the cart was never loaded.

### Testing instructions

1- Add items to your cart.
2- From your browser console, ping `store/cart/items` 
```js
await (await fetch("https://EXAMPLE/wp-json/wc/store/cart/items")).json()
```
3- You should see a result of the current cart items.

### Changelog

> Fix - Make sure cart is initialized before the CartItems route is used in the Store API.

### DIFF
--- a/src/StoreApi/Routes/CartItems.php
+++ b/src/StoreApi/Routes/CartItems.php
@@ -8,7 +8,7 @@ use Automattic\WooCommerce\Blocks\StoreApi\Utilities\CartController;
  *
  * @internal This API is used internally by Blocks--it is still in flux and may be subject to revisions.
  */
-class CartItems extends AbstractRoute {
+class CartItems extends AbstractCartRoute {
 	/**
 	 * Get the path of this REST route.
 	 *


opr:
Hi @ettoredn - thank you for this pull request! Would you be able to describe the issue you're facing along with the steps to reproduce it? This will be really helpful to give us more context about your proposed changes. Thanks in advance!

ettoredn:
Given any non empty cart:

image

the endpoint /wc/store/cart/items always returns error because, as I said, the cart is not loaded from the session (via AbstractCartRoute overrides):

image

senadir:
Thank you, I noticed the issue today and your PR fixes it.

I will update your PR body and merge it.

senadir (approved):

PR #3498, opened 2020-12-02
Introduce pluginArea in Cart and Checkout

### LINKED ISSUE #3442: Integration Tracking (WC Subscriptions): Create a registration API that uses registerPlugin.
As part of the integration process, 3PD need to:
- Write some code in a slotFill.
- Give us that code so we can render it inside our Block.

This issue tracks the second part, it should be done according to this post pca54o-y6-p2
We're going to register a `pluginArea` and accept a generic `registerPlugin`.

Because of the way `registerPlugin` work and the fact that it can be bypassed, I don't think we should provide any abstraction and directly use `registerPlugin`.
When rendering each value in `registerPlugin`, we can wrap it in a ErrorBoundary, we do the same with SlotFills and this should provide enough protection.

senadir:
This PR does two things:
- Introduces `PluginArea` in Cart and Checkout, the position inside the block doesn't matter as along as it's inside any context, it also wraps them in an BlockErrorBoundary.
- Modifies `BlockErrorBoundary` to accept a `render` function, we use it to render nothing so errors are not visible for guests/shoppers.

The ErrorBoundary is limited in that it protects the block from a broken integration but doesn't protect other plugins from a broken plugin, such protection should be done on an individual level, We already do that on SlotFills (see https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3361).
Such a thing needs to be done on an individual level via a render prop and `PluginArea` doesn't offer this flexibility but I will push a PR to Gutenberg to add this.

Also, because of the way `PluginArea` works, all registered plugins would get rendered in our Area, even ones that are not entered to be shown in Cart or Checkout, anyone misbehaving there would break the rest as well. I pushed a PR to protect against that https://github.com/WordPress/gutenberg/pull/27438 and that's also pending review.

I didn't see any value in adding an abstraction layer above `registerPlugin` because it provides everything we need, however if we ever end up using the latest version of `@wordpress/plugins` instead of the `wp.plugins` one, we lose access to scope and would need to publish our own `registerPlugin` abstraction.

### Testing instructions 
1- In any plugin that loads a JS file in Cart or Checkout, add this code.
If you have WC Subscriptions, add this code to `assets/js/frontend/wcs-cart.js`
```js
const el = wp.element.createElement;
const registerPlugin = wp.plugins.registerPlugin;
registerPlugin("my-plugin", {
  render: () => {
    console.log("I'm loading!");
    return el("p", {}, "This won't get rendered");
  },
});
```
2- In the file enqueue folder (in the above case woocommerce-subscriptions.php line 378), add `"wp-element", "wp-plugins"` as dependencies.
3- Visit Cart block, you should see "I'm loading!" in console log.
4- Edit the rendered element so that instead of `"p"`, you call an undefined variable.
5- Visit Cart block again, if you're admin, you should see an error, if you're a guest, you should see nothing.

Fixes #3442

### DIFF
--- a/assets/js/base/components/block-error-boundary/index.js
+++ b/assets/js/base/components/block-error-boundary/index.js
@@ -39,10 +39,14 @@ class BlockErrorBoundary extends Component {
 			showErrorMessage,
 			text,
 			errorMessagePrefix,
+			render,
 		} = this.props;
 		const { errorMessage, hasError } = this.state;
 
 		if ( hasError ) {
+			if ( render ) {
+				return render( { errorMessage } );
+			}
 			return (
 				<BlockError
 					errorMessage={ showErrorMessage ? errorMessage : null }
@@ -85,6 +89,10 @@ BlockErrorBoundary.propTypes = {
 	 * Text preceeding the error message.
 	 */
 	errorMessagePrefix: PropTypes.string,
+	/**
+	 * Render function to show a custom error component.
+	 */
+	render: PropTypes.func,
 };
 
 BlockErrorBoundary.defaultProps = {
--- a/assets/js/base/components/block-error-boundary/index.js
+++ b/assets/js/base/components/block-error-boundary/index.js
@@ -39,13 +39,13 @@ class BlockErrorBoundary extends Component {
 			showErrorMessage,
 			text,
 			errorMessagePrefix,
-			render,
+			renderError,
 		} = this.props;
 		const { errorMessage, hasError } = this.state;
 
 		if ( hasError ) {
-			if ( render ) {
-				return render( { errorMessage } );
+			if ( typeof renderError === 'function' ) {
+				return renderError( { errorMessage } );
 			}
 			return (
 				<BlockError
@@ -92,7 +92,7 @@ BlockErrorBoundary.propTypes = {
 	/**
 	 * Render function to show a custom error component.
 	 */
-	render: PropTypes.func,
+	renderError: PropTypes.func,
 };
 
 BlockErrorBoundary.defaultProps = {
--- a/assets/js/base/context/cart-checkout/checkout/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/index.js
@@ -1,4 +1,11 @@
 /**
+ * External dependencies
+ */
+import { PluginArea } from '@wordpress/plugins';
+import { CURRENT_USER_IS_ADMIN } from '@woocommerce/block-settings';
+import BlockErrorBoundary from '@woocommerce/base-components/block-error-boundary';
+
+/**
  * Internal dependencies
  */
 import { PaymentMethodDataProvider } from '../payment-methods';
@@ -31,6 +38,15 @@ export const CheckoutProvider = ( {
 				<ShippingDataProvider>
 					<PaymentMethodDataProvider>
 						{ children }
+						{ /* If the current user is an admin, we let BlockErrorBoundary render
+						the error, or we simply die silently. */ }
+						<BlockErrorBoundary
+							renderError={
+								CURRENT_USER_IS_ADMIN ? null : () => null
+							}
+						>
+							<PluginArea />
+						</BlockErrorBoundary>
 						<CheckoutProcessor />
 					</PaymentMethodDataProvider>
 				</ShippingDataProvider>
--- a/assets/js/blocks/cart-checkout/cart/full-cart/index.js
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/index.js
@@ -34,6 +34,7 @@ import Title from '@woocommerce/base-components/title';
 import { getSetting } from '@woocommerce/settings';
 import { useEffect } from '@wordpress/element';
 import { decodeEntities } from '@wordpress/html-entities';
+import { PluginArea } from '@wordpress/plugins';
 import { CartProvider } from '@woocommerce/base-context';
 
 /**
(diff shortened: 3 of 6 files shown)


senadir:
Tagging @nerrad beacuse this is a public API and you might have something to add.

assets/js/base/components/block-error-boundary/index.js
```diff
@@ -39,10 +39,14 @@ class BlockErrorBoundary extends Component {
 			showErrorMessage,
 			text,
 			errorMessagePrefix,
+			render,
```
nerrad: I wonder if the name of this prop should be a bit more specific here. It could be confusing given we have a `render` function that could potentially invoke _another_ function named `render` coming in from props (even though I know they are two different things). 

So maybe `renderError` or something like that?

assets/js/base/components/block-error-boundary/index.js
```diff
@@ -39,10 +39,14 @@ class BlockErrorBoundary extends Component {
...
 		} = this.props;
 		const { errorMessage, hasError } = this.state;
 
 		if ( hasError ) {
+			if ( render ) {
```
nerrad: Should validate this is a function before invoking.

nerrad (commented):
I left a comment but I have a few other questions/comments:

- Extensions would use `registerPluginArea` to add their integration and the registered `render` function in plugin area would then implement any fills right?
- Instead of putting `<PluginArea/>` in the block component. what about putting it in the `CartProvider` and `CheckoutProvider` along with the children? 
- Given this note you made:
> Also, because of the way PluginArea works, all registered plugins would get rendered in our Area, even ones that are not entered to be shown in Cart or Checkout, anyone misbehaving there would break the rest as well. I pushed a PR to protect against that WordPress/gutenberg#27438 and that's also pending review.

Does this mean we can't reliably use this in the editor context because if extensions register a plugin, it will show up in other plugin areas in the Editor? Or, it shouldn't matter in the editor context if we're using the already defined `<PluginArea/>` component in the editor context (which means we should probably not expose our implementation in the editor context?).

Have you verified that `PluginArea` and `registerPlugin` works consistently for our needs across the WordPress versions we support (and the `@wordpress/plugin` package version included in those WP releases)?

senadir:
> Extensions would use registerPluginArea to add their integration and the registered render function in plugin area would then implement any fills right?

The way plugins work is that there is a global `plugins` variable that holds things passed from `registerPlugin`.
PluginArea would loop through `plugins` and render everything in it.
PluginArea is `display:none` so it needs slotFills as well to make sense.

> Instead of putting <PluginArea/> in the block component. what about putting it in the CartProvider and CheckoutProvider along with the children?

Seems like a good idea.

> Does this mean we can't reliably use this in the editor context because if extensions register a plugin, it will show up in other plugin areas in the Editor? Or, it shouldn't matter in the editor context if we're using the already defined <PluginArea/> component in the editor context (which means we should probably not expose our implementation in the editor context?).

Let me unpack this question first.
Given the above explanation, in the editor context, we have several `PluginArea` (2 from GB, 1 from us).
So third party developers who are trying to integrate into Gutenberg using `registerPlugin` would also have their `render` function rendered in our `PluginArea`.
Same goes for 3PD integrating with our PluginArea, they would their `render` in GB's `PluginArea`. 
WordPress/gutenberg#27438 solves this by adding a scoping factor.
This isn't a major concern because:
- `PluginArea` is hidden by default and needs slotFills ↩️
- Anyone integrating into our plugin would be using out SlotFills, anyone integrating into GB would use GB's slotFills, so there is no chance of interfering.

The only concern is if the user had a plugin with a broken `registerPlugin` render function, that function would break our pluginArea, but would also break the editor.
If that broken function is loaded in frontend (which is unlikely but still possible with poor code), it would break our pluginArea.
A good mitigation against this is to do individual `ErrorBoundary` for each `registerPlugin` call, which is not possible right now.
We do implement individual `ErrorBoundary` for slotFills so we're protected there, but the code could be broken outside the slotFill.

Bottom line is: there's a small rare risk, but it would only take down the PluginArea and not the whole block. 

>  it shouldn't matter in the editor context if we're using the already defined <PluginArea/> component in the editor context (which means we should probably not expose our implementation in the editor context?).

Yes, we can not expose `PluginArea` in editor and things would still work, but we would be piggybacking on GB's PluginArea which could be removed or changed so it's better to do ours.

Other than that, there's no other benefit of having `PluginArea` in our code, since all of those benefits are gained by slotFills (like possible access to context, rerender on change, and event delegation). 

> Have you verified that PluginArea and registerPlugin works consistently for our needs across the WordPress versions we support (and the @wordpress/plugin package version included in those WP releases)?

I verified that wp.plugins is loaded on 5.3 (didn't go deeper) which is sufficient since we're going to support 5.4 and up anyway.
The API didn't change much since the package was introduced so no harm there (we're running 2.23, WP 5.3 runs [2.7](https://github.com/WordPress/wordpress-develop/blob/5.3.0/package.json#L106)).

assets/js/base/components/block-error-boundary/index.js
senadir: Did some validation, my first instinct was to use `isValidElement` but that would return false if I return a function so I'm just checking if this is a function.

opr (approved):
Tested and this is working great for me. I agree with Darren about the rename of `renderError` and the check to make sure it's a function. 👍🏼  Works as advertised.

PR #3500, opened 2020-12-02
Hide unsupported payment methods in checkout if the cart contains a subscription

### LINKED ISSUE #3252: Integration Tracking (WC Subscriptions): Ensure subscription orders use a supported payment method when using checkout block
Not all payment gateways [support WooCommerce Subscriptions](https://docs.woocommerce.com/document/subscriptions/enabling-payment-gateways-for-subscriptions/).

For example, on my store, in the shortcode checkout, onlyStripe and PayPal available when the order includes a subscription. 

Screen Shot 2020-10-08 at 1 39 03 PM

Screen Shot 2020-10-08 at 1 39 31 PM

In checkout block, all the payment methods are shown.

Screen Shot 2020-10-08 at 1 47 39 PM

We'll need to add support for this in the checkout block payments components.

haszari:
Fixes #3252

**Work In Progress** This is a proof-of-concept. Will likely need refinement and feedback before merging, or may be closed in favour of an alternative approach.

### Problem/goal
Only certain payment methods are supported for orders with subscriptions. This PR implements one way of hiding unsupported payment methods.

### Overview of the approach
- Payment gateways in Woo have a [`supports` array of strings](https://github.com/woocommerce/woocommerce/blob/master/includes/abstracts/abstract-wc-payment-gateway.php#L417) to indicate feature support.
- This is a core extensibility API. Extensions can declare and check for support for any feature (string) they wish; there is no predetermined list of capabilities.
- In this PR, the `supports` array is exposed to the client via the payment method data JSON blob, and then merged into the client-side config object for the payment gateway.
  - Note: this is only implemented for Stripe CC for now.
- The Store API `cart-items` endpoint now allows extensions to add data about cart items. WooCommerce Subscriptions use this [to add info about any subscriptions associated with a cart item - `items[].extensions.subscriptions`](https://github.com/woocommerce/woocommerce-subscriptions/pull/3839).
- When payment methods are refreshed in the checkout (or cart), this PR adds a new check. If there are any subscriptions in the cart, and the payment method doesn't support subscriptions, it's excluded (hidden).

#### Potential enhancements, improvements
1. The logic to detect subscriptions, and filter payment methods, is "hard-coded" in WooCommerce Blocks. There's no extensibility API (yet); Woo Blocks has specific knowledge about how `gateway.supports.subscriptions` and `cartItems.extensions.subscriptions relate.
2. There are possibly more use-cases supported in Woo Subscriptions that are not supported here. For example, [`multiple_subscriptions`](https://github.com/woocommerce/woocommerce-subscriptions/blob/trunk/includes/gateways/class-wc-subscriptions-payment-gateways.php#L91) or the merchant overriding settings to [`_accept_manual_renewals`](https://github.com/woocommerce/woocommerce-subscriptions/blob/trunk/includes/gateways/class-wc-subscriptions-payment-gateways.php#L83).

Next step for 1 & 2 is to explore extensibility APIs so this logic can be delegated to extensions (e.g. Woo Subscriptions in this case). Also confirm if this is a common enough need that it warrants extensibility; do we want to open this up, how much.

3. Gateway `supports` is [filterable - `woocommerce_payment_gateway_supports`](https://github.com/woocommerce/woocommerce/blob/master/includes/abstracts/abstract-wc-payment-gateway.php#L418). We're currently ignoring this. 

If `woocommerce_payment_gateway_supports` is regularly used/essential, we could loop through all `supports` values when generating the `supports` array for JS.

4. Implement this in more payment methods (PayPal, explore implications for WCPay). We should be able to export `supports` as JSON automatically for all payment gateways.

### Alternatives
1. Implement a more generic `order_type` (or `order_tags`?) field in Store API, as an extensibility point. This would allow Woo Subs to extend the API to capture info like "this cart contains `subscriptions`" (or `multiple_subscriptions` etc). Then generically consult this and `supports` automatically. This was [proposed by @nerrad in issue](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3252#issuecomment-738248611).
2. Implement a filter-style API in JS, similar to the previous approach. Extensions would be able to disable payment methods, based on cart contents. E.g. `shouldAllowPayment( paymentGateway, cartItems )` or `shouldAllowPayment( paymentGatewaySupports, cartItems )`.

I didn't prototype (1) above because it felt too limited for the Woo Subs use case. Subscriptions has more logic in `get_available_payment_gateways`, though some of this may not be relevant to checkout. Also we have already extended the API to add subs data, so it didn't feel right to add another extensibility point to allow extension to summarise that same data (to add `subscriptions` and/or `multiple_subscriptions`).

(2) feels too flexible for the initial iteration. We may iterate towards something like this; ideally we can limit the extensibility API so it's robust, simple to use, and doesn't allow abuse (e.g. hide other payment methods, conflicts between extensions).

1. Clone this branch and build.
2. Clone feature branch from WooCommerce Subscriptions [`feature/checkout-block-simple-multiple-subscriptions`](https://github.com/woocommerce/woocommerce-subscriptions/tree/feature/checkout-block-simple-multiple-subscriptions), build.
3. Ensure both extensions are active, checkout block is on checkout page, and store has a subscription product.
4. Add normal products to cart.
1. View cart with a range of payment methods enabled, ensure they all work correctly.
1. Add a subscription to cart.
1. View cart with a range of payment methods enabled, ensure they all work correctly - should only see Stripe CC if a subscription product is in the cart.

### Changelog

> TBC

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.js
@@ -52,7 +52,7 @@ const usePaymentMethodRegistration = (
 	const { selectedRates, shippingAddress } = useShippingDataContext();
 	const selectedShippingMethods = useShallowEqual( selectedRates );
 	const paymentMethodsOrder = useShallowEqual( paymentMethodsSortOrder );
-	const { cartTotals, cartNeedsShipping } = useStoreCart();
+	const { cartItems, cartTotals, cartNeedsShipping } = useStoreCart();
 	const canPayArgument = useRef( {
 		cartTotals,
 		cartNeedsShipping,
@@ -85,6 +85,10 @@ const usePaymentMethodRegistration = (
 			};
 		};
 
+		const subscriptionCartItems = cartItems.filter(
+			( cartItem ) => cartItem?.extensions?.subscriptions?.billing_period
+		);
+
 		for ( let i = 0; i < paymentMethodsOrder.length; i++ ) {
 			const paymentMethodName = paymentMethodsOrder[ i ];
 			const paymentMethod = registeredPaymentMethods[ paymentMethodName ];
@@ -97,12 +101,22 @@ const usePaymentMethodRegistration = (
 				const canPay = await Promise.resolve(
 					paymentMethod.canMakePayment( canPayArgument.current )
 				);
-				if ( canPay ) {
-					if ( canPay.error ) {
-						throw new Error( canPay.error.message );
-					}
-					addAvailablePaymentMethod( paymentMethod );
+				if ( ! canPay ) {
+					continue;
 				}
+
+				if ( canPay.error ) {
+					throw new Error( canPay.error.message );
+				}
+
+				if (
+					subscriptionCartItems.length > 0 &&
+					! paymentMethod?.supports?.subscriptions
+				) {
+					continue;
+				}
+
+				addAvailablePaymentMethod( paymentMethod );
 			} catch ( e ) {
 				if ( CURRENT_USER_IS_ADMIN || isEditor ) {
 					const errorText = sprintf(
@@ -135,6 +149,7 @@ const usePaymentMethodRegistration = (
 		noticeContext,
 		paymentMethodsOrder,
 		registeredPaymentMethods,
+		cartItems,
 	] );
 
 	// Determine which payment methods are available initially and whenever
--- a/assets/js/blocks-registry/payment-methods/payment-method-config.js
+++ b/assets/js/blocks-registry/payment-methods/payment-method-config.js
@@ -21,7 +21,10 @@ export default class PaymentMethodConfig {
 		this.canMakePayment = config.canMakePayment;
 		this.paymentMethodId = config.paymentMethodId || this.name;
 		this.supports = {
+			// Supply a default for `savePaymentInfo`.
 			savePaymentInfo: config?.supports?.savePaymentInfo || false,
+			// Pass through any `supports` feature string (open-ended API).
+			...config?.supports,
 		};
 	}
 
--- a/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
@@ -57,7 +57,11 @@ const stripeCcPaymentMethod = {
 	),
 	supports: {
 		savePaymentInfo: getStripeServerData().allowSavedCards,
-		...getStripeServerData().supports,
+		// Copy all server-provided `supports` features into JS config.
+		...getStripeServerData().supports.reduce( ( features, featureKey ) => {
+			features[ featureKey ] = true;
+			return features;
+		}, {} ),
 	},
 };
 
--- a/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
@@ -57,6 +57,7 @@ const stripeCcPaymentMethod = {
 	),
 	supports: {
 		savePaymentInfo: getStripeServerData().allowSavedCards,
+		...getStripeServerData().supports,
 	},
 };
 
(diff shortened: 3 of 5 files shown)


nerrad:
Thanks Rua for doing an exploration of things. As you've already noted, there's some problems with the approach in this PR. Notably:

- tightly coupled to Subscriptions implementation.
- `supports` is only enhanced on the payment method data for `WC_Gateway_Stripe` which is just one payment method available from the Stripe extension (you did add a comment block acknowledging that).

For the solution I'd like to see, I believe we can reach the following targets:

- Client side implementation should be _agnostic_ to what extension is integrating the "supports" data around what payment methods show or don't show. 
- The responsibility for integration by extensions should be server side only. I really don't think we need any client side integration hook points for this particular need. There still will need to be client side logic needing added by our code, but it can be a black box to the extensions.
- The API interface should be fairly straightforward and build on what we've already implemented for extensibility.

`WC_Payment_Gateway` children are able to indicate a `supports` property from classes that is populated with an array of strings that "tag" support for a certain feature. Currently they way this is extendable is by:

- overriding the property in the gateway class (payment gateway controls).
- overriding the `WC_Payment_Gateway::supports()` method (payment gateway controls) which receives a "feature" string and returns whether that string exists in the `supports` property (which is a public method).
- overriding the public `supports` property on any gateway instance (any plugin can do). Unfortunately doable since both the gateway instance is available and the property is _public_.
- hooking into `woocommerce_payment_gateway_supports` which is a filter around what is returned from `WC_Payment_Gateway::supports()`that can be used to inject custom return value for the feature string requested (any extension can use). The problem is that since the method containing this filter can be overridden. There's no guarantee the filter will execute for a given gateway.

As you can see from the above quick audit, the `supports` property is inherently fragile and extremely easy for plugins/custom code to break or manipulate in unexpected ways :(

From the perspective of how it is used I'd like to move completely away from us having to rely on this fragile implementation and instead implement a solution that allows extensions to use whatever they want to indicate payment method support given the following information:

- the available payment method options currently provided to the checkout flow.
- the current order/cart state.

So what I'd like to propose the establishment of a pattern that attaches an array of `supported_payment_methods` directly to the order details in a response. This information should in turn be extendable extensions to control whether a payment method is for use given the known data at the time of the order. Practically this means:

### We expose an array of `supported_payment_methods` on any response involving payment methods.

This would be a simple array of payment method identifier strings which would default to the Active payment methods.

Client side, `PaymentMethodRegistry` would use this array to determine what payment methods to are _available to show_ in the calculation for what appears in the UI. 

An advantage of having this as a part of the REST API responses is that anytime order details change, this array would be recalculated and in turn will affect what is shown in the UI.

### This array of `supported_payment_methods` is passed through `ExtendRestAPI` before being returned in the response.

We can have an interface here that allows extensions to register callbacks that will filter what is included/excluded in the array and we can do validation after each callback is executed to verify:

- We still have an array of strings.
- We don't have an array with a string that is unexpected (i.e. contains a payment method that is not registered in Woo).

Example interface that could be added to `ExtendRestApi` (subject to naming/improvement, just a poc):

```php
public function register_payment_support_callback( $callback ) {
	if ( ! is_callable( $callback ) ) {
		$this->throw_exception( '$data_callback must be a callable function.' );
	}
	$this->payment_supports_callbacks[] = $callback;
}

public function get_payment_supports( WC_Order $order, array $payment_supports ) {
	foreach ( $this->payment_supports_callbacks as $payment_supports_callback ) {
		$original_payment_supports = $payment_supports;
		// this should be wrapped in a try/catch in case extensions didn't implement the callback with the correct arguments or some other error happens.
		$payment_supports = $payment_supports_callback( $order, $payment_supports );
		// following conditions would need validated
		// - is array
		// - is array of strings
		// - no unexpected strings (i.e a string that is not a registered payment method identifier)
		// if any of the above fail, then reset `$payment_supports` to $original_payment_supports
		if ( $validation_fails ) {
			$payment_supports = $original_payment_supports;
		}
	}
	return array_unique( $payment_supports );
}
```

Then extensions would simply use this interface to register what payment methods are supported for the given order on the fly as the response is prepared using the given data. So they might use the `supports` property from payment methods or the supports() method, but this code doesn't really care. The API for `supports` could be improved down the road and it wouldn't impact this code. Essentially we provide the interface for modifying the array of supported payment methods but leave the actual determination of that up to the extension registering with the API.

What I didn't account for here (which I think we will need to account for in the final implementation) is passing in a `WC_Cart` instance to the `ExtendRestAPI::get_payment_supports` method (and passed through to registered callbacks) because we'll need this for the express payment method options in the cart context as well.

One advantage of this pattern, is we should be able to re-implement it for other endpoints we create involving payment methods. Of course, this would also satisfy the requirements I outlined above as well.

PR #3502, opened 2020-12-03
Fix:3501 Omit keys in `customerPaymentMethods` that have an empty list as a value.

### LINKED ISSUE #3501: CustomerPaymentMethods shows empty array for stored methods that have their gateway disabled.
## Describe the bug
If there are stored payment methods, and they were saved for a payment gateway that is now disabled the `customerPaymentMethods` from `PaymentMethodDataContext` can return the payments object with the sored method key and a value of an empty array.

This was found while working on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3439 . Disabling Stripe is causing unexpected code flow.

## To reproduce

1. Use Stripe
2. Save a CC with a Stripe payment
3. Disable Stripe payment
4. observe `customerPaymentMethods` value. It will have this component.

```js
{
  cc: [],
}
```

## Expected behavior
When there are no payment methods then the key should not be added to the methods object.

* Core version: [e.g. 5.5]
* WooCommerce version: [e.g. 4.8]
* WooCommerce Blocks version: [e.g. 4.0.0-dev]

budzanowski:
When there are stored payment methods for a gateway that got disabled the `PaymentMethodDataContext` was still adding a key entry to the value returned by `customerPaymentMethods`. 
This caused wrong behaviour for code constructs similar to this:
```js
const customerHasSavedCards = Object.keys( customerPaymentMethods ).length > 0;
```
The fix is to simply not add entries to the object when the value is an empty array.

Fixes #3501 

1. Follow the instructions from #3501, this time the object should be empty:
```js
{}
```

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -93,13 +93,13 @@ const getCustomerPaymentMethods = ( availablePaymentMethods = [] ) => {
 	}
 	const enabledCustomerPaymentMethods = {};
 	paymentMethodKeys.forEach( ( type ) => {
-		const methods = customerPaymentMethods[
-			type
-		].filter( ( paymentMethod ) => {
-			return Object.keys( availablePaymentMethods ).includes(
-				paymentMethod.method.gateway
-			);
-		} );
+		const methods = customerPaymentMethods[ type ].filter(
+			( paymentMethod ) => {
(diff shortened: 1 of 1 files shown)


senadir (approved):
Logic looks good to me. To break down this:

- We added `paymentMethodKey` to `enabledCustomerPaymentMethods` then verified if it had active methods or not, sometimes it doesn't and `enabledCustomerPaymentMethods` would have an empty value.
- Now, we do the verification before doing the array assignment making sure we only fill in working payment methods.

PR #3503, opened 2020-12-03
Add Formatter Classes to the Store API for Extensions to Consume

mikejolley:
This is to help implement subscriptions support (https://github.com/woocommerce/woocommerce-subscriptions/pull/3904).

When extending the REST API, extensions may need access to certain formatting functions to avoid re-implementing them in the extension itself. This applies mainly to monetary and currency values where our Store API uses specific formats.

This PR attempts to solve this by exposing formatter classes via the `ExtendRestAPI` class. I've included currency, money, and HTML formatters in this PR.

Example usage:

```php
$this->extend->formatters->money->format( $value )
```

Formatters are registered via Bootstrap and injected into the `ExtendRestAPI` class.

Each individual formatter has public methods to `format`, `set_option`, and `get_option`. This ensures all formatters are compatible with each other, and in the absence of a specific formatter there are no fatal errors from calling undefined methods.

If you do a GET request on the cart endpoint, you should see correct results. This endpoint has implemented the new formatters via `prepare_money_response`.

I have included tests to check formatter functionality, and registration.

### Changelog

> Added formatting classes to the Store API for extensions to consume.

cc'ing @nerrad to ensure he is happy with this pattern of usage.


src/StoreApi/Formatters.php
```diff
@@ -0,0 +1,45 @@
...
+	 *
+	 * @param string $name Name of the formatter.
+	 * @return AbstractFormatter Formatter class instance.
+	 */
+	public function __get( $name ) {
+		if ( ! isset( $this->formatters[ $name ] ) ) {
+			return new DefaultFormatter();
+		}
+		return new $this->formatters[ $name ]();
+	}
```
senadir: Does this mean we're going to just die silently when someone reaches an unknown formatter? 
Can't we do this for everything but throw an error for admin users?

mikejolley: This doesn't die silently. This will return a default formatter which passes through values. This avoids extensions shooting themselves in the foot by using unregistered formatters, or breakages if a formatter is removed or renamed.

senadir: This is my definition of dying silenctly (which is wrong).
I think the current behaviour is great and that's what I'm doing everywhen when it makes sense, I was suggesting on top of that we also throw an error for admin users only.
Same we do in extendRestApi
```php
private function throw_exception( $exception_or_error ) {
		if ( is_string( $exception_or_error ) ) {
			$exception = new Exception( $exception_or_error );
		} else {
			$exception = $exception_or_error;
		}
		// Always log an error.
		wc_caught_exception( $exception );
		if ( defined( 'WP_DEBUG' ) && WP_DEBUG && current_user_can( 'manage_woocommerce' ) ) {
			throw $exception;
		}
	}
```

src/StoreApi/Formatters.php
```diff
@@ -0,0 +1,45 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\StoreApi;
+
+use Exception;
+use Schemas\AbstractSchema;
+use Automattic\WooCommerce\Blocks\Domain\Services\ExtendRestApi;
```
senadir: Those are not needed here?

senadir (approved):
I did a review and this looks good, I have a couple of comments about throwing errors for admin and extra code.

mikejolley:
Thanks @senadir Just waiting for blessing from @nerrad 👍🏻

src/Domain/Services/ExtendRestApi.php
```diff
@@ -18,13 +19,34 @@ class ExtendRestApi {
...
+	 * @param string $name Property name.
 	 */
-	public function __construct( Package $package ) {
-		$this->package = $package;
+	public function __get( $name ) {
+		if ( in_array( $name, [ 'formatters' ], true ) ) {
+			return $this->$name;
+		}
+		return null;
 	}
```
nerrad: Using the getter method is clever, but I would prefer that there be an explicit getter for formatters. One problem with magic method usage like this is that IDEs can't resolve the methods so it makes the method more opaque. We can also explicitly add typing for what is returned which helps for things like automatic documentation (which Woo Core is using) and also IDE intellisense (i.e. `@returns FormatterInterface`)

src/StoreApi/Formatters.php
```diff
@@ -0,0 +1,48 @@
...
+	 * @return AbstractFormatter Formatter class instance.
+	 */
+	public function __get( $name ) {
+		if ( ! isset( $this->formatters[ $name ] ) ) {
+			if ( defined( 'WP_DEBUG' ) && WP_DEBUG && current_user_can( 'manage_woocommerce' ) ) {
+				throw new Exception( $name . ' formatter does not exist' );
+			}
+			return new DefaultFormatter();
+		}
+		return new $this->formatters[ $name ]();
```
nerrad: What's the intention of lazy instantiation here? Is there any benefit over just registering the formatter instance (and thus allowing to fail quicker if the instance registered is not of `FormatterInterface`)?

src/StoreApi/Formatters/FormatterInterface.php
```diff
@@ -0,0 +1,34 @@
...
+	 */
+	public function get_option( $name, $default = '' );
+
+	/**
+	 * Sets a formatter option.
+	 *
+	 * @param string $name Option name.
+	 * @param string $value Option value.
+	 */
+	public function set_option( $name, $value );
```
nerrad: I see this interface is intended to allow a way for changing how a formatter works. One potential problem I see with the current implementation is that any code implementing a formatter could not rely on what is set by default and would explicitly need to always set any configuration options so I wonder if this interface is needed. Should we instead just have `format` receive a second argument for any formatter configuration (for on the fly config) It could either be an instance of some sort of configuration interface or a simple array (and default to empty)?

src/StoreApi/Schemas/AbstractSchema.php
```diff
@@ -232,45 +232,13 @@ protected function get_store_currency_properties() {
...
-			'currency_code'               => get_woocommerce_currency(),
-			'currency_symbol'             => $symbol,
-			'currency_minor_unit'         => wc_get_price_decimals(),
-			'currency_decimal_separator'  => wc_get_price_decimal_separator(),
-			'currency_thousand_separator' => wc_get_price_thousand_separator(),
-			'currency_prefix'             => $prefix,
-			'currency_suffix'             => $suffix,
-		];
+	protected function prepare_currency_response( $values ) {
+		return $this->extend->formatters->currency->format( $values );
```
nerrad: For now this is sufficient. But it'd be nice to have AbstractSchema explicitly have `Formatters` as an injected dependency. This is a very loosely held opinion.

nerrad (commented):
I really like what you've done here Mike with the Formatters classes. Besides the inline comments I made I think we should also have `Formatters` live in the `src/Domain/Services` folder/namespace as it seems better suited as a general service class (usable not just by the `StoreApi`) and a good potential candidate for moving to WooCommerce core at some point.

src/StoreApi/Formatters.php
mikejolley: I didn't want code that set options to interfere with other usage.

mikejolley:
@nerrad I think I disagree with you on moving this to Services. The formatters are specifically designed to format data for API responses, for example, HtmlFormatter encodes special characters, and CurrencyFormatter matches defined Schemas. I don't think they quality as general purpose.

src/StoreApi/Formatters/FormatterInterface.php
mikejolley: Well if we're not lazy instantialising we can do that instead. So long as the methods have the same signature across all formatters it will be ok.

src/StoreApi/Schemas/AbstractSchema.php
senadir: Wouldn't we also need to expose it to plugins?

src/StoreApi/Formatters/FormatterInterface.php
senadir: Mike's comment above should clear this because no one is sharing the same instance.

nerrad:
> I think I disagree with you on moving this to Services.

Yup I see that. Formatting seems like a general useful utility service. We could feasibly allow registering additional formatters that are not specific to RestAPI response formatting?

I understand this gets into the weeds of potential premature architecture for future use that may never materialize so I'm happy to keep in StoreAPI as an `@internal` thing for now.

src/StoreApi/Formatters.php
nerrad: Ohhh, I see, and this applies to my other comment which would be irrelevant because options are set per instance. I think I'd prefer the options to be configured on the fly with an argument on `format` as it seems like it'd be a more straightforward API (and there'd only ever be one instance of the formatter in a request).

src/StoreApi/Formatters/FormatterInterface.php
nerrad: It does 👍 (as noted in my response to the comment).

src/StoreApi/Schemas/AbstractSchema.php
nerrad: I'm not suggesting _instead of_ being a dependency in ExtendRestAPI. I'm just saying it'd also be a dependency on AbstractSchema. But again, loosely held opinion and is fine as is.

src/Domain/Services/ExtendRestApi.php
mikejolley: Replaced with `get_formatter( $name )`

mikejolley:
Taken care of feedback and pushed the same updates to https://github.com/woocommerce/woocommerce-subscriptions/pull/3904

When this goes green, :shipit:

nerrad (approved):
Thanks for responding to and applying feedback :shipit:

PR #3520, opened 2020-12-08
Rename Cart & Checkout totals components

### LINKED ISSUE #2712: Simplify naming of directories of the Cart & Checkout totals components
We could simplify the naming of the directories under `/assets/js/base/components/cart-checkout/totals`:

* `subtotals-item`→ `subtotal`
* `totals-coupon-code-input` → `coupon`
* `totals-discount-item` → `discount`
* `totals-fees-item` → `fees`
* `totals-footer-item` → `footer-item`
* `totals-item` → `item`
* `totals-shipping-item` → `shipping`
* `totals-taxes-item` → `taxes`

Reasons for that:
* There is no need to repeat `totals`, it's already the name of the parent class.
* Lets reserve the suffix `-item` for `totals-item` and `totals-footer-item`, which apply a specific style depending on their position.
* This will match the class names introduced in #2711.

Aljullu:
Fixes #2712.

This PR renames some components/folders according to the guidelines in #2712 and with the goal to make component names & folder structure to be aligned with CSS class names.

### Screenshot
imatge

If there is any issue, it will probably be apparent when running `npm run build` or `npm run start`, but it's also a good idea to go to the Cart or Checkout block and verify the sidebar looks correct.

In order to get all rows to appear you need to:
* Have a shipping method and in the frontend add a product that needs shipping to the Cart.
* Have taxes enabled and show prices excluding tax (WooCommerce > Settings > Taxes ).
* Have a coupon that adds a discount and introduce it in the frontend.
* Add this code snippet to a PHP file (for example: `woocommerce-gutenberg-products-block.php`) to get the `Fees` entry to appear:
```PHP
add_action( 'woocommerce_cart_calculate_fees', 'add_fee', 10 );
function add_fee( $cart ) {
	$cart->add_fee( __( 'Fee', 'woo-gutenberg-products-block' ), 100, true );
}
```


opr (approved):
I agree with the rationale, described in the original issue and it certainly makes the codebase a little tidier.

Here's what I'm seeing and it all looks great:

Screenshot 2020-12-08 at 20 32 32

Let's 🚢 it.

PR #3524, opened 2020-12-08
Update Release Template and Docs

mikejolley:
After doing a release, I took notes and made some changes:

- I have removed the release docs from the plugin since these are now in the team handbook on p2
- In the release checklists, I clarified the process of updating the changelog to make it clear that versions and stable tags should be left alone
- I replaced the sample date with YYYY-MM-DD so this date is not missed accidentally
- I broke the testing notes section into sub tasks

Not much here to review :)


nerrad (approved):
Thanks Mike for the improvements :). The changes look good to me. I'm still a bit on the fence whether we should move the release notes to our handbook and out of the repository. But I'm willing to try for a few releases and see how things go (we can always move back if we find it doesn't work out well).

Besides updating the related links in our team handbook, please also surface this in our async meeting notes so the team is aware of the changes.

PR #3535, opened 2020-12-09
Update WP versions used in Travis e2e tests

### LINKED ISSUE #3534: Update e2e tests in Travis to cover WP versions from 5.4 to 5.6
We are planning to update the minimum required version of WP to 5.4 in WC Blocks 4.1 (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3516), so we need to update the e2e tests to exclude WP 5.3 and include WP 5.6.

Aljullu:
Fixes #3534.

This PR removes WP 5.3 from Travis e2e testing and instead adds WP 5.6. I also changed the `with Gutenberg plugin` version to target 5.6 instead of 5.5.

Verify Travis tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,13 +93,6 @@ jobs:
               - npm run test
           env:
               - WOOCOMMERCE_BLOCKS_PHASE=3
-        - name: E2E Tests (WP 5.3)
-          script:
-              - npm run test:e2e
-          env:
-              - WP_VERSION=5.3-branch
-              - E2E_TESTS=1
-              - WOOCOMMERCE_BLOCKS_PHASE=3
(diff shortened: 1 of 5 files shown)


mikejolley (approved):
LGTM!

Do we need to run tests on intermediate versions, or could we get away with min and max supported for most PRs etc?

Aljullu:
> Do we need to run tests on intermediate versions, or could we get away with min and max supported for most PRs etc?

Good point. AFAIK we never had an intermediate version failing without the min/max versions failing as well, so I don't think we lose anything removing the intermediate versions.

I will merge this PR as-is and we can handle removing intermediate versions in a follow-up if needed.

PR #3536, opened 2020-12-09
Switch to Fragment Short Syntax (`<>`)

### LINKED ISSUE #1761: Switch to Fragment short syntax (from <Fragment> to <>)
There has been some discussion in a couple of PRs in parallel whether we should use `<Fragment>` or `<>`:
* https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1561#discussion_r369029350
* https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1736#discussion_r380131662

This issue is to centralize the discussion and, in case we agree to switch from `<Fragment>` to `<>`, enable the appropriate [ESLint rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-fragments.md).

Some things to notice:
* We have been consistently using `<Fragment>` until now.
* Gutenberg is inconsistent about this and uses both, `<Fragment>` and `<>`.
* Using `<Fragment>` means we need to import it. Considering we currently have three different packages where it can be imported from (`@wordpress/element`, `wordpress-element` and `react`), that might be a source of unexpected bugs that could be prevented if we used the short syntax.
* [Fragments in React docs](https://reactjs.org/docs/fragments.html)

mikejolley:
This PR replaces usage of `<Fragment>` with the shorthand `<>`—this also allows us to remove some imports. 

[Fragment documentation can be found here](https://reactjs.org/docs/fragments.html). Our only remaining use of Fragment is as a fallback for the `Label` component, and where a `key` prop is needed.

I have enabled the eslint rule `react/jsx-fragments` which will catch usage of Fragment, **however**, this only catches `React.Fragment`, not the fragments imported from the `@wordpress/elements` package. I'm not sure if this is fixable or not.

I think this change can be left out of the changelog. `('<>')/`

Closes #1761

Given we've been rolling out shorthand `<>` in recent updates, I don't envision this having negative impact anywhere, and since the updated components are used in Blocks covered by e2e I think we're safe. That said, some smoke testing of updated components could be helpful. I tested a few of them (Product Tag Block since that change was a little larger) in particular. In all cases you should only need insert the block and check the console for errors.

The following Blocks are affected:

- Active Filters
- Price Slider
- Filter By Attribute
- Featured Cateogry
- Featured Product
- Handpicked products
- Best Sellers
- New Products
- Product categories
- Products by Category
- Onsale products
- Products by Tag

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,5 +20,6 @@ module.exports = {
 	rules: {
 		'woocommerce/feature-flag': 'off',
 		'react-hooks/exhaustive-deps': 'error',
+		'react/jsx-fragments': [ 'error', 'syntax' ],
 	},
 };
--- a/assets/js/base/components/block-error-boundary/index.js
+++ b/assets/js/base/components/block-error-boundary/index.js
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import PropTypes from 'prop-types';
-import { Component, Fragment } from 'react';
+import { Component } from 'react';
 
 /**
  * Internal dependencies
@@ -20,10 +20,10 @@ class BlockErrorBoundary extends Component {
 		) {
 			return {
 				errorMessage: (
-					<Fragment>
+					<>
 						<strong>{ error.status }</strong>:&nbsp;
 						{ error.statusText }
-					</Fragment>
+					</>
 				),
 				hasError: true,
 			};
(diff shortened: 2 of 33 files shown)


Aljullu (approved):
I tested it in the editor and frontend in a page with all WC Blocks and everything looked good. LGTM!

I see an ESLint warning in [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3536/files#diff-615d49ba906702d5afd8c7c4e079b5984633a74a1cd6eee0c1a74ef56c6a7d5bR245). Not introduced in this PR, but do you think we should create an issue to fix it?

PR #3537, opened 2020-12-09
Update minimum required WP version to 5.4

### LINKED ISSUE #3516: Update minimum required WordPress version to 5.4
In line with our L2 support policy, with the release of WooCommerce Blocks 4.1.0 we can update the minimum required WordPress version to 5.4.

Aljullu:
Fixes #3516.

This PR updates the minimum WP required version to 5.4 in all relevant files and removes a couple of checks which are no longer needed.

Changes are quite safe, so the only think that it might make sense testing is that blocks that were only available on WP>=5.3 are still available:

1. Open the inserter when creating/editing a post or page and verify All Products, Price Filter, Attribute Filter, Active Filters, Checkout and Cart blocks are available.

### Changelog

> Bumped the minimum WP required version to 5.4.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,13 +40,13 @@ We release a new version of WooCommerce Blocks onto WordPress.org every few week
 >
 > We **recommend** you always keep WordPress and WooCommerce core up to date in order to ensure your store is running with the most recent fixes and enhancements to help your store be successful.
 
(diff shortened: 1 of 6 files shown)


mikejolley (approved):
LGTM 👍🏻

PR #3540, opened 2020-12-10
Add Close stale issues workflow

mikejolley:
This PR implements the [close stale issues action](https://github.com/actions/stale). It's configured to run daily at 9am UTC and will label issues and PRs with no activity over 60 days as stale. These are then closed in a further 10 days unless unlabeled, or edited. When marked stale, an explanatory comment will also be posted.

Note, issues labelled `Priority: Critical` and `Priority: High` are excluded, but open to changing this if needed.

### DIFF
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -1,0 +1,22 @@
+name: "Close stale issues"
+
+on:
+  schedule:
+  # Runs daily at 9am UTC
+  - cron: "0 9 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        days-before-stale: 60
+        days-before-close: 10
+        remove-stale-when-updated: true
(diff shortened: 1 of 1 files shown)


nerrad (approved):
Looks good to me and configuration makes sense! Thanks Mike, let's :shipit: 

Do we need to create the label that the workflow will use?

mikejolley:
> Do we need to create the label that the workflow will use?

I don't know :) I will just in case. A nice stale brown/green color.

haszari:
I'm getting failure emails (with success in body!) from this stale workflow whenever I push to an open PR. I think this is a bug, shouldn't run on PRs?

Screen Shot 2020-12-17 at 8 21 06 AM

nerrad:
Looks like there's a yaml syntax error which is causing this. I'm looking into this right now.

nerrad:
ya there was a syntax error I introduced with changes I made recently to improve messaging. I think I've fixed it.

mikejolley:
@nerrad I noticed the heading is off too because there is no space between `internal` and the `####`

nerrad:
> I noticed the heading is off too because there is no space between internal and the ####

Yup, see it - and fixed (hopefully - we'll find out next run).

PR #3541, opened 2020-12-10
Update usage of legacy packages (inc. ServerSideRender)

### LINKED ISSUE #3513: Remove workarounds for `@wordpress/components` update no longer needed with required WP 5.4 version update.
With WooCommerce Blocks 4.1.0 release we are able to bump our minimum required version of WP to 5.4 due to our L2 support policy. As a result, some workarounds we have implemented in our Webpack builds as a part of #3457 are no longer necessary. Specifically related to the workarounds for:

- `@wordpress/primitives`
- `@wordpress/warning`

And _possibly_ `@wordpress/compose` (should be verified).

I've put as high priority because I think we should remove this workaround code as soon as feasible to prevent possible gotcha's down the road if and when we start using these packages.

### LINKED ISSUE #3515: Import ServerSideRender from the `@wordpress/server-side-render` package.
In WordPress 5.3, the `ServerSideRender` component was moved to the `@wordpress/server-side-render` package (and external). We should make sure all our imports of this component come from the correct package.

mikejolley:
This PR updates our imports of `ServerSideRender ` to the correct package, `@wordpress/server-side-render`. This PR also removes our workarounds for:

- `@wordpress/primitives`
- `@wordpress/warning`

`@wordpress/compose` is needed until the min version is 5.5. I added a note to the docblock about this.

Fixes #3515
Fixes #3513

This affects all SSR blocks but I don't think this is high risk. If build complets, check that SSR blocks still function in the editor and frontend. And example of an SSR block is `Products by Tag`.

In the page source, you should see the external JS for server-side-render being included.

I believe e2e tests would flag anything when ran on WP 5.4.

### Changelog

> Removed compatibility with packages in WordPress 5.3.

### DIFF
--- a/bin/webpack-helpers.js
+++ b/bin/webpack-helpers.js
@@ -6,12 +6,11 @@ const path = require( 'path' );
 const NODE_ENV = process.env.NODE_ENV || 'development';
 const FORCE_MAP = process.env.FORCE_MAP || false;
 
-// Some packages are not available in legacy versions of WordPress, so we don't
-// want to extract them.
+// If a package is not available, or missing functionality, in an old but __supported__ version of WordPress, it should be listed here.
+// Some packages are not available in legacy versions of WordPress, so we don't want to extract them.
 const requiredPackagesInWPLegacy = [
-	'@wordpress/compose', // WP 5.3 version doesn't include `useResizeObserver`.
-	'@wordpress/primitives', // Not included in WP 5.3.
-	'@wordpress/warning', // Not included in WP 5.3.
+	// The package included in WP 5.4 version doesn't include `useResizeObserver`. This can be removed when we support 5.5+.
+	'@wordpress/compose',
 ];
 
 const wcDepMap = {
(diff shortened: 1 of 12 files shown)


bin/webpack-helpers.js
```diff
@@ -6,12 +6,11 @@ const path = require( 'path' );
 const NODE_ENV = process.env.NODE_ENV || 'development';
 const FORCE_MAP = process.env.FORCE_MAP || false;
 
-// Some packages are not available in legacy versions of WordPress, so we don't
-// want to extract them.
+// If a package is not available, or missing functionality, in an old but __supported__ version of WordPress, it should be listed here.
+// Some packages are not available in legacy versions of WordPress, so we don't want to extract them.
```
haszari, line 10: Nice clarification of this comment 📘

haszari (approved):
Code all makes sense. Nice to tidy up some of this legacy / bw compat machinery.

Did some quick smoke tests on 5.6 (local dev) and 5.4 (jurassic ninja site, production zip build). Edited the storefront homepage, which includes various SSR blocks, then tested front end. All 👌 - :shipit:

PR #3544, opened 2020-12-10
Create GitHub actions for automated tests

### LINKED ISSUE #3542: Move tests from Travis to GitHub actions
Travis recently announced that it will no longer be free for open source projects (https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing).

At the moment Travis runs our tests and builds and deploys Storybook. Since we are already running a few actions on GitHub it makes sense to migrate our tests (and Storybook build) to GitHub actions.

We may be able to use the workflow configs they're using on Gutenberg to inform the way we set ours up.

- [Gutenberg E2E Tests action](https://github.com/WordPress/gutenberg/actions/runs/413103677/workflow)
- [Gutenberg Unit Tests action](https://github.com/WordPress/gutenberg/actions/runs/413103667/workflow)
- [Gutenberg Storybook deploy](https://github.com/WordPress/gutenberg/actions/runs/412882115/workflow)

CC @haszari since @nerrad mentioned you were interested in working on this too.

opr:
Fixes #3542 and incorporates @haszari 's PHP test workflow found in #3584

1. Push a commit to this branch that breaks one or more of the PHP tests.
2. Verify the "PHP, JavaScript, and E2E tests" workflow fails.
3. Revert the commit and push a new commit to the branch that breaks one of the JS unit tests.
4. Verify the "PHP, JavaScript, and E2E tests" workflow fails.
5. Revert the commit and push a new commit to the branch that breaks one of the JS E2E tests.
6. Verify the "PHP, JavaScript, and E2E tests" workflow fails.
7. Revert the commit and leave the branch as you found it.


.github/workflows/end-to-end-tests.yml
```diff
@@ -0,0 +1,100 @@
...
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Cache node modules
```
haszari: Nice! I'm assuming this will cache `node_modules` to speed up individual jobs. Should we cache `composer` deps too?

.github/workflows/end-to-end-tests.yml
```diff
@@ -0,0 +1,100 @@
+name: End-to-End Tests
+
+on:
+  push:
+    branches: [trunk]
+  pull_request:
+    branches: [trunk]
```
haszari: Does this run on all pushes to all PRs/branches? I'm thinking that's what we want, so this is another check against PRs as you work.

haszari: Wahey! I see this is running against this PR, I think `pull_request` is doing what we need 🌮 

Screen Shot 2020-12-22 at 12 58 32 PM

opr: Yep! This will run every time something is pushed to a branch with an open PR to `trunk` and it will also run when any code is pushed to `trunk` (so when the PR is merged I guess these checks will run again)

.github/workflows/end-to-end-tests.yml
opr: It doesn't cache `node_modules` but rather the `npm` cache. This, I believe keeps all the packages locally, so when `npm ci` runs, rather than fetching from npm it just pulls them from the cache. The `node_modules` directory seems to persist between jobs too, so no need to `npm ci` on each job, we can just do it once at the start.

This cache persists between runs as far as I'm aware.

Good idea about composer too, I added that to the cache.

.github/workflows/php-js-e2e-tests.yml
```diff
@@ -0,0 +1,123 @@
+name: PHP, JavaScript, and E2E tests
```
haszari: Suggest simplifying this to `Automated tests` - more compact/generic (and rename the file too?). Not a blocker, just a personal preference.

.github/workflows/php-js-e2e-tests.yml
```diff
@@ -0,0 +1,123 @@
...
+      - name: Npm install and build
+        run: |
+          npm ci
+          FORCE_REDUCED_MOTION=true npm run build:e2e-test
+          
+      - name: Composer install
+        run: |
+          composer install
+          
+      - name: Run PHP tests
```
haszari: Should we split out the different types of tests to different jobs? That seems appropriate to me; or if we want to test with different conditions (e.g. WP/PHP versions) maybe make those different jobs.

For example:

```
Automated tests
   - job: JS unit tests
   - job: PHP unit/WP tests
   - job: e2e tests
```

or 

```
Automated tests
   - job: WP bleeding edge + PHP 8.0
   - job: WP 5.6 + PHP 7.4
   - job: WP 5.4 + PHP min
```

haszari: Note that the existing Travis phpunit is run under different conditions, so we should probably replicate those (unless they're out of date/stale). Also `WOOCOMMERCE_BLOCKS_PHASE` may be important - this is our feature flag which enables all the latest blocks (currently cart & checkout).

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/4111e1e533031774c17f4af5817c895a36196304/.travis.yml#L71-L88

.github/workflows/php-js-e2e-tests.yml
```diff
@@ -0,0 +1,123 @@
+name: PHP, JavaScript, and E2E tests
+
+on:
+  push:
+    branches: [trunk]
+  pull_request:
+    branches: [trunk]
```
haszari: Probably don't need the branch specifier for `pull_request` - we rarely open PRs against non-`trunk`, but if we do, it's worthwhile running tests :)

.github/workflows/php-js-e2e-tests.yml
```diff
@@ -0,0 +1,123 @@
...
+
+jobs:
+  E2E_Tests:
+    name: End to end tests
+
+    runs-on: ubuntu-latest
```
haszari, line 11: I don't have a particular preference for OS, unless there's an option that will speed these up/use less power 🔋

.github/workflows/php-js-e2e-tests.yml
```diff
@@ -0,0 +1,123 @@
...
+
+    strategy:
+      fail-fast: false
```
haszari: This seems appropriate, since this is an "omnibus" of various tests.

haszari: Hmm .. in my testing, I only saw the first test fail, so maybe better to fail fast.

haszari (commented):
This is looking great! 

In my testing I didn't see e2e test fail, but I think this is because the unit tests are actually broken (?).

Overall I think this is close and we should get this merged soon. There are a bunch of considerations or improvements we could look at, though most probably as follow up.

### Blocker - fully migrate away from Travis.
- Get the tests all passing (not sure what's up here, php tests are failing ?).
- Let's ensure we've got equivalent coverage for the existing Travis jobs (e.g. different wordpress/php versions). Or make an explicit decision to drop some combinations of jobs.
- Delete the travis config (i.e. complete the migration), or log specific follow up to complete migration away from travis.

### Consideration - how to split into different jobs
This is where it gets fuzzy - some of these may interact with each other. This is more a list of ideas, no need to follow up here, but might be relevant for the above stuff.

- Fail fast vs. press on - I don't think this is doing anything useful, but it might if we have separate jobs.
- Jobs vs steps - I think it's worth splitting this into distinct jobs. Hopefully this would give us clear fail signals when multiple tests fail.
- How to break down the jobs - by type (e2e | php | js) or by environment (PHP / WP version).

Also there's some wrinkles to be aware of around environments / versions - we have the host (ubuntu) and also wp-env (e2e tests) and a custom docker (phpunit) to consider. I'm not clear on the dependencies of each type of test, e.g. WP/PHP, and how our env vars affect each environment. So we might need to consider this when we're deciding how to break up the tests into jobs, or isolate specific test runs.

.github/workflows/php-js-e2e-tests.yml
opr: I've actually gone ahead and made each type of test, and each E2E test its own job, that way we can run them in parallel for a faster build, and we get an easier way to understand which specific environment failed without digging through logs.

This is beneficial too as one failure won't stop other environments from running.

.github/workflows/php-js-e2e-tests.yml
opr: Yep, this only applied to matrices, and we're not using these. The new way of splitting everything into its own job will show us individual failure points.

opr:
> Blocker - fully migrate away from Travis.
Delete the travis config (i.e. complete the migration), or log specific follow up to complete migration away from travis.

Yes I agree, Nadir did some digging and found the setting where we can disable this. We can do it once this PR is merged p1608740454254900-slack-C8X6Q7XQU 

> Get the tests all passing (not sure what's up here, php tests are failing ?).

This was fixed in #3594

> Let's ensure we've got equivalent coverage for the existing Travis jobs (e.g. different wordpress/php versions). Or make an explicit decision to drop some combinations of jobs.

I did some investigation and found that the Travis job for PHP 5.6/Unit-tests is not actually using PHP 5.6, it looks like it uses 7.4. Based on https://docs.woocommerce.com/document/update-php-wordpress/ I don't think we should be testing against PHP 5.6. 

Currently we test the following in travis (what the configuration file claims it is doing notwithstanding):

- PHP Unit tests using PHP 7.4
- JavaScript Unit Tests
- E2E Tests using PHP 7.4 & WordPress 5.4
- E2E Tests using PHP 7.4 & WordPress 5.5
- E2E Tests using PHP 7.4 & WordPress 5.6
- E2E Tests using PHP 7.4 & WordPress 5.6 & The Gutenberg plugin

I have replicated all of these here.

> How to break down the jobs - by type (e2e | php | js) or by environment (PHP / WP version).
Fail fast vs. press on - I don't think this is doing anything useful, but it might if we have separate jobs.

Why not both? Now everything runs as an independent job. We can see more specifically what's failing too which is a benefit.

Unfortunately this does mean we do have to install and set up the environment each time, however thanks to the `Setup` job, the subsequent jobs run faster because of caching.

> Also there's some wrinkles to be aware of around environments / versions - we have the host (ubuntu) and also wp-env (e2e tests) and a custom docker (phpunit) to consider.

Now each test runs it its own job, so nothing is shared (unless retrieved from the cache) between jobs. E2E tests run in their own wp-env, the settings for which are added to `.wp-env.override.json` in each job.

haszari (approved):
Awesome work @opr  - this is looking great!

PR #3547, opened 2020-12-14
Replace IconButton component with Button

### LINKED ISSUE #3511: Fix `IconButton` deprecations for supporting WordPress 5.4+
WordPress 5.4 included an update to the `@wordpress/components` package that deprecates the usage of the `IconButton` component. The recommended replacement is to use `Button` instead.

Since we can bump the minimum WordPress version support from WordPress 5.3 to 5.4 as a part of the WooCommerce Blocks 4.1.0 release (in line with our L2 support policy), we should take care of these deprecations.

This will also address in part the following issues that were created as a part of our WP 5.6 audit:

- #3438
- #3475
- #3479
- #3485

### LINKED ISSUE #3546: Layout looks broken in All Products block edit screen
## Describe the bug
For some reason it looks like the All Products block edit screen is not loading the styles, so it looks completely off.

## To reproduce
1. Add the All Products block to a page.
2. Click the edit button (pencil icon):
imatge
3. Notice bottom buttons are not vertically centered, text on top isn't properly aligned either and the image is too big.

Aljullu:
Fixes #3511.
Fixes #3546.

#### Fix for #3511
1. Add the Featured Product to a post or page.
2. Choose a product and select the block.
3. Verify the Edit media button looks good and ensure there isn't any JS message in the console with a deprecation message for the `IconButton` component.
imatge
4. Repeat steps 1-3 with Featured Category block.
5. Add the All Products block to a page.
6. Click the edit button (pencil icon) and verify the Reset Layout button looks good and there aren't JS errors in the console.
imatge

#### Fix for #3546
1. Add the All Products block to a page (or use the one you just added in the steps above).
2. Click the edit button (pencil icon) and verify the layout looks like in the screenshot on the right.

| Before | After |
| --- | --- |
| imatge | imatge |

### Changelog

> Fix All Products block edit screen.


haszari:
~~Just checking - are we good for backward compat with this change? I'm assuming `IconButton` is available in all our supported WP versions.~~

🎉 Aha I see from the linked issue that we are all good #3511

senadir (approved):

PR #3548, opened 2020-12-14
Rename version compare functions to isWpVersion and isWcVersion and add docblocks

Aljullu:
`compareWithWpVersion` and `compareWithWooVersion` are a bit confusing, so this PR renames them to `isWpVersion` and `isWcVersion` and adds a doc block to each function.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/block.js
+++ b/assets/js/blocks/cart-checkout/checkout/block.js
@@ -27,7 +27,7 @@ import {
 	CHECKOUT_ALLOWS_GUEST,
 	CHECKOUT_ALLOWS_SIGNUP,
 } from '@woocommerce/block-settings';
-import { compareWithWooVersion, getSetting } from '@woocommerce/settings';
+import { isWcVersion, getSetting } from '@woocommerce/settings';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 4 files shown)


mikejolley (approved):
This is useful thanks 👍🏻 

I still wonder if we should refactor this however. For example `isWpVersion = ( version, operator = '=' ) => ...`

Example, `isWpVersion( '5.6', '<= )` is the WordPress version less than or equal to 5.6. 

Thats more readable IMO than `compareWithWpVersion( '5.6', '>' )`  to achieve the same thing.

Aljullu:
> I still wonder if we should refactor this however. For example `isWpVersion = ( version, operator = '=' ) => ...`

I like this new function name! It sounds clearer to me. I will update this PR accordingly, just one suggestion:

What do you think about changing the order of params: `isWpVersion = ( operator, version ) => ...`. It makes the function to be read in a more human-way `isWpVersion( '<=', '5.6' )`. I feel it makes it even clearer that it returns true if WP version is <=5.6. It has the drawback that `operator` can't be optional because it goes first, but I think it's worth it in order to make it easier to understand... Sounds good?

mikejolley:
@Aljullu I prefer optional operator and making it inline with version_compare, but thats me (PHP dev) and others may disagree :)

Aljullu:
@mikejolley I updated this PR with the new function names and I kept the params order as you suggested, I didn't have a strong opinion on that so I'm fine either way. :ok_hand:

mikejolley (approved):
<3 thanks for making those changes, this is way better now!

PR #3550, opened 2020-12-15
GitHub Stale Action: Tweak message for stale activity and bump close boundary

nerrad:
The new stale label action workflow works great but I think it could use a couple tweaks:

- I've temporarily bumped the close trigger threshold to 20 days from 10. This gives some extra allowance for the lack of triaging that will happen over the holidays. I intentionally left the messaging around when things are closed because it will be bumped back to 10 after the holidays.
- I noticed that some non team authors of issues are alarmed with the automation and add a spurious comment just to keep their issue open (which means the stale label gets removed and it likely won't surface in the triaging). So I've tweaked the wording of the messages a bit to hopefully help reduce the unnecessary commenting so that there's time for it to get through triage.

### DIFF
--- a/.github/workflows/close-stale-issues.yml
+++ b/.github/workflows/close-stale-issues.yml
@@ -12,11 +12,11 @@ jobs:
     - uses: actions/stale@v3
       with:
         repo-token: ${{ secrets.GITHUB_TOKEN }}
-        days-before-stale: 90
-        days-before-close: 21
+        days-before-stale: 60
+        days-before-close: 20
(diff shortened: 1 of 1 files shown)


mikejolley (approved):
We should up the emoji game in these messages, but this is ok :) 👍🏻 👏🏻

nerrad:
> We should up the emoji game in these messages, but this is ok :) 👍🏻 👏🏻

Most definitely! 🎸  I'll look into that when I bump the trigger back 🚀 .

PR #3551, opened 2020-12-15
Remove customer saved payment methods from context if they are not enabled.

budzanowski:
### Info

If a user has a saved payment method for a gateway but the option to use saved payment methods is disabled the context still includes the saved methods. This can happen if the merchant has disabled that option after it was initially available. Currently, it is up to the consumer of the context to do the check. IMO this should be the task of the context. This PR adds the check if saved payment methods are allowed to the `paymentMethodDataContext`. This way no saved methods are propagated unless their gateway is enabled and it has the usage of saved payment methods options enabled.

This PR also removes the check that is no longer necessary after the context change.

### Testing
This is related to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3439. That PR relies heavily on the idea of saved payment methods for the UI configuration. Different views are presented depending if the saved methods are present or not.

For this PR a valid test would be to check if the saved payment methods in checkout still work as they did.
Saved not saved, disabled, and enabled in different configurations.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -94,7 +94,7 @@ const getCustomerPaymentMethods = ( availablePaymentMethods = {} ) => {
 	const enabledCustomerPaymentMethods = {};
 	paymentMethodKeys.forEach( ( type ) => {
 		const methods = customerPaymentMethods[ type ].filter(
-			( { method : { gateway } } ) => {
+			( { method: { gateway } } ) => {
 				const isAvailable = gateway in availablePaymentMethods;
 				return isAvailable && availablePaymentMethods[ gateway ].supports?.savePaymentInfo;
 			}
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -81,11 +81,11 @@ export const usePaymentMethodDataContext = () => {
  * Gets the payment methods saved for the current user after filtering out
  * disabled ones.
  *
- * @param {Object[]} availablePaymentMethods List of available payment methods.
+ * @param {Object} availablePaymentMethods List of available payment methods.
  * @return {Object} Object containing the payment methods saved for a specific
  *                  user which are available.
  */
-const getCustomerPaymentMethods = ( availablePaymentMethods = [] ) => {
+const getCustomerPaymentMethods = ( availablePaymentMethods = {} ) => {
 	const customerPaymentMethods = getSetting( 'customerPaymentMethods', {} );
 	const paymentMethodKeys = Object.keys( customerPaymentMethods );
 	if ( paymentMethodKeys.length === 0 ) {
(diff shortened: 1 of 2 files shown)


assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -95,9 +95,17 @@ const getCustomerPaymentMethods = ( availablePaymentMethods = [] ) => {
 	paymentMethodKeys.forEach( ( type ) => {
 		const methods = customerPaymentMethods[ type ].filter(
 			( paymentMethod ) => {
-				return Object.keys( availablePaymentMethods ).includes(
-					paymentMethod.method.gateway
-				);
+				const { gateway } = paymentMethod.method;
+				const isGatewayAvailable = Object.keys(
```
mikejolley: This line uses `Object.keys`. Is `availablePaymentMethods` an array of an object? The doc block says it's an array of objects, and the default value of `availablePaymentMethods` is array. I see this is what was in place before.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -95,9 +95,17 @@ const getCustomerPaymentMethods = ( availablePaymentMethods = [] ) => {
...
+					availablePaymentMethods
+				).includes( gateway );
+				if ( ! isGatewayAvailable ) {
+					return false;
+				}
+				const supportsSavedPayments =
+					availablePaymentMethods[ gateway ].supports
+						?.savePaymentInfo;
+				return supportsSavedPayments;
```
mikejolley: Based on your changes, I wonder if we could simplify this down if the gateway exists in the list of `availablePaymentMethods`

```suggestion
				return (
					typeof availablePaymentMethods[
						paymentMethod.method.gateway
					] !== 'undefined' &&
					availablePaymentMethods[ paymentMethod.method.gateway ]
						.supports?.savePaymentInfo
				);
```

mikejolley (approved):
This appears to be working 👍🏻  I made a suggestion to simplify the logic, and a comment about the expected type of `availablePaymentMethods` which may need addressing also. Pre-approving so it doesn't block you.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
budzanowski: We could, but I would like to avoid it in this case. A agree that the brevity of code is a good quality in general. The context and deeply embedded nature of this suggest that it should be painfully obvious what this does for the future. This is the second time in one month I am revisiting the logic and adding additional checks. I bet that this is not the last time we will do updates here.

budzanowski:
@mikejolley thank you for the review. I will check what is the type of `availablePaymentMethods` and I will adjust either code or comments :)

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
mikejolley: Hmm, I found the original code hard to understand to be honest.

Maybe you could play around with it slightly, for example:

```javascript
const gateway = paymentMethod.method.gateway;
const isGatewayAvailable = typeof availablePaymentMethods[ gateway ] !== 'undefined';
const supported = isGatewayAvailable ? availablePaymentMethods[ gateway ].supports?.savePaymentInfo : false;
return !! supported;
```

Still more compact but easier to read.

budzanowski: OK I will take a look.

budzanowski: I did a bit of research on how to approach this in a most idiomatic way. For the check, if the gateway exists I have used the `in` operator. It feels like it was designed with cases like this in mind
```js
const isAvailable = gateway in availablePaymentMethods;
```
For the second test I have used this simple construct:
```js
return (
	isAvailable &&
	availablePaymentMethods[ gateway ].supports?.savePaymentInfo
);
```
I have replaced the `?` in your suggestions with `&&` to shortcircuit the evaluation. We don't need to worry about the potential `undefined` from the optional chaining operator. The returned value is used by `Array.prototype.filter` and by definition in callback description, we have `Return a value that coerces to true to keep the element, or to false otherwise.` Since `undefined` coerces to false it is a good way to indicate that we don't want to include that gateway method.
@mikejolley thanks for the pushback. It resulted in a much better and cleaner code.

As for the parameter type issue: I am creating a separate PR to fix that, there will be one additional change that would clutter the scope and sense of this PR.

PR #3552, opened 2020-12-15
Fix Address Validation in the Store API and client

### LINKED ISSUE #3525: WooCommerce Checkout Block Validation
I noticed while using the WooCommerce Checkout Block that if a user fills in billing & shipping information, chooses a shipping method, then changes the zip code to something invalid (space, completely invalid text) shipping options often no longer appear, and the order can be submit and payment captured with invalid shipment details (literally, a space as the zip code).

Seems that due to the order-draft storing data, there may be need for additional validation on the front end.

## To reproduce
1. Build a cart
2. Go to checkout & enter Shipping Address (check use same address for billing) details
3. Select shipping option (if not pre-selected)
4. Change Shipping Address to Invalid (change anything, but change the zip code to a space ' ')
5. Click Checkout
6. Watch as form submitted and accepted, payment captured, order in processing, possibly to a wrong address, depending on the intention of a user.

## Additional Notes
Can also be reproduced by building a cart, changing contents & address.
Real life situation: customer builds a cart with an existing address stored, needing to ship to a new address, the order may have already been accepted in the cart when the shipping options might be displayed

## Expected behavior
1. Build a cart
2. Go to checkout & enter Shipping Address (check use same address for billing) details
3. Select shipping option (if not pre-selected)
4. Change Shipping Address to Invalid (change anything, but change the zip code to a space ' ')
5. Click Checkout
6. Invalid shipping address & shipping option selected, checkout not submit, payment not captured

* Core version: 5.5.3
* WooCommerce version: 4.7.0 & 4.8.0
* WooCommerce Blocks version: 3.9 & 4.0

* OS: macOS
* Browser: Safari, Chrome, Firefox

## EDIT:
On inspection of the code, and further front end testing, there is quite a bit of validation, but on some test cases that I myself cannot reproduce, we are able to get orders through using US addresses with no postal code or state even when they are required.

mikejolley:
We have some validation issues on the checkout within the client and the Store API.

The main issue in #3525 comes from whitespace—we are not trimming values before running client side validation. The fix for that is [in this commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/76ed78527c3c48f4fbff98f5a9b917475de40e65).

The StoreAPI does not currently validate anything. It sanitizes values, but does not check they are valid before placing an order. The changes in this PR solve this by:

1. Adding a custom sanitization callback to the address schema—this ensures all address properties are set and sanitizes using appropriate functions.
2. Adding a custom validation callback to the address schema—this checks values are valid according to the schema. For example, a valid country code, a valid postcode, a valid email address.
3. Adding extra validation to the Checkout Route and OrderController—this checks values are valid according to the store setup. For example, an allowed country code, required fields are all set. Anything that should block the order going through.

There are follow ups to improve things further, adding more validation to the client so these errors never occur during a regular flow. [For example, by validating postcodes.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3326)

Fixes #3525

#### White space

For the white space issue, it's easier to test without this PR first to see what happened, and then test it worked.

1. Add something to the cart.
2. Go to checkout.
3. Enter a space character in the postcode field.
4. See if it validates or not. Before this PR it does not—it allows checkout submission.

#### Validation via API

For validation, best to post to the API.

First add something to the cart via the API:

POST `https://one.wordpress.test/wp-json/wc/store/cart/add-item`
Body:
```
{
	"id": 32,
	"quantity":1
}
```

POST https://one.wordpress.test/wp-json/wc/store/checkout

Body:
```
{
	"payment_method": "bacs",
	"billing_address": {
	"first_name": "Mike",
		"last_name": "Jolley",
		"address_1": "Test Address",
		"city": "Test City",
		"country":"US",
		"state":"AL",
		"postcode":"90210",
		"email": "test@test.com"
	},
	"shipping_address": {
		"first_name": "Mike",
		"last_name": "Jolley",
		"address_1": "Test Address",
		"city": "Test City",
		"country":"US",
		"state":"AL",
		"postcode":"90210"
	}
}
```

Leave out certain values to see validation errors. e.g. use an invalid email address, remove the state, enter an invalid state, use a non-existing country etc. Each will produce an error 400 and show a message.

#### Validation via Checkout

Also confirm checkout works as usual and this validation does not block. One thing you can do is checkout using a US address and enter a non-numeric zip code, e.g. ABC. You'll see error notices as the address updates and also if you try to place an order.

### Changelog

> Correctly validate fields containing white space on the checkout.
> Add address validation to values posted to the Checkout via StoreApi.


assets/js/base/utils/errors.js
```diff
@@ -28,3 +33,27 @@ export const formatError = async ( error ) => {
...
 };
+
+/**
+ * Given an API response object, formats the error message into something more human readable.
+ *
+ * @param  {Object}   response           Response object.
+ * @return {string}   Error message.
+ */
+export const formatStoreApiErrorMessage = ( response ) => {
+	if ( response.data && response.code === 'rest_invalid_param' ) {
```
mikejolley, line 44: `rest_invalid_param` is an error code WP uses for errors from the schema validation callbacks. The error messages from these are not great and hide the actual issues. This ensures something actionable is surfaced to the user.

src/StoreApi/Routes/Checkout.php
```diff
@@ -73,12 +73,6 @@ public function get_args() {
 					'context' => $this->get_context_param( [ 'default' => 'view' ] ),
 				],
 			],
-			[
-				'methods'             => WP_REST_Server::EDITABLE,
```
mikejolley, line 77: I had to swap these around because the WP REST API for some reason would not do sanitisation and validation with editable first. I think it's because the methods overlap (POST).

src/StoreApi/Routes/Checkout.php
```diff
@@ -95,7 +89,7 @@ public function get_args() {
 										'type' => 'string',
 									],
 									'value' => [
-										'type' => 'string',
+										'type' => [ 'string', 'boolean' ],
```
mikejolley, line 92: Once I fixed schema validation, I noticed an error from the payments due to a "true" param from a payment method. This prevents it breaking due to schema mismatch.

src/StoreApi/Routes/Checkout.php
```diff
@@ -261,6 +261,17 @@ protected function get_route_post_response( WP_REST_Request $request ) {
 	 */
 	protected function get_route_error_response( $error_code, $error_message, $http_status_code = 500, $additional_data = [] ) {
 		switch ( $http_status_code ) {
+			case 400:
+				return new WP_Error(
+					$error_code,
+					$error_message,
+					array_merge(
+						$additional_data,
```
mikejolley, line 269: Ensures additional data is supported on all error codes.

src/StoreApi/Utilities/OrderController.php
```diff
@@ -155,26 +175,217 @@ public function validate_order_before_payment( \WC_Order $order ) {
...
+		}
+	}
+
+	/**
+	 * Validates customer address data based on the locale to ensure required fields are set.
 	 *
+	 * @throws RouteException Exception if invalid data is detected.
+	 * @param \WC_Order $order Order object.
+	 */
+	protected function validate_addresses( \WC_Order $order ) {
```
mikejolley, line 215: This ensures addresses are valid in the order object itself before payment.

mikejolley:
I just pushed a change which removes the debounced fields with local state cc @Aljullu This fixed the other issue reported in #3525 and correctly clears fields when county changes. I didn't notice much of a difference when typing. I increased the debounce for the server updates to compensate.

src/StoreApi/Utilities/OrderController.php
```diff
@@ -155,26 +175,217 @@ public function validate_order_before_payment( \WC_Order $order ) {
...
+	}
+
+	/**
+	 * Check all required address fields are set and return errors if not.
+	 *
+	 * @param string $country Country code.
+	 * @param array  $allowed_countries List of valid country codes.
+	 * @return boolean True if valid.
+	 */
+	protected function validate_allowed_country( $country, $allowed_countries ) {
```
Aljullu: Should this method default to false if `$allowed_countries` is not an array or it's an empty array? I was allowed to do a purchase in a store with these settings:

imatge

src/StoreApi/Utilities/OrderController.php
```diff
@@ -155,26 +175,217 @@ public function validate_order_before_payment( \WC_Order $order ) {
...
+				'label'    => __( 'Country/Region', 'woo-gutenberg-products-block' ),
+				'required' => true,
+			],
+			'city'       => [
+				'label'    => __( 'City', 'woo-gutenberg-products-block' ),
+				'required' => true,
+			],
+			'state'      => [
+				'label'    => __( 'State/County', 'woo-gutenberg-products-block' ),
+				'required' => true,
```
Aljullu, line 340: I tried to do a purchase from Afghanistan as a test and I wasn't allowed because the API was expecting a state:

imatge

Can you reproduce that too?

src/StoreApi/Schemas/AbstractAddressSchema.php
```diff
@@ -0,0 +1,181 @@
...
+			],
+		];
+	}
+
+	/**
+	 * Sanitize and format the given address object.
+	 *
+	 * @param array            $address Value being sanitized.
+	 * @param \WP_REST_Request $request The Request.
+	 * @param string           $param THe param being sanitized.
```
Aljullu: Typo. :slightly_smiling_face: 

```suggestion
	 * @param string           $param The param being sanitized.
```

src/StoreApi/Schemas/AbstractAddressSchema.php
```diff
@@ -0,0 +1,181 @@
...
+	 * @return array
+	 */
+	public function sanitize_callback( $address, $request, $param ) {
+		$address               = array_merge( array_fill_keys( array_keys( $this->get_properties() ), '' ), (array) $address );
+		$address['country']    = wc_strtoupper( wc_clean( $address['country'] ) );
```
Aljullu: Out of curiosity, why `$address['country']` is not wrapped with `wp_unslash()` as well?

Aljullu (commented):
I tested following the testing steps and everything looked good. However, I think I found a couple of issues (first two comments below :point_down: ) that might need to be addressed before merging.

Regarding the `DebouncedValidatedTextInput`, I spent some time with my old Android phone trying to detect if there was a performance regression but I couldn't notice it, so I'm fine removing it. :ok_hand:

src/StoreApi/Utilities/OrderController.php
mikejolley: Ill test this. Maybe a problem combining this with the locale.

src/StoreApi/Schemas/AbstractAddressSchema.php
mikejolley: Oversight probably :)

src/StoreApi/Utilities/OrderController.php
mikejolley: Found it, there was a typo in the locale variable name 🤦🏻  Works now :)

Aljullu (approved):
Re-tested and the couple of issues I reported are fixed. Code changes look fine too, so LGTM.

assets/js/base/components/text-input/validated.js
```diff
@@ -107,8 +114,8 @@ const ValidatedTextInput = ( {
 			}
 			ref={ inputRef }
 			onChange={ ( val ) => {
-				hideValidationError( errorId );
 				onChange( val );
+				hideValidationError( errorId );
```
nerrad: I'm curious, why the change of sequence here?

assets/js/base/utils/errors.js
```diff
@@ -28,3 +33,27 @@ export const formatError = async ( error ) => {
...
+	}
+
+	if ( response?.message ) {
+		return response?.message;
+	}
+
+	return __(
+		'Something went wrong. Please contact us to get assistance.',
+		'woo-gutenberg-products-block'
+	);
```
nerrad: This could probably be simplified to just be something like:

```suggestion
	return response?.message || __(
		'Something went wrong. Please contact us to get assistance.',
		'woo-gutenberg-products-block'
	);
```

src/StoreApi/Schemas/AbstractAddressSchema.php
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * AddressSchema class.
+ *
+ * Provides a generic address schema for composition in other schemas.
+ *
+ * @internal This API is used internally by Blocks--it is still in flux and may be subject to revisions.
+ * @since 4.1.0
+ */
+abstract class AbstractAddressSchema extends AbstractSchema {
```
nerrad, line 14: I don't want to block the pull request so this is mostly just food for thought.

I have seen validation and sanitization implemented via the [strategy design pattern](https://refactoring.guru/design-patterns/strategy/php/example) where you inject the Sanitization/Validation strategy used (that implements a defined interface). So for example there could be a `ValidationInterface` and `SanitizationInterface` and various implementations of those interfaces (of which `Address` would be one). `RequestAddressValidation` or `RequestAddressSanitization` in turn could be compositions of other things like `EmailValidation` etc.

Then, your schemas would just receive the validation and sanitization strategies to use injected into the constructors.

This could feasibly allow for the validation and sanitization strategies to be re-used in multiple places needing validation or sanitization.

I realize implementing something like this right now might be overkill or premature, but thought I'd mention anyways in case you come across places where this could potentially be implemented in the future.

src/StoreApi/Utilities/OrderController.php
```diff
@@ -103,26 +104,45 @@ public function sync_customer_data_with_order( \WC_Order $order ) {
...
-			try {
-				$this->validate_coupon_email_restriction( $coupon, $order );
-			} catch ( \Exception $error ) {
-				$coupon_errors[ $coupon_code ] = $error->getMessage();
-			}
-			try {
-				$this->validate_coupon_usage_limit( $coupon, $order );
-			} catch ( \Exception $error ) {
-				$coupon_errors[ $coupon_code ] = $error->getMessage();
+			foreach ( $validators as $validator ) {
```
nerrad: Might be a good place for usage of [`array_walk`](https://www.php.net/manual/en/function.array-walk.php) to keep things a bit more concise?

src/StoreApi/Utilities/OrderController.php
```diff
@@ -155,26 +175,210 @@ public function validate_order_before_payment( \WC_Order $order ) {
...
+	}
+
+	/**
+	 * Check all required address fields are set and return errors if not.
 	 *
+	 * @param array     $address Address array.
+	 * @param string    $address_type billing or shipping address, used in error messages.
+	 * @param \WP_Error $errors Error object.
+	 */
+	protected function validate_address_fields( $address, $address_type = 'billing', \WP_Error &$errors ) {
```
nerrad: Why passing $errors by reference here? `$errors` is still a pointer to the `\WP_Error` instance so it's not creating a new copy of `\WP_Error`?

nerrad (approved):
Looks great Mike! I love the fact that there is now validation and sanitization on the StoreAPI! 

I didn't test - trusting Albert's and your results on that.

I did leave a few comments but none of them are blocking and you can action however you want. Most questions are just for clarification/understanding.

assets/js/base/components/text-input/validated.js
mikejolley: O, I think it was because in onBlur I made a change where something was called before the onBlur from props and I changed this to match, but I ended up reverting one and not the other. I will change it back; it's not related to this work.

src/StoreApi/Schemas/AbstractAddressSchema.php
mikejolley: I think shared validation would be a nice idea for the logic now within OrderController! 

At schema level the validation and sanitised is directly coupled to the defined schema, meaning if schema changes, sanitisation and validation also needs changing, so I think there is less need for reuse. To clarify; the validation callbacks within the Schema classes validate against the defined Schema. e.g. Schema defines an enum, and the validation validates against that same enum.

mikejolley:
Did some additional testing to ensure express payments still go through and it's working :) Merging!

src/StoreApi/Routes/CartUpdateCustomer.php
```diff
@@ -72,16 +72,20 @@ public function get_args() {
...
-						'description' => __( 'Shipping address.', 'woo-gutenberg-products-block' ),
-						'type'        => 'object',
-						'context'     => [ 'view', 'edit' ],
-						'properties'  => $this->shipping_address_schema->get_properties(),
+						'description'       => __( 'Shipping address.', 'woo-gutenberg-products-block' ),
+						'type'              => 'object',
+						'context'           => [ 'view', 'edit' ],
+						'properties'        => $this->shipping_address_schema->get_properties(),
+						'sanitize_callback' => [ $this->shipping_address_schema, 'sanitize_callback' ],
+						'validate_callback' => [ $this->shipping_address_schema, 'validate_callback' ],
```
senadir, line 88: Even though we're copying billing to shipping in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3552/files#diff-c8a67488bba0001a6420427352799810b838b6b9a7f18626427e69e7e47fb317R109 those two lines are still running and causing #3592

PR #3554, opened 2020-12-15
Fix select inputs when dark mode is enabled in Twenty Twenty-One

### LINKED ISSUE #3518: Twenty Twenty One: selects are barely visible in dark mode
## To reproduce
1. Install and activate Twenty Twenty One.
2. Go to Customize > Colors & Dark mode and check the Dark mode support checkbox.
3. In the frontend, make sure the dark mode is enabled and go to a page with the Checkout block.
4. Open any select and notice the text is barely visible:

imatge

## Expected behavior
There should be more contrast between the select background and content.

opr:
### Description

This PR adds a new rule to` assets/js/base/components/select/style.scss` for when the class `is-dark-theme` exists (it is put on the body by Twenty Twenty-One theme).

This PR is required because without it, when the theme is set to dark mode (irrespective of the browser/OS dark mode preference) the text in the select is barely visible. 

Fixes #3518 

#### Accessibility

#### Before
imatge

#### After when block has dark mode inputs enabled
Screenshot 2020-12-15 at 18 37 09

#### After when block doesn't have dark mode inputs enabled.
Screenshot 2020-12-15 at 18 37 09

1. Install and activate Twenty Twenty One.
2. Go to Customize > Colors & Dark mode and check the Dark mode support checkbox.
3. `npm run build`.
5. Add a checkout block to a page, ensure items are in your basket.
6. Visit the page, enable dark mode (bottom right of screen), and open the country dropdown.
7. Ensure text is readable.
8. Change checkout block settings to enable dark mode inputs.
9. Check again and ensure the colour of the text has changed to white on black and that it is is still readable.

### Changelog

> Fixed text visibility in select inputs when using Twenty Twenty-One theme's dark mode.

### DIFF
--- a/assets/js/base/components/select/style.scss
+++ b/assets/js/base/components/select/style.scss
@@ -118,3 +118,18 @@
 		display: none;
 	}
 }
+
+// This class is set on the body by Twenty Twenty-One when dark mode is active.
+.is-dark-theme {
+	// If the theme is in dark mode, as well as the block, then this selector will match.
+	.has-dark-controls {
+		.components-custom-select-control__item {
+			color: $input-text-dark;
+		}
+	}
+
+	// If the theme is in dark mode, but the block isn't, then this selector will match.
+	.components-custom-select-control__item {
+		color: $input-text-active;
+	}
+}
--- a/assets/js/base/components/select/style.scss
+++ b/assets/js/base/components/select/style.scss
@@ -119,17 +119,18 @@
 	}
 }
 
-// This class is set on the body by Twenty Twenty-One when dark mode is active.
-.is-dark-theme {
-	// If the theme is in dark mode, as well as the block, then this selector will match.
-	.has-dark-controls {
-		.components-custom-select-control__item {
-			color: $input-text-dark;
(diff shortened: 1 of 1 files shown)


assets/js/base/components/select/style.scss
```diff
@@ -118,3 +118,18 @@
 		display: none;
 	}
 }
+
+// This class is set on the body by Twenty Twenty-One when dark mode is active.
+.is-dark-theme {
```
Aljullu: I didn't do a full-review, but just a quick note: when writing styles for a specific theme, we usually wrap them in a class, like so: `.theme-twentytwentyone { ... }`. You can take the [`CheckboxControl` styles](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/components/checkbox-control/style.scss#L65) as a reference. :slightly_smiling_face:

opr: Thanks for pointing that out, I've updated the pull with this change now.

mikejolley (approved):
Tests well, :shipit:

PR #3557, opened 2020-12-16
add fallback screen reader styles

### LINKED ISSUE #3265: Add styles for `.screen-reader-text` to cover themes that don't have that style defined
In [this comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3169#issuecomment-706653184) a user of the blocks reported an issue which upon investigation revealed that the theme for his site is not styling the `screen-reader-text` class.

It looks like [WooCommerce core](https://github.com/woocommerce/woocommerce/blob/16faf915e0055e48d4f0c2f647696716df957bbb/assets/css/woocommerce.scss#L41-L48) adds these styles, presumably as a fallback for themes that don't add the it (the blame for that code indicates so too).

This issue is for adding something similar for WooCommerce Blocks as well for cases where blocks are loaded on pages not loading the `woocommerce.scss` styles.

haszari:
Fixes #3265 

This PR adds styling for screen-reader text elements. Usually this is handled by the theme. However if the theme doesn't have support the screen reader text may be rendered inline (which is not intended); these fallback styles ensure screen reader text is hidden by default, and provides `:focus` styles for when the text has focus and is being read.

The specifics of the styles are pasted direct from underscores, with some colours changed. Feedback welcome on how to generalise these so they are a minimal, safe fallback (i.e. should not be opinionated).

#### Accessibility

I've tested with Safari VoiceOver but I couldn't focus the element in normal use. Keen for some testing of this with various browsers/platforms and screen reader tech :)

### Screenshots

Focused in chrome and safari:

Chrome

Safari

Notes: 

- I faked `:focus` using browser dev tools.
- When unfocused, screen reader text should not be visible.

For comparison, here's a screenshot of the issue:

Screen Shot 2020-12-16 at 2 27 46 PM

1. Install a theme that doesn't provide screen reader styles. I built a new `Bad Theme` using [Underscores](https://underscores.me) and deleted the screen reader styles.
2. Add checkout block to a page and stuff to cart.
3. View checkout on front end - confirm `.screen-reader-text` is hidden by default and is helpful when using a screen reader.
4. Test with other themes to ensure these styles don't conflict or cause issues.

### Changelog

> Fix: Added fallback styling for screen reader text.

### DIFF
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -52,6 +52,7 @@ $fontSizes: (
 
 // Hide an element from sighted users, but available to screen reader users.
 @mixin visually-hidden() {
+	border: 0;
 	clip: rect(1px, 1px, 1px, 1px);
 	clip-path: inset(50%);
 	height: 1px;
@@ -61,6 +62,29 @@ $fontSizes: (
 	/* Many screen reader and browser combinations announce broken words as they would appear visually. */
 	overflow-wrap: normal !important;
 	word-wrap: normal !important;
+	padding: 0;
+	position: absolute !important;
+	width: 1px;
+}
+
+@mixin visually-hidden-focus-reveal() {
+	background-color: #fff;
+	border-radius: 3px;
+	box-shadow: 0 0 2px 2px rgba(0, 0, 0, 0.6);
+	clip: auto !important;
+	clip-path: none;
+	color: $input-text-active;
+	display: block;
+	font-size: 0.875rem;
+	font-weight: 700;
+	height: auto;
+	left: 5px;
+	line-height: normal;
+	padding: 15px 23px 14px;
+	text-decoration: none;
+	top: 5px;
+	width: auto;
+	z-index: 100000;
 }
 
 @mixin reset-box() {
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -303,3 +303,38 @@
 		}
 	}
 }
+
+// Default screen-reader styles. Included as a fallback for themes that don't have support.
+// These styles are based on the styles included in _underscores theme.
+// https://underscores.me
+.screen-reader-text {
+	border: 0;
+	clip: rect(1px, 1px, 1px, 1px);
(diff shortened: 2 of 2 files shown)


Aljullu (approved):
> I've tested with Safari VoiceOver but I couldn't focus the element in normal use. Keen for some testing of this with various browsers/platforms and screen reader tech :)

If you were using the <kbd>Tab</kbd> key, I think that's to be expected. Screen readers usually provide a different navigation method which is more "line-by-line" so you can navigate to non-focusable items. With [Orca](https://help.gnome.org/users/orca/stable/) it's done with the arrow keys, I guess it's probably the same in other screen readers?

I tested removing the `screen-reader-text` styles from Storefront and I can confirm this PR adds a fallback so screen reader elements are hidden.

In order to test `:focus` styles, I converted the `<FormStep>` `legend` element into an `a` and it looked great:

imatge

My only concern is that we are introducing a new set of 'screen-reader' styles, which differs from the ones in other projects:
  * [WooCommerce](https://github.com/woocommerce/woocommerce/blob/16faf915e0055e48d4f0c2f647696716df957bbb/assets/css/woocommerce.scss#L41-L48)
  * [WooCommerce Blocks mixin](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/css/abstracts/_mixins.scss#L53-L64)
  * [Storefront](https://github.com/woocommerce/storefront/blob/trunk/assets/css/base/_base.scss#L507-L531)

Not blocking merge, but do you think we could use one of those :point_up_2: as the base instead of introducing a new one from underscores? :slightly_smiling_face:

haszari:
> My only concern is that we are introducing a new set of 'screen-reader' styles … we could use one of those 👆

Good thinking @Aljullu , I should have looked more carefully for prior art!

I've switched to using our (most local) `visually-hidden()` mixin. I also made a new mixin for the companion focus style, since they are coupled.

Aljullu (approved):
Nice to see a new mixing for the focus state. I didn't re-test it but code changes look good. :shipit:

PR #3568, opened 2020-12-17
respect merchant setting for Stripe CC `title`

### LINKED ISSUE #2919: Support Stripe payment method title store setting
**Describe the bug**
Stripe Credit Card has a `Title` setting. Currently we're using a hard-coded `Credit Card` title for the tab - it would be good to use the setting. This allows stores to optimise the wording as appropriate, or disambiguate multiple payment methods.

Screen Shot 2020-07-28 at 3 28 01 PM

**To Reproduce**
1. Set up checkout block.
2. Set up Stripe, enable Stripe Credit Card.
3. Edit settings, change title (or leave default title, which is different from the title in blocks).
4. View checkout block and checkout shortcode on front end - note the shortcode uses the store setting, block doesn't.

Screen Shot 2020-07-28 at 3 26 41 PM

Screen Shot 2020-07-28 at 3 26 58 PM

haszari:
Fixes #2919 

This PR exposes the `title` setting to the client and uses it for the tab label. This respects the customisation option already available in legacy checkout (and other payment methods in blocks too).

### Screenshots
Here I've gone with `Stripes Timez`…

Screen Shot 2020-12-18 at 12 55 14 PM

1. Set up Stripe CC and checkout block.
2. View front end, add stuff to cart, view checkout.
3. Stripe should work correctly.
4. Customise title in `WooCommerce > Settings > Payments > Stripe` and save.
5. View checkout again, should use custom title in checkout block. In general should work similar to shortcode checkout.

### Changelog

> Fix: Checkout Block – Support customised `title` setting for Stripe Credit Card payment method.

### DIFF
--- a/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
+++ b/assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
@@ -36,11 +36,11 @@ const StripeComponent = ( props ) => {
 const StripeLabel = ( props ) => {
 	const { PaymentMethodLabel } = props.components;
 
-	return (
-		<PaymentMethodLabel
-			text={ __( 'Credit / Debit Card', 'woo-gutenberg-products-block' ) }
-		/>
-	);
+	const labelText = getStripeServerData().title
+		? getStripeServerData().title
+		: __( 'Credit / Debit Card', 'woo-gutenberg-products-block' );
+
+	return <PaymentMethodLabel text={ labelText } />;
 };
 
 const cardIcons = getStripeCreditCardIcons();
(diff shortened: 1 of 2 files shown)


haszari:
Perfect random reviewer - @budzanowski  this will hopefully merge nicely with #3439 🧯 

FYI @nerrad I grabbed this in cooldown, seems like a neat safe improvement.

assets/js/payment-method-extensions/payment-methods/stripe/credit-card/index.js
```diff
@@ -36,11 +36,11 @@ const StripeComponent = ( props ) => {
...
 	const { PaymentMethodLabel } = props.components;
 
-	return (
-		<PaymentMethodLabel
-			text={ __( 'Credit / Debit Card', 'woo-gutenberg-products-block' ) }
-		/>
-	);
+	const labelText = getStripeServerData().title
+		? getStripeServerData().title
+		: __( 'Credit / Debit Card', 'woo-gutenberg-products-block' );
```
budzanowski, line 41: Maybe we can skip the `?` check and use directly `getStripeServerData().title` in this place since we are already testing this on the API side?

haszari: This gives us an extra check on client, we shouldn't rely on the server default IMO - just more robust to have a late check. Also the behaviour for empty string (`title === ''`) might be different in JS vs PHP.

budzanowski (approved):

PR #3569, opened 2020-12-18
Add 'hidden' class name when using the 'hidden' attribute

Aljullu:
Some themes might add custom styles which override the browser styles for the `hidden` attribute. For example, the Artisan theme has a `div { display: block; }` rule, that made the `hidden` elements to be displayed anyway. This PR adds the `hidden` class to those elements in order to fix it.

Some more context in p9h1Lv-1KE-p2#comment-6046.

0. Install [Artisan](https://woocommerce.com/products/artisan/) theme.
2. Go to the Cart or Checkout blocks and verify you can expand/contract the Coupon Code panel.

_Before:_
Peek 2020-12-18 12-20

_After:_
Peek 2020-12-18 12-35

3. Edit a product and in the Inventory tab check the option so it can only be sold individually.
4. Create a post or page with the Add to cart atomic block.
5. In the frontend, visit that page and verify there is no number input.

_Before:_
imatge

_After:_
imatge

### Changelog

> Fixed Coupon Code panel not expanding/contracting in some themes.

### DIFF
--- a/assets/js/atomic/blocks/product-elements/add-to-cart/shared/quantity-input.js
+++ b/assets/js/atomic/blocks/product-elements/add-to-cart/shared/quantity-input.js
@@ -1,4 +1,9 @@
 /**
+ * External dependencies
+ */
+import classnames from 'classnames';
+
+/**
  * Quantity Input Component.
  *
  * @param {Object}         props          Incoming props for component
@@ -11,7 +16,10 @@
 const QuantityInput = ( { disabled, min, max, value, onChange } ) => {
 	return (
 		<input
-			className="wc-block-components-product-add-to-cart-quantity"
+			className={ classnames(
+				'wc-block-components-product-add-to-cart-quantity',
+				{ hidden: max === 1 }
+			) }
 			type="number"
 			value={ value }
 			min={ min }
--- a/assets/js/base/components/panel/index.js
+++ b/assets/js/base/components/panel/index.js
@@ -42,7 +42,9 @@ const Panel = ( {
 				</button>
 			</TitleTag>
 			<div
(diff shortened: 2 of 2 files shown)


haszari (approved):
Tested and confirmed the issue in Artisan and that this fixes it. Code looks reasonable.

However .. this feels like it belongs in Artisan theme. Perhaps we should report the issue (alongside the other various wrinkles!) to `Organic Themes` via marketplace team. Could also add a theming guideline that [`hidden` attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden) should work in Woo themes.

The class `hidden` used to hide the coupon form is defined in artisan styles - we don't provide this in blocks (e.g. `.hidden { display: none }`). So adding the class in blocks relies on the theme providing something like that.

I'm thinking either we:

- Explicitly add `hidden` class and our own `.hidden {}` css rule as a fallback so hiding is more robust in various themes/browsers.
- _OR_
- Continue using `hidden` attribute only and add guidelines for themers (and report issues).

I'm marking as approved - up to you how to proceed 👍

senadir (approved):

Aljullu:
Good points @haszari. I agree that's an issue in the Artisan theme. Initially, I thought most themes would have styles targetting the `.hidden { ... }` class name so I thought using it would be fine (something like `screen-reader-text`). But taking a look at Storefront, Tweny Twenty and Twenty Twenty One, none of them has styles for the `.hidden` class, so that's probably not the way to proceed.

I opened a new PR that reverts the changes in this PR, while enforcing `display: none` for the hidden panel contents: #3582.

Issue #2200, opened 2020-04-13
postcss-loader is stripping SCSS sourcemap url


senadir:
**Describe the bug**
It seems that even though `sass-loader` is generating a source map, it is getting stripped out in the final build, a suspect for that could be `postcss-loader` from what I [read here](https://stackoverflow.com/a/53122155).

**To Reproduce**

1. run `npm start` to generate files
2. notice that sourcemap files for styles are being created `build/styls.css.map`.
3. open a style file like `build/style.css` and scroll to the end of it, there is no sourceMapURL comment to include the source map.

haszari:
To clarify the impact of this … I'm assuming the missing sourcemap means it's more painful debugging styles. Does this affect both release and builds @senadir ? Let's tweak the title so it highlights the problem/impact (rather than the technical details).

senadir:
Sourcemaps are not included in production builds, they're a development tool so they're only available in `npm start`.

>  Let's tweak the title so it highlights the problem/impact (rather than the technical details).

I'm not sure I understand, this is a purely technical problem that affects the development team on a technical level, so I think the title is fine, how would you change it?

haszari:
> I'm not sure I understand, this is a purely technical problem that affects the development team on a technical level, so I think the title is fine, how would you change it?

I'm thinking something like `Debugging SCSS is broken - minified css, source maps broken`. Focusing on the outcome that we want - debugging SCSS should be easy, the SCSS should show in the dev tools. 

Anyway I might take a look at this in cooldown :)

haszari:
Confirmed the issue - when debugging using `npm start` the css files have no source maps in dev tools, you can only see minified CSS. I was looking at our 3 css files - `editor.css` `style.css` `vendors-style.css`.

I noticed two issues:

1. The css files don't reference a source map at all (the issue mentioned in description). I think there should be a comment something like this: `/*# sourceMappingURL=style.css.map */`.
2. The actual source map files look empty to me. For example, this is what my `style.css.map` contains: `{"version":3,"sources":[],"names":[],"mappings":"","file":"style.css","sourceRoot":""}`.

So looks like there's a problem generating the maps.

haszari:
Saw this warning that might be relevant: 

```
Building Styles [=================== ] 93% after chunk asset optimization (37.1 seconds)
Without `from` option PostCSS could generate wrong source map and will not find Browserslist config. 
Set it to CSS file path or to `undefined` to prevent this warning.
[...]

PR #3573, opened 2020-12-20
add sourceMap: true option in various style related webpack loaders

haszari:
Related #2200 

This PR adds `sourceMap: true` config option to all our style-related webpack config. This goes some way to fixing our sourcemap problems; the previous sourcemaps were empty. 

However it's not a complete solution to #2200 - there's still an issue linking the generated css file to the source map. I don't know how to solve this, we could try not using `MiniCssExtractPlugin` in dev build maybe. Note I also tried adding `devtool: 'source-map'` to our webpack config, this didn't make any difference.

Raising this PR for review anyway, I think this is worth merging as it's a step forward 🤷

### DIFF
--- a/bin/webpack-configs.js
+++ b/bin/webpack-configs.js
@@ -464,11 +464,23 @@ const getStylingConfig = ( options = {} ) => {
 					test: /\/node_modules\/.*?style\.s?css$/,
 					use: [
 						MiniCssExtractPlugin.loader,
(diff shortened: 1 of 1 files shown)


senadir:
Because our build process changed a bit since this PR was presented and it doesn't yet offer a solution, creating a source map without actually using it is a net negative for me because it increases the build time without any visible outcome, I will close this for now and might reopen it and build from it if we decide to dedicate time to fix this.

PR #3576, opened 2020-12-21
Fix data type used for paymentData.paymentMethods handling.

### LINKED ISSUE #3574: Fix paymentData.paymentMethods used as a wrong data type.
`paymentData.paymentMethods` is an `Object` but it was used as an `Array` in following places:
- https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js#L88
- https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js#L187

 There were no visible problems with this but it could potentially cause problems in the future.

budzanowski:
This PR adjusts the `paymentData.paymentMethods` assumed type in `getCustomerPaymentMethods` it is an `Object` not an `Array`. Besides the parameter and documentation the code was adjusted to reflect the correct type.

Fixes #3574 

Check that saved payment methods still work. Two cases:
1. User has saved payment methods
2. User has no saved payment methods

In general, this error was not causing issues since there was a proper check later in the code ( now removed since it is no longer necessary ), but this could potentially cause issues in the future if used improperly.

Props to @mikejolley for the comment in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3551#discussion_r545888046 that led to this PR.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -88,9 +88,6 @@ export const usePaymentMethodDataContext = () => {
 const getCustomerPaymentMethods = ( availablePaymentMethods = {} ) => {
 	const customerPaymentMethods = getSetting( 'customerPaymentMethods', {} );
 	const paymentMethodKeys = Object.keys( customerPaymentMethods );
-	if ( paymentMethodKeys.length === 0 ) {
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
LGTM

Issue #3438, opened 2020-11-23
WP 5.6 – deprecation warnings in various blocks


haszari:
Note: This is a megabug - adding all the deprecation warnings I see when testing against 5.6 here. If needed we can split these off to separate issues later.

## Describe the bug
Editing the `Featured Category` or `Featured Product` block shows some deprecation warnings in browser dev console:

- `[Warning] Using Toolbar without label prop is deprecated. Please use ToolbarGroup component instead. See: https://developer.wordpress.org/block-editor/components/toolbar/ (deprecated.min.js, line 2)`
- `[Warning] wp.components.IconButton is deprecated. Please use wp.components.Button instead. (deprecated.min.js, line 2)`
- `[Warning] Using custom components as toolbar controls is deprecated. Please use ToolbarItem or ToolbarButton components instead. See: https://developer.wordpress.org/block-editor/components/toolbar-button/#inside-blockcontrols (deprecated.min.js, line 2)`

Also similar issues with `Hand-picked Products` and `Products by Tag`:

- `[Warning] Using Toolbar without label prop is deprecated. Please use ToolbarGroup component instead. See: https://developer.wordpress.org/block-editor/components/toolbar/ (deprecated.min.js, line 2)`
- `[Warning] wp.components.IconButton is deprecated. Please use wp.components.Button instead. (deprecated.min.js, line 2)`

<img width="794" alt="Screen Shot 2020-11-23 at 1 16 22 PM" src="https://user-images.githubusercontent.com/4167300/99921266-1bcda500-2d8e-11eb-9b78-5250ef2fa8f9.png">

<img width="753" alt="Screen Shot 2020-11-23 at 1 16 11 PM" src="https://user-images.githubusercontent.com/4167300/99921258-11131000-2d8e-11eb-8cd0-30e7a979232b.png">

`Product Categories List` and `All Reviews` are both using `Button .isDefault`:

- `Button isDefault prop is deprecated. Please use isSecondary instead.`

## To reproduce
1. Go to block editor, open dev tools console.
2. Add a `Featured Category` block.
3. Click out of editing the inner button block to edit the  `Featured Category` block.
4. See console deprecated warnings.

### WordPress (please complete the following information):
* Core version: 5.6-RC.1
* Gutenberg version (if installed): n/a
* WooCommerce version: 4.8.0-beta.1
* WooCommerce Blocks version: 3.8.0

Aljullu:
[...]

PR #3579, opened 2020-12-21
use ReplaceMediaFlow in featured category and feature product

senadir:
Part of #3438

This PR updates our approach to uploading images by using `ReplaceImageFlow` (h/t @haszari ) in both Featured Product and Featured Category and result in a better experience.

1. Insert Featured Category and Featured Product
2. Try updating the image via media library or directly uploading.
3. There should be no console errors or warnings.
4. Your upload should work fine.

### Changelog

> Add the ability to directly upload an image in Featured Category and Featured Product blocks.

### DIFF
--- a/assets/js/blocks/featured-category/block.js
+++ b/assets/js/blocks/featured-category/block.js
@@ -102,7 +102,6 @@ const FeaturedCategory = ( {
 					} }
 					allowedTypes={ [ 'image' ] }
 					value={ mediaId }
-					disabled={ ! category }
 				/>
 			</BlockControls>
 		);
--- a/assets/js/blocks/featured-category/block.js
+++ b/assets/js/blocks/featured-category/block.js
@@ -101,7 +101,6 @@ const FeaturedCategory = ( {
 						} );
 					} }
 					allowedTypes={ [ 'image' ] }
-					value={ mediaId }
 				/>
 			</BlockControls>
 		);
--- a/assets/js/blocks/featured-category/block.js
+++ b/assets/js/blocks/featured-category/block.js
@@ -7,8 +7,7 @@ import {
 	BlockControls,
 	InnerBlocks,
 	InspectorControls,
-	MediaUpload,
-	MediaUploadCheck,
+	MediaReplaceFlow,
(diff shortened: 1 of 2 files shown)


assets/js/blocks/featured-category/block.js
```diff
@@ -92,29 +90,19 @@ const FeaturedCategory = ( {
...
+					mediaURL={ mediaSrc }
+					accept="image/*"
+					onSelect={ ( media ) => {
+						setAttributes( {
+							mediaId: media.id,
+							mediaSrc: media.url,
+						} );
+					} }
+					allowedTypes={ [ 'image' ] }
+					value={ mediaId }
```
Aljullu: I think `value` can be removed since it's not in the list of props accepted by this component.

assets/js/blocks/featured-product/block.js
```diff
@@ -145,29 +144,20 @@ const FeaturedProduct = ( {
...
+					mediaURL={ mediaSrc }
+					accept="image/*"
+					onSelect={ ( media ) => {
+						setAttributes( {
+							mediaId: media.id,
+							mediaSrc: media.url,
+						} );
+					} }
+					allowedTypes={ [ 'image' ] }
+					value={ mediaId }
```
Aljullu: Same. :slightly_smiling_face:

Aljullu (approved):
Really nice addition to two of our oldest blocks. I left a comment about a prop that I think is unnecessary. Also, I wonder if we should use another `name` for the button, something like "Replace image" instead of "Replace" to make it clearer. But up to you, everything else looks good to me. :ok_hand:

senadir:
> Also, I wonder if we should use another name for the button, something like "Replace image" instead of "Replace" to make it clearer. But up to you, everything else looks good to me. 👌

I'm afraid it might turn out to be very wordy and take up a lot of space, I believe this was discussed in the [original PR](https://github.com/WordPress/gutenberg/pull/16200) but I'm not sure.

haszari:
Tested and confirmed this is working well in WP 5.4 👍

PR #3584, opened 2020-12-22
Migrate WordPress PHP unit (integration) tests to GitHub actions (and remove travis)

haszari:
_In progress_

Related #3542

This PR implements a GitHub action workflow for running our PHP tests. The goal is to run our PHP tests on all commits to PRs as a check.

There's a bit more to do before merge:

- Implement PHP tests for different versions of PHP. We had this in Travis, so we should replicate same conditions (or revisit if this is out of date). 
- Review and implement other conditions - specifically `WP_VERSION`, `WOOCOMMERCE_BLOCKS_PHASE`, not sure what `PHP_UNIT` is doing.
- Remove the Travis config (i.e. migrate from Travis to GitHub Actions).

And some other things to consider:

- Should we share / cache state to speed these up? 
- Can we share state with other actions?
- What platform do we want to run on - debian, ubuntu, macOS etc?
- Is the workaround installing build deps for `nodegit` providing value, or could we address that more elegantly?

1. Scroll down to view the checks for this PR, see the new one: **`PHP Tests`**.
2. Push some code to this branch that breaks a PHP test 😈 
3. See the test fail 💇‍♂️ 
4. (Revert your changes.)

### DIFF
--- a/.github/workflows/php-wp-unit-tests.yml
+++ b/.github/workflows/php-wp-unit-tests.yml
@@ -16,6 +16,13 @@ jobs:
         with:
           node-version: 14.x
 
+      - name: Set up PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
+          coverage: none
+          tools: composer, cs2pr
+
       - name: Install devtools for nodegit (dep) build 🤨
         run: |
           sudo apt-get update
--- a/.github/workflows/php-wp-unit-tests.yml
+++ b/.github/workflows/php-wp-unit-tests.yml
@@ -5,19 +5,19 @@ on:
 
 jobs:
   PHP_Tests:
-    name: 'PHP WP Tests'
+    name: PHP WP Tests
     runs-on: ubuntu-latest
 
-  steps:
-    - uses: actions/checkout@v2
+    steps:
+      - uses: actions/checkout@v2
 
-    - name: 'Install Node.js'
-      uses: actions/setup-node@v1
-      with:
-        node-version: 14.x
+      - name: Install Node.js
+        uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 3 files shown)


opr:
Looks great Rua!

> Should we share / cache state to speed these up?

We can cache npm files with this approach https://github.com/actions/cache/blob/main/examples.md#node---npm - please note that caching `node_modules` is not recommended.

> Can we share state with other actions?

We could add these steps into the E2E Tests workflow, they used to be done that way in Travis, and it cuts down on the npm and composer installs we do. I don't see any reason why these should be separate going forward.

> What platform do we want to run on - debian, ubuntu, macOS etc?

I don't have an opinion on Ubuntu vs. Debian, so maybe someone with better knowledge can weigh in. Whenever I've used GitHub actions in other projects I've always defaulted to Ubuntu. MacOS [costs more](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions) too. I don't think there's any benefit to running the tests on MacOS vs Ubuntu/Debian (since we can run the tests on MacOS easily ourselves).

I believe that if the tests pass or fail on any single platform, there's not much value running them on other platforms. I don't think it proves anything and it wastes our minutes.

> Is the workaround installing build deps for nodegit providing value, or could we address that more elegantly?

I removed this in the JS e2e tests, I can't see any difference between our workflows, so I don't know why yours should be failing without this step 🤔 I will try to investigate today.

**FYI**: I've copied this into the E2E workflow and it looks great, see it in action in #3544

haszari:
@opr I agree with all your feedback/thoughts 🙂 

Closing this in favour of #3544 - I think combining all auto tests into one action is good idea.

PR #3586, opened 2020-12-22
Hide categories without visible products on ProductCategory block (#3564)

### LINKED ISSUE #3564: Hide categories with no visible products
## Describe the bug
Currently, the Product Categories block has an option to hide empty categories. However, this option doesn't work as expected.

To consider a category empty, it only checks whether there are products associated to this category. However, these products might not be visible (hidden or outofstock, if the don't show out of stock products is on). Thus, the category shows on the list (or dropdown), customer clicks on it and then there are products in the archive page, which is very frustrating.

## To reproduce

1. Add the "Product categories list" block to a page. Ensure the "Show empty categories" option is off. (Empty categories should be hidden).
2. Add a new category.
3. Create a new product, add it to the category.
4. Change the product's catalogue visibility to hidden.
5. Visit the page and see that the category is still listed.
6. Try this with out of stock products too when the hide out of stock products option is enabled.

## Expected behavior
The category list should not show any categories that do not contain at least one visible product.

_edited by @opr_

alantygel:
This PR over the ProductCategories block checks each if category has visible products before returning the list of categories, when `$attributes['hasEmpty']` is false. It takes into account the `woocommerce_hide_out_of_stock_items` option. If it is checked, out of stock products are not taken into account.

Fixes #3564

1. Create two Products (P1 and P2) and two Categories (C1 and C2). Set each product in one different category (P1 -> C1, P2 -> C2).
2. Create a page and insert the ProductCategory block.
3. Check that both categories will show on the block.
4. Set P1 as Visibility: hidden and P2 as out of stock
5. If out of stock products are set to be hidden, no category will show; else, only P2 will show.

### Changelog

> Categories without visible products are not shown in Product Categories block

### DIFF
--- a/src/BlockTypes/ProductCategories.php
+++ b/src/BlockTypes/ProductCategories.php
@@ -133,6 +133,40 @@ class ProductCategories extends AbstractDynamicBlock {
 			return [];
 		}
 
+		if ( ! $attributes['hasEmpty'] ) {
+
+			if ( get_option( 'woocommerce_hide_out_of_stock_items' ) === 'yes' ) {
+				$terms = [ 'outofstock', 'exclude-from-catalog' ];
+			} else {
+				$terms = [ 'exclude-from-catalog' ];
+			}
+
+			foreach ( $categories as $i => $category ) {
+				$args = array(
+					'post_type' => 'product',
+					'tax_query' => array( // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_tax_query
+						'relation' => 'AND',
+						array(
+							'taxonomy' => 'product_cat',
+							'field'    => 'term_id',
+							'terms'    => $category->term_id,
+						),
+						array(
+							'taxonomy' => 'product_visibility',
+							'field'    => 'name',
+							'terms'    => $terms,
+							'operator' => 'NOT IN',
+						),
+					),
+				);
+
+				$posts = get_posts( $args );
+				if ( ! $posts ) {
+					unset( $categories[ $i ] );
+				}
+			}
+		}
+
 		return $hierarchical ? $this->build_category_tree( $categories ) : $categories;
 	}
 


Aljullu (commented):
Thanks for noticing the issue and opening a PR to fix it, @alantygel!

While I agree that the Product Categories block should not display categories with hidden/out of stock products, there are some things that make me not to be sure about this solution:
* [As you noted in the issue](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3564#issuecomment-749256732), this seems to add a bit of overhead. Specially for stores with many categories, because for each one of them we need to run a query.
* If the block is set to show product count, out of stock and excluded from catalog products are still counted. In the GIF below, category A has an out of stock product and category B has a product hidden from stock:
Peek 2020-12-28 10-24

AFAIK, WooCommerce Core doesn't provide any way to get a list of product categories where empty categories are filtered out even if they contain out of stock or hidden from catalog products. This is probably the reason why WooCommerce Core _Product Categories List_ widget suffers the same issue described in #3564.

So I think there are two ways to move forward from here:

* Write a custom query for Product Categories block which gathers categories and filters empty ones all in one single query. It should also set `count` to the correct values.
* A much better solution would be to do that but in a function in WooCommerce Core. In the process, we could also fix the _Product Categories List_ widget. Once that function is in WooCommerce Core, we can simply consume it from the block.

alantygel:
Great, @Aljullu, thanks for the careful review.

I fully understand what you mean and agree with you. The ideal fix would somehow filter the `get_term` function to adjust the count parameter on woocommerce. 

I'm not sure if I'm able to do this, but I'll open the issue there.

Aljullu:
Sounds good, I see you already created the issue (https://github.com/woocommerce/woocommerce/issues/28680), thanks @alantygel! I will close this PR then.

PR #3593, opened 2020-12-23
sync shipping with billing when shipping is disabled

### LINKED ISSUE #3592: "Country is required" error when doing a digital checkout.
## Describe the bug
#3552 introduced validation for our API, it seems this validation is running on shipping fields even if we don't have shipping in that order (shipping not set/digital order).

## To reproduce

1. disable shipping or checkout a digital order.
2. try to checkout, you would get a "country is required" error, even if you have country set.

## Expected behavior
The error shouldn't be visible because the country is set and sent.

h/t to @opr for finding this.

senadir:
This PR syncs shipping with billing when shipping is disabled.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3592

1. Disable Shipping and try to checkout.
2. You should not see a validation issue.
3. Your shipping address should be the same as the billing address on the backend.
4. After you already did a successful checkout, try doing another one with shipping disabled.
5. You should see no error, the order should have your billing address as a shipping address.

### DIFF
--- a/assets/js/base/hooks/cart/use-store-cart.js
+++ b/assets/js/base/hooks/cart/use-store-cart.js
@@ -113,7 +113,7 @@ export const useStoreCart = ( options = { shouldSelect: true } ) => {
 			const billingAddress = decodeAddress( cartData.billingAddress );
 			const shippingAddress = cartData.needsShipping
 				? decodeAddress( cartData.shippingAddress )
-				: defaultShippingAddress;
+				: billingAddress;
 			return {
 				cartCoupons: cartData.coupons,
 				cartItems: cartData.items || [],
--- a/assets/js/base/hooks/checkout/use-checkout-address.js
+++ b/assets/js/base/hooks/checkout/use-checkout-address.js
@@ -60,6 +60,20 @@ export const useCheckoutAddress = () => {
 		[ shippingAsBilling, setShippingAddress, setBillingData ]
 	);
 
+	/**
+	 * Sets billing address data, and also shipping if shipping is disabled.
+	 */
+	const setBillingFields = useCallback(
+		( value ) => {
+			setBillingData( value );
+
+			if ( ! needsShipping ) {
+				setShippingAddress( value );
+			}
+		},
+		[ needsShipping, setShippingAddress, setBillingData ]
+	);
+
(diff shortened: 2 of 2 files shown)


opr (changes requested):
✅ Great fix, these cases work:
- [x] Shipping fully disabled (no methods in back end)
- [x] Shipping enabled and using shipping as billing
- [x] Shipping enabled and entering different addresses for shipping and billing

My orders in the back end all have the correct shipping address for these cases.

❌ However this flow does not:
1. Open an incognito window, add a physical product to cart and continue to checkout, use shipping address for billing. Checkout should succeed.
2. Stay logged in, and add a downloadable product to the cart.
3. Go to checkout, notice that billing should already be filled, and you should see no shipping fields.
4. Try to check out
5. See "Country is required" error

opr (approved):
Worked great for me, all the above cases work ok too.

senadir:
@mikejolley This PR also touches on the code you worked on and is probably a regression, take a look at in when you're back.

mikejolley:
@senadir was this to fix what you raised in this comment? https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3552#discussion_r547893742

Another solution may have been to not send the empty address to the API if disabled, which would avoid syncing data. We could probably add test coverage against this?

senadir:
> @senadir was this to fix what you raised in this comment? #3552 (comment)

Yes

> Another solution may have been to not send the empty address to the API if disabled,

Sync would still happen on the backend, I'd prefer we don't sync in the client-side as well.
Wouldn't emitting sending shipping data trigger a validation error as well?

mikejolley:
@senadir When I tested this, if you left the address completely blank, or left it out of the request, the validation/sanitization would not throw errors. We can tweak this if this approach is preferred - I'd prefer sync on the server side since it enforces the correct data shape for orders rather than us doing it manually in client.

senadir:
> I'd prefer sync on the server side since it enforces the correct data shape for orders rather than us doing it manually in client.

It's also one less thing to think about when implementing Store API in other places.

> We can tweak this if this approach is preferred 

Yes, I think this is better.

PR #3597, opened 2020-12-23
Update input colors and alignment

LevinMedia:
As it stands, we've got some issues with the "Standard" inputs when they're render on a dark background. Both the checkboxes and radio buttons use a dark grey color, making them difficult to see. The standard inputs should work on any color background. This PR updates their styles to match the specs noted in pbIy4N-dY-p2 

This PR also updates some alignment issues in the shipping, payment method, and order notes sections as well.

**Currently in Trunk:** 

screencapture-localhost-checkout-2020-12-23-14_35_09

**Updates in this PR:** 

screencapture-localhost-checkout-2020-12-23-14_33_23

Likewise, this PR also updates the styles of the dark mode checkboxes and radio buttons slightly as well.

**Currently in Trunk:**

screencapture-localhost-checkout-2020-12-23-14_36_33

**Updates in this PR:**

screencapture-localhost-checkout-2020-12-23-14_32_39

Lastly, this PR updates the dropdown menu text color in the standard inputs from white to black, making it readable. 

**Before:**

Screen Shot 2020-12-23 at 2 56 40 PM

**After:**

Screen Shot 2020-12-23 at 2 56 07 PM

### To be addressed in a follow up issue

The credit card fields in the payment method section don't respect the darkmode toggle.


budzanowski:
@LevinMedia would you mind fixing the CSS Linter findings? Ping if you need any help, or I can do it for you if you don't want to dabble in our tests :P

assets/js/base/components/radio-control/style.scss
```diff
@@ -2,14 +2,13 @@
 	@include reset-typography();
 	@include with-translucent-border( 0 0 1px );
 	display: block;
-	padding: $gap-small $gap-small 0 #{$gap-larger * 2};
+	padding: $gap 0 $gap #{$gap-larger + 4};
```
Aljullu: If I'm not wrong, this should be equivalent:
```suggestion
	padding: $gap 0 $gap $gap-largest;
```

assets/js/blocks/cart-checkout/cart/full-cart/style.scss
```diff
@@ -251,7 +251,7 @@ table.wc-block-cart-items {
 		}
 	}
 	.wc-block-components-radio-control__option {
-		padding-left: $gap-large;
+		padding-left: 40px;
```
Aljullu: I think we can remove this ruleset completely now that the [default padding](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3597/files#diff-29bc3ff26316a7c413ce211149319161bc52bf9d4c0d36afba3725ed13a25ec0R5) is the same:

```diff
-	.wc-block-components-radio-control__option {
-		padding-left: 40px;
-	}
```

assets/js/base/components/select/style.scss
```diff
@@ -90,6 +90,7 @@
 		margin: 0;
 		max-height: 300px;
 		overflow: auto;
+		color: #000;
```
Aljullu: I wonder if this color should be `$input-text-active`. Otherwise, the text color of options look a bit darker than the selected value:

imatge

assets/js/base/components/radio-control/style.scss
```diff
@@ -51,28 +49,28 @@
 .wc-block-components-radio-control {
 	.wc-block-components-radio-control__input {
 		appearance: none;
-		background: transparent;
+		background: #fff;
 		border: 2px solid currentColor;
 		border-radius: 50%;
 		display: inline-block;
 		height: 1.25rem;
```
Aljullu: In this file, every time we update the `min-height`/`min-width`, we should update the `height` and `width` values as well. In this case, they should be in relative units so they adapt to the themes or any a11y preference from the user.

I think it should be safe using the `em()` function here. So in this selector, we would need to make these changes:
```diff
.wc-block-components-radio-control__input {
	...
-	height: 1.25rem;
- 	width: 1.25rem;
+	height: em(24px);
+	width: em(24px);
}
```

Same for the other sizes which are modified in this file.

assets/js/base/components/checkbox-control/style.scss
```diff
@@ -7,21 +7,21 @@
 
 	.wc-block-components-checkbox__input[type="checkbox"] {
 		appearance: none;
-		border: 2px solid currentColor;
+		border: 2px solid #50575e;
```
Aljullu: Text inputs use a `#8d96a0` color for the border. Do you think we could use one or the other for both components or is it intended to use different colors for them?

Aljullu (commented):
Nice improvement for themes with dark background, also thanks for fixing the issue with the 'invisible' options text in selects. :ok_hand: 

I left some code suggestions and comments below. Let us know if you want to take care of them yourself or if you want to hand it over to us, I'm happy either way.

Aljullu:
I updated the PR with the changes I suggested yesterday. @budzanowski would you like to review it?

budzanowski (approved):

PR #3598, opened 2020-12-24
Fix nonce issues when adding product to cart from All Products

### LINKED ISSUE #3595: X-WC-Store-API-Nonce is invalid errors because of inconsistent nonce
## Describe the bug
When you try to add a product after a checkout, you get a nonce error.

## To reproduce

1. In a private window, add and checkout, without creating an account.
2. Go back to all products, try adding a product
3. you get a nonce error, this seems to happen with all products only, works fine in cart requests.

You can also reproduce it by:
1. In a private window, visit a page with All Products.
2. Try adding to cart, it would work.
3. Try filtering or switching the page, and then adding the cart, it wouldn't work.

this issue is mostly because of inconsistent nonce between products and cart due to session.
Sometimes it works, sometimes it doesn't.

## Expected behavior
You can add a product with no problem.

senadir:
products endpoint had a different session than cart and checkout, this caused it to return a different nonce from them, causing the Add to cart functionality in all products to not work, ever.
This PR does two things:
- It fixes the different nonce issue by loading cart in Products route, it also extends `AbstractCartRoute` for other cart subroutes that we forget.
- It refreshes the nonce even if an add to cart fails.

Fixes #3595

1. In a private window, go to All Products, try to add to cart, it should work
2. Do it several times with several products.
3. Paginate the block and try to add products again.
4. If you have filters set up, try using filters and then adding to cart.

### DIFF
--- a/assets/js/data/shared-controls.js
+++ b/assets/js/data/shared-controls.js
@@ -55,6 +55,7 @@ export const controls = {
 						} );
 				} )
 				.catch( ( errorResponse ) => {
+					triggerFetch.setNonce( errorResponse.headers );
 					if ( typeof errorResponse.json === 'function' ) {
 						// Parse error response before rejecting it.
 						errorResponse
--- a/src/StoreApi/Routes/CartCoupons.php
+++ b/src/StoreApi/Routes/CartCoupons.php
@@ -8,7 +8,7 @@ use Automattic\WooCommerce\Blocks\StoreApi\Utilities\CartController;
  *
  * @internal This API is used internally by Blocks--it is still in flux and may be subject to revisions.
  */
-class CartCoupons extends AbstractRoute {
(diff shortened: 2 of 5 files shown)


assets/js/data/shared-controls.js
```diff
@@ -55,6 +55,7 @@ export const controls = {
 						} );
 				} )
 				.catch( ( errorResponse ) => {
+					triggerFetch.setNonce( errorResponse.headers );
```
senadir, line 58: We already do this in Checkout processor, but this is not inside that context, so it doesn't happen.

opr (approved):
Worked well for me, using filter products by attribute too.

senadir:
@mikejolley When you're back, do take a look at this and if's the optimal solution.

PR #3599, opened 2020-12-24
fix: shipping_address/billing_address use properties instead on items

MohammedFaragallah:
Thank you for your hard

I think the title says it all


senadir (approved):
Great catch, Thanks @MohammedFaragallah

PR #3602, opened 2020-12-28
Add protection around mock queries in unit tests.

vedanshujain:
This adds additional checks before mocking wc_get_orders query results to prevent overriding refunds query. This is needed with WC core 4.9 since it does an addtional wc_get_orders query to preload and hydrate cache for refunds in order to improve performance. This second call was also getting mocked, causing refunds array to not return empty when it should have been.

Fixes #https://github.com/woocommerce/woocommerce/issues/28668

1. Run phpunit tests with WC 4.9 beta.1, you can do so by  a) downloading plugin zip and replacing your /tmp/wordpress/wp-content/plugins/woocommerce dir, or if you use docker, then by modifying https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/bin/install-wp-tests.sh#L153 to use `latest.zip` instead of `latest-stable.zip`.
OR 
2. Wait for Travis tests to pass.

### Changelog

> NA

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
 	"minimum-stability": "dev",
 	"require": {
 		"composer/installers": "^1.7.0",
-		"automattic/jetpack-autoloader": "^2.0.0"
+		"automattic/jetpack-autoloader": "2.7.1"
 	},
 	"require-dev": {
 		"phpunit/phpunit": "6.5.14",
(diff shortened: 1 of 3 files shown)


senadir:
Rebased with trunk. Good to merge if tests pass.

Aljullu (approved):
Confirming it fixes the tests that were throwing and code changes look good.

There are still two errors which are failing, but those are unrelated to this PR, so merging. Thanks @vedanshujain!

Issue #3603, opened 2020-12-28
Tasks to do when updating `@woocommerce/components`


Aljullu:
Recently, I pushed a couple of PRs that change the styling of `SearchListControl` and `SearchListItem` (https://github.com/woocommerce/woocommerce-admin/pull/5809 and https://github.com/woocommerce/woocommerce-admin/pull/5901). Some of it was in preparation for the widgets editor (#3228) and to fix some styling issues that we got in recent versions of WP. This issue should be used as a list of the tasks that we will need to do when updating `@woocommerce/components`:

* The operator select that is usually rendered after the list, will no longer need to have a margin-top, since that margin will be included in the list that we get from upstream (previously, it was removed in [this line](https://github.com/woocommerce/woocommerce-admin/pull/5901/files#diff-6674882d8a24592c41ba22ae733ede51a8756959648b7b7266ac3ee62e7cc318L207)). In other words, it should be safe to remove these lines (notice those selectors weren't working any way in recent versions of WP):
  https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0cffa3e2eca9fa757597a504085e352acb9e0ecf/assets/js/editor-components/product-attribute-term-control/style.scss#L14
  https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0cffa3e2eca9fa757597a504085e352acb9e0ecf/assets/js/editor-components/product-category-control/style.scss#L14
  https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0cffa3e2eca9fa757597a504085e352acb9e0ecf/assets/js/editor-components/product-tag-control/style.scss#L14
* We will need to add a new prop (`isCompact={ true }`) to all `SearchListControl` components that are rendered in the sidebar.
* It should be safe to remove this border since that was moved upstream (https://github.com/woocommerce/woocommerce-admin/pull/5901#pullrequestreview-554794141):
  https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0cffa3e2eca9fa757597a504085e352acb9e0ecf/assets/css/editor.scss#L51-L54
* Since we can no longer rely on the `.components-panel` class, we will need to rewrite the selector for `.components-select-control__input` from this three files
  https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0cffa3e2eca9fa757597a504085e352acb9e0ecf/assets/js/editor-components/product-attribute-term-control/style.scss#L16
[...]

PR #3604, opened 2020-12-28
SearchListControl styles cleanup

Aljullu:
Small PR that removes a very old CSS selector which was no longer necessary and adds a comment to another selector which we will be able to remove when updating `@woocommerce/components`. Notice I didn't prefix it with `// @todo` because I included it in a meta issue about all the tasks we will need to do when updating `@woocommerce/components` (#3603).

### Screenshots

imatge

imatge

Note there is an issue in the second screenshot: `Display products matching` should have top margin. It's not there because [this selector](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0cffa3e2eca9fa757597a504085e352acb9e0ecf/assets/js/editor-components/product-attribute-term-control/style.scss#L13) is not working with recent WP versions, instead of fixing it downstream, I [upstreamed the fix](https://github.com/woocommerce/woocommerce-admin/pull/5901/files#r545244002), but until we update `@woocommerce/components`, it will not be fixed in WC Blocks. In parallel, we can't remove the broken selector yet, because it _does_ work in WP 5.4.

1. Add a Products by Attribute block in a page or post.
2. Select some attributes and verify there are no visual regressions.
3. Extra points if you test it in WP 5.4 + WC 4.3 (our oldest supported versions). :slightly_smiling_face:

### DIFF
--- a/assets/css/editor.scss
+++ b/assets/css/editor.scss
@@ -48,8 +48,10 @@
 	}
 }
 
-// Adds border to placeholder lists, so they can be distinguished from the white background.
-.components-placeholder .woocommerce-search-list__list {
+// This border has been upstreamed to WooCommerce Admin, so it can be removed
+// from here once @woocommerce/components gets updated.
+// https://github.com/woocommerce/woocommerce-admin/pull/5809
+.woocommerce-search-list__list {
 	border: 1px solid $gray-200;
 }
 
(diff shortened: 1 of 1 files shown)


budzanowski (approved):
It works very nice. No issues found.

budzanowski:
Travis issue :/ WP 5.4 again.

PR #3627, opened 2021-01-04
Convert saved payment methods token to a string

### LINKED ISSUE #3618: Saved payment validation returns error
## Describe the bug
When using saved payments on the checkout block, the checkout process errors out visibly to front end users. The error is:
payment_data[1][value] is not of type string,boolean.

## To reproduce
1. In my account, add a payment method, mark it as default or the only one.
2. Add any product type to the cart
3. Checkout regularly using the saved card as payment type, don't change the saved card.
4. Hit complete checkout
5. See error above checkout

## Expected behavior
Checkout works.

## Screenshots

blocks-validation-error

* Core version: 5.5.3
* WooCommerce version: 4.7.0
* WooCommerce Blocks version: 4.1.0
* Site language: English
* Any other plugins installed: WooCommerce Stripe Gateway 4.5.5

## Additional context
Seems to have been introduced with validation in 4.1.0

Aljullu:
Fixes #3618.

Even though the issue has been there forever, it wasn't reproducible until validation was added to the Checkout route (#3552), so #3618 can be considered a regression from 4.0 → 4.1.

The issue is caused because when making a payment with a Stripe saved credit card, the value of `wc-stripe-payment-token` was set to an integer, but we are [only accepting strings and bools](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/StoreApi/Routes/Checkout.php#L92). This PR forces saved payment method tokens to be converted to strings.

imatge

I wasn't able to reproduce #3618 in a local install, but I could reproduce it in an ephemeral site.

1. Make sure you have WC Blocks >=4.1.0 installed and WooCommerce Stripe Payment Gateway enabled.
2. With a new user, go to My Account > Payment Methods and add a new payment method.
3. Add a product to your cart and go to the checkout page.
4. Fill the details but keep the payment method as the saved card you added in step 2.
5. Try to make the purchase and verify the error from #3618 doesn't appear anymore.

### Changelog

> Fix an error that was blocking checkout with some user saved payment methods.

### DIFF
--- a/assets/js/base/components/payment-methods/saved-payment-method-options.js
+++ b/assets/js/base/components/payment-methods/saved-payment-method-options.js
@@ -47,7 +47,7 @@ const getCcOrEcheckPaymentMethodOption = (
 			setActivePaymentMethod( method.gateway );
 			setPaymentStatus().success( {
 				payment_method: method.gateway,
-				[ savedTokenKey ]: token,
+				[ savedTokenKey ]: token + '',
 				isSavedToken: true,
 			} );
 		},
@@ -81,7 +81,7 @@ const getDefaultPaymentMethodOptions = (
 			setActivePaymentMethod( method.gateway );
 			setPaymentStatus().success( {
 				payment_method: method.gateway,
-				[ savedTokenKey ]: token,
+				[ savedTokenKey ]: token + '',
 				isSavedToken: true,
 			} );
 		},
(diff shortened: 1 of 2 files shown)


senadir:
I'm not sure if this is the best solution, https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2450 started forcing the token to be string because it caused a lot of inconsistencies in our checks so I would prefer we default tokens to be strings everywhere than to accept integers.

Aljullu:
I wasn't aware about the changes in #2450, thanks @senadir! I updated this PR so we force the `token` to be a string. It should be ready for another look.

budzanowski (approved):

opr (approved):

PR #3634, opened 2021-01-04
Add "you don't need lodash/underscore" eslint rule

nerrad:
To review existing usage of lodash and prevent against unnecessary future usage, I've added the handy ["you don't need lodash/underscore" eslint plugin](https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore/#_uniq) to the repo. Besides setting up the plugin, this pull also:

- removes unnecessary uses of `omit` 
- ignores lodash usage in developer related files (storybook stories, webpack config as example), currently the config is just excluding specific rules - we can add others as needed.
- convert usage of `_.assign`, `_.isNil` and `_.isNan`. 
- adds rule ignore comments for files where lodash is imported anyways (in the case of `_.flatten` where I don't want to increase the testing load for this pull with a conversion to something like `Array.flatMap` just yet).

## Why?

ES6/Next has removed the need for a lot of what lodash brought and implementing this helps us to avoid unnecessary syntax usage where a native implementation exists. This will go a long way towards reducing the size of our builds with the potential _removal_ of a lodash dependency all-together eventually.

With this eslint plugin, as we right code, we will be warned right away if we use something from lodash that has a native equivalent helping reduce our dependency and tech debt tied to its usage.

Since we still support IE however, we will have to watch whether bundle size is adversely affected because there is the chance that the compiled version support built by webpack could actually end up greater than just using the lodash equivalent (especially if lodash is already loaded). 

## Testing

The following behaviour was impacted by this change (due to lodash replacement with native equivalent):

- `clearValidationError` from the Validation context provider in the checkout. To test, just ensure that when validation errors are corrected in the form that the error correctly clears.
- `useSynchronizedQueryState` hook. To test make sure that pagination for the all products block works as expected and there are no increases in network requests in this branch.
- `GridLayoutControl`. This component is used by all grid blocks that allow for changing the number of rows or columns. To test, make sure there are no errors while modifying columns or rows using this control and that the correct values persist.
- `withTransformSingleSelectToMultipleSelect` used by `ProductControl`. To test, ensure that the product selection control (found in the initial form on product grid blocks) allows for successful selection of products with no issues (or console errors).

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -25,4 +25,12 @@ module.exports = {
 		'react-hooks/exhaustive-deps': 'error',
 		'react/jsx-fragments': [ 'error', 'syntax' ],
 	},
+	overrides: [
+		{
+			files: [ '**/bin/**.js', '**/storybook/**.js', '**/stories/**.js' ],
+			rules: {
+				'you-dont-need-lodash-underscore/omit': 'off',
+			},
+		},
+	],
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,8 @@
 module.exports = {
-	extends: [ 'plugin:@woocommerce/eslint-plugin/recommended' ],
+	extends: [
+		'plugin:@woocommerce/eslint-plugin/recommended',
+		'plugin:you-dont-need-lodash-underscore/compatible',
+	],
 	globals: {
 		wcStoreApiNonce: 'readonly',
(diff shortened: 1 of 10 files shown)


opr (approved):
This is a great move Darren, I really like the idea of moving away from dependencies if the functionality is available natively.

The code looks great too, all the areas you've modified are working well with no errors in my testing. 👌🏻

PR #3653, opened 2021-01-06
Add major for devDependencies.

### LINKED ISSUE #2811: Configure renovate to update major versions on devDependencies
I've noticed that a number of our devDependencies are _significantly_ behind in updates because currently renovate seems to be configured to only update within the minor release range. Ideally devDependencies would not be pinned either since they are for development, they should be updated more frequently.

budzanowski:
Fixes #2811 

Add major versions update and auto-marege for devDependencies.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -15,7 +15,12 @@
 		{
 			"updateTypes": [ "minor", "patch", "pin", "digest" ],
 			"automerge": true
-		}
+		},
+		{
+			"updateTypes": [ "major", "minor", "patch", "pin", "digest" ],
+			"depTypeList": [ "devDependencies" ],
+			"automerge": true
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -15,7 +15,12 @@
 		{
 			"updateTypes": [ "minor", "patch", "pin", "digest" ],
 			"automerge": true
-		}
+		},
+		{
+			"updateTypes": [ "major", "minor", "patch", "pin", "digest" ],
+			"depTypeList": [ "devDependencies" ],
+			"automerge": true
+		  }
```
mikejolley:
```suggestion
		}
```

mikejolley (approved):

PR #3654, opened 2021-01-06
Create @woocommerce/checkout to export components to a global

### LINKED ISSUE #3613: Integration Tracking (WC Subscriptions): Export components to a global checkout
In order to get #3246 out, we need to a way to export components globally so that they're consumable by other plugins.
Getting an npm package is not a priority right now so getting them to the global `wc` variable should be enough.
The new package should be named checkout and is accessed globally from `wc.checkout`.
When implemented, a new rule should be added to [@woocommerce/dependency-extraction-webpack-plugin](https://github.com/woocommerce/woocommerce-admin/tree/main/packages/dependency-extraction-webpack-plugin) to alias `@woocommerce/checkout` to `wc.checkout`.
I anticipate that wc.checkout would conflict with our block slug, so we should probably name the slug something else when enqueueing the file.
This branch `try/checkout-package` contains an exploration I did some time ago and you can probably build on it or start fresh, up to you.

The new package would contain slotFills and probably some components, so it should be able to build SCSS as well.
We would gradually move components from `assets/js/base/components/cart-checkout` to this package when it makes sense.

Aljullu:
Fixes #3613.

This PR:
* Creates a `@woocommerce/blocks-checkout` alias that directs to the `/packages/` directory.
  * Everything in that package is available for external scripts via the `wc.wcBlocksCheckout` global.
  * It should be used to make some components and utils available to extensions.
* Moves the `TotalsItem` component to that package. This is the only component that has been moved for now as a proof-of-concept.
* In the branch history there is an example of usage (2f50349692f53c5bb143516365eee7a98d580dfd). I removed it since that's not something we want to ship, but it can be used as a reference for follow-up PRs.

cc @senadir who created the branch and started the work in this area.

1. Revert the commit that removes the example code: `git revert 808ef814ce88c2cbda8384700ab779c2dcc6f06f`.
2. Go to the Cart and Checkout and verify there is a `My extension` item in the sidebar:
imatge
3. In your browser console, print the values of `wc.wcBlocksCheckout` and verify it includes `TotalsItem` and `SlotFillProvider`.


bin/webpack-configs.js
```diff
@@ -86,6 +95,12 @@ const getCoreConfig = ( options = {} ) => {
 						},
 					},
 				},
+				{
+					test: /\.s[c|a]ss$/,
+					use: {
+						loader: 'ignore-loader',
+					},
+				},
```
senadir, line 103: does this mean any style loaded by this package is going to be ignored?

senadir (approved):
Looks good, merge ahead.

bin/webpack-configs.js
senadir: I see we're catching them again below with glob.sync.

Aljullu: Yes, styling is handled in a different process. There is some background in #2818.

senadir:
Only change I would do (in a follow up) is to rename `wc.wcBlocksCheckout` to `wc.blocksCheckout` no need for redundant wc

PR #3656, opened 2021-01-06
Improve error displayed to customers when an item's stock status changes during checkout.

### LINKED ISSUE #3412: Checkout block: pending order when product is out of stock, generic error message

opr:
### Description
This PR adds some new exceptions that can be thrown when validating the cart contents. These exceptions all relate to different types of "out of stock"

##### Thrown in `validate_cart_item`
- `PartialOutOfStockException` - Thrown when there _are_ items in stock, but not enough to satisfy the customer's order. The quantity in cart will be reduced to the in-stock amount.
- `TooManyInCartException` - Thrown when products can only be purchased individually but there is more than one in the cart. The quantity in cart will be reduced to 1.
- `OutOfStockException` - Thrown when the product is completely out of stock. The item is removed from cart.
- `NotPurchasableException` - Thrown when the product is not able to be purchased (doesn't exist, doesn't have a price etc.)
##### Thrown in `validate_cart_items`
- `StockValidationException` - Thrown as a "general" exception from `validate_cart_items` when one/some/all of the above exceptions are thrown in `validate_cart_item`. This Exception contains a `WP_Error` object that in turn contains all of the errors thrown by each execution of`validate_cart_item`. This `WP_Error` is then converted into a response and sent to the client so that the individual error notices might be displayed.

Fixes #3412
This supersedes #3469 

### Before
Screenshot 2021-01-06 at 16 06 36

### After
Screenshot 2021-01-06 at 14 45 04

1. **In an incognito window** Add several products to your cart. Add these in differing quantities. Several of these added products need to have a quantity of 3+ in the cart.
2. Open the checkout page and ensure all the products are showing. **Do different combinations of the following steps in a different session to the one where you've added the items, and also without refreshing the checkout page. This is necessary because we can't be logged in as admin when testing step 3 below.**
3. Delete one or more of the products from the back end.
4. Set one or more of the products to "Manage stock?" true, and edit the in-stock amount of one or more of the products ensuring it is below the quantity that is in the cart, but above 0.
5. Set one or more of the products to "Sold individually"
6. Set one or more of the products to be entirely out of stock.
7. Try to check out and ensure you see the correct error messages for the products whose stock you've changed/reduced.
8. Try a few different scenarios with stock levels, number of products affected, number of products in cart etc. 

### Changelog

> Improve error displayed to customers when an item's stock status changes during checkout.

### DIFF
--- a/src/StoreApi/Routes/AbstractRoute.php
+++ b/src/StoreApi/Routes/AbstractRoute.php
@@ -2,6 +2,7 @@
 namespace Automattic\WooCommerce\Blocks\StoreApi\Routes;
 
 use Automattic\WooCommerce\Blocks\StoreApi\Schemas\AbstractSchema;
+use Automattic\WooCommerce\Blocks\StoreApi\Utilities\StockAvailabilityException;
 
 /**
  * AbstractRoute class.
@@ -72,6 +73,8 @@ abstract class AbstractRoute implements RouteInterface {
 			}
 		} catch ( RouteException $error ) {
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode(), $error->getAdditionalData() );
+		} catch ( StockAvailabilityException $error ) {
+			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getError(), $error->getCode(), $error->getAdditionalData() );
 		} catch ( \Exception $error ) {
 			$response = $this->get_route_error_response( 'unknown_server_error', $error->getMessage(), 500 );
 		}
--- a/src/StoreApi/Routes/AbstractRoute.php
+++ b/src/StoreApi/Routes/AbstractRoute.php
@@ -3,6 +3,7 @@ namespace Automattic\WooCommerce\Blocks\StoreApi\Routes;
 
 use Automattic\WooCommerce\Blocks\StoreApi\Schemas\AbstractSchema;
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\StockAvailabilityException;
+use WP_Error;
 
 /**
  * AbstractRoute class.
@@ -74,7 +75,7 @@ abstract class AbstractRoute implements RouteInterface {
 		} catch ( RouteException $error ) {
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode(), $error->getAdditionalData() );
 		} catch ( StockAvailabilityException $error ) {
-			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getError(), $error->getCode(), $error->getAdditionalData() );
+			$response = $this->get_route_error_response_from_object( $error->getError(), $error->getCode(), $error->getAdditionalData() );
 		} catch ( \Exception $error ) {
 			$response = $this->get_route_error_response( 'unknown_server_error', $error->getMessage(), 500 );
 		}
@@ -205,6 +206,19 @@ abstract class AbstractRoute implements RouteInterface {
 	 */
 	protected function get_route_error_response( $error_code, $error_message, $http_status_code = 500, $additional_data = [] ) {
 		return new \WP_Error( $error_code, $error_message, array_merge( $additional_data, [ 'status' => $http_status_code ] ) );
+	}
+
+	/**
+	 * Get route response when something went wrong and the supplied error is a WP_Error. This happens when
+	 *
+	 * @param WP_Error $error_object The WP_Error object containing the error.
+	 * @param int      $http_status_code HTTP status. Defaults to 500.
+	 * @param array    $additional_data  Extra data (key value pairs) to expose in the error response.
+	 * @return WP_Error WP Error object.
+	 */
+	protected function get_route_error_response_from_object( $error_object, $http_status_code = 500, $additional_data = [] ) {
+		$error_object->add_data( array_merge( $additional_data, [ 'status' => $http_status_code ] ) );
+		return $error_object;
 	}
 
 	/**
--- a/src/StoreApi/Routes/AbstractRoute.php
+++ b/src/StoreApi/Routes/AbstractRoute.php
@@ -2,7 +2,7 @@
 namespace Automattic\WooCommerce\Blocks\StoreApi\Routes;
 
 use Automattic\WooCommerce\Blocks\StoreApi\Schemas\AbstractSchema;
-use Automattic\WooCommerce\Blocks\StoreApi\Utilities\StockAvailabilityException;
+use Automattic\WooCommerce\Blocks\StoreApi\Utilities\InvalidStockLevelsInCartException;
 use WP_Error;
 
 /**
@@ -74,7 +74,7 @@ abstract class AbstractRoute implements RouteInterface {
 			}
 		} catch ( RouteException $error ) {
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode(), $error->getAdditionalData() );
-		} catch ( StockAvailabilityException $error ) {
+		} catch ( InvalidStockLevelsInCartException $error ) {
 			$response = $this->get_route_error_response_from_object( $error->getError(), $error->getCode(), $error->getAdditionalData() );
 		} catch ( \Exception $error ) {
 			$response = $this->get_route_error_response( 'unknown_server_error', $error->getMessage(), 500 );
--- a/src/StoreApi/Routes/AbstractRoute.php
+++ b/src/StoreApi/Routes/AbstractRoute.php
@@ -209,7 +209,9 @@ abstract class AbstractRoute implements RouteInterface {
 	}
 
 	/**
-	 * Get route response when something went wrong and the supplied error is a WP_Error. This happens when
+	 * Get route response when something went wrong and the supplied error is a WP_Error. This currently only happens
+	 * when an item in the cart is out of stock, partially out of stock, can only be bought individually, or when the
+	 * item is not purchasable.
 	 *
 	 * @param WP_Error $error_object The WP_Error object containing the error.
 	 * @param int      $http_status_code HTTP status. Defaults to 500.
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -255,59 +255,58 @@ class Checkout extends AbstractRoute {
 	/**
 	 * Get route response when something went wrong.
 	 *
-	 * @param string          $error_code String based error code.
-	 * @param string|WP_Error $error_message User facing error message.
-	 * @param int             $http_status_code HTTP status. Defaults to 500.
-	 * @param array           $additional_data  Extra data (key value pairs) to expose in the error response.
+	 * @param string $error_code String based error code.
+	 * @param string $error_message User facing error message.
+	 * @param int    $http_status_code HTTP status. Defaults to 500.
+	 * @param array  $additional_data  Extra data (key value pairs) to expose in the error response.
 	 * @return WP_Error WP Error object.
 	 */
 	protected function get_route_error_response( $error_code, $error_message, $http_status_code = 500, $additional_data = [] ) {
+		$error_from_message = new WP_Error(
+			$error_code,
+			$error_message
+		);
 		switch ( $http_status_code ) {
-			case 400:
-				if ( is_wp_error( $error_message ) ) {
-					return $error_message;
-				}
-				return new WP_Error(
-					$error_code,
-					$error_message,
-					array_merge(
-						$additional_data,
-						[
-							'status' => $http_status_code,
-						]
-					)
-				);
 			case 409:
-				// If there was a conflict, return the cart so the client can resolve it.
-				$controller = new CartController();
-				$cart       = $controller->get_cart_instance();
-
-				if ( is_wp_error( $error_message ) ) {
-					$error_message->add_data(
-						array_merge(
-							$error_message->error_data,
-							[
-								'status' => $http_status_code,
-								'cart'   => wc()->api->get_endpoint_data( '/wc/store/cart' ),
-							]
-						)
-					);
-					return $error_message;
-				}
-
-				return new WP_Error(
-					$error_code,
-					$error_message,
-					array_merge(
-						$additional_data,
-						[
-							'status' => $http_status_code,
-							'cart'   => wc()->api->get_endpoint_data( '/wc/store/cart' ),
-						]
-					)
-				);
+				// 409 is when there was a conflict, so we return the cart so the client can resolve it.
+				return $this->add_data_to_error_object( $error_from_message, $additional_data, $http_status_code, true );
 		}
-		return new WP_Error( $error_code, $error_message, [ 'status' => $http_status_code ] );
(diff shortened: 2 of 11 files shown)


src/StoreApi/Routes/AbstractRoute.php
```diff
@@ -72,6 +73,8 @@ public function get_response( \WP_REST_Request $request ) {
 			}
 		} catch ( RouteException $error ) {
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode(), $error->getAdditionalData() );
+		} catch ( StockAvailabilityException $error ) {
+			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getError(), $error->getCode(), $error->getAdditionalData() );
```
mikejolley: We may need to replace or refactor `get_route_error_response` since it's inherited from the Abstract and we're changing the params here (string to WP_Error Object).

Perhaps a new method is needed? `get_route_error_response_from_object`?

src/StoreApi/Routes/Checkout.php
```diff
@@ -277,6 +282,19 @@ protected function get_route_error_response( $error_code, $error_message, $http_
 				$controller = new CartController();
 				$cart       = $controller->get_cart_instance();
 
+				if ( is_wp_error( $error_message ) ) {
```
mikejolley: If we go with the `get_route_error_response_from_object ` suggestion in another comment, we should abstract this chunk of code out and use it in both; appending the cart data should be shared between both.

src/StoreApi/Routes/Checkout.php
```diff
@@ -253,15 +255,18 @@ protected function get_route_post_response( WP_REST_Request $request ) {
 	/**
 	 * Get route response when something went wrong.
 	 *
-	 * @param string $error_code String based error code.
-	 * @param string $error_message User facing error message.
-	 * @param int    $http_status_code HTTP status. Defaults to 500.
-	 * @param array  $additional_data  Extra data (key value pairs) to expose in the error response.
+	 * @param string          $error_code String based error code.
+	 * @param string|WP_Error $error_message User facing error message.
```
mikejolley: I think we should only handle strings in this method to avoid complexity.

src/StoreApi/Utilities/CartController.php
```diff
@@ -3,7 +3,9 @@
 
 use Automattic\WooCommerce\Blocks\StoreApi\Routes\RouteException;
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\NoticeHandler;
+use Automattic\WooCommerce\Blocks\Utils\ArrayUtils;
 use Automattic\WooCommerce\Checkout\Helpers\ReserveStock;
+use WP_Error;
```
mikejolley, line 8: Looks like all the backslashes in the rest of the codebase are unnecessary 🤦🏻 👍🏻

src/StoreApi/Utilities/CartController.php
```diff
@@ -344,6 +463,12 @@ public function get_cart_item_errors() {
 				$this->validate_cart_item( $cart_item );
 			} catch ( RouteException $error ) {
 				$errors[] = new \WP_Error( $error->getErrorCode(), $error->getMessage() );
+			} catch ( NotPurchasableException $error ) {
+				$errors[] = new \WP_Error( $error->getErrorCode(), $error->getMessage() );
```
mikejolley: `\WP_Error` can just be `WP_Error` because its imported with the use operator.

src/StoreApi/Utilities/NotPurchasableException.php
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @param string $error_code       Machine-readable error code, e.g `woocommerce_invalid_product_id`.
+	 * @param string $product_name     The name of the product that can only be purchased individually.
+	 * @param array  $additional_data  Extra data (key value pairs) to expose in the error response.
+	 */
+	public function __construct( $error_code, $product_name, $additional_data = [] ) {
+		$this->error_code      = $error_code;
+		$this->product_name    = $product_name;
+		$this->additional_data = array_filter( (array) $additional_data );
+		parent::__construct();
```
mikejolley: Should we pass params to the parent constructor here?

src/StoreApi/Utilities/OutOfStockException.php
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @param string $error_code       Machine-readable error code, e.g `woocommerce_invalid_product_id`.
+	 * @param string $product_name     The name of the product that is out of stock.
+	 * @param array  $additional_data  Extra data (key value pairs) to expose in the error response.
+	 */
+	public function __construct( $error_code, $product_name, $additional_data = [] ) {
+		$this->error_code      = $error_code;
+		$this->product_name    = $product_name;
+		$this->additional_data = array_filter( (array) $additional_data );
+		parent::__construct();
```
mikejolley: Should we pass params to the parent constructor here?

src/StoreApi/Utilities/PartialOutOfStockException.php
```diff
@@ -0,0 +1,73 @@
...
+	 *
+	 * @param string $error_code       Machine-readable error code, e.g `woocommerce_invalid_product_id`.
+	 * @param string $product_name     The name of the product that is partially out of stock.
+	 * @param array  $additional_data  Extra data (key value pairs) to expose in the error response.
+	 */
+	public function __construct( $error_code, $product_name, $additional_data = [] ) {
+		$this->error_code      = $error_code;
+		$this->product_name    = $product_name;
+		$this->additional_data = array_filter( (array) $additional_data );
+		parent::__construct();
```
mikejolley: Should we pass params to the parent constructor here?

src/StoreApi/Utilities/NotPurchasableException.php
```diff
@@ -0,0 +1,73 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\StoreApi\Utilities;
+
+/**
+ * NotPurchasableException class.
+ *
+ * @internal This API is used internally by Blocks, this exception is thrown when an item in the cart is not able to be
+ * purchased.
+ */
+class NotPurchasableException extends \Exception {
```
mikejolley: All of these new Exception classes appear to be almost identical—perhaps we should have that code in one class and extend it? It may make catching the exceptions more straightforward too (if you can catch the parent class, would have to test to ensure this works).

src/StoreApi/Utilities/CartController.php
```diff
@@ -288,20 +416,11 @@ public function validate_cart_item( $cart_item ) {
...
-							'There are only %1$s units of &quot;%2$s&quot; in stock.',
-							$qty_remaining,
-							'woo-gutenberg-products-block'
-						),
-						wc_format_stock_quantity_for_display( $qty_remaining, $product ),
-						$product->get_name()
-					),
-					400
+
+				WC()->cart->set_quantity( $cart_item['key'], $qty_remaining, false );
```
mikejolley, line 405: Is this desirable every time the validate function runs on all requests, or should it be it's own method/optional? 

I'm not saying this is wrong, I'm just wondering if this is something we definitely want to do in all use cases or if we want to enable this somehow. I guess that might impact the error notices though..feel free to ignore.

src/StoreApi/Utilities/CartController.php
```diff
@@ -213,17 +215,144 @@ public function validate_add_to_cart( \WC_Product $product, $request ) {
...
 	}
 
+	// phpcs:disable Squiz.Commenting.FunctionCommentThrowTag.WrongNumber -- this method _does_ throw a RouteException
+	// in convert_notices_to_exceptions.
 	/**
 	 * Validate all items in the cart and check for errors.
 	 *
-	 * @throws RouteException Exception if invalid data is detected.
+	 * @throws StockAvailabilityException Exception if invalid data is detected due to insufficient stock levels.
+	 * @throws RouteException Exception if invalid data is detected but not due to insufficient stock levels.
```
mikejolley: Since this is thrown outside of this method, I think you can remove this line and the phpcs exception. It's documented within convert_notices_to_exceptions

src/StoreApi/Utilities/CartController.php
```diff
@@ -213,17 +215,144 @@ public function validate_add_to_cart( \WC_Product $product, $request ) {
...
+					},
+					$out_of_stock_products
+				);
+
+				$error->add(
+					409,
+					sprintf(
+						/* translators: %s: product names  */
+						_n(
+							'"%s" is out of stock and cannot be purchased.',
```
mikejolley: I like what you've done with these messages but one change needed would be to wrap each product name with `""` rather than the full list. e.g. "Product 1 and Product 2" would become "Product 1" and "Product 2".

src/StoreApi/Utilities/CartController.php
```diff
@@ -213,17 +215,144 @@ public function validate_add_to_cart( \WC_Product $product, $request ) {
...
+			} catch ( NotPurchasableException $error ) {
+				$not_purchasable_products[] = $error;
+			} catch ( PartialOutOfStockException $error ) {
+				$partial_out_of_stock_products[] = $error;
+			} catch ( OutOfStockException $error ) {
+				$out_of_stock_products[] = $error;
+			}
+		}
+
+		if (
```
mikejolley: I think this section can be streamlined. If you create the WP_Error object regardless you can evaluate each error type separately, each using `$error->add` as needed, and then finally you would only need to do a check like this:

```
if ( $error->has_errors() ) {
			WC()->cart->calculate_totals();
			throw new StockAvailabilityException(
				'woocommerce_stock_availability_error',
				$error
			);
}
```

src/StoreApi/Utilities/CartController.php
```diff
@@ -213,17 +215,144 @@ public function validate_add_to_cart( \WC_Product $product, $request ) {
...
+						/* translators: %s: product names  */
+						_n(
+							'"%s" is out of stock and cannot be purchased.',
+							'"%s" are out of stock and cannot be purchased.',
+							count( $out_of_stock_products ),
+							'woo-gutenberg-products-block'
+						),
+						ArrayUtils::natural_language_join( $out_of_stock_product_names )
+					)
+				);
```
mikejolley: This appears to be repeated for each error type, the only difference is the error message text. This seems like it could be easily extracted into a separate method. e.g. `product_stock_error_message( $singular, $plural, $product_names )`. Just an idea, what you have works.

src/StoreApi/Utilities/CartController.php
```diff
@@ -213,17 +215,144 @@ public function validate_add_to_cart( \WC_Product $product, $request ) {
...
+			}
+
+			if ( count( $not_purchasable_products ) > 0 ) {
+				$not_purchasable_product_names = array_map(
```
mikejolley: You can use https://github.com/woocommerce/woocommerce/blob/686872a2372a71ed35b90a8ff790b046d0b40118/includes/wc-core-functions.php#L2055 if you like: `wc_list_pluck( $not_purchasable_products, 'getProductName' )`

mikejolley (commented):
Nice job @opr — this appears to be working well on first try! I have added some inline comments, most are minor. Once addressed I can do another round of testing.

src/StoreApi/Routes/Checkout.php
opr: Thanks for the suggestion, I've now added `add_data_to_error_object` to `Checkout` I don't think it needs to be defined in `AbstractRoute` because it's not used anywhere else except the Checkout. If other routes start returning `WP_Error` objects then we can move it but I think that's out of scope for this PR.

src/StoreApi/Utilities/NotPurchasableException.php
opr: The reason the exceptions are separate like this is because we need to catch them separately in `validate_cart_items` so we can generate the correct error messages. I agree that we should create a new exception to reduce code duplication in these exceptions, but I don't think we can catch the parent exception for the reason stated above.

I propose we rename what is currently `StockAvailabilityException` which is the **ultimate** exception, it contains the `WP_Error` object with all of the error messages inside, to something like `InvalidQuantitiesInCartException`. Then we can use the name `StockAvailabilityException` as the parent of the `OutOfStockException`, `PartialOutOfStockException`, `NotPurchasableException`, and `TooManyInCartException` exceptions.

src/StoreApi/Utilities/CartController.php
opr: I actually don't believe this is needed at all. Removing it now.

src/StoreApi/Routes/AbstractRoute.php
opr: Good call, I wasn't sure if that was overkill, so I didn't do it, but since you suggested it too I have implemented it.

src/StoreApi/Utilities/CartController.php
opr: Yup I like the idea, thanks Mike, I've implemented `add_product_names_to_message` which takes `$singular_error`, `$plural_error`, and `$items`, returning a string.

src/StoreApi/Utilities/CartController.php
opr: Cheers, I've added this functionality to `natural_language_join`: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8094d9f98e7459e79d89c83eb116db9d82508fba/src/Utils/ArrayUtils.php#L17-L24

src/StoreApi/Routes/AbstractRoute.php
```diff
@@ -204,6 +208,19 @@ protected function get_route_error_response( $error_code, $error_message, $http_
 		return new \WP_Error( $error_code, $error_message, array_merge( $additional_data, [ 'status' => $http_status_code ] ) );
 	}
 
+	/**
+	 * Get route response when something went wrong and the supplied error is a WP_Error. This happens when
```
mikejolley: > This happens when

Stray comment here.

src/StoreApi/Utilities/CartController.php
```diff
@@ -213,17 +215,130 @@ public function validate_add_to_cart( \WC_Product $product, $request ) {
...
+				'There are not enough %s in stock. The quantities in your cart have been reduced.',
+				'woo-gutenberg-products-block'
+			);
+
+			$error->add(
+				409,
+				$this->add_product_names_to_message( $singular_error, $plural_error, $partial_out_of_stock_products )
+			);
+
+			if ( $error->has_errors() ) {
```
mikejolley: I think this is nested too deep? It's within the `if ( count( $partial_out_of_stock_products ) > 0 ) {` conditon.

mikejolley (changes requested):
Found a few small issues; I'll do a final test once those are fixed :) Thanks!

src/StoreApi/Utilities/CartController.php
opr: Good spot, I didn't see this when testing because I had partial out of stock products.

mikejolley (approved):
Thanks for taking care of feedback. LGTM now 👍🏻  Nice job!

PR #1892, opened 2020-03-06
Include address fields in the AddressForm component

mikejolley:
This PR moves away from using core data for address fields, into the AddressForm component. This ensures labels are correct (based on optional/required status), and that the data is available at all times.

Fixes #1879
Fixes #1888

Extensibility is hindered, however, this is something we can iterate on in the future.

### How to test the changes in this Pull Request:

1. Go to checkout
2. Try changing countries 
3. Confirm fields change, labels update etc

PR #3662, opened 2021-01-08
Extensibility for Block Checkout Address Fields

### LINKED ISSUE #3633: Sates added with "woocommerce_states" won't show up on the checkout block
## Describe the bug
If we add states to a country by using the `woocommerce_states` filter, for example like this:
`add_filter( 'woocommerce_states', 'woocommerce_portugal_states' );
function woocommerce_portugal_states( $states ) {
	$states["PT"] = array(
		'AC' => __( 'Azores',           'portugal-states-distritos-for-woocommerce' ),
		'AV' => __( 'Aveiro',           'portugal-states-distritos-for-woocommerce' ),
		'BJ' => __( 'Beja',             'portugal-states-distritos-for-woocommerce' ),
		'BR' => __( 'Braga',            'portugal-states-distritos-for-woocommerce' ),
		'BG' => __( 'Bragança',         'portugal-states-distritos-for-woocommerce' ),
		'CB' => __( 'Castelo Branco',   'portugal-states-distritos-for-woocommerce' ),
		'CM' => __( 'Coimbra',          'portugal-states-distritos-for-woocommerce' ),
		'EV' => __( 'Évora',            'portugal-states-distritos-for-woocommerce' ),
		'FR' => __( 'Faro',             'portugal-states-distritos-for-woocommerce' ),
		'GD' => __( 'Guarda',           'portugal-states-distritos-for-woocommerce' ),
		'LR' => __( 'Leiria',           'portugal-states-distritos-for-woocommerce' ),
		'LS' => __( 'Lisbon',           'portugal-states-distritos-for-woocommerce' ),
		'MD' => __( 'Madeira',          'portugal-states-distritos-for-woocommerce' ),
		'PR' => __( 'Portalegre',       'portugal-states-distritos-for-woocommerce' ),
		'PT' => __( 'Oporto',           'portugal-states-distritos-for-woocommerce' ),
		'ST' => __( 'Santarém',         'portugal-states-distritos-for-woocommerce' ),
		'SB' => __( 'Setúbal',          'portugal-states-distritos-for-woocommerce' ),
		'VC' => __( 'Viana do Castelo', 'portugal-states-distritos-for-woocommerce' ),
		'VR' => __( 'Vila Real',        'portugal-states-distritos-for-woocommerce' ),
		'VS' => __( 'Viseu',            'portugal-states-distritos-for-woocommerce' ),
	);
	return $states;
}`

They won't show up on the blocks checkout:
image

As they show up for countries that have their states shipped with WooCommerce itself:
image

## To reproduce
Install this plugin https://wordpress.org/plugins/portugal-states-distritos-for-woocommerce/ and try to checkout with a Portuguese address both on the regular and on the blocks checkout.

* Core version: 5.7-alpha-49862
* WooCommerce version: 4.9.0-rc.1
* WooCommerce Blocks version: 4.1.0
* Any other plugins installed: https://wordpress.org/plugins/portugal-states-distritos-for-woocommerce/

mikejolley:
This is an extensibility enhancement for the Checkout block which fixes #3633.

When developing Checkout we decided to include a list of country locales in Blocks to support an improved format, and allow us to fully control the field locales. This however meant that plugins/users extending core address fields, or adding custom locales (including country states/regions), were not honoured.

This PR proposes a change which imports the locales from core, changes them to our new format, and merges them to create the correct address field config. In order of priority it merges Default Fields > Locale Fields > Overrides at block level.

Fixes #3633

1. Change the country on the checkout and check fields are updated/sorted. Japan is a good example because it re-orders all fields. US has states. UK has a text input for state.
2. Install https://wordpress.org/plugins/portugal-states-distritos-for-woocommerce/ Confirm states appear for portugal
3. Complete checkout - no errors, correct address stored.

### Changelog

> Dev - Block Checkout will now respect custom address locales and custom country states via core filter hooks.


assets/js/base/components/cart-checkout/address-form/index.js
```diff
@@ -76,18 +76,14 @@ const AddressForm = ( {
...
-	const sortedAddressFields = addressFields.sort(
-		( a, b ) => a.index - b.index
-	);
+
 	const countryValidationError =
 		getValidationError( 'shipping-missing-country' ) || {};
+
+	const addressFormFields = useMemo( () => {
+		return prepareAddressFields( fields, fieldConfig, values.country );
+	}, [ fields, fieldConfig, values.country ] );
```
Aljullu: This `useMemo` seems to run on every render, so we should either:
* memoize `fields` as well, so dependencies don't change in every render (`values.country` is a string so it's not a problem, `fieldConfig` seems to be memoized in `address-step.js` already)
* drop `useMemo()` if the performance gain is not worth it?

src/Assets.php
```diff
@@ -196,6 +196,7 @@ public static function get_wc_block_data( $settings ) {
 					$checkout->is_registration_enabled(),
 					FILTER_VALIDATE_BOOLEAN
 				),
+				'checkoutCountryLocale'         => wc()->countries->get_country_locale(),
```
Aljullu: Can we load this setting dynamically, so it's only loaded when Cart and Checkout blocks are rendered? Something like what we do with `shippingCountries`:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/Cart.php#L74-L76

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/Checkout.php#L96-L98

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
```diff
@@ -5,542 +5,73 @@
...
+	if ( localeField.label !== undefined && ! localeField.optionalLabel ) {
+		fields.optionalLabel = sprintf(
+			/* Translators: %s Field label. */
+			__( '%s (optional)', 'woo-gutenberg-products-block' ),
+			localeField.label
+		);
+	}
+
+	if ( localeField.priority ) {
+		fields.index = parseInt( localeField.priority, 10 ) / 10;
```
Aljullu: There shouldn't be any problem for blocks if priority has increments of 10, no? I _think_ there is no need for the division here.

Aljullu (approved):
> When developing Checkout we decided to include a list of country locales in Blocks to support an improved format, and allow us to fully control the field locales. This however meant that plugins/users extending core address fields, or adding custom locales (including country states/regions), were not honoured.

For reference, this is the PR that introduced the change: #1892.

The solution introduced in this PR looks good to me, it seems to be a middle point between only using the data from Core and replicating it in JS.

I confirm it fixes #3633 and I couldn't find any regressions while testing.

I left a couple of in-line comments which I think might be worth fixing, but pre-approving.

assets/js/base/components/cart-checkout/address-form/index.js
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3662/commits/c55f58cc0f0e8ea4e7cb42757124a988f3c06d71

src/Assets.php
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3662/commits/67a8a52d39566fe78a23df7d59b9b93e77935a16

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3662/commits/9a54feffffbab44f49654330b3a6a379499530e5

mikejolley:
@Aljullu I've fixed tests and resolved feedback; ok to merge?

mikejolley:
I know this was approved, but the feedback and test fixes results in more commits. Ideally needs a final look over and green light before it merges. I'm punting to 4.5.

assets/js/base/components/cart-checkout/address-form/index.js
```diff
@@ -1,3 +1,4 @@
 export { default as AddressForm } from './address-form';
 export { default as countryAddressFields } from './country-address-fields';
 export { default as defaultAddressFields } from './default-address-fields';
+export { default as prepareAddressFields } from './prepare-address-fields';
```
Aljullu: Do you think we need to export `prepareAddressFields`? It seems to be used only inside `address-form.js`, and it imports it directly from the file, so the export should be unnecessary.

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
```diff
@@ -5,542 +5,73 @@
...
-	state: {
-		required: false,
-	},
-};
-
-/**
- * Used to hide the state field.
+ * Notes:
+ * 	Ignores "class", "type", "placeholder", and "autocomplete"--blocks handles these visual elements.
+ * 	Core "priority" goes up in increments of 10, but blocks expects 1, 2, 3 etc.
```
Aljullu: This comment is outdated now, no?

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
```diff
@@ -5,542 +5,73 @@
...
+
+	if ( localeField.label !== undefined && ! localeField.optionalLabel ) {
+		fields.optionalLabel = sprintf(
+			/* Translators: %s Field label. */
+			__( '%s (optional)', 'woo-gutenberg-products-block' ),
+			localeField.label
+		);
+	}
+
+	if ( localeField.priority ) {
```
Aljullu, line 50: Might the priority be `0`?

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
```diff
@@ -5,542 +5,73 @@
...
- */
-const hiddenPostcode = {
-	postcode: {
-		required: false,
-		hidden: true,
-	},
-};
+	if ( localeField.label !== undefined ) {
+		fields.label = localeField.label;
+	}
```
Aljullu, line 32: Nitpick: are these if clauses necessary? If `localeField.label` is undefined, I think it's the same assigning it to `fields.label` than not doing it.
```suggestion
	fields.label = localeField.label;
```

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
```diff
@@ -5,542 +5,73 @@
...
+
+	if ( localeField.hidden === true ) {
+		fields.required = false;
+	}
+
+	return fields;
 };
 
-export default countryAddressFields;
+const coreAddressFieldConfig = Object.fromEntries(
```
Aljullu: For some reason `Object.fromEntries()` is not correctly transpiled and is causing a JS error in IE11:

imatge

Aljullu (commented):
Sorry for missing the ping @mikejolley! I did another review and it's testing well. There is a IE11 issue that needs to be fixed and besides that I left some inline comments.

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
mikejolley: Correct me if I am wrong, but that would mean `fields` would have a `label ` key, even through it's not needed and the value would be undefined. This could cause breakage later on if this object is spread.

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
mikejolley: fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3662/commits/dea1b55e07c428e22cc661b751e6e23752598ce9

assets/js/base/components/cart-checkout/address-form/index.js
mikejolley: fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3662/commits/8469e1cc37ff0e807792b9770c9eda3d79fd4828

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
mikejolley: Added a polyfill in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3662/commits/ba343adcf5fd2f843b225aebe340cce9b664c851

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
mikejolley: It could be set to anything, but I think it's reasonable to force a positive integer for this.

mikejolley:
@Aljullu I have not had much luck with browserstack. Does the polyfill work for you?

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
Aljullu: Aaah, yeah, good point. Ignore my comment.

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
Aljullu: It doesn't seem to work for me. :confused: 

imatge

I remember @senadir [added a ponyfill](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/1945/commits/cf28a2fa6690a1f2d0db9dbb087b6752b54fb300) which maybe we could use here?

mikejolley: Good shout, that worked for me also.

mikejolley: Added in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3662/commits/5abf6fa14e4f952ef58a3f895213b22531dc649e

assets/js/base/components/cart-checkout/address-form/country-address-fields.js
```diff
@@ -5,542 +5,74 @@
...
+	}
+
+	return fields;
 };
 
-export default countryAddressFields;
+const coreAddressFieldConfig = Object.entries( coreLocale )
+	.map( ( [ country, countryLocale ] ) => [
+		country,
+		Object.fromEntries(
```
Aljullu: There is still a `fromEntries` here.

Aljullu (approved):
Added an inline comment about another `fromEntries` that needs to be refactored, but after fixing it, everything should work fine so pre-approving. :ok_hand:

Issue #3659, opened 2021-01-07
Integration Tracking (WC Subscriptions): display subscriptions details in cart line item


Aljullu:
We want to display the subscription details (_Free trial_, _Sign up fee_, etc.) below the product description in the cart line item:

![imatge](https://user-images.githubusercontent.com/3616980/103906573-674ecf00-5100-11eb-98c6-1c38b236be2d.png)

They should be displayed like any other metadata info and variations.

Aljullu:
This is closed by #3665 and woocommerce/woocommerce-subscriptions#3943. There seems to be a bug in the implementation, but I created a follow-up issue to investigate it: #3731.

PR #3665, opened 2021-01-08
Cart & Checkout blocks: print all item data and update variation designs

### LINKED ISSUE #3138: Cart and Checkout: print all item data, not only variations (extension compatibility)
See pca54o-gF-p2 for some investigation background.

In order to support Product-Add-Ons extension (and other extensions that use the `woocommerce_get_item_data` filter), we need the Cart and Checkout blocks to render all item data, not only variations.

Steps to fix this:
* Update the API endpoint that returns Cart items so it returns item data in addition to (or instead of only returning) variation data.
* Update components that render variations in C&C, so they render any generic item data.

## Screenshots

| Cart shortcode | Cart block |
| --- | --- |
| imatge-1 | imatge-2 |

_V-Neck T-Shirt is using variations, while Beanie with Logo is using Product Add-Ons. Notice the Cart block shows variations correctly, but doesn’t show the Add-Ons._

### LINKED ISSUE #3630: Update how cart item meta data is shown
Based on the new design iteration for Cart and Checkout, we need to tweak how cart item meta data is shown on cart and checkout.
 
> On desktop viewports, the attribute and addon titles should be in bold. The terms are rendered in regular body font. On mobile devices everything is rendered in the regular body font.
> In the checkout meta, the attribute titles are always rendered using the regular font style so that they don’t distract from the totals in the right column.

figma file: n90hX6hG3wwQbYwsYsrJnj-fi-5119%3A0

screenshot
image
image
image

Aljullu:
This PR:
* prints the cart item data added by extensions to the Cart and Checkout blocks (fixes #3138):
  * this will be useful to display subscription details like sign up fees, free trials, etc. (see follow-up  #3659)
  * some other extensions also add cart item data, like Product Add-Ons
* updates the styles of variation data in Cart and Checkout blocks so they are aligned with the lastest designs (fixes #3630).

### Screenshots

In the screenshots below:

* Magazine is a subscription product.
* Hoodie is a variable product.
* Cap is a product with add-ons.

_Cart:_
imatge

_Checkout:_
Screenshot

Variations:

1. Add a variable product to your cart and go to the Cart and Checkout pages.
2. Verify variations are displayed with the new design (a list instead of separated by `/`).

WC Subscriptions:

1. Currently, WC Subscriptions doesn't use cart item data yet, but you can checkout this WIP branch to test it: `fix/subscription-details-cart-item-data`.
2. Create a subscription product with free trial and/or sign up fee.
3. Add that product to your cart and go to the Cart and Checkout pages.
4. Verify the free trial and sign up fee are displayed.

WC Product Add-Ons:

1. Create a product with a product add-on.
2. Add that product to your cart and go to the Cart and Checkout pages.
3. Verify add-on data is displayed.

### Changelog

> Extensibility: show item data in Cart and Checkout blocks and update the variation data styles.


assets/js/base/components/cart-checkout/product-details/index.js
```diff
@@ -0,0 +1,50 @@
...
+const ProductDetails = ( { details = [] } ) => {
+	if ( details.length === 0 ) {
+		return null;
+	}
+
+	return (
+		<ul className="wc-block-components-product-details">
+			{ details.map( ( detail ) => {
+				return (
+					<li key={ detail.name }>
```
budzanowski: Would it make sense to have a class here that would be generated using the name? This would allow a fine-grained ( for each detail independently ) styling feature.

assets/js/base/components/cart-checkout/product-metadata/index.js
```diff
@@ -22,9 +23,12 @@ const ProductMetadata = ( {
 				shortDescription={ shortDescription }
 				fullDescription={ fullDescription }
 			/>
-			<ProductVariationData
-				className="wc-block-components-product-metadata__variation-data"
-				variation={ variation }
+			<ProductDetails details={ itemData } />
```
budzanowski, line 26: Why do we need two `<ProductDetails>`?

budzanowski (approved):
It looks very nice. I have left two questions, if there is no follow up please ignore them. Approving already.

assets/js/base/components/cart-checkout/product-details/index.js
Aljullu: Good idea, added in bc9fdb738a9a584fbca14aca673911ab6b60892c.

assets/js/base/components/cart-checkout/product-metadata/index.js
Aljullu: According to the designs there should be some margin between the item data added by extensions and variations, so I thought an easy (and semantic) way of doing that was to render them in two different lists. But I don't have a strong opinion on this, alternatively we could render all in the same list and then add the margin with CSS.

budzanowski: OK if it is design-driven.

PR #3667, opened 2021-01-11
WIP Recurring totals in Cart

senadir:
This PR ignores our build process and slotFills and tries to build recurring totals directly in our repo, **therefore it shouldn't be merged as is**
It's here for sharing only and might later change to expose certain components.


senadir:
Now that both https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3671 and 3961-gh-woocommerce/woocommerce-subscriptions are out, we can close this.

Issue #3617, opened 2020-12-30
Integration Tracking (WC Subscriptions): Provide a way for plugins to still capture and process payment even if cart value is $0


senadir:
This is a second use case compared to https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3616. It's very common for WCSubs to create an order with $0 value (free trial) and still capture payment for when a renewal date is up.

Currently in Checkout block, if the cart value is $0, you don't get any payment method (because `cart->needs_payment() = false`) but you still need to capture payment.
You also can't place the order because `order->needs_payment() = true` and so you get "No payment method provided." error.

## Steps to reproduce
- Create A Subscription Product with a free trial and no sign up fees.
- Add it to cart, the total should be 0.
- Checkout, you won't see any payment method.

## Expected Behavior
- Payment method should be visible.

Both this issue and https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3616 are two faces for the same coin, somehow, WCSubs changes the value of `order->needs_payment()` but doesn't do that for Cart and so we have this in which we decide if we want to capture payment using `cart->needs_payment()` and decide if we want to process payment using `order->needs_payment()`.
Ideally, we shouldn't jump to a solution before giving it a deep though first, should we switch both to use `cart` or `order`, or should we keep it as is and change WCSubs to also override `cart` value.
Not assigning this to you @mikejolley but just pinging you for some wisdom.

mikejolley:
This might be tricky because some gateways are not going to allow a 0 payment - stripe for example has a min amount of 0.30p. Whilst we can work around this by using `needs_payment ` instead of the order total, I don't see how we can make gateways process the non-payment...

senadir:
Subscription has a way around this I think, they sometimes charge and refund $0.5 or can disable completely if it doesn't work.
Core checkout allows this currently, so it's possible and we just need to allow payment collection.

opr:
As mentioned in my PR #3670 the cart we send to the client on page load does not reflect the true state of the cart after the extension (subscriptions) has extended the API. So this is not a case of `cart->needs_payment` being false, but more like a problem that we're not seeing the complete picture when we load the page.

Because of Nadir's work in subs, we extend the API to return new data relevant to subscriptions, and during this there is a side-effect that updates the cart. I think this will be unavoidable. The methods relevant to this extension are executed inline, during the creation of the cart, so at this point, the response is already 99% created.

[...]

PR #3670, opened 2021-01-12
Get extended data before returning cart

opr:
In this PR, I have moved the execution of `$this->get_extended_data( self::IDENTIFIER );` in `CartSchema` so that this happens _before_ the cart is returned.

Previously when this was executed inline, any changes made to the cart by extensions were not reflected. By doing it before the cart is returned, we can ensure we're working with the most up to date cart.

This is related to #3617 and _may_ close it.

1. Check out `woocommerce-subscriptions` branch `feature/checkout-block-simple-multiple-subscriptions` and this branch in blocks. Ensure subscriptions is active and set up on your test site.
2. Add a subscription product with a trial period, so that the cart value would be £0. Add this to your cart.
2. Optionally visit the shortcode checkout to verify the cart is set up correctly, and your payment method is still required in spite of the cart being £0.
3. Visit the blocks checkout and ensure payment methods are visible. Please note some non-subscriptions compatible payment methods will show. Choose one you know works with subs, for example Stripe.
4. Try checking out. It should work OK. You may also want to check your payment gateway to see if the Setup Intent (as it is called in Stripe) was set up correctly.

### Changelog

> Ensure extensions can fully modify the cart before it is returned by the API.

### DIFF
--- a/src/StoreApi/Schemas/CartSchema.php
+++ b/src/StoreApi/Schemas/CartSchema.php
@@ -303,6 +303,9 @@ class CartSchema extends AbstractSchema {
 		// calculated so we can avoid returning costs and rates prematurely.
 		$has_calculated_shipping = $cart->show_shipping();
 
+		// Get the extended data before returning the cart - this is because the cart may be modified by extensions.
+		$extended_data = $this->get_extended_data( self::IDENTIFIER );
+
 		return [
 			'coupons'                 => array_values( array_map( [ $this->coupon_schema, 'get_item_response' ], array_filter( $cart->get_applied_coupons() ) ) ),
 			'shipping_rates'          => $has_calculated_shipping ? array_values( array_map( [ $this->shipping_rate_schema, 'get_item_response' ], $controller->get_shipping_packages() ) ) : [],
@@ -332,7 +335,7 @@ class CartSchema extends AbstractSchema {
 				]
 			),
 			'errors'                  => $cart_errors,
-			self::EXTENDING_KEY       => $this->get_extended_data( self::IDENTIFIER ),
+			self::EXTENDING_KEY       => $extended_data,
 		];
 	}
 


src/StoreApi/Schemas/CartSchema.php
```diff
@@ -303,6 +303,9 @@ public function get_item_response( $cart ) {
 		// calculated so we can avoid returning costs and rates prematurely.
 		$has_calculated_shipping = $cart->show_shipping();
 
+		// Get the extended data before returning the cart - this is because the cart may be modified by extensions.
+		$extended_data = $this->get_extended_data( self::IDENTIFIER );
```
Aljullu, line 307: To clarify why I'm proposing. I would run the filter here:
```PHP
apply_filters( 'woocommerce_calculated_total', $core_cart->get_total( 'total' ), $core_cart );
```

And remove this line from Subscriptions:

https://github.com/woocommerce/woocommerce-subscriptions/blob/fd2d01da50a4149e07d586a59ac73fda8e4e3cc6/includes/class-wc-subscriptions-extend-store-endpoint.php#L179

Aljullu (commented):
I missed some of the initial PRs of the Subscriptions - Checkout integration, so forgive me if this has already been discussed, but when reviewing this PR, I was a bit confused about what are the expectations of `get_extended_data()`. I expected it would return some data to extend the API response, but I didn't expect it to modify the internals of the Cart at all. It seems to do that in [this line](https://github.com/woocommerce/woocommerce-subscriptions/blob/fd2d01da50a4149e07d586a59ac73fda8e4e3cc6/includes/class-wc-subscriptions-extend-store-endpoint.php#L179).

I feel this adds complexity to the code, because the callback has two responsibilities 1) returning a value to extend the API and 2) modifying the internal state of the application. From my point of view, these two responsibilities shouldn't be handled by the same method.

Instead of that, I wonder if we could call the `woocommerce_calculated_total` filter, so totals are updated before returning? This filter seems to be the one [currently used for the shortcode integration](https://github.com/woocommerce/woocommerce-subscriptions/blob/trunk/includes/class-wc-subscriptions-cart.php#L69), so I would expect this solution would also facilitate integrating any other extension that uses the same filter.

I know we are moving away from filters where they can be dangerous, but if the same changes can be accomplished in the extend api callbacks, I don't see any reason not to use the existing filters.

Below, I left the code changes I propose, but happy to discuss them. :slightly_smiling_face:

src/StoreApi/Schemas/CartSchema.php
opr: Thanks Albert! Yes your thoughts make a lot of sense, and I agree in principle that we should try to avoid side effects when extending the API. Executing this filter is good, but we are only doing this because _we know_ (as the authors of both plugins) that the cart totals need to be recalculated here.

I am thinking about this in terms of future integrations, if the API is extended, and a new extension modifies the cart and causes a side effect (but one whose changes won't be picked up by `woocommerce_calculated_total`) the cart returned by the API won't be correct, will it? 

Moreover, if the site isn't running subscriptions, we would still be executing `apply_filters( 'woocommerce_calculated_total', $core_cart->get_total( 'total' ), $core_cart );` unnecessarily.

I think my changes leave the plugin to be as agnostic as possible about what is extending it.

Aljullu: Thanks for the response @opr!

> I am thinking about this in terms of future integrations, if the API is extended, and a new extension modifies the cart and causes a side effect (but one whose changes won't be picked up by `woocommerce_calculated_total`) the cart returned by the API won't be correct, will it? 

Yes, the issue is not specific to the `woocommerce_calculated_total` filter. I was probably too fast suggesting the changes before. Instead of calling the filter directly in WC Blocks, it might be better to call a WC method which includes that filter (ie: `WC_Cart->calculate_totals()`).

I didn't dig too much into this, but my idea is that extensions would not need to run side effects inside the `get_extended_data()` callback, instead of that, WC Blocks would call the necessary filters/methods before.

> Moreover, if the site isn't running subscriptions, we would still be executing `apply_filters( 'woocommerce_calculated_total', $core_cart->get_total( 'total' ), $core_cart );` unnecessarily.

That's a good point, but the Cart and Checkout shortcodes [call `WC()->cart->calculate_totals();`](https://github.com/woocommerce/woocommerce/blob/master/includes/shortcodes/class-wc-shortcode-checkout.php#L275) (which applies `woocommerce_calculated_total`) before rendering to ensure they have the most recent values. I wonder if we should do something similar for the Rest API endpoints instead of expecting extensions to call the necessary functions to update the totals. If not, that seems something that extensions will need to account when adding support to WC Blocks because the behavior is different from shortcodes, so it might be good to document it in some way?

senadir: I'm yet to read the full PR and convo above but I disagree about this point
```
// Get the extended data before returning the cart - this is because the cart may be modified by extensions.
		$extended_data = $this->get_extended_data( self::IDENTIFIER );
```
extended data should the be last thing we fetch, all code hooking into this should get the latest and up to date cart, and the code can't have side effects, and if it has, those are going to get negletedd.
side effects can live in other filters and actions, but anything trying to extend data should only worry about shaping its data.

opr: >  it might be better to call a WC method which includes that filter (ie: `WC_Cart->calculate_totals()`)

I think this is a good solution, but would this remove the need for an extension to ever run side effects? I am trying to think about possible side effects that would modify the cart, and that would not be corrected by this method running... Can you think of any?

> my idea is that extensions would not need to run side effects inside the get_extended_data() callback

I think we're not worried about extensions _needing_ to run side effects, but that blocks should tolerate them when they happen, and reflect the most up to date cart. As asked above, are there any conceivable situations where the cart will be modified in such a way that `WC_Cart->calculate_totals()` won't fix it?

opr: I've just tested what you suggested - Adding `$cart->calculate_totals();` before we return the cart, removing the side effect from https://github.com/woocommerce/woocommerce-subscriptions/blob/fd2d01da50a4149e07d586a59ac73fda8e4e3cc6/includes/class-wc-subscriptions-extend-store-endpoint.php#L179 and moving the call to `get_extended_data` back inline. It works fine.

I do think we need to check and be sure that there are no situations we can think of where a cart can be modified by an extension, and that `$cart->calculate_totals();` won't pick this up.

Aljullu: > I do think we need to check and be sure that there are no situations we can think of where a cart can be modified by an extension, and that `$cart->calculate_totals();` won't pick this up.

That's a good point. What I like about calling `$cart->calculate_totals()` is that this seems to be how the shortcodes work, so we get some confidence because it makes our implementation closer to what is already being used by extensions.

Regarding `get_extended_data()`, in my opinion we should be explicit that side effects are not allowed. Like @senadir says, WC core already has several filters/actions that can be used to modify the cart and I don't think `get_extended_data()` callbacks are the right tool for that.

There will be some cases in which the Rest API doesn't apply some core filters/actions. But I think we should evaluate those on a per-case basis: do we want to update the Rest API to support those filters or we want to provide some other way to extend it (via SlotFill, for example)?

senadir: I'm not calling `wc()->cart->calculate_totals();` again the is best outcome here, we're already doing that in Abstract Route before we return, it seems to be an issue with Subscriptions not running its code on the right filter isteadd.

opr:
As discussed in p1610546239193900-slack-C8X6Q7XQU we can close this PR if 3956-gh-woocommerce/woocommerce-subscriptions and #3678 are merged. Waiting on the outcome of both of these now.

PR #3671, opened 2021-01-12
Move Total components to checkout package

senadir:
In this PR I moved several components to Checkout Package, we still have a circular dependency issue but we're avoiding it now with https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3695.
I moved 

- `Subtotal`
- `TotalsTaxes`
- `TotalsShipping`
- `TotalsFees`
- `Panel`
- several utils in `price`.

I also added `className` to all of them so we can pass down classes.

I added `extesions` to useStoreCart.

I passed cart data to slotFills.

## Testing changes
- This won't work unless you follow the testing instructions here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3695
- Make sure Cart and Checkout are working correctly.


assets/js/blocks/cart-checkout/cart/full-cart/index.js
```diff
@@ -149,7 +150,7 @@ const Cart = ( { attributes } ) => {
 					currency={ totalsCurrency }
 					values={ cartTotals }
 				/>
-				<ExperimentalOrderMeta.Slot />
+				<ExperimentalOrderMeta.Slot cart={ cart } />
```
nerrad: I don't think we should expose `cart.receiveCart` on this prop.

assets/js/blocks/cart-checkout/checkout/sidebar/index.js
```diff
@@ -75,7 +76,7 @@ const CheckoutSidebar = ( {
 				currency={ totalsCurrency }
 				values={ cartTotals }
 			/>
-			<ExperimentalOrderMeta.Slot />
+			<ExperimentalOrderMeta.Slot cart={ cart } />
```
nerrad: It'd be nice to avoid having to invoke the hook in this component just so `ExperimentalOrderMeta` can receive the cart as a prop. Could we instead have `ExperimentalOrderMeta` be a wrapper component around the slot and then just expose the data to the slot? Something like:

```js
const ExperimentalOrderMetaSlot = ( props ) => {
  // you could potentially get data from other hooks/context in here to pass along as well?
  return <ExperimentalOrderMeta.Slot cart={ useStoreCart() } { ...props } />
}
```

assets/js/blocks/cart-checkout/cart/full-cart/index.js
senadir: yeah I need to shake down that value before returning it.

assets/js/blocks/cart-checkout/checkout/sidebar/index.js
senadir: We have a wrapper component `Slot` so we can do that there.

senadir:
Size increase didn't sit well with me so I filled this for cooldown https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3697

nerrad:
I rebased on trunk and also pushed 1ae191d which fixes the `wc-blocks-checkout` handle circular dependency issue for the checkout package. We have to decide whether we're okay with exposing `ShippingRatesControl` and `Packages` on the `@woocommerce/checkout` package or whether we want to refactor things differently to avoid the dependency issues.

I haven't tested completely yet, but just got the code up for you to take a look at @senadir and work on next steps.

packages/checkout/shipping/shipping-rates-control/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as ShippingRatesControl } from './shipping-rates-control';
+export { default as Packages } from './packages';
```
senadir, line 2: Don't think we need to export package.

nerrad: Sounds good, you can remove the export here then :)

senadir:
I don't think we can avoid exposing `ShippingRatesControl`. Also I'm not sure but I think @mikejolley  need that component to create shipping controls in Subs, so we would end up exposing it anyway.

Another solution is to simply refactor `TotalsShipping` to simply not be responsible for also showing the selector, and to be frank, we're not using `ShippingRatesControl` and its parent `ShippingRateSelector` at all in WCS, they just happen to be there by nature of how coupled `TotalsShipping`.

nerrad:
One option we could take is make sure it's clear that this entire package is experimental (possibly by prefixing the global property with `__experimental`). Then we can revisit this once we've done a few integrations to decide the approach. The tradeoff of course is that it complicates how extensions manage various scenarios when we graduate this to non-experimental (but that likely can be mitigated by us having the old global be a shim for the new name when we make it stable).

senadir:
I wait until we finish this plugin to decide on how to continue on that, but would probably go the `__experimental` approch.

PR #3672, opened 2021-01-12
Subscriptions Shipping Packages Support

mikejolley:
This PR adds integration points and schema changes to support subscriptions recurring cart shipping methods implemented in https://github.com/woocommerce/woocommerce-subscriptions/pull/3949

There are 2 experimental filters in this PR:

1. `__experimental_woocommerce_store_api_cart_recalculate_totals` to force recalculation so subscriptions populates it's cart(s)
2. `__experimental_woocommerce_store_api_cart_shipping_packages` to allow subscriptions to populate the packages from it's recurring carts.

I am not 100% happy with how this is implemented but this is as far as I can get without changing approach. cc @senadir 

As discussed, the alternative to this would be to instead allow subscriptions to return these packages in it's own API data set, and then allow it to render the shipping options itself. There would be more work involved in this but it would avoid the need for `__experimental_woocommerce_store_api_cart_shipping_packages`.

This is not complete, but I am marking for review to gain feedback.

### Screenshots

Screenshot 2021-01-12 at 16 16 26

### DIFF
--- a/src/StoreApi/Schemas/AbstractSchema.php
+++ b/src/StoreApi/Schemas/AbstractSchema.php
@@ -198,23 +198,6 @@ abstract class AbstractSchema {
 	}
 
 	/**
-	 * Apply a schema get_item_response callback to an array of items and return the result.
-	 *
-	 * @param AbstractSchema $schema Schema class instance.
-	 * @param array          $items Array of items.
-	 * @return array Array of values from the callback function.
-	 */
-	protected function get_item_responses_from_schema( AbstractSchema $schema, $items ) {
-		$items = array_filter( $items );
-
-		if ( empty( $items ) ) {
-			return [];
-		}
-
-		return array_values( array_map( [ $schema, 'get_item_response' ], $items ) );
-	}
-
-	/**
 	 * Returns consistent currency schema used across endpoints for prices.
 	 *
 	 * @return array
--- a/src/StoreApi/Schemas/AbstractSchema.php
+++ b/src/StoreApi/Schemas/AbstractSchema.php
@@ -198,6 +198,23 @@ abstract class AbstractSchema {
 	}
 
 	/**
+	 * Apply a schema get_item_response callback to an array of items and return the result.
+	 *
+	 * @param AbstractSchema $schema Schema class instance.
+	 * @param array          $items Array of items.
+	 * @return array Array of values from the callback function.
+	 */
+	protected function get_item_responses_from_schema( AbstractSchema $schema, $items ) {
+		$items = array_filter( $items );
+
+		if ( empty( $items ) ) {
+			return [];
+		}
+
+		return array_values( array_map( [ $schema, 'get_item_response' ], $items ) );
+	}
+
+	/**
 	 * Returns consistent currency schema used across endpoints for prices.
 	 *
 	 * @return array
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -459,7 +459,6 @@ class CartController {
 	 * Get shipping packages from the cart with calculated shipping rates.
 	 *
 	 * @todo this can be refactored once https://github.com/woocommerce/woocommerce/pull/26101 lands.
-	 *
 	 * @param bool $calculate_rates Should rates for the packages also be returned.
 	 * @return array
 	 */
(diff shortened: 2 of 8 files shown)


src/StoreApi/Schemas/AbstractSchema.php
```diff
@@ -179,6 +179,23 @@ function( $property ) {
...
+	 */
+	protected function get_item_responses_from_schema( AbstractSchema $schema, $items ) {
+		$items = array_filter( $items );
+
+		if ( empty( $items ) ) {
+			return [];
+		}
+
+		return array_values( array_map( [ $schema, 'get_item_response' ], $items ) );
+	}
```
senadir: This seems to be an abstraction only?

src/StoreApi/Routes/AbstractCartRoute.php
```diff
@@ -36,7 +36,9 @@ protected function maybe_recalculate_totals() {
 				]
 			)
 		);
-		if ( ! hash_equals( $current_hash, $new_hash ) ) {
+		$changed      = ! hash_equals( $current_hash, $new_hash );
+
+		if ( apply_filters( '__experimental_woocommerce_store_api_cart_recalculate_totals', $changed ) ) {
```
nerrad: What use cases exist for something using this hook point?

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -309,18 +309,21 @@ public function get_item_response( $cart ) {
...
 		$has_calculated_shipping = $cart->show_shipping();
 
+		// Get shipping packages to return in the response from the cart.
+		$shipping_packages = $has_calculated_shipping ? $controller->get_shipping_packages() : [];
+
 		return [
-			'coupons'                 => array_values( array_map( [ $this->coupon_schema, 'get_item_response' ], array_filter( $cart->get_applied_coupons() ) ) ),
-			'shipping_rates'          => $has_calculated_shipping ? array_values( array_map( [ $this->shipping_rate_schema, 'get_item_response' ], $controller->get_shipping_packages() ) ) : [],
+			'coupons'                 => $this->get_item_responses_from_schema( $this->coupon_schema, $cart->get_applied_coupons() ),
+			'shipping_rates'          => $this->get_item_responses_from_schema( $this->shipping_rate_schema, $shipping_packages ),
```
nerrad: `$has_calculated_shipping` no longer needs considered here?

src/StoreApi/Utilities/CartController.php
```diff
@@ -471,22 +470,26 @@ public function get_shipping_packages( $calculate_rates = true ) {
 			return [];
 		}
 
-		$packages = $cart->get_shipping_packages();
+		$packages = apply_filters( '__experimental_woocommerce_store_api_cart_shipping_packages', $cart->get_shipping_packages() );
```
nerrad, line 473: I'm concerned about how this increases the surface area of how extensions could break the shipping data (i.e. remove existing packages etc). If this approach is taken, ideally we'd surface via an explicit registry that extensions would register callbacks to for enhancing the packages array and we'd merge any packages returned from callbacks in the registry when assembling this data.

senadir: Yes that's going to be the way to go if we want to continue with this approch.

src/StoreApi/Schemas/AbstractSchema.php
mikejolley: I've extracted this into https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3679

src/StoreApi/Schemas/CartSchema.php
mikejolley: Yes but I've extracted this to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3679 because it's not related to this work.

src/StoreApi/Routes/AbstractCartRoute.php
mikejolley: Subs requires a calculation otherwise the recurring cart data is all missing. Looks like we may not need this if we remove the cache - other things like fees also cannot be cached because they are dynamic https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3678

src/StoreApi/Utilities/CartController.php
mikejolley: This is why I was pushing for this to be handled by subscriptions under a different key - adding packages in this context outside of the context of the actual cart doesn't make total sense to me.

mikejolley:
I'm marking this PR blocked whilst we look into the alternative API/rendering approach we discussed.

mikejolley:
Closing this one in favour of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3687

PR #3678, opened 2021-01-13
Support for multiple fee rows in the cart

### LINKED ISSUE #3519: Fees are groupped under a 'Fees' row, instead of being listed individually in the Cart and Checkout blocks
## To reproduce
1. Add this code snippet to `woocommerce-gutenberg-products-block.php`:
```PHP
add_action( 'woocommerce_cart_calculate_fees', 'add_fees', 10 );
function add_fees( $cart ) {
	$cart->add_fee( __( 'Packaging 1', 'woo-gutenberg-products-block' ), 100, true );
	$cart->add_fee( __( 'Packaging 2', 'woo-gutenberg-products-block' ), 50, true );
}
```
2. Go to a page with the Checkout shortcode and notice fees are listed individually:
imatge
3. Go to a page with the Checkout block and notice fees are added up and listed in a 'Fees' row:
imatge

## Expected behavior
I would expect the Checkout block to behave like the shortcode and list each fee individually.

The Cart block is also affected.

mikejolley:
Adds fee data to the API response, useStoreCart hook, and cart/checkout blocks.

Note: Fees are added inline and not persisted. Due to this, cart totals must be calculated otherwise those rows will not be returned. This may degrade performance slightly because caching is not supported.

Fixes #3519

### Screenshots

Screenshot 2021-01-13 at 14 00 44

1. Use the code snippet in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3519
2. Go to the cart page and see fee rows.
3. Go to checkout page and see fee rows.
4. Test with and without prices including tax in WC Tax settings.

### Changelog

> Show itemised fee rows in the cart/checkout blocks.

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/fees/index.js
+++ b/assets/js/base/components/cart-checkout/totals/fees/index.js
@@ -6,36 +6,41 @@ import { DISPLAY_CART_PRICES_INCLUDING_TAX } from '@woocommerce/block-settings';
 import PropTypes from 'prop-types';
 import { TotalsItem } from '@woocommerce/blocks-checkout';
 
-const TotalsFees = ( { currency, values } ) => {
-	const { total_fees: totalFees, total_fees_tax: totalFeesTax } = values;
-	const feesValue = parseInt( totalFees, 10 );
-
-	if ( ! feesValue ) {
-		return null;
-	}
-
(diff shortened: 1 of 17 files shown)


senadir (approved):
Looks and test good, should probably add this to StoreAPI docs.

PR #3683, opened 2021-01-14
Update Jetpack autoloader to 2.7.1

Aljullu:
Cherry-picked commit from @senadir's in #3602. It updates Jetpack autoloader and fixes an error in the tests.

While tests might still fail, verify the error is not related to Jetpack autoloader

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
 	"minimum-stability": "dev",
 	"require": {
 		"composer/installers": "^1.7.0",
-		"automattic/jetpack-autoloader": "^2.0.0"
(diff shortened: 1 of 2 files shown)


nerrad (commented):
Fixes look fine, but looks like there's some phpunit tests (not related to this fix but surfaced as a result of the fix) needing addressed as well.

_Edit:_ Just realized you already knew phpunit tests might fail.

Aljullu:
> _Edit:_ Just realized you already knew phpunit tests might fail.

Yes, working on it right now. :slightly_smiling_face: 

As discussed in Slack, I will merge this PR first and open a follow-up PR with the fixes for the failing tests.

PR #3686, opened 2021-01-14
Change payment processing for subscriptions

### LINKED ISSUE #3271: Integration Tracking (WC Subscriptions): Support saved payment methods for Stripe
When checking out a subscription product, saved payment methods won't show up, and you can't save the current credit card (not sure if those are two separate issues or can be fixed from a single source).

### Steps to reproduce
- Add a subscription product to your cart.
- Try to checkout with stripe.
- You won't see the checkbox to save the card.
- If you already have a card, you won't see it as well.

### Expected behavior
You should be able to use and save cards.
image
image

### LINKED ISSUE #3616: Integration Tracking (WC Subscriptions): Allow zero orders to checkout without payment.
It's very common of Subscriptions to be initially zero dollars, this happens if you crossgrade a subscription and your payment details and method is already saved.
This means checkout wouldn't have any totals ($0) and doesn't need to capture any payment method, checking out, in this case, would return an error (no payment method selected).

This is needed for Subscription and only happen there, if you try to checkout a zero dollar regular product, it would pass fine.

## Steps to reproduce
- In WooCommerce Settings, go to Subscriptions and scroll to Switching and check Allow Switching | Between Subscription Variations.
- Create a variable Subscription Product with two variations of the same price.
- Add the product to your cart and checkout.
- Go to `my-account/subscriptions` and view this new subscription.
- Scroll down to select Downgrade/Upgrade.
- Change the variation and go to checkout.
- Your checkout total is $0, so you don't get any payment methods, your payment method is already saved, so you don't need to add a payment method, placing the order fails with "No payment method provided."

## Expected Behavior
- Checkout should still proceed if no payment method is available

### LINKED ISSUE #3620: Integration Tracking (WC Subscriptions): Switching a subscription creates a new one instead of updating the existing one
When switching a subscription, your is updated with the new details, however, if you checkout via our blocks, this doesn't happen, and a new subscription is created instead.

## To reproduce

- In WooCommerce Settings, go to Subscriptions and scroll to Switching and check Allow Switching | Between Subscription Variations.
- Create a variable Subscription Product with two variations.
- Add the product to your cart and checkout.
- Go to my-account/subscriptions and view this new subscription.
- Scroll down to select Downgrade/Upgrade.
- Change the variation and go to checkout.
- Instead of updating the original subscription, a new one is created, you can verify that by visiting my-account/subscriptions and seeing the new and old one.

## Expected behavior
The original subscription should be updated.

opr:
Saved payment methods are now correctly displayed - previously they were hidden based on whether or not the 'Save payment method' checkbox was visible, by running the `wc_stripe_display_save_payment_method_checkbox` filter.

The checkbox display was ultimately decided by the Stripe extension, which set the checkbox to not display if a subscription product was in the cart. The logic was because the payment method should always be saved when buying a subscription, so displaying the checkbox was unnecessary.

To fix this I have changed the naming of some props used when registering payment methods:
- Previously, we used `savePaymentInfo` to inform the UI whether we should show saved cards **and** the checkbox. 
- I have now changed this to `showSavedCards` and `showSaveOption` to give more granular control over the UI to the payment methods. In our Stripe integration, we set `showSavedCards` based on the value of the `saved_cards` option in the Stripe extension, and we set `showSaveOption` based on the result of the `wc_stripe_display_save_payment_method_checkbox` filter.
- Change `get_allow_saved_cards` method in the `Stripe` class to be named `get_show_saved_cards` and return whether the `saved_cards` option is enabled, without running the checkbox filter mentioned above.
- Add a method to the Stripe class `get_show_save_option` which will return the result of `get_show_saved_cards` filtered through `wc_stripe_display_save_payment_method_checkbox`
- Change the `PaymentMethodTab` component to accept a boolean `showSaveOption` prop. It is this prop that determines if the checkbox should be displayed or not.
- Update the payment method registration code to reflect the new names of these props.
- Update the payment method config and validation to reflect the new names of these props. I also included a shim to set `supports.showSavedCards` to the same value as `supports.savePaymentInfo` for backward compatibility. If any payment methods are registering and still using `savePaymentInfo` then a deprecation notice will be emitted using `@wordpress/deprecated`.
- Updated test files to reflect the new names of these props.

I have also modified the Checkout API so that it will fire a new action `__experimental_woocommerce_blocks_checkout_update_order_meta` which is handled in Subscriptions - this is required as extensions should be given the chance to update the metadata for the orders.

Fixes #3616
Fixes #3620 
Fixes #3271 
Related PR in subs: 3963-gh-woocommerce/woocommerce-subscriptions

### How to test the changes in this Pull Request

1. Check out this branch and `add/actions-hooks-for-update-meta-from-blocks` on `woocommerce-subscriptions`
1. Enable a payment gateway that supports subs and blocks, for example Stripe 💳
2. Add a non-subscription product to your cart. 
3. Go to the checkout block and when using the Stripe payment method verify that the `Save payment method` checkbox shows.
4. Empty your cart.
4. Create a subscription product that has multiple variations so that you may up/down/crossgrade between them. 
3. Add a subscription for this product to your cart and check out using the checkout block, and ensure any saved cards show as options.
4. Ensure the checkout block asks you for payment, or allows you to use an existing method if you have one saved.
5. Ensure the subscription is created correctly with the correct variation, renewal dates, and price. Also ensure the payment method used to create the subscription was recorded.
6. Go to My account -> Subscription -> and begin the process to up/down/crossgrade the subscription.
7. Go to the checkout block and ensure you are not asked for payment. The payment method should already be saved on the subscription. (If you have prorate upgrades and downgrades turned on you may actually be asked for payment here)
8. Check out, and ensure the checkout completes and you can see the order summary showing your subscription on the client.
9. Visit the subscription in the admin dashboard and ensure it has the correct details for the order you just created.
10. Ensure there is a singular additional order in the Orders list for the *grade we just performed.

### DIFF
--- a/assets/js/base/components/payment-methods/payment-method-tab.js
+++ b/assets/js/base/components/payment-methods/payment-method-tab.js
@@ -55,8 +55,7 @@ const PaymentMethodTab = ( { children, showSaveOption } ) => {
 };
 
 PaymentMethodTab.propTypes = {
-	allowsSaving: PropTypes.bool,
-	requiresSaving: PropTypes.bool,
+	showSaveOption: PropTypes.bool,
 	children: PropTypes.node,
 };
 
--- a/assets/js/base/components/payment-methods/payment-method-tab.js
+++ b/assets/js/base/components/payment-methods/payment-method-tab.js
@@ -21,17 +21,13 @@ import PaymentMethodErrorBoundary from './payment-method-error-boundary';
  * @param {Object}  props              Incoming props for the component.
  * @param {boolean} props.allowsSaving Whether that payment method allows saving
  *                                     the data for future purchases.
- * @param {boolean} props.displaySavePaymentMethodCheckbox Whether the payment method should display the option to save
+ * @param {boolean} props.requiresSaving Whether the payment method should display the option to save
  * 														   the details entered by the customer.
  * @param {Object}  props.children     Content of the payment method tab.
  *
  * @return {*} The rendered component.
  */
-const PaymentMethodTab = ( {
-	children,
-	allowsSaving,
-	displaySavePaymentMethodCheckbox,
-} ) => {
+const PaymentMethodTab = ( { children, allowsSaving, requiresSaving } ) => {
 	const { isEditor } = useEditorContext();
 	const {
 		shouldSavePayment,
@@ -42,28 +38,26 @@ const PaymentMethodTab = ( {
 	return (
 		<PaymentMethodErrorBoundary isEditor={ isEditor }>
 			{ children }
-			{ customerId > 0 &&
-				allowsSaving &&
-				displaySavePaymentMethodCheckbox && (
-					<CheckboxControl
-						className="wc-block-components-payment-methods__save-card-info"
-						label={ __(
-							'Save payment information to my account for future purchases.',
-							'woo-gutenberg-products-block'
-						) }
-						checked={ shouldSavePayment }
-						onChange={ () =>
-							setShouldSavePayment( ! shouldSavePayment )
-						}
-					/>
-				) }
+			{ customerId > 0 && allowsSaving && ! requiresSaving && (
+				<CheckboxControl
+					className="wc-block-components-payment-methods__save-card-info"
+					label={ __(
+						'Save payment information to my account for future purchases.',
+						'woo-gutenberg-products-block'
+					) }
+					checked={ shouldSavePayment }
+					onChange={ () =>
+						setShouldSavePayment( ! shouldSavePayment )
+					}
+				/>
+			) }
 		</PaymentMethodErrorBoundary>
 	);
 };
 
 PaymentMethodTab.propTypes = {
 	allowsSaving: PropTypes.bool,
-	displaySavePaymentMethodCheckbox: PropTypes.bool,
+	requiresSaving: PropTypes.bool,
 	children: PropTypes.node,
 };
 
--- a/assets/js/base/components/payment-methods/payment-method-tab.js
+++ b/assets/js/base/components/payment-methods/payment-method-tab.js
@@ -18,16 +18,15 @@ import PaymentMethodErrorBoundary from './payment-method-error-boundary';
 /**
  * Component used to render the contents of a payment method tab.
  *
- * @param {Object}  props              Incoming props for the component.
- * @param {boolean} props.allowsSaving Whether that payment method allows saving
- *                                     the data for future purchases.
- * @param {boolean} props.requiresSaving Whether the payment method should display the option to save
- * 														   the details entered by the customer.
- * @param {Object}  props.children     Content of the payment method tab.
+ * @param {Object}  props                Incoming props for the component.
+ * @param {boolean} props.showSaveOption Whether that payment method allows saving
+ *                                       the data for future purchases and should
+ *                                       display the checkbox to do so.
+ * @param {Object}  props.children       Content of the payment method tab.
  *
  * @return {*} The rendered component.
  */
-const PaymentMethodTab = ( { children, allowsSaving, requiresSaving } ) => {
+const PaymentMethodTab = ( { children, showSaveOption } ) => {
 	const { isEditor } = useEditorContext();
 	const {
 		shouldSavePayment,
@@ -38,7 +37,7 @@ const PaymentMethodTab = ( { children, allowsSaving, requiresSaving } ) => {
 	return (
 		<PaymentMethodErrorBoundary isEditor={ isEditor }>
 			{ children }
-			{ customerId > 0 && allowsSaving && ! requiresSaving && (
+			{ customerId > 0 && showSaveOption && (
 				<CheckboxControl
 					className="wc-block-components-payment-methods__save-card-info"
 					label={ __(
--- a/assets/js/base/components/payment-methods/payment-method-tab.js
+++ b/assets/js/base/components/payment-methods/payment-method-tab.js
@@ -21,11 +21,17 @@ import PaymentMethodErrorBoundary from './payment-method-error-boundary';
  * @param {Object}  props              Incoming props for the component.
  * @param {boolean} props.allowsSaving Whether that payment method allows saving
  *                                     the data for future purchases.
+ * @param {boolean} props.displaySavePaymentMethodCheckbox Whether the payment method should display the option to save
+ * 														   the details entered by the customer.
  * @param {Object}  props.children     Content of the payment method tab.
  *
  * @return {*} The rendered component.
  */
-const PaymentMethodTab = ( { children, allowsSaving } ) => {
+const PaymentMethodTab = ( {
+	children,
+	allowsSaving,
+	displaySavePaymentMethodCheckbox,
+} ) => {
 	const { isEditor } = useEditorContext();
 	const {
 		shouldSavePayment,
@@ -36,25 +42,28 @@ const PaymentMethodTab = ( { children, allowsSaving } ) => {
 	return (
 		<PaymentMethodErrorBoundary isEditor={ isEditor }>
 			{ children }
-			{ customerId > 0 && allowsSaving && (
-				<CheckboxControl
-					className="wc-block-components-payment-methods__save-card-info"
-					label={ __(
-						'Save payment information to my account for future purchases.',
-						'woo-gutenberg-products-block'
-					) }
-					checked={ shouldSavePayment }
-					onChange={ () =>
-						setShouldSavePayment( ! shouldSavePayment )
-					}
-				/>
-			) }
+			{ customerId > 0 &&
+				allowsSaving &&
+				displaySavePaymentMethodCheckbox && (
+					<CheckboxControl
+						className="wc-block-components-payment-methods__save-card-info"
+						label={ __(
+							'Save payment information to my account for future purchases.',
+							'woo-gutenberg-products-block'
+						) }
+						checked={ shouldSavePayment }
+						onChange={ () =>
+							setShouldSavePayment( ! shouldSavePayment )
+						}
+					/>
+				) }
 		</PaymentMethodErrorBoundary>
 	);
 };
 
 PaymentMethodTab.propTypes = {
 	allowsSaving: PropTypes.bool,
+	displaySavePaymentMethodCheckbox: PropTypes.bool,
 	children: PropTypes.node,
 };
 
--- a/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
+++ b/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
@@ -95,7 +95,7 @@ const getCustomerPaymentMethods = ( availablePaymentMethods = {} ) => {
 				const isAvailable = gateway in availablePaymentMethods;
 				return (
 					isAvailable &&
-					availablePaymentMethods[ gateway ].supports?.savePaymentInfo
+					availablePaymentMethods[ gateway ].supports?.showSavedCards
 				);
 			}
 		);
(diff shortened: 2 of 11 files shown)


budzanowski:
> This is not a reliable check, and I don't think it should be the job of the client to decide which payment methods to display. We should rely on the server-side to send back only valid payment methods. This may need testing to ensure only valid payment methods for subscriptions can be selected.

With the check removed, we will surface the saved payment methods even if they are not allowed. If we would want to remove the check here we should ensure that the payment methods are not sending the saved options otherwise we are exposing them to the user even if they are not working. This is how I understand the problem.

nerrad:
> If we would want to remove the check here we should ensure that the payment methods are not sending the saved options otherwise we are exposing them to the user even if they are not working.

I agree with Bartek here. Thomas did you confirm that the server is only surfacing saved payment methods for payment methods that are active? If not, we will need to make sure this is fixed on the server side.

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js
```diff
@@ -91,13 +91,7 @@ const getCustomerPaymentMethods = ( availablePaymentMethods = {} ) => {
 	const enabledCustomerPaymentMethods = {};
 	paymentMethodKeys.forEach( ( type ) => {
 		const methods = customerPaymentMethods[ type ].filter(
-			( { method: { gateway } } ) => {
-				const isAvailable = gateway in availablePaymentMethods;
-				return (
-					isAvailable &&
-					availablePaymentMethods[ gateway ].supports?.savePaymentInfo
-				);
-			}
```
senadir: It's still fuzzy to me what this check does exactly? and why did it had `supports?.savePaymentInfo`?

src/StoreApi/Routes/Checkout.php
```diff
@@ -375,6 +374,23 @@ private function create_or_update_draft_order() {
...
+		 * - To keep the interface focused (only pass $order, not passing request data).
+		 * - This also explicitly indicates these orders are from checkout block/StoreAPI.
+		 *
+		 * @see https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3686
+		 * @internal This Hook is experimental and may change or be removed.
+		 *
+		 * @param WC_Order $order Order object.
+		 */
+		do_action( '__experimental_woocommerce_blocks_checkout_update_order_meta', $this->order );
+
```
senadir, line 374: We did a lot of testing yesterday in our call so I'm confident that this works, I can't comment yet on the payment change until I understand it a bit better but approving this part.

senadir (approved):

assets/js/blocks-registry/payment-methods/payment-method-config.js
```diff
@@ -86,5 +87,14 @@ export default class PaymentMethodConfig {
 				'If the payment method includes the `supports.savePaymentInfo` property, it must be a boolean'
 			);
 		}
+		if (
+			config.supports &&
+			typeof config.supports.requiresSaving !== 'undefined' &&
```
budzanowski: If you use optional chaining here you can get rid of the previous line.

assets/js/base/components/payment-methods/payment-method-tab.js
```diff
@@ -21,11 +21,13 @@ import PaymentMethodErrorBoundary from './payment-method-error-boundary';
  * @param {Object}  props              Incoming props for the component.
  * @param {boolean} props.allowsSaving Whether that payment method allows saving
  *                                     the data for future purchases.
+ * @param {boolean} props.requiresSaving Whether the payment method should display the option to save
+ * 														   the details entered by the customer.
```
nerrad: nit, comment block spacing here seems off.

assets/js/base/components/payment-methods/payment-method-tab.js
```diff
@@ -36,7 +38,7 @@ const PaymentMethodTab = ( { children, allowsSaving } ) => {
 	return (
 		<PaymentMethodErrorBoundary isEditor={ isEditor }>
 			{ children }
-			{ customerId > 0 && allowsSaving && (
+			{ customerId > 0 && allowsSaving && ! requiresSaving && (
```
nerrad: hmm this conditional seems a bit wonky to me. Don't we want something like:
```suggestion
			{ customerId > 0 && ( allowsSaving || requiresSaving ) && (
```

In other words, "If a customer is logged in, and the payment method allows saving or requires saving, then show the checkbox".

src/Payments/Integrations/Stripe.php
```diff
@@ -104,12 +105,22 @@ public function get_payment_method_data() {
 	 * @return bool True if merchant allows shopper to save card (payment method) during checkout).
 	 */
 	private function get_allow_saved_cards() {
-		$saved_cards = isset( $this->settings['saved_cards'] ) ? $this->settings['saved_cards'] : false;
+		return isset( $this->settings['saved_cards'] ) ? 'yes' === $this->settings['saved_cards'] : false;
```
nerrad: nice little improvement here :)

src/Payments/Integrations/Stripe.php
```diff
@@ -104,12 +105,22 @@ public function get_payment_method_data() {
...
+	 * @return bool True if the save payment checkbox should be displayed to the user.
+	 */
+	private function get_requires_saving() {
+		$saved_cards = $this->get_allow_saved_cards();
 		// This assumes that Stripe supports `tokenization` - currently this is true, based on
 		// https://github.com/woocommerce/woocommerce-gateway-stripe/blob/master/includes/class-wc-gateway-stripe.php#L95 .
 		// See https://github.com/woocommerce/woocommerce-gateway-stripe/blob/ad19168b63df86176cbe35c3e95203a245687640/includes/class-wc-gateway-stripe.php#L271 and
 		// https://github.com/woocommerce/woocommerce/wiki/Payment-Token-API .
-		return apply_filters( 'wc_stripe_display_save_payment_method_checkbox', filter_var( $saved_cards, FILTER_VALIDATE_BOOLEAN ) );
+		return ! apply_filters( 'wc_stripe_display_save_payment_method_checkbox', filter_var( $saved_cards, FILTER_VALIDATE_BOOLEAN ) );
```
nerrad: hmm this seems to indicate to me that I might be confused about what `requiresSaving` means. I'm a bit wary of the implicit connection between `requiresSaving` and `allowsSavedCards`. So essentially what we're saying here is that if anything filters this option so it returns false, that means the the saved payment methods ARE required (double negative resulting in that).  I wonder if we should introduce a new filter that allows for explicitly setting this rather than implicitly?

nerrad (commented):
Thanks for tackling this Thomas! I have some feedback that might be due to a lack of fully understanding on my part. As noted in one of the inline comments, I have some concern about the implicit assumptions being made by the usage of the filter.

src/Payments/Integrations/Stripe.php
senadir: > I wonder if we should introduce a new filter that allows for explicitly setting this rather than implicitly?

double down on this.

nerrad: Thomas and I chatted about this in our call and reviewed existing behaviour in the shortcode flow. It turns out we think we can improve this by retaining the explicit use of the filtered value and non filtered value and update the name of the properties to make it more clear how things are used. He's in the process of doing that now. My concerns and wariness with the current approach ended up being more an artifact of how the filtered value was returned (switching boolean) and the naming of the properties/values involved.

assets/js/blocks-registry/payment-methods/payment-method-config.js
opr: Great idea, let's add this as a cooldown task to tidy this up so we don't pollute the PR. Thanks for the tip!

opr: Never mind, I see how small this change is, I thought it was going to be much larger. I've added it to the PR!

opr:
Thanks for the feedback everyone, I have updated the testing instructions and PR body to include more information about what changes I have made. Would you mind taking another look and giving it a quick spin?

assets/js/blocks-registry/payment-methods/payment-method-config.js
```diff
@@ -21,7 +26,11 @@ export default class PaymentMethodConfig {
 		this.canMakePayment = config.canMakePayment;
 		this.paymentMethodId = config.paymentMethodId || this.name;
 		this.supports = {
-			savePaymentInfo: config?.supports?.savePaymentInfo || false,
+			showSavedCards:
+				config?.supports?.showSavedCards ||
+				config?.supports?.savePaymentInfo || // Kept for backward compatibility if methods still pas this when registering.
```
senadir:
```suggestion
				config?.supports?.savePaymentInfo || // Kept for backward compatibility if methods still pass this when registering.
```

assets/js/blocks-registry/payment-methods/payment-method-config.js
```diff
@@ -78,12 +87,24 @@ export default class PaymentMethodConfig {
...
+			typeof config.supports?.showSavedCards !== 'boolean'
+		) {
+			throw new TypeError(
+				'If the payment method includes the `supports.showSavedCards` property, it must be a boolean'
+			);
+		}
+		if ( typeof config.supports?.savePaymentInfo !== 'undefined' ) {
+			deprecated(
+				'Passing savePaymentInfo when registering a payment method is deprecated, instead you should pass showSavedCards and showSaveOption'
+			);
```
senadir: `deprecated` will append a deprecated since... logic into your call, so you should have it in this pattern

```
	deprecated( 'Passing a callback to registerPaymentMethod()', {
			alternative: 'a config options object',
			plugin: 'woocommerce-gutenberg-products-block',
		} );
```

Since we don't intend to remove this anytime soon, no version should be supplied.

Full docs: https://github.com/WordPress/gutenberg/tree/master/packages/deprecated

senadir (approved):
I did a code review, I will do a test review later, I really appreciate the new changed and clarity they bring.

assets/js/base/components/payment-methods/payment-method-tab.js
```diff
@@ -55,6 +56,7 @@ const PaymentMethodTab = ( { children, allowsSaving } ) => {
 
 PaymentMethodTab.propTypes = {
 	allowsSaving: PropTypes.bool,
+	requiresSaving: PropTypes.bool,
```
nerrad: prop types need updated.

assets/js/blocks-registry/payment-methods/payment-method-config.js
```diff
@@ -21,7 +26,11 @@ export default class PaymentMethodConfig {
 		this.canMakePayment = config.canMakePayment;
 		this.paymentMethodId = config.paymentMethodId || this.name;
 		this.supports = {
-			savePaymentInfo: config?.supports?.savePaymentInfo || false,
+			showSavedCards:
+				config?.supports?.showSavedCards ||
+				config?.supports?.savePaymentInfo || // Kept for backward compatibility if methods still pass this when registering.
```
nerrad, line 31: ++ on the comment

nerrad (approved):
Great work @opr - I agree with Nadir, this is much clearer 👏. 

I have one comment around prop types that should be addressed before merging and also we should probably add something about the `supports`  config option and the current available properties in our documentation found at `docs/extensibility/payment-method-integration.md`. 

I'll pre-approve for merging based on the code review. I have not tested, trusting in yours and Nadir's testing of things.

budzanowski (approved):

senadir:
I'm going to rebase, test, and merge this PR.

PR #3690, opened 2021-01-15
Fix broken PHPunit tests

### LINKED ISSUE #3688: Investigate PHP test failures
There are a couple of PHP tests which are failing both in GitHub actions and locally when running `npm run phpunit`.

After some investigation, it looks like the issue is that tests are running against an old version of our codebase (probably the one included in core).

Things we tried:
* Modifying `bin/install-wp-tests.sh` so it added `define( 'JETPACK_AUTOLOAD_DEV', true )` to `wp-config.php`. This didn't make any difference, `wp-config.php` seems to be ignored by tests.
* Modifying the WC version used to [download the zip in `bin/install-wp-tests.sh`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/bin/install-wp-tests.sh#L153) to 4.8 seems to make our codebase to be used for tests, so the issue might have been introduced in WC 4.9.0-beta.1.

nerrad:
Fixes #3688

On investigating this, it looks like our phpunit tests have been running against the blocks code in WooCommerce core since Jetpack autoloader 2.0 update since we were never accounting for blocks dev build being tested (which requires the `JETPACK_AUTOLOAD_DEV` constant to be defined).

To address this:

- I added a `var_dump` to `CartSchema` in blocks to verify when phpunit tests were run they were executing against Blocks plugin classes. They weren't.
- In 5dc1c4d I defined the `JETPACK_AUTOLOAD_DEV` constant in the tests bootstrap and re-ran the tests, this time the debugging was being hit. So that fixed the incorrect classes being autoloaded. 
- With the fix to the above, this surfaced more errors in failing tests which correspond with what @opr [commented on here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3688#issuecomment-760393523), so I fixed that in c37ada2.

If phpunit tests pass in the Github actions this should be good to go.

### DIFF
--- a/src/StoreApi/Schemas/CartSchema.php
+++ b/src/StoreApi/Schemas/CartSchema.php
@@ -338,7 +338,7 @@ class CartSchema extends AbstractSchema {
 			'needs_shipping'          => $cart->needs_shipping(),
 			'has_calculated_shipping' => $has_calculated_shipping,
 			'fees'                    => $this->get_item_responses_from_schema( $this->fee_schema, $cart->get_fees() ),
-			'totals'                  => $this->prepare_currency_response(
+			'totals'                  => (object) $this->prepare_currency_response(
 				[
 					'total_items'        => $this->prepare_money_response( $cart->get_subtotal(), wc_get_price_decimals() ),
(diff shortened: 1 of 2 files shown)


budzanowski:
Very nice.

budzanowski (approved):

PR #3691, opened 2021-01-17
Switch phpunit tests to use wp-env

nerrad:
We're already using `wp-env` for our e2e testing and this pull is an attempt to start using `wp-env` for phpunit testing as well. Some benefits:

- consistency in tools used for automated testing in our repository (as opposed to having some things with a custom docker config and other things using `wp-env`).
- less maintenance, `wp-env` is receiving attention from a much wider developer base than our custom configurations.
- easier access to nice features in `wp-env` and new features as they are developed.
- cleaner log output (the current docker phpunit test has a enormous amount of log output before even getting to the phpunit tests).
- slightly easier configuration for GitHub workflows. For instance, we should be able to do custom matrix testing (WP_version/PHP Version/WC_Version).


nerrad:
Need to make sure `bin/woocommerce` is added to the `.eslintignore` so that local linting won't pick up the eslint config file in the temp cloned woo repo.

nerrad:
There's some current bugs with `wp-env` that will need fixed before we can have multiple php version support:

- Currently PHP 7.4 is the php version that executes the tests and there's a bug where it will install PHPUnit 7.x.x image when phpversion is not specified in .wp-env config.
- If I specify PHP 7.0  I get [this bug](https://github.com/WordPress/gutenberg/issues/27783)
- if I specify PHP 7.2+ in the config, then I get this error:

> Error: Looks like you're using PHPUnit 8.5.13. WordPress requires at least PHPUnit 5.4 and is currently only compatible with PHPUnit up to 7.x.
> Please use the latest PHPUnit version from the 7.x branch.

I think this is because I currently have things set-up to use the `wp-phpunit/wp-phpunit` package which [does not have support for PHPUnit 8.x yet](https://github.com/wp-phpunit/docs) and PHPUnit 8.x is what [is installed when the PHP version is specified](https://github.com/WordPress/gutenberg/blob/01d20a2d46aa202265a972df6bbd8c457284e97e/packages/env/lib/build-docker-compose-config.js#L154-L155). So we need to have control over what version of PHPUnit container is installed.
- I tried running PHP8 and that doesn't even finish setting up certain containers (`tests-cli` as an example). So there's some sort of bug with specifying PHP 8.0 for php version.

Currently using `wp-env` for phpunit only works when PHP version is not specified (thus PHP 7.4).  It's really not much different from what we already have running but I think we might want to hold off making this change until we get the value of begin able to specify php versions for the builds. I might try to debug this and submit some fixes myself at some point.

mikejolley:
Superceded by https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4567

PR #3692, opened 2021-01-17
add circular dependency plugin

senadir:
This plugins watch our files for any circular dependencies issues and log them to the console.

Plugin repo: https://github.com/aackerman/circular-dependency-plugin

### How to test:
- Running `npm start` should produce some errors, they're non-blocking so you should use blocks fine.

### DIFF
--- a/bin/webpack-configs.js
+++ b/bin/webpack-configs.js
@@ -12,6 +12,7 @@ const WebpackRTLPlugin = require( 'webpack-rtl-plugin' );
 const chalk = require( 'chalk' );
 const { kebabCase } = require( 'lodash' );
 const CreateFileWebpack = require( 'create-file-webpack' );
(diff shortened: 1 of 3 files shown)


senadir:
Even thou I had set this to warn instead of error, it still breaks our actions, might need to tweak with plugin config to not emit an error.

senadir:
Based on a previous conversation with @nerrad, we're not going to merge this as is, but probably have it available on a special command (like our bundle analysis command) to be run when debugging this issue, therefore, I'm going to change this to a draft and unassign myself until I have some time to revisit it or if anyone else wants to pick it up.

nerrad:
I'm going to close this PR because the plugin is being added as a part of #4025 - thanks Nadir for surfacing this plugin and doing the initial work on it here!

PR #3693, opened 2021-01-17
Fix error with eslint workflow artifact download issues.

nerrad:
Recently the ESLint workflow hasn't been working with the following errors in the log:

```
Error: HttpError: Failed to generate URL to download artifact
Error: Failed to generate URL to download artifact
```

I spent some time trying to debug this (including seeing if deleting artifacts would do anything) but was unsuccessful. In the end I ended up switching to use a [different action](https://github.com/ataylorme/eslint-annotate-action) that seems to be more resilient and uses eslint reporting for doing the checks. 

Annotation is still supported.

In the process I cleaned up the workflow a bit with a little bit better naming of steps. 

The only slight downside to this change is that it lints all files not just changed, so that means that there will be a section in the annotations for unchanged files causing linting warnings (which might not be a bad thing because it keeps the warnings from being forgotten).

I tested deliberately causing a eslint fail and this caught it nicely. I then forced a reset of the history for the pull so you won't see that test in the current history. Once Github actions pass this can be merged.

### DIFF
--- a/.github/workflows/js-css-linting.yml
+++ b/.github/workflows/js-css-linting.yml
@@ -6,37 +6,74 @@ on:
     branches: [trunk]
 
 jobs:
-  check:
-    name: All
-
+  Setup:
+    name: Setup for Jobs
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+      - uses: actions/checkout@v2
+      - name: Cache node modules
+        uses: actions/cache@v2
+        with:
+          path: node_modules
+          key: ${{ runner.OS }}-build-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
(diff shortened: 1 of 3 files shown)


senadir (approved):
Do we still get to keep the comment? or is it replaced with code comments?

anyway looks good to me.

nerrad:
> Do we still get to keep the comment? or is it replaced with code comments?

Unfortunately no - the new action doesn't leave any comments in the pr, it just annotates the code changes. For now, I'm okay with that because we still get the functional value out of the eslint checks and if there's a failure - having it annotated inline with the code changes is a useful way to parse.

During the next cooldown there will be opportunity for further exploration of alternatives but I just wanted to get this fixed now to avoid blocking our current work.

senadir:
I prefer annotation instead of a summary comment, I mostly asked to confirm that this is not a leftover from testing around.

Issue #3694, opened 2021-01-17
base-components, base-hook, and base-context has circular dependency on itself


senadir:
When working on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3671 I run into a circular dependency issue, it seems that wc-blocks-checkout was marking itself as a dependency.
Upon further inspection, it seems that any import from `base-components` `base-hooks` or `base-context` (which depends on `base-hook`) would cause this.
After further inspection, I managed to print this error list (using the code in this PR https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3692) and it printed 115 circular dependency issue.

Examples of those issues
```
    ERROR in Circular dependency detected:
    assets/js/base/components/text-input/index.js -> assets/js/base/components/text-input/validated.js -> assets/js/base/context/index.js -> assets/js/base/context/cart-checkout/index.js -> assets/js/base/context/cart-checkout/payment-methods/index.js -> assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js -> assets/js/base/hooks/index.js -> assets/js/base/hooks/payment-methods/index.js -> assets/js/base/hooks/payment-methods/use-payment-method-interface.js -> assets/js/base/components/cart-checkout/index.js -> assets/js/base/components/cart-checkout/address-form/index.js -> assets/js/base/components/text-input/index.js
    
    ERROR in Circular dependency detected:
    assets/js/base/components/text-input/validated.js -> assets/js/base/context/index.js -> assets/js/base/context/cart-checkout/index.js -> assets/js/base/context/cart-checkout/payment-methods/index.js -> assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.js -> assets/js/base/hooks/index.js -> assets/js/base/hooks/payment-methods/index.js -> assets/js/base/hooks/payment-methods/use-payment-method-interface.js -> assets/js/base/components/cart-checkout/index.js -> assets/js/base/components/cart-checkout/address-form/index.js -> assets/js/base/components/text-input/index.js -> assets/js/base/components/text-input/validated.js
    
    ERROR in Circular dependency detected:
[...]

PR #3695, opened 2021-01-17
Add circular dependency check and fix it

senadir:
This PR adds a check around our enqueue call, it checks if the current script being registered has a dependency on its own.
When this happens and left unfixed, PHP will emit a `maximum function nesting level of '256' reached aborting` error and the script won't load.
This PR will throw an exception.

```
Script with handle %s had a dependency on itself. This is an indicator that your JS code has a circular dependency that can cause bugs.
```

### How to test

- Check out this branch https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3671
- check that Cart block is not loading, with a nesting level warning below.
- cherry-pick this PR into that branch `git cherry-pick 2450048`
- Refresh the page, a more descriptive error is shown.

### DIFF
--- a/src/Assets/Api.php
+++ b/src/Assets/Api.php
@@ -87,7 +87,7 @@ class Api {
 		}
 
 		if ( in_array( $handle, $dependencies, true ) ) {
-			if ( Package::feature()->is_development_environment() ) {
+			if ( $this->package->feature()->is_development_environment() ) {
 				$dependencies = array_diff( $dependencies, [ $handle ] );
 				// phpcs:ignore
 				trigger_error( sprintf( 'Script with handle %s had a dependency on itself which has been removed. This is an indicator that your JS code has a circular dependency that can cause bugs.', $handle ), E_USER_WARNING );
--- a/src/Assets/Api.php
+++ b/src/Assets/Api.php
@@ -84,6 +84,14 @@ class Api {
 			$version = $this->get_file_version( $relative_src );
 		}
 
+		if ( in_array( $handle, $dependencies, true ) ) {
+			$dependencies = array_diff( $dependencies, [ $handle ] );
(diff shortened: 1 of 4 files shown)


src/Assets/Api.php
```diff
@@ -84,6 +84,14 @@ public function register_script( $handle, $relative_src, $dependencies = [], $ha
 			$version = $this->get_file_version( $relative_src );
 		}
 
+		if ( in_array( $handle, $dependencies, true ) ) {
+			$dependencies = array_diff( $dependencies, [ $handle ] );
+			if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
+				// phpcs:ignore
+				trigger_error( sprintf( 'Script with handle %s had a dependency on itself which we removed. This is an indicator that your JS code has a circular dependency that can cause bugs.', $handle ), E_USER_WARNING );
```
Aljullu: If the dependency is removed automatically, I wonder if it makes more sense avoiding the pronoun ("has been removed" instead of "we removed").
```suggestion
				trigger_error( sprintf( 'Script with handle %s had a dependency on itself which has been removed. This is an indicator that your JS code has a circular dependency that can cause bugs.', $handle ), E_USER_WARNING );
```

Aljullu (approved):
> This is introduced to bypass error found in #3694 so we can load our scripts, but it doesn't close that issue because it is preferred that we fix the circular dependency because it might cause memory limits or unexpected bugs.

:100: 

This is testing as described and code look good, I added a suggestion about changing the error message, but not a big deal so up to you if you apply it or not.

I see linting is failing, but I assume that has been fixed in #3693.

senadir:
The warning can get extremely lengthy, we should probably switch to something less "in your face" giving that this issue will take some time to be fully fixed.
image

PR #3703, opened 2021-01-19
Catch errors during cart validation

### LINKED ISSUE #3412: Checkout block: pending order when product is out of stock, generic error message

### LINKED ISSUE #3699: Empty error when a product in cart gets out of stock and the user updates the quantity of another product
Found when testing WC Blocks 4.3.0 (#3698).

## To reproduce

1. Add Cap and Belt (for example) to your cart.
2. Go to the Cart page.
3. In the admin, change Belt inventory status to be out of stock.
4. Increase the Cart quantity in the cart.
5. Notice Belt disappears from the page and you get an empty error message:

imatge

The response seems to include the `woocommerce_rest_cart_product_no_stock` error but without any message:

imatge

## Expected behavior
The error should show the message.

opr:
This PR extends the work done in #3656 - there was a bug found where the errors would not display properly in the cart block so this was reverted from trunk for the release. This PR includes those changes, but makes further changes to the way we handle the different types of out of stock error and refactors the way the errors are generated so that they can work in both the cart and checkout blocks.

This PR adds some new exceptions that can be thrown when validating the cart contents. These exceptions all relate to different types of "out of stock"

##### Thrown in `validate_cart_item`
- `PartialOutOfStockException` - Thrown when there _are_ items in stock, but not enough to satisfy the customer's order. The quantity in cart will be reduced to the in-stock amount.
- `TooManyInCartException` - Thrown when products can only be purchased individually but there is more than one in the cart. The quantity in cart will be reduced to 1.
- `OutOfStockException` - Thrown when the product is completely out of stock. The item is removed from cart.
- `NotPurchasableException` - Thrown when the product is not able to be purchased (doesn't exist, doesn't have a price etc.)
##### Thrown in `validate_cart_items`
- `StockValidationException` - Thrown as a "general" exception from `validate_cart_items` when one/some/all of the above exceptions are thrown in `validate_cart_item`. This Exception contains a `WP_Error` object that in turn contains all of the errors thrown by each execution of`validate_cart_item`. This `WP_Error` is then converted into a response and sent to the client so that the individual error notices might be displayed.

It also introduces the following code changes:

1. `get_error_message_for_stock_exception_type` function in the `CartController` which will return an appropriate error message based on the type of `StockAvailabilityException` and whether a single product or multiple products are causing the exception.
2. `stock_exceptions_to_wp_errors` in `CartController` - this takes all of the exceptions generated when checking the cart items' validities and turns them into `WP_Error`s
3. Add `get_route_error_response_from_object` to `AbstractRoute` - this will create an error response when the error is already a `WP_Error`.
4. Clean up code in `Routes/Checkout` and abstract adding the cart to the response to its own function so we don't repeat code
5. Creation of a `natural_language_join` function in the `ArrayUtils` class - used to join an array with a (translatable) natural language conjunction between the second to last and last item.

Fixes #3699
Fixes #3412

### Cart Before
image

### Cart After
image

### Checkout Before
Screenshot 2021-01-06 at 16 06 36

### Checkout After
Screenshot 2021-01-06 at 14 45 04

1. **In an incognito window** Add several products to your cart. Add these in differing quantities. Several of these added products need to have a quantity of 3+ in the cart.
2. Open the checkout page and ensure all the products are showing. **Do different combinations of the following steps in a different session to the one where you've added the items, and also without refreshing the checkout page. This is necessary because we can't be logged in as admin when testing step 3 below.**
3. Delete one or more of the products from the back end.
4. Set one or more of the products to "Manage stock?" true, and edit the in-stock amount of one or more of the products ensuring it is below the quantity that is in the cart, but above 0.
5. Set one or more of the products to "Sold individually"
6. Set one or more of the products to be entirely out of stock.
7. Try to check out and ensure you see the correct error messages for the products whose stock you've changed/reduced.
8. Try a few different scenarios with stock levels, number of products affected, number of products in cart etc. 

### Changelog

> Improve error displayed to customers when an item's stock status changes during checkout.

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -231,15 +231,13 @@ const CheckoutProcessor = () => {
 						id: 'checkout',
 					} );
 
-					if ( Array.isArray( response.additional_errors ) ) {
-						response.additional_errors.forEach(
-							( additionalError ) => {
-								addErrorNotice( additionalError.message, {
-									id: additionalError.error_code,
-								} );
-							}
-						);
-					}
+					response.additional_errors?.forEach?.(
+						( additionalError ) => {
+							addErrorNotice( additionalError.message, {
+								id: additionalError.error_code,
+							} );
+						}
+					);
 
 					dispatchActions.setHasError();
 					dispatchActions.setAfterProcessing( response );
--- a/assets/js/base/context/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/cart-checkout/checkout/processor/index.js
@@ -230,6 +230,17 @@ const CheckoutProcessor = () => {
 					addErrorNotice( formatStoreApiErrorMessage( response ), {
 						id: 'checkout',
 					} );
+
+					if ( Array.isArray( response.additional_errors ) ) {
+						response.additional_errors.forEach(
+							( additionalError ) => {
+								addErrorNotice( additionalError.message, {
+									id: additionalError.error_code,
+								} );
+							}
+						);
+					}
+
 					dispatchActions.setHasError();
 					dispatchActions.setAfterProcessing( response );
 					setIsProcessingOrder( false );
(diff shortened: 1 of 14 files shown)


budzanowski:
This looks very nice!

budzanowski (approved):

PR #3704, opened 2021-01-19
Fix circular dependencies

nerrad:
Through work @senadir has been doing in #3671, #3692, and #3671, he has discovered a number of circular dependency issues in our code structure. This pull request is a result of an audit I did of our code where I did find some issues with how we were importing various things.

A couple cautions:

- the circular dependency plugin that was added in #3692 is great (and I used it to catch many of the initial issues before just manually reviewing) but it doesn't work so well with wildcard exports. So some of the issues it's catching don't look to be legit for our build process.
- It's very likely that some of these issues were hidden practically in our actual builds because of a combination of our build process and how some of the packages references are actually externals. So this means we may see slight increases in built files with these changes (I'm hoping for decreases). It also means that the practical benefit of this pull request is more stable import chains. 
- This doesn't fix the original circular dependency issue Nadir was facing in #3671 with the assets dependencies php side. That was actually due to a separate issue with a circular dependency with `@woocommerce/checkout` and some of the work he was doing in that PR (which I confirmed separately). After this pull is merged into trunk and into that PR, we'll address the maining root cause there.

### What were the circular dependencies?

One of the things we have to watch for that create circular dependencies is importing a component from a package that exports from the file you are writing that import in. So for example, say you have a folder structure like:

```
- base
--- components
----- index.js
----- a.js
----- b.js
```

Your files might have something like this.

```js
// in index.js
export * from 'a';
export * from 'b';

// in a.js
import { bThing } from './';

export const aThing = ( d ) => bThing( d );

// in b.js
export const bThing = ( e ) => console.log( e );
```

Do you see the problem here? `a.js` is importing `bThing` from the `index.js` file _which is also exporting from `a.js`_. So this creates a circular dependency. To fix, you just need to make sure you do something like this in `a.js`:

```js
import { bThing } from './b';

export const aThing = ( d ) => bThing( d );
```

In our case, it's easy to miss things like this because we also use aliases to avoid deeply nested, hard to read, relative paths in our imports. But we had cases where we were essentially duplicating the same pattern with the aliases too. Let's repeat our same example, but instead, let's say that we have an alias `@basecomponents` for the imports that points to `base/components/index.js`.

```js
// in index.js
export * from 'a';
export * from 'b';

// in a.js
import { bThing } from '@basecomponents';

export const aThing = ( d ) => bThing( d );

// in b.js
export const bThing = ( e ) => console.log( e );
```

See how this could slip by unnoticed? As a rule of thumb, when using an aliased package import, make sure you are not using it in a nested situation like the above where you are essentially importing from a file that exports the file you are importing into.

## To Test

The changes in this code impact:

- All Products Block
- Cart and Checkout Blocks.

To verify there is no breakage:

- Manual smoke testing of All Products behaviour and Cart and Checkouts behaviour (including full purchase process).
- E2E tests should pass
- Unit tests should pass.


senadir (approved):
This tests great for me and there's no change in code, I didn't validate it against the webpack plugin or with the checkout code, I'd prefer we merge this and get to rebasing ASAP.

PR #3706, opened 2021-01-19
add missing aria-label for stars image in review-list-item component

### LINKED ISSUE #1675: Reviews blocks: rating has role="img" but missing aria-label
Ratings shown in the reviews blocks are not accessible. They have `role="img"` but are missing an `aria-label` prop to define them:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/base/components/review-list-item/index.js#L126

Instead of directly solving it there, I think it would make more sense to reuse the logic from [`atomic/components/product/rating/index.js`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/master/assets/js/atomic/components/product/rating/index.js), which was made accessible in #1656 and has exactly the same markup (only class names differ).

ralucaStan:
This PR adds an aria-label for the stars image in the review-list-item component for a11y purposes.

Fixes #1675 

#### Accessibility

1. Insert *All reviews* block into a page
2. Use the developer tools to inspect one of the review items in the Reviews list
3. Select the div containing the start image
4. Make sure it has `role="img"` & `aria-label="Rated X out of 5"`
5. Check that the child span has the same text content as the `aria-label` attribute

### Changelog

> Add missing aria-label for stars image in the review-list-item component.

### DIFF
--- a/assets/js/base/components/reviews/review-list-item/index.js
+++ b/assets/js/base/components/reviews/review-list-item/index.js
@@ -124,22 +124,19 @@ function getReviewRating( review ) {
 	const starStyle = {
 		width: ( rating / 5 ) * 100 + '%' /* stylelint-disable-line */,
 	};
+	const ratingText = sprintf(
+		/* Translators: %f is referring to the average rating value */
+		__( 'Rated %f out of 5', 'woo-gutenberg-products-block' ),
+		rating
+	);
 	return (
 		<div className="wc-block-review-list-item__rating wc-block-components-review-list-item__rating">
 			<div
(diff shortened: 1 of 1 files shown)


ralucaStan:
@Aljullu I tried testing this with Voice Over to see if I get the reader to read the text inside the aria-label.
Because it's a div it's not an element that receives focus so it was not read to me "Rated X out of 5".

What is the expected result here? Do we want this element to receive focus so that it's read to the user? 
Curious if you were able to test this with a screen reader app?

Aljullu (approved):
This is testing well. I tested with Orca screen reader and it correctly read the rating. Code also looks good, good job! :ok_hand:

PR #3719, opened 2021-01-21
Introduce payment gateway feature check for cart content.

### LINKED ISSUE #3252: Integration Tracking (WC Subscriptions): Ensure subscription orders use a supported payment method when using checkout block

budzanowski:
This PR implements a payment gateways capability check for the content in the cart. This means that a payment gateway will not be shown in the available payments methods list ( at checkout ) if not all of the required features are provided by that payment gateway. 

It consists of three parts:

1. Required features integration.
2. Payment gateway features list
3. Glue code that compares required features to features provided by the payment gateways. 

### 2 Payment gateway features list
Each payment gateway needs to advertise what type of payment operations it supports.

### 3 Glue code that compares required features to features provided by the payment gateways. 
Before the Checkout block displays the available payments methods the `usePaymentMethodRegistration` check which payment methods are valid for this cart setup. This PR adds additional features. It crosschecks what is required by the cart and what is required by 

Fixes #3252

TODO:
- [x] Move features check to can make payment function.
- [x] Prepare subscriptions PR that injects subscription requirements to carts with subscription type items. 3966-gh-woocommerce/woocommerce-subscriptions
- [x] Make sure that all payment gateways we integrate are supported.
- [x] _experimental for filters.

1. Use 3966-gh-woocommerce/woocommerce-subscriptions
2. Setup payment methods that do not support subscriptions and payment methods that do
3. add non-subscription product to the cart
4. View the checkout block - all payment methods should be there
5. Add subscription product to the cart
6. View the checkout block - only payment methods that support subscriptions should be available
7. Enable manual payments for renewal
8. View the checkout block - all payments should be visible, no matter if subscription items are in the cart or not.
9. Also check with express payments.

### Changelog

Update - Filter payment methods by features required by the content of the cart.

### DIFF
--- a/src/Domain/Services/ExtendRestApi.php
+++ b/src/Domain/Services/ExtendRestApi.php
@@ -62,6 +62,13 @@ final class ExtendRestApi {
 	private $extend_data = [];
 
 	/**
+	 * Array of payment requirements
+	 *
+	 * @var array
+	 */
+	private $payment_requirements = [];
+
+	/**
 	 * An endpoint that validates registration method call
 	 *
 	 * @param array $args {
@@ -133,6 +140,60 @@ final class ExtendRestApi {
 			$registered_data[ $namespace ] = $data;
 		}
 		return (object) $registered_data;
+	}
+
+	/**
+	 * Registers and validates payment requirements callbacks.
+	 *
+	 * @param array $args {
+	 *     Array of registration data.
+	 *
+	 *     @type callable $data_callback Callback executed to add payment requirements data.
+	 * }
+	 *
+	 * @throws Exception On failure to register.
+	 * @return boolean True on success.
+	 */
+	public function register_payment_requirements( $args ) {
+		if ( ! is_callable( $args['data_callback'] ) ) {
+			$this->throw_exception( '$data_callback must be a callable function.' );
+		}
+
+		$this->payment_requirements[] = $args['data_callback'];
+
+		return true;
+	}
+
+	/**
+	 * Returns the additional payment requirements.
+	 *
+	 * @param array $initial_requirements list of requirements that should be added to the collected requirements.
+	 * @return array Returns a list of payment requirements.
+	 * @throws Exception If a registered callback throws an error, or silently logs it.
+	 */
+	public function get_payment_requirements( array $initial_requirements = [ 'products' ] ) {
+		$requirements = $initial_requirements;
+		if ( empty( $this->payment_requirements ) ) {
+			return $initial_requirements;
+		}
+
+		foreach ( $this->payment_requirements as $callback ) {
+			$data = [];
+
+			try {
+				$data = $callback();
+
+				if ( ! is_array( $data ) ) {
+					throw new Exception( '$data_callback must return an array.' );
+				}
+			} catch ( Throwable $e ) {
+				$this->throw_exception( $e );
+				continue;
+			}
+			$requirements = array_merge( $requirements, $data );
+		}
+
+		return array_unique( $requirements );
 	}
 
 	/**
--- a/src/Payments/Integrations/PayPal.php
+++ b/src/Payments/Integrations/PayPal.php
@@ -71,7 +71,7 @@ final class PayPal extends AbstractPaymentMethodType {
 		return [
 			'title'       => $this->get_setting( 'title' ),
 			'description' => $this->get_setting( 'description' ),
-			'supports'    => apply_filters( '__experimental_woocommerce_blocks_payment_gateway_features_list', $this->get_supported_features(), $this->get_name() ),
+			'supports'    => $this->get_supported_features(),
 		];
 	}
 
@@ -81,7 +81,8 @@ final class PayPal extends AbstractPaymentMethodType {
 	 * @return string[]
 	 */
 	public function get_supported_features() {
-		$gateway = new WC_Gateway_Paypal();
-		return array_filter( $gateway->supports, array( $gateway, 'supports' ) );
+		$gateway  = new WC_Gateway_Paypal();
+		$features = array_filter( $gateway->supports, array( $gateway, 'supports' ) );
+		return apply_filters( '__experimental_woocommerce_blocks_payment_gateway_features_list', $features, $this->get_name() );
 	}
 }
--- a/src/Payments/Integrations/PayPal.php
+++ b/src/Payments/Integrations/PayPal.php
@@ -1,9 +1,7 @@
 <?php
 namespace Automattic\WooCommerce\Blocks\Payments\Integrations;
 
-use Exception;
-use WC_Stripe_Payment_Request;
-use WC_Stripe_Helper;
+use WC_Gateway_Paypal;
 use Automattic\WooCommerce\Blocks\Assets\Api;
 
 /**
@@ -74,5 +72,15 @@ final class PayPal extends AbstractPaymentMethodType {
 			'title'       => $this->get_setting( 'title' ),
 			'description' => $this->get_setting( 'description' ),
 		];
+	}
+
+	/**
+	 * Returns an array of supported features.
+	 *
+	 * @return string[]
+	 */
+	public function get_supported_features() {
+		$gateway = new WC_Gateway_Paypal();
+		return array_filter( $gateway->supports, array( $gateway, 'supports' ) );
 	}
 }
--- a/src/Payments/Integrations/PayPal.php
+++ b/src/Payments/Integrations/PayPal.php
@@ -71,6 +71,7 @@ final class PayPal extends AbstractPaymentMethodType {
 		return [
 			'title'       => $this->get_setting( 'title' ),
 			'description' => $this->get_setting( 'description' ),
+			'supports'    => apply_filters( 'woocommerce_blocks_payment_gateway_features_list', $this->get_supported_features(), $this->get_name() ),
 		];
 	}
 
--- a/src/Payments/Integrations/PayPal.php
+++ b/src/Payments/Integrations/PayPal.php
@@ -71,7 +71,7 @@ final class PayPal extends AbstractPaymentMethodType {
 		return [
 			'title'       => $this->get_setting( 'title' ),
 			'description' => $this->get_setting( 'description' ),
-			'supports'    => apply_filters( 'woocommerce_blocks_payment_gateway_features_list', $this->get_supported_features(), $this->get_name() ),
+			'supports'    => apply_filters( '__experimental_woocommerce_blocks_payment_gateway_features_list', $this->get_supported_features(), $this->get_name() ),
 		];
 	}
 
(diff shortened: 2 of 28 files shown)


src/Payments/PaymentMethodRegistry.php
```diff
@@ -153,7 +153,8 @@ public function get_all_registered_script_data() {
 		$payment_methods = $this->get_all_active_registered();
 
 		foreach ( $payment_methods as $payment_method ) {
-			$script_data[ $payment_method->get_name() . '_data' ] = $payment_method->get_payment_method_data();
+			$script_data[ $payment_method->get_name() . '_data' ] =
+				array_merge( $payment_method->get_payment_method_data(), $payment_method->get_supported_features() );
```
senadir: why can't this be part of `get_payment_method_data`?

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -302,6 +302,12 @@ public function get_properties() {
 					'properties' => $this->force_schema_readonly( $this->error_schema->get_properties() ),
 				],
 			],
+			'payment_requirements'    => [
+				'description' => __( 'List of required payment gateway features to process the order.', 'woo-gutenberg-products-block' ),
+				'type'        => 'array',
+				'context'     => [ 'view', 'edit' ],
+				'readonly'    => true,
+			],
```
senadir, line 310: Conceptually, this makes more sense to be within `CheckoutSchema` but due to limitations (we don't get an updated `checkout` response on each change in cart like shipping, items, coupons...), we have to leave it here.

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -356,6 +362,7 @@ public function get_item_response( $cart ) {
 				]
 			),
 			'errors'                  => $cart_errors,
+			'payment_requirements'    => apply_filters( 'woocommerce_blocks_store_api_cart_payment_requirements', array( 'products' ) ),
```
senadir: I don't think we should be using filters here, we should use ExtendRestAPI, here's why:
**- Filters code can break**: any exception would bubble up to our cart response and return a broken response (breaking the cart).
**- Filters can break other plugins**: developers might forget to append to the array passed, might hardcode an array, causing the code applied by other plugins to be superseded.
**- Filters are not type-safe**: developers can return a string (instead of an array of string), return an object, or any other shape that might break our frontend integration of it, causing all payment gateways to not works.

Generally, when developing new integration points, we go with the assumption that if a developer can break code via our API, they eventually will.

ExtendRestAPI is a glorified filter, that simply tries to fix the above problems, without using filters, because if you use filters, people would bypass your code and directly inject into the filter.

Any integration point we introduce should be:
- Lazy loaded, only executing the developer code once we need it.
- Wrapped in a try-catch, to ignore and remove the code if it has errors and only surface the error to admins.
- Type checked, so that simple issues are caught early, type errors should trigger an exception that is handled like the above.

senadir (changes requested):
I don't know if this PR is ready for review or not, I skipped JS code for now but left a comment in the PHP code so we don't invest too much time in a dead-end.

src/StoreApi/Schemas/CartSchema.php
budzanowski: I agree. My first approach was with `CheckoutSchema` but hit a wall after I realized how things are set up :)

src/StoreApi/Schemas/CartSchema.php
budzanowski: I understand the concerns and I know what issues are associated with using filters. I do honestly believe that we should not use ExtendRestAPI in such a simple scenario. This is overkill. That being said I rely on your judgment and will change the implementation.

So your proposal means that that the per extension requirement will be part of the checkout endpoint response. In the case of subscriptions it will be:
`extensions->subscriptions->payment_requirements: []` Do you think that we should indicate that a particular extension has requirements or should we parse all entries in extensions and look for a particular `key`? All the items from different plugins should be combined into one value. All of this should happen inside `useStoreCart`. Is that correct?

src/Payments/PaymentMethodRegistry.php
budzanowski: I was working on a change that required yet another filter in this place for filtering payment gateway features. You can check it out it here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3719/commits/a59aa05ed1cd41ace8fd67f2002e0614e0d1eaf4 Such filter is unavoidable if we want to allow PayPal integration by subscriptions. PayPal integration does not support payment gateways features required by subscriptions. Subscriptions plugin enhances the PayPal integration in a way it starts to support `subscriptions` features - that is very convoluted and it took me some time to figure it out.
I was thinking about the implementation over the weekend and I think that we should move the filter out of this place and move it just to our PayPal integration code. This will free up the logic - we will be able to remove `get_supported_features` and utilize `get_payment_method_data` for this.

src/Payments/PaymentMethodRegistry.php
```diff
@@ -153,7 +153,11 @@ public function get_all_registered_script_data() {
 		$payment_methods = $this->get_all_active_registered();
 
 		foreach ( $payment_methods as $payment_method ) {
-			$script_data[ $payment_method->get_name() . '_data' ] = $payment_method->get_payment_method_data();
+			$script_data[ $payment_method->get_name() . '_data' ] =
+				array_merge(
+					$payment_method->get_payment_method_data(),
+					array( 'supports' => apply_filters( 'woocommerce_blocks_payment_gateway_features_list', $payment_method->get_supported_features(), $payment_method->get_name() ) )
```
opr: This is a nitpick I suppose but should this be prefixed as `__experimental_`?

budzanowski: Noted.

src/StoreApi/Schemas/CartSchema.php
senadir: > checkout endpoint response. In the case of subscriptions it will be: `extensions->subscriptions->payment_requirements: []`

No, the data will always stay here as is, the API could be shaped like this

```
$extend->register_payment_requirements( [ 'subscriptions', 'multiple_ subscriptions' ] );
```

senadir: so instead of people calling the filter, they call the above code, have in run inside `woocommerce_blocks_loaded` action or something, just like the rest of API.

budzanowski: Done. I used your suggestion from Slack about how to shape this. It does not care about the namespace or the endpoint. I don't think that those parameters matter in this scenario. Should they be added just in case?

src/Domain/Services/ExtendRestApi.php
```diff
@@ -135,6 +142,60 @@ public function get_endpoint_data( $endpoint, array $passed_args = [] ) {
...
+	 *
+	 * @throws Exception On failure to register.
+	 * @return boolean True on success.
+	 */
+	public function register_payment_requirements( $args ) {
+		if ( ! is_callable( $args['data_callback'] ) ) {
+			$this->throw_exception( '$data_callback must be a callable function.' );
+		}
+
+		$this->payment_requirements[] = $args['data_callback'];
```
senadir, line 162: Why are we using callbacks if we're not passing any variables, for me it would make more sense for this to accept a string or an array of strings.

src/Payments/Integrations/PayPal.php
```diff
@@ -73,6 +71,17 @@ public function get_payment_method_data() {
 		return [
 			'title'       => $this->get_setting( 'title' ),
 			'description' => $this->get_setting( 'description' ),
+			'supports'    => apply_filters( '__experimental_woocommerce_blocks_payment_gateway_features_list', $this->get_supported_features(), $this->get_name() ),
```
senadir: Wouldn't be more cleaner to have this filter inside `get_supported_features`?

Apart from that, this is still fuzzy to me, why does WCS has to declare support? how is it done in Stripe? and why can't we do it the same way it is in Stripe?

budzanowski: PayPal does not support subscriptions on its own. The only way it supports it is by all the thins Subscription plugin adds on top of it to make it work with subscription payments. Without the Subscription plugin, PayPal only supports products and refunds:
https://github.com/woocommerce/woocommerce/blob/a76f9d9096b54c4a8e4d206c52ad5b3764b46d66/includes/gateways/paypal/class-wc-gateway-paypal.php#L34-L37
This is why the plugin injects the additional capabilities and since we are using the capabilities check we need a place where we can capture the capabilieties.

budzanowski: > Wouldn't be more cleaner to have this filter inside get_supported_features?

I think that this is the same level of clean to have it here but I am happy to move it if you think that it will be better.

src/Domain/Services/ExtendRestApi.php
budzanowski: We are using callbacks because we want the code to be only executed if it is needed. This way we only fetch the values when we prepare the response. That is what I understand from your tips:
> Lazy loaded, only executing the developer code once we need it.

Besides, it is nice that the signature for this is similar to the other extension point that we have. It makes things less confusing.

senadir: I see your point, in my mind, adding a requirement is a static task, but it actually requires some logic and queries on the current state of things (cart, order, settings...) so a callback is good.

src/Payments/Integrations/PayPal.php
senadir: I don't have strong feelings here, but I think in Core, the filter is applied before the return statement of a function, this means if we wanted to reach those values again in another place, you would simply call the function instead of the function + filter.
Unless you think the only place the filter should be applied is in that line alone, in which case it would make sense to leave it there, but I don't think this is the case.

senadir (approved):

senadir:
@budzanowski Just making sure that nothing is blocking you here? I know that you're busy with porter currently so feel free to get back to this when you're available or unassign it if you don't know when you might get to it.

budzanowski:
@senadir I just need to find a spare 2-3 hours to finish this :) I know what needs to be done:
- small code adjustment around the filter
- on or two typedef fixes
- rebase
- merge

src/StoreApi/Schemas/CartSchema.php
mikejolley: Shouldn't this be an enum field?

PR #3720, opened 2021-01-21
Move error to admin notice

senadir:
This moves the circular dependency warning to an admin notice.

A warning took a lot of screen, modified headers, and broke down POST requests.

### How to test:
- run `npm start` to generate the circular dependency.
- On cart, checkout, or any frontend page, you shouldn't see any error.
- in admin, you should see the error notice.
- run `npm run build` the error is now gone.


budzanowski (approved):
Nice.

PR #3721, opened 2021-01-22
Decode fees name so special characters are rendered correctly

Aljullu:
### Screenshots
_Before:_
imatge

_After:_
imatge

1. Add this code snippet to any PHP file (ie: `/woocommerce-gutenberg-products-block.php`):
```PHP
add_action( 'woocommerce_cart_calculate_fees', 'add_fees', 10 );
function add_fees( $cart ) {
	$cart->add_fee( __( 'Fee - Number 1', 'woo-gutenberg-products-block' ), 100, true );
	$cart->add_fee( __( 'Fee - Number 2', 'woo-gutenberg-products-block' ), 100, true );
}
```
2. Add a product to yout cart and visit the Cart or Checkout pages.
3. Verify the dash is rendered correctly.

### Changelog

> Ensure special characters are displayed properly in the Cart sidebar.

### DIFF
--- a/assets/js/base/hooks/cart/use-store-cart.js
+++ b/assets/js/base/hooks/cart/use-store-cart.js
@@ -118,9 +118,9 @@ export const useStoreCart = ( options = { shouldSelect: true } ) => {
 			const shippingAddress = cartData.needsShipping
 				? decodeValues( cartData.shippingAddress )
 				: billingAddress;
-			const cartFees = Array.isArray( cartData.fees )
-				? cartData.fees.map( ( fee ) => decodeValues( fee ) )
-				: [];
+			const cartFees = cartData.fees.map( ( fee ) =>
+				decodeValues( fee )
+			);
 			return {
 				cartCoupons: cartData.coupons,
 				cartItems: cartData.items || [],
--- a/assets/js/base/hooks/cart/use-store-cart.js
+++ b/assets/js/base/hooks/cart/use-store-cart.js
@@ -27,8 +27,8 @@ const defaultBillingAddress = {
 	phone: '',
 };
 
-const decodeAddress = ( address ) =>
-	mapValues( address, ( value ) => decodeEntities( value ) );
+const decodeValues = ( object ) =>
+	mapValues( object, ( value ) => decodeEntities( value ) );
 
 /**
  * @constant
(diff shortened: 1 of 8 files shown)


mikejolley:
@Aljullu Do you think this should be at cart hook level, so the data is correctly decoded for all consumers? We do that with addresses and error messages.

Aljullu:
@mikejolley absolutely, good point! Fixed in e7f2a8cb5deba300d07bb7682ad485a129656e43.

assets/js/base/hooks/cart/use-store-cart.js
```diff
@@ -113,14 +113,17 @@ export const useStoreCart = ( options = { shouldSelect: true } ) => {
...
 			const { receiveCart } = dispatch( storeKey );
-			const billingAddress = decodeAddress( cartData.billingAddress );
+			const billingAddress = decodeValues( cartData.billingAddress );
 			const shippingAddress = cartData.needsShipping
-				? decodeAddress( cartData.shippingAddress )
+				? decodeValues( cartData.shippingAddress )
 				: billingAddress;
+			const cartFees = Array.isArray( cartData.fees )
+				? cartData.fees.map( ( fee ) => decodeValues( fee ) )
+				: [];
```
ralucaStan: Don't think this isArray check is necessary, as `fees` will always be coming back as an array. For example it is `[]` when no fees are available.

ralucaStan (approved):
I tested it and it works as expected. I left a small comment, pre approving. 👍

assets/js/base/hooks/cart/use-store-cart.js
Aljullu: Hmmm, I'm not sure about this one. Cart data comes directly from the server, so there is the danger that the data we receive is not in the form we expect it, so I think it's good to have this kind of protections? At least, we have the same for `cartItems`, `cartItemErrors`, `extensions` and `shippingRates`:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f880fdbe9cfb00c38854a5d2647b64fab435699c/assets/js/base/hooks/cart/use-store-cart.js#L122-L135

However, I also noticed that removing this protection caused several tests to fail. In this case, I think the issue was with tests and I fixed them here 6247633bd973e15168059e96a04762cb125de6c9.

Let me know if you agree with the resolution @ralucaStan.

mikejolley (approved):
LGTM 👍🏻  Prefer handling the array rather than null to match our schema.

PR #3722, opened 2021-01-22
Fix image link in all products block.

### LINKED ISSUE #1654: All Products block: invalid link when image has no alt text
axe shows this error when using the _All Products_ block and at least one of the product images has no `alt` text:

> Ensures links have discernible text

Ideally we should:
* Add `aria-hidden` to the link that wraps the image when the image has no `alt` text and the product card has the `ProductTitle` block.
* Add a placeholder `alt` text when the image has no `alt` and the product card doesn't have the `ProductTitle` block.
* Do nothing when the image has an `alt` text.

ralucaStan:
Changes introduced by this PR:
- add the product name as alternative text to the product image
- prevent rendering two image tags when the product is missing an image. (show only the placeholder)
- add a data-label to anchors for products without a product image

Fixes #1654 

#### Accessibility

1. Add an All product block and make sure you have at least a product without an image
2. Use slow 3G Network setting to see that the placeholder image is loaded until the product image is loaded
3. Reset Network setting and check that product images have alt attributes
4. Check that a product without an image displays a placeholder image, with no alt attribute

### Changelog

> Fix - Improved accessibility of product image links in the products block by using correct aria tags and hiding empty image placeholders.


assets/js/atomic/blocks/product-elements/image/block.js
```diff
@@ -69,37 +81,21 @@ const Block = ( {
...
-							product={ product }
-						/>
-					) }
-					<Image
-						image={ image }
-						onLoad={ () => setImageLoaded( true ) }
-						loaded={ imageLoaded }
-						showFullSize={ imageSizing !== 'cropped' }
+			<ParentComponent { ...( showProductLink && anchorProps ) }>
+				{ !! showSaleBadge && (
```
budzanowski, line 84: I see that it was here before but `showSaleBadge` has `PropTypes.bool` so I don't see a reason for `!!` maybe we can remove it?

assets/js/atomic/blocks/product-elements/image/block.js
```diff
@@ -29,10 +29,11 @@ import './style.scss';
  * @param {string} [props.saleBadgeAlign] How should the sale badge be aligned if displayed.
  * @return {*} The component.
  */
-const Block = ( {
+export const Block = ( {
 	className,
 	imageSizing = 'full-size',
-	productLink = true,
+	// @todo Rename productLink to showProductLink to better reflect it's a Boolean
```
budzanowski: Do we still need this todo?

budzanowski:
I see that in the snapshot and in the tests the URLs are pointing to:
`http://localhost:10003/wp-content/uploads/2021/01/...` If I am not mistaken that is your Local setup. I don't really think that it is available in our continuous environment.
I see how the tests work( I think :) ), you fire the `fireEvent.load( image );` manually - this triggers the logic because on its own it would not work because the images are not there in the tests setup. It would probably be quite hard to change the tests to actually load images and test the logic. If that is the case then I would suggest changing the URLs ( because it looks like they don't really matter without the manual trigger logic ) in the tests and snapshots to something that would feel test environment agnostic because if someone would like to work with those test in the future they will have a hard time grasping what is `localhost:10003`

budzanowski:
This looks great in general. Nice job.

assets/js/atomic/blocks/product-elements/image/block.js
ralucaStan: I think this should still be handled at project level, as this prop name is used in other parts as well. For example, the component receives `showSaleBadge` and `productLink`. Both are booleans, and trigger a different rendered output. But the shape of the prop name is different and I found that confusing. With the current name, `productLink` appears to be a string, an actual link value.

Do you think I should update the comment with more details?

ralucaStan:
> I see that in the snapshot and in the tests the URLs are pointing to:
> `http://localhost:10003/wp-content/uploads/2021/01/...`

Good catch, I did not notice this. I am thinking that it's actually better here to write some assertions and remove the snapshot testing, instead of just changing that image source.

assets/js/atomic/blocks/product-elements/image/block.js
budzanowski: I would remove `@todo` here and just create a simple issue and tag it as a refactor. You can quickly explain what is it about there.

ralucaStan:
changes pushed. I have no experience with React Testing Library so please feel free to comment on improving the tests.

budzanowski (approved):
Nice! :ship

assets/js/atomic/blocks/product-elements/image/block.js
ralucaStan: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3756

assets/js/atomic/blocks/product-elements/image/block.js
ralucaStan: [Issue](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3756) created

PR #3732, opened 2021-01-25
Honor hidden property of cart item data and add support for experimenal property

Aljullu:
This PR makes it so:
* If a cart item data has the `hidden` property set to true, it will be hidden.
* `hidden` can be overridden with the `__experimental_woocommerce_blocks_hidden` property, which only has effect in blocks.

### Screenshots
imatge

imatge

0. Checkout `fix/details-disappearing` branch in WC Subscriptions (see relevant PR: 3968-gh-woocommerce/woocommerce-subscriptions).
1. Add a subscription with free trial and signup fee to your cart.
2. Go to the Cart block and verify the free trial and signup fee details are rendered.
4. Open the Cart shortcode and verify free trial and signup fee metadata is not rendered.

### Changelog

> Hidden cart item meta data will not be rendered in the Cart and Checkout blocks.


assets/js/base/components/cart-checkout/product-details/index.js
```diff
@@ -12,6 +12,8 @@ import './style.scss';
 
 // Component to display cart item data and variations.
 const ProductDetails = ( { details = [] } ) => {
+	details = details.filter( ( detail ) => ! detail.hidden );
```
opr, line 19: Are we worried about checking if `details` is arraylike here? I know it defaults to one, but what if a non-arraylike is passed in?

opr (approved):
Tests great, approving 👌🏼  I left a comment, feel free to address it if you think it is necessary to do so.

assets/js/base/components/cart-checkout/product-details/index.js
Aljullu: Good point. I also took the chance to add snapshot testing.

opr: Cool! This looks great now 🙌🏻

PR #3734, opened 2021-01-26
Add subtotal to cart and checkout and update the CartLineItem component with new styles

### LINKED ISSUE #3629: Add subtotal to Cart and Checkout
Based on the new design iteration for Cart and Checkout, we need to add subtotal below Product title in cart and checkout.

> Display the single item price under the product title. This was added to improve visibility into how product totals are calculated. This will also simplify how we display prices for subscription products

figma file: n90hX6hG3wwQbYwsYsrJnj-fi-5119%3A0

screenshot
image
image
image

### LINKED ISSUE #3631: Remove quantity column in Cart and move Quantity selector and delete button to below Product Title
Based on the new design iteration for Cart and Checkout, we need to move the quantity selector and the remove link to below the price.

> The Quantity column was removed to provide more space for the product title, description, price etc.
>
> The quantity selector was moved to the bottom of the left column.
>
> Mobile considerations:
> In the cart, the delete icon is removed from the upper right corner, and the “remove” link is added under the quantity selector to match desktop. This provides additional space for the product title and price.
>
> The quality select should be aligned to the right of the image.

screenshot
image
image

opr:
As per the designs in this figma file: eaQcoMs7E8Ys0YgxfkOKxR-fi-0%3A1 this PR:

- Moves the quantity picker to below the product metadata in the cart block
- Removes the trash icon on mobile in the cart block and lets the "remove item" text be visible on mobile
- Adds a sale badge under the item total when the quantity is more than one and the item is on sale.
- Removes the link to the product on the checkout sidebar.
- Shows the individual item price below the product name in the checkout sidebar.
- Moves the "total" of each product (`individual total * quantity`) to its own "column" on the sidebar.

I have added some design changes here because adding the subtotal below the item name without making the other design changes caused this to look very messy.

Fixes #3629
Fixes #3631

#### Accessibility

#### Before
Screenshot 2021-01-25 at 17 10 55

#### After 
Screenshot 2021-01-25 at 17 09 58

#### Before
Screenshot 2021-01-26 at 11 16 40

#### After
Screenshot 2021-01-26 at 11 45 58

1. Add items to your cart, some with sale prices, add multiple items etc.
2. See that the subtotal for each item is shown below the item's name, and the overall total is shown to the right. (Overall total is `item price * quantity`).
3. Try different themes out to see how the new design works there.

### Changelog

> Design tweaks to the cart page which move the quantity picker below each cart item and improve usability on mobile.


assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -34,59 +33,71 @@ const OrderSummaryItem = ( { cartItem } ) => {
...
 	} )
-		.multiply( quantity )
 		.convertPrecision( currency.minorUnit )
 		.getAmount();
-	const isProductHiddenFromCatalog =
-		catalogVisibility === 'hidden' || catalogVisibility === 'search';
+	const convertedLinePrice = Dinero( {
+		amount: parseInt( prices.raw_prices.price, 10 ),
+		precision: parseInt( prices.raw_prices.precision, 10 ),
+	} ).convertPrecision( currency.minorUnit );
```
Aljullu: Are there any dangers on using `.convertPrecision()` here and then using this value to calculate `linePrice`? I feel we should do all calculations first and only call `.convertPrecision()` when the value is going to be rendered, but not before it's used in further calculations.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -34,59 +33,71 @@ const OrderSummaryItem = ( { cartItem } ) => {
...
-				<div className="wc-block-components-order-summary-item__quantity">
-					<Label
-						label={ quantity }
-						screenReaderLabel={ sprintf(
-							/* translators: %d number of products of the same type in the cart */
-							__( '%d items', 'woo-gutenberg-products-block' ),
-							quantity
-						) }
-					/>
+			<div className="wc-block-components-order-summary-item__product-details">
```
Aljullu: Is this div needed for something?

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -8,7 +8,6 @@ import QuantitySelector from '@woocommerce/base-components/quantity-selector';
 import ProductPrice from '@woocommerce/base-components/product-price';
 import ProductName from '@woocommerce/base-components/product-name';
 import { useStoreCartItemQuantity } from '@woocommerce/base-hooks';
-import { Icon, trash } from '@woocommerce/icons';
```
Aljullu, line 11: I think the [trash icon](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/icons/library/trash.js) can now be removed.

assets/js/base/components/cart-checkout/order-summary/style.scss
```diff
@@ -69,7 +73,6 @@
 
 .wc-block-components-order-summary-item__description {
 	padding-left: $gap-large;
-	padding-top: $gap;
 	padding-bottom: $gap;
 	line-height: 1.375;
```
Aljullu: I wonder if we should remove this `line-height` rule. Currently the product name and price are not vertically aligned because of it:
imatge

assets/js/base/components/quantity-selector/style.scss
```diff
@@ -12,12 +12,13 @@
 
 .wc-block-components-quantity-selector {
```
Aljullu, line 13: Nitpick: Quantity selector is closer to the product description than to the `Remove item` link.
imatge

Aljullu (commented):
Lots of improvements in this PR, good job @opr! I left some inline comments, I think they are quite minor but posting them here to know your thoughts.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
opr: I added it to allow the "left column" (product image, name, subtotal, details) to be separated from the "right column" (the total price), but I see that it's actually not necessary, thanks for pointing it out.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
opr: Good point, I have fixed this. Thanks 😸

assets/js/base/components/cart-checkout/order-summary/style.scss
opr: Yep, looks nicer without 👍🏻

assets/js/base/components/quantity-selector/style.scss
opr: To fix this I've removed the margin from the `last-of-type` of `.wc-block-components-product-details` and added margin to the bottom of `.wc-block-components-product-metadata` - do you think this was the right approach? I had to add `0.75em` of margin because the Remove item text is quite small, so the gap looked bigger than the gap above.

assets/js/base/components/cart-checkout/product-details/style.scss
```diff
@@ -3,6 +3,10 @@
 	list-style: none;
 	margin: 0.5em 0;
 	padding: 0;
+
+	&:last-of-type {
+		margin: 0;
```
Aljullu: I'm just curious, why removing also the top margin? Shouldn't this be `margin-bottom: 0`?

Aljullu (approved):
Thanks for the updates @opr! I added one last question about the details margin, but LGTM.

assets/js/base/components/cart-checkout/product-details/style.scss
opr: It was just because the previous `product-details` element has a margin-bottom, and so the margin-top "overlaps" and doesn't actually do anything. I figured we could save a couple of bytes by doing it this way.

Aljullu: Oh, but if there is only one product details list, then there is no margin-top, no? I think this is reproducible if you compare a variable subscription (has two details list) with a simple subscription (has one details list). The margin between the product description and the first details list will be different.

Aljullu:
@opr one more thing that I completely forgot about! :facepalm: We will need to update the skeleton markup. Currently when loading it shows the layout with the quantity selector in the middle column:

https://user-images.githubusercontent.com/3616980/105971758-2efd3980-608b-11eb-85b9-35126cb7fa17.mp4

The markup can be found here:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/96ef975f8cdc0cf6af07eff2e0ae284b39aa0ec7/src/BlockTypes/Cart.php#L141

assets/js/base/components/cart-checkout/product-details/style.scss
opr: Ah yea, you're correct. Thanks for pointing this out 🤦🏻‍♂️ fix pushed.

Aljullu (approved):
LGTM :shipit:

PR #3737, opened 2021-01-26
Hide All Products Block from Widget Areas

### LINKED ISSUE #3724: Changes for Block Based Widget Editor
WordPress 5.7 is currently slated to surface the new [Blocks Widget Editor](https://github.com/WordPress/gutenberg/projects/27) in WordPress core. While there's still some uncertainty whether that will happen, this Epic collects all the issues and work slated for work on preparing Woo Blocks for this change.

Other Resources:

- [Overview issue for Block Widgets](https://github.com/WordPress/gutenberg/issues/24687)
- P2 Post: pca54o-KL-p2

### LINKED ISSUE #3726: Hide All Products Block from usage in Block Widget editor.
Currently there are some issues when the All Products block is added in a widget area and also when in the post content of a page showing the widget:

- pagination doesn't sync between the two instances
- sort select doesn't sync between the two instances

Currently we only allow one instance of the All Products block in post content because we haven't yet implemented behaviour to support multiple instances in the same view. Until we can work out support for this and how we want things to behave, the short term solution is to disallow the All Products block from being used in a widget area.

It's unclear whether there is technical ability to do this currently. If Gutenberg doesn't have an API interface that allows this type of control, then we'll have to work with the GB team to see if we can add it.

### LINKED ISSUE #3727: Hide all filter blocks from Block Widget Editor
Currently the filter blocks will not have any impact on the default shop page in WooCommerce core. They also do not have feature parity with the filter widgets that do impact the shop page. To prevent user confusion, for the short term we need to hide the filter blocks from the Block Widget editor (and they'll only be available for use in block based themes or post_content).

This includes:

- Filter Products by Price
- Filter Products by Attribute
- Active Product Filters

Similar to #3726, if the functionality for hiding blocks just in the block widget editor is not available in Gutenberg, then we will need to work with the team to add that functionality.

mikejolley:
This PR offers a simplistic solution to hiding certain blocks in the new Widget Areas that are part of the Gutenberg Feature Plugin, and coming in WordPress <del>5.7</del> <ins>5.8</ins>.

The new widget functionality does allow legacy widgets to be removed via the `widgets_to_exclude_from_legacy_widget_block ` filter, but does not have a filter for blocks. This probably makes sense because blocks are the future, and widgets are legacy.

Originally I thought we'd be able to utilise the `allowed_block_types` filter, however on inspection this filter is only part of the editor for post types, [ran here](https://github.com/WordPress/WordPress/blob/b3b8942dfcb451eddb5559b63c1043fce5d9449e/wp-admin/edit-form-blocks.php#L161).

The new widget areas page has no such filter,[ the page content being generated from here](https://github.com/WordPress/gutenberg/blob/b529dd7a45cd8a3b4ecd2c7921795b6c6db88d5e/lib/widgets-page.php#L13). There is also no block level 'supports' flag we can use either.

So instead, we need to determine if we're on the widgets screen in WP Admin and remove the blocks that way. Blocks are registered on `init` which is too early for `get_current_screen`, but we do have access to the `$pagenow` global. If `$pagenow` is themes (where the widgets page is registered) we simply do not register the blocks that do not support widgets.

Fixes #3727
Fixes #3726
Closes #3724

1. Ensure Gutenberg feature plugin is enabled and you can see the new Widget Areas screen under Appearance > Widgets.
2. Go to the screen and try to insert "All Products", "Price Filter", "Attribute Filter", and "Active Filter" blocks. You shouldn't be able to see them in the inspector!
3. Edit a page and try to insert "All Products", "Price Filter", "Attribute Filter", and "Active Filter" blocks. It should work as expected.

### Changelog

> Hide the All Products Block from the new Gutenberg Widget Areas until full support is achieved.

### DIFF
--- a/src/Library.php
+++ b/src/Library.php
@@ -40,7 +40,8 @@ class Library {
 	 * Register blocks, hooking up assets and render functions as needed.
 	 */
 	public static function register_blocks() {
-		global $wp_version;
+		global $wp_version, $pagenow;
+
 		$blocks = [
 			'AllReviews',
 			'FeaturedCategory',
@@ -68,6 +69,18 @@ class Library {
 		}
 		if ( Package::feature()->is_experimental_build() ) {
 			$blocks[] = 'SingleProduct';
+		}
+		/**
+		 * This disables specific blocks in Widget Areas by not registering them.
+		 */
+		if ( 'themes.php' === $pagenow ) {
+			$blocks_to_unset = [
+				'AllProducts',
+				'PriceFilter',
+				'AttributeFilter',
+				'ActiveFilters',
+			];
+			$blocks          = array_diff( $blocks, $blocks_to_unset );
 		}
 		foreach ( $blocks as $class ) {
 			$class    = __NAMESPACE__ . '\\BlockTypes\\' . $class;


Aljullu (approved):
Thanks for the explanation and the links @mikejolley! Ideally, I feel Gutenberg should provide an API so plugins can hide some blocks in the widget editor. Do you think we should open an issue in Gutenberg's repo and explain our usecase so this can be considered?

I took some time to research how the _More_ and _Page break_ blocks from Gutenberg are hidden just to make sure we were not missing anything. It looks like they use the [`parent` property](https://github.com/WordPress/gutenberg/blob/37163dcba1a2720522681fef5d258231d7890ea6/packages/block-library/src/nextpage/block.json#L5) to restrict where those blocks can be added. The issue with that approach is that it doesn't seem to work with reusable blocks and that we would need to explicitly declare all other parents we want to allow (ie: Empty cart block).

Given that currently there isn't a perfect solution, I think we can go with the fix in this PR. I tested it and I can confirm that All Products and filter blocks are no longer available.

If the user had one already added in a widget area, it will show as unsupported in the editor (even though it continues working in the frontend):

imatge

I think this behavior is fine.

mikejolley:
@Aljullu Thanks. I've logged an issue (link above). Parent does allow for fine grained control but you need to know all areas it could be used. I think just not loading these blocks will be better for now, and have less overhead. We shouldn't register things unless needed.

PR #3738, opened 2021-01-26
Fix color for BlockErrorBoundry

### LINKED ISSUE #3668: BlockErrorBoundary text not visible in dark themes
## To reproduce

1. Chose a dark theme or a theme that allows you changing the colors (like Storefront) and set the background to black and text to white.
2. Force an error to the Cart or Checkout block. You can do that adding this snippet to [line 74 in `assets/js/blocks/cart-checkout/cart/block.js `](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c2fa2ce5bd98e49275a27fdfedafbe2ac9d7d592/assets/js/blocks/cart-checkout/cart/block.js#L74):
```JS
throw new Error( 'test' );
```
3. Notice the error text doesn't have enough contrast:
_Dark background store:_
imatge
_Light background store:_
imatge

## Expected behavior
Text should be visible.

ralucaStan:
The new color ensures readability on both light and dark themes.
Also the PR changes the colors used in the component, using colors defined in [WordPress/gutenberg](https://github.com/WordPress/gutenberg/blob/master/packages/base-styles/_colors.scss)

Fixes #3668 

#### Accessibility

### Screenshots

Before:
imatge
After: 
Screenshot 2021-01-26 at 15 11 02

1. Chose Storefront and set the background to black and text to white.
2. Force an error to the Cart or Checkout block. Add on line 74 `throw new Error( 'test' );`  in `assets/js/blocks/cart-checkout/cart/block.js`
3. Notice the error text is now visible

### Changelog

> Fix - Set correct text color in BlockErrorBoundry notices.

### DIFF
--- a/assets/js/base/components/block-error-boundary/style.scss
+++ b/assets/js/base/components/block-error-boundary/style.scss
@@ -1,15 +1,16 @@
 .wc-block-components-error {
 	display: flex;
-	background-color: #f3f3f4;
-	border-left: 4px solid #6d6d6d;
+	background-color: $gray-100;
+	border-left: 4px solid $gray-300;
 	padding: $gap-larger $gap;
 	align-items: center;
 	justify-content: center;
 	flex-direction: column;
+	color: $gray-700;
 }
 
 .wc-block-components-error__header {
-	@include font-size(larger);
+	@include font-size( larger );
 	font-weight: bold;
 	margin: 0;
 }
--- a/assets/js/base/components/block-error-boundary/style.scss
+++ b/assets/js/base/components/block-error-boundary/style.scss
@@ -10,7 +10,7 @@
 }
 
 .wc-block-components-error__header {
-	@include font-size( larger );
+	@include font-size(larger);
 	font-weight: bold;
 	margin: 0;
 }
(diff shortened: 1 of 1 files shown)


assets/js/base/components/block-error-boundary/style.scss
```diff
@@ -1,15 +1,16 @@
...
 	padding: $gap-larger $gap;
 	align-items: center;
 	justify-content: center;
 	flex-direction: column;
+	color: $gray-700;
 }
 
 .wc-block-components-error__header {
-	@include font-size(larger);
+	@include font-size( larger );
```
opr: Values in parentheses shouldn't be padded according to the [WordPress coding standards](https://make.wordpress.org/core/handbook/best-practices/coding-standards/css/#values)

opr (approved):
This tests perfectly, great work @ralucaStan! I left a comment about parentheses that should be fixed, but pre-approving anyway since this isn't a blocker.

assets/js/base/components/block-error-boundary/style.scss
mikejolley: Prettier should catch this @ralucaStan so might be another IDE config issue.

PR #3749, opened 2021-01-27
Update @wordpress/scripts and eslint and stylint configurations

nerrad:
In #3742 Renovate automatically updated the `@woocommerce/eslint-plugin` to the latest available version, however the indirect dependency for `@wordpress/eslint-plugin` remained at an earlier version than what is used by `@wordpress/eslint-plugin` because `@wordpress/scripts` also depends on `@wordpress/eslint-plugin`.  So in order to ensure our linting is running on the latest ruleset exposed by `@wordpress/eslint-plugin`, `@wordpress/scripts` needed updated as well.

I've left some inline comments with the code to outline why the different changes. 

## To test

- All GitHub action code checks should pass.
- IDE (Vscode) linting should work okay (with prettier too!). Also not sure if other folks are using it but I didn't realize that I could install [this Vscode extension](https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint) to get handy styelinting in the code.


.eslintrc.js
```diff
@@ -25,6 +25,11 @@ module.exports = {
 		'react-hooks/exhaustive-deps': 'error',
 		'react/jsx-fragments': [ 'error', 'syntax' ],
 		'@wordpress/no-global-active-element': 'warn',
+		// @todo Remove temporary conversion of eslint rules to warn instead of error.
+		//       These are new rules surfacing as a result of a eslint plugin update and the intent
+		//       is to update them in their own individual prs to make for easier review.
+		'import/no-extraneous-dependencies': 'warn',
+		'import/no-unresolved': 'warn',
```
nerrad, line 32: Temporarily set as warnings for now. We'll have to return to errors in follow-ups. In particular the `import/no-unresolved` rule seems to fail with all of our aliased packages, so that may be something we'll need to see if we can customize/exclude somehow.

.prettierrc.js
```diff
@@ -0,0 +1,3 @@
+// Import the default config file and expose it in the project root.
+// Useful for editor integrations.
+module.exports = require( '@wordpress/prettier-config' );
```
nerrad, line 3: This is required for IDE support.

.stylelintrc.json
```diff
@@ -1,11 +1,12 @@
 {
-	"extends": "stylelint-config-wordpress",
+	"extends": "@wordpress/stylelint-config",
```
nerrad, line 2: This was a breaking change needing fixed from the `@wordpress/scripts` update because `@wordpress/scripts` no longer bundles `stylint-config-wordpress`.

.stylelintrc.json
```diff
@@ -1,11 +1,12 @@
...
 	"rules": {
 		"at-rule-empty-line-before": null,
 		"at-rule-no-unknown": null,
 		"comment-empty-line-before": null,
 		"declaration-block-no-duplicate-properties": null,
 		"declaration-property-unit-whitelist": null,
+		"declaration-property-unit-allowed-list": null,
```
nerrad, line 9: This rule produced errors. I tried to convert to warning unsuccessfully. We'll need to handle this in a follow-up. Actually the follow-up should re-examine all of our exceptions here and see why we have them and possibly setup prettier to fix our style files. Currently I can't turn on the automattic formatter for prettier for stylesheets in my IDE.

package.json
```diff
@@ -104,7 +104,7 @@
 		"@wordpress/html-entities": "2.8.0",
 		"@wordpress/i18n": "3.15.0",
 		"@wordpress/is-shallow-equal": "1.8.0",
-		"@wordpress/scripts": "12.2.1",
+		"@wordpress/scripts": "^13.0.0",
```
nerrad, line 107: Needed to update this in order to get the latest `@wordpress/eslint-plugin` ruleset.

package.json
```diff
@@ -119,6 +119,7 @@
 		"cross-env": "6.0.3",
 		"cssnano": "4.1.10",
 		"deep-freeze": "0.0.1",
+		"eslint-plugin-import": "^2.22.1",
```
nerrad, line 122: This was a new dependency needed with the `@wordpress/scripts` update (or might be a side-effect of pulling in the updated `@wordpress/eslint-plugin` ruleset. It might be an artifact of new eslint requirements too.

nerrad: Speak of the devil: https://github.com/WordPress/gutenberg/pull/28525

budzanowski (approved):

Issue #3735, opened 2021-01-26
Integration Tracking (WC Subscriptions): create an API so extensions can hook into Cart and Checkout blocks


Aljullu:
In #3716 we added `__EXPERIMENTAL_TOTAL_LABEL_FILTER`, but we need to make sure extensions use this filter (and more filters added in currently open PRs) properly. In order to do so, we will create an API similar to the one used to register payment methods to abstract filters. We might not even need to use filters once we have this API! Lastly, we should add some protections to the returned items.

PR #3750, opened 2021-01-27
Allow extensions to extend cart item price

### LINKED ISSUE #3605: Allow extensions to modify how the product price from cart line items is displayed
Some extensions need to tweak how the price of cart line items is displayed. For example, Subscriptions needs to display a `Due today` label in some cases:

screen-shot-2020-11-30-at-8 27 23-pm

Aljullu:
Fixes #3605.

Related PR in WC Subscriptions: 3971-gh-woocommerce/woocommerce-subscriptions

This PR does a couple of things:
* Introduces a new filter (available through the `cartItemPrice` key) that allows extensions to modify the way cart item prices are displayed in the Cart and Checkout blocks. Notice this filter passes a string and expects a string, so extensions can add text before or after the price, but they can't replace it with another component. There is also a check to verify the filtered string contains the price, so extensions can't remove it by mistake.
* In f51199b854a03cfc35d40b3d5a8f16c75642f5d7, it changes the value used to display the cart item price. Until now, we were using the formula `product_price * quantity`. But we already get a `totals` property from the API with the value calculated so we can use that one. This has a couple of consequences:
  * Subscriptions sets the cart item total to the sign-up fee. That was lost in blocks because we were ignoring the total and instead were calculating it in the client side. With this PR it should be fixed.
  _(Sidenote: the client side calculation of cart item totals was also problematic in PAO because it was ignoring the price of the add-ons)_
  * :warning: Given that we are no longer making the item price calculation client side, optimistic updates no longer work. Debounce is still there, only change is that optimistic updates will not be visible.

I'm adding the :warning: emoji because this is a feature loss. I think optimistic updates need to be re-thought from the base. The way it worked until now was showing the user a value that was different from the one in the server. IMO this is very dangerous and has proven problematic in Subscriptions and PAO so far.

### Screenshots

imatge

imatge

1. Checkout out this branch in WC Subscriptions: `fix/cart-item-price-filter`.
2. Add a Subscription product to your cart with a sign-up fee.
3. Go to the Cart and Checkout blocks and verify the price for that cart item is shown as `Due today $1.23` (in Cart) and `$1.23 due today` (in Checkout).
4. Test with Subscriptions without sign-up fee, with regular products, etc. to verify all cases work fine.


assets/js/base/components/product-price/index.js
```diff
@@ -119,12 +119,26 @@ const ProductPrice = ( {
 		}
 	);
 
+	if ( ! format.includes( '<price/>' ) ) {
+		format = '<price/>';
+		// eslint-disable-next-line no-console
+		console.error( 'Price formats need to include the `<price/>` tag.' );
```
Aljullu: I'm planning to improve this and other protection checks in #3735.

senadir:
@Aljullu Is it possible to have this PR on hold until we reach a consensus to how are we going to handle optimistic updates?

Aljullu:
PR updated so it uses the extendibility API introduced in #3774.

assets/js/base/components/product-price/index.js
senadir: This should now be removed?

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -100,16 +112,31 @@ const CartLineItemRow = ( { lineItem = {} } ) => {
...
+	const productPriceFormat = __experimentalApplyCheckoutFilter( {
+		filterName: 'cartItemPrice',
+		defaultValue: '<price/>',
+		arg: {
+			cartItem: lineItem,
+			block: 'cart',
+		},
+		validation: ( value ) =>
+			typeof value === 'string' && value.includes( '<price/>' ),
+	} );
```
senadir, line 138: I'm going to update this in a follow up.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -73,6 +76,18 @@ const OrderSummaryItem = ( { cartItem } ) => {
...
+	const productPriceFormat = __experimentalApplyCheckoutFilter( {
+		filterName: 'cartItemPrice',
+		defaultValue: '<price/>',
+		arg: {
+			cartItem,
+			block: 'checkout',
+		},
+		validation: ( value ) =>
+			typeof value === 'string' && value.includes( '<price/>' ),
+	} );
```
senadir, line 89: I'm going to update this in a follow up.

senadir (approved):

PR #3755, opened 2021-01-27
Enable defaults for new import rules and configure

### LINKED ISSUE #3753: Remove temporary conversion of eslint rules to warn inste...
Remove temporary conversion of eslint rules to warn instead of error. These are new rules surfacing as a result of a eslint plugin update and the intent is to update them in their own individual prs to make for easier review.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a112796829f5f13a70f2f495ff7737f54dbcd92c/.eslintrc.js#L28-L35
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in a112796829f5f13a70f2f495ff7737f54dbcd92c when #3749 was merged. cc @nerrad

nerrad:
Fixes: #3753 

In this pull request, the `import/no-unresolved` and `import/no-extraneous-dependencies` rules defaults are enabled and configuration for adding exceptions for both webpack config resolution and externals or indirect dependencies are added.

With our webpack configuration, there are a number of aliases and externals we define that aren't resolved with the default rule configuration, so the changes here make sure _known_ exceptions are accounted for.

## To test

If GitHub workflow checks pass without error, this can be merged.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -48,7 +48,7 @@ module.exports = {
 			'requireindex',
 		],
 		'import/resolver': {
-			Node: {},
+			node: {},
 			webpack: {},
 		},
 	},
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,17 +19,44 @@ module.exports = {
 	},
 	settings: {
 		jsdoc: { mode: 'typescript' },
+		// List of modules that are externals in our webpack config.
+		// This helps the `import/no-extraneous-dependencies` and
+		//`import/no-unresolved` rules account for them.
+		'import/core-modules': [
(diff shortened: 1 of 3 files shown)


nerrad:
whew finally. Lots of weird things happening with the lint config locally vs GitHub actions but I think I was experiencing a red herring locally because initially having `node: {}` didn't work for me locally so I didn't think it'd resolve the issues in the GitHub workflow. I think I just had a bad `node_modules` or something though because i wiped it and re-installed and all was good 🤷

budzanowski (approved):

PR #3766, opened 2021-01-29
Force api call to "store/products" to refresh on browser back button click.

### LINKED ISSUE #3757: Nonce validation issue: "X-WC-Store-API-Nonce is invalid" in All Products Block.
## Describe the bug
When using All Products Block it is possible to hit:
image

Related ticket: 3688475-zd-woothemes
## To reproduce

1. Open incognito window
2. Go to a page with All Products block
3. Add an item to the cart 
4. Go to cart
5. Go back using the browser back button
6. Try to add another item to the cart: not possible + nonce notice

## Expected behavior
The item should be added to the cart and the error notice should not be shown.

* Core version: [e.g. 5.6]
* Gutenberg version (if installed): none
* WooCommerce version: 4.9.0 - 5.0.0-rc1
* WooCommerce Blocks version: none
* Site language: English.
* Any other plugins installed: nope

* OS: macOS
* Browser Chrome
* Version 87.0.4280.141 (Official Build) (x86_64)

## Additional context
This was also tested with WooCommerce 4.8 and the problem was not present there.

budzanowski:
This PR sets up the Cache-Control header for the `store/products` route. This is now set up in a way that tells the browser not to use that request from cache. 

Why was it failing in the first place? Long explanation that I have used for myself to understand what is going on.

When a **fresh ( no store session ) logged-out user** navigates to a page with All Products block the site is loaded with a nonce for a logged-out user. This nonce is store as a global JS value. After adding an item to the car ( using the nonce ), WooCommerce creates a session. This changes how the nonce is generated for the user. This is because of the `nonce_user_logged_out` filter added in WooCommerce that modifies the user-id used for the nonce generation. When a sessing is established this filter substitutes, logged-out user-id, with user-id randomly generated when the session was created, and stored in the COOKIE. This means that after the user adds something to the cart he becomes identified as a different user for the purpose of nonce creation and validation. All Products block still works because in the response to `Add to Cart` we are sending and updating this new nonce - so any next requests can still be performed. Now the user navigates to a different page, next he clicks the browser navigation back button. The page with the All Products block is loaded. But this time not entirely from the server - most of the values come from the browser cache. This means that we are using the nonce for the **fresh ( no store session ) logged-out user** stored in global JS value. That is no longer the correct value because the user has a session so the nonce validation will look for a different user.

We could force the reload of the whole page but that would be not optimal. Instead, we can tell the browser no to cache the `store/products` API requests that All Products is doing to get the products. This API request was also cached by the browser. By disabling it we ensure that although the page body is from the cache and initially has the old nonce after the API call finishes the nonce will be updated to the proper one. From my observation, this has a performance impact only for the browser back button use - the user needs to wait for the product's API to respon.

Fixes #3757

Follow the instruction in #3757 and verify that the issue is no longer there.

### Changelog

### DIFF
--- a/src/StoreApi/Routes/Products.php
+++ b/src/StoreApi/Routes/Products.php
@@ -67,6 +67,14 @@ class Products extends AbstractRoute {
 		$response = ( new Pagination() )->add_headers( $response, $request, $query_results['total'], $query_results['pages'] );
 		$response->header( 'Last-Modified', $product_query->get_last_modified() );
 
+		/*
+		 * Setup Cache-Control not to cache this request response.
+		 * This forces the browser to re-do this request when user clicks the navigation back button.
+		 * This in turn allows the code to send an updated store api nonce.
+		 * Without this logged out users could potentially be using stale nonce.
+		 */
+		$response->header( 'Cache-Control', 'no-store, max-age=0' );
+
 		return $response;
 	}
 


nerrad:
I've added the blocker label to this PR because I think we should make sure the fix gets in for the 4.4.0 release so it's included with WooCommerce Core 5.1.

PR #3767, opened 2021-01-29
Move "no rates found" logic to package level

### LINKED ISSUE #3728: Integration Tracking (WC Subscriptions): Subscriptions products with a trial don't return shipping rates
Discovered this while testing https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3686 and 3963-gh-woocommerce/woocommerce-subscriptions, switched to trunk and `feature/checkout-block-simple-multiple-subscriptions` to make sure it's not a regression but the issue persisted.

When you have only Subscription products with a trial on your cart, no shipping rates are returned, it doesn't matter if those products have a signup fee.
My guess is that free trials have $0 on the due date, sign up fees are counted as fees and won't trigger shipping calculation, hence the empty array.
I didn't test yet to see if this is a regression or was there from the beginning and we never saw it.
Shipping rates appear once you add a product with no trial or a regular product.

### Testing instructions:
- Add a product with a trial to your cart.
- see cart or checkout, enter your address, you won't see any rates, with an error 
image
- Add a regular product or a subscription product with no trial (anything that captures payment directly that isn't a fee), you will see rates.

mikejolley:
Two changes for compatibility with subscriptions.

1. Allow rates to be selected via the API even if shipping is not needed for the cart. Reason: Subscriptions has recurring carts which may need shipping, even if the main cart does not.
2. Move the "no rates found" notice logic to cart level. This ensures that we only see a notice when a package has no rates, not when there are no packages.

cart

Fixes #3728

Requires the feature branch: https://github.com/woocommerce/woocommerce-subscriptions/tree/feature/checkout-block-simple-multiple-subscriptions

See https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3728

### DIFF
--- a/packages/checkout/shipping/shipping-rates-control/shipping-rates-control.js
+++ b/packages/checkout/shipping/shipping-rates-control/shipping-rates-control.js
@@ -50,14 +50,7 @@ const ShippingRatesControl = ( {
 		}
 		const packageCount = getShippingRatesPackageCount( shippingRates );
 		const shippingOptions = getShippingRatesRateCount( shippingRates );
-		if ( shippingOptions === 0 ) {
-			speak(
-				__(
-					'No shipping options were found.',
-					'woo-gutenberg-products-block'
-				)
-			);
-		} else if ( packageCount === 1 ) {
+		if ( packageCount === 1 ) {
 			speak(
 				sprintf(
 					// translators: %d number of shipping options found.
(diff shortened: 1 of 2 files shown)


packages/checkout/shipping/shipping-rates-control/shipping-rates-control.js
```diff
@@ -50,14 +50,7 @@ const ShippingRatesControl = ( {
 		}
```
senadir, line 50: The text "Select shipping options below." that is shown when multiple packages are available is no longer visible when we don't have initial shipping, but that's I'm okay giving up.

The other thing is that when you have single item, you're provided two shipping options, initial one and future one, feels a bit redundant and thinking if we can skip it?

senadir (approved):
I'm seeing different styling for core shipping methods and recurring shipping methods, will investigate more, unrelated to this, this PR works fine!
image

packages/checkout/shipping/shipping-rates-control/shipping-rates-control.js
mikejolley: Does subs core skip it?

senadir: Yes.

PR #3768, opened 2021-01-31
Add TypeScript support and convert cart data store to TypeScript

nerrad:
As a companion to a proposal for starting to use TypeScript in the blocks repo, I've created this pull request. I'm not going to go into the benefits and why we'd want to switch to TS in this pr description (that's for our internal deliberation) however, they are many. One significant thing I already noticed while working on the initial conversion was some minor bugs surfaced as the types were implemented (mostly around not accounting for possible new properties in the Cart API response and some guards around unknown data shapes in the reducer).

Since TypeScript is a fairly trivial enhancement on top of native JS, it's not too difficult to pick up the syntax and should be relatively easy to adopt. 

Some notes about what's introduced in this PR:

- I reviewed a number of WordPress project configs using TS (including Calypso and other projects) and the config I landed on is one that I _think_ will be okay for Blocks. But I'm not an expert here so this is open for modification.
- I stuck with babel-loader for typescript transpilation because that allows us to more easily keep many of the other babel plugins we have in use (along with inheriting @wordpress/* babel configs). AFAICT, this is sufficient, the only caveat is that there won't be any hard ts-loader breaks on type errors during builds. I think this is sufficiently covered by the additional linting put in place however.
- Some new `@types/*` packages were installed for various types being used. 
- There's some typing I implemented that I had trouble with so I just applied `@ts-ignore` to those places for now. As we get more proficient with TS (or others more skilled take a look at the code) we can address.
- We don't have to convert everything over to TypeScript immediately - our existing code can live with the ts files indefinitely. But it is something we'll probably want to chip away at (and/or implement with new code) if we decide to go all in. The benefits will only be fully realized as we type more of our code.
- I'm finding that I really prefer full TS over the jsdoc flavored TS we were using as it's more expressive and less finagling with how to setup typedefs etc.

Some things we may want to consider:

- The benefit of `@param` and `@return` jsdocs is less useful in TypeScript files because it's kind of redundant with the existing typing. I wonder if we should remove those jsdoc rulesets for TS files (and just accept jsdoc descriptions where needed)?
- There may be other eslint rules we can relax/remove in TS files because of the gains with typing - we should watch for them.

## To Test

Anything interacting with the Cart data store will need to be tested because of some minor changes done in the PR. This pretty much means anything with the All Products block and the Cart and Checkout blocks. So full flow tests should be done. However, any problems that exist should surface immediately because of the low level impact of the changes.

## Next steps

- Decide if we want to go all in with TS
- Merge this pull and establish some guidelines for new work being written in TS.
- Slowly chip away at migrating existing code to TS (I would start with anything that is at the base level and then work down the tree - so data stores, base components/hooks/context, and then blocks).


globals.d.ts
```diff
@@ -1 +1 @@
-declare var fetchMock: any;
+declare let fetchMock: never;
```
nerrad: This is probably wrong. Should probably see if there are types for fetch mock instead.

nerrad:
Before merging, it looks like we may need to accommodate `.tsx` extensions for files containing jsx [because of some TypeScript requirements](https://www.typescriptlang.org/docs/handbook/jsx.html). While there are no jsx files in this pr, the configuration will need updated to handle them.

nerrad:
Before this pr is merged, the configuration will need updated to handle `.tsx` files (which is the extension we'd have to use for any TypeScript file containing jsx) because of [technical reasons](https://www.typescriptlang.org/docs/handbook/jsx.html):

> Recall how to write a type assertion:
> `var foo = <foo>bar;`
> This asserts the variable bar to have the type foo. Since TypeScript also uses angle brackets for type assertions, combining it with JSX’s syntax would introduce certain parsing difficulties. As a result, TypeScript disallows angle bracket type assertions in .tsx files.

assets/js/data/cart/actions.ts
```diff
@@ -2,42 +2,82 @@
...
- * @param {Object}   [response={}]    An object containing the response from the
- *                                    request.
- * @return {Object} Object for action.
+ * @param  {CartResponse}      response
+ * @return {ReceiveCartAction}          Action object.
  */
-export function receiveCart( response = {} ) {
+export function receiveCart( response: CartResponse ): ReceiveCartAction {
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+	// @ts-ignore -- unclear how to type a transform from one type to another here.
```
opr: If we change `Cart` to a `type` instead of an `interface` then we can do this:

`const cart: Cart = mapKeys( response, ( _, key ) => camelCase( key ) ) as Cart;`

If there's a reason to stick with `interface` then we can first convert to `unknown` and then to `Cart`. Since we, as devs know that a `Cart` is the same as a `CartResponse` jut with camel cased keys I think it's OK to do this.

```ts
const cart: Cart = ( mapKeys( response, ( _, key ) =>
		camelCase( key )
	) as unknown ) as Cart;
```

assets/js/data/cart/reducers.ts
```diff
@@ -1,26 +1,41 @@
...
+
+type AnyAction = {
+	type: string;
+	error?: ResponseError | null;
+	response?: Cart;
+	couponCode?: string;
+	cartItem?: CartResponseItem | null;
+	cartItemKey?: string;
+	isResolving?: boolean;
+	[ key: string ]: unknown;
```
nerrad: Probably good to remove this otherwise basically any key can be in this type. That way we have to update it if we add additional actions to the store?

opr: I agree this should be removed. I don't think it's too much work to add a new property to this type when we need to. This is speculation though, I can't imagine us adding actions to this reducer so frequently that it becomes a problem.

assets/js/data/cart/reducers.ts
```diff
@@ -1,26 +1,41 @@
...
 
 /**
  * Internal dependencies
  */
 import { ACTION_TYPES as types } from './action-types';
-import { defaultCartState } from '../default-states';
+import { defaultCartState, CartState } from '../default-states';
+import type { ResponseError } from '../types';
+
+type AnyAction = {
```
opr: Just a thought, but maybe we'd prefer a name like `CartAction` here.

assets/js/data/cart/reducers.ts
```diff
@@ -1,26 +1,41 @@
...
+	type: string;
+	error?: ResponseError | null;
+	response?: Cart;
+	couponCode?: string;
+	cartItem?: CartResponseItem | null;
+	cartItemKey?: string;
+	isResolving?: boolean;
+	[ key: string ]: unknown;
+};
```
opr: We need to add 

```ts
	isPendingQuantity: boolean;
	isPendingDelete: boolean;
``` 
to this type.

globals.d.ts
opr: I see this was originally added to stop TS linting errors in [this commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2678/commits/e3c6292bf2c520177e0306629c795d1948b97216). There is [`@types/fetch-mock`](https://www.npmjs.com/package/@types/fetch-mock). I believe it would still be necessary to `import { FetchMock as fetchMock } from 'jest-fetch-mock';` in each test file where this is used. We don't do this now, so it would require a little effort to get working. Is this a bad thing? I don't think so, it is good to be explicit about what external packages we're using.

tsconfig.json
```diff
@@ -42,8 +46,9 @@
 			"@woocommerce/knobs": [ "storybook/knobs" ],
 			"@woocommerce/settings": [ "assets/js/settings/shared" ],
 			"@woocommerce/shared-context": [ "assets/js/shared/context" ],
-			"@woocommerce/type-defs/*": [ "assets/js/type-defs/*" ]
+			"@woocommerce/type-defs/*": [ "assets/js/type-defs/*" ],
+			"@woocommerce/types": [ "assets/js/type-defs" ]
 		}
 	},
-	"include": [ "./assets/js/**/*.js", "./globals.d.ts" ]
+	"include": [ "./assets/js/**/*.js", "./globals.d.ts", "./assets/js/**/*.ts" ]
```
opr: Do we want to remove `.js` from here? It seems with the addition of the `no-implicit-any` eslint rule, the checking is kind of aggressive and in my IDE at least it just paints entire .js files red.

assets/js/data/shared-controls.ts
```diff
@@ -48,19 +59,23 @@ export const controls = {
...
+							// @ts-ignore -- this does exist but doesn't appear to be typed in the api-fetch types.
 							triggerFetch.setNonce( fetchResponse.headers );
 						} )
 						.catch( () => {
 							reject( invalidJsonError );
 						} );
 				} )
 				.catch( ( errorResponse ) => {
+					// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+					// @ts-ignore -- this does exist but doesn't appear to be typed in the api-fetch types.
```
opr, line 72: Maybe it is because we're adding this method to apiFetch? https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b8f1cd87ea21715cb5f334ef63a38e689f73c1d6/assets/js/middleware/store-api-nonce.js#L57

Not sure how we can indicate that we're extending `apiFetch` ourselves though so TS knows about it. I found this: https://stackoverflow.com/questions/38971984/how-to-add-custom-typings-in-typescript-2-0-3-0 but it doesn't seem to let me extend `@wordpress/api-fetch` in a way that stops TS showing an error... something to look at when we know a bit more about this maybe?

opr (commented):
Hey Darren this is awesome, and I'm so glad that we're taking steps toward adopting TypeScript! I've left some comments about things I've noticed while testing, and some errors I spotted.

I see your comment about adding `.tsx` support to the configs. Was there a specific reason you didn't do this already or did you just not get round to it? If you were facing issues then feel free to pass over some information and I'll take a look.

Anyway I hope my comments are useful to you, I'll comment internally on the P2 with my input to the TS consideration.

assets/js/data/cart/actions.ts
nerrad: shweet! I'll give that a go. We can probably just import the Cart interface as a type and do that then?

assets/js/data/cart/reducers.ts
nerrad: Ya that'd be better, I still like the word "any" in there to signify it could be one of many action types. So maybe `AnyCartAction`?

assets/js/data/cart/reducers.ts
nerrad: ooo good catch (and a good example why the existing catch property there is a bad idea).

globals.d.ts
nerrad: Agreed, I'll take a look at doing that in this pr.

tsconfig.json
nerrad: VScode has jsdoc typescript support so I don't get the same behaviour as you in PHPStorm. We have `.js` here to get that implicit type-checking through the jsdocs on the js files. Maybe there's something that could be setup in PHPStorm to improve that?

nerrad:
> I see your comment about adding .tsx support to the configs. Was there a specific reason you didn't do this already or did you just not get round to it?

I didn't know it'd be needed on my initial pass until I did some more reading after the fact. So I just added the comment to remind myself to do before merging.

assets/js/data/shared-controls.ts
nerrad: ohhh ya! That would be it :| - We might be able to declare the module in a `d.ts` file to override/enhance (is that what you tried?).

opr: Yep that's what I tried, I spent an hour on it last night and another hour on it now and couldn't get it to work. I think this is called module augmentation if you want to have a go at it yourself.

package.json
```diff
@@ -40,9 +40,9 @@
 		"lint:ci": "npm run lint:js && npm run lint:css",
 		"lint:css": "stylelint 'assets/**/*.scss'",
 		"lint:css-fix": "stylelint 'assets/**/*.scss' --fix",
-		"lint:js": "wp-scripts lint-js",
-		"lint:js:report": "npm run lint:js -- --output-file eslint_report.json --format json",
-		"lint:js-fix": "eslint assets/js --ext=js,jsx --fix",
+		"lint:js": "wp-scripts lint-js --ext=js,ts",
```
sirreal:
```suggestion
		"lint:js": "wp-scripts lint-js --ext=js,ts,tsx",
```

package.json
```diff
@@ -40,9 +40,9 @@
...
+		"lint:js:report": "npm run lint:js -- --output-file eslint_report.json --ext=js,ts --format json",
```
sirreal:
```suggestion
		"lint:js:report": "npm run lint:js -- --output-file eslint_report.json --ext=js,ts,tsx --format json",
```

package.json
```diff
@@ -189,7 +196,7 @@
 		"*.scss": [
 			"npm run lint:css"
 		],
-		"*.js": [
+		"*.{js,ts}": [
```
sirreal:
```suggestion
		"*.{js,ts,tsx}": [
```

bin/webpack-configs.js
```diff
@@ -123,7 +123,10 @@ woocommerce_blocks_env = ${ NODE_ENV }
 `.trim(),
 			} ),
 		],
-		resolve,
+		resolve: {
+			...resolve,
+			extensions: [ '.js', '.ts' ],
```
sirreal:
```suggestion
			extensions: [ '.js', '.ts', '.tsx' ],
```

bin/webpack-configs.js
```diff
@@ -210,7 +213,10 @@ const getMainConfig = ( options = {} ) => {
 				( result ) => ( result.resource = dashIconReplacementModule )
 			),
 		],
-		resolve,
+		resolve: {
+			...resolve,
+			extensions: [ '.js', '.ts' ],
```
sirreal:
```suggestion
			extensions: [ '.js', '.ts', '.tsx' ],
```

bin/webpack-configs.js
```diff
@@ -310,7 +316,10 @@ const getFrontConfig = ( options = {} ) => {
...
```
sirreal:
```suggestion
			extensions: [ '.js', '.ts', '.tsx' ],
```

bin/webpack-configs.js
```diff
@@ -411,7 +420,10 @@ const getPaymentsConfig = ( options = {} ) => {
...
```
sirreal:
```suggestion
			extensions: [ '.js', '.ts', '.tsx' ],
```

bin/webpack-configs.js
```diff
@@ -570,7 +582,10 @@ const getStylingConfig = ( options = {} ) => {
 			// Remove JS files generated by MiniCssExtractPlugin.
 			new RemoveFilesPlugin( `./build/*style${ fileSuffix }.js` ),
 		],
-		resolve,
+		resolve: {
+			...resolve,
+			extensions: [ '.js', '.ts' ],
```
sirreal:
```suggestion
			extensions: [ '.js', '.ts', '.tsx' ],
```

sirreal (commented):
This is a quick pass to add the `tsx` extension to places I noticed it was missing. Unlike JavaScript where `jsx` is a convention, TypeScript requires TypeScript files that include JSX to use the `.tsx` extension for [technical reasons related to some syntax](https://www.typescriptlang.org/docs/handbook/jsx.html):

> Recall how to write a type assertion:
> `var foo = <foo>bar;`
> This asserts the variable `bar` to have the type `foo`. Since TypeScript also uses angle brackets for type assertions, combining it with JSX’s syntax would introduce certain parsing difficulties. As a result, TypeScript disallows angle bracket type assertions in `.tsx` files.

**EDIT** Sorry for this [very redundant comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3768#issuecomment-770886027) 🙂

nerrad:
Thanks @sirreal for the changes (saved me a step)!

assets/js/data/cart/actions.ts
```diff
@@ -73,26 +113,31 @@ export function receiveRemovingCoupon( couponCode ) {
...
  */
-export function receiveCartItem( response = {} ) {
+export function receiveCartItem(
+	response: CartResponseItem | null = null
+): ResponseCartItemAction {
 	return {
 		type: types.RECEIVE_CART_ITEM,
 		cartItem: response,
 	};
 }
```
sirreal: I find with action creators specifically, it's nice to let TypeScript infer the types. They just return simple objects and it doesn't seem helpful to describe the shape twice. Adding the _`const` assertion_ will make sure the return type has exactly the shape we expect:

```suggestion
export function receiveCartItem(
	response: CartResponseItem | null = null
) {
	return {
		type: types.RECEIVE_CART_ITEM,
		cartItem: response,
	} as const;
}
```

sirreal (commented):
There are a few places I'd recommend adding [`const` assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions). They'll effectively tell TypeScript the object is frozen and it will be able to do more powerful inference. Just add `as const` outside the closing brace.

Here (although in this case I'd remove the object and export the names directly):

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/5785038fcc51141e431a4183b8f67ccb4c4c1c34/assets/js/data/cart/action-types.ts#L1-L12

[Notice the difference](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBAYwgOwM7wIIGEAqBJAeQDkB9HATQAUBRAZTgF44BvAKAEgAlaravANWoksGTjgBccAOTdeAoSLFSANB1l9BJap04FOkmTw1DtuzitYBfANytWoSLEQp0cbPmJkqdYcVo5GFjUjeWFRCWl1UMUcCy4QzVM9AyjEnT0LSzgAQ1RnNBhrIA), input on the left, types on the right:

Screen Shot 2021-02-02 at 18 02 16

In the future, I'd consider restructuring so that your action creators "own" the action types use in your store and you can completely remove `action-types` as the type system can track that your reducers and actions are "in sync". The string literals can be used directly.

In Calypso, we've done several [typed data stores](https://github.com/Automattic/wp-calypso/tree/trunk/packages/data-stores/src) and they're not perfect but the broad strokes may be a good model for the work here.

assets/js/data/types.d.ts
```diff
@@ -0,0 +1,8 @@
+export type ResponseError = {
```
sirreal: This is a type alias (`type Foo = …`), generally [interfaces are preferred](https://www.typescriptlang.org/docs/handbook/advanced-types.html#interfaces-vs-type-aliases). Is there a reason you used an interface here?

```suggestion
export interface ResponseError {
```

assets/js/type-defs/index.d.ts
```diff
@@ -0,0 +1,2 @@
+export * from './cart-response';
```
sirreal: Is there a reason to write `.d.ts` (declaration) files? I would avoid writing these files and just use `.ts`. There's no requirement to include code in `.ts` files.

assets/js/data/types.d.ts
nerrad: Oops nope that was unintentional.

assets/js/type-defs/index.d.ts
nerrad: That's my lack of familiarity with TypeScript showing :). I still don't fully understand the difference between declaration files and regular ts files. I thought this would be a good way to signal this file only contains non compiled code (thus only types, interfaces and declarations).

assets/js/data/shared-controls.ts
nerrad: I'm going to leave this to work on during the meetup. It can be followed up after this pr.

globals.d.ts
nerrad: done in 98c2312

assets/js/data/cart/reducers.ts
nerrad: Removed entirely as a result of work in 263f9e6

assets/js/data/cart/actions.ts
nerrad: done in d2c17a7

assets/js/data/cart/reducers.ts
nerrad: Removed entirely in 263f9e6

assets/js/data/cart/reducers.ts
nerrad: Removed entirely in 263f9e6

assets/js/data/cart/actions.ts
nerrad: done in d2c17a7

assets/js/data/types.d.ts
nerrad: done in ae1b185

assets/js/type-defs/index.d.ts
nerrad: done in b1f1cc0 and 3e5b949.

opr:
I've just pushed some changes:
- Include the `tsc --build` command in the `npm run build` script, this will ensure the project doesn't build if there are TS errors, however development will be able to continue because it doesn't affect the `npm run start` script, do you agree that this is the right approach?
- Add a new `tsconfig.json` file in the `assets/js/data` directory, and reference it in `./package.json` as a TS project
- Add a `tsconfig.base.json` which TS projects can extend, so we don't have to repeat configs.
- Add `"checkJs": false, "allowJs": false` to both configs - this is needed because without it, TS was parsing the JS files and throwing errors, this is probably not what we want right now, we should only let TS check files we've explicitly converted to TS.

assets/js/data/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../../tsconfig.base.json",
+  "compilerOptions": {
+	"checkJs": false,
+	"allowJs": false
+  },
```
nerrad: Speaking of which though, is this even needed? The base config already has these set to false right?

nerrad:
I pushed a couple commits that adds a separate command for the `tsc` check and reverts the `allowJs` and `checkJs` changes.

I have a few thoughts on enforcing the `tsc` check on builds. 

- In principle, I like the idea because this will ensure any issues surface that might have been missed in development.
- In practice this will only surface issues if the developer or someone testing runs `npm run build` instead of `npm run start` on a pr _or_ for any Github workflow executing `npm run build` (eg our bundle size checker workflow).
- The problem with the `allowJs` and `checkJs` flag changes is that it means IDEs will no longer show type errors in JS files. So effectively we've lost the passive built-in type-checking (via typescript flavored jsdoc) for a large part of our codebase when working on it.  **I don't think that's a good tradeoff at the moment**. There have been times where I have made changes in JS files I'm working on because of those surfaced errors.

Here's what I propose:

- we remove the `allowJS` and `checkJS` flag changes so that IDEs still surface type issues in JS files (I've already done this)
- remove `ts:check` from the `build` command (I've already done this)
- we introduce a Github workflow that runs the new `ts:check` command, but the workflow won't be _required_. That way it can still fail on PRs, but doesn't block merge. Bonus if we can have that workflow only run on changed files in the PR.
- For _now_, we can add a checklist item to the releases for verifying that the `ts:check` workflow doesn't surface any errors in typescript files.

Then when we are confident we have enough of our significant files/flows switched over to ts for blocking builds with fails, we can go ahead and add the check to the build command.

If someone is really ambitious, we could also audit and fix (those that are easy with a quick fix, those that are hard with a ts-ignore for now) the 3k+ errors showing up when running `tsc` and then we can enforce it in build command. I imagine a large number of those 3k+ errors come from missing types on imported packages maybe (especially for tests).

Another option would be to be more explicit with checking finished "projects". So example we could add this to the `build` command `npm run ts:check assets/js/data && ...` and that will just check that project. Once more "projects" are configured we continue to add the paths there (or abstract to something like `ts:check:projects` command).

What are your thoughts on the above?

opr:
Ok these changes look good, I didn't realise this would actually stop those JSDoc type errors from surfacing, I figured this would only affect the typescript compilation, the rest of your points I agree with. I can work on adding the workflow next week.

What do you think of changing the `tsc:check` script to be `tsc --allowJs false --checkJs false`, we would not be able to use `--build` if we did this, but based on this link https://www.typescriptlang.org/docs/handbook/project-references.html#build-mode-for-typescript build mode doesn't give that much benefit if we're running it as a one-off does it? 

I think this will be a good change as it will allow the IDE to still check JS files during development, but when running this command, only TS files would be checked. By the way we could also add `ts:check` as a pre-commit check, what do you think?

nerrad:
I think we should keep `ts:check` (or some variation) as is so we have a way of giving visibility to all the errors in the repository (including js files) which should go down over time as we convert things over to typescript. We can try your other suggestions to help with only checking typescript files (I like the idea of adding it to pre-commit hook too).

So maybe we could have something like:

- `ts:check`: will work the way you are suggesting (it only checks typescript files).
- `ts:check-all`: is the same as the current command in the code (`tsc --build`).

assets/js/data/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "extends": "../../../tsconfig.base.json",
   "compilerOptions": {},
-  "include": [ ".", "../type-defs" ],
+  "include": [ "." ],
```
nerrad: Don't we need to keep the `../type-defs` inclusion here for the type check? I'll add it back in but limit it to `.ts` files only in that folder and that way we're not inflating the count due to the js files in the folder.

nerrad:
Alright @opr , here's what I did:

- checked your code 👍 except for the `type-defs` folder inclusion which I added back in with a modification in a1f96cb
- fixed the bug I found with the `isApplyingCoupon` state (from a bug I introduced in a slight cart store refactor) in efeb112
- tested the behaviour of cart and checkout blocks and all products block to validate no other issues were introduced in this pr. afaict everything is all good.

Assuming there are no failing GH actions caused by this PR and you think the above changes are okay, then I'm okay with getting this merged.

PR #3770, opened 2021-02-01
Improve nonce handling by rejecting stale values

### LINKED ISSUE #3757: Nonce validation issue: "X-WC-Store-API-Nonce is invalid" in All Products Block.

mikejolley:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3766 fixes the cache issue reported in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3757 by disabling caches.

As an alternative, this PR adds some additional logic to the nonce middleware which allows the nonce to be stored within browser localStorage, and some new logic to detect if a nonce is older or newer than the current.

Fixes #3757
Closes #3766

1. Start an incognito window.
2. Add some thing to cart from the products block.
3. Go to cart page.
4. Use browser back button back to the page with products block.
5. Add to cart.

If the add to cart worked, this fix is working. Also check browser localStorage and you should see a new field called storeApiNonce.

### Changelog

> Prevent "X-WC-Store-API-Nonce is invalid" error when going back to a page with the products block using the browser back button.

### DIFF
--- a/assets/js/middleware/store-api-nonce.js
+++ b/assets/js/middleware/store-api-nonce.js
@@ -3,9 +3,10 @@
  */
 import apiFetch from '@wordpress/api-fetch';
 
-// @ts-ignore wcStoreApiNonce is window global
-// Cache for the initial nonce initialized from hydration.
-let nonce = wcStoreApiNonce || '';
+// Stores the current nonce for the middleware.
+let currentNonce = window.localStorage.getItem( 'storeApiCurrentNonce' ) || '';
+let previousNonce =
+	window.localStorage.getItem( 'storeApiPreviousNonce' ) || '';
 
 /**
  * Returns whether or not this is a non GET wc/store API request.
@@ -29,9 +30,26 @@ const isStoreApiGetRequest = ( options ) => {
  */
 const setNonce = ( headers ) => {
 	const newNonce = headers?.get( 'X-WC-Store-API-Nonce' );
-	if ( newNonce ) {
-		nonce = newNonce;
+	updateNonce( newNonce );
+};
+
+/**
+ * Updates the stored nonce and localStorage so it is persisted between page loads.
+ *
+ * @param {string} newNonce Incoming nonce string.
+ */
+const updateNonce = ( newNonce ) => {
+	// If the "new" nonce matches the current or previous nonces, we don't need to update. It might be unchanged, or it might be coming from cache.
+	if ( newNonce === currentNonce || newNonce === previousNonce ) {
+		return;
 	}
+
+	previousNonce = currentNonce;
+	currentNonce = newNonce;
+
+	// Update the persisted values.
+	window.localStorage.setItem( 'storeApiCurrentNonce', currentNonce );
+	window.localStorage.setItem( 'storeApiPreviousNonce', previousNonce );
 };
 
 /**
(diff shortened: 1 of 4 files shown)


assets/js/middleware/store-api-nonce.js
```diff
@@ -28,12 +37,44 @@ const isStoreApiGetRequest = ( options ) => {
...
+		return;
+	}
+
+	// Only update the nonce if newer. It might be coming from cache.
+	if ( currentTimestamp && timestamp < currentTimestamp ) {
+		return;
+	}
+
+	currentNonce = nonce;
+	currentTimestamp = timestamp || Date.now() / 1000;
```
opr: Perhaps add a comment for why we're dividing by 1000 here, without looking at L68 of Assets.php it's not _immediately_ obvious. Not a big deal I think but may be helpful.

opr (approved):
This tests great, nice one Mike. The code looks good to me too.

assets/js/middleware/store-api-nonce.js
```diff
@@ -3,9 +3,18 @@
...
-// @ts-ignore wcStoreApiNonce is window global
-// Cache for the initial nonce initialized from hydration.
-let nonce = wcStoreApiNonce || '';
+// Stores the current nonce for the middleware.
+let currentNonce = '';
+let currentTimestamp = 0;
+
+try {
+	const storedNonceValue = window.localStorage.getItem( 'storeApiNonce' );
+	const storedNonce = storedNonceValue ? JSON.parse( storedNonceValue ) : {};
```
nerrad, line 12: Any security concerns about storing the nonce value in localStorage? Would we be able to get away with just storing the nonce timestamp (and assume if that hasn't changed that the nonce in the request is okay)?

nerrad (commented):
As noted in my inline comment, I wonder if you can get away with just storing the timestamp in localStorage?

assets/js/middleware/store-api-nonce.js
mikejolley: @nerrad I don't think there is a security concern because the nonces are also injected into the page (so readable from JS), and they are to stop CSRF not used as secure tokens or anything. We need both here because the nonce injected on the page is out of date if you go back using the browser button. The new nonce needs to persist somewhere.

nerrad: Makes sense, in that case - this gets the 👍 from me (note, I have not tested though...so probably should get confirmed by someone else in testing).

nerrad (approved):

assets/js/middleware/store-api-nonce.js
mikejolley: Thanks, added in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3770/commits/7aa18669a36a7bdb3c5c91f4cc47a26869edef7c

ralucaStan:
works as expected on Firefox, Chrome and Safari.

PR #3763, opened 2021-01-28
Add filters for subtotal string and sale badge

opr:
Credit where credit is due: this PR is based on the work done in #3750 by @Aljullu 

The second part of this PR is in the subscriptions repository: 3972-gh-woocommerce/woocommerce-subscriptions

Apart from the changes already added by Albert, this PR will:

- Introduce two new filters:
    - `EXPERIMENTAL_CART_ITEM_SALE_BADGE_FILTER`
    - `EXPERIMENTAL_CART_ITEM_SUBTOTAL_FILTER`
- These will be used in the Cart and Checkout blocks to allow extensions to modify the information displayed in the `ProductPrice` below the item title, and the `SaleBadge`s that appear on the cart lines.
- Update styling for the cart table, specifically it removes the `table-layout: fixed` property and the fixed-width on the total column which will allow the `SaleBadge` component in that column to display without overflowing the container. 

Fixes #3747

#### Before
image

#### After
image

### How to test the changes in this Pull Request:

1. Check out `add/subtotal-price-filter` in subscriptions.
1. Add a non-subscription product to your cart with a sale price.
2. Add a subscription product to your cart with a sale price.
3. Add varying lengths/types of subscription product to your cart, every 2 weeks, every 6 months etc. 
4. Open the cart block and verify the sale badge shows the period for subscriptions products, but not for regular products.
5. Verify the subtotal for subscription products shows under each product name and contains the period.
6. Verify the non-subscription product does not show a period for the subtotal.
7. Check the cart and ensure the billing frequency is present under the product name too.

PR #3774, opened 2021-02-01
Replace filters with an extendibility API to hook into Cart and Checkout blocks

### LINKED ISSUE #3735: Integration Tracking (WC Subscriptions): create an API so extensions can hook into Cart and Checkout blocks

Aljullu:
Fixes #3735.

This PR refactors the filter added in #3716 so instead of using `@wordpress/hooks`, it uses our own API. This adds an abstraction layer that should make life easier for extensions and it allows us to more easily make internal changes without breaking 3rd party extensions.

There are a couple of PRs that will need to be updated if we go ahead with this approach: #3750 and #3763. cc @opr 

1. Check out this branch in WC Subscriptions: `fix/filters-api` (see associated PR: 3979-gh-woocommerce/woocommerce-subscriptions).
2. Add a subscription product to your cart + any other product which is not a subscription.
3. Go to the Cart or Checkout blocks and verify the sidebar `Total` label is `Total due today`.
4. Remove the subscription from your cart and verify the `Total due today` label changes to `Total`.


packages/checkout/registry/index.js
```diff
@@ -0,0 +1,56 @@
...
+ * @param {any}      defaultValue Default value to filter.
+ * @param {Object}   [args]       Arguments to pass to registered functions.
+ * @param {Function} [validate]   Function that needs to return true when the
+ *                                filtered value is passed in order for the
+ *                                filter to be applied.
+ * @return {any} Filtered value.
+ */
+export const __experimentalApplyCheckoutFilter = (
+	filterName,
+	defaultValue,
```
ralucaStan, line 48: As discussed, maybe it's good to add a default value here `defaultValue = ''`.
A bulletproof way of passing these arguments is using an object 
```
{ 
    filterName: ...,
    defaultValue: ...,
    arg:  ...,
    validate: ...
}
```

packages/checkout/registry/index.js
```diff
@@ -0,0 +1,56 @@
...
+	args = {},
+	validate = () => true
+) => {
+	const filters = getCheckoutFilters( filterName );
+	let value = defaultValue;
+	filters.forEach( ( filter ) => {
+		const newValue = filter( value, args );
+		value = validate( newValue ) ? newValue : value;
+	} );
+	return value;
```
ralucaStan, line 63: This works in this case because we have one filter: `totalLabel`. 
But if there were more filters the value returned would be of the last filter applied. Is this the expected result, or should the filters have some sort of priority?

ralucaStan (commented):
As discussed  I added a comment about applying multiple filters and the expected result. 👍

packages/checkout/registry/index.js
senadir: Currently, that's an accepted tradeoff, we don't plan on introducing priority for the time.

assets/js/base/components/cart-checkout/totals/footer-item/index.js
```diff
@@ -24,9 +25,14 @@ const SHOW_TAXES = TAXES_ENABLED && DISPLAY_CART_PRICES_INCLUDING_TAX;
...
-		__EXPERIMENTAL_TOTAL_LABEL_FILTER,
-		__( 'Total', 'woo-gutenberg-products-block' )
+	const { extensions } = useStoreCart();
+	const label = __experimentalApplyCheckoutFilter(
+		'totalLabel',
+		__( 'Total', 'woo-gutenberg-products-block' ),
+		{
+			extensions,
+		},
+		__experimentalValidateElementOrString
```
senadir: Using an object would make reading this much easier.

packages/checkout/index.js
```diff
@@ -1,10 +1,9 @@
 export * from './totals';
 export * from './utils';
 export * from './shipping';
+export * from './registry';
+export * from './registry/validations.js';
```
senadir: `'./registry/validations.js'` would probably make more sense to be inside `utils`

packages/checkout/registry/index.js
```diff
@@ -0,0 +1,56 @@
...
+export const __experimentalApplyCheckoutFilter = (
+	filterName,
+	defaultValue,
+	args = {},
+	validate = () => true
+) => {
+	const filters = getCheckoutFilters( filterName );
+	let value = defaultValue;
+	filters.forEach( ( filter ) => {
+		const newValue = filter( value, args );
```
senadir: This is too dangerous to just execute like this, any malfunction in the 3PD code would break our block.
You can test this by changing `extensions` to `extension` in WCS code, the whole cart block is now broken.

The error can be at runtime or can happen with a version mismatch (so the 3PD import something that is no longer available) or just anything.

A very important principle when developing the new extensibility points is to not trust the developer.

The filter call should be wrapped in a try/catch closure that will skip that particular filter something wrong happens.

Skipping filters is only possible on raw values, we can't know for sure if an element is faulty or not until we evaluate it, meaning we would break other extensions.

We could wrap our `__experimentalApplyCheckoutFilter` call in a BlockErrorBoudary but that would still neglect the rest of filters.

We can also nest all filters into each other somehow, but I'm not sure how would that happen.

I'd suggest we skip element for now and only have raw values until we have a strong case for elements and then we can think about it very well.

packages/checkout/registry/index.js
```diff
@@ -0,0 +1,56 @@
...
+		value = validate( newValue ) ? newValue : value;
```
senadir: validate only checks the returned value but doesn't guarantee no error was thrown before reaching the return point.

senadir (changes requested):
Thank you for your effort here Albert, I still think we should apply some changes before shipping this, mainly removing the ability to pass Elements and guarding the executed functions.

packages/checkout/registry/index.js
```diff
@@ -0,0 +1,56 @@
...
+ * @return {any} Filtered value.
+ */
+export const __experimentalApplyCheckoutFilter = (
+	filterName,
+	defaultValue,
+	args = {},
+	validate = () => true
+) => {
+	const filters = getCheckoutFilters( filterName );
+	let value = defaultValue;
```
senadir, line 53: There is no enforcing that the new string includes the original or default value, is this intended?

senadir:
@Aljullu I rebased this PR to trunk so it wouldn't conflict more (and to resolve conflict on my PR).

packages/checkout/registry/index.js
Aljullu: Should be fixed in the latest commits. :+1:

assets/js/base/components/cart-checkout/totals/footer-item/index.js
Aljullu: Done. :+1:

packages/checkout/index.js
Aljullu: Given the changes in 094eea75ef254eeb9f61354b3bcbdfb9041c839a (we only accept strings for this filter), I ended up removing that file altogether.

packages/checkout/registry/index.js
Aljullu: Good points, thanks for the feedback. In 660c0266af2494d2eb48469c89175b16b1b349ae I wrapped the filter function in a `try { ... } catch { ... }` block and in 094eea75ef254eeb9f61354b3bcbdfb9041c839a I updated the validation so only strings are accepted. Let me know if changes make sense.

packages/checkout/registry/index.js
Aljullu: Should be fixed in the commits linked in the previous comment.

packages/checkout/registry/index.js
Aljullu: For this particular filter, this is intended because we allow extensions to completely replace the `Total` label with whatever they want. For other filters that change price format, for example, we will need to check that it includes the original string. The good thing is that each filter can specify a different `validation` function.

senadir: I guess that makes sense, it does cause a lot of noise in the component, but I think we can clean it by time.

Aljullu:
Thanks for the feedback @ralucaStan and @senadir. I made several changes to apply the feedback, so I think this is ready for another look.

packages/checkout/registry/index.js
```diff
@@ -0,0 +1,64 @@
...
+} ) => {
+	const filters = getCheckoutFilters( filterName );
+	let value = defaultValue;
+	filters.forEach( ( filter ) => {
+		try {
+			const newValue = filter( value, arg );
+			value = validation( newValue ) ? newValue : value;
+		} catch ( e ) {
+			// eslint-disable-next-line no-console
+			console.log( e );
```
senadir, line 60: Logging errors is not very practical, it wouldn't trigger an error on the console (easily missed) and would happen to all users or admins.
I will write a P2 to suggest how we can catch and expose errors, for now, how about you check if the current use is an admin, fi so, you throw, or else you neglect the error and move on.
```suggestion
			if ( CURRENT_USER_IS_ADMIN ) {
				throw e;
			}
```

packages/checkout/registry/test/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+	test( 'should return default value if there are no filters', () => {
+		const value = 'Hello World';
+		const newValue = __experimentalApplyCheckoutFilter( {
+			filterName,
+			defaultValue: value,
+		} );
+
+		expect( newValue ).toBe( value );
+	} );
```
senadir, line 20: Another test we can add (we need to mock `CURRENT_USER_IS_ADMIN`) is passing in invalid code (that throws an exception) and making sure previous filter still passed.

packages/checkout/registry/index.js
Aljullu: We will probably need to revisit this after the P2 discussion, but for now I made the change you suggested (throw when the user is an admin) but still kept the `console.error` for costumers: 171276669b2a55dc0f6cc170cd4304d8c6681d4b. Let me know how it looks.

packages/checkout/registry/test/index.js
Aljullu: I had a bad time mocking `CURRENT_USER_IS_ADMIN` so it had different values in two different tests. I ended up splitting them in two files: 1d5b636f0c15c94872417deeed916c35cfe70af5. There might better ways of doing this (let me know if you find how!), but couldn't figure it out how after spending some time on this.

senadir (approved):
:shipit:

PR #3775, opened 2021-02-02
Increase width of Cart's product column on <= medium screens

opr:
When viewing the cart block in a way that causes it to be styled with the `is-medium` class or smaller, the product column in the cart block is too narrow.

#### Before 
image

#### After
image

1. Activate Twenty Twenty-One
2. Add products with a long name to your cart
3. View the cart block and see that the product column isn't too narrow.
4. Try this in storefront too but resize your window so the cart block displays at medium or smaller.
5. Try this in other themes too


mikejolley (approved):
Thanks, this fixed it. I will cherry pick into the release branch.

PR #3777, opened 2021-02-02
Add TypeScript via projects

sirreal:
Use a [composite project](https://www.typescriptlang.org/docs/handbook/project-references.html) to separate code and gradually introduce types.

Try it:

```
./node_modules/.bin/tsc --build assets/js/data
```

Note, this project has a much more reasonable 72 errors to address compared to 3000+ when running against the entire project in #3768.

cc: @nerrad


tsconfig.base.json
```diff
@@ -0,0 +1,54 @@
+{
+	"compilerOptions": {
+		"outDir": "./build",
+		"sourceMap": true,
+		"baseUrl": ".",
+		"module": "esnext",
+		// Don't emit; allow Babel to transform files
+		"noEmit": true,
```
sirreal, line 8: I think this should `emitDeclarationOnly` to better support a composite repo. I believe that's how referencing projects access type information. The declaration maps may also be helpful.

Issue #3723, opened 2021-01-22
Audit and refactor cyclical dependencies.


nerrad:
## Background

With work done in #3671, #3692, and #3691, @senadir discovered a number of circular dependency issues in our code structure. In #3704 I did an audit of some of our code base (focusing primarily on `@woocommerce/base-components`, `@woocommerce/base-context` and `@woocommerce/base-hooks`) to fix obvious circular dependencies happening when importing into file from a file that also exports the same thing (so more straightforward `a --> b ---> a` dependencies). 

While the work I did in #3704 and f9b04b2 fixed _some_ issues, we discovered that we still had the issue where `wc-blocks-checkout` was being detected as a dependency for the `wc-blocks-checkout` bundle on _development_ builds. _Production_ builds are fine because of tree shaking.

## What were the circular dependencies?

One of the things we have to watch for that create circular dependencies is importing a component from a package that exports from the file you are writing that import in. So for example, say you have a folder structure like:

```
- base
--- components
----- index.js
----- a.js
----- b.js
```

Your files might have something like this.

```js
// in index.js
export * from 'a';
export * from 'b';

// in a.js
import { bThing } from './';

export const aThing = ( d ) => bThing( d );

// in b.js
export const bThing = ( e ) => console.log( e );
```

Do you see the problem here? `a.js` is importing `bThing` from the `index.js` file _which is also exporting from `a.js`_. So this creates a circular dependency. To fix, you just need to make sure you do something like this in `a.js`:

```js
import { bThing } from './b';

export const aThing = ( d ) => bThing( d );
```

In our case, it's easy to miss things like this because we also use aliases to avoid deeply nested, hard to read, relative paths in our imports. But we had cases where we were essentially duplicating the same pattern with the aliases too. Let's repeat our same example, but instead, let's say that we have an alias `@basecomponents` for the imports that points to `base/components/index.js`.

```js
// in index.js
export * from 'a';
export * from 'b';

// in a.js
import { bThing } from '@basecomponents';

export const aThing = ( d ) => bThing( d );

// in b.js
export const bThing = ( e ) => console.log( e );
```

See how this could slip by unnoticed? As a rule of thumb, when using an aliased package import, make sure you are not using it in a nested situation like the above where you are essentially importing from a file that exports the file you are importing into.

## Yet... there's more

[...]

PR #3778, opened 2021-02-02
Fix broken shipping Calculator

### LINKED ISSUE #3697: Investigate size increase in @woocommerce/blocks-checkout
While working on moving components to `@woocommerce/blocks-checkout` package, its size increased to 50kb, and there was a total of 9kb increased between all files, this didn't make sense because if anything, the total size should decrease because we removed duplicated components between cart and checkout.

This PR introduced the size change https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3671

We should investigate if this increase is legit and if there's anything we can do to reduce it.

### LINKED ISSUE #3758: Move ShippingRatesControl, TotalsShipping, and ShippingRa...
Move ShippingRatesControl, TotalsShipping, and ShippingRateSelector to our base components folder. they&#x27;re too coupled with our base implementation to be exported and we should only export Package.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/80e5b6595316e93864420e66b2292aad2eadf4d2/packages/checkout/shipping/shipping-rates-control/shipping-rates-control.js#L28-L41
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in 80e5b6595316e93864420e66b2292aad2eadf4d2 when #3687 was merged. cc @mikejolley

nerrad:
Closes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3697
Closes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3758

In testing the 4.4.0 release (#3773 ), it was discovered the Shipping Calculator was broken. After some debugging I discovered that the Shipping context provider (and `useShippingAddress` hook) was being built into two different scopes. One for the checkout block built file, and one for the `wc-blocks-checkout` (aliased to `@woocommerce/blocks-checkout`) file. This meant that when the `ShippingCalculator` was pulled into the `wc-blocks-checkout` file, it was referencing a different context provider than what was wrapping the actual checkout block.

A quick solution for this was to implement the work outlined in #3758 which is what I did in this PR, but I'm not sure if this will have the side-effect of breaking the Subscriptions integration work. If yes, then we'll have to have a conversation about what approach we'll take for a fix for the 4.4.0 release and if we do this pull for a temp fix giving us some room to do a more robust refactor to address the issues (might need to pull out the contexts into their own built file shared by others).

There's definitely some relation here to what is outlined in #3723 as well.

## To Test

- [x] Verify that anything shipping related works as expected with the cart and the checkout block flows (make sure you test with built production file) and there are no errors in the console that impact user experience.

Might not be blocking if this fails so 4.4.0 can be shipped, but definitely needs addressed immediately after in that case.
- [x] Verify anything needed by Subscriptions integration is not affected.

**Note:** I have not done exhaustive tests other than verifying this appears to fix the shipping calculator issue in the cart. However, I did see this error in the console which suggests this might break the Subscriptions integration:

Image 2021-02-02 at 3 22 02 PM


nerrad:
You can actually get a peek into the effects of the original bug by looking at the bundle size differences. `build/blocks-checkout` decreased by 33.2kb whereas the build/cart-frontend, build/cart, build/checkout-frontend, and build/checkout files only increased a total of 20.02kb. So the extra `~13kb` that was in `build/blocks-checkout` was all the duplicate context and dependencies pulled in from `base/context` and elsewhere.

senadir:
This is breaking some of our integration, mainly the recurring totals section.

senadir:
Subscription issues are fixed by 3985-gh-woocommerce/woocommerce-subscriptions and are non-blocking for this.

budzanowski (approved):
The code looks good.
Did some quick testing and found some issues with Subscriptions integration, but they should be fixed with 3985-gh-woocommerce/woocommerce-subscriptions, so thi does not impact this PR,

mikejolley (approved):

PR #3787, opened 2021-02-04
Implement a dynamic SlotFill Component for cart item row additions.

### LINKED ISSUE #3619: Integration Tracking (WC Subscriptions): Indicate if the cart/summary item is a downgrade/upgrade/crossgrade
When a customer switches his subscription, some sort of indication should be shown on Cart and Order Summary that the current item is a switch, Checkout shortcode appends (downgrade/upgrade/crossgrade) to the item subtotal, in our case we have "Due today" prepended to the subtotal, so I'm not sure what's the best place to add the switch keyword.

pinging @LevinMedia

budzanowski:
This PR implements a dynamic SlotFill generator for cart items. It can be placed in multiple locations at the same time.
This makes it usable as a SlotFill that can be placed on each cart line. It has a component that aggregates the fills. Its behavior is modeled after the standard Fill component. In this implementation, it only accepts the `BadgeGenerator` that was created for the purpose of the design. With a simple change, we can open the Slots for components of any type or we can extend the list of allowed components. It should also be also possible to implement a filter functionality for the admin UI that would allow limiting the badge display only to a selected group. 

Fixes #3619 
This can be tested using https://github.com/woocommerce/woocommerce-subscriptions/tree/update/wc-blocks-issue-3619 to work. Using that code the following badge feature is implemented:
image

This should also easily solve the Resubscribe display issue.

### DIFF
--- a/packages/checkout/cart-item-dynamic-meta/index.js
+++ b/packages/checkout/cart-item-dynamic-meta/index.js
@@ -1,0 +1,90 @@
+/**
+ * External dependencies
+ */
+import { createSlotFill } from 'wordpress-components';
+import { Children, cloneElement, useEffect } from '@wordpress/element';
+import classnames from 'classnames';
+import { CURRENT_USER_IS_ADMIN } from '@woocommerce/block-settings';
+
+/**
+ * Internal dependencies
+ */
+import BlockErrorBoundary from '../error-boundary';
+const components = [];
+
+function CartItemDynamicMeta( { FillComponent } ) {
+	useEffect(() => {
+		console.log( 'create CartItemDynamicMeta' );
+		return () => {
+			console.log( 'destroy CartItemDynamicMeta' );
+		};
+	  }, []);
+	return (
+		<FillComponent>
+			{ ( fillProps ) => {
+				console.log( 'fill', fillProps );
+				return Children.map( components, ( fill ) => {
+					return (
+						<BlockErrorBoundary
+							renderError={
+								CURRENT_USER_IS_ADMIN ? null : () => null
+							}
+						>
+							{ cloneElement( fill, fillProps ) }
+						</BlockErrorBoundary>
+					);
+				} );
+			} }
+		</FillComponent>
+	);
+}
+
+function SlotProvider( { SlotComponent, className, extensions } ) {
+	console.log('slot');
+	return (
+		<SlotComponent
+			bubblesVirtually
+			className={ classnames(
+				className,
+				'wc-block-components-cart-item-dynamic-meta'
+			) }
+			fillProps={ extensions }
+		/>
+	);
+}
+
+export function ExperimentalCartItemMeta( { children } ) {
+	useEffect(() => {
+		if ( Array.isArray( children ) ) {
+			components.push( ...children );
+		} else {
+			components.push( children );
+		}
+		console.log( 'create' );
+		return () => {
+			components.length = 0;
+			console.log( 'destroy' );
+		};
+	  }, []);
+	return null;
+}
+
+const ExperimentalCartItemDynamicMeta = ( { slotName, extensions, className = '' } ) => {
+	if ( slotName === '' ) {
+		return null;
+	}
+	console.log( {slotName, extensions});
+	const { Fill, Slot } = createSlotFill( slotName );
+
+	return (
+		<>
+			<CartItemDynamicMeta FillComponent={ Fill }/>
+			<SlotProvider 
+				className={ className }
+				SlotComponent={ Slot }
+				extensions={ extensions }/>
+		</>
+	)
+}
+
+export default ExperimentalCartItemDynamicMeta;
--- a/packages/checkout/cart-item-dynamic-meta/index.js
+++ b/packages/checkout/cart-item-dynamic-meta/index.js
@@ -2,73 +2,82 @@
  * External dependencies
  */
 import { createSlotFill } from 'wordpress-components';
-import { Children, cloneElement, useEffect } from '@wordpress/element';
+import {
+	Children,
+	cloneElement,
+	useEffect,
+	useState,
+} from '@wordpress/element';
 import classnames from 'classnames';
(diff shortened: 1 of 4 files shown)


packages/checkout/cart-item-dynamic-meta/index.js
```diff
@@ -0,0 +1,132 @@
+/**
```
senadir, line 1: This is fuzzy for me so please bear with me. Why are you creating a slot and a fill, and then render them yourself?

Why go through the trouble of creating a registry with `ExperimentalCartItemMeta` and loop it? this is exactly what Slot does on the inside, we're just recreating the implementation twice (and having it run twice).

packages/checkout/cart-item-dynamic-meta/index.js
```diff
@@ -0,0 +1,132 @@
...
+		} else {
+			componentsRegistry[ id ] = [ children ];
+		}
+		return () => {
+			console.log( componentsRegistry );
+			componentsRegistry[ id ].length = 0;
+		};
+	}, [ id, children ] );
+	return null;
+};
```
senadir, line 46: It seems here you're collecting all registered badges, Slot would do that for you.

packages/checkout/cart-item-dynamic-meta/index.js
```diff
@@ -0,0 +1,132 @@
...
+	const badgeParameters = badgeCallback( props );
+	if ( badgeParameters === null ) {
+		return null;
+	}
+	return (
+		<ProductBadge className={ badgeParameters?.className }>
+			{ badgeParameters?.text }
+		</ProductBadge>
+	);
+};
```
senadir, line 70: In a tightly coupled implementation, we would avoid using slotFills all together because they would limit the value of slotFills (being able to insert any component) and bring the inconvenient of needing to have them inside a react component.
In this case, a simple registration API would solve the issue, while forcing a tight type (like registerPlugin), so something like.

```js
const registerItemBadge = wc.blocksCheckout;

registerItemBadge("switch-badge", (extensions) => {
  const { swtich_type } = extensions.subscriptions;
  let text = "";
  switch (subscriptions?.switch_type) {
    case "upgraded":
      text = __("Upgrade", "woocommerce-subscriptions");
      break;

    case "downgraded":
      text = __("Downgrade", "woocommerce-subscriptions");
      break;

    case "crossgraded":
      text = __("Crossgrade", "woocommerce-subscriptions");
      break;

    default:
      break;
  }
  if (swtich_type) {
    return {
      className: `wc-subscriptions-badge wc-subscriptions-badge-${swtich_type}`,
      text,
    };
  }
});
```

packages/checkout/cart-item-dynamic-meta/index.js
```diff
@@ -0,0 +1,132 @@
...
+											: () => null
+									}
+								>
+									{ cloneElement( fill, fillProps ) }
+								</BlockErrorBoundary>
+							);
+						}
+					);
+				} }
+			</Fill>
```
senadir, line 126: This trades down the value of a slotFill, when using a slotFill, you're not the primary or only consumer of fill.

packages/checkout/cart-item-dynamic-meta/index.js
```diff
@@ -0,0 +1,132 @@
...
+		</div>
+	);
+};
+
+export { ExperimentalCartItemMeta };
+export default ExperimentalCartItemDynamicMeta;
```
senadir, line 132: It's preferable that you either have a single export that is default, or  all of your exports are named. (the first is named, the second is default).

senadir (changes requested):
Thank you for taking the time to implement this Bartek, sadly I don't think we want to implement a badge system right now, and even if we want to do it, I don't think the current implementation is the best one. Other than complicating the code, we're reimplementing the behaviour of SlotFill in several places and using it wrongly.

packages/checkout/cart-item-dynamic-meta/index.js
budzanowski: This is necessary if you want to have more than one slot.
Imagine that you have a list of cart items. You want to render a SlotFill for each line. With single SlotFill it is not possible. Consider for example on order-meta SlotFill. It has only one slot and if you would place it in two places on the UI only one of those slots would display the Fills. You also can't pass different data per car line to the SingularSlotFill.

So we need to generate a SlotFill per line. And it needs to be unique. Now we need to feed each unique SlotFeel. This is why we need a registry. This way we can ensure that each slot receives the same set of registered components. If this is the same set so why wee need more than one. This is because each slot can receive individual parameters and pass them to fills.

I can try to explain it from the other side. WooCommerce extension says that it needs to add a subscription switch badges to each cart item that has a subscription switch product. This is a component to which you can register your product badge. We take this component put it in the registry. Then for each line in the cart, we create a SlotFill where we copy this component. This component reacts to cart line item props and decides if a badge is needed or not.

You can dynamically add more lines and it will still work because each line is unique.

packages/checkout/cart-item-dynamic-meta/index.js
budzanowski: I collect all badges to be able to feed all of my Slots. Single SlotFeel implementation is not able to render different Slots with different parameters.

packages/checkout/cart-item-dynamic-meta/index.js
budzanowski: I don't understand, could you please elaborate?

packages/checkout/cart-item-dynamic-meta/index.js
budzanowski: This is exactly what is happening with BadgeGenerator. Though the wit single change we are opening the slots for any other component or a set of components.

packages/checkout/cart-item-dynamic-meta/index.js
budzanowski: OK, I did not know that TIL.

nerrad:
I'm going to close this for now. This is outlined in a post internally and we can reference the pull in the future if we decide to revisit.

PR #3790, opened 2021-02-04
Introduce price-format package.

senadir:
This PR introduces for `@woocommerce/price-format` a new package that has formatting helpers, extracted from `@woocommerce/blocks-checkout` to be used in All products and filters.

### How to test
- run `WOOCOMMERCE_BLOCKS_PHASE=1 npm run build` make sure All products et al work fine.
- run `npm run build` make sure cart and checkout work fine.


mikejolley (approved):
Tested both cases, so this looks good 👍🏻  I also checked the built assets and only cart/checkout now reference the checkout package.

PR #3796, opened 2021-02-05
Use default "features" values for payment methods.

budzanowski:
This is related to https://github.com/Automattic/woocommerce-payments/issues/1284

We did not have predefined features value assuming that all the payments integrations will need to provide it on their own.

1. Payment methods for WC Pay 1.9.0 should work

### Changelog

> Fix 3rd Party payment method integrations.


mikejolley (approved):
I tested with a modified Stripe and this worked. I'm happy with this, but lets get another review and set the milestone possible to 4.4.2

nerrad (approved):
Looks good to me. :shipit:

PR #3797, opened 2021-02-05
Update Cart and Checkout sidebar design

### LINKED ISSUE #3632: Update Cart and Checkout sidebar design
Based on the new design iteration for Cart and Checkout, we need to tweak some designs on Cart and Checkout design.

> The top of the cart meta box (right column) was been re-designed to match and align with the cart items (left column).
The label and price font weights in the cart meta box were transposed. The prices are now displayed using bold body font. The labels now use regular weight.
The horizontal separators that separated the shipping method options were removed.

image
image

Aljullu:
Fixes #3632.

This PR updates the Cart and Checkout sidebar so it matches the newest designs, that required making some changes to the Shipping rates control and payment methods too.

I tried to keep commits organized so every code change is explained by its commit description, but if something is still not clear, I can give a more verbose explanation. :slightly_smiling_face: 

### Screenshots

_Cart sidebar with no shipping options:_
imatge

_Cart sidebar with 2 shipping rates:_
imatge

_Checkout sidebar:_
imatge

_Checkout shipping options & payment methods:_
imatge

Go to the Cart and Checkout block and verify designs look good. Some ideas of things to try:
* Test in a situation where there are +1 shipping rate options, only 1 or when there are no available options.
* Test when there is more than one shipping package and when there is only one.
* Test in several themes.
* Test with a different browser font size.
* Verify things look good in the editor as well.
* Test mobile too.

### Changelog

> Improve design of cart and checkout sidebars.

### DIFF
--- a/assets/js/base/components/cart-checkout/order-summary/index.js
+++ b/assets/js/base/components/cart-checkout/order-summary/index.js
@@ -23,6 +23,7 @@ const OrderSummary = ( { cartItems = [] } ) => {
 		<Panel
 			className="wc-block-components-order-summary"
 			initialOpen={ isLarge }
+			hasBorder={ true }
 			title={
 				<span className="wc-block-components-order-summary__button-text">
 					{ __( 'Order summary', 'woo-gutenberg-products-block' ) }
--- a/assets/js/base/components/cart-checkout/order-summary/style.scss
+++ b/assets/js/base/components/cart-checkout/order-summary/style.scss
@@ -1,10 +1,3 @@
-.wc-block-components-order-summary {
-	.wc-blocks-components-panel__button {
-		margin-top: 0;
-		padding-top: 0;
-	}
-}
-
 .wc-block-components-order-summary__content {
 	display: table;
 	width: 100%;
@@ -17,6 +10,10 @@
 	padding-bottom: 1px;
 	padding-top: $gap;
 	width: 100%;
+
+	&:first-child {
+		padding-top: 0;
+	}
 
 	&:last-child {
 		> div {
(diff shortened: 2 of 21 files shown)


senadir:
Still reviewing the PR, but I found a weird spacing issue when we have recurring totals.
image

Aljullu:
Good catch @senadir! I created a PR in Subscriptions to fix it: 3997-gh-woocommerce/woocommerce-subscriptions. I also proposed to create a new component to abstract this kind of styles so it's easier for extensions to add things to the sidebar + more difficult for us to break their styles.

senadir:
Was the narrow styling in recurring totals a side effect or intended behaviour? 
design:
image
storefront:
image

Aljullu:
> Was the narrow styling in recurring totals a side effect or intended behaviour?

@senadir side-effect of a too broad CSS selector. Should be fixed in the WC Subscriptions PR. :+1:

opr (changes requested):
Hey nice work! It mostly looks great, but I noticed in the checkout block when an item name is long, it can sit flush with the price.
image

We can fix this by adding `padding-right` to `.wc-block-components-order-summary-item__description` I think? If we do this we should probably remove it from `.wc-block-components-product-name` too since it doesn't seem to be doing anything there.

Aljullu:
That's interesting, I wasn't able to reproduce that issue in Firefox but I could in Chrome. In any case, it should be fixed in 62b95a314942e86d139117c27b2820479d4b4493 with the changes you suggested.

opr (commented):
When viewing a product with a signup fee in Twenty Twenty it is indented. I will make the change needed to solve this.
image

opr (approved):
@Aljullu I am approving this now - as mentioned it tested well and the bits that didn't have been addressed. When you've seen my change about TwentyTwenty and agree with it then feel free to 🚢

PR #3803, opened 2021-02-05
Align Place Order button to the right

### LINKED ISSUE #3795: Place Order button should be aligned to the right
## To reproduce
1. Go to the Checkout block and scroll down to the Place Order button.
2. Compare its aligned with the designs:

_Block:_
imatge

_Designs:_
imatge

## Expected behavior
The Place Order button should be aligned to the right.

Aljullu:
Fixes #3795.

### Screenshots

_Before:_
imatge

_After:_
imatge

1. Go to the Checkout block and verify the Place Order button is aligned to the right and looks correct.
2. In mobile, verify it takes the entire width.
3. Optionally, test some other themes in addition to Storefront.

### Changelog

> Align place order button to the right of the block.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/style.scss
+++ b/assets/js/blocks/cart-checkout/checkout/style.scss
@@ -16,7 +16,6 @@
 	display: flex;
 	justify-content: space-between;
 	align-items: center;
-	flex-direction: row-reverse;
 	margin-left: 9px;
 
 	.wc-block-components-checkout-place-order-button {
--- a/assets/js/blocks/cart-checkout/checkout/style.scss
+++ b/assets/js/blocks/cart-checkout/checkout/style.scss
@@ -14,15 +14,15 @@
 
 .wc-block-checkout__actions {
 	display: flex;
-	flex-direction: column;
-	margin: 0 0 $gap-large*2;
-	padding-left: 9px;
+	justify-content: space-between;
+	align-items: center;
+	flex-direction: row-reverse;
+	margin-left: 9px;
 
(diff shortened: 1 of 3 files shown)


assets/js/blocks/cart-checkout/checkout/style.scss
```diff
@@ -14,15 +14,15 @@
 
 .wc-block-checkout__actions {
 	display: flex;
-	flex-direction: column;
-	margin: 0 0 $gap-large*2;
-	padding-left: 9px;
+	justify-content: space-between;
+	align-items: center;
+	flex-direction: row-reverse;
```
opr: Why do we need to do this? Can't we change the order of the back to cart button and the checkout button in the markup and keep the default `flex-direction`?

opr (commented):
This tests great, and the code looks pretty good, but I'm curious about the `flex-direction: row-reverse;` - I can approve after a discussion about that.

assets/js/blocks/cart-checkout/checkout/style.scss
Aljullu: Yeah, absolutely. I went for a completely CSS solution but I can't see any reason to prefer that over changing the order of the buttons in the markup. Changed in 8a0d62af687124a0bba8d5b2eea14679d68c83a2.

opr: Nice, that's great. This was just a query about why you did it this way, in case I'd missed something that made this necessary. Thanks for the change :)

opr (approved):

PR #3807, opened 2021-02-07
Allow extensions to extend Cart Item Name in Cart and Checkout.

senadir:
This PR is based on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3774 and shouldn't be merged until that one is merged fully.

This PR adds a new filter for the Product price, currently only in Cart line item and order summary (not in shipping items because that's in a different context all together).

I suspect we might tweak this a bit once we merge #3774 

### Testing instructions 
- Checkout 3991-gh-woocommerce/woocommerce-subscriptions
- Follow steps there like switching or resubscribing.


budzanowski:
Left a small nitpick, but it looks good.

budzanowski (approved):

PR #3810, opened 2021-02-08
Validate order items on post, not on draft order creation

### LINKED ISSUE #3447: If coupons become invalid between applying to a cart and checking out, the order continues without notifying the customer
## Describe the bug

In some cases, when using coupons that become invalid between applying them to the basket and completing the checkout, the order complete, the customer is not notified, and the discount is not applied.

##### 🐛 Limited-use coupons
Tested with a single-use coupon being applied to cart A, but without completing checkout. Create a new incognito session and apply the coupon to cart B. Complete check out on cart B. Return to cart A and complete checkout. Checkout fails, but an account is created.

##### 🐛 Expiring coupons
When a coupon expires after applying to the cart but before completing the checkout, the order completes without the discount being applied and the user is still created. Core behaviour is: Returns user to the checkout page with an error message, no order or account is created.

##### 🐛 Coupons with minimum spend
When a coupon is modified to have a minimum spend value after it has been applied to a cart, the checkout continues without applying the discount. The order is created and the discount is not applied. Core behaviour is: Returns user to the checkout page with an error message, no order or account is created.

##### 🐛 Coupons with product restriction
If a coupon is modified to have a product restriction after it has been applied to a cart, then the checkout continues without applying the discount. The order is created and the discount is not applied. Core behaviour is: Returns user to the checkout page with an error message, no order or account is created.

## To reproduce

1. Create a coupon with no expiry date, minimum spend restriction, product restrictions, or usage limit.
2. Add a product to the cart, go to the checkout block, then apply the coupon.
3. Repeat these steps several times, with different coupons and new baskets each time.
- Modify the coupon's expiry date to be in the past.
- Modify the allowed products for the coupon, to ensure it is not applicable to the cart you're testing against.
- Create a coupon with a single use. Apply it to this test cart. Then use it on another, unrelated cart and complete the checkout.
- Add a minimum order value to the coupon, that is above the cart's value.
4. Complete checkout

## Expected behaviour
The checkout should fail, and inform the shopper that their coupon was invalid. Additionally, no account should be created if the "Create account?" checkbox is on.

mikejolley:
If coupons become invalid between applying to a cart and checking out, right now the coupons will be removed/invalidated without informing the user. This is because the check is ran when the draft order is created on page load. Any errors are not surfaced at this point by the API.

To resolve this, we simply defer the validation until the user attempts to checkout. Instead of the order going through, a proper error is returned from the API as the invalid coupons are removed.

Fixes #3447

### Screenshots

This is an example notice from the API:

Screenshot 2021-02-08 at 11 42 08

There are 4 test cases in #3447.

### Changelog

> If coupons become invalid between applying to a cart and checking out, show the user a notice when the order is placed.

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -195,6 +195,14 @@ class Checkout extends AbstractRoute {
 	 */
 	protected function get_route_post_response( WP_REST_Request $request ) {
 		/**
+		 * Validate items etc are allowed in the order before the order is processed. This will fix violations and tell
+		 * the customer.
+		 */
+		$cart_controller = new CartController();
+		$cart_controller->validate_cart_items();
+		$cart_controller->validate_cart_coupons();
+
+		/**
 		 * Obtain Draft Order and process request data.
 		 *
 		 * Note: Customer data is persisted from the request first so that OrderController::update_addresses_from_cart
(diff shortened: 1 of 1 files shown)


opr (approved):
Nice fix Mike, simple and tests perfectly.

PR #3829, opened 2021-02-11
Refactor block types to include script registration and integration classes

### LINKED ISSUE #3532: Abstract Payments/API class to be used by other plugins
Based on the discussion in https://github.com/woocommerce/woocommerce-subscriptions/pull/3906#discussion_r537126352 we want to:
> abstract some of the current server side Payment API interface to wrap a new generic extension interface that is doing much of what the Payment API interface is doing now.

The new interface should allow extensions to:
- Load data via `wcSettings` and `AssetDataRegistry`.
- enqueue styles and scripts for Cart and Checkout.

Some of the core functionality of this file [`src/Payments/Api.php`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Payments/Api.php) should be moved to that abstract class.

mikejolley:
This is a fairly hefty PR which refactors our Block Type classes (and makes some changes to the Payments Integration classes).

## 1. Removes editor script registration from the Assets class to the block type classes

This keeps block type definitions together with the script registration, avoids the need to do additional checks for build type when registering assets (because blocks are not registered if they do not support the current build), and allows each block type to define it's own dependencies.

## 2. Abstracts `PaymentMethodRegistry` and `PaymentMethodTypeInterface` with `IntegrationRegistry` and `IntegrationInterface`

Since we needed something similar for other integrations (block types) I created `IntegrationRegistry` and `IntegrationInterface` and moved the code that can be reused to those.

`AbstractPaymentMethodType` takes care of the differences and this seems to be used by WC Pay etc so shouldn't break anything.

## 3. `AbstractBlock` refactor

This class will now:

- Register the editor script
- Register the frontend script
- Enqueue the frontend script when the block is rendered 
- Use the `IntegrationRegistry` based on the block name, and with this append registered dependencies and data whenever assets are registered or enqueued.

@senadir before this is merged I want to update subscriptions with it to ensure it covers all use cases. We will hook in via:

`woocommerce_blocks_{registry_identifier}_registration` is the hook where 3rd parties can register themselves. In the case of subscriptions it will be both:

- woocommerce_blocks_cart_block_registration
- woocommerce_blocks_checkout_block_registration

Fixes #3532

This touches on quite a lot of code so we need to ensure:

1. All tests pass, which should cover block rendering issues
2. Smoke test all blocks (frontend and editor) to ensure they render. This will confirm scripts are still loaded.
3. Payment methods should still show up during checkout. Test full checkout flow.
4. Check console for 404s to ensure no invalid scripts are loaded.
5. Check experimental block types like atomic blocks still function

### Changelog

> Refactor block type registration to support 3rd party integrations.

### Dev note:

An important note that internally, this release has modified how `AbstractBlock` (the base class for all of our blocks) functions, and how it loads assets. `AbstractBlock` is internal to this project and does not seem like something that would ever need to be extended by 3rd parties, but note if you are doing so for whatever reason, your implementation would need to be updated to match.

### DIFF
--- a/src/Integrations/IntegrationRegistry.php
+++ b/src/Integrations/IntegrationRegistry.php
@@ -128,6 +128,25 @@ class IntegrationRegistry {
 	 *
 	 * @return string[]
 	 */
+	public function get_all_registered_editor_script_handles() {
+		$script_handles          = [];
+		$registered_integrations = $this->get_all_registered();
+
+		foreach ( $registered_integrations as $registered_integration ) {
+			$script_handles = array_merge(
+				$script_handles,
+				$registered_integration->get_editor_script_handles()
+			);
+		}
+
+		return array_unique( array_filter( $script_handles ) );
+	}
+
+	/**
+	 * Gets an array of all registered integration's script handles.
+	 *
+	 * @return string[]
+	 */
 	public function get_all_registered_script_handles() {
 		$script_handles          = [];
 		$registered_integrations = $this->get_all_registered();
@@ -135,7 +154,7 @@ class IntegrationRegistry {
 		foreach ( $registered_integrations as $registered_integration ) {
 			$script_handles = array_merge(
 				$script_handles,
-				is_admin() ? $registered_integration->get_admin_script_handles() : $registered_integration->get_script_handles()
+				$registered_integration->get_script_handles()
 			);
 		}
 
--- a/src/Integrations/IntegrationRegistry.php
+++ b/src/Integrations/IntegrationRegistry.php
@@ -20,20 +20,17 @@ class IntegrationRegistry {
 	protected $registered_integrations = [];
 
 	/**
-	 * Integration Registry Constructor.
+	 * Initializes all registered integrations.
+	 *
+	 * Integration identifier is used to construct hook names and is given when the integration registry is initialized.
 	 *
 	 * @param string $registry_identifier Identifier for this registry.
 	 */
-	public function __construct( $registry_identifier = '' ) {
+	public function initialize( $registry_identifier = '' ) {
 		if ( $registry_identifier ) {
 			$this->registry_identifier = $registry_identifier;
 		}
-	}
 
-	/**
-	 * Initializes all registered integrations.
-	 */
-	public function initialize() {
 		if ( empty( $this->registry_identifier ) ) {
 			_doing_it_wrong( __METHOD__, esc_html( __( 'Integration registry requires an identifier.', 'woo-gutenberg-products-block' ) ) );
 			return false;
--- a/src/Integrations/IntegrationRegistry.php
+++ b/src/Integrations/IntegrationRegistry.php
@@ -1,0 +1,163 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\Integrations;
+
+/**
+ * Class used for tracking registered integrations with various Block types.
+ */
+class IntegrationRegistry {
+	/**
+	 * Integration identifier is used to construct hook names and is given when the integration registry is initialized.
+	 *
+	 * @var string
+	 */
+	protected $registry_identifier = '';
+
+	/**
+	 * Registered integrations, as `$name => $instance` pairs.
+	 *
+	 * @var IntegrationInterface[]
+	 */
+	protected $registered_integrations = [];
+
+	/**
+	 * Integration Registry Constructor.
+	 *
+	 * @param string $registry_identifier Identifier for this registry.
+	 */
+	public function __construct( $registry_identifier = '' ) {
+		if ( $registry_identifier ) {
+			$this->registry_identifier = $registry_identifier;
+		}
+	}
+
+	/**
+	 * Initializes all registered integrations.
+	 */
+	public function initialize() {
+		if ( empty( $this->registry_identifier ) ) {
+			_doing_it_wrong( __METHOD__, esc_html( __( 'Integration registry requires an identifier.', 'woo-gutenberg-products-block' ) ) );
+			return false;
+		}
+
+		/**
+		 * Hook: integration_registration.
+		 *
+		 * Runs before integrations are initialized allowing new integration to be registered for use. This should be
+		 * used as the primary hook for integrations to include their scripts, styles, and other code extending the
+		 * blocks.
+		 *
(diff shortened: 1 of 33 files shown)


src/Integrations/IntegrationRegistry.php
```diff
@@ -0,0 +1,179 @@
...
+	 * Integration identifier is used to construct hook names and is given when the integration registry is initialized.
+	 *
+	 * @param string $registry_identifier Identifier for this registry.
+	 */
+	public function initialize( $registry_identifier = '' ) {
+		if ( $registry_identifier ) {
+			$this->registry_identifier = $registry_identifier;
+		}
+
+		if ( empty( $this->registry_identifier ) ) {
```
budzanowski, line 34: This should be first in the function block.

src/Integrations/IntegrationRegistry.php
```diff
@@ -0,0 +1,179 @@
...
+		/**
+		 * Hook: integration_registration.
+		 *
+		 * Runs before integrations are initialized allowing new integration to be registered for use. This should be
+		 * used as the primary hook for integrations to include their scripts, styles, and other code extending the
+		 * blocks.
+		 *
+		 * @param IntegrationRegistry $this Instance of the IntegrationRegistry class which exposes the IntegrationRegistry::register() method.
+		 */
+		do_action( 'woocommerce_blocks_' . $this->registry_identifier . '_registration', $this );
```
budzanowski, line 48: Could we only send `::register()` and not the entire class instance?

src/Integrations/IntegrationRegistry.php
```diff
@@ -0,0 +1,179 @@
...
+	 * @return boolean|IntegrationInterface Returns the unregistered integration instance if unregistered successfully.
+	 */
+	public function unregister( $name ) {
+		if ( $name instanceof IntegrationInterface ) {
+			$name = $name->get_name();
+		}
+
+		if ( ! $this->is_registered( $name ) ) {
+			/* translators: %s: Integration name. */
+			_doing_it_wrong( __METHOD__, esc_html( sprintf( __( 'Integration "%s" is not registered.', 'woo-gutenberg-products-block' ), $name ) ) );
```
budzanowski, line 98: Once we call this Integration and another time we call it a Class. I am talking about translator comments on the lines 66 vs 97. Can this be consistent or this is actually something different?

src/Integrations/IntegrationRegistry.php
```diff
@@ -0,0 +1,179 @@
...
+	 * Retrieves all registered integrations.
+	 *
+	 * @return IntegrationInterface[]
+	 */
+	public function get_all_registered() {
+		return $this->registered_integrations;
+	}
+
+	/**
+	 * Gets an array of all registered integration's script handles.
```
budzanowski: The comment needs `editor` word somewhere otherwise this is the same thing as for `get_all_registered_script_handles`

budzanowski:
I can't really follow all the changes since these are 3 different things at the same time. The only thing I was able to spot were simple nitpicks. I will just try to test as much as possible.

What is the impact on existing third-party integrations if they are using AbstractBlock? The class constructor signature was changed. Is this a public API? We have it documented in the `Contributing` section of the documentation.

budzanowski:
I have tried to test as much as I could, and it all looks good.

budzanowski (approved):

src/Integrations/IntegrationRegistry.php
mikejolley: This is matching the pattern in the Payments Registry. There are some other useful methods like unregistration that may be needed.

mikejolley:
> What is the impact on existing third-party integrations if they are using AbstractBlock? The class constructor signature was changed. Is this a public API? We have it documented in the Contributing section of the documentation.

@budzanowski I wasn't totally sure on this. Yes it's "public" because it's accessible, but it's something internal to our blocks and has not been documented as an extendibility API or anything such as this. I would argue it's internal, even if the scope requires it to be public.

budzanowski (approved):

src/Integrations/IntegrationRegistry.php
mikejolley: The conditional above this sets `$this->registry_identifier` based on the incoming class constructor param, so I think this ism in the correct place.

mikejolley:
Thanks for the review @budzanowski 👍🏻 

@senadir did you also get a chance to review this? 

@nerrad did you have any thoughts on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3829#issuecomment-779153654?

src/Integrations/IntegrationRegistry.php
budzanowski: I should have been more explicit. I am talking about the **return early** pattern.

There is no reason to set $this->registry_identifier if $registry_identifier is an empty sting.

```php
if ( empty( $registry_identifier ) ) {
    _doing_it_wrong( __METHOD__, esc_html( __( 'Integration registry requires an identifier.', 'woo-gutenberg-products-block' ) ) );
		return false;
}

$this->registry_identifier = $registry_identifier;
```

mikejolley: I see. The thing is, this works so that `$this->registry_identifier` may be set on the class itself. So it needs to check both $this->registry_identifier and $registry_identifier once resolve. Does that make sense?

budzanowski: Ah OK, now I understand. If `$registry_identifier` is set it overwrites what is in `$this->reg...`. Later we just check if we have anything either set on the class or set through the function parameter. Thank you for the explanation.

nerrad:
> @nerrad did you have any thoughts on #3829 (comment)?

I agree with your assessment. However, I think it should still be highlighted in a dev note in whatever release has these changes (would you be able to pre-write one in the PR description for the next release Mike?)

mikejolley:
Dev note:

An important note that internally, this release has modified how `AbstractBlock` (the base class for all of our blocks) functions, and how it loads assets. `AbstractBlock` is internal to this project and does not seem like something that would ever need to be extended by 3rd parties, but note if you are doing so for whatever reason, your implementation would need to be updated to match.

PR #3832, opened 2021-02-11
Update jetpack autoloader dependency to 2.9.1

mikejolley:
Fixes a conflict with WooCommerce core by using version 2.9.1 of the jetpack autoloader, as used in https://github.com/woocommerce/woocommerce/blob/master/composer.json

^2.9.1 covers us from >2.9.1 <3.0.0 so should be safe to use, and is recommended in https://getcomposer.org/doc/articles/versions.md#caret-version-range-.

This will be cherry picked into the 4.4.x branch.

1. Composer install
2. Check Blocks plugin still loads
3. Confirm correct version is loaded in WC > Status

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
 	"minimum-stability": "dev",
 	"require": {
 		"composer/installers": "^1.7.0",
-		"automattic/jetpack-autoloader": "2.7.1"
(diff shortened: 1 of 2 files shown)


nerrad (approved):
LGTM 👍

PR #3835, opened 2021-02-11
Enhance checkout filter system.

senadir:
This PR is a follow up to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3750.

### Error messages on the validation
instead of skipping bad filters, we're going to throw a specific error for each error (not a string, doesn't contain label...).
Error being thrown would be caught by the filter system and highlighted to admins or skipped for users, just like other exceptions.
Error messages are now living in `util/validation`.

### Refactor passed in arguments
We now pass extensions as a top level prop to the filter, this is to align with the rest of our APIs like SlotFills and to make it easier for plugin author to access extended data.
I also renamed some arguments like blocks to be context and mixed usage of cartItem|lineItem to lineItem.

### Performance guards
We previously called all registered filters each the component did a recalculation (about 8-10 times on a single render), this is risky because we're calling 3PD code each time and we don't know how expensive it can be, therefore, I memozied all filters calls.
I tried to do that inside the filter directly (by making it a hook) but it didn't really work, probably because the passed in arguments are too complex.

### Cherry pick lost commits
Previous PRs had a couple of commits from @Aljullu that never made it to the merged PR for some reason, I was still able to see them, so I cherry picked them into this PR.

## Testing
PR needs 3971-gh-woocommerce/woocommerce-subscriptions in WCS.
Each step is a commit, starting from `throw errors on validation` commit.
- Make sure all filters are working fine, add subscription products to your cart and make sure "total due today" is showing, "x every y" and "due today x" are showing on each price, subtotal, and saving badge,
- Try returning a wrong value in WCS, if you're an admin, you should see the error, if you're a customer, you should only see a console error.


assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -65,28 +74,38 @@ const OrderSummaryItem = ( { cartItem } ) => {
...
-	} );
+	const productPriceFormat = useMemo(
+		() =>
+			__experimentalApplyCheckoutFilter( {
+				filterName: 'cartItemPrice',
+				defaultValue: '<price/>',
+				extensions,
+				arg: {
+					lineItem: cartItem,
+					context: 'checkout',
```
Aljullu: Is it intentional to differentiate between `context: 'summary'` and `context: 'checkout'`? Not blocking but I think it might be good to document which contexts will exist so it's easier for us (and 3rd-parties) to know which context to use and also to avoid introducing new ones by mistake.

Another idea would be to save them in constants exported from `@woocommerce/blocks-checkout`, that would help to avoid typos.

packages/checkout/registry/index.js
```diff
@@ -32,32 +37,38 @@ const getCheckoutFilters = ( filterName ) => {
...
- * @param {Object}   o              Object of arguments.
- * @param {string}   o.filterName   Name of the filter to apply.
- * @param {any}      o.defaultValue Default value to filter.
- * @param {any}      [o.arg]        Argument to pass to registered functions. If
- *                                  several arguments need to be passed, use an
- *                                  object.
+ * @param {Object} o              Object of arguments.
+ * @param {string} o.filterName   Name of the filter to apply.
+ * @param {any}    o.defaultValue Default value to filter.
+ * @param {Object} [o.extensions]   Values extend to REST API response.
```
budzanowski, line 49: Comment part indentation.

packages/checkout/registry/index.js
```diff
@@ -32,32 +37,38 @@ const getCheckoutFilters = ( filterName ) => {
...
+ * @param {any}    [o.arg]        Argument to pass to registered functions. If
+ * several arguments need to be passed, use an
+ * object.
```
budzanowski: That can go on the previous line.

Aljullu (approved):
Nice changes @senadir, much cleaner API and I like that validation shows clearer messages now. :100: 

I left a single comment about the `context` argument, but everything else is looking good to me.

budzanowski:
The pattern of filter usage + `useMeme` suggests that we could maybe move the `useMemo` inside the filter.
I also have left a small formatting comment. Besides that this looks good :)

budzanowski (approved):

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
opr: I think it's good to provide the context, that's why in the first version of our filters (before we had this API) I wanted to pass the block type. It doesn't cost us anything to send it, and it has the potential to be extra helpful to anything extending blocks to know the context.

+1 for the documentation and exporting the context names as constants though 👍🏼

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -126,38 +135,53 @@ const CartLineItemRow = ( { lineItem = {} } ) => {
...
-			typeof value === 'string' && value.includes( '<price/>' ),
-	} );
+	const subtotalPriceFormat = useMemo(
+		() =>
+			__experimentalApplyCheckoutFilter( {
+				filterName: 'subtotalPriceFormat',
+				defaultValue: '<price/>',
+				extensions,
+				arg: {
+					lineItem,
```
opr: If we're going to pass the context in the arg for one filter (`cartItemPrice`), then we should ensure we do it for all filters imo. Even if the filter only gets applied in one place, we may eventually do it elsewhere, so it would be good to have this as a standardised parameter we pass through.

packages/checkout/utils/validation/index.js
```diff
@@ -0,0 +1,51 @@
...
+	if ( ! value.includes( label ) ) {
+		throw Error(
+			sprintf(
+				// translators: %s value passed to filter.
+				__(
+					'Returned value must include <price/>, you passed "%s"',
+					'woo-gutenberg-products-block'
+				),
+				value
+			)
```
opr: Since the label may change, do we need to change this to

```suggestion
			sprintf(
				// translators: %$1s placeholder that returned value must include. %$2s value passed to filter.
				__(
					'Returned value must include %$1s, you passed "%$2s"',
					'woo-gutenberg-products-block'
				),
				label,
				value
			)
```

opr (dismissed):

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
senadir: > Is it intentional to differentiate between context: 'summary' and context: 'checkout'?

No, just something I forget to update, checkout is for the main checkout area, the summary is for the sidebar (applies for both cart and checkout sidebars).

> Another idea would be to save them in constants exported from @woocommerce/blocks-checkout, that would help to avoid typos.

I agree but I'm afraid it might complicate things beyond how complicated they're.

I know they're optional and people can just use a string, but once you implement it, you will also have it in a tutorial, and that would be an extra struggle.

Do you think the benefit of having a constant here is worth the extra overhead?

senadir:
> The pattern of filter usage + useMeme suggests that we could maybe move the useMemo inside the filter.

I did try that but was unsuccessful, will try again.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
senadir: > I think it's good to provide the context, that's why in the first version of our filters (before we had this API) I wanted to pass the block type. It doesn't cost us anything to send it, and it has the potential to be extra helpful to anything extending blocks to know the context.

I think context is helpful for filters or items that are used in several places (cart item being used in several places) but would it make sense for things used just once in a single place? feels a bit redundant.

I get the argument that it doesn't cost much to have, but afraid it might cause confusion.
It should be mandatory once you introduce the same filter somewhere else.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
senadir: Applying the same filter elsewhere could be a breaking change so I don't see value in prematurely passing in context when we only have the filter in once price.

I did update our context calls for several filters but not all filters (skipped TotalsFilter because it's a single component so we can't really know the context).

Aljullu (commented):
I tested the last changes and they are working fine, but I feel code became more complex and the `arg` comparison seems run on every render.

I wonder if we could undo the last changes in `__experimentalApplyCheckoutFilter` and instead create a hook wrapper around it. Something along these lines:

```JSX
const useCheckoutFilter = ( {
	filterName,
	defaultValue,
	extensions,
	arg = null,
	validation = null,
} ) => {
	return useMemo( () => {
		return __experimentalApplyCheckoutFilter( {
			filterName,
			defaultValue,
			extensions,
			arg,
			validation,
		} );
	}, [ filterName, defaultValue, extensions, arg, validation ] );
};
```

It could then be consumed like this:

```JSX
const label = useCheckoutFilter( {
	filterName: 'totalLabel',
	defaultValue: __( 'Total', 'woo-gutenberg-products-block' ),
	extensions,
	arg,
	validation: mustBeString,
} );
```

In some cases we might need to memoize `arg` or `validation`, but I think that should be handled by the consumer only in cases when it's needed. In fact, in many cases we will be able to define `arg` and `validation` outside of the component, so they would have the same reference across renders.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
Aljullu: > Do you think the benefit of having a constant here is worth the extra overhead?

A couple of benefits I see on using constants is making it more difficult to make typos and that if some day we want to merge two contexts or rename one of them, we will be able to do so because extensions will not be using the context name directly but a constant that we can remap to a different value. But I don't have a strong opinion on this, so I'm fine either way.

senadir:
Filled an issue for the performance problem.
https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3855

Ideally, we would like to solve this before going public.

packages/checkout/registry/index.js
```diff
@@ -53,45 +52,13 @@ const getCheckoutFilters = ( filterName ) => {
 export const __experimentalApplyCheckoutFilter = ( {
 	filterName,
 	defaultValue,
-	extensions = {},
-	arg: _arg = null,
-	validation: _validation = () => true,
+	extensions,
+	arg = null,
+	validation = () => true,
```
Aljullu: `() => true` will be considered a different function on each render and that will cause `useMemo()` to always rerun. That's not an issue now because all ours filters specify a `validation` argument, but I think it's still worth fixing it.

packages/checkout/registry/index.js
```diff
@@ -53,45 +52,13 @@ const getCheckoutFilters = ( filterName ) => {
 export const __experimentalApplyCheckoutFilter = ( {
```
Aljullu, line 58: Given that this function now contains a hook, I think we should rename it to `useExperimentalApplyCheckoutFilter` or something similar. React + ESLint apply some extra-checks to functions whose name starts with `use` so they follow the [rules of hooks](https://reactjs.org/docs/hooks-rules.html). The name will need to be updated in the [docs](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/blocks/feature-flags-and-experimental-interfaces.md) as well.

Aljullu (approved):
Thanks for making these changes @senadir. I'm looking forward to #3855 which will help avoid some unnecessary re-calculations. I left two comments that I think might be worth fixing before merge, but pre-approving because everything else looks good to me and it's testing well.

packages/checkout/registry/index.js
senadir: It seems the rule of hooks ignores functions that start with an underscore so we should be good?
I'd prefer we leave it as is for now, I didn't spot any issues with it.

PR #3845, opened 2021-02-15
Only use SHOP_URL if set

### LINKED ISSUE #2600: Unset shop pages are using `false` as value
Today while testing, I saw that some shop pages are using `false` as a value if that page doesn’t exist, instead, they should either use an empty string or home page link.

@mikejolley suggested we should default to the home page, this is done here
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f77c7f387ea413cdfe1c2022b105e9bcf553b386/src/Assets.php#L206

However, we have some conditional rendering using those links, so we should keep them into account

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f77c7f387ea413cdfe1c2022b105e9bcf553b386/assets/js/base/components/cart-checkout/policies/index.js#L21

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f77c7f387ea413cdfe1c2022b105e9bcf553b386/assets/js/base/components/cart-checkout/policies/index.js#L37

Also, if we use an empty string, we should remember it will render, as `""` doesn’t evaluate to false in React, we should use `SHOP_URL.length` as this will evaluate to 0 if it's an empty string.

**To Reproduce**

1. Have a store with no shop page.
2. in empty cart, notice that "go to store" will take you to `/false`

**Expected behavior**
Those links should either not render, or render to the home page or the current page.

mikejolley:
This is a small tweak to ensure we only use SHOP_URL if it's actually set. This cannot affect existing blocks because they store content as inner blocks on creation.

Fixes #2600

1. In WC > Settings > Products unset the shop page and save.
2. Insert a NEW cart block
3. Confirm the empty cart template does not include a link to browse shop.

### Changelog

> Hide Browse Shop link in cart block empty state when there is no shop page.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/empty-cart-edit/index.js
+++ b/assets/js/blocks/cart-checkout/cart/empty-cart-edit/index.js
@@ -12,6 +12,69 @@ import PropTypes from 'prop-types';
 import iconDataUri from './icon-data-uri.js';
 import './style.scss';
 
+const templateItemBrowseStore = SHOP_URL
+	? [
+			'core/paragraph',
+			{
+				align: 'center',
+				content: sprintf(
+					// Translators: %s is the link to the store product directory.
+					__(
+						'<a href="%s">Browse store</a>.',
+						'woo-gutenberg-products-block'
+					),
+					SHOP_URL
+				),
+				dropCap: false,
+			},
+	  ]
+	: null;
+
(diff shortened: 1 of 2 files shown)


opr (approved):
Code changes are sound and it tests well 🚢

PR #3846, opened 2021-02-15
Remove shadows from text buttons and gradient background from selects in some themes

Aljullu:
This PR:
* Sets `text-shadow` to none in the `text-button` mixin (buttons that we want to display as plain text).
* Resets the `background` property of the button of select components. Until now, it was only setting the `background-color` to white, but that didn't reset gradients (which are treated as `background-image`).

One theme that allows seeing both changes in action is [Bookshop](https://woocommerce.com/products/bookshop/).

1. Install and activate Bookshop.
2. Go to the Cart block and verify the `Delete item` button doesn't have text shadow:

| Before | After |
| --- | --- |
| imatge | imatge |

3. Go to the Checkout block and verify Country and State selects don't have a gradient in the background:

| Before | After |
| --- | --- |
| imatge | imatge |

### DIFF
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -124,6 +124,7 @@ $fontSizes: (
 	background: transparent;
 	box-shadow: none;
 	display: inline;
+	text-shadow: none;
 
 	&:hover,
 	&:focus,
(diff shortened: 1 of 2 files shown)


mikejolley (approved):
:shipit:

PR #3853, opened 2021-02-16
Fix product price not displaying properly when product is on sale

### LINKED ISSUE #3852: Sale price not displaying correctly in cart and checkout
## Describe the bug
If a product is on sale, then the `<ProductPrice>` component does not show the strikethrough price properly.

## To reproduce

1. Set a product to be on sale and add it to the cart
2. Go to the cart and checkout blocks and observe only the sale price is displayed.

## Expected behavior
The regular price should be displayed, but styled with strikethrough.

## Screenshots
The price under the title should say: ~~£24.00~~ £21.60
image

opr:
This PR fixes an issue where the Sale price of a product would not display properly, the reason this wasn't happening is because in the ProductPrice component, we choose which component to render based on a set of conditions, one of which is whether the price is less than the regular price. This is fine, but there are also other conditions that evaluate as true, meaning the component we chose gets overridden later in the execution path. By using `else if` we can ensure only the first true condition is executed.

Fixes #3852

### Screenshots
Before | After
-------|------
image | image

1. Set a product to be on sale and add it to the cart
2. Go to the cart and checkout blocks and make sure you can see the struckthrough regular price and the sale price is displayed normally.

### DIFF
--- a/assets/js/base/components/product-price/index.js
+++ b/assets/js/base/components/product-price/index.js
@@ -126,7 +126,6 @@ const ProductPrice = ( {
 	}
 
 	const isDiscounted = regularPrice && price !== regularPrice;
-
 	let priceComponent = (
 		<span
 			className={ classNames(
@@ -148,9 +147,7 @@ const ProductPrice = ( {
 				regularPriceStyle={ regularPriceStyle }
 			/>
 		);
-	}
-
-	if ( minPrice !== null && maxPrice !== null ) {
+	} else if ( minPrice !== null && maxPrice !== null ) {
 		priceComponent = (
 			<PriceRange
 				currency={ currency }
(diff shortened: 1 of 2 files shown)


assets/js/base/components/product-price/index.js
```diff
@@ -148,9 +147,7 @@ const ProductPrice = ( {
 				regularPriceStyle={ regularPriceStyle }
 			/>
 		);
-	}
-
-	if ( minPrice !== null && maxPrice !== null ) {
+	} else if ( minPrice !== null && maxPrice !== null ) {
```
mikejolley, line 150: This price handles ranges. Did you test this fix on a discounted product with variations? I will do so now..

senadir (approved):
Logic checks out.

assets/js/base/components/product-price/index.js
senadir: In the previous versions of this code, we would return on each condition, so this refactor should be the same as the original behaviour.

mikejolley: Seems to be working with variables. Doesn't appear that individual variations in the cart ever have ranges. Trying to figure out when this case is used.

mikejolley (approved):

PR #3859, opened 2021-02-17
Handle out of stock product visibility setting in All Products block.

### LINKED ISSUE #2824: Product block don't respect out-of-stock product visibility setting
**Describe the bug**
There is a setting (checkbox) in WooCommerce settings to display/not display out-of-stock products in the shop catalog. Regardless of this setting the product block display all products.

**To Reproduce**

1. Set WooCommertce to not show out-of-stock products in the shop (here: /wp-admin/admin.php?page=wc-settings&tab=products&section=inventory)
2. check to see: the product block still display them

**Expected behavior**
Product block should respect this setting and not display out-of-stock product if that option is checked.

* Core version: 5.4.2
* WooCommerce version: 4.2.2
* WooCommerce Blocks version: 2.9.0
* Site language: German

ralucaStan:
A new `hideOutOfStockItems` setting flag has been added to reflect the user setting `Hide out of stock items from the catalog`. When true, the products call will include a field for `stock_status: ['instock', 'onbackorder']`.

Fixes #2824 

1. Add All Products block to a page
2. Mark one product as `Out of stock`, and one as `On back order`
3. Select WC > Settings > Products > Inventory > Hide out of stock items from the catalog
4. Go to the products block and see that the `Out of stock` product is missing, but the `On back order` is visible

### PHP Changes
Not sure if the code here is the best, feel free to suggest improvements. Also make sure the code is not missing any edge cases.

### Changelog

> Handle out-of-stock product visibility setting in All Products block.

### DIFF
--- a/src/StoreApi/Routes/Products.php
+++ b/src/StoreApi/Routes/Products.php
@@ -309,10 +309,14 @@ class Products extends AbstractRoute {
 
 		$params['stock_status'] = array(
 			'description'       => __( 'Limit result set to products with specified stock status.', 'woo-gutenberg-products-block' ),
-			'type'              => 'string',
-			'enum'              => array_keys( wc_get_product_stock_status_options() ),
+			'type'              => 'array',
+			'items'             => array(
+				'type' => 'string',
+			),
+			'default'           => [],
 			'sanitize_callback' => 'sanitize_text_field',
 			'validate_callback' => 'rest_validate_request_arg',
+
 		);
 
 		$params['attributes'] = array(
--- a/src/StoreApi/Routes/Products.php
+++ b/src/StoreApi/Routes/Products.php
@@ -308,15 +308,15 @@ class Products extends AbstractRoute {
 		);
 
 		$params['stock_status'] = array(
-			'description'       => __( 'Limit result set to products with specified stock status.', 'woo-gutenberg-products-block' ),
-			'type'              => 'array',
-			'items'             => array(
-				'type' => 'string',
+			'description' => __( 'Limit result set to products with specified stock status.', 'woo-gutenberg-products-block' ),
+			'type'        => 'array',
+			'items'       => array(
+				'type'              => 'string',
+				'enum'              => array_keys( wc_get_product_stock_status_options() ),
+				'sanitize_callback' => 'sanitize_text_field',
+				'validate_callback' => 'rest_validate_request_arg',
 			),
-			'default'           => [],
-			'sanitize_callback' => 'sanitize_text_field',
-			'validate_callback' => 'rest_validate_request_arg',
-
+			'default'     => [],
 		);
 
 		$params['attributes'] = array(
--- a/src/StoreApi/Utilities/ProductQuery.php
+++ b/src/StoreApi/Utilities/ProductQuery.php
@@ -316,7 +316,7 @@ class ProductQuery {
 
 		if ( $wp_query->get( 'stock_status' ) ) {
 			$args['join']   = $this->append_product_sorting_table_join( $args['join'] );
-			$args['where'] .= $wpdb->prepare( ' AND wc_product_meta_lookup.stock_status = %s ', $wp_query->get( 'stock_status' ) );
+			$args['where'] .= ' AND wc_product_meta_lookup.stock_status IN ("' . implode( '","', array_map( 'esc_sql', $wp_query->get( 'stock_status' ) ) ) . '")';
 		}
 
 		if ( $wp_query->get( 'min_price' ) || $wp_query->get( 'max_price' ) ) {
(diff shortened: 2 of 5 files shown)


src/StoreApi/Routes/Products.php
```diff
@@ -309,10 +309,14 @@ public function get_collection_params() {
...
 		$params['stock_status'] = array(
 			'description'       => __( 'Limit result set to products with specified stock status.', 'woo-gutenberg-products-block' ),
-			'type'              => 'string',
-			'enum'              => array_keys( wc_get_product_stock_status_options() ),
+			'type'              => 'array',
+			'items'             => array(
+				'type' => 'string',
+			),
+			'default'           => [],
 			'sanitize_callback' => 'sanitize_text_field',
```
Aljullu: I think `sanitize_callback` needs to be inside `items`. Otherwise, it was not working for me. :confused:

Aljullu (approved):
Nice to see this being fixed. I needed to make a change to the code to get this working, but after that it was testing fine. I commented below the change I needed to do.

I would like @mikejolley to take a second look to this PR since I'm never 100% confident when reviewing code that queries the database, but to me it's looking good.

Also wondering if this is a breaking change to the store API. Should we keep supporting `stock_status` being a string in addition to an array?

src/StoreApi/Routes/Products.php
```diff
@@ -309,10 +309,14 @@ public function get_collection_params() {
 
 		$params['stock_status'] = array(
 			'description'       => __( 'Limit result set to products with specified stock status.', 'woo-gutenberg-products-block' ),
-			'type'              => 'string',
-			'enum'              => array_keys( wc_get_product_stock_status_options() ),
```
mikejolley, line 313: Sorry if I wasn't clear yesterday. We should keep `enum` if possible, I just wasn't sure if it needs to be defined within or outside of items. Ideally we can keep this in place so invalid values are rejected.

mikejolley:
@Aljullu Supporting both a single string or an array would be interesting. I know you can define an array of types, but I am unsure who this would behave in conjunction with `enum` and `items`. It would probably be better to keep it consistent as an array however.

@ralucaStan Is any other code inside blocks using this query?

There are some docs too in the StoreAPI folder which need updating if this changes from string to array.

src/StoreApi/Utilities/ProductQuery.php
```diff
@@ -316,7 +316,7 @@ public function add_query_clauses( $args, $wp_query ) {
 
 		if ( $wp_query->get( 'stock_status' ) ) {
 			$args['join']   = $this->append_product_sorting_table_join( $args['join'] );
-			$args['where'] .= $wpdb->prepare( ' AND wc_product_meta_lookup.stock_status = %s ', $wp_query->get( 'stock_status' ) );
+			$args['where'] .= ' AND wc_product_meta_lookup.stock_status IN ("' . implode( '","', array_map( 'esc_sql', $wp_query->get( 'stock_status' ) ) ) . '")';
```
mikejolley, line 319: 👍🏻

ralucaStan: @mikejolley I did not find another block using this query.

Thanks for the heads-up, I've updated the docs file.

src/StoreApi/Routes/Products.php
ralucaStan: Changes pushed, please check again. I also moved `sanitize_callback` and `validate_callback` under items.

mikejolley (approved):
Good job 👍🏻  Also magically still works with strings and comma separated value, so must be some magic behind the scenes to convert those values automatically :)

Magganpice:
My product blocks are still displaying out of stock products in WooCommerce 5.1.0. Am I missing something?

fishblizz:
Is this issue the same as when you don't put a price on a variation? Because they show up aswell.

ralucaStan:
Hi there @fishblizz,
thank you for reaching out. Could you please describe in detail the bug you are referring to? Some steps to reproduce the bug would be helpful. Depending on them we might need to open another issue.

fishblizz:
Hi @ralucaStan, my reference is this issue wich kinda got forwarded to Github: https://wordpress.org/support/topic/product-with-variations-shows-product-without-a-price/page/2/#post-14787001

ralucaStan:
HI @fishblizz, 
thanks for your comment. The issue from your link is different from this one and has a separate [GitHub ticket on WC](https://github.com/woocommerce/woocommerce/issues/27192).

fishblizz:
Hi @ralucaStan, thank you for reaction and confirmation that the other ticket is still actual. I will leave this one and move to the one you've pointed out.

Magganpice:
This issue is still there. Wrong products are still displayed. Not sensitive to the "display out of stock" setting in WooCommerce settings. So the block still shows out of stock products. Is the issue not properly understood here?

frontdevde:
> This issue is still there.

@Magganpice I wasn't able to reproduce the issue with the latest version of WooCommerce and WooCommerce Blocks. Could you please provide us with detailed steps on how to reliably reproduce the issue? We'd also appreciate it if you could share which versions of WordPress, WooCommerce, and WooCommerce Blocks you are using. Thank you!

PR #3864, opened 2021-02-18
Don't load wc blocks data on every route.

### LINKED ISSUE #3748: Don't load `wc_blocks_data` on every route unnecessarily.
There's some work being discussed on improving performance issues spotted by a developer in the community in https://github.com/woocommerce/woocommerce-admin/issues/5690 and I noticed in doing some of my own investigation that it looks like WooCommerce Blocks is [loading a bunch of data unnecessarily](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/befd4cfc136b096e0fec4e5e296faee82360969b/src/Assets.php#L26) on any route that has code using the `AssetDataRegistry` interface. There's a few things we need to do here.

- [x] Switch from using the `woocommerce_shared_settings` filter as it was only intended as a temporary back-compat shim and implement adding data via `AssetDataRegistry::add` (which should be done by blocks needing it on demand).  **priority 1**
- [ ] ~~Refactor `Assets` into a new initializer/loader/controller for registering block assets and inject AssetDataRegistry and `Assets/Api` as dependencies (then get rid of the old class). **priority 2**~~
- [x] Remove `BackCompatAssetDataRegistry` (given the length of time it's been there it should be safe to do so now). **priority 3**
- [x] Remove the `woocommerce_shared_settings` filter (or potentially just have it throw a deprecated notice and not filter anything). We'll need to make sure no other extensions are using it before we do (WC Admin for example).  **low priority**

**I am moving the:**
- Refactor `Assets` into a new initializer/loader/controller for registering block assets and inject AssetDataRegistry and `Assets/Api` as dependencies (then get rid of the old class). **priority 2** 
to a separate issue to make the output PR less bloated -> #3868

budzanowski:
Fixes: #3748 

There are a couple of connected but different things happening in this PR:
 1. Removal of `BackCompatAssetRegistry`: this was used for WC Admin <= v0.19.0. This is an 18 months old version so this should be safe to remove.
 2. Decoupling loading of blocks common data from the `woocommerce_shared_settings` filter. In the new approach, the common data is registered by each block ( previously checking if it was not already registered ) in AbstractBlock::enqueue_data. This is managed by the default render_callback. This ensures that the data creation function will be called only if Gutenberg will render a block.
 3. Deprecation of woocommerce_shared_settings

### Testing:

This reorganizes how data required by blocks are loaded. Therefore this requires a rather thorough smoke testing.

The only truly visible item from the outside should be the filter deprecation. All other items are internal refactor and are skipped from the changelog.

### Changelog

> Filter woocommerce_shared_settings is now deprecated.


budzanowski:
I have intended to rebase but ended up with merge. Fingers automation 😞

assets/js/settings/blocks/constants.js
```diff
@@ -3,94 +3,64 @@
...
-export const WC_BLOCKS_BUILD_URL = getSetting( 'wcBlocksBuildUrl', '' );
-export const SHIPPING_COUNTRIES = getSetting( 'shippingCountries', {} );
-export const ALLOWED_COUNTRIES = getSetting( 'allowedCountries', {} );
-export const SHIPPING_STATES = getSetting( 'shippingStates', {} );
-export const ALLOWED_STATES = getSetting( 'allowedStates', {} );
-export const SHIPPING_METHODS_EXIST = getSetting(
-	'shippingMethodsExist',
-	false
-);
+const blockConstants = getSetting( 'block_constants', {} );
```
senadir, line 6: This is a breaking change, I know we're not using this publicly yet but it should be highlighted at least.
For example, I used `getSetting( 'displayCartPricesIncludingTax', false );` which is not longer available.
What's the reason for moving all data under `block_constants`?

budzanowski: With the refactor each block tells the assets registry that it needs the constants. Until a block rendered happens the AssetsRegistery will not trigger the constants creation. This was the sole purpose of the refactor - no to trigger the wc_data_creation on every route. **And now with that in mind the reason for the `block_constant`:** since this is no longer global and each block does it is own registering we need a way to easily check if the data was already requested - so we will be able to avoid unnecessary multiple registers. This is managed by the AbstractBlock:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3864/files#diff-d1b778e863fef808d0b34565afe0c8bc172cc3d9e3e99324b402a19577f1fbf1R333-R336

```PHP
// Enqueue common data.
if ( ! $this->asset_data_registry->exists( 'block_constants' ) ) {
	$this->asset_data_registry->add( 'block_constants', [ self::class, 'get_wc_block_data' ] );
}
```

We have only one key that we need to check. Any other dynamic approach way would require adjusting the AssetsDataRegistery framework for this special case of blocks common data which I think we should not do.

assets/js/atomic/blocks/product-elements/summary/block.js
```diff
@@ -4,7 +4,7 @@
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import Summary from '@woocommerce/base-components/summary';
-import { getSetting } from '@woocommerce/settings';
+import { WORD_COUNT_TYPE } from '@woocommerce/block-settings';
```
nerrad, line 7: Let's switch back to use `getSetting` here. I'd like us to move away from using `block-settings` and use `getSetting` instead. That helps make the dependency much clearer.

This comment applies to all the other places modified to use `getSetting`.

assets/js/settings/blocks/constants.js
nerrad: I agree with Nadir here, this is a pretty significant breaking change and should be reconsidered. I'm not sure I agree with the solution in this refactor. I agree there needs to be a way for each block to register what data it needs (without triggering the error from duplicate keys) but I would prefer to see that happen at the block level. To avoid having to check if the data is available before registering (via `AssetsDataRegistry::exists`) - we could add a new helper method `addIfNotExists` or something like that which absorbs the necessary `exists` check and avoids registering duplicate data on the same key.

Then blocks that share data could potentially do the registration via a method on the abstract block class.

Some variation of the above should work and allows us to preserve back compat.

src/Assets/AssetDataRegistry.php
```diff
@@ -127,11 +127,13 @@ protected function initialize_core_data() {
...
 		 */
-		$settings = apply_filters(
-			'woocommerce_shared_settings',
-			$this->data
+		$message = __(
+			'The filter should not be used for assets registration. Please read woocommerce-gutenberg-products-block/docs/contributors/block-assets.md for more information on how to proceed.',
+			'woo-gutenberg-products-block'
 		);
 
+		$settings = apply_filters_deprecated( 'woocommerce_shared_settings', [ $this->data ], '4.6.0', '', $message );
```
nerrad, line 135: Nice!

src/Domain/Bootstrap.php
```diff
@@ -153,12 +152,8 @@ function ( Container $container ) {
...
 			AssetDataRegistry::class,
 			function( Container $container ) {
-				$asset_api        = $container->get( AssetApi::class );
-				$load_back_compat = defined( 'WC_ADMIN_VERSION_NUMBER' )
-					&& version_compare( WC_ADMIN_VERSION_NUMBER, '0.19.0', '<=' );
-				return $load_back_compat
-					? new BackCompatAssetDataRegistry( $asset_api )
-					: new AssetDataRegistry( $asset_api );
+				$asset_api = $container->get( AssetApi::class );
+					return new AssetDataRegistry( $asset_api );
```
nerrad, line 156: Glad to see this removed 👏

nerrad (changes requested):
@budzanowski  I realize you won't be following up on this given your rotation, but just thought I'd leave my review here. I might pick this up this week so we can get it merged.  Thanks for the work you did on this!

Issue #1962, opened 2020-03-13
Move base/components into subfolders if they are only relevant to one specific block


Aljullu:
Initially discussed in #1949.

In `base/components` we have several components which are only relevant to specific blocks, so we would like to move them into subfolders to make that clearer.

Some suggestions:
* [x] `address-form`, `shipping-calculator`, `shipping-location`, `shipping-rates-control` and `totals` could go under `cart-checkout` (#2034).
* [x] `checkout` subfolder and `cart-checkout` could be merged (#2034).
* [x] `review-list`, `review-list-item` and `review-sort-select` could go under `reviews` (#2034).
* [ ] `filter-submit-button`, `product-list`, `product-list-item`, `product-sort-select` could go under `all-products-and-filters` (I'm not 100% sure about this one, it seems a bit too broad and it doesn't match a folder under `blocks`).
* [ ] `payment-methods` should go under `cart-checkout`.

ralucaStan:
@Aljullu regarding the `all-products-and-filters` suggestion, I agree it seems too broad.

I identified these dependencies: 

- filter-submit-button -> attribute filter & price filter (so I suggest not to move it)
- product-list -> all products
- product-list-item -> product-list
- product-sort-select -> product-list

I would group the product list related components:
- weather we move `product-sort-select` and `product-list-item` under `product-list` and keep them in base components
- or we move all `product-list,-list-item,-sort-select` under `all products` in block, but I did not see this as a pattern.

What do you think? Should we also bring this up with the team, component structure is always an issue?

Aljullu:
I would lean towards this option:

> weather we move `product-sort-select` and `product-list-item` under `product-list` and keep them in base components

In the future, some PHP-based product grid blocks might end up being refactored to use the same components as All Products, so a `product-list` folder which is generic for all blocks that display products sounds better than having a different folder for each block.

ralucaStan:
Ok, this option made more sense for me as well. I will create a PR for it.

PR #3866, opened 2021-02-18
Move payment-methods folder from base components to cart-checkout block

ralucaStan:
Part of #1962.

This PR moves the  `payment-methods` folder, as discussed in #1962. It doesn't close the issue completely, as there are other folders to be moved. 

What is done in this PR:
- `/base/components/payment-methods` is now under `blocks/cart-checkout`

1. Verify the new folder structure and whether it makes sense.
2. Make sure the project still builds and do some smoke testing to verify _Cart_, _Checkout_ and blocks continue working.


mikejolley:
There appears to be a conflict already :)

ralucaStan:
> There appears to be a conflict already :)

Should be good now, it was a small change on a comment in one of the moved files.

mikejolley (approved):
Built and works. I think the new location makes more sense than the old. I was going to suggest moving payment-methods under checkout but then I remembered there are express methods sometimes on the cart, so this is fine 👍🏻

PR #3869, opened 2021-02-19
Replace uppercase Translators with translators & make comment style consistent

### LINKED ISSUE #3764: Replace all uppercase Translators string with translators
According to the [WordPress i18n documentation](https://developer.wordpress.org/apis/handbook/internationalization/internationalization-guidelines/#descriptions), the comment used to give context to translators should be written as `translators:` in lowercase. 

We have several instances of `Translators:` being used in our repository.

We should replace `Translators:` with `translators:` across the codebase.

opr:
When translating text, you should add a comment to the preceding line explaining what the translation will be used for, if you think extra context needs to be given to translators.

The way this is done is to comment like: `/* translators: %s is the product name */` as per the [WordPress documentation](https://codex.wordpress.org/I18n_for_WordPress_Developers) - in our codebase we had a mixture of uppercase `Translators` strings, and `// translators:` style comments - while this still _worked_ it sometimes formatted the translation hint incorrectly, showing the `Translators` word when it shouldn't have been. Not a huge problem but it would be nice to keep the plugin polished.

I also changed all comment styles to `/* ... */` rather than `// ... ` to keep this consistent across our codebase.

Fixes #3764

1. View code changes and ensure only comment lines have changed
2. Try translating some strings using LocoTranslate or some other translation method, you should only see the translation hint, e.g. `%s is the product name` not `Translators: %s is the product name`

### DIFF
--- a/assets/js/atomic/blocks/product-elements/add-to-cart/shared/add-to-cart-button.js
+++ b/assets/js/atomic/blocks/product-elements/add-to-cart/shared/add-to-cart-button.js
@@ -127,7 +127,7 @@ const ButtonComponent = ( {
 		>
 			{ isDone && quantityInCart > 0
 				? sprintf(
-						// translators: %s number of products in cart.
+						/* translators: %s number of products in cart. */
 						_n(
 							'%d in cart',
 							'%d in cart',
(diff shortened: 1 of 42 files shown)


Aljullu (approved):
LGTM :ship:

PR #3879, opened 2021-02-22
Show total sale badge in medium carts & make it display below price

opr:
When the price and sale badge are short, they appear on the same line. During testing, the product sale badge was usually long enough to cause it to break onto a new line so this went unnoticed.

By setting the class `.wc-block-components-product-price` to be block it ensures the badge will always display below the price.

I also noticed that the sale badge didn't display when the block size was medium, so I have allowed it to display there. This was especially noticeable in the TwentyTwenty theme when the block size is always medium.

#### Before
image

#### After
image

1. Set your currency settings to display without a decimal place, like this: image
1. Add a product that is on sale to your cart, if you've got an up to date WooCommerce Subscriptions repo (on branch: `feature/checkout-block-simple-multiple-subscriptions`) then add a subscription product that's on sale, too.
2. View the Cart block and ensure the price and sale badge display well.
3.

### Changelog

> Ensure the sale badge is displayed well when a product price is short in the Cart block.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
@@ -150,6 +150,14 @@ table.wc-block-cart-items {
 	display: flex;
 }
 
+.is-small {
+	.wc-block-cart-item__total {
+		.wc-block-components-sale-badge {
+			display: none;
+		}
+	}
+}
+
 .is-medium,
 .is-small,
 .is-mobile {
@@ -201,10 +209,6 @@ table.wc-block-cart-items {
 
 				.wc-block-components-formatted-money-amount {
 					display: inline-block;
-				}
-
-				.wc-block-components-sale-badge {
-					display: none;
 				}
 			}
 		}
--- a/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/style.scss
@@ -150,7 +150,7 @@ table.wc-block-cart-items {
 	display: flex;
 }
 
-.wc-block-cart-item__total-price-and-sale-badge {
+.wc-block-cart-item__total-price-and-sale-wrapper {
 	display: flex;
 	flex-direction: column;
 	align-items: flex-end;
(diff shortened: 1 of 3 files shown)


ralucaStan:
This solution works but I see that ProductPrice has this class and there might be cases where it's needed to be inline.

What we could do is style locally, depending on the need.

So I am thinking of wrapping the price and disscount in a div and setting that as a flex container, with a column direction (style is inline for demo purpose only):
```html
<td class="wc-block-cart-item__total">
  <div style="
      display: flex;
      flex-direction: column;
      align-items: flex-end;
  ">
    <span class="price wc-block-components-product-price">...</span>
    <div class="wc-block-components-product-badge wc-block-components-sale-badge">...</div>
  </div>
</td>
```

This way the elements will display correctly here. 
What do you think? @opr

assets/js/blocks/cart-checkout/cart/full-cart/style.scss
```diff
@@ -150,6 +150,14 @@ table.wc-block-cart-items {
 	display: flex;
 }
 
+.is-small {
```
ralucaStan: I wonder why this is hidden on mobile?

ralucaStan (changes requested):
Let me know what you think of the proposal to handle the display locally.

assets/js/blocks/cart-checkout/cart/full-cart/style.scss
opr: This looks like it was hidden on mobile because of #1587 - I think it looks too crowded on mobile if it's added back, but on medium it looks OK. @mikejolley since you authored the original PR there, what do you think about adding the sale badge back on medium?

opr:
Thanks for the review @ralucaStan I've made the change you suggested, I would like to hear your thoughts on my choice of class name, `wc-block-cart-item__total-price-and-sale-badge` doesn't feel that great to me, but if you think it works then fine. Any other suggestions for this class name are welcome though.

ralucaStan:
@opr how about `wc-block-cart-item__total-price-container` or `wc-block-cart-item__total-price-and-sale-wrapper`

ralucaStan (approved):

PR #3881, opened 2021-02-22
Add to cart batching using Data Loader

mikejolley:
WIP

Batching/defer of add to cart requests to mitigate https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2094


mikejolley:
See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3899

PR #3889, opened 2021-02-25
Update Label component to TSX

opr:
The changes in this PR are:
- Add `tsconfig.json` to the `components` directory.
- Update the root `tsconfig.json` to include this project
- Rename `label/index.js` to `label/index.tsx` and convert to TypeScript.

I have not removed the JSDoc comment because I don't know if there's any alternative for typing the props we pass in.

While using JSDoc alongside TS I get this 
image
 I found https://github.com/TypeStrong/typedoc/issues/278#issuecomment-240758927 (Maybe not applicable) and https://stackoverflow.com/questions/51641378/is-there-a-way-to-document-fields-in-an-interface-in-typescript but removing the JSDoc comments and replacing them with the examples from the linked GitHub issue does not work. Neither in PHPStorm or VSCode, so maybe we want to leave the JSDoc comments in place, but maybe remove the lint that requires us to type them?

1. View a component in the browser that uses `Label` and ensure it displays properly
2. Add a `Label` component somewhere in the codebase and ensure intellisense shows the correct props and types


assets/js/base/components/label/index.tsx
```diff
@@ -1,27 +1,29 @@
 /**
  * External dependencies
  */
-import PropTypes from 'prop-types';
-import { Fragment } from 'react';
+import { Fragment } from '@wordpress/element';
```
nerrad, line 4: We should make sure we're importing from `@wordpress/element` as much as possible (hence the change I made here in 8899541)

assets/js/base/components/label/index.tsx
```diff
@@ -1,27 +1,29 @@
...
 import classNames from 'classnames';
+import type { HTMLAttributes } from 'react';
```
nerrad, line 6: Importing this as a type from `react` because:

- `HTMLAttributes` is not exported from `@types/wordpress__element`.
- We don't have anything other than the interface imported here.

(I made the change in 8899541)

assets/js/base/components/label/index.tsx
```diff
@@ -1,27 +1,29 @@
...
+
+interface LabelProps {
+	label?: string;
+	screenReaderLabel?: string;
+	wrapperElement?: string;
+	wrapperProps?: HTMLAttributes< HTMLElement >;
+}
```
nerrad, line 13: Contrary to what we discussed in the meetup, I think we should be explicitly defining interfaces and implementing. In an internal slack convo:

> ..the inline object there is actually a type alias I believe. It will behave very similarly but interfaces are preferred whenever possible.

See also [this reference](https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections).

The implication being that using interfaces can have some performance benefits.

I think this only applies in some scenarios, but given this seems to be the convention being pointed to in A8C implementation of TS and is easier to recommend over thinking about whether to use or not - I think it makes sense to default to defining interfaces.

I made the change here in 8899541.

nerrad (approved):
Looks good! I added some changes as noted.

I also rebuilt the `package-lock.json` because it had been built using `npm 7+` and we don't support that version of npm yet.

I'm approving on the assumption you've tested a build and ensured nothing is broken.

nerrad (changes requested):
Cancelling my approval because of the failing test (which I'm looking into now).

nerrad (approved):
Okay 7c5bdb8 fixes the jest configuration. So back to ✅ for this (again assuming you've tested a build with this branch to make sure nothing is broken with anything using the label component).

nerrad:
> I have not removed the JSDoc comment because I don't know if there's any alternative for typing the props we pass in.

I went ahead and removed this in my commit because I've already turned off the jsdoc rule for param comments in .ts files:

```
rules: {
    '@typescript-eslint/no-explicit-any': 'error',
    'no-use-before-define': 'off',
    '@typescript-eslint/no-use-before-define': [ 'error' ],
    'jsdoc/require-param': 'off',
},
```

Aljullu (approved):

PR #3890, opened 2021-02-25
Improvements to payment methods integration docs

### LINKED ISSUE #3000: Improve documentation on PHP side payment method registration
The [documentation for PHP side payment method integration](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/docs/extensibility/payment-method-integration.md) with the Checkout block can be improved. While it includes some instructions about what hooks to implement etc, it's still vague on actual registration of new payment method integrations.

Ideally it'd be good to have an example payment method integration as a reference. The PM's we integrate in the blocks plugin are helpful but also include usage of the container etc which really isn't needed by Payment methods. Also our usage of `AssetApi` can be a bit confusing to implementers because `AssetApi` is only usable in the Blocks plugin context.

nerrad:
Fixes: #3000 

This PR is the result of a pass through our existing documentation and improving it to include more details around the server side payment method integration with Cart and Checkout blocks. There are also some improvements/fixes for the client documentation as well.

We will be making this more public so I've also removed references to this being a work in progress which means we are committing more publicly to this being the official API interface for payment method integrations.


docs/extensibility/payment-method-integration.md
```diff
@@ -2,23 +2,25 @@
...
 - [Server Side Integration](#server-side-integration)
   - [Processing Payment](#processing-payment)
   - [Registering Assets](#registering-assets)
+  - [Hooking into the Checkout processing by the Store API.](#hooking-into-the-checkout-processing-by-the-store-api)
+  - [Putting it all together](#putting-it-all-together)
 
 ## Client Side integration
 
-The client side integration consists of an API for registering both _express_ payment methods (those that consist of a one button payment process initiated by the shopper such as Stripe ApplePay or GooglePay), and payment methods such as _cheque_, Paypal Standard, or Stripe Credit Card.
+The client side integration consists of an API for registering both _express_ payment methods (those that consist of a one button payment process initiated by the shopper such as Stripe ApplePay or GooglePay), and payment methods such as _cheque_, PayPal Standard, or Stripe Credit Card.
```
loranallensmith: Looks like you may want some serial commas for the express methods you mention here:

```suggestion
The client side integration consists of an API for registering both _express_ payment methods (those that consist of a one-button payment process initiated by the shopper such as Stripe, ApplePay, or GooglePay), and payment methods such as _cheque_, PayPal Standard, or Stripe Credit Card.
```

docs/extensibility/payment-method-integration.md
```diff
@@ -2,23 +2,25 @@
...
+
+In both cases, the client side integration is done using registration methods exposed on the `blocks-registry` API. You can access this via the `wc` global in a WooCommerce environment (`wc.wcBlocksRegistry`).
 
-In both cases, the client side integration is done using registration methods exposed on the `blocks-registry` API. You can access this via the `wc` global in a WooCommerce environment (`wc.wcBlocksRegistry`). You'll see that we also Webpack configured in the blocks repository to expose this API on `@woocommerce/blocks-registry` which you can implement in your own build process as well.
+> Note: In your build process, you could do something similar to what is done in the blocks repository which [aliases this API as an external on `@woocommerce/blocks-registry`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e089ae17043fa525e8397d605f0f470959f2ae95/bin/webpack-helpers.js#L16-L35).
```
loranallensmith, line 23: It could be worth mentioning the [WC Dependency Extraction Webpack Plugin](https://www.npmjs.com/package/@woocommerce/dependency-extraction-webpack-plugin) as an option here too.  I think it accounts for the wcBlocksRegistry as part of its default settings.

docs/extensibility/payment-method-integration.md
```diff
@@ -62,15 +64,15 @@ const options = {
...
+-   `content` (required): This should be a React node that will output in the express payment method area when the block is rendered in the frontend. It will be cloned in the rendering process. When cloned, this React node will receive props from the payment method interface to checkout that will allow your component to interact with checkout data (more on this later).
```
loranallensmith: I have a bit of trouble following the wording here.  Are you saying that the cloned node receives props from the interface between the payment method and the checkout block?  Or are you saying that the node receives props from the payment method interface, which it needs in order to interact with checkout data during the checkout process?

docs/extensibility/payment-method-integration.md
```diff
@@ -62,15 +64,15 @@ const options = {
...
+-   `canMakePayment` (required): A callback to determine whether the payment method should be available as an option for the shopper. The function will be passed an object containing data about the current order. Return a boolean value - true if payment method is available for use. If your gateway needs to perform async initialization to determine availability, you can return a promise (resolving to boolean). This allows a payment method to be hidden based on the cart, e.g. if the cart has physical/shippable products (example: [`Cash on delivery`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e089ae17043fa525e8397d605f0f470959f2ae95/assets/js/payment-method-extensions/payment-methods/cod/index.js#L48-L70)); or for payment methods to control whether they are available depending on other conditions. **Keep in mind this function could be invoked multiple times in the lifecycle of the checkout and thus any expensive logic in the callback provided on this property should be memoized.**
```
loranallensmith, line 69: This tip about memoization is :sparkles:!  :smiley:

docs/extensibility/payment-method-integration.md
```diff
@@ -62,15 +64,15 @@ const options = {
...
 -   `paymentMethodId`: This is the only optional configuration object. The value of this property is what will accompany the checkout processing request to the server and used to identify what payment method gateway class to load for processing the payment (if the shopper selected the gateway). So for instance if this is `stripe`, then `WC_Gateway_Stripe::process_payment` will be invoked for processing the payment.
```
loranallensmith: I think these existing docs might be missing a word:
```suggestion
-   `paymentMethodId`: This is the only optional configuration object. The value of this property is what will accompany the checkout processing request to the server and is used to identify what payment method gateway class to load for processing the payment (if the shopper selected the gateway). So for instance if this is `stripe`, then `WC_Gateway_Stripe::process_payment` will be invoked for processing the payment.
```

docs/extensibility/payment-method-integration.md
```diff
@@ -94,9 +96,9 @@ registerPaymentMethod( options );
 
 The options you feed the configuration instance are the same as those for express payment methods with the following additions:
 
--   `label`: This should be a react node that will be used to output the label for the tab in the payment methods are. For example it might be `<strong>Credit/Debit Cart</strong>` or you might output images.
+-   `label`: This should be a React node that will be used to output the label for the tab in the payment methods are. For example it might be `<strong>Credit/Debit Cart</strong>` or you might output images.
```
loranallensmith: I think there might be a missing word here:

```suggestion
-   `label`: This should be a React node that will be used to output the label for the option where the payment methods are. For example it might be `<strong>Credit/Debit Cart</strong>` or you might output images.
```

docs/extensibility/payment-method-integration.md
```diff
@@ -111,35 +113,35 @@ A big part of the payment method integration is the interface that is exposed fo
...
--   `components`: The properties on this object are exposed components that can be implemented by your payment method for various common interface elements used by payment methods. Currently the available components on this property are: `ValidationInputError` (a container for holding validation errors which typically you'll include after any inputs), and `CheckboxControl`(which is usually used for indicating to save the payment method). **Note: this last one is subject to change**
+-   `components`: The properties on this object are exposed components that can be implemented by your payment method for various common interface elements used by payment methods. Currently the available components on this property are: `ValidationInputError` (a container for holding validation errors which typically you'll include after any inputs), [`PaymentMethodLabel`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e089ae17043fa525e8397d605f0f470959f2ae95/assets/js/payment-method-extensions/payment-methods/paypal/index.js#L37-L40) (use this component for the payment method label, including an optional icon)
```
loranallensmith: It might be worth creating bullets for these components to improve readability:

```suggestion
-   `components`: The properties on this object are exposed components that can be implemented by your payment method for various common interface elements used by payment methods. Currently the available components on this property are:
  - `ValidationInputError` — a container for holding validation errors which typically you'll include after any inputs
  - [`PaymentMethodLabel`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e089ae17043fa525e8397d605f0f470959f2ae95/assets/js/payment-method-extensions/payment-methods/paypal/index.js#L37-L40) — use this component for the payment method label, including an optional icon
```

I wasn't sure about the context of the link for the `PaymentMethodLabel` component.  It looks like it leads to an example of how it's used.  It might be worth linking to the [component's definition](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f3b2b53435ffbc9ac49215846bad08ce74c7124b/assets/js/base/components/cart-checkout/payment-method-label?/index.js) in the name instead and then separately link to the example with text along the lines of "You can see an example of this component in the PayPal payment method integration.

(👆 and ditto for the `ValidationInputError` component too)  :smiley:

docs/extensibility/payment-method-integration.md
```diff
@@ -111,35 +113,35 @@ A big part of the payment method integration is the interface that is exposed fo
...
 
 ## Server Side Integration
 
-> **Note:** This portion of the doc is still in development and there could be changes to the API here.
-
 ### Processing Payment
 
-Currently, the checkout block has legacy handling for payment processing (by converting incoming `payment_data` provided by the client side payment method to `$_POST` and calling the payment gateway's `process_payment` function ). So unless your payment method hooks into the core checkout `process_checkout` function in any way, this will take care of processing your payment for you.
+Currently, the checkout block has legacy handling for payment processing (by converting incoming `payment_data` provided by the client side payment method to `$_POST` and calling the payment gateway's `process_payment` function ). So unless your payment method hooks into the core checkout `process_checkout` function in any way, this will take care of processing your payment for you server side if you already have a WooCommerce Payment method extension integrated with the existing shortcode checkout flow.
```
loranallensmith: I found the wording a bit challenging to follow here.  What about something like this:

```suggestion
The checkout block currently has legacy handling for payment processing.  It converts incoming `payment_data` provided by the client-side payment method to `$_POST` and calls the payment gateway's `process_payment` function.  If you already have a WooCommerce Payment method extension integrated with the existing shortcode checkout flow, the checkout block's legacy handling will take care of processing your payment for you on the server side.  However, If your payment method hooks into the core checkout `process_checkout` function in any way, you will need to account for this behavior and make appropriate adjustments.  (See the section below about hooking into the checkout process via the Store API.)
```

docs/extensibility/payment-method-integration.md
```diff
@@ -111,35 +113,35 @@ A big part of the payment method integration is the interface that is exposed fo
...
 
 ### Registering Assets
 
-Since implementing the correct loading of your client side asset registration is tricky with the variables around the block loading process, there is an API for ensuring you can register any assets and data to pass along to your client side payment method from the server.
+Since implementing the correct loading of your client side asset registration is tricky with some dependencies on how related assets load in the request, there is an API for ensuring you can register any assets and data to pass along to your client side payment method from the server.
```
loranallensmith: I had a bit of trouble understanding the wording here.  Are you saying that the abstract class (and interface) below are available to help payment method developers make sure that their client-side scripts get registered/enqueued/served correctly alongside related scripts for WC core, Blocks, etc.?

docs/extensibility/payment-method-integration.md
```diff
@@ -152,3 +154,31 @@ This contains various details about the payment result returned to the client an
 -   set the status to return for the payment method (one of `success`, `failure`, `pending`, `error`).
 -   set a redirect url.
 -   set any additional payment details (in case you need to return something for your client to further process with).
+
+### Putting it all together
```
loranallensmith, line 160: I :heart: the way you've broken this section down into an explanation of steps followed by an example of the result.

loranallensmith (commented):
Thanks so much for all your work updating these docs @nerrad!  🙌  Everything is looking really great to me.  I made a few inline comments/suggestions, most of which relate to wording or punctuation.  :smiley:

I tried to add a comment to the paragraph around Line 148 of `payment-method-integration.md` but I think it was too far from the diff, so I'll mention it here.  It might be worth adjusting the word ordering in that explanation of `PaymentContext`.  Here's a recommendation:

```
This contains various details about the payment extracted from the checkout processing request. Notable is the `payment_data` property that will contain an associative array of data your client-side payment method provided to checkout. It also contains a string value for `payment_method` which contains the `paymentMethodId` value for the active payment method used during checkout processing. So you can use this to determine whether your payment method processes this data or not.
```

Overall this is looking really great!  I appreciate the ping on this PR and can't wait for us to share these updated docs with more folks.

docs/extensibility/payment-method-integration.md
nerrad: I thought about that, but in this case the Payment Methods API explicitly takes care of this via how it actually registers the registered payment method asset handle as a [dependency of the blocks assets](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0d432032706987ee5d17659850491f559c845499/src/Payments/Api.php#L64-L69) . 

I'm wary about introducing too much into this doc about how developers would register their scripts without it straying into more detailed tutorial territory.

docs/extensibility/payment-method-integration.md
nerrad: The latter. Your question indicates this can be worded better - I'll give another go at it :)

docs/extensibility/payment-method-integration.md
nerrad: Yes, basically what happens is the payment method registers their handle and the payment method registry surfaces these handles to be registered as dependencies on the actual block assets. The reason this is needed because while the `WP_Scripts` dependency handling works well for indicating what scripts need to load _before_ my JS file, there's no built in way for me to indicate that my script must load _before_ another registered script in another extension. 

I see how I can reword this a bit so I'm going to give a go.

docs/extensibility/payment-method-integration.md
nerrad: I ended up adding this in c5b4b09 :)

nerrad:
Thanks @loranallensmith  for all the great feedback on this! I've applied most of your suggestions.

docs/extensibility/payment-method-integration.md
```diff
@@ -48,10 +50,10 @@ The options you feed the configuration instance should be an object in this shap
 
 ```js
 const options = {
-	name: 'my_payment_method',
-	content: <div>A react node</div>,
-	edit: <div> A react node </div>,
-	canMakePayment: () => true,
+  name: 'my_payment_method',
```
mikejolley, line 53: Non-blocking, we could include the "details" content below inline in code comments within this example so the description is right next to the example. Might make it easier to follow.

mikejolley (approved):
Nice job @nerrad I only made a small comment, nothing else from me.

docs/extensibility/payment-method-integration.md
nerrad: Ya I thought about that, the difficulty is that the details linking out won't work (because code blocks stop the link parsing in md).

PR #3893, opened 2021-02-26
Followups and improvements on TS implementation in the Cart data store.

nerrad:
This PR is a followup to #3768 that improves the TS implementation in the Cart data store. In this PR:

- `@automattic/data-stores` is installed as a dev package to gain access to some helpful types/generics for use in our data stores (in particular [mapped-types](https://github.com/Automattic/wp-calypso/blob/61acc957638fc7009632af4c0c058aa6bc0e82b7/packages/data-stores/src/mapped-types.ts) which are implemented in the Cart Store registration (see 5d93996).
- Fixing TSLint warnings for the actions related to the `explicit-module-boundary-types` rule.  There's now only one warning for that lint rule. I'm thinking we might want to turn off this rule for typescript files (it's not handling the `as const` declaration well on non-arrow functions).
- Improving some typing in the reducer, controls and main entry point to remove usage of the `@ts-ignore` (as things are now properly typed! 🎉 ).

## To test

The changes here should have nil impact on production runtime. However, it does impact anything using the cart store so smoke testing of Cart and Checkout block (ensure coupons work and various interactivity related to cart data works) should be done.

### DIFF
--- a/assets/js/data/cart/index.ts
+++ b/assets/js/data/cart/index.ts
@@ -3,6 +3,7 @@
  */
 import { registerStore } from '@wordpress/data';
 import { controls as dataControls } from '@wordpress/data-controls';
+import type { DispatchFromMap, SelectFromMap } from '@automattic/data-stores';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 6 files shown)


assets/js/data/cart/index.ts
```diff
@@ -11,19 +12,25 @@ import { STORE_KEY } from './constants';
...
 export const CART_STORE_KEY = STORE_KEY;
+
+declare module '@wordpress/data' {
+	function dispatch(
+		key: typeof CART_STORE_KEY
+	): DispatchFromMap< typeof actions >;
+	function select(
+		key: typeof CART_STORE_KEY
+	): SelectFromMap< typeof selectors >;
+}
```
opr, line 36: Will we need to do this for collections and query-state too? Will this cause problems, or would we move this augmentation to a "central" location and just add those types to it?

assets/js/data/cart/reducers.ts
```diff
@@ -152,4 +153,6 @@ const reducer = (
 	return state;
 };
 
+export type State = ReturnType< typeof reducer >;
```
opr, line 156: Nice 😎

opr (approved):
This is great Darren, I think that disabling the `explicit-module-boundary-types` lint rule is probably fine here, especially since we're using `as const` I don't think there will be a problem. The app tests fine with these changes too. I have left a comment with a question about module augmentation.

I am pre-approving this on the condition that the module augmentation in `cart/index.ts` can be moved and further augmented without issue.

assets/js/data/cart/index.ts
nerrad: > Will we need to do this for collections and query-state too?

 Yes.

> Will this cause problems, or would we move this augmentation to a "central" location and just add those types to it?

Not sure yet. We may have to change how we export the stores and do something similar to how calypso does it for some of their data stores.

PR #3897, opened 2021-03-01
Inherit line height in Cart item total

opr:
This PR will change the line height of the price and sale badge (and any other items that may appear in the total column of a Cart line item). Previously, this line height being set was causing the sale badge in this column to appear differently to the sale badge under the product name.

#### Before
image

#### After
image

1. Add an item that's on sale to your cart then go to the cart block. Increase its quantity to 2 or more.
2. Verify the sale badges are the same height.
3. Try this with a subscription product and our branch of subscriptions checked out. The text in the sales badges will be longer but should still display well.
4. Try this in other themes to ensure the badges display the same in them.

### Changelog

> Ensure sale badges have a uniform height in the Cart block.


mikejolley (approved):
👍🏻

PR #3899, opened 2021-03-01
Rest API batching support

### LINKED ISSUE #2094: All Products: add to cart can fail when adding multiple products to cart in quick succession
### To reproduce

- Set up a page with all products block, showing lots of products which can be added to cart directly (no variations/options).
- Click add to cart for multiple products in quick succession. 

Notice only the first product correctly shows the updated `1 in cart ` feedback text. On the other product(s), the button switches back initial `Add to cart` text.

mikejolley:
This PR implements batching in the Store API, and a dataloader class in the client to batch requests within a certain interval.

Batching works for POST, PUT, DELETE API requests. GET requests are not affected.

As well as adding the batching, [this commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/d0b901b06174aa514efaf23bec1482db13f6b276) creates a control to handle fragment updates and adds a timeout. This prevents multiple fragment updates after a batch of events have ran.

NOTE: I am requested feedback/code review, but some things are left to do before it can be merged:

- [x] Check what happens if nonce changes on items in the batch.
- [x] Batch is only supported on 5.6+ so we need the client to account for this also.
- [x] Reduce batch timeout from 3000 - I have left this high to make testing easier.

Fixes #2094

1. Go to the products page.
2. Add to items to the cart really quickly. If you focus a button with the keyboard you can hit enter and click another button.
3. Check the network request for a "batch" API call. You'll see 2 responses in an array.
4. Smoke test cart page functionality/cart/checkout. All should be unaffected.


senadir (changes requested):
I didn't do a code review yet.

While testing cart, I can't do any action, the response is coming back with `woocommerce_rest_missing_nonce` error but no notice is visible.

I'm getting the same error in All Products with an empty red notice
image

mikejolley:
@senadir I think you have a build issue, I'm not seeing that at all, and tests don't indicate a problem.

mikejolley:
I've pushed some changes to make nonce-validation run earlier so it cannot change during a batch request.

assets/js/data/cart/actions.ts
```diff
@@ -145,6 +145,17 @@ export const shippingRatesBeingSelected = ( isResolving: boolean ) =>
 		isResolving,
 	} as const );
 
+/**
+ * Returns an action object for updating legacy cart fragments.
+ *
+ * @return {Object} Action object.
+ */
+export function updateCartFragments() {
```
opr: nit: If you write this as a function expression rather than a function declaration you would not get a lint warning.

opr (commented):
This looks good to me, I tested it and see the requests are batched. I left one nitpick comment about TypeScript. 

Code wise everything looks fine, but could we add some PHP unit tests for the batch route?

mikejolley:
@opr thanks! I've added some tests for this batch endpoint and reduced the timeout to 200ms. Should be good to go if you're happy.

opr (approved):
> @opr thanks! I've added some tests for this batch endpoint and reduced the timeout to 200ms. Should be good to go if you're happy.

Nice, the tests look good Mike! Thanks 🥳

mikejolley:
This PR is failing e2e tests in the cart, but testing manually I see no issues. I think this is just e2e timeout-related rather than a new bug. I will merge...

nerrad:
@mikejolley I had to revert (done through git client not GitHub) this work in trunk for the following reasons:

- The e2e test failures were legit. The Cart Update middleware added in #3960 is broken with these changes.
- Updating the quantity in the cart for a product failed with this JSON response:

```json
{
  "responses": [
    {
      "body": {
        "code": "woocommerce_rest_missing_nonce",
        "message": "Missing the X-WC-Store-API-Nonce header. This endpoint requires a valid nonce.",
        "data": { "status": 401 }
      },
      "status": 401,
      "headers": { "Allow": "POST" }
    }
  ]
}
```

There may be other issues too with other cart updates but I didn't check. I felt it safest to revert for now so this isn't blocking other work while it's being addressed.

I did test adding something like this for the [middleware](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/middleware/cart-update.ts) (which fixes the timestamp issue for local storage surfacing in e2e test failures and confirmed in manual testing), but it does not address the nonce issue (not expected to and I didn't dig further to find out the potential cause for the nonce issue). I thought I'd share the code here so it can be used for fixes.

```js
const isCartUpdatePostRequest = ( options: APIFetchOptions ) => {
	const url = options.url || options.path;
	if ( ! url || ! options.method || options.method !== 'POST' ) {
		return false;
	}
	return /wc\/store\/cart\//.exec( url ) !== null;
	let isCartRoute = !! /wc\/store\/cart\//.exec( url );

	// Might be batch route, if so, let's check if cart route is one of the
	// batch endpoints.
	if ( ! isCartRoute && !! /wc\/store\/batch/.exec( url ) ) {
		const requests = options?.data?.requests || [];
		for ( const request of requests ) {
			const path = request?.path || '';
			isCartRoute = !! /wc\/store\/cart\//.exec( path );
			if ( isCartRoute ) {
				break;
			}
		}
	}

	return isCartRoute;
};
```

nerrad:
@opr I think we may also want to revisit [these e2e tests](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/e2e/specs/frontend/cart.test.js#L52-L149) you added for the work in #3960. One problem with them is that when the timeout happens for the test not able to retrieve the data from the `localStorage` the error message for the test is ambiguous and makes it a bit harder to capture the reason for why it may be happening. One thing we could try is to wrap the execution of `waitForFunction` in a try/catch and if there is a timeout error we could fail with a better message?

PR #3907, opened 2021-03-02
Sync cart item quantity if its Implicitly changed.

senadir:
The main way of changing a cart item quantity is via the quantity selector, but quantity can be changed imperatively from the server, or from another source by directly calling the datastore. This PR fixes it so it's always synced.

This happens because we keep a local copy of the item quantity in a state for optimistic updates.

- We can move that state to the component level instead of the hook level and sync there.
- We can move the optimistic update section to the data store.
- We can react to changes in the data store and sync them back to the state in our hook.

The 1st option would require us to do the 3rd anyway, so I opted for the 3rd now.

### testing

- Add an item to your cart.
- In console, run `wp.data.dispatch('wc/store/cart').changeCartItemQuantity(wcSettings.cartData.items[0].key, 10)` this would change the quantity of your first item to 10.
- Make sure the quantity selector is now 10.
- Try updating it from the selector.

### changelog

> Fix a bug in which cart item quantity didn't react to the changes from server.

### DIFF
--- a/assets/js/base/hooks/cart/use-store-cart-item-quantity.js
+++ b/assets/js/base/hooks/cart/use-store-cart-item-quantity.js
@@ -38,6 +38,12 @@ export const useStoreCartItemQuantity = ( cartItem ) => {
 	// Store quantity in hook state. This is used to keep the UI
 	// updated while server request is updated.
 	const [ quantity, changeQuantity ] = useState( cartItemQuantity );
+	// Quantity can be updated from somewhere else, this is to keep it in sync.
+	useEffect( () => {
+		if ( cartItemQuantity !== quantity ) {
+			changeQuantity( cartItemQuantity );
+		}
+	}, [ cartItemQuantity, quantity ] );
 	const [ debouncedQuantity ] = useDebounce( quantity, 400 );
 	const previousDebouncedQuantity = usePrevious( debouncedQuantity );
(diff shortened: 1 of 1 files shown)


Aljullu (commented):
> quantity can be changed imperatively from the server, or from another source by directly calling the datastore

Do you know any situation when this might happen? I would like to get into the user's shoes when testing this, since in the other hand I suspect that changing quantity values without user input might cause some confusion.

For now, I had trouble testing this. It seems like changing the quantity from the UI is not working anymore? Can you reproduce?

quantity-selector-not-working

senadir:
> Do you know any situation when this might happen? I would like to get into the user's shoes when testing this, since in the other hand I suspect that changing quantity values without user input might cause some confusion.

This came out from [this issue (long thread).](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3895)
- Some grouped products will all change quantity together.
- A coupon or some option might trigger quantity change in the server.
- Calling data layer would update the data.

Overall, regardless of those options, the selectQuantity source of truth should be the store, the original behaviour would cause it to be out of sync.

senadir:
Fixed the issue @Aljullu

Aljullu (approved):
Thanks for the explanation, makes sense! This is working great now. :ship:

PR #3908, opened 2021-03-02
Update extensions types and fix inconsistent type from server.

senadir:
Extensions data wasn't inside our preview data, so it would break Cart and Checkout in the editor if you had WCS Subscriptions integration on.

It also updates some types to the correct format.

I also added a new param to `extendRestApi::register_endpoint_data` `schema_type` which can be either `ARRAY_N` or `ARRAY_A` to signal if the data schema should be an object or an array of items.
The value is optional and defaults to `ARRAY_A`.

### Testing
- Enable WCS in our feature branch `feature/checkout-block-simple-multiple-subscriptions` and build `npm run build`.
-  Visit the editor and insert cart or checkout.

### DIFF
--- a/assets/js/type-defs/cart.ts
+++ b/assets/js/type-defs/cart.ts
@@ -123,6 +123,7 @@ export interface CartItem {
 	variation: Array< CartVariationItem >;
 	prices: CartItemPrices;
 	totals: CartItemTotals;
+	extensions: ExtensionsData;
 }
 
 export interface CartTotalsTaxLineItem {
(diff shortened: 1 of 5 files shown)


assets/js/type-defs/cart.ts
```diff
@@ -160,9 +161,7 @@ export interface CartErrorItem {
 	message: string;
 }
 
-export interface CartExtensionItem {
-	[ key: string ]: unknown;
-}
+export type ExtensionsData = Record< string, unknown >;
```
mikejolley: This type won't work for empty data set (`{}`).

mikejolley:
```suggestion
export type ExtensionsData = Record< string, unknown > | Record< string, never >;
```

mikejolley (approved):

PR #3913, opened 2021-03-03
Add types to hooks

senadir:
This PR adds types to some hooks to learn more TS.

`assets/js/base/hooks/cart/use-store-cart-coupons.ts`
`assets/js/base/hooks/cart/use-store-cart-item-quantity.ts`
`assets/js/base/hooks/cart/use-store-cart.ts`
`assets/js/base/hooks/customer/use-customer-data.ts`
`assets/js/base/hooks/customer/utils.ts`
`assets/js/base/hooks/use-container-queries.ts`
`assets/js/base/hooks/use-previous.ts`
`assets/js/base/hooks/use-shallow-equal.ts`
`assets/js/base/hooks/use-store-add-to-cart.ts`
`assets/js/base/hooks/use-throw-error.ts`
`assets/js/base/hooks/use-trigger-fragment-refresh.ts`
`assets/js/type-defs/hooks.ts`

### Testing instructions
- Smoke test cart and checkout to make sure there is no regression because I renamed some variables.


assets/js/base/hooks/cart/use-store-cart.ts
```diff
@@ -7,9 +7,19 @@ import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
...
+	CartResponseFeeItem,
+	CartResponseBillingAddress,
+	CartResponseShippingAddress,
+} from '@woocommerce/types';
 
-const defaultShippingAddress = {
+declare module '@wordpress/html-entities' {
+	// eslint-disable-next-line no-shadow
+	export function decodeEntities< T >( coupon: T ): T;
+}
```
senadir: We need this because `decodeEntities` is wrong.
What's there
`(html: string): string` but it actually takes everything and works fine with any type so it should be 
`<T>(html: T): T`

assets/js/base/hooks/cart/use-store-cart.ts
```diff
@@ -21,20 +31,49 @@ const defaultShippingAddress = {
...
+const decodeValues = (
+	object: Record< string, unknown >
+): Record< string, unknown > => {
+	return Object.fromEntries(
+		Object.entries( object ).map( ( [ key, value ] ) => [
+			key,
+			decodeEntities( value ),
+		] )
+	);
+};
```
senadir: Moved from lodash to `fromEntries` and `entries` cc @Aljullu is using `fromEntries` still a problem?

assets/js/base/hooks/customer/use-customer-data.ts
```diff
@@ -18,31 +26,39 @@ import { useStoreCart } from '../cart';
...
 		billingAddress: initialBillingAddress,
 		shippingAddress: initialShippingAddress,
-	} = useStoreCart();
+	}: CustomerData = useStoreCart();
 
 	// State of customer data is tracked here from this point, using the initial values from the useStoreCart hook.
-	const [ customerData, setCustomerData ] = useState( {
-		billingData: initialBillingAddress,
+	const [ customerData, setCustomerData ] = useState< CustomerData >( {
+		billingAddress: initialBillingAddress,
```
senadir: In this hook only, I renamed `billingData` to `billingAddress`

assets/js/type-defs/cart-response.ts
```diff
@@ -145,11 +145,11 @@ export interface CartResponseFeeItemTotals extends CurrencyResponseInfo {
 	total_tax: string;
 }
 
-export interface CartResponseFeeItem {
+export type CartResponseFeeItem = {
```
senadir, line 148: Had to change this to type so it works well with Record and `decodeValues`

assets/js/base/hooks/cart/use-store-cart.ts
mikejolley: @senadir yes it is. We have to use a polyfill. See `deriveSelectedShippingRates`

assets/js/base/hooks/cart/use-store-cart.ts
```diff
@@ -68,12 +107,14 @@ export const defaultCartData = {
  *
  * @return {StoreCart} Object containing cart data.
  */
-export const useStoreCart = ( options = { shouldSelect: true } ) => {
+export const useStoreCart = (
+	options: { shouldSelect: boolean } = { shouldSelect: true }
+): StoreCart => {
 	const { isEditor, previewData } = useEditorContext();
 	const previewCart = previewData?.previewCart || {};
```
opr, line 114: Just to note that this line is showing a TS error, it will be fixed when we type `assets/js/base/context/editor/index.js`

assets/js/base/hooks/cart/use-store-cart-item-quantity.ts
```diff
@@ -71,8 +67,11 @@ export const useStoreCartItemQuantity = ( cartItem ) => {
 	const previousIsPendingDelete = usePrevious( isPendingDelete );
 
 	const removeItem = () => {
+		const removeItemPromise = ( removeItemFromCart(
+			cartItemKey
+		) as unknown ) as Promise< void >;
```
opr: What do you think of the way I've done this? It stops the type error on the `.then` but I'm not 100% sure if this is the right way to go.

The actions are generator functions, but I don't understand how to type them so they can be used like Promises when we get them from `useDispatch`.

assets/js/base/hooks/customer/use-customer-data.ts
opr: Makes sense 👍🏼

assets/js/base/hooks/customer/use-customer-data.ts
```diff
@@ -97,7 +113,7 @@ export const useCustomerData = () => {
 		}
 		previousCustomerData.current = debouncedCustomerData;
 		updateCustomerData( {
-			billing_address: debouncedCustomerData.billingData,
+			billing_address: debouncedCustomerData.billingAddress,
 			shipping_address: debouncedCustomerData.shippingAddress,
 		} )
 			.then( () => {
```
opr: This is again the same issue with the action being typed properly, noting it in a comment here so we can review it when I find out how to type these.

assets/js/base/hooks/use-previous.ts
```diff
@@ -3,15 +3,21 @@
...
  * @param {Function} [validation] Function that needs to validate for the value
  *                                to be updated.
  */
-export const usePrevious = ( value, validation ) => {
-	const ref = useRef();
+export function usePrevious< T >(
+	value: T,
+	validation?: Validation< T >
+): T | undefined {
+	const ref = useRef< T >();
```
opr, line 20: Great use of generics 🥳

assets/js/base/hooks/use-shallow-equal.ts
```diff
@@ -16,10 +16,10 @@ import isShallowEqual from '@wordpress/is-shallow-equal';
  * @return {*} The previous cached instance of the value if the current has
  *                 shallow equality with it.
  */
-export const useShallowEqual = ( value ) => {
-	const ref = useRef();
-	if ( ! isShallowEqual( value, ref.current ) ) {
+export function useShallowEqual< T >( value: T ): T | undefined {
+	const ref = useRef< T >();
+	if ( ref.current === undefined || ! isShallowEqual( value, ref.current ) ) {
```
opr, line 21: @senadir I added this because without it, we were possibly passing undefined to `isShallowEqual` which is not allowed according to its type definitions 🤓

assets/js/base/hooks/use-store-add-to-cart.ts
```diff
@@ -51,7 +60,7 @@ export const useStoreAddToCart = ( productId ) => {
 
 	const addToCart = ( quantity = 1 ) => {
 		setAddingToCart( true );
-		addItemToCart( productId, quantity )
+		addItemToCart< Promise< boolean > >( productId, quantity )
```
opr: Noting this for when we figure out how to type actions properly.

assets/js/base/hooks/use-throw-error.ts
```diff
@@ -10,16 +10,12 @@ import { useState, useCallback } from '@wordpress/element';
...
-
-	const throwError = useCallback(
-		( error ) =>
-			setState( () => {
-				throw error;
-			} ),
-		[]
-	);
-
+	const throwError = useCallback( ( error: Error ): void => {
```
opr: Can we just return without using the `throwError` variable? Seems redundant.

opr (commented):
Great work Nadir, this mostly looks fine, but I am really struggling to figure out how we are going to type the actions, since they are generator functions, we can't specify that they return promises (because they don't). I don't have a great amount of experience with `@wordpress/data` and especially not using it with TypeScript. 

You can see [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3913/files#diff-86c68631a2dc83678f26a5eb3701684e1fee8b9238928d28513c459429cb26a0R69-R78) that I've cast `removeItemFromCart` to a Promise, but I am not convinced this is the correct way of doing things.

Asked the question in p1614972508072600-slack-CLCNELZE1

opr:
Added some more commits here, the [mapped-types.ts](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3913/files#diff-7e630e6f8dfcf5354f81558d9ac0828bf697a9c519747ed6430c88747cfab4dd) file is really useful for correctly getting the types of actions that are generator functions. You can see this working by looking at [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/87adc877a6875bcb98634f223895580c2fdacaa0/assets/js/base/hooks/cart/use-store-cart-item-quantity.ts#L71) in your IDE and noticing there are no TS errors, and also that the function has the correct return type ( `Promise< void >` ) in the tooltip.

assets/js/base/hooks/customer/use-customer-data.ts
mikejolley: @senadir we should leave it as billingData if it contains the email and tel. Does it here?

assets/js/base/hooks/cart/use-store-cart.ts
```diff
@@ -7,9 +7,19 @@ import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
...
+	StoreCart,
+	CartResponseTotals,
+	CartResponseFeeItem,
+	CartResponseBillingAddress,
+	CartResponseShippingAddress,
+} from '@woocommerce/types';
 
-const defaultShippingAddress = {
+declare module '@wordpress/html-entities' {
+	// eslint-disable-next-line no-shadow
```
mikejolley: is this still needed now it's configured in eslint config?

assets/js/base/hooks/customer/use-customer-data.ts
mikejolley: I believe we call it billingData in places because it contains email + tel, not just address. I think we should leave naming alone until reviewing this across the client.

assets/js/base/hooks/cart/use-store-cart.ts
senadir: As of when? I can see it being turned off only for mapped file.

assets/js/base/hooks/customer/use-customer-data.ts
senadir: In our typedef and API response, billingAddress always contain email and tel, I can't find any instance in which billingAddress has only the address, yet we have two namings for it.
I can roll this back but just wanted to be consistent with the API and types.

assets/js/base/hooks/cart/use-store-cart.ts
senadir: Fixed with a new `fromEntriesPolyfill` function that's used in both places.

assets/js/base/hooks/customer/use-customer-data.ts
senadir: Updated this with some gymnastics, still not convinced this should be billingData.

mikejolley (approved):
Thanks for fixing those issues, tested and seems ok 👍🏻

PR #3920, opened 2021-03-04
 Hide payment methods that have missing dependencies and display an error in the admin

### LINKED ISSUE #3800: Prevent errant payment methods from keeping Cart and Checkout blocks from loading.
In order for payment methods to work correctly with cart and checkout blocks, there is a delicate asset dependency dance that needs to be done. Payment method integration scripts need to load after `wc-blocks-registry` (which contains the client side API registration), but _before_ the main cart and checkout block scripts. The WordPress script registry system works well with controlling the loading order of things a script is dependent on loading before it loads, but _not_ for controlling when a script must load _before_ another script.

In order to handle the "dance of asset dependencies" correctly for payment methods, the Payment Method Registration API interface for Cart and Checkout blocks has a mechanism that actually adds the script handle provided by the registered payment method as a dependency on the cart and checkout block scripts. This removes the need for payment methods to have to worry about this dance and gives us greater control over managing it (if WP interface changes etc).

However, a flaw with this approach was recently discovered when @jrodger was attempting to test another issue with WooCommerce Payments integration with the blocks. The Cart and Checkout blocks were not available in the block picker to insert into a page. In debugging this, I discovered that the `wc-payment-method-wcpay` handle registered with the api interface as the script handle for the WooPay integration [has `stripe` as a dependency](https://github.com/Automattic/woocommerce-payments/blob/be5c9193730ba603f03b525e9663fe1d80f96db3/includes/class-wc-payments-blocks-payment-method.php#L36-L45) and `stripe` is not registered in this request context. Thus the `wc-payment-method-wcpay` asset would not load which in turn prevented the Cart and Checkout block scripts from loading.

So the problem here, is twofold:

- It's relatively easy for payment methods to cause Cart and Checkout blocks to not load if their registered script has dependencies that aren't registered in the same context.
- This is a hard issue to debug because there's no obvious errors giving clues about what the problem is. So unless the developer has domain knowledge about this dependency dance, this can lead to a lot of wasted time figuring out the issue.

### Suggested Approaches

Without getting too prescriptive, I think the solution here involves:

- somehow checking the dependencies of the handle that is registered by a payment method and ensuring those dependencies are also registered.

When there is a problem detected...
- surface the problematic dependency handle as an error.
- remove the registered payment method handle from being a dependency on the blocks.

Aljullu:
Fixes #3800.

In some occasions, a payment method might be registered but its dependencies might not (read #3800 for a more in-depth explanation). In those cases, this PR makes it so:
* The payment method with missing dependencies is removed from Cart/Checkout script dependencies. This way, Cart and Checkout block continue working and the broken payment method is silently hidden.
* It adds an error notice in `wp-admin` and `error_log()`'s the error. If `WP_DEBUG` is set to true, it also includes a link to our docs.

There are two different set-ups that you can try:
* One consists on recreating the issue we faced with WC Pay. You will need to install WC Pay and remove [these lines](https://github.com/Automattic/woocommerce-payments/blob/trunk/includes/class-wc-payments-blocks-payment-method.php#L37-L43), so `stripe` is not registered.
* Another solution is to add a fake dependency in any other payment method. Ie, in [Cheque](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Payments/Integrations/Cheque.php):
```diff
	public function get_payment_method_script_handles() {
		$this->asset_api->register_script(
			'wc-payment-method-cheque',
			'build/wc-payment-method-cheque.js',
+			[ 'missing-dependency' ]
		);
		return [ 'wc-payment-method-cheque' ];
	}
```

1. Go to any `wp-admin` page and verify an error notice is displayed for each faulty payment method:
imatge
2. Create a post or page and verify the Cart and Checkout blocks are available. If you add them, you will notice the faulty payment methods are not available.
3. In the frontend, verify the Cart and Checkout blocks load as usual (but the payment methods with missing dependencies are not loaded).

### Changelog

> Prevent errant payment methods from keeping Cart and Checkout blocks from loading


src/Payments/Api.php
```diff
@@ -174,4 +175,58 @@ public function process_legacy_payment( PaymentContext $context, PaymentResult &
...
+	 * If not, remove that payment method script from the list of dependencies
+	 * of Cart and Checkout block scripts so it doesn't break the blocks and show
+	 * an error in the admin.
+	 */
+	public function verify_payment_methods_dependencies() {
+		global $wp_scripts;
+		$payment_method_scripts = $this->payment_method_registry->get_all_active_payment_method_script_dependencies();
+
+		foreach ( $payment_method_scripts as $payment_method_script ) {
+			$deps = $wp_scripts->registered[ $payment_method_script ]->deps;
```
mikejolley, line 196: Is this array item guaranteed to exist or do you need a guard?

src/Payments/Api.php
```diff
@@ -174,4 +175,58 @@ public function process_legacy_payment( PaymentContext $context, PaymentResult &
...
 	}
+
+	/**
+	 * Verify all dependencies of registered payment methods have been registered.
+	 * If not, remove that payment method script from the list of dependencies
+	 * of Cart and Checkout block scripts so it doesn't break the blocks and show
+	 * an error in the admin.
+	 */
+	public function verify_payment_methods_dependencies() {
+		global $wp_scripts;
```
mikejolley: I think we can avoid the global by calling `$wp_scripts = wp_scripts();`?

src/Payments/Api.php
```diff
@@ -174,4 +175,58 @@ public function process_legacy_payment( PaymentContext $context, PaymentResult &
...
+
+		foreach ( $payment_method_scripts as $payment_method_script ) {
+			$deps = $wp_scripts->registered[ $payment_method_script ]->deps;
+			foreach ( $deps as $dep ) {
+				if ( ! wp_script_is( $dep, 'registered' ) ) {
+					// Log and show errors.
+					error_log(  // phpcs:ignore
+						sprintf(
+							/* translators: 1: handle of the payment gateway. 2: handle of the dependency that is missing. */
+							esc_html__( 'Payment gateway with handle %1$s has been deactivated because its dependency %2$s is not registered.', 'woo-gutenberg-products-block' ),
```
mikejolley: Do we need to localise debug strings? This is extra burden for translators, and we need to make sense of logs.

src/Payments/Api.php
```diff
@@ -174,4 +175,58 @@ public function process_legacy_payment( PaymentContext $context, PaymentResult &
...
+						'admin_notices',
+						function() use ( $payment_method_script, $dep ) {
+							echo '<div class="error"><p>';
+							printf(
+								/* translators: 1: handle of the payment gateway. 2: handle of the dependency that is missing. */
+								esc_html__( 'Payment gateway with handle %1$s has been deactivated because its dependency %2$s is not registered.', 'woo-gutenberg-products-block' ),
+								esc_html( $payment_method_script ),
+								'<code>' . esc_html( $dep ) . '</code>'
+							);
+							if ( defined( 'WP_DEBUG' ) && WP_DEBUG ) {
```
mikejolley: Why are we only showing docs with DEBUG on?

src/Payments/Api.php
```diff
@@ -174,4 +175,58 @@ public function process_legacy_payment( PaymentContext $context, PaymentResult &
...
+					error_log(  // phpcs:ignore
+						sprintf(
+							/* translators: 1: handle of the payment gateway. 2: handle of the dependency that is missing. */
+							esc_html__( 'Payment gateway with handle %1$s has been deactivated because its dependency %2$s is not registered.', 'woo-gutenberg-products-block' ),
+							esc_html( $payment_method_script ),
+							esc_html( $dep )
+						)
+					);
+					add_action(
+						'admin_notices',
```
mikejolley: Is this the best place to surface errors? I am aware that notices are easy to add but I think it does add a lot of noise within admin.

I think debug log is good, and perhaps we could add an inline script to log to the console too. Those would be visible by developers and it avoids polluting admin notices.

If that works well we should do the same for the circular dependency notice too.

mikejolley (changes requested):
This worked (I tested the 2nd case), but I have some concerns regarding notices. Thoughts on that?

src/Payments/Api.php
Aljullu: Added a guard in 1845b12b57ca5c26a400a8b80413fd655d2e89a0.

src/Payments/Api.php
Aljullu: Yup, much better. Changed in 5d46dbedbaf3f1e5eadee2070656ebb4aed18a40.

src/Payments/Api.php
Aljullu: Makes sense not localizing them. Changed in 23a26315014f8a4415194686889596c15721fbf3.

src/Payments/Api.php
Aljullu: The idea was not to bother merchants with a link that was only useful to payment method developers. But now that we no longer use a notice in the admin, this is not relevant anymore and we can print the URL in all logs. :+1:

src/Payments/Api.php
Aljullu: Good idea! I moved this error from being a notice to a `console.error()` in a9bd620ff1a0eaa5d8af9fd3923ef0657de7f600. For now, I didn't change the circular dependency notice, we can take care of it in a follow-up and maybe even creating a util so we can easily log in the browser console from PHP code.

Aljullu:
Thanks for the review @mikejolley. I applied your feedback, this is ready for another review.

mikejolley (approved):
<3 :shipit:

PR #3922, opened 2021-03-04
Add billing data to payment method registration, convert cart/checkout context provider to typescript

### LINKED ISSUE #2906: Checkout: add billing address to the canMakePayment property of the payment methods
See #2858 

Payment gateways should have access to the billing address alongside with the shipping address (that is already made available) in the `canMakePayment` property to be able to use those details to determine if the payment method is available or not.

Let's see this scenario: a person from Portugal is buying a gift for a friend in the US. The shipping address will be in the US but he will deactivate the "Use same address for billing" option and enter his Portuguese address on the billing address. I should allow him to use a Portuguese payment gateway.

The `usePaymentMethodRegistration` hook on `use-payment-method-registration.js` that makes `cartTotals`, `cartNeedsShipping`, .... available is related to **payment**, so it should also make available the **billing** address, don't you agree?

mikejolley:
This commit: https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/08b483db46875c2447c6661d9da8796cee9e8a78 fixes #2906 by including billing data in the data that is passed down to payment methods.

Fixes #2906

...every other commit is me implementing typescript for the cart-checkout context provided and event emitter. The naming changes from `Subscriber` to `Observer` were suggested by @nerrad for consistently.

1. Smoke test checkout and the payment process.
2. I smoke tested payments via stripe and noticed no regression.

### Dev Note

> For Payment Method integrations with Block Checkout, the `canPayArgument` now includes `billingData`. This consists of the customer billing address, email, and phone number, so you can disable or enable a payment method conditionally based on those properties.

### Changelog

> Provided billing data to payment method extensions so they can decide if payment is possible.


nerrad (commented):
I reviewed 08b483d which adds `billingData` to the payment method interface API and it looks great. Thanks Mike. However, we should probably:

- create a devnote mentioning the interface addition (maybe changelog entry is enough, but I'd like to highlight this in the post.
- update payment method integration docs to include mention of the new prop?

mikejolley:
@nerrad I've added a dev note, but I couldn't find docs that covered the properties of `canMakePayment`. The wiki mentions this, but it does not list the fields it receives.

nerrad:
> @nerrad I've added a dev note, but I couldn't find docs that covered the properties of canMakePayment. The wiki mentions this, but it does not list the fields it receives.

It's in [this documentation](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/payment-method-integration.md#express-payment-methods---registerexpresspaymentmethod-options-) - which is actually missing references to _any_ properties it receives (other than a general "The function will be passed an object containing data about the current order").

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.ts
```diff
@@ -62,6 +64,7 @@ const usePaymentMethodRegistration = (
 	const canPayArgument = useRef( {
 		cartTotals,
 		cartNeedsShipping,
+		billingData,
```
ralucaStan, line 67: Unrelated to changes, but got me thinking why we send billing**Data** and shipping**Address**.  From what I saw [in cart type definitions](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/647ef09a1dbf0861168bc47b3f9827da87250df6/assets/js/type-defs/cart.ts#L67) we should rename this to billingAddress. In [this documentation ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/payment-method-integration.md#express-payment-methods---registerexpresspaymentmethod-options-) I saw both `billingData` and `billingAddress` but I'm not sure what the difference is.

tsconfig.json
```diff
@@ -6,6 +6,8 @@
 		{ "path": "./assets/js/data" },
 		{ "path": "./assets/js/icons" },
 		{ "path": "./assets/js/base/components" },
-		{ "path": "./packages" }
+		{ "path": "./packages" },
+		{ "path": "./assets/js/base/components" },
```
ralucaStan: `{ "path": "./assets/js/base/components" },` appears twice

assets/js/base/context/shared/event-emit/emitters.ts
```diff
@@ -50,21 +49,23 @@ export const emitEvent = async ( observers, eventType, data ) => {
...
 	const observersByType = getObserversByPriority( observers, eventType );
 	for ( const observer of observersByType ) {
 		try {
 			const response = await Promise.resolve( observer.callback( data ) );
-			if ( typeof response !== 'object' ) {
+			if ( typeof response !== 'object' || response === null ) {
 				continue;
 			}
-			if ( typeof response.type === 'undefined' ) {
+			if ( ! response.hasOwnProperty( 'type' ) ) {
```
ralucaStan, line 68: `if (!response?.type) {}` could also work here

assets/js/base/context/cart-checkout/payment-methods/types.ts
```diff
@@ -0,0 +1,152 @@
...
+export interface PaymentMethodConfig {
+	// A unique string to identify the payment method client side.
+	name: string;
+	// A react node for your payment method UI.
+	content: ReactNode;
+	// A react node to display a preview of your payment method in the editor.
+	edit: ReactNode;
+	// A callback to determine whether the payment method should be shown in the checkout.
+	canMakePayment: (
+		cartData: ObjectType
```
ralucaStan, line 24: Not necessary now, but I think cartData can be defined in more detail if we look at what we send
```
{
			cartTotals,
			cartNeedsShipping,
			billingData,
			shippingAddress,
			selectedShippingMethods,
			paymentRequirements,
		}
```

assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.tsx
```diff
@@ -237,53 +219,39 @@ export const PaymentMethodDataProvider = ( { children } ) => {
...
-			 *                                             payment method.
-			 */
-			failed: ( errorMessage, paymentMethodData, billingData = null ) => {
-				if ( billingData ) {
+			failed: (
+				errorMessage,
+				paymentMethodData,
+				billingData = undefined
+			) => {
+				if ( typeof billingData !== undefined ) {
```
ralucaStan: I think `if ( billingData ) {}` worked well as well.

ralucaStan (commented):
👍  really useful work in this PR. I added some comments as well.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.ts
mikejolley: Billing has extra data (email and telephone number) in addition to the address.

ralucaStan: Should we change [ this]( https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/647ef09a1dbf0861168bc47b3f9827da87250df6/assets/js/type-defs/cart.ts#L67)  to CartBillingData then? @nerrad what do you think?

nerrad: I may be missing something Raluca, but what's the rationale for changing it from `billingData` to `CartBillingData` here?

ralucaStan: I just noticed we are using different names for the same data structure. My proposal is to leave `billingData` as is here (it has billingAddress + email + phone). I was thinking that the type definition should be updated for consistency reasons from `CartBillingAddress` to `CartBillingData`.

tsconfig.json
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3922/commits/280abf6985fc94e7f3b954129fe0f2208a0bf21f

assets/js/base/context/shared/event-emit/emitters.ts
mikejolley: Typescript complains that type does not exist on response with your suggestion. `hasOwnProperty` was the only way I could suppress that warning.

assets/js/base/context/cart-checkout/payment-methods/use-payment-method-registration.ts
ralucaStan: ^ change is not relevant for this PR, it was a suggestion for normalizing data naming

assets/js/base/context/cart-checkout/payment-methods/types.ts
mikejolley: Yeah. I won't make that change now because of scope, but you're right.

mikejolley:
Address that feedback @ralucaStan - I've left the optional items for follow ups so this can get merged before it becomes stale :)

ralucaStan (approved):

assets/js/base/context/shared/event-emit/emitters.ts
ralucaStan: I had a hunch that Ts might be behind this one

PR #3925, opened 2021-03-05
Update/typescript checkout shipping 

ralucaStan:
PR adding types to /packages/checkout/shipping directory. 

Checkout branch and make sure code is free of TS warnings errors.
 Testing is limited to smoke test the Cart and Checkout blocks to verify there are no regressions.

Just a note that this PR already typed FormattedMonetaryAmount component. So I just kept my file and ignored the change brought to the file here https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/aa82c92d5883c09d80aed60291d13661164e3437#diff-d6ebd1dae536334d33ae51da0820ecd0896c24c57a34f5ddaebd0f8b1c98c12cL36


assets/js/base/components/radio-control/option-layout.tsx
```diff
@@ -28,15 +34,15 @@ const OptionLayout = ( {
...
+						id={ id && `${ id }__description` }
 						className="wc-block-components-radio-control__description"
 					>
 						{ description }
 					</span>
 				) }
 				{ secondaryDescription && (
 					<span
-						id={ id ? `${ id }__secondary-description` : null }
+						id={ id && `${ id }__secondary-description` }
```
mikejolley, line 45: 👍🏻 good improvement, not sure why null was there originally.

assets/js/base/components/formatted-monetary-amount/index.tsx
```diff
@@ -64,7 +74,7 @@ const FormattedMonetaryAmount = ( {
 		className
 	);
 	const numberFormatProps = {
-		displayType,
+		displayType: 'text' as NumberFormatProps[ 'displayType' ],
```
mikejolley: I think we should move this one to our `FormattedMonetaryAmount` props since we do use this in other places and we always set the default value to 'text'. The type can then go within `FormattedMonetaryAmountProps`.

assets/js/base/components/formatted-monetary-amount/index.tsx
```diff
@@ -27,27 +32,32 @@ const currencyToNumberFormat = ( currency ) => {
...
+}
+
 /**
  * Formatted price component.
  *
  * Takes a price and returns a formatted price using the NumberFormat component.
  *
  * @param {Object} props Component props.
- * @param {string=} props.className CSS class used.
+ * @param {string} props.className CSS class used.
```
mikejolley: We can remove this docblock and inline the comments now (within `FormattedMonetaryAmountProps`)

packages/checkout/shipping/shipping-rates-control/render-package-rate-option.tsx
```diff
@@ -3,33 +3,35 @@
  */
 import { decodeEntities } from '@wordpress/html-entities';
 import { DISPLAY_CART_PRICES_INCLUDING_TAX } from '@woocommerce/block-settings';
-import FormattedMonetaryAmount from '@woocommerce/base-components/formatted-monetary-amount';
 import { getCurrencyFromPriceResponse } from '@woocommerce/price-format';
+import FormattedMonetaryAmount from '@woocommerce/base-components/formatted-monetary-amount';
```
mikejolley, line 7: This one gives me an error. `Casing of @woocommerce/base-components/formatted-monetary-amount does not match the underlying filesystem.`

mikejolley (changes requested):
I tested cart and checkout block (particularly shipping) and noticed no regressions 👍🏻 

I have added a few comments, and some issues I spotted in TS console.

Other than that I also have 2 questions which may not require action, but we should add to our hack.md file as notes.

1. Should we rename all `.js` to `.ts` in packages we convert? I noticed in `shipping-rates-control` the index file was left as a JS file.
2. When do we use type-defs vs a types.ts file within the package/component directory? I think we may all be handling this is different ways. Is it just based on if they are shared? Personally I prefer types.ts so it's in the same context as the component using it.

ralucaStan:
> 1. Should we rename all `.js` to `.ts` in packages we convert? I noticed in `shipping-rates-control` the index file was left as a JS file.

I think that's the way to go, this also removes uncertainty whether a file was converted or not.  I can add a note in the knowledge file

> 2. When do we use type-defs vs a types.ts file within the package/component directory? I think we may all be handling this is different ways. Is it just based on if they are shared? Personally I prefer types.ts so it's in the same context as the component using it.

I can see how a file could use a shared type and something specific. I also prefer local types.ts files for specific types, but this is a good topic to bring to the team.

packages/checkout/shipping/shipping-rates-control/render-package-rate-option.tsx
ralucaStan: Does this error appear only in your editor and not in the CI? If you don't have other casing on this path `assets/js/base/components/formatted-monetary-amount/index.tsx` it might be a false positive. Found a [similar issue](https://github.com/benmosher/eslint-plugin-import/issues/720). Could you maybe try to update to eslint-plugin-import@v2.8.0 and see if it still happens.

ralucaStan:
@mikejolley changes pushed

assets/js/base/components/formatted-monetary-amount/index.tsx
```diff
@@ -1,21 +1,33 @@
...
 /**
  * Internal dependencies
  */
 import './style.scss';
 
+interface FormattedMonetaryAmountProps {
+	className?: string;
+	displayType?: NumberFormatProps[ 'displayType' ];
+	value: number; // Value of money amount.
+	currency: Currency | Record< string, never >; //Currency configuration object.
```
mikejolley:
```suggestion
	currency: Currency | Record< string, never >; // Currency configuration object.
```

assets/js/base/components/formatted-monetary-amount/index.tsx
```diff
@@ -1,21 +1,33 @@
...
+	onValueChange?: ( unit: number ) => void; //Function to call when value changes.
```
mikejolley:
```suggestion
	onValueChange?: ( unit: number ) => void; // Function to call when value changes.
```

mikejolley (approved):

PR #3932, opened 2021-03-08
Only return a single handle for block styles

mikejolley:
For background, see https://github.com/WordPress/gutenberg/pull/29473

Gutenberg only expects a single style handle, not an array of handles. We already have vendor styles set up as a dependency of block styles, so enqueueing multiple handles is not needed. Behaviour does not change with this PR.

### Changelog

> Fix block style handle definition.


ralucaStan (approved):
👍

senadir:
I feel like we can skip changelog for this PR.

PR #3937, opened 2021-03-08
Show compatibility notice when adding the Cart and Checkouts blocks in the editor

### LINKED ISSUE #3017: Implement a compatibility notice when a user views or inserts Cart or Checkout for the first time.

Aljullu:
Fixes: #3017.
Supersedes: #3018.

Related PR in Gutenberg that intends to remove the page control from the `Guide` when there is only one page: https://github.com/WordPress/gutenberg/pull/29629.

Some notes about this PR:
* The notice is disabled in WP 5.4, in that version of WordPress the `Guide` component we use had a different design, which would require us to have the image in a different size, and also the props were different. Alternatively, we could load it from bundled `wordpress-components`, but that would increase the assets size. Given that we will be dropping support for WP 5.4 in the next weeks, I considered it was not worth it.
* The notice is shown per-user, so each user will see the notice the first time they add the Cart or Checkout blocks or edit a page that contains them. The idea is to behave similarly to the _Welcome guide_ that appears the first time a user opens the block editor.
* The notice needs to be dismissed for each block. Ie: if the user adds the Cart block and dismisses the notice, they will still see a notice when adding the Checkout block.
* My initial idea was to create a store that saves the values into the local storage, similar to `core/edit-post` in Gutenberg. But it seemed a bit overwhelming considering the goal is to save one single value. Instead, I went ahead with a simpler implementation using `useLocalStorageState()`.

### Screenshots
Cart and Checkout compatibility notice

1. Create a post or page and add the Cart and Checkout blocks.
2. Verify a compatibility notice appears. Press the <kbd>Tab</kbd> key to verify the focus is inside the modal.
3. Verify the compatibility notice can be dismissed and the block is focused.
4. Verify reloading the page or adding the same block in another page or post doesn't show the notice again.


assets/js/editor-components/compatibility-notice/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * Internal dependencies
+ */
+import WooImage from './woo-image';
+
+interface CompatibilityNoticeProps {
+	blockName: string;
+}
+
+export default function CompatibilityNotice( {
```
ralucaStan: I wonder if we should rename to CartCheckoutCompatibilityNotice as it is specific to these two. I think that if we need to add to other blocks in the future, we should change the logic to store only one key in localStorage `wc-blocks_compatibility_notice: ["cart","checkout","future_block"]` instead of keeping individual keys. But for now this works as expected.

assets/js/base/hooks/use-local-storage-state.ts
```diff
@@ -0,0 +1,21 @@
...
+		return valueInLocalStorage
+			? JSON.parse( valueInLocalStorage )
+			: initialValue;
+	} );
+	useEffect( () => {
+		window.localStorage.setItem( key, JSON.stringify( state ) );
+	}, [ key, state ] );
+
+	return [ state, setState ];
+};
```
ralucaStan, line 36: Should we add some try catch statements here to be safe with the JSON methods and the local storage operations?

ralucaStan (commented):
It works as expected. Please see my comments and let me know what you think. One thing I'd change is adding the try/catch as commented. Nice work 👍

assets/js/editor-components/compatibility-notice/index.tsx
Aljullu: Great suggestions! In a1556e6ee7177078c60c3bbc6b0d13b6ef4c945c I renamed the component and in 06f2054afebf033e670bef9fc136a1e519876e6b I refactored it so it uses an array like the one you suggested.

assets/js/base/hooks/use-local-storage-state.ts
Aljullu: Good point, done in 7d810f0c49bdc7d4850dd9307e29bff572b44691.

Aljullu:
Thanks for the review @ralucaStan! I applied the feedback and this is ready for another look.

Aljullu:
Rebased and updated the tests. All green now. :green_circle:

assets/js/editor-components/compatibility-notices/use-compatibility-notice.ts
```diff
@@ -0,0 +1,32 @@
...
+): [ boolean, () => void ] => {
+	const [ dismissedNotices, setDismissedNotices ] = useLocalStorageState(
+		`wc-blocks_dismissed_compatibility_notices`,
+		initialDismissedNotices
+	);
+	const [ isVisible, setIsVisible ] = useState( false );
+
+	const isDismissed = dismissedNotices.includes( blockName );
+	const dismissNotice = () => {
+		const dismissedNoticesSet = new Set( dismissedNotices );
```
ralucaStan, line 20: Sets are not supported in ie11. Not sure if we have to support it or our configuration of @babel/preset-env supports this, but I'd double-check it in the browser. I saw it's used once in the project outside of notices.

ralucaStan (approved):
Nice work @Aljullu ! I've added one comment regarding Sets but outside of that is good to go.

assets/js/editor-components/compatibility-notices/use-compatibility-notice.ts
Aljullu: Good catch. I think Babel takes care of transpilling this, we have a couple of other instances of Set in the codebase. I just tested with BrowserStack to double-check and everything seemed to work fine with IE11. :ok_hand:

PR #3938, opened 2021-03-08
Fix Storybook after TypeScript changes

Aljullu:
`npm run storybook` was failing with an error:

```
ERR! TypeError: Cannot read property 'paths' of undefined
ERR!     at Object.<anonymous> (/wp-content/plugins/woocommerce-gutenberg-products-block/storybook/webpack.config.js:17:55)
```

(I think the issue was introduced in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3768 cc @nerrad)

This PR fixes it:
* Storybook's webpack config now points to `tsconfig.base.json`, where `compilerOptions` where moved.
* I needed to remove the comments from the JSON file in order not to throw when reading it.
* I also needed to add the `@woocommerce/price-format` alias.

1. Run `npm run storybook` and verify it generates the storybook and you can view it in the browser.

### DIFF
--- a/storybook/webpack.config.js
+++ b/storybook/webpack.config.js
@@ -12,7 +12,7 @@ const {
 	getMainConfig,
 	getStylingConfig,
 } = require( '../bin/webpack-configs.js' );
-const tsConfig = require( '../tsconfig.json' );
+const tsConfig = require( '../tsconfig.base.json' );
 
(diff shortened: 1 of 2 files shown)


ralucaStan (approved):
Works as expected

PR #3945, opened 2021-03-10
Add docs on how to run e2e tests with a specific environment locally

Aljullu:
It took me some time to discover how to get e2e tests running locally to simulate the tasks which are run automatically in each PR. Ie: in #3937 tests were passing with WP 5.5. and 5.6, but were failing with WP 5.4 and 5.6+GB, so I needed to reproduce those environments locally. This PR adds docs on how to do that.


opr (approved):
Thanks for adding this Albert!

PR #3952, opened 2021-03-11
Remove extra padding from payment methods with no description

Aljullu:
This PR:
* Fixes an issue that was displaying too much bottom padding in the selected payment method if it had no description (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3952/commits/8a5816ff846a56c5cd056cf95d31ac014a463b7b).
* Fixes a typo in a code snippet in payment method integration docs (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3952/commits/2e4b719f3a49330b5c58615d50b39eb598e8ed4a).

### Screenshots

_Before:_
Screenshot before
(notice the extra padding that appears when selecting check payments)

_After:_
Screenshot before

1. Go to WooCommerce > Settings > Payments, edit one of the payment methods and remove any text in its description field.
2. Go to the Checkout block and select the payment method you modified in step 1.
3. Verify there isn't extra padding at the bottom.

### DIFF
--- a/assets/js/blocks/cart-checkout/payment-methods/style.scss
+++ b/assets/js/blocks/cart-checkout/payment-methods/style.scss
@@ -235,6 +235,10 @@
 
 .wc-block-components-radio-control-accordion-content {
 	padding: 0 $gap em($gap) $gap;
+
+	&:empty {
+		display: none;
+	}
 }
 
 .wc-block-checkout__order-notes {
--- a/assets/js/payment-method-extensions/payment-methods/bacs/index.js
+++ b/assets/js/payment-method-extensions/payment-methods/bacs/index.js
@@ -26,7 +26,7 @@ const label = decodeEntities( settings.title ) || defaultLabel;
  * Content component
  */
 const Content = () => {
-	return <div>{ decodeEntities( settings.description || '' ) }</div>;
+	return decodeEntities( settings.description || '' );
 };
 
 /**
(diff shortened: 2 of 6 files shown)


ralucaStan (changes requested):
This does not work for me, because the div is not empty, it has an empty div inside. I suggested a fix, let me know if that works for you

assets/js/blocks/cart-checkout/payment-methods/style.scss
```diff
@@ -235,6 +235,10 @@
 
 .wc-block-components-radio-control-accordion-content {
 	padding: 0 $gap em($gap) $gap;
+
+	&:empty {
```
ralucaStan, line 239: this does not work for me because the div has an empty div inside and it's not empty.
But if I move the styles to the child div it works as expected.

```
.wc-block-components-radio-control-accordion-content {
	div {
		padding: 0 $gap em($gap) $gap;

		&:empty {
			display: none;
		}
	}
}
```

ralucaStan: let me know if it works for you

ralucaStan (changes requested):
Please see my comment

Aljullu:
Aaah, yeah, good catch! That happens if you remove the description from the UI, I updated the PR testing steps accordingly.

Your solution would work, but I'm wondering if we should instead remove those `<div>` from the payment methods altogether. They don't seem to be used for anything. I made the change in f52425210aa78601fa6a172dd0965ba8d09dcf9a, let me know how it looks.

ralucaStan (approved):
Works as expected. Removing that extra div was the best solution. 👍

Issue #3935, opened 2021-03-08
GA: Emit Events when viewing product blocks


mikejolley:
Blocked by #3933

Some blocks need to fire a "view" event. This is for Google Analytics.

- Product (list)
- Single product
- Clicking a link in a product block

PR #3953, opened 2021-03-11
Google Analytics Integration - Action Hooks and initial tracking events

### LINKED ISSUE #3934: GA: Cart Events
Blocked by #3933 

Some actions need to emit native JavaScript events for 3rd parties to pick up. These include:

- [x] Add to cart
- [x] Remove item from cart
- [x] Change cart item quantity

mikejolley:
This is my first pass on the Google Analytics Integration within blocks. This requires the [Google Analytics Integration Extension](https://github.com/woocommerce/woocommerce-google-analytics-integration) to test functionality, setup with a G-X type ID for gtag support, and to test events are firing you can use the [Google Analytics Debugger.](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en) I am also logging events to the console to aid in development, but these will be removed before shipping.

The PHP side code includes the gtag script. There is a check to see if this script is already enqueued, but it will never be unless [this issue is resolved in the extension](https://github.com/woocommerce/woocommerce-google-analytics-integration/issues/186). For now we work around this.

Events from blocks are triggered using the [`@wordpress/hooks`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-hooks/) package. Event names have the prefix `experimental__` (underscore prefix is not allowed).

A separate script with it's own entry point is included for the actual GA events. This hooks in using `addAction` and currently covers the following event:

- All Products Block View triggers a `view_item_list` event
- All Products Block add to cart button triggers an `add_to_cart` event
- Cart block remove item triggers a `remove_cart_item` event
- Cart block change qty triggers a custom `change_cart_quantity` event

I chose to trigger events at component level rather than hook level to offer more control over event names and how tracking is done without impacting other potential consumers of the hooks.

Closes #3934 
Part of #3935 

Seeking feedback before moving on to other events.

- Install the [Google Analytics Integration Extension](https://github.com/woocommerce/woocommerce-google-analytics-integration)
- Setup the extension with a G-X type ID from your GA account 
- Install [Google Analytics Debugger](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en) extension and enable it
- Go to the store and trigger some events using the All Products Block and Cart block. See list of events above.
- The console will show when events are fired, and the Google Analytics Debugger will also show logs


senadir:
It seems we have a slight blocker before we ship this, this PRs adds 400kb of gzipped JS to all product block, I'm just marking it here but will take a look once we finalize on the solution because it could be removed while we polish.

assets/js/base/components/product-list/product-list.js
```diff
@@ -137,15 +140,23 @@ const ProductList = ( {
...
-	// reset pagination to the first page.
+	// If the product list changes, trigger an event.
+	useEffect( () => {
+		if ( products.length > 0 ) {
+			dispatchStoreEvent( 'list-products', {
+				products,
+				listName: listType,
+			} );
+		}
+	}, [ products, listType, dispatchStoreEvent ] );
```
senadir: What's listType? I can see it's `All Products` but I don't get why.

assets/js/base/hooks/cart/use-store-cart-item-quantity.ts
```diff
@@ -31,56 +30,45 @@ export const useStoreCartItemQuantity = (
...
-				return false;
-			}
-
-			const store = select( storeKey );
-			return store.isItemPendingDelete( cartItemKey );
-		},
-		[ cartItemKey ]
-	);
-	const previousIsPendingDelete = usePrevious( isPendingDelete );
+	const previousIsPending = usePrevious( isPending );
```
senadir, line 58: Nice abstraction!

assets/js/base/hooks/cart/use-store-cart-item-quantity.ts
```diff
@@ -93,34 +81,37 @@ export const useStoreCartItemQuantity = (
 	] );
 
 	useEffect( () => {
```
senadir, line 83: Why a single Effect instead of two? I see they're completely decoupled and you're not reusing any logic between them.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
```diff
@@ -243,12 +247,24 @@ const CartLineItemRow = ( { lineItem = {} } ) => {
...
 					<button
 						className="wc-block-cart-item__remove-link"
-						onClick={ removeItem }
+						onClick={ () => {
+							removeItem();
+							dispatchStoreEvent( 'remove-cart-item', {
+								product: lineItem,
+								quantity,
+							} );
+						} }
```
senadir, line 267: This doesn't seem to be debounced so if I click it 10 times in a row, it will trigger the event 10 times.

assets/js/base/hooks/use-store-events.ts
```diff
@@ -0,0 +1,36 @@
...
+		eventName: string,
+		eventParams: Partial< Record< string, unknown > >
+	) => void;
+} => {
+	const dispatchStoreEvent = useCallback(
+		(
+			eventName: string,
+			eventParams: Partial< Record< string, unknown > >
+		) => {
+			doAction( `${ namespace }-${ eventName }`, eventParams );
```
senadir: I think we need to wrap this somehow in a try-catch error boundary, or have some control over the actions triggered here.
Throwing a code in `assets/js/extensions/google-analytics/index.ts` for example would break the whole block.

senadir (changes requested):
This PR is good as far as I see, I don't have any comments on the PHP code because I didn't wrap my head around it yet, I don't know why we're doing async registration.

I also think we need to protect our event emitter system from bad 3PD code that can leak all the way to our block and break it.

The other thing is to investigate the sudden surge in size for All Products that I don't really understand.

assets/js/base/components/product-list/product-list.js
mikejolley: The product list components can be shared (they are base components). I wanted a way to say "this list is from the all products block", so I'm passing the block name.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.js
mikejolley: Correct. I decided to track button clicks instead of successful calls to the API. This tracks user behaviour instead of client behaviour.

assets/js/base/hooks/cart/use-store-cart-item-quantity.ts
senadir: Having them in a single effect or two effects wouldn't change much, the effect would run more often but it's light enough now to no care about splitting it, so I was just wondering.

assets/js/base/hooks/use-store-events.ts
mikejolley: Whilst that is true, is our intention to **always** wrap action and filter hooks in try catch? If so, let's abstract it further in a follow up, noting our code is the only consumer of these experimental hooks at present so there won't be breakage we don't know about.

assets/js/base/hooks/cart/use-store-cart-item-quantity.ts
mikejolley: I did them together because the dependencies of the effect are the same.

mikejolley:
@senadir fixed the size issue. It was because I imported the block index file. Found another way of identifying the list using the parentName from inner block context.

assets/js/base/components/product-list/product-list.js
mikejolley: I've swapped this for Inner Block context name.

assets/js/base/hooks/use-store-events.ts
senadir: isn't the goal to have those hooks public eventually? I don't mind handling this in a follow up as long as we handle it before releasing them, even in experimental. While wrapping SlotFills and CheckoutFilters we ended up changing the implementation to accommodate for try-catch (like limiting filters to strings only).

mikejolley: I've added a try catch around the action call so it's isolated.

senadir (approved):
Tested the new changes and they work great, I will add more granular control to the hooks callbacks later so each 3PD runs in its own try-catch boundary instead of sharing it with the rest of callbacks.

PR #3958, opened 2021-03-12
Allow FormattedMonetaryAmount to deal with string based numbers

### LINKED ISSUE #3954: Product Price block is not rendering (in All Products block inner content)
I noticed this when I did a fresh `npm run build` (production build) of current trunk to reproduce another issue I'm currently debugging. My existing All Products block implementation was not rendering the contents for the Product Price inner block. Some brief details (I didn't do any extensive debugging):

- I'm running WordPress 5.7 in my environment (GB plugin is not active)
- I'm running the production build of Woo Blocks (via `npm run build`) - I've also ensured `composer install` was run.
- WooCommerce 5.0 branch from the repo is checked out and built.
- I've verified that the Woo Block files are being loaded.
- The product price element block is also available in the block picker, but when I choose it, there is no price output for the chosen product.

I'm not sure if this also impacts other element blocks, but those should be verified (especially any that are exposed for use in the All Products block).

mikejolley:
Some of the typescript changes touched `FormattedMonetaryAmount` where we asserted value would be numeric. Prices for products are numeric, but are strings. This causes the price component to render as null.

To fix this, I've added string to number conversion in `FormattedMonetaryAmount` and updated the incoming types.

Fixes #3954

1. View all Products Block
2. Confirm prices are displayed
3. Confirm prices still display in cart/checkout to verify no regressions

### DIFF
--- a/assets/js/base/components/formatted-monetary-amount/index.tsx
+++ b/assets/js/base/components/formatted-monetary-amount/index.tsx
@@ -17,7 +17,7 @@ import './style.scss';
 interface FormattedMonetaryAmountProps {
 	className?: string;
 	displayType?: NumberFormatProps[ 'displayType' ];
-	value: number; // Value of money amount.
+	value: number | string; // Value of money amount.
 	currency: Currency | Record< string, never >; // Currency configuration object.
 	onValueChange?: ( unit: number ) => void; // Function to call when value changes.
 }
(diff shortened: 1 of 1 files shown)


ralucaStan (approved):
Nice catch 👍

ralucaStan:
I recommend adding a comment in ProductPrice that prices are string so we don't lose track of this info when we type the file

senadir (approved):
I had the same issue in TotalsItem not rendering the price when it's a string.

PR #3960, opened 2021-03-12
Ensure cart changes remain after using back button to get back to the cart

### LINKED ISSUE #3827: Mobile: If products are added on the cart page, navigating back from checkout to cart does not show the newly added products
## Describe the bug

On mobile, if you add additional products to the cart while on the cart page (e.g. from a cross-sell section on the page), if later you navigate back from the checkout page to the cart, you only see the original products, not the ones added while on the page. 
The products are not actually removed from the cart - If you go back to checkout again you will see them there - but they're not displayed in the cart unless you refresh the page.
On mobile only. Works correctly on desktop.
Reproduced on both mobile Safari and Firefox.

## To reproduce

1. Add a product to the cart and go to to the cart page
2. Add another product from product carousels on the cart page.
3. Click 'proceed to checkout'
4. Use browser back button or a back gesture to navigate back to the cart page
5. Observe that only the original product is listed, the new product added on the cart page isn't shown.
6. If you refresh the cart page, you will see the correct contents displayed.

## Expected behavior
Navigating back to the cart page should show the correct contents of the cart.

## Screenshots
https://user-images.githubusercontent.com/71322363/107629594-424b0000-6c5a-11eb-859d-9f79757c32f8.MP4

Steps followed in the attached video:
STEP 1: Added item to cart and navigated to cart page
STEP 2: Added item to from products on cart page
STEP 3: Proceeded to checkout
STEP 4: Used gesture to go back to cart (product missing)
STEP 5: Returned to checkout page (correct cart - both products included)
STEP 6: Used browser back button to go back to cart (product missing)
STEP 7: Refreshed cart page (correct cart - both products included)

* Core version:  5.5.3
* WooCommerce version: 4.9.2
* WooCommerce Blocks version: 4.4.0. Also tested on 4.3.0 with the same results.
* Site language: English
* Any other plugins installed: Yes (multiple)

* OS: -
* Browser: -
* Version: -

* Device:  iPhone 6
* OS: 12.4.9
* Browser  firefox, safari
* Version: firefox 31.0 (3537), safari - ?

opr:
### Description
If the customer alters their cart, i.e. changes the quantity or adds an item, then navigates away from the page, for example by clicking the "Proceed to checkout" button, and then returns to the cart via the back button in their browser, then their changes will not be shown. The client and the server are out of sync.

This happens because we print the cart data onto the page at page load, and the cart hydrates itself based on this. When the user returns to the page using the back button, the data that were printed onto the page are stale, so the cart hydrates itself with the data that were present when the page was first loaded.

#### Source Code
- Added a new action to `assets/js/data/cart/actions` called: `cartDataIsStale`. This will set a flag in the store indicating the cart data is out of sync with the server. This is based on the check detailed below.
- Add `cartUpdateMiddleware` to `assets/js/middleware/cart-update.ts` which will write the timestamp of the last cart update to the browser's local storage.
- Update the `CartSchema` to add a new field: `generated_timestamp` this will allow us to know when the data were generated so we can compare it to the last time the cart was updated (which is stored in local storage).
- Add a check to the `withStoreCartApiHydration` hook that will compare the `generated_timestamp` time to the one in local storage to determine if the cart needs to be fetched from the API. 
- The cart data will not be rehydrated from `wcSettings` if it is determined that the data is stale.

#### Test code
This PR also introduces some new e2e tests to test this functionality. I needed to create some utility functions to enable testing the front end. The details of this follow.

- Install `@woocommerce/e2e-utils` to the project as a dev dependency. This is so we can use the `shopper.addToCart` flow.
- Add `visitPostOfType` function. This will cause the e2e browser to visit a page in the WordPress dashboard by name and type. This is needed so we can navigate to a `product` page by name. It could also be useful for other tests in future.
- Add `getBlockPagePermalink` function - this will cause the e2e browser to return the permalink of the page it is on if it the page is using the block editor. The analogue to this function is `getNormalPagePermalink` which will do the same but on a page that doesn't use the block editor (for example a product page in WooCommerce).
- Add tests for the cart on the front-end.

Fixes #3827

1. Add items to your cart then go to the cart block.
2. Change the quantity of an item, you may try changing multiple items, with various different quantities. **Make a note of the values!**
3. Press the "Proceed to checkout" button.
4. Use the browser's back button to get back to your cart.
5. Ensure the quantities displayed in the cart match what you expect them to from step 2.
6. Run the e2e tests and ensure all suites still pass.

### Changelog

> Ensure the cart stays up to date if the user navigates using the back/forward buttons in the browser.


assets/js/data/cart/action-types.ts
```diff
@@ -4,6 +4,7 @@ export const ACTION_TYPES = {
 	REPLACE_ERRORS: 'REPLACE_ERRORS',
 	APPLYING_COUPON: 'APPLYING_COUPON',
 	REMOVING_COUPON: 'REMOVING_COUPON',
+	CART_DATA_STALE: 'CART_DATA_STALE',
```
ralucaStan: I see that predominantly the actions start with a verb, so maybe this could be renamed to SET_CART_DATA_STALE_STATE.

assets/js/data/cart/reducers.ts
```diff
@@ -149,6 +149,16 @@ const reducer: Reducer< CartState > = (
 					updatingSelectedRate: !! action.isResolving,
 				},
 			};
+			break;
+		case types.CART_DATA_STALE:
+			state = {
+				...state,
+				metaData: {
+					...state.metaData,
+					cartDataIsStale: !! action.isCartDataStale,
```
ralucaStan: ``cartDataIsStale: action.isCartDataStale,`` works as well as isCartDataStale will always be Boolean

assets/js/data/cart/selectors.ts
```diff
@@ -59,6 +59,16 @@ export const isApplyingCoupon = ( state: CartState ): boolean => {
 	return !! state.metaData.applyingCoupon;
 };
 
+/**
+ * Returns true if cart is stale, false if it is not, and undefined if the check is not complete.
```
ralucaStan: This will never return undefined so I guess the comment is outdated.
There is an issue tho' with not having a default value for `cartDataIsStale` on metaData as this `state.metaData.cartDataIsStale` will throw an error. 

I also see a TS warning 
Screenshot 2021-03-23 at 16 11 24

assets/js/hocs/with-store-cart-api-hydration.js
```diff
@@ -13,13 +13,50 @@ import { useSelect } from '@wordpress/data';
...
+		) {
+			const lastCartUpdateRaw = window.localStorage.getItem(
+				'lastCartUpdate'
+			);
+			const lastCartUpdate =
+				lastCartUpdateRaw === null
+					? null
+					: JSON.parse( lastCartUpdateRaw );
+
+			if ( lastCartUpdate?.timestamp === undefined ) {
```
ralucaStan: If you have only one value you care about here you could directly store the timestamp and avoid working with objects. 
`lastCartUpdateTimestamp: X`

I also recommend to rename the key to `wc-blocks_cart_update_timestamp` to follow the naming pattern used for `wc-blocks_dismissed_compatibility_notices` that easily identifies the items coming from blocks.

assets/js/hocs/with-store-cart-api-hydration.js
```diff
@@ -13,13 +13,50 @@ import { useSelect } from '@wordpress/data';
...
+				cartDataIsStale( false );
+				return;
+			}
+
+			const needsUpdateFromAPI =
+				lastCartUpdate.timestamp >
+				cartData.current?.generated_timestamp;
+
+			if ( needsUpdateFromAPI ) {
+				cartDataIsStale();
```
ralucaStan: Seeing this reinforces my comment to rename the action and call here something like `setCartDataStaleState(true)` as I think it brings more clarity to the code.

assets/js/hocs/with-store-cart-api-hydration.js
```diff
@@ -13,13 +13,50 @@ import { useSelect } from '@wordpress/data';
...
+				return;
+			}
+			cartDataIsStale( false );
```
ralucaStan: I guess this line is needed because there is no default value set under metadata for cartDataIsStale. Would it be an issue to start with cartData.metaData.cartDataIsStale: false?

assets/js/middleware/cart-update.ts
```diff
@@ -0,0 +1,42 @@
...
+			'lastCartUpdate',
+			JSON.stringify( {
+				timestamp: Date.now() / 1000,
+			} )
+		);
+	}
+	return next( options, next );
+};
+
+apiFetch.use( cartUpdateMiddleware );
```
ralucaStan, line 45: 👍

ralucaStan (commented):
Thank you so much for the overview of the changes, they were really helpful for testing the PR. 👍 for the extensive work on testing.

I added some comments, let me know what you think.

assets/js/data/cart/action-types.ts
opr: I disagree with adding `_STATE` to the end, and I decided that `SET_IS_CART_DATA_STALE` reads better. Do you agree?

assets/js/data/cart/selectors.ts
opr: This should be fixed now!

assets/js/hocs/with-store-cart-api-hydration.js
opr: Thanks, I based it on the local storage key used to track the store API nonce update which is called `storeApiNonce` - do you think it's worth adding a cooldown issue to rename this to `wc-blocks_store_api_nonce`?

assets/js/hocs/with-store-cart-api-hydration.js
opr: I've tidied the logic up here, please re-review and see if it makes more sense!

assets/js/data/cart/action-types.ts
ralucaStan: 👍

assets/js/hocs/with-store-cart-api-hydration.js
```diff
@@ -13,13 +18,45 @@ import { useSelect } from '@wordpress/data';
...
+
+			const needsUpdateFromAPI =
+				! isNaN( cartGeneratedTimestamp ) &&
+				! isNaN( lastCartUpdate ) &&
+				lastCartUpdate > cartGeneratedTimestamp;
+
+			if ( needsUpdateFromAPI ) {
+				setIsCartDataStale();
+			}
+		}
```
ralucaStan, line 61: What works better here is to don't do `|| ''` for lastCartUpdateRaw. Test if there is a value and then do all the logic. This way we don't do unnecessary operations. 
```js
const lastCartUpdateRaw = window.localStorage.getItem( LAST_CART_UPDATE_TIMESTAMP_KEY );
        
if (lastCartUpdateRaw) {
        const lastCartUpdate = parseFloat( lastCartUpdateRaw );
		const cartGeneratedTimestamp = parseFloat(
		    cartData.current.generated_timestamp
	    );
        
	const needsUpdateFromAPI =
		! isNaN( cartGeneratedTimestamp ) &&
		! isNaN( lastCartUpdate ) &&
		lastCartUpdate > cartGeneratedTimestamp;
        
			        if ( needsUpdateFromAPI ) {
				        setIsCartDataStale();
			        }
}
			
```

assets/js/hocs/with-store-cart-api-hydration.js
ralucaStan: sure, sounds good

assets/js/data/cart/selectors.ts
```diff
@@ -59,6 +59,16 @@ export const isApplyingCoupon = ( state: CartState ): boolean => {
...
 };
 
+/**
+ * Returns true if cart is stale, false if it is not.
+ *
+ * @param {CartState} state The current state.
+ * @return {boolean} True if the cart data is stale.
+ */
+export const isCartDataStale = ( state: CartState ): boolean => {
+	return !! state.metaData.isCartDataStale;
```
ralucaStan: Don't think !! is needed here, these are booleans

ralucaStan (approved):
Looks good, I added a comment for a small improvement. Good to merge after that

PR #3961, opened 2021-03-12
Add new config property for payment method registration that allows for additional saved customer token handling.

### LINKED ISSUE #3956: Provide payment methods with a way to interact with submitted saved payment method client side.
Some payment methods, such as Stripe, need to be able to do requests to a payment method API for saved payment methods client side before submitting to the server for final processing OR they may need to handle a response from the server requiring more information before the saved payment token is processed (for example 3DS via payment intents).

Currently in the case of Stripe, this 3DS processing is captured and handle fine by the Stripe payment method because it is mounted in the checkout during the various events that fire. However, in the case of saved payment methods, the registered client component for Stripe is _not_ active, and thus it's not able to interact with responses from the server.

One thing I'm going to try as a solution is to introduce a new configuration property on the payment method API that payment methods use to provide a component rendering null (so the component doesn't have any visible ui, it simply contains the logic the payment method provides for handling any scenarios it has with saved payment tokens), that receives the same information other registered components receives along with details about any saved token etc. This will allow the payment method to hook into our checkout events and do any of the necessary handling from either the pre checkout processing or post checkout processing from the server (i.e in the case of Stripe, triggering the 3DS modal).

nerrad:
Fixes #3956 

See the reference issue for some background on the need for the changes in this pull request. Essentially this pull request fulfills the need for some payment methods that support saved customer tokens for future payments (such as Stripe) to handle any extra authorization or related logic when those selected saved tokens are processed.

For example, 3DS type of payment cards usually require some sort of additional authorization from the customer (2FA, text etc) to complete the payment. Via the Stripe API this is done using [payment intents](https://stripe.com/docs/payments/payment-intents). While the current Stripe Payment method integration for the checkout flow supports payment intents on _new_ purchases using a 3DS type card, it never had access to client side handling of server responses indicating further action was needed when the payment was being made with a _saved_ 3DS type card.

The solution introduced in this PR, is to expose a new configuration property on the payments registration function for payment methods to provide a component that handles any additional logic related to saved payment methods. In turn, this registered component is rendered for any **selected** saved payment method token that matches the `name` of the registered payment method.

A few additional notes for reviewers:

- Given this is a significant bug that (fortunately) has seemingly gone unnoticed in the wild but nevertheless impacts potential sales for merchants, we need to get this fix out sooner. This meant that I avoided any TypeScript work in the affected components for now.
- I'm not sure on the name of the config property, if you have suggestions for something other than `savedTokenComponent`, I'm all ears (maybe `savedTokenHandlerComponent`?)
- I did this in concert with implementation changes in the [corresponding Stripe PR I started](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1467) for migrating the stripe integration. This helped to confirm the new API is sufficient for the fixes needed in the Stripe integration. I _did not_ make the changes in the bundled implementation with C&C blocks figuring that customers should be encouraged to be on the latest Stripe extension anyways. I'm on the fence whether this is the best route to take and welcome feedback here.
- The new configuration property is optional and has a default set for registered payment methods registered without it. That preserves back-compat (and the scenario where most payment methods won't need this).

## Testing

First, the main thing to verify, is that this branch doesn't impact the current behaviour of payment methods being used for purchases. So smoke testing the checkout flow for bundled payment methods to ensure they work fine without errors (including saved payment methods and purchasing in incognito) is important.

- Test making a purchase with cheque payment method.
- Test making a purchase with bundled Stripe (with non 3DS type card).
- Test making a purchase with express payment method (with non 3DS type card).
- Test making a purchase with a saved payment method (this test is excluded for incognito mode because you need to be logged in).

The next set of tests should be done against a build of the Stripe extension using [this pr](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1467).  You also need to be logged in for this test.

- complete a checkout using a 3DS test card (`4000 0000 0000 3063`) with the "Save payment information to my account for future purchases" checkbox checked and verify you get the additional authorization modal for authorizing and completing the purchase. After completing the purchase, verify the order in the background shows the payment completed successfully.
- Do a new purchase on the frontend and select the card you used above from the saved payment method options (it should be available as an option in the checkout). When submitting the checkout, you should see a 3DS modal pop up for additional verification and continuing should complete the order (verify the payment was successful in the backend).
- Also try the above, but cancel the modal instead of authorizing, and validate that an error message shows in the checkout. You should also be able to pay with an alternative payment method at that point.

### Changelog

> A new configuration property is available to registered payment methods for additional logic handling of saved payment method tokens. See the [updated payment method integration docs for details](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/payment-method-integration.md#payment-methods---registerpaymentmethod-options-).

### DIFF
--- a/docs/extensibility/payment-method-integration.md
+++ b/docs/extensibility/payment-method-integration.md
@@ -6,6 +6,12 @@ The checkout block has an API interface for payment methods to integrate that co
 
 - [Client Side integration](#client-side-integration)
   - [Express payment methods - `registerExpressPaymentMethod( options )`](#express-payment-methods---registerexpresspaymentmethod-options-)
+    - [`name` (required)](#name-required)
+    - [`content` (required)](#content-required)
+    - [`edit` (required)](#edit-required)
+    - [`canMakePayment` (required):](#canmakepayment-required)
+    - [`paymentMethodId`](#paymentmethodid)
+    - [`supports:features`](#supportsfeatures)
   - [Payment Methods - `registerPaymentMethod( options )`](#payment-methods---registerpaymentmethod-options-)
   - [Props Fed to Payment Method Nodes](#props-fed-to-payment-method-nodes)
 - [Server Side Integration](#server-side-integration)
@@ -72,12 +78,12 @@ This should be a React node that will output in the express payment method area 
 #### `edit` (required)
 This should be a React node that will be output in the express payment method area when the block is rendered in the editor. It will be cloned in the rendering process. When cloned, this React node will receive props from the payment method interface to checkout (but they will contain preview data).
 
-#### `canMakePayment` (required): 
-    
+#### `canMakePayment` (required):
+
 A callback to determine whether the payment method should be available as an option for the shopper. The function will be passed an object containing data about the current order.
 
 ```
-canMakePayment( { 
+canMakePayment( {
     cartTotals: CartTotals,
     cartNeedsShipping: boolean,
     shippingAddress: CartShippingAddress,
@@ -87,7 +93,7 @@ canMakePayment( {
 } )
 ```
 
-Returns a boolean value - true if payment method is available for use. If your gateway needs to perform async initialization to determine availability, you can return a promise (resolving to boolean). This allows a payment method to be hidden based on the cart, e.g. if the cart has physical/shippable products (example: [`Cash on delivery`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e089ae17043fa525e8397d605f0f470959f2ae95/assets/js/payment-method-extensions/payment-methods/cod/index.js#L48-L70)); or for payment methods to control whether they are available depending on other conditions. 
+Returns a boolean value - true if payment method is available for use. If your gateway needs to perform async initialization to determine availability, you can return a promise (resolving to boolean). This allows a payment method to be hidden based on the cart, e.g. if the cart has physical/shippable products (example: [`Cash on delivery`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e089ae17043fa525e8397d605f0f470959f2ae95/assets/js/payment-method-extensions/payment-methods/cod/index.js#L48-L70)); or for payment methods to control whether they are available depending on other conditions.
 
 **Keep in mind this function could be invoked multiple times in the lifecycle of the checkout and thus any expensive logic in the callback provided on this property should be memoized.**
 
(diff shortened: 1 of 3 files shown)


docs/extensibility/payment-method-integration.md
```diff
@@ -125,6 +131,7 @@ registerPaymentMethod( options );
 
 The options you feed the configuration instance are the same as those for express payment methods with the following additions:
 
+- `savedTokenComponent`: This should be a React node that contains logic handling any processing your payment method has to do with saved payment methods if your payment method supports them. This component will be rendered whenever a customer's saved token using your payment method for processing is selected for making the purchase.
```
Aljullu, line 134: > This component will be rendered whenever a customer's saved token using your payment method for processing is selected for making the purchase.

Is this true? My understanding is that the component is always rendered, not only when the saved payment method is selected.

Aljullu (approved):
Good catch with this issue. Code changes look good to me and I couldn't find any regression when testing.

These are all the paths I tested:
* WC Blocks in `add/payment-method-saved-token-handler-config` and Stripe payment gateway 4.9.0:
  * Check payment method.
  * Express payment method.
  * New credit card.
  * Saved credit card.
  * New 3DS credit card.
  * Saved 3DS credit card (I could reproduce the issue: the validation modal didn't appear and the payment was not processed).

* WC Blocks in `add/payment-method-saved-token-handler-config` and Stripe payment gateway in `feature/support-block-checkout`:
  _Note: for some reason, I wasn't able to get the Stripe payment gateway built from your branch to appear in blocks when testing in my ephemeral site, so I had to test it locally and because of that I couldn't test express payment methods with it. But I verified that everything else worked:_
  * Check payment method.
  * New credit card.
  * Saved credit card.
  * New 3DS credit card.
  * Saved 3DS credit card (in this case, the validation modal appeared and the payment was processed).
  
> I did not make the changes in the bundled implementation with C&C blocks figuring that customers should be encouraged to be on the latest Stripe extension anyways. I'm on the fence whether this is the best route to take and welcome feedback here.

Do we know when this fix will be released in the Stripe payment gateway extension? I think depending on that we should make a decision.

docs/extensibility/payment-method-integration.md
nerrad: Yes, [it's true](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3961/files#diff-04e6a0d14db1a0e96784bf8db371e0e8e576c2bbc067012bd002cd54f5123477R166-R174). The component is only rendered when the associated payment method for the `activePaymentMethod` related to the saved token is selected.

However, your question got me thinking and I realized (with testing) that  the saved token handler component is also rendered when the _payment method_ is active in the payment method area (i.e. in this case, new Stripe CC).  So I addressed that in 2e5827a.

nerrad:
> Do we know when this fix will be released in the Stripe payment gateway extension?

The PR is actively in review/testing now (this is how the issue was surfaced) so I imagine as soon as the PR is merged it will go out with the next Stripe extension release.

I think if we were seeing a number of customer reports of this we'd want to put the fix in the embedded integration but given the bug has been here since we first released the blocks last May (with no reports of the bug that I'm aware), I think we're okay to wait for the Stripe extension release? Either way, we can always do a future PR adding the fix if we want/need to in the future.

PR #3967, opened 2021-03-16
Add Google Analytics Product Block and Checkout Events

### LINKED ISSUE #3935: GA: Emit Events when viewing product blocks

### LINKED ISSUE #3951: GA: Checkout Events
For the Google Analytics Integration, the following gtag events should be added (from https://developers.google.com/analytics/devguides/collection/gtagjs/events) to checkout:

- add_payment_info
- begin_checkout
- [checkout_progress](https://developers.google.com/gtagjs/reference/event#checkout_progress) and set_checkout_option
- purchase

### LINKED ISSUE #3959: GA: Emit product view events in SSR Blocks
We need to emit events in the SSR Product Blocks render function to track views. I did a quick test and it looks like JavaScript added to the render() method _does_ run, so it should be possible to use gtag directly in those instances.

mikejolley:
This PR adds events for product block views, add to cart links, search functionality, exceptions, and checkout. It demonstrates how actions can be used from blocks themselves, or triggered by inline JS in the case of SSR and non-dynamic blocks.

I've moved the GA code to it's own utility class now that more PHP side code is needed.

Closes #3935
Closes #3951
Closes #3959

See the setup instructions in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3953. Please ensure the Google Analytics Browser Extension is not in 'debug mode' as this will cause events to not fire.

1. View a single product block. Check a view item event fires.
2. Click the add to cart button in single product block. Check add to cart event fires.
3. Search for something using the product search block. Check for the event.

To test the checkout functionality,

1. Visit checkout with items in your cart. You will see a begin_checkout event
2. Start editing fields. Depending on the field you edit (which step) you will see a checkout_progress event. e.g. adding billing information will add checkout_progress with a number indicating the step you are on.
3. Select a shipping rate. See that a set_checkout_option event fired.
4. Select a payment method. See that a set_checkout_option fired.
5. Leave a required field blank and place order. See _exception_ event.
6. Place order. See add_payment_info event.

To test SSR blocks:

1. Insert product grid block into a page
2. View the page. Check a view_item_list event fired in GA.


mikejolley:
I accidentally forgot to branch when adding checkout events, so I'm just going to add that here also 🤦🏻

ralucaStan:
Things I noticed while following the testing steps:

> Visit checkout with items in your cart. You will see a begin_checkout event

When I have a saved payment I see two events here: 
```
Tracking event begin_checkout
Tracking event set_checkout_option
```

> Start editing fields. Depending on the field you edit (which step) you will see a checkout_progress event. e.g. adding billing information will add checkout_progress with a number indicating the step you are on.

~~I haven't seen any number, only the `checkout_progress`.~~ This was about the log from utils.js

 Also, I'm wondering if it's good to show a number instead of a let's say id; I'm just thinking the order might change in the future. 

> Leave a required field blank and place order. See exception event.

For me, the exception event is never the only one being triggered
```
Tracking event add_payment_info
Tracking event exception
```

> Place order. See add_payment_info event. 

I see this but I wonder why we show `add_payment_info` and not `purchase`.


> View the page. Check a view_items event fired in GA.

For All products, I see `view_item_list`, but I guess that's what you meant here.




The rest works as expected.

assets/js/extensions/google-analytics/index.ts
```diff
@@ -13,89 +17,217 @@ import {
...
+	} ): void => {
+		trackEvent( 'add_to_cart', {
+			event_category: 'ecommerce',
+			event_label: __( 'Add to Cart', 'woo-gutenberg-products-block' ),
+			items: [ getProductFieldObject( product, quantity ) ],
+		} );
+	}
+);
+addAction(
+	`${ actionPrefix }-set-cart-item-quantity`,
```
ralucaStan: I think it's good to follow a pattern here as well as you did for `${ actionPrefix }-checkout-....`, so `${ actionPrefix }-cart-set-item-quantity`

assets/js/extensions/google-analytics/index.ts
```diff
@@ -13,89 +17,217 @@ import {
...
-		`${ actionPrefix }-remove-cart-item`,
-		namespace,
-		trackRemoveCartItem
-	);
-}
+/**
+ * Product Events.
+ */
+addAction(
+	`${ actionPrefix }-render-product-list`,
```
ralucaStan: `${ actionPrefix }-product-list-render`?

assets/js/extensions/google-analytics/index.ts
```diff
@@ -13,89 +17,217 @@ import {
...
+	( { paymentMethodSlug }: { paymentMethodSlug: string } ): void => {
+		trackCheckoutOption( {
+			step: 6,
+			option: __( 'Payment Method', 'woo-gutenberg-products-block' ),
+			value: paymentMethodSlug,
+		} )();
+	}
+);
+addAction( `${ actionPrefix }-checkout-submit`, namespace, (): void => {
+	trackEvent( 'add_payment_info' );
```
ralucaStan, line 109: I think this should be `trackEvent( 'purchase' );`

assets/js/extensions/google-analytics/index.ts
```diff
@@ -13,89 +17,217 @@ import {
...
-	} );
-};
+/**
+ * Track customer progress through steps of the checkout. Triggers the event when the step changes.
+ *
+ * 1 - Contact information
+ * 2 - Shipping address
+ * 3 - Billing address
+ * 4 - Shipping options
+ * 5 - Payment options
```
ralucaStan: On line 76 you send a 6 for step and I don't see step 5 being send anywhere
```
trackCheckoutOption( {
			step: 6,
			option: __( 'Payment Method', 'woo-gutenberg-products-block' ),
			value: paymentMethodSlug,
		} )();
```

ralucaStan:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3935 mentions 
> fire a "view" event.  [...] when Clicking a link in a product block

I could not find anything related to this in the changes.

assets/js/extensions/google-analytics/index.ts
```diff
@@ -13,89 +17,217 @@ import {
...
+			search_term: searchTerm,
+		} );
+	}
+);
 
-initialize();
+/**
+ * Exception events.
+ */
+addAction(
```
ralucaStan, line 293: I didn't find this documented in the tickets or the PR description. Let's add this at least in the PR description and mention a testing step for it

ralucaStan (changes requested):
👏 for this big PR. I've only checked in detail the front-end side, I'd recommend getting a second 👁️ for the PHP changes.

assets/js/extensions/google-analytics/index.ts
mikejolley: @ralucaStan Purchase is the event fired after payment - it's implemented by the GA extension on the "thanks" page. We shouldn't use that here because some payment methods send you off site to make payment.

ralucaStan: This info could be added as a comment here to prevent any confusion.

mikejolley: Added in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3967/commits/be337d6f0a6310774537eab92e281328c935b807

assets/js/extensions/google-analytics/index.ts
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3967/commits/0bab0de1e09cadf3f04b123b36d625b54b8c09ea

assets/js/extensions/google-analytics/index.ts
mikejolley: I've added a mention for exceptions now. It was also in testing step 5. It's only used in checkout.

mikejolley:
> When I have a saved payment I see two events here:

That might be expected if it selects the payment method automatically. I think it's ok to track events even if they fire together. 

> Also, I'm wondering if it's good to show a number instead of a let's say id; I'm just thinking the order might change in the future.

@ralucaStan The docs don't seem to indicate this is possible. They expect a number. I think this is for the merchants benefit anyhow; the numbers should align without however they have set their checkout up.

> For me, the exception event is never the only one being triggered

Do you think this was maybe also related to saved methods?

> fire a "view" event. [...] when Clicking a link in a product block
> I could not find anything related to this in the changes.

Clicking a link (post title) in a product block should track it as a view. Did that work for you?

assets/js/extensions/google-analytics/index.ts
mikejolley: See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3967/commits/de625a8e1eec8b7b55877a3e77b533264e61ffd3

assets/js/extensions/google-analytics/index.ts
mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3967/commits/de625a8e1eec8b7b55877a3e77b533264e61ffd3

mikejolley:
@ralucaStan Some questions above, but majority of feedback has been addressed, thanks :)

ralucaStan:
> Do you think this was maybe also related to saved methods?

This happens as well without having saved payment methods.

mikejolley:
@ralucaStan I tested that again and I see exception + add_payment_info. We get both because the checkout is submitted (add_payment_info) and then the exception is thrown afterwards. I think thats ok?

ralucaStan:
> @ralucaStan I tested that again and I see exception + add_payment_info. We get both because the checkout is submitted (add_payment_info) and then the exception is thrown afterwards. I think thats ok?

I think so too but it wouldn't hurt to have another opinion here as I am not experienced with GA events. This should not hold back the PR tho'

ralucaStan (approved):
👍 really happy with how this turned work. Really good work.

mikejolley:
Thanks @ralucaStan. 

> I'd recommend getting a second 👁️ for the PHP changes.

@nerrad would you check the PHP side please?

src/Domain/Services/GoogleAnalytics.php
```diff
@@ -0,0 +1,197 @@
...
+					"
+					const forms = document.querySelectorAll( '.wc-block-product-search form' );
+
+					for ( const form of forms ) {
+						form.addEventListener( 'submit', () => {
+							const field = form.querySelector( '.wc-block-product-search__field' );
+
+							if ( field && field.value ) {
+								wp.hooks.doAction( 'experimental__woocommerce_blocks-product-search', { searchTerm: field.value } );
+							}
```
nerrad: So this hook is actually being set by GoogleAnalytics and not blocks code? Then the GA script is going to fire on this hook point?

I would expect that instead of this, the rendered blocks would have the hook point that _any_ script could then hook into.

The concern here is this seems too tightly coupled to GoogleAnalytics. Where would this code live when GA specific stuff is moved to the extension?

nerrad (commented):
Reviewing just the PHP side of things, I have a few questions/concerns.

I'm not sure I see the necessity of the `experimental__woocommerce_blocks_product_grid_render_products` filter here and I'm wary about the signal this gives that it will be possible for folks to filter the rendered products at this level. It appears that what the `GoogleAnalytics` class is adding via this filter could maybe be done via a client side hook that is setup as an inline script added for product grid which fires a `wp.hooks.doAction` and then gtag just hooks in on that action? That would expose the action for other client side things that might want to use it and we can remove the need for a filter. Likely that will be more future proof too as we move blocks away from the SSR rendering we're currently doing?

The same applies to `experimental__product-search_render_callback` which actually seems to be firing `wp.hooks.doAction`. As per my comment there, this doesn't seem like something that needs coupled to Google Analytics code, but instead could live directly in the block itself as an inline script that GA client side code would hook into?

If anything I've suggested above isn't clear, please ping me and we can discuss :)

mikejolley:
> The same applies to experimental__product-search_render_callback which actually seems to be firing wp.hooks.doAction. As per my comment there, this doesn't seem like something that needs coupled to Google Analytics code, but instead could live directly in the block itself as an inline script that GA client side code would hook into?

@nerrad I added an action to the render method of all blocks as an insertion point for extensions to add inline scripts when blocks are used. GA extension uses that hook to insert it's inline code that handles the event and fires a hook action for logging to GA.

So Blocks, on the PHP side fires `experimental__product-search_render_callback` when the block is rendered.
GA.php code listens for that hooks and injects it's client side code (form listener and wp.hook).
GA client side code uses the wp.hook to trigger the event when the form is submitted.

Bare in the mind the GA php and client side code will be in an extension eventually, so nothing else can be done inline. 

Also bare in mind, product search is not a dynamic block. It's just HTML.

Does that make sense now?

mikejolley:
@nerrad For the Grid blocks (`experimental__woocommerce_blocks_product_grid_render_products`), I was also on the fence about that one but the problem is that its Server Side Rendered, so the client lacks the product data needed to make the request to GA.

I added the hook so the GA integration could access product data and inject it's inline handling accordingly. Do you have an alternative idea for that?

nerrad:
> @nerrad I added an action to the render method of all blocks as an insertion point for extensions to add inline scripts when blocks are used. GA extension uses that hook to insert it's inline code that handles the event and fires a hook action for logging to GA.

Right I understand that, but what I'm suggesting instead:

### For the `experimental__woocommerce_blocks_product_grid_render_products` filter implementation:

Instead, add inline JS invoking a `doAction` hook passing _along the data_ you want to make accessible to any client code (product details and categories for instance). So you can still expose the dynamic data server side via the block. Then you remove the necessity of the php action hook and extensions just hook in client side. So GA would simply hook in client side to `addAction` and grab the data for the `gTag` invocation. A couple benefits of this approach:

- We aren't exposing a php filter that presumably makes the block rendering more fragile.
- Gives us flexibility if/when we eventually migrate these grid blocks to use Product Element blocks.
- We have more control over the data we expose for extensions hooking in client side (and can potentially add to it if needed).

One downside though is the potential extra overhead of always generating this extra data for blocks and sending over the wire (which wouldn't be just limited to GA in this case).  Even then though, the argument could be made that we have more control over optimizing this in the future whereas with a php filter...anything goes.

### For the `experimental__product-search_render_callback` action implementation

I still don't see why this particular code needs to be added by GA. It feels like a good enhancement of the search block to just fire a `doAction` with the search term in our _built_ client code that extension client scripts can hook into.

mikejolley:
@nerrad I have not done the other one yet, but is this what you had in mind? https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3967/commits/6d3ddcaa08c4b66bef3ffbec550409faca3b31e5

nerrad:
> @nerrad I have not done the other one yet, but is this what you had in mind? 6d3ddca

Yes, exactly! No PHP filter needed, and GA just hooks in client side and registers an action. 

One downside to this approach (which existed with the php filters as well) is that the server is building the client code. What we could do in the future is decouple things a bit so the client invokes the action with data it receives from a client registry (eg wp.data) which initially might be populated/hydrated by data passed through via the `AssetDataRegistry` but in a headless environment might come from a REST API endpoint.

src/Domain/Services/GoogleAnalytics.php
mikejolley: Some of the hooks/points seem pretty unique to GA which is why I was trying to generalise the hooks and have GA handle most of the more niche stuff like this one. I can't think of much uses besides GA for tracking this form submission ¯\_(ツ)_/¯

mikejolley:
@nerrad Ok, ready for your re-review.

src/Domain/Services/GoogleAnalytics.php
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Service class to integrate Blocks with the Google Analytics extension,
+ */
+class GoogleAnalytics {
+	/**
+	 * Constructor
+	 */
+	public function __construct() {
+		if ( ! Package::feature()->is_experimental_build() ) {
```
nerrad, line 24: I think we should be passing in `\Automattic\WooCommerce\Blocks\Domain\Package` as a dependency in the constructor instead of using this static class. That way the `GoogleAnalytics` class can not only be more easily tested by mocking the incoming `Package` instance but it also more clearly indicates the dependencies.

src/Domain/Services/GoogleAnalytics.php
```diff
@@ -0,0 +1,99 @@
...
+		add_action( 'init', array( $this, 'register_assets' ) );
+		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
+		add_filter( 'script_loader_tag', array( $this, 'async_script_loader_tags' ), 10, 3 );
+	}
+
+	/**
+	 * Register scripts.
+	 */
+	public function register_assets() {
+		$asset_api = Package::container()->get( AssetApi::class );
```
nerrad: I realize this is likely in place because of the thoughts around extracting this eventually into the GA plugin but until then, I think it'd be better to explicitly list `AssetApi` as a dependency in the constructor. While it'll have to be reworked anyways (either way) when moved to GA because the AssetApi class only works for paths in the blocks plugin, I think by moving to the constructor now, we're promoting the kind of patterns we want to see in code living in the repo?

This applies for `enqueue_scripts` as well.

src/Domain/Services/GoogleAnalytics.php
```diff
@@ -0,0 +1,99 @@
...
+	 *
+	 * @param AssetApi $asset_api Asset API class Instance.
+	 */
+	public function enqueue_scripts( $asset_api ) {
+		$asset_api = Package::container()->get( AssetApi::class );
+
+		// Require Google Analytics Integration to be activated.
+		if ( ! class_exists( 'WC_Google_Analytics_Integration' ) ) {
+			return;
+		}
```
nerrad: I wonder if we should shift this check to the [bootstrap code](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3967/files#diff-2648b8b5c4005de7e41408364bf3219dbd03e6d11802d5eba3be021ca7c79440R232-R237) instead and don't even instantiate GA if it's not available? If we do that, you might need to delay the initial `get` from the container in the Bootstrap until a late priority on the `plugins_loaded` hook?

Advantage of moving to bootstrap loading is it avoids some overhead of instantiating the class unnecessarily. However, this is a weak reason so I'm okay with you not doing anything here.

nerrad (approved):
I've reviewed the PHP code again and the changes you made here Mike definitely look better for how the hooks are exposed and GA integrates. Thanks for applying my feedback there.

I have a few more comments around the PHP dependencies and some implementation details that I'd like considered but I consider non-blocking for this PR so it's a call I'm fine with you making on what you address.

I'm pre-approving based on the code review, but I haven't tested.

src/Domain/Services/GoogleAnalytics.php
mikejolley: I'm not going to fix this one because the experimental build check will be removed before I wrap up the project in a new PR.

PR #3982, opened 2021-03-19
Refactor Payment Methods Integration API to fire `onPaymentProcessing` event with saved tokens.

### LINKED ISSUE #3980: Improve Payment Methods integration API so either `onPaymentProcessing` event fires for savedTokenHandler components or `onCheckoutBeforeProcessing` allows updating `POST`
I was notified by the Square payment integration team (hi @mattallan 👋 ) that their saved token handling works a bit differently than the payment method integration API assumes and requires the ability to get the 3DS verification token before the initial checkout submission. The current `onCheckoutSuccessAfterProcessing` event won't work here because the Square extension still needs to process this buyer_verification_token server side after the authorization (to create the payment) and currently `onCheckoutSuccessAfterProcessing` will not have a path that re-submits to the server _unless_ there is an error and `retry` is true. That ui/ux workflow won't work for the Square needs here.

## Options

There are a few options that will resolve Square needs here:

### 1. Provide a way for `onCheckoutSuccessAfterProcessing` callbacks to trigger doing a `resubmit` to the checkout server.

This scenario would expose a way for handlers to indicate they've updated something in the checkout that requires a resubmission to the server for re-validating whatever was updated.

Pros:
- likely the least amount of impact to the current flow and preserves the current behaviour around what events update the POSTED data.
- checkout state is already updated by the handlers via the `setComplete` action (but only if `! shouldRetry( response)` - which is kind of odd here, because the state won't get used anywhere in this scenario).

Cons:
- Re-submitting at this point seems kind of counter-intuitive given the checkout has _already_ been processed. The current flow already accounts for scenarios where if some sort of error occurs at this point, then the checkout state goes to error state and then idle (if possible shopper to fix error state).
- Potential side-effects on the checkout server processing side if it's already been processed successfully and then re-submitted?

### 2. Expose `onPaymentProcessing` event for checkout submission including savedPaymentTokens

This scenario would refactor the existing flow so that the `onPaymentProcessing` event fires as a part of saved tokens being included in the checkout submission.

The reason it doesn't currently is because the assumption was made in the design that if there is a saved token, that means there is no _client side payment processing_ needed, and thus the payment status for a saved token is `SUCCESS` (which is a status saying client side payment processing completed successfully). 

Pros:
- The original assumption that client side processing isn't needed for saved tokens is flawed, and including this as a legit event would keep the flow consistent anywhere payment methods are involved.
- Modification of what gets sent through POST is updated more cleanly by payment methods as a part of an event that makes sense for payment methods.

Cons:
- There could be some difficulty here in ensuring we preserve the submission of saved tokens _without_ a handler where the payment status is set to 'SUCCESS' for the `onPaymentProcessing` event. However, I think we could resolve this by simply having a default handler set for saved tokens that don't have one explicitly defined. Alternatively (which might be better), when the `onPaymentProcessing` event goes through all the subscribed observers, and the PAYMENT_STATUS has not been updated to success, we could trigger the update to success so that checkout proceeds (probably should be doing this anyways if we're not).

### 3. Allow for updating the checkout state with the `onCheckoutBeforeProcessing` event.

Currently this event doesn't impact any updating of state (other than validation/error state) and is mostly used for initial checkout form validation. With this option, we'd also allow updating more of the state (in both contexts, when checkout returns to idle, and when checkout is set to Processing).

Pros:

- exposes another event where state can be updated.

Cons:

- The naming of this event is actually a bit unfortunate given it mostly has the purpose of validation of things in the checkout. So even though it does fire before sending to the server, it only handles validationErrors and messages that are the response from event subscriber callbacks. Might be good to consider renaming to `onCheckoutValidationBeforeProcessing` to make the event more clear (we could deprecate the old event function and just redirect to the new name).
- Updating the state (beyond what is currently done) during this process will likely have unintended consequences because it currently was designed explicitly for validation of the checkout before passing on to other stages in the checkout flow (not for updating what gets sent to the server).

## What do do?

I'm leaning towards Option 2 here because that seems like the most optimal improvement that resolves this situation AND also preserves the original intent behind the design of the flow. It also helps ensure anything to do with Payment methods in the checkout flow behaves consistently (event saved tokens) where if there is some form of payment method, the `onPaymentProcessing` event will always fire.

As an aside, I also think it be a good idea to rename the `onCheckoutBeforeProcessing` event to `onCheckoutValidationBeforeProcessing`.

nerrad:
Fixes: #3980 

Currently there exists what is arguably a flaw in the checkout event system in that `onPaymentProcessing` event can be observed by express payment methods and regular payment methods but _not_ by saved payment methods (implementing the [`savedTokenComponent`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3961) configuration property). This inconsistency is a flaw in that some payment methods implementing this component will have no way to reliably update the `paymentMethodData` in the checkout state before it is passed along on the checkout endpoint request for server side processing. You can read more about where this can manifest in #3980.

A significant reason for why this inconsistency existed is because when the system was initially designed, an assumption was made that saved tokens would always be processed server side and payment method logic would never need to interact on the client. This was a faulty assumption which I started rectifying in #3961 and now finished with this PR here.

As a result, this PR implements the following changes to ensure the `onPaymentProcessing` event is accessible to the `savedTokenComponent`:

- Implemented a `started` action creator that returns an action object that when dispatched will allow for updating the payment method data in the checkout state. 
- The above allows the checkout flow to "initialize" selected saved payment method tokens to a `started` status for the payment method and store data about the saved token in the checkout state. This replaces the existing logic that was setting the payment method status for saved tokens to "success".
- Some slight improvements of types (TS) to account for the fact that both `started` and `success` payment method status changes allow for _retaining_ existing payment method data in the state if it's not provided by the action. This improves performance because there are many places where just the payment status needs updated by checkout _without changing anything else in the state_. 
- The above change does however mean that any place we change the status to success or started where the payment method data does need wiped, must be done _explicitly_ (by passing in an empty object to the dispatched action). Thus one change was needed here for express payment methods. Also where express payment methods restored the saved token status (if the modal was closed without completing), the action dispatched was switched from `success` to `started` in order to reflect the changes made for saved token handling.

Outside of the necessary TypeScript type changes for existing typescript files, I didn't implement typescript anywhere else mostly because there's some time sensitivity around the need for this API change.

Regarding documentation, I did do a pass through the existing documentation and I don't see anywhere that needs updated. Indeed, the existing documentation already pointed to the `onPaymentProcessing` event being available for `savedTokenComponent` so the changes here are actually more in line with what the documentation already indicates!

## To Test

The changes here implement any payment method type processing, so it's important to verify there is no impact to existing payment method handling by these changes. This means setting up an environment where you have the core payment methods and Stripe (along with Stripe express payments, either Chrome Pay or Apple Pay) available for testing.

* [ ] Verify that you can pay with a saved token.
* [ ] Verify that you can pay with an express payment method (Apple Pay or Chrome Pay - either working should be sufficient).
* [ ] Verify that you can pay with a Stripe Credit Card
* [ ] Verify that you can pay with any other Woo core payment method.
* [ ] Try all of the above (except saved token) in incognito mode.

Besides the above smoke tests, try intentionally confusing or breaking the system. Some example scenarios:

* [ ] When checkout loads with available multiple saved tokens (which you might have to setup in previous purchases), try switching between saved tokens and different payment methods, and back to a saved token before completing the purchase. Ensure the saved payment method you selected was used for the payment.
* [ ] Try different payment methods and triggering a scenario that fails payment or causes an error. In the case of Stripe CC (or express payment) there are a [bunch of test cards](https://stripe.com/docs/testing) you can access for triggering different payment failure conditions. After a payment method failure, ensure you can pay with an alternative method of payment that should succeed without issues.
* [ ] Trigger starting an express payment method, selecting different shipping addresses and different rates and ensure it completes successfully after submitting.
* [ ] Do the above, but instead of completing the payment, just cancel and close the modal. Ensure that you can make a payment with the saved payment method token after doing so without issue.
* [ ] Optional (I tested this fairly throughly): For this test you'll have to use Stripe built with [this pr](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1467). Verify that if your saved token is using a 3DS card always 3DS auth ( such as `4000002760003184`), then the 3DS validation process still works when attempting to make a payment using the token.

The above all validates that the changes in this PR don't break _existing_ behaviour. There's no easy way to test (currently) the actual implementation exposing `onPaymentProcessing` event for savedTokenComponents without manually implementing a dummy test. I did this to verify it works as expected. However, this PR will also be tested as a part of the work being done for the Square extension.

### Changelog

> Payment methods implementing the `savedTokenComponent` configuration property will now have the `onPaymentProcessing` event available to the registered component.

### DIFF
--- a/assets/js/base/context/cart-checkout/payment-methods/reducer.ts
+++ b/assets/js/base/context/cart-checkout/payment-methods/reducer.ts
@@ -24,7 +24,7 @@ const reducer = (
 	state = DEFAULT_PAYMENT_DATA_CONTEXT_STATE,
 	{
 		type,
-		paymentMethodData = undefined,
+		paymentMethodData,
 		shouldSavePaymentMethod = false,
 		errorMessage = '',
 		paymentMethods = {},
--- a/assets/js/base/context/cart-checkout/payment-methods/reducer.ts
+++ b/assets/js/base/context/cart-checkout/payment-methods/reducer.ts
@@ -10,7 +10,7 @@ import type { PaymentMethodDataContextState } from './types';
 import type { ActionType } from './actions';
 
 const hasSavedPaymentToken = (
-	paymentMethodData: Record< string, unknown >
+	paymentMethodData: Record< string, unknown > | undefined
 ): boolean => {
 	return !! (
 		typeof paymentMethodData === 'object' && paymentMethodData.isSavedToken
@@ -24,7 +24,7 @@ const reducer = (
 	state = DEFAULT_PAYMENT_DATA_CONTEXT_STATE,
 	{
 		type,
-		paymentMethodData = {},
+		paymentMethodData = undefined,
 		shouldSavePaymentMethod = false,
 		errorMessage = '',
 		paymentMethods = {},
--- a/assets/js/base/context/cart-checkout/payment-methods/reducer.ts
+++ b/assets/js/base/context/cart-checkout/payment-methods/reducer.ts
@@ -36,6 +36,11 @@ const reducer = (
 				? {
 						...state,
 						currentStatus: STATUS.STARTED,
+						paymentMethodData:
+							paymentMethodData || state.paymentMethodData,
+						hasSavedToken: hasSavedPaymentToken(
+							paymentMethodData || state.paymentMethodData
+						),
 				  }
 				: state;
 		case STATUS.ERROR:
@@ -64,7 +69,7 @@ const reducer = (
 						paymentMethodData:
 							paymentMethodData || state.paymentMethodData,
 						hasSavedToken: hasSavedPaymentToken(
-							paymentMethodData
+							paymentMethodData || state.paymentMethodData
 						),
 				  }
 				: state;
--- a/assets/js/blocks/cart-checkout/payment-methods/saved-payment-method-options.js
+++ b/assets/js/blocks/cart-checkout/payment-methods/saved-payment-method-options.js
@@ -50,7 +50,7 @@ const getCcOrEcheckPaymentMethodOption = (
 		onChange: ( token ) => {
 			const savedTokenKey = `wc-${ method.gateway }-payment-token`;
 			setActivePaymentMethod( method.gateway );
-			setPaymentStatus().success( {
+			setPaymentStatus().started( {
 				payment_method: method.gateway,
 				[ savedTokenKey ]: token + '',
 				isSavedToken: true,
@@ -84,7 +84,7 @@ const getDefaultPaymentMethodOptions = (
 		onChange: ( token ) => {
 			const savedTokenKey = `wc-${ method.gateway }-payment-token`;
 			setActivePaymentMethod( method.gateway );
-			setPaymentStatus().success( {
+			setPaymentStatus().started( {
 				payment_method: method.gateway,
 				[ savedTokenKey ]: token + '',
 				isSavedToken: true,
@@ -114,12 +114,9 @@ const SavedPaymentMethodOptions = () => {
 
 	const updateToken = useCallback(
 		( token ) => {
-			if ( token === '0' ) {
-				setPaymentStatus().started();
-			}
 			setActiveSavedToken( token );
 		},
-		[ setActiveSavedToken, setPaymentStatus ]
+		[ setActiveSavedToken ]
 	);
 
 	useEffect( () => {
--- a/assets/js/base/context/cart-checkout/payment-methods/actions.ts
+++ b/assets/js/base/context/cart-checkout/payment-methods/actions.ts
@@ -49,9 +49,21 @@ export const failed = ( {
 export const success = ( {
 	paymentMethodData,
 }: {
-	paymentMethodData: Record< string, unknown >;
+	paymentMethodData?: Record< string, unknown >;
 } ): ActionType => ( {
 	type: STATUS.SUCCESS,
+	paymentMethodData,
+} );
+
+/**
+ * Used to dispatch a payment started status update.
+ */
+export const started = ( {
+	paymentMethodData,
+}: {
+	paymentMethodData?: Record< string, unknown >;
+} ): ActionType => ( {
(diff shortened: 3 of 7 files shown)


assets/js/blocks/cart-checkout/payment-methods/saved-payment-method-options.js
```diff
@@ -114,12 +114,9 @@ const SavedPaymentMethodOptions = () => {
 
 	const updateToken = useCallback(
 		( token ) => {
-			if ( token === '0' ) {
-				setPaymentStatus().started();
-			}
 			setActiveSavedToken( token );
 		},
-		[ setActiveSavedToken, setPaymentStatus ]
+		[ setActiveSavedToken ]
```
nerrad, line 119: This logic hasn't been needed ever since we switched away from tabs and radios to only radios for payment method selection. There is no longer a possibility for a token with a value of `'0'` in the checkout flow.

assets/js/base/context/cart-checkout/payment-methods/reducer.ts
```diff
@@ -24,7 +24,7 @@ const reducer = (
 	state = DEFAULT_PAYMENT_DATA_CONTEXT_STATE,
 	{
 		type,
-		paymentMethodData = {},
+		paymentMethodData = undefined,
```
Aljullu: What is the benefit of setting this parameter default to `undefined` instead of not setting it?
```suggestion
		paymentMethodData,
```

Aljullu (approved):
Code changes look good, and following the testing steps from the PR description worked well. For express payment methods I tested with Chrome Pay. I also did some extra testing trying to break the system without luck, so LGTM.

assets/js/base/context/cart-checkout/payment-methods/reducer.ts
nerrad: doh ya :) it'll be undefined anyways if the property isn't present.

PR #3989, opened 2021-03-22
Split unit tests up from e2e tests

opr:
Fixes #

This PR will split the unit tests and E2E tests up so that when an E2E test fails, due to flakiness, we don't have to run the unit tests again.

1. Ensure the automated tests run for this PR.
2. Push a commit that will cause the unit tests and E2E tests to fail. 
3. Ensure the unit tests and E2E tests fail.
4. Revert the commit and ensure the tests pass again.


Aljullu (approved):
Changes look good and I could verify tests can fail. :ship:

PR #3990, opened 2021-03-22
Expose the shouldSavePayment method value to registered payment methods

### LINKED ISSUE #3988: Add `shouldSavePayment` to the list of props passed to payment method components
Hi team 👋 I'm working on more Square Cart/Checkout block integration and I've hit another roadblock.

When handling 3D Secure requests, we need to send a `verifyBuyer()` request to Square from the client-side which needs to include an [`intent`](https://github.com/woocommerce/woocommerce-square/blob/8a31ab58b544be970a64889f90855d08d226d157/assets/js/frontend/wc-square.js#L497) value. This intent value should be set to `'STORE'` if the payment nonce is going to be used to store/save a new customer card. The issue I've ran into is that I can't see any way to determine whether the "save card" checkbox is selected from within our Payment Method component.

Is there any chance this value could be added to the [list of props](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/9fb27fe9b0b47d918258b552d438d945d213cf20/assets/js/base/hooks/payment-methods/use-payment-method-interface.js#L144) sent to payment method components?

nerrad:
Fixes #3988

This is a very low impact change that exposes the `shouldSavePayment` value as a prop to registered payment methods. Existing payment methods will not be impacted because they have no dependency on the prop.

Exposing this prop is needed by payment methods that need to know the value of this checked item client side before sending info to the server. In the case of Square, saved payment methods involving intents are processed client side before the token is passed through to the server (which differs from Stripe which is able to do all the processing server side - an assumption incorrectly expanded to cover all payment methods made in initial interface design).

## To Test

Just smoke test existing payment methods to ensure there is no breakage with existing behaviour. It should surface immediately if there is.

### Changelog

> Registered payment methods now have access to the `shouldSavePayment` prop in their components (which indicates whether the shopper has checked the save payment method checkbox).

### DIFF
--- a/assets/js/base/hooks/payment-methods/use-payment-method-interface.js
+++ b/assets/js/base/hooks/payment-methods/use-payment-method-interface.js
@@ -98,6 +98,7 @@ export const usePaymentMethodInterface = () => {
 		activePaymentMethod,
 		onPaymentProcessing,
 		setExpressPaymentError,
+		shouldSavePayment,
 	} = usePaymentMethodDataContext();
 	const {
 		shippingErrorStatus,
@@ -194,5 +195,6 @@ export const usePaymentMethodInterface = () => {
 		onSubmit,
 		activePaymentMethod,
 		setExpressPaymentError,
+		shouldSavePayment,
 	};
 };
--- a/assets/js/type-defs/registered-payment-method-props.js
+++ b/assets/js/type-defs/registered-payment-method-props.js
@@ -218,6 +218,9 @@
  *                                                                 when the express payment method
  *                                                                 modal closes and control is
  *                                                                 returned to checkout.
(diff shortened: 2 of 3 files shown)


senadir:
changes only include the typedef change, did you miss a commit?

nerrad:
> changes only include the typedef change, did you miss a commit?

Doh, I didn't save the other change for the commit I made 🤦 - I force pushed the corrected commit.

senadir (approved):
Codes check out, testing checks out. We should probably add this to https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/payment-method-integration.md#props-fed-to-payment-method-nodes as well.

nerrad:
Great call on the documentation update. Thanks!

PR #3991, opened 2021-03-22
Update data-controls and expect console logs from deprecated select control

mikejolley:
This should hopefully fix failing tests in trunk..


nerrad (approved):
Looks gtm - I like that the requirement of the expect makes it clear that we know about this change coming in data-controls (and hopefully will prompt to update when we can).

PR #3996, opened 2021-03-24
Fix click on Proceed to checkout bug when Coupon error is visible.

### LINKED ISSUE #3968: Proceed to checkout button does not work if pressed when there is a coupon validation error
## Describe the bug
In the Cart block, after I receive a coupon validation error I click on the Proceed to checkout button. It does not take me to the next step, but the button receives focus, and the error of the coupon disappears. I have to click again for the button to work and take me to Checkout.

## To reproduce

1. Make sure you use the Cart block
2. Add any product to the cart and go to the Cart block.
3. Enter an invalid coupon code.
4. See coupon validation error appear
5. Click on the "Proceed to Checkout" button
6. Notice that you are still in the Cart block
7. Click again to proceed to checkout

## Expected behavior
The button should work from the first click and I should see the Checkout block.

## Screenshots
Before I click the button
Screenshot 2021-03-16 at 15 21 02
After I click the button
Screenshot 2021-03-16 at 15 35 28

* Core version: 5.7
* WooCommerce version: 5.1.0
* WooCommerce Blocks version: 4.6.0
* Site language: EN
* Any other plugins installed: -

* OS: macOS
* Browser Chrome & Firefox. On Safari it works as expected

ralucaStan:
The bug happened because of the layout shift produced by the error message disappearance.

This PR adds an in-place fix for the Coupon code field. It adds an empty space equal to the space taken up by the error message. 
If this becomes an issue for other parts of the project we will need to implement a project-wide sollution.

Fixes #3968 

### Screenshots

The ticket has the original behavior.

Screenshot 2021-03-24 at 15 45 21
Screenshot 2021-03-24 at 15 45 13

Changes seen here.

Make sure you use the Cart block
Add any product to the cart and go to the Cart block.
Enter an invalid coupon code.
See coupon validation error appear
Click on the "Proceed to Checkout" button
Notice you are directed to the Checkout

### Changelog

> Fix the Proceed to checkout button click bug happening when the Coupon error is visible in the Cart block

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/coupon/index.js
+++ b/assets/js/base/components/cart-checkout/totals/coupon/index.js
@@ -106,6 +106,12 @@ const TotalsCoupon = ( {
 						propertyName="coupon"
 						elementId={ textInputId }
 					/>
+					{ /* prevent layout shift caused the by error message */ }
+					{ ! validationError && (
+						<div className="wc-block-components-validation-error">
+							<p></p>
+						</div>
+					) }
 				</div>
 			</LoadingMask>
 		</Panel>
--- a/assets/js/base/components/cart-checkout/totals/coupon/index.js
+++ b/assets/js/base/components/cart-checkout/totals/coupon/index.js
@@ -108,7 +108,10 @@ const TotalsCoupon = ( {
 					/>
 					{ /* prevent layout shift caused the by error message */ }
 					{ ! validationError && (
-						<div className="wc-block-components-validation-error">
+						<div
+							className="wc-block-components-validation-error"
+							aria-hidden="true"
+						>
 							<p></p>
 						</div>
 					) }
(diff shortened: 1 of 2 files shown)


mikejolley:
Just making a note that I suggested an alternative to empty elements in using e.preventDefault on the place order button click.

If thats not viable, another thing we could consider is making the notice only dismiss after re-entry to the coupon field rather than on blur.

ralucaStan:
While testing your proposal I noticed that the coupon is actually the only with an error that has a layout shift. All the other inputs in cart & checkout don't have it. So for consistency's sake and for a clean solution I will try to implement the same behaviour for the coupon input
Screenshot 2021-03-26 at 13 47 56
In checkout
Screenshot 2021-03-26 at 13 47 06

ralucaStan:
Fixed this by making sure the validation error looks & feels like other validations in the checkout & cart forms. I had to add a small space on the bottom (just CSS) because all the input fields need 1.5em space under them.

ralucaStan:
@mikejolley ready for review

mikejolley (approved):
I'm approving this on the basis that it solves the issue and matches the fields on checkout. I think these could use some work in the future though, the styling doesn't seem great—it's cramped and would break with longer messages I expect.

👍🏻

PR #4000, opened 2021-03-26
[Prototype] Add terms and conditions checkbox

### LINKED ISSUE #3258: Consider including checkbox for "I have read and agree to terms [terms]" with customizable text when there is a terms and conditions page.
The current shortcode checkout flow includes customizable Privacy Policy and Terms and Conditions text that is configurable via the customizer:

Which results in the following in the shortcode checkout ui:

This checkbox must be checked before finalising the order.

While the checkout block _does_ allow for setting the terms and conditions page (and shows a link to it when set), it does not have parity with the current feature in the shortcode checkout flow.

The current implementation was done _intentionally_ as a result of discussion in this thread: p6q8Tx-1gJ-p2 (internal link). However, there are some aspects of the discussion in that thread that still might have been missed in the initial iteration.

opr:
This PR is a prototype which adds a checkbox to accept the store's terms and conditions in the Checkout block.

The following changes have been made:

- Added an attribute to the Checkout block called `requireTermsAndConditions` and a toggle control to change its value. When toggled on, the block will display the terms and conditions checkbox.
- Added a `TextControl` option which appears when `requireTermsAndConditions` is true. This allows the user to edit the text of the checkbox label. Updating it updates the block preview in realtime.
- Added `terms_and_conditions_text` to `wcSettings` - this is the result of  wc_get_terms_and_conditions_checkbox_text()`
- Add a new method to `Checkout` called `validate_terms_and_conditions` - this validates based on the value of `terms_and_conditions_accepted` in the request _not_ being explicitly false. [There is a comment describing this here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/compare/add/terms-and-conditions-checkbox?expand=1#diff-af2c90fa556cc086b780c8fad99b68373d87fd6007e6e2ff1b4c68ebe9ccb551R387)
- Adds an action, selector, and reducer key for `isTermsAndConditionsCheckboxPristine` to track whether the checkbox has been changed. This is so we can show an error if it is not checked, but hide it if the user hasn't interacted with the checkbox yet.
- Add a new `TermsAndConditionsCheckbox` component which gets rendered on the `CheckoutForm` if the Checkout block's `requireTermsAndConditions` is true.
- Pass whether the checkbox is checked to the API. The way this is checked is by the presence of an error keyed by `terms_and_conditions` in the `useValidationContext` hook. When the checkbox is checked, this error is cleared, and re-added when it is unchecked.

Fixes #3258

#### Accessibility

#### General functionality
1. Go to WooCommerce > Settings > Advanced and ensure you have a Terms page set up.
9. Go to **Customizer** > **WooCommerce** > **Checkout** and ensure the **Terms and Conditions** text contains `[terms]`.
1. Go to the checkout block in the editor, try toggling the checkbox for required terms and conditions. See that it shows up in the editor when the toggle is on, and it does not show when the toggle is off.
2. Enable the toggle and save the page.
3. Add an item to your cart then go to the Checkout page on the front-end. See that the checkbox for terms and conditions has appeared. Ensure no error shows up.
4. Submit the form without changing the checkbox.
5. Ensure checkout does not process and a validation error appears under the terms and conditions checkbox.
6. Refresh the page.
7. Tick then untick the checkbox, ensure the error appears.
8. Tick the checkbox again and complete the checkout process. Ensure it works.

#### Customisation
1. Go to **Customizer** > **WooCommerce** > **Checkout** and modify the **Terms and Conditions** text. Optionally remove `[terms]` from it.
10. Refresh the Checkout block and ensure the new text appears.
11. If you removed `[terms]` verify there is no link in the checkbox label.
12. Put `[terms]` back if you removed it.
11. Rename the Terms and conditions page to something else, then verify the link in the checkbox label is updated.

### Automated tests 🚧 
As this is a prototype I will not add automated tests _yet_. When this work is more refined and we are happy with the direction we can add them.

### Changelog

### DIFF
--- a/assets/js/base/context/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/cart-checkout/checkout-state/index.js
@@ -64,10 +64,13 @@ const CheckoutContext = createContext( {
 		setCustomerId: ( id ) => void id,
 		setOrderId: ( id ) => void id,
 		setOrderNotes: ( orderNotes ) => void orderNotes,
+		setTermsAcceptedCheckboxPristine: ( termsAcceptedCheckboxPristine ) =>
+			void termsAcceptedCheckboxPristine,
 	},
 	hasOrder: false,
 	isCart: false,
 	shouldCreateAccount: false,
+	isTermsAcceptedCheckboxPristine: true,
 	setShouldCreateAccount: ( value ) => void value,
 } );
 
@@ -155,6 +158,10 @@ export const CheckoutStateProvider = ( {
 				void dispatch( actions.setOrderId( orderId ) ),
 			setOrderNotes: ( orderNotes ) =>
 				void dispatch( actions.setOrderNotes( orderNotes ) ),
+			setIsTermsAcceptedCheckboxPristine: () =>
+				void dispatch(
+					actions.setIsTermsAcceptedCheckboxPristine( false )
+				),
 			setAfterProcessing: ( response ) => {
 				// capture general error message if this is an error response.
 				if (
@@ -336,6 +343,7 @@ export const CheckoutStateProvider = ( {
 		checkoutState.customerId,
 		checkoutState.customerNote,
 		checkoutState.processingResponse,
+		checkoutState.isTermsAcceptedCheckboxPristine,
 		previousStatus,
 		previousHasError,
 		dispatchActions,
@@ -350,9 +358,9 @@ export const CheckoutStateProvider = ( {
 	] );
 
 	const onSubmit = useCallback( () => {
+		dispatch( actions.setIsTermsAcceptedCheckboxPristine( false ) );
 		dispatch( actions.setBeforeProcessing() );
 	}, [] );
-
 	/**
 	 * @type {CheckoutDataContext}
 	 */
@@ -378,6 +386,8 @@ export const CheckoutStateProvider = ( {
 		shouldCreateAccount: checkoutState.shouldCreateAccount,
 		setShouldCreateAccount: ( value ) =>
 			dispatch( actions.setShouldCreateAccount( value ) ),
+		isTermsAcceptedCheckboxPristine:
+			checkoutState.isTermsAcceptedCheckboxPristine,
 	};
 	return (
 		<CheckoutContext.Provider value={ checkoutData }>
--- a/assets/js/blocks/cart-checkout/checkout/attributes.js
+++ b/assets/js/blocks/cart-checkout/checkout/attributes.js
@@ -53,6 +53,10 @@ const blockAttributes = {
 		type: 'boolean',
 		default: HAS_DARK_EDITOR_STYLE_SUPPORT,
 	},
+	requireTermsAndConditions: {
+		type: 'boolean',
+		default: false,
+	},
 };
 
 export default blockAttributes;
--- a/assets/js/blocks/cart-checkout/checkout/attributes.js
+++ b/assets/js/blocks/cart-checkout/checkout/attributes.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies
  */
+import { __ } from '@wordpress/i18n';
 import { HAS_DARK_EDITOR_STYLE_SUPPORT } from '@woocommerce/block-settings';
 
 const blockAttributes = {
@@ -56,6 +57,14 @@ const blockAttributes = {
 	requireTermsAndConditions: {
 		type: 'boolean',
 		default: false,
+	},
+	termsAndConditionsText: {
+		type: 'string',
+		// translators: [terms] is the link to the Terms and Conditions page.
+		default: __(
+			'I have read and agree to the website [terms].',
+			'woocommerce'
+		),
 	},
 };
 
--- a/assets/js/blocks/cart-checkout/checkout/attributes.js
+++ b/assets/js/blocks/cart-checkout/checkout/attributes.js
@@ -3,6 +3,17 @@
  */
 import { __ } from '@wordpress/i18n';
 import { HAS_DARK_EDITOR_STYLE_SUPPORT } from '@woocommerce/block-settings';
+import { getSetting } from '@woocommerce/settings';
+
+const termsAndConditionsTextFromDb = getSetting( 'termsAndConditionsText' );
+let defaultTermsAndConditionsText = termsAndConditionsTextFromDb;
+
+if ( ! termsAndConditionsTextFromDb ) {
+	defaultTermsAndConditionsText = __(
+		'I have read and agree to the website [terms].',
+		'woocommerce'
+	);
+}
 
 const blockAttributes = {
 	isPreview: {
@@ -61,10 +72,7 @@ const blockAttributes = {
 	termsAndConditionsText: {
 		type: 'string',
 		// translators: [terms] is the link to the Terms and Conditions page.
-		default: __(
-			'I have read and agree to the website [terms].',
-			'woocommerce'
-		),
+		default: defaultTermsAndConditionsText,
 	},
 };
 
--- a/assets/js/blocks/cart-checkout/checkout/edit.js
+++ b/assets/js/blocks/cart-checkout/checkout/edit.js
@@ -9,6 +9,7 @@ import {
 	ToggleControl,
 	CheckboxControl,
 	Notice,
+	TextControl,
 	Disabled,
 } from '@wordpress/components';
 import BlockErrorBoundary from '@woocommerce/base-components/block-error-boundary';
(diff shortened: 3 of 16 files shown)


nerrad:
Thanks Thomas for picking this up. One thing I think we should try to move away from is dependency on anything in the customizer as it's quite likely the Customizer will eventually be phased out of WordPress given the shift to Full Site Editing and Block Themes (which will not use the Customizer).

So I'm wondering if there was some way we could have the text edited inline with the block itself (and maybe prepopulate initially (if empty) from whatever is saved in the option the customizer currently saves to.

opr:
> Thanks Thomas for picking this up. One thing I think we should try to move away from is dependency on anything in the customizer as it's quite likely the Customizer will eventually be phased out of WordPress given the shift to Full Site Editing and Block Themes (which will not use the Customizer).
> 
> So I'm wondering if there was some way we could have the text edited inline with the block itself (and maybe prepopulate initially (if empty) from whatever is saved in the option the customizer currently saves to.

Yep, makes sense. With this in mind I have added a `TextControl` to the block configuration options which allows the user to set the terms themself. If this option is left empty it will use the value of the `woocommerce_checkout_terms_and_conditions_checkbox_text` option in the `wp_options` table. If _this_ option is empty, there is a default value to fall back on which is: `I have read and agree to the website [terms].`

assets/js/base/context/cart-checkout/checkout-state/index.js
```diff
@@ -378,6 +386,8 @@ export const CheckoutStateProvider = ( {
 		shouldCreateAccount: checkoutState.shouldCreateAccount,
 		setShouldCreateAccount: ( value ) =>
 			dispatch( actions.setShouldCreateAccount( value ) ),
+		isTermsAcceptedCheckboxPristine:
```
mikejolley, line 389: pedantic, but "is" and "accepted" both indicate positive states so we could probably use tersmAccepted instead of isTermsAccepted throughout (you did this already in some places e.g. `setTermsAcceptedCheckboxPristine`).

assets/js/blocks/cart-checkout/checkout/attributes.js
```diff
@@ -1,7 +1,19 @@
...
  * External dependencies
  */
+import { __ } from '@wordpress/i18n';
 import { HAS_DARK_EDITOR_STYLE_SUPPORT } from '@woocommerce/block-settings';
+import { getSetting } from '@woocommerce/settings';
+
+const termsAndConditionsTextFromDb = getSetting( 'termsAndConditionsText' );
+let defaultTermsAndConditionsText = termsAndConditionsTextFromDb;
+
+if ( ! termsAndConditionsTextFromDb ) {
```
mikejolley, line 11: can we add an inline comment explaining why the default from the options DB? This might be helpful when block checkout becomes default, because that core setting can be deprecated.

assets/js/blocks/cart-checkout/checkout/attributes.js
```diff
@@ -1,7 +1,19 @@
...
+	defaultTermsAndConditionsText = __(
+		'I have read and agree to the website [terms].',
```
mikejolley, line 13: Doubts about how usable `[terms]` actually is. Is there a placeholder pattern established anywhere else in blocks?

The alternative would be a separate view link using the page title as the link text. This would avoid the user needing to add the [terms] placeholder within the string.

assets/js/blocks/cart-checkout/checkout/attributes.js
```diff
@@ -53,6 +65,15 @@ const blockAttributes = {
 		type: 'boolean',
 		default: HAS_DARK_EDITOR_STYLE_SUPPORT,
 	},
+	requireTermsAndConditions: {
```
mikejolley, line 68: Do we have handling for when there is no TOS page? The [terms] link won't work in those cases right?

assets/js/blocks/cart-checkout/checkout/edit.js
```diff
@@ -277,6 +280,38 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
 					}
 				/>
 			</PanelBody>
+			<PanelBody title={ __( 'Terms and Conditions', 'woocommerce' ) }>
+				<ToggleControl
+					label={ __( 'Terms and conditions acceptance required' ) }
+					help={ __(
+						"Require users to agree to your site's terms and conditions before completing the checkout process."
```
mikejolley, line 287:
```suggestion
						"Require users to agree to your store terms and conditions before completing checkout."
```

assets/js/blocks/cart-checkout/checkout/edit.js
```diff
@@ -277,6 +280,38 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
...
+							'woocommerce'
+						) }
+						value={ termsAndConditionsText }
+						onChange={ ( value ) => {
+							setAttributes( {
+								termsAndConditionsText: value,
+							} );
+						} }
+						help={ __(
+							'Use the tag [terms] in this box to include a link to your terms and conditions page. If left blank, a default message will be shown.',
```
mikejolley, line 309: Just curious if we could get these controls inline in the preview, or maybe that works only when using inner blocks?

assets/js/blocks/cart-checkout/checkout/terms-and-conditions-checkbox/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+		string
+	>( defaultTermsAndConditionsText );
+
+	useEffect( () => {
+		let termsText = text?.trim();
+		if ( ! termsText ) {
+			termsText = defaultTermsAndConditionsText;
+		}
+		if (
+			typeof termsText === 'string' &&
```
mikejolley, line 51: Should this type check be at the top before trim()?

assets/js/blocks/cart-checkout/checkout/terms-and-conditions-checkbox/index.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+	const isCheckboxInvalid =
+		! isTermsAcceptedCheckboxPristine &&
+		getValidationError( 'terms_and_conditions' );
+
+	const setValidationError = () => {
+		setValidationErrors( {
+			terms_and_conditions: {
+				message: __(
+					'You must agree to the terms and conditions to continue.',
```
mikejolley, line 67: In the current PR, the terms text can be changed into anything. Does that affect this notice? Should we inherit anything from the terms page title?

src/StoreApi/Routes/Checkout.php
```diff
@@ -217,6 +217,7 @@ protected function get_route_post_response( WP_REST_Request $request ) {
 		 * Validate items etc are allowed in the order before the order is processed. This will fix violations and tell
 		 * the customer.
 		 */
+		$this->validate_terms_and_conditions( $request );
```
mikejolley, line 220: I'm not sure about this change...

I don't think we should block API requests like this, especially if there is no 'schema' for it. I'd suggest removing this completely and only using client-side validation in the app.

mikejolley (commented):
Left some minor feedback, with the exception of the API change. I think we should only do client-side validation of this checkbox.

nerrad:
While the current UI is a solution that works (note I haven't actually tested, just commenting on the approach), I do have concerns around whether this is an approach we actually want to ship given it's a bit contrary to the Gutenberg paradigm where any content changes are made inline with the block (as opposed to in inspector controls). 

Terms & Conditions feels like something that should ideally be a product element block that can be inserted as an inner block for the Checkout block. Yet, it's not something the Checkout block really supports yet (and I'm weighing whether we should do a spike to see if we _can_ support that). 

Given that, I'm wary about prematurely introducing a pattern here that we don't want to be using long term. So I think we should pause until we can identify a way to insert blocks in the Cart and Checkout blocks.  Thomas this has been awesome work (that we'll likely still be able to use parts of when we resume) and at a minimum helped surface some more needs we may have for a suitable ux here.

Thoughts?

opr:
> While the current UI is a solution that works (note I haven't actually tested, just commenting on the approach), I do have concerns around whether this is an approach we actually want to ship given it's a bit contrary to the Gutenberg paradigm where any content changes are made inline with the block (as opposed to in inspector controls).
> 
> Terms & Conditions feels like something that should ideally be a product element block that can be inserted as an inner block for the Checkout block. Yet, it's not something the Checkout block really supports yet (and I'm weighing whether we should do a spike to see if we _can_ support that).
> 
> Given that, I'm wary about prematurely introducing a pattern here that we don't want to be using long term. So I think we should pause until we can identify a way to insert blocks in the Cart and Checkout blocks. Thomas this has been awesome work (that we'll likely still be able to use parts of when we resume) and at a minimum helped surface some more needs we may have for a suitable ux here.
> 
> Thoughts?

Darren and I chatted about this last week, and I agree that using the inspector controls is not the ideal solution, and having the text for the terms and conditions field editable inline, allowing the user to type freely and add links and formatting to text would definitely be the better solution. 

With this in mind I'll close this issue, and we can refer back to it when/if we do a spike to see if we can use inner blocks in the checkout block.

PR #4004, opened 2021-03-26
Add initial e2e tests to Checkout block frontend

Aljullu:
This PR adds some initial tests to the Checkout block frontend. They are based on the [Checkout shortcode tests included in WC Core](https://github.com/woocommerce/woocommerce/blob/439d26be827e97b51f107e8454495b2a083153fe/tests/e2e/core-tests/specs/shopper/front-end-checkout.test.js), but with some modifications to target the blocks.

### DIFF
--- a/tests/e2e/specs/frontend/checkout.test.js
+++ b/tests/e2e/specs/frontend/checkout.test.js
@@ -14,6 +14,7 @@ import {
  */
 import {
 	getNormalPagePermalink,
+	scrollTo,
 	shopper,
 	visitPostOfType,
 } from '../../../utils';
@@ -133,13 +134,7 @@ describe( `${ block.name } Block (frontend)`, () => {
 			twoProductPrice
 		);
 
-		await page.evaluate( () => {
-			// Disable smooth scrolling so it scrolls instantly.
-			document.querySelector( 'html' ).style.scrollBehavior = 'auto';
-			document
-				.querySelector( '.wc-block-components-radio-control__input' )
-				.scrollIntoView( { block: 'center', inline: 'center' } );
-		} );
+		await scrollTo( '.wc-block-components-radio-control__input' );
 
 		await expect( page ).toClick(
 			'.wc-block-components-payment-method-label',
--- a/tests/e2e/specs/frontend/checkout.test.js
+++ b/tests/e2e/specs/frontend/checkout.test.js
@@ -132,6 +132,14 @@ describe( `${ block.name } Block (frontend)`, () => {
 			twoProductPrice
 		);
 
+		await page.evaluate( () => {
+			// Disable smooth scrolling so it scrolls instantly.
+			document.querySelector( 'html' ).style.scrollBehavior = 'auto';
+			document
+				.querySelector( '.wc-block-components-radio-control__input' )
+				.scrollIntoView( { block: 'center', inline: 'center' } );
+		} );
+
 		await expect( page ).toClick(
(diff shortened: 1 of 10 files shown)


tests/e2e/specs/frontend/checkout.test.js
```diff
@@ -0,0 +1,162 @@
...
+	visitPostOfType,
+} from '../../../utils';
+
+const block = {
+	name: 'Checkout',
+};
+
+const simpleProductName = 'Woo Single #1';
+const singleProductPrice = '$21.99';
+const twoProductPrice = '$43.98';
```
ralucaStan: I recommend, for more clarity, testing for `2 * singleProductPrice` instead of having this variable

tests/e2e/specs/frontend/checkout.test.js
```diff
@@ -0,0 +1,162 @@
...
+
+if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( `skipping ${ block.name } tests`, () => {} );
```
ralucaStan, line 33: is this equivalent to 
```

if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 ) {
	test.only( `skipping ${ block.name } tests`, () => {} );
}

is this something that can be removed?
```

tests/e2e/specs/frontend/checkout.test.js
```diff
@@ -0,0 +1,162 @@
...
+		await shopper.goToCheckoutBlock();
+		await shopper.productIsInCheckoutBlock(
+			simpleProductName,
+			`2`,
+			twoProductPrice
+		);
+
+		await page.evaluate( () => {
+			// Disable smooth scrolling so it scrolls instantly.
+			document.querySelector( 'html' ).style.scrollBehavior = 'auto';
```
ralucaStan: is this something we should save as an utility?

ralucaStan (approved):
Looks good, congrats on the effort. I approved as I have just added some small comments.

tests/e2e/specs/frontend/checkout.test.js
Aljullu: Good idea! Done in f633a4a2d871a38ba87d195ea6d6a05b7d403d2d.

tests/e2e/specs/frontend/checkout.test.js
Aljullu: Done in 14c21452fcc11922efc41c33f6c1dedd086f2748. :slightly_smiling_face:

tests/e2e/specs/frontend/checkout.test.js
Aljullu: We have the same in [Cart](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/e2e/specs/backend/cart.test.js#L26-L28) & [Checkout](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/e2e/specs/backend/checkout.test.js#L25-L27) backend tests and [Cart frontend tests](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/e2e/specs/frontend/cart.test.js#L25-L27). This is required so in tests that simulate the core package (not the feature plugin) those tests are not run given that Cart and Checkout blocks are not available, so I don't think this can be removed.

PR #4005, opened 2021-03-28
Strip tags from exceptions that are converted from notices

### LINKED ISSUE #3977: Raw HTML sometimes appears in checkout error message when shortcode cart shipping calculator is used
## Describe the bug
When you are using the shortcode cart's shipping calculator, and don't enter a ZIP code, when you try to check out in the checkout block the API will return this error:
image

Thanks to @waclawjacek for finding this!

## To reproduce

1. Enable WooCommerce Shipping and Tax
1. Go to the shortcode cart.
2. Use the shipping calculator, select United States as your country, but leave ZIP code empty.
2. Proceed to checkout.
3. Fill in all fields, including ZIP code.
4. See error.

## Expected behaviour
The checkout should work when the ZIP code is entered on the checkout block.

## Screen capture
View reproduction here: https://d.pr/i/RrLNVa

* Core version: 5.7
* WooCommerce version: 5.1.0

opr:
When an extension adds a notice that contains HTML using `wc_add_notice` and then it is converted using the `NoticeHandler::convert_notices_to_exceptions` method, then the resulting exception contains HTML which is displayed incorrectly on the front-end. The changes in this PR will strip the tags from notices before returning.

There is also a simple PHP unit test included with this PR.

Fixes #3977 - but it does _not_ fix the underlying issue causing the error to display in the first place.

#### Before
image

#### After
image

**You will need WooCommerce Shipping and Tax set up and linked to your WPCOM account. If you don't have this, then you can test on _my_ local site using my ngrok tunnel. Just message me when you're ready to test.**

1. Enable WooCommerce Shipping and Tax
2. Go to the shortcode cart.
3. Use the shipping calculator, select United States as your country, but leave ZIP code empty.
4. Proceed to checkout.
5. Fill in all fields, including ZIP code.
6. See error.

### Changelog

> Prevent error messages returned by the API from displaying raw HTML.

### DIFF
--- a/tests/php/StoreApi/Utilities/NoticeHandler.php
+++ b/tests/php/StoreApi/Utilities/NoticeHandler.php
@@ -7,21 +7,16 @@ namespace Automattic\WooCommerce\Blocks\Tests\StoreApi\Utilities;
 
 use Automattic\WooCommerce\Blocks\StoreApi\Routes\RouteException;
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\CartController;
+use Automattic\WooCommerce\Blocks\StoreApi\Utilities\NoticeHandler;
 use PHPUnit\Framework\TestCase;
 use \WC_Helper_Product as ProductHelper;
 
 class NoticeHandlerTests extends TestCase {
 
-	public function test_api_does_not_return_html() {
-		// We expect validate_cart_items to throw
+	public function test_convert_notices_to_exceptions() {
 		$this->expectException(RouteException::class);
 		$this->expectExceptionMessage('This is an error message with Some HTML in it.');
-		$class = new CartController();
-		add_action( 'woocommerce_check_cart_items', function() {
-			wc_add_notice( '<strong>This is an error message with <a href="#">Some HTML in it</a>.', 'error' );
-		} );
-		$product = ProductHelper::create_simple_product();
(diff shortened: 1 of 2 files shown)


tests/php/StoreApi/Utilities/NoticeHandler.php
```diff
@@ -0,0 +1,22 @@
...
+use Automattic\WooCommerce\Blocks\StoreApi\Utilities\CartController;
+use Automattic\WooCommerce\Blocks\StoreApi\Utilities\NoticeHandler;
+use PHPUnit\Framework\TestCase;
+use \WC_Helper_Product as ProductHelper;
+
+class NoticeHandlerTests extends TestCase {
+
+	public function test_convert_notices_to_exceptions() {
+		$this->expectException(RouteException::class);
+		$this->expectExceptionMessage('This is an error message with Some HTML in it.');
```
Aljullu: The formatting looks off in these two lines:
```suggestion
		$this->expectException( RouteException::class );
		$this->expectExceptionMessage( 'This is an error message with Some HTML in it.' );
```

Aljullu (approved):
I tested this PR hardcoding a notice just before `NoticeHandler::convert_notices_to_exceptions()` is called and I could reproduce the issue in `trunk` and verify tags are stripped in this branch. :ship:

PR #4010, opened 2021-03-29
Remove deprecated shortcode save handling from SSR Blocks

mikejolley:
This PR removes the `deprecated` save method from SSR Blocks which converts the legacy shortcode rendering to the newer dynamic rendering. 

[This was introduced in 2019](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/558) and appears to be used because before dynamic blocks were implemented, blocks acted as a wrapper for shortcodes.

This will invalidate any blocks created and unsaved since 2019, however, given these have been deprecated a few years it won't have much impact. I tested to see what happened to these blocks. The block post content is:

```
<!-- wp:woocommerce/product-new -->
<div class="wp-block-woocommerce-product-new">[products limit="3" columns="3" orderby="date" order="DESC"]</div>
<!-- /wp:woocommerce/product-new -->
```

So whilst the block does invalidate, the frontend content is unaffected.

Smoke test affected blocks still insert and redner

1. handpicked
2. new products
3. best sellers
4. product cagegory
5. on sale
6. top rated
7. by attribute

### Changelog

> Removed legacy handling for SSR blocks that rendered shortcodes.


ralucaStan (approved):
👍

PR #4013, opened 2021-03-30
Avoid get_billing and get_shipping method when updating the order

### LINKED ISSUE #3975: Address data not sent to API on checkout block after using shortcode cart's address shipping calculator
## Describe the bug
When using a combination of the shortcode cart, and the checkout block, if you use the shipping calculator on the shortcode block, but update the address without making any changes, then your pre-filled address data will not be sent to the API when checking out.

Thanks to @waclawjacek for originally finding this.

## To reproduce
This can be a little tricky to reproduce, you may have to complete an order first to get the address to save. See this loom for a reproduction.
https://www.loom.com/share/5e97fe53b2e74c76a47ea96fe6274294

1. Go to the shortcode cart, ensure the shipping calculator has a saved address in.
2. If it does not, then you may need to complete an order.
2. Open the shipping calculator and save the address without making any changes.
2. Proceed to checkout (you should be using the checkout _block_) and try to complete the order without changing any address fields. (You may need to fill some required fields)
3. Observe the error: `Sorry, we do not ship orders to the provided country ()`

## Expected behavior
Given the address is valid when arriving on the checkout block, the order should complete when the "Place order" button is pressed.

* Core version: 5.7
* WooCommerce version: 5.1.0

mikejolley:
There is a core bug where the `get_shipping` and `get_billing` methods only return **changes**. This means when we go to update the order, some fields are missing which leads to the issue described in #3975

See https://github.com/woocommerce/woocommerce/pull/29538

This PR works around the issue by avoiding these methods and just using individual setters from the customer object.

Fixes #3975

@nerrad @ralucaStan this should be merged before 4.8 IMO.

See instructions in #3975.

Smoke test checkout taking care that addresses are saved correctly to the order.

### Changelog

> Fix customer address country saving to orders in certain circumstances.

### DIFF
--- a/src/StoreApi/Utilities/OrderController.php
+++ b/src/StoreApi/Utilities/OrderController.php
@@ -453,28 +453,29 @@ class OrderController {
 	 * @param \WC_Order $order The order object to update.
 	 */
 	protected function update_addresses_from_cart( \WC_Order $order ) {
-		$customer_billing = wc()->customer->get_billing();
-		$customer_billing = array_combine(
-			array_map(
-				function( $key ) {
-					return 'billing_' . $key;
-				},
-				array_keys( $customer_billing )
-			),
-			$customer_billing
+		$order->set_props(
+			[
+				'billing_first_name'  => wc()->customer->get_billing_first_name(),
+				'billing_last_name'   => wc()->customer->get_billing_last_name(),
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Great investigation with this issue @mikejolley! I tested and I was no longer able to reproduce the issue, code changes look good as well.

PR #4020, opened 2021-03-31
Enable Google Analytics Integration 

### LINKED ISSUE #3175: Integrate with Google Analytics
Currently Google Analytics (WooCommerce core option) integrates with the Cart and Checkout flow to trigger events for GA. We need to ensure that purchases through the cart and checkout block flow triggers the appropriate events (and any related extensibility points are available for those trigger points).

As much as possible split up the work into independent chunks, for now use this as a tracking issue.

mikejolley:
This PR removes the experimental flag from the GA extension code. If using the GA extension, block events will also be tracked.

Closes #3175 

### Changelog

> Added compatibility with the Google Analytics Integration. Block events, including cart and checkout, can now be tracked.

### Dev note

Blocks are now compatible with the Google Analytics Integration: https://woocommerce.com/products/woocommerce-google-analytics/ If using Google Analytics with GTAG support (and a `G-` prefixed site ID), block events will also be tracked. This includes:

- Product searches in the Product Search Block
- Product views in the product grid blocks and All Products Block
- Add to cart events
- Cart item changes
- Checkout progress events

### DIFF
--- a/docs/blocks/feature-flags-and-experimental-interfaces.md
+++ b/docs/blocks/feature-flags-and-experimental-interfaces.md
@@ -1,13 +1,13 @@
 We have feature gating system setup in our plugin that defines what is accessible to the public and what is not, it has three phases:
 
-*   **Core flag `WOOCOMMERCE_BLOCKS_PHASE=1`:** anything that is not hidden behind a flag, falls under this category, and it contains all of the code that runs on WooCommerce Core plugin.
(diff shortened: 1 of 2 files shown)


nerrad (approved):
🚢

nerrad:
I just noticed when looking through the files as a part of this feature unveiling that it doesn't appear any of the new experimental hook slugs are [in this doc](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/blocks/feature-flags-and-experimental-interfaces.md). It'll be especially important to include them in this doc given the prefix differs from what the doc references (so followup searches in the future won't surface them).

Up to you whether you do that in this PR (preferable) or in a followup (if so asap), but I think we should get that doc updated?

PR #4025, opened 2021-04-01
Fix Circular Dependencies During Builds

### LINKED ISSUE #3723: Audit and refactor cyclical dependencies.

mikejolley:
This is a large PR that fixes circular dependency(CD) issues during our build process, and the CD admin notices that were added to track the issue.

This fixes #3723 🥳 

I tried to split each fix into a separate PR. The majority of the changes just move things around and then import using the path instead of the alias. 

The "checkout" package was a little more complicated ([see this commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4025/commits/4ddf221005b9a65f7f42ddc9fe43ce305d194619)) and involved passing hooks as props instead of importing from the `base-context` package. 

The implication of this is that we'll need a small modification to the subscriptions integration (@senadir) to make it use the prop instead of importing. This affects `SubscriptionsRecurringPackages` which must accept `useSelectShippingRate` prop and pass it to the shipping packages component.

I have tested this works and raised a PR here: https://github.com/woocommerce/woocommerce-subscriptions/pull/4059

**Testing this PR**

This PR moves things around meaning there is no single thing to test. This needs smoke testing to confirm:

- No build errors
- Product blocks still function
- Cart updates still function
- Checkout can be completed
- Payment methods still work during checkout
- Express payment methods still work during checkout
- With https://github.com/woocommerce/woocommerce-subscriptions/pull/4059, subscriptions still work during cart and checkout, including shipping rates
 
Because this potentially can grow stale fast, cc'ing @nerrad and @senadir here who may be interested in reviewing. Please let me know if I have missing anything or something needs more explanation.


bin/webpack-configs.js
```diff
@@ -12,6 +12,7 @@ const WebpackRTLPlugin = require( 'webpack-rtl-plugin' );
 const chalk = require( 'chalk' );
 const { kebabCase } = require( 'lodash' );
 const CreateFileWebpack = require( 'create-file-webpack' );
+const CircularDependencyPlugin = require( 'circular-dependency-plugin' );
```
senadir, line 14: Should probably revert this or keep it under a flag, see https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3692

mikejolley: I asked @nerrad about this. I didn't spot any obvious slow down—did you?

nerrad: If there's no slowdown it's fine to leave it.

One thing I just thought of however, does this plugin consider `.ts` and `.tsx` files in its processing? Also, the `exclude` argument seems odd to me, why are we excluding files named `a.js` (I'm assuming this was just copied from their docs).

mikejolley: This runs on `npm start` only now. Other builds are not affected.

assets/js/atomic/blocks/product-elements/button/block.js
```diff
@@ -5,7 +5,10 @@ import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import { __, _n, sprintf } from '@wordpress/i18n';
 import { useEffect, useRef } from '@wordpress/element';
-import { useStoreAddToCart, useStoreEvents } from '@woocommerce/base-hooks';
+import {
+	useStoreEvents,
+	useStoreAddToCart,
+} from '@woocommerce/base-context/hooks';
```
nerrad, line 11: It's interesting (and odd) to me that this increased the size of the build for this atomic block by 33% (an additional 506 bytes)! At this point this is just an observation but I'm curious why?

assets/js/settings/shared/index.js
```diff
@@ -9,6 +9,8 @@ import compareVersions from 'compare-versions';
 import { getSetting } from './get-setting';
 
 export * from './default-constants';
+export { default as defaultAddressFields } from './default-address-fields';
+export { default as countryAddressFields } from './country-address-fields';
```
nerrad: While moving `defaultAddressFields` and `countryAddressFields` to `@woocommerce/settings` makes sense, it does increase the size of the `wc.wcSettings` build significantly (a 45% increase or 1.08kb). This in turn has high impact on projects using `@woocommerce/settings` that _might not need_ these two exports (it's a big bump for their dependency weight). 

This is one increase that I think we need to avoid. I wonder if this is something we could do some WebPack lazy loading on to reduce the footprint? If not, can we move these exports back where they were and defer this change to a followup where we can work on a solution that limits the increase to the footprint of `@woocommerce/settings`?

assets/js/base/components/cart-checkout/shipping-rates-control/index.js
```diff
@@ -84,6 +88,22 @@ const ShippingRatesControl = ( {
...
+	const slotFillProps = {
+		className,
+		collapsible,
+		noResultsMessage,
+		renderOption,
+		extensions,
+		cart,
+		components: {
+			ShippingRatesControlPackage,
+		},
```
nerrad, line 103: Ya I like this change, it exposes a specific component for extensions to use in their fill and is something we have control over.

The only downside is the extensions need to understand what props to pass to the component, but that's not really a change from what was present already anyways 😄. 

I do wonder, however, if we can improve on this API a bit where we expose a local state setter that extensions would just use to set the details of their packages and then we'd use the state to assemble the UI with components (including ShippingRatesControlPackage). Some advantages:

- we wouldn't have to expose any components, and extension developers wouldn't have to worry about prop usage for exposed components.
- we should be able to validate the provided data from the extensions using the exposed setter.

The above doesn't need to be addressed for this PR but I do think we might want to consider this as part of a followup before removing the experimental prefix from the interface (by "consider" I mean weigh whether there is any merit to this idea, or whether what we are doing now is okay).

assets/js/base/components/cart-checkout/shipping-rates-control/packages.js
```diff
@@ -0,0 +1,55 @@
...
+ * @param {React.ReactElement} props.noResultsMessage Rendered when there are no rates in a package.
+ * @param {boolean} props.collapsible If the package should be rendered as a
+ * collapsible panel.
+ * @param {boolean} props.collapse If the panel should be collapsed by default,
+ * only works if collapsible is true.
+ * @param {boolean} props.showItems If we should items below the package name.
+ * @param {Function} [props.renderOption] Function to render a shipping rate.
+ * @return {React.ReactElement|null} Rendered components.
+ */
+const Packages = ( {
```
nerrad: I'm not sure there is need to move this component out into it's own file since it's not really being imported into multiple places (and at a minimum could have been considered separately maybe?). Having it directly in context of where it is consumed could be useful. If/when we move things like this into their own files it'd probably be good to create the file in `.ts` too :). With that said, this is a non-blocking comment.

assets/js/base/context/hooks/cart/test/use-store-cart.js
```diff
@@ -3,16 +3,16 @@
...
 import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
 
 /**
  * Internal dependencies
  */
 import { defaultCartData, useStoreCart } from '../use-store-cart';
+import { useEditorContext } from '../../../providers/editor-context';
 
-jest.mock( '@woocommerce/base-context', () => ( {
+jest.mock( '../../../providers/editor-context', () => ( {
```
nerrad, line 15: nice change!

assets/js/base/context/providers/store-notices/components/store-notices-container.js
```diff
@@ -24,8 +23,7 @@ const getWooClassName = ( { status = 'default' } ) => {
 	return '';
 };
 
-const StoreNoticesContainer = ( { className, notices } ) => {
-	const { removeNotice } = useStoreNoticesContext();
+const StoreNoticesContainer = ( { className, notices, removeNotice } ) => {
```
nerrad, line 26: 🤔 was the primary reason for this change to remove the need for the context dependency? If so 👏 - with this change it does mean this component doesn't _have_ to live in the context folder tree but it probably still makes sense for it to be here for now.

assets/js/icons/icon/index.tsx
```diff
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { cloneElement, isValidElement } from 'wordpress-element';
+import { cloneElement, isValidElement } from '@wordpress/element';
```
nerrad, line 4: Huh, odd - why did we ever directly import this in the first place?

assets/js/icons/library/arrow-back.js
```diff
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { SVG } from 'wordpress-components';
+import { SVG } from '@wordpress/components';
```
nerrad: If the icons are used on the frontend, doesn't this mean _all_ of `@wordpress/components` will get loaded which means an increase in overall weight on the assets loaded (this comment applies to other files where this change was made too).

bin/webpack-helpers.js
```diff
@@ -124,6 +125,7 @@ const requestToHandle = ( request ) => {
 module.exports = {
 	NODE_ENV,
 	FORCE_MAP,
+	CIRCULAR_DEPS,
```
nerrad: I appreciate the brevity, but I wonder if we could pick a name that makes it a bit clearer what this does - something like `CIRCULAR_DEPS_CHECK`?

tsconfig.base.json
```diff
@@ -1,55 +1,61 @@
...
+			"@woocommerce/block-hocs": [ "assets/js/hocs" ],
+			"@woocommerce/blocks-registry": [ "assets/js/blocks-registry" ],
+			"@woocommerce/blocks-checkout": [ "packages/checkout" ],
+			"@woocommerce/price-format": [ "packages/prices" ],
+			"@woocommerce/block-settings": [ "assets/js/settings/blocks" ],
+			"@woocommerce/e2e-tests": [ "node_modules/woocommerce/tests/e2e" ],
+			"@woocommerce/icons": [ "assets/js/icons" ],
+			"@woocommerce/resource-previews": [ "assets/js/previews" ],
+			"@woocommerce/knobs": [ "storybook/knobs" ],
+			"@woocommerce/price-format": [ "packages/prices" ],
```
nerrad: This key is already in the config.

nerrad (commented):
Whew! Some great work here Mike! A few things I really like about this PR:

- We have a tool enabled that will help us catch cyclical/circular dependencies in the future.
- It's more clear now with certain aliased imports what indirect dependencies might exist. For instance importing hooks from `@woocommerce/base-context/hooks` makes it clear this is a hook that has a dependency on context whereas `@woocommerce/base-hooks` is not dependent on context. This will also help when it comes to determining where something should live.
- You fixed the circular dep issue 🎉 🌮 

Besides what I've left as inline feedback, a couple other observations:

- I know the size of this PR couldn't be avoided - however, there are some changes I _think_ weren't necessary for this PR (as an example all the `wordpress-components` > `@wordpress/components` change for icons. I know it seems like a quick win to just make those changes when you see them but it really does make an already big PR bigger for the reviewer to handle. There's also potential side-effects that are hard to measure in an isolated way because of all the other changes in the PR.
- With the changes in d34e790 passing through components instead of hooks, can you update the description for the PR to reflect them (so future selves stumbling on the PR won't be confused 😄 ).

## Testing

I smoke tested critical flows while both logged in and logged out and also tested various Stripe testing cards to verify error handling etc is still in place. **I did not test the Subscriptions integration.**

All the testing went really well with nothing obvious standing out except that I get this error if I submit checkout with a missing required address field:

CleanShot 2021-04-07 at 16 38 57@2x

I'm not sure if that's introduced in this PR or not, but just pointing it out.

I think most (if not all - I already forget lol) of my feedback is non-blocking but I still am holding off on the ✅ pending your responses.

mikejolley:
> I know the size of this PR couldn't be avoided - however, there are some changes I think weren't necessary for this PR (as an example all the wordpress-components > @wordpress/components change for icons. I know it seems like a quick win to just make those changes when you see them but it really does make an already big PR bigger for the reviewer to handle. There's also potential side-effects that are hard to measure in an isolated way because of all the other changes in the PR.

@nerrad if you run the CD plugin on trunk you'll see that these changes were in response to a CD issue. As was everything in this PR. Importing from `wordpress-components` caused a CD.

assets/js/icons/library/arrow-back.js
mikejolley: Possibly, but this usage in the icon library did cause a CD issue. I will look it up tomorrow.

assets/js/icons/icon/index.tsx
mikejolley: No idea. I checked the changelogs to ensure it was available on supported versions.

assets/js/base/context/providers/store-notices/components/store-notices-container.js
mikejolley: yup, wasn't necessary as it's easy enough to pass down.

assets/js/base/components/cart-checkout/shipping-rates-control/packages.js
mikejolley: There was a reason I did this, I can't quite remember right now but I think it was because this component is exported as well as the component in the index file, and having both default exports from one file was difficult to follow. I can revert and see what breaks if needed. I don't want to convert to TS in this PR.

assets/js/base/components/cart-checkout/shipping-rates-control/index.js
mikejolley: Since this doesn't vary from the current implementation much (we just switch from import to prop) I don't want to tackle this here, but I am cc'ing @senadir as there may be other reasons why it was done this way initially.

assets/js/settings/shared/index.js
mikejolley: @nerrad I can certainly try. In case it wasn't clear, these had to be moved because they were consumed from multiple places which caused a CD issue. Settings is a shared location.

tsconfig.base.json
mikejolley: Was from a rebase. 👍🏻

assets/js/atomic/blocks/product-elements/button/block.js
mikejolley: Could it be that the /hooks index file exports more things that previous, since we moved things from base hooks. Thats the only change I can think of that would impact this.

opr (approved):
I was able to test everything except express payment methods (they also don't work for me in trunk at the moment so not an issue introduced in this PR)

The following were tested:
- Add to cart
- Modify quantities and remove items from cart.
- Change shipping method on cart page.
- Check out with saved credit card (Stripe).
- Check out using new credit card (Stripe).
- Check out using COD & bank transfer.
- Change a subscription item's shipping options during checkout.
- Check out with subscription products.
- Try to check out with invalid data in the form, error message is displayed

From a code point of view, the only comments I have to add is that the work I'm doing on postcode validation and I think @ralucaStan 's work on renaming `onCheckoutBeforeProcessing` will have some conflicts. (I'm actually pretty sure that everyone else's branches will have conflicts too).

With this in mind I'd probably prefer this to be merged first, then for us to resolve the conflicts on _our_ branches rather than the other way around... do you have any preference in the order that this is done?

mikejolley:
> With this in mind I'd probably prefer this to be merged first, then for us to resolve the conflicts on our branches rather than the other way around... do you have any preference in the order that this is done?

I already have one to resolve. I think it would be easier to do this one first because as I'm moving files there is a risk I may miss changes during rebase since I'll be unaware of specific changes in files :)

assets/js/base/components/cart-checkout/shipping-rates-control/packages.js
mikejolley: Done in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4025/commits/e362f1c542aaced890f954780aae9d270e05fada

mikejolley:
@nerrad that bug appears on trunk too so logged here https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4047

opr:
> > With this in mind I'd probably prefer this to be merged first, then for us to resolve the conflicts on our branches rather than the other way around... do you have any preference in the order that this is done?
> 
> I already have one to resolve. I think it would be easier to do this one first because as I'm moving files there is a risk I may miss changes during rebase since I'll be unaware of specific changes in files :)

Yup, sorry when I said _this_ I meant the PR we're commenting on now and let the authors of any other affected PRs rebase 👍🏻

assets/js/icons/library/arrow-back.js
mikejolley: Fixed by removing module replacements https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4025/commits/a9471a5b8234a67593c8813078d1bd6982b574c3

assets/js/atomic/blocks/product-elements/button/block.js
nerrad: Hmm, perhaps, but then that could mean tree shaking isn't being applied. That would be a good thing to try and determine in a followup (might be even more savings gains there!).

assets/js/settings/shared/index.js
nerrad: Yup understood, but given the use of this package by other teams I think we should try to avoid this increase.

assets/js/base/components/cart-checkout/shipping-rates-control/index.js
nerrad: Yup I'm fine with that.

assets/js/settings/shared/index.js
mikejolley: The core locale was only used in one place so I moved it there https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4025/commits/2167f81d966d6ec0bc55f54d5fd4e8459f52d11f

mikejolley:
I tested chrome pay is working.

nerrad (approved):
Thanks for addressing all the feedback Mike! Let's get this 🚢

PR #4028, opened 2021-04-01
Fix Featured Product and Featured Category button alignment

### LINKED ISSUE #3788: Featured Category Block: Buttons have .aligncenter class but are left aligned
I tested this in 2019 theme using the defaults on the Featured Category Block:

Screenshot 2021-02-04 at 13 24 23

I can see the button has the correct aligncenter class, but the parent element has width:100% which forces left alignment. This only appears to affect the frontend of the store.

Aljullu:
Fixes #3788.

We were relying on some styles from Gutenberg which are no longer present in recent versions.

### Screenshots

_Before:_
imatge

_After:_
imatge

1. Add a Featured Product and Featured Category blocks.
2. Without making any change on them, preview them in the frontend and verify the button is centered.
3. Change the alignment of the button to the left, right, back to the center, etc. and verify it works.

### Changelog

> Fix button alignment in Featured Product and Featured Category blocks.

### DIFF
--- a/assets/js/blocks/featured-category/style.scss
+++ b/assets/js/blocks/featured-category/style.scss
@@ -82,6 +82,10 @@
 		}
 	}
 
+	.wp-block-button.aligncenter {
+		text-align: center;
+	}
+
 	&.has-background-dim::before {
 		content: "";
 		position: absolute;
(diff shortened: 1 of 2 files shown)


ralucaStan (approved):
👍

PR #4032, opened 2021-04-02
Remove useCheckoutRedirectUrl()

Aljullu:
I can't find any usage of this hook. Everything seems to work fine after removing it (tested bank transfer, Stripe + 3DS, saved cards and express payment methods).

Smoke test the checkout flow with different payment methods and verify everything keeps working as usual.


mikejolley (approved):
👍🏻 Confirmed as unused. Good catch.

PR #4038, opened 2021-04-06
Prevent unwanted parts of address being displayed

### LINKED ISSUE #3979: State (address field) is sometimes kept in the address when switching to an address that doesn't have states
## Describe the bug
This was discovered by @waclawjacek - thanks!

If you use the shipping calculator on the shortcode cart and enter a US address with State, then successfully check out using the blocks cart, if you change your address using the shortcode cart shipping calculator to a country without states (e.g. Poland) then the US state remains in the address if you switch to the cart _block_.

I appreciate this is a contrived case, but it could reveal other underlying issues of data not being cleared out properly.

## To reproduce

1. Add an item to the cart.
2. Go to the shortcode cart - use the shipping calculator to enter a US address. (`State: California, City: Beverly Hills, ZIP: 90210`)
3. Proceed to checkout and successfully complete. If you get errors, ensure the form is filled properly, and submit again. This is a known issue in #3975 
4. Add another item to the cart and go back to the shortcode cart, change your address on the shipping calculator to a Polish one. (`City: Gdansk, Postcode: 80-000`)
4. Go to the cart _Block_ and see the state `CA` remains.

Loom of reproduction: https://www.loom.com/share/c3b277dda5064598999c35b29576bc20

## Expected behavior
The address shown in the cart block should not contain remnants of any old addresses.

* Core version: 5.7
* WooCommerce version: 5.1.0

opr:
The changes introduced in this PR will cause parts of an address that should be hidden (based on the settings in `countryLocale`) to be set to an empty string in the address object.

The reason we can't simply set the unwanted parts of the address to empty strings when updating it is because in core there is the notion of `changes` to the address object, which when applied only updates fields for which there are new values (see source here: https://github.com/woocommerce/woocommerce/blob/trunk/includes/abstracts/abstract-wc-data.php#L764-L767 - e.g. so when `state` is empty on a new address, but present in the old address, it will not be updated) I'm not sure if this is something that _should_ be changed in core - it appears that the logic used for this is used by several different classes and adding checks around it for parts of an address would decrease performance.

Core has some [similar functionality](https://github.com/woocommerce/woocommerce/blob/trunk/includes/class-wc-countries.php#L563) to display the address in the shortcode cart's shipping calculator.

**Changes**
- Add the `emptyHiddenAddressFields` function
- Add tests for `emptyHiddenAddressFields`.
- Ensure the address is formatted before posting it to the API in the checkout, and before displaying it in the shipping calculator.

Fixes #3979

1. Add an item to the cart.
2. Go to the shortcode cart - use the shipping calculator to enter a US address. (`State: California, City: Beverly Hills, ZIP: 90210`)
3. Proceed to checkout and successfully complete. If you get errors, ensure the form is filled properly, and submit again. This is a known issue in #3975 
4. Add another item to the cart and go back to the shortcode cart, change your address on the shipping calculator to a Polish one (or any other that doesn't use states). (`City: Gdansk, Postcode: 80-000`)
4. Go to the cart _Block_ and see there is no `CA` shown in the address.
5. Check out and inspect the request made to the API. Ensure the posted data does not contain the state, or any other information that should not be posted.

### Changelog

> Prevent parts of old addresses being displayed in the shipping calculator when changing countries.

### DIFF
--- a/assets/js/base/components/cart-checkout/shipping-location/index.js
+++ b/assets/js/base/components/cart-checkout/shipping-location/index.js
@@ -5,7 +5,6 @@ import PropTypes from 'prop-types';
 import { __, sprintf } from '@wordpress/i18n';
 import { getSetting } from '@woocommerce/settings';
 import { decodeEntities } from '@wordpress/html-entities';
-import { emptyHiddenAddressFields } from '@woocommerce/base-utils';
 
 /**
  * Shows a formatted shipping location.
@@ -19,7 +18,6 @@ const ShippingLocation = ( { address } ) => {
 		return null;
 	}
 
-	address = emptyHiddenAddressFields( address );
 	const shippingCountries = getSetting( 'shippingCountries', {} );
 	const shippingStates = getSetting( 'shippingStates', {} );
 	const formattedCountry =
--- a/assets/js/base/components/cart-checkout/shipping-location/index.js
+++ b/assets/js/base/components/cart-checkout/shipping-location/index.js
@@ -5,6 +5,7 @@ import PropTypes from 'prop-types';
 import { __, sprintf } from '@wordpress/i18n';
 import { getSetting } from '@woocommerce/settings';
 import { decodeEntities } from '@wordpress/html-entities';
+import { emptyHiddenAddressFields } from '@woocommerce/base-utils';
 
 /**
  * Shows a formatted shipping location.
@@ -17,6 +18,8 @@ const ShippingLocation = ( { address } ) => {
 	if ( Object.values( address ).length === 0 ) {
 		return null;
 	}
+
+	address = emptyHiddenAddressFields( address );
 	const shippingCountries = getSetting( 'shippingCountries', {} );
 	const shippingStates = getSetting( 'shippingStates', {} );
 	const formattedCountry =
--- a/assets/js/base/utils/address.js
+++ b/assets/js/base/utils/address.js
@@ -1,4 +1,10 @@
 /**
+ * External dependencies
+ */
+import defaultAddressFields from '@woocommerce/base-components/cart-checkout/address-form/default-address-fields';
+import prepareAddressFields from '@woocommerce/base-components/cart-checkout/address-form/prepare-address-fields';
+
+/**
  * pluckAddress takes a full address object and returns relevant fields for calculating
  * shipping, so we can track when one of them change to update rates.
  *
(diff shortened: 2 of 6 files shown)


ralucaStan:
I followed the testing steps and things look good.


I also wanted to test the Checkout block as well, because @mikejolley  [fixed recently](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4013/files) an issue involving it, and the shipping calculator in cart shortcode. 


I found a bug that is only on this branch. On trunk, there aren't problems

The steps are similar to yours but with the mention that the first order is submitted with the checkout shortcode.

- Add an item to the cart.
- Go to the shortcode cart - use the shipping calculator to enter a US address. (State: California, City: Beverly Hills, ZIP: 90210)
- Proceed to **checkout with shortcode** checkout and successfully complete.
- Add another item to the cart and go back to the shortcode cart, change your address on the shipping calculator to a Polish one (or any other that doesn't use states). (City: Gdansk, Postcode: 80-000)
- Go to the cart Block and see there is no CA shown in the address.
- Go to checkout Block and fill in missing fields.
- Submit the form and see an error for the email field, although the input is not invalid.
- Look at the failed POST and see that email is field is missing, so resubmitting the form does not fix the issue

ralucaStan:
I've noticed you are pushing changes to package-lock.json, but no changes to package.json. I think it's safer to remove this file from the PR.

assets/js/base/components/cart-checkout/shipping-location/index.js
```diff
@@ -20,6 +21,9 @@ const ShippingLocation = ( { address } ) => {
 	if ( Object.values( address ).length === 0 ) {
 		return null;
 	}
+
+	address = emptyHiddenAddressFields( address );
+
```
ralucaStan: I am wondering if we should do the cleaning here. I'm thinking this is a presentation component that should just render the address it receives so it should just receive a cleaned address. What do you think?

```<ShippingLocation address={ emptyHiddenAddressFields(shippingAddress)} />```

**Comment by opr** [comment]
> I've noticed you are pushing changes to package-lock.json, but no changes to package.json. I think it's safer to remove this file from the PR.

Thanks, I included this because it was causing ci to fail (for some reason) but now it's working OK after removing it.

File: assets/js/base/utils/address.js [hunk 2]
```diff
@@ -21,3 +27,20 @@ export const pluckAddress = ( {
...
+ *
+ * @param {Object} address The address to empty fields from.
+ * @return {Object} The address with hidden fields values removed.
+ */
+export const emptyHiddenAddressFields = ( address ) => {
+	const fields = Object.keys( defaultAddressFields );
+	const addressFields = prepareAddressFields( fields, {}, address.country );
+	addressFields.forEach( ( field ) => {
+		if ( field.hidden ) {
+			address[ field.key ] = '';
```
ralucaStan: Here you are actually modifying the original address object passed to this function because you are doing operations on the parameter.

The function should return a new object, so you can actually rewrite the function as bellow:

```js
export const emptyHiddenAddressFields = ( address ) => {
	const fields = Object.keys( defaultAddressFields );
	const addressFields = prepareAddressFields( fields, {}, address.country );

	return addressFields.reduce( ( cleanAddress, field ) => {
		cleanAddress[ field.key ] = field.hidden ? '' : address[ field.key ];
		return cleanAddress;
	}, {} );
};
```

ralucaStan (commented):
Nice catch with this PR. I have added some comments. Let me know what you think.

assets/js/base/components/cart-checkout/shipping-location/index.js
opr: Thanks Raluca, I have moved this to the `useStoreCart` hook.

assets/js/base/utils/address.js
opr: As discussed in Slack we have implemented a modified version of this. Thanks for catching it!

ralucaStan (approved):
👍 Nice work on fixing this annoying bug

PR #4039, opened 2021-04-06
Rename onCheckoutBeforeProcessing to onCheckoutValidationBeforeProcessing

### LINKED ISSUE #3984: Rename `onCheckoutBeforeProcessing` to `onCheckoutValidationBeforeProcessing`
This idea came up as I considered the problem outlined in #3980 when I realized the `onCheckoutBeforeProcessing` is tightly coupled to the validation when the checkout has been submitted (as a quick way to alert the user there is something missing/wrong before submitting to the server).

Given its use-case, I suggested that it would make sense to change the name of this event to something like `onCheckoutValidationBeforeProcessing` to more clearly articulate its purpose and behaviour.

For implementation, we will need to make sure we keep back compat for current implementation of this hook outside of our own code (and throw a deprecation notice for when `onCheckoutBeforeProcessing` is called instead of the new function. Not sure we'd ever remove the deprecated event (or at least far in the future) but at least we can stop promoting it (in docs) and elsewhere it's outlined.

ralucaStan:
This commit also adds a deprecation warning when using onCheckoutBeforeProcessing.
See the issue below for details.

Fixes #3984

The renaming part could be tested by smoke testing the Checkout Block:

1. Add something to checkout
2. Submit without a required field
3. Observe validation happens
4. Fix validation and place order

To see the warning you have to call yourself the old `onCheckoutBeforeProcessing`. What I did:

1. I added this component to [Order Summary](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/components/cart-checkout/order-summary/index.js) and included it as a child.
2. I went to the Checkout block and tried to submit it without some required fields.
3. I checked the console for the deprecation warning
4. I then fixed the warnings and  successfully submitted the order

```js
const Component = () => {
	const { onCheckoutBeforeProcessing } = useCheckoutContext();
	useEffect( () => {
		const unsubscribe = onCheckoutBeforeProcessing( () => {
			console.log( 'My custom validation callback' );
			return true;
		} );
		return unsubscribe;
	}, [ onCheckoutBeforeProcessing ] );
	return null;
};
```

One thing that could also help with testing is checking the registered observers by adding `console.log(observersByType)` in [emitters.ts](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/shared/event-emit/emitters.ts#L26). This is logged after submitting with missing inputs.

### Changelog

> Rename onCheckoutBeforeProcessing to onCheckoutValidationBeforeProcessing.

### Dev note

> `onCheckoutBeforeProcessing` has been renamed to `onCheckoutValidationBeforeProcessing` to more clearly articulate its coupling to the validation happening when the Checkout form is submitted. `onCheckoutBeforeProcessing` will be deprecated in the future.

### DIFF
--- a/assets/js/base/context/providers/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/providers/cart-checkout/checkout-state/index.js
@@ -124,14 +124,19 @@ export const CheckoutStateProvider = ( {
 	 * @deprecated use onCheckoutValidationBeforeProcessing instead
 	 */
 	const onCheckoutBeforeProcessing = useMemo( () => {
-		deprecated( 'onCheckoutBeforeProcessing', {
-			alternative: 'onCheckoutValidationBeforeProcessing',
-			plugin: 'WooCommerce Blocks',
-		} );
-
-		return emitterObservers( observerDispatch )
+		const callback = emitterObservers( observerDispatch )
 			.onCheckoutValidationBeforeProcessing;
+
+		return function ( ...args ) {
+			deprecated( 'onCheckoutBeforeProcessing', {
+				alternative: 'onCheckoutValidationBeforeProcessing',
+				plugin: 'WooCommerce Blocks',
+			} );
+
+			return callback( ...args );
+		};
 	}, [ observerDispatch ] );
+
 	const onCheckoutValidationBeforeProcessing = useMemo(
 		() =>
 			emitterObservers( observerDispatch )
--- a/assets/js/base/context/providers/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/providers/cart-checkout/checkout-state/index.js
@@ -123,7 +123,7 @@ export const CheckoutStateProvider = ( {
 	/**
 	 * @deprecated use onCheckoutValidationBeforeProcessing instead
 	 *
-	 * To prevent the deprecation message to be shown at render time
+	 * To prevent the deprecation message being shown at render time
 	 * we need an extra function between useMemo and emitterObservers
 	 * so that the deprecated message gets shown only at invocation time.
 	 * (useMemo calls the passed function at render time)
--- a/assets/js/base/context/providers/cart-checkout/checkout-state/index.js
+++ b/assets/js/base/context/providers/cart-checkout/checkout-state/index.js
@@ -127,6 +127,7 @@ export const CheckoutStateProvider = ( {
 	 * we need an extra function between useMemo and emitterObservers
 	 * so that the deprecated message gets shown only at invocation time.
 	 * (useMemo calls the passed function at render time)
+	 * See: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4039/commits/a502d1be8828848270993264c64220731b0ae181
 	 */
 	const onCheckoutBeforeProcessing = useMemo( () => {
 		const callback = emitterObservers( observerDispatch )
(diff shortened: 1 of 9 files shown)


assets/js/type-defs/contexts.js
```diff
@@ -120,16 +120,16 @@
...
+ *                                                                               fire when the validation of the submitted checkout
+ *                                                                               data happens, before it's sent off to the
+ *                                                                               server.
  * @property {function(function(),number=)} onCheckoutAfterProcessingWithSuccess Used to register a callback that will
  *                                                                               fire after checkout has been processed
  *                                                                               and there are no errors.
  * @property {function(function(),number=)} onCheckoutAfterProcessingWithError   Used to register a callback that will
  *                                                                               fire when the checkout has been
  *                                                                               processed and has an error.
- * @property {function(function(),number=)} onCheckoutBeforeProcessing           Used to register a callback that will
```
opr, line 129: Removing this causes a TypeScript warning - because this is still used in `usePaymentMethodInterface` but according to the type-defs does not exist on `useCheckoutContext`.

opr (approved):
Looks great! Works well too, I just have one (non-blocking) comment about TypeScript warnings - it's up to you to decide if we care about this warning for now, or if you want to add the property back to the type-def.

assets/js/type-defs/contexts.js
ralucaStan: I prefer not to handle this here, I've added a deprecated comment where usePaymentMethodInterface is declared, once that file is converted to .ts I image it should be fine

ralucaStan:
I fixed the tests with commit https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4039/commits/a502d1be8828848270993264c64220731b0ae181, see the description for details.

PR ready to retest @opr

opr (changes requested):
The latest change looks good, clever way of solving it! I know you did add the comment in the commit message about the reasoning behind this, but could you add an inline comment explaining it too? Without knowing the context this code looks pretty strange.

ralucaStan:
> The latest change looks good, clever way of solving it! I know you did add the comment in the commit message about the reasoning behind this, but could you add an inline comment explaining it too? Without knowing the context this code looks pretty strange.

Good idea, I added a comment.

assets/js/base/context/providers/cart-checkout/checkout-state/index.js
```diff
@@ -122,6 +122,11 @@ export const CheckoutStateProvider = ( {
 	}, [ observers ] );
 	/**
 	 * @deprecated use onCheckoutValidationBeforeProcessing instead
+	 *
+	 * To prevent the deprecation message to be shown at render time
```
opr:
```suggestion
	 * To prevent the deprecation message being shown at render time
```

assets/js/base/context/providers/cart-checkout/checkout-state/index.js
```diff
@@ -122,6 +122,11 @@ export const CheckoutStateProvider = ( {
...
+	 * we need an extra function between useMemo and emitterObservers
+	 * so that the deprecated message gets shown only at invocation time.
+	 * (useMemo calls the passed function at render time)
```
opr, line 129:
```suggestion
	 * (useMemo calls the passed function at render time)
	 * See: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4039/commits/a502d1be8828848270993264c64220731b0ae181
```

opr (approved):
Approved again! There's one grammar change and I would recommend adding the commit link for additional context. 🥳

PR #4044, opened 2021-04-07
Cart: fix headline alignment in the empty state of the cart block

### LINKED ISSUE #3903: "Your cart is currently empty!" message is not centered by default
## To reproduce
1. Create a new page with the Cart block and publish it.
2. Visit it in the frontend with an empty cart.
3. Notice the _Your cart is currently empty!_ message is not centered.

## Expected behavior
The message should be centered instead of being left-aligned.

## Screenshots
imatge

## Environment
I could reproduce the issue in WC Blocks 4.6 and 4.5.2.

---

This issue sounds similar to #3788 (which affects Featured Product/Category), even though they might be unrelated.

frontdevde:
This PR fixes an issue where headlines in the empty state of the cart block are left aligned instead of center aligned as intended in the design. This issue is caused by the headline blocks using the align option instead of the text align option.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3903

### Screenshots

|Design|Before|After|
|-|-|-|
|Screenshot 2021-04-07 at 15 09 15|Screenshot 2021-04-07 at 15 19 57|Screenshot 2021-04-07 at 15 18 52|

1. Prerequisite: ensure you have an empty cart.
2. Create a new page and add the Cart block.
2. Check that the empty state looks as expected in the editor (see After state above).
3. Publish the page and check the empty state in the frontend.

### Changelog

> Fix headline alignment in the empty state of the cart block.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/empty-cart-edit/index.js
+++ b/assets/js/blocks/cart-checkout/cart/empty-cart-edit/index.js
@@ -42,7 +42,7 @@ const templateItems = [
 	[
 		'core/heading',
 		{
-			align: 'center',
+			textAlign: 'center',
 			content: __(
 				'Your cart is currently empty!',
 				'woo-gutenberg-products-block'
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
LGTM, thanks for fixing this @frontdevde. :tada: 

For reference, this change fixes the alignment of newly added Cart blocks but it doesn't affect existing Cart blocks, which will continue with whichever alignment they had.

PR #4050, opened 2021-04-08
Load translation file for JS files that has translatable strings

### LINKED ISSUE #4037: Translation not fully loading from translate.wordpress.org
i have an issue on the cart page that the words: subtotal, shipping and save arent translated and also cannot be translated using a plugin like Loco Translate as show in the attached image.
(the wordpress and also the plugins are up to date, i just double checked)
Capture

senadir:
Fixes #4037 

While registering some of our lib scripts (middleware, store, 3rd packages), we didn't load the JSON translation file for those scripts; this results in some translation not loading on the front-end.

This PR enables that loading. I checked each package to see if it has translatable strings and enable it.

### How to test
- Change website language to something other than English; Spanish is a good example of a 100% translated language.
- Go to Dashboard -> Updates and scroll down to download the new translations immidiatly.
- Visit Cart or Checkout block.
- Strings like Subtotal and Taxes should be translated.
### Before 
image

### After
image

### DIFF
--- a/src/Assets.php
+++ b/src/Assets.php
@@ -46,16 +46,16 @@ class Assets {
 
 		// Shared libraries and components across multiple blocks.
 		$asset_api->register_script( 'wc-blocks-middleware', 'build/wc-blocks-middleware.js', [], false );
-		$asset_api->register_script( 'wc-blocks-data-store', 'build/wc-blocks-data.js', [ 'wc-blocks-middleware' ], false );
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Nice fix @senadir, thanks for investigating this! I could reproduce the issue and verify it's fixed.

opr (approved):

PR #4051, opened 2021-04-09
Use font color in payment methods border

Aljullu:
Changes the border color of the payment methods options so it inherits the font color. This way, it follows the same styling all other components of the Cart and Checkout blocks have.

### Screenshots

| Before | After |
| --- | --- |
| imatge | imatge |

1. Using Storefront, in the admin go to Appearance > Customize.
2. Set the font color to something different than gray/black. For example, red.
3. Apply the changes and go to the Checkout page in the frontend.
4. Verify the payments methods of the Checkout block have a border with the same color as the text.

### DIFF
--- a/assets/js/blocks/cart-checkout/payment-methods/style.scss
+++ b/assets/js/blocks/cart-checkout/payment-methods/style.scss
@@ -195,20 +195,18 @@
 		font-weight: bold;
 	}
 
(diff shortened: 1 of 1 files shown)


mikejolley (approved):
🧙🏻 :shipit:

PR #4055, opened 2021-04-13
Move block settings to core and blocktype settings

mikejolley:
This is part of #3748 and is the base for https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4059

This moves all block settings registered through the `woocommerce_shared_settings` hook to the core settings list, or where more appropriate, the block types themselves.

I have added inline todos for settings which should be moved to block types, but require a little more effort to do so.

Also a small change within `asset_data_registry`—you can know use the 3rd arg as `true` to register a setting only if it is not already registed without an exception being thrown. This avoids adding inline exists checks for all settings.

In a follow up, more settings will be moved to the block types, and we'll start moving away from constants to getSetting usage.

Smoke test cart/checkout/product blocks load on the frontend and within the editor. Since this change mostly just moves things to the core settings array, we just need to ensure there is no obvious regression.


ralucaStan (approved):
Front-end changes look good, I've also tested the blocks and things look good. I recommend getting another 👁️ on the PHP side.

mikejolley:
Thanks @ralucaStan I've rebased this one into https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4059 to target trunk.

PR #4056, opened 2021-04-13
Check if Cart and Checkout are registered before removing payment methods

### LINKED ISSUE #29656: Fatal error: WooCommerce 5.2 and Elementor
Hello guys,

I updated the latest version of the WooCommerce plugin and when I try to edit any page with Elementor page builder I get a fatal error. Errors only appear when you enabled any Payment methods, in my case I enabled Direct bank transfer.

Error log file:

Warning: Undefined array key "wc-cart-block" in /public_html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/src/Payments/Api.php on line 221

Warning: Undefined array key "wc-cart-block-frontend" in /public_html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/src/Payments/Api.php on line 221

Warning: Undefined array key "wc-checkout-block" in /public_html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/src/Payments/Api.php on line 221

Warning: Undefined array key "wc-checkout-block-frontend" in /public_html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/src/Payments/Api.php on line 221

Warning: Attempt to read property "deps" on null in /public_html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/src/Payments/Api.php on line 221

Fatal error: Uncaught TypeError: array_diff(): Argument #1 ($array) must be of type array, null given in /public_html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/src/Payments/Api.php:222 Stack trace: #0 /public_html/wp-content/plugins/woocommerce/packages/woocommerce-blocks/src/Payments/Api.php(222): array_diff(NULL, Array)...

Best regards,
Nenad

senadir:
Fixes https://github.com/woocommerce/woocommerce/issues/29656

Elementor empties `$wp_scripts` and `$wp_styles` when rendering its editor (which is very dangerous), this breaks our integrations which must run after `wp_enqueue_scripts` and at the beginning of `wp_print_scripts`.

In this PR, I check if our scripts are still registered before validating payment methods dependencies.

This might have caused a regression with #3920 so more testing should happen using that PR steps.

### Testing steps
- On a website with PHP 8, install Elementor.
- Create a new page, and select "Edit with Elementor", there should be no fatal errors.

### Regression testing.
- Follow testing steps in #3920, mainly, add a faux dep to an enabled payment method and check if Cart and Checkout render fine in frontend and backend.

### DIFF
--- a/src/Payments/Api.php
+++ b/src/Payments/Api.php
@@ -216,6 +216,12 @@ class Api {
 
 					$cart_checkout_scripts = [ 'wc-cart-block', 'wc-cart-block-frontend', 'wc-checkout-block', 'wc-checkout-block-frontend' ];
 					foreach ( $cart_checkout_scripts as $script_handle ) {
+						if (
+							! array_key_exists( $script_handle, $wp_scripts->registered ) ||
+							! property_exists( $wp_scripts->registered[ $script_handle ], 'deps' )
+						) {
+							continue;
+						}
 						// Remove payment method script from dependencies.
 						$wp_scripts->registered[ $script_handle ]->deps = array_diff(
 							$wp_scripts->registered[ $script_handle ]->deps,


nerrad (approved):
Code looks okay to me. I didn't test though.

Aljullu (approved):
I tested #3920 and it looks like it was not regressed. I also followed the testing steps in this PR with PHP 7.4 and couldn't find any error in the logs.

LGTM

PR #4059, opened 2021-04-13
Move Block Type Settings into Block Type Classes

mikejolley:
Note this PR is based on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4055 not trunk. This is part of #3748

**This PR changes how settings are defined for blocks.**

1. **Core Settings.** These are shared/global and useful to blocks but also other projects. Core settings are exported from the `@woocommerce/settings` package.
2. **Block Config Settings.** These are shared between multiple blocks e.g. build properties. They are injected on all pages.
3. **Block Type Settings.** These are defined by block types themselves.

**Block Constants** now only exports the Block Config (build) and derived settings e.g. a page permalink.

Block Type Settings are not defined as constants, and instead use `getSetting` inline. These should not be constant because they will not exist on all requests.

In addition to standardising and reviewing the current settings, this PR:

- Implements TypeScript in the settings packages
- Moves some settings from block to core to block type and vice versa
- Ensures settings are added via AssetDataApi rather than using other methods.

1. Smoke test product blocks - All Products and SSR blocks - confirm max/min/default rows and columns work as expected
2. Smoke test cart and checkout
3. Tests should pass - I updated globals so that the suite still works.

This PR is large because it swaps constants to getSetting across all files.


mikejolley:
@nerrad Just a glance from you on the general approach would be useful; this is quite different from the original PR in how it tackles core/block/blocktype settings.

@senadir Sorry for the size. 100 files but the majority are renaming constants.

Another one with the potential to grow stale 😢

src/Assets.php
```diff
@@ -57,6 +60,29 @@ public static function register_assets() {
...
+		$asset_data_registry->add( 'wordCountType', _x( 'words', 'Word count type. Do not translate!', 'woo-gutenberg-products-block' ) );
+		$asset_data_registry->add( 'placeholderImgSrc', wc_placeholder_img_src() );
+		$asset_data_registry->add( 'isLargeCatalog', $product_counts->publish > 100 );
+		$asset_data_registry->add( 'productCount', array_sum( (array) $product_counts ) );
+		$asset_data_registry->add(
+			'restApiRoutes',
+			[
+				'/wc/store' => array_keys( Package::container()->get( RestApi::class )->get_routes_from_namespace( 'wc/store' ) ),
+			]
+		);
```
nerrad: I appreciate that technically this doesn't matter because this is included in WC Core anyways, but going by the earlier comment that states:

```
// Pass in settings specific to blocks
```

I don't think these data items are really that specific to blocks. Blocks may be the primary consumers of these settings value but there's nothing about them that seems to indicate they are expressly needed by blocks (i.e. they could be used in other contexts?)

nerrad (commented):
Mike thanks for doing this! In general **I like the approach you are taking here PHP side for passing through the data**. I have some additional feedback on what my thinking is for the direction we should be heading towards client side. This is not feedback I expect to be acted on in this PR (it's already big enough) but I do think it'd be good to consider for future followups.

### Standardizing around `getSetting`

I'd personally like to standardize around just using `getSetting` in the JS and move away from constants for the values altogether (we'll have to keep those exported on `@woocommerce/settings` for back-compat though). There's a few reasons for this:

- it establishes `getSetting` as the official API for retrieving these type of values (from a window global).
- it eliminates some of the unnecessary indirection around `@woocommerce/block-settings` which we've seen is confusing for new developers coming into the repo when this is essentially just a wrapper around `getSetting` anyways.

The main _downside_ of course is having to know the specific setting string to pass into the function. One way to get around that would be to export constants for the settings strings and then you still get the discovery of these strings via the constants but still enforce the `getSetting` contract. So the API might be something like:

```js
import { IS_LARGE_CATALOG, getSetting } from '@woocommerce/settings';
const isLargeCatalog = getSetting( IS_LARGE_CATALOG, false );
```

### Consider creating and exporting a `useSetting` hook and eventually replacing usage of `getSetting` with this hook.

One problem we have right now with `getSetting` is it tightly couples us to the availability of the passed through settings value from the server. If we switched to a custom hook instead, that allows us to still use the global for now, but also switch to a settings endpoint in the future. So we'd provide and implement an interface that is more React friendly from the get go.

Something like this could work initially:

```js

// placeholder for eventual known setting key to endpoint map.
const getEndpoint = ( settingKey ) => '';

const useSetting = ( { settingKey, fallBack, settingSource = 'fetch', endpoint = '' } ) => {
    endpoint = settingSource === 'fetch' && ! endpoint
      ? getEndpoint( settingKey )
      : '';
    const canFetch = settingSource === 'fetch' && endpoint;
    // for now we're just always returning getSetting, but the above logic is just to show how we could
    // eventually determine how to get the settings data automatically.
    return getSetting( settingKey, fallBack );
}
```

Then that gives us some flexibility in the future for also providing a `fetch` source for the setting in a headless context. Any components using this would be _much_ easier to update so they are only using the global if it's present.

Caveat, this is _just to communicate the concept_, I'm not suggesting the interface would be exactly as it is above. Also, if we do this, we probably would want to export the hook on a `@woocommerce/settings-hook` or something similar because the custom hook will likely in the future have React dependencies and in the WP context `@woocommerce/settings` might not always be used in code that also requires React (so I think we'll want to avoid creating that dependency out of the box).

src/Assets.php
mikejolley: The idea was we'd include more niche settings here (for blocks to consume) without polluting _core settings_ and causing future bw compat issues. Comment aside, are you still ok with these being located here?

nerrad: Yes I understand that. But from a consumer perspective, given this plugin is included in core as a package, these values are always going to be present anytime the settings global is output for a route because the data is being added on `init`. So effectively, what you're aiming for here isn't happening. Any route that outputs `wc-settings` is going to have this data (unless I'm missing something).  

If we truly want to distinguish these values from "core" settings, then we'd want to only register this data when any block is rendered (as opposed to when registering block scripts). 

Also, since the keys of the settings aren't namespaced that well (`productCount` for instance), we could nest those settings inside an associative array representing blocks and then just expose the entire object as a registered setting? Something like:

```js
$woo_blocks_data = [
    'wordCountType' => _x( 'words', 'Word count type. Do not translate!', 'woo-gutenberg-products-block' ) ),
    'placeholderImgSrc' => wc_placeholder_image_src(),
    // etc...
];
$asset_data_registry->add( 'woo-blocks-data', $woo_blocks_data );
```

mikejolley: I'm not sure what direction to take here then. These could all be defined in the core settings section but that seems wrong for block specific settings. Could we somehow populate settings with a closure and avoid window globals? I'm not sure if that's possible or worth. Ideally we should be able to add private namespaced settings for blocks without worrying about direct access and BW compatibility, because the window global bypasses the API completely.

assets/js/editor-components/utils/index.js
```diff
@@ -21,8 +20,9 @@ const getProductsRequests = ( {
 	search = '',
 	queryArgs = [],
 } ) => {
+	const isLargeCatalog = getSetting( 'productCount', 0 ) > 100;
```
nerrad: Oh nice, good way to eliminate an unnecessary property on the global 👏

assets/js/settings/blocks/constants.ts
```diff
@@ -3,30 +3,27 @@
...
-const IS_LARGE_CATALOG = getSetting( 'isLargeCatalog' );
-const PLACEHOLDER_IMG_SRC = getSetting( 'placeholderImgSrc', '' );
-const PRODUCT_COUNT = getSetting( 'productCount', 0 );
-const REST_API_ROUTES = getSetting( 'restApiRoutes' );
-const WC_BLOCKS_ASSET_URL = getSetting( 'wcBlocksPluginUrl', '' ) + '/assets/';
-const WC_BLOCKS_BUILD_URL = getSetting( 'wcBlocksPluginUrl', '' ) + '/build/';
-const WOOCOMMERCE_BLOCKS_PHASE = getSetting( 'wcBlocksPhase', 1 ) as number;
-const WORD_COUNT_TYPE = getSetting( 'wordCountType', 'words' );
+declare global {
+	interface Window {
```
nerrad: nifty, so this doesn't have to be `window` (lowercase)?

src/Assets.php
```diff
@@ -60,37 +60,43 @@ public static function register_assets() {
...
+			var wcBlocksMiddlewareConfig = {
+				storeApiNonce: '" . esc_js( wp_create_nonce( 'wc_store_api' ) ) . "',
+				wcStoreApiNonceTimestamp: '" . esc_js( time() ) . "'
+			};
 			",
 			'before'
 		);
+
+		// Blocks config is available globally for all blocks scripts.
+		wp_enqueue_script( 'wc-blocks-config' );
```
nerrad: This will load `wc-blocks-config` on every route (not just routes that happen to have blocks on it). Instead should we make this handle a dependency of `wc-blocks` (or if you implement my earlier suggestion it will got loaded automatically on the right route).

src/Assets.php
```diff
@@ -60,37 +60,43 @@ public static function register_assets() {
...
+		wp_add_inline_script(
+			'wc-blocks-config',
+			"
+			var wcBlocksConfig = {
+				pluginUrl: '" . esc_js( plugins_url( '/', __DIR__ ) ) . "',
+				buildPhase: '" . esc_js( Package::feature()->get_flag() ) . "'
+			};
+			",
+			'before'
+		);
```
nerrad: It's still not clear to me in this case why this data needs to be passed through to the client this way. You could accomplish the same here by just doing the following in `AbstractBlock::enqueue_data`:

```suggestion
		$aset_data_registry->add(
			'wcBlocksConfig',
			[
			    'pluginUrl' => plugins_url( '/', __DIR__ ),
			    'buildPhase' =>  Package::feature()->get_flag()
			]
		);
```

That lets the asset data registry take care of escaping and everything for you? Then because it's in `AbstractBlock::enqueue_data` the resulting json object will only be available on a route that has a block in it?

src/BlockTypes/AbstractBlock.php
```diff
@@ -335,6 +335,20 @@ protected function enqueue_data( array $attributes = [] ) {
...
+			* enter 'characters_excluding_spaces' or 'characters_including_spaces'. Otherwise, enter 'words'.
+			* Do not translate into your own language.
+			*/
+			$this->asset_data_registry->add( 'wordCountType', _x( 'words', 'Word count type. Do not translate!', 'woo-gutenberg-products-block' ), true );
+		}
+
+		if ( ! $this->asset_data_registry->exists( 'productCount' ) ) {
+			$product_counts = wp_count_posts( 'product' );
+			$this->asset_data_registry->add( 'productCount', array_sum( (array) $product_counts ) );
+		}
```
nerrad, line 360: If you move `pluginUrl` and `buildPhase` into here, you could have `productCount` and `wordCountType` be a part of the `wcBlocksConfig` object too. Then they are appropriately namespaced and any client consumer just taking a look at the `wcSettings` global will see these are properties of `wcSettings.wcBlocksConfig` which makes it clearer they aren't a core config.

nerrad (commented):
Thanks Mike, I'm really liking the changes. I still have a few comments, mostly around the new `wc-blocks-config` (empty) script and data.

Other than that this is really close to merge.

src/Assets.php
mikejolley: I found that wc-blocks is not on every route, so would have to be a dependency on all scripts. Cart and checkout for example.

src/Assets.php
mikejolley: I'm concerned that these build variables are used everywhere - are we certain block types cover all possible routes?

nerrad: Well, we currently have all our blocks registered server side right and as far as I can tell every block at a minimum extends `AbstractBlock` so that means `enqueue_data` is going to be fired server side for every registered block?

Your concern is definitely noted and important to consider though if we have any of our JavaScript running on routes that might not have a block (or block components)?

assets/js/settings/blocks/constants.ts
mikejolley: TS expects this casing - it shows inline errors otherwise.

assets/js/blocks/price-filter/edit.js
```diff
@@ -151,7 +152,7 @@ export default function ( { attributes, setAttributes } ) {
 
 	return (
 		<>
-			{ getSetting( 'productCount', 0 ) === 0 ? (
+			{ blocksConfig.productCount === 0 ? (
```
nerrad, line 155: This is _slightly_ different than the previous implementation because `undefined !== 0` (before we defaulted to `0` if it was undefined). Practically, we probably wouldn't see this in a WordPress context but still worth noting I think.

assets/js/blocks/products/all-products/edit.js
```diff
@@ -283,7 +284,7 @@ class Editor extends Component {
 		const blockTitle = this.getTitle();
 		const blockIcon = this.getIcon();
 
-		if ( ! getSetting( 'hasProducts', true ) ) {
+		if ( blocksConfig.productCount === 0 ) {
```
nerrad, line 287: If `blocksConfig.productCount` isn't defined this condition won't trigger.

tests/e2e/specs/backend/__fixtures__/product-search.fixture.json
```diff
@@ -1 +1,4 @@
...
\ No newline at end of file
+{
+	"title": "Product Search Block",
+	"pageContent": "<!-- wp:woocommerce/product-search {\"formId\":\"wc-block-product-search-1\"} -->\n<div class=\"wp-block-woocommerce-product-search\"><div class=\"wc-block-product-search\"><form role=\"search\" method=\"get\" action=\"http://localhost:8889/\"><label for=\"wc-block-product-search-1\" class=\"wc-block-product-search__label\">Search</label><div class=\"wc-block-product-search__fields\"><input type=\"search\" id=\"wc-block-product-search-1\" class=\"wc-block-product-search__field\" placeholder=\"Search products…\" name=\"s\"/><input type=\"hidden\" name=\"post_type\" value=\"product\"/><button type=\"submit\" class=\"wc-block-product-search__button\" label=\"Search\"><svg aria-hidden=\"true\" role=\"img\" focusable=\"false\" class=\"dashicon dashicons-arrow-right-alt2\" xmlns=\"http://www.w3.org/2000/svg\" width=\"20\" height=\"20\" viewbox=\"0 0 20 20\"><path d=\"M6 15l5-5-5-5 1-2 7 7-7 7z\"></path></svg></button></div></form></div></div>\n<!-- /wp:woocommerce/product-search -->"
```
nerrad, line 3: So what changed here and in the `product-search.text.js.snap` interestingly enough was the url:

Before:
```
http://localhost:8889
```
After:
```
http://localhost:8889/
```

It looks like this might have been caused by the `esc_url` added around the `homeUrl` property? I wonder if we should use `untrailingslashit` on it too?

nerrad (approved):
Thanks Mike for addressing the feedback and working through the various approaches of this issue.  I do have a few things that popped up as observations in the latest review that are slight changes in behaviour and I think should be addressed but I'm not sure they are blocking so I'll leave up to you to make the call. 

Pre-approving based on my review but note I haven't tested.

tests/e2e/specs/backend/__fixtures__/product-search.fixture.json
mikejolley: It's unrelated to the esc. There was an escape before this PR too. Notice the home_url has always been called with a `/` param. I'm not 100% sure what changed this. It's possible the test suite didn't have access to the correct variable prior to this refactor.

assets/js/blocks/products/all-products/edit.js
mikejolley: Understood. I don't think it matters in this context because we're defining the settings for this block in the block type class.

PR #4068, opened 2021-04-15
Fix failing e2e tests

opr:
This PR updates the version of `@wordpress/e2e-test-utils` and adds some checks to existing tests and utility functions to make them more robust.

- `tests/e2e/utils.js` - Because of [this PR](https://github.com/WordPress/gutenberg/pull/29759/files) in Gutenberg, the button for adding blocks, or toggling the inserter had its aria-label renamed. This caused our utility function `closeInserter` to fail. I changed this function to check for the existence of the 'Add block' button, and if it doesn't exist check for the 'Toggle block inserter' button
- ` tests/e2e/specs/backend/reviews-by-product.test.js` - Added code to ensure a product was checked before counting the reviews. I noticed that a product to show reviews for was randomly checked in some cases and when the code to click the list item was executed it became unchecked.
- `Snapshots and fixtures` - for reasons I cannot identify, a snapshot and fixture for the product search block have changed. The diff is small and only the `wc-block-product-search-` number has changed.

Fixes E2E tests failing.

1. Run e2e tests.
2. Push a commit to make them fail.
3. Observe them failing
4. Revert commit and observe them passing.


mikejolley (approved):
Tests passed here and changes look good 👍🏻

PR #4075, opened 2021-04-19
Rest API batching support (2)

### LINKED ISSUE #2094: All Products: add to cart can fail when adding multiple products to cart in quick succession

mikejolley:
Raising https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3899 again.

**When testing you must ensure you have reran `npm install` for the dataloader dependency.**

This PR implements batching in the Store API, and a dataloader class in the client to batch requests within a certain interval.

Batching works for POST, PUT, DELETE API requests. GET requests are not affected.

As well as adding the batching, [this commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/d0b901b06174aa514efaf23bec1482db13f6b276) creates a control to handle fragment updates and adds a timeout. This prevents multiple fragment updates after a batch of events have ran.

NOTE: I am requested feedback/code review, but some things are left to do before it can be merged:

- [x] Check what happens if nonce changes on items in the batch.
- [x] Batch is only supported on 5.6+ so we need the client to account for this also.
- [x] Reduce batch timeout from 3000 - I have left this high to make testing easier.

Fixes #2094

1. Go to the products page.
2. Add to items to the cart really quickly. If you focus a button with the keyboard you can hit enter and click another button.
3. Check the network request for a "batch" API call. You'll see 2 responses in an array.
4. Smoke test cart page functionality/cart/checkout. All should be unaffected.
5. Check that on the cart page, changing qty persists and the API request is successful.
6. Check that on the cart page, removing an item from the cart is successful.

### Changelog

> Added support to the Store API for batching requests. This allows multiple POST requests to be made at once to reduce the number of separate requests being made to the API.


assets/js/middleware/cart-update.ts
```diff
@@ -12,11 +12,35 @@ import { LAST_CART_UPDATE_TIMESTAMP_KEY } from '../data/cart/constants';
...
+
+	if ( isBatch ) {
+		const requests = options?.data?.requests || [];
+
+		return requests.some( ( request ) => {
+			const requestUrl = request.path || '';
+
+			return cartRegExp.exec( requestUrl ) !== null;
+		} );
+	}
```
nerrad: I like your implementation much better here (over the suggestion I gave) 👏

src/StoreApi/Routes/AbstractCartRoute.php
```diff
@@ -56,6 +56,14 @@ public function get_response( \WP_REST_Request $request ) {
 		$this->cart_controller->load_cart();
 		$this->calculate_totals();
 
+		if ( $this->requires_nonce( $request ) ) {
+			$nonce_check = $this->check_nonce( $request );
+
+			if ( is_wp_error( $nonce_check ) ) {
+				return $this->add_nonce_headers( $this->error_to_response( $nonce_check ) );
+			}
+		}
```
nerrad: I'm curious about why you moved this out of the `validate_callback` function and eliminated that function?

tests/e2e/config/jest.setup.js
```diff
@@ -11,8 +11,8 @@ import {
 	visitAdminPage,
 } from '@wordpress/e2e-test-utils';
 
-// Set the default test timeout to 30s.
-let jestTimeoutInMilliSeconds = 30000;
+// Set the default test timeout to 40s.
+let jestTimeoutInMilliSeconds = 40000;
```
nerrad: Do we still need this timeout increase now? It shouldn't be needed?

nerrad (approved):
I only got a chance to do a review of the code and it looks good to me. I do have a question about removing `validate_callback` function in favour of an inline nonce check on the response but that is non-blocking.

If needed I can run through some testing tomorrow, but I'll pre-approve based on the code review.

src/StoreApi/Routes/AbstractCartRoute.php
mikejolley: I found that when you use these validation callbacks, WP does the error handling for the response which prevents us from appending updated nonce headers (unless you use a filter hook). I moved it back here so we can control the response.

assets/js/middleware/cart-update.ts
```diff
@@ -12,11 +12,35 @@ import { LAST_CART_UPDATE_TIMESTAMP_KEY } from '../data/cart/constants';
...
+	const isBatch = batchRegExp.exec( url ) !== null;
+
+	if ( isCart ) {
+		return true;
+	}
+
+	if ( isBatch ) {
+		const requests = options?.data?.requests || [];
+
+		return requests.some( ( request ) => {
```
opr: Could we do this so TS doesn't complain about request being implicitly `any`?

```suggestion
		return requests.some( ( request: { path: string } ) => {

```

tests/e2e/config/jest.setup.js
mikejolley: Nope, Ill revert this.

assets/js/middleware/cart-update.ts
opr: I think maybe `Record< string, unknown > & { path: string }` is more correct, but doesn't add any value

PR #4077, opened 2021-04-19
Make error messages in front-end cart E2E tests better

opr:
This PR will cause calls to `page.waitForFunction` that time out (and subsequently throw an error) to be caught and allow a more descriptive error to be displayed.

1. Edit `assets/js/data/cart/constants.ts` and change the value of `LAST_CART_UPDATE_TIMESTAMP_KEY` to something else.
2. Run the tests and ensure the error message shown to you tells you that the value couldn't be found in local storage.

### DIFF
--- a/tests/e2e/specs/frontend/cart.test.js
+++ b/tests/e2e/specs/frontend/cart.test.js
@@ -77,15 +77,22 @@ describe( `${ block.name } Block (frontend)`, () => {
 	} );
 
 	it( 'Shows the freshest cart data when using browser navigation buttons', async () => {
-		await jest.setTimeout( 60000 );
+		//await jest.setTimeout( 60000 );
 		await page.goto( cartBlockPermalink );
-		await page.waitForFunction( () => {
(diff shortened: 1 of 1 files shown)


tests/e2e/specs/frontend/cart.test.js
```diff
@@ -77,15 +77,22 @@ describe( `${ block.name } Block (frontend)`, () => {
 	} );
 
 	it( 'Shows the freshest cart data when using browser navigation buttons', async () => {
-		await jest.setTimeout( 60000 );
+		//await jest.setTimeout( 60000 );
```
mikejolley: Is this a mistake?

mikejolley (approved):
Worked as described 👍🏻 One line to fix.

tests/e2e/specs/frontend/cart.test.js
opr: I think it should be removed entirely, actually. Thanks for catching it. Running tests again with it removed to ensure they finish in time.

PR #4087, opened 2021-04-21
Check for element before trying to click it in cart backend test

opr:
I noticed that the backend cart tests sometimes failed because they were trying to click an element that didn't exist. I believe this flakiness was down to a race condition, so I have added a check to stall the test until the element is available.

If the selector never resolves, we can output an error message saying the block didn't load. This would indicate something wrong with the block, rather than the tests.

Not really sure how we can test this apart from merging it and seeing how it works on future PRs, I have run the GitHub actions twice here and both times all tests have been green.


ralucaStan (approved):
🤞

PR #4088, opened 2021-04-21
Convert Cart Table to TypeScript

opr:
### Notable changes made in this PR:
- `ProductLowStockBadge`
  - Changed `lowStockRemaining` prop to be a number. The API returns either a number (how many units remain) or `null`. If it is `null` this component will not be rendered.
- `QuantitySelector`
  - Changed a call to `isNaN` to one of the type guards: `isNumber` - this ensures TS correctly infers the variable to be a `number`.
- `useStoreCartItemQuantity`
  - Add a function called `cartItemHasQuantityAndKey` this is required because it is possible for the cart item to be `{}` so we must infer that this object has the keys `quantity` and `key` - if it doesn't they can have default values, but we need to know it has them at any rate.
- Type guards - these are added in a few other PRs but none of them have been merged into trunk yet. We borrowed them from https://github.com/hqoss/guards but adapted `isObject` to work with our repo and added our own: `objectHasProp`.
- `CartLineItemRow`
  - Removed the `parseInt` from the precision, this is because the API will always send back a number. It can't be a string.
- `CartItem` and `CartResponseItem`
  - These types were incomplete, the API sends back a little more data than was in the existing types.
 
1. Smoke test the cart and checkout blocks


assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
```diff
@@ -27,6 +27,13 @@ import {
 import Dinero from 'dinero.js';
 import { DISPLAY_CART_PRICES_INCLUDING_TAX } from '@woocommerce/block-settings';
 import { useCallback, useMemo } from '@wordpress/element';
+import type { CartItem } from '@woocommerce/type-defs/cart';
+
+/**
+ * Internal dependencies
+ */
+import { Currency } from '../../../../../../packages/prices/types';
+import { objectHasProp } from '../../../../base/utils/type-guards';
```
Aljullu: Can we use aliases in these imports?

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
```diff
@@ -184,7 +198,11 @@ const CartLineItemRow = ( { lineItem = {} } ) => {
 			{ /* If the image has no alt text, this link is unnecessary and can be hidden. */ }
 			<td
 				className="wc-block-cart-item__image"
-				aria-hidden={ ! firstImage.alt }
+				aria-hidden={
+					( objectHasProp( firstImage, 'alt' ) &&
+						! firstImage.alt ) ||
+					true
+				}
```
Aljullu: I'm a bit confused by this, is the last `|| true` necessary? Also, the `objectHasProp( firstImage, 'alt' )` check seems to change the way it works if `firstImage` doesn't have an `alt` attribute. To keep the behavior unchanged, shouldn't it be like this?
```suggestion
				aria-hidden={
					! objectHasProp( firstImage, 'alt' ) || ! firstImage.alt )
				}
```

Aljullu (commented):
Looking and testing good, good job porting all these components to TS. :clap: 

I see several `Missing return type on function.` warnings from the ESLint GitHub Action, is that something that can be fixed before merge? I also left a couple of comments/questions below.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
opr: Yes we can! I have updated the pull now.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
opr: No you're right, it's not - I misinterpreted the error message TS was giving me when creating this, but your solution is great.

opr:
@Aljullu I've made a few changes, these include removing `PropTypes` from components that are now typed, and explicitly stating the return type of components.

Aljullu (approved):
Thanks for the updates @opr! I didn't test it again, but code changes look good to me. :ship:

PR #4090, opened 2021-04-22
Switch to `rest_preload_api_request` for API hydration in cart and checkout blocks.

### LINKED ISSUE #3736: WC()->api->get_endpoint_data vs rest_preload_api_request
Investigate whether it's worth using `rest_preload_api_request` instead of `WC()->api->get_endpoint_data`.

https://developer.wordpress.org/reference/functions/rest_preload_api_request/

mikejolley:
Implements `rest_preload_api_request` for api hydration in place of `WC()->api->get_endpoint_data`.  Added a `hydrate_api_request` helper to the asset data registry to prevent the same data being added multiple times.

`rest_preload_api_request` works pretty similar to the previous method, however, it has the advantage of also returning `headers` which we can use in our hydration HOC (cc @opr).

Fixes #3736

1. Visit cart - ensure block loads (no regression)
2. Repeat for checkout.
3. Viewing source on either page you should see `preloadedApiRequests` encoded in the page/settings with the various routes populated.

### Changelog

> Switched to `rest_preload_api_request` for API hydration in cart and checkout blocks.

### DIFF
--- a/assets/js/hocs/with-store-cart-api-hydration.js
+++ b/assets/js/hocs/with-store-cart-api-hydration.js
@@ -17,13 +17,19 @@ import { LAST_CART_UPDATE_TIMESTAMP_KEY } from '../data/cart/constants';
  * Makes cart data available without an API request to wc/store/cart/.
  */
 const useStoreCartApiHydration = () => {
-	const cartData = useRef( getSetting( 'cartData' ) );
+	const preloadedApiRequests = useRef(
+		getSetting( 'preloadedApiRequests', {} )
+	);
 	const { setIsCartDataStale } = useDispatch( CART_STORE_KEY );
 
 	useSelect( ( select, registry ) => {
-		if ( ! cartData.current ) {
+		const cartData =
+			preloadedApiRequests.current[ '/wc/store/cart' ]?.body || false;
+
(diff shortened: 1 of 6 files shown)


assets/js/hocs/with-store-cart-api-hydration.js
```diff
@@ -17,13 +17,19 @@ import { LAST_CART_UPDATE_TIMESTAMP_KEY } from '../data/cart/constants';
...
-	const cartData = useRef( getSetting( 'cartData' ) );
+	const preloadedApiRequests = useRef(
+		getSetting( 'preloadedApiRequests', {} )
+	);
 	const { setIsCartDataStale } = useDispatch( CART_STORE_KEY );
 
 	useSelect( ( select, registry ) => {
-		if ( ! cartData.current ) {
+		const cartData =
+			preloadedApiRequests.current[ '/wc/store/cart' ]?.body || false;
```
ralucaStan: Only `const cartData = preloadedApiRequests.current[ '/wc/store/cart' ]?.body;` is enough, as it will return undefined if no `body` is found, which will cause a return in the if bellow.

ralucaStan (approved):
Looks good, I have added a small comment.

PR #4091, opened 2021-04-22
Update `Button` component to ts

opr:
This PR changes the `Button` component to TypeScript, and adds an alias of `@types/wordpress__components` to match the alias _we_ have for `@wordpress/components`.

1. Open any component using `Button` (cart and checkout for example) and ensure they work correctly.

### DIFF
--- a/assets/js/base/components/button/index.tsx
+++ b/assets/js/base/components/button/index.tsx
@@ -24,7 +24,7 @@ const Button = ( {
 	showSpinner = false,
 	children,
 	...props
-}: ButtonProps ) => {
+}: ButtonProps ): JSX.Element => {
 	const buttonClassName = classNames(
 		'wc-block-components-button',
 		className,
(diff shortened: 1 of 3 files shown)


Aljullu (approved):
LGTM

assets/js/base/components/button/index.tsx
```diff
@@ -2,27 +2,29 @@
...
- * @param {React.ReactChildren} props.children    Child components passed in.
- * @param {Object}              props.props       Rest of incoming props.
  */
-const Button = ( { className, showSpinner = false, children, ...props } ) => {
+const Button = ( {
+	className,
+	showSpinner = false,
+	children,
+	...props
+}: ButtonProps ) => {
```
Aljullu: Not sure why in the other PR I see an ESLint warning in this line while in this PR there isn't. :confused: 

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4098/files#diff-3d98a12115787e07b0261408773117bb47a24318d3165adc24a0729ac3846133R27

opr: Yeah this is strange, it never shows in my IDE either. I've fixed it now anyway :)

PR #4092, opened 2021-04-22
Deprecate the woocommerce_shared_settings hook

### LINKED ISSUE #3748: Don't load `wc_blocks_data` on every route unnecessarily.

mikejolley:
This is pulled from #3864 and closes #3748

This deprecates `woocommerce_shared_settings` with a notice. Tests have been updated to ignore warnings - we cannot test for the notice yet because WP tests do not support PHPUnit 9.x.

Add a filter somewhere, for example, main plugin file.

```
add_filter( 'woocommerce_shared_settings', function( $settings ) { return $settings; } );
```

Visit the store. 

Check the debug log:

Screenshot 2021-04-22 at 15 53 30

### Dev Note

Blocks surfaced a hook named `woocommerce_shared_settings` which could be used to change, add, or remove settings that are passed from the server to the client. This is now considered deprecated—a warning will be logged to the browser error console if used.

We recommend instead using the Asset Data Registry directly, so for example:

```
use Automattic\WooCommerce\Blocks\Assets\AssetDataRegistry;

Package::container()->get( AssetDataRegistry::class )->add( $key, $value )
```

Doing this lessons the risk of conflicting or overwriting data, since the filter allows any setting to be modified which could lead to unintentional side effects.

### Changelog

> Replaced usage of the `woocommerce_shared_settings` hook. This will be deprecated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,7 +16,7 @@
 		"automattic/jetpack-autoloader": "^2.9.1"
 	},
 	"require-dev": {
-		"phpunit/phpunit": "6.5.14",
+		"phpunit/phpunit": "7.5.20",
 		"woocommerce/woocommerce-sniffs": "0.1.0"
 	},
 	"autoload": {
--- a/composer.lock
+++ b/composer.lock
@@ -4,32 +4,39 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "8dfc916e2f18216a443b7824aa4c2baa",
+    "content-hash": "1afc770bcd0a752692a22bb62bb60d81",
     "packages": [
         {
             "name": "automattic/jetpack-autoloader",
-            "version": "v2.9.1",
+            "version": "2.10.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/Automattic/jetpack-autoloader.git",
-                "reference": "d6ca2cc26ad6963e1be19b3338a9e98f40d9bd88"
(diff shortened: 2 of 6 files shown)


Aljullu (approved):
LGTM

PR #4094, opened 2021-04-22
Create AssetsController and BlockTypesController (replaces Assets.php and Library.php)

### LINKED ISSUE #3868: Refactor `Assets` into a new initializer/loader/controller for registering block assets.
This is a spinoff from #3748:

Refactor `Assets` into a new initializer/loader/controller for registering block assets and inject AssetDataRegistry and `Assets/Api` as dependencies (then get rid of the old class). **priority 2** 
to a separate issue to make the output PR less bloated.

mikejolley:
Refactors Assets and Library with dependency injection. Closes #3875 and fixes #3868

I've marked this as in need of a dev note since we're removing classes, however, they are internal so unlikely to be used directly. cc @nerrad 

Will add inline notes for other changes.

General smoke test to ensure classes are still loaded. Check all block types are registered and existing blocks load within admin and frontend.

### Dev Note

> Two internal classes (`Automattic\WooCommerce\Blocks\Assets` and `Automattic\WooCommerce\Blocks\Library`) have been deprecated in this release, with functionality being moved to `AssetsController` and `BlockTypesController`. Since these were internal bootstrapping classes we don't anticipate any external usage, but noting that these classes will be removed in a future release of Blocks.

### Changelog

> Introduced AssetsController and BlockTypesController classes (which replace Assets.php and Library.php).


src/AssetsController.php
```diff
@@ -0,0 +1,151 @@
...
+		$this->register_style( 'wc-block-vendors-style', plugins_url( $this->api->get_block_asset_build_path( 'vendors-style', 'css' ), __DIR__ ), wp_style_is( 'wp-edit-post' ) ? [ 'wp-edit-post' ] : [] );
+	}
+
+	/**
+	 * Add body classes to the frontend and within admin.
+	 *
+	 * @param string|array $classes Array or string of CSS classnames.
+	 * @return string|array Modified classnames.
+	 */
+	public function add_theme_body_class( $classes ) {
```
mikejolley, line 96: Note I combined the 2 functions into one which handles $classes based on type instead of 2 functions for similar logic.

src/BlockTypesController.php
```diff
@@ -8,45 +8,70 @@
...
+	}
+
+	/**
+	 * Initialize class features.
 	 */
-	public static function init() {
-		add_action( 'init', array( __CLASS__, 'register_blocks' ) );
-		add_action( 'init', array( __CLASS__, 'define_tables' ) );
+	protected function init() {
+		add_action( 'init', array( $this, 'register_blocks' ) );
```
mikejolley, line 48: Note, we don't need to define tables any more now that the stock table is in core.

src/Assets/Api.php
```diff
@@ -121,29 +121,6 @@ function() use ( $handle ) {
...
-	 * @since 2.5.0
-	 * @since 2.6.0 Changed $name to $script_name and added $handle argument.
-	 * @since 2.9.0 Made it so scripts are not loaded in admin pages.
-	 * @deprecated 4.5.0 Block types register the scripts themselves.
-	 *
-	 * @param string $script_name  Name of the script used to identify the file inside build folder.
-	 * @param string $handle       Optional. Provided if the handle should be different than the script name. `wc-` prefix automatically added.
-	 * @param array  $dependencies Optional. An array of registered script handles this script depends on. Default empty array.
-	 */
-	public function register_block_script( $script_name, $handle = '', $dependencies = [] ) {
```
mikejolley, line 136: Note; I removed these as they have been deprecated for several versions and are not being used internally.

nerrad:
Just a quick note that I think we should still leave the classes in place with any public method and add depecrated calls/comment blocks. Once this is deployed to WooCommerce core in one version with the deprecations, the next release of blocks after that could remove the classes altogether (which will land in the next WooCommerce core release).

Some rationale here is that there are some deploy scenarios for WooCommerce core in some environments where fatals could be introduced so we have stagger actual class removal.

src/AssetsController.php
```diff
@@ -0,0 +1,151 @@
...
+namespace Automattic\WooCommerce\Blocks;
+
+use Automattic\WooCommerce\Blocks\Package;
+use Automattic\WooCommerce\Blocks\Assets\Api as AssetApi;
+use Automattic\WooCommerce\Blocks\Assets\AssetDataRegistry as AssetDataRegistry;
+
+/**
+ * AssetsController class.
+ *
+ * @internal
```
nerrad, line 12: Should we have a `@since $VID:$` here?

src/AssetsController.php
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * Initialize class features.
+	 */
+	protected function init() {
+		add_action( 'init', array( $this, 'register_assets' ) );
+		add_action( 'body_class', array( $this, 'add_theme_body_class' ), 1 );
+		add_action( 'admin_body_class', array( $this, 'add_theme_body_class' ), 1 );
+		add_action( 'woocommerce_login_form_end', array( $this, 'redirect_to_field' ) );
```
nerrad: This seems like an odd thing to include in this class. Is this an opportunity to move it out into the new `BlockTypesController` (I know it already existed in `Assets` prior to the move)?

src/AssetsController.php
```diff
@@ -0,0 +1,151 @@
...
+		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
+		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
+	}
+
+	/**
+	 * Register block scripts & styles.
+	 *
+	 * @since 2.5.0
+	 * Moved data related enqueuing to new AssetDataRegistry class as part of ongoing refactoring.
```
nerrad: This comment block isn't really needed now?

src/BlockTypesController.php
```diff
@@ -8,45 +8,70 @@
...
  * @internal
  */
-class Library {
+class BlockTypesController {
 
 	/**
-	 * Initialize block library features.
+	 * Instance of the asset API.
+	 *
+	 * @var AssetApi
```
nerrad, line 21: We should include the `@since $VID:$` tag here.

src/AssetsController.php
```diff
@@ -0,0 +1,151 @@
...
+use Automattic\WooCommerce\Blocks\Package;
+use Automattic\WooCommerce\Blocks\Assets\Api as AssetApi;
+use Automattic\WooCommerce\Blocks\Assets\AssetDataRegistry as AssetDataRegistry;
+
+/**
+ * AssetsController class.
+ *
+ * @internal
+ */
+class AssetsController {
```
nerrad: Should we make this final?

src/BlockTypesController.php
```diff
@@ -8,45 +8,70 @@
...
 
 /**
- * Library class.
- * Initializes blocks in WordPress.
+ * BlockTypesController class.
  *
  * @internal
  */
-class Library {
+class BlockTypesController {
```
nerrad: Should we make this final?

src/BlockTypesController.php
```diff
@@ -69,20 +94,20 @@ public static function register_blocks() {
 		];
 
 		if ( Package::feature()->is_feature_plugin_build() ) {
-			$blocks[] = 'Checkout';
-			$blocks[] = 'Cart';
+			$block_types[] = 'Checkout';
+			$block_types[] = 'Cart';
```
nerrad, line 111: 👍 to the variable name here

nerrad (commented):
Code wise this looks good, thanks Mike! 

I have a few items for you to look as inline comments that I'm interested in your take on.  I also [have this comment I made earlier](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4094#issuecomment-824968302) that I think should be addressed.

mikejolley:
@nerrad thanks, changes made. Just waiting on tests :)

src/Library.php
```diff
@@ -0,0 +1,44 @@
+<?php
+namespace Automattic\WooCommerce\Blocks;
+
+use Automattic\WooCommerce\Blocks\BlockTypes\AtomicBlock;
+use Automattic\WooCommerce\Blocks\Package;
+use Automattic\WooCommerce\Blocks\Assets\AssetDataRegistry;
+use Automattic\WooCommerce\Blocks\Assets\Api as AssetApi;
+use Automattic\WooCommerce\Blocks\Integrations\IntegrationRegistry;
```
nerrad: probably don't need all these imports now?

nerrad (approved):
Changes look good, just a minor note about the unnecessary imports in the deprecated Library class.

Good to go after tests are 🟢  and you're happy with your own testing of this.

src/Library.php
nerrad: Ignore this, it was an artifact of the commits I was viewing and how the changes were being shown, I see they don't actually exist in the file.

PR #4097, opened 2021-04-23
Remove work-around to load our stylesheets after wp-edit-post

### LINKED ISSUE #3225: Remove fix to load our stylesheets after editor CSS. See ...
Remove fix to load our stylesheets after editor CSS. See #3068 for the rationale of this fix. It should be no longer github.com/WordPress/gutenberg/issues/20797).

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e07de12ed23db91cd977be4423132f3151573d87/src/Assets.php#L39-L51
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in e07de12ed23db91cd977be4423132f3151573d87 when #3219 was merged. cc @Aljullu

Aljullu:
Closes #3225.

### Background

In the past, some of the styles of the Checkout block were conflicting with those from the editor, that affected the Country selector, which was rendering with a smaller size in the editor than in the frontend (see #3068 for more details). To fix that, in #3219 we forced our styles to be loaded after `wp-edit-post` styles in order to fix the specificity battle.

However, I recently tested it and it looks like we no longer need that. So this PR takes care of removing that work-around from the code.

### Screenshots

imatge

1. Open the Checkout block in the editor and verify the `Country/Region` selector has the same high as the `City` field next to it, rounded corners, etc.
2. Bonus points if you can test it in different versions of WordPress (I tested 5.5 and 5.7 with and without Gutenberg) and different themes.

### DIFF
--- a/src/AssetsController.php
+++ b/src/AssetsController.php
@@ -37,14 +37,13 @@ final class AssetsController {
 		add_action( 'init', array( $this, 'register_assets' ) );
 		add_action( 'body_class', array( $this, 'add_theme_body_class' ), 1 );
 		add_action( 'admin_body_class', array( $this, 'add_theme_body_class' ), 1 );
-		add_action( 'wp_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
-		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_scripts' ) );
 	}
 
 	/**
 	 * Register block scripts & styles.
 	 */
 	public function register_assets() {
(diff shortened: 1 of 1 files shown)


mikejolley:
Changes look good @Aljullu, I wasn't sure if this was still needed. Just needs rebase (move changes to AssetsController).

Aljullu:
Rebased. :ok_hand:

mikejolley (approved):

PR #4098, opened 2021-04-23
Convert Product components to TS

opr:
Note this also includes some changes from #4091 - I plan to rebase when that has been merged.

This PR converts the following components to TypeScript:
- `Policies`
- `ProductBackorderBadge`
- `ProductBadge`
- `ProductDetails`
- `ProductLowStockBadge`
- `ProductImage`
- `ProductMetadata`
- It also adds a new type: `ProductResponseItemData` which defines how `itemData` on a cart item is shaped. (This is used by things like Product Add-Ons).

1. To begin make three products:
1. To make the Low Stock badge appear, edit a product and tick **Inventory > Enable stock management at product level** and make the low stock threshold lower than the current stock level.
2. To make a Product Backorder badge appear edit a product and check **Inventory > Enable stock management at product level** and make the stock level 0, In `Allow backorders` choose `Allow, but notify customer`
3. Add some Add-ons to a product using Product Add-ons.
4. Add each of these products to your cart and go to the cart block, ensure the cart works and all content appears as expected.


assets/js/base/components/cart-checkout/product-details/index.tsx
```diff
@@ -1,17 +1,20 @@
...
 import { decodeEntities } from '@wordpress/html-entities';
+import type { ProductResponseItemData } from '@woocommerce/type-defs/product-response';
 
 /**
  * Internal dependencies
  */
 import './style.scss';
 
+interface ProductDetailsProps {
+	details: ProductResponseItemData[];
```
Aljullu, line 14: According to this component's code, the `name` property of each item is optional, but the `ProductResponseItemData` type has it as always existing. I didn't investigate which one is right, but I think it should be consistent?

Aljullu: Hmm, I see the issue was there in the PropTypes as well. Looking at the code I think they were wrong. :thinking:

assets/js/base/components/cart-checkout/product-low-stock-badge/index.tsx
```diff
@@ -8,6 +8,9 @@ import { __, sprintf } from '@wordpress/i18n';
  */
 import ProductBadge from '../product-badge';
 
+interface ProductLowStockBadgeProps {
```
Aljullu, line 11: This type is not used in the component definition.

assets/js/base/components/cart-checkout/product-metadata/index.tsx
```diff
@@ -10,6 +11,13 @@ import ProductDetails from '../product-details';
 import ProductSummary from '../product-summary';
 import './style.scss';
 
+interface ProductMetadataProps {
```
Aljullu, line 14: Same here. :slightly_smiling_face:

Aljullu (commented):
Thanks for working on porting all these components to TypeScript! Overall looking good, I left some comments inline of things that I think need to be tweaked. Also, there are several `Missing return type on function.` warnings in the code view in this PR.

assets/js/base/components/cart-checkout/product-details/index.tsx
opr: I think the `name` property is not actually optional, the code just caters for it being an empty string. If you look here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4098/files#diff-61ea5241262bf8f97af73e78e1d15467091815e69a744abb33b04dfa2e0ec098R40, allowing `name` to be undefined would result in trying to concatenate undefined with some other string. 

The `name` could be an empty string, that's where this ternary operator comes in:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4098/files#diff-61ea5241262bf8f97af73e78e1d15467091815e69a744abb33b04dfa2e0ec098R33

Aljullu: Aaah, makes sense, thanks!

Aljullu (approved):
Code changes look good. I also did a round of testing and everything seems to be working fine. :ship:

PR #4135, opened 2021-04-26
Convert shipping components to typescript

opr:
This PR will convert the following components to TypeScript:
- `ShippingCalculatorAddress`
- `ShippingCalculator`
- `ShippingLocation`
- `ShippingRatesControl`
- `ShippingOptionsStep`

There were a few instances where the type `Rate` was used, but this was the same as a `CartShippingRateItemShippingRate` so I removed it.

1. General smoke testing of cart and checkout blocks, especially focussing on the shipping aspect of these blocks


assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
```diff
@@ -7,9 +7,10 @@ import { decodeEntities } from '@wordpress/html-entities';
 import Label from '@woocommerce/base-components/label';
 import Title from '@woocommerce/base-components/title';
 import type { ReactElement } from 'react';
-import type { Rate, PackageRateOption } from '@woocommerce/type-defs/shipping';
+import type { PackageRateOption } from '@woocommerce/type-defs/shipping';
 import { Panel } from '@woocommerce/blocks-checkout';
 import { useSelectShippingRate } from '@woocommerce/base-context/hooks';
+import type { CartShippingRateItemShippingRate } from '@woocommerce/type-defs/cart';
```
mikejolley: This gets quite confusing naming wise when we start throwing in "item" since item can refer to many things. Would it make sense to stick with just "package" and "rate" terminology? e.g. CartShippingPackage which contains CartShippingPackageRate[]? Or something along those lines.

assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
```diff
@@ -34,21 +35,28 @@ interface Destination {
 	country: string;
 }
 
+export interface PackageData {
+	destination: Destination;
+	name: string;
+	// eslint-disable-next-line camelcase
+	shipping_rates: CartShippingRateItemShippingRate[];
```
mikejolley: This wasn't changed in this PR, but do you think we should translate shipping_rates to shippingRates after receiving data from the API or leave in snake case?

assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
```diff
@@ -34,21 +35,28 @@ interface Destination {
...
-	renderOption: ( option: Rate ) => PackageRateOption;
-	collapse: boolean;
-	packageData: {
-		destination: Destination;
-		name: string;
-		// eslint-disable-next-line camelcase
-		shipping_rates: Rate[];
-		items: PackageItem[];
-	};
+	/* PackageId can be a string, WooCommerce Subscriptions uses strings for example, but WooCommerce core uses numbers */
```
mikejolley, line 51: Subscriptions may not be the only fwiw. Core allows strings but uses numbers internaly.

assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
```diff
@@ -12,16 +11,78 @@ import {
...
 /**
  * Internal dependencies
  */
-import ShippingRatesControlPackage from '../shipping-rates-control-package';
+import ShippingRatesControlPackage, {
+	PackageRateRenderOption,
+} from '../shipping-rates-control-package';
+
+interface PackagesProps {
+	// eslint-disable-next-line camelcase
```
mikejolley: I don't think this is needed - next line is not invalid.

assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
```diff
@@ -12,16 +11,78 @@ import {
...
+	showItems?: boolean;
+	noResultsMessage: ReactElement;
+	renderOption: PackageRateRenderOption;
+}
 
 /**
- * @typedef {import('react')} React
+ * Renders multiple packages within the slotfill.
+ *
+ * @param {Object} props Incoming props.
```
mikejolley, line 36: Is this docblock needed now Packages is typed?

assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
```diff
@@ -40,7 +101,7 @@ const ShippingRatesControl = ( {
 	collapsible = false,
 	noResultsMessage,
 	renderOption,
-} ) => {
+}: ShippingRatesControlProps ) => {
```
mikejolley: There is an eslint warning here—missing return type on the function.

assets/js/blocks/cart-checkout/checkout/form/shipping-options-step.tsx
```diff
@@ -50,7 +54,7 @@ const renderShippingRatesControlOption = ( option ) => {
 	};
 };
 
-const ShippingOptionsStep = () => {
+const ShippingOptionsStep = (): JSX.Element | null => {
```
mikejolley, line 57: JSX.Element vs ReactNode - which is better? `ReactNode` covers null types.

assets/js/type-defs/customer.ts
```diff
@@ -0,0 +1,52 @@
...
+	city: string;
+	state: string;
+	postcode: string;
+	country: string;
+}
+
+export type ShippingAddress = Address;
+export type BillingAddress = Address;
+
+export interface AddressFieldConfiguration {
```
mikejolley: This type exists in js/settings/shared/default-address-fields.js - should we import from there instead? Same for the other types in this file.

mikejolley (commented):
Left some feedback. I notice we have some duplicate types and I'm not sure if this is intentional.

assets/js/blocks/cart-checkout/checkout/form/shipping-options-step.tsx
opr: I initially started with ReactNode, but then ran into issues like this:

image

Also, when the return type is not explicitly specified, TypeScript infers that it is is a `JSX.Element`. I would prefer to stick with that on the basis that it doesn't break, and TS thinks it's that anyway.

assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
opr: Sure, thanks 😄 This was just an example of a time when they are strings explaining why it was typed as `string | number`

assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
opr: Not _needed_ but  I don't think we should actively remove them. It being there gives more explanation of each prop so I don't see its existence as a negative thing.

assets/js/type-defs/customer.ts
opr: Cool, I added this file as part of another (WIP) PR and copied it over, looks like your changes made it through before mine did. I'm happy to remove this file, however I like the type aliases for `BillingAddress` and `ShippingAddress` and `KeyedAddressFieldConfiguration` I added. I'll rename and move them over.

opr: Actually, on second thought, we need to differentiate between the `Address` with the entered values in, and the `Address` with configuration data in. This is why I had both `Address` and `AddressFields`. I will add the type back that is the address with string values entered by the user.

assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
opr: I've removed the word Item where it seemed appropriate and replaced it with `Package` and `Rate` - is this what you had in mind?

assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx
opr: I think we should - but not in this PR, I'm not sure how long it'll take to find all the places this happens in, and this PR is pretty big already. So, I'll make a new issue for it to be done next cooldown.

mikejolley (approved):
Changes good. I've re-ran tests so this can merge once passed.

PR #4143, opened 2021-04-28
Add inline console warning if shared settings hook is consumed

### LINKED ISSUE #4137: `woocommerce_shared_settings` deprecation notice
[The shared settings notice was deprecated](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4092), but it still used by WC Admin.

We need to decide if we can:

- Leave notice in place and add dev note somewhere
- Soft-deprecate the filter (no notice) and inform teams before hard deprecating the filter
- Deprecate with a notice only in feature plugin context
- Leave filter in place, supported, but just don't use it in blocks internally.

mikejolley:
Fixes #4137

### Screenshots

Screenshot 2021-04-28 at 11 49 26

1. Visit an admin screen or checkout
2. View warning in console

This hook is consumed by woocommerce-admin so will be consumed until changed in that project.

### DIFF
--- a/src/Assets/AssetDataRegistry.php
+++ b/src/Assets/AssetDataRegistry.php
@@ -202,10 +202,7 @@ class AssetDataRegistry {
 		// Surface a deprecation warning in the error console.
 		if ( has_filter( 'woocommerce_shared_settings' ) ) {
 			$error_handle  = 'deprecated-shared-settings-error';
-			$error_message = __(
-				'`woocommerce_shared_settings` filter in Blocks is deprecated. See https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/contributors/block-assets.md',
-				'woo-gutenberg-products-block'
-			);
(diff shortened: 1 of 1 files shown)


src/Assets/AssetDataRegistry.php
```diff
@@ -197,15 +197,25 @@ protected function initialize_core_data() {
...
+			$error_message = __(
+				'`woocommerce_shared_settings` filter in Blocks is deprecated. See https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/contributors/block-assets.md',
+				'woo-gutenberg-products-block'
+			);
+			// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.NotInFooter,WordPress.WP.EnqueuedResourceParameters.MissingVersion
+			wp_register_script( $error_handle, '' );
+			wp_enqueue_script( $error_handle );
+			wp_add_inline_script(
+				$error_handle,
+				sprintf( 'console.warn( "%s" );', $error_message )
```
nerrad, line 211: Given the `$error_message` is a translated string, do we need to escape it to catch potential malicious translations (maybe just do `esc_html__` for the php translation function?)

nerrad (approved):
Thanks Mike, this looks good. I do have a comment about escaping the translated string. I'll pre-approve so you can go ahead and merge once that's addressed.

PR #4145, opened 2021-04-28
Add "Filter Products by Stock" block

### LINKED ISSUE #4096: Stock status filter feature
No problems!

Created stock status filter in our fork, we just need to understand is it possible to contribute our changes to this repo and what we need to do for that. For example filter, I examine attribute filter.

## Additional context
The editor side is like another filter. For now, it has an attribute filter icon :)
And it gets disabled if used once.

image

The editor side rendered block.
image

Block inspector controls.
image

And finally the front side combination with other filters. ( Added to active filter too. )
image

grogou:
This PR introduces a new block: "Filter Products by Stock". When this is used in conjunction with the `All Products` block, shoppers will be able to filter out products by different stock levels. This block supports custom stock levels too.

Fixes #4096 

#### Accessibility

### Screenshots

image
image

1. Add 'Filter Products by Stock", "All Products" and "Active Filters" blocks to a page.
2. Ensure `Hide out of stock products from catalogue` is **disabled** in **WooCommerce > Settings > Products > Inventory**
2. Set the stock status of some products to `out of stock`, some to `on backorder`.
3. Go to the page on the front end and verify that you can filter products by the three different stock levels.
4. For any product that is `out of stock` change it back to `in stock`. All products should now either be `on backorder` or `in stock`.
5. Reload the page and ensure the option to filter out of stock products is not there.
5. Set some products back to being `out of stock`.
6.  Change the `Hide out of stock products from catalogue` to **enabled** in **WooCommerce > Settings > Products > Inventory**
7. Reload the page and ensure the option to filter out of stock products is not there.
8. Add the `Filter Products by Price` and `Filter Products by Attribute` block to the page. 
9. Verify that the filters work together and the correct products are shown in the `All Products` block.

### Changelog

>Added "Filter Products by Stock" block


senadir:
👋🏼  I will (along side others) review your PR to help you get it merged.

It seems you're running npm 7+ which caused the package-lock.json to be updated, can you downgrade to npm 6 (which ever is installed with node 14), delete package-lock.json and run `npm install` again.
Another option is not commit the lock file at all since your PR isn't installing any new dependencies.

grogou:
Done!

grogou:
@senadir we are pulled new changes from trunk and fixed conflicts for stock filter.

senadir:
Hey @grogou, sorry for taking so long to come back at you on this PR. I did start a review on this, functionally, everything seems to work fine. Code wise, I didn't finish my review yet (about 70% left). I'm moving slowly due to other higher priority tasks and the fact that this part of codebase hasn't been touched for the past two years, so I need to brush up on my knowledge of it.

assets/js/base/components/product-list/product-list.js
```diff
@@ -57,14 +57,31 @@ const generateQuery = ( {
...
+	if ( hideOutOfStockItems ) {
+		let newStockStatus = [ 'instock', 'onbackorder' ];
+		if ( productStockStatus.length > 0 ) {
+			newStockStatus = productStockStatus;
+			if ( productStockStatus.includes( 'outofstock' ) ) {
+				newStockStatus = productStockStatus.filter(
+					( slug ) => slug !== 'outofstock'
+				);
+			}
+		}
```
opr: Let's try to find a different solution to this - right now this forces the `instock` and `onbackorder` stock options to be applied, and there's no way to remove them.

I understand what this is trying to achieve, but I think the UX can be improved. If you try to uncheck both options, then they both get automatically reapplied, this honestly confused me to begin with, and I think a shopper would be equally confused.

Ping @ralucaStan because #3859 added this and wondered if you have any thoughts on other ways around this. Attached is a screenshot of how it looks on the front-end, you cannot uncheck both boxes, only one at a time.

image

I think it'd be better if they were (somehow) both unchecked to begin with, similar to how the Filter by Attribute block works.

assets/js/blocks/stock-filter/block.js
```diff
@@ -0,0 +1,285 @@
...
+ */
+import Label from './label';
+import { previewOptions } from './preview';
+import './style.scss';
+
+const STOCK_STATUS_OPTIONS = getSetting( 'stockStatusOptions', [] );
+const hideOutOfStockItems = getSetting( 'hideOutOfStockItems', false );
+if ( hideOutOfStockItems ) {
+	delete STOCK_STATUS_OPTIONS.outofstock;
+}
```
opr: This seems to be working OK, it doesn't actually affect anything except what gets rendered as a selectable option, right?

assets/js/base/components/product-list/product-list.js
```diff
@@ -57,14 +57,31 @@ const generateQuery = ( {
 		}
 	};
 
+	// eslint-disable-next-line react-hooks/rules-of-hooks
```
opr: Maybe we could pass in `productStockStatus` to the `generateQuery` function so we don't need to disable this eslint rule? We'd also have to update the code below to directly assign to `productStockStatus` rather than using the `setProductStockStatus` function?

assets/js/base/components/product-list/product-list.js
```diff
@@ -140,6 +157,10 @@ const ProductList = ( {
 		'attributes',
 		[]
 	);
+	const [ productStockStatus, setProductStockStatus ] = useQueryStateByKey(
+		'stock_status',
+		[]
+	);
```
opr: As mentioned above, should we move this further up the function, and pass `productStockStatus` to `getQuery`?

assets/js/blocks/stock-filter/block.js
```diff
@@ -0,0 +1,285 @@
...
+		return productStockStatusQuery;
+	}, [ productStockStatusQuery ] );
+
+	const currentCheckedQuery = useShallowEqual( checkedQuery );
+	const previousCheckedQuery = usePrevious( currentCheckedQuery );
+	// Track Stock query changes so the block reflects current filters.
+	useEffect( () => {
+		if (
+			! isShallowEqual( previousCheckedQuery, currentCheckedQuery ) && // checked query changed
+			! isShallowEqual( checked, currentCheckedQuery ) // checked query doesn't match the UI
```
opr: Making a suggestion to update these comments to use sentence case 🙂 

```suggestion
			! isShallowEqual( previousCheckedQuery, currentCheckedQuery ) && // Checked query changed.
			! isShallowEqual( checked, currentCheckedQuery ) // Checked query doesn't match the UI.
```

assets/js/base/components/product-list/product-list.js
opr: This also causes trouble with screen readers, if you remove the `On backorder` filter (from the Active filters block) and then remove the `In stock` filter, then the screen reader announces "Removed in stock filter" but the actual state of the app shows that both stock status filters are applied.

assets/js/blocks/stock-filter/block.js
```diff
@@ -0,0 +1,285 @@
...
+							/* translators: %1$s and %2$s are stock statuses (for example: 'instock'...). */
+							__(
+								'%1$s filter replaced with %2$s.',
+								'woo-gutenberg-products-block'
+							),
+							filterAddedName,
+							filterRemovedName
+						)
+					);
+				} else if ( filterAddedName ) {
```
opr: I don't think this is needed, since the stock filters are never replaced (only removed, and re-added). It looks like this logic exists only for filters when they're set to use a `AND` option, this block doesn't have that option.

assets/js/blocks/stock-filter/edit.js
```diff
@@ -0,0 +1,131 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+
+import { InspectorControls } from '@wordpress/block-editor';
```
opr: Suggesting to remove this empty line.

```suggestion
import { __ } from '@wordpress/i18n';
import { InspectorControls } from '@wordpress/block-editor';
```

assets/js/blocks/stock-filter/label.js
```diff
@@ -0,0 +1,41 @@
...
+					wrapperProps={ {
+						className: 'wc-block-stock-filter-list-count',
+					} }
+				/>
+			) }
+		</>
+	);
+};
+
+export default StockFilterLabel;
```
opr: I think it would be better to take this and `AttributeFilterLabel` and make a more generalised version of this component?

src/StoreApi/Schemas/ProductCollectionDataSchema.php
```diff
@@ -98,6 +98,29 @@ public function get_properties() {
...
+				'description' => __( 'Returns number of products with each stock status.', 'woo-gutenberg-products-block' ),
+				'type'        => [ 'array', 'null' ],
+				'context'     => [ 'view', 'edit' ],
+				'readonly'    => true,
+				'items'       => [
+					'type'       => 'object',
+					'properties' => [
+						'status' => [
+							'description' => __( 'Status', 'woo-gutenberg-products-block' ),
+							'type'        => 'integer',
```
opr: Should this be `string`?

opr (changes requested):
Thanks for making this PR @grogou 👏🏼! Really great work.

In terms of functionality, it works really well for the most part, but I have noticed an issue relating to UX. At the moment, when adding this block to a page while the option: `Hide out of stock items from the catalog` is true, then the `instock` and `onbackorder` stock statuses are selected, and it's impossible to unselect them both. This looks like it's caused by [this block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4145/files#diff-3a616c1f74d8a938addc8e6a5bf365304512ce5cc9b16ad4bace0611911ec328R66-R75) - I understand this is just a refactor of something that was added in #3859 so I'm not blaming you for this 😁 

I'll raise this with the team, and see if there is a potential workaround, however, to manage your expectations: given that our team has a lot of other high-priority work on right now I can't promise a time that we will get back to you on this, sorry.

I also think you're going to be opening another PR for a `Filter by category` block, right? If so, then I think my suggestion of generalising the `AttributeFilterLabel` and `StockFilterLabel` components should be considered even more strongly.

Thanks again for creating this PR! 🙌🏼

assets/js/blocks/stock-filter/block.js
```diff
@@ -0,0 +1,285 @@
...
+import './style.scss';
+
+const STOCK_STATUS_OPTIONS = getSetting( 'stockStatusOptions', [] );
+const hideOutOfStockItems = getSetting( 'hideOutOfStockItems', false );
+if ( hideOutOfStockItems ) {
+	delete STOCK_STATUS_OPTIONS.outofstock;
+}
+const initialOptions = Object.entries(
+	STOCK_STATUS_OPTIONS
+).map( ( [ slug, name ] ) => ( { slug, name } ) );
```
nerrad: One problem with doing this is it mutates the `stockStatusOptions` object. So anywhere in the code that `getSetting( 'stockStatusOptions' )` is called will have the mutated object. We don't force immutability of values returned from `getSetting` (something we might want to consider doing in future iterations). 

A better option here might be to to do something like this to remove the `outofstock` property via something like:

```suggestion
const hideOutOfStockItems = getSetting( 'hideOutOfStockItems', false );
const { outofstock, ...otherStockStatusOptions } = getSetting( 'stockStatusOptions', {} );
let STOCK_STATUS_OPTIONS = hideOutOfStockItems ? otherStockStatusOptions : { outofstock, ...otherStockStatusOptions };
// filter added to handle if there are slugs without a corresponding name defined.
const initialOptions = Object.entries(
	STOCK_STATUS_OPTIONS
).map( ( [ slug, name ] ) => ( { slug, name } ) ).filter( ( status ) => !! status.name );
```

assets/js/base/components/product-list/product-list.js
grogou: This case is fixed. I do it from back side with MySql, in my opinion it is better solution for this case, we don't need to do query from JS to hide out of stocks, we just need to send filtered data if hide option is on. And this fix is fixed UX issue.

assets/js/blocks/stock-filter/block.js
grogou: Related to this https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4145#discussion_r664799423 suggestion.

assets/js/base/components/product-list/product-list.js
grogou: New solution with sql fixed this case.

assets/js/base/components/product-list/product-list.js
grogou: Related to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4145#discussion_r669743408

assets/js/blocks/stock-filter/block.js
grogou: Fixed as you suggested!

assets/js/blocks/stock-filter/block.js
grogou: Fixed as you suggested!

assets/js/blocks/stock-filter/edit.js
grogou: Fixed as you suggested!

assets/js/blocks/stock-filter/label.js
grogou: For this case created `filter-element-label` component and change both filters label component as you suggested.

src/StoreApi/Schemas/ProductCollectionDataSchema.php
grogou: Fixed!

assets/js/blocks/stock-filter/block.js
grogou: Yes it is my bad, this case fixed as you suggested. Just changed `let STOCK_STATUS_OPTIONS` to `const` and sorted initial options alphabetically fro better UI performance.

src/StoreApi/Utilities/ProductQuery.php
```diff
@@ -317,6 +317,9 @@ public function add_query_clauses( $args, $wp_query ) {
 		if ( $wp_query->get( 'stock_status' ) ) {
 			$args['join']   = $this->append_product_sorting_table_join( $args['join'] );
 			$args['where'] .= ' AND wc_product_meta_lookup.stock_status IN ("' . implode( '","', array_map( 'esc_sql', $wp_query->get( 'stock_status' ) ) ) . '")';
+		} elseif ( 'yes' === get_option( 'woocommerce_hide_out_of_stock_items' ) ) {
+			$args['join']   = $this->append_product_sorting_table_join( $args['join'] );
+			$args['where'] .= ' AND wc_product_meta_lookup.stock_status NOT IN ("outofstock")';
```
opr, line 324: cc @nerrad - During a 1:1, we discussed whether the "hide out of stock products" should be a block attribute - to enable merchants to control blocks and whether they may display products that are out of stock, even if this is contrary to the global store setting. I cannot remember the outcome of our discussion, or if we even had a solid decision, so I'm pinging you here for your thoughts.

In its current state (in this PR) the product query is essentially "hard coded" to respect the store setting, so merchants will not be able to show any products that are out of stock.

If we _do_ want to add this as a block attribute, I'd argue that we should do it in a different PR rather than here as it would significantly increase the scope and complexity of this PR. What do you think?

assets/js/blocks/stock-filter/block.js
```diff
@@ -0,0 +1,278 @@
...
+
+const hideOutOfStockItems = getSetting( 'hideOutOfStockItems', false );
+const { outofstock, ...otherStockStatusOptions } = getSetting(
+	'stockStatusOptions',
+	{}
+);
+const STOCK_STATUS_OPTIONS = hideOutOfStockItems
+	? otherStockStatusOptions
+	: { outofstock, ...otherStockStatusOptions };
+// filter added to handle if there are slugs without a corresponding name defined.
```
opr:
```suggestion
// Filter added to handle if there are slugs without a corresponding name defined.
```
Sorry for the nitpick!

assets/js/blocks/attribute-filter/style.scss
```diff
@@ -25,14 +16,10 @@
...
-
-		.wc-block-attribute-filter-list-count {
-			float: right;
-		}
 	}
 
-	.is-single .wc-block-attribute-filter-list-count,
-	.wc-block-dropdown-selector .wc-block-dropdown-selector__list .wc-block-attribute-filter-list-count {
+	.is-single,
+	.wc-block-dropdown-selector .wc-block-dropdown-selector__list {
```
opr: Oops, looks like this change inadvertently affects dropdown filters - this is what I'm seeing on my test site with an attribute filter dropdown.

opr (commented):
Thanks for addressing the feedback @grogou! The new `FilterElementLabel` looks and works great.

I've left a couple of comments in this round, one specifically about styling relating to changes made with the attribute filter, and one that needs input from @nerrad - the SQL solution you've implemented now works really well, but I'm not sure if we want to impose this hard limit on the merchant, or if we want to allow them to choose to hide out of stock products on a block-by-block basis.

assets/js/blocks/attribute-filter/style.scss
grogou: I brought back it to old state․ Sorry for inattentive part.

grogou: This happened because my background is white))

image

src/StoreApi/Utilities/ProductQuery.php
nerrad: > If we do want to add this as a block attribute, I'd argue that we should do it in a different PR rather than here as it would significantly increase the scope and complexity of this PR. What do you think?

That seems like a fair approach. I trust your judgement here. I'd suggest creating a followup issue though to prevent it getting forgotten if the PR can't be done right away.

opr:
Howdy @garymurray and @vivialice - pinging you here to ask your opinion from a product/UX point of view.

Right now this block strictly obeys the `Hide out of stock items from the catalog` setting from WC, and does not let the customer choose to filter by "Out of stock" products at all.

I was wondering if we are OK with this, _or_ if we want a block-level option on the "Filter by stock" block to allow shoppers to select "Out of stock" products. If they did, then this would override the store setting.

vivialice:
Hey! @opr are you saying we could add a setting on the block level for merchants to allow customers to filter out of stock items? I'm not against this. As long are we aren't showing that filter option without the merchant knowing/explicitly agreeing to it. Showing out of stock items could be useful in conjunction with stock alerts - not sure if these exist already but assuming they do via extension at least.

opr:
> Hey! @opr are you saying we could add a setting on the block level for merchants to allow customers to filter out of stock items? I'm not against this. As long are we aren't showing that filter option without the merchant knowing/explicitly agreeing to it. Showing out of stock items could be useful in conjunction with stock alerts - not sure if these exist already but assuming they do via extension at least.

Hey @vivialice thanks for answering so quickly! Yes, so the setting would exist on the "Filter Products by Stock" block, and that would control whether it displays like this:

| Off | On |
|---|---|
| image |  image |

This option would only be available in the block settings if the core `Hide out of stock items from the catalog` is enabled/`true`.

garymurray:
> Right now this block strictly obeys the Hide out of stock items from the catalog setting from WC, and does not let the customer choose to filter by "Out of stock" products at all.

> I was wondering if we are OK with this, or if we want a block-level option on the "Filter by stock" block to allow shoppers to select "Out of stock" products. If they did, then this would override the store setting.

Obeying the core setting is the right approach for now imo - otherwise we creating conflicting settings. Ultimately the long term decision needs to be made around what options are made at the block level and which ones at the core level when blocks are the default way to build your site.

So, for now, if you enable the `Hide out of stock items from the catalog` setting in WC Core - then there should not be an option to filter by out-of-stock items within the block - however if you leave `Hide out of stock items from the catalog` unchecked, then the option to filter by out-of-stock items should exist.

opr:
> Ultimately the long term decision needs to be made around what options are made at the block level and which ones at the core level when blocks are the default way to build your site.

Thanks @garymurray & @vivialice - with this in mind then, maybe we can go ahead and merge this as-is (following a final review round), I'll P2/create a GH issue to track the block-level setting option. Sound like a plan?

garymurray:
@opr Sounds good.

src/StoreApi/Utilities/ProductQueryFilters.php
```diff
@@ -47,6 +47,72 @@ public function get_filtered_price( $request ) {
...
+	/**
+	 * Get stock status counts for the current products.
+	 *
+	 * @param \WP_REST_Request $request The request object.
+	 * @return array status=>count pairs.
+	 */
+	public function get_stock_status_counts( $request ) {
+		global $wpdb;
+		$product_query         = new ProductQuery();
+		$stock_status_options  = array( 'instock', 'outofstock', 'onbackorder' );
```
opr:
```suggestion
		$stock_status_options  = array_map( 'esc_sql', array_keys( wc_get_product_stock_status_options() ) );
```

If you make this change, it will allow us to get custom stock statuses.

opr (changes requested):
Hey @grogou this is excellent, we're so close to being able to merge this! Again thanks for addressing my feedback earlier.

I've made a suggestion that will allow custom stock statuses to be used, and one to ensure our SQL query for stock status counts is sanitised. 

By the way, to add custom stock statuses for testing, please add this code to your theme's `functions.php` file (or wherever suits you)

```php
	// Add new statuses

	return $status;
}
```

Maybe you could rebase onto `trunk` too to resolve the conflict in `assets/js/blocks/attribute-filter/label.js` ?

As soon as this is addressed I will approve and merge this PR 🥳

src/StoreApi/Utilities/ProductQueryFilters.php
```diff
@@ -47,6 +47,72 @@ public function get_filtered_price( $request ) {
...
+	 *
+	 * @return false|string
+	 */
+	private function generate_stock_status_count_query( $status, $product_query_sql, $stock_status_options ) {
+		if ( ! in_array( $status, $stock_status_options, true ) ) {
+			return false;
+		}
+		global $wpdb;
+
+		return "
```
opr:
```suggestion
		global $wpdb;
		$status = esc_sql( $status );
		return "
```

grogou:
Hey @opr. Thank you so much for your job on this PR! Your suggestions is commited!

opr:
@grogou awesome I see you've actioned those suggestions, thank you! I edited my comment earlier but you may have missed it, my final suggestion was:

Maybe you could rebase this branch onto `trunk` too to resolve the conflict in `assets/js/blocks/attribute-filter/label.js` ?

grogou:
Sure, on it.

grogou:
@opr sorry, but rebase is take to long to do it, the package-lock.json always get conflicted, so maybe you can resolve it from your side?

opr (approved):

grogou:
@opr  conflict resolved! Thanks for your help!

opr:
@grogou Hello! Seems like something has gone a bit wrong while we've been trying to resolve this conflict. I think you'll need to reset your branch to a known good commit (Maybe e7bd0e6e1e171a041ac17b8b11abc812585f7645 ?) and force-push it to your remote.

Then you'll need to reapply the changes you made to resolve the conflict and push them to your remote again. After this is done, we should be in a position to merge into `trunk`. Sorry for the confusion here!

grogou:
@opr I think we done! finally.

opr:
The JS/CSS linting job is failing because of `Error: Resource not accessible by integration` - this is because it is a PR from a fork. I have run the linting job locally and can confirm it is fine.

@grogou Thank you for all of your efforts on this PR, and thank you for your patience while we reviewed it. This block is a great addition. Congratulations on getting it merged! 🥳  🚀

grogou:
@opr Thank you too for help me do this PR. Think we will work together again.

grogou:
@opr and we forgot about block icon, do we need to change it to another one?

opr:
> @opr and we forgot about block icon, do we need to change it to another one?

Oh, yes good idea, since this PR has been merged you'll need to open another one with the changes! @vivialice any ideas what icon we should use for this new block?

opr:
@grogou FYI your block is now released into the wild in WooCommerce Blocks 5.8.0 :) https://developer.woocommerce.com/2021/08/31/woocommerce-blocks-5-8-0-release-notes/

Thanks so much for the work you put into this and for being patient while we reviewed it!

grogou:
@opr Thanks, it was pleasure for me to communicate with WC team and do my first contribution, me and our team are going to have more contribution in the near future.

PR #4158, opened 2021-04-30
Improvements to `emitEventWithAbort`.

### LINKED ISSUE #4048: Improve `emitEventWithAbort` handling of "success" type responses
The intended purpose of the [`emitEventWithAbort` emitter](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f0fc6c5fff2dcc5ee2ddb6ac76eec13b86701afa/assets/js/base/context/shared/event-emit/emitters.ts#L56) was to allow for immediately surfacing responses that have errors and skipping any other registered subscribers until the errors are handled. 

However, it's currently designed so that any object with a `type` property aborts the loop:

```js
		try {
			const response = await Promise.resolve( observer.callback( data ) );
			if ( typeof response !== 'object' || response === null ) {
				continue;
			}
			if ( ! response.hasOwnProperty( 'type' ) ) {
				throw new Error(
					'If you want to abort event emitter processing, your observer must return an object with a type property'
				);
			}
			return response;
		} catch ( e ) {
			// We don't handle thrown errors but just console.log for troubleshooting.
			// eslint-disable-next-line no-console
			console.error( e );
			return { type: 'error' };
		}
	}
```

The problem is this makes (what I think) is a faulty assumption that if a subscribed callback returns success as the response type that it means no other subscribed callbacks need to execute. Currently this creates a gotcha that allows extensions to effectively prevent any other registered extension from reacting to the checkout response unintentionally by returning a success type. An example would be one payment method having handling for 3DS payments in which it correctly aborts it's processing early if the checkout response doesn't indicates it doesn't need it (by returning `{ type: success }` but then the checkout was actually submitted for _a different payment method_ that _also_ has a subscribed event handler for 3DS processing.

On the surface, the solution appears to be simply modifying the logic of the emitter so that `{type: success}` doesn't abort early from the loop, however, implementations of this emitter (for [`onPaymentProcessing`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f0fc6c5fff2dcc5ee2ddb6ac76eec13b86701afa/assets/js/base/context/cart-checkout/payment-methods/payment-method-data-context.tsx#L367-L373) and [`onCheckoutWithSuccess`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f0fc6c5fff2dcc5ee2ddb6ac76eec13b86701afa/assets/js/base/context/cart-checkout/checkout-state/index.js#L305-L306))  are also updating the state from the response which only works currently because `success` type aborts.

Ultimately I think we need a solution which satisfies the following conditions:

- assumes multiple subscribed observers might be registered that all have to process on the event.
- only aborts on the first observer error.
- updates state for success responses where we want to allow updating the state (I also think we need to review the current implementations to make sure we don't update "protected" state values).

nerrad:
Fixes: #4048 

As highlighted in #4048, the `emitEventWithAbort` emitter has a flaw in that it was aborting on `{ type: 'success' }` responses from subscribed observers. This creates complications where there are multiple observers subscribed as outlined with the example given in the issue. 

As implemented in this PR:

- `emitEventWithAbort` now only aborts if there is a thrown error or any subscribed observers return a fail or error type response (`{ type: 'error' }` or `{ type: 'failure' }`.
- If there is a failure by any of the observers, it doesn't matter if any of the previous observer calls up to that point executed successfully, there will be _no_ update to the checkout state.
- for the `onPaymentProcessing` event, if there are multiple success events (and no error events) then _only the response from the last observer_ will be used to update the state.

On the latter point, there's still some potential for payment method extensions having poor logic to result in unexpected behaviour on checkout submission. However, I think the behaviour I've implemented here is the best approach for now given there's no way to know which observer should "win" in terms of updating the state. 

I didn't do any conversion to TypeScript because this should get included in the next release of Blocks to support the various improvements made to the Stripe extension in the migration work there. Also, there are more and more payment methods integrating with the blocks so this prevents a class of bugs surfacing for them (especially payment methods that have additional verification processes).

## To Test

This impacts all payment processing in the checkout. So it does mean testing of the flows needs to be fairly thorough:

- Test all supported payment methods that completing a transaction works as expected.
- For Stripe, test various cards that fail (both for express payment methods and cc) - you [can get test card numbers here](https://stripe.com/docs/testing).
- Try various scenarios such as a payment card failing for express payment method and then being able to complete using a different payment method (or vice versa).
- Test with 3DS card for Stripe CC.
- Ensure saved payment methods work.
- Test express payment methods in the cart context.

### Testing pending Stripe integration migration (not for inclusion in blocks release testing notes)

Using the [branch for the Stripe integration migration here](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1467), build that version and test with this branch of blocks. Besides repeating the tests above with the Stripe integration included in blocks in this branch for Stripe you'll also want to verify behaviour of:

- payment with a token that is attached to a 3DS card (example `4000000000003063` which forces 3DS confirmation modal on every transaction).
- payment with a 3DS card via express payment methods. Unless you have a personal 3DS card, the only way I've been able to test this is to add the test card using the Chrome payment settings and then do the transaction with an incognito browser instance (which won't use Google Pay even if it's configured for your browser). This will ensure that just ChromePay is used and allows you to test with dummy cards.
- payment with a 3DS card for Stripe CC.

### DIFF
--- a/assets/js/base/context/event-emit/test/emitters.js
+++ b/assets/js/base/context/event-emit/test/emitters.js
@@ -30,6 +30,13 @@ describe( 'Testing emitters', () => {
 				'observerPromiseWithResolvedValue',
 				{ priority: 10, callback: observerPromiseWithResolvedValue },
 			],
+			[
+				'observerSuccessType',
+				{
+					priority: 10,
+					callback: jest.fn().mockReturnValue( { type: 'success' } ),
+				},
+			],
 		] );
 	} );
 	describe( 'Testing emitEvent()', () => {
@@ -39,11 +46,11 @@ describe( 'Testing emitters', () => {
 			expect( console ).toHaveErroredWith( 'an error' );
 			expect( observerA ).toHaveBeenCalledTimes( 1 );
 			expect( observerB ).toHaveBeenCalledWith( 'foo' );
-			expect( response ).toBe( true );
+			expect( response ).toEqual( [ { type: 'success' } ] );
 		} );
 	} );
 	describe( 'Testing emitEventWithAbort()', () => {
-		it( 'does not abort on any return value other than an object with a type property', async () => {
+		it( 'does not abort on any return value other than an object with an error or fail type property', async () => {
 			observerMocks.delete( 'observerPromiseWithReject' );
 			const observers = { test: observerMocks };
 			const response = await emitEventWithAbort(
@@ -54,17 +61,16 @@ describe( 'Testing emitters', () => {
 			expect( console ).not.toHaveErrored();
 			expect( observerB ).toHaveBeenCalledTimes( 1 );
 			expect( observerPromiseWithResolvedValue ).toHaveBeenCalled();
-			expect( response ).toBe( true );
+			expect( response ).toEqual( [ { type: 'success' } ] );
 		} );
-		it( 'Aborts on a return value with an object that has a type property', async () => {
+		it( 'Aborts on a return value with an object that has a a fail type property', async () => {
 			const validObjectResponse = jest
 				.fn()
-				.mockReturnValue( { type: 'success' } );
+				.mockReturnValue( { type: 'failure' } );
 			observerMocks.set( 'observerValidObject', {
 				priority: 5,
 				callback: validObjectResponse,
 			} );
-			observerMocks.delete( 'observerPromiseWithReject' );
 			const observers = { test: observerMocks };
 			const response = await emitEventWithAbort(
 				observers,
@@ -74,7 +80,7 @@ describe( 'Testing emitters', () => {
 			expect( console ).not.toHaveErrored();
 			expect( validObjectResponse ).toHaveBeenCalledTimes( 1 );
 			expect( observerPromiseWithResolvedValue ).not.toHaveBeenCalled();
-			expect( response ).toEqual( { type: 'success' } );
+			expect( response ).toEqual( [ { type: 'failure' } ] );
 		} );
 		it( 'throws an error on an object returned from observer without a type property', async () => {
 			const failingObjectResponse = jest.fn().mockReturnValue( {} );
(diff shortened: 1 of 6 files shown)


nerrad:
I've made this a blocker for 5.1.0 only because we'll want to make sure this gets released so that the Stripe integration migration work can potentially have this as a version requirement.

assets/js/base/context/providers/add-to-cart-form/form-state/index.js
```diff
@@ -206,17 +206,31 @@ export const AddToCartFormStateContextProvider = ( {
 	 */
 	useEffect( () => {
 		if ( addToCartFormState.status === STATUS.AFTER_PROCESSING ) {
+			// @todo: This data package differs from what is passed through in
+			// the checkout state context. Should we introduce a "context"
+			// property in the data package for this emitted event so that
+			// observers are able to know what context the event is firing in?
```
nerrad, line 212: I think this context provider was preliminary work for the add-to-cart form element block which is currently not released. So I've excluded it from testing notes. It's possible this will be revisited when we resume work on that block.

assets/js/base/context/event-emit/test/emitters.js
```diff
@@ -54,17 +61,16 @@ describe( 'Testing emitters', () => {
 			expect( console ).not.toHaveErrored();
 			expect( observerB ).toHaveBeenCalledTimes( 1 );
 			expect( observerPromiseWithResolvedValue ).toHaveBeenCalled();
-			expect( response ).toBe( true );
+			expect( response ).toEqual( [ { type: 'success' } ] );
 		} );
-		it( 'Aborts on a return value with an object that has a type property', async () => {
+		it( 'Aborts on a return value with an object that has a a fail type property', async () => {
```
Aljullu, line 66:
```suggestion
		it( 'Aborts on a return value with an object that has a fail type property', async () => {
```

Aljullu (approved):
I did a round of testing and only found one issue (#4165). I could reproduce it in `trunk` too, so it doesn't seem to be a regression from this PR. I also had to skip the 3DS card in express payment method step because I had trouble setting it up. I will spend some more time trying to get it to work, but I don't want to block this PR because of that.

The rationale of code changes makes sense. :+1: 

There is an e2e test failing, but doesn't seem to be related to this PR.

ralucaStan:
@nerrad rebasing trunk should fix the failing e2e test

PR #4162, opened 2021-05-02
Don't clear email and phone fields when using separate billing address.

### LINKED ISSUE #4155: When you click the 'Use same address for billing', the email input field gets emptied
When you click the 'Use same address for billing', the email input field gets emptied.

senadir:
When unchecking "Use shipping as billing", the email and address fields are wiped out due to a code order, it seems @nerrad [anticipated this in his review](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3358/files#r524545700) but @mikejolley confirmed it didn't happen, not sure what changed between now and then but I updated the spread order so it doesn't override any existing email and billing fields.

I went along and removed the empty email and phone fields from the previous billing address, to not affect the current values.

Fixes #4155

1. On checkout (guest and logged in), fill out your email, shipping fields and phone number.
2. Uncheck "use shipping for billing"
3. Make sure email and phone still have their values.
4. Submit the order and make sure those values persisted on the server.

### Changelog

> Fix issue in which email and phone fields are cleared when using a separate billing address.

### DIFF
--- a/assets/js/base/context/hooks/use-checkout-address.js
+++ b/assets/js/base/context/hooks/use-checkout-address.js
@@ -88,7 +88,7 @@ export const useCheckoutAddress = () => {
 				setBillingData( shippingAddress );
 			} else {
 				const {
-					// We need to pluck out email and phone from previous billing data because they can be empty, causing the current email and phone to get emptied.
+					// We need to pluck out email and phone from previous billing data because they can be empty, causing the current email and phone to get emptied. See issue #4155
 					/* eslint-disable no-unused-vars */
 					email,
 					phone,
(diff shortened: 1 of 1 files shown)


senadir:
Confirmed that this was working fine in 3.9.0 so this is a recent regression.

opr (commented):
Works great when logged in, but in an incognito window, logged out, this isn't working for me, the email and phone are still removed.

> We could even emit email: undefined because billingData would always contain an email and phone keys (with "" value).

Do you mean `previousBillingData`? If leaving `email: undefined` is redundant we should remove it 🙂

senadir:
@opr this is ready for another review.

senadir:
I tested #3358 and this PR doesn't cause a regression there.

opr (approved):
Working excellently! 🚢

PR #4164, opened 2021-05-03
Add cart data to filters

senadir:
This PR adds cart data to all of our filters, it makes sense to provide this since a lineItem data alone might be enough to extend that item (sometimes you need other items as well).

### Testing instructions 
Currently, we don't have any other plugin using filters except Subscriptions, [so load trunk](https://github.com/woocommerce/woocommerce-subscriptions), run `npm install` and `npm run build` and make sure the filters are working fine.

This might cause a performance decrease by having useStoreCart in all child items, but this is yet to be confirmed so I won't beat myself about it until we have a process to measure performance decrease.  

cc @franticpsyx

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/footer-item/index.js
+++ b/assets/js/base/components/cart-checkout/totals/footer-item/index.js
@@ -28,11 +28,11 @@ const TotalsFooterItem = ( { currency, values } ) => {
 	// Prepare props to pass to the __experimentalApplyCheckoutFilter filter.
 	// We need to pluck out receiveCart.
 	// eslint-disable-next-line no-unused-vars
-	const { extensions, receiveCart, ...cart } = useStoreCart();
+	const { receiveCart, ...cart } = useStoreCart();
 	const label = __experimentalApplyCheckoutFilter( {
 		filterName: 'totalLabel',
 		defaultValue: __( 'Total', 'woo-gutenberg-products-block' ),
-		extensions,
+		extensions: cart.extensions,
 		arg: { cart },
 		// Only accept strings.
 		validation: mustBeString,
--- a/assets/js/base/components/cart-checkout/totals/footer-item/index.js
+++ b/assets/js/base/components/cart-checkout/totals/footer-item/index.js
@@ -24,11 +24,16 @@ const SHOW_TAXES =
 
 const TotalsFooterItem = ( { currency, values } ) => {
 	const { total_price: totalPrice, total_tax: totalTax } = values;
-	const { extensions } = useStoreCart();
+
+	// Prepare props to pass to the __experimentalApplyCheckoutFilter filter.
+	// We need to pluck out receiveCart.
+	// eslint-disable-next-line no-unused-vars
+	const { extensions, receiveCart, ...cart } = useStoreCart();
 	const label = __experimentalApplyCheckoutFilter( {
 		filterName: 'totalLabel',
 		defaultValue: __( 'Total', 'woo-gutenberg-products-block' ),
 		extensions,
+		arg: { cart },
 		// Only accept strings.
 		validation: mustBeString,
 	} );
--- a/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
@@ -9,6 +9,7 @@ import ProductName from '@woocommerce/base-components/product-name';
 import {
 	useStoreCartItemQuantity,
 	useStoreEvents,
+	useStoreCart,
 } from '@woocommerce/base-context/hooks';
 import {
 	ProductBackorderBadge,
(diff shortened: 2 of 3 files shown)


assets/js/base/components/cart-checkout/totals/footer-item/index.js
```diff
@@ -24,11 +24,16 @@ const SHOW_TAXES =
...
+
+	// Prepare props to pass to the __experimentalApplyCheckoutFilter filter.
+	// We need to pluck out receiveCart.
+	// eslint-disable-next-line no-unused-vars
+	const { extensions, receiveCart, ...cart } = useStoreCart();
 	const label = __experimentalApplyCheckoutFilter( {
 		filterName: 'totalLabel',
 		defaultValue: __( 'Total', 'woo-gutenberg-products-block' ),
 		extensions,
+		arg: { cart },
```
mikejolley, line 36: There is a data mismatch between this and the data passed in `productPriceValidation` because extensions is plucked out.

Do we need to ensure what we expose to others is consistent across all hooks?

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
```diff
@@ -114,12 +115,18 @@ const CartLineItemRow = ( {
...
+	// eslint-disable-next-line no-unused-vars
+	const { receiveCart, ...cart } = useStoreCart();
 	const arg = useMemo(
 		() => ( {
 			context: 'cart',
 			cartItem: lineItem,
+			cart,
 		} ),
-		[ lineItem ]
+		[ lineItem, cart ]
```
mikejolley, line 129: Will this affect memorization in a negative way? Is memo needed?

assets/js/base/components/cart-checkout/totals/footer-item/index.js
mikejolley, line 36: Why is this called arg? Just curious.

mikejolley (commented):
Added some questions inline. I think the concept is ok but you might want to reach out for others feedback before committing to this approach.

manospsyx:
Thanks for looking into this @senadir :) 

> This PR adds cart data to all of our filters

A change like this is very welcome -- it certainly gives third parties a more completed picture about how these extensibility points can be leveraged 👍 

As more people become interested in working with the new blocks, it might be a good idea to consider ways to "stabilize" the API. Perhaps add support for setting callback priorities, too? I'm wondering -- why wasn't `@wordpress/hooks` used here? I understand this is not the right PR to discuss this, but I'm bringing this up here because experimental/temporary APIs tend to discourage involvement: If a change to filters is warranted/planned, then every major addition to the current filters (like this one) is creating debt for later (both for your team and for third parties). 

Thanks again!

assets/js/base/components/cart-checkout/totals/footer-item/index.js
senadir: That's a good point, I will make sure to unify them.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
senadir: memo is needed because lineItem was changing a lot, I didn't confirm if we fixed it or not. cart shouldn't affect memo because it's stable.

assets/js/base/components/cart-checkout/totals/footer-item/index.js
senadir: short for arguments, I guess `args` would be better but I don't think it warrants a deprecation process.

senadir:
> Perhaps add support for setting callback priorities
> why wasn't @wordpress/hooks used here?

I still haven't given prioritizing enough mental space yet, but I promise that I will before stable on them.

For `@wordpress/hooks` we used them at the start of the project but they didn't work, `addFilter` didn't work for us at that time, I'm going to scan them again to see if they can work for our use case or not, and I will test `addHook` to see if it works or not. Our requirements for a filter engine:
- It evaluates each hooked filter individually before committing the value, therefore if the filter is broken, it doesn't affect other ones or the final result.
- It works outside React context, meaning the user doesn't need to set it up with a react hook or inside a component for it to work.
- It should be as light as possible because it will run several times in the page for several elements as each element change.

This is the [initial implementation](https://github.com/woocommerce/woocommerce-subscriptions/pull/3965/files) (if you have access to WC Subscriptions repository)

```jsx
/**
  * External dependencies
  */
 import { useEffect } from '@wordpress/element';
 import { addFilter, removeFilter } from '@wordpress/hooks';
 import { __EXPERIMENTAL_TOTAL_LABEL_FILTER } from '@woocommerce/blocks-checkout';

 /**
  * Filters the total label in cart and checkout to append Due today.
  *
  */
 export const TotalsLabelFilter = () => {
 	useEffect( () => {
 		addFilter(
 			__EXPERIMENTAL_TOTAL_LABEL_FILTER,
 			'woocommerce-subscriptions/set-totals-label',
 			() => __( 'Total due today', 'woocommerce-subscriptions' )
 		);

 		return () => {
 			removeFilter(
 				__EXPERIMENTAL_TOTAL_LABEL_FILTER,
 				'woocommerce-subscriptions/set-totals-label'
 			);
 		};
 	}, [] );

 	return null;
 };
```

And in this PR, we switched to our own custom engine
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3774

mikejolley (approved):
Thanks for changing the params passed to the filters. I think this is gtg 👍🏻

manospsyx:
Thanks for looking into this folks, and @senadir -- really appreciate the enlightening comments and background information you are sharing :) Onwards!

PR #4166, opened 2021-05-04
Add checkout filter for coupon names

### LINKED ISSUE #4151: [Points and Rewards] Implement a filter for coupon names
As per pbIy4N-1r0-p2#comment-2208 we need a way to allow extensions to filter coupon names, and possibly replace them with something else.

We already have a system in place to do this with the price labels, so implementing one for coupon codes could follow a similar pattern.

The registration API should allow extensions to add functions in the following shape:
`( couponName: string ) => string`

opr:
This PR will introduce a new filter location. It will be used to allow extensions to change the text in the `RemovableChip` components used to represent a discount that has been applied to the cart.

The filter's name will be `couponCode`, and the value passed to the filter function will be the coupon code. The context (summary) will also be passed to the filter function, along with the rest of the coupon data (`totals` and `discount_type`).

Documentation for this filter has been added too.

Fixes #4151 

### Screenshots

|Before|After|
|---|---|
|||

1. Check out 422-gh-woocommerce/woocommerce-points-and-rewards 
2. Run npm run build in both repositories.
3. In the dashboard, go to **WooCommerce > Points and rewards > Manage points** and add some points to your user account.
4. Add items to your cart.
5. Go to the shortcode cart, apply the points discount using the notice and, observe the discount in the cart.
6. Go to the Cart Block and ensure the coupon applied reads `Points redemption` and not the long system generated coupon code.
7. Try removing the coupon and ensure the discount is removed, and no errors are shown

### Changelog

> Add couponName filter to allow extensions to modify how coupons are displayed in the Cart and Checkout summary.

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/discount/index.js
+++ b/assets/js/base/components/cart-checkout/totals/discount/index.js
@@ -5,7 +5,11 @@ import { __, sprintf } from '@wordpress/i18n';
 import LoadingMask from '@woocommerce/base-components/loading-mask';
 import { RemovableChip } from '@woocommerce/base-components/chip';
 import PropTypes from 'prop-types';
-import { TotalsItem } from '@woocommerce/blocks-checkout';
+import {
+	__experimentalApplyCheckoutFilter,
+	mustBeString,
+	TotalsItem,
+} from '@woocommerce/blocks-checkout';
 import { getSetting } from '@woocommerce/settings';
 
 /**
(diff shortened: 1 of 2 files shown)


assets/js/base/components/cart-checkout/totals/discount/index.js
```diff
@@ -53,34 +57,48 @@ const TotalsDiscount = ( {
...
-								/>
-							) ) }
+							{ cartCoupons.map( ( cartCoupon ) => {
+								const filteredCouponCode = __experimentalApplyCheckoutFilter(
+									{
+										validation: mustBeString,
+										arg: {
+											context: 'summary',
+											coupon: cartCoupon,
+										},
```
senadir, line 67: I will add cart data and extensions data here in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4164 but we would need them in all filters going forward.

senadir (approved):
Codes checks out.

PR #4167, opened 2021-05-04
Add documentation for filters

opr:
This PR adds documentation for the filters we introduced to the Cart and Checkout blocks. The documentation contains a list of filters, and two examples of how to use them.

It also contains a change to a type definition: I have renamed the first argument in the CheckoutFilterFunction type from `label` to `value`. Since we're not always filtering a label this makes more sense. This only exists as a hint to anyone using this type, the value `label` was never used by anything outside of this type def, so it does not need to be changed anywhere else.

1. Please read the documentation from a third party developer's point of view, and try out the worked examples. (The examples assume you're savvy enough to enqueue scripts, so it does not cover that.)

### Changelog

> Add documentation for the experimental filtering API.

### DIFF
--- a/docs/extensibility/available-filters.md
+++ b/docs/extensibility/available-filters.md
@@ -83,7 +83,7 @@ The word 'Total' that precedes the amount due, present in both the Cart _and_ Ch
 
 | Filter name  | Description | Return type  |
 |---|---|---|
-| `totalLabel`  | This is the label for the car total. It defaults to 'Total' (or the word for 'Total' if using translations). | `string`
+| `totalLabel`  | This is the label for the cart total. It defaults to 'Total' (or the word for 'Total' if using translations). | `string`
 
 There are no additional arguments passed to this filter.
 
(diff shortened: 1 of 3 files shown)


docs/extensibility/available-filters.md
```diff
@@ -0,0 +1,154 @@
...
+
+Each of these filters has the following additional arguments passed to it: `{ context: 'summary', cartItem: CartItem }` ([CartItem](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c00da597efe4c16fcf5481c213d8052ec5df3766/assets/js/type-defs/cart.ts#L113))
+
+### Totals footer item (in Cart and Checkout)
+
+The word 'Total' that precedes the amount due, present in both the Cart _and_ Checkout blocks, is also passed through filters.
+
+| Filter name  | Description | Return type  |
+|---|---|---|
+| `totalLabel`  | This is the label for the car total. It defaults to 'Total' (or the word for 'Total' if using translations). | `string`
```
mikejolley:
```suggestion
| `totalLabel`  | This is the label for the cart total. It defaults to 'Total' (or the word for 'Total' if using translations). | `string`
```

mikejolley (approved):
Looks good to me. I added a suggestion for a small typo.

👍🏻

PR #4176, opened 2021-05-06
Feature gate PaymentApi

Aljullu:
A user [reported](https://wordpress.org/support/topic/payment-gateway-with-handle-wc-payment-method-stripe-has-been-deactivated-2/) to be getting a `Payment gateway with handle 'wc-payment-method-stripe' has been deactivated because its dependency 'wc-settings' is not registered.` error even though they were not using WooCommerce Blocks. While I wasn't able to reproduce the issue with Stripe itself, there are a couple of things that could be improved here:

* The error is confusing because at no moment it says the payment gateway has only been deactivated in the blocks (it should keep working as usual in the shortcodes).
* If the site doesn't have WooCommerce Blocks enabled (so Cart and Checkout blocks are not available), it's not needed to perform this verification. I remember in the past it was discussed feature-gating PaymentsApi but we never implemented it (until this PR :slightly_smiling_face: ).

#### Regression testing in the feature plugin

Testing means making sure there are no regressions with https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3920:
1. Add a fake dependency in any payment method. Ie, in [Cheque](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Payments/Integrations/Cheque.php):
```diff
	public function get_payment_method_script_handles() {
		$this->asset_api->register_script(
			'wc-payment-method-cheque',
			'build/wc-payment-method-cheque.js',
+			[ 'missing-dependency' ]
		);
		return [ 'wc-payment-method-cheque' ];
	}
```
2. Go to any `wp-admin` page and verify an error notice logged in your PHP error logs and in the browser console.
3. Create a post or page and verify the Cart and Checkout blocks are available. If you add them, you will notice the faulty payment methods are not available.
4. In the frontend, verify the Cart and Checkout blocks load as usual (but the payment methods with missing dependencies are not loaded).

#### PaymentApi not loaded in packaged version

Until this is released in WC Core, the only way to test is to modify several files to make sure `PaymentApi` is not loaded:

* In [`src/Domain/Services/FeatureGating.php`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Domain/Services/FeatureGating.php#L101) modify the `is_feature_plugin_build()` method to always return false. This will simulate the version that is packaged in WC Core.
* Add `die( 'This should not die' );` inside [`src/Payments/Api.php`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Payments/Api.php).
* Visit any WooCommerce page in the admin and verify `This should not die` is never displayed.

### Changelog

> Improve error message displayed when a payment method didn't have all its dependencies registered.


solstudioim:
Another user [also reported](https://wordpress.org/support/topic/payment-gateway-deactivated-error/?) similar behavior with the dependency being `wc-blocks-registry`

frontdevde (approved):
I tested as per testing instructions and can confirm that I'm now seeing the more descriptive error message that notifies me the issue is related to the Cart and Checkout blocks. 

Thank you for adding the additional testing instructions on how to test the non-feature-plugin case as well. I can confirm it also worked as expected. 

LGTM 👍

PS: The only thing that came to mind was that if the idea is for the gated code to only run when the Cart/Checkout blocks are active, whether there would be a way to gate for the blocks being present/active rather than checking for the feature plugin state (which the blocks will eventually be promoted out of).

Aljullu:
Thanks for the review @frontdevde!

> PS: The only thing that came to mind was that if the idea is for the gated code to only run when the Cart/Checkout blocks are active, whether there would be a way to gate for the blocks being present/active rather than checking for the feature plugin state (which the blocks will eventually be promoted out of).

That's a good point. I'm not 100% if we want to restrict the PaymentApi to be available only if Cart and Checkout blocks are in use. For example, in the future, we could allow 3rd-party blocks/extensions to use that API even if Cart and Checkout blocks are not active.

PR #4179, opened 2021-05-07
Add lodash as a devDependency

ralucaStan:
Lodash is missing as a dependency in package.json. 
This PR adds it to devDependecies, as we don't want it to be part of our build and we rely on using the global.


Aljullu (approved):
LGTM

There is a failing e2e, but I don't think it's related to this pull.

PR #4182, opened 2021-05-07
Check if variation data is iterable before trying to format in CartItemSchema

opr:
When using force sells, I get this warning in the Cart block when adding a product with a synced force sell:
```
Warning: Invalid argument supplied for foreach() in /var/www/html/wp-content/plugins/woocommerce-gutenberg-products-block/src/StoreApi/Schemas/CartItemSchema.php on line 399
```
By checking if the variation data is iterable before supplying it to the foreach we can avoid this warning.

### Screenshots

| Before | After |
|---|---|
| image |  image |

1. Install WooCommerce Force Sells (https://woocommerce.com/products/force-sells/ or https://github.com/woocommerce/woocommerce-force-sells/)
2. Edit a product to add a "Synced force sell" 
image
3. Add this product to your cart
4. Go to the Cart block and ensure no warnings are visible
5. Do the same for the Checkout block

### Changelog

> Stopped a warning being shown when using WooCommerce Force Sells and adding a product with a Synced Force Sell to the cart.

### DIFF
--- a/src/StoreApi/Schemas/CartItemSchema.php
+++ b/src/StoreApi/Schemas/CartItemSchema.php
@@ -392,6 +392,10 @@ class CartItemSchema extends ProductSchema {
 	protected function format_variation_data( $variation_data, $product ) {
 		$return = [];
 
+		if ( ! is_iterable( $variation_data ) ) {
+			return $return;
+		}
+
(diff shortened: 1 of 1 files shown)


ralucaStan (approved):
👍

PR #4186, opened 2021-05-11
Set order status to pending when no payment is needed before completing order

### LINKED ISSUE #4168: Confirmation Emails are not sending
## Describe the bug
When an order is made, no processing / new order or any other confirmation emails are sending. I have tried to deactive all plugins and use just woocommerce and the default woocommerce shortcode checkout, and emails send. I have SMTP set up, and have checked all required solutions such as double checking settings, removing any custom code, etc.

## To reproduce
To reproduce the bug, 
1) Fresh installation of Wordpress.
1) Set up a Woocommerce store with woocommerce blocks. 
2) Ensure all email settings are correct. 
3) Create one test product. 
4) Order the product. 
5) No confirmation email should come through.

TO confirm this is with Woocommerce blocks, now:
1) Make another cart page with the old woocommerce cart shortcode [woocommerce_cart]
2) Make another checkout page with the old woocommerce checkout shortcode [woocommerce_checkout]
3) Ensure Woocommerce advanced settings to locate cart and checkout are pointed to the new pages.
4) Order the test product again.
5) Email confirmation should arrive immediately.

## Expected behavior
After an order is made, there are a number of emails that should be sent out to both admin and customer at different stages of the order process.

Divi  4.9.4
Wordpress 5.7.1
Woocoomerce 5.2.2
Woocommerce blocks 5.0.0

mikejolley:
The default order status for Block Checkout is "draft". Before a payment is made, the order status is changed to "pending". After payment, the order status is changed again to "onhold", "processing", or "complete". This in turn triggers transactional emails, by core, for the following events:

```php
add_action( 'woocommerce_order_status_pending_to_processing_notification', array( $this, 'trigger' ), 10, 2 );
add_action( 'woocommerce_order_status_pending_to_completed_notification', array( $this, 'trigger' ), 10, 2 );
add_action( 'woocommerce_order_status_pending_to_on-hold_notification', array( $this, 'trigger' ), 10, 2 );
add_action( 'woocommerce_order_status_failed_to_processing_notification', array( $this, 'trigger' ), 10, 2 );
add_action( 'woocommerce_order_status_failed_to_completed_notification', array( $this, 'trigger' ), 10, 2 );
add_action( 'woocommerce_order_status_failed_to_on-hold_notification', array( $this, 'trigger' ), 10, 2 );
add_action( 'woocommerce_order_status_cancelled_to_processing_notification', array( $this, 'trigger' ), 10, 2 );
add_action( 'woocommerce_order_status_cancelled_to_completed_notification', array( $this, 'trigger' ), 10, 2 );
add_action( 'woocommerce_order_status_cancelled_to_on-hold_notification', array( $this, 'trigger' ), 10, 2 );
```

In the case of free orders (no payment needed), we were just changing status directly to completed which resulted in a "draft_to_completed" transition, thus bypassing transactional emails.

The fix is to make the order pending before completing it.

Fixes #4168

1. Add a VIRTUAL product costing $0 to the cart. The order total should be 0.
2. Go to Block checkout and place the order.
3. Check mailhog (if using WP Local) and confirm the new order emails (x2) were received.

### Changelog

> Fix sending of confirmation emails for orders when no payment is needed.

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -264,7 +264,13 @@ class Checkout extends AbstractCartRoute {
 		/**
 		 * Process the payment and return the results.
 		 */
-		$payment_result = $this->order->needs_payment() ? $this->process_payment( $request ) : $this->process_without_payment( $request );
+		$payment_result = new PaymentResult();
+
+		if ( $this->order->needs_payment() ) {
+			$this->process_payment( $request, $payment_result );
+		} else {
+			$this->process_without_payment( $request, $payment_result );
+		}
 
 		return $this->prepare_item_for_response(
 			(object) [
@@ -478,33 +484,35 @@ class Checkout extends AbstractCartRoute {
 	 * For orders which do not require payment, just update status.
 	 *
 	 * @param WP_REST_Request $request Request object.
-	 * @return PaymentResult
+	 * @param PaymentResult   $payment_result Payment result object.
 	 */
-	private function process_without_payment( WP_REST_Request $request ) {
+	private function process_without_payment( WP_REST_Request $request, PaymentResult $payment_result ) {
+		// Transition the order to pending, and then completed. This ensures transactional emails fire for pending_to_complete events.
+		$this->order->update_status( 'pending' );
 		$this->order->payment_complete();
 
-		$result = new PaymentResult( 'success' );
-		$result->set_redirect_url( $this->order->get_checkout_order_received_url() );
-
-		return $result;
+		// Mark the payment as successful.
+		$payment_result->set_status( 'success' );
+		$payment_result->set_redirect_url( $this->order->get_checkout_order_received_url() );
 	}
 
 	/**
 	 * Fires an action hook instructing active payment gateways to process the payment for an order and provide a result.
 	 *
 	 * @throws RouteException On error.
+	 *
 	 * @param WP_REST_Request $request Request object.
-	 * @return PaymentResult
+	 * @param PaymentResult   $payment_result Payment result object.
 	 */
-	private function process_payment( WP_REST_Request $request ) {
+	private function process_payment( WP_REST_Request $request, PaymentResult $payment_result ) {
 		try {
-			$result  = new PaymentResult();
+			// Transition the order to pending before making payment.
+			$this->order->update_status( 'pending' );
+
+			// Prepare the payment context object to pass through payment hooks.
 			$context = new PaymentContext();
 			$context->set_payment_method( $this->get_request_payment_method_id( $request ) );
 			$context->set_payment_data( $this->get_request_payment_data( $request ) );
-
-			// Orders are made pending before attempting payment.
-			$this->order->update_status( 'pending' );
 			$context->set_order( $this->order );
 
 			/**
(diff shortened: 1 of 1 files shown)


src/StoreApi/Routes/Checkout.php
```diff
@@ -264,7 +264,13 @@ protected function get_route_post_response( WP_REST_Request $request ) {
 		/**
 		 * Process the payment and return the results.
 		 */
-		$payment_result = $this->order->needs_payment() ? $this->process_payment( $request ) : $this->process_without_payment( $request );
+		$payment_result = new PaymentResult();
```
mikejolley, line 267: Passing this through to the private methods to make the payment methods clearer.

src/StoreApi/Routes/Checkout.php
```diff
@@ -478,53 +484,52 @@ private function update_order_from_request( WP_REST_Request $request ) {
 	 * For orders which do not require payment, just update status.
 	 *
 	 * @param WP_REST_Request $request Request object.
-	 * @return PaymentResult
+	 * @param PaymentResult   $payment_result Payment result object.
 	 */
-	private function process_without_payment( WP_REST_Request $request ) {
+	private function process_without_payment( WP_REST_Request $request, PaymentResult $payment_result ) {
+		// Transition the order to pending, and then completed. This ensures transactional emails fire for pending_to_complete events.
+		$this->order->update_status( 'pending' );
```
mikejolley, line 491: This is the important line; changing the order status to pending.

frontdevde (approved):
Tested this PR as per testing instructions and can confirm that now the New Order email is being sent even when the digital item costs 0$. 

|Before|After|
|-|-|
|Markup on 2021-05-11 at 14:56:48|Markup on 2021-05-11 at 14:57:16|

LGTM 👍

senadir (approved):
Can confirm this fixes the issue for me as well.

image

nerrad:
Hey @mikejolley, something occurred to me when reviewing some pending questions we have around stock reservations that has some relevance to this change. In another document (pca54o-7v-p2) this question was asked:

> What happens if an order becomes pending after being a checkout-draft?

The biggie in relation to that question was:

> But what reserves stock if a draft order becomes pending payment?

Have stock reservations been checked as a part of testing the changes in this PR?

mikejolley:
@nerrad Stock is reserved by the Store API when the draft order is created. Stock holds are valid for both pending and draft orders. Therefore this is not a problem. Also note the order is only pending for literally a few milliseconds - it's set to completed right after :)

This is where it's reserved: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fdc74b8700eaaf6f3d07ddef0e5f4070b25b14ea/src/StoreApi/Routes/Checkout.php#L423-L424

And this is the line that checks for reservations, which uses both pending and draft status: https://github.com/woocommerce/woocommerce/blob/b19500728b4b292562afb65eb3a0c0f50d5859de/src/Checkout/Helpers/ReserveStock.php#L209

I think the question may be referring to the 'hold stock' setting in woo core which reserves pending orders for much much longer, so users have a chance to pay via PayPal and receive an IPN response. This may still be something that needs reviewing for orders requiring payment, but this change is about fixing the behaviour of free orders so is not affected.

PR #4187, opened 2021-05-11
Use value rather than name for list items within the ProductDetails component

### LINKED ISSUE #4081: React warning when using product addons with checkboxes
## Describe the bug
When viewing the cart, if you have an item using checkboxes for Product Addons and more than one is checked the warning `Encountered two children with the same key` is shown.

## To reproduce

1. Create a product with checkbox addons.
2. View the product page, select two of the checkboxes and add the product to the cart.
3. View the cart block.
4. See error in console.

## Expected behavior
There should be no error.

* Core version: 5.7.1
* WooCommerce version: 5.3.0-dev

mikejolley:
Keys were based on "name" and "display", falling back to "name". With product addons, this produced keys such as `Added extrasAdded extras`.

"display" is actually a variation of the value prop, so to make the items more unique we should actually be using the value prop. This produces keys like `Added extrasValue` and solves the issue.

Fixes #4081

See instructions in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4081. Ensure SCRIPT_DEBUG is true in your environment.

### Changelog

> Fix duplicate react keys in ProductDetails component.

### DIFF
--- a/assets/js/base/components/cart-checkout/product-details/index.tsx
+++ b/assets/js/base/components/cart-checkout/product-details/index.tsx
@@ -37,7 +37,7 @@ const ProductDetails = ( {
 					: '';
 				return (
 					<li
-						key={ detail.name + ( detail.display || detail.name ) }
+						key={ detail.name + ( detail.display || detail.value ) }
 						className={ className }
 					>
 						{ detail.name && (


senadir (approved):
I was only able to reproduce when they had the same price or no price, since price string was also used in creating the key.
Checks out!

PR #4188, opened 2021-05-11
Remove the need for the `canMakePayment` callback in the editor context

### LINKED ISSUE #4101: Checkout payment method `canMakePayment` applies to Gutenberg preview
## Describe the bug

I'd like to always display an express payment method preview in the Gutenberg editor for the cart and checkout blocks, but sometimes not display it to the shopper (e.g when it should not be available due to product requirements).

From the [docs](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/main/docs/extensibility/payment-method-integration.md):

> canMakePayment (required): A callback to determine whether the payment method should be available as an option for the shopper.

The `canMakePayment` callback applies not only for the shopper, but also for the Gutenberg editor preview in wp-admin. I don't want to initialize the payment API in the blocks editor to confirm if the payment method should be available.

In my opinion, the `canMakePayment` callback should be dissociated from the Gutenberg preview. If I don't want to display the preview, I should simply be able to return `null` in the `edit` prop.

Am I missing something, or is that really how it's supposed to work?

Is there an easy way to return `true` in `canMakePayment` if the payment method is being rendered in the editor?

## To reproduce

Register an express payment method similar to this:

```JavaScript
registerExpressPaymentMethod(
    {
	name: 'my_payment_method',
	content: <div>A react node</div>,
	edit: <div>A react node</div>,
	canMakePayment: () => 'undefined' !== typeof enqueuedAPIObject,
	paymentMethodId: 'new_payment_method',
    }
);
```

As an example, note that I just want to make it available to shoppers when the `enqueuedAPIObject` global variable is defined, but as a result, `<div>A react node</div>` isn't displayed in Gutenberg

## Expected behavior

The `edit` preview is never visible in `wp-admin`. If I want to make it visible, I need to do some special logic there to tell if the component is being rendered in the editor.

## Additional context

I'm working on Apple Pay, and apparently, the problem exists in the Stripe integration as well. The preview is never visible.

mikejolley:
`canMakePayment` currently runs in the editor context for payment methods which is unnecessary, especially considering the callback is passed sample data. Instead, as requested in #4101, it should evaluate as true in editor context so that the given `edit` component is rendered regardless. 

If `edit` returns null, nothing will be rendered, which still gives developers control.

Fixes #4101

I tested this by enabling the Stripe plugin and ensuring the preview (apple pay) appeared correctly in the editor.

I also confirmed that `canMakePayment` still worked on the frontend, and that it wasn't running in the editor context.

### Changelog

> For payment methods, only use `canMakePayment` in the frontend (not the editor) context.

### DIFF
--- a/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
+++ b/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
@@ -103,11 +103,15 @@ const usePaymentMethodRegistration = (
 				continue;
 			}
 
-			// In front end, ask payment method if it should be available.
+			// See if payment method should be available. This always evaluates to true in the editor context.
 			try {
-				const canPay = await Promise.resolve(
-					paymentMethod.canMakePayment( canPayArgument.current )
-				);
+				const canPay = isEditor
+					? true
+					: await Promise.resolve(
+							paymentMethod.canMakePayment(
+								canPayArgument.current
+							)
+					  );
 
 				if ( !! canPay ) {
 					if (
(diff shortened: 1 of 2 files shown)


senadir (approved):
Codes looks right, testing works fine
image

PR #4197, opened 2021-05-12
Sync customer data during checkout with draft orders.

### LINKED ISSUE #3150: Capture details of checkout in draft order

No problem. When using the checkout block as a non-logged in user, no details are initially filled in.
If a required field is left empty and the customer clicks the 'Place Order' button, a draft order is created.

The draft order does not include any information that was actually filled in during checkout.

Capturing _any_ of the information that was filled in on the checkout block in the draft order could prove helpful for stores following up on those draft orders.

I believe GDPR should be taken into consideration, but a similar solution can be found in AutomateWoo's settings. **Enable pre-submit data capture**: https://automatewoo.com/docs/abandoned-cart/#capturing-guest-data

If nothing else, offering a hook (if possible) to enable this would be helpful as well.

No alternatives that I can think of at this moment.

## Additional context

https://wordpress.org/support/topic/draft-order-created-automatically-on-failed-checkout-checkout-block/

### LINKED ISSUE #3661: Implement code needed for MailPoet integration
From some conversations with the team, they shared the things they need for integrating and it looks like this is all server side.

The following hooks:
```
woocommerce_add_to_cart
woocommerce_cart_item_removed
woocommerce_after_cart_item_quantity_update
woocommerce_before_cart_item_quantity_zero
woocommerce_cart_emptied
woocommerce_cart_item_restored
```

And then this was also mentioned:

> I think all we really need is a hook, from which we can get an email entered in the checkout. If the above-mentioned hooks stay the same, we’ll use the new hook to update the email address to send abandoned cart to.

Once we work on this, we will need to clarify at what point they need the email, but I'm just documenting this for planning. I think most of the cart hooks should already be firing but we'll just need to confirm.

mikejolley:
Currently, when you enter checkout, a draft order is created. This is populated with any existing session data, but the order is not updated unless you re-enter checkout, or place the order using the button.

This PR improves how checkout data is persisted on the server by:

1. Syncing with the server when the billing email changes, not just the address
2. Syncing any existing draft order with customer data when customer data updates

This means if a customer field changes on checkout that is synced with the server, the order also gets updated to keep it current. This makes it usable by plugins that need access to the billing email before the order is placed.

Fixes #3661
Fixes #3150

1. Go to checkout
2. Fill out billing email
3. Go to admin, Orders > Drafts. See if the billing email is populated in the draft order.

### Changelog

> Sync customer data during checkout with draft orders.

### DIFF
--- a/src/StoreApi/Routes/CartUpdateCustomer.php
+++ b/src/StoreApi/Routes/CartUpdateCustomer.php
@@ -119,9 +119,50 @@ class CartUpdateCustomer extends AbstractCartRoute {
 		);
 		wc()->customer->save();
 
-		$cart->calculate_shipping();
-		$cart->calculate_totals();
+		$this->calculate_totals();
+		$this->maybe_update_order();
 
 		return rest_ensure_response( $this->schema->get_item_response( $cart ) );
+	}
+
+	/**
+	 * If there is a draft order, update customer data there also.
+	 *
+	 * @return void
+	 */
+	protected function maybe_update_order() {
+		$draft_order_id = wc()->session->get( 'store_api_draft_order', 0 );
+		$draft_order    = $draft_order_id ? wc_get_order( $draft_order_id ) : false;
+
+		if ( ! $draft_order ) {
+			return;
+		}
+
+		$draft_order->set_props(
+			[
+				'billing_first_name'  => wc()->customer->get_billing_first_name(),
+				'billing_last_name'   => wc()->customer->get_billing_last_name(),
+				'billing_company'     => wc()->customer->get_billing_company(),
+				'billing_address_1'   => wc()->customer->get_billing_address_1(),
+				'billing_address_2'   => wc()->customer->get_billing_address_2(),
+				'billing_city'        => wc()->customer->get_billing_city(),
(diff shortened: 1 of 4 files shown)


src/StoreApi/Routes/CartUpdateCustomer.php
```diff
@@ -119,9 +119,50 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
 		);
 		wc()->customer->save();
 
-		$cart->calculate_shipping();
```
senadir, line 122: `calculate_shipping` was redundant?

senadir (approved):
Works like charm 👨🏼‍🍳

src/StoreApi/Routes/CartUpdateCustomer.php
mikejolley: Cart class has a calculate_totals method which does the same thing.

mikejolley:
Failing tests are all `TimeoutError: waiting for selector `.wc-block-checkout` failed: timeout 30000ms exceeded` which I believe is unrelated and common to all PRs right now. Merging.

PR #4199, opened 2021-05-12
Fix cart and checkout frontend e2e tests

ralucaStan:
This PR fixes the frail cart and checkout frontend e2e tests.

- I added the logic to prevent CartCheckoutCompatibilityNotice appearance
- I handled clicking on a link and navigating to another page with Promise.all[] and page.waitForNavigation(), to make sure the navigation happens
- I search for text on-page instead of a selector, seems more stable and is more user-oriented testing

I've also updated the testing docs, to include wp-env clean command.

### DIFF
--- a/tests/e2e/specs/frontend/cart.test.js
+++ b/tests/e2e/specs/frontend/cart.test.js
@@ -130,10 +130,11 @@ describe( `${ block.name } Block (frontend)`, () => {
 
 		// This is to ensure we've clicked the right cart button.
 		expect( selectedValue ).toBeGreaterThan( 1 );
-
 		await scrollTo( '.wc-block-cart__submit-button' );
-		await page.click( '.wc-block-cart__submit-button' );
-		await page.waitForNavigation();
+		await Promise.all( [
+			page.waitForNavigation(),
+			page.click( '.wc-block-cart__submit-button' ),
+		] );
 		await page.waitForSelector( '.wc-block-checkout' );
 		await page.goBack( { waitUntil: 'networkidle0' } );
 
--- a/tests/e2e/specs/frontend/cart.test.js
+++ b/tests/e2e/specs/frontend/cart.test.js
@@ -36,12 +36,6 @@ describe( `${ block.name } Block (frontend)`, () => {
 
 	beforeAll( async () => {
 		await page.evaluate( () => window.localStorage.clear() );
-		//prevent CartCheckoutCompatibilityNotice from appearing
-		await page.evaluate( () => {
-			localStorage.removeItem(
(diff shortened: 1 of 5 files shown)


mikejolley:
@ralucaStan this is looking promising. I restored the PHPUnit test and rebased, since that was fixed in trunk now.

tests/e2e/specs/frontend/cart.test.js
```diff
@@ -114,10 +131,12 @@ describe( `${ block.name } Block (frontend)`, () => {
...
 		expect( selectedValue ).toBeGreaterThan( 1 );
-
 		await scrollTo( '.wc-block-cart__submit-button' );
-		await page.click( '.wc-block-cart__submit-button' );
-		await page.waitForSelector( '.wc-block-checkout' );
+		await Promise.all( [
+			page.waitForNavigation(),
+			page.click( '.wc-block-cart__submit-button' ),
+		] );
+		await page.evaluate( () => window.find( 'Checkout Block' ) );
```
opr: What is this line for? I understand a value is returned by `page.evaluate` but it isn't being assigned to anything.

Maybe it would be better to use [`page.waitForFunction`](https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-pagewaitforfunctionpagefunction-options-args) here? It would keep retrying until the function returns true if the goal is to wait for the `Checkout Block` text to be on the page?

ralucaStan: Nice catch Thomas, the assertion was missing here. I change a bit the code to actually test that the h1 on the page contains the text `Checkout`. For some reason on the CI the title is `Checkout` and on the local env the title is `Checkout block`. We will need to investigate why the block navigates to different pages on different envs, but for this PR the purpose is to pass the test.

tests/e2e/specs/frontend/cart.test.js
```diff
@@ -36,6 +36,18 @@ describe( `${ block.name } Block (frontend)`, () => {
 
 	beforeAll( async () => {
 		await page.evaluate( () => window.localStorage.clear() );
+		//prevent CartCheckoutCompatibilityNotice from appearing
+		await page.evaluate( () => {
+			localStorage.removeItem(
+				'wc-blocks_dismissed_compatibility_notices'
+			);
+		} );
```
senadir: This seems redundant giving that you're writing to this key in the next block

senadir (approved):
This should be good to go if tests are passing, left a small comment.

senadir:
Tests still pass after removing the redundant `removeItem` code.

PR #3983, opened 2021-03-19
Add types to Checkout State Provider, and introduce Context Module Functions

mikejolley:
This PR implements Typescript in the Checkout State Provider. Doing this revealed a bunch of typos and bugs! 

As well as this, I added some shared helpers to base-utils to help with Typescript support (isObject, objectHasProp, isNull).

Finally, inspired by epic react I added some new functions exported from the context provider which take `dispatch` as a parameter and do something. These were previously exported by the context provider, but with this method we only need to export the dispatcher which means the functions can be tree-shaken.

- `submitCheckout`
- `resetCheckout`
- `setRedirectUrl`
- `setHasError`
- `incrementCalculating`
- `decrementCalculating`
- `setCustomerId`
- `setOrderId`
- `setOrderNotes`
- `setShouldCreateAccount`
- `setAfterProcessing`

So example usage would be:

```javascript
import {
	useCheckoutContext,
	setOrderNotes,
} from '@woocommerce/base-context/cart-checkout/checkout-state';

...

const { dispatch } = useCheckoutContext();

setOrderNotes( dispatch, value );
```

**To test:**

Smoke test the checkout process, including making payment. Checkout state provider is only used during checkout.

PR #4200, opened 2021-05-12
Convert checkout state context provider to Typescript

mikejolley:
I've pulled all the Typescript changes from #3983 into this new PR. This is only typing, and fixing issues as a result of that.

Once merged, Ill make the rest of the changes in #3983 in a fresh PR.

Requires checkout smoke test; try to get validation errors, send incorrect data on submission, and test payment/placing orders.


mikejolley:
@nerrad was a reviewer on #3983 so I've added him to this one too.

mikejolley:
Annoyingly, even with git move this hasn't shown the diff properly :(

mikejolley:
Noticed a missing change so moved this back in progress until I review.

mikejolley:
Sorted. There were 2 similar methods in my old PR and in trunk. I've consolidated them into one.

assets/js/base/context/providers/cart-checkout/checkout-state/actions.ts
```diff
@@ -0,0 +1,78 @@
...
+	orderNotes?: string;
+}
+
+/**
+ * All the actions that can be dispatched for the checkout.
+ */
+export const actions = {
+	setPristine: (): ActionType => ( {
+		type: ACTION.SET_PRISTINE,
+	} ),
```
nerrad: If we do something like the following, we shouldn't have to define the `ActionType` interface and the types will be more narrowly defined to the actual action?

```suggestion
	setPristine: () => ( {
		type: ACTION.SET_PRISTINE,
	} as const ),
```

I make this suggestion because this was what was recommended to me for action objects when I was doing the TS conversion for the cart data store ([example](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a5add04cd927090c356136eaf2920b4d6f1827cc/assets/js/data/cart/actions.ts#L31-L39)).

assets/js/base/context/providers/cart-checkout/checkout-state/constants.ts
```diff
@@ -0,0 +1,103 @@
...
+	isCalculating: false,
+	isProcessing: false,
+	isBeforeProcessing: false,
+	isAfterProcessing: false,
+	hasError: false,
+	redirectUrl: '',
+	orderId: 0,
+	orderNotes: '',
+	customerId: 0,
+	onCheckoutAfterProcessingWithSuccess: () => () => () => void null,
```
nerrad: I'm curious why three functions here instead of two (same for other events in this default state)?  The `useMemo` invocation in the Checkout would end up _returning_ a structure like `() => () => void null` right (but I know you pass something like `() => () => () => void null` _into_ the `useMemo` as the arg)?

So I guess it looks like the default _should_ be `() => () => void null`?

assets/js/base/context/providers/cart-checkout/checkout-state/event-emit.ts
```diff
@@ -0,0 +1,62 @@
...
+ * for the event emitter and will return a dispatcher for removing the
+ * registered callback (useful for implementation in `useEffect`).
+ *
+ * @param {Function} observerDispatch The emitter reducer dispatcher.
+ * @return {Object} An object with the various payment event emitter registration functions
+ */
+const useEventEmitters = (
+	observerDispatch: React.Dispatch< ActionType >
+): EventEmittersType => {
+	const eventEmitters = useMemo(
```
nerrad, line 42: Oh this is a nifty change to move the memoization into a custom hook! Nice 👏🏻

assets/js/base/context/providers/cart-checkout/checkout-state/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+		message: '',
+		paymentStatus: '',
+		redirectUrl: '',
+		paymentDetails: {},
+	} as PaymentResultDataType;
+
+	// payment_result is present in successful responses.
+	if ( 'payment_result' in response ) {
+		paymentResult.paymentStatus = response.payment_result.payment_status;
+		paymentResult.redirectUrl = response.payment_result.redirect_url;
```
nerrad, line 28: Currently `response.payment_result.redirect_url` might not be present? So `CheckoutResponseSuccess` should probably have this defined as optional (see comment I left with that type) and thus there would need to be a check to make sure it's present here.

```suggestion
		paymentResult.redirectUrl = response.payment_result?.redirect_url;
```

Of course if you've discovered that `redirectUrl` is always in the response then this should be fine, but this  does significantly change behaviour because prior to this the `redirectUrl` was only used if it was non-truthy:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a5add04cd927090c356136eaf2920b4d6f1827cc/assets/js/base/context/providers/cart-checkout/checkout-state/reducer.js#L117-L124

Won't the change in this PR mean that if a `paymentResult.redirectUrl` is set to an empty string then whatever is set as the redirectUrl in the state will get cleared given the reducer now explicitly checks for undefined?

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
```diff
@@ -183,35 +144,17 @@ export const CheckoutStateProvider = ( {
...
 					dispatch(
-						actions.setProcessingResponse(
-							prepareResponseData( response.payment_result )
-						)
+						actions.setRedirectUrl( paymentResult.redirectUrl )
 					);
 				}
-				void dispatch( actions.setAfterProcessing() );
+				dispatch( actions.setProcessingResponse( paymentResult ) );
+				dispatch( actions.setAfterProcessing() );
```
nerrad, line 157: I really appreciate the cleanup around the logic here however it appears there's some changes. One of the major changes here is that `paymentResult` is always created with defaults and `setProcessingResponse` action is always dispatched.

While I did notice that the reducer has the following change:

```js
case ACTION.SET_PROCESSING_RESPONSE:
			newState = {
				...state,
				processingResponse: ( data as PaymentResultDataType ) || null,
			};
			break;
```

It still appears to me that processing response will never be null after this event fires because `paymentResult` always has defaults (due to the returned value from `getPaymentResultFromCheckoutResponse`)?

Some changes that I can think of as a result:

| Previous | Now |
| ------- | ------- |
state only updated twice if an observer returned a response with a `payment_result` property | state is always updated twice regardless if the observer is not a payment method updating payment related state
there is _optional_ handling for payment method data that may be provided by the observer subscribed to the event | it is assumed there's always payment method data for updating in the state (if the response doesn't have it, default values are used).

I realize I might have read the code wrong or missed something here - I'm hoping you can point that out to me.

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
```diff
@@ -261,20 +204,21 @@ export const CheckoutStateProvider = ( {
 			return;
 		}
 
-		const handleErrorResponse = ( observerResponses ) => {
+		const handleErrorResponse = ( observerResponses: unknown[] ) => {
 			let errorResponse = null;
 			observerResponses.forEach( ( response ) => {
-				const { message, messageContext } = response;
 				if (
+					isObject( response ) &&
```
nerrad: `isErrorResponse` and `isFailResponse` already [check for whether `response.type` exists](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a5add04cd927090c356136eaf2920b4d6f1827cc/assets/js/base/context/hooks/use-emit-response.js#L7-L9) and thus whether it's an object or not implicitly.

I tried removing this check and I see it results in ts errors on the `response` usage inside the conditional block. It'd probably be preferable to see if this could be fixed by converting `use-emit-response.js` to ts and updating the `isResponseOf` function to have the appropriate guards but I appreciate the desire to not go too far down the rabbit trail here. Maybe a good place to note for a followup?

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
```diff
@@ -285,7 +229,7 @@ export const CheckoutStateProvider = ( {
 				redirectUrl: checkoutState.redirectUrl,
 				orderId: checkoutState.orderId,
 				customerId: checkoutState.customerId,
-				customerNote: checkoutState.customerNote,
+				orderNotes: checkoutState.orderNotes,
```
nerrad, line 231: Nice cleanup for consistency here 👏🏻

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
```diff
@@ -0,0 +1,195 @@
...
+					? {
+							...state,
+							redirectUrl: url,
+					  }
+					: state;
+			break;
+		case ACTION.SET_PROCESSING_RESPONSE:
+			newState = {
+				...state,
+				processingResponse: ( data as PaymentResultDataType ) || null,
```
nerrad: As mentioned in another comment, I don't see how `null` would ever get set here with the current implementation of the action dispatched on this reducer? 

Also I'm not quite clear on how `as` used here affects the condition. Does this mean if `data` is present (and not undefined) it will be set on `processingResponse` asserted as the type of `PaymentResultDataType` or with this type of construct is this a way of using TypeScript to assert if `data` is of the `PaymentResultDataType` and if it's not, set to `null` (used as a typeguard)? I couldn't find any documentation on it and when I tried an [example in TS playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAsgogQQCpQLxQN4Cgq6gYQAkBJfAaWIDkAuKAZ2ACcBLAOwHMAaHPAITgCqAMVoMWHbgF8A3FiwBjAPasGUAIZpMMuUpXAoAI00AKderqxEKAJRQAPnagByOooC2EYAAs27KBAAbOggnWQVlVwCIADoAxXZTI2tpIA) it looks like as long as `data` is an object, `processingResponse` will get set to it?

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
```diff
@@ -0,0 +1,195 @@
...
+		case ACTION.SET_COMPLETE:
+			newState =
+				state.status !== STATUS.COMPLETE
+					? {
+							...state,
+							status: STATUS.COMPLETE,
+							redirectUrl:
+								data !== undefined &&
+								data.redirectUrl !== undefined
+									? ( data.redirectUrl as string )
```
nerrad: Same question here as what I asked earlier. What is the intent of using `as` here?

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
```diff
@@ -0,0 +1,195 @@
...
+			break;
+		case ACTION.DECREMENT_CALCULATING:
+			newState = {
+				...state,
+				calculatingCount: Math.max( 0, state.calculatingCount - 1 ),
+			};
+			break;
+		case ACTION.SET_CUSTOMER_ID:
+			newState =
+				customerId !== undefined
```
nerrad, line 151: This is a change here from the logic in the previous reducer code. Have you checked that there are no cases where customerId might explicitly be cleared from the state when this action is dispatched (this also applies to a similar change made with the `SET_ORDER_ID` action type dispatched)?

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
```diff
@@ -0,0 +1,195 @@
...
+				orderId !== undefined
+					? {
+							...state,
+							orderId,
+					  }
+					: state;
+			break;
+		case ACTION.SET_SHOULD_CREATE_ACCOUNT:
+			if (
+				shouldCreateAccount !== undefined &&
```
nerrad, line 169: This is also a logic change from the previous code. Have you verified there are no places in existing code where this might have been undefined and updated the state to clear it (which might be a bug worked around).

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
```diff
@@ -0,0 +1,195 @@
...
+				shouldCreateAccount !== state.shouldCreateAccount
+			) {
+				newState = {
+					...state,
+					shouldCreateAccount,
+				};
+			}
+			break;
+		case ACTION.SET_ORDER_NOTES:
+			if ( orderNotes !== undefined && state.orderNotes !== orderNotes ) {
```
nerrad, line 179: I like the check for undefined here but I also wonder if there's any logic where this action is dispatched that might need updated given the change of behaviour (i.e. not passing in orderNotes property).

I realize that this is partially in place because `CheckoutStateContextState` type explicitly requires `orderNotes` to be present and as a string. However, it appears we have a potential problem with the current type inference (or maybe even how we're creating our actions) because the `setOrderNotes` action is indicating it could return an `ActionType` which _might_ not have `orderNotes` but the action creator explicitly types `orderNotes` as an expected string argument. 

I'm not sure of a solution yet so as long as we're sure there's no unexpected side-effects from the logic change with explicitly checking if `orderNotes` is undefined I'm okay with leaving this as is.

I realize this also applies to all of my other comments about the added `undefined` checks too.

assets/js/base/context/providers/cart-checkout/checkout-state/actions.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Internal dependencies
+ */
+import { ACTION } from './constants';
+import type { PaymentResultDataType } from './types';
+
+export interface ActionType {
+	type: ACTION;
+	data?:
+		| Record< string, unknown >
+		| Record< string, never >
```
opr, line 29: Just wondering what `Record< string, never >` is actually supposed to represent? Is it needed here?

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
```diff
@@ -31,51 +41,15 @@ import { useStoreNotices } from '../../../hooks/use-store-notices';
...
-		setCustomerId: ( id ) => void id,
-		setOrderId: ( id ) => void id,
-		setOrderNotes: ( orderNotes ) => void orderNotes,
-	},
-	hasOrder: false,
-	isCart: false,
-	shouldCreateAccount: false,
-	setShouldCreateAccount: ( value ) => void value,
-} );
+const CheckoutContext = createContext( DEFAULT_CHECKOUT_STATE_DATA );
```
opr, line 50: Nice 😎

assets/js/base/context/providers/cart-checkout/checkout-state/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+): PaymentResultDataType => {
+	const paymentResult = {
+		message: '',
+		paymentStatus: '',
+		redirectUrl: '',
+		paymentDetails: {},
+	} as PaymentResultDataType;
+
+	// payment_result is present in successful responses.
+	if ( 'payment_result' in response ) {
```
opr, line 26: Would it be wise to add additional checks about what `response` is here? Or are we always guaranteed an object?

opr (approved):
Smoke test was fine, no errors. Just left a couple of comments but I don't think they're blocking.

assets/js/base/context/providers/cart-checkout/checkout-state/actions.ts
mikejolley: This allows for empty objects (`{}`).

nerrad:
Just noting I've got an in progress review on this PR.

nerrad (commented):
Really impressive update here Mike 👏🏻 . I especially like how you've reorganized the provider a bit and did things like creating `useEventEmitters` so you can isolate the code and memoization specific to it inside the custom hook and outside of the provider. There's some great improvements in readability here.

I do have some concerns noted in my comments primarily around:

- apparent changes in behaviour for dispatched actions in the `setAfterProcessing` dispatcher.
- apparent changes in behaviour in some of the reducer logic for various actions.

Some of my feedback might expose my newness to TS but I did try to validate some of my thinking via TS playground first. 

I haven't tested this PR at all.

assets/js/base/context/providers/cart-checkout/checkout-state/actions.ts
mikejolley: I'm not sure if this always works—one of the `as const` examples failed eslint checks when the function contained more logic (see https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4206/files#diff-a942a5198733d5fa79775b811f7cfe45bbf4d144bdf209a3334694286a6e49e4).

Why is const better than typing the props? @opr any thoughts here?

nerrad: For reference, [this is where it was suggested](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3768#pullrequestreview-581575534) ([and here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3768#discussion_r568776097))

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
mikejolley: I will add a todo. As you said, it's easy to go down a rabbit hole with this and I don't want to outside of cooldown :)

mikejolley: Change of plan, was not as difficult as I thought. Added in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4200/commits/0a55c8327097b7eb99a4e70199d46b2ec811b42e

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
mikejolley: There are not, no. The usage we have comes from the API and in those cases, they return `0` rather than undefined. `0` would be correct for both the customer ID and order ID if there are none or the user is a guest. We should enforce the prop.

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
nerrad: Nice!

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
mikejolley: It's type unknown, but I can perhaps just swap the undefined check for a type of string check.

mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4200/commits/52fb87931333ea4d9c55c03bd3a4ab202b1b8021

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
mikejolley: Confirmed it's only passed boolean values.

assets/js/base/context/providers/cart-checkout/checkout-state/actions.ts
mikejolley: "const assertions" 

I changed this in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4200/commits/c1b972e8ba3ba1da03270a20b4549f265f909efa

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
mikejolley: It looks like the dispatcher function always accepts a string value here too. `setOrderNotes` is used on one of the components and accepts a value. The undefined check is to satisfy typescript.

assets/js/base/context/providers/cart-checkout/checkout-state/utils.ts
mikejolley: It appears to always be an object—everywhere I see usage of `setAfterProcessing` it's dealing with objects. Those files are not typed yet.

assets/js/base/context/providers/cart-checkout/checkout-state/utils.ts
mikejolley: Added the truth check (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4200/commits/b2dc37e084bd2c7c69134fcb78d91da5e93177a3) and a followup todo to see what the implications of this are (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4200/commits/dd70b86396e04a98e26531bb397069bf7474b029)

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
mikejolley: The old code may have lead you to think there was a mistake here, but I'm fairly confident there is not. 

`setAfterProcessing` is passed an API response. This is either an error response or a success response. 

We know that a successful response will always contain the `payment_result` property. This is part of the API schema. So in this case, state is updated.

An error response always contains the `message` property. You can see here, the old code took this and manually set `payment_result` and then updated state. https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a5add04cd927090c356136eaf2920b4d6f1827cc/assets/js/base/context/providers/cart-checkout/checkout-state/index.js#L186-L196

`getPaymentResultFromCheckoutResponse` has just consolidated the above into a single function with an expected shape and a fallback error message if the response was invalid.

Am I missing something else?

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
mikejolley: Looks like the null can be removed. 

The `as PaymentResultDataType` tells Typescript that data is of that type, rather than `Record<string, unknown>` - the reducer function accepts various different types of data:

```
data?:
		| Record< string, unknown >
		| Record< string, never >
		| PaymentResultDataType;
```

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
nerrad: 👍🏻 Explanation helps. I think I missed that `setAfterProcessing` is currently only called by the checkout processor (which is passing in the API response). There is the assumption that going forward checkout processor will be the only consumer of this dispatcher though. I _think_ the addition of type safety here with TypeScript could help with this though?

nerrad (approved):
Thanks for addressing all my feedback Mike! I'm relying on yours and Thomas's testing of the pr and giving my 👍🏻 for the code. Really great work on this!

mikejolley:
Did a final smoke test with free orders, paid orders, and express payment (stripe, in edge) and all appears to be working :) Merging!

PR #4202, opened 2021-05-13
Don't use fallback in getSetting if value is falsy

### LINKED ISSUE #4201: Disabling Coupons Doesn’t hide the Coupon Code Option in the Checkout Block
## Describe the bug
Even when coupons are disabled under **WooCommerce > Settings > General**, there is still an option to add a coupon in the Checkout block.

## To reproduce

1. Under **WooCommerce > Settings > General**, make sure that "Enable coupons" is unchecked
2. Create a page, add a Checkout block to it, and save it
3. View the page from the front end

## Expected behavior
The option to add a coupon should be hidden.

## Screenshots
Coupons Are Disabled In General Settings
Coupon Setting

Coupons Are Still an Option in the Checkout Block
Page with Checkout Block

### WordPress:
* Core version: 5.7.1
* WooCommerce version: 5.2.2
* WooCommerce Blocks version: 5.1.0
* Site language: en_US
  * Companion Plugin: by Osk – 1.18

### Desktop:
* OS: macOS 11.3.1 (20E241)
* Browser: Chrome 90.0.4430.93 (Official Build) (x86_64)

## Additional context
* This was reported in [this WooCommerce Blocks forum thread on WordPress.org](https://wordpress.org/support/topic/can-not-disable-coupon-code-option/)
* A JN site displaying the issue can be seen [here](https://melted-goldfish.jurassic.ninja/product/test-product/). It will be destroyed around May 18, 2021. For ease of checking, there's a link there in the menu to a page with a Checkout block.

Aljullu:
Fixes #4201.

With the changes introduced in #4059, any setting which had a falsy value would not be evaluated and the fallback would be used. For example, `getSetting( 'couponsEnabled', true )` would always evaluate to `true` even if coupons were disabled in the store.

Adding it to the 5.2.0 milestone since this is fixing a regression.

1. Go to WooCommerce > Settings > General and uncheck Enable the use of coupon codes.
2. Go to the Cart or Checkout blocks and verify there is no Coupon panel in the sidebar.

### Changelog

> Prevent Coupon code panel from appearing in stores were coupons are disabled.

### DIFF
--- a/assets/js/settings/shared/test/get-setting.js
+++ b/assets/js/settings/shared/test/get-setting.js
@@ -11,6 +11,9 @@ describe( 'getSetting', () => {
 	it( 'returns expected value for existing setting', () => {
 		expect( getSetting( 'adminUrl', 'not this' ) ).toEqual( ADMIN_URL );
 	} );
+	it( "returns expected value for existing setting even if it's a falsy value", () => {
+		expect( getSetting( 'currentUserIsAdmin', true ) ).toBe( false );
+	} );
 	it( 'filters value via provided filter callback', () => {
 		expect( getSetting( 'some value', 'default', () => 42 ) ).toBe( 42 );
 	} );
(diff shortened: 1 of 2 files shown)


senadir (commented):
Code and testing checks out.

senadir (approved):

PR #4203, opened 2021-05-13
Set default store currency to USD in tests

Aljullu:
After a change in WC Core that changes the default currency from `GBP` to `USD` (https://github.com/woocommerce/woocommerce/pull/29752), we need to update a PHP test which relies on it.

1. Verify PHP tests pass.


mikejolley (approved):
Failing tests are e2e and unrelated so approving 👍🏻

PR #4204, opened 2021-05-13
Checkout State Provider: Export Context Module Functions

mikejolley:
Taken from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3983 Based on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4200 which has the Typescript updates.

Inspired by epic react I added some new functions exported from the context provider which take dispatch as a parameter and do something. These were previously exported by the context provider, but with this method we only need to export the dispatcher which means the functions can be tree-shaken.

- submitCheckout
- resetCheckout
- setRedirectUrl
- setHasError
- incrementCalculating
- decrementCalculating
- setCustomerId
- setOrderId
- setOrderNotes
- setShouldCreateAccount
- setAfterProcessing

Example usage:

```
import {
	useCheckoutContext,
	setOrderNotes,
} from '@woocommerce/base-context/cart-checkout/checkout-state';

...

const { dispatch } = useCheckoutContext();

setOrderNotes( dispatch, value );
```

EpicReact was fuzzy on the tangible benefits ([linking through to this tweet](https://twitter.com/dan_abramov/status/1125774170154065920) and citing potential performance improvements), but the reasons given seem to include:

- The context provider should only pass down necessary props via the context API, and in this case, `dispatch` is the only thing required
- Having exported functions that accept the dispatch object means that the exports can be tree shaken
- Whilst the functions do require dispatch, it's arguably easier to just import what you need, rather than getting everything via context
- I guess you could also argue that in the future if similar context providers had similar or overlapping functionality, they could share the same functions but accept different dispatch props.

Looking for feedback on this approach, and are we happy how these functions are exported?

Smoke test the checkout process, including making payment. Checkout state provider is only used during checkout.

### DIFF
--- a/assets/js/base/context/hooks/cart/use-store-cart-item-quantity.ts
+++ b/assets/js/base/context/hooks/cart/use-store-cart-item-quantity.ts
@@ -13,7 +13,11 @@ import type { CartItem, StoreCartItemQuantity } from '@woocommerce/types';
  * Internal dependencies
  */
 import { useStoreCart } from './use-store-cart';
-import { useCheckoutContext } from '../../providers/cart-checkout';
+import {
+	useCheckoutContext,
+	incrementCalculating,
+	decrementCalculating,
+} from '../../providers/cart-checkout/checkout-state';
 import {
 	isNumber,
 	isObject,
@@ -55,7 +59,7 @@ export const useStoreCartItemQuantity = (
 		quantity: cartItemQuantity = 1,
 	} = verifiedCartItem;
 	const { cartErrors } = useStoreCart();
-	const { dispatchActions } = useCheckoutContext();
+	const { dispatch: checkoutContextDispatch } = useCheckoutContext();
 
 	// Store quantity in hook state. This is used to keep the UI updated while server request is updated.
 	const [ quantity, setQuantity ] = useState< number >( cartItemQuantity );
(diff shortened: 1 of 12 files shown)


opr (commented):
I don't see anything _wrong_ with this approach, however could you explain better the benefits of doing it this way? I see that now it's tree-shakeable but a description of how that materially benefits us/the merchant/the shopper would be a great addition to this PR.

mikejolley:
@opr I added some notes. It's not totally obvious, just a different way or approaching the same problem I think :) It was from https://epicreact.dev/modules/advanced-react-patterns/context-module-functions-intro

mikejolley:
This has been rebased; should be easier to review now.

mikejolley:
@nerrad do you still think it's worth doing this or just bin it? It doesn't seem that anyone is really opinionated or keen on this. Its stale again.

nerrad:
Sorry for not looking at this sooner Mike, I've taken another look as a result of your ping. In general I think this is a good pattern and at first I thought it'd be okay to switch over to it. However after rethinking this, I'm not sure we get many of the benefits from this pattern with the current implementation because tree-shaking won't apply much with the current way we build the checkout files (everything dependent on context is built to the same file anyways). This seems reflected in the build size report as well (change is minimal). It's hard to see much performance benefit without having any before/after measurements as well. 

Besides that, we're now exposing an implementation detail by passing along the dispatch function which makes it a bit harder to discover how to dispatch actions affecting the checkout unless documentation/code is read and folks either discover the action types available, or the new context module functions receiving this dispatch.

I think if the new functions were re-usable in different contexts this might have more benefit but for now it seems like there's no significant rationale for making the switch. It's just a different approach?

PR #4211, opened 2021-05-16
Fix duplicate plugins by using wp.components global in the editor.

### LINKED ISSUE #3839: Adding a Cart or Checkout Block causes block editor "plugins" to duplicate
## Describe the bug
When editing a page in the block editor that has a "plugin" for the block editor (like Yoast or classic editor), assets seem to get duplicated when adding a cart or checkout block.
Seems that this is limited to only the cart and checkout, but I have not exhausted all blocks in the plugin.
This shows on the "pinned" bar and within the Kebab menu, under the "plugins" section

## To reproduce
1) Add a plugin to a WC Blocks environment that has a block editor "plugin"
      Yoast SEO or Classic Editor are good primary examples. 

2) Edit any page (can be random or assigned cart/checkout pages)
3) Add a cart or checkout block
4) See as the pinned plugin has duplicated, or the plugin menu items have doubled.

## Expected behavior
"Plugins" are not duplicated.

## Screenshots
image

* Core version: 5.5.3 & 5.6
* WooCommerce version: 4.7.0 & 5.0.0
* WooCommerce Blocks version: 4.2.0 & 4.4.2
* Any other plugins installed: Yoast SEO & Classic Editor, issue persists when other is off in both cases.

* OS: macOS
* Browser Safari & Chrome

senadir:
Fixes #3839

This PR fixes the issue with duplicate plugins in the editor.

It does that by creating two output files of `blocks-checkout` one that includes wp.components and one that uses the external value.

The first is loaded in frontend, the second is loaded inside wp-admin.

### Testing instructions
- Install Jetpack or Yoast SEO and enable it.
- Install WC Subscriptions 3.1 or a branch that contains our built integration code.
- Load the editor with Cart or Checkout block, make sure both blocks render fine.
- There should be a single Yoast or Jetpack icon and a single panel only, not two.
- Load the block in frontend, add subscription product to your cart.
- Subscription integration (recurring totals, future shipping) should load fine.
- There should be no console issues.

### DIFF
--- a/bin/webpack-configs.js
+++ b/bin/webpack-configs.js
@@ -22,6 +22,8 @@ const {
 	CHECK_CIRCULAR_DEPS,
 	requestToExternal,
 	requestToHandle,
+	requestToExternalInsideGB,
+	requestToHandleInsideGB,
 	findModuleMatch,
 } = require( './webpack-helpers' );
 
@@ -40,6 +42,20 @@ const sharedPlugins = [
 	} ),
 ].filter( Boolean );
 
+const mainBlocksPlugins = [
+	CHECK_CIRCULAR_DEPS === 'true'
+		? new CircularDependencyPlugin( {
+				exclude: /node_modules/,
+				cwd: process.cwd(),
+				failOnError: 'warn',
+		  } )
+		: false,
+	new DependencyExtractionWebpackPlugin( {
+		injectPolyfill: true,
+		requestToExternal: requestToExternalInsideGB,
+		requestToHandle: requestToHandleInsideGB,
+	} ),
+].filter( Boolean );
 const getProgressBarPluginConfig = ( name, fileSuffix ) => {
 	const isLegacy = fileSuffix && fileSuffix === 'legacy';
 	const progressBarPrefix = isLegacy ? 'Legacy ' : '';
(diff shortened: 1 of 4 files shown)


bin/webpack-helpers.js
```diff
@@ -9,10 +9,7 @@ const CHECK_CIRCULAR_DEPS = process.env.CHECK_CIRCULAR_DEPS || false;
 
 // If a package is not available, or missing functionality, in an old but __supported__ version of WordPress, it should be listed here.
 // Some packages are not available in legacy versions of WordPress, so we don't want to extract them.
-const requiredPackagesInWPLegacy = [
-	// The package included in WP 5.4 version doesn't include `useResizeObserver`. This can be removed when we support 5.5+.
-	'@wordpress/compose',
-];
```
senadir, line 15: we support 5.5+ now.

mikejolley (approved):
This does fix the problem and does not break subscriptions, so this is a 👍🏻  You may want a 2nd code review however to check the changes to the build process.

bin/webpack-configs.js
```diff
@@ -137,6 +153,43 @@ woocommerce_blocks_env = ${ NODE_ENV }
 	};
 };
 
+// This is meant to fix issue #3839 in which we have two instances of SlotFillProvider context. Should be deleted once wordpress/gutenberg#27462.
```
nerrad, line 156: maybe add a `@todo` to this line so an issue is created when this PR is merged? That will make sure we don't forget to remove this when it's no longer needed.

bin/webpack-helpers.js
```diff
@@ -9,10 +9,7 @@ const CHECK_CIRCULAR_DEPS = process.env.CHECK_CIRCULAR_DEPS || false;
 
 // If a package is not available, or missing functionality, in an old but __supported__ version of WordPress, it should be listed here.
 // Some packages are not available in legacy versions of WordPress, so we don't want to extract them.
```
nerrad, line 11: This comment block can probably be removed too?

nerrad (approved):
Changes make sense to me 👍🏻 - I like that this is noted as a temp change until the relevant  I didn't test so I'm leaning on yours and Mike's testing here.

Should also make sure things work as expected in WordPress 5.5 as well (if that hasn't been done already).

bin/webpack-helpers.js
senadir: Not really, `requiredPackagesInWPLegacy` is still relevant for us to use it when we need it, currently, we just don't need it. So I think it makes sense to leave it there. If the comment was relevant to `@wordpress/compose` then I would delete it.

PR #4216, opened 2021-05-17
Fix/e2e fe tests

ralucaStan:
Fix checkout test 
- make tests independent & test also for empty cart case
- make sure Order summary is expanded with better selector

Remove unnecessary localStorage operations in tests.

This PR also touches some jest and puppeteer settings:
- increase puppeteer timeout on CI to prevent timeouts
- add ts to moduleFileExtensions in jest, to allow imports from ts files

Most importantly it removes the interception of console.log in tests with the method `observeConsoleLogging()`.
Reasons for removing:
-  pollutes our test logs with runtime console warnings
-  prevents console.log from inside tests
- it's mostly legacy code, as big part of the logic was done for Puppeteer 1.6.1, but since 3.0.0, `message.text()` returns string so the workaround is not necessary.

### DIFF
--- a/tests/e2e/config/jest.setup.js
+++ b/tests/e2e/config/jest.setup.js
@@ -170,7 +170,7 @@ function observeConsoleLogging() {
 			return;
 		}
 
-		const logFunction = OBSERVED_CONSOLE_MESSAGE_TYPES[ type ];
+		// const logFunction = OBSERVED_CONSOLE_MESSAGE_TYPES[ type ];
 
 		// As of Puppeteer 1.6.1, `message.text()` wrongly returns an object of
 		// type JSHandle for error logging, instead of the expected string.
@@ -195,7 +195,7 @@ function observeConsoleLogging() {
 		// failure.
 
 		// eslint-disable-next-line no-console
-		console[ logFunction ]( text );
+		// console[ logFunction ]( text );
 	} );
 }
 
--- a/tests/e2e/config/jest.setup.js
+++ b/tests/e2e/config/jest.setup.js
@@ -170,7 +170,7 @@ function observeConsoleLogging() {
 			return;
 		}
 
-		// const logFunction = OBSERVED_CONSOLE_MESSAGE_TYPES[ type ];
+		const logFunction = OBSERVED_CONSOLE_MESSAGE_TYPES[ type ];
 
 		// As of Puppeteer 1.6.1, `message.text()` wrongly returns an object of
 		// type JSHandle for error logging, instead of the expected string.
@@ -195,7 +195,7 @@ function observeConsoleLogging() {
 		// failure.
 
 		// eslint-disable-next-line no-console
-		// console[ logFunction ]( text );
+		console[ logFunction ]( text );
 	} );
 }
 
--- a/tests/e2e/config/jest.setup.js
+++ b/tests/e2e/config/jest.setup.js
@@ -1,10 +1,9 @@
 /**
  * External dependencies
  */
-import { get } from 'lodash';
 import {
 	enablePageDialogAccept,
-	isOfflineMode,
+	// isOfflineMode,
 	setBrowserViewport,
 	switchUserToAdmin,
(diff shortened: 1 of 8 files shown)


tests/e2e/specs/frontend/checkout.test.js
```diff
@@ -36,12 +36,7 @@ describe( `${ block.name } Block (frontend)`, () => {
 	let productPermalink;
 
 	beforeAll( async () => {
-		//prevent CartCheckoutCompatibilityNotice from appearing
-		await page.evaluate( () => {
-			localStorage.removeItem(
-				'wc-blocks_dismissed_compatibility_notices'
-			);
-		} );
+		// prevent CartCheckoutCompatibilityNotice from appearing
```
senadir, line 40: Unrelated but is this needed here? CartCheckoutCompatibilityNotice shouldn't appear in checkout to begin with.

senadir (approved):
This looks good so far. Let's see if tests pass.

tests/e2e/specs/frontend/checkout.test.js
ralucaStan: goToCheckoutBlock method goes to the editor, on the Checkout block page to get the permalink in order to navigate to that url

ralucaStan:
@senadir please take another look. 
- I've fixed the failing test,  basically cart was not being emptied before the checkout test, so the empty cart state was never displayed
- I've also added some jest and puppeteer settings with things I've noticed during debugging. Please see the PR description again

tests/e2e/config/jest.setup.js
```diff
@@ -1,18 +1,16 @@
...
-	isOfflineMode,
 	setBrowserViewport,
 	switchUserToAdmin,
 	switchUserToTest,
 	visitAdminPage,
 } from '@wordpress/e2e-test-utils';
 
 // Set the default test timeout.
-let jestTimeoutInMilliSeconds = 40000;
+let jestTimeoutInMilliSeconds = 60000;
```
senadir, line 14: Waiting 60 seconds for a timeout seems excessive, anything that is logical and takes more than 40 seconds?

tests/e2e/config/jest.setup.js
```diff
@@ -108,104 +102,12 @@ function removePageEvents() {
 	} );
 }
 
-/**
- * Adds a page event handler to emit uncaught exception to process if one of
- * the observed console logging types is encountered.
- */
-function observeConsoleLogging() {
```
senadir: Now that you deleted this, it means we won't catch console errors right? is this something that we want? couldn't we just skip `warning` type?

senadir (approved):
So far this looks good to me.

I get the value of cleaning console warning but we would be missing errors emitted by our own blocks unless I'm wrong?

tests/e2e/config/jest.setup.js
ralucaStan: I actually wanted to make it bigger & make sure that if the timeout expires then the element isn't really found on the page. This setting would eliminate the slow UI as a reason and highlight that there is an issue with the page. As this only affects failing tests I don't think the impact is that big. We could try it out and see the outcome.

ralucaStan:
> So far this looks good to me.
> 
> I get the value of cleaning console warning but we would be missing errors emitted by our own blocks unless I'm wrong?

@senadir You make a good point here. I've allowed `console.error` to surface, but only for debugging reasons. It's good to see them if they appear in the logs. The old comment mentioned that `console.error` breaks the tests, but that is no longer the case.

tests/e2e/config/jest.setup.js
ralucaStan: @senadir You make a good point here. I've allowed console.error to surface, but only for debugging reasons. It's good to see them if they appear in the logs. The old comment mentioned that console.error breaks the tests, but that is no longer the case.

senadir:
Ready to ship!

PR #4226, opened 2021-05-18
Convert TextInput and ValidatedTextInput to TypeScript

opr:
Before I get to work on allowing `TextInput` to be used as a `number` input (p1621348983029400-slack-C8X6Q7XQU) I converted the components to TypeScript.

1. Smoke test of the Cart and Checkout blocks.
2. Enter invalid information into checkout fields, ensure the appropriate validation error message is shown.
3. Ensure the error message goes away when you fix the error.


assets/js/base/components/text-input/validated-text-input.tsx
```diff
@@ -39,8 +62,8 @@ const ValidatedTextInput = ( {
 		getValidationErrorId,
 	} = useValidationContext();
 
-	const textInputId = id || 'textinput-' + instanceId;
-	errorId = errorId || textInputId;
+	const textInputId = id !== undefined ? id : 'textinput-' + instanceId;
+	const errorIdString = errorId !== undefined ? errorId : textInputId;
```
opr, line 66: This change was made because TS still couldn't properly infer that after this line, `errorId` was a string. It still thought it would be `string | undefined`.

packages/checkout/label/index.tsx
```diff
@@ -3,13 +3,13 @@
  */
 import { Fragment } from '@wordpress/element';
 import classNames from 'classnames';
-import type { ReactElement, HTMLAttributes } from 'react';
+import type { ReactElement, HTMLProps } from 'react';
 
-interface LabelProps {
+interface LabelProps extends HTMLProps< HTMLElement > {
```
opr, line 8: This lets us pass extra props to `Label` which are not defined in the `LabelProps` interface. The specific prop I needed to add this for was `htmlFor` but it makes sense to allow all valid props through.

assets/js/base/components/text-input/validated-text-input.tsx
```diff
@@ -17,6 +16,29 @@ import { withInstanceId } from '@woocommerce/base-hocs/with-instance-id';
...
+interface ValidatedTextInputPropsWithId {
+	instanceId?: string;
+	id: string;
+}
+
+interface ValidatedTextInputPropsWithInstanceId {
+	instanceId: string;
+	id?: string;
+}
+
```
opr, line 28: Essentially this means: if `id` is passed, `instanceId` is optional, and vice/versa.

assets/js/base/components/text-input/validated-text-input.tsx
```diff
@@ -39,8 +61,8 @@ const ValidatedTextInput = ( {
 		getValidationErrorId,
 	} = useValidationContext();
 
-	const textInputId = id || 'textinput-' + instanceId;
-	errorId = errorId || textInputId;
+	const textInputId = id !== undefined ? id : 'textinput-' + instanceId;
```
Aljullu: I thought ESLint would [discourage](https://eslint.org/docs/rules/no-undefined) using `undefined` directly. For some reason that rule is not enabled, but I wonder if we should be extra cautious and convert this check to use `typeof`:

```suggestion
	const textInputId = typeof id !== 'undefined' ? id : 'textinput-' + instanceId;
```

What do you think?

assets/js/base/components/text-input/validated-text-input.tsx
```diff
@@ -91,15 +113,19 @@ const ValidatedTextInput = ( {
...
 	useEffect( () => {
 		return () => {
-			clearValidationError( errorId );
+			clearValidationError( errorIdString );
 		};
-	}, [ clearValidationError, errorId ] );
+	}, [ clearValidationError, errorIdString ] );
 
-	const errorMessage = getValidationError( errorId ) || {};
+	// TODO - When useValidationContext is converted to TypeScript, remove this cast and use the correct type.
```
Aljullu: In the past we unified all "to do" comments to use this form: `@todo`. That makes it easier to search them in the codebase. :slightly_smiling_face:

Aljullu (approved):
This is testing well and code changes look good, nice to see more components being ported to TS! I left a couple of minor comments, but LGTM.

assets/js/base/components/text-input/validated-text-input.tsx
opr:
```suggestion
	// @todo - When useValidationContext is converted to TypeScript, remove this cast and use the correct type.
```

PR #4228, opened 2021-05-18
Update Express Payments Loading UI

### LINKED ISSUE #4159: Block checkout and cart block UI while express payment method is processing

While using the Payment Request Button in Stripe, the UI is blocked by the Payment Request API dialog or Apple Pay dialog, but after the payment is confirmed, there's a small delay until you get redirected to the Order received page. This allows shoppers to be able to click the Payment Request Button again, leading to bad UX.

There's also a bigger delay when you use 3D Secure cards like `4000002500003155`, after you authorize the transaction and wait for a response.

If you click the button again before the confirmation + redirect, it prevents the `window.location` change. So you might get stuck at the checkout page after the order has been processed.

More information in [this GitHub thread](https://github.com/woocommerce/woocommerce-gateway-stripe/pull/1467#discussion_r624078652).

The credit card payment button gets blocked when you use the express payment method, but the same should apply for all payment methods, even after the Payment Request API processing has completed, but still waiting 3DS confirmation.

For express payment methods, perhaps adding an overlay mask would suffice. 

Any design inputs are welcome.

mikejolley:
This PR makes some improvements to the Express Payments area and related components, notably:

- Split out the loading spinner from our button component so it can be used in the loading mask. This means we use one style of spinner animation instead of 2 :)
- Fixed LoadingMask to:
   - a) Block mouse interaction when masked
   - b) Not force children to re-render if loading state changes
- `useShallowEqual` now returns a default value instead of undefined. Updated typescript hints to match.
- Added loading mask to the express payments area
- Prevented excessive re-renders in express payments area by adding memorization to it's dependencies

Fixes #4159

To test this properly you'll need Stripe setup locally in sandbox mode. You can test the express payments with Chrome/Edge Pay, or Apple Pay in Safari.

1. Add an item to the cart
2. Go to checkout and wait for express payment methods to display
3. Toggle to change shipping; express payments should show the loading spinner and be blocked until the request is complete
4. Pay using an express payment method
5. After clicking Pay, the express payments area will show a loading spinner. When the request is complete and the checkout starts to redirect, the express payments area will remain blocked (no spinner).

### Changelog

> Show loading state in the express payments area whilst payment is processing or the page is redirecting.


frontdevde (commented):
EDIT: As of the latest rebase on May 27, I can't reproduce this issue anymore.

~~I'm seeing an issue with Chrome Pay on this branch that I'm not seeing on `trunk`. I have doubled checked this by first finding the issue on this branch. Then switching to `trunk`, doing a clean install I wasn't able to reproduce the issue. Then switching back to this branch doing a clean install the issue resurfaced again.~~

~~I've sent you a DM with a screen recording that I won't share here due to it containing private data. But I'll describe the issue in case we need someone else to see if they can reproduce it.~~

~~**What happened**~~

~~In Chrome, I click the `Buy Now` button. This triggers the Chrome modal for express payment. The `Pay` button is greyed out. I select a `Delivery Address`. This leads to a spinner showing in the Chrome modal for 60 seconds before the modal closes itself / crashes. I click `Buy Now` again, the modal reopens. The `Pay` button is still grey. I select a `Delivery Address`, this time the selection is instant. The `Pay` button turns blue and I can check out successfully. Now the payment is registered in the Stripe backend.~~

~~**What should've happened**~~
~~The modal shouldn't have closed itself. Instead, the selection should've updated and the `Pay` button should've become active.~~

assets/js/blocks/cart-checkout/payment-methods/express-payment-methods.js
```diff
@@ -48,32 +55,42 @@ const ExpressPaymentMethods = () => {
...
+		}
+
+		return entries.map( ( [ id, paymentMethod ] ) => {
+			const expressPaymentMethod = isEditor
+				? paymentMethod.edit
+				: paymentMethod.content;
+			return isValidElement( expressPaymentMethod ) ? (
+				<li key={ id } id={ `express-payment-method-${ id }` }>
+					{ cloneElement( expressPaymentMethod, {
+						...currentPaymentMethodInterface.current,
```
nerrad: I think there may be some risk here with changes in `currentPaymentMethodInterface` not surfacing to payment methods. This could explain some of the unusual behaviour reported in testing.  The reason being that if there are any changes in the data exposed through the `paymentMethodInterface` that also doesn't impact any of the dependencies for this effect, then the payment method element will be still using stale data from the interface.

nerrad (commented):
I'll still do some testing (I haven't yet) but I did see one potential spot that could result in issues (depending on what else triggers re-rendering in the stack).

nerrad (changes requested):
Testing isn't going too well for me in Edge using the Browser pay modal:

When there is a shopper cached in the browser instance (i.e. address details etc will get filled in on loading the checkout). I get an error when setting the shipping address in the browser pay modal. I've verified the same conditions do not produce an error in trunk. I highly suspect this is due to how we're passing through the `paymentMethodInterface` data. Under the right conditions the data is stale.

CleanShot 2021-05-28 at 07 52 57

The other issue is that this isn't adding a loading transition or preventing the express payment button from being clicked again when a card requiring 3DS confirmation is used:

CleanShot 2021-05-28 at 07 56 21

mikejolley:
I removed the `useMemo` and logged a todo to fix the error some people encountered during payment. 

The other issue, 3D Secure in latest Stripe, is not resolved. I added loading state for `waitingForRedirect` status but didn't realise 3DS has it's own process. Whilst the 3DS modal is active, the status is idle, so we have no way to track it currently. We need a new status to solve this.

I've moved this back in progress and removed my assignment. Ill pick it up again next week if no-one gets to it before I'm back from AFK.

mikejolley:
Supporting Stripe changes for `onError` can be found in this branch: https://github.com/woocommerce/woocommerce-gateway-stripe/compare/fix/block-error-handling?expand=1

Without this, the loading state will not clear once you initiate an express payment.

assets/js/base/components/loading-mask/style.scss
```diff
@@ -1,6 +1,7 @@
 .wc-block-components-loading-mask {
 	position: relative;
 	min-height: 18px + $gap;
+	pointer-events: none;
```
mikejolley, line 4: This prevents mouse interaction with a blocked area.

assets/js/base/components/loading-mask/index.js
```diff
@@ -4,12 +4,12 @@
 import { __ } from '@wordpress/i18n';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import { Spinner } from 'wordpress-components';
 
 /**
  * Internal dependencies
  */
 import './style.scss';
+import Spinner from '../spinner';
```
mikejolley, line 12: We had two types of spinner across blocks so I swapped to our version with a different loading icon.

assets/js/base/context/providers/cart-checkout/payment-methods/actions.ts
```diff
@@ -2,106 +2,77 @@
...
+	paymentMethods?: PaymentMethods | ExpressPaymentMethods;
 	shouldSavePaymentMethod?: boolean;
 }
 
 /**
- * Used to dispatch a status update only for the given type.
+ * All the actions that can be dispatched for payment methods.
  */
-export const statusOnly = ( type: STATUS ): { type: STATUS } => ( { type } );
+export const actions = {
```
mikejolley, line 18: I refactored this a bit to match checkoutState (return an actions object rather than individual functions). It just makes it slightly clearer when consuming from context.

assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
```diff
@@ -90,103 +62,58 @@ export const PaymentMethodDataProvider = ( {
...
-					paymentMethods as PaymentMethods
-				)
-			);
-		},
-		[ dispatch ]
-	);
+	const {
+		dispatchActions,
+		setPaymentStatus,
+	} = usePaymentMethodDataDispatchers( dispatch );
```
mikejolley, line 96: I split the dispatchers into a custom hook because this function is large enough.

assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
```diff
@@ -90,103 +62,58 @@ export const PaymentMethodDataProvider = ( {
...
-			dispatch( setShouldSavePaymentMethod( shouldSave ) );
-		},
-		[ dispatch ]
-	);
+	const {
+		activePaymentMethod,
+		activeSavedToken,
+		setActivePaymentMethod,
+		setActiveSavedToken,
+	} = useActivePaymentMethod();
```
mikejolley, line 111: Also split the active method tracking into custom hook to reduce the size of this function. Logic is the same.

assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
```diff
@@ -213,68 +140,24 @@ export const PaymentMethodDataProvider = ( {
...
+			isProcessing: paymentData.currentStatus === STATUS.PROCESSING,
+			isFinished: [
+				STATUS.ERROR,
+				STATUS.FAILED,
+				STATUS.SUCCESS,
+			].includes( paymentData.currentStatus ),
+			hasError: paymentData.currentStatus === STATUS.ERROR,
+			hasFailed: paymentData.currentStatus === STATUS.FAILED,
+			isSuccessful: paymentData.currentStatus === STATUS.SUCCESS,
+			isDoingExpressPayment:
```
mikejolley, line 163: I added this to make it easier to track when we're making payment using an express method. If the status is not pristine, and the method is express, we know this is true. It's being used to block the express area.

assets/js/blocks/cart-checkout/payment-methods/express-payment-methods.js
```diff
@@ -48,32 +55,71 @@ const ExpressPaymentMethods = () => {
 			setPaymentStatus,
 		]
 	);
+
+	/**
+	 * onExpressPaymentClose should be triggered when the express payment process is cancelled or closed.
+	 *
+	 * This restores the active method and returns the state to pristine.
+	 */
 	const onExpressPaymentClose = useCallback( () => {
```
mikejolley, line 65: When we close express payment, we restore previous payment method and set pristine.

assets/js/blocks/cart-checkout/payment-methods/express-payment-methods.js
```diff
@@ -48,32 +55,71 @@ const ExpressPaymentMethods = () => {
...
 		}
 	}, [ setActivePaymentMethod, setPaymentStatus ] );
-	const paymentMethodIds = Object.keys( paymentMethods );
+
+	/**
+	 * onExpressPaymentError should be triggered when the express payment process errors.
+	 *
+	 * This shows an error message then restores the active method and returns the state to pristine.
+	 */
+	const onExpressPaymentError = useCallback(
```
mikejolley, line 78: Error handling is new and allows us to close express payment, restore everything, and show the notice. This can be tested by using Stripe 3d secure and clicking "Fail".

assets/js/blocks/cart-checkout/payment-methods/express-payment/cart-express-payment.js
```diff
@@ -25,17 +39,32 @@ const CartExpressPayment = () => {
...
 	return (
 		<>
-			<div className="wc-block-components-express-payment wc-block-components-express-payment--cart">
-				<div className="wc-block-components-express-payment__content">
-					<StoreNoticesProvider
-						context={ noticeContexts.EXPRESS_PAYMENTS }
-					>
-						<ExpressPaymentMethods />
-					</StoreNoticesProvider>
+			<LoadingMask
```
mikejolley, line 51: This code blocks express payments when payment is in progress :)

assets/js/blocks/cart-checkout/payment-methods/express-payment/checkout-express-payment.js
```diff
@@ -40,34 +52,49 @@ const CheckoutExpressPayment = () => {
...
-							'woo-gutenberg-products-block'
-						) }
-					</Title>
-				</div>
-				<div className="wc-block-components-express-payment__content">
-					<StoreNoticesProvider
-						context={ noticeContexts.EXPRESS_PAYMENTS }
-					>
-						<p>
+			<LoadingMask
```
mikejolley, line 64: Same as cart, this code blocks express payments when payment is in progress :)

assets/js/base/context/providers/cart-checkout/payment-methods/reducer.ts
```diff
@@ -32,17 +32,14 @@ const reducer = (
 ): PaymentMethodDataContextState => {
 	switch ( type ) {
 		case STATUS.STARTED:
-			return state.currentStatus !== STATUS.STARTED
```
mikejolley, line 35: I removed this to fix a side effect of using status started for express methods. If saved methods are used, started status is already set. Therefore, clicking an express method will not clear any payment method data. This change allows it to clear paymentMethodData.

mikejolley: There was a test for this FYI

assets/js/base/components/loading-mask/index.js
```diff
@@ -18,29 +18,27 @@ const LoadingMask = ( {
...
-
 	return (
 		<div
-			className={ classNames(
-				className,
-				'wc-block-components-loading-mask'
-			) }
+			className={ classNames( className, {
+				'wc-block-components-loading-mask': isLoading,
+			} ) }
```
nerrad, line 26: Why the change here to always have a wrapping div (and just changing the classname if loading mask or not)?

assets/js/base/components/loading-mask/style.scss
nerrad: nifty! What about keyboard interaction?

assets/js/base/context/hooks/payment-methods/use-payment-methods.js
```diff
@@ -10,12 +15,26 @@ const usePaymentMethodState = ( express = false ) => {
...
+	const currentPaymentMethods = useShallowEqual( paymentMethods );
+	const currentExpressPaymentMethods = useShallowEqual(
+		expressPaymentMethods
+	);
+	const currentPaymentMethodsInitialized = useShallowEqual(
+		paymentMethodsInitialized
+	);
+	const currentExpressPaymentMethodsInitialized = useShallowEqual(
+		expressPaymentMethodsInitialized
+	);
```
nerrad: `paymentMethodsInitialized` and `expressPaymentMethodsInitialized` are booleans, so the `useShallowEqual` check here is unnecessary.

assets/js/base/context/hooks/use-store-events.ts
```diff
@@ -39,11 +44,13 @@ export const useStoreEvents = (): {
...
 			try {
+				const cartParam = currentStoreCart.current;
+
 				doAction(
 					`experimental__woocommerce_blocks-checkout-${ eventName }`,
 					{
 						...eventParams,
-						storeCart,
+						cartParam,
 					}
```
nerrad: This is a significant change and testing notes don't cover it (and afaict there are no automated tests covering this - although I could be mistaken). Should there be some mention of coverage in testing notes and/or add some automated tests around behaviour?

I do see how this improves the memoization of `dispatchCheckoutEvent` given the high volatility of store cart value. On the surface this seems like there shouldn't be any unintended side-effects of the change given cartParam should always reflect the current store state when `dispatchCheckoutEvent` is invoked.

assets/js/base/context/providers/cart-checkout/payment-methods/constants.ts
```diff
@@ -54,6 +55,7 @@ export const DEFAULT_PAYMENT_METHOD_DATA: PaymentMethodDataContextType = {
 		hasError: false,
 		hasFailed: false,
 		isSuccessful: false,
+		isDoingExpressPayment: false,
```
nerrad, line 58: There is some overlap between `isDoingExpressPayment` and `expressPaymentMethodActive` which is a derived value in the checkout processor component:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/9e4f4f0bc751e34a06f51bf3b84fc454f60b936d/assets/js/base/context/providers/cart-checkout/checkout/processor/index.js#L66-L68

Were you aware of this, and is this something that can/should be reconciled?

This also applies to the logic added for the status in the [payment method data context provider](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4228/files#diff-9fc47288e4cc5aafaa5a6dae4510af2d7139e99d353e47cd1991cc0c70ce8f22R159).

assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-dispatchers.ts
```diff
@@ -0,0 +1,105 @@
...
+			setRegisteredExpressPaymentMethods: ( paymentMethods ) =>
+				void dispatch(
+					actions.setRegisteredExpressPaymentMethods( paymentMethods )
+				),
+			setShouldSavePayment: ( shouldSave ) =>
+				void dispatch(
+					actions.setShouldSavePaymentMethod( shouldSave )
+				),
+		} ),
+		[]
```
nerrad: This ESLint warning should be handled, while the `dispatch` function is provided by the reducer and is [guaranteed to remain the same instance between renders](https://reactjs.org/docs/hooks-reference.html#additional-hooks ), I think there's still value in explicitly indicating it as a dependency to avoid ambiguity and be consistent in our application of dependencies.

assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
```diff
@@ -213,69 +145,63 @@ export const PaymentMethodDataProvider = ( {
...
+				paymentData.currentStatus !== STATUS.PRISTINE &&
+				Object.keys( paymentData.expressPaymentMethods ).includes(
+					activePaymentMethod
+				),
 		} ),
-		[ paymentData.currentStatus ]
+		[
+			paymentData.currentStatus,
+			paymentData.expressPaymentMethods,
```
nerrad: I'm assuming `paymentData.expressPaymentMethods` shouldn't change as often as it would have if you didn't also do the `useShallowEqual` implementation in the payment method initialization logic? Otherwise this would have caused the `currentStatus` object to update more frequently. However, I'm wondering if that's actually a significant problem (and we could remove the memoization of `currentStatus` altogether. I imagine in _most_ (if not all) cases, consumers of `currentStatus` would be referencing the boolean values inside the object and in those cases the referential equality of the entire `currentStatus` object doesn't matter (just a thought, doesn't necessarily warrant changes for now - but something to consider for potential performance improvements when we can measure - memoization has costs too).

assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
```diff
@@ -213,69 +145,63 @@ export const PaymentMethodDataProvider = ( {
...
+			// included in the check here.
+			if (
+				! currentActivePaymentMethod ||
+				! allPaymentMethodKeys.includes( currentActivePaymentMethod )
+			) {
+				setPaymentStatus().pristine();
+				return Object.keys( paymentData.paymentMethods )[ 0 ];
+			}
+			return currentActivePaymentMethod;
+		} );
```
nerrad, line 195: Oh ya, this is a nice improvement from the previous unnecessary `setActivePaymentMethod` wrapper over the state setter in the previous iteration 👏🏻 .  I also like how the custom event dispatcher is also moved to the custom hook.

assets/js/base/context/providers/cart-checkout/payment-methods/reducer.ts
nerrad: Makes sense, and ya it looks like the existing automated test case should cover this. So the reason why the test passed prior to this change is because Express Payment methods didn't change the status?

assets/js/blocks/cart-checkout/checkout/sidebar/test/__snapshots__/index.js.snap
```diff
@@ -58,41 +58,50 @@ exports[`Testing checkout sidebar Shows rate percentages after tax lines if the
...
       >
-        <form
-          class="wc-block-components-totals-coupon__form"
+        <div
+          aria-hidden="false"
+          class=""
         >
           <div
-            class="wc-block-components-text-input wc-block-components-totals-coupon__input is-active"
+            class="wc-block-components-totals-coupon__content"
```
nerrad, line 68: These test snapshots indicate some significant changes to html structure. Will this have impact on any styling customizations that might have been done for theme support? Should that be reflected in pull request notes and testing? (this question also applies to other test snapshot file changes).

assets/js/blocks/cart-checkout/payment-methods/express-payment-methods.js
```diff
@@ -48,32 +55,71 @@ const ExpressPaymentMethods = () => {
...
+			setActivePaymentMethod( previousActivePaymentMethod.current );
+			if ( previousPaymentMethodData.current.isSavedToken ) {
+				setPaymentStatus().started( previousPaymentMethodData.current );
+			}
+		},
+		[ setActivePaymentMethod, setPaymentStatus, setExpressPaymentError ]
+	);
+
+	/**
+	 * @todo Find a way to Memorize Express Payment Method Content
```
nerrad: Term would be more accurately be `Memoize` (see [memoization](https://en.wikipedia.org/wiki/Memoization)) not Memorize (although it's easy to see why "memorize" might be used).

assets/js/blocks/cart-checkout/payment-methods/express-payment-methods.js
```diff
@@ -48,32 +55,71 @@ const ExpressPaymentMethods = () => {
...
 					</li>
 				) : null;
 			} )
 		) : (
-			<li key="noneRegistered">No registered Payment Methods</li>
+			<li key="noneRegistered">
+				{ __(
+					'No registered Payment Methods',
+					'woo-gutenberg-products-block'
+				) }
```
nerrad, line 144: Whew, good catch!

nerrad (commented):
Wow, there's some really nice refactoring in here (really like moving logic out of the payment method context data provider into custom hooks...makes the code easier to read) 👏🏻 

Summary of feedback sprinkled through inline comments:

- Some concerns around change in html structure (as surfaced in test snapshots as well) and how that may impact existing styling. Likely not a problem but still had questions.
- A question around loading interactivity restrictions.
- A _requested_ change around unnecessary use of `useShallowEqual` for boolean values.
- Maybe add some testing notes (or even better, automated tests) to cover change to `dispatchCheckoutEvent`. 
- Might be some opportunity to clean up the overlap between `isDoingExpressPayment` and `expressPaymentMethodActive` (but could be done in a followup).

Besides the above, regarding `onExpressPaymentError`:

- I like the implementation and think it works well.
- I'm not sure on the name `onExpressPaymentError`. I think it shares similarity to the naming structure for other event subscribers but itself _isn't_ a subscriber function. 
- I would like to see `setExpressPaymentError` as the name of the new method and you move in the setting of notices into the new method. I don't think we need to have two separate behaviours here.

I did very basic testing of Stripe (with your branch for the gateway) and as far as I can tell it works great! However, I'm really leaning on your own testing for complete validation of payment processing flow. 

I do plan on trying to setup WooCommerce Payments and test with it too before giving the ✅ .

assets/js/blocks/cart-checkout/checkout/sidebar/test/__snapshots__/index.js.snap
mikejolley: I believe this is the loading wrapper only. I'll comment elsewhere why this is needed.

assets/js/base/components/loading-mask/style.scss
mikejolley: > Elements with pointer-events: none will still receive focus through sequential keyboard navigation using the Tab key.

This blocks mouse interaction only I believe.

Now that loading mask is present, and reflects state if it is confirmed as an edge case issue, we can tackle this in the loading mask component as a follow-up.

assets/js/base/context/providers/cart-checkout/payment-methods/reducer.ts
mikejolley: I think it was that started status immediately changed back to pristine in the previous version, whereas now we need it to stay 'started'.

assets/js/base/components/loading-mask/index.js
mikejolley: I may have mentioned this in Slack the other week when I was stuck. Without the wrapping element, React was unmounting/remounting/rerendering all children when `isLoading` changed which caused bugs further down. With the wrapping div, this doesn't happen.

assets/js/base/context/hooks/payment-methods/use-payment-methods.js
mikejolley: Not sure why I added this; possibly trying to stop rerenders since this uses a ref? Ill try removing it.

assets/js/base/context/providers/cart-checkout/payment-methods/constants.ts
mikejolley: No, I wasn't aware of that one. We can move `expressPaymentMethodActive ` to this provider perhaps but they do have different purposes.

assets/js/base/components/loading-mask/index.js
nerrad: Hmm, given that rationale, I wonder if adding & removing a class results in the same (see this [recent merged PR in GB](https://github.com/WordPress/gutenberg/pull/32567))

mikejolley: Example showing issue https://codesandbox.io/s/es6-forked-up4z8?file=/src/app.js

assets/js/blocks/cart-checkout/checkout/sidebar/test/__snapshots__/index.js.snap
mikejolley: Just to be clear, the markup has not changed in a way that affects styles. It's hard to see in the diff, but there is just a single wrapper being added around these elements.

assets/js/base/context/hooks/use-store-events.ts
mikejolley: I know why I did this (it's used as a dependency in various places and I didn't want to trigger the effects when the cart changed). 

The change was to store the cart as a ref so the callback function can use the ref instead of using cart as a dependency. I did a quick test to make sure I understand refs and callbacks so it will have the latest cart object available (https://codesandbox.io/s/clever-brook-i8fuy?file=/src/app.js).

I know tests would be good, but I'm still not confident writing tests from scratch and I know it will take me the rest of the week to fathom.

assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
mikejolley: Was able to replace this after replacing expressPaymentMethodActive with a boolean in this context provider.

assets/js/base/context/hooks/use-store-events.ts
mikejolley: I've restored the `storeCart` parameter name here; that was an oversight on my part. It's still using the ref.

nerrad (approved):
Looks great Mike! I tested this with WooCommerce Payments and things check out there.

The _only_ remaining issue I noticed is that when the express payment method is processing after completing the 3DS verification, the "Place Order" button is still clickable (and it's possible to edit fields in the checkout. The only thing blocked is the express payment method area. We probably should be blocking the entire checkout?

Up to you whether you handle this in a followup or not but I do think it'll need addressed. Pre-approving pending your decision about the last point.

assets/js/base/context/hooks/use-store-events.ts
nerrad: > I know tests would be good, but I'm still not confident writing tests from scratch and I know it will take me the rest of the week to fathom.

I'm okay with it not blocking the PR, but I do think it'd be worthwhile doing this as a followup. We need more confidence when changing things and automated tests are a good way to get that.

nerrad (approved):
I tested the latest changes with WooCommerce Payments, Stripe Extension, built in blocks stripe integration and also with the Stripe Extension onClose fix branch.  I tested GooglePay (via chrome) and Browser Payment Request (via Edge). As far as I could tell most behaviour works okay.

However, in my testing I am seeing some issues with switching shipping address in the express payment method and that rates being updated. While I'm reproducing this in trunk, it does give me pause to rely on just my testing for this. My setup is:

- display taxes inclusive in Cart and Checkout
- Different shipping zones with different rates based on address and postal code.
- When I trigger GooglePay, I'll either won't get all the expected rate options initially (and a different rate selected) or I will get the expected rate options but switching the address in Google Pay does not update the rates as expected.
- Rates in the Browser Pay modal update just fine.
- I'm reproducing this for both built-in Stripe integration in blocks AND the Stripe extension integration. Further, with the Stripe extension I'm also reproducing on the _shortcode flows_.

So it'd be good to have someone else test and:

- see if you can reproduce what I experienced above in this branch.
- If yes to the above, see if you can reproduce in trunk too. If yes, then not a blocker to merging this PR. 
- If you can't reproduce my issue at all, then let me know, we'll have to sync up on what might be in my configuration that's triggering it (or perhaps I have something borked on my db).

I'm pre-approving because other than the above issue, things test well. I do think additional testing to verify whether the above issue I've reported is valid or not though.

mikejolley:
@nerrad I seem to only have issues when using **this branch** and **Stripe Trunk**. I see matching behaviour between trunk and this branch when using my Stripe onClose handling Branch. Does this line up with what you are seeing?

In the broken setup, there is no initial shipping address.

For the working setup, I get an initial rate, and can rate switch, and paying has the correct amount, but I do see the wrong amount on the checkout in the background.

mikejolley:
@nerrad I got it working by adding `useCallback` - those functions were not stable so it must have been upsetting google pay. Fixed now. I get shipping in Google Pay, but it won't accept test cards to run through the full flow.

Check your side?

nerrad:
Just leaving a comment here to note that Mike and I did some synchronous testing together to validate and clear the issues I reported. All was good.

PR #4238, opened 2021-05-19
Move `TextInput` to checkout package and allow it to be used for input type=number

opr:
As part of the Points and Rewards integration, I found the need to expose several of our UI components. This PR will make `TextInput` available via the `@woocommerce/blocks-checkout` package, and also allow it to be used for number inputs.

To allow additional props to be passed to the `<input>` element, I added a new prop to `ValidatedTextInput` and `TextInput` called `inputProps`, which will be spread onto the `<input>` element in `TextInput`.

Because of accessing the validation context from a different package, the validation functions returned by `useValidationContext` were not available, so I modified the `ValidatedTextInput` component to accept the functions as props, and also the `ValidatedTextInput` component.

1. Smoke test the cart/checkout flow.
2. Try adding valid _and_ invalid coupons in both the Cart and Checkout blocks.
3. Ensure validation still works in the Checkout form, invalid emails, empty fields etc.
4. Modify `assets/js/base/components/cart-checkout/address-form/address-form.js` and add the following at the top of the file and toward the top of the `AddressForm` function respectively:
 ```typescript
```
```typescript
const [ testValue, setTestValue ] = useState( 0 );
```
then output the following just before the closing `</div>` of the returned component:
```typescript
<ValidatedTextInput
  label={ 'Test input' }
  type={ 'number' }
  { ...textInputValidationFunctions }
  onChange={ setTestValue }
  value={ testValue }
  step={ 2 }
  min={ 0 }
  max={ 10 }
/>
``` 
5. The Address form should have an input of type `number` now, and the extra props you put in `inputProps` should be attributes on the element.

### Changelog

> Move `ValidatedTextInput` to the `@woocommerce/blocks-checkout` package.


packages/checkout/text-input/text-input.tsx
```diff
@@ -26,6 +26,10 @@ interface TextInputProps
 	autoComplete?: string;
 	onChange: ( newValue: string ) => void;
 	onBlur?: ( newValue: string ) => void;
+	inputProps?: Omit<
```
ralucaStan: I'd recommend not having this extra level of inputProps. Ideally, we should use the component with
`<TextInput type="number" max="" min="" />` for a more intuitive API. Just like [here](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/text-control/index.js).

packages/checkout/text-input/text-input.tsx
```diff
@@ -49,18 +53,18 @@ const TextInput = forwardRef< HTMLInputElement, TextInputProps >(
...
 		const [ isActive, setIsActive ] = useState( false );
-
 		return (
 			<div
 				className={ classnames(
 					'wc-block-components-text-input',
 					className,
 					{
-						'is-active': isActive || value,
+						'is-active': isActive || value || value === 0,
```
ralucaStan: If this `value === 0` only reffers to the number type than I'd add that in the condition as well: `value === 0 && type === 'number'`

assets/js/base/components/cart-checkout/address-form/address-form.js
```diff
@@ -74,10 +77,21 @@ const AddressForm = ( {
...
+		hideValidationError,
 	} = useValidationContext();
 
+	const textInputValidationFunctions = {
+		getValidationError,
+		getValidationErrorId,
+		setValidationErrors,
+		clearValidationError,
+		hideValidationError,
+		validationInputError: ValidationInputError,
```
ralucaStan: Looks like ValidationInputError is a component so I would not pass it here along with these functions. From what I see it's usually passed as `inputErrorComponent`:
	
```
	<InlineCard
			onChange={ onChange }
			inputErrorComponent={ ValidationInputError }
		/>
```

ralucaStan (commented):
I've added some comments. Let me know what you think.

packages/checkout/text-input/text-input.tsx
opr: I suppose this is better, the extra level of props was kind of pointless now that I think about it. Cheers 👏🏼

assets/js/base/components/cart-checkout/address-form/address-form.js
opr: 😍 Excellent spot and suggestion, thanks!

ralucaStan (approved):
👍

PR #4240, opened 2021-05-19
Add key to each `CartTotalItem`

### LINKED ISSUE #4148: Add item type to cartTotalItems
It is common in European countries to display item price including tax. It would be helpful if the `cartTotalItems` that is part of the `usePaymentMethodInterface` included the item type. If the item type is known, it will assist with payment method logic.

In the function `prepareTotalItems`, add a `type` property that describes what the line item is. The function already has access to the line item type via the `factory` function so that could be used to add the type to the line item. 

Example:
```javascript
const factory = ( label, property ) => {
    const value = parseInt( totals[ property ], 10 );
    const tax = parseInt( totals[ property + '_tax' ], 10 );	
    return {
        label,
        value,
        valueWithTax: value + tax,
        type: property
    };
};
```

- Using the label of the item to determine its type is unreliable because of translations. 
- The tax item could be determined by adding the difference of `item.tax - item.valueWithTax` and comparing that value to all line items. But that doesn't tell you what the other line items are.

mikejolley:
Adds a `key` prop to each `CartTotalItem` generated by `prepareTotalItems`.

Whilst doing this, I converted `prepareTotalItems` to Typescript, and moved our guards to `assets/js/types/type-guards` which uses the `@woocommerce/types` alias for ease of access.

Fixes #4148

Smoke test cart and checkout. The key prop is added in `prepareTotalItems` but is not visible. I'm not sure how best to test this; perhaps console logging `usePaymentMethodInterface`?

### Changelog

> Added a key prop to each `CartTotalItem` within `usePaymentMethodInterface `


packages/prices/utils/price.ts
```diff
@@ -2,13 +2,11 @@
...
- * Internal dependencies
- */
-import type { Currency } from '../types';
+import type {
+	Currency,
+	CurrencyResponse,
+	CartShippingPackageShippingRate,
+} from '@woocommerce/types';
 
 type SymbolPosition = 'left' | 'left_space' | 'right' | 'right_space';
```
opr: Maybe we should move this to the `@woocommerce/types` package too and update https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/5261c4e9a733d5292ab40027a6ad6b3d61952e73/assets/js/settings/shared/settings-init.ts#L15 to use this type instead of the strings

assets/js/base/context/hooks/payment-methods/use-payment-method-interface.ts
```diff
@@ -171,10 +95,25 @@ export const usePaymentMethodInterface = () => {
 			displayPricesIncludingTax: getSetting(
 				'displayCartPricesIncludingTax',
 				false
-			),
+			) as boolean,
```
opr, line 98: not blocking for this PR, but wondering if we'd prefer to change `getSetting` to accept a type argument, so we could change this to 
```typescript
getSetting< boolean >(
				'displayCartPricesIncludingTax',
				false
			)
```
I don't know if it makes very much difference, but it seems to be a little cleaner. 🤷🏻

opr (approved):
I can see those keys when I log `usePaymentMethodInterface` 👏🏼 . Couple of non-blocking comments.

assets/js/base/context/hooks/payment-methods/use-payment-method-interface.ts
opr: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4250 issue created for this so no need to action here :)

PR #4241, opened 2021-05-19
Add preliminary extensibility points for Product Bundles support (Cart block)

### LINKED ISSUE #3895: Add support for Product Bundles
For many years in core the only way to depict a hierarchical (parent/child) relationship between two cart items (=bundling) was to handle things an a plugin-specific manner, which has always been a pain to support and maintain, and led to an inconsistent user experience, depending on the plugin used and the active theme.

The move to an API-driven experience unavoidably "locks down" the extensibility of the cart/checkout pages, and moves a lot of the responsibility to decide "what's possible" to the core team. It's a radical change that requires everyone to work a lot more closely together than before. I'm sure extending a product such as WC Subscriptions to work with these blocks will make this statement clearer :)

I am not sure if this is something that can/should ever be formalized. To start with, I am opening an issue to track what's needed here, and perhaps propose a couple changes that would allow plugins like Product Bundles to be supported by the Cart/Checkout blocks without touching the React components downstream.

(Disclaimer: These have been validated/requested by users in the last decade that Product Bundles has existed. There's a good chance I might be missing something that's vital to others.)

Being able to visually represent a parent/child relationship between cart items basically means developers need to:

* depict the grouping relationship between parent + child items (indentation, font size, etc -- see "Additional context" below);
* have the option to keep certain child item quantities synced with the parent item quantity (= update child item quantities when the parent quantity changes, based on response);
* have the option to hide child item quantity inputs;
* have the option to display an aggregated parent+child item price + total at the parent level;
* have the option to hide child item prices + totals;
* have the option to exclude certain child items from the displayed cart item count;
* have the option to completely hide certain child items.

Without looking at the existing cart item component(s), I assume that there are many things about the current cart item component definition that disregard these concepts. At the bare minimum, one solution that doesn't require an opinionated rewrite of these components would be to:

1. Allow developers to control the visibility of certain cart items -- perhaps include a `visibility` prop that takes the `woocommerce_cart_item_visible` filter into account, or override `CartSchema::get_item_responses_from_schema` to include only "visible" cart items?
2. Allow the displayed cart item `prices` + `totals` to be filtered inside `CartItemSchema::get_item_response`. To my understanding, these affect nothing but the "view" side of things.
3. Give developers a way to inject custom `tr` classes -- perhaps pass a classes prop down to the component from the API, and honor the existing `woocommerce_cart_item_class` filter. I understand "classes" are not object properties, as they affect the view side of things directly -- but we may need to treat them as such to provide some basic creative freedom to third party developers.
4. When changing the quantity of a cart item, update the quantity of all cart items based on the quantity that comes with the response. 

## Additional context

Two examples, for reference.

### Case 1

**PHP Cart**

Image 2021-02-28 at 10 47 47 a m

**React Cart**

Image 2021-02-28 at 10 48 56 a m
 
### Case 2

**PHP Cart**

Image 2021-02-28 at 10 49 37 a m

Note the "Subtotal" string and "↳" symbol that prepends prices. The need to filter/change subtotals to include additional information is a very real one. For example, WooCommerce Subscriptions includes extra data such as sign up fees and subscription details. I can't wrap my head around a solution when working with an API that is designed to pass raw prices to a front-end.

PS: If the solution here is to let people filter things using JS Filters, then I can't help but think we are re-building the same kind of experience using tools that come with a barrier of entry + complexity increased by the 10fold :) We built a version of hooks in Composite Products 5 years ago, and although those JS hooks gave a lot of creative freedom to developers in a largely familiar way, a lot of hand-holding was necessary to help people understand what to do with them.

**React Cart**

Image 2021-02-28 at 10 52 48 a m

manospsyx:
This PR introduces some new extensibility points to add Cart/Checkout Blocks support in Product Bundles, Composite Products, and likely other extensions, such as Min/Max Quantities.

Ref #3895 

This doesn't close #3895 yet - it includes some of the basic extensibility points we've identified after studying the _documented_ stories and scenarios of PB. As the development of these integrations moves forward, it is likely that more missing bits and pieces will appear.

Ideally, some of these changes should/could have been opened in individual PRs, however:

- To make reviewing easier I have clearly separated each change in its own commit.
- Since all of these changes are required by #3895 it makes a lot of sense to review them together. With the exception of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/af91ae8ba159c496da3661c215854eda4ab19118 all other commits are probably low-risk and rather small in scope. Let me know if you’d rather see a separate PR for each commit.

### Screenshots

Image 2021-05-19 at 5 34 57 p m

1. Clone PB and checkout branch `issue-758` (ping me for repo access). Alternatively, I can send a link to a pre-release version (still in flux).
2. Create a Bundle like [this](https://user-images.githubusercontent.com/1783726/118832114-04005500-b8c9-11eb-97af-2978861bc3fc.png). Leave the **General > Regular/Sale Price** fields empty, but make sure that at least one bundled item is **Priced Individually**.
3. Add the bundle to your cart.

At this point:

- Bundled line items should appear indented.
- Bundled product quantities should not be freely editable and will stay in sync with "container" quantity.
- When the "container" item quantity is changed, bundled item quantities should be updated to reflect the new quantities returned by the API. This has been discussed in #3895.
- The "container" item should list the aggregated bundled product prices + subtotals.

A couple notes:

> When the "container" item quantity is changed, bundled item quantities should be updated to reflect the new quantities returned by the API

See https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/af91ae8ba159c496da3661c215854eda4ab19118 -- I am not entirely confident with my approach here. My goal here was to make sure that the quantity passed down from that API in `lineItem` is fed into the `QuantitySelector` component unless there's a pending quantity update triggered by user input. See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/files#diff-d4fa4fa6c42594eee9e65823c4d7fb7c92ba45afa28d4696baced7131bbc6770R303

To know whether "there's a pending quantity update triggered by user input" some modifications to the `QuantitySelector` component were necessary, because there's some built-in throttling (debouncing) there that prevents `isPending.quantity` from returning `true` immediately after a quantity increment/decrement: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/files#diff-164a1373b934919b2a907a1fede87c8c91c3298de7ca55a2bd35ed98a814d3c0R144

This part needs a solid review. 

> Bundled product quantities should not be freely editable.

Min/step handling in the quantity selector component had to go all the way to the API. This is a vital requirement that in my mind shouldn't be left to third parties to implement (this needs to be supported at component level anyway). Note that I haven't added any filters in the API schema class yet (maybe I should?): https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/files#diff-4a0ace31a5843c5b16d88d6081f20dee749c9d5309dff0b7c0504bd1482217dbR327-R328

cc @nerrad 

### Changelog

> Added 'cartItemClass' filter for cart items.
> Introduced product title formatting using the 'productNameFormat' filter.
> Added cart item min quantity and quantity step support in the store API and quantity selector component.
> Wrapped cart item product td elements in div for greater styling flexibility.


assets/js/base/components/quantity-selector/index.tsx
```diff
@@ -78,19 +81,25 @@ const QuantitySelector = ( {
...
+				min={ minimum }
 				value={ quantity }
 				onKeyDown={ quantityInputOnKeyDown }
 				onChange={ ( event ) => {
-					let value =
-						! isNumber( event.target.value ) || ! event.target.value
-							? 0
-							: parseInt( event.target.value, 10 );
+
+					let valueIsNumber = event.target.value && ( isNumber( event.target.value ) || parseInt( event.target.value, 10 ).toString() === event.target.value );
```
manospsyx:
```
let valueIsNumber = event.target.value && parseInt( event.target.value, 10 ).toString() === event.target.value;
```

This should be enough -- I believe `isNumber( event.target.value )` was incorrectly used here as `event.target.value` is always `string`, not `number`.

If you do a quick test on `trunk` you'll notice that every time you enter a value inside the input field, it always gets converted to `1`, because `isNumber( event.target.value )` always returns `false`.

manospsyx:
I noticed some JS Unit tests are failing: I'm not very familiar with Jest, snapshot testing, React, or TS -- apologies if there's more I could have done here that I didn't :)

senadir:
>  noticed some JS Unit tests are failing: I'm not very familiar with Jest, snapshot testing, React, or TS -- apologies if there's more I could have done here that I didn't :)

No worries, we can look into that once we finish reviewing, it's most luckily that we need to update some snapshots given the HTML structure changed.

manospsyx:
@senadir I see, thank you! 

The diff does look scary on GH, until you realize it's just the effect of indentation. I believe the only markup change is this one here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/04dfe4f0ac0decd2302845c4c24f059d3492e274 -- adds a wrapper element.

[This commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/d0ee06666d7174c5dc3dcbda2843a8eb0621d918) is only adding the necessary indentation, with the added bonus of messing up the diff. It will be a lot easier on your eyes if you look at the commits one by one, and ignore that one ;)

senadir:
👋🏼 I'm going to review this PR next week, is it possible to grant me access to the PB repository?

manospsyx:
> I'm going to review this PR next week, is it possible to grant me access to the PB repository?

Thanks so much :) Access granted.

All work on cart/checkout blocks support is in `issue-758` -- [this diff](https://github.com/somewherewarm/woocommerce-product-bundles/compare/issue-745...issue-758) should give you a quick overview of how we have approached things. 

To build the `assets` directory, you'll need to:

```
npm install
grunt build
```

Then, you should be good to go. 

Work on the Checkout Block is still underway but should be completed next week -- the approach we'll be aiming for there will be to list bundled items as a container item meta, similar to how we've done it in the mini-cart: https://a.cl.ly/v1ubdwr8

Feel free to ping me here or in the WC Community Slack channel if you come across any roadblocks :)

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
```diff
@@ -116,6 +116,11 @@ const CartLineItemRow = ( {
 		[]
 	);
 
+	const cartItemClassNameValidation = useCallback(
+		( value ) => mustBeString( value ),
+		[]
+	);
+
```
senadir: This type of validation shouldn't be needed since we implemented same type validation on the actual filter already in https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/ad33f7d16a527b239c78937a0a90b12674b36ef8/packages/checkout/registry/index.ts#L86-L98

However, I'd probably not tackle this filter this way (as a string) but more as an array, `classnames` handles array just as well.

For the validation, to make sure other extensions don't mistakenly remove your value, we should make sure the previous value still exist.

```ts
/**
 * Checks if a value of type array passed contains all previous values
 */
export const mustContainPreivousValues = < T >(
	value: T[],
	previousValue: T[]
): true | Error => {
	if ( ! previousValue.every( ( item: T ) => value.includes( item ) ) ) {
		throw Error(
			__(
				'Returned value must include all original items.',
			)
		);
	}
	return true;
};
```

This requires a bit more changes in `__experimentalApplyCheckoutFilter` as well, which I can push later along side the above function if you feel this is right.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -52,6 +52,11 @@ const OrderSummaryItem = ( { cartItem } ) => {
 		[]
 	);
 
+	const productNameValidation = useCallback(
+		( value ) => mustBeString( value ) && mustContain( value, '<name/>' ),
```
senadir, line 56: As I said in my previous comment, `mustBeString( value )` shouldn't be needed.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
```diff
@@ -247,6 +247,9 @@ const CartLineItemRow = ( {
 				) }
 			</td>
 			<td className="wc-block-cart-item__product">
+				<div
+					className={ classnames( 'wc-block-cart-item__wrap' ) }
```
senadir: We prefer to not change the markup or add additional divs unless needed, what's the value here in adding this wrapper that was not possible with `wc-block-cart-item__product`?

assets/js/base/components/quantity-selector/index.tsx
senadir: Parsing and then calling `toString` doesn't make sense to me here,  are trying to test if calling `parseInt` on a non-number would return `NaN` which you would then parse back to compare

What exactly is the point of this logic?

I'd be lying if I said the original was as obvious 😅 4 lines of codes has been changed by 4 different people.

> I believe isNumber( event.target.value ) was incorrectly used here as event.target.value is always string, not number.

I agree, I believe this is a regression in [this PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4088) @opr 

@franticpsyx before expanding on this function logic, I'd prefer we tame it down a bit. This can be first fixing the `isNumber` issue (by using `isNaN` instead).
@opr is the `isNumber` call needed here? switching to `isNaN( parseInt( event.target.value, 10 ) )` doesn't throw any typescript errors for me. It seems `isNaN is typed wrongly in that it only accepts `number` while it should accept everything.

@franticpsyx Would this logic still work? What I go from your changes is that you're trying to use `quantity` instead of `0`, in that case, this should be enough?

```js
let value = parseInt( event.target.value, 10 );
value =
	isNaN( value ) || ! event.target.value
		? quantity
		: value;
```

assets/js/base/components/quantity-selector/index.tsx
```diff
@@ -78,19 +81,25 @@ const QuantitySelector = ( {
...
+					let value         = ! valueIsNumber ? quantity : parseInt( event.target.value, 10 );
+
 					if ( hasMaximum ) {
-						value = Math.min( value, maximum );
+						value = Math.min( value, Math.floor( maximum / step ) * step );
```
senadir: This assumes `maximum` isn't already a multiple of `step`? I can see cases where this might happen, can you add a comment expanding on this logic?

assets/js/base/components/quantity-selector/index.tsx
```diff
@@ -78,19 +81,25 @@ const QuantitySelector = ( {
...
+					}
+
+					if ( hasStep ) {
+						value = value % step ? quantity : value;
```
senadir: `hasStep` seems redudant and unexplantory, 1 is still a step, you should probably delete `hasStep` and move the logic here so it's `if ( step > 1 )`

Also, a comment here explaining why what's happening is preferred, since it took me a bit of time to get it.

From what I'm seeing, if the value inserted is not a multiple of `step`, we refuse it.

senadir: Another issue I'm seeing is that we no longer can type values, if `step` is 3, you can't type in 12, or 15, because 1 is smaller than 3 and so our value would be refused.

What you should do is more `hasStep` logic to `onBlur` or debounce it somehow.

src/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -50,8 +50,20 @@ public function get_properties() {
...
+			],
 			'quantity_limit'       => [
-				'description' => __( 'The maximum quantity than can be added to the cart at once.', 'woo-gutenberg-products-block' ),
+				'description' => __( 'The maximum quantity that can be added to the cart at once.', 'woo-gutenberg-products-block' ),
+				'type'        => 'integer',
+				'context'     => [ 'view', 'edit' ],
+				'readonly'    => true,
+			],
+			'quantity_step'       => [
+				'description' => __( 'The quantity step of this item in the cart.', 'woo-gutenberg-products-block' ),
```
senadir, line 66: I can't really give an honest review if this makes sense or not, cc @mikejolley do you feel like adding those two fields make sense here?

Use cases for this are plugins that has bundles or purchase limitation.

Such limitation would be created on the server, currently, a call would be made but the server would return a notice.

src/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -312,6 +324,8 @@ public function get_item_response( $cart_item ) {
 			'key'                  => $cart_item['key'],
 			'id'                   => $product->get_id(),
 			'quantity'             => wc_stock_amount( $cart_item['quantity'] ),
+			'quantity_min'         => 1,
+			'quantity_step'        => 1,
```
senadir: What's the point of hardcoding this, how can plugins like yours access this? is it available somewhere on the cart item?

senadir (changes requested):
Hey @franticpsyx

I did an overall review, I have some comments left about things to change.
I was pulled into other priority work last week and didn't have a chance to review it, it seems my invite was expired so I couldn't test this with PB, can you send a new invite.

The only commit I didn't test yet was `Honor api changes to cart item quantities when rendering cart line items`, I feel it might not be needed once we fix https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4289 or we might need to change it here.

assets/js/base/components/quantity-selector/index.tsx
opr: Yes I agree, we shouldn't be using `isNumber` here after all. `isNaN` works 👍🏼

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
manospsyx: > I'd probably not tackle this filter this way (as a string) but more as an array

Sure thing, happy to change this. How would you like to handle array validation?

manospsyx: > For the validation, to make sure other extensions don't mistakenly remove your value

Removing classes could be desirable in certain cases. Such kind of interplay is common, as it allows complex integrations to be maintain on one side only. 

For example a three-way integration between 2 extensions and WC core may require extension B to un-do something that extension A did -- extension A shouldn't necessarily be aware of B's existence.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
manospsyx: This builds upon the existing validation pattern of `productPriceValidation`: This is a "product title" formatting filter, not a classname filter.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
manospsyx: A wrapper div element like that is a life-saver there, as it provides much-needed flexibility, for example with flexbox.

Padding on `td` elements such as `wc-block-cart-item__product` cannot be reliably overridden -- themes typically define their own table-cell padding, which makes it impossible for plugins to consistently indent all elements contained in that `td`.

assets/js/base/components/quantity-selector/index.tsx
manospsyx: `isNaN` should be fine.

`parseInt( event.target.value, 10 ).toString() === event.target.value` is just a more verbose attempt to do the same, while only accepting integer values. As far as I could tell, the codebase in its current form was not designed to handle floats (?) all the way from that API to the front-end.

assets/js/base/components/quantity-selector/index.tsx
manospsyx: > This assumes `maximum` isn't already a multiple of `step`?

It was written to account for that, yes. I'm not sure if this is front-end jurisdiction, though.

assets/js/base/components/quantity-selector/index.tsx
manospsyx: > Another issue I'm seeing is that we no longer can type values

`quantity` is the "current" input quantity. If `step` is 3, you should be able to type 3, 6, 9, 12 because `value % step` would equal 0, and therefore `value` would preserve its value.

src/StoreApi/Schemas/CartItemSchema.php
manospsyx: Not quite sure if these need to be part of the schema. If not, then we'll probably need to filter the values fed into the `QuantitySelector` component from `CartLineItemRow`.

src/StoreApi/Schemas/CartItemSchema.php
manospsyx: I wasn't 100% certain if these should be part of the API, and therefore didn't add any filters there.

See

> Note that I haven't added any filters in the API schema class yet (maybe I should?)

In our initial implementation we are filtering the response via WP REST API filters -- eventually I'd love to see a way to filter certain cart line fields on the PHP side. For example, product permalinks commonly need to be amended with query arg params that restore their cart configuration when clicked in the cart (WC core already does that with variations).

src/StoreApi/Schemas/CartItemSchema.php
manospsyx: > Such limitation would be created on the server, currently, a call would be made but the server would return a notice.

That's why I initially thought that quantity input field attributes like this should be part of the API -- to let the API validate the values. I'm not sure if that's sound thinking now that I'm looking at it again.

If this stays out of the API, how should third parties validate quantity changes?

manospsyx:
> The only commit I didn't test yet was Honor api changes to cart item quantities when rendering cart line items, I feel it might not be needed once we fix #4289 or we might need to change it here.

Did you perhaps mean to link to another issue? I'm not sure how the flickering issue described in #4289 could be related with the issue I described + attempted to solve in https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/af91ae8ba159c496da3661c215854eda4ab19118.

The problem is that quantities passed down from that API in `lineItem` are not reflected in the updated cart view at all. 

When Bundle "container" quantities get updated, "child item" quantities are synced in the back-end with the container value. The synced value gets passed down to the app correctly, but the way quantity inputs maintain state does not account for this.

This affects pretty much every plugin that synchronizes cart item quantities one way or another -- you could experiment by writing some back-end code that keeps that quantity of Product B synced with the quantity of Product A by a pre-defined multiple. You will notice that when you update the quantity of Product A in `trunk`, Product B's quantity stays the same in the front-end. You need to do a full cart page refresh to see the right quantity next to Product B. The work in https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/af91ae8ba159c496da3661c215854eda4ab19118 sorts it out, but I'm not confident with it as I have a hunch that there might be a better way to do it.

@senadir happy to address your feedback about the cart item classes filter and perhaps make another couple tweaks based on some of your other points. 

As you said, we may need to wait for more folks to chime in to determine how to handle some of the more philosophical aspects of this (e.g. define min/step fields in API or let third parties filter those?).

senadir:
> Did you perhaps mean to link to another issue? I'm not sure how the flickering issue described in #4289 could be related with the issue I described + attempted to solve in af91ae8.

Flickering means the datastore is not waiting for the new value to be updated, which could cause the value to not reflect on the frontend, I made sure to fix this but it seems to still happen.

For the next steps, I'm going to attempt and fix #4289 which would hopefully reduce the need to add the extra logic you had.

> happy to address your feedback about the cart item classes filter and perhaps make another couple tweaks based on some of your other points.

I'm going to push a commit to this branch that contains the updated logic around `ItemClassName`.


We still have the issue of item quantity not being writable if the step is larger than 1. Try setting step to 3 and write 12 as a value. So we need to simplify the logic in `onChange` and handle that issue.

For `step` and `min`, cc @nerrad when you have time to review PHP changes in commit 66b053424c11ca310069ec3a0883884a3e541c82

nerrad (commented):
Thanks for taking this on Manos and contributing!

Forgive me if this is something that has already been thought about and accounted for, but there's a few questions I have around the current work here:

- For the `productNameFormat` filter, is this something that could be filtered server side in the StoreAPI?
- Nadir and I chatted about the additions to the CartItemSchema for `quantity_min` and `quantity_step` and we think it'd be better to have these live as a part of `WC_Cart` and then the StoreAPI would simply read from this source. Filtering etc could be provided there and the logic is better encapsulated as a part of the Cart.
- One thing to consider is how `step` and `min` values apply to calculations around available stock and allowed backorders. It isn't clear that this is something that has been considered here?

manospsyx:
Hey @nerrad 👋 

Thanks to both for helping out :)

> For the productNameFormat filter, is this something that could be filtered server side in the StoreAPI?

The filter makes modifications like [this one](https://a.cl.ly/6qur7APO) possible. Without it, third parties can only change the product name that gets wrapped in the permalink. It's a small difference, but one that ensures that the product link is preserved in the right context.

> Nadir and I chatted about the additions to the CartItemSchema for quantity_min and quantity_step and we think it'd be better to have these live as a part of `WC_Cart` and then the StoreAPI would simply read from this source. Filtering etc could be provided there and the logic is better encapsulated as a part of the Cart.

Usually we want to:

1. Build UIs that enforce certain rules/behaviors.
2. Introduce validation in the back end to make sure that people don't get out of line :) 

If I understand correctly, you suggest that back-end validation could be handled via `WC_Cart` hooks? This would definitely work without "polluting" the API.

At the same time, developers will need to have a way to set those `input` attributes. Do you think it would be a good idea to make those filterable at the point they are passed into the `QuantitySelector` component?

 > One thing to consider is how `step` and `min` values apply to calculations around available stock and allowed backorders. It isn't clear that this is something that has been considered here?

Min should always be below max, and the "accepted" value in-between min and max. There are some special cases that we'd need to handle here, for example:

- min = 5
- max = 15
- step = 20

This wouldn't play out very well. When making min/max/step filterable, we can't just trust developers to enter good values -- we'll need to decide which ones to accept when they are not all in agreement. 

I'd probably require that 

1. 0 <= min
2. 0 < max
3. max < max_stock_qty
4. min <= max
5. min <= step <= max

I'm not sure how the code currently handles (2), but we should be able to handle the rest. If (4) isn't satisfied, we'll make min=max, and if (5) is not satisfied, we could ignore step.

It's up to developers to make sure they double-check their logic if they expect it to work in its entirety.

I could add those missing bits; but how would you suggest that we let extension code modify min/max/step? Via JS filters?

manospsyx:
@senadir 

> Try setting step to 3 and write 12 as a value.

I see what you mean now. I expected that the existing code would run only when clicking out of the input -- it seems to be wired up slightly differently than I'm used to. I'd need some serious React schooling to contribute meaningfully here, but will give it a shot.

senadir:
> I see what you mean now. I expected that the existing code would run only when clicking out of the input -- it seems to be wired up slightly differently than I'm used to. I'd need some serious React schooling to contribute meaningfully here, but will give it a shot.

That's okay, the event you need to hook into is `onBlur`, that event fires after the user is done editing and unfocus the input, but it's going to be complicated since you also need to revert any change that happened in onChange.

I can look into that once we decide on how we should proceed with `step` and `min`.

senadir:
@franticpsyx I'm unable to push to `somewherewarm/woocommerce-gutenberg-products-block` can you enable contributors on that repo or just add me to it?

manospsyx:
In jQuery I'd normally handle this like:

```
.on( 'input change', 'input', function( event ) {

	var $input = $( this ),
		qty    = parseFloat( $input.val() ),
		min    = parseFloat( $input.attr( 'min' ) ),
		max    = parseFloat( $input.attr( 'max' ) );

	if ( 'change' === event.type ) {
		// Rewrite input values here.
	}
}
```

The callback wired up to `onChange=` seems to fire when I type.

EDIT: Interesting, React seems to internally hook onChange to `input` events!

> I'm unable to push to `somewherewarm/woocommerce-gutenberg-products-block` can you enable contributors on that repo or just add me to it?

Sure thing, let me try that!

Been working on removing the min/step API fields, if @nerrad and you both agree I could make a commit to add filters with proper validation for min/max/step.

senadir:
> Been working on removing the min/step API fields, if @nerrad and you both agree I could make a commit to add filters with proper validation for min/max/step.

I'm actually not so sure about that, I had a communication mismatch with Darrent about this, so could you please hold on any changes until we're fully cleared on the next step here.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
```diff
@@ -180,33 +176,36 @@ const CartLineItemRow = ( {
...
+		Array< string >
+	>( {
 		filterName: 'cartItemClass',
-		defaultValue: '',
+		defaultValue: [],
 		extensions,
 		arg,
-		validation: cartItemClassNameValidation,
+		validation: ( value, previousValue ) =>
+			mustContainPreivousValues< string >( value, previousValue ),
```
manospsyx, line 189: @senadir just a tiny typo here: `mustContainPreivousValues` => `mustContainPreviousValues`

manospsyx:
For reference, I started trying out something like this in `cart-line-item-row.tsx`:

```
	const quantityMaxValidation = useCallback(
		( value ) => {
			if ( value <= 0 || value > quantityLimit ) {
				throw Error(
					sprintf(
						/* translators: %s is type of value passed */
						__(
							'Returned value must greater than zero, and less than or equal to %1$d. You passed %2$d.',
						),
						quantityLimit, value
					)
				);
			}

			return true;
		},
		[]
	);

	const quantityMax = __experimentalApplyCheckoutFilter( {
		filterName: 'cartItemQuantityMax',
		defaultValue: quantityLimit,
		extensions,
		arg,
		validation: quantityMaxValidation,
	} );

	const quantityStepValidation = useCallback(
		( value ) => {
			if ( value <= 0 || value > quantityMax ) {
				throw Error(
					sprintf(
						/* translators: %s is type of value passed */
						__(
							'Returned value must greater than zero, and less than or equal to %1$d. You passed %2$d.',
						),
						quantityMax, value
					)
				);
			}
			return true;
		},
		[]
	);

	const quantityStep = __experimentalApplyCheckoutFilter( {
		filterName: 'cartItemQuantityStep',
		defaultValue: 1,
		extensions,
		arg,
		validation: quantityStepValidation,
	} );

	const quantityMinValidation = useCallback(
		( value ) => {
			if ( value < quantityStep || value % quantityStep > 0 || value > quantityMax ) {
				throw Error(
					sprintf(
						/* translators: %s is type of value passed */
						__(
							'Returned value must be a multiple of %1$d, greater than or equal to %2$d, and less than or equal to %3$d. You passed %4$d.',
						),
						quantityStep, quantityStep, quantityMax, value
					)
				);
			}
			return true;
		},
		[]
	);

	const quantityMin = __experimentalApplyCheckoutFilter( {
		filterName: 'cartItemQuantityMin',
		defaultValue: quantityStep,
		extensions,
		arg,
		validation: quantityMinValidation,
	} );
```

- The max quantity is constrained by `quantityLimit` which provides the initial value.
- The step is constrained by the max quantity and must be greater than zero.
- The min quantity must be greater or equal to step and lower or equal to max. 

These filtered values are then passed into the `QuantitySelector` component.

To validate quantity changes in the back end, I'd probably use `woocommerce_cart_item_set_quantity` in a way that tracks whether the request originated from the Cart API, and throw a `RouteException`. I haven't tested this to see if the error would be picked up and displayed correctly in the front-end though.

senadir:
Hey Manos, given the current state of PR, do we need those filters? If you can define Step and Min on the server, should that be enough?

manospsyx:
> Hey Manos, given the current state of PR, do we need those filters? If you can define Step and Min on the server, should that be enough?

@senadir noap, we don't need both. I added those filters to give you an idea of an alternative approach that would also get us covered from an extensibility perspective.

I'm torn between the two -- it's one of those things that may justify a chat between the API Lead + PM. The more I think about it, the more I tend to believe that extensibility requirements like this shouldn't define/shape the Cart API :)

senadir:
Hey Manos,
I had a chat this morning with Darren to define the requirements exactly, this is what we came out with:

### Store API
Adding step and min quantity felt like an edge but a bit justified one for the time being. I'd like us to add them to the store API but reserve the right to remove them if we believe they should live elsewhere. I'm going to check other plugins that control quantities and see the requirements for each plugin.

### Source of Min and Step
This is something I'm not fully certain about. I'm inclined to believe that this information should live on the cart item level and not the product level. Given your use case, having it on the item level makes sense since a product can be bought individually which a regular 1 step or within a bundle. 

So, in the API, we're going to fetch them like this

```php
'quantity_min'  => isset( $cart_item['quantity_min'] ) ? isset( $cart_item['quantity_min'] ) : 1,
'quantity_step' => isset( $cart_item['quantity_step'] ) ? isset( $cart_item['quantity_step'] ) : 1
```

### Changing the data
By having them on the item level, we don't need new filters, any filter that already accesses the cart data or reacts to it can set them, an example is `woocommerce_cart_add_item` or `woocommerce_cart_loaded_from_cart`

```php
add_filter(
	'woocommerce_add_cart_item',
	function( $cart_item ) {
			$cart_item['quantity_step'] = 3;
			$cart_item['quantity_min']  = 3;
		return $cart_item;
	}
);
add_action(
	'woocommerce_cart_loaded_from_session',
	function( $cart ) {
		foreach ( $cart->cart_contents as $key => $cart_item ) {
			$cart->cart_contents[ $key ]['quantity_step'] = 3;
			$cart->cart_contents[ $key ]['quantity_min']  = 3;
		}
	}
);
```

### Frontend
I'd push against using any extra frontend filters beyond what's already available, plugins need to manipulate the step or min quality on the frontend. I can see a need to change the way quality is set but that's a different topic.

So what would we do is iron out the frontend implementation so that it's resilient to any step value or min quantity.

manospsyx:
I'm 100% onboard with this.

> I'm going to check other plugins that control quantities and see the requirements for each plugin.

The most popular one that could leverage this from your lineup would be **Min/Max Quantities**. Another one would be **Force Sells**. Others: Composite Products, Mix and Match Products, and likely more grouping/bundling plugins in the wild.

> So, in the API, we're going to fetch them like this

💯 

> So what would we do is iron out the frontend implementation so that it's resilient to any step value or min quantity.

Let me take a look and see what's missing!

manospsyx:
Hey @senadir 

I have committed some more code based on your earlier suggestions and findings. 

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/44129a9a64f3752bf4d76eca7341c4f54aca133f -- here I'm providing initial values to `quantity_min` and `quantity_step` in the back-end, based on your suggestion to check the cart item data array for the existence of these keys.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/fae7d2a82afe182e0eee0e5e235b2e9cb6ab0c3e -- added a check to ensure we don't set a min/step value for sold individually products.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/be537e8941aa0fd5fc9604533fd9b17582721afe -- based on the previous discussion and since we're formalizing the use of these fields in the API, I thought we may also want to add some validation.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/81486f330d0518a09d996d75872c505e7ee37e4d -- this is an attempt to:

- do a sanity check with min/max/step values passed into the `QuantitySelector` component;
- fix the shortcomings of validating quantity input values while typing. 

Re the latter -- I have tried to validate the input value when the element loses focus, as you suggested: Not sure how this will play out with touchscreens, though. The need to decouple the value rendered by the Component from the current/validated qty value meant that we needed to introduce an additional property, resetQuantity, which onBlur will fall back to if the input value is invalid. Hope it's not all wrong :) 

See https://a.cl.ly/YEuZKmBk

helgatheviking:
Awesome work @manospsyx !!

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx
manospsyx: @senadir I haven't managed to find a good CSS workaround for this one -- even involved CSS guru @xristos3490 .

We'll need this wrapper to do stuff [like this](https://user-images.githubusercontent.com/1783726/118831630-97855600-b8c8-11eb-8906-c7f9ada769e7.png).

senadir:
Hey Manos, I think we can close this now given we have other most of the work in other PRs.

PR #4248, opened 2021-05-20
Add Slot/Fill to discounts area in cart sidebar

### LINKED ISSUE #4205: Add SlotFill to the `Discounts` area of the new Cart and Checkout sidebar
As part of the integration with WooCommerce Points and Rewards, we would like to move away from the current way Notices are being used by Points and Rewards to alert users to how many points they can redeem. (In Blocks only)

We had this design recommendation:

To implement this we will need to add a SlotFill into which Points and Rewards can insert the input field.

opr:
This PR adds a Slot/Fill to the Cart sidebar (The same functionality is required in Checkout, but this will come in another PR). Extensions can register components to be fill this Slot, it is rendered just underneath the `TotalsCoupon`  component.

The PR also includes a (partial) revert of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4238 (specifically, the part where `ValidatedTextInput` was moved to the checkout package) but keeps the part where styles, and a spread were added to `ValidatedTextInput` to handle the cases when this component is used as `type="number"`.

**⚠️ Please ignore the changes in `packages/checkout/utils/update-cart-from-api.ts`** - these are changed again in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4298 which is based off this branch.

Fixes #4205 

1. Check out `feature/add-inputs-for-points-redemption` from https://github.com/woocommerce/woocommerce-points-and-rewards/ and ensure you've run `npm run build` in that project.
2. Go to the file `includes/class-wc-points-rewards-extend-store-endpoint.php` in Points and Rewards and comment out the following lines (66-72):
```php
self::$extend->register_update_callback(
  array(
    'endpoint'          => CartExtensionsSchema::IDENTIFIER,
    'namespace'         => self::IDENTIFIER,
    'callback_function' => array( 'WC_Points_Rewards_Cart_Checkout', 'rest_apply_discount' ),
  )
);
```
3. Ensure you have added points to your user account ( WooCommerce -> Points and Rewards) and that you're logged in 
4. Add items to your cart and visit the Cart block. See that the Points and Rewards input is present. Note that it won't function on this branch. This PR is just to get the display working.

### Changelog

> Add Slot in the Discounts section of the cart sidebar to allow third party extensions to render their own components there.


assets/js/blocks/cart-checkout/cart/full-cart/index.tsx
```diff
@@ -114,6 +114,18 @@ const Cart = ( { attributes }: CartProps ) => {
...
+		extensions,
+		cart,
+		contexts: {
+			validation: useValidationContext(),
+			storeCart: useStoreCart(),
+		},
+		components: {
+			ValidationInputError,
+		},
+	};
```
nerrad: Instead of doing this which is a side-effect of exporting `ValidatedTextInput` from the checkout package (which wouldn't have access to the same validation context instance). Could you avoid removing the encapsulation of the validation logic in `ValidatedTextInput` and have the component passed through on the slot-fill here? So you'd end up with something like:

```suggestion
	const discountsSlotFillProps = {
		extensions,
		cart,
		contexts: {
			storeCart: useStoreCart(),
		},
		components: {
			ValidatedTextInput
		},
	};
```

The concern I have with the current implementation is it exposes more of the implementation details to extensions and makes it harder for us to make changes in the future.

nerrad: I also have a high level of concern around exposing the entire cart implementation details here. I realize points and rewards redemption requires some ability to update the cart but this again exposes implementation details that makes it harder for us to modify how this works in the future - especially when in the [current implementation](https://github.com/woocommerce/woocommerce-points-and-rewards/pull/426/files#diff-887f8916eac678e0237153f2b3c99f800a46465825202cb0e4f1d30d713cfb28R35-R39) there's some pretty high dependency on [making sure the right shape for the cart data](https://github.com/woocommerce/woocommerce-points-and-rewards/pull/426/files#diff-184b109d487795364c256a34343c3fa20257a671debff646fa7e4bc31ebbe638R202) is fed to the `receiveCart` dispatcher.

This is an interface that would be fairly hard for extensions to understand and apply correctly without deep knowledge of internals.

I don't have any immediate alternatives to suggest but I'd be wary about shipping with the current implementation.

opr: Thanks Darren, could you clarify what you mean about the `ValidatedTextInput` change here? This component renders a `ValidationInputError` which is (at the moment) passed to `ValidatedTextInput` as a prop. Previously it was imported from the `base-context` package, but since moving the input to the checkout package, the context instance is different one, so errors don't show up. Are you suggesting to move `ValidatedTextInput` back out of the checkout package and instead pass it through the slot?

Also, right now I no longer need the `useStoreCart` context, following the addition of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4248/files#diff-e9030d5520892c9926d801219c89bbc21b4863a510ce18409f74b0a148f1270eR1-R24, BUT, the caveat is, after updating the cart, the extension needs to request fresh data to be loaded, which has a pretty poor UX especially on slower devices or connections: https://www.loom.com/share/8a9871aeb0b5494a84cfc54f84c00dfd

nerrad: > Are you suggesting to move ValidatedTextInput back out of the checkout package and instead pass it through the slot?

Yes this. That would also remove the need to expose implementation details from the validation context right?

nerrad: > BUT, the caveat is, after updating the cart, the extension needs to request fresh data to be loaded, which has a pretty poor UX especially on slower devices or connections: https://www.loom.com/share/8a9871aeb0b5494a84cfc54f84c00dfd

Ya that's not the best UX. I wonder if instead of doing this, we could provide some API that signals the checkout should do some sort of "update_cart" post to the server on the StoreAPI that extensions could react to server side and make any necessary updates? So it'd work similarly to when quantity is updated etc except extensions are simply triggering the post and response? So the rough flow would be something like:

1. Extension updates a value attached to it's state in the cart store.
2. Extension uses some way of signaling it's state needs updated in on the server.
3. This triggers a checkout "update" POST.
4. Extension is hooked into the StoreAPI "update" POST for capturing the data it needs to process server side and preparing the response.
5. StoreAPI returns the updated Cart response (and the extension could react to value changes in it's cart store state)?

opr: Thanks for your feedback @nerrad, as we have discussed in our 1:1 and in [this PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4298), this is the approach I think we'll continue with. 

I have reverted the changes in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4238 and instead opted to pass this through the slot, due to us not wanting to pass validation contexts through the slot.

This PR is now ready for a full review 🎉

assets/js/base/components/cart-checkout/totals/coupon/style.scss
```diff
@@ -17,6 +17,10 @@
 		padding-right: $gap-large;
 		white-space: nowrap;
 	}
+
+	.wc-block-components-totals-coupon__button.no-margin {
```
Aljullu: Not blocking, but I'm not a big fan of 3rd-party extensions using our internal class names, this introduces another breakage point every time we change the styles of this class. Ideally, I would go for one of these approaches:

* Extract the coupon button (or the entire Coupon form layout) into its own component and export it from `@woocommerce/checkout`. This way extensions could create a button with the same styles without the need to call the CSS class name directly.
* In this specific case, maybe just copy & paste the styles of `.wc-block-components-totals-coupon__button` into the extension CSS with another selector name?

packages/checkout/panel/style.scss
```diff
@@ -6,6 +6,13 @@
 	}
 }
 
+.wc-block-components-panel.has-border.no-top-border {
+	@include with-translucent-border( 1px 0 );
+	&::after {
+		border-top-width: 0;
+	}
+}
+
```
Aljullu: You should be able to accomplish the same with the mixin:

```suggestion
.wc-block-components-panel.has-border.no-top-border {
	@include with-translucent-border( 0 0 1px );
}
```

I also think the selector could be renamed from `.wc-block-components-panel.has-border.no-top-border` to `.wc-block-components-panel.has-bottom-border`.

And, ideally this would be exported from a `TotalsWrapper` component so 3rd-party extensions don't need to call our CSS class names. Sorry if I'm too annoying with that! :smile: It's not blocking this PR, though.

Aljullu (approved):
Changes look good. I left a couple of CSS suggestions but I'm approving because none is blocking.

One thing I noticed is that when I run `npm run build` in Points and Rewards, I can't see the rewards panel, it only works when I run `npm run start`. Is that a known issue?

Another thing is that text doesn't seem to wrap correctly:
imatge

assets/js/base/components/text-input/text-input.tsx
```diff
@@ -87,6 +88,7 @@ const TextInput = forwardRef< HTMLInputElement, TextInputProps >(
 							: ariaDescribedBy
 					}
 					required={ required }
+					{ ...rest }
```
senadir: What was the goal of doing `{ ...rest }`? this is quite the change and exposes a lot of implementation detail to the user.
It also means the user might overwrite several events like onChange, onFocus, and onBlur.
You should make sure to merge events or just be wary of what you're exposing.

assets/js/blocks/cart-checkout/cart/full-cart/index.tsx
```diff
@@ -114,6 +114,14 @@ const Cart = ( { attributes }: CartProps ) => {
 		cart,
 	};
 
+	const discountsSlotFillProps = {
+		extensions,
+		cart,
+		components: {
+			ValidatedTextInput,
+		},
```
senadir: I'm still not sold on this `components` pattern tbh. Why are we doing it this way? Do people need this exact input with its validation context and all? can't just they get a regular input and handle validation themselves?

packages/checkout/discounts-meta/index.js
```diff
@@ -0,0 +1,32 @@
...
+import { createSlotFill } from '../slot';
+
+const slotName = '__experimentalDiscountsMeta';
+
+const {
+	Fill: ExperimentalDiscountsMeta,
+	Slot: DiscountsMetaSlot,
+} = createSlotFill( slotName );
+
+const Slot = ( { className, extensions, cart, components, contexts } ) => {
```
senadir: Context was deleted as far as I remember.

Aljullu:
Oh, one more thing: are the up and down arrows expected in the number input? They collide with the placeholder. I would personally remove them, but leaving the decision to you:

imatge

senadir (commented):
I do have some concerns about the amount of things we're exposing here and if they make sense for us to do that. I did leave some questions.

packages/checkout/discounts-meta/index.js
opr: Yep, you're right it was. Removing 🗑️

assets/js/base/components/text-input/text-input.tsx
opr: It was so we could have step, min, and max attributes on the input field, an alternative we could offer is to just check if those specific props are passed, check if the type is number then spread only those three onto the input 🤔 let me try that.

opr: Updated with this change. Seems to be working fine

packages/checkout/panel/style.scss
opr: Yep, the `TotalsWrapper` change is on my to-do list! I think for now I can completely remove `no-top-border` and work on the `TotalsWrapper` changes. In the meantime there will be a double border.

image

assets/js/blocks/cart-checkout/cart/full-cart/index.tsx
opr: I guess I got too bogged down here, I think there's actually no reason to pass this component through... let me get rid of it and see how it goes 👌🏼 .

nerrad: Just noting that Thomas and I chatted about this today. A couple reasons why this was passed through was to:

- Expose (what would likely be a common type of) component for extension developers to use to preserve consistent styling and behaviour.
- Make it easier for extension developers to implement this kind of functionality.

However, the tradeoffs are that we are establishing a contract to preserve this component and kicks off a pattern that might be hard to maintain in the future if extension developers have need for other types of components in the checkout (checkbox etc). There's some element here of just including this component because Points & Rewards needs it.

So for now, we agreed that we'll hold off exposing this and Thomas will implement the functionality directly in Points & Rewards. I do think we'll still need to address the issue of exposing some components/style system for extension developers to use so they can inherit/preserve the current (and any evolution of the) design of the blocks which makes for a better shopper experience. The work Thomas will be doing in Points and Rewards implementation will unblock that integration but will be more inherently fragile in relation to potential changes in the Cart and Checkout design down the road (for example when we potentially start including Global Styles from Gutenberg to allow for more customization options by merchants etc). So I think we will need to address this in a followup.

senadir: > Expose (what would likely be a common type of) component for extension developers to use to preserve consistent styling and behaviour.
> Make it easier for extension developers to implement this kind of functionality.

Why wouldn't having the component in `@woocommerce/blocks-checkout` be enough? previously we passed shippingSelector from there to preserve an already passed context I guess, but that pattern is up for reconsideration during this cooldown in which we might delete the shipping methods slotFill for a registration API.

In my mind, `components` was used to pass components that are harder to move to the checkout package or are very opinionated to a specific case.

I'm not against exposing components in general via a components package, in that case, we have a real boundary of "all code in this folder is public" type of protection, making it easier for us to maintain it.

nerrad: This is exactly the sort of discussion that we need to work through separately now that we are starting to see integrations that could use more of these components (whether published via package, or passed through as an interface prop etc).

Aljullu (commented):
Thanks for the updating this PR @opr! Changes look good to me. I noticed there is an issue in Firefox (input field lets you write numbers and on unfocus the label collides with the input):

imatge

I would say we don't need to fix this since this is a [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1398528) and IMO an edge-case most users will not encounter. But wanted to raise it in here in case you think differently.

I will leave the final approval of this PR to @senadir given that he made other comments.

opr:
> Thanks for the updating this PR @opr! Changes look good to me. I noticed there is an issue in Firefox (input field lets you write numbers and on unfocus the label collides with the input):
> 
> imatge
> 
> I would say we don't need to fix this since this is a [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1398528) and IMO an edge-case most users will not encounter. But wanted to raise it in here in case you think differently.
> 
> I will leave the final approval of this PR to @senadir given that he made other comments.

Yep, you can also write the letter `e` in the box and it will do the same in chromium, I don't think there's anything we can do to stop this besides revert to a `type="text"` input and `parseInt` on the value, but this gives degraded performance to mobile users as it won't display the numerical keyboard. I think this is an edge case we are OK with leaving as fixing it would result in a worse overall experience for more users.

assets/js/blocks/cart-checkout/cart/full-cart/index.tsx
opr: I already tried moving the `ValidatedTextInput` component to the checkout package, but it didn't work with the validation context, because they were in different packages it seemed they couldn't work together. Then we tried passing the context through the slot, and that worked, but wasn't the right approach to take as we didn't want to expose this context to third parties.

For now I'm sticking with what Darren has already said - we will duplicate the component in P&R until we can revisit this and find a way to expose `ValidatedTextInput` to extensions.

senadir (approved):
All good to me.
I wonder if it just made sense to create a new number field instead of having the logic inside the text field. I'd leave this to a follow up for now.

Thank you for addressing the feedback Thomas!

PR #4268, opened 2021-05-25
WIP - Checkout i2 Feature Branch Tracking

### LINKED ISSUE #4444: Checkout i2: Prepare Feature Branch for Merge (with experimental flag)
To avoid the Checkout i2 Feature Branch getting too large, we should consider merging it to trunk with an experimental flag.

Prerequisites (to be investigated as part of this issue):

- Ensure any changes in this branch do not impact the stable Checkout Block
- Ensure Checkout i2 is marked experimental and does not conflict with the existing Checkout Block
- Feature parity between Checkout i2 and Checkout Blocks - ensure Checkout i2 is functional
- Code review/testing Checkout functionality

Once merged we can adopt a proper PR/Review workflow for all Checkout i2 work.

mikejolley:
We'd like to merge the new Checkout i2 block into trunk flagged "experimental" to avoid this PR getting much larger now that many of the features are implemented. The block itself is not yet complete.

You can scan over the Checkout i2 block files if you wish, but as it's incomplete, focussing on the non-Checkout i2 block changes would be best when reviewing, since those parts can impact other blocks. 

Closes #4444

This is a summary of changes outside of the Checkout i2 block that should be reviewed:

- Store API - API level support for `shipping phone` in addition to `billing phone`. Both can be sent via the API, and both will be stored to meta data. The different values will not be visible in admin until core support is added.
- `blocks-registry` - This is consumed by Checkout i2 to register (experiential) blocks for insertion into checkout areas. It's the extensibility interface. This should not impact other blocks, and this has test coverage.
- icons - Additions for i2, and tweaks to existing icons to match.
- `ExternalLinkCard` - Editor component used by i2 to link to WordPress admin pages.
- Transforms - There is a new transform added to the old checkout to migration that is only present in experimental builds.
- `withScrollToTop` HOC - Converted to a functional component to avoid Typescript errors.
- `CustomerDataContext` - Moved some methods to this context so they are available outside of component state.
- `useCheckoutAddress` - Support for shipping/billing phone
- `ShippingRatesControl` - Avoid slot fill when previewing, because the slot fill is only available on the frontend.

**Testing**

1. Smoke test current checkout - check for regressions, paying attention that phone field still saves after placing an order, and shipping rates are displayed correctly.
2. Do a non-dev build and confirm Checkout i2 is not available.
3. Do a dev build and confirm Checkout i2 is available.
4. Try a transform between Checkout -> Checkout i2.
5. Trigger an error on the old checkout - ensure it scrolls to the error notice (to test withScrollToTop).


assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
```diff
@@ -172,14 +173,26 @@ const ShippingRatesControl = ( {
...
 					packages={ shippingRates }
 					noResultsMessage={ noResultsMessage }
 					renderOption={ renderOption }
 				/>
-			</ExperimentalOrderShippingPackages>
+			) : (
+				<>
+					<ExperimentalOrderShippingPackages.Slot
+						{ ...slotFillProps }
+					/>
```
nerrad, line 186: So this means this slot won't be available in the Editor context? Won't that impact previewing what extensions might add in this area or is that not a problem for this specific slot.

assets/js/base/components/checkbox-control/index.js
```diff
@@ -55,9 +55,10 @@ const CheckboxControl = ( {
 				<path d="M9 16.2L4.8 12l-1.4 1.4L9 19 21 7l-1.4-1.4L9 16.2z" />
 			</svg>
 
-			<span className="wc-block-components-checkbox__label">
-				{ label }
-			</span>
+			<span
+				className="wc-block-components-checkbox__label"
+				dangerouslySetInnerHTML={ { __html: label } }
```
nerrad: Given this is a base component, it could be a gotcha to consumers of this component that the label is non escaped client side. Why did this need to be implemented here?

Also, is the usage of `RawHTML` not an option here (assuming the rationale for non escaped content here is okay)?

assets/js/base/context/hooks/use-checkout-address.js
```diff
@@ -78,23 +57,22 @@ export const useCheckoutAddress = () => {
...
+	// This either sets the billing address to the shipping address, or restores the billing address to it's previous state.
 	useEffect( () => {
 		if ( currentShippingAsBilling.current !== shippingAsBilling ) {
 			if ( shippingAsBilling ) {
 				previousBillingData.current = billingData;
 				setBillingData( shippingAddress );
 			} else {
 				const {
-					// We need to pluck out email and phone from previous billing data because they can be empty, causing the current email and phone to get emptied. See issue #4155
+					// We need to pluck out email from previous billing data because they can be empty, causing the current email and phone to get emptied. See issue #4155
```
nerrad: Doe the mention of phone at the end of this comment block also need removed?

assets/js/base/context/hooks/use-customer-data.ts
```diff
@@ -130,10 +130,15 @@ export const useCustomerData = (): {
...
-		setCustomerData( ( prevState ) => ( {
-			...prevState,
-			shippingAddress: newData,
-		} ) );
+		setCustomerData( ( prevState ) => {
+			return {
+				...prevState,
+				shippingAddress: {
+					...prevState.shippingAddress,
+					...newData,
```
nerrad, line 138: What if a non required field is intentionally left blank on edit? Won't this prevent that value from being persisted if it previously had content?

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
```diff
@@ -55,8 +55,7 @@ export const useCheckoutContext = (): CheckoutStateContextType => {
 
 /**
  * Checkout state provider
- * This provides provides an api interface exposing checkout state for use with
- * cart or checkout blocks.
+ * This provides provides an api interface exposing checkout state for use with cart or checkout blocks.
```
nerrad:
```suggestion
 * This provides an API interface exposing checkout state for use with cart or checkout blocks.
```

assets/js/base/hocs/with-scroll-to-top/index.js
```diff
@@ -1,77 +1,83 @@
...
-/**
- * HOC that provides a function to scroll to the top of the component.
- *
- * @param {Function} OriginalComponent Component being wrapped.
- */
-const withScrollToTop = ( OriginalComponent ) => {
-	class WrappedComponent extends Component {
-		constructor() {
-			super();
+const WithScrollToTopComponent = ( { OriginalComponent, ...props } ) => {
```
nerrad: Nitpick, I think the "with" prefix can be reserved for actual HOCs and this could just be `ScrollToTopComponent`?

assets/js/base/hocs/with-scroll-to-top/index.js
```diff
@@ -1,77 +1,83 @@
...
+	const scrollPointRef = useRef( null );
 
-			this.scrollPointRef = createRef();
+	const scrollToTopIfNeeded = useCallback( () => {
```
nerrad: To avoid the memoization costs, it looks like this function (along with potentially others in this refactor) could be moved outside the component and invoked with an argument for the calculation?

assets/js/blocks/cart-checkout/checkout/index.js
```diff
@@ -13,6 +17,36 @@ import edit from './edit';
...
+							return createBlock(
+								'woocommerce/checkout-i2',
+								attributes
+							);
+						},
+					},
+				],
+			},
+	  }
+	: {};
```
nerrad, line 48: nifty, I wondered how this might be done!

assets/js/editor-components/external-link-card/index.tsx
```diff
@@ -0,0 +1,57 @@
+/**
+ * External dependencies
+ */
+import { __ } from '@wordpress/i18n';
+import { Icon, chevronRight } from '@woocommerce/icons';
+import { VisuallyHidden } from '@wordpress/components';
```
nerrad, line 6: Is there a WP version requirement for this component?

assets/js/editor-components/external-link-card/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+			rel="noreferrer"
+		>
+			<span className="wc-block-editor-components-external-link-card__content">
+				<strong className="wc-block-editor-components-external-link-card__title">
+					{ title }
+				</strong>
+				{ description && (
+					<span
+						className="wc-block-editor-components-external-link-card__description"
+						dangerouslySetInnerHTML={ { __html: description } }
```
nerrad: Can we trust this content? Why not use `RawHTML`? Any concern about this component being consumed by other things in the future that may not realize `description` is not escaped?

assets/js/editor-components/external-link-card/index.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Internal dependencies
+ */
+import './editor.scss';
+
+/**
+ * Show a link that displays a title, description, and optional icon. Links are opened in a new tab.
+ */
+const ExternalLinkCard = ( {
```
nerrad, line 16: It'd be good to do a storybook story for this new component.

bin/webpack-entries.js
```diff
@@ -45,6 +45,10 @@ const blocks = {
 	checkout: {
 		customDir: 'cart-checkout/checkout',
 	},
+	'checkout-i2': {
```
nerrad, line 48: I wonder if long term we should think of another slug for referencing this or maybe what we have is sufficient?

packages/checkout/blocks-registry/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+```js
+registerCheckoutBlock( blockName, options );
+```
+
+The options you feed the configuration instance should be an object in this shape (see `CheckoutBlockOptions` typedef):
+
+```js
+const options = {
+	component: () => <div>A Functional Component</div>,
```
nerrad:
```suggestion
	component: () => <div>A Function Component</div>,
```

packages/checkout/blocks-registry/index.ts
```diff
@@ -0,0 +1,194 @@
+/**
+ * External dependencies
+ */
+import { BlockConfiguration } from '@wordpress/blocks';
+import { registerBlockComponent } from '@woocommerce/blocks-registry';
+import { registerExperimentalBlockType } from '@woocommerce/block-settings';
+import type { LazyExoticComponent } from 'react';
```
nerrad, line 7: TIL about this type!

packages/checkout/blocks-registry/index.ts
```diff
@@ -0,0 +1,194 @@
...
+};
+
+let registeredBlocks: RegisteredBlocks = {
+	fields: [],
+	totals: [],
+	contactInformation: [ 'core/paragraph' ],
+	shippingAddress: [ 'core/paragraph' ],
+	billingAddress: [ 'core/paragraph' ],
+	shippingMethods: [ 'core/paragraph' ],
+	paymentMethods: [ 'core/paragraph' ],
```
nerrad, line 30: Why is paragraph a default here?

packages/checkout/blocks-registry/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * Asserts that an option is of the given type. Otherwise, throws an error.
+ *
+ * @throws Will throw an error if the type of the option doesn't match the expected type.
+ */
+const assertOption = (
+	options: Record< string, unknown >,
+	optionName: string,
+	expectedType: string
```
nerrad: Should we more strongly type on the acceptable values for `expectedType` (especially since `actualType` is derived from `typeof` on the option value)?

packages/checkout/blocks-registry/index.ts
```diff
@@ -0,0 +1,194 @@
...
+
+/**
+ * Get a list of blocks available within a specific area.
+ */
+export const getRegisteredBlocks = (
+	area: keyof RegisteredBlocks
+): Array< string > => {
+	assertValidArea( area );
+	return registeredBlocks[ area ];
+};
```
nerrad: It appears that this method is exposed publicly (via the checkout package). Do we have some concern that this then makes the array of registered blocks available for mutation by third parties (which could lead to unexpected results). I wonder if this is something we should protect against by either cloning or freezing the returned value?

src/StoreApi/Routes/CartUpdateCustomer.php
```diff
@@ -117,6 +117,15 @@ protected function get_route_post_response( \WP_REST_Request $request ) {
...
 			)
 		);
+
+		$shipping_phone_value = isset( $shipping['phone'] ) ? $shipping['phone'] : null;
+
+		if ( is_callable( [ wc()->customer, 'set_shipping_phone' ] ) ) {
+			wc()->customer->set_shipping_phone( $shipping_phone_value );
+		} else {
+			wc()->customer->update_meta_data( 'shipping_phone', $shipping_phone_value );
+		}
```
nerrad, line 128: Should we add a comment block here (todo) as a reminder to remove this when the core functionality is added in WC 5.6 (your PR in Woo Core got merged)?

_This comment also applies to other schema and route classes where the shipping phone value is set or retrieved_.

src/StoreApi/Schemas/ShippingAddressSchema.php
```diff
@@ -36,6 +36,12 @@ class ShippingAddressSchema extends AbstractAddressSchema {
 	 */
 	public function get_item_response( $address ) {
 		if ( ( $address instanceof \WC_Customer || $address instanceof \WC_Order ) ) {
+			if ( is_callable( [ $address, 'get_shipping_phone' ] ) ) {
+				$shipping_phone = $address->get_shipping_phone();
+			} else {
+				$shipping_phone = $address->get_meta( $address instanceof \WC_Customer ? 'shipping_phone' : '_shipping_phone', true );
+			}
```
nerrad, line 43: So one potential problem I see here (and elsewhere similar retrieval logic is found), is that should we release Checkout I2 in the feature plugin (non-experimental) before `get_shipping_phone` is available in core, and then later the store updates to a version of WooCommerce core that _does_ have `get_shipping_phone` is there risk here that the existing shipping phone stored in the `shipping_phone` or `_shipping_phone` meta fields would not be referenced leading to displaying/saving an empty value because `get_shipping_phone` doesn't return any value? 

As far as I can tell there's no migration routine for handling this sort of scenario and if we leave this in place I wonder if we should have something like this logic instead:

```suggestion
			$shipping_phone = $address->get_meta( $address instanceof \WC_Customer ? 'shipping_phone' : '_shipping_phone', true );
			if ( empty( $shipping_phone ) && is_callable( [ $address, 'get_shipping_phone' ] ) ) {
				$shipping_phone = $address->get_shipping_phone();
			}
```

It does result in two db queries but reduces the risk of losing information?

_This comment applies similarly in other places where there is logic conditionally retrieving the shipping phone value._

nerrad (commented):
Whew this was a big PR :). For the most part this looks pretty great.

I am wondering, with the [pending L2->L0 policy change](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4484) (likely merged this week), should we gate checkout and cart blocks availability to a hard WordPress 5.8 requirement? 

In my review:
- I did not review anything in the `blocks/cart-checkout/checkout-i2` path under the assumption that was already reviewed as a part of the feature plugin.
- I did not test any of this.

assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
mikejolley: It would yes, but there is a known conflict/issue @senadir has been wrestling with within the editor context which this works around. Slots are either non-functional or cause duplication of other components. This does not impact frontend.

assets/js/base/components/checkbox-control/index.js
mikejolley: Labels can contain HTML for example links. I believe this was added to support the terms and conditions block.

assets/js/base/context/hooks/use-checkout-address.js
mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/fd6e36a4fad4dc596cc6ecc983a608beb093f172

bin/webpack-entries.js
mikejolley: Would this not just change to checkout once it's no longer experimental?

packages/checkout/blocks-registry/index.ts
mikejolley: Demo purposes—it's likely the defaults will be changed, or registered, once user testing/design have reviewed this.

assets/js/base/context/hooks/use-customer-data.ts
mikejolley: Nope, and doesn't test that way. The empty field is still defined. I tested this by setting the value and checking request to server, then clearing the field and checking again. Works fine.

assets/js/base/hocs/with-scroll-to-top/index.js
mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/f56a7d58a8a2572fc7dcf177e35c85fc9b717bfa

assets/js/editor-components/external-link-card/index.tsx
mikejolley: It was added in 2019, and present in WP 5.5. I didn't look further back.

packages/checkout/blocks-registry/index.ts
mikejolley: See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/c01888bee0e4071a89d3dd5ec5b86ae7243cc23e

bin/webpack-entries.js
nerrad: Maybe, but how would block transforms work for that?

src/StoreApi/Schemas/ShippingAddressSchema.php
mikejolley: I think the way I have it round is better for two reasons:

1. We're defaulting to the core method when it's available. _It looks like this will be available in WC 5.6_ 
2. The fallback logic I used is using the correct meta key which will match the core implementation. That's why one is prefixed and one is not.

Customer data store in core: https://github.com/woocommerce/woocommerce/blob/trunk/includes/data-stores/class-wc-customer-data-store.php#L305

Order data store: https://github.com/woocommerce/woocommerce/blob/trunk/includes/data-stores/class-wc-order-data-store-cpt.php#L137

Because we use the same meta key, there will be no data loss.

src/StoreApi/Routes/CartUpdateCustomer.php
mikejolley: Added in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/9935cd1d1a4c78754cf49ab0c87673e27fc34619

packages/checkout/blocks-registry/index.ts
mikejolley: I don't understand this point. This is returning a value from `registeredBlocks` which is not exported. The only way to edit this is to go through `registerCheckoutBlock` which is exported as an extensibility API. Can you explain further?

src/StoreApi/Schemas/ShippingAddressSchema.php
nerrad: Ahh, that makes sense, I missed the piece about the two implementations sharing the same meta key 👍🏻 (Very smart btw!)

assets/js/editor-components/external-link-card/index.tsx
mikejolley: I think we could from gateways but I've removed this https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/90871773c2cb6fab10993bf55c8cfd7968b77caf - we can revisit or strip tags if we come across a use case in testing.

assets/js/editor-components/external-link-card/index.tsx
mikejolley: Done in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/65cec3af57644c652eaa5131e232f621ed38be98

assets/js/base/hocs/with-scroll-to-top/index.js
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/e62c42b6858e26e9e7685c4666ca795e311e79f8

assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
mikejolley: This is going to come up again with #4447 so let's tackle it then.

assets/js/base/components/checkbox-control/index.js
mikejolley: Updated in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/a3c4fb01f88f308b263f5e99df0cb76c4abc84c9

bin/webpack-entries.js
mikejolley: Block Transforms wouldn't be used. If the Blocks end up with totally different HTML when saved, we would instead use the [Deprecation API](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/) to convert instances of Checkout i1 to instances of Checkout i2.

assets/js/base/context/hooks/use-customer-data.ts
nerrad: Still seems suspect but given you've tested we can roll with it.

assets/js/base/hocs/with-scroll-to-top/index.js
nerrad: Nice!

bin/webpack-entries.js
nerrad: Ahh right, perfect. So we'd likely just do that when we're ready to remove the experimental flag right?

packages/checkout/blocks-registry/index.ts
nerrad: Yup. `packages/checkout/index.js` [exports everything from `packages/checkout/blocks-registry/index.ts`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/files#diff-f9bb35272358eefbb66edc83a6f637c19796d51e97ff3bf96f6112cea2f6ef2bR6). Thus `getRegisteredBlocks` is exposed on whatever global is attached to the `packages/checkout` build. `getRegisteredBlocks` will return a mutable array for the provided area thus exposing the internals of the registered blocks that can be mutated.

So if I'm understanding correctly, consumer code could do something like this (mutating any registered blocks):

```js
const registeredBlockFields = wc.blocksCheckout.getRegisteredBlocks( 'fields' );
registeredBlockFields[] = 'some/other-block';

// or
delete( registeredBlockFields[0] );
```

nerrad (approved):
Thanks for handling the feedback Mike (especially around the concerns for dangerouslySetInnerHTML usage)! I still have a question around the potential (that I see) for the registered blocks array in a checkout area to be mutated. I'm okay with it being handled in a followup if you agree something should be done.

Also, there's an open question I asked in my previous review that I don't see an answer for:

> I am wondering, with the pending L2->L0 policy change (likely merged this week), should we gate checkout and cart blocks availability to a hard WordPress 5.8 requirement?

The above could be done in a followup too.

With that in mind I'm pre-approving on the assumption my concerns will either be addressed in followups or in this PR before merge.

assets/js/base/components/checkbox-control/index.js
senadir: Yes, this was the only way to support labels with links. Gutenberg will give you back raw HTML so we can't just pass children.

packages/checkout/blocks-registry/index.ts
mikejolley: Ok. I couldn't repro but I spreaded to be safe in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/commits/2bfe6deda4ef3f94972466290b3d7484e13c3182

mikejolley:
> I am wondering, with the pending L2->L0 policy change (likely merged this week), should we gate checkout and cart blocks availability to a hard WordPress 5.8 requirement?

I think we'll need to discuss this and work in a follow-up. My concern is that this check would be shifting as new versions are released, and it might make more sense to do detection higher up rather than on a per-block basis. For example, having a version check in the isExperimentalBuild function so that experimental functionality requires L0.

mikejolley:
Merging 🥳

nerrad:
> I think we'll need to discuss this and work in a follow-up. 

Yup I'm fine with a followup!

> My concern is that this check would be shifting as new versions are released, and it might make more sense to do detection higher up rather than on a per-block basis. For example, having a version check in the isExperimentalBuild function so that experimental functionality requires L0.

Right that makes sense while something is still behind the experimental flag. As soon as it is surfaced in the plugin though I think we'll want to gate C&C blocks to the WP version supported at the time of release. Mostly because I'm expecting we will probably be surfacing to _new WooCommerce users_ in Woo Core while Woo Core still has L2 support outside the range of what the blocks will require. But that's also something we could also plan for when we come to it I suppose.

My _primary_ concern is that we whenever this is surfaced to feature plugin users we have some way of indicating what is the minimum version this will require to help inform the approach we take when we plan for surfacing in WooCommerce core.

PR #4279, opened 2021-05-26
Fix display of itemised taxes

opr:
In the current (new design) of the Cart and Checkout sidebars, when the merchant had the `Show tax totals` setting set to `Itemised` the tax lines did not have a price next to them. This was correct (as per the designs) but it was a regression in terms of UX.

Following this slack thread: p1622027500383600-slack-C8X6Q7XQU I:
- When `Itemised` taxes were enabled, remove the `Taxes` item, and in its place output the individual rates with the associated prices
- Updated the CSS to reduce the vertical space between tax lines, and ensure the first and last individual lines do not have margins above, and below them.

### Screenshots
| Before | After |
|---|---|
| image | image |

1. Go `WooCommerce -> Settings -> Tax` and set `Display prices during basket and checkout` to `Excluding Tax`.
2. Set `Display tax totals` to `Itemised`
3. Set up a few products to use different tax rates.
4. Add them to the cart and verify you see each tax rate shown in the sidebar of both the Cart and Checkout blocks and that the totals are correct.
5. Set `Display tax totals` back to `As a single total` and check that the individual rates are no longer shown, but a `Taxes` total is shown in their place.
6. Check the blocks in the editor context too and ensure they display correctly.

### Changelog

### DIFF
--- a/packages/checkout/totals/taxes/index.tsx
+++ b/packages/checkout/totals/taxes/index.tsx
@@ -40,29 +40,38 @@ const TotalsTaxes = ( {
 		return null;
 	}
 
-	const itemisedTaxItems: ReactElement | null = getSetting(
+	const showItemisedTaxes = getSetting(
 		'displayItemizedTaxes',
 		false
-	) ? (
-		<>
-			{ taxLines.map( ( { name, rate }, i ) => {
+	) as boolean;
+
+	const itemisedTaxItems: ReactElement | null = showItemisedTaxes ? (
+		<div
+			className={ classnames(
+				'wc-block-components-totals-taxes',
+				className
+			) }
+		>
+			{ taxLines.map( ( { name, rate, price }, i ) => {
 				const label = `${ name }${
 					showRateAfterTaxName ? ` ${ rate }` : ''
 				}`;
 				return (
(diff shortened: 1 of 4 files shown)


packages/checkout/totals/taxes/index.tsx
```diff
@@ -40,29 +40,38 @@ const TotalsTaxes = ( {
...
+			) }
+		>
+			{ taxLines.map( ( { name, rate, price }, i ) => {
 				const label = `${ name }${
 					showRateAfterTaxName ? ` ${ rate }` : ''
 				}`;
 				return (
 					<TotalsItem
 						key={ `tax-line-${ i }` }
-						className="wc-block-components-totals-taxes__tax-line"
```
senadir, line 55: Any reason we changed this classname? was it introduced recently in your PR or was it there since before?

senadir (approved):
Tests well as per the screenshots

packages/checkout/totals/taxes/index.tsx
opr: It's new as of the previous PR, but I changed it to be more descriptive of what the element actually represents.

PR #4282, opened 2021-05-26
Add documentation for the `couponName` filter

opr:
Because this this was somehow omitted from #4166, I am re-adding it in this PR. I know we said that we should avoid using TS syntax in our example documentation, so we can go back and revisit these docs during a cooldown. For now this PR is to simply include a file that was already written, for the purposes of documenting a new filter.

1. Please refer to #4166 and verify that the documentation included in this PR makes sense.

### Changelog


Aljullu (approved):
LGTM

PR #4284, opened 2021-05-27
Check if there are tax lines before rendering the container for them

opr:
When taxes are set to be shown as `Itemised` there is a div that contains the individual lines, this div was previously rendered even if there are no tax lines to be shown. This PR will check if there are lines to output, and only render the div with borders if there are.

Introduced in #4279 😞 

### Screenshots

| Before | After |
|---|---|
| image | image |

1. Go WooCommerce -> Settings -> Tax and set Display prices during basket and checkout to Excluding Tax.
2. Set Display tax totals to Itemised
2. Add an item to the cart with a zero-rated tax rate.
3. Check there are no borders without anything in them, like the before screenshot above.

### Changelog

> Fix a display issue when itemised taxes are enabled, but no products in the cart are taxable.

### DIFF
--- a/packages/checkout/totals/taxes/index.tsx
+++ b/packages/checkout/totals/taxes/index.tsx
@@ -45,29 +45,30 @@ const TotalsTaxes = ( {
 		false
 	) as boolean;
 
-	const itemisedTaxItems: ReactElement | null = showItemisedTaxes ? (
-		<div
-			className={ classnames(
-				'wc-block-components-totals-taxes',
-				className
-			) }
-		>
-			{ taxLines.map( ( { name, rate, price }, i ) => {
-				const label = `${ name }${
-					showRateAfterTaxName ? ` ${ rate }` : ''
-				}`;
-				return (
-					<TotalsItem
-						key={ `tax-line-${ i }` }
-						className="wc-block-components-totals-taxes__grouped-rate"
-						currency={ currency }
-						label={ label }
-						value={ parseInt( price, 10 ) }
-					/>
-				);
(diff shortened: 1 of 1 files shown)


packages/checkout/totals/taxes/index.tsx
```diff
@@ -45,29 +45,30 @@ const TotalsTaxes = ( {
...
-						currency={ currency }
-						label={ label }
-						value={ parseInt( price, 10 ) }
-					/>
-				);
-			} ) }{ ' ' }
-		</div>
-	) : null;
+	const itemisedTaxItems: ReactElement | null =
+		showItemisedTaxes && taxLines.length > 0 ? (
```
opr, line 49: This is the only change here, the rest is formatting.

frontdevde (approved):
Tested as per testing instructions and it works as expected! 👍

When there's only a zero tax rated product in the Cart, in `trunk` it renders the empty container, on this branch it doesn't. I also tested with a taxed product and there the tax still renders as expected.

|Before|After|
|-|-|
|Screenshot 2021-05-31 at 15 38 12|Screenshot 2021-05-31 at 15 39 39|

senadir:
When the cart is not eligible for shipping as well, I'm getting a double border
image
This doesn't happen if coupons are also disabled
image
This is probably unrelated to this PR but just wanted to highlight while testing this.

opr:
Thanks @senadir and @frontdevde - the defect you found is noted and tracked here https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4267

PR #4291, opened 2021-05-31
Update Cart Widget when adding an item in All Products

### LINKED ISSUE #4287: Cart Widget data not updated when adding item to the cart in All Products block
This appears to be a regression introduced recently. We had some logic that was supposed to take care of triggering the refreshing of cart-fragments in the Cart Widget data whenever an item is added to the cart via the All Products block. However, I noticed recently while testing something else, that this seems to have stopped working:

CleanShot 2021-05-28 at 08 06 05

From a bit of digging, it looks like this was introduced with adding REST API batch support in fdc74b87 - my guess is we just forgot to add handling for the cart button update with the response change update.

## To reproduce

1. Go to a page with the All Products block in the content.
2. Click on the Add to Cart button for a product.
3. View the Add to Cart widget and the total should be updated to reflect the change.

senadir:
In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4075 we refactored our data store, a new control file was created inside the `cart` store that had only one controller, which managed triggering the update fragment, we didn't import that control and so it wasn't registered, this PR fixes it.

Fixes #4287

1. In storefront or any theme that support Cart widgets, go to a page with All Products
2. Add a product to cart from there.
3. The widget should be updated.

### Changelog

> Fix a bug in which Cart Widget didn't update when adding items from All Products block

### DIFF
--- a/assets/js/data/cart/index.ts
+++ b/assets/js/data/cart/index.ts
@@ -12,13 +12,14 @@ import * as selectors from './selectors';
 import * as actions from './actions';
 import * as resolvers from './resolvers';
 import reducer, { State } from './reducers';
-import { controls } from '../shared-controls';
+import { controls as sharedControls } from '../shared-controls';
+import { controls } from './controls';
 
 registerStore< State >( STORE_KEY, {
 	reducer,
 	actions,
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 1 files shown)


opr (approved):
👏🏼 Works well, code looks fine too.

PR #4293, opened 2021-05-31
Fix the flickering of the Proceed to Checkout button on quantity update in the Cart Block

### LINKED ISSUE #4289: "Proceed to Checkout" button flickers when changing quantity or removing item
On Cart, The "Proceed to Checkout" button seems to flicker instead between disabled and enabled when changing quantity or removing an item, it works fine when changing the shipping method

## Expected behavior
- The button should be disabled from when you click the quantity or remove button, up until the server response and value is persisted.

## Current Behavior
- The button flickers when you make the action immediately switches from disabled to enabled again.

## Screenshot

https://user-images.githubusercontent.com/6165348/120106339-88b66300-c154-11eb-81a9-b6d32b85506f.mov

senadir:
In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3314 we introduced `previousIsPendingQuantity` and `previousIsPendingDelete` to fix an issue with checkout button flickering on quantity changes.

After refactoring it over time, the issue surfaced again, deleting the code added in that PR fixed the issue.

Currently, we set isCalculating to true if we have quantity pending in our store, or if we're deleting an item. The other new part of this PR is that we also set isCalculating in the debouncing period, this is to fix a delay between clicking update and the button is greyed out.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4289

1. Smoke test Cart block, make sure changing quantity works fine.
2. Make sure removing items works fine.
3. When changing an item quantity, make sure the button is disabled immediately as you change the quantity and is re-enabled once the server request finishes

### Changelog

> Fix the flickering of the Proceed to Checkout button on quantity update in the Cart Block

### DIFF
--- a/assets/js/base/context/hooks/cart/use-store-cart-item-quantity.ts
+++ b/assets/js/base/context/hooks/cart/use-store-cart-item-quantity.ts
@@ -83,7 +83,6 @@ export const useStoreCartItemQuantity = (
 		},
 		[ cartItemKey ]
 	);
-	const previousIsPending = usePrevious( isPending );
 
 	const removeItem = () => {
 		return cartItemKey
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Works fine for me and having split the `useEffect()` in two is great. :ship:

PR #4294, opened 2021-05-31
Remove the mustBeString validation utility and its usage

senadir:
This is a refactoring issue, [we recently introduced same-type-validation in our Checkout filters](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/11a9b23d737a6fefeff0569debc22749d0ff5cc2/packages/checkout/registry/index.ts#L86-L98), making `mustBeString` useless here because it's no longer being used. This PR removes `mustBeString` and its usage, it also makes `validation` an optional value.

### Testing steps
- Install and enable WooCommerce Subscriptions, add subscriptions products to your cart.
- Smoke test Cart and Checkout, make sure there are no console errors.

### DIFF
--- a/packages/checkout/registry/index.ts
+++ b/packages/checkout/registry/index.ts
@@ -74,7 +74,7 @@ export const __experimentalApplyCheckoutFilter = < T >( {
 	/** Object containing arguments for the filter function. */
 	arg: CheckoutFilterArguments;
 	/** Function that needs to return true when the filtered value is passed in order for the filter to be applied. */
-	validation: ( value: unknown ) => true | Error;
+	validation?: ( value: T ) => true | Error;
 } ): T => {
 	return useMemo( () => {
 		const filters = getCheckoutFilters( filterName );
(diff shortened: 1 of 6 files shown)


packages/checkout/registry/index.ts
```diff
@@ -74,7 +74,7 @@ export const __experimentalApplyCheckoutFilter = < T >( {
 	/** Object containing arguments for the filter function. */
 	arg: CheckoutFilterArguments;
 	/** Function that needs to return true when the filtered value is passed in order for the filter to be applied. */
-	validation: ( value: unknown ) => true | Error;
+	validation?: ( value: T ) => true | Error;
```
senadir, line 77: `value` type isn't uknown, we know for sure that it's the same as defaultValue and return value `T`

Aljullu (approved):
Looks good to me, nice cleanup!

> we recently introduced same-type-validation in our Checkout filters

For reference, I think you are referring to this check, right?

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/11a9b23d737a6fefeff0569debc22749d0ff5cc2/packages/checkout/registry/index.ts#L86

senadir:
Yes, I linked to it in the post.

PR #4298, opened 2021-06-01
Add cart/extensions endpoint to allow third party extensions to process server-side and get a new cart

### LINKED ISSUE #4254: Create a function that allows extensions to make an API request, and signal that the cart should also be reloaded
When extensions update the cart server-side, we need the fresh cart data to be fetched client-side. The following solutions have already been explored:
1. Let extensions set cart data themselves
2. Give extensions a way to tell Blocks that the cart needs to be re-fetched

Neither of these solutions are suitable for the following reasons, respectively:
1. If an extension adds cart data that is the incorrect shape, the entire block will break
2. If this happens, there will be the initial delay where the cart is updated by the extension, then a second delay when Blocks fetches new API data.

To solve this I am proposing a new method in the `blocks-checkout` package, named something like `restRequestWithCartUpdate`. The API would be the same as `@wordpress/api-fetch` but the request made by the extension would be batched with a request to fetch new cart data, so only one request needs to be made to the server.

opr:
This PR is based on `feature/coupon-area-slot-fill` as it forms part of the work I'm doing for WooCommerce Points and Rewards.

--- 

This PR will add the ability for extensions to register functions/callbacks to be executed by Blocks when the `cart/extensions` endpoint is hit.

This PR is necessary to allow third-party extensions to tell Blocks they need to do some server-side processing which will update the cart.

The way it should work is an extension will register its callbacks with Blocks, via the `register_update_callback` method. When the `wc/store/cart/extensions` endpoint is requested, there should be at the minimum, a `namespace` and `endpoint` parameter passed, as well as any additional data. If any callbacks are registered, matching the `namespace` and `endpoint` values, they are executed, and then the cart is returned by the API.

 ---

The changes made in this PR are:
- Delete `updateCartFromAPI` JS function and replace it with `extensionCartUpdate` - this is a JS function intended for use by extensions to signal to Blocks that it should POST to the `/cart/extensions` endpoint, then pass the result to `receiveCart`. This function returns a promise so extensions can know when the request is finished processing.
- The args passed to this function should be: 
```typescript
{
  namepsace: string,
  data: object,
}
```
- Add the `cart/extensions` route. Extensions may signal to blocks that they need something to be done server-side which will result in a modified cart, and Blocks will make the request on their behalf.
- Add two methods to `src/Domain/Services/ExtendRestApi.php`: `register_update_callback` and `get_update_callback`. Extensions will use `register_update_callback` to tell Blocks which `callable`s to execute when the `cart/extensions` endpoint is hit. The `get_update_callback` will be used internally only, to determine which callbacks ought to be executed, based on the passed namespace and endpoint.
- Add `CartExtensionsSchema` - This is added to allow us to control the flow of the programme and to execute the callbacks before returning the cart.
- Add PHPUnit tests for these new endpoints and test their error handling.

Fixes #4254

1. In an extension that extends the REST API, you could add
```php
self::$extend->register_update_callback(
  array(
    'namespace'         => self::IDENTIFIER,
    'callback_function' => array( 'WC_Points_Rewards_Cart_Checkout', 'rest_apply_discount' ),
  )
);
```
Changing the `callback_function` to a callable of your choice. Subscriptions might be a good one to test with.
2. On the front-end use the `extensionCartUpdate` function to call the endpoint, and ensure the callable you registered with `register_update_callback` is executed.

### Changelog

> Add the ability for extensions to register callbacks to be executed by Blocks when the cart/extensions endpoint is hit. Extensions can now tell Blocks they need to do some server-side processing which will update the cart.


nerrad:
The direction is looking good here. However, have you considered the following:

- What happens when potentially multiple extensions invoke `extensionCartUpdate` client side and there are several running requests? As a possible solution, we have batch request handling in our `wp.data` store, so I wonder if we could create a special action generator creator that could be exposed for extensions to _dispatch_ a cart update? It'd be pretty similar to what the logic you have right now but fulfilled as a store action instead? That would get the advantage of potentially debouncing multiple requests (depending on what extensions are doing).
- Something else we may need to consider is what happens if the component triggering the request is unmounted during the request. Might be edgecase, but there's potential for memory leaks here.

opr:
Thanks Darren, indeed batching is something I was going to look at doing, I will work on that next.

> Something else we may need to consider is what happens if the component triggering the request is unmounted during the request. Might be edgecase, but there's potential for memory leaks here.

How do you see this memory leak happening? If the component responsible for calling `extensionCartUpdate` unmounts, then all that'll happen is an updated cart will be fetched from the API and the store updated, this method doesn't interact with the extension, or calling component at all. Unless you're thinking about the promise resolving/rejecting?

nerrad:
The memory leak could happen if something calls `extensionCartUpdate` in a `useEffect` hook which kicks off the request and then the dispatch gets invoked on the response update when the component is unmounted. As I'm typing this, I'm realizing this likely won't be a problem because the dispatch will eventually result in `wp.data` cart store being update which is state that lives outside of the component triggering the request.

packages/checkout/utils/extension-cart-update.ts
```diff
@@ -0,0 +1,23 @@
...
+ * When executed, this will invalidate the getCartData selector, causing a request to be made
+ * to the API. This is in place to allow extensions to signal that they have modified the cart,
+ * and that it needs to be reloaded in the client.
+ */
+export const extensionCartUpdate = (
+	args: ExtensionCartUpdateArgs
+): Promise< CartResponse > => {
+	const { applyExtensionCartUpdate } = dispatch( STORE_KEY );
+	return applyExtensionCartUpdate( args );
+};
```
senadir, line 24: I think the docs are wrong here, this passes the data to create the request, not just the invalidation.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -52,7 +53,7 @@ public function get_formatter( $name ) {
 	 *
 	 * @var array
 	 */
-	private $endpoints = [ CartItemSchema::IDENTIFIER, CartSchema::IDENTIFIER ];
+	private $endpoints = [ CartItemSchema::IDENTIFIER, CartSchema::IDENTIFIER, CartExtensionsSchema::IDENTIFIER ];
```
senadir: `CartItemSchema::IDENTIFIER` and `CartSchema::IDENTIFIER` are endpoints you can add data to, `CartExtensionsSchema::IDENTIFIER` is the endpoint that will get the data, it doesn't make sense for it to be here.

src/Domain/Services/ExtendRestApi.php
```diff
@@ -118,6 +126,66 @@ public function register_endpoint_data( $args ) {
...
+	public function register_update_callback( $args ) {
+		if ( ! is_string( $args['namespace'] ) ) {
+			$this->throw_exception( 'You must provide a plugin namespace when extending a Store REST endpoint.' );
+		}
+
+		if ( ! is_string( $args['endpoint'] ) || ! in_array( $args['endpoint'], $this->endpoints, true ) ) {
+			$this->throw_exception(
+				sprintf( 'You must provide a valid Store REST endpoint to extend, valid endpoints are: %1$s. You provided %2$s.', implode( ', ', $this->endpoints ), $args['endpoint'] )
+			);
+		}
```
senadir: I don't get the point of this option, I'm going to assume the only way people can register update callbacks is going to be in `cart/extensions`, so why add it to the options if it's going to be static?

src/Domain/Services/ExtendRestApi.php
```diff
@@ -118,6 +126,66 @@ public function register_endpoint_data( $args ) {
...
+	public function get_update_callback( $endpoint, $namespace ) {
+		if ( ! is_string( $namespace ) ) {
+			$this->throw_exception( 'You must provide a plugin namespace when extending a Store REST endpoint.' );
+		}
+
+		if ( ! is_string( $endpoint ) || ! in_array( $endpoint, $this->endpoints, true ) ) {
+			$this->throw_exception(
+				sprintf( 'You must provide a valid Store REST endpoint to extend, valid endpoints are: %1$s. You provided %2$s.', implode( ', ', $this->endpoints ), $endpoint )
+			);
+		}
```
senadir: I get that you copied this from `get_extend_data` but like I said above, this option doesn't make sense.

src/StoreApi/SchemaController.php
```diff
@@ -95,6 +96,16 @@ protected function initialize() {
...
+		$this->schemas[ CartExtensionsSchema::IDENTIFIER ]   = new CartExtensionsSchema(
+			$this->extend,
+			$this->schemas[ CartItemSchema::IDENTIFIER ],
+			$this->schemas[ CartCouponSchema::IDENTIFIER ],
+			$this->schemas[ CartFeeSchema::IDENTIFIER ],
+			$this->schemas[ CartShippingRateSchema::IDENTIFIER ],
+			$this->schemas[ ShippingAddressSchema::IDENTIFIER ],
+			$this->schemas[ BillingAddressSchema::IDENTIFIER ],
+			$this->schemas[ ErrorSchema::IDENTIFIER ]
+		);
```
senadir: Why initiate it with all of those classes? the ones in Cart should be enough. Unless you're explicitly using it in the class, I don't think you should include it.

senadir (approved):
I like the direction of this PR Thomas, I have some comments but they're cosmetic/simple ones and shouldn't require a lot of change.

I'd like us to test faulty requests, bad namespace, wrong types, args that contain errors, and see what breaks and at what level, this would help inform us if we want to do more try/catch

src/Domain/Services/ExtendRestApi.php
```diff
@@ -118,6 +126,66 @@ public function register_endpoint_data( $args ) {
...
+
+		$method = $this->callback_methods[ $endpoint ][ $namespace ]['callback_function'];
+		if ( ! is_callable( $method ) ) {
+			$this->throw_exception(
+				sprintf( 'There is no valid callback registered for: %1$s on the %2$s endpoint.', $namespace, $endpoint )
+			);
+		}
+		return $method;
+	}
```
senadir: Who executes `$method` here? I think it should be executed at `ExtendRestApi` level and for it to wrapped in a try/catch. We should return an Error Response to users. a Try Catch would just make sure the registration doesn't fail breaking the whole API.

src/StoreApi/SchemaController.php
opr: It's because `CartExtensionsSchema` inherits `CartSchema` - I actually reviewed this and found a better way of doing it. Committing now.

src/Domain/Services/ExtendRestApi.php
opr: Correct, I've removed it, I don't foresee any other routes they could need to update the cart on for now.

src/Domain/Services/ExtendRestApi.php
opr: Agreed, removed.

packages/checkout/utils/extension-cart-update.ts
opr: Thanks, you're right. This is from an old iteration of this function. I have updated the comment now.

src/Domain/Services/ExtendRestApi.php
opr: `$method` is executed by `get_item_response` in `CartExtensionSchema` I have added a try catch there in 468bae3d20a46ebf9334f4f62d36809e0cc64941. This would cause the promise returned by `extensionCartUpdate` to reject. Extensions can then catch this and handle it.

image

Is this what you were thinking of or did you have something else in mind?

opr:
> I like the direction of this PR Thomas, I have some comments but they're cosmetic/simple ones and shouldn't require a lot of change.
> 
> I'd like us to test faulty requests, bad namespace, wrong types, args that contain errors, and see what breaks and at what level, this would help inform us if we want to do more try/catch

Thanks for your review Nadir, I appreciate your comments. I agree with all the suggestions so I have made them, I also refactored a little bit around what `CartExtensionSchema` inherits, so the constructor for it is now simpler.

I am also including some changes here to `ValidationInputError` and `ValidatedTextInput` as they are minor.

There is now a try/catch in the `CartExtensionSchema` which handles Exceptions thrown by callbacks.

If an extension calls `extensionCartUpdate` and doesn't pass the correct namespace, the API returns an error response. If the args/types are wrong, well then it would be up to the extension to handle this, and throw an Exception. I will be writing documentation surrounding this later. When you say faulty requests, what do you mean? The requests are made here, in Blocks, but anyway if they fail, the Promise will reject and extensions can catch it.

Feel free to try this out yourself by using the `feature/add-inputs-for-points-redemption` branch in Points and Rewards.

senadir:
> When you say faulty requests, what do you mean? The requests are made here, in Blocks, but anyway if they fail, the Promise will reject and extensions can catch it.

Faulty requests are requests with the wrong shape of data or containing fatal code.
I think the handling here is correct since if a request returns an error, it's up to the plugin to catch it, and if they don't, our SlotFill boundaries should catch it. The goal is that errors shouldn't surface all the way to the block causing it to break.

opr:
@senadir I have now added some unit tests, and renamed `callback_function` to `callback` it seems better.

Please would you give this another look over and see if you agree it can be merged? Thanks 🙌🏼

senadir (approved):
Thank you for addressing feedback and 💯 on the new tests!

PR #4303, opened 2021-06-02
Hide Cart and Checkout blocks from the widget editor

### LINKED ISSUE #4190: Block Widgets: Cart showing in sidebar when in Checkout
_This issue surfaced as part of our audit of the current state of using WooCommerce Blocks as Block Widgets in the new Widgets Editor and Customizer (pca54o-1pT-p2)._

Given that the sidebar widgets show everywhere, the Cart will still be present in the sidebar after having clicked Proceed to checkout on the Cart block. When using the Core Checkout, one can then still interact with the cart but the values aren’t updated in the Checkout. This experience could be confusing from a user’s perspective.

screenshot-2021-05-08-at-12 21 15

### LINKED ISSUE #4191: Block Widgets: Order completion summary duplicated in the sidebar
_This issue surfaced as part of our audit of the current state of using WooCommerce Blocks as Block Widgets in the new Widgets Editor and Customizer (pca54o-1pT-p2)._

Now that the shortcode in the widget area is displaying correctly, the user will see a duplicated order summary when completing an order with the Checkout block.

markup-on-2021-05-08-at-123130

Aljullu:
Closes #4190.
Closes #4191.

After some more discussion (pca54o-1pT-p2#comment-1929) we decided to hide the Cart block and Checkout blocks from the widget editor. In the future we will have a 'Minicart block' (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4138) which will serve the use-case of showing the cart contents in a widget area. Regarding the Checkout block, we don't see any use-case for it in a widget area.

1. Using a classic theme (ie: Storefront) with Gutenberg enabled, go to Appearance > Widgets.
2. Verify you can't add the Cart block or the Checkout block. (You will see a widget called Cart, that's the widget from WC core and is not related to this PR, so ignore it).
3. Now go to Appearance > Customize > Widgets and open a widget area.
4. Verify you can't add the Cart block or the Checkout block either.

### Changelog

> Hide the Cart and Checkout blocks from the new block-based widget editor.

### DIFF
--- a/src/BlockTypesController.php
+++ b/src/BlockTypesController.php
@@ -139,6 +139,8 @@ final class BlockTypesController {
 					'PriceFilter',
 					'AttributeFilter',
 					'ActiveFilters',
+					'Cart',
+					'Checkout',
 				]
 			);
 		}


ralucaStan (approved):
Works as expected 👏

jeffpaul:
@Aljullu @ralucaStan we got a report related to this in WordPress core, [see here](https://core.trac.wordpress.org/ticket/53433).  Could you confirm if what Olga Gleckler reported in WP Core is in fact intended as a result from this PR or not?

Aljullu:
Thanks for the ping @jeffpaul, I answered directly in the ticket. The issues were not related to this PR but instead to this two: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3626 and https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4362. The first one is fixed and the second one currently under our radar.

PR #4305, opened 2021-06-02
Fix broken phpunit tests

nerrad:
Recently PHPUnit tests started failing in the repository and I narrowed it down to the fact that for the test environment setup:

- The latest version of WooCommerce _released_ (from WP.org) is installed.
- The WooCommerce test utilities are retrieved from the `trunk` branch of WooCommerce core.

The above resulted in the test utilities referencing a class that is not available in the current released version of Woo Core (5.3) which resulted in the reported fatals.

For now, to get tests running, this PR has reconfigured the test setup so it synchronizes the test utilities used with the version of WooCommerce core being tested. The synchronization is accomplished via:

- updating `install-wp-tests.sh` script to use the `WOO_VERSION` environment variable to determine what to download for Woo.
- update docker configuration to pass through environment variable correctly.
- add environment variable to `.env` file for local Woo test runs.
- add environment variable to GitHub workflow for phpunit test runs.

## Tradeoffs.

While this does fix the immediate problem, there are some tradeoffs with this approach:

- the test utilities are currently being pulled in via a checked out tag in the WooCommerce repo. This requires cloning the entire WooCommerce repository increasing the overall impact to test run time by 1-2 min. There might be another way of doing this but in the interest of time I didn't dig too deep here
- the docker configuration we're currently using seems to have a lot of unused parts and could likely be cleaned up a bit. What I added here complicates things a bit more.
- Woo Core version is now hardcoded instead of always pulling the "latest". That means that we'll need to manually update the Woo Core version we test against whenever there's a new Woo Core release.

## Benefits

While it's listed as a tradeoff, we do have a benefit now of being able to explicitly specify Woo Core version to run phpunit tests against. This could be useful for increasing our test coverage to run against all Woo Core versions we support. However, I'm not sure how needed this is given the L2 policy keeps the version support range fairly small anyways.

## Next Steps

Some potential followups after merging this:

- Investigate whether we can automatically infer Woo Core version from the latest release on the downloaded WP.org package and use that to know what tag to checkout from the Woo Core repo. This could eliminate the manual version bumps we'd have to do to maintain this suite.
- Investigate alternative ways to obtain the correct Woo Core test utilities for the version of Woo Core being tested against other than cloning the entire repository. This will become especially problematic when WC Admin (and potentially other products) are merged directly into Woo Core and the size of the clone increases.
- Switch our test suite to use `wp-env` for phpunit tests (see the [pr I started here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3691)). With that change we'll be able to more easily switch PHP versions being tested with and even start testing against PHP8. Right now we're only testing against one version of PHP. The actual test reporting will be much better too (there currently is a lot of noise in the action logs from docker setup which makes it more time consuming to scroll to the actual phpunit test results).
- Potentially introduce testing against more than one WooCommerce version (eg. latest and earliest matching our requirements) as a part of a test matrix. Although this is likely not as important as PHP version testing.

### DIFF
--- a/.env
+++ b/.env
@@ -13,3 +13,4 @@ WORDPRESS_LOGIN=admin
 WORDPRESS_PASSWORD=password
 WORDPRESS_EMAIL=admin@woocommercecoree2etestsuite.com
 WP_VERSION=5.4
+WOO_VERSION=5.3.0
--- a/tests/bin/Dockerfile
+++ b/tests/bin/Dockerfile
@@ -1,5 +1,7 @@
 FROM wordpress:latest
 
+ARG WOO_INSTALL_VERSION=latest
+
 CMD echo "Installing dependencies..."
 
 RUN apt-get update && \
@@ -9,6 +11,7 @@ CMD echo "Installing tests..."
 
 ENV WP_TESTS_DIR=/tmp/wordpress-tests-lib
 ENV WP_CORE_DIR=/tmp/src/wordpress
+ENV WOO_VERSION=$WOO_INSTALL_VERSION
 
 COPY install-wp-tests.sh /usr/local/bin/dockerInit
 RUN chmod +x /usr/local/bin/dockerInit
--- a/tests/bin/install-wp-tests.sh
+++ b/tests/bin/install-wp-tests.sh
@@ -11,6 +11,7 @@ DB_PASS=$3
 DB_HOST=${4-localhost}
 WP_VERSION=${5-latest}
 SKIP_DB_CREATE=${6-false}
+WOO_VERSION=${WOO_VERSION-latest}
 
 # directories
 TMPDIR=${TMPDIR-/tmp}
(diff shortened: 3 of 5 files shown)


tests/bin/Dockerfile
```diff
@@ -9,6 +11,7 @@ CMD echo "Installing tests..."
 
 ENV WP_TESTS_DIR=/tmp/wordpress-tests-lib
 ENV WP_CORE_DIR=/tmp/src/wordpress
+ENV WOO_VERSION=$WOO_INSTALL_VERSION
```
nerrad, line 14: This has to be set here because `install-wp-tests` is executed as a RUN command in the docker file and in that uses an intermediate container that will not use environment variables set in the `docker-compose.yml` file.

tests/bin/install-wp-tests.sh
```diff
@@ -149,21 +150,25 @@ install_db() {
...
 
-	# Script Variables
-	BRANCH=$TRAVIS_BRANCH
-	REPO=$TRAVIS_REPO_SLUG
-
-	# Get github version
-	if [ "$TRAVIS_PULL_REQUEST_BRANCH" != "" ]; then
-		BRANCH=$TRAVIS_PULL_REQUEST_BRANCH
-		REPO=$TRAVIS_PULL_REQUEST_SLUG
+	git clone "https://github.com/woocommerce/woocommerce.git" "$TMPDIR/woocommerce-git"
```
nerrad, line 163: Unfortunately the entire repo needs cloned to support checking out specific tags. This impacts the run time for the tests by ~1-2min (depending on network connection) and also the bytes transferred. Ideally we'll find a different solution for this but for now this helps unblock failing tests.

.env
```diff
@@ -13,3 +13,4 @@ WORDPRESS_LOGIN=admin
 WORDPRESS_PASSWORD=password
 WORDPRESS_EMAIL=admin@woocommercecoree2etestsuite.com
 WP_VERSION=5.4
```
Aljullu, line 15: Should `WP_VERSION` be updated here? We no longer support WP 5.4.

Aljullu (approved):
Thanks for investigating and fixing this @nerrad! Code changes look good and tests are passing locally and in this PR.

I agree with the proposed next steps, for now I will merge this PR as-is so we unblock other PRs waiting for tests to pass.

.env
nerrad: This is one of the unused parts I mentioned here:

> the docker configuration we're currently using seems to have a lot of unused parts and could likely be cleaned up a bit. What I added here complicates things a bit more.

The actual test suite currently ignores this and always uses the latest WP.

nerrad:
I'm not creating a followup issue for this item:

> Potentially introduce testing against more than one WooCommerce version (eg. latest and earliest matching our requirements) as a part of a test matrix. Although this is likely not as important as PHP version testing.

I don't think it's required for now.

PR #4312, opened 2021-06-05
Change `couponName` filter to `coupons` and update documentation

### LINKED ISSUE #4311: Adding more than one coupon to the cart results in an error
## Describe the bug
When more than one coupon is added to the cart, this error happens:

```Warning: React has detected a change in the order of Hooks called by TotalsDiscount. This will lead to bugs and errors if not fixed. For more information, read the Rules of Hooks: https://fb.me/rules-of-hooks

   Previous render            Next render
   ------------------------------------------------------
1. useMemo                    useMemo
2. undefined                  useMemo
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This is because `__experimentalApplyCheckoutFilter` uses `useMemo`. In the `TotalsDiscount` component we run this filter once for each coupon in the cart. This results in the number of hooks being executed changing between renders.

## To reproduce

1. Ensure you have two coupons available in your store.
1. Go to the Cart block and apply one.
2. Apply the second and see the error.

## Expected behavior
To be able to add multiple coupons without the Cart or Checkout breaking, and to also still allow extensions to filter coupons before displaying them in the sidebar.

opr:
#### Description
There is a bug where the coupon filter doesn't work if more than one coupon is added to the cart. This PR will change the way the coupon filter works to ensure the filter is only invoked once per render. Unfortunately this is a breaking change.

Another way to solve this would be to implement our own memoization function, and use that in `__experimentalApplyCheckoutFilter` rather than `React.useMemo`, or to simply remove memoization altogether (not desirable).

---

#### Changes made in this PR
- Add a `text` property to each `coupon` in the cart. This is required because when extensions change how the coupon is displayed, we need to keep a record of the code, so we can remove it from the cart if the shopper desires.
- Remove the `couponNames` filter, and replace it with `coupons`. Move the filter outside of the loop so it is always called once per render.
- Add deprecation notice if an extension has included `couponNames` in its filters object.
- Add documentation about new filter.

Fixes #4311

1. Ensure two coupons are available on your store. **Name one of them `testcoupon`.**
2. Go to the Cart block and add them both.
3. Ensure they work, and can be removed.
4. Try with three+ coupons.
5. Add this code somewhere in your application (I added it to the bottom of `packages/checkout/registry/index.ts`)
```typescript
__experimentalRegisterCheckoutFilters( 'blocks', {
	couponName: ( value ) => {
		return value;
	},
} );
```
6. Ensure the deprecation warning is shown in the console. `couponName is deprecated. Please use coupons instead.`
7. Remove that code and add:
```typescript
__experimentalRegisterCheckoutFilters( 'blocks', {
	coupons: ( value ) => {
		return value.map( ( coupon ) => {
			if ( coupon.code !== 'testcoupon' ) {
				return coupon;
			}
			return {
				...coupon,
				text: 'this is custom coupon text!',
			};
		} );
	},
} );
```
8. Reload the page and ensure the coupon that was previously labelled `testcoupon` is now shown as `this is custom coupon text!`

### Changelog

> Remove `couponName` filter and replace it with `coupons` filter.

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/discount/index.js
+++ b/assets/js/base/components/cart-checkout/totals/discount/index.js
@@ -41,12 +41,20 @@ const TotalsDiscount = ( {
 		? discountValue + discountTaxValue
 		: discountValue;
 
+	const filteredCartCoupons = __experimentalApplyCheckoutFilter( {
+		arg: {
+			context: 'summary',
+		},
+		filterName: 'coupons',
+		defaultValue: cartCoupons,
+	} );
+
 	return (
 		<TotalsItem
 			className="wc-block-components-totals-discount"
 			currency={ currency }
 			description={
-				cartCoupons.length !== 0 && (
+				filteredCartCoupons.length !== 0 && (
 					<LoadingMask
 						screenReaderLabel={ __(
 							'Removing coupon…',
@@ -56,30 +64,19 @@ const TotalsDiscount = ( {
 						showSpinner={ false }
 					>
 						<ul className="wc-block-components-totals-discount__coupon-list">
-							{ cartCoupons.map( ( cartCoupon ) => {
-								const filteredCouponCode = __experimentalApplyCheckoutFilter(
-									{
-										arg: {
-											context: 'summary',
-											coupon: cartCoupon,
-										},
-										filterName: 'couponName',
-										defaultValue: cartCoupon.code,
-									}
-								);
-
+							{ filteredCartCoupons.map( ( cartCoupon ) => {
 								return (
 									<RemovableChip
 										key={ 'coupon-' + cartCoupon.code }
 										className="wc-block-components-totals-discount__coupon-list-item"
-										text={ filteredCouponCode }
+										text={ cartCoupon.code }
 										screenReaderText={ sprintf(
 											/* translators: %s Coupon code. */
 											__(
 												'Coupon: %s',
 												'woo-gutenberg-products-block'
 											),
-											filteredCouponCode
+											cartCoupon.code
 										) }
 										disabled={ isRemovingCoupon }
 										onRemove={ () => {
@@ -92,7 +89,7 @@ const TotalsDiscount = ( {
 												'Remove coupon "%s"',
 												'woo-gutenberg-products-block'
 											),
-											filteredCouponCode
+											cartCoupon.code
 										) }
 									/>
 								);
--- a/assets/js/base/components/cart-checkout/totals/discount/index.js
+++ b/assets/js/base/components/cart-checkout/totals/discount/index.js
@@ -69,14 +69,14 @@ const TotalsDiscount = ( {
 									<RemovableChip
 										key={ 'coupon-' + cartCoupon.code }
 										className="wc-block-components-totals-discount__coupon-list-item"
-										text={ cartCoupon.code }
+										text={ cartCoupon.text }
 										screenReaderText={ sprintf(
 											/* translators: %s Coupon code. */
 											__(
 												'Coupon: %s',
 												'woo-gutenberg-products-block'
 											),
-											cartCoupon.code
+											cartCoupon.text
 										) }
 										disabled={ isRemovingCoupon }
 										onRemove={ () => {
@@ -89,7 +89,7 @@ const TotalsDiscount = ( {
 												'Remove coupon "%s"',
 												'woo-gutenberg-products-block'
 											),
-											cartCoupon.code
+											cartCoupon.text
 										) }
 									/>
 								);
--- a/assets/js/base/components/cart-checkout/totals/discount/index.js
+++ b/assets/js/base/components/cart-checkout/totals/discount/index.js
@@ -89,7 +89,7 @@ const TotalsDiscount = ( {
 												'Remove coupon "%s"',
 												'woo-gutenberg-products-block'
 											),
-											cartCoupon.text
+											cartCoupon.label
 										) }
 									/>
 								);
--- a/assets/js/base/components/cart-checkout/totals/discount/index.js
+++ b/assets/js/base/components/cart-checkout/totals/discount/index.js
@@ -69,14 +69,14 @@ const TotalsDiscount = ( {
 									<RemovableChip
 										key={ 'coupon-' + cartCoupon.code }
 										className="wc-block-components-totals-discount__coupon-list-item"
-										text={ cartCoupon.text }
+										text={ cartCoupon.label }
 										screenReaderText={ sprintf(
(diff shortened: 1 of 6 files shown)


docs/extensibility/available-filters.md
```diff
@@ -87,20 +87,24 @@ The word 'Total' that precedes the amount due, present in both the Cart _and_ Ch
...
 There are no additional arguments passed to this filter.
 
-### Coupon names
+### Coupons
 
 The current functionality is to display the coupon codes in the Cart and Checkout sidebars. This could be undesirable
-if you dynamically generate a coupon code that is not user friendly. It may, therefore, be desirable to change the way
-this code is displayed. To do this, the filter `couponName` exists. 
+if you dynamically generate a coupon code that is not user-friendly. It may, therefore, be desirable to change the way
+this code is displayed. To do this, the filter `couponName` exists.
```
ralucaStan: I think you missed one replace here: *the filter `coupons` exists*  instead of `couponName`

docs/extensibility/available-filters.md
```diff
@@ -87,20 +87,24 @@ The word 'Total' that precedes the amount due, present in both the Cart _and_ Ch
...
+
+This filter must _not_ be used to alter the value/totals of a coupon. This will not carry through to the Cart totals.
 
 | Filter name  | Description | Return type  |
 |---|---|---|
-| `couponName`  | This is the coupon code of a coupon currently applied to the cart. Its value will be the same as the code the customer entered to apply the discount. | `string`
+| `coupons`  | This is an array of coupons currently applied to the cart. | `CartCoupon[]`
 
-The additional argument supplied to this filter is: `{ context: 'summary', coupon: CartCoupon }`. A `CartCoupon` has the following shape:
+The additional argument supplied to this filter is: `{ context: 'summary' }`. A `CartCoupon` has the following shape:
```
ralucaStan, line 104: I am wondering if coupon was actually the only additional argument and maybe this line needs to go away?
About `CartCoupon`, as there is no longer any reference to it in this paragraph, I would remove the sentence from here and make it standalone.

ralucaStan: Unrelated to the changes, I've looked at the readme and there is no example of how one should actually use this filter. It would be great if you could add one, two examples with the coupons filter.

docs/extensibility/available-filters.md
```diff
@@ -87,20 +87,24 @@ The word 'Total' that precedes the amount due, present in both the Cart _and_ Ch
...
 
 ```typescript
-interface CartCoupon {
+CartCoupon {
   code: string
   discount_type: string
```
ralucaStan, line 110: As discussed, let's also add the text field in the definition, as developers can change it, it should be listed here.

assets/js/base/context/hooks/cart/use-store-cart.ts
```diff
@@ -175,8 +175,16 @@ export const useStoreCart = (
 				decodeValues( fee )
 			);
 
+			// Add a text property to the coupon to allow extensions to modify
+			// the text used to display the coupon, without affecting the
+			// functionality when it comes to removing the coupon.
+			const cartCoupons = cartData.coupons.map( ( coupon ) => ( {
+				...coupon,
+				text: coupon.code,
```
ralucaStan: Wondering if `label` would be more appropriate than `text`. For example, we have the Total label that the user can change, so to keep it in the same registry I'd say a `coupon label` makes it more clear that we are talking about what text is displayed. What do you think?

ralucaStan (changes requested):
This looks good, I really like the new field idea. I also recommend mentioning in the changelog the bug this PR is fixing from a user's perspective. Nice work 👍

assets/js/base/components/cart-checkout/totals/discount/index.js
```diff
@@ -92,7 +89,7 @@ const TotalsDiscount = ( {
 												'Remove coupon "%s"',
 												'woo-gutenberg-products-block'
 											),
-											filteredCouponCode
+											cartCoupon.text
```
ralucaStan: Shouldn't this be `cartCoupon.label` as well?

docs/extensibility/available-filters.md
```diff
@@ -201,6 +206,39 @@ __experimentalRegisterCheckoutFilters( 'my-hypothetical-price-plugin', {
...
+```typescript
+const filterCoupons = ( coupons ) => {
+  return coupons.map( ( coupon ) => {
+  	// Regex to match autocoupon then unlimited undersores and numbers
+    if ( ! coupon.text.match( /autocoupon(?:_\d+)+/ ) ) {
+      return coupon;
+    }
+    return {
+      ...coupon,
+      text: sprintf(
```
ralucaStan: You need to replace text with label here as well.

docs/extensibility/available-filters.md
```diff
@@ -201,6 +206,39 @@ __experimentalRegisterCheckoutFilters( 'my-hypothetical-price-plugin', {
...
+        pointsLabelPlural
+      ),
+    };
+  } );
+};
+```
+
+| Before | After |
+|---|---|
+| <img src="https://user-images.githubusercontent.com/5656702/123768988-bc55eb80-d8c0-11eb-9262-5d629837706d.png" /> | * |
```
ralucaStan: There is no after photo here from what I see.

assets/js/base/components/cart-checkout/totals/discount/index.js
opr: Great spot! Updated

docs/extensibility/available-filters.md
```diff
@@ -201,6 +206,39 @@ __experimentalRegisterCheckoutFilters( 'my-hypothetical-price-plugin', {
 |---|---|
 | <img src="https://user-images.githubusercontent.com/5656702/117035086-d5488300-acfb-11eb-9954-feb326916168.png" width=400 /> | <img src="https://user-images.githubusercontent.com/5656702/117035616-70415d00-acfc-11eb-98d3-6c8096817e5b.png" width=400 /> |
 
+### Change the name of a coupon
```
ralucaStan, line 209: Great job adding this example, I'm sure it's gonna be super useful

docs/extensibility/available-filters.md
opr: This isn't the case, the context is passed as the only additional argument, the coupon is no longer passed. The current state of this file does correctly describe what is passed to the filter function.

ralucaStan (approved):
Nice work!

PR #4313, opened 2021-06-05
Add snackbar filter

opr:
This PR will add a filter that will let extensions modify the notices queued to be shown in the snackbar.

It also makes the `arg` property on the arguments passed to `__experimentalApplyCheckoutFilter` optional, as this is not always necessary to pass.

1. Add a coupon to your site called `testcoupon`.
1. Add the following code to the bottom of `packages/checkout/registry/index.ts`:
```typescript

__experimentalRegisterCheckoutFilters( 'blocks', {
	snackbarNotices: ( value ) => {
		return value.filter( ( notice ) => {
			return ! notice.content.match( /testcoupon/ );
		} );
	},
} );
```
2. When adding and removing this coupon ensure the snackbar is not shown.
3. Remove the coupon and add a different one. (Ensure you remove the coupon first, due to a known bug where adding two breaks the cart: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4312)
4. When adding a coupon that is not `testcoupon` ensure the snackbar _is_ shown.

### Changelog

> Add filter to allow extensions to control the notification snackbar.

### DIFF
--- a/assets/js/base/context/providers/store-notices/components/snackbar-notices-container.js
+++ b/assets/js/base/context/providers/store-notices/components/snackbar-notices-container.js
@@ -3,6 +3,7 @@
  */
 import { SnackbarList } from 'wordpress-components';
 import classnames from 'classnames';
+import { __experimentalApplyCheckoutFilter } from '@woocommerce/blocks-checkout';
 
 const SnackbarNoticesContainer = ( {
 	className,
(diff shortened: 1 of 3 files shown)


Aljullu (approved):
This is testing good and code changes look fine as well.

PR #4317, opened 2021-06-07
Prevent PHP notice for variable products without enabled variations.

sebcode:
`ProductSchema::get_variations` throws a PHP notice due to an invalid SQL query for products without enabled variations, for example for this query: `SELECT post_id as variation_id, meta_key as attribute_key, meta_value as attribute_value FROM wp_postmeta WHERE post_id IN () AND meta_key IN ('attribute_pa_size')`

1. Create a variable product with one disabled variation and at least one variation attribute.
2. Create a new blog post with a "Product category" block and select the category that contains the newly created product.
3. Visiting the new blog post throws a PHP notice (`NOTICE: PHP message: WordPress database error You have an error in your SQL syntax ...`)

### Changelog

> Prevent PHP notice for variable products without enabled variations.

### DIFF
--- a/src/StoreApi/Schemas/ProductSchema.php
+++ b/src/StoreApi/Schemas/ProductSchema.php
@@ -556,7 +556,12 @@ class ProductSchema extends AbstractSchema {
 		}
 		global $wpdb;
 
-		$variation_ids               = $product->get_visible_children();
+		$variation_ids = $product->get_visible_children();
+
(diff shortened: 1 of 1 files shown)


opr (approved):
Thanks for your PR! I've just tested this and verified the notice is gone, the block functions as normal. The code change is great, and the formatting is fine as well.

🎉

PR #4323, opened 2021-06-08
Update Reviews blocks so they use block styles

### LINKED ISSUE #1676: Reviews blocks: author and date text doesn't have enough contrast
Text color is `#808080` and background is white, so there [isn't enough contrast](https://webaim.org/resources/contrastchecker/?fcolor=808080&bcolor=FFFFFF).

**Screenshots**
imatge

### LINKED ISSUE #4302: Update a block so it uses block styles and block settings
[WordPress 5.8 will include block styles and block settings](https://github.com/WordPress/gutenberg/issues/20331#issuecomment-848685044). We don't have a plan to update all our blocks yet, but we would like to start experimenting with these features. This issue is about updating one or two blocks to use block styles and block settings.

This issue should allow us to answer these two questions:

* How much effort is required to update one block so it uses block styles and block settings?
* Is it possible to make these updates and keep blocks compatible with older versions of WordPress?

Aljullu:
Fixes #1676.
Fixes #4302.

This PR:
1. Adds global styles to Reviews blocks so its possible to change the text color and font size.
2. Updates Reviews blocks styling so there are no hard-coded colors and so everything aligns well with different font sizes.
3. Updates component `EditorContainerBlock` to a functional component.

### Screenshots

| Default display (before) | Default display (after) |
| --- | --- |
| imatge | imatge |

Tweaking some colors and font size in Storefront:
imatge

1. In a classic theme (ie: Storefront):
	1. Add the All Reviews block to a post or page.
	2. Verify you can change the text color and font size.
	3. Repeat the process for the Reviews by Category and Reviews by Product blocks.
1. In a block-based theme (ie: [TT1 Blocks](https://wordpress.org/themes/tt1-blocks/)) with Gutenberg enabled:
	1. Go to Appearance > Site editor, click on the Global styles icon and verify the All Reviews block is shown and you can tweak its styles:
imatge
	2. Add the All Reviews block to a post or page.
	3. Verify it honors the styles you set in the Site editor.
	4. Change the styles in the post/page editor and verify they have priority over the styles from the Site editor.
	3. Repeat the process for the Reviews by Category and Reviews by Product blocks.

### Changelog

> Added global styles to All Reviews, Reviews by Category and Reviews by Product blocks. Now it's possible to change the text color and font size of those blocks.

### DIFF
--- a/assets/js/base/components/reviews/review-list-item/index.js
+++ b/assets/js/base/components/reviews/review-list-item/index.js
@@ -14,11 +14,7 @@ import './style.scss';
 function getReviewImage( review, imageType, isLoading ) {
 	if ( isLoading || ! review ) {
 		return (
-			<div
-				className="wc-block-review-list-item__image wc-block-components-review-list-item__image"
-				width="48"
-				height="48"
-			/>
+			<div className="wc-block-review-list-item__image wc-block-components-review-list-item__image" />
 		);
 	}
 
@@ -34,8 +30,7 @@ function getReviewImage( review, imageType, isLoading ) {
 				<img
 					aria-hidden="true"
 					alt=""
-					src={ review.reviewer_avatar_urls[ '48' ] || '' }
-					srcSet={ review.reviewer_avatar_urls[ '96' ] + ' 2x' }
+					src={ review.reviewer_avatar_urls[ '96' ] || '' }
 				/>
 			) }
 			{ review.verified && (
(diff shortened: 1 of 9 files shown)


mikejolley (approved):
Wow that new editor is super confusing 😅 

Works great. I tested on both and it works as described. Only comment from my side would be, do you think it would make sense to scale the avatar with the text size?

Other than that, good to go 👍🏻

Aljullu:
> Only comment from my side would be, do you think it would make sense to scale the avatar with the text size?

Good idea. Implemented in a65715548d6d44e5d9436d7f8dbe677bd95c1f50. I had to remove `srcSet` because now that the image size depends on the font size, we can't know in advance which size is needed.

PR #4324, opened 2021-06-08
Document and update naming of Block Asset Handles

### LINKED ISSUE #2660: Establish a consistent naming schema for asset handles.
This issue is prompted from a [comment made by @haszari](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2648#issuecomment-640275006) which appropriately draws attention to the potential for regression if we don't namespace the handles we used for built assets registered with WP.

For back-compatibility reasons, I don't think we should change any of the established script handles (except maybe `wc-vendors` and `wc-shared-context` (as that seems like something other WC plugins - or core itself - might use). Since there seem to be two patterns of how namespacing is used, I propose the following convention (to keep things consistent with existing patterns):

- If the asset is for a block, then use `block` as the suffix. Example: `wc-checkout-block`
- If the asset is some other bundled script, then use `wc-blocks-` as the prefix. (Note: we have some uses of `wc-block-` prefix right now...I'm on the fence whether we change them, for back compat reasons).

Should there be agreement, we could document this in our `/docs` folder.

mikejolley:
This PR establishes the pattern of naming block handles as `wc-{blockname}-block`, and other scripts as `wc-blocks-{script}` to avoid conflicts.

To help map handles -> built scripts I've also made script filenames match. I did not do this for the blocks themselves because it caused some breakage in Atomic Blocks and complicates how blocks are setup in webpack.

Closes #2660

- Tests should pass
- Check existing block types still show up in the editor and can be inserted without regression.
- Console should be error-free
- Specifically test single-product block works to verify Atomic Blocks are not broken

### Dev Note

Some script handles have been changed in this version for consistency. It is unlikely that these handles are being used directly by 3rd parties because they are only used within the blocks, however for reference:

- `wc-block-vendors-style` -> `wc-blocks-vendors-style`
- `wc-block-editor` -> `wc-blocks-editor-style`
- `wc-block-style` -> `wc-blocks-style`
- `wc-vendors` -> `wc-blocks-vendors`
- `wc-shared-context` -> `wc-blocks-shared-context`
- `wc-shared-hocs` -> `wc-blocks-shared-hocs`

In addition, all block script handles now have the `-block` suffix.

### Changelog

> Made script and style handles consistent.


Aljullu (approved):
Works great and code changes look good. Nice cleanup! :clap: 

@mikejolley do you think this PR needs a dev note or something similar? My understanding is that it will impact any theme/plugin that relies on our script/style handles.

mikejolley:
@Aljullu I'm uncertain — the scripts and styles we've changed are our own block handles (the external packages, price, and checkout, were not changed). Can you think of a use case where someone would use these handles?

Aljullu:
> @Aljullu I'm uncertain — the scripts and styles we've changed are our own block handles (the external packages, price, and checkout, were not changed). Can you think of a use case where someone would use these handles?

Yeah... I'm not sure. I'm thinking on a store that uses WooCommerce but none of WooCommerce Blocks, which might have unregistered our styles/scripts? Not sure if that's a case we need to account for, though, sounds very edge-case. Leaving the decision to you.

mikejolley:
Added a note 👍🏻

PR #4325, opened 2021-06-08
Refactor Render Parent Block Function for Atomic Blocks

mikejolley:
This is a change pulled from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268/files to add more flexibility to the `renderParentBlock` function. I will add inline notes regarding the specific changes.

This also converts the function to Typescript.

1. Insert the Single Product Block on a page
2. Choose a product
3. Check it renders correctly on the frontend


assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+/**
+ * Renders a block component in the place of a specified set of selectors.
+ */
+export const renderParentBlock = ( {
+	Block,
+	selector,
+	blockName,
+	getProps = () => ( {} ),
+	blockMap,
```
mikejolley, line 106: This now accepts the `blockMap` instead of looking up the block map itself. This allows someone using this function to pass in their own block map (block name -> component conversion) which is something we required for Checkout i2.

assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -0,0 +1,126 @@
...
+	blockWrapper,
```
mikejolley, line 107: This lets you pass in a wrapper for inner blocks. In the case of Checkout i2, we'll pass in a wrapper with some props that are exposed to children.

opr (approved):
LGTM - Code changes seem fine too, not sure you need to explicitly declare the types on the map in `renderInnerBlocks` but it doesn't hurt to, so 👍🏼

PR #4340, opened 2021-06-10
Change some snake cased properties to camelCase - mostly related to Shipping and Totals

### LINKED ISSUE #4150: Find places where snake_cased names are received from the API and convert them to camelCase
When we get data from the API, the keys are in snake_case, but in JavaScript we prefer camelCase. We need to inhibit eslint warnings when we continue to use snake_cased variable names. An example of this happening can be found at the following URL.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/components/cart-checkout/shipping-rates-control-package/index.tsx#L45

We should immediately convert snake_cased data to camelCase before doing anything with it.

To solve this issue we should identify instances where we have used `// eslint-disable-next-line camelcase`, change the types to be camelCase, and ensure the snake_cased data is mapped to camelCase as soon as it's received from the API.

### LINKED ISSUE #4250: Change `getSetting` to accept a generic type to avoid casting the result
At present, when using `getSetting` we have to cast the result to a specific type to appease TypeScript. We could change `getSetting` to accept a generic type argument which will allow us to specify the type returned by the function. 

I suggest changing `getSetting` to this (note only the generic argument and return type is changing)

```typescript
export const getSetting = < T >(
	name: string,
	fallback: unknown = false,
	filter = ( val: unknown, fb: unknown ) =>
		typeof val !== 'undefined' ? val : fb
): T => {
	const value = name in allSettings ? allSettings[ name ] : fallback;
};
```

and then finding uses of `getSetting` and ensuring the type argument is used rather than casting.

opr:
This PR changes the casing of properties in the shipping rates section of the cart from `snake_case` to `camelCase`. The cart_totals property was also changed to be camel cased.

- Convert `has-shipping-rate` to TypeScript
- Convert `derive-selected-shipping-rates` to TypeScript
- Create `mapKeysToCamelCase` function and unit tests for it. This function is designed to change any object's keys to camel case, even nested objects within it. It leaves non-plain objects alone, but does check items in an array, and converts them too if any members are a plain object.
- Add `isPlainObject` type guard. This differentiates between things like `Map`, `Array`, etc.
- Add tests for it.
- Update JS unit tests to account for changes in casing in the dummy data
- Add a new `PreviewCart` type - this is a hybrid of the `Cart` and `PreviewCart` types, when we finish converting all of the properties to camel case this can be removed in favour of simply using `Cart`.
- Add CSS fix for image sizes in the checkout editor.

Fixes #4150 and fixes #4250

1. Please extensively smoke test the cart and checkout front-end _and_ editor screen.
2. Ensure you test lots of different combinations of tax/shipping/payment settings.
3. Ensure the correct currency is displayed when in the cart and checkout.
3. Ensure you are billed the correct amount when paying.


opr:
Closing following the latest comment in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4150 and in slack p1623837562194000-slack-C8X6Q7XQU

PR #4342, opened 2021-06-11
Add a workflow for the assign-milestone automation

mikejolley:
Includes the assign-milestone automation added in https://github.com/woocommerce/automations/pull/41

To test, approve this PR. It should be assigned a milestone automatically when approved. Other workflows should continue to run normally (label etc).

If this assigns to the wrong milestone, please unapproved so we can tweak somehow.

### DIFF
--- a/.github/workflows/approved-with-labels.yml
+++ b/.github/workflows/approved-with-labels.yml
@@ -22,4 +22,4 @@ jobs:
               with:
                   github_token: ${{ secrets.GITHUB_TOKEN }}
                   automations: assign-milestone
-                  milestone_bump_strategy: minor
+                  milestone_bump_strategy: none
--- a/.github/workflows/approved-with-labels.yml
+++ b/.github/workflows/approved-with-labels.yml
@@ -1,14 +1,24 @@
 on: pull_request_review
-name: Add Labels to Approved Pull Requests
+name: Approved Pull Requests
 jobs:
-  labelWhenApproved:
-    name: Label when approved
-    runs-on: ubuntu-latest
-    steps:
-    - name: Label when approved
-      uses: pullreminders/label-when-approved-action@master
-      env:
-        APPROVALS: "1"
-        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        ADD_LABEL: "status: ready to merge"
-        REMOVE_LABEL: "status:%20needs%20review"
+    labelWhenApproved:
+        name: Add Labels
+        runs-on: ubuntu-latest
+        steps:
+            - name: Ready to merge label
+              uses: pullreminders/label-when-approved-action@master
+              env:
+                  APPROVALS: '1'
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  ADD_LABEL: 'status: ready to merge'
+                  REMOVE_LABEL: 'status:%20needs%20review'
+    milestoneWhenApproved:
+        name: Add Milestone
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
+            - name: Latest milestone
+              uses: woocommerce/automations@v1
+              with:
+                  github_token: ${{ secrets.GITHUB_TOKEN }}
+                  automations: assign-milestone
--- a/.github/workflows/approved-with-labels.yml
+++ b/.github/workflows/approved-with-labels.yml
@@ -22,3 +22,4 @@ jobs:
               with:
                   github_token: ${{ secrets.GITHUB_TOKEN }}
                   automations: assign-milestone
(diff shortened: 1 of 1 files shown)


Aljullu (approved):

Aljullu:
@mikejolley this got added to the 5.3.1 milestone instead of 5.4.0. In the way we work in WC Blocks, it would be great if this workflow excluded minor versions. Is that possible to implement and do you think it's worth it?

mikejolley:
@Aljullu I have updated the action, so if you approve again we should see the correct milestone applied.

Aljullu (approved):

mikejolley:
> Could not rediscover milestone by title: 5.5

5.5 wasn't created yet 🤦🏻

Aljullu:
Hmm, I might be misunderstanding this, but in `package.json` we want to have the next version number + `-dev` (ie now it is `5.4.0-dev`). So we don't need to add `1` to the version number [here](https://github.com/woocommerce/automations/pull/44/files#diff-3d2b59189eeedc2d428ddd632e97658fe310f587f7cb63b01f9b98ffc11c0197R160-R166), otherwise it will look for the milestone of the version after the next one (ie, in this PR it's trying to find the 5.5 milestone).

mikejolley:
@Aljullu yeah that's annoying. We're doing something different to Gutenberg where I took the idea from. I'm not sure if we should account for this or stop bumping package versions before a release.. @nerrad

nerrad:
>  I'm not sure if we should account for this or stop bumping package versions before a release.. @nerrad

I think bumping the version helps because we know what development version it's for. Perhaps we could add some basic configuration for the action to control how it handles the version retrieved from `package.json`? Some simple strategies on a flag?

For example being able to add something like:

```yaml
steps:
    - uses: actions/checkout@v2
    - name: Latest milestone
      uses: woocommerce/automations@v1
      with:                  
        github_token: ${{ secrets.GITHUB_TOKEN }}
        automations: assign-milestone
        bump_strategy: 'ignore';
```

Where `bump_strategy` could be initially just handle two options:

- `ignore` meaning it will just use the version as is from `package.json`
- `minor` meaning it will bump the minor version (the current bumping logic in the action).

To better support the structure of the automations architecture, we might actually do something like this (so `config` could be any arbitrary configuration values passed through to the various automations:


```yaml
steps:
    - uses: actions/checkout@v2
    - name: Latest milestone
      uses: woocommerce/automations@v1
      with:                  
        github_token: ${{ secrets.GITHUB_TOKEN }}
        automations: assign-milestone
        config: 
            bump_strategy: 'ignore';
```

mikejolley:
> To better support the structure of the automations architecture, we might actually do something like this (so config could be any arbitrary configuration values passed through to the various automation:

Is there any prior art for this? TBH I was trying to avoid adding config support because it's increasing scope and I'm not that familiar with the automation repo yet :)

nerrad:
The [automations configuration object](https://github.com/woocommerce/automations/blob/trunk/lib/automations/release/index.js) can define a `getConfig` function that will be [called to retrieve any configuration](https://github.com/woocommerce/automations/blob/51052514802d6389a912e9dd5106cffcdd613681/lib/index.js#L60) and the return value from the function [gets passed into the runner as the third argument](https://github.com/woocommerce/automations/blob/51052514802d6389a912e9dd5106cffcdd613681/lib/index.js#L64).

You could feasibly use the `getInput` method from `@actions/core` to get the configuration property from the yaml (see how this is being used for the [`automations` value here](https://github.com/woocommerce/automations/blob/51052514802d6389a912e9dd5106cffcdd613681/lib/index.js#L20)) in your custom `getConfig` callback for the milestone automation and use that to derive the bump strategy.

mikejolley (commented):
@nerrad it doesn't look like you can have nested inputs; unless you know different? I may need to add them at the same level as `automations` and define them in the actions.yaml file.

mikejolley:
I've added milestone_bump_strategy minor to this. If approved it should inherit the next minor release milestone.

Aljullu (approved):

Aljullu (changes requested):
@mikejolley it seems to still have the issue that adds `1` to the minor version even though the version in `package.json` is already the next version:

```
assign-milestone: Current plugin version is 5.6.NaN
assign-milestone: Could not rediscover milestone by title: 5.7
```

In this case, the milestone version to use should be 5.6.

mikejolley:
Ah thats right, we need a bump policy of "none" to use the package.json. Apologies, picking this up after a few weeks and forgot.

Aljullu (approved):
Let's try it now. :slightly_smiling_face:

Aljullu:
@mikejolley seems to be working great now. Thanks for working on this, great to see more things being automated. :clap:

PR #4345, opened 2021-06-14
Load woocommerce.css in Site editor

Aljullu:
Fixes #4332.

### Screenshots

| Before | After |
| --- | --- |
| Screenshot where rating stars are replaced by S | Schreenshot where rating stars are rendered correctly |

### How to test the changes in this Pull Request:

1. Check out this branch in WooCommerce core: `fix/woocommerce-css-editor` (related PR: https://github.com/woocommerce/woocommerce/pull/30093).
2. Enable a block-based theme (ie: TT1 Blocks) and Gutenberg.
3. Go to the Site Editor.
4. Add a Reviews block and verify rating stars are rendered correctly.

### Changelog

> Fix rating stars not being shown in the Site Editor.

PR #4350, opened 2021-06-15
Use old wc-block-style handle

Aljullu:
In #4345 we added `woocommerce-general` style as a dependency of `wc-blocks-style`. However, in 5.3.0, that stylesheet had a different handle: `wc-block-style`, so when cherry-picking that PR into `release/5.3.0`, it stopped working.

This PR updates the fix introduced in #4345 so it uses the old handle.

**Note: this PR should not be merge into `trunk`.**

Same as #4345:

1. Make sure you are using the latest version of WooCommerce `trunk`.
2. Enable a block-based theme (ie: TT1 Blocks) and Gutenberg.
3. Go to the Site Editor.
4. Add a Reviews block and verify rating stars are rendered correctly.

| Before | After |
| --- | --- |
| Screenshot where rating stars are replaced by S | Schreenshot where rating stars are rendered correctly |

### DIFF
--- a/src/AssetsController.php
+++ b/src/AssetsController.php
@@ -127,7 +127,7 @@ final class AssetsController {
 	 */
 	public function update_block_style_dependencies() {
 		$wp_styles = wp_styles();
-		$style     = $wp_styles->query( 'wc-blocks-style', 'registered' );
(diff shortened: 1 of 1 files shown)


opr (approved):
Working fine for me! 🚢

PR #4348, opened 2021-06-15
Release/5.3.1

Aljullu:
This is the patch release pull request for WooCommerce Blocks plugin `5.3.1`.

## Changelog

---
```
- Fix Product Categories List block display in Site Editor (#4335).
- Make links in the Product Categories List block unclickable in the editor (#4339).
- Fix rating stars not being shown in the Site Editor (#4345).
```

---

## Communication

This release introduces some fixes in preparation for WordPress 5.8.

### Prepared Updates

The following documentation, blog posts, and changelog updates are prepared for the release:

**Release announcement:** https://developer.woocommerce.com/2021/06/15/woocommerce-blocks-5-3-1-release-notes/

## Quality

> This section is for any notes related to quality around the release Please include any extra details with each item as needed. This can include notes about why something isn't checked or expanding info on your response to an item.

* [ ] Changes in this release are covered by Automated Tests.

      > This section is for confirming that the release changeset is covered by automated tests. If not, please leave some details on why not and any relevant information indicating confidence without those tests.

     * [ ] Unit tests
     * [ ] E2E tests
     * [ ] for each supported WordPress and WooCommerce core versions.

* This release has been tested on the following platforms:
     * [x] mobile
     * [x] desktop

* [ ] This release affects public facing REST APIs.
    * [ ] It conforms to REST API versioning policy.

* [ ] This release impacts **other extensions** or **backward compatibility**.
    * [ ] The release changes the signature of public methods or functions
        * [ ] This is documented (see: *Enter a link to the documentation here*)
    * [ ] The release affects filters or action hooks.
        * [ ] This is documented (see: *Enter a link to the documentation here*)

* [x] Link to **testing instructions** for this release: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/49a88e911a8d9513f75d987a3ef17dedc2e6af67/docs/testing/releases/531.md

* [ ] The release has a negative performance impact on sites.
    * [ ] There are new assets (JavaScript or CSS bundles)
    * [ ] There is an increase to the size of JavaScrip or CSS bundles) *please include rationale for this increase*
    * [ ] Other negative performance impacts (if yes, include list below)

* [ ] The release has positive performance impact on sites. If checked, please document these improvements here.

------

PR #4351, opened 2021-06-15
Add 5.3.1 testing instructions

Aljullu:
Given that #4348 has commits we don't want to merge to `trunk` (see #4350), I'm creating this PR which cherry-picks the version updates and testing instructions from 5.3.1.


opr (approved):
🙌🏼

PR #4360, opened 2021-06-16
Fix error regarding duplicate keys in TotalsFees

opr:
This PR fixes the `Warning: Each child in a list should have a unique "key" prop.` experienced when fees are present on the cart.

This occurs when the fee does not have an ID, we can repair this by using the fee's ID as its key, but if it's not present then we can use the fee's name and its index in the fees list to generate a unique key.

1. Add fees to your cart (using the extra fees extension https://woocommerce.com/products/extra-fees-for-woocommerce/ set up an order fee, or any other combination of fees).
2. Go to the cart/checkout blocks and ensure no console warning is shown.

### Changelog

> Fix a warning shown when fees are included in the order.

### DIFF
--- a/packages/checkout/totals/fees/index.tsx
+++ b/packages/checkout/totals/fees/index.tsx
@@ -26,7 +26,7 @@ const TotalsFees = ( {
 }: TotalsFeesProps ): ReactElement | null => {
 	return (
 		<>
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
LGTM

PR #4362, opened 2021-06-16
Build: Fix errors related to WP5.8-beta2 updates and drop IE11 support.

nerrad:
After updating to WordPress 5.8-beta2, we noticed that all our builds were breaking in the browser with an error containing something like this (variations of actual text, but all referencing undefined 'mark'):

```
 errors.js:15 Uncaught TypeError: Cannot read property ‘mark’ of undefined
```

This can be traced to a missing polyfill as a result of [this change in WP core](https://core.trac.wordpress.org/ticket/52941#comment:24) for the WP 5.8 branch.

While there are some efforts to restore this polyfill in WordPress(which will likely negate the urgency of this Pull Request), I'm still publishing this as something that we'll want to work on anyways (and we can decide whether it needs to go in the Woo Core 5.5 release as a part of Woo Blocks 5.3 branch depending on the outcome of the WP core changes and if there are any other polyfill related issues).

Unfortunately, at the moment I can't own the completion of this work, so I've pushed this in progress PR which functionally fixes all the issues by:

- updating WP related babel packages.
- switching to `@wordpress/babel-preset-default` in our Webpack configuration and removing async generator.
- fixing some reference errors exposed by the changes (likely because previously Webpack polyfills were hiding the issue).

Also:

- This drops IE11 support and adopts what the latest `@wordpress/browserlist-config` settings are.
- Cleans up a bit of our package dependencies and Webpack configuration for babel config.

I've tested this on WP 5.8-beta2 and WP 5.7.2 and things seem to work okay but note that the Stripe Extension and WooCommerce Payments are affected by this bug, so if you have either of those active you will still experience similar errors in the checkout/cart blocks.

## Next Steps

- Do more thorough testing against supported WP versions.
- Test across browsers we support to make sure there are no unexpected bugs surfacing (I tested latest Chrome).
- Audit our Webpack config and package dependencies. I have a feeling we can likely get rid of most of the babel plugins we're using directly in the webpack config with the switch to `@wordpress/babel-preset-default`. We might just need to keep the prop-types removal plugin (and maybe the class properties one, but should check if that's only needed for IE11 support...) for production builds.
- Follow the linked WP core issue and see if any fixes they apply to restore the polyfill also fixes using Woo Blocks trunk with WP 5.8. That will help decide whether we need to cherrypick this PR into the Woo Blocks 5.3.1 branch (there will be other issues to deal with if that's the case as earlier versions of Woo Core would likely have blocks break in WP 5.8).
- ~I haven't checked whether our automated test suite fails or not related to this. So any fixes in test builds will need addressed.~ (they pass)
- Same as ^^ for Storybook builds.


mikejolley:
Is this PR still needed?

nerrad:
It's no longer needed to address any errors, but the performance gains from dropping IE11 support in our builds would be nice.

mikejolley:
I extracted the IE11 stuff into an updated PR here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5212

PR #4363, opened 2021-06-16
Allow HTML in the All Products Block Product Titles

### LINKED ISSUE #1685: WC All Products Block Titles are not rendering HTML
**Describe the bug**
The title in the new All Products Block does not render HTML. They are shown as plain text.

**To Reproduce**
1. Add additional HTML tags via the woocommerce_product_title filter
2. Insert the "All Products" Block

**Screenshots**
Screenshot of All Product Block Title

**Expected behavior**
HTML tags are getting rendered

mikejolley:
Uses `dangerouslySetInnerHTML` to allow HTML in product titles. Titles coming from the StoreAPI are ran through `wp_kses_post`. Note that core also allows HTML in product titles and does not face the same problem.

Fixes #1685

### Screenshots

Using a strong tag:

Screenshot 2021-06-16 at 17 08 12

1. Edit a product and add HTML to the title e.g. `<strong>Something</strong>`
2. View the product within the All Products Block
3. Confirm the strong tag is rendered, rather than the HTML as a string

### Changelog

> Allow HTML in All Products Block Product Titles.


opr (approved):
Looks good to me, is the title run though `wp_kses_post` when saving _and_ fetching? I don't suppose it matters, does it? As long as it's done at some point.

mikejolley:
Saving is handled in the editor, and we're applying it on fetch :)

PR #4367, opened 2021-06-17
Add screen reader text to price range

### LINKED ISSUE #1669: Make price ranges accessible
In many blocks we display price ranges:

imatge
imatge
imatge

Currently, screen readers read both prices without specifying what relation is there between them. Instead, it should be read as "From €18,00 to €45,00", or something similar.

The solution we apply for that, should work for all blocks that display prices, including JS-rendered and PHP-rendered blocks.

mikejolley:
Adds screen reader text to price ranges stating: "Price between X and X".

Fixes #1669

#### Accessibility

1. View all products block
2. Find a product with a price range (grouped product). e.g. Logo Collection in default content.
3. Use voiceover or screen reader to read the price. It should announce "Price between X and X". Some tools will also read the other price but this is wrapped in aria-hidden so can be ignored.

### Changelog

> Add screen reader text to price ranges

### DIFF
--- a/assets/js/base/components/product-price/index.js
+++ b/assets/js/base/components/product-price/index.js
@@ -1,11 +1,12 @@
 /**
  * External dependencies
  */
-import { __ } from '@wordpress/i18n';
+import { __, sprintf } from '@wordpress/i18n';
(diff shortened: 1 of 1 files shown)


opr (approved):
Working well for me! 🚢

PR #4369, opened 2021-06-18
Switch from Select to Combobox for Country and State Inputs

### LINKED ISSUE #1752: Migrate CountryInput and StateInput to ComboboxControl

mikejolley:
This PR implements a new Combobox control which essentially allows text input to search a list of values. This is useful for country and state fields and is similar to what Woo Core has on the core checkout.

The control being used is the [ComboBox Control from Gutenberg](https://developer.wordpress.org/block-editor/reference-guides/components/combobox-control/).

Fixes #1752

Looking for feedback and ideas to further polish. To get the label working I needed to simulate onBlur. Autofill was also challenging. I ended up making it search when the input changes which causes an immediate update so I am not 100% sure this is a good solution.

#### Accessibility

Tested Keyboard controls.

### Screenshots

Screenshot 2021-06-18 at 13 50 00
Screenshot 2021-06-18 at 13 49 33
Screenshot 2021-06-18 at 13 48 53

1. Try searching for values in the country input. Value should persist on change.
2. Repeat for state input (for countries with states)
3. Test autofill populates the field
4. Test on a mobile browser (I used mobile view on Edge)
5. Test on a variety of themes

I went through all the TwentyX themes and Storefront.

### Changelog

> Checkout: Switch from select element to combobox for country and state inputs so contents are searchable


mikejolley:
Tests will need updating due to the different country select markup.

assets/js/base/components/combobox/style.scss
```diff
@@ -0,0 +1,156 @@
+.wc-block-components-form .wc-block-components-combobox,
```
Aljullu, line 1: Why is this selector required? If it is for specificity reasons, I would add a comment explaining that, and maybe using `.wc-block-components-combobox.wc-block-components-combobox` would be better so styles have the same specificity independently if it's inside a form component or not?

assets/js/base/components/country-input/country-input.tsx
```diff
@@ -9,7 +9,7 @@ import classnames from 'classnames';
 /**
  * Internal dependencies
  */
-import { ValidatedSelect } from '../select';
```
Aljullu, line 12: It looks like `ValidatedSelect` (and `Select`) weren't used anywhere else. Do you think we should delete them?

Aljullu (commented):
I really love how it works and looks. I think this is a big improvement to the user experience. :clap: 

I tested several flows and features:

* Desktop, mobile and screen reader.
* Several themes.
* Autofill.
* Validation.

It worked great in all of them!

The only thing that worries me a bit is the size increase in cart and checkout frontend scripts. But I think the increase is worth it considering how the combobox simplifies the filling experience.

> To get the label working I needed to simulate onBlur.

If it's only for styling purposes, I wonder if we could use the [`:focus-within` pseudoclass](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within). Now that we don't need to support IE11, this is a new tool we can make use of.

assets/js/base/components/combobox/style.scss
mikejolley: I based this off of the select box control's CSS which has the same.

assets/js/base/components/country-input/country-input.tsx
mikejolley: That was a question I had :) Can we? I don't think these are exported so I think perhaps yes.

assets/js/base/components/combobox/style.scss
Aljullu: Aaah, I see. I would make the change in both, but not blocking this PR, then.

assets/js/base/components/country-input/country-input.tsx
Aljullu: If they are not used anywhere neither exported, I would delete them, yes. We can always recover them with git. :slightly_smiling_face:

mikejolley:
@Aljullu Feedback taken care of :) The focus-within change worked really well, thanks for that.

assets/js/base/components/cart-checkout/address-form/test/index.js
```diff
@@ -162,17 +159,11 @@ describe( 'AddressForm Component', () => {
 		inputAddress( secondaryAddress );
 		// Only update `country` to verify other values are reset.
 		inputAddress( { country: primaryAddress.country } );
-
-		expect( screen.getByLabelText( cityRegExp ).value ).toBe( '' );
 		expect( screen.getByLabelText( stateRegExp ).value ).toBe( '' );
-		expect( screen.getByLabelText( postalCodeRegExp ).value ).toBe( '' );
```
Aljullu, line 168: I thought the expected behavior was that when the country changed, all other address fields would be reset. That's especially important for countries which don't have a state or postal code, so we don't carry over values from the previous address. Was this behavior changed on purpose?

Aljullu (commented):
Thanks for updating this PR @mikejolley! I didn't repeat the tests in all devices, but I did it on desktop and it's working fine.

Something that I think would improve the user experience is auto-selecting the first option of the combobox by default, would that be possible?

Selecting a country

In the example above, I'm writing `estonia` and pressing <kbd>Enter</kbd>, but the only suggestion available is not autoselected. Instead, I would need to press <kbd>Key down</kbd> + <kbd>Enter</kbd> or select it with the mouse. In `trunk`, however, writing a country name and pressing <kbd>Enter</kbd> works, so this is kind of a regression (not sure if blocking, though).

assets/js/base/components/cart-checkout/address-form/test/index.js
mikejolley: Yeah, I will check this again, but when you're filling out the form, if fields clear because you input country last that's super annoying. It also breaks autocomplete. Ideally (and what I think I checked) fields should be cleared in the background if they are not needed. We don't need to do this in client/form.

mikejolley:
@Aljullu That issue seems to be coming from the combo control: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/combobox-control/index.js#L112-L117

If you have a value already, then type, then hit enter, the above code triggers and returns your selection to the original value.

You'll see that if you just type, or click out of the input, our custom logic triggers and keeps the new value selected.

It doesn't look like we have control over the behaviour or enter. Any suggestions?

assets/js/base/components/cart-checkout/address-form/test/index.js
mikejolley: @Aljullu the new logic is here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4369/files#diff-54ee17137efc6b9816bcbf193608f7a40333290f7deee587370a1dbcad331f9fR140-R147 It will clear the value if the field is hidden for the locale. Otherwise, there is no need to clear.

assets/js/base/components/cart-checkout/address-form/address-form.tsx
```diff
@@ -137,6 +137,13 @@ const AddressForm = ( {
 		<div id={ id } className="wc-block-components-address-form">
 			{ addressFormFields.map( ( field ) => {
 				if ( field.hidden ) {
+					// clear hidden field value.
+					if ( values[ field.key ] ) {
+						onChange( {
+							...values,
+							[ field.key ]: '',
+						} );
+					}
```
Aljullu: That means `onChange()` is called on each render, right? Could this be refactored so it's run inside a `useEffect()` when the country or state change?

assets/js/base/components/cart-checkout/address-form/test/index.js
Aljullu: > Yeah, I will check this again, but when you're filling out the form, if fields clear because you input country last that's super annoying. It also breaks autocomplete. Ideally (and what I think I checked) fields should be cleared in the background if they are not needed. We don't need to do this in client/form.

Makes sense. :ok_hand:

Aljullu (approved):
> @Aljullu That issue seems to be coming from the combo control: https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/combobox-control/index.js#L112-L117

Thanks for investigating! Tested the [`ComboboxControl` from the GB storybook](https://wordpress.github.io/gutenberg/?path=/story/components-comboboxcontrol--default) and this is how it works there too.

I still feel this behavior is a bit counter-intuitive. I would open an issue in Gutenberg and see if they agree with us and would accept PRs. If yes, maybe we can add a task to contribute a PR upstream to our next cooldown?

---

I left one comment below about an `onChange()` which runs on every render, but besides that, pre-approving.

mikejolley:
Logged an issue here: https://github.com/WordPress/gutenberg/issues/33920

mikejolley:
@nerrad @senadir @Aljullu I've added something called [Patch Package](https://www.npmjs.com/package/patch-package) which allows us to patch one of our dependencies and have the patch applied after NPM install. I've used this to [apply the fix from my upstream PR](https://github.com/WordPress/gutenberg/pull/33928) so something is in place until that is updated and we bump the version we're using.

Are you happy with this solution?

nerrad:
> Are you happy with this solution?

Nifty tool, first I learned of it! The only question I have is whether this particular usability improvement of combobox warrants the extra complexity to surface it now for all users, or if it's sufficient to just have the behaviour show up for sites running WordPress 5.9 (which is presumably when the fix will show up in WordPress core - sometime in December?) or if they have GB active?

mikejolley:
@nerrad By users do you mean us developers? This particular package is one of our direct imports (`wordpress-components`) which get included in the build. End-users won't notice a difference.

nerrad:
No sorry, I mean users of whatever UI implements the component.

nerrad:
We've chatted about my question in DM, I'm just leaving my feedback in here for reference. I was a bit concerned about the additional complexity for addressing a usability issue (a user typing something in the field powered by combobox and pressing enter and it not selecting the first matching item) given:

- Mike has a fix that _is_ approved and being merged in the relevant package in Gutenberg.
- The combobox is being used directly from the `@wordpress/components` package so presumably we'll be able to get the fix on the next package update of WP components (at first I didn't realize it's directly imported which shortens the window even more for users seeing the fix than what I originally thought).
- I personally wasn't sure usability issue was significant.

However, in conversing with Mike, it appears that the third point is more significant than I thought so I'm happy to defer to the choice to apply the patch until its no longer necessary.

How will we make sure we remove the patch (and package) when it's no longer necessary?

I haven't reviewed the actual code or tested this so I can't give a green check - I mostly just came into this PR to answer the question directly asked of me. But if it's good to go and tested, there's nothing blocking from me.

Aljullu (approved):
I tested it again on desktop and I love how it works now. It makes filling addresses much faster. :heart_eyes: 

Good job @mikejolley! :shipit: 

> How will we make sure we remove the patch (and package) when it's no longer necessary?

Should we open an issue (or add a `@todo` comment) so we don't forget about this?

mikejolley:
Todo added! 👍🏻

assets/js/base/components/combobox/index.tsx
```diff
@@ -0,0 +1,150 @@
...
+			ref={ controlRef }
+		>
+			<ComboboxControl
+				className={ 'wc-block-components-combobox-control' }
+				label={ label }
+				onChange={ onChange }
+				onFilterValueChange={ ( filterValue: string ) => {
+					if ( filterValue.length ) {
+						// If we have a value and the combobox is not focussed, this could be from browser autofill.
+						const activeElement = isObject( controlRef.current )
```
senadir, line 113: why was `isObject` used here? a dom element would never evaluate to true here.

PR #4370, opened 2021-06-18
Storybook to gh-pages deployment

mikejolley:
Adds an NPM script to deploy a dist folder to `gh-pages` which can be used for our Storybook deploys.

Running this will build storybook and update the `gh-pages` branch:

```
npm run storybook:deploy
```

After doing so, the [GitHub pages site will be updated](https://woocommerce.github.io/woocommerce-gutenberg-products-block/) as well as the [`gh-pages` branch](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/gh-pages). I just ran this to test so that is now up to date.

So this is not forgotten, I've added this is a post-deploy script, so every time we do a new release, this will update automatically.


opr (approved):
Nice, this looks good to me, thanks Mike 👏🏼 

Just wondering if we should update `README.md` to have a link to the Storybook?

mikejolley:
@opr 👍🏻  Ill edit it separately.

PR #4377, opened 2021-06-21
Add documentation for `ExperimentalDiscountsMeta`, `cart/extensions` endpoint, and `extensionCartUpdate`

opr:
This PR adds documentation for the `ExperimentalDiscountsMeta` Slot/Fill, the `extensionCartUpdate` JS function, and the `cart/extensions` endpoint, and the `register_update_callback` method on the `ExtendRestApi` class.

1. From a third-party developer's point of view, please read the documentation and see if it makes sense.
2. Check the links added to the documentation work correctly.

### Changelog

> Add documentation for `extensionCartUpdate` method - this allows extensions to update the client-side cart after it has been modified on the server.

### DIFF
--- a/docs/extensibility/available-slot-fills.md
+++ b/docs/extensibility/available-slot-fills.md
@@ -40,7 +40,7 @@ Checkout:
 - `components`: an object containing components you can use to render your own shipping rates, it contains `ShippingRatesControlPackage`.
 
 ## ExperimentalDiscountsMeta
-This slot renders below the `CouponCode` input 
+This slot renders below the `CouponCode` input. 
 
 Cart:
 <img alt="Cart showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122774218-ea27a880-d2a0-11eb-9450-11f119567f26.png" />
--- a/docs/extensibility/available-slot-fills.md
+++ b/docs/extensibility/available-slot-fills.md
@@ -46,7 +46,7 @@ Cart:
 <img alt="Cart showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122774218-ea27a880-d2a0-11eb-9450-11f119567f26.png" />
 
 Checkout:
-<img alt="Checkout showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122774626-425eaa80-d2a1-11eb-8f50-6363eaed09f8.png" />
+<img alt="Checkout showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122779277-a08d8c80-d2a5-11eb-9de9-a5f14170058f.png" />
 
 ### Passed paramters
 - `cart`: `wc/store/cart` data but in `camelCase` instead of `snake_case`. [Object breakdown.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c00da597efe4c16fcf5481c213d8052ec5df3766/assets/js/type-defs/cart.ts#L172-L188)
--- a/docs/extensibility/available-slot-fills.md
+++ b/docs/extensibility/available-slot-fills.md
@@ -46,7 +46,7 @@ Cart:
 <img alt="Cart showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122774218-ea27a880-d2a0-11eb-9450-11f119567f26.png" />
 
 Checkout:
-<img alt="Checkout showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122779277-a08d8c80-d2a5-11eb-9de9-a5f14170058f.png" />
+<img alt="Checkout showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122779606-efd3bd00-d2a5-11eb-8c84-6525eca5d704.png" />
 
 ### Passed paramters
 - `cart`: `wc/store/cart` data but in `camelCase` instead of `snake_case`. [Object breakdown.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c00da597efe4c16fcf5481c213d8052ec5df3766/assets/js/type-defs/cart.ts#L172-L188)
--- a/docs/extensibility/available-slot-fills.md
+++ b/docs/extensibility/available-slot-fills.md
@@ -38,3 +38,16 @@ Checkout:
 - `cart`: `wc/store/cart` data but in `camelCase` instead of `snake_case`. [Object breakdown.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c00da597efe4c16fcf5481c213d8052ec5df3766/assets/js/type-defs/cart.ts#L172-L188)
 - `extensions`: external data registered by third-party developers using `ExtendRestAPI`, if you used `ExtendRestAPI` on `wc/store/cart` you would find your data under your namespace here.
 - `components`: an object containing components you can use to render your own shipping rates, it contains `ShippingRatesControlPackage`.
+
+## ExperimentalDiscountsMeta
+This slot renders below the `CouponCode` input 
+
+Cart:
+<img alt="Cart showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122774218-ea27a880-d2a0-11eb-9450-11f119567f26.png" />
+
+Checkout:
+<img alt="Checkout showing ExperimentalDiscountsMeta location" src="https://user-images.githubusercontent.com/5656702/122774626-425eaa80-d2a1-11eb-8f50-6363eaed09f8.png" />
+
+### Passed paramters
+- `cart`: `wc/store/cart` data but in `camelCase` instead of `snake_case`. [Object breakdown.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c00da597efe4c16fcf5481c213d8052ec5df3766/assets/js/type-defs/cart.ts#L172-L188)
+- `extensions`: external data registered by third-party developers using `ExtendRestAPI`, if you used `ExtendRestAPI` on `wc/store/cart` you would find your data under your namespace here.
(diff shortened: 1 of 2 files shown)


docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,90 @@
+# Updating the cart with the Store API
```
senadir, line 1: To keep consistency with the rest of our codebase, can we frame this doc as:
- Problem: What problem are we seeing here, this should be fairly easy to do since we only create new points After an existing problem.
- Solution: Describes the solution, can reference other documents.
- Basic Usage: The most minimal solution that can do the job.
- Things to consider: Any gotchas or things that people might need to know.
- API Definition: Full API docs in a table, see other documents.
- Putting it all together: A full example touching on all options that can be copy-pasted.

docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,90 @@
...
+If you're an extension developer, and your extension does some server-side processing as a result of some client-side
+input, then you may want to use the `cart/extensions` endpoint to signal to
+the Store API that some processing needs to take place.
```
senadir: I'm struggling to understand this part. As a plugin developer, I fail to understand where my code would fall into this, and what would be Cart/Checkout blocks responsibility.

docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,90 @@
...
+## Example use case
+You are the author of an extension that automatically applies valid coupons to a user's cart when they press a
+button.
+
+Your extension adds a button to the sidebar in the Cart and Checkout blocks using the [`DiscountsMeta`](./available-slot-fills.md) Slot.
+
+When this button is clicked, you want the server to do some work to figure out which coupons the customer is eligible
+for, and then apply them to the cart.
+
+When this is done, you expect to see the updated cart data in the client.
```
senadir: This fits very well in the intro of this document and severs as a good introduction to what problem are you solving here. After an initial read, developers can just skip that section straight into basic usage or API definition.

docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,90 @@
...
+ // You shouldn't initiate your own ExtendRestApi instance using `new ExtendRestApi` but should always use the shared instance from the Package dependency injection container.
+ $extend = Package::container()->get( ExtendRestApi::class );
+ $extend->register_update_callback(
+   [
+    'namespace' => 'my-extensions-unique-namespace',
+    'callback'  => /* Add your callable here */
+   ]
+ )
+} );
+```
```
senadir: This would fit very well in the basic usage section as a sub item.

docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,90 @@
...
+
+The method takes a single argument, an associative array.
+
+The associative array must have the following entries:
+- `namespace` (string) - The unique namespace of your extension.
+- `callback` (Callable) - The function/method (or Callable) that will be executed when the `cart/extensions` endpoint is
+  hit with a `namespace` that matches your extension's. The callable should take a single argument. The data passed into
+  the callback via this argument will be an array containing whatever data you choose to pass to it.
+More information on how data gets passed from the client to the callback will be detailed over the next sections.
```
senadir: This could probably live in the API definition part.

docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,90 @@
...
+
+The callable does not need to return anything, if it does, then its return value will not be used.
+
+## Making the request to the `cart/extensions` endpoint
+When you wish to make a request to the endpoint, you should do it by calling the `extensionCartUpdate` function, available
+from `@woocommerce/blocks-checkout`.
```
senadir: This, with the JS example below, can be moved into the basic usage section with the argument breakdown into API definition.

docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,90 @@
...
+When this function is executed, after a short delay (due to batch requests) a POST request will be made to the
+`cart/extensions` endpoint, the POSTed data will be the value of the `data` key in the argument object.
+
+When the request completes, the API will return the current server-side cart to the client. The client will then update
+the UI.
+
+## Why can't I just POST to `cart/extensions` directly?
+This endpoint returns an object representing the customer's cart, WooCommerce Blocks will then update the client-side
+stores with the new cart data. The methods to do this are not available to third party extensions, which is why you must
+use the `extensionCartUpdate` method.
```
senadir: This can be part of the "Problem" section.
It also doesn't do a good job of explaining why you can't update Cart block yourself, which is to protect the block from getting broken by a malfunctioning plugin.

senadir (changes requested):
Thank you for writing this PR Thomas! it's super valuable to have those docs this early.

However, I strongly suggest we align our documentation to a single template. This is something that Darren, Allen, and I discussed and iterated on to make our docs more approachable/structured to WordPress developers who are new to the JS world and those type of extensibility points on general.

docs/extensibility/extend-rest-api-update-cart.md
opr: Thanks Nadir, I've updated the document now to follow this format. I feel like there's going to be way too much to cover in the `Putting it all together` section, and a lot of irrelevant information not relating to this part of the API. I did not want to go into too much detail about the things not directly related to this, so I've omitted some parts that are covered in other documents, and also not gone into detail about the implementation of my example methods.

I also feel like there's quite a bit of overlap and duplication of the `Basic usage` and `Putting it all together` section, but this is probably OK.

docs/extensibility/extend-rest-api-update-cart.md
opr: I think the new format may help to illustrate the problem a bit better. What do you think?

docs/extensibility/extend-rest-api-update-cart.md
opr: Fixed by the new format, I think.

docs/extensibility/extend-rest-api-update-cart.md
opr: Fixed by the new format.

docs/extensibility/extend-rest-api-update-cart.md
opr: Fixed by the new format.

docs/extensibility/extend-rest-api-update-cart.md
opr: Fixed by the new format.

docs/extensibility/extend-rest-api-update-cart.md
opr: Fixed by the new format, and also more detail about _why_ they can't update it has been added to the Problem section, and also mentioned again in the Things to consider section.

docs/extensibility/available-slot-fills.md
```diff
@@ -38,3 +38,16 @@ Checkout:
 - `cart`: `wc/store/cart` data but in `camelCase` instead of `snake_case`. [Object breakdown.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c00da597efe4c16fcf5481c213d8052ec5df3766/assets/js/type-defs/cart.ts#L172-L188)
 - `extensions`: external data registered by third-party developers using `ExtendRestAPI`, if you used `ExtendRestAPI` on `wc/store/cart` you would find your data under your namespace here.
 - `components`: an object containing components you can use to render your own shipping rates, it contains `ShippingRatesControlPackage`.
+
+## ExperimentalDiscountsMeta
+This slot renders below the `CouponCode` input
```
nielslange: **nitpick**: I'd finish the sentence using a full stop.

docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,208 @@
...
+| `data` | `Object` | No | The data you want to pass to your callback. Anything in the `data` key will be passed as the first (and only) argument to your callback as an associative array.
+
+## Putting it all together
+You are the author of an extension that lets the shopper redeem points that they earn on your website for a discount on
+their order. There is a text field where the shopper can enter how many points they want to redeem, and a submit button 
+that will apply the redemption.
+
+Your extension adds these UI elements to the sidebar in the Cart and Checkout blocks using the [`DiscountsMeta`](./available-slot-fills.md) Slot.
+
+More information on how to use Slots is available in our [Slots and Fills documentation](./slot-fills.md), and thus we
```
nielslange: **nitpick** Instead of

> More information on how to use Slots is available in our [Slots and Fills documentation](./slot-fills.md), and thus we will not be demonstrating how to use them in _this_ document.

I'd write

> More information on how to use Slots is available in our [Slots and Fills documentation](./slot-fills.md).

as adding the second part feels a bit passive-aggressive to me.

nielslange (commented):
Except the two minor suggestions, your changes look good to me. The two suggestions are nitpicking, and you don't necessarily have to apply them. 😉

docs/extensibility/extend-rest-api-update-cart.md
```diff
@@ -0,0 +1,207 @@
...
+    'namespace' => 'extension-unique-namespace',
+    'callback'  => /* Add your callable here */
+   ]
+ );
+} );
+```
+
+and on the client side:
+```typescript
+import { extensionCartUpdate } from '@woocommerce/blocks-checkout';
```
senadir: I think we're currently sticking to using ES5 syntax only. Example https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/3829f6bb5f1ea904553dd069b7688b78b23d8c1c/docs/extensibility/slot-fills.md#L22-L40

This means that this line can be written like this
```suggestion
const { extensionCartUpdate } = wc.blocksCheckout;
```

This allows it be used without a build process.
It's okay to not do that on `Putting it all together` section.

senadir (approved):
Thank you for handling updates and feedback on this PR Thomas. I think we can ship it 🚢

PR #4382, opened 2021-06-22
Add camelcase rule to eslint config

### LINKED ISSUE #4150: Find places where snake_cased names are received from the API and convert them to camelCase

opr:
This PR, inspired by @louwie17's p2 paJDYF-1ZD-p2 will add a configuration to our `.eslintrc.js` file changing the rules about when camelCase must be used. This is a follow on from #4340 where we discovered that changing values from the API to snake_case immediately caused problems with third-party extensions.

You will notice I have added some regex configuration to the `@typescript-eslint/naming-convention` rules, this targets the following variables:

- `Block` - found here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/40a8a72e276ed9f6ccd0050bc970d869a05bb7bf/assets/js/atomic/utils/render-parent-block.tsx#L24 
- `__webpack_public_path__` - found here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7bb3b7ba47e4a5b98bc551f4f23b831c360e358a/assets/js/atomic/blocks/component-init.js#L10
- `API_FETCH_WITH_HEADERS` - found here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/5261c4e9a733d5292ab40027a6ad6b3d61952e73/assets/js/data/shared-controls.ts#L123

I think we could possibly address the `Block` and `API_FETCH_WITH_HEADERS` ones in a follow-up PR.

Fixes #4150

1. Run `npm run lint:js` on the command line and ensure no lint errors are found.
2. Test the "All products" block, ensure pagination works correctly.
3. Test the "Products by attribute" block.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -59,6 +59,13 @@ module.exports = {
 		'react-hooks/exhaustive-deps': 'error',
 		'react/jsx-fragments': [ 'error', 'syntax' ],
 		'@wordpress/no-global-active-element': 'warn',
+		camelcase: [
+			'error',
+			{
+				properties: 'never',
+				ignoreGlobals: true,
+			},
+		],
 	},
 	overrides: [
 		{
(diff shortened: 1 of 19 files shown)


senadir (approved):
I tested this a couple of days ago but forget to approve.

PR #4392, opened 2021-06-23
Handle translations in lazy loaded files

### LINKED ISSUE #4019: "Sale" Not Being Translated on the All Products Block
## Describe the bug
When using the All Products block, "Sale" is translated while editing a page. However, on the front end, it's not.

## To reproduce

1. On WP Admin > Settings, set the Site Language to a language other than English like PT BR - Português do Brasil
2. Make sure to download any translation updates under My Home > Updates
3. Create a page, add the All Products block to it, and save it
4. View the page from the front end

## Expected behavior
"Sale" should be translated on the front end.

## Screenshots
Translated “Sale” on the All Products Block while Editing the Page:
Translated Sale on the All Products Block while Editing the Page

Untranslated “Sale” on the All Products Block from the Front End:
Untranslated Sale on the All Products Block from the Front End

Translated “Sale” on the Hand-picked Products Block from the Front End:
Translated Sale on the Hand-picked Products Block from the Front End

Translated “Sale” on the On Sale Products Block from the Front End:
Translated Sale on the On Sale Products Block from the Front End

### WordPress:
* Core version: 5.7
* Gutenberg version: 10.2.1
* WooCommerce version: 5.1.0
* WooCommerce Blocks version: 4.4.3
* Site language: en_US
  * Akismet Anti-Spam: by Automattic – 4.1.9
  * AMP: by AMP Project Contributors – 2.0.11
  * CoBlocks: by GoDaddy – 2.9.1
  * Crowdsignal Forms: by Automattic – 1.4.5
  * WordPress.com Editing Toolkit: by Automattic – 3.3396
  * Jetpack by WordPress.com: by Automattic – 9.5.2
  * Layout Grid: by Automattic – 1.5
  * Page Optimize: by Automattic – 0.5.1

### Desktop:
* OS: macOS
* Browser: Chrome

## Additional context
* This was reported in #3814871-zen on a WordPress.com site
* Tested on a WordPress.com site too
* [There aren't any untranslated "Sale" terms for PT BR](https://translate.wordpress.org/projects/wp-plugins/woo-gutenberg-products-block/stable/pt-br/default/?filters%5Bterm%5D=sale&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D&sort%5Bby%5D=original&sort%5Bhow%5D=desc&sorts=Apply+Sorting&page=2)
* Also tested with ES - Español ([100% translated](https://translate.wordpress.org/projects/wp-plugins/woo-gutenberg-products-block/stable/es/default/?filters%5Bterm%5D=sale&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D&sort%5Bby%5D=original&sort%5Bhow%5D=desc&sorts=Apply+Sorting&page=2)):
ES

* And JA - 日本語 ([98% translated](https://translate.wordpress.org/projects/wp-plugins/woo-gutenberg-products-block/stable/ja/default/?filters%5Bterm%5D=sale&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D&sort%5Bby%5D=original&sort%5Bhow%5D=desc&sorts=Apply+Sorting&page=2))
JA

### LINKED ISSUE #4420: "In Cart" Not Being Translated on the All Products Block
## Describe the bug
When using the All Products block, "add to cart" is translated. However, "# in cart" is not.

## To reproduce

1. On WP Admin > Settings, set the Site Language to a language other than English like Español or 日本語
2. Make sure to download any translation updates under My Home > Updates
3. Create a page, add the All Products block to it, and save it
4. View the page and add a product to cart

## Expected behavior
"# in Cart" should be translated.

## Screenshots
GHSpanish
GHJapanese

### WordPress:
* Core version: 5.7.2
* WooCommerce version: 5.4.1
* WooCommerce Blocks version: 5.4.0
* Site language: es_ES
 * Companion Plugin: by Osk – 1.18

* OS: macOS
* Browser: Chrome Version 91.0.4472.114 (Official Build) (x86_64)

## Additional context
* Reported in [this WordPress.org forum thread](https://wordpress.org/support/topic/spanish-translation-error-add-to-cart-button-x-items-in-cart/)
* May be related to this report: ["Sale" Not Being Translated on the All Products Block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4019)?
* Test site demonstrating the issue: https://solitary-pheasant.jurassic.ninja/test-page/

mikejolley:
Adds support for loading translations for lazy-loaded components.

`RequireChunkCallbackPlugin` is borrowed from Calypso.

On the blocks side, we use `addRequireChunkTranslationsHandler` to provide a list of Chunk IDs containing translations. `getTranslationChunkFileUrl` is a callback that maps the Chunk ID to a JSON filename in the WP_LANG dir. This consists of the textdomain, locale, and a hash for the script file itself.

When a file is imported by React, `addRequireChunkTranslationsHandler` will run and fetch the JSON file. If successful, we simply pass the body to `setLocaleData` so the translation is then performed by the `@wordpress/i18n` package.

Fixes #4019
Closes #4392

**Testing**

**Testing translations:**

1. Add the all products block to a page and save.
2. Ensure a product is on sale that's visible in this block.
3. Switch to a non US language with translations e.g. Brazillian Portuguese. 
4. View the page on the frontend with the all products block. Ensure "sale" is translated.

**Testing builds:**

Also, test builds (dev and production) because this required an update to Webpack 5x.

**Core plugin**

TBC - We need a way to test under core context because the domain and paths are different. I've added handling for this but it's not tested.

**Changelog**

> Fixed string translations within the All Products Block.


assets/js/base/utils/i18n.ts
```diff
@@ -0,0 +1,94 @@
...
+	options: {
+		// Callback function to get the file name from the ChunkId.
+		getTranslationChunkFileUrl?: ( chunkId: string ) => string;
+		// Text domain.
+		domain?: string;
+		// Array of chunk ids that should have available translations.
+		translatedChunks?: string[];
+	} = {}
+): void => {
+	if ( typeof window.__requireChunkCallback__ === undefined ) {
```
jsnajdr: The result of `typeof` is a string, like `"undefined"`. Comparing to the `undefined` value will always yield `false`.

bin/webpack-configs.js
```diff
@@ -95,10 +97,6 @@ const getCoreConfig = ( options = {} ) => {
 			path: path.resolve( __dirname, '../build/' ),
 			library: [ 'wc', '[name]' ],
 			libraryTarget: 'this',
-			// This fixes an issue with multiple webpack projects using chunking
-			// overwriting each other's chunk loader function.
-			// See https://webpack.js.org/configuration/output/#outputjsonpfunction
-			jsonpFunction: 'webpackWcBlocksJsonp',
```
jsnajdr, line 75: What has changed so that the JSONP functions from different blocks no longer conflict and don't need to be assigned different names?

jsnajdr (commented):
The `RequireChunkCallbackPlugin` has one flaw that prevents it from working correctly when multiple blocks that use it (i.e., multiple independent webpack compilations) are loaded into one HTML document: there is only one global `window.__requireChunkCallback__` object and different blocks will overwrite each other's data.

One solution is to not assign `__requireChunkCallback__` to `window`, but to make it into a special webpack-interpreted global variable. Webpack has many such [module variables](https://webpack.js.org/api/module-variables/), like `__webpack_public_path__`.

Here is the webpack code that finds these variables in the AST and outputs code to interpret them specially: https://github.com/webpack/webpack/blob/master/lib/APIPlugin.js

@yuliyan This could be a valuable improvement of the plugin. And if it gets popular, we could move it to the Gutenberg repo and make it a standard part of the block building workflow.

assets/js/atomic/blocks/component-init.ts
```diff
@@ -0,0 +1,171 @@
...
+		domain,
+		getTranslationChunkFileUrl: ( chunkId ) => {
+			const hash = md5(
+				isFeaturePluginBuild
+					? `build/${ chunkId }.js`
+					: `packages/woocommerce-blocks/build/${ chunkId }.js`
+			);
+			return `${ blocksConfig.languageUrl }/${ domain }-${ blocksConfig.locale }-${ hash }.json`;
+		},
+		translatedChunks: [
```
yuliyan: I'm not sure if having the list of translation chunks hardcoded is the best approach. For languages with incomplete translations, some of the language files for the listed chunks might be missing, which would result in 404 for the translation request.

Would it be possible to generate the translated chunks list from the back-end and provide it to the app using the `blocksConfig` object, e.g.:

```php
$translated_chunks = array_map(
	'basename',
	glob( WP_LANG_DIR . '/plugins/woo-gutenberg-products-block-' . get_locale() . '*.json' )
);
```

senadir: We still have to do some cross-checking to know which file to load I guess right? which means we still need to generate the hash on the frontend again?  `$translated_chunks` returns 227 files for the Japanse translation of WooCommerce blocks.

senadir: I applied this feedback @yuliyan if you want to take another look.

I'm not fully convinced of this approach. Passing the full list of files being translated adds 16kb to the page size (227) while we might need only 4 or 5. So on a later PR I might just pass down atomic blocks (the ones being dynamically loaded).

bin/webpack-configs.js
senadir: `jsonpFunction` is no longer an option in webpack 5, it's now replaced with `chunkLoadingGlobal`.
However, I couldn't find any reference for this `webpackWcBlocksJsonp`. I'm not sure what we're trying to solve here.
The [original PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/903/files) that introduced this (2 years ago) provides very little rationale about what was happening here.
Deleting it doesn't seem to break anything in my testing.

senadir:
@jsnajdr I wasn't able to reliably add a module variable, the webpack page doesn't provide any meaningful info about adding. I tried both `ProvidePlugin` and `DefinePlugin` but they both didn't work for my case.
I ended up just using a locally scoped variable with `let`.
Because we don't yet have a case in which multiple blocks inserted on the same page dynamically load JS, I couldn't test my solution. Inserting the same block twice with different inner block templates worked fine in my case.

assets/js/atomic/blocks/component-init.ts
yuliyan: I didn't expect the files to be that many, but yeah, in this case it doesn't make a lot of sense to provide all that data to the front-end and increase the page size unnecessarily.

As you suggested, it would be better if you could pass only the files that are relevant to the front, maybe something like that could do the job:

```php
$files_with_translations = array_map(
	'md5',
	array(
		'build/atomic-block-components/add-to-cart.js',
		'build/atomic-block-components/button-frontend.js',
		'build/atomic-block-components/rating.js',
		'build/atomic-block-components/title.js',
		'build/atomic-block-components/tag-list.js',
		...
	) // hardcode the atomic block filepaths, or get them dynamically with another glob.
);
```

And then the glob pattern would be: 
```php
glob( WP_LANG_DIR . '/plugins/woo-gutenberg-products-block-' . get_locale() . "-{{$files_with_translations}}" . '*.json', GLOB_BRACE );
```

mikejolley:
@senadir @opr I updated how we pass in supported files. It seems to be working well now. Can you take a final look at this, please?

mikejolley:
@senadir I've tried another approach to this in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4897

Issue #4405, opened 2021-06-28
In a WooCommerce core environment blocks with wc-blocks-checkout dependency don't work


ralucaStan:
## Describe the bug

This issue appeared while investigating https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4391.

 It was [discovered](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4391#issuecomment-868288126) that after the changes from  #4222, where we moved Label to `@woocommerce/blocks-checkout` package, all blocks which were using that component had `wc-blocks-checkout` as a dependency.
In a WooCommerce core environment ( where the feature plugin is not present), `wc-blocks-checkout` is not registered (because the Cart and Checkout blocks are available through the feature-plugin only) so the blocks dependent on it were broken ("All Products", "Filters" and "Reviews").

While reverting #4222  the issue, it surfaced that the All Products block still had a `wc-blocks-checkout` dependency, because it included the StoreNoticesProvider. The provider pulls in the SnackbarNoticesContainer, which includes the `__experimentalApplyCheckoutFilter`, which is a filter from `@woocommerce/blocks-checkout` package.

## Expected behavior

We need to be able to separate Cart & Checkout-related logic from the blocks that are included in WC Core.

ralucaStan:
What we need to discuss is:
1. how do we handle exposing the components that were moved as part of #4222 Label and Button components:
    Ideas proposed: 
    - duplicate components (not ideal, please note that `ValidatedTextInput` component is already duplicated, so we might need including it in the final solution)
    - create a separate package for WC Blocks components (we'd have to consider the overlap with the existing @woocommerce/components library and what that means for how we name or where these components get exposed.)
    
2. how to write the StoreNoticesProvider so that it does not expose C&C functionality to other components using the notices store
    - All products block uses the basic notice, while the C&C filter is used for the snack bar notices. The StoreNoticesContext  is a wrapper for both type of notices.

PR #4222, opened 2021-05-17
Move Button and Label components to `@woocommerce/blocks-checkout` package

opr:
This PR is part of a series of smaller PRs that relocate certain components to the Checkout package. I will do 1-2 components per PR to ensure the diffs are manageable.

This PR contains the changes for the `Label` and `Button` components.

### How to test the changes in this Pull Request:

1. Smoke test of Blocks functionality.
2. Ensure critical flows work, such as adding items to cart from the All Products Block, continuing to Checkout from the Cart Block and placing an order in the Checkout Block.
2. Ensure labels are displaying correctly in blocks, compare to trunk and ensure nothing is different or missing.

### Changelog

> Move Button and Label components to `@woocommerce/blocks-checkout` package

PR #4406, opened 2021-06-28
Export `Button` and `Label` from `wc-blocks-checkout` package

opr:
This PR will revert #4222, but at the same time will make the `Button` and `Label` components available in the `wc-blocks-checkout` package.

By leaving `Button` and `Label` where they are and exporting them from the checkout package, we can allow existing blocks to import from `@woocommerce/base-components` components while extensions can use `@woocommerce/blocks-checkout`.

Partly addresses #4405

1. Check out 430-gh-woocommerce/woocommerce-points-and-rewards and run `npm run build`. Ensure no build errors happen.
2. Ensure the Points and Rewards plugin is activated.
3. Set up your user account so it has points available. (WooCommerce -> Points and Rewards)
3. Add items to your cart.
4. Go to the Cart/Checkout blocks and ensure the Points and Rewards sidebar is showing properly, it should look _similar_ to this:


Aljullu (approved):
LGTM

PR #4407, opened 2021-06-28
Create `StoreNoticesWithSnackbarProvider`

### LINKED ISSUE #4405: In a WooCommerce core environment blocks with wc-blocks-checkout dependency don't work

opr:
This PR will create a new component `StoreNoticesWithSnackbarProvider` which is based on `StoreNoticesProvider`. The snackbar related code is removed from `StoreNoticesProvider`.

Any components that need to manipulate store notices, but not the snackbar (All Products) should be using `StoreNoticesProvider` and any that do need to manipulate the snackbar should use `StoreNoticesWithSnackbarProvider`.

This will resolve #4405 as long as #4406 is merged too.

1. Run `npm run build:deploy` - inspect `build/all-products.asset.php` and verify `wc-blocks-checkout` is not there.
2. Go to the All products block, find a product that you want to add to the cart, don't add it yet!
3. In a new tab, edit this product so it is out of stock.
4. Go back to the All products block, and try to add the product to your cart.
5. Check you get a store notice at the top of the page/block.
6. Add some in-stock products to the cart.
7. Ensure you've got a useable coupon on your store.
8. Go to the Cart/Checkout block, add a coupon.
9. Ensure the snackbar shows.


assets/js/base/context/hooks/test/use-store-notices-with-snackbar.js
```diff
@@ -0,0 +1,75 @@
+/**
+ * External dependencies
+ */
+import { render, act } from '@testing-library/react';
+import { StoreNoticesWithSnackbarProvider } from '@woocommerce/base-context/providers';
```
ralucaStan: I have one small recommendation, namely to rename to `StoreSnackbarNoticesProvider`. You already have `SnackbarNoticesContainer`, so it would be consistent to refer to them as `SnackbarNotices`. I'm thinking if another type of notification appears we could keep the pattern

- NoticesProvider
- SnackbarNotices
- XNotices

assets/js/base/context/providers/store-notices-with-snackbar/context.js
```diff
@@ -0,0 +1,152 @@
...
+	className = '',
+	createNoticeContainer = true,
+	context = 'wc/core',
+} ) => {
+	const { createNotice, removeNotice } = useDispatch( 'core/notices' );
+	const [ isSuppressed, setIsSuppressed ] = useState( false );
+	const { dispatchStoreEvent } = useStoreEvents();
+	const { isEditor } = useEditorContext();
+
+	const createNoticeWithContext = useCallback(
```
ralucaStan: I see now why you had StoreSnackbarNoticesProvider. I think this separate Provider is in the right direction, but by keeping the default notifications part we end up having duplicated logic, which is not ideal. I am thinking to separate the implementations, have a provider just for default notices and one for the snackbar notices. This way there would be no confusion on what we should use, we would not have duplicated logic and the implementations would be decoupled; tests would just focus on the specific type of notification (currently the test for this also tests default notifications, which is already tested in the intial provider).

ralucaStan (changes requested):

assets/js/base/context/providers/store-notices-with-snackbar/context.js
ralucaStan: I should have submitted this as a comment, not as a "request changes".

Aljullu (commented):
This is testing great and confirming it removes the `wc-blocks-checkout` dependency from All Products.

I have one question about the approach taken, though. In the way this PR is created, `StoreNoticesWithSnackbarProvider` is a wrapper around `StoreNoticesProvider`, right? Why not making `StoreNoticesWithSnackbarProvider` to be completely independent of `StoreNoticesProvider`? Would that be possible? The benefit I see is that `StoreNoticesWithSnackbarProvider` would be smaller and if we ever have a block or component that only needs snackbar notices but not regular notices, they will be able to use that provider without the need to import `StoreNoticesProvider`. I think that would also work well with the renaming @ralucaStan proposed: we would have different providers for each kind of notice.

opr:
Thanks @Aljullu and @ralucaStan for your reviews! 🙌🏼 

Could you please be a little clearer about what you mean when mentioning that this is a wrapper around `StoreNoticesProvider`? The intention was for this to be a standalone component that _also_ fills the role of `StoreNoticesProvider`. The reason it needs to also fill the role of `StoreNoticesProvider` is because in the Cart and Checkout contexts, sometimes both snackbar and regular notices need to be shown. If you try and nest them it's not possible because nesting context providers doesn't work. (I will try this again and get the exact error message, I can't remember it off the top of my head)

Just to clarify, the way snackbar notices work is very similar to regular store notices, except they have a `snackbar` type, so a lot of the logic for adding/removing is shared. I will work on addressing the feedback from @ralucaStan to see if I can extract some of the duplicated logic to a mutually convenient place for both of these providers, as well as the rename.

Aljullu:
> Could you please be a little clearer about what you mean when mentioning that this is a wrapper around `StoreNoticesProvider`?

Sure! So my understanding is that when a block wants to make use of snackbar notices, the context & logic for managing regular notices is imported as well because `StoreSnackbarNoticesProvider` depends on `StoreNoticesProvider`. And example of that is that `useStoreNoticesWithSnackbar()` is exporting methods not related to snackbar notices (ie: `addDefaultNotice`, `addErrorNotice`, [etc.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4407/files#diff-c1fe5e372a9bfac5e83846e21fe1aebd6f889e2437653a386a7ac544706d79bcR45-R64)). My question is: is that required for some reason or we could make snackbar notices independent of regular notices? So they would have different contexts and different hooks, and none of them would depend on the other.

How that will affect blocks?
* All Products would be subscribing to StoreNoticesContext (via `useStoreNotices()`) like it currently does.
* Cart and Checkout would be subscribing to StoreNoticesContext and StoreSnackbarNoticesContext (via  `useStoreNotices()` and `useStoreSnackbarNotices()`) because they use both types of notices.
* If in the future we have a block which only uses snackbar notices, it would only need to subscribe to StoreSnackbarNoticesContext.

ralucaStan:
My suggestion had the same reasoning as @Aljullu's above comment: use the corresponding context provider for each notification type. Stacking context providers shouldn't be an issue, @opr let me know if you see that error again and want another 👀

opr:
> * Cart and Checkout would be subscribing to StoreNoticesContext and StoreSnackbarNoticesContext (via  `useStoreNotices()` and `useStoreSnackbarNotices()`) because they use both types of notices.

I am not sure why this didn't work the first time, I got an error message about not being able to nest contexts, but it is working ok now.

I have made it so the new `StoreSnackbarNoticesProvider` can stand alone as a provider, updated the tests and removed as much of the duplicated/unnecessary logic as I could.

You can see that the [`Cart`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4407/files#diff-6c8997eee95477d2d3ff6e07c37a5dc7d03f525088204b3b7753b93be792a1b1R32) and [`Checkout`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4407/files#diff-d1233b55f457bf3bdc331fae331c9029571c86a11181c330ae79218301354e39R59) blocks are now wrapped in an additional provider, but [`AllProducts`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f4d01783bce3363769e6b5a2be7ef0a499578b5f/assets/js/blocks/products/all-products/frontend.js#L18) is not.

assets/js/base/context/hooks/use-store-notices-with-snackbar.js
```diff
@@ -0,0 +1,78 @@
+/**
```
ralucaStan: this file can be removed

ralucaStan (approved):
Good work! I've tested it and works as expected. I left just one comment about removing a leftover file

Aljullu (approved):
Tested on my end and it also works great!

PR #4415, opened 2021-06-30
Add `TotalsWrapper` component

### LINKED ISSUE #4267: Border on `Coupon code` Panel in Cart/Checkout is too thick when there is no shipping required
## Describe the bug
When there is no shipping required on the order, the border below `Coupon code` is too thick, it is made up of two borders, one below the coupon panel, and one above the order total.

## To reproduce

1. Add a virtual product to the cart.
2. View the Cart/Checkout block.
3. See the double border.

## Expected behavior
The border should not be doubled up if there is no shipping on the cart.

## Screenshots
image

opr:
This PR adds a component called `TotalsWrapper` whose purpose is to wrap components in the Cart and Checkout sidebar so that components that have borders (the grey lines between items in the screenshot section) don't have to be concerned with what follows them.

For example, if the store has coupons disabled, then you could end up with a component with a bottom border, and one with a top border together resulting in a "double border".

By wrapping these components in a `TotalsWrapper` then the borders can be consistently managed from one place.

The PR will also slightly change the way Slot/Fills are rendered. The `Slot` will be rendered inside a `TotalsWrapper` but each `Fill` will also have a border and padding applied to it (by CSS), to visually separate them.

Here is an overview of the changes made in this PR:

- Add a `TotalsWrapper` component. This has the optional prop `slotWrapper` to allow CSS to render borders on `Fill`s.
- Add `hasValidFills` function in `slot/index.js` - This is to check that at least one of the `Fill`s about to be rendered is truthy (i.e. not null/undefined) or it won't render the `Slot`.
- Change styles on `OrderSummary`, `ShippingRatesControlPackage`, `TotalsCoupon`, `Cart`, and `CheckoutSidebar` to remove borders and padding because `TotalsWrapper` is now handling it.
- Wrap `Slot`s in `<TotalsWrapper>`
- Change the button responsible for expanding the `Panel` component to only have margins/padding when the panel is open.

Fixes #4267 

### Screenshots
| Before | After |
|---|---|
| image | image |

1. Ensure you're running WC Subscriptions 3.1.3 or later and you've got some different subscriptions products available to add to your cart.
2. Check out `feature/add-inputs-for-points-redemption` on [WC Points and Rewards](430-gh-woocommerce/woocommerce-points-and-rewards)
3. run `npm run build` in your WC Points and Rewards directory.
4. Ensure your user has points available to use (WooCommerce > Points and Rewards)
5. Add some products to your cart, some of which should be subscription products.
6. Go to the Cart block. Ensure the sidebar doesn't have double borders, empty gaps, or inconsistent spacing.
7. Change settings relating to tax, try several different combinations and check the Cart/Checkout sidebar each time to make sure the spacing/borders are all consistent.
8. Disable coupons and do the same.
9. Try any other settings you can think of to introduce/hide different sections on these sidebars.

### Changelog

> Wrap components in the Cart and Checkout sidebar in a TotalsWrapper. This will ensure consistent spacing and borders are applied to items in the sidebar.


senadir (approved):
I tested with and without Subscriptions. everything looks fine!

PR #4417, opened 2021-07-01
Update `snackbarNotices` filter to `snackbarNoticeVisibility`

opr:
After internal discussion at pca54o-1Gs-p2 I wanted to put this PR out there to see what folks think, the changes are made to stop third party extensions being able to modify the particulars of any snackbar notice, and control only whether it is shown to the user or not. An overview of the changes is:

- Change the filter so that only an object keyed by the notice's message is passed to the filter - initially all values in this object will be `true`.
- After the filters have done their work, remove any entries from the array whose value is false.
- Filter the original list of notices so that only notices that have a corresponding entry in the one returned by the filter are shown.

1. Check out `update/snackbar-filtering-function` in [WooCommerce Points and Rewards](https://github.com/woocommerce/woocommerce-points-and-rewards/)
2. Run `npm run build` in the WC P&R directory.
3. Ensure your user account has points (WooCoomerce > Points and Rewards)
4. Add items to your cart and go to the Cart block.
5. Try adding a coupon, ensure the snackbar notice shows for coupon addition.
6. Remove the coupon, ensure the snackbar notice shows for coupon removal.
7. Please don't add more than one coupon, we have a known issue with this, https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4312
8. Use the points redemption panel to redeem some points.
9. Remove the coupon that was added as a result of redeeming points.
10. Ensure no snackbar notice shows.
11. In the WooCommerce Points and Rewards file: [`assets/js/filters.js`](https://github.com/woocommerce/woocommerce-points-and-rewards/blob/0c8eb7539c42764176ee2af19050c0cbe9b7f3e3/assets/js/filters.js#L24) change `snackbarNoticeVisibility` to `snackbarNotices` and run `npm run build`.
12. Add a coupon to the cart again and ensure a deprecation notice is shown in the console.

### Changelog

> Deprecate snackbarNotices filter in favour of snackbarNoticeVisibility to allow extensions to hide snackbar notices in the Cart and Checkout blocks.

### DIFF
--- a/assets/js/base/context/providers/store-snackbar-notices/components/snackbar-notices-container.js
+++ b/assets/js/base/context/providers/store-snackbar-notices/components/snackbar-notices-container.js
@@ -19,10 +19,19 @@ const SnackbarNoticesContainer = ( {
 		( notice ) => notice.type === 'snackbar'
 	);
 
+	const noticeVisibility = snackbarNotices.reduce( ( acc, { content } ) => {
+		acc[ content ] = true;
+		return acc;
+	}, {} );
+
 	const filteredNotices = __experimentalApplyCheckoutFilter( {
-		filterName: 'snackbarNotices',
-		defaultValue: snackbarNotices,
+		filterName: 'snackbarNoticeVisibility',
+		defaultValue: noticeVisibility,
 	} );
+
+	const noticesToDisplay = snackbarNotices.filter(
+		( notice ) => filteredNotices[ notice.content ] === true
+	);
 
 	const wrapperClass = classnames(
 		className,
@@ -31,7 +40,7 @@ const SnackbarNoticesContainer = ( {
 
 	return (
 		<SnackbarList
-			notices={ filteredNotices }
+			notices={ noticesToDisplay }
 			className={ wrapperClass }
 			onRemove={ removeNotice }
 		/>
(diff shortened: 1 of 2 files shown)


mikejolley (approved):
This appears to be working as described :) Approving. Rebase is needed.

Was the "multiple coupon" issue you mentioned the block error boundary when removing a coupon when points are applied? 

I noticed a notice which is unrelated to this change but something in points. Are you aware or should I log a new issue?

> Notice: Array to string conversion in /plugins/woocommerce-points-and-rewards/includes/class-wc-points-rewards-extend-store-endpoint.php on line 86

opr:
> This appears to be working as described :) Approving. Rebase is needed.

Cheers, Mike. Done!

> Was the "multiple coupon" issue you mentioned the block error boundary when removing a coupon when points are applied?

Yep, this is fixed now following the rebase.

> I noticed a notice which is unrelated to this change but something in points. Are you aware or should I log a new issue?
> 
> > Notice: Array to string conversion in /plugins/woocommerce-points-and-rewards/includes/class-wc-points-rewards-extend-store-endpoint.php on line 86

Strange, I'm not getting this notice, this line is an array of tags that should be allowed in [`strip_tags`](https://www.php.net/manual/en/function.strip-tags.php), but I think I can change this to a string, which version of PHP are you using?

mikejolley:
> Strange, I'm not getting this notice, this line is an array of tags that should be allowed in strip_tags, but I think I can change this to a string, which version of PHP are you using?

7.3.5

opr:
> > Strange, I'm not getting this notice, this line is an array of tags that should be allowed in strip_tags, but I think I can change this to a string, which version of PHP are you using?
> 
> 7.3.5

Makes sense then since this change to `strip_tags` was made in 7.4.0 - I'll change it to a string in P&R 👍🏼

PR #4418, opened 2021-07-01
Remove source assets from ZIP and releases

### LINKED ISSUE #4299: Remove source files from released code.
This issue is for removing JavaScript and CSS source files from the release of the plugin. 

This will have the advantage of:

- reducing the size of the release build (the `assets` folder is 5MB unzipped) - both what is deployed to WordPress.org and included as a package in WooCommerce core.
- potentially reducing confusion around folks not realizing there is a build process and editing those files.
- eliminating the extra work involved for WP.org glotpress instance to scan for strings in unused files and also the extra source entries.

Some known tradeoffs:

- Translators won't have the ability to fallback on easily readable code for the translated strings (to help with context etc). So we will need to be extra careful about including translator comments etc to help with this.
- Debugging with `SCRIPT_DEBUG` set to `true` won't have useful stack traces on production builds (source maps etc). I think this is already the case anyways because of our build process.

## Rough preliminary task list:

* [x] Currently, all translator comments are stripped by our Webpack build process so they don't show in built files. This was an issue in Gutenberg too that was solved by [this custom Terser plugin](https://github.com/WordPress/gutenberg/pull/22990) which we'll need to implement in our repo (and test).
* [x] Audit the `assets` directory and make sure that no paths are referenced as a part of built files (images etc). Anything that is will need refactored to be moved outside of the assets folder.
* [x] #4408

senadir:
This PR removes the `assets` folder from the ZIP file, thereby removing it from releases. Hopefully this translates well to tagged released, I haven't been able to test this really. 
The `source files` zip in tags would still contain assets, but the `woo-gutenberg-products-block.zip` won't, I'm assuming that's the file that wp.org would use.

This is a non-destructive PR so even if it was included, we can still rectify it.

Fixes #4299

1. Run `npm run package-plugin:zip-only`
2. Verify the zip doesn't contain the assets folder.
3. Use the zip in a website, make sure blocks load.

### Changelog

> Remove source files from releases.

### DIFF
--- a/bin/copy-plugin-files.sh
+++ b/bin/copy-plugin-files.sh
@@ -1,5 +1,6 @@
 cd "$1" || exit
 rsync ./ "$2"/ --recursive --delete --delete-excluded \
+	--exclude=assets/ \
 	--exclude=".*/" \
 	--exclude="*.md" \
 	--exclude=".*" \


opr (approved):
Works fine for me. I uploaded the zip to a JN site and the blocks are working ok. 🚢 it!

PR #4419, opened 2021-07-01
Block Widgets: fix product grid display

### LINKED ISSUE #4189: Block Widgets: Grid blocks have styling issues related to # of columns displayed
_This issue surfaced as part of our audit of the current state of using WooCommerce Blocks as Block Widgets in the new Widgets Editor and Customizer (pca54o-1pT-p2)._

This issue affects all our grid-based blocks. By default, the block settings are set to display with three columns. In the widget editor preview, the block shows with two columns though. Ideally, the number of columns shown in the preview should match the number of columns in the setting. In the Customizer the preview correctly shows three columns.

Placed in the sidebar of Storefront, the block shows three columns on the Frontend when there’s clearly not enough space. Here it probably should drop down to two columns.

|Widget editor|Sidebar Frontend|
|-|-|
|screenshot-2021-05-08-at-11 07 22|screenshot-2021-05-08-at-11 13 18|

frontdevde:
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4189

This PR fixes an issue where the product grid in all blocks it is used in displays the wrong number of columns.

The issue is caused by [this generic Gutenberg @reset mixin](https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/_mixins.scss#L375) overriding the box-sizing of all children of the `.blocks-widgets-container` in the [edit widgets styles here](https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-widgets/src/style.scss#L27).

|Before|After|
|-|-|
|Screenshot 2021-07-01 at 13 18 04|Screenshot 2021-07-01 at 13 29 19|

1. Ensure the Gutenberg feature plugin is enabled and up to date.
2. Go to the Widgets editor in Appearance > Widgets. 
3. Go to Sidebar widgets and insert blocks that use the product grid e.g. _Top Rated Products_.
4. Confirm that the number of columns displayed equals the number set in the block settings (Default: 3).

### Changelog

> Ensure product grids display as intended in the widget editor.

### DIFF
--- a/assets/js/base/components/product-list/style.scss
+++ b/assets/js/base/components/product-list/style.scss
@@ -55,6 +55,11 @@
 	list-style: none;
 }
 
+// The block widgets editor's generic @reset overrides box sizing on all children
+.blocks-widgets-container .wc-block-grid__product {
+	box-sizing: border-box;
+}
+
 .theme-twentytwenty .wc-block-grid,
 .wc-block-grid {
 	&.has-aligned-buttons {


assets/js/base/components/product-list/style.scss
```diff
@@ -55,6 +55,11 @@
 	list-style: none;
 }
 
+// The block widgets editor's generic @reset overrides box sizing on all children
+.blocks-widgets-container .wc-block-grid__product {
```
Aljullu, line 59: Would the fix work if we dropped `.blocks-widgets-container` from the selector? In general, I think we should try to have exactly the same CSS in the frontend and the editor (block editor, site editor or widget editor). Otherwise, things get a bit tricky for themes because there are selectors with different specificity in frontend _vs._ editor so they need to write different styles for each environment.

Aljullu (approved):
Nice investigation figuring out what was causing this issue. Confirming the fix works! :ok_hand: 

I just left a comment below about the selector, but besides that, pre-approving.

assets/js/base/components/product-list/style.scss
frontdevde: I fully agree that we should aim for having the same CSS for both the frontend and editor. In this case, the issue is that Gutenberg has block widgets editor-specific styles that override the box-sizing of `.wc-block-grid__product`. If we were to drop the class, it'd be the same as the selector that currently gets overridden by the new block widgets editor styles. This is why we, unfortunately, need the extra specificity here.

PR #4422, opened 2021-07-02
Fix filtering by product type on Store API

VoHoTv:
I moved the $tax_query variable above the code where filtering happens for the product type. The reason for this change is to avoid resetting the $tax_query after a product type is specified for filtering. Which leaves you with an unfiltered product type result.

### DIFF
--- a/src/StoreApi/Utilities/ProductQuery.php
+++ b/src/StoreApi/Utilities/ProductQuery.php
@@ -40,6 +40,9 @@ class ProductQuery {
 		if ( ! empty( $request['sku'] ) ) {
 			$args['post_type'] = [ 'product', 'product_variation' ];
 		}
+		
+		// Taxonomy query to filter products by type, category, tag, shipping class, and attribute.
+		$tax_query = [];
 
 		// Filter product type by slug.
 		if ( ! empty( $request['type'] ) ) {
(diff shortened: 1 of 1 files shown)


senadir:
Hey, @VoHoTv Thank you for your PR!
Can you add testing notes so I can reproduce the original issue?

VoHoTv:
> Hey, @VoHoTv Thank you for your PR!
> Can you add testing notes so I can reproduce the original issue?

Hi, sure. To reproduce the original error go to to /wp-json/wc/store/products?type=simple and you will see that the product type filter does not work and instead lists products without regard to the product type.

senadir:
Thank you! works as expected, will merge after tests pass.

src/StoreApi/Utilities/ProductQuery.php
```diff
@@ -40,6 +40,9 @@ public function prepare_objects_query( $request ) {
 		if ( ! empty( $request['sku'] ) ) {
 			$args['post_type'] = [ 'product', 'product_variation' ];
 		}
+
```
frontdevde:
```suggestion

```

src/StoreApi/Utilities/ProductQuery.php
```diff
@@ -40,6 +40,8 @@ public function prepare_objects_query( $request ) {
 		if ( ! empty( $request['sku'] ) ) {
 			$args['post_type'] = [ 'product', 'product_variation' ];
 		}
+		// Taxonomy query to filter products by type, category, tag, shipping class, and attribute.
```
frontdevde, line 44: Removed the whitespace but also removed the line break in the process. Adding the line break back in.

```suggestion

		// Taxonomy query to filter products by type, category, tag, shipping class, and attribute.
```

senadir (approved):

PR #4426, opened 2021-07-05
Fixing the rotation of the loading icon

vmatt:
### Fixes 
If there the screen size is small, the "Add to basket" text could break into two lines. Width height: auto, the loading icons element height is not as big as it's width, thus the loading icon is not rotating at it's centre.

### Screenshots

before
after

1. Open a WP page with Best selling products
2. Click on Add to basket (it's different if you're not using the US-EN locale)
3. It's spinning around it's centre.

### Changelog

Fixed the rotation of the add to cart loading icon, if the add to cart text is break to two or more lines.

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -81,7 +81,7 @@
 			margin-left: 0.5em;
 			display: inline-block;
 			width: auto;
-			height: auto;
+			height: 100%;
 		}
 	}
 }


ralucaStan:
Hi there @vmatt,
thank you for taking the time to look into this issue. 
Could you share what theme you used for testing? 
I tested with Storefront and there is no issue, see image below. There might be some code from the theme you used that is actually causing this bug.

Screenshot 2021-07-07 at 15 11 20

vmatt:
> Hi there @vmatt,
> thank you for taking the time to look into this issue.
> Could you share what theme you used for testing?
> I tested with Storefront and there is no issue, see image below. There might be some code from the theme you used that is actually causing this bug.
> 
> Screenshot 2021-07-07 at 15 11 20

Hi,
Authority Pro
In devtools set screen width to 650px
[Check here](https://internationalbullies.com/)

vmatt:
The failed test completely irrelevant for this fix. Please merge.

ralucaStan:
Hi there @vmatt,
thank you for your patience.

`height: auto` actually makes sense, as we want the element to have the height of the containing child. With `height: 100%` we are telling it to take the height of the parent.

I've taken a closer look I've noticed that you are using some custom CSS and are not using the All products block. 

Considering all the above, I am gonna go ahead and close this issue.

PR #4427, opened 2021-07-05
Make payment method icons display well even if theme tries to override their height/width

### LINKED ISSUE #4371: There is a bug on the shopping cart page when the resolution changes
I have seen through hotjar that my customers are facing this problem.

The user came with the tablet to the page to buy the product, the resolution of the tablet is 748x480.

Capture

The thumbnails below the "process to checkout" button were magnified full screen and the user could not do anything. I lost that customer because of it.

I had to manually add my own CSS style to customize the thumbnails.

Please fix this in the next version.

opr:
Note: I couldn't replicate the issue reported in #4371 without adding custom CSS, such as:
```css
.wp-block-woocommerce-cart img,
.wp-block-woocommerce-checkout img {
    height: auto !important;
    width: 100% !important;
}
```
I added this through an option in the theme, but you could also add it to `{your-theme}/style.css`.

This PR will add a max-height and max-width to payment method icons, we've seen a couple of instances of themes adding 100% width to images which have been applied to the payment icons.

It also adds the icons to a flex container to ensure they display in a row and wrap onto the next line if there are too many icons to fit on one line.

It also changes the display of the individual payment method icons on the Checkout page, it sets `object-fit: cover` and `object-position: left` so that even if themes have set 100% width on images, any payment method image won't stretch.

Fixes #4371 (I can't replicate #4371 but have emulated using custom CSS.)

### Screenshots

Cart
| Before | After |
|---|---|
| image | image |

Checkout
| Before | After |
|---|---|
| image | image |

1. Add some custom CSS to your theme, do something that will "accidentally" catch all images in the Cart/Checkout blocks, e.g:
```css
.wp-block-woocommerce-cart img,
.wp-block-woocommerce-checkout img {
    height: auto !important;
    width: 100% !important;
}
```
2. Go to the Cart and Checkout blocks and ensure that the payment method icons display correctly and are not too large.
3. Experiment with a few different themes and ensure this works for each one you test.

### Changelog

> Ensure payment method icons are constrained to a reasonable size in the Cart and Checkout blocks.

### DIFF
--- a/assets/js/base/components/cart-checkout/payment-method-icons/style.scss
+++ b/assets/js/base/components/cart-checkout/payment-method-icons/style.scss
@@ -1,15 +1,19 @@
 .wc-block-components-payment-method-icons {
-	display: block;
 	text-align: center;
 	margin: 0 0 #{$gap - 2px};
+	display: flex;
+	flex-direction: row;
+	flex-wrap: wrap;
+	justify-content: center;
 
 	.wc-block-components-payment-method-icon {
 		display: inline-block;
 		margin: 0 4px 2px;
 		padding: 0;
 		width: auto;
+		max-width: 38px;
 		height: 24px;
-		vertical-align: middle;
+		max-height: 24px;
 	}
 
 	&--align-left {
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Looking great, thanks @opr! :clap:

PR #4428, opened 2021-07-06
Fix all review memory leak on block transform

### LINKED ISSUE #4304: All Reviews block: unmount state update error
We're getting the following reproducible error when interacting with the All reviews block:

> Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.

**Steps to reproduce:**
I was able to reliably reproduce the issue with the following steps:
- Install the Gutenberg plugin
- Create an empty page. 
- Add the All Reviews block. 
- In the toolbar select the transform options and quickly hover back and forth over the various transform options.
- Deselect the block

ralucaStan:
There was a memory leak when previewing transform options for the All reviews block.
The preview window was fetching the reviews, but the user could move away from the transform option before the ratings loaded and the preview was shown. This PR makes sure that the preview window is still opened before updating the state of the component with the new reviews. 

Fixes #4304 

- Install WooBlocks plugin
- Create an empty page and add the All Reviews block. 
- Select the block and in the toolbar select the transform option
- Hover on one option and hover out before the preview is loaded
- Check the console is free of error warnings

### Changelog

> Fix memory leak when previewing transform options for the All reviews block

### DIFF
--- a/assets/js/base/hocs/with-reviews.js
+++ b/assets/js/base/hocs/with-reviews.js
@@ -47,6 +47,8 @@ const withReviews = ( OriginalComponent ) => {
 
 		delayedAppendReviews = this.props.delayFunction( this.appendReviews );
 
+		isMounted = false;
+
 		state = {
 			error: null,
 			loading: true,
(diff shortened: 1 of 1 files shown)


opr (approved):
Cool fix, Raluca! Works well and the code seems fine to me 👏🏼

PR #4454, opened 2021-07-09
Add add-hoc Block locking from WooCommerce Blocks side.

### LINKED ISSUE #4453: Create interim solution to disable removal of default Blocks

senadir:
This PR adds an interim solution for locking blocks from being removed or moved. It does this by using the same registration system introduced in [#32457](https://github.com/WordPress/gutenberg/pull/32457) via `supports` option and locks the blocks.
- It adds a classname to the body when a locked block is selected, this is to hide the remove and drag buttons from the block toolbar.
- It intercepts the `DELETE` and `BACKSPACE` keydown events on the block and prevent them from propagating to Gutenberg.

Fixes #4453

1. Currently, in this PR, only express payment methods is locked.
2. Try deleting that block or moving it, it shouldn't work.
3. Try deleting other block, they all work fine.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout-i2/edit.tsx
+++ b/assets/js/blocks/cart-checkout/checkout-i2/edit.tsx
@@ -12,7 +12,10 @@ import {
  * Internal dependencies
  */
 import { Columns } from './columns';
-import { AddCurrentlySelectedBlockClass } from './hacks';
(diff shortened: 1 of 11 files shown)


mikejolley (approved):
This does the job. Done some cleanup and made sure the follow-ups are logged. Let's merge 👍🏻

PR #4457, opened 2021-07-12
Checkout i2: Enable shipping toggle in the editor

### LINKED ISSUE #4443: Checkout i2: "use same address for billing" toggle in the editor
In the editor, the toggle is checked, but the billing form is visible:

This does not reflect the appearance on the frontend.

Given the address field config is now shared between shipping and billing addresses, we can probably have this toggle functional in the editor. So you can toggle it off to edit the form step contents, whilst still previewing how the checkout should look on the frontend.

mikejolley:
For Checkout i2, to make the editor better reflect the frontend, this re-enables the "Use same address" checkbox so that you can preview how it looks with this on and off.

When the billing section is visible, everything remains editable, so the toggle is not persisted and does not affect saving. Since the billing and shipping sections now **share** address field configs, it doesn't matter if billing is hidden by default.

Note; whilst the sections can be moved right now, this will be disabled in the final version, so the position of the billing and shipping sections does not matter at this point if moving blocks around.

Fixes #4443

1. Add Checkout i2 to a page
2. Try toggling the "Use same address" checkbox to show/hide billing address
3. Confirm address fields are still disabled.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-billing-address-block/block.tsx
+++ b/assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-billing-address-block/block.tsx
@@ -1,11 +1,13 @@
 /**
  * External dependencies
  */
-import { useMemo, useEffect } from '@wordpress/element';
+import { useMemo, useEffect, Fragment } from '@wordpress/element';
+import { Disabled } from '@wordpress/components';
 import {
-	useStoreEvents,
 	useCheckoutAddress,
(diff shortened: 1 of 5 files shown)


frontdevde (approved):
Tested as per testing instructions on both the feature branch and this branch. Can confirm that now with this branch the shipping toggle is enabled in the editor while it wasn't on the feature branch. The text fields are disabled as expected in both cases.

I've noticed that the block doesn't show on the Frontend of the site. That's the case on the feature branch for me too though, so I assume it's known/unrelated and I'm ✅ with that in mind.

Issue #4245, opened 2021-05-20
Research spike: investigate performant ways to render interactive blocks in the frontend


Aljullu:
We have several blocks which import React to add frontend interactivity. While that works for blocks which are the main page content (ie: Cart and Checkout), it might be a performance issue for blocks that are rendered in all pages (ie: Minicart), we should investigate alternative approaches for this (Preact, Web Components, vanilla JS...).

### Outcome

A successful outcome of this research would be having a decision on what's the best solution to build interactive blocks without React or any other heavy dependency. Ideally, we would also have an example block we can use as an example. Ie: and Add to cart block (which will also require #4246).

### Resources

Some resources that might be good to read before diving into this:

* p1HpG7-7Aj-p2
* p1620710803455000-slack-C45SNKV4Z
* p7H4VZ-32U-p2
* p9dueE-Lm-p2
* p1HpG7-9PA-p2

Aljullu:
Conclusions of this research spike:

* **Favor Vanilla JS over any other library or framework when possible:** using Vanilla JS in simple blocks makes bundle sizes smaller because less dependencies need to be loaded. We should consider this approach the 'default' one and only use heavy dependencies when really needed. That might happen when developing a block that requires complex interactions or data management that would be hard to build from scratch
* **Don't use Preact:** while we observed some bundle size improvements when building the Reviews blocks with Preact instead of `@wordpress/element`, it was noted that a solution at the framework level would be more desirable. We decided to stick with `@wordpress/*` packages as we have been doing for now, so we benefit of future improvements of those packages.
* **Apply lazy loading when possible:** when developing new blocks, let's look for opportunities to apply lazy-loading. We can load a very small JS file at the beginning to handle user interaction and only load other dependencies like `@wordpress/element` when they are really needed. A perfect example of this is the Mini Cart block: its initial state can be achieved with PHP + vanilla JS and only load other dependencies once the user clicks on the button or adds a product to the cart.
* **Don't do PHP prerendering:** one thing that was considered to improve  perceived loading speed when rendering `@wordpress/element`-based blocks was to prerender their contents in PHP. That presented some issues like having to duplicate the rendering logic from JSX components in PHP, and not having a way to honor JS-based extensibility points in the PHP prerender. We considered those drawbacks to be important enough to avoid pursuing this solution.

More info here: pca54o-20m-p2.

PR #4463, opened 2021-07-14
Use Preact to render Reviews blocks frontend

Aljullu:
Part of #4245.

This PR adds a new Webpack config that generates frontend scripts using Preact instead of React and updates the Reviews blocks to use it.

Currently, Preact is bundled into the built file, but if in the future we use Preact in more blocks, we can consider externalizing it so it's only downloaded once. Jetpack also uses Preact for the Jetpack Search, so the win of externalizing would be even greater for sites using Jetpack and WC.

I did a quick test on the amount of scripts required to render a page only containing the All Reviews blocks in `trunk` and in this branch with Storefront:

| Branch | Requests | Size |
| --- | ---: | ---: |
| `trunk` | 43 | 696,06 KB |
| This branch | 25 | 320,96 KB |

Excluding all scripts unrelated to the block (jQuery, WC core scripts, etc.), it looks like this:

| Branch | Requests | Size |
| --- | ---: | ---: |
| `trunk` | 30 | 554,79 KB |
| This branch | 12 | 179,69 KB |

<details>
<summary>Scripts required in `trunk` to render the All Reviews block in the frontend</summary>

List of dependencies

</details>

<details>
<summary>Scripts required in this branch to render the All Reviews block in the frontend</summary>

List of dependencies

</details>

#### For devs (no need to include this in the release testing steps)

1. Add a Reviews block to a post or page and in the frontend verify these scripts are not loaded:
  * `/wp-content/plugins/gutenberg/build/element/index.min.js`
  * `react`
  * `react-dom`
2. This PR removes `"./assets/js/settings/blocks/**"` from the list of `sideEffects`. That was introduced in this PR: #2042. I wasn't able to reproduce the issues mentioned in that PR, but would be good to double-check. cc @nerrad 

#### For users (include this in the release testing steps)

1. Smoke test All reviews, Reviews by product and Reviews by Category blocks in the editor and the frontend: try with different attributes, changing the order of reviews, clicking on `Load more`, etc. and verify there are no regressions.
2. Smoke test other blocks with frontend scripts: All Products & filters, Cart and Checkout. Verify there are no regressions.

### Changelog

> Reduced the amount of JavaScript code required to render Reviews blocks, that will speed up page loads when a Reviews block is rendered.


Aljullu:
Closing this PR because we decided not to add Preact to any of our blocks and instead continue relying on `@wordpress/element` with the hope performance improvements will come at the framework level in the future. More info here: pca54o-20m-p2.

PR #4469, opened 2021-07-14
Fix SKU search in the Store API

mikejolley:
I discovered that the SKU searching code in the Store API **was never ran**. This is due to using the wrong variable name (WP Query uses `s`, but we were looking for `search` ).

Since we actually want to control the search, we should map this to `search` so we can search on the product title and SKU without WP_Query interfering with things.

I have added a test to check SKU search is functional.

Try some search requests via the Store API, for example:

`https://store.local/wp-json/wc/store/products?search=woo-album`

Ensure matched results match your query. Try searching for both titles and SKUs.

### Changelog

> Fixed the SKU search on the /wc/store/products endpoint.


ralucaStan:
This tests out good.
I recommend somebody else's 👀 to double-check the PHP code.

Aljullu (approved):
Confirming it's testing well and changes look safe based on the discussion we had yesterday.

PR #4471, opened 2021-07-16
Remove utils dead files

Aljullu:
Something that I found while working on #4463. It looks like we had a file that was never used.

#### For devs (no need to include this in the release testing steps)

1. Verify `npm run build`, `npm run start` and `npm test` work fine (e2e tests are failing but it's unrelated to this PR, I think @ralucaStan is investigating it in #4464).
2. Verify GitHub Actions show no change in bundle sizes. That proves those files weren't imported from anywhere.

#### For users

1. Do some basic smoke testing with several blocks to verify there aren't evident regressions.


mikejolley (approved):
👍🏻  checks out. I think these were used by some deprecated block handling we removed.

Issue #3350, opened 2020-10-30
Export Woo Blocks e2e tests as a package so they can be pulled into Woo Core for e2e tests.


nerrad:
We currently have e2e testing flows that are equivalent to smoke tests for blocks. It'd be useful to export our e2e tests as a package so that WooCommerce core can have the package as a dependency and use it for executing Woo Blocks e2e tests in the Woo Core environment (with blocks included as a package). Then whenever we do a Woo Blocks package bump pull request in Woo Core, if there are new e2e tests in the blocks e2e package, we can also bump that dependency as well.

The outcome here is to have blocks e2e tests running in WooCommerce core which should hopefully reduce the amount of manual smoke testing as well.

PR #4484, opened 2021-07-20
Update relevant files due to L2->L0 policy change for minimum required versions.

nerrad:
For a number of reasons (which will be documented in a separate post on developer.woocommerce.com before merging this PR), we have decided to change the minimum version support policy for WooCommerce Blocks to L0 from L2. This effectively means that beginning with the next release of this plugin (after merging this PR), the latest version of WordPress (currently WP 5.8) and WooCommerce (currently 5.5) will be _required_.

There will be some impact on current users of the plugin but the main impact would mostly be they won't be able to update the feature plugin to the latest version until their site satisfies the minimum version requirements.

When included in WooCommerce core as a package, WooCommerce blocks will adhere to the Woo core version policy _for existing features_ and functionality already available in WooCommerce core via the feature plugin. However, _new blocks or functionality_ coming from the package inclusion _might_ be behind a WordPress version gate in keeping with the version followed when the new block/feature was being developed.

Note: This PR description will be updated with a link to the developer.woocommerce.com post about this policy change once it goes up (and I'll also be linking to the post within some of the files changed in the PR).

## Changes included

For the purpose of review, here are the files/impact of these changes:

- Plugin header, and readme files that both inform and (via built-in WordPress and WooCommerce logic) restrict the usage of the plugin.
- Main plugin file (which restricts enabling on an unsupported WordPress version). The main plugin file is not loaded when the code is loaded as a package in Woo Core so should not impact usage there.
- `src/Domain/Bootstrap.php` - Woo Blocks didn't have any restriction for Woo Core version, this was added to the bootstrap. It reads from the plugin header so we don't have to keep this updated (just the plugin header).
- Automated tests are updated to test against the supported L0 supported versions only. Unfortunately, we're unable to set things up for L2 in the automated tests because there's no way to ensure blocks is active for earlier WP versions (as far as I could tell there's no way to filter the Blocks plugin header values). I did try to do this because it's advantageous to ensure existing blocks released in Woo Core still support L2 of WordPress.
- PHPCS config. This updates the PHPCS lint rules to account for the latest version of WordPress supported by the L2 policy in core. Why not L0? We might work on code in a block or feature plugin that exists already in Woo Core and in that case this is a preventative measure against accidentally breaking that contract with the use of functions or code only existing in L0 within those things. We may have to iterate on this rule a bit if we find it leads to a lot of lint ignore rules in the codebase.

**Note: I also made some updates to the PHPUnit bootstrap so we could run the latest Woo version for PHPUnit.** In #4305, I added some logic so we'd have the ability to pin PHPUnit tests to specific Woo versions to get around errors we started seeing happening. In this PR, I was able to successfully debug the cause of the errors that was preventing us from using the latest Woo Core version and fixed it so we can now run the latest Woo Core version for PHPunit tests.

## To test

For testing the below scenarios you can manually set the "requires values" in `woocommerce-gutenberg-products-block.php` file to higher versions than what's installed in your local dev environment.

The only user facing impact of these changes is if a user is on a non-supported version of WordPress and/or WooCommerce when they are notified about a pending update. They _shouldn't_ be able to update in that scenario.

For users that update via FTP or some other direct method, they shouldn't be able to activate the plugin.

WooCommerce inclusion via package should not impact usage there (which will be tested as a part of the next package bump for WooCommerce 5.7).

## Re: GH workflow checks not completing.

Keep in mind that there are some required checks configured in the repo settings that no longer get executed due to the changes in this PR. After this PR is merged, we can update the required checks in the repo settings.


nerrad:
Accompanying blog content draft here: 181-gh-woocommerce/woocommerce-developer-advocacy

nerrad:
I need to make sure that we have a test for the required WC version because there currently is no hardcoded behaviour in Woo Core that will prevent activating the plugin if the required version isn't met.

src/Domain/Bootstrap.php
```diff
@@ -100,7 +100,31 @@ protected function init() {
...
+							<div class="notice notice-error is-dismissible">
+								<p><?php esc_html_e( 'WooCommerce Blocks requires a more recent version of WooCommerce and has been paused. Please update WooCommerce to the latest version to continue enjoying WooCommerce Blocks.', 'woo-gutenberg-products-block' ); ?></p>
+							</div>
+							<?php
+						}
+					}
+				);
+			}
+		}
+		return $has_needed_dependencies;
```
nerrad: There is currently nothing restricting the plugin from operating if the active WooCommerce version is lower than that required by Woo Blocks. This code implements that.

tests/bootstrap.php
```diff
@@ -83,6 +85,58 @@ function wc_test_includes() {
...
+ * to replace the registered read-only container with a fully configurable one for testing.
+ * To this end we hack a bit and use reflection to grab the underlying container that the read-only one stores
+ * in a private property.
+ *
+ * Additionally, we replace the legacy/function proxies with mockable versions to easily replace anything
+ * in tests as appropriate.
+ *
+ * @throws \Exception The Container class doesn't have a 'container' property.
+ */
+function wc_blocks_initialize_dependency_injection() {
```
nerrad, line 101: This was copied from the WooCommerce core tests bootstrap file. It registers `MockableLegacyProxy::class` in place of `LegacyProxy::class` for the purpose of tests and fixes the errors we were getting with running our phpunit tests.

phpcs.xml
```diff
@@ -9,7 +9,7 @@
 	<exclude-pattern>languages/woo-gutenberg-products-block.php</exclude-pattern>
 
 	<!-- Configs -->
-	<config name="minimum_supported_wp_version" value="4.7" />
+	<config name="minimum_supported_wp_version" value="5.6" />
```
mikejolley, line 12: I think we should add a comment here or to the release checklist suggesting the appropriate value here. I'd suggest the min version supported by WC core rather than L0 (which I think you have?)

.github/patch-initial-checklist.md
```diff
@@ -8,8 +8,8 @@ The release pull request has been created! This checklist is a guide to follow f
   * [ ] Add the version and date to the changelog section within `readme.txt`, e.g. `= {{version}} - YYYY-MM-DD =`
   * [ ] Copy the changelog from the pull request description above into this new section
 * [ ] Update compatibility sections (if applicable). __Note:__ Do not change the stable tag or plugin version; this is automated.
-  * [ ] Update _Requires at least_, _Tested up to_, and _Requires PHP_ sections at the top of `readme.txt`.
-  * [ ] Update _Requires at least_, _Requires PHP_, _WC requires at least_, and _WC tested up to_ at the top of `woocommerce-gutenberg-products-block.php`
+  * [ ] Update _Requires at least_, _Tested up to_, and _Requires PHP_ sections at the top of `readme.txt`. Note, this should also be the latest WordPress version available at time of release.
```
mikejolley: Can we (Follow up?) automate this as part of the release process and `version-changes.sh`?

mikejolley:
> Automated tests are updated to test against the supported L0 supported versions only. Unfortunately, we're unable to set things up for L2 in the automated tests because there's no way to ensure blocks is active for earlier WP versions (as far as I could tell there's no way to filter the Blocks plugin header values). I did try to do this because it's advantageous to ensure existing blocks released in Woo Core still support L2 of WordPress.

Do we need to add some kind of basic integration test in woo core to ensure blocks (included with core) exist/load without errors? Apologies if this exists already.

src/Domain/Bootstrap.php
```diff
@@ -100,7 +100,31 @@ protected function init() {
 	 * @return boolean
 	 */
 	protected function has_core_dependencies() {
-		return class_exists( 'WooCommerce' ) && function_exists( 'register_block_type' );
+		$has_needed_dependencies = class_exists( 'WooCommerce' ) && function_exists( 'register_block_type' );
```
mikejolley: `register_block_type` was introduced in WP 5.0. Is a check still relevant?

src/Domain/Bootstrap.php
```diff
@@ -100,7 +100,31 @@ protected function init() {
...
+			);
+			if ( isset( $plugin_data['RequiredWCVersion'] ) && version_compare( \WC()->version, $plugin_data['RequiredWCVersion'], '<' ) ) {
+				$has_needed_dependencies = false;
+				add_action(
+					'admin_notices',
+					function() {
+						if ( should_display_compatibility_notices() ) {
+							?>
+							<div class="notice notice-error is-dismissible">
+								<p><?php esc_html_e( 'WooCommerce Blocks requires a more recent version of WooCommerce and has been paused. Please update WooCommerce to the latest version to continue enjoying WooCommerce Blocks.', 'woo-gutenberg-products-block' ); ?></p>
```
mikejolley: To avoid confusion with the blocks in Woo core, perhaps we can be more explicit here? i.e. "The WooCommerce Blocks Feature Plugin requires..."

src/Domain/Bootstrap.php
```diff
@@ -100,7 +100,31 @@ protected function init() {
...
+				]
+			);
+			if ( isset( $plugin_data['RequiredWCVersion'] ) && version_compare( \WC()->version, $plugin_data['RequiredWCVersion'], '<' ) ) {
+				$has_needed_dependencies = false;
+				add_action(
+					'admin_notices',
+					function() {
+						if ( should_display_compatibility_notices() ) {
+							?>
+							<div class="notice notice-error is-dismissible">
```
mikejolley: This notice has `is-dismissible` class, but is not dismissable. It will return on refresh.

nerrad:
> Do we need to add some kind of basic integration test in woo core to ensure blocks (included with core) exist/load without errors? Apologies if this exists already.

There isn't anything that exists yet and I think this should be done as a followup (maybe via collaborating with the quality team). This is something that #3350 would really help with. For now, I think we have some coverage with the GlobalStep testing that is done for each release.

src/Domain/Bootstrap.php
```diff
@@ -100,7 +100,31 @@ protected function init() {
 	 * @return boolean
 	 */
 	protected function has_core_dependencies() {
-		return class_exists( 'WooCommerce' ) && function_exists( 'register_block_type' );
+		$has_needed_dependencies = class_exists( 'WooCommerce' ) && function_exists( 'register_block_type' );
+		if ( $has_needed_dependencies ) {
+			$plugin_data = \get_file_data(
```
mikejolley, line 105: Will using `get_file_data` on every page load cause performance issues? There is no caching here. Does this have to be dynamic/based on headers, or could we add a release step to update a hardcoded version string here to compare with the WC Core version?

tests/bootstrap.php
mikejolley: I don't full understand what is happening here, but assuming tests pass 👍🏻

woocommerce-gutenberg-products-block.php
```diff
@@ -18,7 +18,7 @@
 
 defined( 'ABSPATH' ) || exit;
 
-$minimum_wp_version = '5.5';
+$minimum_wp_version = '5.8';
```
mikejolley, line 21: Regarding my comment in the `has_core_dependencies` method where we pull metadata from a file, we could potentially include minium_wc_version in the same way as this `$minimum_wp_version`.

mikejolley (commented):
Added some feedback/comments on the approach. I have not tested this branch _yet_. I will also look over the dev blog draft.

phpcs.xml
nerrad: Good callout, ya I switched to the min version supported by WC core.

.github/patch-initial-checklist.md
nerrad: Yup I'd love to automate. We can create a followup issue that can be done sometime.

src/Domain/Bootstrap.php
nerrad: Oh, good catch! Nope, we don't need it.

src/Domain/Bootstrap.php
nerrad: 🤔 I copied what is currently used by the wp core version check notice in the main plugin file. So probably should just remove the classname here and with the WP core version check as well.

src/Domain/Bootstrap.php
nerrad: I'll double-check on this. The advantage of headers is it's one less thing to update (compared with having a hardcoded string). I don't think the performance hit should be significant because WP only pulls the first 8kb of the files and the parsing is fairly straightforward. I also am only passing in the single array value to check for (which is implemented in the regex pattern).

woocommerce-gutenberg-products-block.php
nerrad: I had initially thought of that, but if we do that we'd have to either pass it through to the Bootstrap (or maybe Package class) as an argument, we make it `$minimum_wp_version` a global, or we do the Woo core version check in a callback on `plugins_loaded` implementing the `use` keyword on the action callback to pass through the $minimum_wp_version. I didn't think of trying the last item until just now so I guess we could go with that to avoid the extra code (and potential performance hit) in the Bootstrap class.

src/Domain/Bootstrap.php
mikejolley: That's fair, but hardcoded values will still perform better, less can go wrong, and it will be easier to read. If we're going to go ahead with automated version changes in the deploy script (as a follow up), it shouldn't matter if there is one more version string to define. We're already doing something similar for `$minimum_wp_version`.

nerrad: Yup agreed, as noted in my feedback on a later comment I'm going to try implementing that.

.github/patch-initial-checklist.md
nerrad: See #4511

src/Domain/Bootstrap.php
nerrad: Actually, I just remembered why I took this route. I had to put this in the bootstrap class because otherwise the code change complexity involved in _aborting the loading of Woo Blocks_ is greater. With the WP check, the main file can just return early, but the WC check has to happen (at a minimum) after `plugins_loaded`. So I either have to deactivate the feature plugin when minimum WC version isn't supported, or do some refactoring to support early aborting. All viable solutions but would likely involve more code changes.

nerrad: I just did some rough performance checks on this code using this:

```php
protected function has_core_dependencies() {
		$has_needed_dependencies = class_exists( 'WooCommerce' ) && function_exists( 'register_block_type' );
		$start_time              = hrtime( true );
		if ( $has_needed_dependencies ) {
			$plugin_data = \get_file_data(
				$this->package->get_path( 'woocommerce-gutenberg-products-block.php' ),
				[
					'RequiredWCVersion' => 'WC requires at least',
				]
			);
			if ( isset( $plugin_data['RequiredWCVersion'] ) && version_compare( \WC()->version, $plugin_data['RequiredWCVersion'], '<' ) ) {
				$has_needed_dependencies = false;
				add_action(
					'admin_notices',
					function() {
						if ( should_display_compatibility_notices() ) {
							?>
							<div class="notice notice-error is-dismissible">
								<p><?php esc_html_e( 'WooCommerce Blocks requires a more recent version of WooCommerce and has been paused. Please update WooCommerce to the latest version to continue enjoying WooCommerce Blocks.', 'woo-gutenberg-products-block' ); ?></p>
							</div>
							<?php
						}
					}
				);
			}
		}
		$end_time      = hrtime( true );
		$time_executed = 'Time spent:' . ( $end_time - $start_time );
		var_dump( $time_executed );
		return $has_needed_dependencies;
	}
```

The _most_ time I saw reported on multiple refreshes across different pages (including frontend/backend) was 4565661 nano seconds (or 0.004565661 seconds). IMO that is extremely negligible and in keeping with the performance impact of including PHP files.

woocommerce-gutenberg-products-block.php
nerrad: See my [latest comment here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4484#discussion_r677697085) (and the following performance checks I did). I think we're okay with sticking with the current implementation given:

- it retains one place to change minimum required version for Woo.
- the alternative will likely lead to more code churn/refactoring.

phpcs.xml
nerrad: see a495b6b

src/Domain/Bootstrap.php
nerrad: see 70300d2

nerrad:
Ready for another review.

.github/patch-initial-checklist.md
```diff
@@ -8,8 +8,10 @@ The release pull request has been created! This checklist is a guide to follow f
...
+  * [ ] Update _Requires at least_, _Requires PHP_, _WC requires at least_, and _WC tested up to_ at the top of `woocommerce-gutenberg-products-block.php`. Note, this should include requiring the latest WP and WC versions at the time of the plugin release.
```
Aljullu: This is the checklist for patch releases, right? Shouldn't patch releases support the same WP and WC versions from the release they are based from?

README.md
```diff
@@ -40,19 +39,16 @@ If you want to see what we're working on for future versions, or want to help ou
 
 We release a new version of WooCommerce Blocks onto WordPress.org every few weeks, which can be used as an easier way to preview the features.
 
-> Note: The plugin follows a policy of supporting the "L2" strategy for version support. What this means is that the plugin will support the most recent two minor versions of WordPress, and the most recent two minor versions of WooCommerce core at the time of a release.
+> Note: The plugin follows a policy of supporting the "L0" strategy for version support. What this means is that the plugin will require the most recent version of WordPress, and the most recent version of WooCommerce core at the time of a release. You can read more about [this policy here](https://developer.woocommerce.com/?p=9998)
```
Aljullu: Nit, I would add a dot at the end of the sentence:
```suggestion
> Note: The plugin follows a policy of supporting the "L0" strategy for version support. What this means is that the plugin will require the most recent version of WordPress, and the most recent version of WooCommerce core at the time of a release. You can read more about [this policy here](https://developer.woocommerce.com/?p=9998).
```

Aljullu (commented):
I'm not sure if I'm testing it wrong, but I noticed something weird. In a WP 5.8 install, when I set `$minimum_wp_version = '5.9';` in `woocommerce-gutenberg-products-block.php` I correctly get the error notice in the plugins screen. However, all assets are still loaded from `/wp-content/plugins/woocommerce-gutenberg-products-block/build/` instead of `/wp-content/plugins/woocommerce/packages/woocommerce-blocks/build/` and the Cart & Checkout blocks are still available. Is that the expected behavior?

I also left a couple of minor comments below. I also agree with what was discussed in this PR about the risks of losing automatic testing for legacy versions of WordPress that we still support via WC core. @nerrad do you think #3350 should be added to next cycle if it's not taken during this cooldown?

nerrad:
> I'm not sure if I'm testing it wrong, but I noticed something weird. In a WP 5.8 install, when I set $minimum_wp_version = '5.9'; in woocommerce-gutenberg-products-block.php I correctly get the error notice in the plugins screen. However, all assets are still loaded from /wp-content/plugins/woocommerce-gutenberg-products-block/build/ instead of /wp-content/plugins/woocommerce/packages/woocommerce-blocks/build/ and the Cart & Checkout blocks are still available. Is that the expected behavior?

I think this is expected if you are using the feature plugin from the GH repo because the `JETPACK_AUTOLOAD_DEV` constant in our `wp-config.php` file is basically instructing Core to use the feature plugin paths.

> do you think #3350 should be added to next cycle if it's not taken during this cooldown?

Yes, I've added it to the Cycle 5 pipeline on the planning board. Good call-out 🙇🏻

.github/patch-initial-checklist.md
nerrad: oh right! I'll fix.

Aljullu (approved):
Thanks for answering my question and applying the feedback @nerrad. Everything looks good to me now, so if @mikejolley doesn't have anything else to add, :shipit:

mikejolley (approved):
I tested the new code by changing version strings, and I tested without `JETPACK_AUTOLOAD_DEV` to ensure it loads the correct version. I also ran the build and tests locally. Changes look good. 🚢

nerrad:
As an additional comment, worth noting that the WP check won't happen when included as a package in Woo Core because Woo Core doesn't load the feature plugin main file. So, for any features we implement that _do_ have specific WordPress version requirements, we'll have to make sure those features are gated to the WP version.

PR #4496, opened 2021-07-22
Add checkout term field validation

### LINKED ISSUE #4319: Terms and Conditions Block
This will be an inner block for Checkout i2. There will be 2 versions or layout options for this block:

- Checkbox: Mandatory checkbox with custom text beside it
- Text: Text that explains that by proceeding you agree to TOS

Notes:
- We need to decide between pre-def text, or completely custom
- Pre-def text can contain links to the TOS and privacy pages, which are selectable in the UI.
- If not pages are selected, a warning is shown to the admin.

See https://teamrubik.wordpress.com/2021/06/04/cart-and-checkout-i2-design-exploration-progress-post-2/

Edit: This has been started in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4268 and the block has been defined, but two things need to be added:

- [x] Checkbox Validation
- [x] Validation in editor if no links are added within the block text

mikejolley:
Adds client-side validation to the terms and conditions block in Checkout i2.

Closes #4319

2 supporting changes were needed to make it scroll to the correct element on error; setting the error state, and making it scroll to elements with the `has-error` class name. I added TypeScript to the `useStoreNotices` hook to clarify some confusing naming.

### Screenshots

Screenshot 2021-07-22 at 16 06 02

1. Setup Checkout i2 on a page and ensure the terms and conditions block is present
2. Go to the frontend and submit the checkout without checking the box
3. Checkout should jump to the input and focus it. The styling is red.

### DIFF
--- a/assets/js/base/context/providers/cart-checkout/checkout/processor/index.js
+++ b/assets/js/base/context/providers/cart-checkout/checkout/processor/index.js
@@ -111,12 +111,7 @@ const CheckoutProcessor = () => {
 
 	const checkValidation = useCallback( () => {
 		if ( hasValidationErrors ) {
-			return {
-				errorMessage: __(
-					'Some input fields are invalid.',
-					'woo-gutenberg-products-block'
-				),
-			};
+			return false;
 		}
 		if ( currentPaymentStatus.hasError ) {
 			return {
--- a/assets/js/base/components/checkbox-control/style.scss
+++ b/assets/js/base/components/checkbox-control/style.scss
@@ -45,6 +45,26 @@
 		}
 	}
 
+	&.has-error {
+		color: $alert-red;
+
+		a {
+			color: $alert-red;
+		}
+		.wc-block-components-checkbox__input {
+			&,
+			&:hover,
+			&:focus,
+			&:active {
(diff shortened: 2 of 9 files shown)


assets/js/base/context/providers/cart-checkout/checkout/processor/index.js
```diff
@@ -111,12 +111,7 @@ const CheckoutProcessor = () => {
 
 	const checkValidation = useCallback( () => {
 		if ( hasValidationErrors ) {
-			return {
```
mikejolley, line 114: I don't think this is necessary since inputs show inline validation notices. Agree?

senadir: I agree that the notice feels redundant. I believe design added it. We can check later.

senadir (approved):
:LGTM:

senadir (changes requested):
Missed testing one thing. This doesn't support when there's no checkbox (only text). You can't place an order.

mikejolley:
Good catch @senadir resolved.

PR #4498, opened 2021-07-23
Add missing class to Cart skeleton title

Aljullu:
Adds a missing class to the `<h2>` title of the Cart skeleton so it has the same margin as the rendered block.

1. Load the Cart block and verify the margin under the `Your cart` title doesn't change between the rendered block and the loading skeleton.

| Before | After |
| --- | --- |
| Before | After |

### Changelog

> Remove unnecessary margin from Cart block loading skeleton to avoid content jump.


ralucaStan (approved):
👍

PR #4503, opened 2021-07-23
initialize checkout alongside checkout-i2

### LINKED ISSUE #4500: Run Checkout registration action inside CheckoutI2
Third-party integrations (like WCS) hook into our blocks using dynamic hooks like `woocommerce_blocks_checkout_block_registration`.

Currently, for them to hook into Checkout i2 they need to register with `woocommerce_blocks_checkout-i2_block_registration` as well. This is only temporary until we migrate Checkout i2 to be Checkout.

But we still want to test the integration of those plugins with our block so we need to call `woocommerce_blocks_checkout_block_registration` inside the init process of Checkout i2.

mikejolley:
Runs `woocommerce_blocks_checkout_block_registration` when checkout-i2 is initialised. This ensures i2 is compatible with existing i1 integrations.

cc @senadir 

Closes #4500

### DIFF
--- a/src/BlockTypes/CheckoutI2.php
+++ b/src/BlockTypes/CheckoutI2.php
@@ -15,6 +15,27 @@ class CheckoutI2 extends AbstractBlock {
 	protected $block_name = 'checkout-i2';
 
 	/**
+	 * Initialize this block type.
+	 *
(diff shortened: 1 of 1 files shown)


senadir (approved):
Tested and it works fine with WCS.

Issue #4447, opened 2021-07-08
Checkout i2: Slot/Fill Areas


mikejolley:
As well as supporting new blocks that can be inserted by merchants (#4306, #4446) there is a need for extensions to forcibly add content to various areas within checkout. An example of this would be a newsletter plugin that wants to add a checkbox automatically without the merchant needing to insert something manually.

One solution to this would be to introduce some extra slot/fill areas. I propose adding one of these within the "checkout actions" block which can be used as a pattern if more areas are needed in the future. The area would be above the place order button, at the bottom of the checkout.

This may need further discussion with stakeholders.

~Tasks~
- [ ] ~Discuss and confirm placement of slot fill area(s)~
- [ ] ~Create slot fill area~
- [ ] ~Document usage~

Edit: [See comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4447#issuecomment-885011894) We're going to attempt a non-slot fill solution just using inner blocks.

mikejolley:
I wonder if it would be a good idea to insert an inner block (per form step) that contains the slot fill. This would mean the merchant can "move" the extensions slot fills around somewhat, whilst allowing extensions to force content to appear. @senadir thoughts?

senadir:
In a recent chat, Mike and I discussed the ability to force insert blocks registered by third-party developers.
In our finding, third-party blocks are a couple of types:
- Optional blocks that merchants can add after enabling the plugin.
- Required blocks that must be available after enabling the plugin.

The first part is already covered with https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4446.
The second part is a bit complicated, initially, we wanted to do this Slot Fills, this can be done by:
- Having a pseudo like inner block per form step that has a slot fill, plugins can register to it and merchant can move (the comment above).

This creates a limitation in which a developer might need a block to be required and fixed (WCS totals, Smart coupons line item...).
This would make using the SlotFill hard because we would need to provide a slot fill before an after each block, which would make supporting them a nightmare.

A solution we want to explore is to get away from SlotFills and use blocks for this. For the required blocks, we want to:
- Force insert them into existing Checkout blocks (via plugin_activate, plugin_deactivate hooks), or take the user to the checkout editing page and show a warning to update the template.
- We need to have inner block dynamic templates to support those new blocks.

A developer can decide where should his block live (by inserting it on the block template array) and he can lock his block using the locking ability we developed to hold that block in place.

[...]

PR #4504, opened 2021-07-23
Checkout i2: Template Syncing to force block usage

mikejolley:
This is an experiment to solve #4447


senadir:
All code is living in #4606 now, plus other stuff. Closing this for now. Thank you for the brilliant code, Mike!

PR #4507, opened 2021-07-26
Fix e2e tests: Remove paypal from fixture data

mikejolley:
Fixes 2 broken tests:

1. e2e - Used PayPal fixture data, [but PayPal is now removed from core for new installs](https://github.com/woocommerce/woocommerce/pull/29971).
2. JS tests - Breakage in withScrollToTop hoc due to recent changes and settimeout usage.

This can merge once tests are passing.


ralucaStan (approved):
👏

assets/js/blocks/cart-checkout/checkout-i2/block.tsx
```diff
@@ -120,9 +120,12 @@ const ScrollOnError = ( {
...
 			showAllValidationErrors();
-			scrollToTop( {
-				focusableSelector: 'input:invalid, .has-error input',
-			} );
+			// Scroll after a short timeout to allow a re-render. This will allow focusableSelector to match updated components.
+			setTimeout( () => {
+				scrollToTop( {
+					focusableSelector: 'input:invalid, .has-error input',
+				} );
+			}, 50 );
```
nerrad, line 128: Should return `clearTimeout` from the effect to prevent memory leak on unmount.

nerrad (changes requested):
Thanks for doing this Mike! There is one comment I've left that will need addressed.

PR #4508, opened 2021-07-26
Update documentation to describe updated snackbar filter

opr:
In #4417 I updated how the snackbar filter works. This PR updates the related documentation and adds an example of how the filter can be used.

1. Please read the documentation and ensure it makes sense and reads well.
2. Feel free to try out the example if you like! I would recommend trying to filter notices based on a specific coupon value.

### Changelog

> Update documentation for the snackbarNoticeVisibility filter.

### DIFF
--- a/docs/extensibility/available-filters.md
+++ b/docs/extensibility/available-filters.md
@@ -127,26 +127,24 @@ There is a snackbar at the bottom of the page used to display notices to the cus
 
 <img src="https://user-images.githubusercontent.com/5656702/120882329-d573c100-c5ce-11eb-901b-d7f206f74a66.png" width=300 />
 
-It may be desirable to hide this (by removing it from the array) or to change the text in the notice.
+It may be desirable to hide this if there's a notice you don't want the shopper to see.
 
 | Filter name  | Description | Return type  |
 |---|---|---|
-| `snackbarNotices`  | An array of notices waiting to be shown in the snackbar | `SnackbarNotice[]`
+| `snackbarNoticeVisibility`  | An object keyed by the content of the notices slated to be displayed. The value of each member of this object will initially be true. | `object`
 
-These are the relevant members of a `SnackbarNotice` object.
+The filter passes an object whose keys are the `content` of each notice.
 
-```typescript
-SnackbarNotice {
-  content: string;
-  explicitDismiss: boolean;
-  icon: string | null;
-  isDismissable: boolean;
-  onDismiss: Function;
-  spokenMessage: string;
-  status: string;
-  type: string
+If there are two notices slated to be displayed ('Coupon code "10off" has been applied to your basket.', and 'Coupon
+code "50off" has been removed from your basket.'), the value passed to the filter would look like so:
+
+```javascript
+{
+  'Coupon code "10off" has been applied to your basket.': true, 
+  'Coupon code "50off" has been removed from your basket.': true
 }
 ```
+To reiterate, the _value_ here will determine whether this notice gets displayed or not. It will display if true.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


docs/extensibility/available-filters.md
```diff
@@ -127,26 +127,24 @@ There is a snackbar at the bottom of the page used to display notices to the cus
...
-  content: string;
-  explicitDismiss: boolean;
-  icon: string | null;
-  isDismissable: boolean;
-  onDismiss: Function;
-  spokenMessage: string;
-  status: string;
-  type: string
+If there are two notices slated to be displayed ('Coupon code "10off" has been applied to your basket.', and 'Coupon
+code "50off" has been removed from your basket.'), the value passed to the filter would look like so:
```
nielslange, line 139: The break between lines 138 and 139 looks to haven been added accidentally. (That said, when looking closer, I see that other sections also show line breaks at places where they should not belong.)

docs/extensibility/available-filters.md
```diff
@@ -243,6 +241,26 @@ __experimentalRegisterCheckoutFilters( 'automatic-coupon-extension', {
 |---|---|
 | <img src="https://user-images.githubusercontent.com/5656702/123768988-bc55eb80-d8c0-11eb-9262-5d629837706d.png" /> | ![image](https://user-images.githubusercontent.com/5656702/124126048-2c57a380-da72-11eb-9b45-b2cae0cffc37.png) |
 
+### Hide a snackbar notice containing a certain string
+Let's say we want to hide all notices that contain the string `auto-generated-coupon`.
+
+We would do this by setting the value of the `snackbarNoticeVisibility` to false for the notices we want to hide.
```
nielslange, line 247: **nitpic** Coming from support, I tend not to use pronouns (_we_, _you_, etc.), but would write instructions neutral.  Instead of 

> Let's say we want to hide all notices that contain the string `auto-generated-coupon`.
> We would do this by setting the value of the `snackbarNoticeVisibility` to false for the notices we want to hide.

I'd write

> To hide all notices that contain the string `auto-generated-coupon`, the value of the `snackbarNoticeVisibility` need to be set to false, for the notices that should be hidden.

That said, I see that other sections use a similar wording. Hence, my prefix _nitpic_.

nielslange (dismissed):
LGTM and I just pointed out two minor suggestions, of which one is a bit nitpicking.

docs/extensibility/available-filters.md
opr: Thanks Niels, this is a good thing to keep in mind! I'll update these instances in this PR, but not the others. Instead, I will create a cooldown issue to review documentation and re-write it in a neutral way without using personal pronouns.

docs/extensibility/available-filters.md
opr: I added the linebreaks there because my IDE indicates a "maximum line length" of 120
Here's what I see
image

The thin line on the right is the 120 character mark. (Tables have to exceed it or their formatting breaks)

nielslange: > I added the linebreaks there because my IDE indicates a "maximum line length" of 120

Ah, I see what you mean. For source code, I'd stick to the "maximum line length", while for *.md files, I'd ignore it and keep paragraphs in one line, in case one if viewing the *.md file in a different reader, e.g. in an IDE. That said, when viewing the *.md file in GitHub, no strange line break is visible. I'll mark this feedback as resolved. 😀

docs/extensibility/available-filters.md
nielslange: That sounds like a good approach. I would also only touch the section I highlighted and not any other section as this would exceed the scope of this change. I'll mark this feedback as resolved as well.

mikejolley:
I notice this is RTM but hasn't been. Is it awaiting some more changes?

Issue #4138, opened 2021-04-27
Create Mini Cart block


Aljullu:
We want to create a Mini Cart block which can be used as a replacement of the Cart widget and also as a Mini Cart link that can be added inside the Navigation block and which shows a dropdown of the cart contents on hover.

## Screenshots

**Current Mini Cart in Storefront**

![Mini Cart in Storefront](https://user-images.githubusercontent.com/3616980/116245294-1de6c600-a769-11eb-8724-9803b62c894d.png)

**Designs of the future Mini Cart block**

<img width="1440" alt="Mini Cart block design" src="https://user-images.githubusercontent.com/3616980/129868644-03fe8dfc-6f60-42e4-a287-41b998f5ca96.png">

_(designs are not final yet)_

Aljullu:
Closing this issue given that the block has successfully been released in WC Blocks and WC core.

PR #4510, opened 2021-07-27
Add Mini Cart block in experimental builds

Aljullu:
Part of #4138 and #4245.

This PR adds to the experimental builds an initial version of a MiniCart block which has the following characteristics:

* Only a small JavaScript file is downloaded & parsed on each page where the Mini Cart block is present. That script is used to detect mouse hover/blur on the mini cart button and to preload + append the other scripts.
* Once the page is loaded, all other scripts are preloaded. That means they don’t block rendering the page and, instead, they are downloaded in the background after the page finishes loading.
* Once the user interacts with the Mini Cart, those scripts are appended to the page and, because of that, parsed and executed.

The benefits of this approach are:

* We don't block page rendering to download the scripts required to render the block: scripts are downloaded in the background.
* Files are only parsed and executed when necessary.
* Using the same components and hooks as in the Cart block, we can reuse all the logic. Extensibility points can be reused as well.

This approach was discussed in pca54o-1PL-p2 (conclusions can be found in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4245#issuecomment-899394073).

2. Visit that post or page in the frontend and verify the Mini Cart button appears and, when hovered, Cart contents are rendered.

This PR is an initial step, so don't expect the block to look or work perfectly. UI, interactions, etc. will be improved in subsequent PRs. This epic has some of the next steps we will be working on: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4138.


nerrad:
Besides the non-blocking improvements this PR suggest is possible, would you be able to get some performance metrics using performance tab in Google Chrome (I think you can measure Web Vitals now there too) to get some idea of the impact between non-lazy and lazy loaded?

Aljullu:
I quickly created a new branch that removes the lazy-loading (`add/mini-cart-block-prototype-no-lazy-loading`). Then, I loaded a page containing the Mini Cart block and measured three events: Largest Content Paint, DOMContentLoaded and Load. Results can be seen below (I repeated the test three times on each configuration to get more meaningful data):

With lazy loading:

| Largest Contentful Paint | DOMContentLoaded | Load |
| --- | --- | --- |
| 0,856 | 0,823 | 0,891 |
| 0,820 | 0,776 | 0,845 |
| 0,829 | 0,838 | 0,899 |

Without lazy loading:

| Largest Contentful Paint | DOMContentLoaded | Load |
| --- | --- | --- |
| 1,071 | 1,02 | 1,05 |
| 0,863 | 0,951 | 0,972 |
| 0,911 | 0,985 | 1,01 |

Then, I applied a Network throttling of `Slow 3G` to try to make the differences more apparent:

With lazy loading:

| Largest Contentful Paint | DOMContentLoaded | Load |
| --- | --- | --- |
| 6,819 | 10.41 | 12.84 |
| 6,836 | 10.46 | 12.84 |
| 6,853 | 10,50 | 12,88 |

Without lazy loading:

| Largest Contentful Paint | DOMContentLoaded | Load |
| --- | --- | --- |
| 6,873 | 17,57 | 18,22 |
| 6,840 | 17,53 | 18,16 |
| 6,824 | 17,54 | 18,16 |

What's interesting is that _Largest Contentful Paint_ doesn't seem to change when lazy loading is enabled or not. However, differences in DOMContentLoaded and the load event look quite big.

assets/js/blocks/cart-checkout/mini-cart/preload-script.ts
```diff
@@ -0,0 +1,26 @@
...
+		const prefetchLink = document.createElement( 'link' );
+		prefetchLink.href = version ? `${ src }?${ version }` : src;
+		prefetchLink.rel = 'preload';
+		prefetchLink.as = 'script';
+		prefetchLink.id = `${ handle }-js-prefetch`;
+		document.head.appendChild( prefetchLink );
+	}
+};
+
+export default preloadScript;
```
ralucaStan: This looks like something that could live outside the mini-cart block, especially since we will want to use prefetching where possible.

assets/js/blocks/cart-checkout/mini-cart/lazy-load-script.ts
```diff
@@ -0,0 +1,77 @@
...
+				`#${ handle }-js-before`
+			);
+			if ( handleTranslationsElements.length === 0 ) {
+				const handleTranslations = document.createElement( 'script' );
+				handleTranslations.innerHTML = translations;
+				handleTranslations.id = `${ handle }-js-translations`;
+				document.body.appendChild( handleTranslations );
+			}
+		}
+		// Before
```
ralucaStan: It's not clear what `Before` relates to. It would be helpful if this comment would explain what is happening here.

assets/js/blocks/cart-checkout/mini-cart/lazy-load-script.ts
```diff
@@ -0,0 +1,77 @@
...
+	src,
+	version,
+	after,
+	before,
+	translations,
+}: lazyLoadScriptParams ): Promise< void > => {
+	return new Promise( ( resolve, reject ) => {
+		// Translations
+		if ( translations ) {
+			const handleTranslationsElements = document.querySelectorAll(
```
ralucaStan: is the `-before` suffix something that is used exclusively by translations? I see that both appear in the lazyLoadScriptParams interface, but it's not instantly clear what this API does. Some details about after, before and  translations  param would hep.

assets/js/blocks/cart-checkout/mini-cart/lazy-load-script.ts
```diff
@@ -0,0 +1,77 @@
...
+		);
+		if ( handleScriptElements.length > 0 ) {
+			resolve();
+		} else {
+			const handleScript = document.createElement( 'script' );
+			handleScript.src = version ? `${ src }?${ version }` : src;
+			handleScript.id = `${ handle }-js`;
+			handleScript.onerror = reject;
+			handleScript.onload = () => {
+				// After
```
ralucaStan: This comment isn't also really helpful. I recommend replacing it with something more detailed or removing it

src/BlockTypes/MiniCart.php
```diff
@@ -0,0 +1,250 @@
...
+			return '';
+		}
+		$cart_controller     = new CartController();
+		$cart                = $cart_controller->get_cart_instance();
+		$cart_contents_count = $cart->get_cart_contents_count();
+		$cart_contents       = $cart->get_cart();
+
+		// Force mobile styles.
+		return '<div class="wc-block-mini-cart is-mobile">
+			<button class="wc-block-mini-cart__button">' .
```
ralucaStan, line 190: The button also needs some [a11y aria-attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#associated_aria_roles_states_and_properties), as well as the expanded area. This could be fixed in another PR, as I know this is not the focus of the current one

assets/js/blocks/cart-checkout/mini-cart/frontend.ts
```diff
@@ -0,0 +1,79 @@
...
+			'.wc-block-mini-cart__contents'
+		);
+
+		if ( ! miniCartButton || ! miniCartContents ) {
+			// Markup is not correct, abort.
+			return;
+		}
+
+		const showContents = async () => {
+			miniCartContents.removeAttribute( 'hidden' );
```
ralucaStan, line 52: Here you would need to also handle the [aria attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#associated_aria_roles_states_and_properties) for the expanded/collapsed state.

assets/js/blocks/cart-checkout/mini-cart/frontend.ts
```diff
@@ -0,0 +1,79 @@
...
+	const miniCartBlocks = document.querySelectorAll( '.wc-block-mini-cart' );
+
+	if ( miniCartBlocks.length === 0 ) {
+		return;
+	}
+
+	const dependencies = getSetting(
+		'mini_cart_block_frontend_dependencies',
+		{}
+	) as Record< string, dependencyData >;
```
ralucaStan, line 27: I see that the dependencies list is pretty long and it has things that are already on the page like react, react-dom. I wonder if we could make a list of dependencies to skip. 


```
["lodash", "moment", "regenerator-runtime", "wp-polyfill", "wp-hooks", "wp-i18n", "react", "react-dom", "wp-deprecated", "wp-escape-html", "wp-element", "wc-blocks-registry", "wc-settings", "wp-dom-ready", "wp-a11y", "wp-autop", "wp-blob", "wp-block-serialization-default-parser", "wp-dom", "wp-is-shallow-equal", "wp-keycodes", "wp-priority-queue", "wp-compose", "wp-redux-routine", "wp-data", "wp-html-entities", "wp-primitives", "wp-shortcode", "wp-blocks", "wp-date", "wp-rich-text", "wp-warning", "wc-blocks-checkout", "wp-url", "wp-api-fetch", "wp-data-controls", "wp-notices", "wc-blocks-middleware", "wc-blocks-data-store", "wc-price-format", "wp-plugins", "wp-wordcount", "wc-block-mini-cart-component-frontend"]


```

ralucaStan (changes requested):
Hi Albert, really nice work in this commit. I left some comments, my focus was mostly on the FE part, so I recommend another 👀 for the PHP side.

assets/js/blocks/cart-checkout/mini-cart/preload-script.ts
Aljullu: Good point. Moved to `base-utils` in 81781f74d2c56ae043f840ff9558d9995d176b73.

assets/js/blocks/cart-checkout/mini-cart/lazy-load-script.ts
Aljullu: You're right, I improved the documentation of `lazyLoadScript` and `preloadScript` in 3487ae5adb0f34acd6be5eb7947bc118294cbd3d. Hopefully things should be more clear now. Let me know how it looks. :slightly_smiling_face:

src/BlockTypes/MiniCart.php
Aljullu: Aaah, good point, thanks! Let's work on this in #4608 if that's ok.

assets/js/blocks/cart-checkout/mini-cart/frontend.ts
Aljullu: Something that can be fixed in #4608 as well.

assets/js/blocks/cart-checkout/mini-cart/frontend.ts
Aljullu: Which dependencies would you skip? Things like `react` or `react-dom`, might or might not be already on the page, so we can't safely skip them, no?

assets/js/blocks/cart-checkout/mini-cart/lazy-load-script.ts
Aljullu: Docs improved in 3487ae5adb0f34acd6be5eb7947bc118294cbd3d.

assets/js/blocks/cart-checkout/mini-cart/lazy-load-script.ts
Aljullu: Docs improved in 3487ae5adb0f34acd6be5eb7947bc118294cbd3d.

Aljullu:
Thanks for reviewing this @ralucaStan. I applied your feedback (except a couple of comments that I leave for #4608). Let me know how everything looks now.

assets/js/blocks/cart-checkout/mini-cart/frontend.ts
ralucaStan: My comment was a bit confusing. I was worried that we will load resources twice on the page, so I was suggesting skipping the ones already loaded. I didn't know the already loaded scripts will have ids, like id='react-js', so no change needed here.

ralucaStan (approved):
Thank you for updating the comments. Really nice work! 👏

assets/js/blocks/cart-checkout/mini-cart/frontend.ts
Aljullu: In b10cb4924fa7e49d79376f500f4ae914890e130b, I refactored the `lazyLoadScript()` function a bit to make it easier to understand this part.

PR #4518, opened 2021-07-28
Add inbox notification to surface cart and checkout blocks to select merchants

opr:
This PR will add the code necessary to show an inbox message in WooCommerce Admin's dashboard letting merchants know about the Cart and Checkout blocks.

To enable us to target only a percentage of users (and to keep the users we _do_ target consistent) I have used a `wp_options` entry to give each merchant a randomly generated "probability" of being chosen, this number is between 0-100. When the code runs, if this option doesn't exist in the database, the random number will be added. We can then use this for percentage-based targeting since only 10% of users will have numbers 0-10, 50% will have 0-50 etc.

I have also included a list of extensions that will preclude the merchant from being shown the notification (based on [the list of known incompatible extensions](https://docs.woocommerce.com/document/cart-checkout-blocks-support-status/#section-7)). We need to include `woocommerce-gutenberg-products-block` here too because we don't want to show the notification to merchants already running the feature plugin. At the moment this is commented out for testing purposes.

### Screenshots
image

1. Ensure your store is set to US or UK (or add your country locale to the `ELIGIBLE_COUNTRIES` array in `InboxNotifications.php`
1. Open the WooCommerce Admin dashboard (`/wp-admin/admin.php?page=wc-admin`)
2. Check `wp_options` table and verify an entry has been made for `wc_blocks_surface_cart_checkout_probability` - if the number is above 10, change it to be below 10. (This ensures we only target 10% of users)
2. If you've got any of the incompatible plugins active, then you should not see the notification in your inbox.
3. Disable any incompatible plugins and refresh the admin page, you should see a notification about the Cart and Checkout.
4. To delete the notification, add this code to the top of `create_surface_cart_checkout_blocks_notification` ` Notes::delete_notes_with_name( self::SURFACE_CART_CHECKOUT_NOTE_NAME );`
5. Re-test with a store in an ineligible location, or with incompatible plugins active.

### DIFF
--- a/src/Domain/Bootstrap.php
+++ b/src/Domain/Bootstrap.php
@@ -76,7 +76,7 @@ class Bootstrap {
 		$this->register_payment_methods();
 
 		add_action(
-			'woocommerce_init',
+			'admin_init',
 			function() {
 				InboxNotifications::create_surface_cart_checkout_blocks_notification();
 			},
--- a/src/Domain/Bootstrap.php
+++ b/src/Domain/Bootstrap.php
@@ -74,7 +74,15 @@ class Bootstrap {
 		}
 		$this->register_dependencies();
 		$this->register_payment_methods();
-		InboxNotifications::create_surface_cart_checkout_blocks_notification();
+
+		add_action(
+			'woocommerce_init',
+			function() {
+				InboxNotifications::create_surface_cart_checkout_blocks_notification();
+			},
+			10,
+			0
+		);
 
 		$is_rest = wc()->is_rest_api_request();
 
--- a/src/Domain/Bootstrap.php
+++ b/src/Domain/Bootstrap.php
@@ -5,6 +5,7 @@ use Automattic\WooCommerce\Blocks\AssetsController as AssetsController;
 use Automattic\WooCommerce\Blocks\Assets\Api as AssetApi;
 use Automattic\WooCommerce\Blocks\Assets\AssetDataRegistry;
 use Automattic\WooCommerce\Blocks\BlockTypesController;
+use Automattic\WooCommerce\Blocks\InboxNotifications;
 use Automattic\WooCommerce\Blocks\Installer;
 use Automattic\WooCommerce\Blocks\Registry\Container;
 use Automattic\WooCommerce\Blocks\RestApi;
@@ -73,6 +74,7 @@ class Bootstrap {
 		}
 		$this->register_dependencies();
 		$this->register_payment_methods();
+		InboxNotifications::create_surface_cart_checkout_blocks_notification();
 
 		$is_rest = wc()->is_rest_api_request();
 
--- a/src/InboxNotifications.php
+++ b/src/InboxNotifications.php
@@ -14,7 +14,7 @@ class InboxNotifications {
 
 	const SURFACE_CART_CHECKOUT_NOTE_NAME          = 'surface_cart_checkout';
 	const SURFACE_CART_CHECKOUT_PROBABILITY_OPTION = 'wc_blocks_surface_cart_checkout_probability';
-	const PERCENT_USERS_TO_TARGET                  = 40;
+	const PERCENT_USERS_TO_TARGET                  = 10;
 	const INELIGIBLE_EXTENSIONS                    = [
 		'automatewoo',
 		'mailchimp-for-woocommerce',
@@ -123,7 +123,7 @@ class InboxNotifications {
 		$note->add_action(
 			'learn_more',
 			'Learn More',
-			'https://woocommerce.com'
+			'https://woocommerce.com/woocommerce-checkout'
 		);
 		$note->save();
 
--- a/src/InboxNotifications.php
+++ b/src/InboxNotifications.php
@@ -67,13 +67,6 @@ class InboxNotifications {
 			return;
 		}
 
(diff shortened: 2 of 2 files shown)


src/Domain/Bootstrap.php
```diff
@@ -74,6 +75,15 @@ protected function init() {
...
 
+		add_action(
+			'admin_init',
+			function() {
+				InboxNotifications::create_surface_cart_checkout_blocks_notification();
+			},
+			10,
+			0
+		);
+
```
opr, line 86: Is there a better place to add this?

src/InboxNotifications.php
```diff
@@ -0,0 +1,142 @@
...
+		'woocommerce-conditional-shipping-and-payments',
+		'woocommerce-dynamic-pricing',
+		'woocommerce-eu-vat-number',
+		'woocommerce-follow-up-emails',
+		'woocommerce-gateway-amazon-payments-advanced',
+		'woocommerce-gateway-authorize-net-cim',
+		'woocommerce-google-analytics-pro',
+		'woocommerce-memberships',
+		'woocommerce-paypal-payments',
+		'woocommerce-points-and-rewards',
```
opr:
```suggestion
```

opr: Reminder for me to remove this since Points and Rewards is compatible now :)

src/InboxNotifications.php
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		// Next check the store is located in one of the eligible countries.
+		$raw_country = get_option( 'woocommerce_default_country' );
+		$country     = explode( ':', $raw_country )[0];
+		if ( ! in_array( $country, self::ELIGIBLE_COUNTRIES, true ) ) {
+			return;
+		}
+
+		// Finally, check if the store's generated % chance is below the % of users we want to surface this to.
```
frontdevde, line 117: Would it make sense to check this earlier? If we're excluding 90% of users to start with, we might not want to run the previous checks for those users.

Maybe we could consider changing the order of some of the checks i.e. the ones that exclude the most users first. For example, given the current set of countries, at first glance, it seems like that check would limit the set of eligible users more than the limited list of incompatible extensions. Or if we'd want to prevent a potentially unnecessary DB write we could also consider moving the simple read checks in front of the probability check.

frontdevde (approved):
Tested a host of different scenarios based on the testing instructions and can confirm everything worked as expected.

The inbox notification shows for probabilities below 10 while not showing for probabilities above.

|Probability 8 | Probability 100 |
|-|-|
|Screenshot 2021-09-10 at 17 31 07| Screenshot 2021-09-10 at 17 30 49|

The inbox notification shows for a UK/US-based store while not showing for other countries e.g. Germany.

| UK | Germany |
|-|-|
|Screenshot 2021-09-10 at 17 31 07|Screenshot 2021-09-10 at 17 30 49|

The inbox notification shows when no unsupported extension is active while not showing when one is active.

| No unsupported extensions | mailpoet active |
|-|-|
|Screenshot 2021-09-10 at 17 31 07|Screenshot 2021-09-10 at 17 30 49|

In addition, I also tested if increasing the percentage of users to target works as expected and can confirm it did. 

One thing I noticed is that if the inbox notice shows once and then you switch to a non-eligible country, activate an incompatible plugin, or change the `wc_blocks_surface_cart_checkout_probability` > 10, the inbox notice still continues to show. That doesn't seem ideal but I assume that's how inbox notifications generally work?

Other than that, LGTM 👍

PS: I really appreciated the detailed code comments that made it easy to follow. Keep it up!👏

opr:
> One thing I noticed is that if the inbox notice shows once and then you switch to a non-eligible country, activate an incompatible plugin, or change the wc_blocks_surface_cart_checkout_probability > 10, the inbox notice still continues to show. That doesn't seem ideal but I assume that's how inbox notifications generally work?

According to item 2 in the docs https://github.com/woocommerce/woocommerce-admin/blob/38fbb4aedd0f21cf21ddaa6af93e48564fd35324/docs/examples/activity-panel-inbox.md#inbox-design-constraints-and-best-practices we shouldn't delete notes - How do you feel about breaking this rule in the case that they activate an ineligible plugin? I think keeping the notice after a country change is fine, after all we don't have any actual limits on this anyway.

opr:
Hi @frontdevde thanks for your review!

I have made some changes now:
- Add the `delete_surface_cart_checkout_blocks_notification` which will delete the note.
- Make the "eligible plugin check" the first check to run - if an ineligible plugin is found, then run the delete method. This will cater for the situation where a merchant activates an incompatible plugin. The notice should be deleted in this case.
- Make the percentage check run last, so we can target a more accurate percentage of users who meet the criteria, and also prevent a `wp_options` write if we don't need to.

It would be great if you could take a look at this and give it a 👍🏼  if you agree with the changes!

frontdevde (approved):
Appreciate you acting on the feedback. Retested with activating/deactivating an unsupported extension and LGTM 👍 

|Before|Activating mailpoet|Deactivating mailpoet|
|-|-|-|
|Screenshot 2021-09-13 at 12 30 04|Screenshot 2021-09-13 at 12 31 42|Screenshot 2021-09-13 at 12 32 03|

PR #4522, opened 2021-07-30
Add cart item data to cart and validator cart data

dangtienngoc:
I'm building the Mobile app and using /store/cart API and [Product Add-Ons](https://woocommerce.com/products/product-add-ons/)

**The problem "cart_item_data" is not used also the filter hook does not pass cart item data.**

I try fixed and there are 2 points that need update.

The payload:

image

### DIFF
--- a/src/StoreApi/Routes/CartAddItem.php
+++ b/src/StoreApi/Routes/CartAddItem.php
@@ -87,9 +87,10 @@ class CartAddItem extends AbstractCartRoute {
 		$cart   = $this->cart_controller->get_cart_instance();
 		$result = $this->cart_controller->add_to_cart(
 			[
-				'id'        => $request['id'],
-				'quantity'  => $request['quantity'],
-				'variation' => $request['variation'],
+				'id'             => $request['id'],
+				'quantity'       => $request['quantity'],
+				'variation'      => $request['variation'],
+				'cart_item_data' => $request['cart_item_data'],
 			]
 		);
 
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -204,7 +204,8 @@ class CartController {
 			$this->get_product_id( $product ),
 			$request['quantity'],
 			$this->get_variation_id( $product ),
-			$request['variation']
+			$request['variation'],
+			$request['cart_item_data']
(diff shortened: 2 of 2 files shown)


opr:
Hey @dangtienngoc thanks for opening this Pull Request! At the moment I'm not sure what _exactly_ it solves, so I'm not sure how to review it correctly.

Please could you edit the body of this PR and:
- [ ] write a description of the changes you've made and why,
- [ ] Write a description of what problem this solves (as well as steps to reproduce the problem if necessary) and,
- [ ] a set of testing instructions so that we can verify these changes solve the problem.

Once you've done this I'll be happy to take a look 🚀

senadir:
Hey there @dangtienngoc. Thank you for your PR. Alongside what Thomas said, we also need other things.
This PR changes how the add to cart endpoint works, before fully accepting this, we need to:
- Document this change.
- Make sure it doesn't break existing behaviour (when cart data is not present).

cc @mikejolley as well to give an eye.

dangtienngoc:
@senadir 

Actually, the function Add to cart already take the param cart_item_data also used to generate cart key

image

But in the API `wp-json/wc/store/cart/add-item` with Function handle request not pass the cart_item_data

image

That means when I pass param `cart_item_data` through REST API It will not work.

As I mentioned above I used the plugin https://woocommerce.com/products/product-add-ons/ To validate add-ons options need pass it via the filter.

How it works via the mobile app is a long story.
- I have written a plugin and handle cart sessions to the database only via REST API & cart_key

image

- I'm disabled the nonce check via REST API & cart_key exists. Because everything works in a box

- Finally to checkout I restore the cart from the database.

mikejolley:
The `cart_item_data` param in the `CartController` is for backwards compatibility with core. That's not reason enough to include it as-is within the Store API IMO, and if it were included, we'd likely want to have a more robust shape for it and then map it to the original. Let's get more information as [requested here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4522#issuecomment-889798618) and then we can decide the path forward. Other projects may need something similar to this eventually.

dangtienngoc:
@mikejolley I found the solution, not need to update the Wc block and Woocommerce core.

So I close the pull request

PR #4530, opened 2021-08-02
Use .hidden instead of .screen-reader-text for toggling helper texts

### LINKED ISSUE #4525: Replace `.screen-reader-text` with `.hidden` for elements that are not relevant for screen reader
## Describe the bug

The following sections use the CSS class `.screen-reader-text` to toggle the visibility of helper texts, but these helper texts are not relevant to screen reader:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/20389f280acd3663f22eac3a1c0b132a64e9fd0c/assets/js/editor-components/product-category-control/index.js#L152-L156

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/20389f280acd3663f22eac3a1c0b132a64e9fd0c/assets/js/editor-components/product-tag-control/index.js#L153-L157

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/20389f280acd3663f22eac3a1c0b132a64e9fd0c/assets/js/editor-components/product-attribute-term-control/index.js#L179-L183

## To reproduce

Look up the sections, that I pointed out before.

## Expected behavior

Helper texts, of which the visibility should be toggled, should use the default WP CSS class `.hidden`, which is defined in https://github.com/WordPress/WordPress/blob/a4bdae8122df0685f70a6d637b61750ab193bafa/wp-admin/css/common.css#L166-L175. 

## Current behavior

Currently, helper texts, of which the visibility should be toggled, are using the CSS class `.screen-reader-text `.

## Additional context

https://make.wordpress.org/accessibility/handbook/markup/the-css-class-screen-reader-text/#use states:

> The screen-reader-text class is used:
> 
> - to hide text visually that is primarily needed for screen reader users.
> - for skip links to ensure keyboard navigation visibility when the link gets focus.

nielslange:
Fixes #4525 

#### Accessibility

n/a

### Screenshots

<table>
<tr>
<td valign="top">Screenshot 2021-08-02 at 14 19 28</td>
<td valign="top">Screenshot 2021-08-02 at 14 19 48</td>
<td valign="top">Screenshot 2021-08-02 at 14 20 06</td>
</tr>
</table>

#### Products by Tag block

1. Install and activate the WooCommerce plugin
2. Install and activate this plugin
3. Create min. 2 product tags
4. Create one test product and add the product tags to it
5. Create a test page
6. Add the **Products by Tag** block
7. Select tags
8. When >= 2 product tags are selected, the message 'Pick at least two tags to use this setting.' is visible
9. When < 2 product tags  are selected, the message 'Pick at least two tags to use this setting.' is not visible

#### Products by Category block

1. Install and activate the WooCommerce plugin
2. Install and activate this plugin
3. Create min. 2 product categories with min 
4. Create one test product and add the tags to it
5. Create a test page
6. Add the **Products by Category** block
7. Select product categories
8. When >= 2 product categories are selected, the message 'Pick at least two categories to use this setting.' is visible
9. When < 2 product categories are selected, the message 'Pick at least two categories to use this setting.' is not visible

#### Products by Attribute block

1. Install and activate the WooCommerce plugin
2. Install and activate this plugin
3. Create min. 1 product attribute with min. 2 values
4. Create one test product and add the product attribute to it
5. Create a test page
6. Add the **Products by Attribute** block
7. Select product attribute values
8. When >= 2 product attribute values are selected, the message 'Pick at least two attributes to use this setting.' is visible
9. When < 2 product attribute values are selected, the message 'Pick at least two attributes to use this setting.' is not visible

### Note

This PR only replaces `.screen-reader-text` with `.hidden`, while it does not change the existing functionality. However, the current functionality seems to be implemented the other way around. 

### Changelog

> Replace .screen-reader-text with .hidden for elements that are not relevant to screen readers

### DIFF
--- a/assets/js/editor-components/product-attribute-term-control/index.js
+++ b/assets/js/editor-components/product-attribute-term-control/index.js
@@ -176,11 +176,7 @@ const ProductAttributeTermControl = ( {
 				isHierarchical
 			/>
 			{ !! onOperatorChange && (
-				<div
-					className={
-						selected.length < 2 ? 'hidden' : ''
-					}
-				>
+				<div className={ selected.length < 2 ? 'hidden' : '' }>
 					<SelectControl
 						className="woocommerce-product-attributes__operator"
 						label={ __(
--- a/assets/js/editor-components/product-attribute-term-control/index.js
+++ b/assets/js/editor-components/product-attribute-term-control/index.js
@@ -176,7 +176,7 @@ const ProductAttributeTermControl = ( {
 				isHierarchical
 			/>
 			{ !! onOperatorChange && (
-				<div className={ selected.length < 2 ? 'hidden' : '' }>
+				<div hidden={ selected.length < 2 }>
 					<SelectControl
 						className="woocommerce-product-attributes__operator"
 						label={ __(
--- a/assets/js/editor-components/product-attribute-term-control/index.js
+++ b/assets/js/editor-components/product-attribute-term-control/index.js
@@ -178,7 +178,7 @@ const ProductAttributeTermControl = ( {
 			{ !! onOperatorChange && (
 				<div
 					className={
-						selected.length < 2 ? 'screen-reader-text' : ''
+						selected.length < 2 ? 'hidden' : ''
 					}
 				>
 					<SelectControl
--- a/assets/js/editor-components/product-category-control/index.js
+++ b/assets/js/editor-components/product-category-control/index.js
@@ -149,11 +149,7 @@ const ProductCategoryControl = ( {
 				isSingle={ isSingle }
 			/>
 			{ !! onOperatorChange && (
-				<div
-					className={
-						selected.length < 2 ? 'hidden' : ''
-					}
-				>
+				<div className={ selected.length < 2 ? 'hidden' : '' }>
 					<SelectControl
 						className="woocommerce-product-categories__operator"
 						label={ __(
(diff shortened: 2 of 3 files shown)


Aljullu (approved):
Good catch with this issue @nielslange! The usage of the `screen-reader-text` class here seems very weird and an important accessibility bug.

I'm approving, but suggesting one improvement in case you want to apply it before merging: do you think we should use the [`hidden` HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) instead of the `hidden` CSS class? The attribute seems to have more semantic meaning and it automatically applies `display: none`.

nielslange:
> I'm approving, but suggesting one improvement in case you want to apply it before merging: do you think we should use the [`hidden` HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) instead of the `hidden` CSS class? The attribute seems to have more semantic meaning and it automatically applies `display: none`.

Thanks for your review and approving the suggested change. I wasn't aware of the [`hidden` HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden) and I agree that using this element matches better. I just updated the PR accordingly.

Aljullu (approved):
Thanks for updating this PR @nielslange! LGTM :ship:

PR #4532, opened 2021-08-03
Update Dependencies to Latest Versions supporting React 16x

mikejolley:
This PR updates our dependencies and refreshes locks. This is instead of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4527 which caused tests to fail due to multiple versions of react being present in Peer Dependencies.

- I've updated our `@wordpress` dependencies to the latest versions prior to the React 17x requirement.
- I removed our import of `wordpress-compose` because `withInstanceId` and `useResizeObserver` are available in WP 5.5.
- removed what imports from `wordpress-components` I could based on the existence of components in Gutenberg v7.6.0 which had components package v9.2.0 (this is the version in WP 5.5). I left out direct import of 'wordpress-components' version alone so it's clearer when we can remove that import in the future, and to fix an increase in build size.
- One of the dependencies broke builds because of our use of the nullish coalescing operator so I had to add `@babel/plugin-proposal-nullish-coalescing-operator`
- Updated Storybook to latest version

Closes #4527

- Test full and dev builds
- Due to the increase in package versions (from WordPress), please smoke test block functionality, including legacy blocks, all products block, single products, and cart/checkout.
- Test storybook build and view


mikejolley:
Something is affecting build size negatively, so let's put this on hold until investigated.

mikejolley:
Build issue resolved; it was because of the bump to v13 for 'wordpress-components'.

mikejolley:
@nerrad is this something you'd be able to look at please? I know PRs that touch package.json go out of date fairly quickly.

nerrad:
> removed what imports from wordpress-components I could based on the existence of components in Gutenberg v7.6.0 which had components package v9.2.0 (this is the version in WP 5.5). I left out direct import of 'wordpress-components' version alone so it's clearer when we can remove that import in the future, and to fix an increase in build size.

This decreases the blocks build size but does it increase the total loaded over the wire on the frontend because we're enqueuing the entire `@wordpress/components` external?

nerrad:
> I removed our import of wordpress-compose because withInstanceId and useResizeObserver are available in WP 5.5.

Are we sure the direct import was related to version support? It also takes advantage of tree-shaking on the frontend potentially making the overall size of downloaded assets be less (i.e. not needing to load all of `@wordpress/compose` on the frontend).

nerrad:
> One of the dependencies broke builds because of our use of the nullish coalescing operator so I had to add @babel/plugin-proposal-nullish-coalescing-operator

This might be because IE11 is still supported in the blocks build config.

assets/js/blocks/attribute-filter/block.js
```diff
@@ -227,6 +220,13 @@ const AttributeFilterBlock = ( {
 		]
 	);
 
+	// Track checked STATE changes - if state changes, update the query.
+	useEffect( () => {
+		if ( ! blockAttributes.showFilterButton ) {
+			onSubmit( checked );
+		}
+	}, [ blockAttributes.showFilterButton, checked, onSubmit ] );
```
nerrad, line 228: Odd that this created a changeset. I assume it was spacing changes maybe?

nerrad (commented):
I did some smoke testing and things look okay. My primary concern with this PR is the potential impact to _overall_ KB's loading over the wire on the frontend with both `@wordpress/components` and `@wordpress/compose` being loaded in their entirety as externals vs the tree-shaking we had before the changes. Have you checked to measure that impact?

assets/js/blocks/attribute-filter/block.js
mikejolley: This was using `onSubmit ` before it was defined. I had an error when testing a build that didn't support ie11. We'll need this change in place at some point.

mikejolley:
@nerrad If you look at trunk, you'll notice several of the frontend asset files already require both `wp-compose` and `wp-components`. If these files are already enqueued, there is no reason to tree shake or import directly because it's just duplicate code.

nerrad:
So here's what I discovered:

On `trunk`:

- All Products + filter blocks loads `wp.components`
- Cart Block _does not_ load `wp.components`.
- Checkout Block _does not_ load `wp.components`.

On this branch:

- Cart Block _does_ load `wp.components` 
- Checkout Block _does_ load `wp.components`.

So IMO there's a regression here with about an additional `~411Kb` added to the bytes sent over the wire on routes with the Cart and Checkout block (I was testing with original C&C blocks, not I2).

If I recall correctly, one of the primary motivator of using the `wp-components` alias for things loaded on the frontend was to reduce the impact of loading all of the `wp.components` global because of it's hefty weight. At one point I'm fairly certain All Products block and Filter blocks weren't loading it but appears that at some point we regressed there too.

`@wordpress/compose` is loaded in all cases both in this branch and on `trunk`, so there's no regression for that package usage. However with its impact at ~35kb, it's not as concerning as the change with `wp.components`.

So if we merge as is, this will have significant impact on page load of Cart and Checkout blocks vs what's in trunk currently. I'm not sure I'm comfortable with that. Was the change needed for some other reason that needs to be considered as well?

Also, as an aside, I just did my testing against WordPress 5.8. Have you tested the older blocks against WordPress 5.6 to make sure they work in that context?

assets/js/blocks/attribute-filter/block.js
nerrad: ahhh! I missed the line number change.

mikejolley:
@nerrad My motivation here was that the external was loaded (and cached) anyway, so using it would reduce the size of our built files. If the reality is that wp-components should _never_ load on frontend, I will revert those changes. 

For things like `<SVG>`, do we need to bother importing a component? It looks like it's there for React.Native support.

nerrad:
> For things like <SVG>, do we need to bother importing a component? It looks like it's there for React.Native support.

Uncertain. Currently there's probably not much need given we don't export most of our components for consumption so there's little chance of them being used in a React.Native context. That could potentially be a followup performance enhancement?

mikejolley:
I've reverted the move to `@wordpress/components` and gone through the existing imports in our base directory. Any imports of `@wordpress/components` in the frontend context have been switched to `wordpress-components`. This has fixed the existence of `wp-components` in our frontend assets, so consider this issue resolved.

- Tests are passing.
- I tested 5.6 to ensure the change to `compose` has no negative impact. Things look fine under that version.
- Build size is mostly unaffected.

I'm going to go ahead and merge this before it becomes stale.

PR #4545, opened 2021-08-04
Twenty Twenty: Adjust font styles for filter and product grid blocks

### LINKED ISSUE #2630: Twenty Twenty: force smaller font sizes for filter blocks
**To Reproduce**
1. Add a _All Products_ block + filter blocks.
2. Notice the title font size of the filter blocks is very big. That's because the theme styles are used, which are a little too big in Twenty Twenty.

**Expected behavior**
We can write styles targeting Twenty Twenty with the class `.theme-twentytwenty { ... }`, so we would like to force a smaller font size when 2020 is active.

**Screenshots**
_Editor:_
imatge

_Frontend:_
imatge

nielslange:
Fixes #2630 

#### Accessibility

n/a

### Screenshots

<table>
<tr>
<td>Editor before:
<br><br>

#2630-editor-before
</td>
<td>Editor after:
<br><br>

#2630-editor-after
</td>
</tr>
</table>

<table>
<tr>
<td>Frontend before:
<br><br>

#2630-frontend-before
</td>
<td>Frontend after:
<br><br>

#2630-frontend-after
</td>
</tr>
</table>

1. Install and activate the WooCommerce plugin
2. Install and activate this plugin
3. Create a new page
4. Add the column block (1/3 + 2/3) to the created page
5. Add the _Active Product Filters_, _Filter Products by Price_ & _Filter Products by Attribute_ blocks to the 1/3 column
6. Add the _All Products_ & _Products by Category_ blocks to the 2/3 column
7. Check the font sizes of the elements in the editor
8. Check the font sizes of the elements in the frontend
9. Check out this PR
10. Check the font sizes of the elements in the editor again
11. Check the font sizes of the elements in the frontend again

### Changelog

> Twenty Twenty: Adjust font styles for filter and product grid blocks


Aljullu:
Confirming it fixes the issue and looks much better. However, I'm a bit wary of this kind of selectors: `div[data-type$="-filter"] h3`. They depend on an implementation detail from Gutenberg: what if GB stopped using `data-type` attributes in a future version? And what if we ever had a block whose name also ended in `filter` but we didn't want those styles to apply?

After some time running in circles with this, I can't find an easy way to fix this issue only with CSS, so I think the best solution would be to add a specific class name for the title, and then use that one as our CSS selectors. For example, for the Price Filter block title I would do:

* In [`assets/js/blocks/price-filter/block.js` L54](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/price-filter/block.js#L154) add `className="wc-block-price-filter__title"` to the title:
```diff
-<TagName>{ attributes.heading }</TagName>
+<TagName className="wc-block-price-filter__title">
```
* In [`assets/js/blocks/price-filter/edit.js` L160](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/price-filter/edit.js#L160) add the same class name to the title:
```diff
<BlockTitle
+    className="wc-block-price-filter__title"
     headingLevel={ headingLevel }
```
* Then, I think there is a flaw in the [`BlockTitle`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/editor-components/block-title/index.js#L13) component. It sets the `className` to the `<PlainText>` component, but I don't understand why. IMO that `className` should be applied to `<TagName>`.
* If you do that, you should be able to change your selector from `div[data-type$="-filter"] h3` to `.wc-block-price-filter__title`. This has the benefit that we know this class will always be there and the selector will always work. It also opens the possibility of having the same CSS selector that works in the frontend & the editor components.

:point_up: the same process will need to be done for the Filter Products by Attributes and the Active Filters blocks. I didn't test, but hopefully it will work. :pray:

nielslange:
@Aljullu Thanks for your review, and apologies for my delayed response (due to being AFK). I just went ahead and updated the PR according to your suggestions. Feel free to review it again, when you have time.

Aljullu (commented):
This is looking great, thanks for updating this PR @nielslange! One thing that I would like to see if we can achieve is getting the same class names in the editor and the frontend. I tested the Filter Products by Price block and noticed that the `.wc-block-price-filter__title` is only applied in the editor and not the frontend:

| Editor | Frontend |
| --- | --- |
| imatge | imatge |

In addition to that, IMO `.wc-block-price-filter__title` should be applied to the `<h3>` tag and not the `<textarea>` as it's currently happening. I think if you apply steps 1 and 3 from [my previous comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4545#issuecomment-893573208), you should be able to fix these issues. Let me know if you have any questions or I can help in any way.

If we accomplish that, we would be able to get rid of the `.wc-block-active-filter__fragment` class, which is only present in the editor. And because of that, the markup in the editor and the frontend will be more similar, making things easier for theme devs and ourselves.

nielslange:
> This is looking great, thanks for updating this PR @nielslange! One thing that I would like to see if we can achieve is getting the same class names in the editor and the frontend. I tested the Filter Products by Price block and noticed that the `.wc-block-price-filter__title` is only applied in the editor and not the frontend.

Thanks for your feedback, @Aljullu. I applied your suggestions and ensured that the `h3` elements both in the editor and the frontend are using the same CSS classes. Feel free to review this PR again and let me know if you spot anything else that needs to be adjusted.

assets/css/editor.scss
```diff
@@ -60,3 +60,26 @@
...
+			@include font-size(regular);
+		}
+
+		.wc-block-grid__product-add-to-cart.wp-block-button {
+			.wp-block-button__link {
+				@include font-size(small);
+				padding: em($gap-small);
+				text-decoration: none;
+			}
+		}
```
Aljullu: Do you think we can remove this ruleset? The one in `style.scss` will already have effect in the editor, so it should be fine defining all styles there.

assets/css/editor.scss
```diff
@@ -60,3 +60,26 @@
...
 	padding: 0;
 }
+
+.theme-twentytwenty {
+	.wp-block {
+		h2.wc-block-grid__product-title {
+			@include font-size(large);
+		}
+
+		.wc-block-grid__product-price,
```
Aljullu: Is `.wc-block-grid__product-price` needed here? It didn't seem to make any difference on my end, but maybe I'm missing something?

assets/css/editor.scss
```diff
@@ -60,3 +60,26 @@
 	margin: 0;
 	padding: 0;
 }
+
+.theme-twentytwenty {
+	.wp-block {
+		h2.wc-block-grid__product-title {
+			@include font-size(large);
```
Aljullu: [In `style.scss` font-size is set to `regular`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4545/files#diff-cd886530abfa495dc6782189ca779f1304bb73bc1758e5d3f0968448dfb67abeR221) but here it's `large`. We should try to unify those to use the same.

assets/js/editor-components/block-title/index.js
```diff
@@ -21,22 +21,19 @@ const BlockTitle = ( {
 } ) => {
 	const TagName = `h${ headingLevel }`;
 	return (
-		<TagName>
+		<TagName
+			className={ classnames(
+				'wc-block-editor-components-title',
```
Aljullu: Sorry if I wasn't clear here. I think we should keep `wc-block-editor-components-title` class into the `PlainText` component. That class is used to make the textarea background transparent:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0811719ed486c3952a1e022490b3715766379f77/assets/js/editor-components/block-title/editor.scss#L1-L6

Aljullu (commented):
This is getting closer, thanks for the updates @nielslange!

One new thing we will need to handle in this PR is replicating the changes you did for the Filter Products by Attribute block, into the new Filter Products by Stock block, which was introduced in the last WC Blocks version. This is how it currently looks:
imatge

assets/css/editor.scss
nielslange: > Do you think we can remove this ruleset? The one in `style.scss` will already have effect in the editor, so it should be fine defining all styles there.

Thanks for pointing out that `style.scss` will also affect the editor styles. I removed the section that you mentioned, and the styles in the editor remained the same.

assets/css/editor.scss
nielslange: I am not sure why I added this in the first place, but after removing it, the styles remained the same.

assets/css/editor.scss
nielslange: I added this initially as the product titles in the editor are slightly smaller than the ones in the frontend when using the same font size. For consistency reasons, I'm using the same font size in the editor and frontend. Please let me know if the font size difference between the editor and the frontend is acceptable.

assets/js/editor-components/block-title/index.js
nielslange: > Sorry if I wasn't clear here. I think we should keep wc-block-editor-components-title class into the PlainText component. That class is used to make the textarea background transparent.

My bad as well. I just adjusted this section based on your feedback. The CSS class for the `<Tagname>` will be injected via prop, while the element `<PlainText>` uses the hard-coded CSS class `.wc-block-editor-components-title`.

assets/css/editor.scss
```diff
@@ -60,3 +60,17 @@
 	margin: 0;
 	padding: 0;
 }
+
+.theme-twentytwenty {
+	.wp-block {
+		h2.wc-block-grid__product-title {
```
Aljullu: You will need to remove the `h2` from here. Otherwise, when editing the All Products block and changing the Product Title tag from `h2` to `h3`, the title becomes huge again.

Aljullu (commented):
Thanks for updating this PR @nielslange, this is really close to being ready. I left one comment below, besides that, I would make the changes you did in the Filter Products by Attribute block to the Filter Products by Stock block too. That block was introduced after you created this PR, but I think it would be great updating it in this PR. What do you think?

assets/css/editor.scss
nielslange: Good catch! I removed the `h2` from the CSS selector to affect other heading levels as well.

PR #4548, opened 2021-08-04
Implement memoization in several selectors and hooks

### LINKED ISSUE #3855: Implement memoization for wp.data
While implementing memoization for https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3835 to avoid rerunning the same 3PD code several times, we ended up discovering that our wp.data store is not memorizing its selector and therefore creating new instance several times, up to 13 times in each render cycle.

After asking around, this is the conclusion we had:
- Our selector should never return a raw `[]` or `{}` but should always use `EMPTY_OBJECT` and `EMPTY_ARRAY` (https://github.com/WordPress/gutenberg/blob/4872d1983db04d685de1816a81177caba5a36903/packages/editor/src/store/selectors.js#L48-L64).
Example: instead of
```js
export const getCartErrors = ( state ) => {
	return state.errors || [];
};
```
Do
```js
export const getCartErrors = ( state ) => {
	return state.errors || EMPTY_ARRAY;
};
```
Instead of
```js
export const useStoreCart = ( options = { shouldSelect: true } ) => {
	const { shouldSelect } = options;

	const results = useSelect(
		( select ) => {
			const store = select( storeKey );
			const cartData = store.getCartData();

			return {
				cartItems: cartData.items || [],
				cartItemErrors: cartData.errors || [],
				extensions: cartData.extensions || {},
				shippingRates: cartData.shippingRates || [],
				paymentRequirements: cartData.paymentRequirements || [],
			};
		},
		[ shouldSelect ]
	);
	return results;
};
```
Do
```js
export const useStoreCart = ( options = { shouldSelect: true } ) => {
	const { shouldSelect } = options;

	const results = useSelect(
		( select ) => {
			const store = select( storeKey );
			const cartData = store.getCartData();

			return {
				cartItems: cartData.items || EMPTY_ARRAY,
				cartItemErrors: cartData.errors || EMPTY_ARRAY,
				extensions: cartData.extensions || EMPTY_OBJECT,
				shippingRates: cartData.shippingRates || EMPTY_ARRAY,
				paymentRequirements:
					cartData.paymentRequirements || EMPTY_ARRAY,
			};
		},
		[ shouldSelect ]
	);
	return results;
};
```
- All of our selectors should be memoized to avoid returning new values several times in a row, we do this by making sure our objects don't change at the reducer level, or if the object is constructed on the selector, it is memoized.
- instead of using raw default values inside our selectors, we define those values in a const and use that.
Example: instead of 
```js
/**
 * Retrieves cart meta from state.
 *
 * @param {Object} state The current state.
 * @return {Object} The data to return.
 */
export const getCartMeta = ( state ) => {
	return (
		state.metaData || {
			applyingCoupon: '',
			removingCoupon: '',
		}
	);
};
```
Do
```js
const defaultCartMeta = {
	applyingCoupon: '',
	removingCoupon: '',
};

export const getCartMeta = ( state ) => {
	return state.metaData || defaultCartMeta;
};
```

More reading:
https://riad.blog/2020/02/14/a-journey-towards-a-performant-web-editor/

Aljullu:
Fixes #3855.

This PR implements some techniques to avoid unnecessary re-renders and side effects when using some of our selectors and hooks.

#### For devs (don't include this in the release testing steps)

It's possible to verify that some effects are no longer re-run on each interaction. For example:

1. Add `console.log( 'effect run' );` before [`context/providers/cart-checkout/shipping/index.js` L121](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b8451c90d68be626ba9a41c71a5629a600fe5c72/assets/js/base/context/providers/cart-checkout/shipping/index.js#L121):

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b8451c90d68be626ba9a41c71a5629a600fe5c72/assets/js/base/context/providers/cart-checkout/shipping/index.js#L121

2. Go to the Cart block in the frontend.
3. Increase or decrease the quantity of a product.
4. Verify `effect run` is not logged every time.
5. If you repeat the test in trunk, you will notice `effect run` is logged every time the quantity changes.

#### For users (include this in the release testing steps)

1. Smoke test the Cart and Checkout blocks: try removing products, changing their quantity, modifying shipping address, adding coupons, etc.
2. Verify there are no regressions and everything is working fine.

### Changelog

> Cart and Checkout block performance improvements

### DIFF
--- a/assets/js/base/context/hooks/cart/test/use-store-cart.js
+++ b/assets/js/base/context/hooks/cart/test/use-store-cart.js
@@ -17,6 +17,7 @@ jest.mock( '../../../providers/editor-context', () => ( {
 } ) );
 
 jest.mock( '@woocommerce/block-data', () => ( {
+	...jest.requireActual( '@woocommerce/block-data' ),
 	__esModule: true,
 	CART_STORE_KEY: 'test/store',
 } ) );
--- a/assets/js/base/context/hooks/cart/test/use-store-cart.js
+++ b/assets/js/base/context/hooks/cart/test/use-store-cart.js
@@ -88,6 +88,8 @@ describe( 'useStoreCart', () => {
 		hasCalculatedShipping: true,
 		extensions: {},
 		errors: [],
+		receiveCart: undefined,
+		paymentRequirements: [],
 	};
 	const mockCartTotals = {
 		currency_code: 'USD',
--- a/assets/js/base/context/hooks/cart/use-store-cart.ts
+++ b/assets/js/base/context/hooks/cart/use-store-cart.ts
@@ -84,6 +84,27 @@ const decodeValues = (
 		] )
 	);
 
+// If the value passed is not an object or it's an empty object, return
+// `EMPTY_OBJECT`, otherwise, return the value.
+const getObjectOrEmptyConstant = (
+	val: Record< string, unknown >
+): Record< string, unknown > => {
+	if ( typeof val === 'object' && Object.keys( val ).length > 0 ) {
+		return val;
+	}
+	return EMPTY_OBJECT;
+};
+
+// If the value passed is not an array or it's an empty array, return
+// `EMPTY_ARRAY`, otherwise, return the value.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const getArrayOrEmptyConstant = ( val: any[] ): any[] => {
+	if ( Array.isArray( val ) && val.length > 0 ) {
+		return val;
+	}
+	return EMPTY_ARRAY;
+};
+
 /**
  * @constant
  * @type  {StoreCart} Object containing cart data.
@@ -178,7 +199,7 @@ export const useStoreCart = (
 				? decodeValues( cartData.shippingAddress )
 				: billingAddress;
 			const cartFees =
-				cartData?.fees?.length > 0
+				getArrayOrEmptyConstant( cartData.fees ).length > 0
 					? cartData.fees.map( ( fee: CartResponseFeeItem ) =>
 							decodeValues( fee )
 					  )
@@ -188,7 +209,7 @@ export const useStoreCart = (
 			// the text used to display the coupon, without affecting the
 			// functionality when it comes to removing the coupon.
 			const cartCoupons: CartResponseCouponItemWithLabel[] =
-				cartData?.coupons?.length > 0
+				getArrayOrEmptyConstant( cartData.coupons ).length > 0
 					? cartData.coupons.map(
 							( coupon: CartResponseCouponItem ) => ( {
 								...coupon,
@@ -199,27 +220,27 @@ export const useStoreCart = (
 
 			return {
 				cartCoupons,
-				cartItems: cartData.items || EMPTY_ARRAY,
+				cartItems: getArrayOrEmptyConstant( cartData.items ),
 				cartFees,
 				cartItemsCount: cartData.itemsCount,
 				cartItemsWeight: cartData.itemsWeight,
 				cartNeedsPayment: cartData.needsPayment,
 				cartNeedsShipping: cartData.needsShipping,
-				cartItemErrors:
-					cartData?.errors?.length > 0
-						? cartData.errors
-						: EMPTY_ARRAY,
+				cartItemErrors: getArrayOrEmptyConstant( cartData.errors ),
 				cartTotals,
 				cartIsLoading,
 				cartErrors,
 				billingAddress: emptyHiddenAddressFields( billingAddress ),
 				shippingAddress: emptyHiddenAddressFields( shippingAddress ),
-				extensions: cartData.extensions || EMPTY_OBJECT,
-				shippingRates: cartData.shippingRates || EMPTY_ARRAY,
+				extensions: getObjectOrEmptyConstant( cartData.extensions ),
+				shippingRates: getArrayOrEmptyConstant(
+					cartData.shippingRates
+				),
 				shippingRatesLoading,
 				cartHasCalculatedShipping: cartData.hasCalculatedShipping,
-				paymentRequirements:
-					cartData.paymentRequirements || EMPTY_ARRAY,
+				paymentRequirements: getArrayOrEmptyConstant(
+					cartData.paymentRequirements
+				),
 				receiveCart,
 			};
 		},
--- a/assets/js/base/context/hooks/cart/use-store-cart.ts
+++ b/assets/js/base/context/hooks/cart/use-store-cart.ts
@@ -7,8 +7,15 @@ import { isEqual } from 'lodash';
 import { useRef } from '@wordpress/element';
 import {
 	CART_STORE_KEY as storeKey,
-	EMPTY_ARRAY,
-	EMPTY_OBJECT,
+	EMPTY_CART_COUPONS,
+	EMPTY_CART_ITEMS,
+	EMPTY_CART_FEES,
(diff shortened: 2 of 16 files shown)


assets/js/base/context/hooks/cart/use-store-cart.ts
```diff
@@ -173,39 +198,49 @@ export const useStoreCart = (
...
+							( coupon: CartResponseCouponItem ) => ( {
+								...coupon,
+								label: coupon.code,
+							} )
+					  )
+					: EMPTY_ARRAY;
 
 			return {
 				cartCoupons,
-				cartItems: cartData.items || [],
```
ralucaStan, line 192: Is cartData.items ever `undefined` or `null`? Do we ever set it or is this just a safety measurement? From what I saw in the state we always start with `[]`, so `|| []` is never reached. Which means that this fix here is also never reached.

ralucaStan: Same observation about the changes below regarding cartData.errors, cartData.extensions, cartData.shippingRates and cartData.paymentRequirements.

assets/js/data/cart/selectors.ts
```diff
@@ -46,7 +47,7 @@ export const getCartMeta = ( state: CartState ): CartMeta => {
  * @return {Array<ResponseError>} Array of errors.
  */
 export const getCartErrors = ( state: CartState ): Array< ResponseError > => {
-	return state.errors || [];
+	return state.errors || EMPTY_ARRAY;
 };
```
ralucaStan, line 50: I think using a global EMPTY_ARRAY is really error prone and could affect all the data in the store, as it could easily be mutated. If in the end it turns out that this is the way to go, I would recomment using something like EMPTY_CART_ERRORS instead of EMPTY_ARRAY:

```
const EMPTY_CART_ERRORS = [];

export const getCartErrors = ( state: CartState ): Array< ResponseError > => {
	return state.errors || [];
	return state.errors || EMPTY_CART_ERRORS;
};
```

- a localized variable, instead of a global one. We don't export it and there is no danger of sharing it among data.
- a specific variable for a specific data branch

ralucaStan (commented):
From what I understood from @Aljullu, _when we return an empty array from our selectors, any component that is subscribed to them and uses them as a useEffect() or useCallback() dependency, will be running that effect or callback on every render (because [] === [] //false)._

I am curious why we need to return empty [] or {}. From what I noticed our store has all the cart data already initialized to empty values, like '', [] or {}.
For example, when we select the data `cartData.errors || []`, cartData.errors is not `undefined` nor `null`, so that empty array is never reached. 

So I need a bit of context of why we actually need to return in our selectors || [] or || {}, because maybe the problem lies somewhere else, in the place where we actually set a cart data to null/undefined instead of resetting it to the default cart value for that data,  like '', [] or {}.

When it comes to using this EMPTY_OBJECT solution I think we should be really careful, because this introduces a high degree of risk. It's a global variable, shared across multiple data with a high risk of mutation. I see a lot of places where EMPTY_OBJECT is used and this could also be something that we might end up doing without really understanding the need.  I presented in a comment a solution with something localized, more specific see EMPTY_CART_ERRORS. 



Lastly, I want us to comment as a team about what we are trying to solve here. Do we have a really big performance issue that is worth bringing this amount of unpredictability into the project?

assets/js/base/context/hooks/cart/use-store-cart.ts
Aljullu: That's a good point... I investigated this and I think those default values were only needed for tests to pass, which means that the issue was in the tests configuration and not in the code itself.

I made some changes in `assets/js/base/context/hooks/cart/test/use-store-cart.js` and `assets/js/previews/cart.ts` and now tests pass without requiring those defaults values to exist.

assets/js/data/cart/selectors.ts
Aljullu: I changed all `EMPTY_ARRAY` and `EMPTY_OBJECT` instances for localized variables, that means there are many more variables now, but I think it's ok. :ok_hand:

Aljullu:
@ralucaStan I updated this PR addressing your feedback, let me know how it looks now.

> Lastly, I want us to comment as a team about what we are trying to solve here. Do we have a really big performance issue that is worth bringing this amount of unpredictability into the project?

I don't think there is a performance issue we want to fix specifically, I just think this is a good practice to implement even if it doesn't have a real performance impact. Another thing to consider is that some of those values might be used by extensions, so even if in WC Blocks there isn't a performance issue, there might be in 3rd-party extensions relying on those values.

assets/js/data/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const EMPTY_CART_COUPONS: [  ] = [];
+export const EMPTY_CART_ITEMS: [  ] = [];
+export const EMPTY_CART_FEES: [  ] = [];
+export const EMPTY_CART_ITEM_ERRORS: [  ] = [];
+export const EMPTY_CART_ERRORS: [  ] = [];
+export const EMPTY_SHIPPING_RATES: [  ] = [];
+export const EMPTY_PAYMENT_REQUIREMENTS: [  ] = [];
+export const EMPTY_EXTENSIONS: Record< string, unknown > = {};
+export const EMPTY_TAX_LINES: [  ] = [];
```
ralucaStan, line 16: 👏 this actually triggers a warning when trying to modify any of this values

assets/js/data/default-states.ts
```diff
@@ -16,12 +27,15 @@ export interface CartState {
 	errors: Array< ResponseError >;
 }
 
+export const EMPTY_PENDING_QUANTITY = [];
```
ralucaStan: Let's type these as well.
```
export const EMPTY_PENDING_QUANTITY : [ ] = [];
export const EMPTY_PENDING_DELETE : [ ] = [];
```

assets/js/base/context/hooks/cart/test/use-store-cart.js
```diff
@@ -17,6 +17,7 @@ jest.mock( '../../../providers/editor-context', () => ( {
 } ) );
 
 jest.mock( '@woocommerce/block-data', () => ( {
+	...jest.requireActual( '@woocommerce/block-data' ),
```
ralucaStan, line 20: What is the purpose of this? Is this something we should do for other mocks in our tests in another PR, I see not all of them use them.

assets/js/base/context/hooks/cart/use-store-cart.ts
ralucaStan: Congrats on investigating this.

ralucaStan (approved):
I've left a few comments and requested a small change. After that is addressed this could be merged from my side

assets/js/base/context/hooks/cart/test/use-store-cart.js
Aljullu: If this line wasn't there, any exported constant that wasn't mocked, would be undefined. In this case, all new `EMPTY_XYZ` arrays would be undefined because they are not explicitly added to the mock.

I'm not sure if there is any drawback of doing this change to all other mocks, but it makes tests to be closer to the way production code runs, so in general I think it's better to have it, even if not needed.

assets/js/data/default-states.ts
Aljullu: Good catch! This is done in 8282985fbfea5dd37babd55694cc36121ce8e527.

PR #4551, opened 2021-08-05
Fix unfiltered html usage of blocks

### LINKED ISSUE #4181: Audit Blocks for dependency on `unfiltered_html` capability attached to the user that inserts and saves content with the block (and work on alternatives)
We have some blocks that add html to the `post_content` of a page or post that is stripped out by WordPress unless the user role has `unfiltered_html` capabilities. An example of a block doing this is the Product Search block. 

This issue has the goal of accomplishing the following:

## Audit our blocks for which are affected.

A quick way to do this is to create a user with the "Author" WordPress role and use that user to add all blocks to a page. Go to the frontend and see which blocks don't render/behave as expected.

## Plan how we'll resolve this.

There's a couple approaches that immediately come to mind (there may be more). Also, note that it's possible different approaches might be needed across the blocks that come up in the audit.

### Option A: Only make affected blocks available to users editing content with `unfiltered_html` capability.

This probably would be the most trivial to implement but some significant cons are:

- it removes the availability of these blocks for users on sites/stores with reduced capabilities for security reasons.
- it'd be fairly easy for someone without the `unfiltered_html` capability to edit a post that _was_ saved with the block by a user with the right capabilities and bork the content as a result. This would also be an initially hard thing to troubleshoot what went wrong.

### Option B: Change what we save to the post content to be placeholder and hook in server side or client side to dynamically add any additional html usually stripped out by `wp_kses_post` on post save.

While this would be a bit more effort to implement, it would have the benefit of retaining usage of the blocks to _anyone_ on the site that has publishing capabilities. The downside with this approach (at least in the case of server side rendering only) is that the user experience in the editor will be degraded a bit with using this type of block and/or potentially having multiple places to track/preserve layout changes (vs using the saved `post_content` only).

mikejolley:
This fixes the unfiltered HTML issue for the product search block by using server side rendering on the frontend. The edit view remains as-is to allow greater flexibility in editing.

Fixes #4181

### Screenshots

Screenshot 2021-08-05 at 13 10 35

1. Add `define( 'DISALLOW_UNFILTERED_HTML', true );` to wp-config.php, or use a user account with author role.
2. Add the product search block to a page and save.
3. View the search block on the frontend. Confirm it works as intended.

To test migration,

1. Before using this patch, create and save a search block as an admin user,
2. Checkout this PR and build.
3. Load the post. Confirm there is no invalid block error.

### Changelog

> Add server side rendering to search block so the block can be used by non-admins.

### DIFF
--- a/src/BlockTypes/ProductSearch.php
+++ b/src/BlockTypes/ProductSearch.php
@@ -79,8 +79,7 @@ class ProductSearch extends AbstractBlock {
 					array_filter(
 						[
 							'wc-block-product-search',
-							$attributes['align'] ? 'align' + $attributes['align'] : '',
-							$attributes['className'],
+							$attributes['align'] ? 'align' . $attributes['align'] : '',
 						]
 					)
 				),
--- a/src/BlockTypes/ProductSearch.php
+++ b/src/BlockTypes/ProductSearch.php
@@ -31,6 +31,19 @@ class ProductSearch extends AbstractBlock {
 	 * @return string Rendered block type output.
 	 */
 	protected function render( $attributes, $content ) {
+		static $instance_id = 0;
+
+		$attributes = wp_parse_args(
+			$attributes,
+			array(
+				'hasLabel'    => true,
+				'align'       => '',
+				'className'   => '',
+				'label'       => __( 'Search', 'woo-gutenberg-products-block' ),
+				'placeholder' => __( 'Search products…', 'woo-gutenberg-products-block' ),
+			)
+		);
+
 		/**
 		 * Product Search event.
 		 *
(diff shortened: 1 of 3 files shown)


src/BlockTypes/ProductSearch.php
```diff
@@ -57,6 +70,59 @@ protected function render( $attributes, $content ) {
...
+		$input_id           = 'wp-block-search__input-' . ( ++$instance_id );
+		$wrapper_attributes = get_block_wrapper_attributes(
+			array(
+				'class' => implode(
+					' ',
+					array_filter(
+						[
+							'wc-block-product-search',
+							$attributes['align'] ? 'align' + $attributes['align'] : '',
+							$attributes['className'],
```
ralucaStan: I have tested with a custom class and the class appears twice in the dom: 
`<div class="wc-block-product-search test-my-class test-my-class wp-block-woocommerce-product-search">`

ralucaStan: I have tried to change the alignment and when I pick Wide width a PHP error appears on the frontend in Query monitor `A non-numeric value encountered`:

```
wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTypes/ProductSearch.php:82
	.	Automattic\W\B\B\ProductSearch->render() wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTypes/AbstractBlock.php:85
	.	Automattic\W\B\B\AbstractBlock->render_callback() wp-content/plugins/gutenberg/lib/compat.php:132
	.	{closure}() wp-includes/class-wp-block.php:221
	.	WP_Block->render() wp-includes/blocks.php:864
	.	render_block() wp-includes/blocks.php:902
	.	do_blocks() wp-includes/class-wp-hook.php:303
	.	apply_filters('the_content') wp-includes/post-template.php:253
	.	the_content() wp-content/themes/storefront/inc/storefront-template-functions.php:352
	.	storefront_page_content() wp-includes/class-wp-hook.php:303
	.	do_action('storefront_page') wp-content/themes/storefront/content-page.php:18
	.	load_template('wp-content/themes/storefront/content-page.php') wp-includes/template.php:716
	.	locate_template() wp-includes/general-template.php:204
	.	get_template_part('content','page') wp-content/themes/storefront/page.php:24
```

ralucaStan (commented):
I have tested and it works as expected. I added a couple of comments with what I found while testing. I also recommend another 👁️ for the PHP side.

src/BlockTypes/ProductSearch.php
mikejolley: This is what happens when we context switch between php and JS :D

mikejolley:
Thanks @ralucaStan, all fixed pending your approval.

ralucaStan (approved):
Works as expected

PR #4561, opened 2021-08-09
Call validation before submitting the order

### LINKED ISSUE #4495: Form autofill doesn't work for Checkout block.
## Describe the bug
When auto-filling the checkout form and trying to place an order, you will get an error that you need to fill out fields "Please fill out this field.", despite them being filled. This persist unless you fill all fields manually.

## To reproduce
0. On a browser, make sure you have a saved address (chrome, edge, safari, firefox all support this).
1. On a private window, add an item to cart and go to checkout.
2. Click on a field and autofill it.
3. Try to place an order
4. You will get an error.

## Expected behavior
Your order should be placed with your informations.

## Screenshots

https://user-images.githubusercontent.com/6165348/126644140-0ed5359a-e737-4e88-b364-1cecebbfbeb5.mov

I tested on both Safari and Edge (chome based) on Mac.

senadir:
This PR fixes the autofill issue, see #4495 for what's happening.
TL;DR, we only run validation on mount and blur, autoFill doesn't trigger blur, meaning we didn't run any validation. Additionally, all inputs mount with an error if empty, meaning when we submit an order, errors aren't cleared.

This PR runs validation on `isBeforeProcessing`.

Fixes #4495

1. On a private window, autoFill the form.
2. Submit the order, it should pass fine.
3. Again, on a private window, autoFill the order.
4. Mess up the email field or postcode with an invalid value.
5. Try submitting, make sure it shows the error.
6. Fix the field, make sure there is no jumping around.
7. Submit again, it should pass.

### Changelog

> Fix autofill triggering validation errors for valid values in Checkout block

### DIFF
--- a/assets/js/base/components/text-input/validated-text-input.tsx
+++ b/assets/js/base/components/text-input/validated-text-input.tsx
@@ -117,6 +117,9 @@ const ValidatedTextInput = ( {
 		}
 	}, [ isPristine, setIsPristine, validateOnMount, validateInput ] );
 
+	/**
+	 * @todo Remove extra validation call after refactoring the validation system.
+	 */
 	useEffect( () => {
 		if ( isBeforeProcessing ) {
 			validateInput();
--- a/assets/js/base/components/text-input/validated-text-input.tsx
+++ b/assets/js/base/components/text-input/validated-text-input.tsx
@@ -65,7 +65,7 @@ const ValidatedTextInput = ( {
 		getValidationErrorId,
 	} = useValidationContext();
 
-	const { onCheckoutValidationBeforeProcessing } = useCheckoutContext();
+	const { isBeforeProcessing } = useCheckoutContext();
 
 	const textInputId =
 		typeof id !== 'undefined' ? id : 'textinput-' + instanceId;
@@ -118,10 +118,10 @@ const ValidatedTextInput = ( {
 	}, [ isPristine, setIsPristine, validateOnMount, validateInput ] );
 
 	useEffect( () => {
-		const unsubscribe = onCheckoutValidationBeforeProcessing( () => validateInput() );
-		return unsubscribe;
-	}, [ onCheckoutValidationBeforeProcessing, validateInput ] );
-
+		if ( isBeforeProcessing ) {
+			validateInput();
+		}
+	}, [ isBeforeProcessing, validateInput ] );
 	// Remove validation errors when unmounted.
 	useEffect( () => {
 		return () => {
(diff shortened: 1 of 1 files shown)


senadir:
@nerrad what are your concerns on having `useCheckoutContext` being used in the input field? It would cause issues if we used this input outside Cart and Checkout.

ralucaStan:
> @nerrad what are your concerns on having `useCheckoutContext` being used in the input field? It would cause issues if we used this input outside Cart and Checkout.

I don't think we should have business logic inside the input component, especially if we are looking at exporting them at some point. Because there is no event we could catch on autofill, I'm thinking we could we have something like a `forceValidation` prop, which reruns validation once the prop value is true?

nerrad:
> I don't think we should have business logic inside the input component, especially if we are looking at exporting them at some point. Because there is no event we could catch on autofill, I'm thinking we could we have something like a forceValidation prop, which reruns validation once the prop value is true?

I agree with Raluca here. If there's need to force revalidation via some logic, it makes sense to abstract the logic outside of the input and use props to trigger that re-validation.

senadir:
I agree that we should have some sort of external trigger that calls validation, but having a prop that forces validation doesn't make sense to me.
This prop feels ephemeral to me and doesn't make sense as a prop.
Being a declarative UI, how should a component like this behave? `<ValidatedInput value={value} onChange={onChange} forceRevalidate={ true } />`
Should this keep revalidating on each mount, should it revalidate on mount, on what level? `forceRevalidate` isn't an event to react to, and isn't a static prop, it feels very imperative thing.

I might be missing something, but do we have other examples or precedents of using such pattern?

IMHO, we should probably be rethinking of how we do validation to make it more accessible. We store validation errors outside components yet can decide if we can delete them inside the component, which is already a lot of coupling. Libraries like formik have the validation schema (function) that is accessible from the from level. This allows you to run validation inside the component reacting to certain events like onChange or onBlur, and also outside the component on the form level on demand or at certain lifecycle events.

Such workaround should be currently acceptable until we find a better refactor. Having an extra `forceValidation` would only cause us to go even deeper into the validation issue.

I did attempt to add that prop to the component and it caused significant prop drilling since inputs are really deep into the tree.

ralucaStan:
@senadir if we could move the logic at the form level it would be great. 

I agree that the input should just receive info about the state it should render (error, valid, etc) and the form is the place where the state should be available.

mikejolley (approved):
I tested this using an invalid email address and just autofill and both cases worked as described. Using Edge in private mode.

PR #2756, opened 2020-06-19
Add more E2E tests (WIP)

senadir:
This is a WIP, I struggled to get this rebased, it's a really old branch, finishing it right now is not a super-priority.

PR #4562, opened 2021-08-10
Add E2E tests for Filter blocks (Price, Attribute, and Active Filters)

### LINKED ISSUE #4643: Add an e2e test to all blocks to verify they can be added in the editor
In order to rely more on automatic testing and less on manual smoke testing every time we do a release, we should add a test to all blocks that verifies that the block is available and can be added in the editor. Similar to this one:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/42beb9b1a4166d279f5dff3ec737542b4dfde0ac/tests/e2e/specs/backend/product-search.test.js#L27-L29

In the docs, you can read more about [e2e testing](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/testing/when-to-employ-e2e-testing.md) and learn [how to run our e2e tests](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/contributors/javascript-testing.md#how-to-run-end-to-end-tests).

## Status

List of blocks that include a test like this:
* [x] Featured Product Block (#4771)
* [x] Featured Category Block (#4773)
* [x] Hand-picked Products Block (#4788)
* [x] Best Selling Products Block (#4785)
* [x] Top Rated Products Block (#4786)
* [x] Newest Products Block (#4789)
* [x] On Sale Products Block (#4872)
* [x] Products by Category Block (#4872)
* [x] Products by Tag Block (#4872)
* [x] Products by Attribute Block (#4872)
* [x] Product Categories List Block (#4915)
* [x] Reviews by Product (#2808)
* [x] Reviews by Category (#2808)
* [x] All Reviews (#2799)
* [x] Product Search (#2707)
* [x] All Products (#1716)
* [x] Filter Products by Attribute (#4562)
* [x] Filter Products by Price (#4562)
* [x] Filter Products by Stock (#4682)
* [x] Active Product Filters (#4562)
* [x] Cart (#2707)
* [x] Checkout (#2707)
* [x] Mini Cart (#4677)
* [x] Single Product (#1240)

opr:
Closes #4643

This PR will introduce E2E tests for the `Filter Products by Price` block.

The changes included are:
- Add a class to the `div` containing the Filter Products by Price block in the editor (`wp-block-woocommerce-price-filter`). Previously there was not one so our testing utils that rely on class names couldn't target the block.
- Add a class to the `div` containing the Filter Products by Attribute block in the editor (`wc-block-attribute-filter`).
- Add [`pptr-testing-library`](https://www.npmjs.com/package/pptr-testing-library) so we can use the queries provided by it.
- Add `clearAndFillInput` util so we can type into input boxes.
- Add attributes (Capacity and Shade) to the E2E test store.
- Create two new products that use the Capacity attribute.
- Ensure attributes are added correctly to the product by ID (see https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4562/files#diff-2fccf9f205d4c46e7393da8019992bb5bd730aa2a3e854420bba4cdf661ee3ffR362-R398) - This creates the attributes in the store then returns their IDs. We can't know ahead of time what the ID will be so it needs to be fetched from the API after creation.
- Add tests for `Active Filters`, `Filter Products by Price`, and `Filter Products by Attribute`.

See #2395 huge thanks to @senadir for the initial work done in that PR 🙌🏼 

1. Ensure the Filter Products by Price block works in the editor and on the front-end still.
2. Ensure the Filter Products by Attribute block works in the editor and on the front-end still.
2. Run the E2E tests - verify all pass.
3. Wait for tests to go green here ✅ 

### Changelog

> Add end-to-end tests to cover the `Active Filters`, `Filter Products by Attribute`, and `Filter Products by Price` blocks.


tests/e2e/specs/backend/active-filters.test.js
```diff
@@ -0,0 +1,77 @@
...
+	it( 'can only be inserted once', async () => {
+		await insertBlockDontWaitForInsertClose( block.name );
+		expect( await getAllBlocks() ).toHaveLength( 1 );
+	} );
+
+	it( 'renders without crashing', async () => {
+		await expect( page ).toRenderBlock( block );
+	} );
+
+	it( 'allows title can be manipulated', async () => {
```
ralucaStan: This test description could be changed to `allows changing the block's title`

tests/e2e/specs/backend/active-filters.test.js
```diff
@@ -0,0 +1,77 @@
...
+		await openDocumentSettingsSidebar();
+		await clearAndFillInput(
+			'.wp-block[data-type="woocommerce/active-filters"] textarea.wc-block-editor-components-title',
```
ralucaStan: I love that we now have the pptr-testing-library that allows us to search for content from a user's point of view. With Testing library I know that it's encouraged to search the page for texts the user sees, so I wonder if we could drop searching by this selector and use 

`$document.findByPlaceholderText('Active filters')`

tests/e2e/specs/backend/active-filters.test.js
```diff
@@ -0,0 +1,77 @@
...
+			'Active filters'
+		);
+		await page.click(
+			'.components-toolbar button[aria-label="Heading 3"]'
+		);
+	} );
+
+	it( 'Display style to change', async () => {
```
ralucaStan: `allows changing the Display Style`?

ralucaStan (commented):
Congrats on this PR, your efforts will benefit the project greatly. 

I have added some small comments. 
I salute the addition of pptr-testing-library and bringing us closer to some of Testing library's tools. 

I think that moving forward we should also try to test according to React Testing Library's testing principles. I think this PR would be a nice starting point and showcase of using them.

I recommend to try and change in the new tests you've added selecting elements by css selectors and using pptr-testing-library selectors instead (getByText, findByLabelText, etc...) that allow selecting from an user point of view.

tests/e2e/specs/backend/active-filters.test.js
opr: Hmm, I agree it would be _better_, but I'm not confident this is a good time to start using that in place of what we're doing now with selectors. My reasoning is: if we go ahead with rewriting the `clearAndFillInput` method, we will need to add some kind of fallback for when it's not possible to select something by these "text" values from `pptr-testing-library`. This then means we're juggling two things, puppeteer `elementHandles` and dom-testing-library's `HTMLElement`. Also, all tests using this will need to be rewritten to keep track of the current text. This is because some inputs don't have placeholders, or labels, see Filter by Price title, it's a `textarea` with no label or placeholder, so we'll need to know the text beforehand, then keep track of the the text it changes to.

I apologise if I've misunderstood the way nodes/elements are returned by both of these libraries, please correct me if I'm thinking about this wrong.

On top of this, based on what I read here (https://testing-library.com/docs/example-input-event/) we wouldn't be able to simulate typing into an input, which is sometimes required, because changing the `.value` property of a node isn't enough, in the case of the `Filter Products by Price` title, setting `.value` doesn't persist the new title, it has to be typed "manually".

ralucaStan: Maybe I got a bit carried away because I am a huge fan of the [Testing library's principles](https://testing-library.com/docs/), moving away from testing implementation details (aka search for nested selectors) and, instead, test using what the user sees (texts, labels, etc).

It wasn't fully clear to me why `pptr-testing-library` was added. You mentioned making use of its queries and I thought that would be the default way from now on. I think it would be helpful to explain, maybe in a P2:
- what are the reasons behind adding it
- when this library should be used
- how do we see the adoption phase

I did not know about the limitations around typing so thanks for clearing those out. We can keep `clearAndFillInput` as it is.

tests/e2e/specs/backend/active-filters.test.js
```diff
@@ -0,0 +1,77 @@
...
+			{ text: 'New Title' }
+		);
+
+		expect( await getEditedPostContent() ).toMatchSnapshot();
+		await clearAndFillInput(
+			'.wp-block[data-type="woocommerce/active-filters"] textarea.wc-block-editor-components-title',
+			'Active filters'
+		);
+		await page.click(
+			'.components-toolbar button[aria-label="Heading 3"]'
```
ralucaStan: It is not clear what the test is clicking here. Maybe a comment above could explain this step. Also, this would be like a good use case for searching something by text, if possible, as it is more human-friendly.

Also, Heading 3 is not a human-friendly aria-label so this is something that we might need to fix in another ticket

ralucaStan (commented):
Re-review done

tests/e2e/specs/backend/active-filters.test.js
opr: Yup, it was brought in because I based my work here on Nadir's PR - https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2756

I liked the idea that we didn't have to go and find controls by their css selector, and could use more "user-based" selectors, so I wanted to bring it over to this PR too. In places that selectors had already been used, I wasn't sure if I should go back and rewrite them. I can certainly do that, though.

tests/e2e/specs/backend/active-filters.test.js
opr: Added a comment!

I don't think we have control over the label though, that comes from Gutenberg. It refers to the control in the block sidebar:

tests/e2e/specs/backend/active-filters.test.js
opr: I've rewritten all the places where it makes sense to do so (in files touched by this PR!) to use the `pptr-testing-library` queries. This only amounted to 6 or 7 code changes, though. Another look over would be appreciated!

tests/e2e/fixtures/fixture-loaders.js
```diff
@@ -334,7 +352,76 @@ const deleteBlockPages = ( ids ) => {
 	);
 };
 
+/**
+ * Ensure attributes are removed from the site before setting up. This exists in case
+ * the teardown didn't happen properly for any reason. This function exists mostly as a
+ * dev-ex improvement tool while working on tests.
+ *
+ * @return {Promise<void>}
+ */
```
senadir: Can we call teardown manually to accommodate for this need? Are we just clearing attributes and not everything else? Alternatively, can we add attributes in a more defensive way so an error won't be returned if attributes already exist? If that's not possible, can we call all teardown functions in the setup but guard that in a flag that's only true when working locally?

The issue I have with this is that this adds extra load to tests, and doesn't feel like it's enough, which might result in us adding more teardown calls in setup and hindering our E2E speed.

tests/e2e/specs/backend/active-filters.test.js
```diff
@@ -0,0 +1,106 @@
...
+
+const block = {
+	name: 'Active Product Filters',
+	slug: 'woocommerce/active-filters',
+	class: '.wc-block-active-filters',
+};
+
+if ( process.env.WP_VERSION < 5.3 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( 'skipping all other things', () => {} );
```
senadir: This can be deleted. We don't support 5.3 anymore.

tests/e2e/specs/backend/attribute-filter.test.js
```diff
@@ -0,0 +1,176 @@
...
+const block = {
+	name: 'Filter Products by Attribute',
+	slug: 'woocommerce/attribute-filter',
+	class: '.wc-block-attribute-filter',
+};
+
+if ( process.env.WP_VERSION < 5.3 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( 'skipping all other things', () => {} );
+
```
senadir: Same here.

tests/e2e/specs/backend/active-filters.test.js
```diff
@@ -0,0 +1,106 @@
...
+		);
+		await heading6Button.click();
+		await expect(
+			page
+		).toMatchElement(
+			'.wp-block[data-type="woocommerce/active-filters"] h6 textarea',
+			{ text: 'New Title' }
+		);
+
+		expect( await getEditedPostContent() ).toMatchSnapshot();
```
senadir: I think we can get rid of this given that we use fixtures for this block which already ensures that attributes persist.

tests/e2e/specs/backend/price-filter.test.js
```diff
@@ -0,0 +1,129 @@
...
+				/Heading 3/i
+			);
+			await heading3Button.click();
+			await expect(
+				page
+			).not.toMatchElement(
+				'.wp-block[data-type="woocommerce/price-filter"] h6 textarea',
+				{ text: 'New Title' }
+			);
+		} );
```
senadir: Resetting the block back to its initial state after will get painful quickly. Might be worth exploring this in a follow-up?
https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2783

tests/e2e/specs/backend/price-filter.test.js
```diff
@@ -0,0 +1,129 @@
...
+			const document = await getDocument( page );
+			const showGoButton = await getByLabelText(
+				document,
+				/Filter button/i
+			);
+			await showGoButton.click();
+			await expect( page ).toMatchElement(
+				'button.wc-block-filter-submit-button.wc-block-price-filter__button'
+			);
+			expect( await getEditedPostContent() ).toMatchSnapshot();
```
senadir: As I said, block fixtures should solve the need for this.

senadir (commented):
Lovely test suit Thomas, Thank you so much for reviving #2756
I left some minor comments. I would differ the approval/decline to someone else.

tests/e2e/fixtures/fixture-loaders.js
ralucaStan: I've removed the `ensureCleanAttributes` function and created a new issue for handling the clearing of the attributes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5410

ralucaStan (dismissed):
I've rebased this to trunk to use node v16 and handled the requested changes.

mikejolley:
Removing RTM label and milestone until updated.

tests/e2e/config/setup.js
```diff
@@ -39,14 +39,19 @@ module.exports = async ( globalConfig ) => {
 			createCoupons(),
 			createCategories(),
 			createShippingZones(),
+			createProductAttributes(),
 			enablePaymentGateways(),
-			setupPageSettings(),
```
ralucaStan, line 43: @opr is there a reason this was removed?

opr: I can't remember, can you add it back and does it still work fine if you do?

ralucaStan:
The puppeteer-testing-library, although added by Gutenberg is not actively used.
I have removed it and realised that the added functionality can be easily replaced by `expect-puppeteer`, which Gutenberg also uses in its tests.
This is a package that we specify in the jest configuration but it was not part of our package.json.
The selectors from puppeteer-testing-library have been replaced with `.toClick` and the utility `clearAndFillInput` with `.toFill`

ralucaStan (approved):
I've picked this PR after it was opened for almost 4 months and made some changes:
The puppeteer-testing-library, although added by Gutenberg is not actively used.
I have removed it and realised that the added functionality can be easily replaced by expect-puppeteer, which Gutenberg also uses in its tests.
This is a package that we specify in the jest configuration but it was not part of our package.json.
The selectors from puppeteer-testing-library have been replaced with .toClick and the utility clearAndFillInput with .toFill

PR #4567, opened 2021-08-11
PHPUnit Tests running on wp-env

mikejolley:
This PR migrates our unit testing setup to wp-env. [This was based on the work in this PR from @nerrad.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3691)

- Updates PHPUnit tests to run on wp-env using `npm run test:php`
- Implements [PHP Unit Polyfill](https://github.com/Yoast/PHPUnit-Polyfills)
- Removes reliance on woo cores legacy unit testing framework by moving useful helpers into the test suite. This means we no longer need to clone the repo from GitHub which greatly simplifies our bootstrap.php and makes it less prone to errors and conflicts.

Closes #3691 

**Testing against other versions of WooCommerce and PHP**

Testing is currently limited to the stable PHP version (7.4). See below. In the future, we could easily add tests for other versions to our CI config file.

**Testing against other versions of WooCommerce is possible.** Because of this, I've added 2 tests to the CI; one using the latest (development version from .org) which will include betas, and one using the latest stable version. This is inline with L0 policy. We could add other versions in this matrix if we wanted!

**Future work**

Right now there are 2 limitations:

1. PHP 7.0 testing is not possible because of a permissions issue. When running under PHP 7.0, the wp-env fails to start. I think we can mostly ignore this and not test under 7.0 because a) we test under 7.4 so there are no major differences, and b) 7.0 is EOL.
2. PHP 8.0 testing is not yet possible - waiting on support in the wp-phpunit package. See https://core.trac.wordpress.org/ticket/46149

Once these are in place, we can add an override file in the CI script and run tests across multiple versions.

**Testing**

- Tests should pass in the CI
- Locally, run:

```
npm run wp-env destroy
npm run wp-env start
npm run test:php
```

This will recreate the wp-env environments and run PHP tests. It will use PHP 7.4.


opr (approved):
Nice work! I am really fond of the new API for adding products, it's much easier now. I also like the `ControllerTestCase` class. Great work 👏🏼 

Tests are working locally and in CI, so ✅ from me!

PR #4568, opened 2021-08-12
Fix store notice classes

brettshumaker:
The `StoreNoticesContainer` component is using an incorrect class name for a "success" notice and also applying the general `woocommerce-message` (what's used by WC core for the "success" notice) to all notices. In some instances, this could cause the general `woocommerce-message` styles to be applied instead of the `woocommerce-error` or `woocommerce-info` styles.

This PR changes the use of `woocommerce-success` to `woocommerce-message` and prevents `woocommerce-message` from being applied to all notices.

More info about this issue can be found in https://github.com/Automattic/woocommerce-payments/issues/2494 but, essentially what we've found is that when this plugin and WooCommerce Payments are being used on Storefront, the notices displayed in the Checkout block in the credit card section show the incorrect icons. For example, the failed payment notice has the correct red background color but still has the white checkmark icon, which is incorrect for the `woocommerce-error` type of notice.

Fixes # https://github.com/Automattic/woocommerce-payments/issues/2494

### Screenshots

**Before**
126348588-b74d9d93-aa4e-46ef-b53c-c6c588ca626d

**After**
Screen Shot 2021-08-12 at 6 50 49 AM

_Alternative testing instructions with WooCommerce Stripe Payment Gateway available [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4568#pullrequestreview-728544965)._
 
1. Create a test site.
2. Install and activate WooCommerce Payments and WooCommerce Blocks.
3. Install and activate the Storefront theme.
4. Create a new page with the Checkout block.
5. Add product to cart.
6. Go to the new checkout page created in step 4.
7. Try to purchase with incorrect/incomplete credit card data.
8. Ensure that the icon in the notice is correct.

### Changelog

> Adjusted store notice class names so that error notices show the correct icons

### DIFF
--- a/assets/js/base/context/providers/store-notices/components/store-notices-container.js
+++ b/assets/js/base/context/providers/store-notices/components/store-notices-container.js
@@ -15,7 +15,7 @@ const getWooClassName = ( { status = 'default' } ) => {
 		case 'error':
 			return 'woocommerce-error';
 		case 'success':
-			return 'woocommerce-success';
+			return 'woocommerce-message';
 		case 'info':
 		case 'warning':
 			return 'woocommerce-info';
@@ -42,7 +42,6 @@ const StoreNoticesContainer = ( { className, notices, removeNotice } ) => {
 					{ ...props }
 					className={ classnames(
 						'wc-block-components-notices__notice',
-						'woocommerce-message',
 						getWooClassName( props )
 					) }
 					onRemove={ () => {


Aljullu (approved):
Good catch with this issue! I tested and it works as described. :ok_hand:  I also checked WooCommerce Core codebase and there isn't any instance of `woocommerce-success`, so the usage of that class was probably a mistake.

In order to test this PR without WooCommerce Payments:

1. Set up [WooCommerce Stripe Payment Gateway](https://wordpress.org/plugins/woocommerce-gateway-stripe/) and go to a page with the Checkout block.
2. Try to pay with a [card number that fails](https://stripe.com/docs/testing) like `4000000000000127`.
3. Notice the payment will be rejected.
4. Verify the correct icon (exclamation mark instead of check mark) is rendered in the error message.

Thanks for opening this PR @brettshumaker!

PR #4569, opened 2021-08-12
Delete unneeded double spaces in text strings

NekoJonez:
Deleted an unneeded double space in a string on line 271.

https://github.com/woocommerce/woocommerce/issues/30486


Aljullu (approved):
Good catch @NekoJonez and thanks for opening a PR with the fix! I will merge it now so the change is included in the next release of WC Blocks planned for next Monday.

PR #4570, opened 2021-08-12
Add optimizations around checkout filter calls

Aljullu:
This PR includes several optimizations to reduce the number of times `__experimentalApplyCheckoutFilter()` is called in the Cart and Checkout blocks. Each optimization is unrelated from the others, so it might be easier to review this PR commit by commit.

While there isn't any specific performance issue this PR is fixing, `__experimentalApplyCheckoutFilter` executes 3rd-party code, so I think it's good to avoid as much as possible calling it unnecessarily. In addition to that, several of the changes included in this PR are general good practices, so even if the performance impact was negligible, I still think it makes sense merging them.

I measured how many times `__experimentalApplyCheckoutFilter()` was called in these three scenarios:

**Cart block:**

| | On render | When changing a product quantity |
| --- | ---: | ---: |
| `trunk` | 230 | 251 |
| This branch | 212 | 234 |

**Checkout block (desktop):**

| | On render | When changing shipping country |
| --- | ---: | ---: |
| `trunk` | 220 | 252 |
| This branch | 197 | 220 |

**Checkout block (mobile):** (the big improvement here is because `Panel` contents are no longer rendered when collapsed, and `Order Summary` is collapsed by default on mobile)

| | On render | When changing shipping country |
| --- | ---: | ---: |
| `trunk` | 238 | 258 |
| This branch | 28 | 35 |

Note: the exact numbers will depend on the number of products you have in the cart, whether they are on sale, if you have coupons, etc. so most probably you will get different results, but the improvement should be noticeable anyway.

1. Testing means verifying there are no regressions. Try interacting with the Cart and Checkout blocks (removing products, changing quantity, adding a coupon, filling the Checkout form, etc.) and verify everything works as expected.
2. Verify extensibility still works as usual. You can test it getting WooCommerce Subscriptions and adding a subscription product to your cart.

### Changelog

> Performance improvements in the Cart and Checkout block extensibility points

### DIFF
--- a/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
+++ b/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
@@ -29,18 +29,18 @@ const productPriceValidation = ( value ) => mustContain( value, '<price/>' );
 const OrderSummaryItem = ( { cartItem } ) => {
 	const {
 		images,
-		low_stock_remaining: lowStockRemaining = null,
-		show_backorder_badge: showBackorderBadge = false,
+		low_stock_remaining: lowStockRemaining,
+		show_backorder_badge: showBackorderBadge,
 		name: initialName,
 		permalink,
 		prices,
 		quantity,
 		short_description: shortDescription,
 		description: fullDescription,
-		item_data: itemData = [],
+		item_data: itemData,
 		variation,
 		totals,
-		extensions = {},
+		extensions,
 	} = cartItem;
 
 	// Prepare props to pass to the __experimentalApplyCheckoutFilter filter.
--- a/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
+++ b/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
@@ -13,7 +13,7 @@ import {
 import PropTypes from 'prop-types';
 import Dinero from 'dinero.js';
 import { getSetting } from '@woocommerce/settings';
-import { useCallback, useMemo } from '@wordpress/element';
+import { useMemo } from '@wordpress/element';
 import { useStoreCart } from '@woocommerce/base-context/hooks';
 
 /**
@@ -23,6 +23,8 @@ import ProductBackorderBadge from '../product-backorder-badge';
 import ProductImage from '../product-image';
 import ProductLowStockBadge from '../product-low-stock-badge';
 import ProductMetadata from '../product-metadata';
+
+const productPriceValidation = ( value ) => mustContain( value, '<price/>' );
 
 const OrderSummaryItem = ( { cartItem } ) => {
 	const {
@@ -45,11 +47,6 @@ const OrderSummaryItem = ( { cartItem } ) => {
 	// We need to pluck out receiveCart.
 	// eslint-disable-next-line no-unused-vars
 	const { receiveCart, ...cart } = useStoreCart();
-
-	const productPriceValidation = useCallback(
-		( value ) => mustContain( value, '<price/>' ),
-		[]
-	);
 
 	const arg = useMemo(
 		() => ( {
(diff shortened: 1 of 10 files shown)


senadir:
Thank you for this PR Albert! 234 is still a worrying number. Do you know if we happen to execute third party 234 times or would the majority of this bails?
Having some sort of memoization here might be helpful

Aljullu:
> Thank you for this PR Albert! 234 is still a worrying number. Do you know if we happen to execute third party 234 times or would the majority of this bails?

I only counted the cases where [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/020e3fdb61dac71a2b4fb42a94970365634192b7/packages/checkout/registry/index.ts#L109) was reached, in which case the third-party code will run if there is a filter registered. So yes, we would be executing third-party code >200 times if an extension registered in all possible filters.

The main cause I could find is that the `arg` of many of those filters is not successfully memoized. For example:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/020e3fdb61dac71a2b4fb42a94970365634192b7/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-item-row.tsx#L122-L129

Even though it's wrapped in a `useMemo()`, `lineItem` and `cart` are different on every render. So every time the component is re-rendered, all those filters are fired.

I guess we could do a couple of things to improve this:
* Try to get better at memoizing the values returned by `useStoreCart()`. I think there are some easy wins we can make there (some are already done in #4548), but we would need to memoize **everything** returned by that hook and I'm not sure if it's even possible/fesable.
* Even if we could memoize everything returned by `useStoreCart()`, I wonder if those filters need to have access to the entire `cart` object. Ie: does the `itemName` filter need to be re-run when the billing address changes? Probably not. I think we might be exposing too many things to those filters.

What do you think @senadir?

senadir:
> I only counted the cases where this line was reached, in which case the third-party code will run if there is a filter registered. So yes, we would be executing third-party code >200 times if an extension registered in all possible filters.

This looks concerning to me. We don't know how third party code is going to work. Till now, the filters we wrote were all pure functions because we used the provided tools (extensions, cart from arg...), if a third party code has side effect functions (fetching data), the fetching will happen several times.

> The main cause I could find is that the arg of many of those filters is not successfully memorized.
This is the example I provided in our last NIRL meetup, around this exact issue, I didn't get around to test more if that issue resurfaced.

> Try to get better at memoizing the values returned by useStoreCart().

This is probably something we need to do anyway, the current block rerenders several times (up to 10 times) before ideling.

> Even if we could memoize everything returned by useStoreCart(), I wonder if those filters need to have access to the entire cart object. Ie: does the itemName filter need to be re-run when the billing address changes? Probably not. I think we might be exposing too many things to those filters.

While this might complicate our docs significantly, I guess it's a good win for us. The issue is how do we decide what to expose or not, such process is probably going to be very subjective and 3PD driven. I also suspect this might not be enough because we must expose the `extensions` fields anyway, which is a complex field that would be considered new on each server request.

Do billing address on `useStoreCart` change on each update to the billing field, or would it change on each server response? 
I suspect the blame here mostly falls on functions getting recreated a lot.

I will investigate this issue more to try and pinpoint all the data points that keep changing.

senadir (approved):

mikejolley (approved):
Changes make sense. I tested the subscriptions panel and noticed no regressions.

Aljullu:
> Do billing address on `useStoreCart` change on each update to the billing field, or would it change on each server response?

Billing address is updated on each server response, which happens when changing the values of Country/City/State/Postal Code fields. However, given that the address is not memoized, if you add a coupon or change the quantity of a product, for example, it will be considered a different value even though all fields are still the same.

> I will investigate this issue more to try and pinpoint all the data points that keep changing.

That would be great, thanks!

---

Thanks for the reviews @senadir and @mikejolley!

PR #4580, opened 2021-08-16
Make SortSelect components controlled

Aljullu:
Now that #4463 is closed, I wanted to extract the changes from one commit which I think is a nice clean up: making the `ReviewSortSelect` component controlled, so it doesn't have a `defaultValue` and a `value` props, which is always confusing. I also took the chance to remove the `defaultValue` prop from `SortSelect` and `ProductSortSelect` given that they were never used.

#### Reviews blocks
1. Make sure you have at least a couple of reviews in your store.
2. Create a post or page with the All Reviews, Reviews by Product and Reviews by Category blocks.
3. In the editor, make sure you can change the `Order Product Reviews by` value under `List Settings` and blocks are updated accordingly.
4. In the frontend, make sure you can change the order reviews are displayed in.

#### All Products block
1. Create a post or page with the All Products block.
3. In the editor, make sure you can change the `Order Products By` value under `Content Settings` and the block is updated accordingly.
4. In the frontend, make sure you can change the order products are displayed in.

### DIFF
--- a/assets/js/base/components/product-list/product-sort-select/index.js
+++ b/assets/js/base/components/product-list/product-sort-select/index.js
@@ -10,11 +10,10 @@ import SortSelect from '@woocommerce/base-components/sort-select';
  */
 import './style.scss';
 
-const ProductSortSelect = ( { defaultValue, onChange, readOnly, value } ) => {
+const ProductSortSelect = ( { onChange, readOnly, value } ) => {
 	return (
 		<SortSelect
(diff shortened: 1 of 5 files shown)


senadir (approved):
Tests well.
Unrelated, while testing, I saw the sort component style is broken in the editor. I created an issue here https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4596
image

PR #4582, opened 2021-08-16
Remove top rule so validation messages are positioned correctly below

mikejolley:
At some point validation message styling was broken causing it to overlap the fields below. This could be due to a style change in the parent. Removing the `calc` rule fixes it.

### Screenshots

Screenshot 2021-08-16 at 13 14 25

1. Cause a validation error on checkout
2. Confirm the error notice is shown below the field


ralucaStan (approved):
The error messages look as expected

PR #4585, opened 2021-08-16
Add label for Block Title component to satisfy accessibility

### LINKED ISSUE #1680: Filter blocks (editor): missing label for <textarea> element
imatge

Filter blocks allow editing the title via a `<textarea>`, but it's missing a label.

tjcafferkey:
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/1680

#### Accessibility

1. In a new page add Filter Products By Attribute (and Price) block
2. Choose an attribute and then click "Done"
3. Inspect the blocks title within the editor
4. Observe `<label />` element which has `for` attribute value which corresponds to the `<textarea />` `id` value

### Changelog

> Add label element to `<BlockTitle>` component

### DIFF
--- a/assets/js/editor-components/block-title/index.js
+++ b/assets/js/editor-components/block-title/index.js
@@ -4,17 +4,29 @@
 import PropTypes from 'prop-types';
 import { PlainText } from '@wordpress/block-editor';
 import classnames from 'classnames';
+import { withInstanceId } from '@wordpress/compose';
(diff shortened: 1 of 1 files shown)


mikejolley (approved):
Thanks for submitting this patch. I can confirm this fixes the issue in the editor 👍🏻

PR #4588, opened 2021-08-17
Add padding in 2020 and 2021

mikejolley:
Fixes a styling regression affecting 2021 and 2020 themes caused by https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4568

These themes do not style the `woocommerce-error` container, they style an inner `li` element. We don't use lists. To fix this, we can apply the padding in the component.

We should likely take full control over notice styling in the future and deprecate core classes.

### Screenshots

Before:
Screenshot 2021-08-17 at 11 31 23

After:
Screenshot 2021-08-17 at 11 27 32

Using 2021 and 2020 themes:

1. Go to checkout
2. Enter invalid postcode and submit order
3. Check error notice styling

### DIFF
--- a/assets/js/base/context/providers/store-notices/components/style.scss
+++ b/assets/js/base/context/providers/store-notices/components/style.scss
@@ -22,3 +22,10 @@
 		margin-top: 1em;
 	}
 }
+
+.theme-twentytwentyone,
+.theme-twentytwenty {
+	.wc-block-components-notices__notice {
+		padding: 1.5rem 3rem;
+	}
+}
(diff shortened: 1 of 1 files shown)


assets/js/base/context/providers/store-notices/components/style.scss
```diff
@@ -22,3 +22,10 @@
 		margin-top: 1em;
 	}
 }
+
+.theme-twentytwentyone,
+.theme-twentytwenty {
+	.wc-block-components-notices__notice {
+		padding: 1.5rem 3rem;
+	}
+}
```
senadir: Seems like a weird place for this to live, can we add a `@todo` here to move it to WooCommerce Core?

senadir (approved):

PR #4591, opened 2021-08-17
Define parent in registerCheckoutBlock

### LINKED ISSUE #4492: Sample block should only be visible in correct areas, not...
Sample block should only be visible in correct areas, not top level.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2bfe6deda4ef3f94972466290b3d7484e13c3182/assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-sample-block/index.tsx#L19-L30
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in 2bfe6deda4ef3f94972466290b3d7484e13c3182 when #4268 was merged. cc @mikejolley

mikejolley:
Update `registerCheckoutBlock` to include an empty `parent` property. This ensures a block registered via `registerCheckoutBlock` is only available in defined areas, not globally.

Fixes #4492

1. Dev build only (`npm start`).
2. "sample block" should not be available in the block inspector at top level.
3. Insert checkout i2.
4. "sample block" should be available in the contact information section as an inner block.

### DIFF
--- a/packages/checkout/blocks-registry/index.ts
+++ b/packages/checkout/blocks-registry/index.ts
@@ -175,18 +175,28 @@ export const registerCheckoutBlock = (
 	assertOption( options, 'areas', 'array' );
 	assertBlockComponent( options, 'component' );
 
+	/**
+	 * If provided with a configuration object, this registers the block with WordPress.
+	 */
 	if ( options?.configuration ) {
 		assertOption( options, 'configuration', 'object' );
 		registerExperimentalBlockType( blockName, {
 			...options.configuration,
 			category: 'woocommerce',
+			parent: [],
 		} );
 	}
 
+	/**
+	 * This enables the inner block within specific areas. See RegisteredBlocks.
+	 */
 	options.areas.forEach( ( area ) =>
 		registerBlockForArea( area, blockName )
 	);
 
+	/**
+	 * This ensures the frontend component for the checkout block is available.
+	 */
 	registerBlockComponent( {
 		blockName,
 		component: options.component,


packages/checkout/blocks-registry/index.ts
```diff
@@ -175,18 +175,28 @@ export const registerCheckoutBlock = (
...
 
+	/**
+	 * If provided with a configuration object, this registers the block with WordPress.
+	 */
 	if ( options?.configuration ) {
 		assertOption( options, 'configuration', 'object' );
 		registerExperimentalBlockType( blockName, {
 			...options.configuration,
 			category: 'woocommerce',
+			parent: [],
```
senadir, line 186: Can't we extrapolate this from `options.areas`?

mikejolley: It doesn't seem like we need to. Those areas still work.

senadir: I'm not sure I understand?
I mean that instead of waiting for 3PD to pass the parent value, we implicitly provide it from what's in `options.areas`

mikejolley: Parent _should_ be empty. We're not accepting this prop from 3PD, it's being overwritten as `[]`.

3PD use the area prop to define where they want blocks to appear. Not parent.

senadir: Ah right, sorry. My understanding is that having `[]` would disable the block from all inserters.
I don't understand yet how can we show sample block only on our innerBlocks, is it because `allowedBlocks` override `parent`?

senadir (approved):

packages/checkout/blocks-registry/index.ts
mikejolley: Yup it certainly looks that way. My testing instructions cover this—sample block can be inserted into the contact information section.

senadir: Yeah, I can see that. Approved.

PR #4606, opened 2021-08-18
Checkout i2: Handle forced blocks in editor

senadir:
Experiment spun from #4504 to solve #4447.
By adding `force` to the block settings object, we can force it to render even if it's not available.

We need to have it on `settings` because that's the only way we can access it on block (by using `getBlockType`) The next option would be to have that data returned alongside `getRegisteredBlocks`. It would certainly save us from having to call `getBlockType`.

Still experimental, on Checkout i2 load on editor, sample block would be force inserted at the end.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout-i2/form-step/additional-fields.tsx
+++ b/assets/js/blocks/cart-checkout/checkout-i2/form-step/additional-fields.tsx
@@ -6,7 +6,6 @@ import {
 	RegisteredBlocks,
 	getRegisteredBlocks,
 } from '@woocommerce/blocks-checkout';
-import { useRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
@@ -21,12 +20,9 @@ export const AdditionalFields = ( {
 } ): JSX.Element => {
 	const registeredBlocks = getRegisteredBlocks( area );
 	const { 'data-block': clientId } = useBlockProps();
-	const { current: template = [] } = useRef< Array< string > >( [
-		...registeredBlocks,
-	] );
-	useForcedLayout( {
+	const template = useForcedLayout( {
 		clientId,
-		template,
+		template: registeredBlocks,
 	} );
 	return (
 		<div className="wc-block-checkout__additional_fields">
--- a/assets/js/blocks/cart-checkout/checkout-i2/form-step/additional-fields.tsx
+++ b/assets/js/blocks/cart-checkout/checkout-i2/form-step/additional-fields.tsx
@@ -1,25 +1,36 @@
 /**
  * External dependencies
  */
-import { InnerBlocks } from '@wordpress/block-editor';
+import { InnerBlocks, useBlockProps } from '@wordpress/block-editor';
 import {
 	RegisteredBlocks,
 	getRegisteredBlocks,
 } from '@woocommerce/blocks-checkout';
+import { useRef } from '@wordpress/element';
 
 /**
  * Internal dependencies
  */
 import './editor.scss';
+import { useForcedLayout } from '../use-forced-layout';
 
 export const AdditionalFields = ( {
 	area,
 }: {
 	area: keyof RegisteredBlocks;
 } ): JSX.Element => {
+	const registeredBlocks = getRegisteredBlocks( area );
+	const { 'data-block': clientId } = useBlockProps();
+	const { current: template = [] } = useRef< Array< string > >( [
+		...registeredBlocks,
+	] );
+	useForcedLayout( {
+		clientId,
+		template,
+	} );
 	return (
 		<div className="wc-block-checkout__additional_fields">
-			<InnerBlocks allowedBlocks={ getRegisteredBlocks( area ) } />
+			<InnerBlocks allowedBlocks={ template } />
 		</div>
 	);
 };
--- a/assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-fields-block/edit.tsx
+++ b/assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-fields-block/edit.tsx
@@ -1,17 +1,15 @@
 /**
  * External dependencies
  */
-import { useLayoutEffect, useRef } from '@wordpress/element';
 import { useBlockProps, InnerBlocks } from '@wordpress/block-editor';
-import { useSelect, useDispatch } from '@wordpress/data';
-import { createBlock, getBlockType } from '@wordpress/blocks';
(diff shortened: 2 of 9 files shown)


assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-fields-block/edit.tsx
```diff
@@ -23,21 +26,69 @@ const ALLOWED_BLOCKS = [
...
+			return;
+		}
+
+		const innerBlocks = currentBlock.innerBlocks || [];
+
+		// Missing check to see if registered block is 'forced'
+		currentRegisteredBlocks.current.forEach( ( blockName: string ) => {
+			const block = getBlockType( blockName );
+			if (
+				block.force &&
```
senadir: This is what I added from 4504

packages/checkout/blocks-registry/index.ts
```diff
@@ -180,6 +180,18 @@ export const registerCheckoutBlock = (
...
+				lock: {
+					type: 'object',
+					default: {
+						remove: !! options.configuration.force,
+						move:
+							options.configuration.attributes?.lock?.default
+								?.move,
+					},
+				},
+			},
```
senadir: Add lock attribute to the block in case it's not already there, but making sure we don't override move lock. This needs more cleaning tomorrow.

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-sample-block/index.tsx
```diff
@@ -39,7 +45,14 @@ registerCheckoutBlock( 'woocommerce/checkout-sample-block', {
...
 		},
-		attributes: {},
+		attributes: {
+			lock: {
+				type: 'object',
+				default: {
+					remove: true,
+				},
+			},
+		},
```
senadir, line 53: Forcing is currently separated from locking, but if you force a block to render, you should prevent it from being deleted as well. This can be handled at the registration in `registerCheckoutBlock` by automatically adding those attributes.

assets/js/blocks/cart-checkout/checkout-i2/form-step/additional-fields.tsx
```diff
@@ -1,25 +1,36 @@
...
 import './editor.scss';
+import { useForcedLayout } from '../use-forced-layout';
 
 export const AdditionalFields = ( {
 	area,
 }: {
 	area: keyof RegisteredBlocks;
 } ): JSX.Element => {
+	const registeredBlocks = getRegisteredBlocks( area );
+	const { 'data-block': clientId } = useBlockProps();
```
mikejolley, line 22: Can you explain the naming here? It looks like usage is of clientId, so why does naming not reflect this?

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-fields-block/edit.tsx
```diff
@@ -23,21 +24,30 @@ const ALLOWED_BLOCKS = [
...
+
 	const {
 		addressFieldControls: Controls,
 	} = useCheckoutBlockControlsContext();
+	const registeredBlocks = getRegisteredBlocks( 'fields' );
+	const { current: template = [] } = useRef< Array< string > >( [
+		...ALLOWED_BLOCKS,
+		...registeredBlocks,
+	] );
+	useForcedLayout( {
```
mikejolley: Just thinking out loud, do you think there would be greater utility in moving the template ref logic above to this hook and returning the template?

assets/js/blocks/cart-checkout/checkout-i2/use-forced-layout.ts
```diff
@@ -0,0 +1,63 @@
...
+				),
+			};
+		},
+		[ clientId, currentTemplate ]
+	);
+	/**
+	 * If the current inner blocks differ from the registered blocks, push the differences.
+	 *
+	 */
+	useLayoutEffect( () => {
```
mikejolley, line 45: 👍🏻

mikejolley (approved):
@senadir Tested in the editor and works like a charm. I added some small qu/feedback below. Preapproving.

assets/js/blocks/cart-checkout/checkout-i2/form-step/additional-fields.tsx
senadir: `useBlockProps` exposes `clientId` in the `data-block` attribute, we're pulling that value here. Other option including prop drilling it from above in the Edit function.
`useBLockProps` data is meant to be spread over a div, so it's not named `clientId`.

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-fields-block/edit.tsx
senadir: I recall doing this but hitting a problem with some other block. I will try agian.

senadir: Working fine. I remember also moving `getRegisteredBlocks` to the hook which was the problem.

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-sample-block/index.tsx
senadir: I was actually wrong here @mikejolley Forcing doesn’t depend on a `force: true` attribute anymore ,it depends on having `attributes.lock.remove: true` which would cause the block to be inserted. It reduces the surface area for the API and works as expected. This is only for blocks registered with `registerCheckoutBlock`.

assets/js/blocks/cart-checkout/checkout-i2/form-step/additional-fields.tsx
mikejolley: I guess my question is more, why `data-block` and not `data-client-id`

senadir: Well, that's what's returned by `useBlockProps` I'm renaming `data-block` to `clientId`, not the other way around. Or do you want me to use `data-client-id` instead of `clientId`?

mikejolley: ah I understand now. Sorry, I misread. Leave it alone 👍🏻

Issue #4448, opened 2021-07-08
Checkout i2: Test extension integration


mikejolley:
Creating this issue to track progress and serve as a goal for multiple related/blocking issues. This is blocked by #4446, #4306, #4447.

Once the above integration areas/interfaces are available, we need to create a test integration in the form of a Newsletter subscription checkbox. In doing so, we'll need to work on other interfaces and API support for fields.

**Tasks:**
- [x] Successfully register and render a checkbox field on the checkout
- [ ] Add support for validation callbacks (this is not specifically needed for newsletter subscription but we should test that a field could be made required whilst working on this)
- [x] Post the value of the checkbox to the server (via Store API) and ensure an extension can capture and handle this value on the server-side. 
   - [x] Error handling - Can extensions throw a validation error from the server-side?
   - [x] Success handling - Can extensions use the value and/or store it to the order?

I'm increasing the estimate for this based on the assumption work will need to be done on the Store API side as well as client side validation.

mikejolley:
@senadir I think we should start this early on next cycle so we can scope out anything that we're missing, such as API changes.

mikejolley:
I think we can call this done given we've added support at API level and successfully built an integration outside of the repo https://github.com/opr/newsletter-test/blob/main/newsletter-test.php

PR #4607, opened 2021-08-19
Experimental newsletter subscription checkbox block POC and Store API supporting changes

mikejolley:
This PR introduces some changes to the Store API to facilitate nested fields, an experimental newsletter subscription block, and additional hooks. This is part of #4448

**`useCheckoutExtensionData` hook**

This hook allows custom, namespaced data to be stored to the checkout-state context provider. Any values stored here are then passed to the Store API under the `extensions` top-level parameter. This has been implemented in the _newsletter_ block in this PR. It will need exporting externally for extensions to consume it.

_Testing:_

- Insert Newsletter block on checkout i2, save
- Go to frontend
- Check the box and submit order
- Check network inspector and confirm the newsletter API key was set on the request

**Store API changes**

In reviewing the Store API I noticed that nested parameters were never validated, so any given `arg_options` are ignored. WP core does not have a solution for this so I introduced a recursive validation function. Test this by editing the newsletter API registration in the main blocks plugin file in this PR:

```php
					'schema_callback' => function() {
						return array(
							'newsletter' => array(
								'description' => __( 'Subscribe to newsletter opt-in.', 'woo-gutenberg-products-block' ),
								'type'        => 'boolean',
								'context'     => array(),
								'arg_options' => array(
									'validate_callback' => function( $value ) {
										return new \WP_Error( 'api-error', ( $value ? 'true' : 'false' ) . ' was posted to the newsletter validation callback' );
									},
								),
							),
						);
					},
```

**Newsletter Block**

I've added a test newsletter block available within the contact section in order to test the above changes. This needs extracting into an extension to ensure all functionality is available externally. This is a follow-up item.


assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/register-components.ts
```diff
@@ -9,6 +9,12 @@ import { WC_BLOCKS_BUILD_URL } from '@woocommerce/block-settings';
 // eslint-disable-next-line no-undef,camelcase
 __webpack_public_path__ = WC_BLOCKS_BUILD_URL;
 
+/**
+ * Internal dependencies
+ */
+import './checkout-sample-block';
+import './checkout-newsletter-subscription-block'; // Register these blocks here so they load in editor and frontend contexts.
```
mikejolley: @senadir this was the frontend issue

senadir (approved):
I'm approving changes in this PR. We still need some documentation and more testing. The newsletter block didn't work for me. I will be testing it against @opr implementation.

Before merging this PR, I will delete the newsletter block.

opr:
@senadir you may clone https://github.com/opr/newsletter-test and activate the plugin to test the integration. I have just pushed my changes that remove the block from this branch.

Note that the checkout won't submit - there's a purposeful exception thrown on https://github.com/opr/newsletter-test/blob/main/newsletter-test.php#L53 - just remove this to allow the checkout form to submit

I also passed down the result of the `useCheckoutSubmit` and `useCheckoutExtensionData` hooks to the children components in db1d715f94c52f4347a7fc26edf29291e45356a4

assets/js/blocks/cart-checkout/checkout-i2/frontend.tsx
```diff
@@ -36,12 +40,16 @@ const Wrapper = ( {
 	// we need to pluck out receiveCart.
 	// eslint-disable-next-line no-unused-vars
 	const { extensions, receiveCart, ...cart } = useStoreCart();
+	const checkoutSubmitData = useCheckoutSubmit();
+	const checkoutExtensionData = useCheckoutExtensionData();
```
opr: Not 100% confident about the naming of these variables, is it descriptive enough about what it contains?

opr: The reason I am choosing to pass the entire object returned by the hook is because how can we know what the consuming components will need?

senadir: We should skip exposing `useCheckoutSubmit`. It exposes irrelevant data and it's an internal hook that we use to aid showing the submit button, we might give up this hook for other ways to handle the submit button later and having it on public API causes a support liability. 

I see in https://github.com/opr/newsletter-test/blob/main/assets/js/checkout-newsletter-subscription-block/block.js#L10 that you use it for `isDisabled` which IMO something we can skip anyway because clicking the checkbox again wouldn’t really effect a checkout at the submit process.

If it proves that we need `isDisabled`, we can look into exposing Checkout lifecycle related in a dedicated hook, probably a subset of what's exposed from `useCheckoutContext` but without `dispatchActions`.

PR #4608, opened 2021-08-20
Add Drawer component

### LINKED ISSUE #4597: Mini Cart block: Create native Drawer component
In the first iteration of the Mini Cart block, it will open in a drawer that will slide in from the right side of the screen (left side in RTL languages). This issue is about creating a component we can use for this. Ideally, this component should not require React given that we want the drawer to open instantly even if React is not yet executed.

## Tasks
* **Functionality:**
  * [ ] Open the drawer when clicking the Mini Cart button.
  * [ ] Make it possible to close the drawer when clicking on the close button.
  * [ ] Make it possible to close the drawer when clicking outside of it.
* **Styling:**
  * [ ] Slide in the drawer from the right (from the left in RTL languages).
* **Accessibility:**
  * [ ] Capture focus inside the drawer and move it back to the content when closed.
  * [ ] Alert screen readers when the drawer is opened/closed.

### LINKED ISSUE #4600: Mini Cart block: prerender when the user hovers the Mini Cart button, but only open when the user clicks
In the current Mini Cart block prototype (#4510), the Mini Cart contents are displayed when the user hovers the button, but this is not the desired behavior. Instead, it should only open when the user clicks on it. However, we can take advantage of the `mouseover` event to start pre-rendering the contents.

Aljullu:
Fixes #4597.
Fixes #4600.

Follow-up of #4510.

This PR introduces a `Drawer` component to be used by the Mini Cart block. Styles are not final yet, but the interaction should be mostly done.

According to the designs, the `<body>` contents should move along the drawer. However, I had some issues adding `transform: translateX()` to the `<body>` element (admin bar appeared displaced, transitions seemed to be clunky...). I think we can leave it for a follow-up or reconsider the designs if we see it gives too many problems.

### Screenshots

https://user-images.githubusercontent.com/3616980/130203163-500d6d14-b347-40f7-a4a4-66f80e66ce2c.mp4

2. Visit that post or page in the frontend and verify the Mini Cart button appears.
3. Click on the Mini Cart button and verify the Drawer slides in from the right side of the screen. Press <kbd>Tab</kbd> several times to ensure the focus is inside the Drawer.
4. Reload the page and try clicking the button quickly (without hovering it for too long) and verify the 'skeleton' appears.
5. Reload the page again and try hovering the button, waiting for a couple of seconds, and then clicking it. Verify the 'skeleton' doesn't show up and instead Mini Cart contents appear directly because they were pre-loaded while the button was hovered.

### Known issues
* #4618: leaving it for a follow-up because it might require debugging some Gutenberg components/hooks.

### DIFF
--- a/assets/js/base/components/drawer/style.scss
+++ b/assets/js/base/components/drawer/style.scss
@@ -1,5 +1,21 @@
 $drawer-width: 480px;
 
+.wc-block-components-drawer__screen-overlay {
+	background-color: rgba(0, 0, 0, 0.15);
+	bottom: 0;
+	left: 0;
+	position: fixed;
+	right: 0;
+	top: 0;
+	transition: all 0.3s;
+	z-index: 999;
+}
+
+.wc-block-components-drawer__screen-overlay--is-hidden {
+	background: rgba(0, 0, 0, 0);
+	pointer-events: none;
+}
+
 .wc-block-components-drawer {
 	@include with-translucent-border(0 0 0 1px);
 	background: #fff;
@@ -12,7 +28,6 @@ $drawer-width: 480px;
 	top: 0;
 	transform: translateX(-$drawer-width);
 	width: $drawer-width;
-	z-index: 999;
 }
 
 .wc-block-components-drawer--with-slide-out {
@@ -61,7 +76,7 @@ $drawer-width: 480px;
 	}
 }
 
-.wc-block-components-drawer--is-hidden {
+.wc-block-components-drawer__screen-overlay--is-hidden .wc-block-components-drawer {
 	transform: translateX(0);
 }
 
--- a/assets/js/base/components/drawer/style.scss
+++ b/assets/js/base/components/drawer/style.scss
@@ -107,6 +107,7 @@ $drawer-width: 480px;
 // Same styles as `Title` component.
 .wc-block-components-drawer .components-modal__header-heading {
 	@include reset-box();
-	@include font-size(large);
+	// We need the font size to be in rem so it doesn't change depending on the parent element.
+	@include font-size(large, 1rem);
 	word-break: break-word;
 }
--- a/assets/js/base/components/drawer/style.scss
+++ b/assets/js/base/components/drawer/style.scss
@@ -1,0 +1,97 @@
+$drawer-width: 480px;
+
+.wc-block-components-drawer {
+	@include with-translucent-border(0 0 0 1px);
+	background: #fff;
+	display: block;
+	height: 100%;
+	left: 100%;
+	overflow: auto;
+	position: fixed;
+	right: 0;
+	top: 0;
+	transform: translateX(-$drawer-width);
+	width: $drawer-width;
+	z-index: 999;
+}
+
+.wc-block-components-drawer--with-slide-out {
+	transition: all 0.3s;
+}
+
+.wc-block-components-drawer--with-slide-in {
+	animation-duration: 0.3s;
+	animation-name: slidein;
+}
+
+@keyframes slidein {
+	from {
+		transform: translateX(0);
+	}
+
+	to {
+		transform: translateX(-$drawer-width);
+	}
+}
+
+@media only screen and (max-width: 480px) {
+	.wc-block-components-drawer {
+		transform: translateX(-100vw);
+		width: 100vw;
+	}
+
+	@keyframes slidein {
+		from {
+			transform: translateX(0);
+		}
+
+		to {
+			transform: translateX(-100vw);
+		}
+	}
+}
+
+@media screen and (prefers-reduced-motion: reduce) {
+	.wc-block-components-drawer--with-slide-out {
+		transition: none;
+	}
+
+	.wc-block-components-drawer--with-slide-in {
+		animation-name: none;
+	}
+}
+
+.wc-block-components-drawer--is-hidden {
+	transform: translateX(0);
+}
+
+.wc-block-components-drawer .components-modal__content {
+	padding: $gap-largest $gap;
+}
+
+.wc-block-components-drawer .components-modal__header {
+	position: relative;
+
+	// Close button.
+	.components-button {
+		@include reset-box();
+		background: transparent;
+		position: absolute;
+		right: 0;
+		top: 0;
+		// Increase clickable area.
+		padding: 1em;
+		margin: -1em;
+
+		> span {
+			@include visually-hidden();
+		}
+	}
+}
+
+// Same styles as `Title` component.
+.wc-block-components-drawer .components-modal__header-heading {
+	@include reset-box();
+	@include font-size(large);
+	word-break: break-word;
+}
--- a/assets/js/base/components/drawer/style.scss
+++ b/assets/js/base/components/drawer/style.scss
@@ -34,6 +34,9 @@ $drawer-width: 480px;
 	transition: all 0.3s;
 }
 
+// We can't use transition for the slide-in animation because the element
+// doesn't exist in the DOM when not open. Instead, use an animation that
+// is triggered when the element is appended to the DOM.
 .wc-block-components-drawer--with-slide-in {
 	animation-duration: 0.3s;
 	animation-name: slidein;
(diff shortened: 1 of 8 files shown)


Aljullu:
Rebased and tests passing, this is ready for review again @ralucaStan.

assets/js/base/components/drawer/style.scss
```diff
@@ -0,0 +1,116 @@
+$drawer-width: 480px;
+
+.wc-block-components-drawer__screen-overlay {
+	background-color: rgba(95, 95, 95, 0.35);
+	bottom: 0;
+	left: 0;
+	position: fixed;
+	right: 0;
+	top: 0;
+	transition: all 0.3s;
```
ralucaStan: To me, the animation feels _snappy_, and I think one reason for this is that the overlay just appears out of the blue. I tried adding a transition for opacity, so its appearance seems smoother and I find it to be better. 

 Also, it's not recommended to just transition all properties, because:
- not all can be transitioned efficiently
- it's good to be specific, so you know what the intention behind the code is
- all properties that might change would also be transitioned, and that's not always the intention.

Here is the CSS I recommend using, with the opacity transition included.
```css
.wc-block-components-drawer__screen-overlay {
	background-color: rgba( 95, 95, 95, 0.35 );
	bottom: 0;
	left: 0;
	position: fixed;
	right: 0;
	top: 0;
	transition: background-color 0.3s, opacity 0.3s;
	z-index: 999;
	opacity: 1;
}

.wc-block-components-drawer__screen-overlay--is-hidden {
	background: rgba( 95, 95, 95, 0 );
	pointer-events: none;
	opacity: 0;
}
```

assets/js/base/components/drawer/style.scss
```diff
@@ -0,0 +1,116 @@
...
+
+.wc-block-components-drawer--with-slide-out {
+	transition: all 0.3s;
+}
+
+// We can't use transition for the slide-in animation because the element
+// doesn't exist in the DOM when not open. Instead, use an animation that
+// is triggered when the element is appended to the DOM.
+.wc-block-components-drawer--with-slide-in {
+	animation-duration: 0.3s;
```
ralucaStan: I recommend creating a constant for this duration, as it's shared across classes, and we would want things to be in sync.

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
```diff
@@ -13,25 +15,114 @@ import {
...
 
 	return (
-		<CartLineItemsTable
-			lineItems={ cartItems }
-			isLoading={ cartIsLoading }
-		/>
+		<>
+			<button
+				className="wc-block-mini-cart__button"
+				onClick={ () => {
```
ralucaStan, line 49: I think we need to debounce this as well. 
I noticed a behavior when I click to open and close really fast. The fullscreen layer quickly appears and disappears. 
I've also noticed that if the page didn't finish loading and I click the button the drawer does not appear.

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
```diff
@@ -13,25 +15,114 @@ import {
...
 
-renderFrontend( {
-	selector: '.wc-block-mini-cart__contents',
-	Block: withStoreCartApiHydration(
-		withRestApiHydration( MiniCartContents )
-	),
-} );
+const renderMiniCartFrontend = () => {
+	// Check if button is focused. In that case, we want to refocus it after we
+	// replace it with the React equivalent.
```
ralucaStan, line 92: 👏

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
```diff
@@ -13,25 +15,114 @@ import {
...
+		document.activeElement.parentNode instanceof HTMLElement
+	) {
+		focusedMiniCartBlock = document.activeElement.parentNode;
+	}
+	/* eslint-enable @wordpress/no-global-active-element */
+
+	renderFrontend( {
+		selector: '.wc-block-mini-cart',
+		Block: withStoreCartApiHydration(
+			withRestApiHydration( MiniCartBlock )
```
ralucaStan, line 109: unrelated to the code above, while testing I've noticed if I have a lot of items in the cart, and I delete a couple and then I click outside the drawer, on the overlay, the mini cart does not disappear.

assets/js/base/components/drawer/style.scss
```diff
@@ -0,0 +1,116 @@
...
+	overflow: auto;
+	position: fixed;
+	right: 0;
+	top: 0;
+	transform: translateX(-$drawer-width);
+	width: $drawer-width;
+}
+
+.wc-block-components-drawer--with-slide-out {
+	transition: all 0.3s;
```
ralucaStan: Also here I recommend specifying exactly what needs to be transitioned.

ralucaStan (changes requested):
I've added some comments mostly around how the animation happens. 
I also recommend getting an opinion from design about the animation parameters and timing.

assets/js/base/components/drawer/style.scss
Aljullu: Good catch with the lack of animation on the overlay. I have been playing around with it a a bit more and I noticed it has the same issue as `.wc-block-components-drawer`. When the Drawer is closed, the element doesn't exist in the DOM, so when it's opened it gets appended but then the transition is not triggered. I had to create a `fadein` animation for it to work. Besides that, I copied your suggestion to use opacity.

You can see the changes here: d620a50001cc64bc4f1efa4e117ad16311bea6b1.

> Also, it's not recommended to just transition all properties, because:

Good points, I will keep that in mind in the future!

assets/js/base/components/drawer/style.scss
Aljullu: Good idea, done as part of d620a50001cc64bc4f1efa4e117ad16311bea6b1.

assets/js/base/components/drawer/style.scss
Aljullu: Done in d620a50001cc64bc4f1efa4e117ad16311bea6b1.

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
Aljullu: I wasn't able to reproduce this issue exactly. But looking at the code I think we can safely ignore clicks when the Drawer is already open. I implemented the changes in e65ce2fa7a7ab1cce51a1933450bd7480fb60dac. Can you check if they help? Otherwise, how would you apply debounce here while making the Drawer to open as fast as possible when clicking on the button?

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
Aljullu: Yes, I noticed that as well. I opened an issue to track this, because I think the issue might be in the GB component we use: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4618.

Aljullu:
Thanks for the review and the suggestions @ralucaStan! This is ready for another look.

assets/js/base/components/drawer/style.scss
ralucaStan: The animation is much smoother now 👏

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
ralucaStan: I think this was the way to go, you make a good point with the debouncing and losing the performance of a fast opening Drawer.

Regarding the bug I've mentioned here is [a video I made](https://www.loom.com/share/7818572367364648b2ffa250a44f5ebb) of it. I think this is an edge case, so if you like to take a look at it in another ticket feel free. I am gonna approve this PR.

ralucaStan (approved):
Ready to 🚢 .

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
Aljullu: We discussed this via Slack and decided to merge this PR as-is and investigate further optimizations later. What's happening here is that the markup is rendered before the `mini-cart-frontend.js` script is loaded, so the button is visible but not interactive for a short period of time (might depend based on network speed).

PR #4610, opened 2021-08-20
Update checkout route hooks to support extensions

mikejolley:
Adds a new hook to the Checkout Route which will be used to update orders based on API request data. This is part of #4448

The new hook is named `__experimental_woocommerce_blocks_checkout_update_order_from_request` and passes on the order object, and the request data. Extensions that pass in custom request data can use this to update the order.

This new hook cannot be tested directory so just check for regressions; place an order via checkout, and test requests to the /wc/store/checkout route. Unit tests must also pass.

### Changelog

> Introduced the `__experimental_woocommerce_blocks_checkout_update_order_from_request` hook to the Checkout Store API.


src/StoreApi/Routes/Checkout.php
```diff
@@ -2,12 +2,6 @@
 namespace Automattic\WooCommerce\Blocks\StoreApi\Routes;
 
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\InvalidStockLevelsInCartException;
-use \Exception;
```
mikejolley, line 5: I removed these and just prefixed them inline. We had a mixture below.

src/StoreApi/Routes/Checkout.php
```diff
@@ -288,46 +275,44 @@ protected function get_route_post_response( WP_REST_Request $request ) {
...
-	 * @return WP_Error WP Error object.
+	 * @return \WP_Error WP Error object.
 	 */
 	protected function get_route_error_response( $error_code, $error_message, $http_status_code = 500, $additional_data = [] ) {
-		$error_from_message = new WP_Error(
+		$error_from_message = new \WP_Error(
 			$error_code,
 			$error_message
 		);
-		switch ( $http_status_code ) {
```
mikejolley, line 298: This switch statement was pointless with 1 possible outcome.

src/StoreApi/Routes/Checkout.php
```diff
@@ -472,23 +457,35 @@ private function update_customer_from_request( WP_REST_Request $request ) {
...
+		 *
+		 * This hook gives extensions the chance to update orders based on the data in the request. This can be used in
+		 * conjunction with the ExtendRestAPI class to post custom data and then process it.
+		 *
+		 * @internal This Hook is experimental and may change or be removed.
+		 *
+		 * @param \WC_Order $order Order object.
+		 * @param \WP_REST_Request $request Full details about the request.
+		 */
+		do_action( '__experimental_woocommerce_blocks_checkout_update_order_from_request', $this->order, $request );
```
mikejolley, line 477: This is the new hook.

src/StoreApi/Routes/Checkout.php
```diff
@@ -608,39 +603,33 @@ private function get_request_payment_data( WP_REST_Request $request ) {
...
 	 *
-	 * @internal CreateAccount class includes feature gating logic (i.e. this may not create an account depending on build).
-	 * @internal Checkout signup is feature gated to WooCommerce 4.7 and newer; Because it requires updated my-account/lost-password screen in 4.7+ for setting initial password.
-	 *
 	 * @throws RouteException API error object with error details.
-	 * @param WP_REST_Request $request Request object.
+	 * @param \WP_REST_Request $request Request object.
 	 */
-	private function process_customer( WP_REST_Request $request ) {
-		if ( defined( 'WC_VERSION' ) && version_compare( WC_VERSION, '4.7', '>=' ) ) {
```
mikejolley, line 618: We no longer support versions lower than 4.7.

opr (approved):
Code changes all look good to me, the refactor you did to introduce null coalescing is cool, and I think the code is just as readable as it was before adding it.

Tested the cart and checkout flow and it's working fine for me. ✅

senadir (approved):
Code checks out and testing checks out.
We should start documenting those hooks, I couldn't find any docs except the feature flag file which I added this hook to in 359b70ae3
So I opened #4629 to track documentation progress.

PR #4630, opened 2021-08-25
Rename productLink to showProductLink

### LINKED ISSUE #3756: Rename productLink to showProductLink
For product elements atomic blocks image and title we use a prop called `productLink`.
This has true/false values and is used to determine if the element is an anchor or not. 

The current name does not reflect that the prop returns a Boolean value (rather a string) and adds confusion to the codebase.

I propose renaming the variable to `showProductLink`, to also align with other props, for example, `showSaleBadge`.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/43b2f2277b7edeb08bca9f3d0db71062f5c6c3b6/assets/js/atomic/blocks/product-elements/image/block.js#L142-L144

nielslange:
### Note

The atomic image and title blocks use a prop called `productLink` which has the type `boolean`. To ensure that developers understand that this prop contains a `boolean` instead of a `string`, @ralucaStan suggested renaming the prop `productLink` to `showProductLink`.

Fixes #3756

#### Accessibility

n/a

### Screenshots

n/a

#### Automated testing

1. Run `npm run test` to ensure that all test cases pass
2. Check out this PR
3. Run `npm run test` again to ensure that all test cases still pass 

#### Manual testing

1. Create a test page
2. Add the `All Products` block
3. Click the pencil icon to edit the block
4. Click on the product image
5. Activate/deactivate the `Link to Product Page` setting in the sidebar
6. Save the changes
7. Look up the test page to ensure that the `Link to Product Page` setting behaves as expected

### Changelog

n/a

### DIFF
--- a/assets/js/atomic/blocks/product-elements/image/test/block.test.js
+++ b/assets/js/atomic/blocks/product-elements/image/test/block.test.js
@@ -56,7 +56,7 @@ describe( 'Product Image Block', () => {
 		test( 'should render an anchor with the product image', () => {
 			const component = render(
 				<ProductDataContextProvider product={ productWithImages }>
-					<Block productLink />
+					<Block showProductLink />
 				</ProductDataContextProvider>
 			);
 
@@ -80,7 +80,7 @@ describe( 'Product Image Block', () => {
 		test( 'should render an anchor with the placeholder image', () => {
 			const component = render(
 				<ProductDataContextProvider product={ productWithoutImages }>
-					<Block productLink />
+					<Block showProductLink />
 				</ProductDataContextProvider>
 			);
 
@@ -103,7 +103,7 @@ describe( 'Product Image Block', () => {
 		test( 'should render the product image without an anchor wrapper', () => {
 			const component = render(
 				<ProductDataContextProvider product={ productWithImages }>
-					<Block productLink={ false } />
+					<Block showProductLink={ false } />
 				</ProductDataContextProvider>
 			);
 			const image = component.getByTestId( 'product-image' );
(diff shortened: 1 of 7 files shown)


assets/js/atomic/blocks/product-elements/image/test/block.test.js
```diff
@@ -56,7 +56,7 @@ describe( 'Product Image Block', () => {
 		test( 'should render an anchor with the product image', () => {
 			const component = render(
 				<ProductDataContextProvider product={ productWithImages }>
-					<Block productLink />
+					<Block showProductLink />
```
Aljullu: Not introduced in this PR, but suggesting it here in case you want to take the chance and improve it. :slightly_smiling_face:  We always try to be explicit about prop values. In this case, it would be:
```suggestion
					<Block showProductLink={ true } />
```
This makes the code easier to understand, because it's easy to get confused by the way React and HTML treat omitted prop/attribute values and the way JavaScript treats object properties. Let me explain: React props (and HTML attributes) without an explicit value default to true (like in this case), while JS object properties default to a variable with the same name (ie: `{ showProductLink }` = `{ showProductLink: showProductLink }`).

Aljullu (approved):
Works like a charm, @nielslange! I left one suggestion to make code a bit clearer, in case you want to pick it up.

assets/js/atomic/blocks/product-elements/image/test/block.test.js
nielslange: Thanks for your review and pointing out to mention the props value explicitly. I applied your suggested change and merged the PR accordingly.

PR #4631, opened 2021-08-25
Extract function from lazyLoadScript to simplify code

Aljullu:
Follow-up from #4510.

Small refactor of the `lazyLoadScript` util. While this PR adds several lines of code, I think the code cleanup it contains makes it easier to understand what the code is doing.

2. Visit that post or page in the frontend and verify the Mini Cart button appears and, when hovered, Cart contents are rendered.

### DIFF
--- a/assets/js/base/utils/lazy-load-script.ts
+++ b/assets/js/base/utils/lazy-load-script.ts
@@ -63,7 +63,7 @@ const appendScript = ( attributes: appendScriptAttributesParam ): void => {
 		}
 	}
 
-	// Now we've assigned all the strings, we can explicitly assign to the
+	// Now that we've assigned all the strings, we can explicitly assign to the
 	// function keys.
 	if ( typeof attributes.onload === 'function' ) {
 		scriptElement.onload = attributes.onload;
--- a/assets/js/base/utils/lazy-load-script.ts
+++ b/assets/js/base/utils/lazy-load-script.ts
@@ -42,8 +42,34 @@ const appendScript = ( attributes: appendScriptAttributesParam ): void => {
 	}
 	const scriptElement = document.createElement( 'script' );
 	for ( const attr in attributes ) {
+		// We could technically be iterating over inherited members here, so
+		// if this is the case we should skip it.
+		if ( ! attributes.hasOwnProperty( attr ) ) {
+			continue;
+		}
 		const key = attr as keyof appendScriptAttributesParam;
-		scriptElement[ key ] = attributes[ key ];
+
+		// Skip the keys that aren't strings, because TS can't be sure which
+		// key in the scriptElement object we're assigning to.
+		if ( key === 'onload' || key === 'onerror' ) {
+			continue;
+		}
+
+		// This assignment stops TS complaining about the value maybe being
+		// undefined following the isString check below.
+		const value = attributes[ key ];
+		if ( isString( value ) ) {
+			scriptElement[ key ] = value;
+		}
+	}
+
+	// Now we've assigned all the strings, we can explicitly assign to the
+	// function keys.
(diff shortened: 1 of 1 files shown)


assets/js/base/utils/lazy-load-script.ts
```diff
@@ -18,6 +31,23 @@ const isScriptTagInDOM = ( scriptId: string ): boolean => {
...
+		return;
+	}
+	const scriptElement = document.createElement( 'script' );
+	for ( const attr in attributes ) {
+		const key = attr as keyof appendScriptAttributesParam;
+		scriptElement[ key ] = attributes[ key ];
+	}
+	document.body.appendChild( scriptElement );
+};
+
```
Aljullu: :point_up_2: This is the code that matters. Given an object of attributes, it creates an `<script>` element with those attributes. It works great, but TypeScript shows an error in line 46:
```JS
scriptElement[ key ] = attributes[ key ];
```
The error:
> Type 'string | OnErrorEventHandlerNonNull | (() => void) | undefined' is not assignable to type 'string & OnErrorEventHandlerNonNull & ((this: GlobalEventHandlers, ev: Event) => any)'.
>  Type 'undefined' is not assignable to type 'string & OnErrorEventHandlerNonNull & ((this: GlobalEventHandlers, ev: Event) => any)'.
>    Type 'undefined' is not assignable to type 'string'.ts(2322)

I feel that TypeScript gets confused by the `for ... in` and it warns that we might be using the wrong attribute values in a script tag.

Some things that I tried:
* Replacing `for ... in` with `Object.keys()` or `Object.entries()`. The issue persists.
* Explicitly setting the `attributes[ key ]` type, but I don't know how to do that inside a loop. We would need to read the type of each specific attribute from `appendScriptAttributesParam`.
* Using [`ScriptHTMLAttributes`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1349b640d4d07f40aa7c1c6931f18e3fbf667f3a/types/react/index.d.ts#L2228) from DefinitelyTyped. I couldn't get it to work, and those types are React-oriented (`onerror` would be `onError`, `innerHTML` doesn't exist...) so probably not something we can use in this case.

Any other ideas on how to solve this? :slightly_smiling_face:

Aljullu: Thanks to @opr for fixing this in 7323018acb45ef301cc8c5dcaaef1e8e4566f286. :partying_face:

assets/js/base/utils/lazy-load-script.ts
```diff
@@ -18,6 +31,49 @@ const isScriptTagInDOM = ( scriptId: string ): boolean => {
 	return scriptElements.length > 0;
 };
 
+/**
+ * Appends a script element to the document body if an script with the same id
```
frontdevde:
```suggestion
 * Appends a script element to the document body if a script with the same id
```

assets/js/base/utils/lazy-load-script.ts
```diff
@@ -18,6 +31,49 @@ const isScriptTagInDOM = ( scriptId: string ): boolean => {
...
+
+		// This assignment stops TS complaining about the value maybe being
+		// undefined following the isString check below.
+		const value = attributes[ key ];
+		if ( isString( value ) ) {
+			scriptElement[ key ] = value;
+		}
+	}
+
+	// Now we've assigned all the strings, we can explicitly assign to the
```
frontdevde:
```suggestion
	// Now that we've assigned all the strings, we can explicitly assign to the
```

frontdevde (approved):
Nice addition, adding the `appendScript` function definitely makes the `lazyLoadScript` function easier to grok.

Code LGTM and functionality also worked as expected 👍

PR #4655, opened 2021-08-31
Add missing checkout i2 inner block icons

### LINKED ISSUE #4573: Checkout i2: Blocks missing icons
Just tracking which icons are currently missing in the i2 prototype. 

**Checkout Fields Block** 

This is a wrapper Block for the left column.

**Order summary**

The Block that holds totals.

**Checkout totals block**

This is a wrapper Block for the right column.

**Checkout actions**

Holds the place order button.

—

cc @vivialice

If we had a generic icon for "column" we could apply that to the checkout totals and checkout fields wrapper blocks? We could use the one in the Gutenberg columns Block:

mikejolley:
Ensures all inner blocks have an icon. This applies column icons to the fields/totals areas, a button icon for checkout actions, and uses the "fields" icon for checkout itself to distinguish it from payments. Custom icons are taken from Figma.

Fixes #4573

### Screenshots

Screenshot 2021-08-31 at 11 10 54
Screenshot 2021-08-31 at 11 10 48
Screenshot 2021-08-31 at 11 10 43

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout-i2/index.tsx
+++ b/assets/js/blocks/cart-checkout/checkout-i2/index.tsx
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Icon, card } from '@woocommerce/icons';
+import { Icon, fields } from '@woocommerce/icons';
 import { registerFeaturePluginBlockType } from '@woocommerce/block-settings';
 import { createBlock } from '@wordpress/blocks';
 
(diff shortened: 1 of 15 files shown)


opr (approved):
Looks great icon-wise, but I'm just wondering about this block:

| Rendered block | Sidebar |
|---|---|
| image | image |

Its title is "Shipping options" but in the sidebar it is referred to as "Shipping methods" I don't think this is a huge deal, but it would be nice to name them the same

The payment block is the opposite:

| Rendered block | Sidebar |
|---|---|
| image |  image |

PR #4659, opened 2021-08-31
Expose the total number of items in each shipping package in the cart response

### LINKED ISSUE #4624: Investigate how to make Package Item Count data available on the client for extensions
The [Conditional Shipping and Payments](https://woocommerce.com/products/conditional-shipping-and-payments/) extension needs to have access to `Package Item Count` data on the client-side. We need to investigate how this data can be surfaced to the client.

opr:
Adds a key called `total_item_count` to each `shipping_rate` in the cart response. This is an integer, the value of which is the total quantity of all items in the package. This change is desirable because it will remove the need of doing this calculation on the client-side for each package.

Fixes #4624 cc @ralucaStan 

1. Add some items to your cart, add several different items, including ones that are virtual!
2. Change the quantity on a few items.
2. Go to the checkout and observe the cart response in dev tools, ensure the counts in `/wc/store/cart` > `shippingRates`  are correct. The count should only be the total number of _shippable_ products, so
3. Add this plugin: https://woocommerce.com/products/woocommerce-advanced-shipping-packages/
4. Set it up like so (click each image to see the full size):
- image
- image
- image
5. Check the cart again, ensure you have products in the cart that will cause it to split into two packages (some below $10 and some above $10).
6. Check the cart response again and ensure the `shippingRates` contains the correct number of products in each package.
7. Check the PHP unit tests pass `npm run test:php`

### Changelog

> Conveniently expose the total number of items in each shipping package in the cart response.

### DIFF
--- a/assets/js/types/type-defs/cart-response.ts
+++ b/assets/js/types/type-defs/cart-response.ts
@@ -75,6 +75,7 @@ export interface CartResponseShippingRate {
 	name: string;
 	destination: ResponseBaseAddress;
 	items: Array< ShippingRateItem >;
+	total_item_count: number;
 	shipping_rates: Array< CartResponseShippingPackageShippingRate >;
 }
 
@@ -192,7 +193,6 @@ export interface CartResponseExtensionItem {
 export interface CartResponse {
 	coupons: Array< CartResponseCouponItem >;
 	shipping_rates: Array< CartResponseShippingRate >;
-	package_item_counts: Array< number >;
 	shipping_address: CartResponseShippingAddress;
 	billing_address: CartResponseBillingAddress;
 	items: Array< CartResponseItem >;
--- a/assets/js/types/type-defs/cart-response.ts
+++ b/assets/js/types/type-defs/cart-response.ts
@@ -192,6 +192,7 @@ export interface CartResponseExtensionItem {
 export interface CartResponse {
 	coupons: Array< CartResponseCouponItem >;
 	shipping_rates: Array< CartResponseShippingRate >;
+	package_item_counts: Array< number >;
 	shipping_address: CartResponseShippingAddress;
 	billing_address: CartResponseBillingAddress;
 	items: Array< CartResponseItem >;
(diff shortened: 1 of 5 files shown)


senadir:
This feels a bit redundant to have. Why couldn’t we just count it using `items.length`?

opr:
> This feels a bit redundant to have. Why couldn’t we just count it using `items.length`?

Because this is not `items.length` this is the total of each `item.quantity` in the package. Yes, we _can_ do this on the client side but it seems wasteful to have to calculate this each time we need it, we can calculate it once on the server side and be done with it.

senadir:
I get that the calculation is wasteful, but it's just a simple `items.reduce((total, item) => total + item.quantity, 0)`, it's cheap to run on the client for who ever needs it. My objection is mostly that we're increasing the surface area of our API by redundant for a very niche use case for a third party plugin. Our Cart response already has a lot of redundant data in several places and I don't think we should be contributing to it.

opr:
Sure I understand that, and agree with you, but I also see it from a developer's point of view. If I were developing an extension I would _probably_ expect the total number of items to be included in this response and not have to derive it from other data. 

By not including this, it requires extension developers to look into our cart response in more detail to figure out how to calculate it. I don't think is necessarily a bad thing but it's certainly an extra (small) hurdle for a developer to overcome, and we should be making it as easy as possible.

senadir:
I see your point of view but it's build on assumption, not actual feedback, and it's built on assumption that I don’t see. As a developer, I don't expect items count to be included, it doesn’t make sense to me to have that value.
For developers to do anything meaningful, they have to look at our cart response in detail to know what they're dealing with anyway.

By opening the gate to those kind of additions, we're only increasing our public API surface area, response size, and amount of things to be on top of.

mikejolley:
Thanks for cc'ing me in @opr. I can see the arguments for and against this. In general, I prefer to leave data out that can be derived, however, we do include useful total values in endpoints when values could be confusing or need special handling. Qty of Items I think falls into this category. Looking at other (other platform) APIs, they seem to do similar things. 

Can we name it `item_count`?

opr:
Closing after discussing in Slack p1630492044143900-slack-C8X6Q7XQU if this really must be added then we can revisit.

PR #4668, opened 2021-09-01
Add extensibility point for extensions to filter payment methods

### LINKED ISSUE #4621: Implement an extensibility point for filtering the payment methods
In order for extensions to filter the payment methods we will need to implement a new extensibility point on the JavaScript side. The filtering method should be called each time there is a change in the checkout block: billing/shipping addresses, shipping method or coupon code.

For the extensibility point, we could enhance the registration functions for payment methods (along with the PaymentMethodConfig js class) to allow for extensions to register additional `canMakePayment` callbacks for specific payment methods. `canMakePayment` callbacks are called when changes are made to the checkout, so this would ensure the need of the extensions to refilter payment methods on state change.

ralucaStan:
This PR introduces a new way for extensions to register their own callbacks to determine wheater a payment method is supported.
Fixes #4621 
[Tests](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4622), [documentation](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4623), and the [Typescript conversion](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4687) will be added in follow-up tickets.

1. Make sure Check payments and Cash on delivery payment methods are activated and appear in the Checkout block
2. Import the registration function somewhere in an extension
```js
import { registerPaymentMethodExtensionCallbacks} from '@woocommerce/blocks-registry';
```
3. Add some callbacks 
```js
	{
		cod: ( arg ) => { return false; },
                 // this zipcode works only for Germany
		cheque: ( arg ) => { return arg.billingData.postcode === "10473";}
	}
);
```
4. Go to the Checkout block
3. Notice that Cash on delivery is no longer available
4. Notice that Check payments are available only when the Postcode for Billing data is 10473

### Changelog

> Add extensibility point for extensions to filter payment methods.


assets/js/blocks-registry/payment-methods/can-make-payment-extensions.js
```diff
@@ -0,0 +1,7 @@
+// Keeps callbacks registered by extensions for different payment methods
```
nerrad: Assuming we want this config object to be something we can add additional functionality to in the future, it might be good to name the file something like `extensions-config.js` instead.

Also, maybe do this in TS from the get-go?

nerrad (commented):
Generally seems to look pretty good!

- This might be a good opportunity to convert the files to TS.
- I'd like to see some automated test coverage for the new extension logic.

I didn't test.

ralucaStan:
> * I'd like to see some automated test coverage for the new extension logic.

There are separate tickets for [tests](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4622) and documentation for this logic.

assets/js/blocks-registry/payment-methods/payment-method-config-helper.js
```diff
@@ -8,3 +8,20 @@ export const canMakePaymentWithFeaturesCheck = ( canMakePayment, features ) => (
 	);
 	return featuresSupportRequirements && canMakePayment( canPayArgument );
 };
+
+// Filter out payment methods by callbacks registered by extensions
+export const canMakePaymentWithExtensions = (
+	canMakePayment,
+	extensionsCallbacks
+) => ( canPayArgument ) => {
+	// validate whether the payment method is available
```
opr: Change to use a capital letter and full stop

assets/js/blocks-registry/payment-methods/payment-method-config-helper.js
```diff
@@ -8,3 +8,20 @@ export const canMakePaymentWithFeaturesCheck = ( canMakePayment, features ) => (
 	);
 	return featuresSupportRequirements && canMakePayment( canPayArgument );
 };
+
+// Filter out payment methods by callbacks registered by extensions
```
opr:
```suggestion
// Filter out payment methods by callbacks registered by extensions.
```

assets/js/blocks-registry/payment-methods/payment-method-config-helper.js
opr:
```suggestion
	// Validate whether the payment method is available.
```

assets/js/blocks-registry/payment-methods/registry.js
```diff
@@ -67,6 +67,33 @@ export const registerExpressPaymentMethod = ( options ) => {
 	}
 };
 
+/**
+ * Registers a callback for a specific payment method to determine if it can make payments
+ *
+ * @param {string} paymentMethodName A unique string to identify the payment method client side.
+ * @param {function():any } callBack Callback defined by extensions to determine if the payment method is supported.
+ */
+export const registerPaymentMethodExtensionCallback = (
```
opr: I am wondering what you think of changing the API to bring it more in line with the filters API. 

The API over there is like so: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/131cb22b9c84574a88b6a4e411fabade5d1ed53c/packages/checkout/registry/index.ts#L34-L37

I was thinking we could do something similar here and change this to look like:

```js
export const registerPaymentMethodExtensionCallbacks = (
	namespace: string,
	paymentMethods: Record< string, PaymentMethodExtensionCallback >,
) 
```

Extensions would then use it like this:

```js
registerPaymentMethodExtensionCallbacks( 'test-extension'
	{
		stripe: ( arg ) => { /* code */ },
		paypal: ( arg ) => { /* code */ },
		cheque: ( arg ) => { /* code */ }
	}
);
```

ralucaStan: I think it's a good suggestion for consistency reasons and also I see a great benefit in tracing back the extensions that declared those callbacks with the namespace. That would increase the complexity a bit of gathering the callbacks, but we'd also have the namespace available to track down the extensions. What do you think?

opr: I was about to write the exact same comment. It's good for error reporting since we can show what extension caused the error. There is a negative tradeoff in that we would have to get the callbacks out of the namespaces in the first place, but this is only minor, and is outweighed by the positives.

ralucaStan: @opr I have added the changes and re-requested review.

assets/js/blocks-registry/payment-methods/payment-method-config.js
```diff
@@ -39,10 +42,24 @@ export default class PaymentMethodConfig {
 			showSaveOption: config?.supports?.showSaveOption || false,
 			features: config?.supports?.features || [ 'products' ],
 		};
-		this.canMakePayment = canMakePaymentWithFeaturesCheck(
-			config.canMakePayment,
+		this.canMakePaymentFromConfig = config.canMakePayment;
+	}
+
+	get canMakePayment() {
```
opr, line 49: For posterity, please could you add the reason why this was moved to a getter in the PR description? We discussed this in a huddle, and I agree with your reasoning 👍🏼

assets/js/blocks-registry/payment-methods/payment-method-config.js
```diff
@@ -39,10 +42,24 @@ export default class PaymentMethodConfig {
...
+		const canPay = canMakePaymentWithFeaturesCheck(
+			this.canMakePaymentFromConfig,
 			this.supports.features
 		);
+
```
opr: I think we'd benefit from describing the purpose of the next statement. You can figure it out by reading, but this makes it a bit quicker :)

```suggestion
// Loop through all calbacks to check if there are any registered for this payment method.
```

assets/js/blocks-registry/payment-methods/registry.js
```diff
@@ -67,6 +67,44 @@ export const registerExpressPaymentMethod = ( options ) => {
...
+		Object.values( callbacks ).some(
+			( callback ) => typeof callback !== 'function'
+		)
+	) {
+		const nonFunctionCallbacks = Object.keys( callbacks ).filter(
+			( key ) => typeof callbacks[ key ] !== 'function'
+		);
+
+		throw new Error(
+			`All callbacks provided to registerPaymentMethodExtensionCallbacks must be functions. The callbacks for :${ nonFunctionCallbacks.join(
```
opr:
```suggestion
			`All callbacks provided to registerPaymentMethodExtensionCallbacks must be functions. The callbacks for: ${ nonFunctionCallbacks.join(
```

The formatting is slightly off here
image

assets/js/blocks-registry/payment-methods/registry.js
```diff
@@ -67,6 +67,44 @@ export const registerExpressPaymentMethod = ( options ) => {
...
+				', '
+			) } were not functions. Error occurred when trying to register callbacks for the ${ namespace } namespace.`
+		);
+	}
+
+	// Set namespace up as an empty object
```
opr:
```suggestion
	// Set namespace up as an empty object.
```

assets/js/blocks-registry/payment-methods/registry.js
```diff
@@ -67,6 +67,44 @@ export const registerExpressPaymentMethod = ( options ) => {
...
+	if (
+		Object.values( callbacks ).some(
+			( callback ) => typeof callback !== 'function'
+		)
+	) {
+		const nonFunctionCallbacks = Object.keys( callbacks ).filter(
+			( key ) => typeof callbacks[ key ] !== 'function'
+		);
+
+		throw new Error(
```
opr: Currently, if one extension registers a bad callback, it stops all subsequent extensions from registering theirs.

Instead of throwing an error here, let's just `console.error` and return early from the function?

I don't think there's anywhere further up the stack that we'd benefit from catching the error.

You can simulate this by adding another call to `registerPaymentMethodExtensionCallbacks` immediately after the initial one, ensuring one of the callbacks is not a function.

```js
registerPaymentMethodExtensionCallbacks( 'oprs-marketplace-extension',
		{
			cod: ( arg ) => { return false; },
			stripe: null,
		}
	);
```

assets/js/blocks-registry/payment-methods/payment-method-config-helper.js
```diff
@@ -8,3 +8,41 @@ export const canMakePaymentWithFeaturesCheck = ( canMakePayment, features ) => (
...
+			try {
+				return namespacedCallbacks[ namespace ]( canPayArgument );
+			} catch ( err ) {
+				// eslint-disable-next-line no-console
+				console.error(
+					`Error when executing callback for ${ paymentMethodName } in ${ namespace }`,
+					err
+				);
+				// every expects a return value at the end of every arrow function and
+				// this ensures that the error is ignored when computing the whole result
```
opr:
```suggestion
				// this ensures that the error is ignored when computing the whole result.
```

assets/js/blocks-registry/payment-methods/payment-method-config-helper.js
```diff
@@ -8,3 +8,41 @@ export const canMakePaymentWithFeaturesCheck = ( canMakePayment, features ) => (
...
 	return featuresSupportRequirements && canMakePayment( canPayArgument );
 };
+
+// Filter out payment methods by callbacks registered by extensions.
+export const canMakePaymentWithExtensions = (
+	canMakePayment,
+	extensionsCallbacks,
+	paymentMethodName
+) => ( canPayArgument ) => {
+	// Validate whether the payment method is available.
```
opr:
```suggestion
	// Validate whether the payment method is available based on its own criteria first.
```

opr (changes requested):
This is fantastic Raluca 👏🏼  I have added a few (admittedly pedantic) comments about comment punctuation. 

I tested with:

- Two different extensions adding valid callbacks. ✅ 
- Two different extensions adding valid callbacks, one returning true for a payment metod, and one returning false for the same, I expected NOT to see the payment method ✅ 
- An extension registering a callback that is not a function - An error is thrown (see comments for improvements here) - If two extensions try to register callbacks, but the first one tries registering a callback that isn't a function, then the second one doesn't get chance to register. ❌ 
- If a callback that is correctly registered, but for some reason is causing an error, other extensions' callbacks can still run ✅ 

For reference, here are the two registrations I'm using:

```js
		{
			cod: ( arg ) => { return false; },
                         // A callback that is not a function. This should cause an error, but other extensions should still be allowed to register.
                         cheque: null,
			stripe: () => { aaa.bbb = 1; return true; }
		}
	);

	registerPaymentMethodExtensionCallbacks( 'oprs-marketplace-extension',
		{
                         // COD should not show because the first callback returns false.
			cod: ( arg ) => { return true; },
                         // Because the stripe callback above causes an error, it should be skipped. Stripe should only show when postcode is 10437 (in Germany).
			stripe: ( arg ) => arg.billingData.postcode === '10473',
                         // Because cheque registration should have failed above, we should still be allowed to see cheque in the payment method list.
			cheque: () => { return true },
		}
	);
```

assets/js/blocks-registry/payment-methods/registry.js
ralucaStan: @opr 👏 for spotting this. 
I am thinking we should not return early, just use `console.error` and proceed to the other callbacks the extension might have. So if one callback is not a function we are registering the ones that actually are. Like so:
```js
export const registerPaymentMethodExtensionCallbacks = (
	namespace,
	callbacks
) => {
	// Set namespace up as an empty object.
	canMakePaymentExtensionsCallbacks[ namespace ] = {};

	Object.entries( callbacks ).forEach(
		( [ paymentMethodName, callback ] ) => {
			if ( typeof callback === 'function' ) {
				canMakePaymentExtensionsCallbacks[ namespace ][
					paymentMethodName
				] = callback;
			} else {
				console.error(
					`All callbacks provided to registerPaymentMethodExtensionCallbacks must be functions. The callback for the ${ paymentMethodName } payment method was not a function. Error occurred when trying to register callbacks for the ${ namespace } namespace.`
				);
			}
		}
	);
};
```

But I do think we should return early if the namespace is already used. So instead of letting a call of registerPaymentMethodExtensionCallbacks that overwrites existing callbacks under the same namespace, just console.error and return. What do you think?

opr: > I am thinking we should not return early

OK :)

> But I do think we should return early if the namespace is already used. So instead of letting a call of registerPaymentMethodExtensionCallbacks that overwrites existing callbacks under the same namespace, just console.error and return. What do you think?

Great idea!

assets/js/blocks-registry/payment-methods/registry.js
```diff
@@ -67,6 +67,42 @@ export const registerExpressPaymentMethod = ( options ) => {
...
+		Object.entries( callbacks ).forEach(
+			( [ paymentMethodName, callback ] ) => {
+				if ( typeof callback === 'function' ) {
+					canMakePaymentExtensionsCallbacks[ namespace ][
+						paymentMethodName
+					] = callback;
+				} else {
+					// eslint-disable-next-line no-console
+					console.error(
+						`All callbacks provided to registerPaymentMethodExtensionCallbacks must be functions. The callback for the ${ paymentMethodName } payment method was not a function. Error occurred when trying to register callbacks for the ${ namespace } namespace.`
```
opr:
```suggestion
						`All callbacks provided to registerPaymentMethodExtensionCallbacks must be functions. The callback for the ${ paymentMethodName } payment method in the ${ namespace} namespace was not a function.`
```

opr (approved):
This is looking great @ralucaStan - I have made one final suggestion about the error message when registering non-function callbacks. After looking at it for some time it felt a little clunky. My suggestion makes it a bit easier to read imo.

Approving because this change is non-blocking and the rest of the PR is working fine now 🥳

opr (commented):
Actually, before we merge, I am wondering about the name of `registerPaymentMethodExtensionCallbacks` why isn't this prefixed with `__experimental`?

Should it be renamed before merging?

nerrad:
> Actually, before we merge, I am wondering about the name of registerPaymentMethodExtensionCallbacks why isn't this prefixed with __experimental?

In this particular case I think I'm okay with skipping experimental for this interface. It's fairly straightforward and I think other alternatives might have more riskiness/uncertainty attached to them. This is just an opinion though - so I'm happy to defer to others here.

ralucaStan:
> Actually, before we merge, I am wondering about the name of `registerPaymentMethodExtensionCallbacks` why isn't this prefixed with `__experimental`?
> 
> Should it be renamed before merging?

I also thought about this, but after looking at the history of the registry file I saw that the exports didn't have an experimental prefix.

PR #4671, opened 2021-09-02
Refactor Checkout Inner Block Registration and Render frontend forced blocks

### LINKED ISSUE #4574: Checkout i2: Forced Inner Blocks/Required Blocks
Spinning this out of #4447

> third-party blocks are a couple of types:
>
> - Optional blocks that merchants can add after enabling the plugin.
> - Required blocks that must be available after enabling the plugin.

We already have a system for registering new optional blocks, and inner block areas for each form step, so the first point is covered.

For _required blocks_ we need a new system so extensions can force insert blocks into the existing checkout layout. Developers decide where the block needs to be inserted, and then it will be locked so it cannot be removed. If it is somehow removed, or missing, the block will be inserted by the system. If an extension is removed, the block needs to be removed by the system.

Todo:

- [x] Add a _force_ attribute to the inner block registration system #4606
- [x] Missing forced blocks must be force inserted by the editor #4606
- [ ] Missing forced blocks must be force inserted on render
- [ ] When an extension is removed, forced/invalid blocks must be removed without causing a block invalidation error
- [x] Template locking should be handled automatically #4606

cc @senadir

mikejolley:
This PR takes care of forced block rendering on the frontend. It does so by looking up forced blocks for an area, and appending it if there are no existing siblings with the same block name.

To achieve this, the registration system has been updated to accept `force` and store `block`, `component`, and `force`. These values are then looked up as needed.

This also makes it possible to register a checkout block without defining it on the PHP side. To do this, I used the`render_block` filter to append `data-block-name`. This applies to all blocks however, not just our blocks. I don't anticipate this causing any issues as it's just an attribute.

Closes #4574

Forced blocks

1. Edit the sample block and make it non forced 
2. Create a checkout block and save it with no sample blocks
3. Make sample block forced again
4. View frontend and confirm sample block is appended to the layout

Block registration

1. Confirm checkout i2 renders
2. Confirm single product block still renders, and can be updated/persisted without regression
3. Confirm all products block still renders, and can be updated/persisted without regression


packages/checkout/blocks-registry/register-checkout-block.ts
```diff
@@ -0,0 +1,77 @@
...
+								type: 'object',
+								default: {
+									remove: true,
+								},
+							},
+					  }
+					: {} ),
+			},
+			category: 'woocommerce',
+			parent: [],
```
senadir: This means someone can't override `lock.remove`. It also overrides `lock.move`. Can't think of a block that would want `lock.remove: false` and `force: true` but all of our blocks are `lock.move: true` and this would override them.

packages/checkout/blocks-registry/test/index.js
```diff
@@ -1,7 +1,11 @@
 /**
  * Internal dependencies
  */
-import { getRegisteredBlocks, registerCheckoutBlock } from '../index';
+import {
+	getRegisteredBlocks,
+	registerCheckoutBlock,
+	InnerBlockAreas,
```
senadir: Any reason why `InnerBlockAreas` instead of `innerBlockAreas` It might get confused for a component.

packages/checkout/blocks-registry/registered-blocks.ts
```diff
@@ -0,0 +1,46 @@
...
+			component: null,
+			force: false,
+		},
+	],
+	'woocommerce/checkout-billing-address-block': [
+		{
+			block: 'core/paragraph',
+			component: null,
+			force: false,
+		},
```
senadir, line 28: Planning to keep paragraph or just for testing?

src/BlockTypesController.php
```diff
@@ -66,6 +67,57 @@ public function register_blocks() {
...
+
+		/**
+		 * WooCommerce Blocks Block Names
+		 *
+		 * This hook defines which block names should have block name and attribute data- attributes appended on render.
+		 *
+		 * @param array $allowed_namespaces List of namespaces.
+		 */
+		$allowed_blocks = (array) apply_filters( '__experimental_woocommerce_blocks_add_data_attributes_to_block', [] );
+
```
senadir, line 98: Did you take into account core blocks when writing this? I know it shouldn’t effect them since they don't save attributes in data but just noting it here.

Paragraph seems to regress for me, our attributes that we provide for third party components are persisting as regular attributes in paragraph 
image

senadir: Not to mention text doesn’t persist.

packages/checkout/blocks-registry/register-checkout-block.ts
```diff
@@ -0,0 +1,77 @@
...
+							lock: {
+								type: 'object',
+								default: {
+									remove: true,
+								},
+							},
+					  }
+					: {} ),
+			},
+			category: 'woocommerce',
```
senadir, line 45: Do we want to force this category for third party blocks? It means blocks would show up with us in the large inspector?

assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -43,53 +111,46 @@ const renderInnerBlocks = ( {
...
+							depth: depth + 1,
+							blockWrapper,
+						} ) }
+						{ renderForcedBlocks(
+							blockName,
+							blockMap,
+							element.children
+						) }
+					</InnerBlockComponent>
+				</InnerBlockComponentWrapper>
```
senadir: This got even more complicated you’re now the only one who can maintain it so we need to document what's happening here later.

senadir (approved):
This is looking good, except some minor docs for later and all what we fixed in chat, this is ready to merge.

src/BlockTypesController.php
senadir: Paragraph not working is now fixed but just wanted to make sure you had that in mind when you wrote this.

packages/checkout/blocks-registry/registered-blocks.ts
mikejolley: I think it might be useful to include some core blocks like Paragraph and Image. I think we'll need to discuss this before going live.

src/BlockTypesController.php
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4671/commits/1afa3e11067e69ee908a3f56fe3b33583d4de17f

packages/checkout/blocks-registry/register-checkout-block.ts
mikejolley: These blocks are disabled in the large inspector because parent is blank. This groups them in the inner block inspector, which I think is good, but we should let it be overriden.

senadir:
There's also this mild CSS issue that we can fix later (extra spacing below steps).
image

packages/checkout/blocks-registry/test/index.js
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4671/commits/80e225c57f841b45de6c4a2a0bae796001f34945

assets/js/atomic/utils/render-parent-block.tsx
mikejolley: Added more explanation in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4671/commits/47b4f328807cc05dae426130e21bbfd2489d733c to make it easier to follow.

packages/checkout/blocks-registry/register-checkout-block.ts
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4671/commits/c27cf0836e42dd73dcfc3292e5f8681630379e9c

packages/checkout/blocks-registry/register-checkout-block.ts
mikejolley: Fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4671/commits/c27cf0836e42dd73dcfc3292e5f8681630379e9c

senadir (approved):
Did a second review. Inline docs are wonderful! they actually helped me figure out what was happening. :shipit:

PR #4673, opened 2021-09-02
Improve accessibility when removing items from cart

### LINKED ISSUE #4618: Mini Cart block: after removing a product, drawer can't be closed clicking outside
## To reproduce
In branch `add/drawer-component` (#4608):

2. Visit that post or page in the frontend and click on the Mini Cart button.
3. In the drawer that appears, remove one of the products from your cart.
4. Click on the overlay and notice the drawer doesn't close.
5. It can be workarounded clicking again somewhere inside the drawer and the outside again.

## Expected behavior
The drawer should close when clicking outside in all cases.

## Screenshots

https://user-images.githubusercontent.com/3616980/130471273-d7e7dbf0-a0e4-4713-925a-864505de6a2f.mp4

## Next steps

The Mini Cart block uses the [`Modal` Gutenberg component](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/modal), which internally uses the [`useFocusOutside()` hook](https://github.com/WordPress/gutenberg/blob/trunk/packages/compose/src/hooks/use-focus-outside/index.js), so the issue might be there.

Aljullu:
This PR introduces several accessibility improvements to the Cart and Mini Cart blocks:

* fdc92d2165bb43204849fcd9bce4c52f970c1cd1: adds a custom label to the Mini Cart drawer close button, so instead of the generic `Close dialog` now it will read `Close mini cart`.
* 4c831c51a502c210034e6769c8e360f0f2e84332: adds a screen reader notice when a product is removed from the cart.
* c6911d497b950472e75b15f12b1c0c7761007839: makes it so when an item is removed from cart, the next item is automatically focused. Until now, focus was going to the `<body>`, and that caused screen readers to read unnecessary data again. In parallel, that was causing the issue in #4618.

Fixes #4618.

#### Accessibility

For a better experience, you can test with the changes in Gutenberg from this PR: https://github.com/WordPress/gutenberg/pull/34684. It improves several focus flows when interacting with cart items and navigating with the keyboard.

**Cart block:**
1. Add several products to your cart and go to a page with the Cart block.
2. With a screen reader + using the keyboard to move around the page try to remove a couple of items.
3. Verify you can successfully remove them and when a product is removed, the screen reader announces it.

**Mini Cart block (you can skip these testing instructions from the release docs):**

1. Add several products to your cart and go to a page with the Mini Cart block.
2. With a screen reader + using the keyboard to move around the page try to remove a couple of items.
3. Verify you can successfully remove them and when a product is removed, the screen reader announces it.
4. Verify that right after clicking the `Remove item` button, you can click on the screen overlay to close the Mini Cart (fix for #4618).

### Changelog

> Accessibility improvements when removing items from cart.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
@@ -29,6 +29,20 @@ const CartLineItemsTable = ( {
 					<CartLineItemRow
 						key={ lineItem.key }
 						lineItem={ lineItem }
+						onRemove={ ( removedRow: HTMLElement | null ) => {
+							if (
+								removedRow?.nextElementSibling instanceof
+								HTMLElement
+							) {
+								removedRow.nextElementSibling.focus();
+							} else if (
+								removedRow?.previousElementSibling instanceof
+								HTMLElement
+							) {
+								removedRow.previousElementSibling.focus();
+							}
+						} }
+						tabIndex={ -1 }
 					/>
 				);
 		  } );
--- a/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
@@ -35,11 +35,6 @@ const CartLineItemsTable = ( {
 								HTMLElement
 							) {
 								removedRow.nextElementSibling.focus();
-							} else if (
-								removedRow?.previousElementSibling instanceof
-								HTMLElement
-							) {
-								removedRow.previousElementSibling.focus();
 							}
 						} }
 						tabIndex={ -1 }
--- a/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
+++ b/assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
@@ -3,7 +3,8 @@
  */
 import { __ } from '@wordpress/i18n';
 import { CartResponseItem } from '@woocommerce/type-defs/cart-response';
-import { useRef } from '@wordpress/element';
+import { createRef, useEffect, useRef } from '@wordpress/element';
+import type { RefObject } from 'react';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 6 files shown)


ralucaStan:
Hi Albert,
I've tested removing items with the keyboard. I had some products in the cart. This is what I've noticed:

- on the first tab, the Close button is focused, but I also see the a11y label 
Screenshot 2021-09-03 at 17 19 08

- after I delete one product using tabs the next item receives focus as expected, but when I tab again the Close button is again focused
- after I delete the last item in the cart and I tab again the focus in on the browser tabs

Expected behaviour would be:
- not see the a11y label
- when an item receives focus after deleting the previous and I tab, it should go to the inner elements of the item
- after I delete the last item the focus should go to the close button, or we might even close the drawer as there is no other action available. I suppose an empty cart message should also be available, in which case the focus should maybe go to the "Continue shopping" button.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
```diff
@@ -29,6 +29,20 @@ const CartLineItemsTable = ( {
...
+							) {
+								removedRow.nextElementSibling.focus();
+							} else if (
+								removedRow?.previousElementSibling instanceof
+								HTMLElement
+							) {
+								removedRow.previousElementSibling.focus();
+							}
+						} }
+						tabIndex={ -1 }
```
ralucaStan, line 45: why is this tabIndex neeed?

ralucaStan (commented):
I have left a couple of comments

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
ralucaStan, line 64: Preferably  `tabIndex={ -1 }` should be avoided. The tabbed navigation in the mini cart worked really well before this PR when navigating from one item the other  after deleting. 

I think removing items from the normal flow is the reason why the issues from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4673#issuecomment-912623764 happen. 

The only issue before this PR was focusing after removing the last item. I am thinking it would be easier just to move the focus to the mini cart wrapper, and not try to focus the previous item. What do you think?

ralucaStan:
If you would like, we could improve a11y a bit more if we also add in the code [attributes that say](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html#rps1_label) that the button for the mini cart opens this item (as long as we can add an id to it, like DRAWER_ID_EXAMPLE):
```
<div class="wc-block-components-drawer__screen-overlay wc-block-components-drawer__screen-overlay--is-hidden" aria-hidden="true">
</div>
```
```
<button
   class="wc-block-mini-cart__button"
   aria-haspopup="true"
   aria-controls="#DRAWER_ID_EXAMPLE"
   aria-expanded="true" // or false depending on the drawer being open
>2 items</button>
```

opr (commented):
As discussed here p1631012690384400-slack-C8X6Q7XQU we should add some text to make the button more descriptive to screen readers. Right now my screen reader just says "9 items" with no further context about what the button is for.

Aljullu:
Thanks for the reviews @ralucaStan and @opr!

> on the first tab, the Close button is focused, but I also see the a11y label

Fixed in 7239ed580a89c5987b7f0de007ce4e983af81677. :+1: 

> after I delete one product using tabs the next item receives focus as expected, but when I tab again the Close button is again focused

Good catch, that seemed to be an issue with `useConstrainedTabbing()` from Gutenberg. I opened a PR upstream that will fix this: https://github.com/WordPress/gutenberg/pull/34684.

> after I delete the last item in the cart and I tab again the focus in on the browser tabs

Should be fixed now if you test with the Gutenberg PR above.

> Preferably  `tabIndex={ -1 }` should be avoided. The tabbed navigation in the mini cart worked really well before this PR when navigating from one item the other after deleting. 

When I tested with a screen reader, after removing an item, the page title was being read by the screen reader. That was because `<body>` was getting the focus. That might depend on the screen reader being used, but it's one of the things I'm trying to fix here, and that's why I try to keep the focus on relevant elements all the time (and thus, why `tabIndex={ -1 }` is needed). I'm happy to discuss other solutions if you have any other idea, though. :)

> I am thinking it would be easier just to move the focus to the mini cart wrapper, and not try to focus the previous item. What do you think?

Good point, focusing the previous item didn't make much sense. I removed that code in aba15f218bd69f0aede99f3f7cda748edec17e0e.

> If you would like, we could improve a11y a bit more if we also add in the code [attributes that say](https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html#rps1_label) that the button for the mini cart opens this item (as long as we can add an id to it, like DRAWER_ID_EXAMPLE):

Yup, sorry for not answering that in the previous PR. There are a couple of things that make me reluctant to use these aria properties here:

* The button can't be focused when the Mini Cart is opened, so its utility might be reduced.
* According to the docs, those aria properties `Indicate the button opens a menu.`, but I don't think the Mini Cart drawer can be considered a menu, to me, it acts more as a dialog and in fact it currently has the `role="dialog"` that is inherited from the `<Modal>` component in Gutenberg.

I'm not 100% sure about what's the best approach here, but following the "no ARIA is better than bad ARIA" mantra, I would avoid adding aria attributes like these if they aren't fixing a specific issue (and we can't do cross-screen-reader testing). What do you think?

---

> we should add some text to make the button more descriptive to screen readers. Right now my screen reader just says "9 items" with no further context about what the button is for.

Opened an issue for this so it can be worked independently of this PR: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4704.

---

Note: it might be easier to review this PR with `Hide whitespace changes` now, because I needed to wrap `CartLineItemRow` with `forwardRef` and that caused big indentation changes.

ralucaStan (approved):
Congrats for the hard work on this 👏

PR #4675, opened 2021-09-02
Enqueue frontend assets on frontend only

### LINKED ISSUE #4609: Checkout i2: Checkout Frontend script loading in Editor
It seems like Checkout frontend script is loading in the editor. This is causing extra registration for some optional blocks (like sample block) and code running twice, which can lead to bugs, not mentioning performance.

mikejolley:
Some blocks (including checkout and checkout i2) were loading both frontend and editor assets in the admin area due to this line:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/341be1f56071fbd4b5ff975e8788d65a09512df2/src/BlockTypes/AbstractBlock.php#L84

This is there to ensure assets are loaded in time for block rendering, but this doesn't seem to be needed in the editor. Adding an `is_admin()` check fixes this.

This did break checkout-i2 sample-block but this was due to missing import. Also resolved.

Fixes #4609

1. Before patching:
  - Load a page containing checkout-i2 in the editor.
  - View source and search for a script named `wc-checkout-i2-block-frontend`
2. Apply the patch and repeat. No `wc-checkout-i2-block-frontend` should be present in the editor.
3. Confirm the block still renders in admin and frontend without regression
4. Smoke test the original cart/checkout blocks are also unaffected by the patch

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/index.tsx
+++ b/assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/index.tsx
@@ -13,3 +13,4 @@ import './checkout-order-summary-block';
 import './checkout-payment-block';
 import './checkout-express-payment-block';
 import './checkout-shipping-methods-block';
+import './checkout-sample-block';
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Good finding! Code changes look good and it's testing well

 I remember working on the same issue a year ago (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2788), but it looks like it was regressed, fortunatelly this PR fixes it. Thanks @mikejolley!

PR #4677, opened 2021-09-03
Add initial e2e tests to Mini Cart block

Aljullu:
Part of #4643.

PR adding some basic e2e tests to the Mini Cart block.

1. Verify tests pass in this PR.


opr (approved):
The tests pass, and it seems like they cover everything given there are no further options on the block, so ✅ from me!

Will we need front-end E2E tests, or can everything front-end wise be tested in JS Unit tests?

Aljullu:
Thanks for the review @opr!

> Will we need front-end E2E tests, or can everything front-end wise be tested in JS Unit tests?

Hopefully, we will be able to test all the frontend side with unit tests, but still difficult to say for sure.

PR #4678, opened 2021-09-03
Try namespacing callbacks added by extensions for `canMakePayment`

opr:
This PR won't be merged, it's an experiment to see how adding `cart` to `canPayArgument` affects things, as well as changing the API for how extensions will register their own callbacks to `canMakePayment`.

Cc @ralucaStan 

1. Add this code somewhere in an extension
2. 
```js
	{
		stripe: ( arg ) => { console.log("callback on woocommerce-marketplace-extension called", arg); throw new Error('Some error thrown by the extension'); return true; },
		cod: ( arg ) => { /* code */ return true; },
		cheque: ( arg ) => { /* code */ return true; }
	}
);
```
2. Check the `stripe` payment method doesn't register, and check the console for an error.
3. Remove the `throw new Error` part of the callback, and check stripe registers.
4. Make any of the callbacks return false and check they don't register.

### Changelog


assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
```diff
@@ -51,18 +51,16 @@ const usePaymentMethodRegistration = (
...
+	const cart = useStoreCart();
+	const { cartTotals, cartNeedsShipping, paymentRequirements } = cart;
 	const canPayArgument = useRef( {
 		cartTotals,
 		cartNeedsShipping,
 		billingData,
 		shippingAddress,
 		selectedShippingMethods,
 		paymentRequirements,
+		cart,
```
ralucaStan, line 63: I think this change should be a part of https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4670, so I'd exclude it from this try branch.

assets/js/blocks-registry/payment-methods/payment-method-config-helper.js
```diff
@@ -8,3 +8,39 @@ export const canMakePaymentWithFeaturesCheck = ( canMakePayment, features ) => (
...
+				if ( typeof callbacks[ paymentMethodName ] === 'function' ) {
+					namespacedCallbacks[ namespace ] =
+						callbacks[ paymentMethodName ];
+				}
+			}
+		);
+		canPay = Object.keys( namespacedCallbacks ).every( ( namespace ) => {
+			try {
+				return namespacedCallbacks[ namespace ]( canPayArgument );
+			} catch ( err ) {
```
ralucaStan, line 34: 👏 👏

ralucaStan (commented):
Hi Thomas, thank you for jumping on this so quickly. 🏅 

I tested it and it works as expected. I will move the relevant code to namespacing on my branch with some small changes.

assets/js/blocks-registry/payment-methods/payment-method-config-helper.js
```diff
@@ -8,3 +8,39 @@ export const canMakePaymentWithFeaturesCheck = ( canMakePayment, features ) => (
...
+	paymentMethodName
+) => ( canPayArgument ) => {
+	// validate whether the payment method is available
+	let canPay = canMakePayment( canPayArgument );
+
+	if ( canPay ) {
+		const namespacedCallbacks = {};
+		Object.entries( extensionsCallbacks ).forEach(
+			( [ namespace, callbacks ] ) => {
+				if ( typeof callbacks[ paymentMethodName ] === 'function' ) {
```
ralucaStan, line 25: I think this check should be part of the registration so it won't be necessary here

opr:
@ralucaStan can this be closed now?

assets/js/blocks-registry/payment-methods/payment-method-config-helper.js
```diff
@@ -8,3 +8,39 @@ export const canMakePaymentWithFeaturesCheck = ( canMakePayment, features ) => (
...
+		canPay = Object.keys( namespacedCallbacks ).every( ( namespace ) => {
+			try {
+				return namespacedCallbacks[ namespace ]( canPayArgument );
+			} catch ( err ) {
+				// eslint-disable-next-line no-console
+				console.error(
+					`Error when executing callback for ${ paymentMethodName } in ${ namespace }`,
+					err
+				);
+				throw err;
```
ralucaStan, line 40: We don't need to throw here as well. We need to continue with the calculation and ignore whatever result the callback would have had if it wouldn't have thrown.

PR #4680, opened 2021-09-03
Don't allow opening the Mini Cart in the Cart and Checkout pages

### LINKED ISSUE #4602: Mini Cart block: don't allow opening the drawer from the Cart & Checkout pages
Similarly to how the Mini Cart currently works in Storefront sites, we don't want the Mini Cart block to allow opening the drawer if the user is in the Cart or Checkout pages given that these pages already contain the contents from the Mini Cart.

Based on that, it would be great if we can also prevent loading any Mini Cart-related script file in those pages, given that they will not be needed.

Aljullu:
Fixes #4602.

1. In the admin, go to Appearance > Widgets and add the Mini Cart block to a widget area (ie: the sidebar).
2. In the frontend, go to the Cart or Checkout pages and verify you can't open the Mini Cart (button is disabled). Verify also `mini-cart-frontend.js` and `mini-cart-component-frontend.js` scripts are not loaded.
3. In the frontend, go to any other page which is not Cart or Checkout and verify the Mini Cart can be opened as usual.

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -49,6 +49,10 @@ class MiniCart extends AbstractBlock {
 	 * @return array|string
 	 */
 	protected function get_block_type_script( $key = null ) {
+		if ( is_cart() || is_checkout() ) {
+			return;
+		}
+
 		$script = [
 			'handle'       => 'wc-' . $this->block_name . '-block-frontend',
 			'path'         => $this->asset_api->get_block_asset_build_path( $this->block_name . '-frontend' ),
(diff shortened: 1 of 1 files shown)


senadir (approved):
Just making sure you're aware, this is limited to any page containing Cart or Checkout block, or actual cart and checkout page?

I know the second one is the most important and chances of the first one happening is low. Are you disabling it for cosmetic reasons (like storefront) or the existence of Cart and Mini Cart might cause one of them to malfunction? When testing both of them. The state is not synchronized. Actions in Cart block are reflected in mini cart, but not the other way. This is not really important but just worth noting.

Aljullu:
Thanks for the review @senadir!

> I know the second one is the most important and chances of the first one happening is low. Are you disabling it for cosmetic reasons (like storefront) or the existence of Cart and Mini Cart might cause one of them to malfunction? When testing both of them. The state is not synchronized. Actions in Cart block are reflected in mini cart, but not the other way. This is not really important but just worth noting.

Ah, good catch. Correct, the current code disables the Mini Cart on the pages defined as the Cart and Checkout, that means it doesn't impact other pages containing the Cart & Checkout blocks or shortcodes. I think it's a good enough approach that doesn't over-complicate the code for an (allegedly) edge case. I will merge it as-is and we can reevaluate if we get reports about this behavior. Does that sound good?

PR #4682, opened 2021-09-03
Add E2E and Unit (PHP and JS) tests for Filter by Stock block.

### LINKED ISSUE #4661: Add tests for the "Filter products by stock status" block
In #4145 a new "Filter Products by Stock" block was added. We need to add E2E tests to cover this block.

These tests should include:
- Adding the block
- Checking that the "Out of stock" option is hidden when the Woo core "Hide out of stock products from the catalog" is enabled
- Check that the block works with custom stock statuses (add these using this code):
```php
	// Add new statuses

	return $status;
}
```
- Check the block works when other filter type blocks are used on the same page.

opr:
This PR introduces E2E, JS unit, and PHP unit tests for the Filter by Stock block.

Besides the code changes to test files, there is one other important change to actual code: I have modified the way that the options for the block are derived and at what point. Originally these options were fetched outside of the component body. Because of this the code was executed as soon as the file was imported. This was not good for testing because we couldn't mock the values returned by `getSetting`. That is because the mocks were set up in the tests after importing the file. The changes made in `assets/js/blocks/stock-filter/block.js` move the options into the component ~~and add them to a `useMemo` with no deps, so this should only be calculated once.~~. 

~~I opted to do this rather than using `useState` because the implication of using this is that the values returned by `getSetting` (and therefore `STOCK_STATUS_OPTIONS`) could change over time, when they will not. I welcome any feedback on this approach though.~~

After discussing with @tjcafferkey we decided that these usages of `useMemo` were "over-optimisations". The length of `STOCK_STATUS_OPTIONS` will _usually_ be 3, so iterating over it is cheap, and not worth memoizing given the complexity this adds to the component.

Because the value of `initialOptions` is used as a dependency to the `useEffect` further down this file, setting these values as internal component state seemed the best way forward to stop this effect running every render.

Fixes #4661

1. Run `npm run test:php` `npm run test` `npm run test:e2e` and verify they all pass.
2. Create a page, add the `Filter by Stock`, `Filter by Price`, `Active filters`, and `All Products` blocks.
3. View the page and ensure the `Filter by Stock` block is working correctly. Verify the counts of each stock level are accurate.
4. Go to **WooCommerce > Settings > Products > Inventory** toggle `Hide out of stock products from catalogue" to be true.
5. View the page again and ensure `Out of stock` is not shown as a stock level option.
6. Wait for tests to go green in this PR.

### DIFF
--- a/assets/js/blocks/stock-filter/block.js
+++ b/assets/js/blocks/stock-filter/block.js
@@ -34,31 +34,24 @@ const StockStatusFilterBlock = ( {
 	attributes: blockAttributes,
 	isEditor = false,
 } ) => {
-	const hideOutOfStockItems = useMemo(
-		() => getSetting( 'hideOutOfStockItems', false ),
-		[]
+	const hideOutOfStockItems = getSetting( 'hideOutOfStockItems', false );
+	const { outofstock, ...otherStockStatusOptions } = getSetting(
+		'stockStatusOptions',
+		{}
 	);
-	const { outofstock, ...otherStockStatusOptions } = useMemo(
-		() => getSetting( 'stockStatusOptions', {} ),
-		[]
-	);
-	const STOCK_STATUS_OPTIONS = useMemo( () => {
-		return hideOutOfStockItems
-			? otherStockStatusOptions
-			: { outofstock, ...otherStockStatusOptions };
-	}, [ hideOutOfStockItems, otherStockStatusOptions, outofstock ] );
+	const STOCK_STATUS_OPTIONS = hideOutOfStockItems
+		? otherStockStatusOptions
+		: { outofstock, ...otherStockStatusOptions };
 
 	const [ checked, setChecked ] = useState( [] );
 	const [ displayedOptions, setDisplayedOptions ] = useState(
 		blockAttributes.isPreview ? previewOptions : []
 	);
 	// Filter added to handle if there are slugs without a corresponding name defined.
-	const initialOptions = useMemo( () => {
-		return Object.entries( STOCK_STATUS_OPTIONS )
-			.map( ( [ slug, name ] ) => ( { slug, name } ) )
-			.filter( ( status ) => !! status.name )
-			.sort( ( a, b ) => a.slug.localeCompare( b.slug ) );
-	}, [ STOCK_STATUS_OPTIONS ] );
+	const initialOptions = Object.entries( STOCK_STATUS_OPTIONS )
+		.map( ( [ slug, name ] ) => ( { slug, name } ) )
+		.filter( ( status ) => !! status.name )
+		.sort( ( a, b ) => a.slug.localeCompare( b.slug ) );
 
 	const [ queryState ] = useQueryStateByContext();
 	const [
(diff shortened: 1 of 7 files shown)


assets/js/blocks/stock-filter/block.js
```diff
@@ -48,10 +34,31 @@ const StockStatusFilterBlock = ( {
 	attributes: blockAttributes,
 	isEditor = false,
 } ) => {
+	const hideOutOfStockItems = useMemo(
```
tjcafferkey: Personally this `useMemo` is an over-optimisation to me as there's no real computation happening.

assets/js/blocks/stock-filter/block.js
```diff
@@ -48,10 +34,31 @@ const StockStatusFilterBlock = ( {
...
+		() => getSetting( 'hideOutOfStockItems', false ),
+		[]
+	);
+	const { outofstock, ...otherStockStatusOptions } = useMemo(
```
tjcafferkey: Personally this `useMemo` is an over-optimisation to me as there's no real computation happening.

assets/js/blocks/stock-filter/block.js
```diff
@@ -48,10 +34,31 @@ const StockStatusFilterBlock = ( {
...
+		() => getSetting( 'stockStatusOptions', {} ),
+		[]
+	);
+	const STOCK_STATUS_OPTIONS = useMemo( () => {
```
tjcafferkey: Again, seems like an over-optimisation personally for the same reasons as above.

assets/js/blocks/stock-filter/block.js
```diff
@@ -48,10 +34,31 @@ const StockStatusFilterBlock = ( {
...
+			? otherStockStatusOptions
+			: { outofstock, ...otherStockStatusOptions };
+	}, [ hideOutOfStockItems, otherStockStatusOptions, outofstock ] );
+
 	const [ checked, setChecked ] = useState( [] );
 	const [ displayedOptions, setDisplayedOptions ] = useState(
 		blockAttributes.isPreview ? previewOptions : []
 	);
+	// Filter added to handle if there are slugs without a corresponding name defined.
+	const initialOptions = useMemo( () => {
```
tjcafferkey: Unless there are a lot of values within `STOCK_STATUS_OPTIONS` it doesn't seem worth using `useMemo`. Just going to quote Kent C Dodds:

```
Performance optimizations are not free. They ALWAYS come with a cost but do NOT always come with a benefit to offset that cost.

Therefore, optimize responsibly.
```
- [Source](https://kentcdodds.com/blog/usememo-and-usecallback)

opr: Under normal circumstances this will have a length of 3, but merchants _can_ add more stock statuses. After thinking about it, the complexity this brings to the component coupled with the overheads of `useMemo` this don't seem like a worthwhile optimisation in my opinion.

ralucaStan:
While testing this PR I've come across a bug I've described in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4741

ralucaStan (approved):

PR #4685, opened 2021-09-03
Expose Validation to 3PD blocks.

### LINKED ISSUE #4502: Checkout i2 - Pass validation context to Inner Blocks to ...
Checkout i2 - Pass validation context to Inner Blocks to avoid exporting in a public package.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b037d76746d346d048d64ea295f497893d75a48a/assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-terms-block/frontend.tsx#L29-L40
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in b037d76746d346d048d64ea295f497893d75a48a when #4496 was merged. cc @mikejolley

senadir:
Introduced `useValidation` which returns `validation` that we pass to plugins. Upon being executed, it returns validation related functions that 3PD can use.

Fixes #4502

1. Add Terms block to Checkout i2
2. Turn it to Checkbox instead of text.
3. Visit Checkout i2, you can toggle the field, leaving it untoggled and submitting would prevent the order.
4. in `assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-terms-block/index.tsx`, inside `parent` add `woocommerce/checkout-billing-address-block`.
5. Open Checkout i2 block, remove Terms block from below block. Uncheck "using same address for billing", add Terms block to billing step, save.
6. Visit checkout, placing order with  checked should work fine.
7. Unchecking "using same address for billing" would prevent the block from checking out if the Terms block isn't checked.
8. Checking "using same address for billing" would remove that requirement.

### DIFF
--- a/assets/js/base/context/hooks/use-validation.ts
+++ b/assets/js/base/context/hooks/use-validation.ts
@@ -15,9 +15,7 @@ import { useValidationContext } from '../providers/validation/';
 /**
  * Custom hook for setting for adding errors to the validation system.
  */
-export const useValidation = (): ( (
-	validationErrorId: string
-) => ValidationData ) => {
+export const useValidation = (): ValidationData => {
 	const {
 		hasValidationErrors,
 		getValidationError,
@@ -25,30 +23,38 @@ export const useValidation = (): ( (
 		hideValidationError,
 		setValidationErrors,
 	} = useValidationContext();
+	const prefix = 'extensions-errors';
 
-	return useCallback(
-		( validationErrorId: string ) => {
-			validationErrorId = `extensions-errors-${ validationErrorId }`;
-			return {
-				hasValidationErrors,
-				getValidationError: () =>
-					getValidationError( validationErrorId ),
-				clearValidationError: () =>
-					clearValidationError( validationErrorId ),
-				hideValidationError: () =>
-					hideValidationError( validationErrorId ),
-				setValidationError: ( error: ValidationContextError ) =>
-					setValidationErrors( {
-						[ validationErrorId ]: error,
-					} ),
-			};
-		},
-		[
-			clearValidationError,
-			getValidationError,
-			hasValidationErrors,
-			hideValidationError,
-			setValidationErrors,
-		]
-	);
+	return {
+		hasValidationErrors,
+		getValidationError: useCallback(
+			( validationErrorId: string ) =>
+				getValidationError( `${ prefix }-${ validationErrorId }` ),
+			[ getValidationError ]
+		),
+		clearValidationError: useCallback(
+			( validationErrorId: string ) =>
+				clearValidationError( `${ prefix }-${ validationErrorId }` ),
+			[ clearValidationError ]
+		),
+		hideValidationError: useCallback(
+			( validationErrorId: string ) =>
+				hideValidationError( `${ prefix }-${ validationErrorId }` ),
+			[ hideValidationError ]
+		),
+		setValidationErrors: useCallback(
+			( errorsObject: Record< string, ValidationContextError > ) =>
+				setValidationErrors(
+					Object.fromEntries(
+						Object.entries(
+							errorsObject
+						).map( ( [ validationErrorId, error ] ) => [
+							`${ prefix }-${ validationErrorId }`,
+							error,
+						] )
+					)
+				),
+			[ setValidationErrors ]
+		),
+	};
 };
--- a/assets/js/base/context/hooks/use-validation.ts
+++ b/assets/js/base/context/hooks/use-validation.ts
@@ -1,0 +1,54 @@
+/**
+ * External dependencies
+ */
+import { useCallback } from '@wordpress/element';
+import type {
+	ValidationData,
+	ValidationContextError,
+} from '@woocommerce/type-defs/contexts';
+
+/**
+ * Internal dependencies
+ */
+import { useValidationContext } from '../providers/validation/';
+
+/**
+ * Custom hook for setting for adding errors to the validation system.
+ */
+export const useValidation = (): ( (
+	validationErrorId: string
+) => ValidationData ) => {
+	const {
+		hasValidationErrors,
+		getValidationError,
+		clearValidationError,
+		hideValidationError,
+		setValidationErrors,
+	} = useValidationContext();
+
+	return useCallback(
+		( validationErrorId: string ) => {
+			validationErrorId = `extensions-errors-${ validationErrorId }`;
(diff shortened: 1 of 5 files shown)


assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-terms-block/frontend.tsx
```diff
@@ -44,27 +45,22 @@ const FrontendBlock = ( {
...
+				message: __(
+					'Please read and accept the terms and conditions.',
+					'woo-gutenberg-products-block'
+				),
+				hidden: false,
 			} );
 		}
-		return () => {
-			clearValidationError( validationErrorId );
-		};
```
senadir: This was the part that caused the infinite loop. I'm not sure why having it would change anything. Deleting it didn't break stuff.

senadir (commented):
I can't seem to get an error to be visible on my checkout. It's blocked thou, is this expected @mikejolley?

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-terms-block/frontend.tsx
mikejolley: Looks like this was here to clear validation errors for a component if it were to be unmounted. A good test case would be for a validation error within the billing address section, and then hiding the billing address section?

assets/js/base/context/hooks/use-validation.ts
```diff
@@ -0,0 +1,54 @@
...
+export const useValidation = (): ( (
+	validationErrorId: string
+) => ValidationData ) => {
+	const {
+		hasValidationErrors,
+		getValidationError,
+		clearValidationError,
+		hideValidationError,
+		setValidationErrors,
+	} = useValidationContext();
```
mikejolley, line 25: I'm fine with this because it exposes what we have. Right now we put the onus on components to implement this, and then expose their validation errors within the context provider. That works fine.

I do however wonder if we should only pass down the current status and let components manage their own validation error + states, with an event or callback to block checkout and show a blocking error message. I don't know if this is something being looked at separately so shouldn't stop this being merged if we just want to roll with the existing system. Did you have any thoughts on this?

senadir: We're going to revisit the whole validation system at some point soon. We still need some way to tell checkout and blocks:
- There's an error.
- The error id and message.
- Various states.

I'm not sure what will come out of that refactor, but we still need a place to collect all existing errors.

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-terms-block/frontend.tsx
senadir: I will test it in that case

senadir: This test case checks out. Will try to understand why is this loop happening.

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-terms-block/frontend.tsx
```diff
@@ -19,23 +18,25 @@ const FrontendBlock = ( {
...
 }: {
 	text: string;
 	checkbox: boolean;
 	instanceId: string;
+	validation: ( validationErrorId: string ) => ValidationData;
 } ): JSX.Element => {
 	const [ checked, setChecked ] = useState( false );
-
 	// @todo Checkout i2 - Pass validation context to Inner Blocks to avoid exporting in a public package.
 	const { isDisabled } = useCheckoutSubmit();
```
nerrad: Is this todo incorrect here? I'm assuming you mean passing the `checkoutSubmit` context to Inner Blocks to avoid exporting?

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-terms-block/frontend.tsx
senadir: Fixed in 8dcc6ed367cb308fa8fd79ad87556d98676fb36e

assets/js/blocks/cart-checkout/checkout-i2/inner-blocks/checkout-terms-block/frontend.tsx
senadir: Yeah, it seems to be left over, fixed in 3e10485

nerrad: I see the change, however this block is still importing `useCheckoutSubmit` which is only available because the block is built as a part of the Woo Blocks repo. Is `useCheckoutSubmit` exposed to extension developers as well? If not, the `checkout-terms-block` would be a good reference for extension developers to learn how to create this type of block so we might want to consider exposing the values of `useCheckoutSubmit` to extensions?

This might already be considered as a followup, or elsewhere. If so feel free to ignore my comment :)

senadir: > Is useCheckoutSubmit exposed to extension developers as well?

No.

> the checkout-terms-block would be a good reference for extension developers to learn how to create this type of block

Yes, but also no. We have [Newsletter Block](https://github.com/opr/newsletter-test) that is optimized to do that. The block is very similar to this one, but is built to be external from the start.

> so we might want to consider exposing the values of useCheckoutSubmit to extensions?

We previously discussed this, `useCheckoutSubmit` was exposed to be used in Newsletter Block (very similar code given both are checkboxes) but I decided we withheld from that.

[My reasoning was](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4607#discussion_r698450400) that `useCheckoutSubmit` is an internal hook that use to render the submit button, it didn’t make sense to expose it for this reason, a better/dedicated hook for checkout states should be exposes if this proves to be needed.

I believe we would need this somehow in the future at some point or another, but we should consider how to expose it (via events, or a hook, or a subset of `useCheckoutContext`).

mikejolley (approved):
👍🏻  Nice job, works well and solves the problem. 🚢

PR #4689, opened 2021-09-07
Mini Cart block: add 'ver' query param to script URLs

Aljullu:
This is a small PR fixing a flaw in the `preloadScript` and the `lazyLoadScript` utils. They were generating script URLs different from the ones generated by WP.

Ie:
`/wp-content/plugins/gutenberg/build/element/index.min.js?51b038d0a9e09d863ec01732ed923941` instead of
`/wp-content/plugins/gutenberg/build/element/index.min.js?ver=51b038d0a9e09d863ec01732ed923941` (notice the missing `ver=`).

That would make the browser cache to consider both scripts different.

1. Create a page with the Mini Cart block. Go to the frontend and click on the button. In your browser devtools, go to the Network tab and search for a script with the word `element`, for example. Write down its location.
Ie: `/wp-content/plugins/gutenberg/build/element/index.min.js?ver=51b038d0a9e09d863ec01732ed923941`
2. Now, create a page with the Cart block. Go to the frontend and in your browser devtools, go to the Network tab and search for a script with the same word as the step before. Compare its location with the one from step 1 and verify they are exactly the same.

### DIFF
--- a/assets/js/base/utils/lazy-load-script.ts
+++ b/assets/js/base/utils/lazy-load-script.ts
@@ -123,7 +123,7 @@ const lazyLoadScript = ( {
 			id: `${ handle }-js`,
 			onerror: reject,
 			onload,
-			src: version ? `${ src }?${ version }` : src,
+			src: version ? `${ src }?ver=${ version }` : src,
 		} );
 	} );
 };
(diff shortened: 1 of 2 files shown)


opr (approved):
Looks good to me and is working as described :)

PR #4696, opened 2021-09-08
Add tests for filtering payment methods

### LINKED ISSUE #4622: Add tests for the new extensibility point for filtering the payment methods
We need to cover with unit tests the new integration for filtering payment options from #4621

### LINKED ISSUE #4661: Add tests for the "Filter products by stock status" block

opr:
This PR introduces tests for the methods introduced in #4668 it does not test other already-existing methods of the payment method config.

There are several different namespaces used when calling `registerPaymentMethodExtensionCallbacks` - this is because there is no way to unregister these filters, or to "reset" the `canMakePaymentExtensionsCallbacks` object between tests (since it's imported).

Fixes #4622

1. Check tests pass `npm run test`
2. Ensure actions go green on this PR


assets/js/blocks-registry/payment-methods/test/registry.js
```diff
@@ -0,0 +1,72 @@
...
+import { registerPaymentMethodExtensionCallbacks } from '@woocommerce/blocks-registry';
+
+/**
+ * Internal dependencies
+ */
+import { canMakePaymentExtensionsCallbacks } from '../extensions-config';
+
+describe( 'registerPaymentMethodExtensionCallbacks', () => {
+	it( 'Logs an error to console if namespace is already registered', () => {
+		registerPaymentMethodExtensionCallbacks(
```
ralucaStan: I think it would be good to also test that the namespace is not overwritten. I think this is the main effect we need to test, and the console is secondary. You do something similar for the next test.

assets/js/blocks-registry/payment-methods/test/payment-method-config-helper.js
```diff
@@ -0,0 +1,99 @@
...
+		);
+	} );
+
+	beforeEach( () => {
+		trueCallback.mockClear();
+		throwsCallback.mockClear();
+		falseCallback.mockClear();
+	} );
+
+	it( 'The function returned returns false if canMakePayment is false already and the callback is not executed', () => {
```
ralucaStan:
```suggestion
	it( 'Returns false without executing the registered callbacks, if the payment method's canMakePayment callback returns false.', () => {
```

assets/js/blocks-registry/payment-methods/test/payment-method-config-helper.js
```diff
@@ -0,0 +1,99 @@
...
+		const canMakePaymentWithExtensionsResult = canMakePaymentWithExtensions(
+			canMakePayment,
+			canMakePaymentExtensionsCallbacks,
+			'cod'
+		)();
+		expect( canMakePaymentWithExtensionsResult ).toBe( false );
+		expect( trueCallback ).not.toHaveBeenCalled();
+	} );
+
+	it( 'Does not unnecessarily execute additional callbacks from other namespaces if an earlier one returns false', () => {
```
ralucaStan:
```suggestion
	it( 'Returns early when a registered callback returns false, without executing all the registered callbacks', () => {
```

ralucaStan (approved):
Thank you so much for taking care of the tests, they look really good. I've approved, although I made some small suggestions to improve the clarity

assets/js/blocks-registry/payment-methods/test/payment-method-config-helper.js
```diff
@@ -0,0 +1,99 @@
...
+		);
+	} );
+
+	beforeEach( () => {
+		trueCallback.mockClear();
+		throwsCallback.mockClear();
+		falseCallback.mockClear();
+	} );
+
+	it( 'Returns false without executing the registered callbacks, if the payment method's canMakePayment callback returns false.', () => {
```
opr:
```suggestion
	it( "Returns false without executing the registered callbacks, if the payment method's canMakePayment callback returns false.", () => {
```

assets/js/blocks-registry/payment-methods/test/registry.ts
```diff
@@ -0,0 +1,97 @@
...
+				cod: () => false,
+			}
+		);
+
+		expect(
+			canMakePaymentExtensionsCallbacks[
+				'overwrite-marketplace-extension'
+			].cod
+		).toEqual( firstCodCallback );
+	} );
```
opr, line 57: @ralucaStan thanks for the review - this test checks that the namespace is not overwritten. I need to leave the `expect( console ).not.toHaveErrored();` line in because if not then the test case fails because of "unexpected" console errors 🤓

assets/js/blocks-registry/payment-methods/test/payment-method-config-helper.ts
```diff
@@ -0,0 +1,99 @@
...
+	it( 'Allows all valid callbacks to run, even if one causes an error', () => {
+		canMakePaymentWithExtensions(
+			() => true,
+			canMakePaymentExtensionsCallbacks,
+			'testpay'
+		)();
+		expect( console ).toHaveErrored();
+		expect( throwsCallback ).toHaveBeenCalledTimes( 1 );
+		expect( trueCallback ).toHaveBeenCalledTimes( 1 );
+	} );
```
senadir, line 98: Does this confirm that `testpay` `canPay` will return true? Meaning it captured the error thrown, ignored it, continued working and will return the true value?

senadir (approved):
Tests are looking really nice! Thank you for writing those Thomas.

assets/js/blocks-registry/payment-methods/test/payment-method-config-helper.ts
opr: Yes, the `testpay` callback registered in `woocommerce-marketplace-extension` throws an error, so this is ignored.

The `testpay` callback in `other-woocommerce-marketplace-extension` returns true so that is the expected outcome.

ralucaStan (approved):

PR #4699, opened 2021-09-09
Fix Checkout i2 mobile styling

### LINKED ISSUE #4576: Checkout i2: Consider mobile styling
The mobile view has lots of wasted space, and elements like form steps do not change much, for example, they still have the left border which makes the form fields narrow. This is the editor in mobile preview mode:

Is the mobile display something we want to think about @vivialice? What about the order of elements (express payment, order summary, fields), is that something we should think about?

At the very least, we should remove the large space below order summary, before the form fields. I'm not sure what is causing that.

mikejolley:
The goal of this PR is to improve the appearance of Checkout i2 on mobile, as well as adding some consistency with the Desktop views. This includes better alignment of fields in the column, and removal of the indentation when viewing on a mobile device or small screen.

Fixes #4576

### Screenshots

Here are some side by side views showing what has changed:

**Mobile appearance:**

| Before | After |
|  Screenshot 2021-09-09 at 14 41 11 |  Screenshot 2021-09-09 at 13 45 25 |
|  Screenshot 2021-09-09 at 14 41 18 |  Screenshot 2021-09-09 at 13 45 40 |

**Desktop appearance:**

| Before | After |
|  Screenshot 2021-09-09 at 14 41 42 |  Screenshot 2021-09-09 at 13 46 10 |
| Screenshot 2021-09-09 at 14 41 47  |  Screenshot 2021-09-09 at 13 46 18 |

**Editor appearance:**

| Before | After |
| Screenshot 2021-09-09 at 14 41 54  |  Screenshot 2021-09-09 at 14 10 38 |
|  Screenshot 2021-09-09 at 14 41 35 | Screenshot 2021-09-09 at 14 10 48  |

### Manual Testing

1. View checkout i2 block
2. Confirm elements are styled as shown above on both desktop, and mobile (using browser tools)
3. Repeat for i1. This will also reflect the changes above

### Changelog

> Updated checkout styling to reduce margins in the mobile layout

### DIFF
--- a/assets/js/base/components/cart-checkout/form-step/style.scss
+++ b/assets/js/base/components/cart-checkout/form-step/style.scss
@@ -5,7 +5,6 @@
 .wc-block-components-form .wc-block-components-checkout-step {
 	position: relative;
 	border: none;
-	// This makes the padding equal on both sides and aligned with elements like checkboxes.
 	padding: 0 0 0 $gap-large;
 	background: none;
 	margin: 0;
--- a/assets/js/base/components/cart-checkout/form-step/style.scss
+++ b/assets/js/base/components/cart-checkout/form-step/style.scss
@@ -5,22 +5,14 @@
 .wc-block-components-form .wc-block-components-checkout-step {
 	position: relative;
 	border: none;
-	padding: 0 0 0 $gap-larger;
+	// This makes the padding equal on both sides and aligned with elements like checkboxes.
+	padding: 0 0 0 $gap-large;
 	background: none;
 	margin: 0;
 
-	.is-large & {
-		padding-right: $gap-large;
-	}
-
-	.wc-block-components-checkout-step__heading::after {
-		content: "";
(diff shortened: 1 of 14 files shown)


senadir:
Tests seems to be failing after restarting them twice so I rebased with trunk and force pushed.

Aljullu (approved):
Looks much better, good job @mikejolley!

I will leave here a couple of things I noticed, but up to you if you want to fix them in this PR or not:

1. In the editor when on mobile, the section titles break in two lines:



2. The desktop skeleton is shown on mobile too:

https://user-images.githubusercontent.com/3616980/132859980-08a9f1b1-7fe8-4458-a98c-cefee0963a51.mp4

mikejolley:
Thanks @Aljullu Ill look at the skeleton when I do https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4681 because I know the current sidebar layout styles are duplicated across both blocks.

PR #4700, opened 2021-09-09
Handle `process_payment` errors the same way as validation

RadoslavGeorgiev:
`Api::process_legacy_payment` already has functionality to convert notices to exceptions, but it was
only called before the gateway's `process_payment`. That call should remain there, but `process_payment`
might add its own notices later (see PR https://github.com/Automattic/woocommerce-payments/pull/2872), which is why another conversion of notices to exceptions is required.

### Screenshots

image

#### Automated Tests

This functionality already exists and is now simply called again a couple of lines later. Additional tests should not be needed.

### Manual Testing

1. Checkout https://github.com/Automattic/woocommerce-payments/pull/2872 for WCPay
2. Try purchasing a product with less than $0.50 USD order total (incl. shipping)
3. See a generic message about something going wrong.
4. Checkout this PR
5. Try checkout again. A full message should be displayed like the screenshot above.

### Performance Impact

The performance impact here is minimal (iterating over a normally empty array).

### Changelog

No changelog entry required.

### DIFF
--- a/src/Payments/Api.php
+++ b/src/Payments/Api.php
@@ -161,6 +161,9 @@ class Api {
 		// Process Payment.
 		$gateway_result = $payment_method_object->process_payment( $context->order->get_id() );
 
+		// If errors were thrown while processing the payment, we need to abort as well.
+		NoticeHandler::convert_notices_to_exceptions( 'woocommerce_rest_payment_error' );
+
 		// Restore $_POST data.
 		$_POST = $post_data;
 


senadir:
Hi there Radoslav! @nerrad @mikejolley and I discussed this and decided not to proceed with this PR for now.

How essential is this to WCPay? we intentionally left out notice handling for payments because:
- Throwing an exception at that point would prevent us from setting `set_status` for the order to set the order as failed.
- We can't reliably know what caused the error, the payment can still be submitted and we end up with double payments.
- We can’t correctly attribute errors to the correct section, WC Core has not way to differentiate if this notice is for payment, shipping, fields, or something else.

Can WCPay throw the error regularly on its `process_payment` function? How much of a blocker is this PR and are there workaround for it?

For more context:

- [This is the origin of that skip.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2189/files?file-filters%5B%5D=.php#r408287266)
- The slack discussion we had p1631196218490100-slack-C8X6Q7XQU that resulted in the above, however, I prefer if we can keep all decisions and context here because this is a public issue that could impact other 3PDs

RadoslavGeorgiev:
Hey @senadir 👋 

> How essential is this to WCPay?

Example in the context of https://github.com/Automattic/woocommerce-payments/pull/2872 (which might need a completely different internal approach, but will display the notice/error in the same fashion):

Stripe has varying minimum amounts, and for USD it's 50 cents. If a purchase is attempted with a lesser amount, we need to clearly indicate why that happened. Ideally we would not show payment fields at all, but an error instead. Unfortunately we cannot know the actual payment amount before shipping details have been entered, and a shipping method has been selected, so we need to handle this issue within `process_payment`.

> Can WCPay throw the error regularly on its process_payment function? How much of a blocker is this PR and are there workaround for it?

Well, `process_payment` is meant for the legacy checkout, meaning that `process_payment` cannot simply throw an error, because it would not be handled anywhere.

---

I just tested my last statement and found out that even legacy checkout supports throwing exceptions, which will be handled as notices. This means that **as far as I'm concerned, we can modify everything we need within `process_payment` and this PR is not needed.**

If anyone else encounters this issue, _just_ throw an exception.

senadir:
I believe we run into the 30 cent issue in Stripe as well, we handled it in the `canMakePayment` function. See PR https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3424

PR #4702, opened 2021-09-09
Add Typescript to /blocks-registry/payment-methods

### LINKED ISSUE #4687: Type the new extensibility point for filtering payment methods
The changes added by [4621](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4621) need to be typed.

ralucaStan:
This PR converts to Typescript the files related to payment methods under `/blocks-registry`. 

Fixes #4687

This is my first TypeScript PR so I'm really opened to advices and improvements.


assets/js/types/type-defs/payments.ts
```diff
@@ -0,0 +1,101 @@
+/**
+ * External dependencies
+ */
+import type { ReactNode } from 'react';
+
+export type ObjectType = Record< string, unknown >;
+export type EmptyObjectType = Record< string, never >;
```
opr: I don't think these should be here, it seems they also exist in `assets/js/base/context/providers/cart-checkout/payment-methods/types.ts` could we move them to a more central place to avoid duplication?

assets/js/types/type-defs/payments.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export interface Supports {
+	showSavedCards?: boolean;
+	showSaveOption?: boolean;
+	features?: string[];
+	// Deprecated, in favour of showSavedCards and showSaveOption
+	savePaymentInfo?: boolean;
+}
+
+// we assign a value in the class for supports.features
```
opr, line 24: I'm wondering if there's a better name than `SupportsInConstructor` or if we even need this type...

Maybe we can:
- Remove this type `SupportsInConstructor`
- Make `Supports.features` mandatory.
- In the `PaymentMethodConfig` `interface` - change the `supports` member to look like this:
```supports: Omit< Supports, 'features' > & { features?: string };```

This is just a suggestion, I would love to hear what you think about this.

File: assets/js/types/type-defs/payments.ts [hunk 3]
```diff
@@ -0,0 +1,101 @@
...
+export interface CanMakePaymentArgument {
+	cartTotals: CartTotals;
+	cartNeedsShipping: boolean;
+	billingData: CartResponseBillingAddress;
+	shippingAddress: CartResponseShippingAddress;
+	selectedShippingMethods: Array< unknown >;
+	paymentRequirements: Array< string >;
+}
+
+export type CanMakePayment = ( cartData: CanMakePaymentArgument ) => boolean;
```
opr: Could we rename this to `CanMakePaymentFunction`?

assets/js/blocks-registry/payment-methods/payment-method-config.ts
```diff
@@ -2,7 +2,24 @@
...
+	ObjectType,
+	SupportsInConstructor,
+	CanMakePayment,
+	CanMakePaymentArgument,
+	PaymentMethodConfigClass,
+} from '@woocommerce/type-defs/payments';
 
+/**
+ * Internal dependencies
+ */
```
opr: Looks like there are two
```js
/**
 * Internal dependencies
 */
```
blocks now

assets/js/types/type-defs/payments.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export interface PaymentMethodConfig {
```
opr: Possibly rename this to `PaymentMethodConfigurationSettings` or something to make it clearer that this is not the same thing as the class?

docs/blocks/feature-flags-and-experimental-interfaces.md
```diff
@@ -55,12 +55,12 @@ We also have individual features or code blocks behind a feature flag, this is a
...
--   `__experimentalDeRegisterExpressPaymentMethod` function used to deregister an express payment method, only used in tests ([experimental function](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b07883b8b76feeb439d655b255507b24fc59e091/assets/js/blocks-registry/payment-methods/registry.js#L74)).
+-   `__experimentalDeRegisterPaymentMethod` function used to deregister a payment method, only used in tests ([experimental function](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b07883b8b76feeb439d655b255507b24fc59e091/assets/js/blocks-registry/payment-methods/registry.ts#L114)).
+-   `__experimentalDeRegisterExpressPaymentMethod` function used to deregister an express payment method, only used in tests ([experimental function](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b07883b8b76feeb439d655b255507b24fc59e091/assets/js/blocks-registry/payment-methods/registry.ts#L120)).
```
opr, line 63: Cool, good spot! 😁

assets/js/blocks-registry/payment-methods/payment-method-config.ts
```diff
@@ -46,15 +70,19 @@ export default class PaymentMethodConfig {
 	}
 
 	// canMakePayment is calculated each time based on data that modifies outside of the class (eg: cart data).
-	get canMakePayment() {
+	get canMakePayment(): (
+		canPayArgument: CanMakePaymentArgument
+	) => boolean {
```
opr: We have a type that describes this function :)
```suggestion
	get canMakePayment(): CanMakePayment {
```

assets/js/blocks-registry/payment-methods/extensions-config.ts
```diff
@@ -0,0 +1,21 @@
...
+export type ExtensionNamespace = keyof CanMakePaymentExtensionsCallbacks;
+export type PaymentMethodName = keyof CanMakePaymentExtensionCallbacks;
+
+export type CanMakePaymentExtensionsCallbacks = Record<
+	PaymentMethodName,
+	CanMakePaymentExtensionCallbacks
+>;
+
+// Keeps callbacks registered by extensions for different payment methods
+/* eslint prefer-const: 0 */
```
opr: Is this the right way to disable it?

assets/js/blocks-registry/payment-methods/extensions-config.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+import { CanMakePayment } from '@woocommerce/type-defs/payments';
+
+type CanMakePaymentExtensionCallbacks = Record< string, CanMakePayment >;
+export type ExtensionNamespace = keyof CanMakePaymentExtensionsCallbacks;
+export type PaymentMethodName = keyof CanMakePaymentExtensionCallbacks;
+
+export type CanMakePaymentExtensionsCallbacks = Record<
+	PaymentMethodName,
+	CanMakePaymentExtensionCallbacks
```
opr: The naming here is quite confusing. I'd suggest renaming `CanMakePaymentExtensionsCallbacks` to `NamespacedCanMakePaymentExtensionsCallbacks`

I also think the key is wrong, shouldn't it be?
```suggestion
export type CanMakePaymentExtensionsCallbacks = Record<
	ExtensionNamespace,
	CanMakePaymentExtensionCallbacks
```

opr (changes requested):
👏🏼  This looks great, thanks for doing this Raluca! There are some cases where I think we can name things better so I've suggested them.

ralucaStan:
@opr Thank you for your review, I've also fixed the failing tests here if you'd like to take a look at https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4702/commits/0a6f8250d4f03a64fe064afa9fc2091c1a61f861.

assets/js/blocks-registry/payment-methods/extensions-config.ts
ralucaStan: As far as I know yes. Do you have anything else in mind?

assets/js/types/type-defs/payments.ts
ralucaStan: I don't think `Supports.features` should be made mandatory because this type will also be available for extensions and then there will be a confusion that features needs to be provided in order to register a payment method.

I could have used `	public supports: Supports & { features: string[] };` in the class, but because this type is used in multiple places I still think it's good to keep the `SupportsInConstructor` type in order to avoid repetition. I am opened to renaming it if you want.

assets/js/blocks-registry/payment-methods/extensions-config.ts
ralucaStan: Great catch, thanks for spotting this, fixed in 930c6886baecbbb8ae388983197fcc9509053b9e

ralucaStan:
@opr thank you so much for your review, you made some great points. 

I've actually added some missing TS definitions so I suggest reviewing the PR as a whole again if that's not too much trouble.
The changes since your last review include
- Add the complete CanMakePaymentReturnType and account for null for PaymentMethod.icons 0bcf0edd4818e37a77a90af59a1df3efb5f58b1b
- Fix failing tests because of savedTokenComponent 6be865ffd6655bdf3e099a68d5b72951789e71a9
- PR feedback e9fd11d53235cc6414daac1dcc84c67909bfb9c1 & 930c6886baecbbb8ae388983197fcc9509053b9e

assets/js/types/type-defs/payments.ts
opr: Hm, ok if `supports.features` is optional then that's fine, but I still think we should change `SupportsInConstructor`'s name. How about renaming like so:

`Supports` -> `SupportsConfiguration`
`SupportsInConstructor` -> `Supports` ?

assets/js/blocks-registry/payment-methods/extensions-config.ts
opr: I thought this would disable that rule for the rest of the file, we don't have anything else affected by it, but we _could_ in future, so let's just disable it for the next line specifically.

`// eslint-disable-next-line prefer-const`

ralucaStan: You are right, I did not notice I wasn't just disabling the next line ✋

assets/js/types/type-defs/payments.ts
```diff
@@ -0,0 +1,114 @@
...
+	cartNeedsShipping: boolean;
+	billingData: CartResponseBillingAddress;
+	shippingAddress: CartResponseShippingAddress;
+	selectedShippingMethods: Array< unknown >;
+	paymentRequirements: Array< string >;
+}
+
+export type CanMakePaymentReturnType =
+	| boolean
+	| Promise< boolean | { error: { message: string } } >;
```
opr, line 40: When is this ever a promise? I don't see anywhere in our code where we handle it, but maybe I'm missing something?

opr: Found it :)
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e9fd11d53235cc6414daac1dcc84c67909bfb9c1/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts#L113

opr:
Thanks for making the changes @ralucaStan great work on this!

When https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4702#discussion_r709309732 has been addressed I will be happy to give this a ✅

assets/js/types/type-defs/payments.ts
ralucaStan: Yes, exactly, and it was also mentioned here https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/providers/cart-checkout/payment-methods/types.ts#L23-L25

assets/js/types/type-defs/payments.ts
ralucaStan: This suggestion for `SupportsConfiguration`  is definitely an improvement. I've also renamed `(Express)PaymentMethodOptions` to `(Express)PaymentMethodConfiguration` for more clarity

opr (approved):
Woo, thanks Raluca! Looking good now. Let's 🚢 it!

PR #4716, opened 2021-09-10
Add e2e tests for Active Product Filters block

nielslange:
### Note

Currently, not every block is covered by e2e tests. This PR aims to add e2e tests for the Active Product Filters block.

Part of #4643 

### Testing

1. Check out this PR.
2. Run `npm run wp-env start && npm run test:e2e`.
3. The e2e test suite shows no failed tests.

### DIFF
--- a/tests/e2e/specs/backend/active-filters.test.js
+++ b/tests/e2e/specs/backend/active-filters.test.js
@@ -1,0 +1,47 @@
+/**
+ * External dependencies
+ */
+import {
+	getAllBlocks,
+	switchUserToAdmin,
+	getEditedPostContent,
+} from '@wordpress/e2e-test-utils';
+import { clearAndFillInput } from '@woocommerce/e2e-utils';
+import { insertBlockDontWaitForInsertClose } from '../../utils.js';
+import { visitBlockPage } from '@woocommerce/blocks-test-utils';
+
+const block = {
+	name: 'Active Product Filters',
+	slug: 'woocommerce/active-filters',
+	class: '.wc-block-active-filters',
+};
+
+describe( `${ block.name } Block`, () => {
+	beforeAll( async () => {
+		await switchUserToAdmin();
+		await visitBlockPage( `${ block.name } Block` );
+	} );
+
+	it( 'can be inserted once', async () => {
+		await insertBlockDontWaitForInsertClose( block.name );
+		expect( await getAllBlocks() ).toHaveLength( 1 );
+	} );
+
+	it( 'renders without crashing', async () => {
+		await expect( page ).toRenderBlock( block );
+	} );
+
+	it( 'can change title label in editor', async () => {
+		await expect( page ).toFill(
+			'.wc-block-active-filters__title > textarea',
+			'Current filters'
+		);
+
+		await clearAndFillInput(
+			'.wc-block-active-filters__title > textarea',
+			'Active filters'
+		);
+
+		expect( await getEditedPostContent() ).toMatchSnapshot();
+	} );
+} );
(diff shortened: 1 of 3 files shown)


tests/e2e/specs/backend/active-filters.test.js
```diff
@@ -0,0 +1,47 @@
...
+			'.wc-block-active-filters__title > textarea',
+			'Current filters'
+		);
+
+		await clearAndFillInput(
+			'.wc-block-active-filters__title > textarea',
+			'Active filters'
+		);
+
+		expect( await getEditedPostContent() ).toMatchSnapshot();
```
ralucaStan, line 45: We don't usually use snapshot testing in our E2E tests. If you look in the snapshot file the new title does not appear there. 
I recommend you remove this line and search for the title again and write an assertion that the block title now has the title you provided `expect( ... ).toHaveTextContent( 'Current filters' );`.

tests/e2e/specs/backend/active-filters.test.js
```diff
@@ -0,0 +1,47 @@
...
+		expect( await getAllBlocks() ).toHaveLength( 1 );
+	} );
+
+	it( 'renders without crashing', async () => {
+		await expect( page ).toRenderBlock( block );
+	} );
+
+	it( 'can change title label in editor', async () => {
+		await expect( page ).toFill(
+			'.wc-block-active-filters__title > textarea',
```
ralucaStan, line 36: Your tests fail on the CI because this selector [is not found](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4716/checks?check_run_id=3568636698#step:11:66). I've looked in the DOM and there was no `.wc-block-active-filters__title` element.  I think you should try with another selector.

ralucaStan (commented):
👏 for adding more E2E tests. I've left a couple of comments. I think it would also benefit the blocks if you would also add some tests for the attributes:
- it allows changing the Display style
- it allows to change the heading level. 

The checkout block also has some tests for attributes you could use as inspiration. Let me know if you get stuck, it's often the case with E2E :)

ralucaStan:
@nielslange actually @opr already [covered this block with tests](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4562), I guess they were not marked in the [initial list](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4643).

nielslange:
> @nielslange actually @opr already [covered this block with tests](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4562), I guess they were not marked in the [initial list](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4643).

Oh, I missed that draft. Thanks for pointing it out. I wanted to get back to you today regarding your review above. I'll close this PR for now and will create e2e tests for the other components that still need tests, in case some components are not covered yet.

tests/e2e/specs/backend/active-filters.test.js
nielslange: To close the loop, the CI tests failed, as I created a PR to add the `.wc-block-active-filters__title` class. That PR had been merged accidentally last week and then reverted, but remained closed. I'll have to create a fresh PR for adding that CSS class.

tests/e2e/specs/backend/active-filters.test.js
nielslange: Thanks for pointing this out. I'll keep this in mind for future e2e tests. For now, I have closed this PR due to @opr drafted PR in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4562.

PR #4718, opened 2021-09-10
Create Cart i2 block

### LINKED ISSUE #4717: Create Cart i2 block from Cart block
We need to copy Cart block into Cart i2 block like we did with Checkout i2. It should work exactly like Cart i1 (mono component) but should work alongside it.

senadir:
This PR creates Cart i2 block from Cart i1. I follow the same steps as Checkout i1 by copying, duplicating classes, and updating stuff to `cart-i2`.

Fixes #4717

### Testing
- run `composer install` to add the new class.
- Visit editor, try inserting Cart i2 block.
- Save page.
- Visit in on frontend, it should work.


assets/js/blocks/cart-checkout/cart-i2/frontend.js
```diff
@@ -0,0 +1,74 @@
...
+		button: (
+			<button className="wc-block-button" onClick={ reloadPage }>
+				{ __( 'Reload the page', 'woo-gutenberg-products-block' ) }
+			</button>
+		),
+	};
+};
+
+renderFrontend( {
+	selector: '.wp-block-woocommerce-cart-i2',
```
senadir, line 70: Only change in this file.

assets/js/blocks/cart-checkout/cart-i2/index.js
```diff
@@ -0,0 +1,51 @@
...
+	save( { attributes } ) {
+		return (
+			<div className={ classnames( 'is-loading', attributes.className ) }>
+				<InnerBlocks.Content />
+			</div>
+		);
+	},
+};
+
+registerFeaturePluginBlockType( 'woocommerce/cart-i2', settings );
```
senadir, line 51: Changed name here.

src/BlockTypes/CartI2.php
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * @internal
+ */
+class CartI2 extends AbstractBlock {
+	/**
+	 * Block name.
+	 *
+	 * @var string
+	 */
+	protected $block_name = 'cart-i2';
```
senadir, line 19: Only change.

senadir (commented):
Most files are copied. I pointed out what I changed between this and Cart i1.

Aljullu (approved):
LGTM.

On `trunk`, there have been [some changes](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4673) recently in `cart-line-items-table.tsx` and `cart-line-item-row.tsx` which are not updated in this PR. That's because I'm using those components in the Mini Cart block as well. @senadir what can we do to keep them in sync so the same component in Cart/Mini Cart and Cart i2 don't diverge? Could we move them to `base/components` so any change you make to those components, you ensure they work in both versions of the Cart block? Or that would be a lot of extra work?

mikejolley:
I'll get this rebased @senadir

mikejolley:
@Aljullu Probably easier to get our new blocks/framework in place first and then consolidate them when i2 replaces i1, as we did with checkout.

PR #4721, opened 2021-09-10
Checkout i2: Add focus styles to form elements

### LINKED ISSUE #4714: Checkout i2: Terms and condition checkbox validation relies only on color
## To reproduce

1. Create a page and add the Checkout i2 block.
2. Select the Terms and Conditions inner block.
3. In the _Display options_, check _Require checkbox_.
4. Visit that page in the frontend and try to submit the form.
5. Notice there is no error message, only the checkbox and the text turned red.

## Expected behavior
For accessibility reasons, I think we shouldn't rely solely on the color of elements and instead add a message or an icon that shows where the error is.

mikejolley:
Adds consistent focus styles to form elements on the checkout and styles error messages with a more visible outline. See screenshots below.

Fixes #4714

#### Accessibility

Added `aria-invalid` to invalid form elements.

### Screenshots

Screenshot 2021-09-10 at 16 42 15
Screenshot 2021-09-10 at 16 42 10
Screenshot 2021-09-10 at 16 41 58

### Testing

1. Leave checkout fields blank
2. Place order and view error styling
3. Click within fields to check focus styles match above screenshots

### DIFF
--- a/assets/js/base/components/checkbox-control/style.scss
+++ b/assets/js/base/components/checkbox-control/style.scss
@@ -25,6 +25,11 @@
 			border-color: $input-border-gray;
 		}
 
+		&:focus {
+			outline: 2px solid $input-border-gray;
+			outline-offset: 2px;
+		}
+
 		&::before,
 		&::after {
 			content: "";
@@ -42,11 +47,18 @@
 				background: $input-background-dark;
 				border-color: $controls-border-dark;
 			}
+
+			&:focus {
+				outline: 2px solid $controls-border-dark;
+				outline-offset: 2px;
+			}
 		}
 	}
 
 	&.has-error {
 		color: $alert-red;
+		outline: 2px solid $alert-red;
+		outline-offset: 0.5em;
 
 		a {
 			color: $alert-red;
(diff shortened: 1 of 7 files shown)


assets/js/base/components/checkbox-control/style.scss
```diff
@@ -42,11 +47,18 @@
...
+				outline: 2px solid $controls-border-dark;
+				outline-offset: 2px;
+			}
 		}
 	}
 
 	&.has-error {
 		color: $alert-red;
+		outline: 2px solid $alert-red;
+		outline-offset: 0.5em;
```
frontdevde, line 61: Just out of curiosity, why is this value using `em` while all the other outline values are using `px`?

frontdevde (approved):
Tested as per testing instructions and everything worked as expected. LGTM 👍

|Error before|Error after|
|-|-|
|Screenshot 2021-09-14 at 12 04 06|Screenshot 2021-09-14 at 12 07 03|

|Focus before|Focus after|
|-|-|
|Screenshot 2021-09-14 at 12 06 45|Screenshot 2021-09-14 at 12 07 33|

assets/js/base/components/checkbox-control/style.scss
mikejolley: I don't think we have a hard and fast rule. I saw several components use .5em for margins and spacing.

Issue #4688, opened 2021-09-07
Mini Cart block: open drawer when a product is added from the All Products block


Aljullu:
With the changes implemented in #4666, the Mini Cart drawer is opened every time a product is added to cart from the Shop page. However, that doesn't work from the All Products block because it doesn't trigger the `added_to_cart` event. This issue is about making the necessary changes to the All Products or Mini Cart blocks so when a product is added to cart from the All Products block, the Mini Cart drawer slides in.

PR #4730, opened 2021-09-13
Avoid exposing receiveCart to extensions

Aljullu:
While working on #4688, I found a couple of instances where we were passing `receiveCart` to slot fills so it was available by extensions. This PR fixes that.

Question: I wonder if we should create a hook that acts as a wrapper of `useStoreCart()` and automatically excludes `receiveCart`, so it can be used safely when making data available to extensions.

### Testing

1. Smoke test Cart & Checkout blocks with an extension like WC Subscriptions and verify there are no JS errors in the console and everything works as expected.

### Changelog

> Removed `receiveCart` method that was exposed in a couple of SlotFills by mistake.

### DIFF
--- a/assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
+++ b/assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
@@ -150,7 +150,7 @@ const ShippingRatesControl = ( {
 	// Prepare props to pass to the ExperimentalOrderShippingPackages slot fill.
 	// We need to pluck out receiveCart.
 	// eslint-disable-next-line no-unused-vars
-	const { extensions, ...cart } = useStoreCart();
+	const { extensions, receiveCart, ...cart } = useStoreCart();
 	const slotFillProps = {
 		className,
 		collapsible,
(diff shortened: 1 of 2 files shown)


assets/js/base/components/cart-checkout/shipping-rates-control/index.tsx
```diff
@@ -150,7 +150,7 @@ const ShippingRatesControl = ( {
 	// Prepare props to pass to the ExperimentalOrderShippingPackages slot fill.
 	// We need to pluck out receiveCart.
 	// eslint-disable-next-line no-unused-vars
-	const { extensions, ...cart } = useStoreCart();
```
senadir, line 153: This was already removed (according to the comment) so it seems like [it slipped in by accident on a recent refactor](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4135). Great catch!

senadir (approved):

senadir:
> I wonder if we should create a hook that acts as a wrapper of useStoreCart() and automatically excludes receiveCart, so it can be used safely when making data available to extensions.

I did think about that, we have a pattern of doing that recently, e.g. `useValidation` and `useCheckoutExtensionData`. However, currently, we don’t have anything extra to add to what's returned to `useCart` so I think we can hold on to this idea until this interface evolves. Given that creating and removing this hook wouldn’t break anything in the future, it's a safe option to add it both now or later, but I would avoid adding complexity to the codebase.

Aljullu:
Makes sense. Thanks for the review @senadir!

PR #4734, opened 2021-09-13
Remove absolute positioning on validation error messages

### LINKED ISSUE #4583: Revisit inline validation message styling
This came up when fixing https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4582

Right now the validation messages are positioned absolutely. 

Screenshot 2021-08-16 at 13 14 25

This prevents a jump in content when validation errors are shown, but comes at the cost of only supporting errors up to 2 lines. Any more than this and it overlaps other content.

Since we cannot be sure there will only ever be 2 lines of error (this can be impacted by font size, display width, mobile, error text) it seems like a fragile approach we should revisit at some point.

### LINKED ISSUE #4715: Checkout i2: coupon code error overlaps bottom border
## To reproduce

1. Create a page with the Checkout i2 block.
2. On the frontend, try to add a coupon that doesn't exist.
3. Notice the error that appears collides with the bottom border of the section.

https://user-images.githubusercontent.com/3616980/132862970-3565458d-137a-4851-9689-f21850dc746e.mp4

mikejolley:
This PR removes the absolute positioning on error messages and fixes #4715. This reverts a change in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3968 which did this to prevent the button press issue. After making this change I did the testing steps in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3968 and could no longer replicate the issue on Chrome or Edge, so this may have been a browser issue that has since been resolved?
 
Fixes #4583
Fixes #4715

### Testing

See steps in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3968 for button issue (test for regression)
See steps in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4715 for styling issue

@ralucaStan can you test this your side and confirm it no longer occurs?

### Changelog

> Fix validation message styling so they never overlap other elements.

### DIFF
--- a/assets/js/base/context/providers/validation/components/validation-input-error/style.scss
+++ b/assets/js/base/context/providers/validation/components/validation-input-error/style.scss
@@ -2,15 +2,10 @@
 	@include font-size(smaller);
 	color: $alert-red;
 	max-width: 100%;
-	position: absolute;
 	white-space: normal;
 
 	> p {
-		align-items: center;
-		display: flex;
-		line-height: 12px; // This allows two lines in a 24px bottom margin.
 		margin: 0;
-		min-height: 24px;
 		padding: 0;
 	}
 }


ralucaStan (approved):
Both blocks tested good.

PR #4737, opened 2021-09-14
Move Mini Cart block component into its own file

Aljullu:
This PR moves `MiniCartBlock` into a separate file, so we have the component in one file and the logic to render that block in the frontend in another one (similar to what we do with Cart and Checkout blocks).

This PR doesn't include any code changes, only moves some code from one file to the other.

### Manual Testing

1. Add a Mini Cart block to a page.
2. Visit that page in the frontend and verify the Mini Cart button renders correctly.
3. Click on it and verify the Mini Cart drawer slides in.


ralucaStan (approved):
Looks good

PR #4740, opened 2021-09-14
Fix Product Search block displaying incorrectly.

### LINKED ISSUE #4738: Visual regression in Product Search block
There is a visual regression on the Product Search block in recent releases. It seems to affect WC Blocks 5.7 and after and WC core 5.7.

I _guess_ it was introduced in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4551.

## To reproduce

1. With WooCommerce core 5.6 and WC Blocks deactivated, add a Product Search block and see how it looks in the frontend:

Before

2. Activate WC Blocks and see how the block looks in the frontend.

After

3. Notice styles changes (the input field is no longer full-width and there is no margin between the input and the button).

opr:
This PR renames the instances of `className` to `class` in the `ProductSearch.php` file. The incorrect attribute was causing the classes not to be applied resulting in incorrect styling.

Fixes #4738 

### Screenshots

| Before | After |
|---|---|
|  | image |

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
  
Not covered by tests as this is a style change, visual inspection is required.
### Manual Testing

1. Add the Product Search block to a page.
2. Ensure it renders at full width with a gap between the input and the button.

### Performance Impact
Nil.

### DIFF
--- a/src/BlockTypes/ProductSearch.php
+++ b/src/BlockTypes/ProductSearch.php
@@ -97,13 +97,13 @@ class ProductSearch extends AbstractBlock {
 		);
 
 		$input_markup  = sprintf(
-			'<input type="search" id="%s" className="wc-block-product-search__field" placeholder="%s" name="s" />',
+			'<input type="search" id="%s" class="wc-block-product-search__field" placeholder="%s" name="s" />',
 			esc_attr( $input_id ),
 			esc_attr( $attributes['placeholder'] )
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Works and looks as expected. Thanks for taking care of this @opr!

PR #4742, opened 2021-09-14
Add types to legacy-events.js

Aljullu:
Refactor `legacy-events.js` so it uses TypeScript (part of working on #4688).

### Manual Testing

1. One way to test that `legacy-events.js` works as expected is to go to a page with the Cart block and remove all items.
2. Once you see the empty cart template, add an item to the Cart.
3. Verify contents update and the full cart template appears.

https://user-images.githubusercontent.com/3616980/133256791-a95802f0-2980-464d-8c1b-01be8ccd1d72.mp4


opr (approved):
Looks good to me, very clean and simple. Tested in browser too and it's working as intended. 🚢 it!

PR #4744, opened 2021-09-14
Add a class name to the filter by product attribute block heading

### LINKED ISSUE #3612: Filter Products by Attribute: Add a CSS class to the heading element
Currently, the Filter Products by Attribute block comes with a title ("Filter by size") which is shown above the product attribute section. Also, this block allows to change a heading level in the Editor. It would be great it the block had a special CSS class for the heading element in order to avoid using complex selectors. For example, we can add the following class to the heading element in this block: `wc-block-attribute-filter-title`.

For example, at the moment, we can use the following selectors to change styles of this section:

### Front-end

`.wp-block-woocommerce-attribute-filter > *:not(.wc-block-attribute-filter)`

### Editor

`.editor-styles-wrapper .wp-block[data-type="woocommerce/attribute-filter"] > div > :first-child`

By having a special class for the heading element, we can simplify the process of styling this element. Instead of using two different versions of CSS rule-sets, we can use just one for both, front end and the editor: `.wc-block-attribute-filter-title { ... }`

The same thing we can apply to the _Filter Products by Price_ and the _Active Product Filters_ blocks.

tjcafferkey:
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3612

#### What
* This PR adds a class to the Filter Products by Attribute block heading.
* `className` prop in `<BlockTitle  />` component gets added to the `<TagName />` element

#### Why
This avoids the need to use a more complex selector to target this heading element. See https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3612 for more context on this.

#### Accessibility

_Removed non-relevant checklist items_

### Screenshots

N/A

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests

### Manual Testing

1. Add "Filter Products by Attribute" block in the Editor and select an attribute to filter by
2. Observe in the Editor the class added to the block heading should be `wc-block-attribute-filter__heading`
3. Save the page, and observe the changes on the frontend. The heading of the block should have a class name `wc-block-attribute-filter__heading`
4. No visual regressions should be introduced

### Changelog

> Filter Products by Attribute: Class added to the heading element of the block.

### DIFF
--- a/assets/js/editor-components/block-title/index.js
+++ b/assets/js/editor-components/block-title/index.js
@@ -21,7 +21,7 @@ const BlockTitle = ( {
 } ) => {
 	const TagName = `h${ headingLevel }`;
 	return (
-		<TagName>
+		<TagName className={ className }>
 			<label
 				className="screen-reader-text"
 				htmlFor={ `block-title-${ instanceId }` }
--- a/assets/js/blocks/attribute-filter/block.js
+++ b/assets/js/blocks/attribute-filter/block.js
@@ -327,7 +327,9 @@ const AttributeFilterBlock = ( {
 	return (
 		<>
 			{ ! isEditor && blockAttributes.heading && (
-				<TagName>{ blockAttributes.heading }</TagName>
+				<TagName className="wc-block-attribute-filter__heading">
+					{ blockAttributes.heading }
+				</TagName>
 			) }
 			<div className="wc-block-attribute-filter">
 				{ blockAttributes.displayStyle === 'dropdown' ? (
--- a/assets/js/blocks/attribute-filter/edit.js
+++ b/assets/js/blocks/attribute-filter/edit.js
@@ -386,6 +386,7 @@ const Edit = ( { attributes, setAttributes, debouncedSpeak } ) => {
 			) : (
 				<div className={ className }>
 					<BlockTitle
+						className="wc-block-attribute-filter__heading"
 						headingLevel={ headingLevel }
 						heading={ heading }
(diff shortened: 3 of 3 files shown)


assets/js/editor-components/block-title/index.js
```diff
@@ -21,7 +21,7 @@ const BlockTitle = ( {
 } ) => {
 	const TagName = `h${ headingLevel }`;
 	return (
-		<TagName>
+		<TagName className={ className }>
```
opr, line 24: What happens if `className` is `undefined`?

tjcafferkey: if `className` is `undefined` then the `class=` attribute wont get added to the element.

ralucaStan (commented):
👏 This tests out as expected. 
I small note, the initial ticket also mentions adding classes for the Filter Products by Price and the Active Product Filters blocks. Do you think we should tackle this in this PR as well?

tjcafferkey:
>the initial ticket also mentions adding classes for the Filter Products by Price and the Active Product Filters blocks. Do you think we should tackle this in this PR as well?

Thanks for the review @ralucaStan. I think considering its a related change and relatively small, the risk of also including it in this PR is small? I'll update this PR with those changes too 😀

tjcafferkey:
Closing this PR and the original issue as seen it's been completed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4749/files

PR #4759, opened 2021-09-16
Use AbortController to cancel the network request on component unmount

### LINKED ISSUE #2100: Cancel inflight api requests if CheckoutProcessor component is unmounted.
In the new `CheckoutProcessor` component, we have a an effect that executes an api fetch request and then dispatches state updates when the promise is resolved or rejected. A potential problem with this is that if the component is unmounted before the promise is resolved, the state setters will fail, and React will log a warning to the console.

The solution is to implement a cancellable apiFetch promise so that we can return the cleanup action on the `useEffect` and cancel inflight requests.

Jetpack implemented a [fix here](https://github.com/Automattic/jetpack/pull/15228) that we can use as a guide.

tjcafferkey:
### Description

* Use AbortController to cancel the `processOrder` network request if the component unmounts. This approach is recommended in the `@wordpress/api-fetch` docs found [here](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-api-fetch/#aborting-a-request)

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2100

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Checkout this branch and copy the code from [this Gist](https://gist.github.com/tjcafferkey/08719dbfa5a3dac7fb033790051a7a41) into file `assets/js/base/context/providers/cart-checkout/checkout-provider.js` overwriting its contents. This will create a button that will force the `<CheckoutProcessor />` to unmount when clicked.
2. On your Woo site add a product to your basket and go to the checkout following the normal checkout process just stopping short of placing your order.
3. Open your DevTools > Network tab and throttle to "Slow 3G"
4. Click place order and as soon as you see the network request being made immediately click "Force Unmount CheckoutProcessor".
5. Observe the "Cancelled" status of the network request in your DevTools

### Changelog

> Cancel order processing network request if `CheckoutProcessor` component unmounts

### DIFF
--- a/assets/js/base/context/providers/cart-checkout/checkout-processor.js
+++ b/assets/js/base/context/providers/cart-checkout/checkout-processor.js
@@ -33,6 +33,12 @@ import { useStoreNotices } from '../../hooks/use-store-notices';
  * Subscribes to checkout context and triggers processing via the API.
  */
 const CheckoutProcessor = () => {
+	const controller = useRef(
+		typeof AbortController === 'undefined'
+			? undefined
+			: new AbortController()
+	);
+
 	const {
 		hasError: checkoutHasError,
 		onCheckoutValidationBeforeProcessing,
@@ -200,6 +206,7 @@ const CheckoutProcessor = () => {
 			data,
 			cache: 'no-store',
 			parse: false,
+			signal: controller.current?.signal,
 		} )
 			.then( ( response ) => {
 				processCheckoutResponseHeaders(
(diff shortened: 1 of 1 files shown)


assets/js/base/context/providers/cart-checkout/checkout-processor.js
```diff
@@ -33,6 +33,12 @@ import { useStoreNotices } from '../../hooks/use-store-notices';
  * Subscribes to checkout context and triggers processing via the API.
  */
 const CheckoutProcessor = () => {
+	const controller = useRef(
+		typeof AbortController === 'undefined'
+			? undefined
+			: new AbortController()
```
nerrad, line 39: We dropped support for IE11 so in theory it should be okay to just assume this is safe to use? What are your thoughts on that?

nerrad (commented):
Nice! Thanks for doing this Tom!

Despite the continued presence of the related issue in the backlog. I'm on the fence whether we need to do this. Also if the checkout processor is unmounted while the request is still processing, I wonder how that would potentially impact the actual behaviour of the checkout flow given the server may complete the order?

tjcafferkey:
@nerrad yeah that's a _very_ good point. If the user hits "Place Order" regardless of whether the component is unmounted or not we don't want to cancel that request to prevent the user from successfully placing an order.

If we are in agreement, I'm happy to close this PR and related issue with the above comment?

nerrad:
> If we are in agreement, I'm happy to close this PR and related issue with the above comment?

Yup 👍🏻. This approach may be useful in other blocks (other than checkout) where we might have in progress requests but definitely makes sense to leave this particular place alone for now.

tjcafferkey:
If the user hits "Place Order" regardless of whether the component is unmounted or not we don't want to cancel that request to prevent the user from successfully placing an order. For this reason I'm closing this PR.

PR #4760, opened 2021-09-16
Switch to external icon for shipping/payment methods

### LINKED ISSUE #4722: Checkout I2: Lacking clear indication Shipping Zones link out to shipping management pages.
## Describe the bug
I would classify this as a UI/UX bug even though I realize this _may_ be implemented according to design. Currently with the Checkout I2 block, when the Shipping Options inner block is selected, the Zones inspector controls have options that appear as items that can be edited inline, however when you click them it opens a new tab in the browser which is a bit jarring given there is no other signal that will happen.

CleanShot 2021-09-10 at 14 50 55

I think there should be some indication that clicking these links will open up a separate browser tab. For now we should probably use the external link icon instead of the right arrow icon? cc @vivialice for feedback.

mikejolley:
Switches icons to make it clear links open a different page.

Fixes #4722

### Screenshots

Before:

Screenshot 2021-09-16 at 13 52 45

After:

Screenshot 2021-09-16 at 13 51 17

1. View the Checkout block in the editor
2. Select the "Shipping Options" or "Payment Methods" inner block
3. See icons in the sidebar

### DIFF
--- a/assets/js/editor-components/external-link-card/index.tsx
+++ b/assets/js/editor-components/external-link-card/index.tsx
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { Icon, chevronRight } from '@woocommerce/icons';
+import { Icon, external } from '@wordpress/icons';
 import { VisuallyHidden } from '@wordpress/components';
 
 /**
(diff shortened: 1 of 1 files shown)


opr (approved):
Works as described 👍🏼

PR #4761, opened 2021-09-16
Validate state on options change

### LINKED ISSUE #4753: Can't checkout in NZ - "state" validation error, possibly when store is in USA and customer in NZ
## Describe the bug
When checking out with the checkout block (i1 or i2), I get an error about an invalid `state`. In NZ we don't have states, and they are not usually required.

Error message: `The provided state is not valid. Must be one of: NL, AK, WA, BP, TK, GI, HB, MW, WE, NS, MB, TM, WC, CT, OT, SL`

can't purchase - invalid state - i2

My woo store is based in USA (don't remember doing that, seems important!), and I'm (shopper/browser) in NZ. 

Screen Shot 2021-09-16 at 9 21 27 AM

## To reproduce
- Store in USA.
- Customer in NZ (use proxy?)
- Use a checkout block (not shortcode).
- Add to cart, attempt to checkout with valid address.

## Expected behavior
Form required inputs and validation agree.

## Screenshots
Here's a couple of shots of the billing address form for me in i1 and i2. 

Screen Shot 2021-09-16 at 9 19 37 AM
Screen Shot 2021-09-16 at 9 19 47 AM

mikejolley:
Validates the value of the state against the current list of states to ensure it's valid.

Fixes #4753

### Testing

1. Setup base location to have a state (US Address)
2. Checkout with a NZ address
3. Do not select a state during checkout. Ensure you can place the order.

This was broken before because the customer would inherit the default store state. E.g. AL (Alabama, US). When checking out from New Zealand, AL would be preserved, but since state is optional, there is no event triggered to update AL to something else. AL is not a valid state for New Zealand.

### Changelog

> Fix state validation if base location has a state, and the address has an optional state.

### DIFF
--- a/assets/js/base/components/state-input/state-input.tsx
+++ b/assets/js/base/components/state-input/state-input.tsx
@@ -3,7 +3,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { decodeEntities } from '@wordpress/html-entities';
-import { useCallback, useMemo } from '@wordpress/element';
+import { useCallback, useMemo, useEffect, useRef } from '@wordpress/element';
 import classnames from 'classnames';
 
 /**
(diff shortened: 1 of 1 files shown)


opr (approved):
LGTM and works well 👍🏼

mikejolley:
Thanks for testing @opr 👏🏻

PR #4764, opened 2021-09-16
Make All Reviews block honor 'ratings enabled' and 'show avatars' preferences

Aljullu:
While investigating build sizes of Reviews blocks, I noticed that All Reviews was missing some code to honor 'ratings enabled' and 'show avatars' preferences. That same code [exists in the other Reviews blocks](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/ReviewsByCategory.php#L31-L42), so it's probably a bug.

### Manual Testing

1. In wp-admin, go to _Settings_ > _Discussion_ > _Avatars_ and uncheck _Show Avatars_.
2. Add an All Reviews block and verify avatars are not shown and instead there is a warning in the sidebar:

3. In wp-admin, go to _WooCommerce_ > _Settings_ and uncheck _Enable star rating on reviews_.
4. In the All Reviews block, verify ratings and the _Sort by_ select aren't displayed.

### Changelog

> Update All Reviews block so it honors 'ratings enabled' and 'show avatars' preferences.

### DIFF
--- a/assets/js/base/components/reviews/review-list-item/index.js
+++ b/assets/js/base/components/reviews/review-list-item/index.js
@@ -158,6 +158,7 @@ const ReviewListItem = ( { attributes, review = {} } ) => {
 				'wc-block-components-review-list-item__item',
 				{
 					'is-loading': isLoading,
+					'wc-block-components-review-list-item__item--has-image': showReviewImage,
 				}
 			) }
 			aria-hidden={ isLoading }
--- a/assets/js/base/components/reviews/review-list-item/style.scss
+++ b/assets/js/base/components/reviews/review-list-item/style.scss
@@ -57,7 +57,7 @@
 	grid-row: 1;
 }
 
-.has-image {
+.wc-block-components-review-list-item__item--has-image {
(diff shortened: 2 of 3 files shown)


tjcafferkey:
Screenshot 2021-09-17 at 09 29 56

Functionally this works as expected. However there's an empty space to the left of the reviewer where the avatars used to be because the block still has the `has-image` class, removing this fixes this visual issue. Do you think we should do this?

Aljullu:
Thanks for the review @tjcafferkey!

> Functionally this works as expected. However there's an empty space to the left of the reviewer where the avatars used to be because the block still has the `has-image` class, removing this fixes this visual issue. Do you think we should do this?

Good point. I don't think we can easily remove the `has-image` class because that one is tied to the `Image` attribute of the block:

Image attribute in Reviews blocks

So instead of that, I just created a new class inside the Review List Item component and then use that one in the CSS. Let me know how it looks!

tjcafferkey (approved):
Nice! Thanks for the changes, this LGTM 🚀  😃

PR #4765, opened 2021-09-16
Show message if user has not yet calculated shipping and no shipping rates exist

### LINKED ISSUE #4752: No Message is shown in Shipping Options Step when there are no rates, and no address
1. In woo core setting settings, enable "Hide shipping until an address is entered"
2. Open checkout in incognito mode as a guest
3. See form step is rendered for shipping, but it's empty

There seems to be some text passed to `ShippingRatesControl` but it's not used. https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8679980306694a164593d8d9bda7edf5ded2e24b/assets/js/blocks/cart-checkout/checkout/form/shipping-options-step.tsx#L105-L108

This affects Checkout i1 and i2.

mikejolley:
When the user has not yet calculated shipping, and there are no rates, the shipping options step shows no content. I think this may have been broken when we introduced slot fills in this area. There was a "no results message", but this is only shown if there are no rates within a package.

Fixes #4752

### Screenshots

Before:
Screenshot 2021-09-16 at 16 13 52

After:
Screenshot 2021-09-16 at 16 11 47

### Testing

1. In WooCommerce Shipping Settings, enable _"Hide shipping until an address is entered"_
2. Open checkout in a new incognito window as a guest
3. Add an item to the cart and go to the checkout
4. Confirm the shipping section shows a message: "Shipping options will be displayed here after entering your full shipping address."

### Changelog

> Show placeholder message in the shipping section when there are no rates.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-shipping-methods-block/block.tsx
+++ b/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-shipping-methods-block/block.tsx
@@ -61,32 +61,44 @@ const Block = (): JSX.Element | null => {
 		return null;
 	}
 
+	const shippingRatesPackageCount = getShippingRatesPackageCount(
+		shippingRates
+	);
+
+	if (
+		! isEditor &&
+		! hasCalculatedShipping &&
+		! shippingRatesPackageCount
+	) {
+		return (
+			<p>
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
Works as described and code looks good. :ship:

PR #4772, opened 2021-09-17
Open Mini Cart drawer when adding to cart from the All Products block

### LINKED ISSUE #4688: Mini Cart block: open drawer when a product is added from the All Products block

Aljullu:
Fixes #4688.

This PR includes several changes:
* When adding a product using the Add to Cart product element block, the events `wc-blocks_adding_to_cart` and `wc-blocks_added_to_cart` are now triggered. The Mini Cart block listens to those events, so it now opens when adding a product to cart from the All Products or Single Product blocks.
* I updated the code that triggers DOM events to create events with `CustomEvent`, that allows us to attach some data to them.
* The `wc-blocks_adding_to_cart` event now accepts an `invalidateCartData` property. If it's set to false, data in the cart data store will not be invalidated when triggered. Useful to avoid the Cart block reloading all its contents when a product is added to cart from a block that is already connected to our data store (ie All Products).
* I removed some IE11 specific code.
* I added a small check inside `showContents()` to load scripts if they hadn't been loaded yet. Should be useful if the `wc-blocks_added_to_cart` event is ever triggered without the `wc-blocks_added_to_cart` having been triggered before.

### Screenshots

https://user-images.githubusercontent.com/3616980/133733589-cf3b8d26-069e-4ac1-a910-59f4cf1df267.mp4

### Manual Testing

**Test that Mini Cart opens when adding a product to cart via the All Products block:**

1. Create a post or page with the Mini Cart block + the All Products block.
2. Add a product to cart using the All Products block.
3. Verify the drawer opens.

**Test that Cart block updates when adding products to cart via a PHP product grid block:**

1. Create a post or page with a PHP product grid block (ie: Handpicked products) and the Cart block.
2. Add a product to cart using the PHP product grid block.
3. Verify the Cart contents are reloaded.

**Test that Cart block updates when adding products to cart via the All Products block:**

1. Create a post or page with the All Products block and the Cart block.
2. Add a product to cart using the All Products block.
3. Verify the Cart contents are updated without causing the full cart to reload (there is no spinner covering the entire block, instead, the new product has just been automatically added).


assets/js/base/utils/legacy-events.ts
```diff
@@ -50,6 +59,23 @@ export const triggerFragmentRefresh = (): void => {
...
 
+export const triggerAddingToCartEvent = (): void => {
+	dispatchEvent( 'wc-blocks_adding_to_cart', {
+		bubbles: true,
+		cancelable: true,
+	} );
+};
+
+export const triggerAddedToCartEvent = ( {
+	invalidateCartData,
```
tjcafferkey: Is it worth giving this a default value of `false` and define a type for it? Given `undefined` is a falsey value anyway (and you are strictly checking against `false` elsewhere in this file) and we can change its usage in this file from:

`yield triggerAddedToCartEvent( { invalidateCartData: false } );`

to 

`yield triggerAddedToCartEvent();`

tjcafferkey (approved):
This works as expected for me so approving it. Just let a comment against some of the code for review.

assets/js/base/utils/legacy-events.ts
Aljullu: > Is it worth giving this a default value of `false`

Good point. I took another look at the code and I think the way I wrote it was a bit confusing, because it's true that here `invalidateCartData` defaults to false, but then in the React components I wrote this check: `! eventDetail || eventDetail.invalidateCartData !== false`, which would default to true if it's undefined.

In 17747aae5ab53091b9d2438df8386ca5fe9e9a98, I refactored the code so now the property is named `preserveCartData` instead of `invalidateCartData`. This way it can default to false in all cases, I think this makes the code easier to understand, but maybe I'm just overcomplicating this? :sweat_smile: 

> define a type for it

The type comes from `AddToCartEventDetail`, right?

tjcafferkey: >I refactored the code so now the property is named preserveCartData instead of invalidateCartData. This way it can default to false in all cases, I think this makes the code easier to understand, but maybe I'm just overcomplicating this?

Thanks! I think this makes more sense and its easier to understand!

>The type comes from AddToCartEventDetail, right?

You're absolutely right, my bad sorry!

All LGTM 🚀 😄

PR #4774, opened 2021-09-17
Add support for extensions to filter express payment methods

### LINKED ISSUE #4733: Allow express payment methods to be filtered by extensions
In #4621 we introduced a new way for extensions to register their own callbacks to determine whether a payment method is supported.

We need to also include payment methods in this extensibility point.

ralucaStan:
This PR adds support for extensions to also filter express payment methods.

Fixes #4733 

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Make sure you use the Checkout block and that you are on https
2. In an extension register an express payment method
```
registerExpressPaymentMethod({
        name: "expressMethod",
        edit: null,
        content: <div>My express test method</div>,
        canMakePayment: () => true,
        paymentMethodId: "expressMethod",
        supports: {
          features: ["products"],
        },
 })
```
3. Go to the Checkout block and see that the express payment method appears
4. In the same extension register a callback to remove the above express method
```
	{
		expressMethod: ( arg ) => { return false; },
	}
);
```
5. Go to the Checkout block and see that the express payment method disappears

### Changelog

> Add support for extensions to filter express payment methods

### DIFF
--- a/assets/js/blocks-registry/payment-methods/payment-method-config-helper.ts
+++ b/assets/js/blocks-registry/payment-methods/payment-method-config-helper.ts
@@ -6,10 +6,11 @@ import type { CanMakePaymentCallback } from '@woocommerce/type-defs/payments';
 /**
  * Internal dependencies
  */
-import type {
+import {
 	NamespacedCanMakePaymentExtensionsCallbacks,
 	PaymentMethodName,
 	ExtensionNamespace,
+	extensionsConfig,
 } from './extensions-config';
 
 // Filter out payment methods by supported features and cart requirement.
@@ -17,7 +18,7 @@ export const canMakePaymentWithFeaturesCheck = (
 	canMakePayment: CanMakePaymentCallback,
 	features: string[]
 ): CanMakePaymentCallback => ( canPayArgument ) => {
-	const requirements = canPayArgument.paymentRequirements || [];
+	const requirements = canPayArgument?.paymentRequirements || [];
 	const featuresSupportRequirements = requirements.every( ( requirement ) =>
 		features.includes( requirement )
 	);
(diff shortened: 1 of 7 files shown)


assets/js/blocks-registry/payment-methods/payment-method-config-helper.ts
```diff
@@ -65,3 +66,22 @@ export const canMakePaymentWithExtensions = (
...
+		PaymentMethodName,
+		CanMakePaymentCallback
+	>[] ).some( ( callbacks ) => paymentMethodName in callbacks )
+		? canMakePaymentWithExtensions(
+				canPay,
+				extensionsConfig.canMakePayment,
+				paymentMethodName
+		  )
+		: canPay;
+};
```
opr, line 87: Nice, I like that this has been moved out, makes things a bit cleaner :)

assets/js/blocks-registry/payment-methods/test/payment-method-config-helper.ts
```diff
@@ -47,53 +104,89 @@ describe( 'canMakePaymentWithExtensions', () => {
...
-			canMakePaymentExtensionsCallbacks,
-			'bacs'
-		)();
-		expect( falseCallback ).toHaveBeenCalledTimes( 1 );
+			const canMakePayment = helpers.getCanMakePayment(
+				args.canMakePaymentConfiguration,
+				args.features,
+				args.paymentMethodName
+			)( canMakePaymentArgument );
+			expect( canMakePayment ).toBe( falseCallback() );
```
opr: Doesn't `falseCallback()` evaluate to simply `false`, so you're testing here `expect( canMakePayment ).toBe( false );` or am I misunderstanding?

opr (changes requested):
This works as advertised 👏🏼  Just a query about the tests. From simply reading the tests, it's not super clear what we're testing, or why we're expecting certain things... I would love it if you added inline comments with an explanation about why we're expecting the result.

assets/js/blocks-registry/payment-methods/test/payment-method-config-helper.ts
ralucaStan: I've updated the test in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4774/commits/630a15ed3e465a247f9db92b7c1bb8d9049d0e37, hope this makes it more clear

opr (approved):
Thanks for making the changes and describing the tests better! 🙌🏼

PR #4775, opened 2021-09-17
Debounce billingData to be used as a dependency on the useEffect hook

### LINKED ISSUE #4766: canMakePayment isn’t tracking billingData changes
When testing the release of 5.9. We stumbled on an issue with canMakePayment, it didn't update when billing address change.
It would rerun because updating billing address sends a server request, and because `cartTotals` isn't well memozied, `canMakePayment` end up running. If we ever fix that memozion part, `canMakePayment` won't rerun.

We discovered this issue when we tried to test `canMakePaymentExtensionsCallbacks` with a use case that triggers a server error. Triggering server erros means `cartTotals` won't update and so `canMakePayment` didn't rerun.

A quick fix to that issue was to add `billingData` to this useEffect hook
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/ca3b3f0fda1582c919b13e538cccca9f6733ab3d/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts#L164-L171

Adding `billingData` there fixed the issue with `canMakePayment` not running on address, but introduced a new issue of `canMakePayment` rerunning on keystroke, which could be dangerous depending on the content of each `canMake` function.

We decided it's best if we debounced or throttled that call before diving into it.

tjcafferkey:
## Description

Debounce `billingData` to be used as a dependency in the useEffect hook. Debouncing is necessary to prevent an excessive amount of useEffect executions.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4766

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Add the following snippet to the top of the changed file (`assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts`) for the purpose of testing only.

```
registerPaymentMethodExtensionCallbacks( 'woocommerce-marketplace-extension', {
	cod: ( arg ) => arg.billingData.first_name === 'Tom',
} );
```

and add `registerPaymentMethodExtensionCallbacks` to the existing import of `'@woocommerce/blocks-registry'`

2. On the checkout input a first name which _is not_ "Tom" in the Billing Address section. Cash on Delivery option should not be available as a payment method.
3. Update this to be "Tom" and observe the change to the payment methods which now should show Cash on Delivery as a valid option.

### Changelog

> usePaymentMethodRegistration: Debounce billingData updates.

### DIFF
--- a/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
+++ b/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
@@ -15,6 +15,7 @@ import type {
 	PaymentMethodConfigInstance,
 	ExpressPaymentMethodConfigInstance,
 } from '@woocommerce/type-defs/payments';
+import { useDebounce } from 'use-debounce';
 
 /**
  * Internal dependencies
@@ -49,6 +50,9 @@ const usePaymentMethodRegistration = (
 	const { isEditor } = useEditorContext();
 	const { selectedRates } = useShippingDataContext();
 	const { billingData, shippingAddress } = useCustomerDataContext();
+
+	// Debouncing billingData to use as a dependency on refreshCanMakePayments and prevent excessive useEffect executions.
+	const [ debouncedBillingData ] = useDebounce( billingData, 750 );
 	const selectedShippingMethods = useShallowEqual( selectedRates );
 	const paymentMethodsOrder = useShallowEqual( paymentMethodsSortOrder );
 	const {
@@ -170,6 +174,7 @@ const usePaymentMethodRegistration = (
 		cartTotals,
 		selectedShippingMethods,
 		paymentRequirements,
+		debouncedBillingData,
 	] );
 
 	return isInitialized;


senadir:
Hey @tjcafferkey 
While this PR is correct and is doing what the issue asked for, the discussion around this evolved and instead of debouncing billingData, we're looking at throttling `refreshMakePayment`.
You can see the lengthy thread here p1632135581442800/1632126365.428300-slack-C8X6Q7XQU

The gist of it is that, while we can debounce `billingData`, it's just one thing that can trigger `refreshMakePayment` 1 too many times, and a better approch would be to throttle `refreshMakePayment`. This would save us from debouncing `billingData`, `cart`, `cartTotals`, and any other variable that changes too much that we add to that hook.

cc @ralucaStan and @nerrad

ralucaStan (approved):
This tests out 👍.

tjcafferkey:
@senadir would it make sense to merge this PR given its a fix and then reopen another issue with the proposed solution you've outlined in your comment? This way it's better than what it _was_ but agreed not ideal.

senadir:
Sure thing! Just mark that line with a `@todo` and an issue would be automatically created for you.

ralucaStan:
Closed by https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4776

PR #4776, opened 2021-09-17
Update canMakePayment to receive cart as argument and make it react to changes in billingData.

### LINKED ISSUE #4670: Add coupon information to data passed to canMakePayment
The Conditional Shipping and Payments extension needs to have access to the coupon code data present in the cart.
It should have this info available after each change in the checkout block in order to determine if a payment method can be used. 

Below is the date needed by the extension.

This can be achieved if we pass coupon data to this [canMakePayment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/ca3b3f0fda1582c919b13e538cccca9f6733ab3d/docs/extensibility/payment-method-integration.md#canmakepayment-required) callback.

### LINKED ISSUE #4766: canMakePayment isn’t tracking billingData changes

ralucaStan:
This PR 
- adds the  `cart` object to the argument we pass to the `canMakePayment` callback
- it adds `billingData` to the dependencies list of the effect that calls `refresCanMakePayment` and closes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4766. 
- It improves the performance of calculating if a payment method is still active after changes in the Checkout block:
   - it debounces refreshMakePayment for the reasons from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4775#issuecomment-923958392. Debounced was chosen instead of throttle because we want to call refreshCanMakePayments once the change event has stopped, with the final value.
- it adds some types related to coupon data made while investigating #4670 

Fixes: 
#4670 
#4766

A new ticket has been created for deprecating cart-derived values from the `canMakePayment` argument https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4809

1. Go to `assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts` and edit for the purpose of testing only:

- add `registerPaymentMethodExtensionCallbacks` to the existing import from `'@woocommerce/blocks-registry'`
- register a callback for COD at the top of the file, after the imports
```
registerPaymentMethodExtensionCallbacks( 'woocommerce-marketplace-extension', {
	cod: ( arg ) => {
		console.log( 'checking COD' );
		return arg.billingData.first_name === 'Alexandra';
	},
} );
```
2. Make the console visible
2. Go to Checkout block and notice that COD payment method is missing and that an initial check was made COD (see console.log())
3. Deselect `Use same address for billing` and write "Alexandra" for First Name in the Billing Address section. Cash on Delivery option should be available as a payment method.
3. Notice that the check for COD is done only after the user finished typing

### Changelog

> Update canMakePayment to receive cart as argument and make it react to changes in billingData.  Improve the performance of calculating canMakePayment after changes in the Checkout block.

### DIFF
--- a/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
+++ b/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
@@ -52,11 +52,13 @@ const usePaymentMethodRegistration = (
 	const selectedShippingMethods = useShallowEqual( selectedRates );
 	const paymentMethodsOrder = useShallowEqual( paymentMethodsSortOrder );
 	const {
+		cartCoupons,
 		cartTotals,
 		cartNeedsShipping,
 		paymentRequirements,
 	} = useStoreCart();
 	const canPayArgument = useRef( {
+		cartCoupons,
 		cartTotals,
 		cartNeedsShipping,
 		billingData,
@@ -68,6 +70,7 @@ const usePaymentMethodRegistration = (
 
 	useEffect( () => {
 		canPayArgument.current = {
+			cartCoupons,
 			cartTotals,
 			cartNeedsShipping,
 			billingData,
@@ -76,6 +79,7 @@ const usePaymentMethodRegistration = (
 			paymentRequirements,
 		};
 	}, [
+		cartCoupons,
 		cartTotals,
 		cartNeedsShipping,
 		billingData,
--- a/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
+++ b/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
@@ -15,6 +15,7 @@ import type {
 	PaymentMethodConfigInstance,
 	ExpressPaymentMethodConfigInstance,
 } from '@woocommerce/type-defs/payments';
+import { useDebouncedCallback } from 'use-debounce';
 
 /**
  * Internal dependencies
@@ -160,12 +161,22 @@ const usePaymentMethodRegistration = (
 		registeredPaymentMethods,
 	] );
 
+	const [ debouncedRefreshCanMakePayments ] = useDebouncedCallback(
+		refreshCanMakePayments,
+		500
+	);
+
 	// Determine which payment methods are available initially and whenever
 	// shipping methods, cart or the billing data change.
 	// Some payment methods (e.g. COD) can be disabled for specific shipping methods.
 	useEffect( () => {
-		refreshCanMakePayments();
-	}, [ refreshCanMakePayments, cart, selectedShippingMethods, billingData ] );
+		debouncedRefreshCanMakePayments();
+	}, [
+		debouncedRefreshCanMakePayments,
+		cart,
+		selectedShippingMethods,
+		billingData,
+	] );
 
 	return isInitialized;
 };
--- a/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
+++ b/assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
@@ -51,14 +51,10 @@ const usePaymentMethodRegistration = (
 	const { billingData, shippingAddress } = useCustomerDataContext();
 	const selectedShippingMethods = useShallowEqual( selectedRates );
 	const paymentMethodsOrder = useShallowEqual( paymentMethodsSortOrder );
-	const {
-		cartCoupons,
-		cartTotals,
-		cartNeedsShipping,
-		paymentRequirements,
-	} = useStoreCart();
+	const cart = useStoreCart();
+	const { cartTotals, cartNeedsShipping, paymentRequirements } = cart;
 	const canPayArgument = useRef( {
-		cartCoupons,
+		cart,
 		cartTotals,
 		cartNeedsShipping,
 		billingData,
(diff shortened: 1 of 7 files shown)


ralucaStan:
@senadir the PR is ready for review.

assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
```diff
@@ -68,6 +67,7 @@ const usePaymentMethodRegistration = (
 
 	useEffect( () => {
 		canPayArgument.current = {
+			cart,
```
senadir, line 70: How about spreading cart here, which would include cartTotals, cartNeedsShipping, paymentRequirements, and also add the rest of items at the same level?

assets/js/blocks/cart-checkout/payment-methods/test/payment-methods.js
```diff
@@ -63,6 +67,22 @@ const resetMockPaymentMethods = () => {
...
+			return Promise.resolve( '' );
+		} );
+		// need to clear the store resolution state between tests.
+		await dispatch( storeKey ).invalidateResolutionForStore();
+		await dispatch( storeKey ).receiveCart( defaultCartState.cartData );
+	} );
+
+	afterEach( () => {
+		fetchMock.resetMocks();
+	} );
```
senadir, line 84: Why was this added?

docs/extensibility/payment-method-integration.md
```diff
@@ -87,11 +87,13 @@ A callback to determine whether the payment method should be available as an opt
 
 ```
 canMakePayment( {
+    cart: Cart,
+    cartCoupons: CartCoupons,
```
senadir: cartCoupons is no longer exposed

assets/js/types/type-defs/payments.ts
```diff
@@ -27,6 +28,8 @@ export interface Supports extends SupportsConfiguration {
 }
 
 export interface CanMakePaymentArgument {
+	cart: Cart;
+	cartCoupons: CartResponseCoupons;
```
senadir: Also here.

senadir (approved):
This checks out, thank you for working on this Raluca!

I left a small comment around spreading cart and some left over code.

The changelog entry also seems to be not part of the quote? you might want to remove the empty line to ensure it's included.

assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
ralucaStan: I would advise against it, as `cart` has [a lot of keys](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7595c3374dddeba8e201f159d9a37c09ae27b309/assets/js/types/type-defs/cart.ts#L173-L189), and that would make the signature really confusing. After deprecating the keys you mentioned it would just be `cart` and `billingData` and that makes it easier to understand where the data comes from.

docs/extensibility/payment-method-integration.md
ralucaStan: Really good catch

assets/js/base/context/providers/cart-checkout/payment-methods/use-payment-method-registration.ts
senadir: Wouldn't having cart as is just complicated one step further? so instead of `arg.cartTotals` you need `arg.cart.cartTotals` especially seeing that canPayArgument is made of cart related fields only.

ralucaStan: It's also made of `billingData` and `shipping data` from the `CustomerDataContext`

senadir: Those are also present in Cart, shipping data is `shippingAddress` and billing data is `billingAddress`.

The only difference here is that `billingAddress` reflects what's currently in state, and what's in `Cart` reflects what comes from the server, so this might not always be in sync. billingData can still replace billingAddress from cart.

This is not a huge point to block the PR on as moving the other direction (spreading) is easier than reversing.

:shipit:

ralucaStan: The data is indeed out of sync. This would be an issue for extensions that need to make calculations on the current `billingData` or `shippingData`.

> This is not a huge point to block the PR on as moving the other direction (spreading) is easier than reversing.
👍

assets/js/blocks/cart-checkout/payment-methods/test/payment-methods.js
ralucaStan: a call was being made by `useStoreCart` and it needed to be mocked. I improved to mock to make it clearer.

ralucaStan:
Please take another look, I've addressed your requests

PR #4778, opened 2021-09-17
Add automation for automatically assigning the team the pr author belongs to for reviewing

nerrad:
Description to come.


nerrad:
I think the CODEOWNERS is read from trunk so this might not work until the CODEOWNERS file is removed from trunk.

nerrad:
Closing again, going to try temporarily deleting CODEOWNERS in trunk

PR #4780, opened 2021-09-20
Setup basic column blocks for Cart i2

senadir:
This start the work on Cart i2 by creating two new inner blocks for columns.
We still don't support empty cart for now.

Part of #4709

### Testing
- Cart is working fine, can be inserted without problem.

Issue #4709, opened 2021-09-10
Break down Cart i2 block into 4 different blocks.


senadir:
We need to break Cart i2 block into those blocks:
- Two columns block.
- A block that contain all items.
- A block that contain total lines.
- A block with Proceed to Checkout block.
- A block with Express Payment blocks.

mikejolley:
@senadir This makes sense. We could look at adding specific blocks later on for things like cross/sells upsells too. Optional.

senadir:
This was handled individually by PRs.

PR #4781, opened 2021-09-20
Add Cart Line Items Block to Cart i2

senadir:
Adds Cart Line Items block (Couldn't use Cart Item Block because it's already used).

Part of #4709, based on #4780

## Testing Steps
- Insert Cart i2.
- Make sure you can see items.
- Make sure it's using preview data.


mikejolley (approved):
I've rebased this and it works :)

PR #4783, opened 2021-09-20
Adds Proceed to Checkout block to Cart i2

senadir:
Adds Proceed to Checkout block (only block with no `cart` in its id).

This block includes the button, and payment methods icons? I am not sure why those two are together.

Part of #4709, based on #4780

## Testing Steps

- Insert Cart i2.
- Make sure you can see button


mikejolley (approved):

mikejolley:
Some linting issues we can fix later.

PR #4789, opened 2021-09-20
Add e2e tests for Newest Products

nielslange:
### Note

Currently, not every block is covered by e2e tests. This PR aims to add e2e tests for the Newest Products block.

Part of #4643 

### Testing

1. Check out this PR.
2. Start Docker Desktop.
3. Run `npm run wp-env start && npm run test:e2e`.
4. The e2e test suite shows no failed tests.

### DIFF
--- a/tests/e2e/config/custom-matchers/__fixtures__/newest-products.fixture.json
+++ b/tests/e2e/config/custom-matchers/__fixtures__/newest-products.fixture.json
@@ -1,0 +1,1 @@
+{"title":"Newest Products Block","pageContent":"<!-- wp:woocommerce/product-new /-->"}
--- a/tests/e2e/specs/backend/product-new.test.js
+++ b/tests/e2e/specs/backend/product-new.test.js
@@ -1,0 +1,30 @@
+/**
+ * External dependencies
+ */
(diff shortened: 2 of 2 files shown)


tjcafferkey:
Shouldn't we also add `/frontend` e2e tests for this block too?

tjcafferkey:
I get the following errors when running the specified npm commands, are there any prerequisite steps I am missing?

Screenshot 2021-10-08 at 14 51 17

nielslange:
> I get the following errors when running the specified npm commands, are there any prerequisite steps I am missing?

My bad. Docker Desktop needs to be up and running, so that `wp-env` can spin up a testing environment there. I'll update the description accordingly.

> Shouldn't we also add /frontend e2e tests for this block too?

For the other blocks, we only check the frontend if it differs much from the backend. Would you say it's beneficial in this case to check the frontend additionally?

tjcafferkey:
>For the other blocks, we only check the frontend if it differs much from the backend. Would you say it's beneficial in this case to check the frontend additionally?

I'd say the more (reliable) tests the better, but I wouldn't reject this PR because of it. We can always follow up with frontend E2E tests if needed.

tjcafferkey (approved):
LGTM

PR #4790, opened 2021-09-20
Add object type to lock attributes

mikejolley:
Some of the inner blocks in Checkout i2 could still be removed. This was due to missing `"type": "object",` in the block json. 

**To test:**
1. Apply patch
2. Ensure express payment block cannot be moved or removed


senadir (approved):
Checks out and works!

Issue #4793, opened 2021-09-21
Cart i2: Add empty cart default inner blocks


senadir:
Empty Cart view is currently broken in Cart i2, so we removed it. We should add it back and make sure it works the same way as Cart i1:
- Toggling it would unmount Full Cart blocks and their selection.
- The value persist (going to be hard to have to innerBlocks on the same level).

senadir:
As of now, those points are fixed:

> Toggling it would unmount Full Cart blocks and their selection.
The value persist (going to be hard to have to innerBlocks on the same level).


What's left is adding the default empty cart inner block layout.

PR #4803, opened 2021-09-21
Add/empty and fillted cart wrapper

mikejolley:
This is part of #4793

Shifts cart-i2 inner blocks down a level into "filled cart block", then introduces a sibling called "empty cart block". There is no view switcher—this just takes care of the structure.

**Testing:**
- Insert a fresh cart-i2 block
- Confirm all inner blocks are rendered
- Confirm placeholder for empty cart block is rendered below the cart view


senadir (approved):
Checks out and works as expected.

PR #4810, opened 2021-09-22
#3906 screen reader fix for order summary block accessibility 

### LINKED ISSUE #3906: Order summary: with a screen reader it's not clear what's the difference between the cart item subtotal and the product price
In the Checkout Order Summary, we display all products with its unitary price and the cart item price. While the alignment helps explain their meaning (unitary price on the left, under the title / cart item price aligned to the right in bold), nothing of this is obvious to screen reader users, which will get two prices read without any clue of which one is which.

Order summary screenshot

ovidiul:
Fixes #3906 

#### Accessibility

### Screenshots

Screen Recording 2021-09-22 at 13 58 25

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Add one or more items to your cart
2. Create a post/page add the `Checkout` block
3. Open the new page from previous step
4. Click on `Order Summary` to expand the tab, or navigate to it using TAB key
5. Navigate using the `TAB` key
6. Optionally you could enable the screen reader

### Changelog

> Improve the Checkout Order Summary block accessibility by making more info available to screen readers.

### DIFF
--- a/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
+++ b/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { __, sprintf } from '@wordpress/i18n';
+import { sprintf, _n } from '@wordpress/i18n';
 import Label from '@woocommerce/base-components/label';
 import ProductPrice from '@woocommerce/base-components/product-price';
 import ProductName from '@woocommerce/base-components/product-name';
@@ -116,7 +116,12 @@ const OrderSummaryItem = ( { cartItem } ) => {
 						label={ quantity }
 						screenReaderLabel={ sprintf(
 							/* translators: %d number of products of the same type in the cart */
-							__( '%d items', 'woo-gutenberg-products-block' ),
+							_n(
+								'%d item',
+								'%d items',
+								quantity,
+								'woo-gutenberg-products-block'
+							),
 							quantity
 						) }
 					/>
@@ -124,27 +129,11 @@ const OrderSummaryItem = ( { cartItem } ) => {
 				<ProductImage image={ images.length ? images[ 0 ] : {} } />
 			</div>
 			<div className="wc-block-components-order-summary-item__description">
-				<span className="screen-reader-text" tabIndex={ 0 }>
-					{ sprintf(
-						/* translators: %d items of product %s */
-						__( '%1$d %2$s', 'woo-gutenberg-products-block' ),
-						quantity,
-						name
-					) }
-				</span>
 				<ProductName
 					disabled={ true }
 					name={ name }
 					permalink={ permalink }
 				/>
-
-				<span className="screen-reader-text" tabIndex={ 0 }>
-					{ sprintf(
-						/* translators: Unit Price %s */
-						__( 'Unit Price: %s', 'woo-gutenberg-products-block' ),
-						formatPrice( priceSingle, priceCurrency )
-					) }
-				</span>
 				<ProductPrice
 					currency={ priceCurrency }
 					price={ priceSingle }
@@ -170,15 +159,18 @@ const OrderSummaryItem = ( { cartItem } ) => {
 					variation={ variation }
 				/>
 			</div>
-			<span className="screen-reader-text" tabIndex={ 0 }>
+			<span className="screen-reader-text">
 				{ sprintf(
-					/* translators: Total Price for %d quantity items: %s */
-					__(
-						'Total Price for %1$d items: %2$s',
+					/* translators: %1$d is the number of items, %2$s is the total price including the currency symbol and %3$s is the item name. */
+					_n(
+						'Total price for %1$d %3$s item: %2$s',
+						'Total price for %1$d %3$s items: %2$s',
+						quantity,
 						'woo-gutenberg-products-block'
 					),
 					quantity,
-					formatPrice( subtotalPrice, totalsCurrency )
+					formatPrice( subtotalPrice, totalsCurrency ),
+					name
 				) }
 			</span>
 			<div className="wc-block-components-order-summary-item__total-price">
--- a/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
+++ b/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
@@ -161,19 +161,22 @@ const OrderSummaryItem = ( { cartItem } ) => {
 			</div>
 			<span className="screen-reader-text">
 				{ sprintf(
-					/* translators: %1$d is the number of items, %2$s is the total price including the currency symbol and %3$s is the item name. */
+					/* translators: %1$d is the number of items, %2$s is the item name and %3$s is the total price including the currency symbol. */
 					_n(
-						'Total price for %1$d %3$s item: %2$s',
-						'Total price for %1$d %3$s items: %2$s',
+						'Total price for %1$d %2$s item: %3$s',
+						'Total price for %1$d %2$s items: %3$s',
 						quantity,
 						'woo-gutenberg-products-block'
 					),
 					quantity,
-					formatPrice( subtotalPrice, totalsCurrency ),
-					name
+					name,
+					formatPrice( subtotalPrice, totalsCurrency )
 				) }
 			</span>
-			<div className="wc-block-components-order-summary-item__total-price">
+			<div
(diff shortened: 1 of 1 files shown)


assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -121,11 +124,27 @@ const OrderSummaryItem = ( { cartItem } ) => {
 				<ProductImage image={ images.length ? images[ 0 ] : {} } />
 			</div>
 			<div className="wc-block-components-order-summary-item__description">
+				<span className="screen-reader-text" tabIndex={ 0 }>
```
Aljullu: We don't need to make these elements focusable, screen readers already offer a way to navigate unfocusable elements.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -151,6 +170,17 @@ const OrderSummaryItem = ( { cartItem } ) => {
 					variation={ variation }
 				/>
 			</div>
+			<span className="screen-reader-text" tabIndex={ 0 }>
+				{ sprintf(
+					/* translators: Total Price for %d quantity items: %s */
```
Aljullu: In the translators comment, you need to explain what each variable will be replaced with. For example, in this case it could be written as:
```suggestion
					/* translators: %1$d is the number of items and %2$s is the total prices including the currency symbol. */
```

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -151,6 +170,17 @@ const OrderSummaryItem = ( { cartItem } ) => {
...
+					__(
+						'Total Price for %1$d items: %2$s',
```
Aljullu: Ideally, we should use sentence case everywhere, even if it's a text for a screen reader:
```suggestion
						'Total price for %1$d items: %2$s',
```

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -121,11 +124,27 @@ const OrderSummaryItem = ( { cartItem } ) => {
...
 					permalink={ permalink }
 				/>
+
+				<span className="screen-reader-text" tabIndex={ 0 }>
+					{ sprintf(
+						/* translators: Unit Price %s */
+						__( 'Unit Price: %s', 'woo-gutenberg-products-block' ),
+						formatPrice( priceSingle, priceCurrency )
+					) }
+				</span>
```
Aljullu: This is a bit tricky because when a product has a discount, we want to announce it to screen readers too. `ProductPrice` already has some logic to handle these cases.

Considering you already added screen reader text to the 'Total price', I wonder if we need to specify this is the unitary price. I think this will be clear now. How does it sound getting rid of this span and just leaving the 'total price' one?

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -121,11 +124,27 @@ const OrderSummaryItem = ( { cartItem } ) => {
 				<ProductImage image={ images.length ? images[ 0 ] : {} } />
 			</div>
 			<div className="wc-block-components-order-summary-item__description">
+				<span className="screen-reader-text" tabIndex={ 0 }>
+					{ sprintf(
+						/* translators: %d items of product %s */
```
Aljullu: I think this translators comment could be made a bit more specific so translators have a better understanding of what these variables are. Something along these lines: `%1$d is the quantity of items and %2$s the name of the product`. How does it sound?

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
Aljullu: Another thing about this string, you will need to account for the case when there is only one item of a specific product. In that case `items` should be `item`. For this, we use `sprintf()` and `_n()`.

You can see an example of how this functions are being used here:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/19d29d11fea059650da48cbca4f35275e88324d6/assets/js/editor-components/product-category-control/index.js#L48-L58

Aljullu (commented):
Thanks for working on this @ovidiul! I like that you also took the chance to improve the accessibility of the count badge. :clap: 

I left some comments below with some suggestions that I would apply. Let me know what you think!

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
ovidiul: Thanks for pointing that out, I've removed it.

Seems [VO key(CTRL+Option) + →] and [VO key + ←] would work for me as opposed to using the TAB key.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
ovidiul: Thanks for the suggestion, I've adjusted the items with the appropriate description.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
ovidiul: That's great, suggestion has been applied here as well. Thanks!

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
ovidiul: We can definitely remove this as it makes the screen reader easier to follow.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
ovidiul: Fixed as well.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
ovidiul: One mention is that I've added the product name in the mix as it makes the Screen Reader easier to understand on my part, so the user gets the full info of quantity, product name, and total price in one sentence.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -151,6 +159,20 @@ const OrderSummaryItem = ( { cartItem } ) => {
...
+						'Total price for %1$d %3$s items: %2$s',
+						quantity,
+						'woo-gutenberg-products-block'
+					),
+					quantity,
+					formatPrice( subtotalPrice, totalsCurrency ),
+					name
+				) }
+			</span>
 			<div className="wc-block-components-order-summary-item__total-price">
```
Aljullu: I would add `aria-hidden="true"` here, otherwise the price is read twice by screen readers.
```suggestion
			<div className="wc-block-components-order-summary-item__total-price" aria-hidden="true">
```

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
```diff
@@ -151,6 +159,20 @@ const OrderSummaryItem = ( { cartItem } ) => {
 					variation={ variation }
 				/>
 			</div>
+			<span className="screen-reader-text">
+				{ sprintf(
+					/* translators: %1$d is the number of items, %2$s is the total price including the currency symbol and %3$s is the item name. */
+					_n(
+						'Total price for %1$d %3$s item: %2$s',
```
Aljullu: Not a big deal, but wondering if we should place variables in order, so `%2$s` is the name and `%3$s` the price?

Aljullu (commented):
This is looking great, good job @ovidiul! :clap: 

I left a couple of minor comments. Once they are addressed, this should be ready to go.

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
ovidiul: Added now!

assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
ovidiul: Done as well!

Aljullu (approved):
This is working like a charm, now! Thanks for working on this @ovidiul and congrats on your first PR merged in WC Blocks! :clap:

PR #4811, opened 2021-09-22
Cart i2: Toolbar based view switcher

### LINKED ISSUE #4794: Cart i2: Update view switcher design
There was a suggestion here (pbIy4N-20c-p2#comment-3239) to update the view switcher design to be in the toolbar now. It would be good to:

- Get a better understanding of how this would look (cc @vivialice and @nerrad).
- Implement it in Cart i2.

mikejolley:
Implements an updated View Switcher as a toolbar item when viewing the cart block, and filled cart/empty cart inner blocks. These components are shared using context.

This has not had design input, but using the toolbar was a suggestion in p2. I've used a dropdown toolbar item to make it align with other controls. I think this should be ok to help us make progress until it's had a full design review because it's behind the experimental flag.

Part of #4793
Fixes #4794

### Screenshots

Screenshot 2021-09-22 at 14 14 16

1. Insert cart i2
2. Select either the top-level cart i2 block, or the filled/empty inner block
3. Try to switch view
4. View should change. Switcher should reflect the updated state.


senadir (commented):
This PR checks out.
I was going to suggest using `__experimentalCaptureToolbars` to absorb `ViewSwitcherComponent` in the main cart block but that didn't work for me.

senadir (approved):

PR #4814, opened 2021-09-23
Fix: Filter button is not aligned with the input field

### LINKED ISSUE #2560: Filter by Attribute: 'Go' button is not aligned with the input field
**To Reproduce**
1. Add a _Filter by Attribute_ block and select 'Display Style: Dropdown' and check 'Filter button'.
2. In the frontend with a wide viewport, notice the right side of the input field and the `Go` button are not aligned.

**Screenshots**
imatge

dinhtungdu:
Fixes #2560

This PR fixes the alignment issue of the Filter Products by Attribute block when using Dropdown as the display type. This PR also fixes the height of the input field, makes it equal to the height of the filter button.

### Manual Testing

1. Add the Filter Products by Attribute block and select 'Display Style: Dropdown' and check 'Filter button'.
2. See the filter button aligned correctly on both editor and front end.

### Changelog

> Fixed: Filter button from Filter Products by Attribute block is not aligned with the input field

### DIFF
--- a/assets/js/blocks/attribute-filter/style.scss
+++ b/assets/js/blocks/attribute-filter/style.scss
@@ -1,7 +1,10 @@
 .wc-block-attribute-filter {
 	margin-bottom: $gap-large;
-	display: flex;
-	flex-wrap: wrap;
+
+	&.style-dropdown {
+		display: flex;
+		gap: $gap;
+	}
 
 	.wc-block-attribute-filter-list {
 		margin: 0;
@@ -19,6 +22,11 @@
 				display: inline-block;
 			}
 		}
+	}
+
+	.wc-block-attribute-filter-dropdown {
+		flex-grow: 1;
+		max-width: unset;
 	}
 
 	.is-single .wc-block-attribute-filter-list-count,
--- a/assets/js/blocks/attribute-filter/style.scss
+++ b/assets/js/blocks/attribute-filter/style.scss
@@ -1,8 +1,11 @@
 .wc-block-attribute-filter {
 	margin-bottom: $gap-large;
+	display: flex;
+	flex-wrap: wrap;
 
 	.wc-block-attribute-filter-list {
 		margin: 0;
+		width: 100%;
 
 		li {
 			text-decoration: underline;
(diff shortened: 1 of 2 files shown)


Aljullu (commented):
This is looking good, thanks for working on this @dinhtungdu! There are a couple of edge cases that I think might be worth addressing in this PR if that sounds good.

If I add the block directly into a page so it occupies the entire width, would it be possible to make the input field to span almost to touch the button (only leaving a small margin between them)?

| Current | Desired |
| --- | --- |
|  |   |

Another case is when adding the block inside the Columns block (with three or more columns), then it's still breaking into two lines:

| Current | Desired |
| --- | --- |
|  |  |

dinhtungdu:
@Aljullu Thanks for the head-up! I updated the PR to cover the cases you mentioned above. Please take another look!

Aljullu (commented):
Thanks for updating this PR @dinhtungdu! I found one last edge case:

If you add the Filter block inside a three-column block, it overflows the available space (you can see it if you make its column to have a colored background):
imatge
I spent some time trying to figure out what's going on here. It looks like [`min-width: 0`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/components/dropdown-selector/style.scss#L67) is [ignored in inputs if their `size` attribute forces them to be wider](https://stackoverflow.com/a/29990524). I wonder if we could set the input `size` to be something very low and then set the `min-width` to 100%. I guess this way it would not overflow?

dinhtungdu:
@Aljullu thanks for the catch! I fixed the issue and tested it on Chrome/Firefox/Safari. Please take another look!

assets/js/blocks/attribute-filter/style.scss
```diff
@@ -18,12 +24,18 @@
 		}
 	}
 
+	.wc-block-attribute-filter-dropdown {
+		flex-grow: 1;
+		max-width: unset;
+		width: 0;
```
Aljullu, line 30: Ah, much better and simpler fix than the one I suggested. Good catch! :ok_hand:

Aljullu (approved):
Thanks for the update! This is ready to merge now. :shipit:

PR #4817, opened 2021-09-23
Cart i2: view switcher shared to children

### LINKED ISSUE #4812: Cart i2: Try using `__experimentalExposeToChildren` to share view switcher component with all child blocks
Context: p1632321579045000-slack-C45SNKV4Z

We currently have a view switcher toolbar button at the top 3 components, Cart, Filled Cart, and Empty Cart. Those blocks are never directly selected because they're just innerBlocks containers, meaning people would always select an innermost block, that doens't have view switcher exposed to it, making the process to discover view switcher hard to impossible to find unless you know what you're looking for.

We should try exposing view switcher button to all inner blocks, using `__experimentalExposeToChildren` and `__experimentalShareWithChildBlocks`.

If we can't make that work, we should look into using view switcher on all blocks, if that's not possible, move it to a permanently visible place like it was in Cart i1.

mikejolley:
Implements `__experimentalExposeToChildren` so that the view switcher is passed down to all children toolbars. When switching, the view is changed and the parent is selected to avoid the selection of a non-visible block.

Fixes #4812

NOTE: Only the top 2 levels of inner blocks will show the toolbar item, unless you first implement https://github.com/WordPress/gutenberg/pull/35078 This makes it so all levels show the toolbar item.

1. Insert Cart i2 Block
2. Toggle between empty/full cart views
3. Select inner blocks. Confirm toolbar item is visible.
4. Toggle between empty/full cart views. Parent block should be selected after a switch.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart-i2/edit.tsx
+++ b/assets/js/blocks/cart-checkout/cart-i2/edit.tsx
@@ -152,15 +152,15 @@ export const Edit = ( {
 		clientId,
 		[
 			{
-				view: 'emptyCart',
-				label: __( 'Empty Cart', 'woo-gutenberg-products-block' ),
-				icon: <Icon srcElement={ removeCart } />,
-			},
-			{
 				view: 'filledCart',
 				label: __( 'Filled Cart', 'woo-gutenberg-products-block' ),
 				icon: <Icon srcElement={ filledCart } />,
 				default: true,
+			},
+			{
+				view: 'emptyCart',
+				label: __( 'Empty Cart', 'woo-gutenberg-products-block' ),
+				icon: <Icon srcElement={ removeCart } />,
 			},
 		]
 	);
--- a/assets/js/blocks/cart-checkout/cart-i2/edit.tsx
+++ b/assets/js/blocks/cart-checkout/cart-i2/edit.tsx
@@ -152,12 +152,12 @@ export const Edit = ( {
 		clientId,
 		[
 			{
-				view: 'filledCart',
+				view: 'woocommerce/filled-cart-block',
 				label: __( 'Filled Cart', 'woo-gutenberg-products-block' ),
 				icon: <Icon srcElement={ filledCart } />,
 			},
 			{
-				view: 'emptyCart',
+				view: 'woocommerce/empty-cart-block',
 				label: __( 'Empty Cart', 'woo-gutenberg-products-block' ),
 				icon: <Icon srcElement={ removeCart } />,
 			},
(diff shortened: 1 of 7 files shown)


assets/js/blocks/cart-checkout/cart-i2/edit.tsx
```diff
@@ -140,24 +141,29 @@ export const Edit = ( {
...
+				label: __( 'Empty Cart', 'woo-gutenberg-products-block' ),
+				icon: <Icon srcElement={ removeCart } />,
+			},
+			{
+				view: 'filledCart',
+				label: __( 'Filled Cart', 'woo-gutenberg-products-block' ),
+				icon: <Icon srcElement={ filledCart } />,
+				default: true,
+			},
+		]
```
senadir: Nitpick, given this is the default option, can it be the first one? When opening the switcher, the first item is focused, It makes it confusing because `emptyCart` is focused and looks like it's selected.

assets/js/blocks/cart-checkout/cart-i2/use-view-switcher.tsx
```diff
@@ -23,30 +24,26 @@ export const useViewSwitcher = (
...
+				text={ currentView.label }
+				icon={
+					<Icon srcElement={ eye } style={ { marginRight: '8px' } } />
+				}
+				controls={ views.map( ( view ) => ( {
+					...view,
+					title: view.label,
+					onClick: () => {
+						setCurrentView( view );
+						selectBlock( clientId );
```
senadir: Can we not select the parent Cart block and just select the current one? So, filled or empty cart block?

senadir (approved):
Does what it says

assets/js/blocks/cart-checkout/cart-i2/use-view-switcher.tsx
mikejolley: How will we get the client ID?

senadir: `selectBlock( getBlock( clientId ).innerBlocks.find(block => block.name === 'woocommerce/filled-cart-block').clientId )`
But like prettier and not a single line.

senadir:
Do you want me to pick up this issue @mikejolley ?

mikejolley:
@senadir nah Ill wrap it up soon.

mikejolley:
@senadir do you want a final look? I made those changes and simplified the API a bit so we map view -> block Name. Makes selection easier too.

senadir (approved):
Works like a charm, thanks for working on this Mike.

Issue #4821, opened 2021-09-23
Cart i2: compare Cart i2 design to Cart i1 and add missing pieces


senadir:
Currently, Cart i2 is missing a couple of titles like cart items and cart totals. There might be other items missing as well and we should add them.

senadir:
The only differences we have are:

[- Missing Cart title, which we agreed to remove.](pbIy4N-24t-p2#comment-3584)
- Some weird spacing, which is logged https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4960

We can close this now.

PR #4823, opened 2021-09-23
Add Cart Totals to cart i2

senadir:
Part of #4821
This adds `Cart Totals` to the correct place.
It also moves a useEffect to the correct place.

### How to test
- Add cart i2.
- Go to frontend.
- You should see `Cart Totals`
- The use effect was missed as part of the refactor, you can try triggering by entering a wrong postcode in shipping calculator.


mikejolley (approved):
I didn't notice it was missing before, but it's there now :)

PR #4826, opened 2021-09-23
Improve mocking of useStoreCart

ralucaStan:
This is a missing commit mentioned in this comment https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4776#discussion_r714857353

### DIFF
--- a/assets/js/blocks/cart-checkout/payment-methods/test/payment-methods.js
+++ b/assets/js/blocks/cart-checkout/payment-methods/test/payment-methods.js
@@ -2,9 +2,6 @@
  * External dependencies
  */
 import { render, screen, fireEvent, waitFor } from '@testing-library/react';
-import { previewCart } from '@woocommerce/resource-previews';
-import { dispatch } from '@wordpress/data';
-import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
 import {
 	registerPaymentMethod,
 	__experimentalDeRegisterPaymentMethod,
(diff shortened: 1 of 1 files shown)


assets/js/blocks/cart-checkout/payment-methods/test/payment-methods.js
```diff
@@ -14,11 +11,20 @@ import {
...
+/**
+ * Internal dependencies
+ */
+import * as useStoreCartHook from '../../../../base/context/hooks/cart/use-store-cart';
+
+// Somewhere in your test case or test suite
+useStoreCartHook.useStoreCart = jest
+	.fn()
+	.mockReturnValue( useStoreCartHook.defaultCartData );
+
```
senadir, line 23: Can we still know if this was called or not? I'm asking because I wanted to copy the code to here but we really need to know if `wc/store/cart` was called more than once, which would be a breaking test.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/6da64165025e7a2afc1782e4b278d72536e7b754/assets/js/blocks/cart-checkout/cart/test/block.js#L54

Is it important here to know if `wc/store/cart` was called more than once?

ralucaStan: To test if this mock was called this should work.
```
	expect( useStoreCartHook.useStoreCart ).toHaveBeenCalledTimes( x );
```

> Can we still know if this was called or not?

Do you mean in the scope of this test? Because I don't think this is something this test should check. After adding cart as part of the argument this test requests `wc/store/cart` and because it was not mocked it was failing. So I just mocked useStoreCart to not make the fetch call and just return a cart object.

For the block test, you actually need to mock the fetch utility to test that an actual call was made. So I would not copy that there as well.

senadir (approved):

PR #4834, opened 2021-09-23
Add documentation for registerPaymentMethodExtensionCallbacks

### LINKED ISSUE #4623: Add documentation for the new extensibility point for filtering payment options
We need to document the extensibility point added in #4621 . The documentation should cover:
- The extensibility point’s purpose and API
- An usage example
- A note for extension developers about how to [extend the StoreAPI](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/README.md) in case extension developers need more cart  related data that is not available in the StoreAPI

ralucaStan:
This PR adds documentation for the `registerPaymentMethodExtensionCallbacks` introduced in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4621

Fixes #4623 

It also improves the type for callbacks registered by extensions in commit https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/86f74afed56c113ab94aed118421c207aa385337: the previous type was CanMakePaymentCallback, used for the payment methods's own `canMakePayment` configuration parameter, that returned true || a Promise. In order to eliminate the confusion, a new type has been created to define the extension callbacks that only return true or false.

### Changelog

> Add documentation for registerPaymentMethodExtensionCallbacks

### DIFF
--- a/docs/extensibility/filtering-payment-methods.md
+++ b/docs/extensibility/filtering-payment-methods.md
@@ -1,0 +1,75 @@
+# Filtering payment methods in the Checkout block
+
+## The problem
+
+You're an extension developer, and your extension is conditionally hiding payment gateways on the checkout step. You need to be able to hide payment gateways on the Checkout block using a front-end extensibility point.
+
+## The solution
+
+WooCommerce Blocks provides a function called `registerPaymentMethodExtensionCallbacks` which allows extensions to register callbacks for specific payment methods to determine if they can make payments.
+
+## Importing
+
+_Aliased import_
+
+```js
+import { registerPaymentMethodExtensionCallbacks } from '@woocommerce/blocks-registry';
+```
+
+_wc global_
+
+```js
+const { registerPaymentMethodExtensionCallbacks } = wc.wcBlocksRegistry;
+```
+
+## Signature
+
+| Parameter   | Description                                                                                                         | Type                                              |
+| ----------- | ------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------- |
+| `name`      | Unique string to identify your extension. Choose something that eliminates a name collision with another extension. | `string`                                          |
+| `callbacks` | An object containing callbacks registered for different payment methods                                             | Record< string, CanMakePaymentExtensionCallback > |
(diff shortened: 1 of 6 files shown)


docs/extensibility/filtering-payment-methods.md
```diff
@@ -0,0 +1,75 @@
...
+| Parameter   | Description                                                                                                         | Type                                              |
+| ----------- | ------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------- |
+| `name`      | Unique string to identify your extension. Choose something that eliminates a name collision with another extension. | `string`                                          |
+| `callbacks` | An object containing callbacks registered for different payment methods                                             | Record< string, CanMakePaymentExtensionCallback > |
+
+Read more below about [callbacks](#callbacks-registered-for-payment-methods).
+
+### Extension name collision
+
+When trying to register callbacks under an extension namespace already used with `registerPaymentMethodExtensionCallbacks`, the registration will be aborted and you will be noticed that you are not using a unique namespace.
```
opr:
```suggestion
When trying to register callbacks under an extension namespace already used with `registerPaymentMethodExtensionCallbacks`, the registration will be aborted and you will be notified that you are not using a unique namespace. This will be shown in the JavaScript console.
```

opr (approved):
Looks great Raluca, beside one typo I made a suggestion for.

I was wondering, could you maybe add a line specifying how you would know what key you have to use in the `callbacks` object? Maybe like 'This is the value of the `$name` property used when registering this payment method with the Blocks API'. Feel free to ignore this if you don't think it adds value.

ralucaStan:
> Looks great Raluca, beside one typo I made a suggestion for.
> 
> I was wondering, could you maybe add a line specifying how you would know what key you have to use in the `callbacks` object? Maybe like 'This is the value of the `$name` property used when registering this payment method with the Blocks API'. Feel free to ignore this if you don't think it adds value.

This is a really good suggestion and I've implemented it in d01b8743778e93b90e91ba0501eab51d9f52092c

PR #4843, opened 2021-09-24
Fix Terms and Conditions block toolbar block not working because of lack of edit wrapping

### LINKED ISSUE #4842: Checkout i2: Terms and Conditions toolbar mounted on wrong block
Terms and Conditions block has an issue in which the toolbar isn't correctly applied to it (it mounts somewhere else)

image

CleanShot 2021-09-24 at 09 00 49

### Steps to reproduce
- Insert Checkout
- Insert Terms and conditions block.
- Try to select it.
- You can't see the toolbar.

senadir:
Terms and Conditions blocks didn't work well (not selectable, no toolbar) because we forget to spread `useBlockProps` on it.

I audited all other inner blocks in Cart i2 and Checkout and it was the only one.

Fixes #4842 

### Testing
1. In Checkout, go to T&C
2. You can select it.
3. You can see toolbar.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/edit.tsx
+++ b/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/edit.tsx
@@ -23,10 +23,11 @@ export const Edit = ( {
 	attributes: { text: string; checkbox: boolean };
 	setAttributes: ( attributes: Record< string, unknown > ) => void;
 } ): JSX.Element => {
+	const blockProps = useBlockProps();
 	const currentText = text || termsCheckboxDefaultText;
 
 	return (
-		<>
+		<div { ...blockProps }>
 			<InspectorControls>
 				<PanelBody
 					title={ __(
@@ -100,7 +101,7 @@ export const Edit = ( {
 					</p>
 				</Notice>
 			) }
-		</>
+		</div>
 	);
 };
 


ralucaStan:
@senadir After following the steps I can see the toolbar appearing.
There is a difference however and what I see depending on what element I select.
1. When clicking on the text `By proceeding...` I don't see the blue border around the block.
Screenshot 2021-09-24 at 15 48 59
1. When clicking on the text `Ensure you add...` the blue border appears.
Screenshot 2021-09-24 at 15 49 04

I don't know what the expected behavior is, but I think the border should also appear for case 1.

I am ok with fixing this in another PR

senadir:
That's the expected behavior, as in 1, you're focusing on the input field, and in 2, you're focusing on the whole block. The blue outline is the focus outline.

ralucaStan (approved):

ralucaStan:
I see in the Contact information section that a smaller border appears when selecting the text, and a bigger one when selecting the block. This is the behavior for other inner blocks. Would this be possible to achieve here as well? 
Screenshot 2021-09-24 at 15 59 43

senadir:
Good catch @ralucaStan, it could be the component we're using. I will take a look.

senadir:
There doens't seem to be much we can do here, or if we should do anything. Contact information and other blocks uses `PlainText` component that is just a simple textarea (no support for formatting) and it gets that default 1px blue border style from Gutenberg. Terms and Conditions block uses RichText which is a div with `contenteditable="true"` so it doens't get that extra border. All of this is different from the block 2px blue border that they get on focus.

There's also a chance that we might replace `PlainText` with `RichText` in the future.

PR #4844, opened 2021-09-24
Enforce valid states from the API

mikejolley:
I came across an issue similar to https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4753 (which was fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4761) where state validation was breaking requests. In this case, I had an invalid billing address in the client state when using the shipping calculator (which only posts shipping addresses). 

Turns out this was because I had an invalid address state coming from the Rest API and then being stored in the client. We should fix this at API level and only return the address state if valid.

Tricky to test this one. You need to have somehow had an invalid state stored in your session. Probably need to just smoke test the customer responses from the API and ensure shipping and billing states are still returned.

### Changelog

> Store API - Ensure returned customer address state is valid.


ralucaStan (approved):
Things look good. I've tested and things looked as expected

PR #4852, opened 2021-09-28
Mini Cart block: improve button accessibility

### LINKED ISSUE #4704: Mini Cart block: improve button accessibility
Issue created from this comment by @opr:

> we should add some text to make the [Mini Cart] button more descriptive to screen readers. Right now my screen reader just says "9 items" with no further context about what the button is for.

I took some time to research what other websites are doing:
* Amazon: "1 item in shopping basket"
* Apple: "Shopping Bag"
* Asos: "Cart 1 item"
* eBay: "Your shopping cart contains 1 item"
* Glossier: "Cart Toggle"
* WayFair: "1 item in Cart"
* WooCommerce (Storefront mini cart): "10,00€ 1 item. View your shopping cart" (`View your shopping cart` comes from the [link title](https://github.com/woocommerce/storefront/blob/7074f6e063bb9cc285d08832a95377b7d2c0369e/inc/woocommerce/storefront-woocommerce-template-functions.php#L91)).

I think we can exclude anything like "View your shopping cart" or "Click to get more details", because screen readers will already announce that's a button, so it should be obvious. However, I think it's good to read the subtotal. So I would keep it like something like this:

"X items in cart, total price of Y"

dinhtungdu:
This PR improves the Mini Cart block button a11y by adding `aria-label` used by screen readers. The label format is "X items in cart, total price of Y".

This PR also updates the `formatPrice` function, makes the second argument optional. I added unit tests to cover that change too.

Fixes #4704 

#### Accessibility

### Screenshots

Screen Shot 2021-09-28 at 09 39 59

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Turn on a screen reader.
2. Add Mini cart block to a page.
3. View the page on the front end, pre `Tab` to navigate to the Mini Cart button.
4. Heard/see the label from the screen reader.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart/block.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart/block.tsx
@@ -13,6 +13,7 @@ import {
 	formatPrice,
 	getCurrencyFromPriceResponse,
 } from '@woocommerce/price-format';
+import { getSetting } from '@woocommerce/settings';
 
 /**
  * Internal dependencies
@@ -82,6 +83,11 @@ const MiniCartBlock = ( {
 		}
 	}, [ isOpen, cartIsLoading, cartItems.length, emptyCartRef ] );
 
+	const subTotal = getSetting( 'displayCartPricesIncludingTax', false )
+		? parseInt( cartTotals.total_items, 10 ) +
+		  parseInt( cartTotals.total_items_tax, 10 )
+		: cartTotals.total_items;
+
 	const ariaLabel = sprintf(
 		/* translators: %1$d is the number of products in the cart. %2$s is the cart total */
 		_n(
@@ -91,10 +97,7 @@ const MiniCartBlock = ( {
 			'woo-gutenberg-products-block'
 		),
 		cartItemsCount,
-		formatPrice(
-			cartTotals.total_price,
-			getCurrencyFromPriceResponse( cartTotals )
-		)
+		formatPrice( subTotal, getCurrencyFromPriceResponse( cartTotals ) )
 	);
 
 	const contents =
--- a/assets/js/blocks/cart-checkout/mini-cart/block.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart/block.tsx
@@ -9,7 +9,10 @@ import { translateJQueryEventToNative } from '@woocommerce/base-utils';
 import { useStoreCart } from '@woocommerce/base-context/hooks';
 import Drawer from '@woocommerce/base-components/drawer';
 import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
-import { formatPrice } from '@woocommerce/price-format';
+import {
+	formatPrice,
+	getCurrencyFromPriceResponse,
+} from '@woocommerce/price-format';
 
 /**
  * Internal dependencies
@@ -88,7 +91,10 @@ const MiniCartBlock = ( {
 			'woo-gutenberg-products-block'
 		),
 		cartItemsCount,
-		formatPrice( cartTotals.total_items )
+		formatPrice(
+			cartTotals.total_price,
+			getCurrencyFromPriceResponse( cartTotals )
+		)
 	);
 
 	const contents =
--- a/assets/js/blocks/cart-checkout/mini-cart/block.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart/block.tsx
@@ -9,6 +9,7 @@ import { translateJQueryEventToNative } from '@woocommerce/base-utils';
 import { useStoreCart } from '@woocommerce/base-context/hooks';
 import Drawer from '@woocommerce/base-components/drawer';
 import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
+import { formatPrice } from '@woocommerce/price-format';
 
 /**
  * Internal dependencies
@@ -23,7 +24,12 @@ interface MiniCartBlockProps {
 const MiniCartBlock = ( {
 	isPlaceholderOpen = false,
 }: MiniCartBlockProps ): JSX.Element => {
-	const { cartItems, cartItemsCount, cartIsLoading } = useStoreCart();
+	const {
+		cartItems,
+		cartItemsCount,
+		cartIsLoading,
+		cartTotals,
+	} = useStoreCart();
 	const [ isOpen, setIsOpen ] = useState< boolean >( isPlaceholderOpen );
 	const emptyCartRef = useRef< HTMLDivElement | null >( null );
 	// We already rendered the HTML drawer placeholder, so we want to skip the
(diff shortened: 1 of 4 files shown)


assets/js/blocks/cart-checkout/mini-cart/block.tsx
```diff
@@ -73,6 +79,18 @@ const MiniCartBlock = ( {
...
+	const ariaLabel = sprintf(
+		/* translators: %1$d is the number of products in the cart. %2$s is the cart total */
+		_n(
+			'%1$d item in cart, total price of %2$s',
+			'%1$d items in cart, total price of %2$s',
+			cartItemsCount,
+			'woo-gutenberg-products-block'
+		),
+		cartItemsCount,
+		formatPrice( cartTotals.total_items )
```
Aljullu: If you set your settings as follows:

1. In WooCommerce > Settings, check this option: 'Enable tax rates and calculations'.
2. In the Taxes tab, select 'No, I will enter prices exclusive of tax' and set 'Display prices during cart and checkout' to 'Including tax'.

You will notice taxes are not included in the accessibility text (but they are in the Storefront legacy mini cart):

prices-mini-cart
_(Storefront mini cart on the left / code of the Mini Cart block on the right)_

I think we should mimic the Storefront mini cart here, so I would use `cartTotals.total_price` instead of `total_items`.

Besides that, why not passing the currency data here? It's included in the `cartTotals` object, so I think it's better if we pass it. I guess you could do something like this:

```suggestion
		formatPrice( cartTotals.total_price, getCurrencyFromPriceResponse( cartTotals ) )
```

packages/prices/utils/price.ts
```diff
@@ -120,7 +120,7 @@ export const getCurrency = (
 export const formatPrice = (
 	// Price in minor unit, e.g. cents.
 	price: number | string,
-	currencyData: Currency
+	currencyData?: Currency
```
Aljullu, line 123: Good catch with this. I noticed there were several other instances where we call `formatPrice` without the second argument.

packages/prices/utils/test/price.js
```diff
@@ -25,4 +25,20 @@ describe( 'formatPrice', () => {
...
+		${ null }      | ${ '' }
+		${ undefined } | ${ '' }
+	`(
+		'correctly formats price given "$value" only',
+		( { value, expected } ) => {
+			const formattedPrice = formatPrice( value );
+
+			expect( formattedPrice ).toEqual( expected );
+		}
+	);
```
Aljullu, line 43: Thanks for adding tests! :clap:

src/BlockTypes/MiniCart.php
```diff
@@ -192,6 +192,7 @@ protected function get_markup() {
 		$cart                = $cart_controller->get_cart_instance();
 		$cart_contents_count = $cart->get_cart_contents_count();
 		$cart_contents       = $cart->get_cart();
+		$cart_contents_total = $cart->get_cart_contents_total();
```
Aljullu: This has the same issue as the React component, if you set the settings as described in my previous message, this price doesn't include the taxes, but we want it to.

Aljullu (commented):
This is mostly testing well, there is only one edge-case with a specific WC settings combination that should be fixed. I described it inline below. :point_down:

assets/js/blocks/cart-checkout/mini-cart/block.tsx
dinhtungdu: I didn't notice we have `getCurrencyFromPriceResponse`, thanks for the head-up!

src/BlockTypes/MiniCart.php
dinhtungdu: I updated this to use `get_cart_subtotal` as Storefront is using it: https://github.com/woocommerce/storefront/blob/3.9.1/inc/woocommerce/storefront-woocommerce-template-functions.php#L93

Aljullu: I don't think we can use the same function as Storefront here, unfortunately. `get_cart_subtotal()` returns some HTML code and I think we should avoid this because then we can't mimic the same markup easily from the React component.

I saw you used `wp_strip_all_tags()`. That would work, but instead I wonder if we could use: `$cart->get_total( 'edit' );`. This is how it's done in the API, so we will make sure the value we display in the React component and in the PHP button is the same.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a09e165253b4f531864eb48705adb84ea73cdfb4/src/StoreApi/Schemas/CartSchema.php#L372-L373

src/BlockTypes/MiniCart.php
```diff
@@ -221,7 +233,7 @@ protected function get_markup() {
 		}
```
Aljullu, line 243: Heads-up that you will also need to update the button that is shown when the Mini Cart is disabled. :point_up:

Aljullu (commented):
Thanks for updating this PR, @dinhtungdu! There is one last thing: in the PHP rendering, the price appears to be 0,00 in my case:

imatge

I investigated this a bit and I think if we use the same function as in the rest API, that should be fixed. I commented it below. :point_down:

src/BlockTypes/MiniCart.php
dinhtungdu: @Aljullu It's actually my mistake, it should be `get_subtotal()` instead of `get_cart_subtotal()`. The purpose of using `wp_strip_all_tags()` is to strip tag generated from `wc_price()`, I don't want to mimic the function of `wc_price`. Is there another way to format the price without HTML tags in WC?

dinhtungdu:
@Aljullu In 8f3c292 I switched to use subtotal again. We need to take `woocommerce_tax_display_cart` option into account here so the store owner can choose how the price is displayed in the mini cart block. The `total` price includes discount and shipping so I don't think we should use it.

Working on this I found a mismatch between real and preview cart data. 

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8f3c29265fae50eeb43f933b261ba6cdd8fca344/assets/js/previews/cart.ts#L200-L202

The issue here is: we change the preview data based on the value of `woocommerce_tax_display_cart`, so the behavior of the preview cart in the editor and the real one is a little bit different. It has little to no difference to end users though. But the `woocommerce_tax_display_cart` should only affect how data is displayed, not the actual data itself.

src/BlockTypes/MiniCart.php
Aljullu: > Is there another way to format the price without HTML tags in WC?

Good point. I couldn't find any. I see WC Admin is using a similar solution:

https://github.com/woocommerce/woocommerce-admin/blob/206384abda75a823c1e85c679c86e4a92b26485a/src/Notes/NewSalesRecord.php#L95

I think we can leave it as-is and open an issue in [WC core repo](https://github.com/woocommerce/woocommerce/issues) explaining that we needed to do this hackish solution and WC Admin is doing it too. This way they can evaluate if it would be worth adding something like that in WC.

Aljullu (approved):
> @Aljullu In 8f3c292 I switched to use subtotal again. We need to take `woocommerce_tax_display_cart` option into account here so the store owner can choose how the price is displayed in the mini cart block. The `total` price includes discount and shipping so I don't think we should use it.

Makes sense. Thanks for catching that! Approving given that this is already working great, but before merging, can you add the `aria-label` to the disabled button as well?

Here:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/158b5bc5f5324f6e8f8b37d33d3b116b064ec172/src/BlockTypes/MiniCart.php#L217-L221

> Working on this I found a mismatch between real and preview cart data.

Ah, good catch! Can you open an issue for this? I don't think it's a high priority issue but better to have it logged. Probably it can be labelled as `good first issue` for new folks onboarding?

PR #4871, opened 2021-09-29
Display the notices registered by the payment gateway's process_payment() method

### LINKED ISSUE #4870: Add specific error message during "card declined" errors when using the blocks-based checkout
## Describe the bug
A generic error message is shown in the checkout block when a card is declined, instead of the actual error message the payment gateway returns.

## To reproduce

1. Make sure you're using the Blocks checkout and WCPay.
2. Add 1 or more products to your cart.
3. Go to the blocks checkout page
4. Pay **through the credit card form** with a card that will be declined, e.g. `4000 0000 0000 0002`.
5. Observe the generic error message.

image
image

## Expected behavior
When my card is declined I expect that I'll get an error message telling me the card was declined.

### WordPress:
* Core version: 5.8.1
* Gutenberg version (if installed):
* WooCommerce version: 5.7.1
* WooCommerce Blocks version: 5.9.0
* Site language: EN

### Desktop:
* OS: macOS
* Browser: chrome
* Version: 93.0.4577.82

## Additional context
Originally posted by @reykjalin in [#1529 (review)](https://github.com/Automattic/woocommerce-payments/pull/1529#pullrequestreview-647456481)
[Original issue](https://github.com/Automattic/woocommerce-payments/issues/1706) by @reykjalin copied from the WooCommerce Payments repo to this one

a-danae:
Call `NoticeHandler::convert_notices_to_exceptions()` after `$payment_method_object->process_payment()` when processing a payment.

This way the notices added by this method are actually displayed. We were displaying a generic error and clearing the notices before. 

Fixes #4870 

#### Accessibility

### Screenshots

Error message before:
before

Error message after:
after

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests

### Manual Testing

1. Make sure you're using the Blocks checkout and WCPay.
2. Add 1 or more products to your cart.
3. Go to the blocks checkout page
4. Pay through the credit card form with a card that will be declined, e.g. 4000 0000 0000 0002.
5. Now the error should say "Error: Card declined" instead of the generic message mentioned in the issue.

### Changelog

> Fix: Payment gateway's error messages not being displayed during checkout.

### DIFF
--- a/src/Payments/Api.php
+++ b/src/Payments/Api.php
@@ -164,9 +164,8 @@ class Api {
 		// Restore $_POST data.
 		$_POST = $post_data;
 
-		// If `process_payment` added notices, clear them. Notices are not displayed from the API -- payment should fail,
-		// and a generic notice will be shown instead if payment failed.
-		wc_clear_notices();
+		// Display the notices added by `process_payment` and abort.
+		NoticeHandler::convert_notices_to_exceptions( 'woocommerce_rest_payment_error' );
 
 		// Handle result.
 		$result->set_status( isset( $gateway_result['result'] ) && 'success' === $gateway_result['result'] ? 'success' : 'failure' );


tjcafferkey (approved):
This LGTM and works as described. Also couldn't find any regressions when testing 👍🏻

leonardola:
Also related to [this issue](https://github.com/Automattic/woocommerce-payments/issues/1127)

leonardola:
@a-danae and @tjcafferkey I noticed that the the line [159](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/6c8f68158f63a0cea34becff3b27f6c28f783211/src/Payments/Api.php#L159) of the changed file is the same as the line [168](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/6c8f68158f63a0cea34becff3b27f6c28f783211/src/Payments/Api.php#L168) maybe we could remove the line 159 and only keep the line 168?

a-danae:
[Related issue](https://github.com/Automattic/woocommerce-payments/issues/2803). Basically, any error coming from WCPay's non-UPE payment methods that's not being displayed in the blocks checkout.

mikejolley:
If a legacy gateway (or any extension for that matter) creates a notice, but payment status is successful, the client would have no knowledge of the payment status which could result in multiple payments or duplicate orders.

We have no way to know that an error notice is coming from a gateway. All notices share the same system. 

This approach was also raised in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4700 and was rejected there. Given the inline comment, I think the generic error fallback was there to ensure a valid payment response was received, rather than an error response.

a-danae:
Gotcha. Checking out another approach then. Thanks for the feedback and for providing more context, @mikejolley

PR #4872, opened 2021-09-30
Add basic E2E tests for products grid blocks

dinhtungdu:
### Note

Currently, not every block is covered by e2e tests. This PR aims to add e2e tests for the On Sale Products, Products by Category, Products by Tag, Products by Attribute blocks.

Part of #4643

### Testing

1. Check out this PR.
2. Start Docker Desktop.
3. Run `npm run wp-env start && npm run test:e2e`.
4. The e2e test suite shows no failed tests.


dinhtungdu:
Note that #4785 is failing because of the same issue as this PR. It looks like the cart or checkout page isn't always created.

Update: Although the test is passing, the inconsistent result issue still remains. The test passed after the third attempt, check the result of previous (failed) attempts [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/actions/runs/1303550958/attempts/2) and [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/actions/runs/1303550958/attempts/1).

dinhtungdu:
Marking this PR is ready for review because all of the tests are passed and the flaky test issue (#4916) doesn't relate to changes introduced in this PR.

tests/e2e/specs/backend/products-by-attribute.test.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+import { getAllBlocks, switchUserToAdmin } from '@wordpress/e2e-test-utils';
+
+import { visitBlockPage } from '@woocommerce/blocks-test-utils';
+
+import { insertBlockDontWaitForInsertClose } from '../../utils.js';
+
+const block = {
+	name: 'Products by Attribute',
+	slug: 'woocommerce/product-tag',
```
mikejolley: Looks like this is the wrong slug?

mikejolley (changes requested):
Thanks for your contribution @dinhtungdu 

I noticed one incorrect slug. Even though the slug is currently not used, it should be updated before merge. Once that's corrected this is good to go :shipit: 👍🏻

tests/e2e/specs/backend/products-by-attribute.test.js
dinhtungdu: @mikejolley It is. I fixed the slug. Can you take another look at this?

mikejolley (approved):
Thanks 👍🏻 Will wait for tests to pass and then merge.

PR #4877, opened 2021-09-30
Honor the `get_product_search_form` filter

Chouby:
The legacy product search widget uses the template tag `get_product_search_form()` which allows plugins to modify the form through the `get_product_search_form` filter.

The product search block uses its own markup to render the form and no alternative filter is provided to allow plugins to modify the form output.

This causes a potential backward compatibility for users migrating from widgets to blocks.

This filter is useful, for example, for multilingual plugins like Polylang and WPML to, in fine, return search results in only one language.

#### Accessibility

No impact.

#### Automated Tests

No impact.

### Performance Impact

No impact.

### DIFF
--- a/src/BlockTypes/ProductSearch.php
+++ b/src/BlockTypes/ProductSearch.php
@@ -117,11 +117,14 @@ class ProductSearch extends AbstractBlock {
 			</div>
 		';
 
-		return sprintf(
-			'<div %s><form role="search" method="get" action="%s">%s</form></div>',
-			$wrapper_attributes,
-			esc_url( home_url( '/' ) ),
-			$label_markup . $field_markup
+		return apply_filters(
+			'get_product_search_form',
+			sprintf(
+				'<div %s><form role="search" method="get" action="%s">%s</form></div>',
+				$wrapper_attributes,
+				esc_url( home_url( '/' ) ),
+				$label_markup . $field_markup
+			)
 		);
 	}
 }


mikejolley:
@Chouby Do the use-cases you mention not store current language in session? I'm curious why this needs to be done via manipulation of form markup. We need to be careful adding filters in case rendering changes and causes a back-compat break in the future.

Chouby:
In the case of multilingual plugins, modifying the html markup allows to add a hidden field with the language information. This language parameter (a simple taxonomy term in the case of Polylang) being passed as an url parameter, the underlying `WP_Query` will automatically get the results in the desired language without doing anything more than adding this additional field.

Generally speaking, all multiligual plugins I have studied determine the language from the information included in the url. If the language (and thus the content displayed) depends on the session, this would mean that a unique url could display a different content. This is something that, I believe, should be avoided, even for search results, as this makes it impossible to share a determined content through the url.

I understand that you want to avoid to add too many new filters. However in this particular case, my suggestion is not to add a new one but to apply the filter already existing for the legacy widget to the block which is supposed to replace it. My proposal is exactly to maintain the backward compatibility and satisfy users who don't understand why something working with the legacy widget doesn't just work now that they are migrating to blocks with WP 5.8.

mikejolley:
The markup that filter filters on is quite different (see https://github.com/woocommerce/woocommerce/blob/0a857088f994def1aca6f3feb3e4a4af964b225f/templates/product-searchform.php) so this is still a risky change if the consumer of this hook is relying on something in that template. 

Whilst to mitigate that we could add an entirely new filter, have you considered appending the LANG input via a small piece of JavaScript code?

mikejolley:
@nerrad Did you have any thoughts on whether or not to introduce this?

nerrad:
I appreciate the use-case, but I don't think we should be filtering the entire rendering given how much that could change the output and potentially break user implementations. Some other thoughts:

- While still not certain, with the store editing work being done, it's entirely possible that we could represent the product search form as a template part. 
- Even if it ends up not getting represented as a template part, this is editable via some attributes now and could be customized via the block editor by additional attributes in the future. The current implementation of the filter in this PR makes it really easy for extensions to clobber the implementation of those attributes leading to poor experience for the user.

Maybe a new filter could be introduced on a narrower subset of the search form. This use-case appears to only need additional hidden fields, so what if this new filter applies on an empty string for additional fields and include that in the rendered markup, what are your thoughts on that Mike? Something like:

```php
$additional_hidden_fields = apply_filters( 'woocommerce_product_search_block_additional_fields', '' );
```

PR #4891, opened 2021-10-04
Convert Title component to TypeScript

tjcafferkey:
#### Description

Converting the `<Title />` component from JS to TypeScript.

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests

### Manual Testing

Please regression test this component to ensure it is still working as expected. You can also do this by comparing the behaviour to that on the `trunk` branch.

1. Checkout the branch and check that the compiler reports no TS errors associated with the converted files.
2. Add block(s) to your page which uses the component. Such as the `All Products` block for example (the more variety the better)
2. Test no regressions with the title components behaviour have been introduced.

### Changelog

> Title: Convert to TypeScript


assets/js/atomic/blocks/product-elements/shared/config.tsx
```diff
@@ -23,12 +25,10 @@ export default {
 		html: false,
 	},
 	parent: isExperimentalBuild()
-		? null
+		? undefined
 		: [ '@woocommerce/all-products', '@woocommerce/single-product' ],
 	save,
-	deprecated: [
-		{
-			save() {},
```
tjcafferkey: **Note to reviewer:** Was this needed? If so can you explain the reason for an empty `save()` here as I removed it in my PR.

senadir:
Is this still needing review from me?

tjcafferkey:
@senadir yes please. I have a conflict to resolve but aside from that it's still OK to be reviewed AFAIK.

assets/js/atomic/blocks/product-elements/title/attributes.ts
```diff
@@ -3,7 +3,7 @@
  */
 import { isFeaturePluginBuild } from '@woocommerce/block-settings';
 
-let blockAttributes = {
+let blockAttributes: Record< string, Record< string, unknown > > = {
```
mikejolley, line 6: I see some of our newer code uses `Record< string, AttributeSource.Attribute >`. Should we do the same here?

assets/js/atomic/blocks/product-elements/title/block.tsx
```diff
@@ -19,6 +19,31 @@ import { useStoreEvents } from '@woocommerce/base-context/hooks';
  */
 import './style.scss';
 
+interface BlockProps {
```
mikejolley: These are the same as attributes right? Just noting the category-list block uses:

```
type Props = Attributes & HTMLAttributes< HTMLDivElement >;
```

which might be shorter than this interface.

assets/js/atomic/blocks/product-elements/title/edit.tsx
```diff
@@ -23,14 +28,31 @@ import Block from './block';
 import withProductSelector from '../shared/with-product-selector';
 import { BLOCK_TITLE, BLOCK_ICON } from './constants';
 
+interface Attributes {
```
mikejolley: This is repeated from `title/block.tsx` - would it be worth defining these types (or even the attributes themselves) in a separate file to be imported here?

assets/js/atomic/blocks/product-elements/title/edit.tsx
```diff
@@ -94,7 +116,7 @@ const TitleEdit = ( {
 							) }
 							colorSettings={ [
 								{
-									value: color.color,
+									value: color,
```
mikejolley, line 114: Was this a bug previously then?

mikejolley (commented):
Added some feedback. Not a TS expert here though :P

assets/js/atomic/blocks/product-elements/title/attributes.ts
tjcafferkey: This will error, `AttributeSource.Attribute` is type `string | undefined`. I'm not sure if there's already a type or a more appropriate type for attributes though?

assets/js/atomic/blocks/product-elements/title/block.tsx
tjcafferkey: Ah yeah good catch, not sure why I did it to one component and not the other 😕

assets/js/atomic/blocks/product-elements/title/edit.tsx
tjcafferkey: Yeah this was previously a bug which this task allowed TypeScript to reveal.

assets/js/atomic/blocks/product-elements/title/edit.tsx
```diff
@@ -22,6 +27,18 @@ import HeadingToolbar from '@woocommerce/editor-components/heading-toolbar';
...
 import withProductSelector from '../shared/with-product-selector';
 import { BLOCK_TITLE, BLOCK_ICON } from './constants';
+import { Attributes } from './types';
+
+interface Props {
+	color: ColorPalette.Color;
+	fontSize: {
+		size: number | undefined;
+	};
+	setFontSize: ( size: number ) => undefined;
```
mikejolley: Just a question; why `undefined` and not `void`? Not saying which is correct.

mikejolley (approved):
Changes look good 👍🏻 I tested all products block and didn't spot regressions in editor or frontend with titles.

assets/js/atomic/blocks/product-elements/title/edit.tsx
tjcafferkey: Good question, it should be `void`. I've updated it.

PR #4892, opened 2021-10-04
Add Checkout Package and Checkout i2 Documentation

### LINKED ISSUE #4840: Cart & Checkout i2: Document how to create inner blocks.
We need to document how to create and register an inner block with Cart and Checkout. Mainly how to use `registerCheckoutBlock`, the different areas available, the locking mechanism, props and values passed to the block, and how to render dynamic vs static blocks.

mikejolley:
This PR adds documentation within the block-checkout package directory, covering the package contents, components, filters, slot fills, and inner blocks added in Checkout i2. This is linked from the main docs directory also.

Please proof and approve if you're happy with what's covered.

Closes #4840


senadir (approved):
From a bird's-eye view, this is looking good. Let's merge and iterate later.

packages/checkout/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+Package components can be accessed via the `wc` global:
+
+```js
+const { ... } = wc.blocksCheckout;
+```
+
+### Aliased imports
+
+Alternatively, you can map this to external to a custom alias using the [WordPress Dependency Extraction Webpack Plugin](https://github.com/WordPress/gutenberg/tree/trunk/packages/dependency-extraction-webpack-plugin):
```
nerrad, line 26: Would the **WooCommerce** Dependency Extraction Webpack Plugin be a better fit here (automatically mapping wc globals)?

pluginslab: From my experience doing the integrations, it has been working as you suggested Darren.

PR #4893, opened 2021-10-05
Add documentation guidelines

nielslange:
## Note

At the moment, we're using different formatting for the documentation. This PR aims to establish a documentation standard so that our documentation is consistency throughout all documents.

### Testing

1. Check out this PR.
2. Look up the file `./docs/contributors/documentation-guidelines.md`.

### DIFF
--- a/docs/contributors/documentation-guidelines.md
+++ b/docs/contributors/documentation-guidelines.md
@@ -60,7 +60,7 @@ Correct H1-H6 headings must be used. Each document can only have one H1 heading.
 
 ## Use semantically correct markup
 
-The markup must be used semantically correct, e.g. a list items must be formatted as list.
+The markup used must be semantically correct, e.g. list markup must only be used to display a list.
 
 ## Use correct spelling
 
--- a/docs/contributors/documentation-guidelines.md
+++ b/docs/contributors/documentation-guidelines.md
@@ -1,0 +1,156 @@
+# Documentation Guidelines
+
+- [Documentation Guidelines](#documentation-guidelines)
+  - [Use Automattic Writing Style Guide](#use-automattic-writing-style-guide)
+  - [Use active instead of passive voice](#use-active-instead-of-passive-voice)
+  - [Use the personal pronoun “you”](#use-the-personal-pronoun-you)
+  - [Don’t use gendered pronouns](#dont-use-gendered-pronouns)
+  - [Filename must resemble the title](#filename-must-resemble-the-title)
+  - [Use correct heading hierarchy](#use-correct-heading-hierarchy)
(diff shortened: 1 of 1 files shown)


docs/contributors/documentation-guidelines.md
```diff
@@ -0,0 +1,156 @@
...
+  - [Explicitly define language for code examples](#explicitly-define-language-for-code-examples)
+  - [Use table of contents for top-level readme](#use-table-of-contents-for-top-level-readme)
+  - [Use internal links](#use-internal-links)
+  - [Sort releases descending](#sort-releases-descending)
+  - [Structure “How to” instructions](#structure-how-to-instructions)
+  - [Link references](#link-references)
+
+## Use Automattic Writing Style Guide
+
+The Automattic Writing Style Guide serves as the foundation of these documentation guidelines.
```
opr: If we can add one, a link would be good here :)

docs/contributors/documentation-guidelines.md
```diff
@@ -0,0 +1,156 @@
...
+The filename of the document must match the title.
+
+**Example**
+
+-   **Title:** Slots and Fills
+-   **Filename:** slots-and-fills.md
+
+## Use correct heading hierarchy
+
+Correct H1-H6 headings must be used. Each document can only have one H1 heading. An H3 heading can only be used within an H2 heading, an H4 heading can only be used within an H3 heading, etc. Headings must not use articles.
```
opr: > Headings must not use articles

Perhaps we could word it more like "If possible, avoid using articles in headings" sometimes it might be necessary.

docs/contributors/documentation-guidelines.md
```diff
@@ -0,0 +1,156 @@
...
+
+**Example**
+
+-   **Incorrect:** The problem
+-   **Correct:** Problem
+
+## Use semantically correct markup
+
+The markup must be used semantically correct, e.g. a list items must be formatted as list.
```
opr:
```suggestion
The markup used must be semantically correct, e.g. list markup must only be used to display a list.
```

opr (approved):
Hey Niels this looks good! Just a couple of changes I think would be good to make. Pre-approving though as the rest is great!

docs/contributors/documentation-guidelines.md
nielslange: Absolutely. I must have missed this.

docs/contributors/documentation-guidelines.md
nielslange: That makes sense. Thanks for this suggestion.

nielslange:
Thanks for your review, @opr. I've updated this PR based on your suggestions and squashed and merged it.

PR #4899, opened 2021-10-05
Improve the engines support in package.json

nerrad:
This is a better representation of the node and npm version support for Woo Blocks (more notably the NPM 7 exclusion) which can help new contributors coming into the repository.

This has no impact on actual builds for the project and is a signal for developers (and automations).


mikejolley (approved):
Just read about what this does exactly and makes sense 👍🏻

PR #4905, opened 2021-10-06
Add a label to the textarea from the editor view of the Product search block

### LINKED ISSUE #1682: Product Search (editor): missing label for <textarea> elements
imatge

`<textarea>` elements are used to enter the label and placeholder, but they are missing a label, which is required for accessibility reasons.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

ovidiul:
This PR fixes #1682  by adding a label field on top of the Label textarea edit field from the Product Search block.

The label text is announced when moving over the textarea field, so to avoid a duplicate announcement from the label tag, a combination of `className="screen-reader-text" aria-hidden="true"` was used on the label tag.

Fixes #1682 

#### Accessibility

### Screenshots

Screenshot 2021-10-06 at 14 53 51
Screenshot 2021-10-06 at 14 58 47

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. create a post and add the search block
2. start Screen reader
3. Move over search label block

### Changelog

> Improve accessibility for the editor view of the Product search block

### DIFF
--- a/assets/js/blocks/product-search/edit.js
+++ b/assets/js/blocks/product-search/edit.js
@@ -80,13 +80,25 @@ const Edit = ( {
 			</InspectorControls>
 			<div className={ classes }>
 				{ !! hasLabel && (
-					<PlainText
-						className="wc-block-product-search__label"
-						value={ label }
-						onChange={ ( value ) =>
-							setAttributes( { label: value } )
-						}
-					/>
+					<>
+						<label
+							className="screen-reader-text"
+							htmlFor="wc-block-product-search__label"
+						>
+							{ __(
+								'Search Label',
+								'woo-gutenberg-products-block'
+							) }
+						</label>
+						<PlainText
+							className="wc-block-product-search__label"
+							id="wc-block-product-search__label"
+							value={ label }
+							onChange={ ( value ) =>
+								setAttributes( { label: value } )
+							}
+						/>
+					</>
 				) }
 				<div className="wc-block-product-search__fields">
 					<TextControl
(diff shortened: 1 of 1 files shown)


assets/js/blocks/product-search/edit.js
```diff
@@ -80,13 +80,26 @@ const Edit = ( {
...
-						className="wc-block-product-search__label"
-						value={ label }
-						onChange={ ( value ) =>
-							setAttributes( { label: value } )
-						}
-					/>
+					<>
+						<label
+							className="screen-reader-text"
+							aria-hidden="true"
```
ralucaStan: I recommend against using `aria-hidden="true"`, because this removes the element from the a11y tree. With `.screen-reader-text` we visually hide the element, but it's still available for screen readers, which is the expected outcome. But using this line removes the element from the screen reader.

ralucaStan (changes requested):
Thank you @ovidiul for taking care of this. I've added a small change request. I've also edited the PR title and the changelog entry, to make them more human-friendly. Let me know what you think.

assets/js/blocks/product-search/edit.js
ovidiul: @ralucaStan agree on this change, as stated in my initial reply, it seems the Label is being read twice by the screen reader(one from the label tag, and once from the textarea linked to it), although this behaviour is not consistent across all browsers.

Change is now done.

ovidiul:
> Thank you @ovidiul for taking care of this. I've added a small change request. I've also edited the PR title and the changelog entry, to make them more human-friendly. Let me know what you think.

Thank you @ralucaStan for your help here, changes look great.

ralucaStan (approved):
👍 🚢

PR #4906, opened 2021-10-06
Convert Product SKU block to TypeScript

frontdevde:
#### Description

Converting the Product SKU component from JS to TypeScript.

**Note:** this PR builds on the changes to the shared config that are being introduced in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4891. Once said PR is merged we'll be able to set the base for this PR to `trunk`.

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests

### Manual Testing

Please regression test this component to ensure it is still working as expected. You can also do this by comparing the behavior to that on the `trunk` branch.

1. Check out the branch and check that the compiler reports no TS errors associated with the converted files.
2. Add block(s) to your page which can use the component. Such as the `Single Product` or `All Products` block for example (the more variety the better).
2. Test no regressions with the Product SKU component's behavior have been introduced.

### Changelog

> Product SKU: Convert block to TypeScript


assets/js/atomic/blocks/product-elements/sku/edit.tsx
```diff
@@ -11,7 +11,7 @@ import Block from './block';
 import withProductSelector from '../shared/with-product-selector';
 import { BLOCK_TITLE, BLOCK_ICON } from './constants';
 
-const Edit = ( { attributes } ) => {
+const Edit = ( { attributes }: Record< string, unknown > ): JSX.Element => {
```
tjcafferkey, line 14: I would perhaps create an `interface` for `Attributes` in a separate `types.ts` file (as these are also being used in `block.tsx`) to be used here. 

And then in `block.tsx` create an interface that combines `Attributes` and `HTMLAttributes< HTMLDivElement >` instead of defining `className` like below for your component.

```
// types.ts
export interface Attributes {
	productId: number;
}
```

```
// edit.tsx
interface Props {
	attributes: Attributes;
}
```

```
// block.tsx
type Props = Attributes & HTMLAttributes< HTMLDivElement >;
```

PR #4907, opened 2021-10-06
Fix button style of Mini Cart block

### LINKED ISSUE #4679: Mini Cart block: update button style
Currently, the Mini Cart block is using a generic button displaying the number of items in the cart:

imatge

However, according to the designs, it should:
1. display the cart subtotal
2. show an icon (same for mobile & desktop, even if in the designs it's different -- icon can be found in the Trello board)
3. have transparent background

imatge

When there is not enough horizontal space, it should collapse and only display the icon with an item counter on top of it:

imatge

For the item counter, we should try to use similar styles as `.wc-block-components-order-summary-item__quantity`, because those proved to work well with a variety of themes. We have two alternatives:

1. Either create a new component (ie: `QuantityBadge` or something like this), that encapsulates the styling.
2. Copy & paste the CSS.

dinhtungdu:
Fixes #4679

This PR style the mini cart button to match the design. A new component named `QuantityBadge` was added to insolate the style of the badge. On screens smaller than 768px, only the badge with the count is displayed. On bigger screens, cart total, count, and the cart icon (without the counter) are displayed.

Update [October 28th]: Working on this I found two bugs of `formatPrice` function, please see the screenshots below or check two issues: #5018 and #5019.

#### Accessibility

### Screenshots
**Editor:**

image

**Front-end - full:**

image

**Front-end - full - after lazy loading:**

image

**Front-end - empty:**

image

**Front-end - empty - after lazy loading:**

image

**Smaller screens (<768px):**

image

### Testing
1. Add a columns block.
2. Add Mini Cart Block into a column.
3. See the design changes: the total, number of items, and the cart icon as the screenshots above.
4. Change the browser width to less than `768px`, see the layout change to only cart icon with a counter badge.

### Changelog

> Fix button style of Mini Cart block for small screens.


dinhtungdu:
Note that, in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4907/commits/d5301bd41037b2ec055daf8662b7dcbad74048e2 I try to import SVG files so we can share icons between PHP and TSX. Before that, I hard-coded the content of the cart icon in two places: `quantity-badge/index.tsx` and `BlockTypes/MiniCart.php`.

dinhtungdu:
Note 2: IMO, this block tends to use inside a column, so I made the button full width. The problem with this approach is when we use the block outside a column, the cart total and count are on the left edge, and the shopping cart icon is pushed to the right.

If the elements of the button are next to each other, we don't need to make the button full width, but according to the design, we have a flex space between the shopping cart icon and the rest.

cc @Aljullu @vivialice

assets/js/blocks/cart-checkout/mini-cart/quantity-badge/index.tsx
```diff
@@ -0,0 +1,14 @@
+/**
+ * Internal dependencies
+ */
+import './style.scss';
+import Cart from './cart.svg';
+
+const QuantityBadge = ( { count }: { count: number } ): JSX.Element => (
+	<span className="quantity-badge">
```
Aljullu: If this component will only be used inside the Mini Cart block (I suppose so given that it's inside the `mini-cart` folder), we should prepend `.wc-block-mini-cart__` to the class name.

This also applies to all other class names, they should always be prefixed. That's important because WC core used generic class names like `.price` or `.sale`, and that caused all kind of conflicts with other plugins and themes in the past, so we are trying to make all class names prefixed, now. :slightly_smiling_face: 

There is some more information about how to prefix CSS class names (and other tips about how CSS is used in the WC Blocks repo) in this document:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/contributors/coding-guidelines.md#css

assets/js/blocks/cart-checkout/mini-cart/quantity-badge/cart.svg
```diff
@@ -0,0 +1,12 @@
+<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg">
```
Aljullu: Ideally, this SVG file should be stored inside `assets/js/icons/library`. You can see how to add icons there in this [docs page](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/assets/js/icons#adding-icons).

If you do so, I think you will no longer need `@svgr/webpack` beucase you will be able to import the SVG from a React component like this:

```
import { Icon, myIcon } from '@woocommerce/icons';
<Icon srcElement={ myIcon } />
```

Aljullu (commented):
I didn't take a full review (leaving this for @frontdevde :smile: ), but left a couple of comments about how we use SVG files and CSS with links to the docs. There are a couple of changes that will need to be done regarding this. Let me know if you have any questions!

Aljullu:
Ah, sorry. I missed your comments. :sweat_smile: 

> Note that, in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4907/commits/d5301bd41037b2ec055daf8662b7dcbad74048e2 I try to import SVG files so we can share icons between PHP and TSX. Before that, I hard-coded the content of the cart icon in two places: `quantity-badge/index.tsx` and `BlockTypes/MiniCart.php`.

Hmm, yes, good point about sharing the icon between JS and PHP. Do you think it would still be possible to have a wrapper inside `assets/js/icons/library` so that icon can be imported like all the others?

> Note 2: IMO, this block tends to use inside a column, so I made the button full width. The problem with this approach is when we use the block outside a column, the cart total and count are on the left edge, and the shopping cart icon is pushed to the right.

Yes, that's a good question... I would keep it full width for now, as you did. I agree with you, I expect this block will usually be used inside a column (ie: a sidebar) or inside the Navigation block, so being full width will be ok. In the worst case, this can be revisited later.

dinhtungdu:
@Aljullu Thanks for the head-up!

> Do you think it would still be possible to have a wrapper inside assets/js/icons/library so that icon can be imported like all the others?

I don't think we can do so because we put the SVG in JS. What I'm doing in the PHP side is get the content of the SVG file and print it.

It's better to follow our convention to put the SVG in the `assets/js/icons/library`. I think we can accept a duplication on the PHP side for now and try to solve that issue if the issue repeats in the future.

assets/js/blocks/cart-checkout/mini-cart/quantity-badge/index.tsx
dinhtungdu: I updated the class and style. Because of the prefix, I moved all nested style rules to the first level.

Aljullu: Sounds good, thanks! I think there are still [a couple of classes](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4907/files#diff-3241a379978fe25150781193c3b72b9542b7f2ca751f437d3a37ffd4e569e71fR25-R26) that need to be updated inside `assets/js/blocks/cart-checkout/mini-cart/edit.tsx`.

frontdevde:
@dinhtungdu Could you please add testing instructions to the PR? Thank you!

assets/js/blocks/cart-checkout/mini-cart/block.tsx
```diff
@@ -128,16 +129,25 @@ const MiniCartBlock = ( {
...
+						_n(
+							'%d item',
+							'%d items',
+							cartItemsCount,
+							'woo-gutenberg-products-block'
+						),
+						cartItemsCount
+					) }
+				</span>
+				<QuantityBadge count={ cartItems.length } />
```
frontdevde: The `cartItemsCount` variable seems more descriptive. Is there a reason for not using it here?
```suggestion
				<QuantityBadge count={ cartItemsCount } />
```

assets/js/icons/library/cart-alt.js
```diff
@@ -0,0 +1,39 @@
+/**
+ * External dependencies
+ */
+import { SVG } from 'wordpress-components';
+
+const cartAlt = (
```
frontdevde: Can we find a more descriptive name than `cartAlt` here? Ideally, we'd want the name alone to make it immediately obvious what the icon is used for.

assets/js/blocks/cart-checkout/mini-cart/quantity-badge/style.scss
```diff
@@ -0,0 +1,30 @@
+.wc-block-mini-cart__quantity-badge {
+	position: relative;
+	display: inline-block;
+	margin-right: 10px;
+	margin-left: 10px;
+}
+
+.wc-block-mini-cart__badge {
+	align-items: center;
```
frontdevde, line 7: It'd be neat if the property order here could vaguely follow the guidelines in the [Property Ordering section](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/css/#property-ordering) of the WordPress CSS Coding Standards. No need to be overly pedantic though, just so we have some basic consistent ordering in place across all the styles you're adding in this PR.

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -4,27 +4,38 @@
...
-					/* translators: %d is the number of products in the cart. */
-					_n(
-						'%d product',
-						'%d products',
-						productCount,
-						'woo-gutenberg-products-block'
-					),
-					productCount
-				) }
+				<span className="amount">{ formatPrice( productTotal ) }</span>
```
frontdevde: I assume the classes in `edit` should probably be the same as in `block`?
```suggestion
				<span className="wc-block-mini-cart__amount">{ formatPrice( productTotal ) }</span>
```

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -4,27 +4,38 @@
...
+				<span className="count">
```
frontdevde:
```suggestion
				<span className="wc-block-mini-cart__count">
```

assets/js/blocks/cart-checkout/mini-cart/style.scss
```diff
@@ -1,3 +1,50 @@
+.wc-block-mini-cart__button {
+	background: transparent;
+	padding-left: 0;
+	padding-right: 0;
+	font-weight: 400;
+	width: 100%;
```
frontdevde: > IMO, this block tends to use inside a column, so I made the button full width. The problem with this approach is when we use the block outside a column, the cart total and count are on the left edge, and the shopping cart icon is pushed to the right.

I agree with the reasoning here to make the button full width. That said, maybe we could use a reasonable max-width to prevent the two sides of the button from being too far apart? @vivialice do you have any thoughts on how the button should behave in scenarios where there's lots of available space?
Screenshot 2021-10-08 at 10 41 08

frontdevde (commented):
Just did an initial first pass at testing the changes.

Visually I wasn't sure if the size of the icon in relation to the text is how we want it? When there's little space between the two, the icon does seem quite large in comparison.
Screenshot 2021-10-08 at 10 53 34

We might also need to address the state where the available space gets too small for the items to display side by side. Options could be to either drop contents below each other or maybe even to collapse to the mobile representation with the quantity badge (if only CSS Container Queries were a thing already to make this trivial 🙂).

Screenshot 2021-10-08 at 10 51 46

assets/js/blocks/cart-checkout/mini-cart/block.tsx
dinhtungdu: No, it's my mistake, thanks for the head-up!

dinhtungdu:
@frontdevde Thanks for the detailed review. I fixed all issues except the design-related ones. For the smaller container width, we can follow the approach Storefront is using, which is not perfect IMO (see the screenshot below). But I would rather wait for @vivialice input. Marking this PR as `need design`.

### Storefront Mini Cart in the small container:

image

assets/js/blocks/cart-checkout/mini-cart/style.scss
vivialice: How come we can't have the cart items and pricing to the left of the icon and just leave the white space to the left?

vivialice: Actually, if we are including the badge, do we need to have the item count to the left? I would just show the price if anything.

dinhtungdu: @vivialice From the engineering perspective, it's possible and easier to always display the badge and the total next to the shopping cart icon. I can see that design in someplace:

image

What is missing to me (if we decide to follow that way) is the design when we display the total including tax.

vivialice: @dinhtungdu Hey thanks for waiting on this - I've mocked up the totals with the alternate labels. I think that's all you need from me on this but let me know if you need any more input from me. I will link to the Figma file in the P2 comment 😄 
Mini Cart Totals

assets/js/blocks/cart-checkout/mini-cart/quantity-badge/style.scss
```diff
@@ -0,0 +1,30 @@
+.wc-block-mini-cart__quantity-badge {
+	align-items: center;
+	display: flex;
+	margin-left: 10px;
+	margin-right: 10px;
+}
+
+.wc-block-mini-cart__badge {
+	align-items: center;
+	background: $white;
```
Aljullu: If we are hardcoding the `background` color, we also need to hardcode the text `color` property (ie: `color: #000`). Otherwise, this won't look good in a dark theme:

imatge

You can reproduce that in Storefront going to the Customizer and setting the background color to black and the text and button color to white.

In addition to that, I think we should add `box-shadow: 0 0 0 2px #fff;`, this is [what we did in the Checkout block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/5d7fd19327f59a3707d8cc62be7df3553efd4443/assets/js/base/components/cart-checkout/order-summary/style.scss#L67) to ensure the badge border was visible in light and dark themes. It's not ideal, but the best way we have right now to make this kind of components to work in light and dark themes.

Ideally, the trolley icon should change its color too based on the font color. But I'm fine leaving it for a follow-up (ie: leaving it until we implement global styles).

assets/js/blocks/cart-checkout/mini-cart/quantity-badge/style.scss
```diff
@@ -0,0 +1,30 @@
+.wc-block-mini-cart__quantity-badge {
+	align-items: center;
+	display: flex;
+	margin-left: 10px;
+	margin-right: 10px;
```
Aljullu: Could we remove the margins here? I know they are applied to the mobile view, but it's unclear to me why they are needed. :thinking:

assets/js/blocks/cart-checkout/mini-cart/style.scss
```diff
@@ -1,3 +1,42 @@
...
+	}
+}
+
+.wc-block-mini-cart__amount {
+	display: none;
+}
+
+.wc-block-mini-cart__quantity-badge {
+	margin-left: auto;
+}
```
Aljullu: Then, we can remove this too.

assets/js/blocks/cart-checkout/mini-cart/style.scss
```diff
@@ -1,3 +1,42 @@
...
+@media screen and (min-width: 768px) {
+	.wc-block-mini-cart__amount {
+		display: initial;
+		font-weight: 600;
+		margin-right: $gap-smaller;
+	}
+
+	.wc-block-mini-cart__quantity-badge {
+		margin-right: 0;
+	}
```
Aljullu: And this. :slightly_smiling_face:

Aljullu (commented):
This is working and looking great! I left some comments regarding the CSS, but overall looking good.

> Update [October 28th]: Working on this I found two bugs of `formatPrice` function, please see the screenshots below or check two issues: #5018 and #5019.

Good catch with these issues! I added them to the Mini Cart epic, I think it would be great to fix them before releasing the Mini Cart block.

assets/js/blocks/cart-checkout/mini-cart/quantity-badge/style.scss
dinhtungdu: The reason for them and the two below is: in the previous design, we only display the counter on mobile, on desktop we hide it and show the icon only. I forgot to remove it. Thanks for the catch!

dinhtungdu:
@Aljullu Thanks for the review, I addressed your comments and add some tweaks. Can you take another look at this?

Test results:
Storefront: 
image

2021:
image

TT1 Block:
image

dinhtungdu:
Note that the unit test is failing because we remove the button text, I'm updating the testcases.

Aljullu (approved):
LGTM :ship:

PR #4908, opened 2021-10-06
Convert product elements shared config to TypeScript

frontdevde:
#### Description

Converting the product elements shared config from JS to TypeScript.

This PR simply extracts the shared config change suggested in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4891 into its own PR so it can be reviewed independently. We have a few product element PRs coming up that depend on this shared config change.

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests

### Manual Testing

Please regression test this component to ensure it is still working as expected. You can also do this by comparing the behavior to that on the `trunk` branch.

1. Check out the branch and check that the compiler reports no TS errors associated with the converted files.
2. Add block(s) to your page which uses product elements Such as the `Single Product` or `All Products` block for example (the more variety the better).
2. Test no regressions with the product elements' behavior have been introduced.

### DIFF
--- a/assets/js/atomic/blocks/product-elements/shared/config.tsx
+++ b/assets/js/atomic/blocks/product-elements/shared/config.tsx
@@ -27,7 +27,14 @@ const sharedConfig: Omit< BlockConfiguration, 'attributes' | 'title' > = {
 		? undefined
 		: [ '@woocommerce/all-products', '@woocommerce/single-product' ],
 	save,
-	deprecated: [],
+	deprecated: [
+		{
+			attributes: {},
+			save(): null {
+				return null;
+			},
+		},
+	],
 };
 
 export default sharedConfig;
--- a/assets/js/atomic/blocks/product-elements/shared/config.tsx
+++ b/assets/js/atomic/blocks/product-elements/shared/config.tsx
@@ -12,6 +12,7 @@ import save from '../save';
 
 /**
  * Holds default config for this collection of blocks.
+ * attributes and title are omitted here as these are added on an individual block level.
  */
 const sharedConfig: Omit< BlockConfiguration, 'attributes' | 'title' > = {
 	category: 'woocommerce-product-elements',
--- a/assets/js/atomic/blocks/product-elements/shared/config.tsx
+++ b/assets/js/atomic/blocks/product-elements/shared/config.tsx
@@ -4,7 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { Icon, grid } from '@woocommerce/icons';
 import { isExperimentalBuild } from '@woocommerce/block-settings';
-import { BlockConfiguration } from '@wordpress/blocks';
+import type { BlockConfiguration } from '@wordpress/blocks';
 /**
  * Internal dependencies
  */
--- a/assets/js/atomic/blocks/product-elements/shared/config.tsx
+++ b/assets/js/atomic/blocks/product-elements/shared/config.tsx
@@ -4,6 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { Icon, grid } from '@woocommerce/icons';
 import { isExperimentalBuild } from '@woocommerce/block-settings';
+import { BlockConfiguration } from '@wordpress/blocks';
 /**
  * Internal dependencies
  */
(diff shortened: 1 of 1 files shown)


assets/js/atomic/blocks/product-elements/shared/config.tsx
```diff
@@ -23,12 +24,10 @@ export default {
 		html: false,
 	},
 	parent: isExperimentalBuild()
-		? null
+		? undefined
 		: [ '@woocommerce/all-products', '@woocommerce/single-product' ],
 	save,
-	deprecated: [
-		{
-			save() {},
```
tjcafferkey, line 31: I am not experienced enough with the codebase/APIs to know if this is safe to remove. Can we get another opinion?

frontdevde, line 31: Question from [original PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4891/files#r723224221):

> **Note to reviewer:** Was this needed? If so can you explain the reason for an empty save() here as I removed it in my PR.

assets/js/atomic/blocks/product-elements/shared/config.tsx
```diff
@@ -12,7 +13,7 @@ import save from '../save';
 /**
  * Holds default config for this collection of blocks.
  */
-export default {
+const sharedConfig: Omit< BlockConfiguration, 'attributes' | 'title' > = {
```
opr, line 17: Maybe add to the docblock comment why you're omitting attributes and title here

assets/js/atomic/blocks/product-elements/shared/config.tsx
```diff
@@ -4,6 +4,7 @@
 import { __ } from '@wordpress/i18n';
 import { Icon, grid } from '@woocommerce/icons';
 import { isExperimentalBuild } from '@woocommerce/block-settings';
+import { BlockConfiguration } from '@wordpress/blocks';
```
opr: If you're just importing a type from a module then you can use the `type` keyword.

```suggestion
import type { BlockConfiguration } from '@wordpress/blocks';
```

assets/js/atomic/blocks/product-elements/shared/config.tsx
senadir: Deleting this would cause a regression if someone inserted All Products block before version 3.2.0, never resaved the page, and accessed the page.

In normal cases I would say this is safe to delete; however, an invalid block might break the inner block structure of All Products. Keeping this here doesn't seem awfully annoying to us?

frontdevde: For continuity's sake, the PR where this change was introduced is https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2607. 
A related review comment can be found [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2607#discussion_r434621502) and the `save` function change [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/2607/files#diff-9c5ee2b6a3a2bcc2832f867a73d1218f77b9feeb2ca85eb29fc8f1239fa86dedR6 ).

frontdevde: See discussion above.

assets/js/atomic/blocks/product-elements/shared/config.tsx
tjcafferkey: Done

assets/js/atomic/blocks/product-elements/shared/config.tsx
```diff
@@ -23,12 +25,17 @@ export default {
...
 	},
 	parent: isExperimentalBuild()
-		? null
+		? undefined
 		: [ '@woocommerce/all-products', '@woocommerce/single-product' ],
 	save,
 	deprecated: [
 		{
-			save() {},
+			attributes: {},
```
tjcafferkey, line 33: @senadir is there any risk of `attributes: {}` being added here? Im not familiar enough with the API and its functionality.

senadir: This worked fine for me without regressing any block.

opr (approved):
Code looks fine to me and since Nadir confirms there is no regression with the change to `deprecated` this can be merged! 🥳

PR #4923, opened 2021-10-11
Revert "Display the notices registered by the payment gateway's process_payment() method (#4871)"

senadir:
This PR reverts https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4871

The PR is previously introduced here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4700.
We originally discussed the approach there and decided not to merge it.

@mikejolley outlined in #4871 why this might be a problem

> If a legacy gateway (or any extension for that matter) creates a notice, but payment status is successful, the client would have no knowledge of the payment status which could result in multiple payments or duplicate orders.
>
> We have no way to know that an error notice is coming from a gateway. All notices share the same system.

### DIFF
--- a/src/Payments/Api.php
+++ b/src/Payments/Api.php
@@ -164,8 +164,9 @@ class Api {
 		// Restore $_POST data.
 		$_POST = $post_data;
 
-		// Display the notices added by `process_payment` and abort.
-		NoticeHandler::convert_notices_to_exceptions( 'woocommerce_rest_payment_error' );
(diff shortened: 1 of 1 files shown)


mikejolley (approved):
👍🏻 

Do we need to make this docblock more clear, since it's been removed via PR twice now :p?

opr (approved):
pca54o-2ji-p2#comment-2808

PR #4925, opened 2021-10-12
Twenty Nineteen: Price slider track fix padding/border

### LINKED ISSUE #3902: Twenty Nineteen: Price slider track is not visible
## To reproduce
1. Set your store theme to Twenty Nineteen.
2. Create a page with the All Products block and the Filter Products by Price block.
3. Notice that the track of the price slider in the Filter Products by Price block is transparent and hovering one of the thumbs hides the other.

## Screenshots
Screenshot showing the bug

## Environment
I could reproduce this with WC Blocks 4.6.0, 4.5.2 and with WC Core 5.0.0 without any version of Blocks installed.

ovidiul:
This PR fixes the visibility of the Price Slider block reported in issue #3902 against the Twenty Nineteen theme.

Fixes #3902

#### Accessibility

### Screenshots

Screenshot 2021-10-12 at 14 39 32

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Merge this PR
2. Switch there to default WordPress Twenty Nineteen
2. Create a page with the All Products block and the Filter Products by Price block.
3. Notice that the track of the price slider in the Filter Products by Price block display is fixed, by making the 0px padding/border take precedence against the default theme style values which are:

```
border: solid 1px #ccc;
padding: 0.36rem 0.66rem;
```

for Twenty Nineteen.

### Performance Impact

I did not find a way to circumvent the use of `!important` flag, as the border and padding values for the input element are taking the default values from the theme style file, so it could be the case for other themes as well.

### Changelog

> Price Slider block display fix for Twenty Nineteen Theme

### DIFF
--- a/assets/js/base/components/price-slider/style.scss
+++ b/assets/js/base/components/price-slider/style.scss
@@ -37,8 +37,8 @@
 }
 @mixin reset {
 	margin: 0;
-	padding: 0;
-	border: 0;
+	padding: 0 !important;
+	border: 0 !important;
 	outline: none;
 	background: transparent;
 	-webkit-appearance: none;
--- a/assets/js/base/components/price-slider/style.scss
+++ b/assets/js/base/components/price-slider/style.scss
@@ -37,8 +37,7 @@
 }
 @mixin reset {
 	margin: 0;
-	/* 
-	Use !important to prevent theme input styles from breaking the component. 
+	/* Use !important to prevent theme input styles from breaking the component.
 	Reference https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3902
 	*/
 	padding: 0 !important;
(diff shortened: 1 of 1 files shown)


gigitux:
LGTM!

What do you think if we add a comment to the code like this:

```
Use !important to prevent theme input styles from breaking the component. 
For example  https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3902
```

So we will remember why we added `!important`

gigitux (approved):

ovidiul:
@gigitux Makes sense from my part, comments added, thanks.

gigitux:
Thanks for adding the comment. 

I will merge this PR even if the test fails, as it’s not related.

PR #4927, opened 2021-10-12
Add hook documentation and build script

### LINKED ISSUE #4629: Write documentation around new actions introduced in the Checkout process.
With #4610, we would now have 4 new hooks in the checkout process. They have no documentation yet except a short description in `docs/blocks/feature-flags-and-experimental-interfaces.md`. Hooks are documented inline but are hard to find so we need some sort of doc page for them + some examples of how to use them.

Hook that needs documentation:
- [ ] `__experimental_woocommerce_blocks_checkout_order_processed`
- [ ] `__experimental_woocommerce_blocks_checkout_update_order_from_request`
- [ ] `__experimental_woocommerce_blocks_checkout_update_order_meta`
- [ ] `__experimental_woocommerce_blocks_payment_gateway_features_list`

mikejolley:
This PR takes care of:

- A new build script to generate hook documentation for actions and filters
   - This also supports an `@examples` tag for linking to an MD file with further code examples
   - Docs are added inside docs/extensibility/filters.md and docs/extensibility/actions.md
   - TOC generation is automatic
- Inline documentation for all filters and actions in our plugin on the PHP side
- Doc examples for some of the experimental filters where needed

Screenshot 2021-10-12 at 17 13 59

Updating docs consists of running the following command:

```bash
npm run build:docs
```

Why do this? Because duplicating content between docs and inline is time-consuming and can become out of sync. Having a script can automate this and allows us to document in one place (inline).

Closes #4629

### Testing

1. Composer install dependencies
2. Npm install dependencies
3. Delete docs/extensibility/filters.md and docs/extensibility/actions.md
4. Run `npm run build:docs`. See that docs were generated.


composer.json
```diff
@@ -38,6 +39,11 @@
 			"phpcbf -p"
 		]
 	},
+	"config": {
+		"platform": {
+			"php": "7.4.24"
```
nielslange, line 44: I'm trying to understand why the PHP version had been added, as I cannot find any reference on https://github.com/johnbillion/wp-hooks-generator that a certain PHP version is required.

May I ask why the PHP version had been added and why this particular version had been selected? Regarding the version, I'm asking as in `woocommerce-gutenberg-products-block.php` we're referring to PHP 7.0 as minimum required version.

docs/extensibility/actions.md
```diff
@@ -0,0 +1,558 @@
+<!-- DO NOT UPDATE THIS DOC DIRECTLY -->
+
+<!-- Use `npm run build:docs` to automatically build hook documentation -->
+
+# Actions
+
+## Table of Contents
+
```
nielslange, line 8: **nitpic** Not a deal-breaker, but I see plenty of double line-breaks below h2 elements. I wonder if we should reduce this to one line-break only.

docs/extensibility/filters.md
```diff
@@ -0,0 +1,831 @@
+<!-- DO NOT UPDATE THIS DOC DIRECTLY -->
+
+<!-- Use `npm run build:docs` to automatically build hook documentation -->
+
+# Filters
+
+## Table of Contents
+
```
nielslange, line 8: **nitpic** Similar to `docs/extensibility/actions.md`, I also see plenty of double line-breaks below h2 elements here.

nielslange (approved):
The PR looks great and works as expected! 

I added three minor comments, but feel free to merge this PR, @mikejolley.

composer.json
mikejolley: Something about adding this composer script made the tests fail because they were trying to use PHP8 which is incompatible with some of our packages. I added this so that when tests are run, composer would run under 7.4. As you can see, tests are passing again.

docs/extensibility/actions.md
mikejolley: This is generated by json2md - looks like a known issue ﻿﻿https://github.com/IonicaBizau/json2md/issues/54

It doesn't affect display of the markdown doc so I'm ok with it.

docs/extensibility/filters.md
mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4927#discussion_r727851860

composer.json
nielslange: I thought about something like that. Makes sense to add it then, Mike. 😄

docs/extensibility/actions.md
nielslange: > This is generated by json2md - looks like a known issue ﻿﻿IonicaBizau/json2md#54

Ah, I see. 

> It doesn't affect display of the markdown doc so I'm ok with it.

Absolutely. I just mentioned it, as I wasn't sure what was causing the double line breaks. All good. 🙌

PR #4938, opened 2021-10-14
Add Mini Cart block tests

Aljullu:
Adds a couple of unit tests to the Mini Cart block.

### Testing

Make sure tests pass in this PR.


tjcafferkey (approved):
Might be good to have some tests eventually to check it renders correctly when there are items in the cart, but can always be done later (if it's not covered already). This LGTM though 🚀

PR #4939, opened 2021-10-14
Draft PR: experiment on an initial basic version of WooCommerce FSE block templates

### LINKED ISSUE #4934: Convert Single Product PHP template into a basic PHP-rendered block template
To better understand the complexity of this endeavor, we’ll start by going through the steps outlined for the _Single Product WooCommerce template_ only. This will allow us to get a clearer picture of the tasks involved in creating more complex block templates, uncover potential gaps and enable us to give a more confident estimate on the overall scope of this project.

frontdevde:
### Notes

This is a Draft PR to experiment on an initial basic version of WooCommerce FSE block templates.

#### Woocommerce
1. On your WordPress site with Woocommerce installed. Update [this line](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-template-loader.php#L110-L112) with the following code

```
return apply_filters( 'wc_has_block_template',
    is_readable(
        get_stylesheet_directory() . '/block-templates/' . $template_name . '.html'
    )
);
```

#### Woo Blocks
1. Install and activate the Gutenberg plugin.
2. Create a build of Woo Blocks in this branch (`npm install & npm run build`);
3. Install a block theme (e.g. [Quadrat](https://github.com/Automattic/themes/tree/trunk/quadrat)). 
4. The theme will need to have `add_theme_support( 'woocommerce' )` defined in order for this to work on the frontend
5. Go to the `Site Editor` where you should find the "Single Product" template containing the placeholder block for the product information. This template content is stored in `templates/block-templates/single-product.html`
6. Navigate to a single product view on the Frontend to see the FSE block template in action. You should be able to verify this as it should say `You are using the block template now` within the page content on the frontend.
7. You can also add more blocks in the editor and save, then view the changes on the Frontend.

### Known issues

* Updating the template via the Site Editor won't save the changes, or reflect them on the frontend.


templates/block-templates/single-product.html
```diff
@@ -0,0 +1,5 @@
+<!-- wp:template-part {"slug":"header","tagName":"header","className":"site-header","layout":{"inherit":true},"theme":"theme-experiments/tt1-blocks"} /-->
```
Aljullu, line 1: I don't think we should hardcode a theme name here (`"theme":"theme-experiments/tt1-blocks"`).

src/BlockTemplateController.php
```diff
@@ -0,0 +1,99 @@
...
+		$this->init();
+		$this->template_dir = plugin_dir_path( __FILE__ ) . '../templates/' . self::TEMPLATE_BASE_PATH;
+	}
+
+	/**
+	 * Initialization method.
+	 */
+	protected function init() {
+		if ( gutenberg_supports_block_templates() ) {
+			add_action( 'init', array($this, 'add_block_templates') );
```
Aljullu, line 22: This kind of things (missing spaces) should be autoformatted by your editor based on the linting rules. We have some [docs here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/contributors/getting-started.md#linting), but let me know if you need help setting it up. :slightly_smiling_face:

Aljullu (commented):
I didn't do a full review, but overall I think this is the correct approach and it's looking great. It's very exciting to be able to edit the Single product page from the Site Editor! I tested it alongside with Product Add-Ons and... it _just works_. :slightly_smiling_face: 

I know this is not final, but just so you know, I get this error in all pages which are not single product:

> Warning: Undefined array key "single-product" in /wp-content/plugins/gutenberg/lib/full-site-editing/template-loader.php on line 143

Is that something you are aware of?

src/BlockTemplateController.php
tjcafferkey: @Aljullu yeah, my editor isn't set up to automatically do it but since it was a experimental PR with no intention of merging, I bypassed the linting when committing to this branch.

templates/block-templates/single-product.html
tjcafferkey: Yeah, you're right. This can be removed AFAIK so thanks for highlighting!

PR #4941, opened 2021-10-14
Remove duplicate data attributes

### LINKED ISSUE #4867: Checkout i2: Alignment and Custom Classnames appended as data attributes
If you set a custom className and an alignment to Checkout i2 and save, the wrapping div would have `data-align` and `data-class-name` appended to it, those two values are honored as classNames so they shouldn't be there.

### Testing Steps
- Add Checkout i2.
- Change aligment.
- Add custom attributes from the advanced settings panel for the block.
- Save the block.
- See HTML
image

Because we don't mount anything to div, its `data-block-name` isn't removed as well.

mikejolley:
We seem to have duplicate data attributes in the saved HTML markup. This is due to `inject_html_data_attributes()` code, as well as the new render filter which applies to all blocks in our namespace:

```php
add_filter( 'render_block', array( $this, 'add_data_attributes' ), 10, 2 );
```

Since `add_data_attributes` applies to all `woocommerce` blocks, we can remove `inject_html_data_attributes`.

Fixes #4867

### Testing

1. Smoke test that the core set of WooCommerce Blocks render on the frontend. I tested product, reviews, checkout, cart.
2. Save the checkout block. View page source. Confirm there are no data-class-name and data-align attributes on the main div.

### Changelog

> Remove duplicate attributes in saved block HTML

### DIFF
--- a/src/BlockTypesController.php
+++ b/src/BlockTypesController.php
@@ -101,11 +101,15 @@ final class BlockTypesController {
 		}
 
 		$attributes              = (array) $block['attrs'];
+		$exclude_attributes      = [ 'className', 'align' ];
 		$escaped_data_attributes = [
 			'data-block-name="' . esc_attr( $block['blockName'] ) . '"',
 		];
 
 		foreach ( $attributes as $key => $value ) {
+			if ( in_array( $key, $exclude_attributes, true ) ) {
+				continue;
+			}
 			if ( is_bool( $value ) ) {
 				$value = $value ? 'true' : 'false';
 			}
--- a/src/BlockTypes/AbstractBlock.php
+++ b/src/BlockTypes/AbstractBlock.php
@@ -320,39 +320,6 @@ abstract class AbstractBlock {
 	}
 
 	/**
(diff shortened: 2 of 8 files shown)


src/BlockTypesController.php
```diff
@@ -101,11 +101,15 @@ public function add_data_attributes( $content, $block ) {
 		}
 
 		$attributes              = (array) $block['attrs'];
+		$exclude_attributes      = [ 'className', 'align' ];
```
senadir, line 104: Works as expected, although a bit limiting as a start. $exclude_attributes can't be extended.

Is it possible for blocks to implement this themselves? as in:
- Ignore the block using a filter.
- Reimplement this function.

senadir (approved):
Works as expected.

src/BlockTypesController.php
mikejolley: tbh I think under our namespace we want all attributes present. These core attributes are outliers. 

Other blocks are not forced to use this handling. `render_block` filter is available to all.

PR #4943, opened 2021-10-14
Allows product grid block contents to be limited by stock level

### LINKED ISSUE #2739: Don't show out of stock single/variation products in blocks
We need an option to disable showing out of stock products, otherwise we are promoting something that we don't have in our stores at the moment. 

The best solution would be an option inside Gutenberg editor, where we can tick "don't show out of stock products"

jonny-bull:
Adds a product stock dropdown to Product Grid blocks. This allows admin users to filter products by their stock status (in stock, out of stock, on back order or any status).

If the site has the 'Hide out of stock items from the catalog' WooCommerce setting enabled, this filter does not display.

The blocks effected are:
* Products by Attribute
* Products by Tag
* Top Rated Products
* Products by Attribute
* On Sale Products
* Newest Products

This has come from a client requesting Product Grid blocks that show only products in stock or on back order, without hiding all out of stock items across the site. This change should allow admin users more control over which products display in the above blocks. For example, creating a product block of products on back order could be used to drive pre-sales or draw attention to products that are due back in stock soon.

This work could also be extendable to include low stock products, allowing for users to create blocks of products that will sell out soon.

Fixes #2739 

#### Accessibility

Interactions added in this PR are duplicates of existing interactions used on product grid blocks.

### Screenshots

https://user-images.githubusercontent.com/53259768/137306957-d2e77691-7f30-4528-914e-103c43f3d7d4.mov

Very important note: the debug string visible in this video is to give a quick and easy visual indicator of what products are in stock and out of stock. It is not included in this PR.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. On a Gutenberg page, add a new product grid block (namely, Products by Attribute, Products by Tag, Top Rated Products, Products by Attribute, On Sale Products or Newest Products)
2. In the right hand edit column, open the 'Stock level' panel
3. Change the drop down menu setting and observe the product list changing to match the new filter
4. Repeat with any variations and combinations of settings

### Performance Impact

When filtering by stock `meta_query` is used by WP_Query to return the selected products. This can product slow queries. 

### Changelog

> Added controls to product grid blocks for filtering by stock levels

### DIFF
--- a/assets/js/blocks/product-new/block.js
+++ b/assets/js/blocks/product-new/block.js
@@ -10,6 +10,7 @@ import PropTypes from 'prop-types';
 import GridContentControl from '@woocommerce/editor-components/grid-content-control';
 import GridLayoutControl from '@woocommerce/editor-components/grid-layout-control';
 import ProductCategoryControl from '@woocommerce/editor-components/product-category-control';
+import ProductStockControl from '@woocommerce/editor-components/product-stock-control';
 import { gridBlockPreview } from '@woocommerce/resource-previews';
 import { getSetting } from '@woocommerce/settings';
 
@@ -26,6 +27,7 @@ class ProductNewestBlock extends Component {
 			contentVisibility,
 			rows,
 			alignButtons,
+			stockStatus,
 		} = attributes;
 
 		return (
@@ -54,6 +56,18 @@ class ProductNewestBlock extends Component {
 						onChange={ ( value ) =>
 							setAttributes( { contentVisibility: value } )
 						}
+					/>
+				</PanelBody>
+				<PanelBody
+					title={ __(
+						'Filter by stock status',
+						'woo-gutenberg-products-block'
+					) }
+					initialOpen={ false }
+				>
+					<ProductStockControl
+						setAttributes={ setAttributes }
+						value={ stockStatus }
 					/>
 				</PanelBody>
 				<PanelBody
--- a/assets/js/blocks/products/edit.js
+++ b/assets/js/blocks/products/edit.js
@@ -2,7 +2,8 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { ToggleControl, SelectControl } from '@wordpress/components';
+import { PanelBody, ToggleControl, SelectControl } from '@wordpress/components';
+import ProductStockControl from '@woocommerce/editor-components/product-stock-control';
 
 export const getSharedContentControls = ( attributes, setAttributes ) => {
 	const { contentVisibility } = attributes;
@@ -27,48 +28,65 @@ export const getSharedContentControls = ( attributes, setAttributes ) => {
 
 export const getSharedListControls = ( attributes, setAttributes ) => {
 	return (
-		<SelectControl
-			label={ __( 'Order Products By', 'woo-gutenberg-products-block' ) }
-			value={ attributes.orderby }
-			options={ [
-				{
-					label: __(
-						'Default sorting (menu order)',
-						'woo-gutenberg-products-block'
-					),
-					value: 'menu_order',
-				},
-				{
-					label: __( 'Popularity', 'woo-gutenberg-products-block' ),
-					value: 'popularity',
-				},
-				{
-					label: __(
-						'Average rating',
-						'woo-gutenberg-products-block'
-					),
-					value: 'rating',
-				},
-				{
-					label: __( 'Latest', 'woo-gutenberg-products-block' ),
-					value: 'date',
-				},
-				{
-					label: __(
-						'Price: low to high',
-						'woo-gutenberg-products-block'
-					),
-					value: 'price',
-				},
-				{
-					label: __(
-						'Price: high to low',
-						'woo-gutenberg-products-block'
-					),
-					value: 'price-desc',
-				},
-			] }
-			onChange={ ( orderby ) => setAttributes( { orderby } ) }
-		/>
+		<div>
+			<SelectControl
+				label={ __(
+					'Order Products By',
+					'woo-gutenberg-products-block'
+				) }
+				value={ attributes.orderby }
+				options={ [
+					{
+						label: __(
+							'Default sorting (menu order)',
+							'woo-gutenberg-products-block'
+						),
+						value: 'menu_order',
+					},
+					{
+						label: __(
+							'Popularity',
+							'woo-gutenberg-products-block'
+						),
+						value: 'popularity',
+					},
+					{
+						label: __(
+							'Average rating',
+							'woo-gutenberg-products-block'
+						),
+						value: 'rating',
+					},
+					{
+						label: __( 'Latest', 'woo-gutenberg-products-block' ),
+						value: 'date',
+					},
+					{
+						label: __(
+							'Price: low to high',
+							'woo-gutenberg-products-block'
+						),
+						value: 'price',
+					},
+					{
+						label: __(
+							'Price: high to low',
+							'woo-gutenberg-products-block'
+						),
+						value: 'price-desc',
+					},
+				] }
+				onChange={ ( orderby ) => setAttributes( { orderby } ) }
+			/>
+			<PanelBody
+				title={ __( 'Stock level', 'woo-gutenberg-products-block' ) }
+				initialOpen={ false }
+			>
+				<ProductStockControl
+					setAttributes={ setAttributes }
+					value={ attributes.stockStatus }
+				/>
+			</PanelBody>
+		</div>
 	);
 };
--- a/assets/js/editor-components/product-stock-control/index.tsx
+++ b/assets/js/editor-components/product-stock-control/index.tsx
@@ -1,0 +1,85 @@
+/**
+ * External dependencies
+ */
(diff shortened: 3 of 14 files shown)


jonny-bull:
This is my first contribution to this plugin, so do point out any rookie errors or incorrect assumptions I've made here.

assets/js/blocks/product-new/block.js
```diff
@@ -56,6 +58,20 @@ class ProductNewestBlock extends Component {
 						}
 					/>
 				</PanelBody>
+				{ getSetting( 'hide_out_of_stock' ) !== 'yes' && (
```
mikejolley: One thing we could do here is populate the default state of the option using this setting, but then override the global once changed. Were there other reasons for turning this off when `hide_out_of_stock` is true aside from the conflict?

assets/js/blocks/product-new/block.js
```diff
@@ -56,6 +58,20 @@ class ProductNewestBlock extends Component {
...
+					<PanelBody
+						title={ __(
+							'Stock level',
```
mikejolley: How about `Filter by stock status` so it's more in line with the other settings here?

assets/js/blocks/products/edit.js
```diff
@@ -27,48 +28,98 @@ export const getSharedContentControls = ( attributes, setAttributes ) => {
...
+							'Price: high to low',
+							'woo-gutenberg-products-block'
+						),
+						value: 'price-desc',
+					},
+				] }
+				onChange={ ( orderby ) => setAttributes( { orderby } ) }
+			/>
+			{ getSetting( 'hide_out_of_stock' ) !== 'yes' && (
+				<SelectControl
```
mikejolley: Could this just be imported from ProductStockControl?

assets/js/editor-components/product-stock-control/index.js
```diff
@@ -0,0 +1,63 @@
...
+import PropTypes from 'prop-types';
+
+/**
+ * A pre-configured SelectControl for product stock settings.
+ *
+ * @param {Object} props Incoming props for the component.
+ * @param {string} props.value
+ * @param {function(any):any} props.setAttributes Setter for block attributes.
+ */
+const ProductStockControl = ( { value, setAttributes } ) => {
```
mikejolley: Ideally, if we're creating new components we do so with TypeScript. That means defining the props here, return type, and dropping the PropTypes object.

assets/js/editor-components/product-stock-control/index.js
```diff
@@ -0,0 +1,63 @@
...
+					label: __( 'In stock', 'woo-gutenberg-products-block' ),
+					value: 'instock',
+				},
+				{
+					label: __( 'Out of stock', 'woo-gutenberg-products-block' ),
+					value: 'outofstock',
+				},
+				{
+					label: __(
+						'On back order',
```
mikejolley: From what I can see, Woo core uses `backorder` not `back order`.

src/BlockTypes/AbstractProductGrid.php
```diff
@@ -50,6 +57,7 @@ protected function get_block_type_attributes() {
 			'align'             => $this->get_schema_align(),
 			'alignButtons'      => $this->get_schema_boolean( false ),
 			'isPreview'         => $this->get_schema_boolean( false ),
+			'stockStatus'       => $this->get_schema_string( 'all' ),
```
mikejolley: This uses `all`, but some of the defaults in this PR use `any`. Which is correct?

mikejolley (commented):
First, thanks for contributing. It's great to see! I added some inline comments/feedback for the current code.

This PR allows filtering to a specific status e.g. In stock vs On Backorder. Do you anticipate multiple statuses at once? E.g. show `instock` and `onbackorder`, but hide `outofstock`? If so we could consider using a multiselect component and array type, rather than select component with string type.

One other thing, we definitely need to refactor this to use https://github.com/woocommerce/woocommerce/blob/5cdf9799613ce00504c16c79dff5064d0ee54677/includes/wc-product-functions.php#L870 otherwise custom statuses will be ignored. Do you agree?

Thanks again. Let me know if you have questions about my comments.

jonny-bull:
@mikejolley thanks for the review. Will be able to pick this up this week

assets/js/blocks/product-new/block.js
jonny-bull: Updated

assets/js/editor-components/product-stock-control/index.js
jonny-bull: Updated here and across the PR.

src/BlockTypes/AbstractProductGrid.php
jonny-bull: Updated to use `any` in all circumstances.

assets/js/blocks/products/edit.js
jonny-bull: Have switched this to use the `<ProductStockControl>` component as requested.

assets/js/blocks/product-new/block.js
jonny-bull: These bits have been removed from all similar components.

Filtering the 'Out of stock' status at ProductStockControl level, rather than in the parent components. This means the filters can still be applied on WC stores with 'hide_out_of_stock' switched on.

assets/js/editor-components/product-stock-control/index.js
jonny-bull: Rewritten for TypeScript. Hopefully I've done it right.

jonny-bull:
@mikejolley comments have been addressed. Let me know if there's more feedback.

One thing I've noticed, and want to fix, is preventing clicking the checkboxes while the edited block is reloading with updated data. Any pointers on that hugely appreciated.

assets/js/editor-components/product-stock-control/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+		<CheckboxList
+			checked={ checkedOptions }
+			options={ displayOptions }
+			onChange={ onChange }
+			isDisabled={ isDisabled }
+		/>
+	);
+};
+
+ProductStockControl.propTypes = {
```
opr: Could we remove this since the props are defined in the `ProductStockControlProps` type?

src/BlockTypes/AbstractProductGrid.php
```diff
@@ -172,6 +181,9 @@ protected function parse_attributes( $attributes ) {
 	 * @return array
 	 */
 	protected function parse_query_args() {
+		// Store the original meta query.
+		$this->meta_query = WC()->query->get_meta_query(); // phpcs:ignore WordPress.DB.SlowDBQuery
```
opr: Can we remove the phpcs:ignore comment?

opr (commented):
Hey @jonny-bull thanks for contributing! I've tested this out and had a look through the code, it all looks good. I've left a couple of comments for you to address but it's nothing major at all.

I have spent some time looking into how we can disable the checkboxes while the block is loading and there's nothing I can see that will allow us to do this. I might be missing something though.

Despite them not being disabled, rapidly changing the options while the block loads still seems to result in the correct output.

src/BlockTypes/AbstractProductGrid.php
jonny-bull: We sure can, with no complaints from PHPCS.

assets/js/editor-components/product-stock-control/index.tsx
jonny-bull: Removed. Thanks for the tip.

jonny-bull:
> Despite them not being disabled, rapidly changing the options while the block loads still seems to result in the correct output.

@opr yeah, I tried to demo this 'bug' to a client this week, but couldn't recreate it. Usually I demo things that work and find bugs, so nice to have it go the opposite way for once.

opr:
> > Despite them not being disabled, rapidly changing the options while the block loads still seems to result in the correct output.
> 
> @opr yeah, I tried to demo this 'bug' to a client this week, but couldn't recreate it. Usually I demo things that work and find bugs, so nice to have it go the opposite way for once.

Awesome, that makes a nice change for sure. Glad it went well! 

I'll approve this PR but wait until Mike is back and get his 👍🏼  to merge it.

note: I'm running the workflows for this PR, but the ESLint action will fail because the job is running from a fork. It looks good from my end though.

opr (approved):

mikejolley (approved):

jonny-bull:
@opr @mikejolley cool. That's two approvals. I don't have access to merge this PR. Can either of you push the button?

mikejolley:
Thanks @jonny-bull, it's merged. And thanks @opr for reviewing.

jonny-bull:
@mikejolley @opr and thanks to you both for the help with my first contribution. If there's more I can do, let me know

mikejolley:
I think we missed something; this affects the All Products Block which is not a grid block. It's caused a little breakage so I will raise a PR to remove it from that particular block for now, and leave it in place for grid blocks only.

PR #4946, opened 2021-10-14
Update blocks-registry docs

### LINKED ISSUE #4799: Cart i2: Validate if extensibility is still working fine.
Block extensibility should work out of the box given we should just be introducing new innerBlock areas, but it should be validated anyway.

senadir:
This PR adds some missing pieces to our docs and removes the validation script from Cart inner blocks.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4799

### DIFF
--- a/packages/checkout/blocks-registry/README.md
+++ b/packages/checkout/blocks-registry/README.md
@@ -91,7 +91,7 @@ If you want your block to appear within the layout of the Checkout without merch
 In the above example, the inner block would be inserted automatically, and would not be movable or removable by the merchant.
 
 ### Passing attributes to your frontend block
-For your block to dynamically render on the frontend, and for you to access your attributes, you need to pass them in the frontend.
+For your block to dynamically render on the frontend and have access to its own attributes, both the block name and the list of block attributes need to be passed via HTML `data-` attributes.
 
 - To render the block on the frontend, you need a `data-block-name` attribute on the HTML with your block name `namespace/block-name`.
 - To access your attributes on frontend, you need to save them as `data-*` attributes on the HTML.
--- a/packages/checkout/blocks-registry/README.md
+++ b/packages/checkout/blocks-registry/README.md
@@ -102,7 +102,7 @@ This function registers a block and it's corresponding component with WooCommerc
 ### Passing attributes to your frontend block
 In order to access your attributes on your frontend component, you need to append them to your saved html.
 
-Blocks whose namespace is `woocommerce` or `woocommerce-checkout` would have this applied to them automaticly.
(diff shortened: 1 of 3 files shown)


packages/checkout/blocks-registry/README.md
```diff
@@ -97,6 +98,39 @@ After registering your block, you need to define which component will replace yo
 
 This function registers a block and it's corresponding component with WooCommerce. The register function expects a JavaScript object with options specific to the block you are registering.
 
+
+### Passing attributes to your frontend block
+In order to access your attributes on your frontend component, you need to append them to your saved html.
+
+Blocks whose namespace is `woocommerce` or `woocommerce-checkout` would have this applied to them automaticly.
```
opr:
```suggestion
Blocks whose namespace is `woocommerce` or `woocommerce-checkout` will have this applied to them automatically, but you can also add this behaviour to your own namespace or individual blocks.
```

packages/checkout/blocks-registry/README.md
```diff
@@ -97,6 +98,39 @@ After registering your block, you need to define which component will replace yo
 
 This function registers a block and it's corresponding component with WooCommerce. The register function expects a JavaScript object with options specific to the block you are registering.
 
+
+### Passing attributes to your frontend block
```
opr, line 93: This doesn't feel like it's in a logical place, could it be moved under the usage section?

Could you also add a more detailed description or example of what exactly this solves, or perhaps describe the way you'd have to do this without the filter?

opr (dismissed):

packages/checkout/blocks-registry/README.md
senadir: I expanded on this, but I believe this docblock should be high up on the tree, before diving into JS implementation. I moved it a step higher.

This is because having this block higher is essential to get the block to actually render on the frontend, missing this step caused a lot of [frustration and debugging when Niels was trying to render his block.](p1634634917287000-slack-C8X6Q7XQU)

packages/checkout/blocks-registry/README.md
```diff
@@ -89,6 +90,41 @@ If you want your block to appear within the layout of the Checkout without merch
 
 In the above example, the inner block would be inserted automatically, and would not be movable or removable by the merchant.
 
+### Passing attributes to your frontend block
+For your block to dynamically render on the frontend, and for you to access your attributes, you need to pass them in the frontend.
```
mikejolley:
```suggestion
For your block to dynamically render on the frontend and have access to its own attributes, both the block name and the list of block attributes need to be passed via HTML `data-` attributes.
```

mikejolley (approved):
Looks good. One suggestion.

PR #4950, opened 2021-10-15
Fix visual border for products image placeholder

### LINKED ISSUE #3354: Product image should have a 1px border if it's the placeholder
WooCommerce [sets a default 1px border](https://github.com/woocommerce/woocommerce/blob/master/assets/css/woocommerce.scss#L196-L198) to placeholder images in the Product grid:

imatge

WC Blocks don't honor those styles and placeholders are shown without border:

_All Products:_
imatge

_Product grid:_
imatge

## Expected result
I would expect the placeholder image to always show a border no matter if it's a WC Core product grid or WC Blocks.

The screenshots above are taken from the Varia theme. The issue is not reproducible in Storefront because it [prevents WC Core default styles from being registered](https://github.com/woocommerce/storefront/blob/trunk/inc/woocommerce/class-storefront-woocommerce.php#L29).

Ideally, when fixing this, we will review other blocks that render product/category images (Cart, Checkout, categories list, reviews, etc.) to ensure all of them are consistent.

ovidiul:
This PR fixes the issue opened in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3354 by adding a border to the placeholder image of "1px solid #f2f2f2" style to match the WooCommerce original theme display. 

The border is applied to placeholder images identified by having an empty `alt` text.

This PR also adds a fallback alternative text as product name value for images missing that value.

Fixes #3354

#### Accessibility

### Screenshots

Blocks that have been tested so far:

## All Products
Screenshot 2021-10-15 at 11 06 06

## Products By Category
Screenshot 2021-10-15 at 11 06 40

## Newest Products
Screenshot 2021-10-15 at 11 07 18

## On Sale Products
Screenshot 2021-10-15 at 11 08 35

## Top Rated Products 
Screenshot 2021-10-15 at 11 09 22

## Order Summary
Screenshot 2021-10-15 at 11 10 52

## Cart & Cart i2
Screenshot 2021-10-15 at 11 10 57

## Checkout
Screenshot 2021-10-15 at 11 11 05

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. merge this PR
2. create a product with no image
3. create a new post and add the WooCommerce blocks and include the product created at step 2. to see its placeholder image and border

### Performance Impact

- estimated no performance impact since these are mostly design changes

### Changelog

> Improve the product images placeholder display by adding a light gray border to it.


mikejolley:
Hi @ovidiul 

Thanks for contributing this fix.

Are you [running our linting rules](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/contributors/getting-started.md#linting)? I noticed in the PHP files short array syntax was removed.  

I also have a question, probably for @Aljullu, do you think we should use box-shadow here instead of borders? Mostly because borders can impact layout, but box-shadow will not. We don't need to style to match core exactly.

assets/css/abstracts/_mixins.scss
```diff
@@ -192,3 +192,8 @@ $fontSizes: (
 	$hex: str-slice(ie-hex-str($color), 4);
 	@return "%23" + unquote("#{$hex}");
 }
+
+// Products image placeholder border style.
+@mixin image-placeholder-border {
```
mikejolley: @Aljullu should probably chime in here too, but I think I'd prefer to have a color defined for borders inside `assets/css/abstracts/_colors` than a mixin, since the mixin only contains one thing.

ovidiul:
> Are you [running our linting rules](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/contributors/getting-started.md?rgh-link-date=2021-10-18T14%3A20%3A02Z#linting)? I noticed in the PHP files short array syntax was removed.
> 

@mikejolley yes, seems the array syntax replacement was actually done by the PHP linter at the time of the commit check.

assets/js/atomic/blocks/product-elements/image/block.js
```diff
@@ -108,7 +107,13 @@ export const Block = ( {
 
 const ImagePlaceholder = () => {
 	return (
-		<img src={ PLACEHOLDER_IMG_SRC } alt="" width={ 500 } height={ 500 } />
+		<img
+			className="product-image-placeholder"
```
mikejolley: This PR introduces 3 new classNames just for placeholders. How would you feel about using something else to target placeholder images, for example, selecting images with empty alt attributes?

src/BlockTypes/AbstractProductGrid.php
```diff
@@ -90,7 +90,7 @@ protected function render( $attributes = array(), $content = '' ) {
 				rawurlencode(
 					wp_json_encode(
 						array_map(
-							[ Package::container()->get( SchemaController::class )->get( 'product' ), 'get_item_response' ],
+							array( Package::container()->get( SchemaController::class )->get( 'product' ), 'get_item_response' ),
```
mikejolley: These changes are unrelated to this fix.

src/BlockTypes/AbstractProductGrid.php
```diff
@@ -395,7 +395,12 @@ protected function render_product( $product ) {
 	 * @return string
 	 */
 	protected function get_image_html( $product ) {
-		return '<div class="wc-block-grid__product-image">' . $product->get_image( 'woocommerce_thumbnail' ) . '</div>'; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
+
+		$placeholder_class = '';
+		if ( ! $product->get_image_id() ) {
+			$placeholder_class = 'wc-block-grid__product-image--placeholder';
+		}
+		return '<div class="wc-block-grid__product-image ' . $placeholder_class . '">' . $product->get_image( 'woocommerce_thumbnail' ) . '</div>'; // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
```
mikejolley: Please use `esc_attr` around the class name even if hardcoded, in case it's made dynamic in the future and missed.

mikejolley:
> yes, seems the array syntax replacement was actually done by the PHP linter at the time of the commit check.

This doesn't happen for me. I wonder if it's using your global ruleset? Did you do composer install in our repo?

src/BlockTypes/AbstractProductGrid.php
ovidiul: Fixed now!

src/BlockTypes/AbstractProductGrid.php
ovidiul: Fixed, thanks for the suggestion.

ovidiul:
> This doesn't happen for me. I wonder if it's using your global ruleset? Did you do composer install in our repo?

You are right, it was using the global ruleset due to the missing vendor install.

assets/css/abstracts/_mixins.scss
ovidiul: Makes sense as well, I've moved the color to `assets/css/abstracts/_colors` and adjusted the references.

assets/js/base/components/cart-checkout/product-image/index.tsx
```diff
@@ -17,8 +22,13 @@ const ProductImage = ( { image = {} }: ProductImageProps ): JSX.Element => {
 	const imageProps = {
 		src: image.thumbnail || PLACEHOLDER_IMG_SRC,
 		alt: decodeEntities( image.alt ) || '',
+		className: '',
```
Aljullu: You shouldn't need to set an empty `className` here.
```suggestion
```

assets/js/atomic/blocks/product-elements/image/style.scss
```diff
@@ -1,4 +1,7 @@
-.editor-styles-wrapper .wc-block-grid__products .wc-block-grid__product .wc-block-components-product-image,
+.editor-styles-wrapper
+.wc-block-grid__products
+.wc-block-grid__product
+.wc-block-components-product-image,
```
Aljullu: Can you review if you have stylelint correctly set up? Selectors should be always in one line.

Aljullu (commented):
Thanks for the contribution @ovidiul, I didn't do a full review, but left a couple of minor comments below that I think would be an improvement.

> I also have a question, probably for @Aljullu, do you think we should use box-shadow here instead of borders? Mostly because borders can impact layout, but box-shadow will not. We don't need to style to match core exactly.

@mikejolley [WooCommerce core is using border](https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/assets/css/woocommerce.scss#L591-L593), so I think we should stick to it for now.

assets/js/atomic/blocks/product-elements/image/block.js
ovidiul: Considering there is a `fallbackAlt` in this block, I agree it's probably safe to remove the extra class name.

Change added now.

assets/js/atomic/blocks/product-elements/image/style.scss
ovidiul: fixed now!

assets/js/base/components/cart-checkout/product-image/index.tsx
ovidiul: fixed as well, the only reason I've added that was because of my editor linting warning as:

> Property 'className' does not exist on type '{ src: string; alt: string; }'.ts(2339)

mikejolley:
Thanks 👍🏻 Should we do the alt change to all placeholders and avoid adding classes? Alt shouldn't be needed for placeholder images so if we can ensure it's blank we can style based on that always.

ovidiul:
> Thanks 👍🏻 Should we do the alt change to all placeholders and avoid adding classes? Alt shouldn't be needed for placeholder images so if we can ensure it's blank we can style based on that always.

Yes, the only conflict I have is [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4950/files#diff-d8e5c485504fe04139c869da3c92500d531258145052b654efc15bb34f6f0ed1L398) where the `$product->image()` might return an empty alt for exiting images as well, and the border would be added to those as well.

`cart-checkout` component and `AbstractProductGrid` both suffer from the same issue.

The solution would be to probably add a fallbackAlt as already implemented [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e23b51119dfa5c11e67ab0b3cbd89d49189a8569/assets/js/atomic/blocks/product-elements/image/block.js#L98) which is the product name.

Screenshot 2021-10-20 at 14 06 49
Screenshot 2021-10-20 at 14 07 17

What do you think about adding the fallbackAlt option for cases when existing product images alt is empty?

mikejolley:
@ovidiul Yeah I agree using the product name would be a good fallback for this.

ovidiul:
@mikejolley fallback for alternative text is now added and so, as confirmed above, images with an empty alt will be considered placeholders and styled accordingly.

assets/js/base/components/cart-checkout/product-image/style.scss
```diff
@@ -0,0 +1,3 @@
+img[alt=""] {
```
mikejolley: This does need greater specificity because it will apply to all images on the page, not just the product image in this context.

Let's give the image component a unique class like `wc-block-components-product-image` so we can target that specifically e.g. `img.wc-block-components-product-image[alt=""]`

assets/js/base/components/cart-checkout/product-image/index.tsx
```diff
@@ -4,21 +4,34 @@
...
+}: ProductImageProps ): JSX.Element => {
 	const imageProps = {
 		src: image.thumbnail || PLACEHOLDER_IMG_SRC,
-		alt: decodeEntities( image.alt ) || '',
+		alt: decodeEntities( image.alt ) || fallbackAlt || '',
 	};
 
+	if ( ! image.thumbnail ) {
+		imageProps.alt = '';
+	}
```
mikejolley: This section might be clearer if we split it all like this:

```js
const imageProps = image.thumbnail ? {
	src: image.thumbnail,
	alt: decodeEntities( image.alt ) || fallbackAlt || 'Product Image',
} : {
    src: PLACEHOLDER_IMG_SRC,
	alt: '',
};
```

mikejolley (commented):
Thanks for those changes. Just 2 small points and then we can ship it :)

assets/js/base/components/cart-checkout/product-image/index.tsx
ovidiul: Great suggestion @mikejolley, change is now added, thank you!

assets/js/base/components/cart-checkout/product-image/style.scss
ovidiul: Fixed as well, also merged the latest code from trunk.

mikejolley (approved):
Thanks, all looks and tests well now. Approving 👍🏻

PR #4854, opened 2021-09-28
Fix free trial subscriptions orders missing payment method meta

ricardo:
Fixes #4853

Issue in Stripe: https://github.com/woocommerce/woocommerce-gateway-stripe/issues/1964

When debugging the issue above in the Stripe plugin, I found that free trial subscriptions purchased via the blocks checkout didn't have the payment method post meta. This caused free trial subscriptions to always require manual renewals after the trial period ended.

This is possibly affecting pre-orders and other types of off-session based orders, although I only tested the issue with the subscriptions extension.

I reproduced the issue in Stripe and WooCommerce Payments, but this is probably happening in all payment gateways that support free trial subscriptions.

Apparently when the code got refactored, it wasn't refactored into its [equivalent previous functionality](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3454/files#diff-af2c90fa556cc086b780c8fad99b68373d87fd6007e6e2ff1b4c68ebe9ccb551L434-L436), so it led to the inconsistency mentioned above.

### Screenshots

**Before: Payment method didn't get saved and required manual renewals**

Free trial subscriptions purchased via checkout block - Before fix

**After: Payment method gets saved and supports automatic renewals**

Free trial subscriptions purchased via checkout block - After fix

#### Automated Tests
* [ ] Changes in this PR are covered by Automated Tests.
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

How to test the changes in this Pull Request:

1. Ensure you have [WooCommerce Subscriptions](https://woocommerce.com/products/woocommerce-subscriptions/) and [Stripe](https://woocommerce.com/products/stripe/) installed and activated.
2. Sign up for a stripe.com account
3. Enable test mode and add test API private and secret keys to your account under **WooCommerce > Settings > Payments > Stripe**.
4. Navigate to **Products > Add new** and type a product name: e.g "Free trial subscription".
5. Set the product data to **Simple subscription**.
6. Add a subscription price and a free trial period.
7. Save the product.
8. As a shopper, add the newly created free trial subscription to the cart and proceed to the checkout block.
9. Complete the purchase with a Stripe test card: `4242424242424242` and any valid CVC and expiration date.
10. Navigate to **WooCommerce > Subscriptions** and notice the newly created subscription is renewable "Via Credit card / debit card".

### Changelog

> Fix missing payment method meta data for free orders

PR #4955, opened 2021-10-18
Use Cart's needs payment check instead of Order's

### LINKED ISSUE #4853: Payment method doesn't get saved for free trial subscriptions
## Describe the bug

The issue was initially being investigated in https://github.com/woocommerce/woocommerce-gateway-stripe/issues/1964, but we found it's not coming from the gateway.

## To reproduce

1. Ensure you have [WooCommerce Subscriptions](https://woocommerce.com/products/woocommerce-subscriptions/) and [Stripe](https://woocommerce.com/products/stripe/) installed and activated.
2. Sign up for a stripe.com account
3. Enable test mode and add test API private and secret keys to your account under **WooCommerce > Settings > Payments > Stripe**.
4. Navigate to **Products > Add new** and type a product name: e.g "Free trial subscription".
5. Set the product data to **Simple subscription**.
6. Add a subscription price and a free trial period.
7. Save the product.
8. As a shopper, add the newly created free trial subscription to the cart and proceed to the checkout block.
9. Complete the purchase with a Stripe test card: `4242424242424242` and any valid CVC and expiration date.
10. Navigate to **WooCommerce > Subscriptions** and notice the newly created subscription is requires a manual renewal.

The above only happens with the checkout block.

## Expected behavior
Free trial subscriptions shouldn't require manual renewals.

## Screenshots

Issue - free trial WooCommerce subscriptions require manual renewals

### LINKED ISSUE #4948: "No payment method provided" error at checkout if cart total is 0
## Describe the bug
Using the Checkout block, if cart total is 0 (either by produce price or use of a coupon), and no payment methods are defined, the checkout cannot be completed. Error **No payment method provided** appears.

I am able to proceed if using the standard `[woocommerce_checkout]`, but not with the new WooCommerce Block version

## To reproduce

1. Disable payment methods
2. Add a $0 product, or use a coupon code with 100% discount
3. Attempt to checkout

## Expected behavior
We should be able to checkout if the cart total is 0.

* Core version: [e.g. 5.8.1]
* WooCommerce version: [e.g. 5.8.0]
* WooCommerce Blocks version: [e.g. 6.1.0]
* Site language: english

* OS: Windows 10
* Browser Brave
* Version 1.30.87 Chromium: 94.0.4606.71 (Official Build) (64-bit)

## Additional context
Checkout works as expected if using the original (non-Blocks) version of WooCommerce Checkout

senadir:
#4854 Removed a check to see if the order needs payment before setting the payment ID on the order. This caused a regression (#4948) In which free orders didn't pass because there are no payment method.

Reverting #4854 fixes #4948 but regresses #4853 , so this PR reverts it and fixes the original issue in both cases.

### What happened?

- We show the payment setup on Checkout based on `$cart->needs_payment()`. Plugins (like WC Subscriptions) hooks into `needs_payment` and sets it to true based on conditions like trial and such. However, when setting a payment method, we check `$order->needs_payment`.

WooCommerce Subscriptions[ does hook into `$order->needs_payment`](https://github.com/woocommerce/woocommerce-subscriptions/blob/d0f64e5e23b8fcc5bdb2fd422a3e7da6a9d76a78/includes/class-wc-subscriptions-order.php#L658-L704) but that didn't work, because that value is false, causing the subscription recurring payment method to be `Manual renewal` even if a payment was still captured.

This PR changes that check into `$cart->needs_payment` instead, why?

I believe, at that moment in the checkout lifecycle, using `$cart->needs_payment` makes more sense than order given that we still didn't submit the order, and that we use that exact check to decide if we should capture payment or not, so they should be consistent.

### Question

There are several other instances in `Routes/Checkout.php` that uses `$order->needs_payment`, should I evaluate them to see if they should be `$cart->needs_payment` or not? In total, there are two other instances left.

Fixes #4948
Reverts #4854 
Fixes #4853

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

We're testing two issues here, #4948, and #4853

#4948
1. Create a free order ($0), Add to Cart.
2. Make sure you have a free shipping method or it's a virtual product.
2. Visit Checkout, Payment and Express Payment steps shouldn't render.
3. Place order, you should not see an error.

#4853
1. Install WC Subscriptions and enable a payment gateway like Stripe or WCPay.
2. Create a product with a trial.
2. Visit Checkout, Payment and Express Payment steps **should render.**
3. Place order, you should not see an error.
4. Visit wp-admin->WooCommerce->Subscriptions
5. See the latest created subscription, it should have Via Credit Card or any payment gateway you did, not vial Manual Renewals.
image

### Changelog

> Fix a bug in free orders and trial subscription products.

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -464,7 +464,7 @@ class Checkout extends AbstractCartRoute {
 		$this->order->set_customer_note( $request['customer_note'] ?? '' );
 
 		$cart = $this->cart_controller->get_cart_instance();
-		$this->order->set_payment_method( $cart->needs_payment() ? $this->get_request_payment_method( $request ) : '' );
+		$this->order->set_payment_method( $request['payment_method'] ?? '' );
 
 		/**
 		 * Fires when the Checkout Block/Store API updates an order's from the API request data.
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -462,8 +462,6 @@ class Checkout extends AbstractCartRoute {
 	 */
 	private function update_order_from_request( \WP_REST_Request $request ) {
 		$this->order->set_customer_note( $request['customer_note'] ?? '' );
-
-		$cart = $this->cart_controller->get_cart_instance();
 		$this->order->set_payment_method( $request['payment_method'] ?? '' );
 
 		/**
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -462,7 +462,9 @@ class Checkout extends AbstractCartRoute {
 	 */
 	private function update_order_from_request( \WP_REST_Request $request ) {
 		$this->order->set_customer_note( $request['customer_note'] ?? '' );
-		$this->order->set_payment_method( $this->order->needs_payment() ? $this->get_request_payment_method( $request ) : '' );
+
+		$cart = $this->cart_controller->get_cart_instance();
+		$this->order->set_payment_method( $cart->needs_payment() ? $this->get_request_payment_method( $request ) : '' );
 
 		/**
 		 * Fires when the Checkout Block/Store API updates an order's from the API request data.
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -462,7 +462,7 @@ class Checkout extends AbstractCartRoute {
 	 */
 	private function update_order_from_request( \WP_REST_Request $request ) {
 		$this->order->set_customer_note( $request['customer_note'] ?? '' );
-		$this->order->set_payment_method( $this->get_request_payment_method( $request ) );
+		$this->order->set_payment_method( $this->order->needs_payment() ? $this->get_request_payment_method( $request ) : '' );
 
 		/**
 		 * Fires when the Checkout Block/Store API updates an order's from the API request data.


senadir:
@mikejolley to give this an eye and answer the question above.

mikejolley:
@senadir Pardon me if missing the point here, but the error in #4948 indicates that we're attempting to take a payment when no payment is required? Setting payment method blank (changed in this PR) is ok. Why do we need an extra check there?

I see our usage as follows:

`$cart->needs_payment()` used by the cart endpoint. This would be useful before checkout, for example, to determine if express payment methods are shown.

`$order->needs_payment()` used by the checkout endpoint. This would be preferred once the draft order has been created, and this will be true or false based on cost.

I don't think we should mix these up between the 2 endpoints.  

> WooCommerce Subscriptions does hook into $order->needs_payment but that didn't work, because that value is false, causing the subscription recurring payment method to be Manual renewal even if a payment was still captured.

https://github.com/woocommerce/woocommerce-subscriptions/blob/d0f64e5e23b8fcc5bdb2fd422a3e7da6a9d76a78/includes/class-wc-subscriptions-order.php#L658-L704

Is there a bug in this function? It would only bail if `true`. Why would it return false in this scenario?

senadir:
> $cart->needs_payment() used by the cart endpoint. This would be useful before checkout, for example, to determine if express payment methods are shown.

> $order->needs_payment() used by the checkout endpoint. This would be preferred once the draft order has been created, and this will be true or false based on cost.

My point around using the cart's method was because we rely on to decide if we want to capture a payment on Checkout ([Payment block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0e50628f52e483da831ed127a021e754a1be4db1/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-payment-block/frontend.tsx#L33-L35), [Express Payment block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/0e50628f52e483da831ed127a021e754a1be4db1/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-express-payment-block/block.tsx#L14-L16)). So having a parity between if we decide to capture payment and if we should set a payment method.

Do you think all plugins who accept zeored orders would hook into both `woocommerce_cart_needs_payment` an `woocommerce_order_needs_payment`? The issue above is one instance of why those conditions might be different. Up to Checkout block, orders weren't created untill you hit submit, meaning `woocommerce_order_needs_payment` was in a different flow than `woocommerce_cart_needs_payment`.

mikejolley:
> My point around using the cart's method was because we rely on to decide if we want to capture a payment on Checkout (Payment block, Express Payment block). So having a parity between if we decide to capture payment and if we should set a payment method.

The checkout one could probably switch if we have an order by that point. 

> Do you think all plugins who accept zeored orders would hook into both woocommerce_cart_needs_payment an woocommerce_order_needs_payment?

I don't have an answer for this but I believe you'd need both..there are flows in core when you'd want both, e.g. paying for an order from My Account.

senadir:
> The checkout one could probably switch if we have an order by that point.

I tried doing that, we can do one of two:
- Add `$order->needs_payment` to `/store/checkout`, however, this won't work as `/store/checkout` won't refresh if you update shipping method, add coupons, or any other amount changing variables, would cause a degradation of quality if we always refresh checkout an cart, batching can solve this somehow.
- Make the `needs_payment` value in `/store/cart` conditional so it uses either `$order->needs_payment` or `$cart->needs_payment` depending on context, seems to be the hardest option.

> I don't have an answer for this but I believe you'd need both..there are flows in core when you'd want both, e.g. paying for an order from My Account.

Should they be assumed to be similar is the issue here, meaning would it always be `$cart->needs_payment() === $order->needs_payment()` or else we need to switch our Checkout conditional rendering to use `$order->needs_payment()`

mikejolley:
Ok, I'm starting to understand this a bit more after seeing the diff in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4854

If we leave the cart/order checks alone and assume that cart data is synced to the order before these checks run and just focus on the change in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4854, that PR is problematic because it assumes a payment method is sent with all requests when I believe it's optional if no payment is required. 

To solve the original "Fix free trial subscriptions orders missing payment method meta" issue we can handle it in 2 ways:

1. Ensure subs filters on order needs payment as well as cart needs payment so this code still runs even if an order has 0 total
2. In subs, use one of the other filters to set the payment_id on the order if total is 0

Do you agree with that @senadir?

This minimises the changes we're making our side. We should revert #4854 because `get_request_payment_method_id` runs validation.

senadir:
2 doesn't seem like a possible solution, right? I don't think subs has access to the payment id if it's not saved on the order.

> Ensure subs filters on order needs payment as well as cart needs payment so this code still runs even if an order has 0 total

This could be possible but it might be a bit risky of a change, but we this is one possible solution, however, it only fixes it for Subs, the issue if `cart->needs_payment !== order->needs_payment` would still persist.

mikejolley:
@senadir only in the client though? Whenever the checkout endpoint is hit (so when you make payment) it syncs with cart: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/StoreApi/Routes/Checkout.php#L370-L382

When else would they differ?

senadir:
> Whenever the checkout endpoint is hit (so when you make payment) it syncs with cart

It only syncs the content, so the core logic of `cart->needs_payment` and `order->needs_payment` would be the same, but this doesn't account for filters and if plugins treats those two differently, which is the issue happening with WooCommerce Subscriptions for example. Going with this assumption would require us to maybe document this change (that `cart->needs_payment` and `order->needs_payment` are used interchangeably in Checkout block and therefore should have the same logic)

mikejolley:
> 2 doesn't seem like a possible solution, right? I don't think subs has access to the payment id if it's not saved on the order.

`__experimental_woocommerce_blocks_checkout_update_order_from_request` has the request data and the order so could be used...

If we are leaving the code as is with the change that saved payment method for all requests, then we need to remove the validation inside `update_order_from_request`, so perhaps:

```
$this->order->set_payment_method( $this->get_request_payment_method( $request ) );
```

to: 

```
$this->order->set_payment_method( $request['payment_method'] ?? '' );
```

OR remove the validation inside `get_request_payment_method_id` if the payment method is empty.

Then we just need to ensure validation runs before the actual payment is attempted. This should satisfy both use case?

nielslange (approved):
I checked both manual test cases against the current trunk and against your PR and your fix works as expected.

As for the points @mikejolley mentioned, I cannot say too much about this yet, so I'd leave that conversation up to you, @senadir. In case this PR remains as it is, feel free to merge it.

senadir:
@mikejolley using
```
$this->order->set_payment_method( $request['payment_method'] ?? '' );
```

Worked like a charm!

src/StoreApi/Routes/Checkout.php
```diff
@@ -462,7 +462,9 @@ private function update_customer_from_request( \WP_REST_Request $request ) {
 	 */
 	private function update_order_from_request( \WP_REST_Request $request ) {
 		$this->order->set_customer_note( $request['customer_note'] ?? '' );
-		$this->order->set_payment_method( $this->get_request_payment_method( $request ) );
+
+		$cart = $this->cart_controller->get_cart_instance();
```
mikejolley: This is no longer needed.

mikejolley (dismissed):

PR #4973, opened 2021-10-20
Migrate Cart i1 to Cart i2

### LINKED ISSUE #4798: Cart i2: setup automatic Depreciation from Cart i1 to Cart i2.
This would require us to:
- Remove all Cart i1 code.
- Move Cart i2 to Cart.
- Setup deprecation.
- Setup `render_callback` to Cart to render basic columns.

senadir:
This PR migrates Cart i1 to Cart i2.

Closes #4798

### Testing steps

Migration process:

1. Before using this PR, create a page containing a cart block, change alignment.
2. Use this PR. Load the page in the editor.
3. Confirm it shows the correct content in both filled/empty states. Do not save.
4. View frontend - confirm cart is functional.
5. Save page in editor.
6. View frontend - confirm cart is functional.

Cart in general:

1. Insert cart into page and smoke test functionality.

Mini Cart:

1. Insert mini cart and smoke test functionality.

### Dev note

> Some Cart block options might not work in the frontend unless you resave the cart block again in the editor.

### Changelog

> Cart v2: The cart block, like checkout block, now supports inner blocks that allow for greater customizability.


assets/js/blocks/cart-checkout/cart/block.js
```diff
@@ -1,43 +1,54 @@
...
-const EmptyCart = ( { content } ) => {
-	useEffect( () => {
-		dispatchEvent( 'wc-blocks_render_blocks_frontend', {
-			element: document.body.querySelector(
-				'.wp-block-woocommerce-cart'
-			),
-		} );
-	}, [] );
-	return <RawHTML>{ content }</RawHTML>;
-};
```
senadir, line 30: This was moved to Empty cart block, no RawHTML needed

assets/js/blocks/cart-checkout/cart/block.js
```diff
@@ -75,19 +86,32 @@ const Block = ( { emptyCart, attributes, scrollToTop } ) => {
 		};
 	}, [ scrollToTop ] );
```
senadir, line 92: I used the updated ScrollonError from Checkout.

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -1,47 +1,54 @@
+/* tslint:disable */
```
senadir, line 1: We will handle TS in a follow up, this is useful to keep a good blame history.

assets/js/blocks/cart-checkout/cart/frontend.js
```diff
@@ -5,70 +5,50 @@ import {
...
-			<StoreNoticesProvider context="wc/cart">
-				<SlotFillProvider>
-					<CartProvider>
-						<Block { ...props } />
-					</CartProvider>
-				</SlotFillProvider>
-			</StoreNoticesProvider>
-		</StoreSnackbarNoticesProvider>
-	);
-};
```
senadir, line 44: This moves to Block.

assets/js/blocks/cart-checkout/cart/index.js
```diff
@@ -29,23 +31,51 @@ const settings = {
...
+					[
+						createBlock( 'woocommerce/filled-cart-block' ),
+						createBlock(
+							'woocommerce/empty-cart-block',
+							{},
+							innerBlocks
+						),
+					],
+				];
+			},
```
senadir: This function takes the previous inner Blocks from Cart i1 and place it in the correct place.

assets/js/blocks/cart-checkout/cart/index.js
```diff
@@ -29,23 +31,51 @@ const settings = {
...
+			isEligible: ( _, innerBlocks ) => {
+				return ! innerBlocks.find(
+					( block ) => block.name === 'woocommerce/filled-cart-block'
+				);
+			},
+		},
+	],
```
senadir, line 105: Because the block didn't really break, we need to force migration.

assets/js/blocks/cart-checkout/cart/inner-blocks/empty-cart-block/edit.tsx
```diff
@@ -94,7 +95,6 @@ export const Edit = ( { clientId }: { clientId: string } ): JSX.Element => {
 			hidden={ currentView !== 'woocommerce/empty-cart-block' }
 		>
 			<InnerBlocks
-				allowedBlocks={ allowedBlocks }
```
senadir, line 97: Empty cart shouldn't be limited by block types.

src/BlockTypes/Cart.php
```diff
@@ -80,7 +80,32 @@ protected function render( $attributes, $content ) {
 		wp_dequeue_script( 'selectWoo' );
 		wp_dequeue_style( 'select2' );
 
-		return $content . $this->get_skeleton();
+		// If the content contains new inner blocks, it means we're in the newer version of Cart.
+		$regex_for_new_block = '/<div[\n\r\s\ta-zA-Z0-9_\-=\'"]*data-block-name="woocommerce\/filled-cart-block"[\n\r\s\ta-zA-Z0-9_\-=\'"]*>/mi';
```
senadir, line 84: We check if we're running Cart i2.

src/BlockTypes/Cart.php
```diff
@@ -80,7 +80,32 @@ protected function render( $attributes, $content ) {
...
+
+		$is_new = preg_match( $regex_for_new_block, $content );
+
+		if ( ! $is_new ) {
```
senadir, line 88: If we're not, then we're going to insert the default template

src/BlockTypes/Cart.php
```diff
@@ -80,7 +80,32 @@ protected function render( $attributes, $content ) {
...
+			// This fallback needs to match the default templates defined in our Blocks.
+			$inner_blocks_html = '$0
```
senadir, line 90: $0 is used here to replace the old block opening, it supports custom calsses.

src/BlockTypes/Cart.php
```diff
@@ -80,7 +80,32 @@ protected function render( $attributes, $content ) {
...
+					<div data-block-name="woocommerce/cart-order-summary-block" class="wp-block-woocommerce-cart-order-summary-block"></div>
+					<div data-block-name="woocommerce/cart-express-payment-block" class="wp-block-woocommerce-cart-express-payment-block"></div>
+					<div data-block-name="woocommerce/proceed-to-checkout-block" class="wp-block-woocommerce-proceed-to-checkout-block"></div>
+					<div data-block-name="woocommerce/cart-accepted-payment-methods-block" class="wp-block-woocommerce-cart-accepted-payment-methods-block"></div>
+				</div>
+			</div>
+			<div data-block-name="woocommerce/empty-cart-block" class="wp-block-woocommerce-empty-cart-block">
+			';
+
+			$content = preg_replace( '/<div class="[a-zA-Z0-9_\- ]*wp-block-woocommerce-cart[a-zA-Z0-9_\- ]*">/mi', $inner_blocks_html, $content );
```
senadir, line 105: Insert template at start.

src/BlockTypes/Cart.php
```diff
@@ -80,7 +80,32 @@ protected function render( $attributes, $content ) {
...
+			$content = substr_replace( $content, '</div></div>', -6 );
```
senadir: insert a closing div at the end.

assets/js/blocks/cart-checkout/cart/full-cart/cart-line-items-table.tsx
```diff
@@ -1,95 +0,0 @@
...
-	lineItems = [],
-	isLoading = false,
-}: CartLineItemsTableProps ): JSX.Element => {
-	const tableRef = useRef< HTMLTableElement | null >( null );
-	const rowRefs = useRef( setRefs( lineItems ) );
-	useEffect( () => {
-		rowRefs.current = setRefs( lineItems );
-	}, [ lineItems ] );
-
-	const onRemoveRow = ( nextItemKey: string | null ) => () => {
```
Aljullu: For example, this function doesn't seem to be in `/assets/js/blocks/cart-checkout/cart/cart-line-items-table/index.tsx`. Could we make it sure it's added there?

assets/js/blocks/cart-checkout/cart/block.js
```diff
@@ -1,43 +1,55 @@
...
+			</CartBlockContext.Provider>
+		</LoadingMask>
+	);
+};
+
+const ScrollOnError = ( { scrollToTop } ) => {
 	useEffect( () => {
-		const invalidateCartData = ( e ) => {
-			const eventDetail = e.detail;
-			if ( ! eventDetail || ! eventDetail.preserveCartData ) {
```
Aljullu: These two lines too. They are needed to avoid a full reload of the Cart contents when adding products from the All Products block.

Aljullu (commented):
I didn't do a full review, but wanted to raise a couple of changes that were implemented recently and that it looks like they are not included in Cart i2. I didn't do a close examination, so there might be more, but these are two that stood out to me. :point_down:

assets/js/blocks/cart-checkout/cart/test/block.js
```diff
@@ -38,38 +32,23 @@ describe( 'Testing cart', () => {
 	} );
 
 	it( 'renders cart if there are items in the cart', async () => {
-		render(
-			<CartBlock
-				emptyCart={ null }
-				attributes={ {
-					isShippingCalculatorEnabled: false,
-				} }
```
senadir, line 46: We no longer need to define emptyCart or attributes here.

assets/js/blocks/cart-checkout/cart/test/block.js
```diff
@@ -38,38 +32,23 @@ describe( 'Testing cart', () => {
...
-		);
+		render( <CartBlock /> );
 		await waitFor( () => expect( fetchMock ).toHaveBeenCalled() );
 		expect(
 			screen.getByText( /Proceed to Checkout/i )
 		).toBeInTheDocument();
 
 		expect( fetchMock ).toHaveBeenCalledTimes( 1 );
-		// ["`select` control in `@wordpress/data-controls` is deprecated. Please use built-in `resolveSelect` control in `@wordpress/data` instead."]
-		expect( console ).toHaveWarned();
```
senadir: This is now fixed.

assets/js/blocks/cart-checkout/cart/test/block.js
```diff
@@ -38,38 +32,23 @@ describe( 'Testing cart', () => {
...
-				attributes={ {
-					isShippingCalculatorEnabled: false,
-				} }
-			/>
-		);
+		render( <CartBlock /> );
+
 		await waitFor( () => expect( fetchMock ).toHaveBeenCalled() );
-		expect( container ).toMatchSnapshot();
+		expect( screen.getByText( /Tax/i ) ).toBeInTheDocument();
```
senadir, line 102: Removed snapshots in favor of RTL selectors.

assets/js/blocks/cart-checkout/cart/test/cart-block.js
```diff
@@ -0,0 +1,46 @@
...
+					/>
+					<ExpressPaymentBlock />
+					<ProceedToCheckoutBlock />
+					<AcceptedPaymentMethodsIcons />
+				</TotalsBlock>
+			</FilledCart>
+			<EmptyCart>
+				<p>Empty Cart</p>
+			</EmptyCart>
+		</Cart>
```
senadir: Since we're running an innerBlocks structure now, we need to construct the block ourselves.

ralucaStan:
Is there a reason we see Accepted Payment Methods grayed out when trying to add a new block on the right column of the cart?
Screenshot 2021-10-21 at 12 59 25

senadir:
@ralucaStan 
> Is there a reason we see Accepted Payment Methods grayed out when trying to add a new block on the right column of the cart?

Yes, that block is optional and so can be inserted and removed, it's also restricted to a single use, so it's grayed out because it's already inserted.

ralucaStan:
Some small Carti2 leftovers: 
- 3 in this [file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/bin/hook-docs/data/actions.json#L201-L210), 
- 1 mention of `BlockTypes/CartI2.php` [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/actions.md)

senadir:
I think those can be updated automatically the next time we run the automation script?

assets/js/blocks/cart-checkout/cart/inner-blocks/cart-order-summary-block/block.tsx
```diff
@@ -50,7 +50,7 @@ const Block = ( {
 	// Prepare props to pass to the ExperimentalOrderMeta slot fill.
 	// We need to pluck out receiveCart.
 	// eslint-disable-next-line no-unused-vars
-	const { extensions, ...cart } = useStoreCart();
+	const { extensions, receiveCart, ...cart } = useStoreCart();
```
ralucaStan, line 53: `// We need to pluck out receiveCart.` Unless we specify why we need to remove it, this is just commenting on the obvious. So I recommend adding a reason for why the prop needs to be removed or just removing the comment

assets/js/blocks/cart-checkout/cart/inner-blocks/cart-order-summary-block/attributes.tsx
```diff
@@ -4,6 +4,10 @@
 import { getSetting } from '@woocommerce/settings';
 
 export default {
+	isShippingCalculatorEnabled: {
```
ralucaStan, line 7: Disabling the shipping calculator does not work for me. 
I've disabled it from the Order summary inner block and saved; the link to change the address still appears in the front-end.
This happens also if I add the block to a new page.

ralucaStan (commented):
@senadir I'm gonna post this review so that you can look into my comments and I'll post individual comments for the rest of the review

assets/js/blocks/cart-checkout/cart/inner-blocks/cart-order-summary-block/block.tsx
senadir: This is rather common comment in our codebase. receiveCart allows you to update the content of cart, we use it in several places but we don't want to provide it for 3pd because we want to keep the integrity of cart.

I can update them all in a follow up.
image

ralucaStan: Please do, adding the reasons you just mentioned would really improve those comments and our understanding.

assets/js/blocks/cart-checkout/cart/inner-blocks/empty-cart-block/edit.tsx
```diff
@@ -13,6 +13,7 @@ import { SHOP_URL } from '@woocommerce/block-settings';
  */
 import { useForcedLayout, getAllowedBlocks } from '../../../shared';
 import iconDataUri from './icon-data-uri.js';
+import './style.scss';
```
ralucaStan, line 16: should we rename this to `editor.scss`? I see that for other inner blocks, the css imported in the editor is editor.scss, and the css for front-end is named `style.scss`.

assets/js/blocks/cart-checkout/cart/inner-blocks/cart-order-summary-block/attributes.tsx
ralucaStan: Probably a side effect of the issue above, but if I have a setting to hide the calculator in Cart i1 this is not carried to Cart i2. 
The other attributes transfer fine.

ralucaStan (commented):
I've finished my review, things I have tested:
- Coupon code
- Transfer of cart attributes from i1 to i2
- Filters in cart 
- Interaction with the editor
- Interaction with the cart on the frontend
- Express payment
- 0 total 
- Shipping calculator
- Inserting custom blocks

Things I've found no to work:
- setting custom classes for inner blocks, guess it will be fixed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4962
- disabling the Shipping calculator/ passing this setting from Cart i1 to Cart i2
- "shows empty cart when changing the view" test was failing randomly for me, I've pushed a change to see if it fixes it

ralucaStan:
> I think those can be updated automatically the next time we run the automation script?

yes, running `npm run build:docs` fixed this. I've pushed [a commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4973/commits/08c1d4bb406d3c2b6dc4bc3f0113adf5b556900d) with the changes

senadir:
> disabling the Shipping calculator/ passing this setting from Cart i1 to Cart i2

I managed to get this to work on the editor, no luck on the frontend. It would make the block too complicated for little gain, we're going to issue a dev note with the block stating that people need to resave if they changed attributes. The attributes are mostly cosmetic and provide good defaults already, so it's not a big issue.

assets/js/blocks/cart-checkout/cart/inner-blocks/cart-order-summary-block/attributes.tsx
senadir: This is now fixed.

assets/js/blocks/cart-checkout/cart/inner-blocks/empty-cart-block/edit.tsx
senadir: styles specific only to editor are named `editor.scss`. Shared styles can still be named `style.scss`, in this case, the styles there are used in both places.

ralucaStan: 👌 Then we should also import the file in frontend.tsx

assets/js/blocks/cart-checkout/cart/inner-blocks/cart-order-summary-block/attributes.tsx
ralucaStan: I tested adding a custom class and it works 👍

senadir:
> The attributes are mostly cosmetic and provide good defaults already, so it's not a big issue.

as per request from @ralucaStan, I'm expanding on this section.

To get old attributes support in the new layout, we need to pass them via `CartBlockContext` and access them in the inner blocks templates and blocks. At that point we need to decide if we should apply the attribute saved with the inner block or coming from above. We need to do this both on frontend and editor.

This essentially introduces extra code for a simple migration code that we later need to delete. I didn't find value in doing this, especially as what we're trying to do is carry over the value of two cosmetic attributes. Having the wrong value in the frontend has little impact (e.g: showing the taxes % or not, showing shipping calculator or not). The attributes value isn't lost, it's saved in the editor block migration code (which is built in for that) and would reapply if the merchants saves the editor page again.
So it's a factor of how much effort/fragility we're creating for the inconvenience of possibly losing the value you applied.

assets/js/blocks/cart-checkout/cart/edit.js
```diff
@@ -162,85 +100,93 @@ const BlockSettings = ( { attributes, setAttributes } ) => {
...
- */
-const CartEditor = ( { className, attributes, setAttributes } ) => {
+export const Edit = ( { className, attributes, setAttributes, clientId } ) => {
+	const { hasDarkControls } = attributes;
+	const { currentView, component: ViewSwitcherComponent } = useViewSwitcher(
+		clientId,
+		[
+			{
+				view: 'woocommerce/filled-cart-block',
+				label: __( 'Filled Cart', 'woo-gutenberg-products-block' ),
```
opr, line 110: Noting that this previously read `Full Cart`

src/BlockTypes/Cart.php
opr: What exactly is the function of this line? I understand it replaces the last 6 chars with `</div></div>` am I correct?

Why not just add `</div></div>` to the end of the `$inner_blocks_html` string?

opr (approved):
Apart from the issues already pointed out by @ralucaStan already about attribute migration and custom classes, there is also the issue of the width of the block, the content doesn't stretch like it did in the old cart. I don't think this is too much of a big deal but just flagging it anyway.

**edit: the width setting works fine on the front-end**

| Old cart | New cart (after migration, before saving) |
|---|---|
| image | image |

Happy to approve on the basis that:
- the migration from old cart to new cart works well and doesn't break the site, and the planned dev note will address what we've pointed out.
- The comment I left about translation is addressed/defended.
- The comment I left about the template is explained.

assets/js/blocks/cart-checkout/cart/edit.js
senadir: Yes, we changed from Full to Filled because Full didn't make sense, while filled assumes there's at least one item.

ralucaStan (approved):
👏

src/BlockTypes/Cart.php
senadir: Yeah that would work.

assets/js/blocks/cart-checkout/cart/inner-blocks/empty-cart-block/edit.tsx
senadir: Implemented in 1c4ac28

Issue #4964, opened 2021-10-19
Cart i2: block has no skeleton and the Empty cart is showed until the block loads


ralucaStan:
## Describe the bug
There is no loading state for the block that shows the cart preview. Because of this, the empty cart view is shown until the block loads.

## Expected behaviour
Show a preview until the block loads.

This was discovered while working on #4793

PR #4976, opened 2021-10-20
Add empty cart template to Cart i2

### LINKED ISSUE #4793: Cart i2: Add empty cart default inner blocks

ralucaStan:
This PR closes #4793 
Know issues: because Cart i2 doen't have a skeleton, the Empty cart will be visible until the block is loaded. This will be fixed as part of #4964 

How to test this PR:

1. Go to a page and add the Cart i2 block
2. Make sure you can select the Empty cart view
3. Go to the front end with an empty cart
4. See that the empty cart is rendered


assets/js/blocks/cart-checkout/cart-i2/inner-blocks/empty-cart-block/style.scss
```diff
@@ -0,0 +1,4 @@
+.wc-block-cart__empty-cart__title,
+.editor-styles-wrapper .wc-block-cart__empty-cart__title {
+	font-size: inherit;
+}
```
senadir, line 4: This is not imported anywhere.

senadir (approved):
Thank you for working on this, besides the conflict, the styles aren't imported, but this is looking fine.

assets/js/blocks/cart-checkout/cart-i2/inner-blocks/empty-cart-block/style.scss
senadir: But it's working just fine 🤔

ralucaStan: Because it is using the CSS from `cart/empty-cart-edit`. Even if I import it, it's still using the one from the cart.

PR #4979, opened 2021-10-21
Add inline note around displayWithTax to clarify why prices change

### LINKED ISSUE #4884: Mismatch behavior between preview and real cart data
## Describe the bug
Working on #4852, I found that we return different preview data based on the `woocommerce_tax_display_cart` option:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8f3c29265fae50eeb43f933b261ba6cdd8fca344/assets/js/previews/cart.ts#L200-L202

While it may have little to no effect on the end-users, it's the wrong usage of  `woocommerce_tax_display_cart` IMO. That option should be used to change how the price is displayed, not the price value itself. 

## Expected behavior
The preview data should be the same regardless of `woocommerce_tax_display_cart`. Blocks should take that option into account while displaying the price.

mikejolley:
Updates cart preview data to explain why displayWithTax changes the response. This matches the Store API behaviour. Also simplifies the totals returned to use round numbers.

Closes #4884 

#### Testing

View the cart block in the editor and check the preview data makes sense. Toggle the setting "Display prices during cart and checkout" in WC > Settings > Tax also and confirm it's still accurate.


senadir (approved):
Testing fine, and snapshot file is back.

assets/js/previews/cart.ts
```diff
@@ -11,7 +11,16 @@ import { getSetting } from '@woocommerce/settings';
  */
 import { previewShippingRates } from './shipping-rates';
 
+/**
+ * Prices from the API may change because of this display setting. This makes the response use either
+ * wc_get_price_including_tax or wc_get_price_excluding_tax, so it is correct that this changes the cart preview data.
+ *
+ * Cart endpoints use displayCartPricesIncludingTax. Other endpoints use displayProductPricesIncludingTax.
```
ralucaStan: For me, this line introduces a bit of confusion, as it does not explain what is the difference between the two.

ralucaStan (approved):
I've tested and it looks ok, the prices are clearly easier to understand now. 
I've left a comment about your comment :) but it's good to go

assets/js/previews/cart.ts
mikejolley: Between wc_get_price_including_tax and wc_get_price_excluding_tax?

ralucaStan: between displayCartPricesIncludingTax and displayProductPricesIncludingTax

assets/js/previews/cart.ts
```diff
@@ -11,7 +11,20 @@ import { getSetting } from '@woocommerce/settings';
...
+/**
+ * Prices from the API may change because of this display setting. This makes the response use either
+ * wc_get_price_including_tax or wc_get_price_excluding_tax. It is correct that this setting changes the cart preview
+ * data.
+ *
+ * WooCommerce core has 2 settings which control this, one for cart (displayCartPricesIncludingTax), and one for the
+ * rest of the store (displayProductPricesIncludingTax). Because of this, Cart endpoints use displayCartPricesIncludingTax
+ * which is the most appropriate.
+ *
+ * Handling the display settings server side help work around rounding/display issues that can arrise from manually
```
frontdevde: Minor typo 
```suggestion
 * Handling the display settings server-side helps work around rounding/display issues that can arise from manually
```

PR #4981, opened 2021-10-21
BlockTemplateController: adds the ability to load and manage block template files

### LINKED ISSUE #4969: Create BlockTemplateController to load Block Templates
## Description
Create a BlockTemplateController _similar_ to that in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4939 to load in Block Templates

## Acceptance criteria
* Given a user is using a compatible Block Theme the class will initialize
* Given a user **does not** have Woocommerce block templates in their theme, it will load in default templates located in `woo-blocks-plugin/templates/block-templates`
* Given a user is using the default block templates they will be accessible via the Site Editor menu item

tjcafferkey:
Added `BlockTemplateController` which will give us the ability to load default block templates in the future. In this specific PR the class is checking if they exist in the theme before loading them from Woo Blocks `templates/block-templates/` directory.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4969

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Ensure you're using a block enabled theme
2. Add the **second** template below to the Woo Blocks path specified above the snippet.
3. Confirm this loads in the Site Editor templates, it should say in the editor "You are using the Woo Blocks block template now"
4. Add the **first** template below to themes directory path specified above the snippet.
5. Confirm this loads in the Site Editor templates, it should say in the editor "You are using the Themes block template now"
6. This code is not yet responsible for loading the templates on the frontend. This will be done as part of https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4969

`/theme-dir/block-templates/single-product.html`
```html
<!-- wp:paragraph --> <p>You are using the Themes block template now</p> <!-- /wp:paragraph -->
```

```html
```

### Changelog

> BlockTemplateController: adds the ability to load and manage block template files.


src/Utils/ThemeUtils.php
```diff
@@ -0,0 +1,25 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\Utils;
+
+/**
+ * ThemeUtils class used for custom utility functions for themes.
+ */
+class ThemeUtils {
```
Aljullu: Can we use the [globals from Gutenberg](https://github.com/WordPress/gutenberg/blob/8326c1b70de443e31e27da299b80c35232015cb9/lib/full-site-editing/full-site-editing.php#L8-L24) for this?

src/BlockTemplatesController.php
```diff
@@ -0,0 +1,103 @@
...
+	 * Add the block template objects to be used.
+	 *
+	 * @param array $query_result Array of template objects.
+	 * @return array
+	 */
+	public function add_block_templates( $query_result ) {
+		$template_files = $this->get_block_templates();
+
+		foreach ( $template_files as $template_file ) {
+			$query_result[] = _gutenberg_build_template_result_from_file( $template_file, 'wp_template' );
```
tjcafferkey: What is the significance of the function starting with an underscore (`_`) here?

src/BlockTemplatesController.php
```diff
@@ -0,0 +1,103 @@
...
+	 *
+	 * @var string
+	 */
+	const TEMPLATES_DIR_NAME = 'block-templates';
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->templates_directory = plugin_dir_path( __FILE__ ) . '../templates/' . self::TEMPLATES_DIR_NAME;
```
Aljullu: Would using `__DIR__` here make the code easier to understand? I was a bit confused by the two dots in `../templates/`.
```suggestion
		$this->templates_directory = plugin_dir_path( __DIR__ ) . 'templates/' . self::TEMPLATES_DIR_NAME;
```

src/BlockTemplatesController.php
```diff
@@ -0,0 +1,103 @@
...
+		return $templates;
+	}
+
+	/**
+	 * Check if the theme has a template. So we know if to load our own in or not.
+	 *
+	 * @param string $template_name name of the template file without .html extension e.g. 'single-product'.
+	 * @return boolean
+	 */
+	public function theme_has_template( $template_name ) {
```
Aljullu, line 102: Do you think we should account for child themes that don't have those templates but the parent theme does? Maybe using [`get_template_directory()`](https://developer.wordpress.org/reference/functions/get_stylesheet_directory/#more-information)?

src/BlockTemplatesController.php
Aljullu: I wonder the same. I would assume that means the function shouldn't be considered public. :thinking: Could we check with somebody from the Gutenberg project about this (via Slack or in their repo)?

Aljullu (commented):
This is working as expected and code looks great. I left some inline comments below, but overall it's looking good.

Regarding the usage of functions prepended by an underscore, I agree we should investigate what that means and whether it's safe to use them in our repo or there are chances they will break in the future.

src/BlockTemplatesController.php
getdave: @tjcafferkey pinged me for info here but unfortunately I don't know the answer either.

I would ask in the `#core-editor` channel in Core slack.

tjcafferkey: Just surfacing the reasoning:

The `_` prefix means the function it shouldn't be used publicly and can change at any time. Because of this I've duplicated the functions into this repo and will kick off a discussion around this to understand what is needed, and if they can become available for public usage.

src/BlockTemplatesController.php
```diff
@@ -0,0 +1,105 @@
...
+		return is_readable( get_template_directory() . '/block-templates/' . $template_name . '.html' );
```
Aljullu: I didn't test it and that might be something we can iterate on, but my understanding is that we would need to checks here (one for the parent theme, one for the child theme):
```suggestion
		return is_readable( get_template_directory() . '/block-templates/' . $template_name . '.html' ) ||
			is_readable( get_stylesheet_directory() . '/block-templates/' . $template_name . '.html' );
```

Aljullu (approved):
LGTM, I left one last comment, but besides that, this is good to go. :shipit:

PR #4986, opened 2021-10-25
Cart i2: migrate alignment options

senadir:
Alignment options changed between Cart i1 and Cart i2. In Cart i2, you need to apply alignment in both Cart and filled cart/empty cart to it to be applied in the editor.

This PR migrates that.

### Testing steps
- Checkout an old tag `git checkout v6.0.0`
- Insert Cart i1.
- Change alignment to full or wide.
- Checkout this branch, refresh the cart in editor, you should see it full.
- Check frontend, it should be respected there.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/attributes.js
+++ b/assets/js/blocks/cart-checkout/cart/attributes.js
@@ -27,4 +27,7 @@ export const blockAttributes = {
 		type: 'boolean',
 		default: true,
 	},
+	align: {
+		type: 'string',
+	},
 };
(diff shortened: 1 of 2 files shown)


ralucaStan:
While testing I've noticed that if I switch to the Empty cart state and I open the alignment option, I see Full width selected, and when I click it, the block is switched to the Filled cart state. 
https://www.loom.com/share/f9099a7c4468430aa5a3dba2b22a7c90

senadir:
That alignment option you're selecting is the one that comes with Cart, the one you're looking for is on the right. I will into the view changing issue.

senadir:
After testing, this seems unrelated to this PR and is coming from the view switcher so we merge regardless

ralucaStan (approved):
I've tested and the align attribute is passed along as expected

senadir:
The issue Raluca reported is coming from Gutenberg. Gutenberg unmounts the block when alignment is set/unset causing our viewSwitcher state to reset.

reported here: https://github.com/WordPress/gutenberg/issues/35949

PR #4992, opened 2021-10-26
Add cart item classname filter

senadir:
This PR adds cartItemClass filter, to allow cart items classes to be customized depending on context.

Cherry-picked from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241

### Testing

1. Not much to test given this is a filter, but you can insert this code somewhere.
```js
const { __experimentalRegisterCheckoutFilters } = window.wc.blocksCheckout;
__experimentalRegisterCheckoutFilters( 'my-hypothetical-deposit-plugin', {
	cartItemClass: () => 'my-class',
} );

```
2. You should see `my-class` on all cart items.

### Changelog

> Introduce `cartItemClass` filter for cart items in Cart and Checkout blocks.

### DIFF
--- a/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
+++ b/assets/js/base/components/cart-checkout/order-summary/order-summary-item.js
@@ -1,6 +1,7 @@
 /**
  * External dependencies
  */
+import classnames from 'classnames';
 import { sprintf, _n } from '@wordpress/i18n';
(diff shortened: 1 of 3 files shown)


mikejolley (approved):
This looks ok, but it would be nice to have better inline documentation for arg, and a docblock before filters explaining purpose like we do in PHP.

PR #4994, opened 2021-10-26
Add `register_payment_requirements` documentation

### LINKED ISSUE #4940: Document register_payment_requirements API
We have an API that allows you to filter out payment requirements that don't support a certain feature `register_payment_requirements`, however, this is not documented anywhere, so we should document it.

It should live with the [Filtering payment methods in the Checkout block](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/filtering-payment-methods.md) document.

This is a gist of how to use it https://gist.github.com/senadir/bbc562e8c5b7170e71052696f55dce64

opr:
This PR adds documentation about how `register_payment_requirements` can be used and uses the example of WooCommerce bookings (though simplified somewhat) to do so.

To test this PR please read the documentation and ensure it makes sense from a developer's point of view.

Fixes #4940

### DIFF
--- a/docs/extensibility/filtering-payment-methods.md
+++ b/docs/extensibility/filtering-payment-methods.md
@@ -101,5 +101,86 @@ If you need data that is not available in the parameter received by the callback
 # Filtering payment methods using requirements
 
 ## The problem
-Your extension has registered a payment method that is capable of taking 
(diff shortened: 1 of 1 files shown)


mikejolley (approved):
The example makes sense, spotted no errors. let's ship it 👍🏻

PR #5002, opened 2021-10-26
Fix duplicate queries in product grids #4695

### LINKED ISSUE #4695: Experimental tracking support data for impressions causing slowdown
Product grid appends inject item [data](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/AbstractProductGrid.php#L93) to body, this line is not cached at all, when you have multiple product grids (in my case 7 * 12 ) causes over extra un cachable queries.

mikejolley:
Adds caching (and cache priming) to the product grid to prevent duplicate queries.

Fixes #4695

The issue here was the [`get_variations` method in the `ProductSchema` class](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/961c0c476d4228a218859c658c42f9b6eebfdec4/src/StoreApi/Schemas/ProductSchema.php#L568) queries all product variations per product, and that custom query is uncached. If the same product is queried multiple times (for instance, if showing multiple product grids) the query will repeat. This results in duplicate queries.

To test this, I added 2 product grids to a page, both containing variable products. This is an example from Query Monitor:

Screenshot 2021-10-27 at 12 14 00
Screenshot 2021-10-27 at 12 13 48

Notice (1), there are 2 calls to each set of variations (duplicate queries). This needs fixing with cache.

Notice (2), there is a separate query for each variable product. This is expected, but could be fixed by priming the cache.

This PR implements both. A cache for each products `$variation_meta_data` query (1), and a cache priming function once the results of the product grid query are returned (2). This eliminates duplicate queries, and  reduces the overall total number of queries on the page:

Screenshot 2021-10-27 at 12 16 49

**Testing**

User facing:

1. Add multiple product grid blocks to a page e.g. `Products by Category`. Ensure there are some variable products visible, or add some new ones
2. Check the Grid Block renders correctly on the frontend.

Dev:

1. Install Query Monitor
2. Check the total number of queries for the above scenario before and after this patch
3. Confirm total number of queries is reduced and there are no duplicate queries for this grid block shown in query monitor
4. Make an API request to a variable product by ID, e.g. `https://store.local/wp-json/wc/store/products/11`. Confirm the `variations` section is still populated.

**Changelog**

> Improved performance of Product Grid Blocks and Products Store API endpoint by caching product variations.

### DIFF
--- a/src/BlockTypes/AbstractProductGrid.php
+++ b/src/BlockTypes/AbstractProductGrid.php
@@ -309,9 +309,80 @@ abstract class AbstractProductGrid extends AbstractDynamicBlock {
 		// Prime caches to reduce future queries.
 		if ( is_callable( '_prime_post_caches' ) ) {
 			_prime_post_caches( $results );
+			$this->prime_product_variations( $results );
 		}
 
 		return $results;
+	}
+
+	/**
+	 * Prime query cache of product variation meta data.
+	 *
+	 * Prepares values in the product_ID_variation_meta_data cache for later use in the ProductSchema::get_variations()
+	 * method. Doing so here reduces the total number of queries needed.
+	 *
+	 * @param int[] $product_ids Product ids to prime variation cache for.
+	 */
+	protected function prime_product_variations( $product_ids ) {
+		global $wpdb;
+		static $primed_product_ids = [];
+
+		$prime_product_ids = array_diff( wp_parse_id_list( $product_ids ), $primed_product_ids );
+
+		if ( ! $prime_product_ids ) {
+			return;
+		}
+
+		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
+		$product_variations      = array_map(
+			function( $row ) {
+				$row['product_id']   = absint( $row['product_id'] );
+				$row['variation_id'] = absint( $row['variation_id'] );
+				return $row;
+			},
+			$wpdb->get_results( "SELECT ID as variation_id, post_parent as product_id from {$wpdb->posts} WHERE post_parent IN ( " . implode( ',', $prime_product_ids ) . ' )', ARRAY_A )
+		);
+		$prime_variation_ids     = array_column( $product_variations, 'variation_id' );
+		$variation_ids_by_parent = array_column( $product_variations, 'product_id', 'variation_id' );
+		$all_variation_meta_data = $wpdb->get_results(
+			$wpdb->prepare(
+				"
+					SELECT post_id as variation_id, meta_key as attribute_key, meta_value as attribute_value
+					FROM {$wpdb->postmeta}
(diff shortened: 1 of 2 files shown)


src/BlockTypes/AbstractProductGrid.php
```diff
@@ -306,14 +306,92 @@ protected function get_products() {
...
+		$prime_product_ids = $this->get_non_cached_ids( wp_parse_id_list( $product_ids ), $cache_group );
+
+		if ( ! $prime_product_ids ) {
+			return;
+		}
+
+		global $wpdb;
+
+		// phpcs:disable WordPress.DB.PreparedSQL.NotPrepared
+		$product_variations      = $wpdb->get_results( "SELECT ID as variation_id, post_parent as product_id from {$wpdb->posts} WHERE post_parent IN ( " . implode( ',', $prime_product_ids ) . ' )', ARRAY_A );
```
opr, line 361: Just wondering why `prime_variation_ids` in the next query is run through `esc_sql` but this one isn't.

src/StoreApi/Schemas/ProductSchema.php
```diff
@@ -590,22 +588,51 @@ function( $defaults, $attribute ) use ( $product ) {
...
 		$attributes_by_variation = array_reduce(
 			$variation_meta_data,
-			function( $values, $data ) {
-				$values[ $data->variation_id ][ $data->attribute_key ] = $data->attribute_value;
+			function( $values, $data ) use ( $default_variation_meta_keys ) {
+				// Whilst the query above only includes keys of $default_variation_meta_data, the cache may have been
+				// primed elsewhere.
+				if ( in_array( $data->attribute_key, $default_variation_meta_keys, true ) ) {
+					$values[ $data->variation_id ][ $data->attribute_key ] = $data->attribute_value;
+				}
```
opr: It took me quite a while to understand what this check was for, even in spite of the comment. Could you expand on the comment a little bit to give more context? (Feel free to disagree if you think it's just me, though!)

opr (approved):
This solves the problem and the code looks solid, couple of non-blocking queries though which you can feel free to ignore or action.

src/BlockTypes/AbstractProductGrid.php
mikejolley: This one has been ran through `wp_parse_id_list`. `$prime_variation_ids` is from the results of the first query.

src/StoreApi/Schemas/ProductSchema.php
mikejolley: In the context of this function, we have `$default_variation_meta_keys` so we know all the values are relevant. The cache prime function I wrote however may have other keys outside of `$default_variation_meta_keys` e.g. attributes from over variable products. Does that sort of make sense?

opr: Yeah it does, thanks Mike 👍🏼

PR #5015, opened 2021-10-27
Graduate `__experimental_woocommerce_blocks_checkout_update_order_from_request` action to stable

### LINKED ISSUE #4997: Graduate `__experimental_woocommerce_blocks_checkout_update_order_from_request` PHP action to stable.
This PHP action [was added](https://href.li/?https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4610) a few months ago. It gives extensions the chance to update orders based on the data in the request.

Things to do:

* [ ] Introduce a new action with the experimental prefix removed.
* [ ] Deprecate the old action (so it will still fire, but have it throw a deprecated developer notice pointing to the new name.
* [ ] Update inline code documentation.
* [ ] Update any other documentation.
* [ ] Check this action off in the list

opr:
This PR will deprecate the `__experimental_woocommerce_blocks_checkout_update_order_from_request` action and replace it with `woocommerce_blocks_checkout_update_order_from_request`. 

It also updates the docs relating to this hook. All references to the now deprecated hook have been _replaced_ with the new hook, besides the one in the experimental interfaces document, which has changed to note that the prefixed hook is now deprecated.

Fixes #4997 

### Manual Testing

1. Add the following code somewhere that it'll be executed, for example your theme's `functions.php` file:
```php
add_action( '__experimental_woocommerce_blocks_checkout_update_order_from_request', function( $order, $request ) { return; }, 10, 2 );

add_action( 'woocommerce_blocks_checkout_update_order_from_request', function( $order, $request ) {
	// For testing purposes, set the customer note to the user-agent from the request.
	$order->set_customer_note( $request->get_header('User-agent') );
	$order->save();
}, 10, 2 );
```
2. Add the following directives to your `wp-config.php` file:
```php
define( 'WP_DEBUG', true );
define( 'WP_DEBUG_LOG', true );
```
3. Add a product to your cart and check out using the Checkout block. Ensure it goes through correctly.
4. Check your `debug.log` file and ensure there's a deprecation message logged for `__experimental_woocommerce_blocks_checkout_update_order_from_request`.
5. Ensure the order contains a customer note that is your browser's user agent.

### Changelog

> Deprecate the __experimental_woocommerce_blocks_checkout_update_order_from_request action in favour of woocommerce_blocks_checkout_update_order_from_request.


src/StoreApi/Routes/Checkout.php
```diff
@@ -474,8 +474,33 @@ private function update_order_from_request( \WP_REST_Request $request ) {
...
+		 *
+		 * @deprecated Deprecated since version x.x.x, use woocommerce_blocks_checkout_update_order_from_request instead.
+		 */
+		do_action_deprecated(
+			'__experimental_woocommerce_blocks_checkout_update_order_from_request',
+			array(
+				$this->order,
+				$request,
+			),
+			'x.x.x',
```
mikejolley: Should this target next release?

src/StoreApi/Routes/Checkout.php
```diff
@@ -474,8 +474,33 @@ private function update_order_from_request( \WP_REST_Request $request ) {
 		 *
 		 * @param \WC_Order $order Order object.
 		 * @param \WP_REST_Request $request Full details about the request.
+		 *
+		 * @deprecated Deprecated since version x.x.x, use woocommerce_blocks_checkout_update_order_from_request instead.
```
mikejolley: The syntax for this should be:

```
@deprecated X.X.X Use woocommerce_blocks_checkout_update_order_from_request instead.
```

mikejolley (approved):
Changes look good here aside from missing versions which I assume you're adding before merging. Remember to regen docs too after doing that! Pre-approving with comments.

src/StoreApi/Routes/Checkout.php
opr: Oh, I misunderstood how woorelease works, I thought it would replace all `x.x.x` strings with the current version, but it seems it's only on `@since x.x.x` and `@version x.x.x` strings. I'll update these and all other PRs to 6.3!

src/StoreApi/Routes/Checkout.php
opr: Thanks Mike 👍🏼

PR #5016, opened 2021-10-27
Return an element instead of a component in useViewSwitcher

### LINKED ISSUE #5012: Return ViewSwitcherComponent as an element instead of component
We had a bug in our viewSwitcher in which focus would be lost on scrolling. This happens because the component we use to render the dropdown `ViewSwitcherComponent` ended up being redefined a lot and react reconciler couldn't keep reference of it.

A fix for this it so return a react element instead of a react component.

React Element: `<Element />`
React Component: `Element = () => { ... }`

So instead of having `ViewSwitcherComponent = () => (` we change it to `ViewSwitcherComponent = ( ... )`.
In the edit component, you would reference `{ ViewSwitcherComponent }` instead of `<ViewSwitcherComponent />`
Another fix would be to take `ViewSwitcherComponent` completely outside `useViewSwitcher` and call it like a regular component and pass attributes to it.

https://user-images.githubusercontent.com/6165348/139052151-fbe6c50a-0255-4801-9a06-a95712cc5c5f.mov

senadir:
Returns an element instead of a component in useViewSwitcher.
Returning a component from a hook causes the reconciler to rebuild the component several times, losing reference to it. Returning a element keeps that reference stable.
I also sneaked a couple of fixes:
- Added `isActive` prop to dropdown elements.
- Switched to `ToolbarGroup` because using `Toolbar` without a label is deprecated.

Fixes #5012

### Manual Testing

1. In the empty cart view, open the view switcher. It shouldn't flicker if it loses hover.
2. The currently active view icon should be highlighted.
3. There shouldn't be a console warning about using Toolbar.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/edit.js
+++ b/assets/js/blocks/cart-checkout/cart/edit.js
@@ -157,7 +157,7 @@ export const Edit = ( { className, attributes, setAttributes, clientId } ) => {
 						setAttributes={ setAttributes }
 					/>
 					<BlockControls __experimentalShareWithChildBlocks>
-						<ViewSwitcherComponent />
+						{ ViewSwitcherComponent }
 					</BlockControls>
 					<CartBlockContext.Provider
 						value={ {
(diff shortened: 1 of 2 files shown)


frontdevde (approved):
This tests well for me and fixes the issue we were seeing. Thank you!

Issue #5005, opened 2021-10-27
"By proceeding with your purchase [...]" Not Translated on Checkout Block


igotdes:
## Describe the bug
When using the Checkout block, "By proceeding with your purchase you agree to our Terms and Conditions and Privacy Policy" is not being translated even if translations for it exist.

## To reproduce

1. On WP Admin > Settings, set the Site Language to a language other than English like Español or Norsk bokmål.
2. Make sure to download any translation updates under My Home > Updates
3. Create a page, add the Checkout block to it, and save it
4. Add a product to cart and view the checkout page created previously

## Expected behavior
"By proceeding with your purchase you agree to our Terms and Conditions and Privacy Policy" should be translated.

## Screenshots

Español - [translation here](https://translate.wordpress.org/projects/wp-plugins/woo-gutenberg-products-block/stable/es/default/?filters%5Bterm%5D=By+proceeding+with+your+purchase+you+agree&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Apply+Filters&sort%5Bby%5D=priority&sort%5Bhow%5D=desc)
![Translation-Checkout-Block-ES](https://user-images.githubusercontent.com/74066241/138988466-f70d2651-5507-441f-955a-fb97bc8475ff.png)

Norsk bokmål - [translation here](https://translate.wordpress.org/projects/wp-plugins/woo-gutenberg-products-block/stable/nb/default/?filters%5Bterm%5D=By+proceeding+with+your+purchase+you+agree&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Apply+Filters&sort%5Bby%5D=priority&sort%5Bhow%5D=desc#)
![Translation-Checkout-Block-NO](https://user-images.githubusercontent.com/74066241/138988451-cbc0da34-4e87-4a71-b6c0-aab12945c4f5.png)

### WordPress:
* Core version: 5.8.1
* Gutenberg version: 11.7.1
* WooCommerce version: 5.8.0
* WooCommerce Blocks version: 6.1.0

### Desktop (please complete the following information):
* OS: macOS
* Browser: Chrome Version 95.0.4638.54 (Official Build) (x86_64)

## Additional context
Reported in #4406259-zen.

[...]

PR #5020, opened 2021-10-28
Correct textdomain from 'woo-gutenberg-product-blocks' to 'woo-gutenberg-products-blocks'

nielslange:
Related #5005 

### Note

Some strings were using the textdomain `woo-gutenberg-product-blocks` instead of `woo-gutenberg-products-blocks` (note the missing `s` for products). Even though these strings have been translated, due to the incorrect textdomain they aren't visible in the plugin.

### Testing instructions

1. Check out the `trunk`.
2. Search for the incorrect text domain `'woo-gutenberg-product-blocks'`.
3. Verify that you get some results.
4. Check out this PR.
5. Search for the incorrect text domain `'woo-gutenberg-product-blocks'` again.
6. Verify that you no longer get results for the incorrect text domain.

### Additional notes

Issue #5005 is not entierly solved by this PR, but require a follow up PR to lazy load some missing translation files on the frontend.

### Changelog

> Replace incorrect with correct text domain.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/constants.js
+++ b/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/constants.js
@@ -7,22 +7,22 @@ import { PRIVACY_URL, TERMS_URL } from '@woocommerce/block-settings';
 const termsPageLink = TERMS_URL
 	? `<a href="${ TERMS_URL }">${ __(
 			'Terms and Conditions',
-			'woo-gutenberg-product-blocks'
+			'woo-gutenberg-products-blocks'
 	  ) }</a>`
-	: __( 'Terms and Conditions', 'woo-gutenberg-product-blocks' );
+	: __( 'Terms and Conditions', 'woo-gutenberg-products-blocks' );
 
 const privacyPageLink = PRIVACY_URL
 	? `<a href="${ PRIVACY_URL }">${ __(
 			'Privacy Policy',
-			'woo-gutenberg-product-blocks'
+			'woo-gutenberg-products-blocks'
 	  ) }</a>`
-	: __( 'Privacy Policy', 'woo-gutenberg-product-blocks' );
+	: __( 'Privacy Policy', 'woo-gutenberg-products-blocks' );
 
 export const termsConsentDefaultText = sprintf(
 	/* translators: %1$s terms page link, %2$s privacy page link. */
 	__(
 		'By proceeding with your purchase you agree to our %1$s and %2$s',
-		'woo-gutenberg-product-blocks'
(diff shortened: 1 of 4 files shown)


opr (changes requested):
Hi @nielslange looks like you've kept the `s` on `blocks`, however, it should read `woo-gutenberg-products-block`.

Could you also add more robust testing instructions for this PR since the current ones don't actually show the issue being fixed. Something like:

1. Add the terms and conditions block to your Checkout block.
3. Change your site language to Spanish and go to the Checkout block.
3. Verify the text in the terms and conditions block is translated.

ralucaStan:
I also recommend changing the changelog entry to something like _Fix missing translations in the Checkout block that were using an incorrect textdomain_. This provides more details and is user-friendly

nielslange:
> Hi @nielslange looks like you've kept the s on blocks, however, it should read woo-gutenberg-products-block.

Good catch, @opr!

> Could you also add more robust testing instructions for this PR since the current ones don't actually show the issue being fixed.

Absolutely. Due to my incorrect fix, I could not see the change initially. That's why I haven't pointed out the testing instructions that you mentioned. I'll correct the testing instructions directly.

nielslange:
> I also recommend changing the changelog entry to something like Fix missing translations that were using an incorrect textdomain. This provides more details and is user-friendly

Great suggestion, @ralucaStan. I've updated the changelog message accordingly. You could now review this PR.

opr:
@nielslange on the block in the editor I see the translated string, but not on the front-end.

| Editor | Front-end |
|---|---|
| image | image |

Any ideas why not?

nielslange:
> Any ideas why not?

@opr I didn't notice that initially. I just looked up this problem and saw that both the editor and the frontend are fetching the translation from the [constants.js](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/constants.js) file. The translations for the editor and the frontend are coming from this part:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/becd87b67103370d1320ddfcdb577b4b909d3e96/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/constants.js#L21-L39

When debugging these parts, I do see the translated version in the editor, but the untranslated version in the frontend. However, the i18n functionality is enqueued in the frontend. I'll investigate what is causing this problem.

senadir (approved):
This does as advertises, which is fix all instances of wrong text domain.
Changelog and testing steps should be updated to reflect the new changes.

nielslange:
@senadir Thanks for your quick review!

@opr I found out that some strings are not visible on the frontend, due to lazy loading. I'll take care of that in a separate PR.

PR #5022, opened 2021-10-28
Remove hydration hocs in favour of apiFetch Middlewares

mikejolley:
Gutenberg provides an apiFetch middleware called `createPreloadingMiddleware` which accepts REST API data. This is very very similar to the hydration code we implemented, but more straightforward to use. 

This PR passes endpoint data to `createPreloadingMiddleware` and removes our custom HOC implementation. This also removes the need to handle stale cart data on page load (implemented in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3960) because `createPreloadingMiddleware` does not exhibit the same problem.

**Testing**

1. Go to the cart block page. Check there were no extra `fetch` requests to `wc/store/cart` in the network inspector.
2. Go to the checkout block page. Ensure there is no error and checkout works as normal.

Testing stale data:

- Go to cart block
- Change cart item quantities. Take note of the amounts.
- Browse to a different page
- Use the browser back button.
- Confirm amounts are still correct and not stale.

@opr confirmed this removes the need for the fix in #3960 
@Aljullu could also use your eyes since it touched a few parts of the mini-cart block


src/Assets/AssetDataRegistry.php
```diff
@@ -339,12 +343,16 @@ public function enqueue_asset_data() {
...
+			$preloaded_api_requests = rawurlencode( wp_json_encode( $this->preloaded_api_requests ) );
+
 			wp_add_inline_script(
 				$this->handle,
-				"var wcSettings = wcSettings || JSON.parse( decodeURIComponent( '"
-					. esc_js( $data )
-					. "' ) );",
+				"
+				var wcSettings = wcSettings || JSON.parse( decodeURIComponent( '" . esc_js( $data ) . "' ) );
+				wp.apiFetch.use( wp.apiFetch.createPreloadingMiddleware( JSON.parse( decodeURIComponent( '" . esc_js( $preloaded_api_requests ) . "' ) ) ) )
```
mikejolley, line 354: createPreloadingMiddleware does the job our hocs were doing previously.

src/BlockTypes/Checkout.php
```diff
@@ -332,13 +332,16 @@ protected function hydrate_customer_payment_methods() {
...
+
 		// Print existing notices now, otherwise they are caught by the Cart
 		// Controller and converted to exceptions.
 		wc_print_notices();
-
 		add_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
-		$this->asset_data_registry->hydrate_api_request( '/wc/store/cart' );
-		$this->asset_data_registry->hydrate_api_request( '/wc/store/checkout' );
+
+		$rest_preload_api_requests = rest_preload_api_request( [], '/wc/store/checkout' );
```
mikejolley, line 342: Checkout still needs manual preloading data because it doesn't go via wp.data. It should, but doesn't yet.

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -377,7 +371,6 @@ public function get_item_response( $cart ) {
 			),
 			'errors'                  => $cart_errors,
 			'payment_requirements'    => $this->extend->get_payment_requirements(),
-			'generated_timestamp'     => time(),
```
mikejolley, line 380: This should have been a header, like the product collection `Last-Modified` header handling. But I found it wasn't needed anyway...

mikejolley:
@senadir you might want a look also

opr:
Hi @mikejolley looks like the changes implemented in #3960 _aren't_ needed anymore, I tested it and it works, the newest cart is always fetched when navigating back and forward using the browser buttons 🙌🏼 

I like this approach as it simplifies our codebase and is easier to reason with.

assets/js/base/context/providers/cart-checkout/checkout-state/constants.ts
```diff
@@ -26,15 +26,15 @@ export enum STATUS {
...
 	string,
-	{ body: Record< string, unknown > }
+	unknown
 >;
 
 const checkoutData = {
 	order_id: 0,
 	customer_id: 0,
-	...( preloadedApiRequests[ '/wc/store/checkout' ]?.body || {} ),
+	...( preloadedCheckoutData || {} ),
```
mikejolley, line 37: Checkout requires preloading still - it uses getSetting and does not have a  wp data store like the cart does.

assets/js/blocks/cart-checkout/mini-cart/with-mini-cart-conditional-hydration.tsx
```diff
@@ -1,35 +0,0 @@
-/**
```
senadir, line 1: I'd prefer if @Aljullu take a look at this.

tests/e2e/specs/frontend/cart.test.js
```diff
@@ -1,195 +0,0 @@
-/**
```
senadir, line 1: Why was this deleted?

senadir (approved):
Nothing like the smell of freshly deleted code. However, I think you deleted a bit too much with `tests/e2e/specs/frontend/cart.test.js`?

I'm pre-approving on the basis that this tests fine, and the code looks good, but I will wait for a review from @Aljullu to confirm the minicart related part.

tests/e2e/specs/frontend/cart.test.js
opr: This test file was only testing the code relating to reloading a "stale" cart - since this PR removes this code, and the possibility for carts to go stale the tests are redundant.

senadir: Lovely!

assets/js/blocks/cart-checkout/mini-cart/with-mini-cart-conditional-hydration.tsx
Aljullu: Mini Cart block changes LGTM. :ok_hand:

mikejolley:
@senadir the tests I removed were for the localstorage timestamp that was removed.

PR #5025, opened 2021-10-28
Mini Cart as template part

Aljullu:
This PR updates how the Mini Cart block works so contents shown inside the drawer belong to a template part. That allows themes to override it and users to edit it from the Template Parts editor.

In summary, the changes are:

* Split the Mini Cart block into two:
  * Mini Cart: displays the button and the drawer with the template part inside it.
  * Mini Cart contents: the main block of the template part, it renders the contents of the drawer.
* Create a Mini Cart template part.
* Updates `BlockTemplatesController` so it can load block template parts alongside block templates.

### Testing

To make it easier to test, checkout this branch from Gutenberg so you can edit and see the template parts added by WooCommerce: https://github.com/WordPress/gutenberg/pull/36379.

Also, make sure to test with a block theme, ie: [Quadrat](https://wordpress.org/themes/quadrat/) or [Tove](https://wordpress.org/themes/tove/).

#### WooCommerce template part

1. Add the Mini Cart block to a page and verify you can open it. Once open, it should show the contents.
2. Go to Appearance > Editor and navigate (using the left sidebar) to the Template Parts screen.
3. Verify a Mini Cart template part is listed.

imatge

#### Theme template part

1. Create a `/block-template-parts/mini-cart.html` file in your theme directory with these contents:

```HTML
<!-- wp:paragraph -->
<p>This is the theme template part.</p>
<!-- /wp:paragraph -->
<!-- wp:woocommerce/mini-cart-contents /-->
```

2. Go back to the page you created before and open the Mini Cart drawer.
3. Verify the `This is the theme template part` text appears in the drawer.

#### User template part

1. Go to Appearance > Editor and navigate (using the left sidebar) to the Template Parts screen.
2. Open the `mini-cart` template and edit its contents (ie, replacing `This is the theme template part.` with `This is the user edited template part.`).
3. Go back to the page with the Mini Cart block. Open the Mini Cart and verify the template you just edited shows up.

#### Classic theme

1. Switch your theme to Storefront (or any other classic theme).
2. Verify you can open the Mini Cart and the default template shows up.


assets/js/base/context/hooks/cart/use-store-cart-event-listeners.ts
```diff
@@ -36,7 +36,7 @@ const setUp = (): void => {
 const addListeners = (): void => {
 	setUp();
 
-	if ( ! window.wcBlocksStoreCartListeners.count ) {
+	if ( window.wcBlocksStoreCartListeners.count === 0 ) {
```
Aljullu, line 39: This change is not directly related to this PR, but still wanted to add it. :slightly_smiling_face: 

`count` is going to be an integer, so it's more precise to compare it to `0` than to check if it's falsy.

dinhtungdu (commented):
This is working great on my end. I just have some minor comments:

- Because we support the template parts now, we need to pass the template type to line [this](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7dacb44d2a0ad853cf95fb14b9444f37936a90e1/src/BlockTemplatesController.php#L177) and [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7dacb44d2a0ad853cf95fb14b9444f37936a90e1/src/BlockTemplatesController.php#L340). Without this change, the edit link of the Mini cart content template is broken because of wrong template type.
- I got a warning about the missing `area` property of the template part: `PHP Warning:  Undefined property: WP_Block_Template::$area in /wp-content/plugins/gutenberg/lib/compat/wordpress-5.9/class-gutenberg-rest-templates-controller.php on line 418`.

Aljullu:
Thanks for the review @dinhtungdu!

> - Because we support the template parts now, we need to pass the template type to line [this](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7dacb44d2a0ad853cf95fb14b9444f37936a90e1/src/BlockTemplatesController.php#L177) and [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7dacb44d2a0ad853cf95fb14b9444f37936a90e1/src/BlockTemplatesController.php#L340). Without this change, the edit link of the Mini cart content template is broken because of wrong template type.

Good catch! Should be fixed in a730af0ba4a60d7a336a128c1ccd0c48a6494313.

>- I got a warning about the missing `area` property of the template part: `PHP Warning:  Undefined property: WP_Block_Template::$area in /wp-content/plugins/gutenberg/lib/compat/wordpress-5.9/class-gutenberg-rest-templates-controller.php on line 418`.

For now, I set it to `uncategorized` in 63222ba32410503e6c05251ae73fe6378e738249, but wondering if we should create a _Mini Cart_ specific area. What do you think?

I think this issue (https://github.com/WordPress/gutenberg/issues/35989) is blocking us, but might be good to take a decision anyway so we can create an issue in our repo and follow-up based on it.

dinhtungdu:
> For now, I set it to uncategorized in 63222ba, but wondering if we should create a Mini Cart specific area. What do you think?

@Aljullu  At the first glance, it seems that the Mini Cart only needs one template part because we only allow one mini cart per page. But given that we have multiple header template parts, and Mini Cart is "usually" added to the header, the cart content style should match with the header (and page style), to do that, we need an area for Mini Cart. So for your question, my answer is yes, we should create a Mini Cart area.

dinhtungdu:
Note that with the latest changes merged to `trunk`, we can create template parts for custom areas. All custom templates created for any custom areas are displayed correctly as well, so I think the https://github.com/WordPress/gutenberg/issues/35989 can be marked as fixed.

image

Aljullu:
TIL, thanks for letting me know @dinhtungdu! I created three follow-up issues based on that: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5194, https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5195 and https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5196. Could you review them and let me know if they make sense to you and look like the correct next steps?

Regarding this PR, what do you think about merging it in the way it currently works and handling anything else in follow-ups?

src/Utils/BlockTemplateUtils.php
```diff
@@ -109,6 +109,9 @@ public static function gutenberg_build_template_result_from_post( $post ) {
 		$template->has_theme_file = $has_theme_file;
 		$template->is_custom      = true;
 		$template->post_types     = array(); // Don't appear in any Edit Post template selector dropdown.
+		if ( 'wp_template_part' === $post->post_type ) {
+			$template->area = 'uncategorized';
+		}
```
dinhtungdu: I think we should get the template area from the database instead of hard-coding it like this. We can get the area of a template part using `get_the_terms( $post, 'wp_template_part_area' );`

src/Utils/BlockTemplateUtils.php
```diff
@@ -137,6 +140,9 @@ public static function gutenberg_build_template_result_from_file( $template_file
 		$template->has_theme_file = true;
 		$template->is_custom      = false; // Templates loaded from the filesystem aren't custom, ones that have been edited and loaded from the DB are.
 		$template->post_types     = array(); // Don't appear in any Edit Post template selector dropdown.
+		if ( 'wp_template_part' === $template_type ) {
+			$template->area = 'uncategorized';
+		}
```
dinhtungdu: With this code, all template parts will have `area` set to `uncategorized`. This is fine for now. But I think we need to add a todo here to improve this in the future. Each template part may have a different area, and we can filter that `area` base on the `slug` IMO.

dinhtungdu (commented):
@Aljullu This is working great on my end, I don't get any PHP error or warning 👍🏽 . I have some minor comments btw, please check the review comment for more detail.

dinhtungdu:
> I created three follow-up issues based on that: #5194, #5195 and #5196. Could you review them and let me know if they make sense to you and look like the correct next steps?

@Aljullu The follow-up issues make a lot of sense to me! The #5196 is actually a very nice UX enhancement! 🎉 

> Regarding this PR, what do you think about merging it in the way it currently works and handling anything else in follow-ups?

I think we should fix https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5025#discussion_r753078689 and leave https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5025#discussion_r753088367 for the next release.

Aljullu:
Thanks for taking another look @dinhtungdu!

> I think we should fix https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5025#discussion_r753078689 and leave https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5025#discussion_r753088367 for the next release. 

Actually, in 5ee58d2142d78592be2256fe50aaf288ca633f90 I tried to fix it in both places, but happy to undo the second change if you are not comfortable with it. In any case, I think that's something that we can keep iterating on in the follow-up issues.

src/BlockTemplatesController.php
```diff
@@ -278,16 +294,25 @@ function( $saved_woo_template ) {
...
+		$directory      = $this->get_templates_directory( $template_type );
+		$template_files = BlockTemplateUtils::gutenberg_get_template_paths( $directory );
 		$templates      = array();
 		foreach ( $template_files as $template_file ) {
+
+			if ( 'wp_template_part' === $template_type ) {
+				$dir_name = self::TEMPLATE_PARTS_DIR_NAME;
+			} else {
+				$dir_name = self::TEMPLATES_DIR_NAME;
+			}
```
dinhtungdu: We should move this check outside of the `foreach` loop.

dinhtungdu (approved):
>Actually, in 5ee58d2 I tried to fix it in both places, but happy to undo the second change if you are not comfortable with it. In any case, I think that's something that we can keep iterating on in the follow-up issues.

@Aljullu https://github.com/woocommerce/woocommerce-gutenberg-products-block/commit/5ee58d2142d78592be2256fe50aaf288ca633f90 looks good to me. Initially, I think it's better to extract the logic for the template part area into its own function (under the `BlockTemplateUtils` class). But given that this PR is already large, and we haven't known if we need other template part areas yet. So I think we so keep it as it is for now, and improve it later if we actually have so many template part areas.

src/BlockTemplatesController.php
Aljullu: Fixed in 51930edda5955e95497df145add49d629d89c26d.

PR #5027, opened 2021-10-28
Run BlockTemplatesController filter only on templates

Aljullu:
The `get_block_templates` filter can be run either on `wp_template` and `wp_template_part`, however, our implementation on `BlockTemplatesController` only needs to filter templates. This PR adds a check so the logic is not run when the filter is called for template parts.

### Manual Testing

Verify there are no regressions from #4981 (testing steps copied):

1. Ensure you're using a block enabled theme
2. Add the **second** template below to the Woo Blocks path specified above the snippet.
3. Confirm this loads in the Site Editor templates, it should say in the editor "You are using the Woo Blocks block template now"
4. Add the **first** template below to themes directory path specified above the snippet.
5. Confirm this loads in the Site Editor templates, it should say in the editor "You are using the Themes block template now"
6. This code is not yet responsible for loading the templates on the frontend. This will be done as part of https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4969

`/theme-dir/block-templates/single-product.html`
```html
<!-- wp:paragraph --> <p>You are using the Themes block template now</p> <!-- /wp:paragraph -->
```

```html
```

### DIFF
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -43,10 +43,12 @@ class BlockTemplatesController {
 	 * Add the block template objects to be used.
 	 *
 	 * @param array $query_result Array of template objects.
+	 * @param array $query Optional. Arguments to retrieve templates.
+	 * @param array $template_type wp_template or wp_template_part.
 	 * @return array
 	 */
-	public function add_block_templates( $query_result ) {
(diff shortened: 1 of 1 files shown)


dinhtungdu (approved):
@Aljullu Look good to me 🎉 !

PR #5031, opened 2021-10-28
Some PR template improvements

nerrad:
While I was hoping to use the new [GitHub issue form syntax](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) for more precise template setup (including forms!), it's not available yet for pull request templates. 

The changes I've made here include:

- Adding an "Other Checks" section to split off from the accessibility section.
- Adding a checkbox for reviews that have two reviewers assigned because of the potential security impact of the changes.
- Added a section for User testing notes if that needs to be different from the other testing notes in scenarios where we want to account for non-developer type testing. This will be useful for inclusion with what gets included in the testing notes saved to the docs folder. This came up in some team discussion around reducing release lead friction during the release process.

This has no impact on production code so mostly just needs a grammar check and any feedback on whether these changes make sense or not.


opr (approved):
Nice 👍🏼  I am very fond of the two reviewers for database related changes checklist item!

PR #5037, opened 2021-10-28
Add minimum quantity and step so Cart Block and Store API.

senadir:
This PR adds `minimum` and `step` concepts to Store API and Cart Block.

This functionality is still mainly plugin land, so we're not actually implementing a way for a merchant to setup a step or a minimum amount, but we're laying the foundation for plugins to use this, in Store API, via two new filters:

- `woocommerce_store_api_item_quantity_minimum`
- `woocommerce_store_api_item_quantity_step`

Following on the steps of `woocommerce_store_api_product_quantity_limit`.

The second addition is honoring those values in Cart Block. Cart Item quantity switcher now honors minimum and step values and would change according to them.

Based on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/66b053424c11ca310069ec3a0883884a3e541c82

### Testing steps
1. Smoke test adding items to cart and changing quantity, either via buttons or directly via the input.
2. Keep you network tap open and make sure no extra requests are being made.
3. Add this code, which would apply to your cart items:
```php
add_filter(
	'woocommerce_store_api_item_quantity_minimum',
	function ( $value, $cart_item ) {
		return 6;
	},
	10,
	2
);

add_filter(
	'woocommerce_store_api_item_quantity_step',
	function ( $value, $cart_item ) {
		return 4;
	},
	10,
	2
);
```

4. Substitute 6 and 4 with whatever value you want.
5. Refresh your cart, change values, quantity should up and down by 4 (or whatever value you set for step).
6. If the initial value isn't a multiplier of 4, after changing the quantity, your new quantity would be a multiplier of 4.
7. Try to type a number inside the quantity selector, trying something that isn't a multiplier of 4 would change to the nearest one, for example, typing 14 would yield back 12.

### Caveats:
- I opted out for filters instead of a registration API. the surface area is very minimal (just a number).
- I decided to keep this plugin land. This means there's no verification on Store API about minimum quantity and if it follows a step or not. We do have verification for maximum quantity, but that's because maximum quantity is a core feature, handled via sold individually and stock attributes.
- @mikejolley and discussed not implementing the step part, and let plugins handle that on server side. This reduces the amount of logic we have, but I believe results in a subpar UI experience overall. I'd prefer, if we can, to handle that on our part.
- 
### Changelog

> Store API and Cart block now support defining a quantity stepper and a minimum quantity.

### DIFF
--- a/assets/js/base/components/quantity-selector/index.tsx
+++ b/assets/js/base/components/quantity-selector/index.tsx
@@ -6,6 +6,7 @@ import { speak } from '@wordpress/a11y';
 import classNames from 'classnames';
 import { useCallback } from '@wordpress/element';
 import { DOWN, UP } from '@wordpress/keycodes';
+import useDebouncedCallback from 'use-debounce/lib/useDebouncedCallback';
 
 /**
  * Internal dependencies
@@ -17,6 +18,7 @@ interface QuantitySelectorProps {
 	quantity?: number;
 	minimum?: number;
 	maximum: number;
+	step?: number;
 	onChange: ( newQuantity: number ) => void;
 	itemName?: string;
 	disabled: boolean;
@@ -27,6 +29,7 @@ const QuantitySelector = ( {
 	quantity = 1,
 	minimum = 1,
 	maximum,
+	step = 1,
 	onChange = () => {
 		/* Do nothing. */
 	},
@@ -39,8 +42,46 @@ const QuantitySelector = ( {
 	);
 
 	const hasMaximum = typeof maximum !== 'undefined';
-	const canDecrease = quantity > minimum;
-	const canIncrease = ! hasMaximum || quantity < maximum;
+	const canDecrease = quantity - step > minimum;
+	const canIncrease = ! hasMaximum || quantity + step < maximum;
+
+	/**
+	 * The goal of this function is to normalize what was inserted,
+	 * but after the customer has stopped typing.
+	 *
+	 * It's important to wait before normalizing or we end up with
+	 * a frustrating expiernce, for example, if the minimun is 2 and
+	 * the customer is trying to type "10", premature normalizing would
+	 * always kick in at "1" and turn that into 2.
+	 */
+	const [ normalizeQuantity ] = useDebouncedCallback(
+		( initialValue: number ) => {
+			// We copy the starting value.
+			let value = initialValue;
+
+			// We check if we have a maximum value, and select the lowest between what was inserted, and the maximum.
+			if ( hasMaximum ) {
+				value = Math.min(
+					value,
+					// the maximum possible value in step increments.
+					Math.floor( maximum / step ) * step
+				);
+			}
+
+			// Select the biggest between what's inserted, the the minimum value in steps.
+			value = Math.max( value, Math.ceil( minimum / step ) * step );
+
+			// We round off the value to our steps.
+			value = Math.floor( value / step ) * step;
+
+			// Only commit if the value has changed
+			if ( value !== initialValue ) {
+				onChange( value );
+			}
+		},
+		// This value is delibrtly smaller than what's in useStoreCartItemQuantity so we don't end up with two requests.
+		300
+	);
 
 	/**
 	 * Handles keyboard up and down keys to change quantity value.
@@ -60,15 +101,15 @@ const QuantitySelector = ( {
 
 			if ( isArrowDown && canDecrease ) {
 				event.preventDefault();
-				onChange( quantity - 1 );
+				onChange( quantity - step );
 			}
 
 			if ( isArrowUp && canIncrease ) {
 				event.preventDefault();
-				onChange( quantity + 1 );
+				onChange( quantity + step );
 			}
 		},
-		[ quantity, onChange, canIncrease, canDecrease ]
+		[ quantity, onChange, canIncrease, canDecrease, step ]
 	);
 
 	return (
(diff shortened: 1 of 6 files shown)


assets/js/base/components/quantity-selector/index.tsx
```diff
@@ -40,8 +42,46 @@ const QuantitySelector = ( {
...
-	const canIncrease = ! hasMaximum || quantity < maximum;
+	const canDecrease = quantity - step > minimum;
+	const canIncrease = ! hasMaximum || quantity + step < maximum;
+
+	/**
+	 * The goal of this function is to normalize what was inserted,
+	 * but after the customer has stopped typing.
+	 *
+	 * It's important to wait before normalizing or we end up with
+	 * a frustrating expiernce, for example, if the minimun is 2 and
```
mikejolley:
```suggestion
	 * a frustrating experience, for example, if the minimum is 2 and
```

assets/js/base/components/quantity-selector/index.tsx
```diff
@@ -78,21 +118,22 @@ const QuantitySelector = ( {
...
-					}
-					value = Math.max( value, minimum );
+					// Inputs values are strings, we parse them here.
+					let value = parseInt( event.target.value, 10 );
+					// parseInt would throw NaN for anything not a number,
+					// so we revert value to the quantity value.
+					value = isNaN( value ) ? quantity : value;
+
 					if ( value !== quantity ) {
+						// we commit this value immideatly.
```
mikejolley:
```suggestion
						// we commit this value immediately.
```

src/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -55,6 +55,18 @@ public function get_properties() {
 				'context'     => [ 'view', 'edit' ],
 				'readonly'    => true,
 			],
+			'quantity_min'         => [
+				'description' => __( 'The minimum quantity than can be added to the cart at once.', 'woo-gutenberg-products-block' ),
+				'type'        => 'integer',
+				'context'     => [ 'view', 'edit' ],
+				'readonly'    => true,
+			],
+			'quantity_step'        => [
```
mikejolley, line 66: I'm still on the fence about this property. You could argue for min and max (limit) in the API because it's something that can be validated against on add to cart requests, but the step is a UI restriction, not something you're using in the API. Also `step` is specific to HTML number inputs. Does it make sense here?

We could use `quantity_increment` and validate that the amount passed is an increment of that. e.g. increment of 2 would prevent a qty of 1 being accepted.

Thoughts? Too much?

Do you also think we should group these properties like `add_to_cart`?

src/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -450,6 +464,38 @@ protected function get_item_data( $cart_item ) {
...
+	 * @return number
+	 */
+	protected function get_item_quantity_min( $cart_item ) {
+		/**
+		 * Filters the quantity minimum for a cart item in Store API.
+		 *
+		 * @param array $cart_item Cart item array.
+		 * @return number
+		 */
+		return apply_filters( 'woocommerce_store_api_item_quantity_minimum', 1, $cart_item );
```
mikejolley, line 481: These filters use $cart_item, `woocommerce_store_api_product_quantity_limit` uses $product. Which is correct?

mikejolley: I see the other is product endpoint only. We should make these filters consistent.

manospsyx:
Hey @senadir -- I have had a chance to quickly set up an example and do some tests.

One issue I noticed is that once I increase the cart item quantity to 2 or higher, [I am no longer able to set it back to 1](https://share.getcloudapp.com/6quvorKY).

On another note, I tried to continue where we left things off in [this PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pulls?page=2&q=is%3Apr+is%3Aopen) and have the following to share:

- The new API properties are a massive help from an integration perspective. I echo @mikejolley's [feedback on the step property](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5037#discussion_r745735099); my impression is that making the step property part of the API optimizes for providing a single extensibility point for those 3 related "properties". Whether we'd like to optimize for that is up for discussion.
- These changes do not address the primary issue I attempted to tackle [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/81486f330d0518a09d996d75872c505e7ee37e4d) and [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241/commits/f35447c052072c3343c3c605b9aef6f5b44df7ca), which is the failure to keep cart item quantities in sync.

Here's some code you can use to play around with this state-related problem:

```
add_action( 'woocommerce_after_cart_item_quantity_update', function( $cart_item_key, $quantity = 0 ) {

	if ( isset( WC()->cart->is_syncing ) && WC()->cart->is_syncing ) {
		return;
	}

	if ( ! empty( WC()->cart->cart_contents[ $cart_item_key ] ) ) {

		$cart_item = WC()->cart->cart_contents[ $cart_item_key ];

		if ( ! $quantity || $quantity < 0 ) {
			$quantity = 0;
		} else {
			$quantity = $cart_item[ 'quantity' ];
		}

		if ( in_array( $cart_item[ 'data' ]->get_name(), array( 'Hoodie', 'Belt' ) ) ) {

			foreach ( WC()->cart->cart_contents as $search_item_key => $search_item ) {

				if ( 'Belt' === $search_item[ 'data' ]->get_name() ) {

					if ( 'Hoodie' !== $cart_item[ 'data' ]->get_name() ) {
						continue;
					}

				} elseif ( 'Hoodie' === $search_item[ 'data' ]->get_name() ) {

					if ( 'Belt' !== $cart_item[ 'data' ]->get_name() ) {
						continue;
					}

				} else {
					continue;
				}

				WC()->cart->is_syncing = true;
				WC()->cart->set_quantity( $search_item_key, $quantity, false );
				WC()->cart->is_syncing = false;
			}

		}
	}
}, 10, 2 );
```

What this code does, in a nutshell: Whenever you change the quantity of an item called "Hoodie", it searches for an item called "Belt", and tries to keep the quantity of the latter in sync with the former. Here's how this works in the legacy cart:

https://user-images.githubusercontent.com/1783726/143051125-e9eff4d8-1897-45e2-ac7a-5c629f864b1b.mp4

And here's what's happening in the block cart:

https://user-images.githubusercontent.com/1783726/143051615-1b3f9cf0-6c88-468d-9d6a-d3afe71e487d.mp4

Note that I need to refresh to get the qty of the "synced" item to update, although it's coming back just fine through the API. 

I'm not sure if this needs to be addressed in this PR (probably not?) but it's still a roadblock, and one that we'll need to overcome to make the new Cart/Checkout Blocks compatible with Force Sells, Product Bundles, Composite Products, Mix and Match Products, and likely other extensions that rely on the `woocommerce_after_cart_item_quantity_update` filter to keep cart item quantities in sync.

senadir:
> One issue I noticed is that once I increase the cart item quantity to 2 or higher, I am no longer able to set it back to 1.

Yes, I noticed this as well, seems like a bug.

> These changes do not address the primary issue I attempted to tackle here and here, which is the failure to keep cart item quantities in sync.

Keeping cart in sync is working, you can see the item subtotal changing, there's a small bug you can see there around reinitaiting state, that needs to be fixed, probably in a another PR.

manospsyx:
> that needs to be fixed, probably in a another PR

Yup, that's probably a separate issue -- definitely a blocking one for what we're trying to achieve! Will open a new issue with the description from the last comment. Thanks!!!

senadir:
Note that work on this was moved to #5406. I will be closing this in favor of that one.

PR #5038, opened 2021-10-29
Updated has_block_template hook to use correct woocommerce_ prefix

tjcafferkey:
Updated the hook name so its correctly prefixed with `woocommerce_` as per this request https://github.com/woocommerce/woocommerce/pull/30997#discussion_r738552712

### Testing

Please follow testing instructions here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4984


frontdevde (approved):
LGTM

Issue #5044, opened 2021-11-01
Integrate AutomateWoo and MailPoet with Checkout Block


mikejolley:
This epic tracks the development of an integration between the Checkout Block, AutomateWoo, and MailPoet.

This currently assumes we're building read-only, non-editable, forced blocks. Estimates will need revising if we add block-level settings to change this.

**Shared**

- [ ] Shared components (newsletter, or checkbox/label combo, like terms and conditions) TBD (3)

**AutomateWoo**

- [x] Build process (5)
- [x] Block creation (2)
- [x] Settings Registration (1)
- [x] Store API Integration (3)

**MailPoet**

- [x] Build process (5)
- [x] Block creation (2)
- [x] Settings Registration (1)
- [x] Store API Integration (3)

PR #5045, opened 2021-11-01
Move checkbox control to checkout package

mikejolley:
This PR moves the `CheckboxControl` component, which is only used in the cart/checkout blocks, to the Checkout package. This will allow extensions to consume it. We intend to use this in the Newsletter plugin integrations. 

This is part of #5044

### Testing

This only requires a smoke test; successful build, and test a block can still use the component, for example, terms and conditions block in Checkout.


src/Assets/AssetDataRegistry.php
```diff
@@ -60,8 +60,8 @@ public function __construct( Api $asset_api ) {
 	 */
 	protected function init() {
 		add_action( 'init', array( $this, 'register_data_script' ) );
+		add_action( 'wp_print_scripts', array( $this, 'enqueue_asset_data' ), 1 );
 		add_action( 'wp_print_footer_scripts', array( $this, 'enqueue_asset_data' ), 1 );
-		add_action( 'admin_print_footer_scripts', array( $this, 'enqueue_asset_data' ), 1 );
```
nerrad: Is there possibility for double printed data with this change? I see that you've added a `! wp_script_is( $this->handle, 'done' )` later in `enqueue_asset_data` but if that was intended to prevent double printed data, that will only work in the admin right?  

Regarding this change impacting the admin, there are at least 2 Woo extensions (but likely more) that I know of that are relying on data exposed on `wcSettings`. This change will need tested with those:

- WC-Admin.
- AutomateWoo.

What happens in the admin if something registers data _after_ `wp_print_scripts` fires in the header? Will it still show up for the view?

Due to the impact, it might be wise to do this in it's own PR so the change has more visibility.

mikejolley: I'm happy to log this and let someone else work on it. I just needed to be unblocked to get MailPoet working locally.

mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5052

senadir:
I will be breaking this PR up to 3 PRs since it does 3 distinct things along with a 4 PR I have.

mikejolley:
@senadir rebased with the single change, ready for review.

senadir (approved):

senadir:
Merging after tests pass.

PR #5048, opened 2021-11-02
Fix Block Templates being rendered correctly

### LINKED ISSUE #5046: Ensure the correct template is loading on the Frontend and the site editor
Today we discovered that we're not loading the correct template on the Frontend as well as the site editor. This has potentially not surfaced before as we were testing with only one template in the template directory.

The goal of this issue is to ensure the correct template is loading on the Frontend and in the site editor and we'll use this issue to track the time spent investigating and resolving the bug we're seeing.

tjcafferkey:
Copy logic from [Gutenberg](https://github.com/WordPress/gutenberg/blob/trunk/lib/full-site-editing/block-templates.php#L417-L445) to ensure the correct templates are being included in the get_block_templates query

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5046

### Manual Testing

**Prerequisite**: Please ensure you have a Block Template Theme activated such as TT1 and also the Gutenberg Plugin installed.

1. Download the [Archive.zip](https://github.com/woocommerce/woocommerce-gutenberg-products-block/files/7459684/Archive.zip) file and unzip. Put the containing .html files in `woo-blocks/templates/block-templates/`
2. Load Product Page (e.g. `/product/album/`) and ensure the `single-product.html` template is rendered.
3. Load Taxonomy Page (e.g. `/product-category/clothing/`) and ensure the `taxonomy-product_cat.html` template is rendered
4. Load the Shop/Archive page and ensure the `archive-product.html` template is rendered.
5. Load Site Editor and ensure templates are also present here.

Once you have followed the above steps, add the same template files to your `theme-dir/block-templates` and change the file contents so you can differentiate the Woo Block template and the Theme template. Repeat the above steps starting from Step 2 and confirm the Theme template is preferred over the Woo Blocks template

### DIFF
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -111,6 +111,7 @@ class BlockTemplatesController {
 			}
 
 			$new_template_item = array(
+				'title' => BlockTemplateUtils::convert_slug_to_title( $template_slug ),
 				'slug'  => $template_slug,
 				'path'  => $template_file,
 				'theme' => get_template_directory(),
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -53,10 +53,37 @@ class BlockTemplatesController {
 			return $query_result;
 		}
 
+		$post_type      = isset( $query['post_type'] ) ? $query['post_type'] : '';
 		$template_files = $this->get_block_templates();
 
+		// @todo: Add apply_filters to _gutenberg_get_template_files() in Gutenberg to prevent duplication of logic.
 		foreach ( $template_files as $template_file ) {
-			$query_result[] = BlockTemplateUtils::gutenberg_build_template_result_from_file( $template_file, 'wp_template' );
+			$template = BlockTemplateUtils::gutenberg_build_template_result_from_file( $template_file, 'wp_template' );
+
+			if ( $post_type && ! $template->is_custom ) {
+				continue;
+			}
+
+			if ( $post_type &&
+				isset( $template->post_types ) &&
+				! in_array( $post_type, $template->post_types, true )
+			) {
+				continue;
+			}
+
+			$is_not_custom   = false === array_search(
+				wp_get_theme()->get_stylesheet() . '//' . $template_file['slug'],
+				array_column( $query_result, 'id' ),
+				true
+			);
+			$fits_slug_query =
+				! isset( $query['slug__in'] ) || in_array( $template_file['slug'], $query['slug__in'], true );
+			$fits_area_query =
+				! isset( $query['area'] ) || $template_file['area'] === $query['area'];
+			$should_include  = $is_not_custom && $fits_slug_query && $fits_area_query;
+			if ( $should_include ) {
+				$query_result[] = $template;
+			}
 		}
 
 		return $query_result;
@@ -84,7 +111,6 @@ class BlockTemplatesController {
 			}
 
 			$new_template_item = array(
-				'title' => ucwords( str_replace( '-', ' ', $template_slug ) ),
 				'slug'  => $template_slug,
 				'path'  => $template_file,
 				'theme' => get_template_directory(),
@@ -92,7 +118,6 @@ class BlockTemplatesController {
 			);
 			$templates[]       = $new_template_item;
 		}
-
 		return $templates;
 	}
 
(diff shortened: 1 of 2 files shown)


src/BlockTemplatesController.php
```diff
@@ -53,10 +53,37 @@ public function add_block_templates( $query_result, $query, $template_type ) {
 			return $query_result;
 		}
 
+		$post_type      = isset( $query['post_type'] ) ? $query['post_type'] : '';
 		$template_files = $this->get_block_templates();
 
+		// @todo: Add apply_filters to _gutenberg_get_template_files() in Gutenberg to prevent duplication of logic.
```
frontdevde, line 59: Agreed that ideally, we'd want to prevent code duplication here. That said, this fix unblocks us now which is important. We should keep it on our list to revisit eventually though.

src/BlockTemplatesController.php
```diff
@@ -84,15 +111,14 @@ public function get_block_templates() {
 			}
 
 			$new_template_item = array(
-				'title' => ucwords( str_replace( '-', ' ', $template_slug ) ),
+				'title' => BlockTemplateUtils::convert_slug_to_title( $template_slug ),
```
frontdevde, line 114: Thank you, I appreciate you iterating on this! Definitely a nice win from a user perspective 🙂

src/BlockTemplatesController.php
```diff
@@ -137,6 +163,18 @@ public function render_block_template() {
 			$this->default_block_template_is_available( 'single-product' )
 		) {
 			add_filter( 'woocommerce_has_block_template', '__return_true', 10, 0 );
+		} elseif (
```
frontdevde, line 166: Correct me if I'm wrong but it looks like we're doing the same in all cases, right?

What are your thoughts on maybe making this a bit more readable with a single `if` and moving the checks to variables? Something along the lines of

```
if ( $show_single_product || $show_taxonomy_product_cat || $show_archive_product ) {
	add_filter( 'woocommerce_has_block_template', '__return_true', 10, 0 );
}
```

src/Utils/BlockTemplateUtils.php
```diff
@@ -128,4 +128,24 @@ public static function gutenberg_get_template_paths( $base_directory ) {
...
+	/**
+	 * Converts template slugs into readable titles.
+	 *
+	 * @param string $template_slug The templates slug (e.g. single-product).
+	 * @return string Human friendly title converted from the slug.
+	 */
+	public static function convert_slug_to_title( $template_slug ) {
+		switch ( $template_slug ) {
+			case 'single-product':
+				return 'Single Product Page';
```
frontdevde: These display names probably need `i18n`, no?

frontdevde (approved):
Tested as per testing instructions and it tests as expected for me!
Great to see this fixed 🥳

I left a couple of comments but they are optional or could be addressed later and don't have to block this PR.

src/BlockTemplatesController.php
tjcafferkey: We could potentially clean this up I agree. Something for a future PR like you suggested so we can merge this in for the other week outstanding.

src/BlockTemplatesController.php
```diff
@@ -53,10 +53,37 @@ public function add_block_templates( $query_result, $query, $template_type ) {
...
 		$template_files = $this->get_block_templates();
 
+		// @todo: Add apply_filters to _gutenberg_get_template_files() in Gutenberg to prevent duplication of logic.
 		foreach ( $template_files as $template_file ) {
-			$query_result[] = BlockTemplateUtils::gutenberg_build_template_result_from_file( $template_file, 'wp_template' );
+			$template = BlockTemplateUtils::gutenberg_build_template_result_from_file( $template_file, 'wp_template' );
+
+			if ( $post_type && ! $template->is_custom ) {
+				continue;
+			}
```
dinhtungdu, line 65: @tjcafferkey the `$template` object returned by `BlockTemplateUtils::gutenberg_build_template_result_from_file()` doesn't have `is_custom` property on WP 5.8.1. The `WP_Block_Template` class shipped with 5.8.1 doesn't have that property (the one in Gutenberg does) so I'm getting this warning:

`
Warning: Undefined property: WP_Block_Template::$is_custom in /wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTemplatesController.php on line 63
`

tjcafferkey: Thanks @dinhtungdu. I will open a PR to fix this by setting the `is_custom` value to `false` in `BlockTemplateUtils::gutenberg_build_template_result_from_file()` seeing as we _know_ this is being loaded from the filesystem so it should be `false`.

`is_custom` templates are ones that have been edited and are being retrieved from the DB.

PR here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5067

dinhtungdu: Resolved in #5067

Issue #4965, opened 2021-10-19
Add global styles to all WC Blocks


Aljullu:
We would like to add [global styles](https://developer.wordpress.org/block-editor/how-to-guides/themes/theme-json/) to all WC Blocks.

Below, there is a list of which styles we would like each block to have enabled:

* [x] Featured Category/Featured Product: font size, text color, gradient and duotone (for the overlay), padding, border (color, width, radius). (@gigitux) #5542 #5555 
* [x] Product Categories List: font size, line height, text color, link color. (@gigitux #5133 )
* [ ] ~Product Search: text color, background color (including gradients), border (color and radius -- it should behave as the Search block from Gutenberg), width.~ (on hold, we might rewrite this block soon)
* [x] Active Filters: text color. (@gigitux #5465) 
* [x] Attribute Filter: text color, border (color and radius -- it should be applied to the fields). (@gigitux #5557)
* [x] Price filter: text color, border (color and radius -- it should be applied to the fields). (@gigitux #5559) 
* [x] ~Reviews (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4323): font size, text color.~
* [x] Mini Cart (button): text color (it should also change the icon color). (@gigitux  #5477) 

**Product element blocks**

* [x] Product title: font size, line height, font family, font weight, text transform, text color, background color (including gradients), margin (~#5009~) (@gigitux #5515 )
* [x] Add to cart button: text color, background color (including gradients), padding, border (color and radius) and width
* [x] Category List: font size, text color @gigitux #5516 
* [x] Image: font size, border (radius) @gigitux  #5562
* [x] ~Price: font size, text color~ (on hold: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4965#issuecomment-1005871877)
* [x] Rating: text color (it should change the color of stars) (@gigitux #5521)
* [x] Sale badge: font size, text color, background color (including gradients), padding, color, radius (@gigitux)  #5565
* [x] Stock indicator: font size, text color (@gigitux) #5525
* [x] Summary: font size, text color (@gigitux) #5524
* [x] Tag list: font size, text color, link color (@gigitux) #5528

Some of them are experimental, so you might need Gutenberg installed to get them working.
  
One challenge of this issue is that we might need to update the blocks to use [`apiVersion: 2`](https://make.wordpress.org/core/2020/11/18/block-api-version-2/).

Aljullu:
[...]

PR #5051, opened 2021-11-02
Fix conflict between Mini Cart and Navigation blocks #4592

### LINKED ISSUE #4952: Mini Cart and Navigation blocks conflict
## Describe the bug
The Mini Cart and Navigation blocks are using `window.onload` event, first for lazy load scripts, second one for init some MicroModal. So if I use these blocks on same page, only work the second rendered block action. In my case I have 

After some research.

I replaced `window.onload` to `window.addEventListener('load',function(){})`
Using this method ensures that the handler is additive and does not replace existing handlers. 
And the 2 blocks works properly.

## Screenshots 
image
this and mini cart is not working. 

## Testing steps
1. Make sure you have Gutenberg and Woo Blocks plugins installed
2. Create a new page and add the Mini cart block first
3. Save the page and go to the site to see the Mini cart block working
4. Go back and add the Navigation block after the mini cart
5. Save the page and go to the front-end to see the page
6. Notice that the Mini Cart block stopped working.

## Expected behaviour
The Mini cart block should work.

gigitux:
Fix conflict between Mini Cart and Navigation blocks.

Fixes #4952 

### Manual Testing

How to test the changes in this Pull Request

Check out this branch
1. Make sure you have Gutenberg and Woo Blocks plugins installed
2. Create a new page and add the Mini cart block first
3. Save the page and go to the site to see the Mini cart block working
4. Go back and add the Navigation block after the mini cart
5. Save the page and go to the front-end to see the page
6. Notice that the Mini Cart & Navigation blocks work as expected.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart/frontend.ts
+++ b/assets/js/blocks/cart-checkout/mini-cart/frontend.ts
@@ -15,7 +15,7 @@ interface dependencyData {
 }
 
 // eslint-disable-next-line @wordpress/no-global-event-listener
-window.onload = () => {
+window.addEventListener( 'load', () => {
 	const miniCartBlocks = document.querySelectorAll( '.wc-block-mini-cart' );
 	let wasLoadScriptsCalled = false;
 
@@ -153,4 +153,4 @@ window.onload = () => {
 			);
 		}
 	} );
-};
+} );


ralucaStan (approved):
👏 Nice work on fixing this. I've tested and it works as expected.

dinhtungdu (commented):
I got some issues testing this PR, they aren't necessarily related to the fix introduced in this PR but raising them here for awareness.

- Testing with Storefront, the Navigation items won't load after [step 4](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4952). This is caused by Mini Cart lazy loading, we don't load the full mini cart in the editor:

> PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in /wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTypes/MiniCart.php:169

I added the fix for it in [`add/4965-mini-cart-global-style`#diff-4c8bd3d0cf](https://github.com/woocommerce/woocommerce-gutenberg-products-block/compare/add/4965-mini-cart-global-style#diff-4c8bd3d0cfbd21fce3d29b5b50c28a5172490edbdabe90867fbf4f36dd5d3cb4R165-R167) for #4965, but the PR is WIP.

- Testing with TT1 Block, we have the warning: 
> Warning: Undefined property: WP_Block_Template::$is_custom in /Users/tenup/Sites/woofse/app/public/wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTemplatesController.php on line 63

I'm testing against WP 5.8.1 and the latest Gutenberg `trunk` commit, it's loading the core `WP_Block_Template` instead of the Gutenberg one.

In both cases, I couldn't save/update the page due to the error/warning above.

cc @Aljullu @tjcafferkey

Aljullu (approved):
This is testing well! Congrats on your first PR @gigitux! :partying_face: 

> Testing with Storefront, the Navigation items won't load after step 4. This is caused by Mini Cart lazy loading, we don't load the full mini cart in the editor:

Good catch! I was able to reproduce this issue in this branch but also in `trunk`, so it shouldn't block this PR.

@dinhtungdu could you fill an issue for it so we can fix it separately from other tasks? I would like to explore different ways of fixing it. For me, the issue seems to be that Stripe has some script dependencies which are not registered in REST requests, so our code crashes because it expects those scripts to exist.

So besides your fix, which works, I would also like to see if we can avoid running [`enqueue_data`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/ddf77b02d7d39f814e9c13c6464845d676ec4647/src/BlockTypes/MiniCart.php#L72) completely on REST requests.

> Testing with TT1 Block, we have the warning:

This one also seems to be unrelated to this PR, I would say it belongs to the [Store Editing Templates](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4926) project. @dinhtungdu could you create an issue for this one too and add it to the Templates epic? Thank you!

dinhtungdu:
@Aljullu I created #5073 to fix the Mini Cart block issue. The later one was fixed it in #5067 : )

PR #5058, opened 2021-11-03
Expose withFilteredAttributes to extensions

senadir:
This PR moves `withFilteredAttributes` to `shared-hocs` package.

### Motivation behind this:
- Third-party blocks wishing to use attributes in frontend will have to render in them in the saved HTML. A limitation there is that default attributes are not available in PHP, meaning they can't be appended in HTML and can't be accessed in frontend.
- To fix the issue above, you need to register your block in PHP, and register your attributes there.
- This solution works fine if all of your attributes are in `block.json`, but for some blocks, default value of attributes is dynamic and coming from settings or something.
- So you need to duplicate the default logic in PHP and JS.
- To avoid doing that, `withFilteredAttributes` takes your default attributes in JS and pass them to your frontend block, it's the cleanest solution.

### DIFF
--- a/assets/js/atomic/blocks/product-elements/add-to-cart/frontend.js
+++ b/assets/js/atomic/blocks/product-elements/add-to-cart/frontend.js
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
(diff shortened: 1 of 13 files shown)


mikejolley (approved):
LGTM

PR #5059, opened 2021-11-03
Fix wc-settings not loading when a script that depend on it is in the header

senadir:
This PR is cherry-picked from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5045

The goal of it is to fix an issue in which a script:
- Is loaded in the header (blocks from `register_block_type( block.json )`  for example).
- Depend on `wc-settings` or something that depend on it like `wc-blocks-checkout`, `wc-price-format`.
would break the page because `wc-settings` inline script would not load.

This PR (by @mikejolley) fixes the issue by forcing scripts that depend on `wc-settings` to print in the footer.

### Testing steps

1. This PR is meant to fix a regression in third party blocks, so you can test it by loading AutomateWoo newsletter block, MailPoet block, or create a new block `npx wp-create-block test-block` and include something from `@woocommerce/blocks-checkout` or `@woocommerce/price-format`.
2. The block should load fine in the editor with Cart and Checkout.
3. Inspect the page, you should have a single `wc-settings` script enqueued.

### Changelog

> Scripts using `wc-settings` or script that depend on it would be enqueued in the footer if they're enqueued in the header.

### DIFF
--- a/src/AssetsController.php
+++ b/src/AssetsController.php
@@ -166,11 +166,27 @@ final class AssetsController {
 	 * @see https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5052
 	 */
 	public function update_block_settings_dependencies() {
-		$wp_scripts = wp_scripts();
+		$wp_scripts     = wp_scripts();
+		$known_packages = [ 'wc-settings', 'wc-blocks-checkout', 'wc-price-format' ];
 
 		foreach ( $wp_scripts->registered as $handle => $script ) {
-			if ( in_array( 'wc-settings', $script->deps, true ) || in_array( 'wc-blocks-checkout', $script->deps, true ) ) {
+			// scripts that are loaded in the footer has extra->group = 1.
+			if ( array_intersect( $known_packages, $script->deps ) && ! isset( $script->extra['group'] ) ) {
+				// Append the script to footer.
 				$wp_scripts->add_data( $handle, 'group', 1 );
+				// Show a warning.
+				$error_handle  = 'wc-settings-dep-in-header';
+				$used_deps     = implode( ', ', array_intersect( $known_packages, $script->deps ) );
+				$error_message = "scripts that has a depenency on [ {$used_deps} ] must be loaded in the footer, {$handle} was loaded in the header but was moved to the footer. See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5059";
+
+				// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.NotInFooter,WordPress.WP.EnqueuedResourceParameters.MissingVersion
+				wp_register_script( $error_handle, '' );
+				wp_enqueue_script( $error_handle );
+				wp_add_inline_script(
+					$error_handle,
+					sprintf( 'console.warn( "%s" );', $error_message )
+				);
+
 			}
 		}
 	}
--- a/src/AssetsController.php
+++ b/src/AssetsController.php
@@ -38,6 +38,7 @@ final class AssetsController {
 		add_action( 'body_class', array( $this, 'add_theme_body_class' ), 1 );
 		add_action( 'admin_body_class', array( $this, 'add_theme_body_class' ), 1 );
 		add_action( 'admin_enqueue_scripts', array( $this, 'update_block_style_dependencies' ), 20 );
+		add_action( 'wp_enqueue_scripts', array( $this, 'update_block_settings_dependencies' ), 1 );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


senadir:
~~This still needs a console log of when this happens~~ This is now done

src/AssetsController.php
```diff
@@ -38,6 +38,8 @@ protected function init() {
 		add_action( 'body_class', array( $this, 'add_theme_body_class' ), 1 );
 		add_action( 'admin_body_class', array( $this, 'add_theme_body_class' ), 1 );
 		add_action( 'admin_enqueue_scripts', array( $this, 'update_block_style_dependencies' ), 20 );
+		add_action( 'wp_enqueue_scripts', array( $this, 'update_block_settings_dependencies' ), 1 );
```
mikejolley: Why priority 1 for these?

src/AssetsController.php
```diff
@@ -149,4 +151,42 @@ public function update_block_style_dependencies() {
...
+				$error_handle  = 'wc-settings-dep-in-header';
+				$used_deps     = implode( ', ', array_intersect( $known_packages, $script->deps ) );
+				$error_message = "Scripts that have a dependency on [$used_deps] must be loaded in the footer, {$handle} was registered to load in the header, but has been switched to load in the footer instead. See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5059";
+				// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.NotInFooter,WordPress.WP.EnqueuedResourceParameters.MissingVersion
+				wp_register_script( $error_handle, '' );
+				wp_enqueue_script( $error_handle );
+				wp_add_inline_script(
+					$error_handle,
+					sprintf( 'console.warn( "%s" );', $error_message )
+				);
```
mikejolley, line 187: 👍🏻

mikejolley (commented):
Question below. Everything else looks ok. Let @nielslange give final approval.

src/AssetsController.php
senadir: I'm not sure, as this is your commit.

mikejolley: I don't recall adding it, it shouldn't be needed. If anything, it should be a late priority.

nielslange (approved):
LGTM! 🚢

senadir:
@mikejolley this code seems to be running on all registered scripts, even the ones not enqueued on the current page. I'm seeing it on the checkout frontend for an editor script. I guess that's not something we can avoid?

image

mikejolley:
@senadir could try using `$wp_scripts->enqueued` instead of `$wp_scripts->registered`? But it might break for dependencies of dependencies.

senadir:
`enqueued` is returning undefined for me 

```
Notice: Undefined property: WP_Scripts::$enqueued in /src/AssetsController.php on line 172
```

senadir:
We can try `queue` but its content is different from `registered`. queue is a flat array of handles that would enqueued.
```
Notice: Trying to get property 'deps' of non-object in /Users/nadir/work/bergs/src/AssetsController.php on line 174
```

mikejolley:
[Maybe it is queue.](https://github.com/WordPress/WordPress/blob/2cb4ebefe2ee98fc36a5962e92590cb0451ad2a6/wp-includes/class.wp-dependencies.php#L33) You can cross-reference the objects in `registered`.

tomalec:
@senadir Could you give some hints on how to act upon that warning?
For AutomateWoo it throws it for the block added in https://github.com/woocommerce/automatewoo/pull/1061

What is the recommended way to register a block in the footer?

See https://github.com/woocommerce/automatewoo/issues/1304

senadir:
If you're loading scripts using the block.json then it will load up in the header sadly, you will need to enqueue your scripts manually.

nerrad:
@senadir or @mikejolley is this (showing a warning for blocks registering scripts via `block.json` configuration that also have a dependency on `wc-settings`) something that can't be addressed in WooCommerce Blocks? I realize the warning is shown after forcing the script to load in the footer. 

There's a couple things we might want to consider:

- rethink the loading strategy of `wc-settings`. Instead of having the entire library enqueued in the footer (which was done to allow for additional settings to be set before the script data is rendered inline) - could we decouple the inline data output from the client API for the data? Then maybe the inline data output could just always be in the footer without requiring `wc-settings` client to also be in the footer?
- There is a `defer` loading strategy [being proposed for block scripts](https://github.com/WordPress/gutenberg/pull/52536) that could potentially be considered for `wc-settings` in combination with the above.

mikejolley:
I'd need to look at this again to refresh memory, but I think we currently just dump a variable on the page with the settings object right? What if that were a callback that added the settings after initial load? I guess this could break some blocks/utils relying on settings being there on load however.

Brianmitchtay:
Seeing this come up here with a user, letting them know not to worry about it. 7761451-zd-a8c

PR #5060, opened 2021-11-03
Fix sanitization callback

senadir:
This PR is cherry-picked from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5045

This fixed an error where data passed to the Store API via post would be incorrectly represented in the $request object after data sanitization. 

Previously it would look over all items in the object recursively but due to this condition, would return the first sanitized value:

```
if ( isset( $property_value['properties'] ) ) {
					$sanitize_callback = $this->get_recursive_sanitize_callback( $property_value['properties'] );
					return $sanitize_callback( $current_value, $request, $param . ' > ' . $property_key );
}
```

Other data in the `extensions` object would then be skipped. 

The fix is to place all values into an array and return them once all properties have been sanitized.

### Testing steps

Since these changes only impact developers, there are no user facing testing steps to perform. I have however added a test case to confirm nested properties are no longer destroyed. Ensure php tests pass with this change.

### DIFF
--- a/tests/php/StoreApi/Routes/Checkout.php
+++ b/tests/php/StoreApi/Routes/Checkout.php
@@ -1,0 +1,225 @@
+<?php
+/**
+ * Controller Tests.
+ */
+
+namespace Automattic\WooCommerce\Blocks\Tests\StoreApi\Routes;
+
+use Automattic\WooCommerce\Blocks\Domain\Services\ExtendRestApi;
+use Automattic\WooCommerce\Blocks\Domain\Package as DomainPackage;
+use Automattic\WooCommerce\Blocks\StoreApi\Formatters;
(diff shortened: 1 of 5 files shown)


mikejolley:
@senadir @opr Added a test case to confirm the fix is in place. Used a package called [Mockery](https://github.com/mockery/mockery) to test the action receives the correct params.

tests/php/StoreApi/Routes/Checkout.php
```diff
@@ -0,0 +1,225 @@
+<?php
```
senadir, line 1: Reviewed this new test and it's looks fine to me! Thank you for writing it.

senadir (commented):
meta-approving giving I'm the PR author but it's actually Mike's PR.

mikejolley:
e2e test fail is unrelated. Merging.

PR #5062, opened 2021-11-03
Allow block templates to be customised and saved

### LINKED ISSUE #5024: Can't save modified WC block template if theme doesn't include it
If the user modifies a WooCommerce block template but the theme doesn't include it, it will fail to save.

## To reproduce
1. Amend WooCommerce Core or checkout this PR https://github.com/woocommerce/woocommerce/pull/30997
2. Add the following snippet in the directory specified:

```html
```

3. Go to Site Editor > Templates and modify the Single product template.
4. Try to save the template. If you take a look at the Network tab of your devtools, you will notice the request returned error 404. If you reload the page, you will notice the changes weren't saved.

opr:
This PR focuses on allowing blocks templates to be customised, saved, and retrieved without errors.

The changes made are:
- Copy the following functions from Gutenberg (because they are private and we need access to them in Blocks) into our `BlockTemplateUtils` class.
  - `gutenberg_add_template_info`
  -  `gutenberg_build_template_result_from_post`
- Wherever a `$template`'s properties are accessed, normalise `$template` to be an object. This is necessary because this is sometimes an array or sometimes a `WP_Block_Template`.
- When building templates from the woo-blocks filesystem, ensure the theme is set to `woocommerce`. This is necessary as the current behaviour is to save the template under the current theme's slug, which seems incorrect.
- Add a `maybe_return_blocks_template` function - this is used to short-circuit the `gutenberg_get_block_templates` method from Gutenberg. It checks the db to see if there's a saved template already. If there is then we can return that.
- Add `get_single_block_template` function. Until now there was no way to get a single woo-blocks block template in our codebase. This function is used within `gutenberg_get_block_templates` to override/augment the data returned by the `gutenberg_get_block_templates` function. This is how we inject _our_ templates into the list of templates gutenberg finds.
- Update `get_block_templates` in `BlockTemplatesController.php` to check the database first to see if any saved templates from the `woocommerce` theme exist. If they do, build a template from the saved posts.
- Update `add_block_templates` in `BlockTemplatesController.php` to only add the templates from files that don't have a customised version in the database.
- If the query passed to the `get_block_templates` filter contains `slug__in` then filter the results of `BlockTemplatesController`'s `get_block_templates` method to only included those that match the queried slugs.
- Add a `remove_theme_templates_with_custom_alternative` - this will cater for the situation described in *Second testing steps for alternative scenario (Theme adds a template for an already customised template)*
 
Fixes #5024 

#### Other Checks

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Use an FSE theme.
2. Add a file named `single-product.html` to `woocommerce-gutenberg-products-block/templates/block-templates/` - insert the following content:
```
<!-- wp:template-part {"slug":"header"} /-->
<!-- wp:paragraph --> <p>single-product</p> <!-- /wp:paragraph -->
<!-- wp:template-part {"slug":"footer"} /-->
```
3. On the site editor change the content of this file, add a block, move things about, etc. 
4. Open the network tab in the inspector and save the page. Ensure there are no errors.
5. Reload the editor, verify your changes have persisted.
6. Delete the `single-product.html` file 
7. Reload the editor, verify your changes have persisted.
8. Go to `Appearance > Templates` and verify the template is saved with a theme of `woocommerce`.
9. Delete the template.
10. Add `single-product.html` with the above contents to `wp-content/themes/<your-theme>/block-templates/` and repeat steps 3-6. 

### Second testing steps for alternative scenario (Theme adds a template for an already customised template)
1. Delete all saved templates from the Appearance > Templates page.
2. Add the `single-product.html` file above to `woo-blocks/templates/block-templates`. Ensure there is *not* a `single-product.html` file in your theme's block-templates directory!
3. Go to the site editor, and verify you can edit the template.
4. Customise it and save it.
5. Add the `single-product.html` file to your theme's block-templates directory. 
6. Reload the editor and ensure your customised version of `single-product` is loaded. 
7. Ensure there is only one entry for the template in the `General templates` list in the editor.

### User Facing Testing
* [ ] Same as above, or
* [x] See steps below.

1. Install a FSE theme, i.e. TT1 Blocks, or Tove.
2. Make make sure you have [this change applied in your WooCommerce plugin](https://github.com/woocommerce/woocommerce/pull/30997/files).
The block theme needs to declare `add_theme_support( 'woocommerce' );` in the `functions.php` file.
3. Go to the site editor and use the arrow at the top of the page to select a template. Select General Templates > Single product.
4. Modify the template, add some blocks, move some things around etc.
5. Save the template.
6. Reload the page, verify the changes have persisted.
7. Go to the dashboard then `Appearance > Templates`. Verify there's a template called `Single product` with the theme `woocommerce`


nerrad:
> Copy the following functions from Gutenberg (because they are private and we need access to them in Blocks) into our BlockTemplateUtils class.

I have just a bit of feedback that you may have already considered/researched. It might be good to check with the GB team regarding the private functions you referenced (and are duplicating logic for) to see:

- Are there going to be public equivalents in WordPress core when WP 5.9 lands (or are they being worked on)?
- Is this something that could be provided in WordPress/Gutenberg to have value for other plugins like Woo that might need similar functionality?

The above doesn't necessarily need to block the work here given the tight timelines we have, but it would be good to followup with so we don't end up diverging too much from what may be a more ideal solution in utilizing an API built into WP.

tjcafferkey:
Hey @nerrad this is something I checked with GB team as we already considered this. They informed us that these were not intended to used publicly outside of Gutenberg for the time being as the APIs for these functions may change at any time unannounced. In the interest of time we have chosen to duplicate these utility functions instead as you've pointed out.

That being said, I plan on writing a P2 post in the aftermath of this project to demonstrate how we have used these duplicate functions and how/why they were useful. Hopefully from this conversation we'll be able to understand what needs to happen to make this available for public usage, if at all possible.

tjcafferkey:
**The below error happened due to the following:**

We need to change references such as `$template_file['slug']` to `$template_file->slug` in the `BlockTemplateController` since you've normalised that data to become an object.

----

Seemed to be going OK (refreshing the Site Editor persisted changes) and then I got to the following step.

>Go to Appearance > Templates and verify the template is saved with a theme of woocommerce.

The template was present, but when I clicked "Edit" on this template I got the following fatal error:

Screenshot 2021-11-04 at 06 41 45

After this error, when I tried to navigate back to the Site Editor via the menu item, I got stuck on the following loading state:

Screenshot 2021-11-04 at 06 42 16

src/Utils/BlockTemplateUtils.php
```diff
@@ -74,6 +74,143 @@ public static function gutenberg_inject_theme_attribute_in_content( $template_co
...
+			$file_path = $theme_dir . '/' . $template_base_paths[ $template_type ] . '/' . $slug . '.html';
+			if ( file_exists( $file_path ) ) {
+				$new_template_item = array(
+					'slug'  => $slug,
+					'path'  => $file_path,
+					'theme' => $theme_slug,
+					'type'  => $template_type,
+				);
+
+				if ( 'wp_template_part' === $template_type ) {
```
tjcafferkey: `$template_type` will never be `wp_template_part` so we can delete this code.

src/Utils/BlockTemplateUtils.php
```diff
@@ -74,6 +74,143 @@ public static function gutenberg_inject_theme_attribute_in_content( $template_co
...
+					return self::gutenberg_add_template_part_area_info( $new_template_item );
+				}
+
+				if ( 'wp_template' === $template_type ) {
```
tjcafferkey: `$template_type` will always be `wp_template`

src/Utils/BlockTemplateUtils.php
```diff
@@ -74,6 +74,143 @@ public static function gutenberg_inject_theme_attribute_in_content( $template_co
...
+		$template->slug           = $post->post_name;
+		$template->source         = 'custom';
+		$template->type           = $post->post_type;
+		$template->description    = $post->post_excerpt;
+		$template->title          = $post->post_title;
+		$template->status         = $post->post_status;
+		$template->has_theme_file = $has_theme_file;
+		$template->is_custom      = true;
+
+		if ( 'wp_template' === $post->post_type && isset( $default_template_types[ $template->slug ] ) ) {
```
tjcafferkey: I think we can delete this as it will never resolve as the Woo templates don't exist in `$default_template_types`

src/Utils/BlockTemplateUtils.php
```diff
@@ -74,6 +74,143 @@ public static function gutenberg_inject_theme_attribute_in_content( $template_co
...
+			$template->is_custom = false;
+		}
+
+		if ( 'wp_template_part' === $post->post_type ) {
```
tjcafferkey: This will never be template part AFAIK

src/Utils/BlockTemplateUtils.php
```diff
@@ -74,6 +74,143 @@ public static function gutenberg_inject_theme_attribute_in_content( $template_co
...
+
+	/**
+	 * Build a unified template object based a post Object.
+	 *
+	 * @param \WP_Post $post Template post.
+	 *
+	 * @return \WP_Block_Template|\WP_Error Template.
+	 */
+	public static function gutenberg_build_template_result_from_post( $post ) {
+		$default_template_types = gutenberg_get_default_template_types();
```
tjcafferkey: No need for this once its usage below has been removed so we can remove this also

src/Utils/BlockTemplateUtils.php
```diff
@@ -83,23 +220,24 @@ public static function gutenberg_inject_theme_attribute_in_content( $template_co
 	 * @return \WP_Block_Template Template.
 	 */
 	public static function gutenberg_build_template_result_from_file( $template_file, $template_type ) {
+		$template_file          = (object) $template_file;
+		$default_template_types = function_exists( 'gutenberg_get_default_template_types' ) ? gutenberg_get_default_template_types() : array();
```
tjcafferkey: This isnt being used so we can remove this.

src/BlockTemplatesController.php
```diff
@@ -38,6 +38,97 @@ public function __construct() {
...
+		$matched_templates = array_values(
+			array_filter(
+				$available_templates,
+				function( $available_template ) use ( $slug ) {
+					$available_template = (object) $available_template;
+					return $available_template->slug === $slug;
+				}
+			)
+		);
+		return count( $matched_templates > 0 ) ? (object) $matched_templates[0] : $template;
```
dinhtungdu: Typed too fast? 😁

```suggestion
		return count( $matched_templates ) > 0 ? (object) $matched_templates[0] : $template;
```

dinhtungdu: This causes a fatal error when I tried to save the modified template at the step 4th.

`PHP Fatal error:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, bool given in /wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTemplatesController.php:131`

dinhtungdu:
After fixing the issue in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5062#pullrequestreview-797401488, I can save the template, but got another after that:

`Warning: Undefined property: WP_Block_Template::$path in /Users/tenup/Sites/woofse/app/public/wp-content/plugins/woocommerce-gutenberg-products-block/src/Utils/BlockTemplateUtils.php on line 226`

The value of `$template_file` at that point:

```
/wp-content/plugins/woocommerce-gutenberg-products-block/src/Utils/BlockTemplateUtils.php:226:
object(WP_Block_Template)[2857]
  public 'type' => string 'wp_template' (length=11)
  public 'theme' => string 'woocommerce' (length=11)
  public 'slug' => string 'single-product' (length=14)
  public 'id' => string 'woocommerce//single-product' (length=27)
  public 'title' => string '' (length=0)
  public 'content' => string '<!-- wp:template-part {"slug":"header","theme":"tt1-blocks"} /-->

<!-- wp:paragraph -->
<p>Hello, world!</p>
<!-- /wp:paragraph -->

<!-- wp:template-part {"slug":"footer","theme":"tt1-blocks"} /-->' (length=271)
  public 'description' => string '' (length=0)
  public 'source' => string 'custom' (length=6)
  public 'wp_id' => int 282
  public 'status' => string 'publish' (length=7)
  public 'has_theme_file' => boolean true
  public 'is_custom' => boolean true
```

The above warning leads to this error:
`Fatal error: Uncaught ValueError: Path cannot be empty in /Users/tenup/Sites/woofse/app/public/wp-content/plugins/woocommerce-gutenberg-products-block/src/Utils/BlockTemplateUtils.php on line 226`

---

Edit: Applying the following change (from @tjcafferkey ) fixes this issue for me:
BlockTemplateController line 160 change:
`$template = BlockTemplateUtils::gutenberg_build_template_result_from_file( $template_file, 'wp_template' );`
to
`$template = BlockTemplateUtils::gutenberg_build_template_result_from_post( $template_file, 'wp_template' );`

tjcafferkey (approved):
LGTM, works as described now without any errors 🎉

frontdevde (changes requested):
While testing as per testing instructions works well and as expected, I'm seeing issues when testing with additional modified templates.

Could you please test the following:

- Add an `archive-product.html` 
```
<!-- wp:template-part {"slug":"header"} /-->
<!-- wp:paragraph --> <p>archive-product</p> <!-- /wp:paragraph -->
<!-- wp:woocommerce/legacy-template {"template":"archive-product"} /-->
<!-- wp:template-part {"slug":"footer"} /-->
```
- Modify the archive product template in the editor & save
- Load a single product page on the Frontend

The expected behavior would be that the single product page still loads the single product template. What I'm seeing in my testing though is that it starts loading the archive product template on the single product page. 

Can you confirm this behavior?

src/Utils/BlockTemplateUtils.php
```diff
@@ -83,23 +199,21 @@ public static function gutenberg_inject_theme_attribute_in_content( $template_co
...
+		$template->slug           = $template_file->slug;
 		$template->type           = $template_type;
-		$template->title          = ! empty( $template_file['title'] ) ? $template_file['title'] : $template_file['slug'];
+		$template->title          = ! empty( $template_file->title ) ? $template_file->title : $template_file->slug;
 		$template->status         = 'publish';
 		$template->has_theme_file = true;
 		$template->is_custom      = false; // Templates loaded from the filesystem aren't custom, ones that have been edited and loaded from the DB are.
-		$template->title          = self::convert_slug_to_title( $template_file['slug'] );
-
+		$template->title          = self::convert_slug_to_title( $template_file->slug );
```
Aljullu: Four lines above we are defining the `title` too, I guess one of those two lines should be removed?

tjcafferkey:
>The expected behavior would be that the single product page still loads the single product template. What I'm seeing in my testing though is that it starts loading the archive product template on the single product page.
>
>Can you confirm this behavior?

This does not happen for me, it works as expected here.

frontdevde (dismissed):

src/BlockTemplatesController.php
```diff
@@ -138,13 +284,19 @@ public function theme_has_template( $template_name ) {
...
 		}
 
 		return is_readable(
 			$this->templates_directory . '/' . $template_name . '.html'
+		) ||
+		array_filter(
+			$this->get_block_templates(),
+			function ( $template ) use ( $template_name ) {
+				return $template->slug === $template_name;
+			}
```
Aljullu: If we only care about templates saved in the database, would this be equivalent?
```suggestion
		) || gutenberg_get_block_template( 'woocommerce//' . $template_name );
```

Not blocking, but I also wonder if this fix should be done [directly in WC core](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/class-wc-template-loader.php#L110). I think that was an overlook from my part. But the WC core function shouldn't ignore templates saved in the database.

src/BlockTemplatesController.php
```diff
@@ -98,25 +206,63 @@ public function get_block_templates() {
 		$template_files = BlockTemplateUtils::gutenberg_get_template_paths( $this->templates_directory );
 		$templates      = array();
 
+		// Check if the template has been saved in the database first.
```
Aljullu: I'm having a bad time understanding this comment. This code is fetching all WC templates from the database right? So what about replacing it with something along these lines?
```suggestion
		// First, retrieve the templates which are saved in the database.
```
I might be misunderstanding the code, in that case, feel free to ignore. :slightly_smiling_face:

src/BlockTemplatesController.php
```diff
@@ -98,25 +206,63 @@ public function get_block_templates() {
...
+		);
+		$check_query         = new \WP_Query( $check_query_args );
+		$saved_woo_templates = $check_query->posts;
+
+		// From a WordPress post, build the template object and add it to our list.
+		foreach ( $saved_woo_templates as $saved_woo_template ) {
+			$templates[] = BlockTemplateUtils::gutenberg_build_template_result_from_post( $saved_woo_template );
+		}
+
 		foreach ( $template_files as $template_file ) {
```
Aljullu, line 279: Not blocking either, but what would you think about breaking `get_block_templates` into smaller functions (`get_block_templates_from_db`, `get_block_templates_from_theme`, `get_block_templates_from_woocommerce`...)? I feel it would make it easier to follow what's going on in each moment.

src/BlockTemplatesController.php
```diff
@@ -56,11 +147,28 @@ public function add_block_templates( $query_result, $query, $template_type ) {
 		$post_type      = isset( $query['post_type'] ) ? $query['post_type'] : '';
 		$template_files = $this->get_block_templates();
 
+		if ( isset( $query['slug__in'] ) ) {
```
Aljullu: Again, not blocking, but wondering if `get_block_templates()` could be updated to accept an array of `slug`s as a parameter so it only returns templates for that specific slug. That way, consumers (like this function or `get_single_block_template`) wouldn't need to filter the list afterwards.

src/BlockTemplatesController.php
```diff
@@ -38,6 +38,97 @@ public function __construct() {
...
+		list( , $slug ) = $template_name_parts;
+
+		// Remove the filter at this point because if we don't then this function will infinite loop.
+		remove_filter( 'pre_get_block_template', array( $this, 'maybe_return_blocks_template' ), 10, 3 );
+
+		// Check if the theme has a saved version of this template before falling back to the woo one. Please note how
+		// the slug has not been modified at this point, we're still using the default one passed to this hook.
+		$maybe_template = gutenberg_get_block_template( $id, $template_type );
+		if ( null !== $maybe_template ) {
+			return $maybe_template;
```
Aljullu, line 71: Shouldn't we add the `pre_get_block_template` filter back again before returning?

Aljullu (commented):
Good job with this PR @opr! I really like how you solved several of the challenges from this PR.

I left some comments below, most of them are not blocking, but if you feel they make sense, we should either fix them or create an issue to track them.

src/BlockTemplatesController.php
opr: Yes your wording makes a bit more sense :)

src/BlockTemplatesController.php
opr: Great point 👏🏼

src/BlockTemplatesController.php
opr: Excellent suggestion, I've defaulted to an empty array so it won't filter anything if it's not explicitly passed in.

src/BlockTemplatesController.php
opr: `get_block_templates_from_theme` won't be necessary because at this point the theme templates are already loaded.

I did split it into `get_block_templates_from_db` and `get_block_templates_from_woocommerce` though, I also had to introduce a new function: `remove_theme_templates_with_custom_alternative` because in the following scenario, duplicate pages would be available in the editor:

1. Theme has no template, but WooCommerce does
2. User edits the template and saves it to DB.
3. Theme updates and includes template.
4. Without this change, the theme would include the template and woo would also include it, from the db causing a duplicate,

src/BlockTemplatesController.php
opr: @Aljullu you are right, WC Core seems to only look at templates on the filesystem. I can address this later. This has changed during the refactor, do you think we should go straight to the Gutenberg function, or do you think it's OK to use our `get_block_templates` with the `$slugs` arg?

src/BlockTemplatesController.php
```diff
@@ -72,55 +172,167 @@ public function add_block_templates( $query_result, $query, $template_type ) {
...
+			$saved_woo_templates
+		);
+	}
+
+	/**
+	 * Gets the templates from the WooCommerce blocks directory, skipping those for which a template already exists
+	 * in the theme directory.
+	 *
+	 * @param string[] $slugs An array of slugs to filter templates by. Templates whose slug does not match will not be returned.
+	 * @param array    $already_found_templates Templates that have already been found, these will customised templates that are loaded from the database.
```
Aljullu: Nit: There seems to be a typo in the comment `these will customised`.

src/BlockTemplatesController.php
```diff
@@ -72,55 +172,167 @@ public function add_block_templates( $query_result, $query, $template_type ) {
...
 			}
 		}
 
+		$query_result = $this->remove_theme_templates_with_custom_alternative( $query_result );
 		return $query_result;
 	}
 
 	/**
-	 * Get and build the block template objects from the block template files.
+	 * Removes templates that were added to a theme's block-templates director, but already had a customised version saved in the database.
```
Aljullu:
```suggestion
	 * Removes templates that were added to a theme's block-templates directory, but already had a customised version saved in the database.
```

src/BlockTemplatesController.php
```diff
@@ -38,6 +38,87 @@ public function __construct() {
 	protected function init() {
 		add_filter( 'get_block_templates', array( $this, 'add_block_templates' ), 10, 3 );
 		add_action( 'template_redirect', array( $this, 'render_block_template' ) );
+		add_filter( 'pre_get_block_template', array( $this, 'maybe_return_blocks_template' ), 10, 3 );
+	}
+
+	/**
+	 * This function checks if there's a blocks template (ultimately it resolves either a saved blocks template from the
+	 * database or a template file in `woo-gutenberg-products/block/templates/block-templates/`)
```
Aljullu:
```suggestion
	 * database or a template file in `woo-gutenberg-products-block/templates/block-templates/`)
```

Aljullu (approved):
LGTM! I left a couple of corrections to comment typos, but everything else is working great on my end, good job!

PR #5065, opened 2021-11-03
Add .prettierignore configuration

### LINKED ISSUE #5064: Add .prettierignore configuration
We need to add a prettier config file to exclude `.scss`files, considering that the project has a specific lint for CSS

gigitux:
This PR adds the `.prettierignore`. In this way, Prettier excludes `.scss` files.

Fixes #5064 

### Manual Testing

Check out this branch
Check that your editor has Prettier configured correctly and it read the configuration of this project

1. Open a `.scss` file.
2. Check that the editor, by adding some space, does not format the code according to Prettier's rules.


ralucaStan:
Adding this file indeed prevents Prettier from running on `.scss` files. But while testing I realised that _I actually like_ having Prettier there, as it is consistent with the rest of the project.

So I was thinking we could fix this another way: let Prettier work, but fix the eslint errors on the precommit hook. Currently we are just linting the staged `.scss` files.

That would mean changing the husky configuration in `package.json` from 
```
	"lint-staged": {
		"*.scss": [
			"npm run lint:css"
		],
```
to 
```
	"lint-staged": {
		"*.scss": [
			"npm run lint:css-fix",
			"git add assets/**/*.scss"
		],
```

This will automatically try to fix any formatting that Prettier introduces that is not aligned with our WP ESlint config. The fixes will be unstaged so we need the ``"git add assets/**/*.scss"`` step to gather the unstaged `.scss` files and add them to the commit as well. 

The only (minor) issue I see is that this might also stage some unstaged `.scss` files that one does not want to include in their commit, because it's not possible to just stage the files where the fix was applied.

I'll run this proposal by the team.

nerrad:
Raluca I like your idea but there's a few things that come to mind:

- As you already mentioned,  a PR might end up having non-related files staged (but I assume this might only happen for a short period.
- While eslint fix will likely be okay, having it run on every commit does introduce the chance of bad fixes. At least with the prettierignore we should be able to configure our IDEs to fix automatically on save for scss files using eslint?
- It's too bad there are mismatches between prettier formatting and eslint rules. I wonder if that's something that could be addressed by configuration (or maybe upstream in the related `@wordpress` linting packages).

I wonder if a good interim step would be to create a separate issue for this and in the meantime we could just merge this PR and see how it works?

ralucaStan:
Thank you @nerrad for your answer, I'll move my suggestion to another ticket. The proposal was inspired by my current workflow: I let Prettier run, and then manually run `npm run lint:css-fix`.  I see now that this might not be the workflow for everybody.

> As you already mentioned, a PR might end up having non-related files staged (but I assume this might only happen for a short period.

The non related files would only  be `*.scss files`. I was thinking that we are not changing a lot of those so the impact would not be so big.

> While eslint fix will likely be okay, having it run on every commit does introduce the chance of bad fixes. 

This will only run on commits that have `.scss` files. 

> At least with the prettierignore we should be able to configure our IDEs to fix automatically on save for scss files using eslint?

I did not think of this, I'll try this

> I wonder if that's something that could be addressed by configuration (or maybe upstream in the related @wordpress linting packages).

I was thinking the same, because there is not much we could do at our project level.

ralucaStan (approved):
Based on the discussion on the PR I'm approving this! 👏 
It works as expected.

PR #5067, opened 2021-11-03
Set is_custom property on template and remove redundant code

tjcafferkey:
* Set `is_custom` property on `$template` to `false`. This will always be false as it's being loaded from the filesystem. Only templates that are edited and therefore saved in the database are considered custom.
* Removed redundant if statement code since `$default_template_types` only contains core WordPress default templates (not WooCommerce) so the removed if statements would never resolve.
* Removed second redundant if statement code since `$template_type` is always going to be `wp_template` and not `wp_template_part`

### Manual Testing

This will mainly be regression testing.

**Prerequisite**: Please ensure you have a Block Template Theme activated such as TT1 and also the Gutenberg Plugin installed.

1. Load the Site Editor (Appearance > Editor) and select the Single Product Page template from the `General templates` section.
2. Confirm you can load the template and it looks as expected in large/small viewports.
3. Confirm the template is being loaded on the frontend.

Once you have followed the above steps, add the same template files to your `theme-dir/block-templates` and change the file contents so you can differentiate between the Woo Block template and the Theme template. Repeat the above steps starting from Step 2 and confirm the Theme template is preferred over the Woo Blocks template

### Changelog

> FSE: Fix missing `is_custom` property for WooCommerce block template objects.

### DIFF
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -111,7 +111,6 @@ class BlockTemplatesController {
 			}
 
 			$new_template_item = array(
-				'title' => BlockTemplateUtils::convert_slug_to_title( $template_slug ),
 				'slug'  => $template_slug,
 				'path'  => $template_file,
 				'theme' => get_template_directory(),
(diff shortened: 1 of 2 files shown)


dinhtungdu (approved):
LGTM 🎉

PR #5069, opened 2021-11-03
pass blockWrapper to forcedBlocks

senadir:
ForcedBlocks in frontend didn't have access to `blockWrapper`, and therefore, didn't have access to props that we pass to them like `cart`, `extensions,` and `checkoutExtensionData`. This PR fixes is.

### Testing steps
1. To test this, you need a forced block, edit the html of a checkout block and remove a core block.
2. Save without switching to visual editor.
3. Inside that block `block.js` component, console log `checkoutExtensionData` from its props.
4. Make sure it has a correct value and it's not undefined.


mikejolley (approved):
👍🏻 makes sense

PR #5072, opened 2021-11-04
Fix label alignment of the product search in the editor

### LINKED ISSUE #3083: Product Search input has different margin in editor and frontend
## To reproduce

Add a Product Search block and notice the input has a margin in the editor but it doesn't in the frontend.

_Frontend:_
imatge

_Editor:_
imatge

:point_up_2: notice the `Search` label is aligned with the box in the frontend, but it isn't in the editor.

---

In Storefront, we have an issue to fix other styling differences for this block between the editor and the frontend (see https://github.com/woocommerce/storefront/issues/1438), but this margin issue seems to be a regression not related to the theme.

gigitux:
This PR fixed the `search` label alignment on the edit page for the block `Product Search`.

Fixes #3083 

### Screenshots

image
*before - frontend page*

image
*before - edit page*

image
*current implementation - frontend page*

image
*current implementation - edit page*

### Manual Testing

Check out this branch

1. Make sure you have Woo Blocks plugins installed
2. Create a new page and add the Product Search block
3. Save
4. Check if the label "Search" position is equal on frontend page and editor page

### Changelog

> Fix label alignment of the product search in the editor.

### DIFF
--- a/assets/js/blocks/product-search/editor.scss
+++ b/assets/js/blocks/product-search/editor.scss
@@ -1,7 +1,12 @@
 .wc-block-product-search__field.input-control {
 	color: #828b96 !important;
 	padding: 0;
+
+	.components-base-control__field {
+		margin-bottom: 0;
+	}
 }
+
 .wc-block-product-search {
 	.wc-block-product-search__fields {
 		.block-editor-rich-text {
--- a/assets/js/blocks/product-search/editor.scss
+++ b/assets/js/blocks/product-search/editor.scss
@@ -1,5 +1,6 @@
 .wc-block-product-search__field.input-control {
 	color: #828b96 !important;
+	padding: 0;
 }
 .wc-block-product-search {
 	.wc-block-product-search__fields {


assets/js/blocks/product-search/editor.scss
```diff
@@ -1,5 +1,6 @@
 .wc-block-product-search__field.input-control {
 	color: #828b96 !important;
+	padding: 0;
 }
```
ralucaStan, line 8: 👍 this fixes the spacing issue.

While testing I found out that we could also easily fix the button size if we remove the margin for the first child div of this wrapper, like in the suggestion below. Feel free to also add this to your PR

```suggestion
	.components-base-control__field {
		margin-bottom: 0;
	}
}
```
Before:
Screenshot 2021-11-05 at 12 40 42
 After: 
Screenshot 2021-11-05 at 12 59 38

ralucaStan (approved):
Looks good, I've also made a suggestion to improve even more the spacing of the block.

assets/js/blocks/product-search/editor.scss
gigitux: Thank you for your suggestion!  
I updated the PR

PR #5093, opened 2021-11-08
Update templates so Legacy Template blocks don't take all available width

### LINKED ISSUE #5092: Legacy Template block is always rendered full width
The templates we have been creating as part of the Store Editing Templates v1 project render the Legacy Template block directly into the content, that causes it to take all the available wide space.

Single product template

I took a look at how that's done in other templates, and they seem to use a Group block with the attribute `{"layout":{"inherit":true}}`, that way it gets the correct width from the theme.

Some themes also add custom class names to that Group block, but given that this depends on the theme, that's not something we can implement on our end.

This issue is about updating the templates under `/templates/block-templates/` and replace:

```HTML
```

with:

```HTML
<!-- wp:group {"layout":{"inherit":true}} -->
<div class="wp-block-group"><!-- wp:woocommerce/legacy-template {"template":"single-product"} /--></div>
<!-- /wp:group -->
```

Aljullu:
Fixes #5092.

This PR updates the templates as described in #5092 so the Legacy Template block is rendered inside a Group block with attribute `{"layout":{"inherit":true}}`. That makes it so the block doesn't occupy the entire width.

### Screenshots

_Before:_

Single product template

_After:_

Single product template

### Testing

1. Make sure your theme doesn't have any WC template and in Appearance > Templates you don't have any modified WC template either.
2. Go to the single product page or the shop page and verify the templates no longer take all available width, but instead they are constrained to the theme width (see screenshots above).

### DIFF
--- a/templates/block-templates/archive-product.html
+++ b/templates/block-templates/archive-product.html
@@ -1,3 +1,5 @@
 <!-- wp:template-part {"slug":"header"} /-->
-<!-- wp:woocommerce/legacy-template {"template":"archive-product"} /-->
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:woocommerce/legacy-template {"template":"archive-product"} /--></div>
+<!-- /wp:group -->
 <!-- wp:template-part {"slug":"footer"} /-->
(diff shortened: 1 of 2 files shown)


tjcafferkey (approved):
This LGTM and tests well. The failing tests look unrelated but feel free to rerun them, I can't see this change being responsible for them failing though.

On the themes I have been testing this on the content seems rather narrow but I guess like you mentioned this is down to the theme.

Issue #4958, opened 2021-10-18
The update to 6.1.0 triggers an error on the checkout page.


DarianDev:
## Describe the bug
The update to 6.1.0 kills the checkout page. 

"Something went wrong…

The checkout has encountered an unexpected error. Try reloading the page. If the error persists, please get in touch with us so we can assist.

Error: Cannot read property ‘call’ of undefined`"

## To reproduce

1. Update to 6.1.0
2. Click on 'Checkout'
3. See error

## Expected behavior
Fix the Error, and display the Checkout Page normally 

### WordPress (please complete the following information):
* Core version: 5.8.1
* Gutenberg version (if installed): [e.g. 4.8.0]
* WooCommerce version: 5.8
* WooCommerce Blocks version: 6.1.0
* Site language: german

### Any other plugins installed:
* Advanced Order Export For WooCommerce: version: 3.1.9, author: AlgolPlus
* Elementor: version: 3.4.5, author: Elementor.com
* Favicon by RealFaviconGenerator: version: 1.3.22, author: Philippe Bernard
* Kadence WooCommerce Email Designer: version: 1.4.8, author: Kadence WP
* Microsoft Advertising Universal Event Tracking (UET): version: 1.0.3, author: Microsoft Corporation
* Orbit Fox Companion: version: 2.10.8, author: Themeisle
* Rank Math SEO: version: 1.0.74, author: Rank Math, Automatische Aktualisierungen aktiviert
* Schema & Structured Data for WP & AMP: version: 1.9.85, author: Magazine3
* WooCommerce: version: 5.8.0, author: Automattic
* WooCommerce Blocks: version: 6.0.2, author: Automattic (latest version: 6.1.0)
* WooCommerce Payments: version: 3.1.0, author: Automattic
* WooCommerce PayPal Payments: version: 1.6.1, author: WooCommerce
* WooCommerce PDF Invoices & Packing Slips: version: 2.10.1, author: Ewout Fernhout
* WooCommerce Pixel Manager: version: 1.11.6, author: woopt
* WooCommerce Stripe Gateway: version: 5.6.2, author: WooCommerce
* WP Rollback: version: 1.7.1, author: Impress.org

### Desktop (please complete the following information):
* OS: Windows and Mac
* Browser: Chrome & Safari = every Browser

### Smartphone (please complete the following information):
* Device: all devices
* OS: all os versions
* Browser: all Browsers

## Additional context

[...]

PR #5094, opened 2021-11-08
Define chunkFilename with contenthash in lazy loaded assets

mikejolley:
This is a potential fix for issues like #4958 where we suspect errors are occurring due to cached lazy loaded scripts.

This busts the cache by appending a `?ver=` hash when webpack loads assets, without adding a hash to asset filenames.

Screenshot 2021-11-08 at 12 55 08

### Testing

These are dev facing testing instructions.

1. Do a build and see the `build/checkout-blocks/` directory. Files should not have a `?ver` suffix.
2. View the checkout block on the frontend, and open up browser tools > network. Confirm that the files are loaded with a `?ver=` hash. This is the cache-busting mechanism at work.
3. Edit one of the lazy-loaded files. Make a note of the `?ver=` hash from the network inspector. Change the contents e.g. add a console.log, and rebuild. Confirm the hash changes.

- Smoke test frontend functionality of cart/checkout blocks which lazy load inner blocks.

### Changelog

> Add cache-busting to lazy loaded checkout assets.

### DIFF
--- a/bin/webpack-configs.js
+++ b/bin/webpack-configs.js
@@ -178,7 +178,7 @@ const getMainConfig = ( options = {} ) => {
 			// i18n system relies on the hash of the filename, so changing that frequently would result in broken
 			// translations which we must avoid.
 			// @see https://github.com/Automattic/jetpack/pull/20926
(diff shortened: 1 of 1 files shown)


senadir (approved):
Works as advertised!

PR #5100, opened 2021-11-09
Add global style support to Mini Cart (button)

dinhtungdu:
This is a part of #4965.

This PR:
- Adds text and background color support to the Mini Cart button.
- Adds option to set the button transparent (default to `true`).
- By default, the text color affects total price, cart amount, and the cart icon color. The badge color doesn't change until the background color is set (black text and border, white background, and box-shadow).
- When the background color is set, the badge background will change and the box-shadow style will be disabled. The color of badge text/border will inherit the text color as well.

Notes: The current limitation of this PR is global style doesn't work for background.

Updates: The global background support was added using `__experimentalSelector`. It's possible to not use experimental features but in my discovery, it only works for the PHP side. See [this](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/try/global-background-style-mini-cart/src/BlockTypes/MiniCart.php#L259-L271).

### Screenshots

https://user-images.githubusercontent.com/5423135/140872646-c4e10cd5-e474-4d22-80a7-c6ed23ef6425.mov

### Testing

Follow the screencast above or:

0. Check out this and build the assets.
1. Add Mini Cart block to a page.
2. Set the text color, see it changes in the Editor.
3. Set the background color, see it changes in the editor.
4. Toggle transparent background, see it changes in the editor.
5. Save the page, see the style on the front-end.
6. Clear all customization for the block.
7. Set the global style to Mini Cart block using the Site Editor.
8. Edit the page created in step one, clear all colors. See the global style applies.
9. Set text color for the Mini cart block, see the text color change, but the background still inherits from the global style.
10. Toggle the transparent button option, see it works well with the global style.

### User Facing Testing
* [x] Same as above


assets/js/blocks/cart-checkout/mini-cart/style.scss
```diff
@@ -1,12 +1,19 @@
-.wc-block-mini-cart {
+.wc-block-mini-cart.wp-block-woocommerce-mini-cart {
+	background-color: transparent !important;
 	display: flex;
 	justify-content: flex-end;
+
+	&.transparent-button .wc-block-mini-cart__button {
+		background: transparent !important;
+	}
```
Aljullu: Given that we now have an attribute to define the button transparency, can we remove `background: transparent` from `.wc-block-mini-cart__button` [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/af3439dd69cb33ba1746b35cf5a5e5b7ddc7300f/assets/js/blocks/cart-checkout/mini-cart/style.scss#L8) and [this one](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/af3439dd69cb33ba1746b35cf5a5e5b7ddc7300f/assets/js/blocks/cart-checkout/mini-cart/style.scss#L15)?

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
```diff
@@ -31,6 +31,8 @@ const renderMiniCartFrontend = () => {
 		getProps: ( el: HTMLElement ) => ( {
 			isDataOutdated: el.dataset.isDataOutdated,
 			isInitiallyOpen: el.dataset.isInitiallyOpen === 'true',
+			backgroundColor: el.dataset.backgroundColor,
```
Aljullu: You will need to define `textColor` here too, right?

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -1,31 +1,94 @@
...
+	attributes,
+	setAttributes,
+}: Props ): ReactElement => {
+	const { transparentButton, backgroundColor, textColor, style } = attributes;
 	const blockProps = useBlockProps( {
-		className: 'wc-block-mini-cart',
+		className: classnames(
+			'wc-block-mini-cart wp-block-woocommerce-mini-cart',
+			{
+				'transparent-button': transparentButton,
```
Aljullu: Can we name this class `is-transparent`? This is a common pattern we use in other class names, to name modifiers starting with `is-`.

assets/js/blocks/cart-checkout/mini-cart/editor.scss
```diff
@@ -0,0 +1,3 @@
+.editor-styles-wrapper .wp-block-woocommerce-mini-cart.wc-block-mini-cart {
+	background-color: transparent !important;
+}
```
Aljullu: Why is this needed?

assets/js/blocks/cart-checkout/mini-cart/index.tsx
```diff
@@ -26,6 +26,9 @@ const settings = {
 	supports: {
 		html: false,
 		multiple: false,
+		color: true,
+		__experimentalSelector:
+			'.wc-block-mini-cart__button, .wc-block-mini-cart__badge',
```
Aljullu, line 41: Everything seemed to work fine for me when I removed these lines. Why was it needed? :thinking:

src/BlockTypes/MiniCart.php
```diff
@@ -219,6 +226,34 @@ protected function get_markup() {
 			$cart_contents_total += $cart->get_subtotal_tax();
 		}
 
+		$wrapper_classes = 'wc-block-mini-cart wp-block-woocommerce-mini-cart';
+		$classes         = '';
+		$style           = '';
```
Aljullu, line 229: I didn't test, so no idea if it would work, but just asking. Did you try attaching the styles in an inline `<style>` tag? This way we wouldn't need to read the styles again from the JS, we would just let the selectors work.

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
Aljullu: Taking another look at this... I wonder if we could read the class names from the elements instead of passing the colors in data attributes. I'm thinking something along these lines:

```JS
quantityBadgeClassnames: el.querySelector('.wc-block-mini-cart__quantity-badge').classList
```
Then, we would just pass that array to the React component so it can apply those classes. This way, we wouldn't need to call `getColorClassName()` from the JS code, which allows us to avoid one dependency: `@wordpress/block-editor`. :slightly_smiling_face:

assets/js/blocks/cart-checkout/mini-cart/style.scss
```diff
@@ -1,12 +1,19 @@
...
 }
 
 .wc-block-mini-cart__button {
 	align-items: center;
 	background: transparent;
 	border: none;
+	color: inherit;
+	cursor: pointer;
```
Aljullu: For accessibility reasons, I think we should leave the default cursor. `pointer` is used for links, but `default` is used for buttons and this behaves more like a button than a link.

src/BlockTypes/MiniCart.php
```diff
@@ -332,4 +367,22 @@ protected function get_cart_item_markup() {
...
 		</tr>';
 	}
+
+	/**
+	 * Get the supports array for this block type.
+	 *
+	 * @see $this->register_block_type()
+	 * @return string;
+	 */
+	protected function get_block_type_supports() {
```
Aljullu, line 380: I'm not sure to understand why this is needed either.

Aljullu (commented):
This is testing great. Good job dealing with the complexities of the Mini Cart block, which is rendered first in PHP and then in JS. :clap: 

I left some comments/questions below.

assets/js/blocks/cart-checkout/mini-cart/editor.scss
dinhtungdu: Oh, I'm sorry, this is from the previous implementation. Thanks for pointing this out!

assets/js/blocks/cart-checkout/mini-cart/style.scss
dinhtungdu: Thanks, I forgot to remove those lines.

src/BlockTypes/MiniCart.php
dinhtungdu: This is needed for inline style attributes instead. When the users choose a preset color, CSS selectors work fine, but if users choose custom colors, we inline style for them.

Initially, I only add inline style and classes for background and inherit the color from the parent element, but with the change to the global style, I do the same for color as well.

assets/js/blocks/cart-checkout/mini-cart/index.tsx
dinhtungdu: If you remove this, the global style doesn't work correctly. Without this, Gutenberg will add global style to `.wp-block-woocommerce-mini-cart` instead of the above selectors. This is fine for text color (by inheriting the parent element) but doesn't work for background color.

src/BlockTypes/MiniCart.php
dinhtungdu: This ensures the global style generated on the front end match with the global style in the Site Editor. We need this because of the `__experimentalSelectors`. Without registering the supports on the PHP side, the global style for Mini Cart on the front end (outside the Site Edittor) will be generated for `.wp-block-woocommerce-mini-cart`.

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
dinhtungdu: Thanks so much, it's a great idea!

src/BlockTypes/MiniCart.php
dinhtungdu: To be more clear, previously, before adding experimental selectors, we add classes and style to the wrapper element. As mentioned above, it doesn't work for the background.

After adding experimental selectors, we can target the CSS selectors to apply the global style. But it has a limitation: we can't target a specific selector for a specific style rule. We apply background style for the button and the badge (or the counter bubble, in the other words), the text color will come with it as well:

```
.wc-block-mini-cart__button, .wc-block-mini-cart__badge {
	background-color: var(--wp--preset--color--pale-pink);
	color: var(--wp--preset--color--vivid-purple);
}
```
Because of that, I decided to apply CSS classes and inline style to both button and badge.

assets/js/blocks/cart-checkout/mini-cart/editor.scss
dinhtungdu: Update: We still need it. The preset style classes in the editor come with `!important` that override our button style.

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
dinhtungdu: I update the related components to follow this approach.

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -1,31 +1,105 @@
...
+
+const MiniCartBlock = ( {
+	attributes,
+	setAttributes,
+}: Props ): ReactElement => {
+	const { transparentButton, backgroundColor, textColor, style } = attributes;
 	const blockProps = useBlockProps( {
-		className: 'wc-block-mini-cart',
+		className: classnames(
+			'wc-block-mini-cart wp-block-woocommerce-mini-cart',
```
Aljullu: `wp-block-woocommerce-mini-cart` shouldn't be necessary, it's already added by `useBlockProps()`, isn't it?
```suggestion
			'wc-block-mini-cart',
```

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -1,31 +1,105 @@
...
 /**
  * Internal dependencies
  */
 import QuantityBadge from './quantity-badge';
 
-const MiniCartBlock = (): ReactElement => {
+interface Attributes {
+	isInitiallyOpen?: boolean;
+	transparentButton: boolean;
+	colorClassNames?: string;
```
Aljullu: I think these attributes should contain `backgroundColor` and `textColor` instead of `colorClassNames`.

assets/js/blocks/cart-checkout/mini-cart/editor.scss
Aljullu: Ah, right, I could reproduce the issue. However, I think that might be a sign that there is something else that we are not doing correctly here. I took a look at the Gutenberg [Table block](https://github.com/WordPress/gutenberg/tree/trunk/packages/block-library/src/table) as an example: it also has to apply the background to an inner element (`<table>`) instead to the wrapper (`<figure>`); but they don't have this issue.

After some investigation, I think the difference is that they use [`__experimentalSkipSerialization`](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/table/block.json#L130). Do you think we could use it too?

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -1,31 +1,105 @@
...
-		className: 'wc-block-mini-cart',
+		className: classnames(
+			'wc-block-mini-cart wp-block-woocommerce-mini-cart',
+			{
+				'is-transparent': transparentButton,
+			}
+		),
 	} );
 
+	const backgroundClass = getColorClassName(
```
Aljullu, line 49: Also looking at the Table block, I discovered that there is a [`__experimentalUseColorProps`](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/table/edit.js#L17) that makes it easier to access the [class names](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/table/edit.js#L478) and [styles](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-library/src/table/edit.js#L482). Given that it's still experimental, I think it's ok if we leave the code in this PR as-is, but wondering if we should add a `@todo` comment so we remember to replace this implementation with `useColorProps()` once it's no longer experimental.

assets/js/blocks/cart-checkout/mini-cart/style.scss
```diff
@@ -1,18 +1,22 @@
-.wc-block-mini-cart {
+.wc-block-mini-cart.wp-block-woocommerce-mini-cart {
+	background-color: transparent !important;
 	display: flex;
 	justify-content: flex-end;
+
+	&.is-transparent .wc-block-mini-cart__button {
+		background: transparent !important;
```
Aljullu: Not a big deal, but I think using `background-color` would be better here? (It would keep it consistent with line 2):
```suggestion
		background-color: transparent !important;
```

Aljullu (commented):
Thanks for updating the PR @dinhtungdu. This is working great, I just left some more comments below, but overall this should be ready to merge. I'm not approving yet because I would like to double-test it once more after the changes are done.

assets/js/blocks/cart-checkout/mini-cart/index.tsx
nerrad: Is the GB team aware that we're using this experimental supports property and why we need to use it? If not, it might be handy to make sure they know somehow so there's awareness of it's usefulness for other projects when it comes time to evaluate graduating it.

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
dinhtungdu: Yeah, it's a copy pasta mistake. Thanks for pointing this out!

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
dinhtungdu: Yes, you're correct! Thank you! : )

assets/js/blocks/cart-checkout/mini-cart/index.tsx
dinhtungdu: @nerrad I'm not sure if they know about it or not. I will add comments explaining why we need them.

Edit: added comments for `__experimentalSelector` and `__experimentalSkipSerialization`.

assets/js/blocks/cart-checkout/mini-cart/editor.scss
dinhtungdu: @Aljullu I added `__experimentalSkipSerialization` in [`27a2584` (#5100)](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5100/commits/27a2584e1ba0b2a88581cd13453b65ee12694dc2)

assets/js/blocks/cart-checkout/mini-cart/index.tsx
Aljullu: Given the discussion in p1637141712378200-slack-C02FL3X7KR6, I think we can go ahead using those `__experimental` features for now, given that the Mini Cart block is gated to the Feature Plugin.

In the future, while we research global styles in a bigger scale for all blocks, we can get a list of all experimental features we need to use and get in contact with Gutenberg devs to know their plans with them.

assets/js/blocks/cart-checkout/mini-cart/block.tsx
```diff
@@ -191,10 +200,13 @@ const MiniCartBlock = ( {
 						getCurrencyFromPriceResponse( cartTotals )
 					) }
 				</span>
-				<QuantityBadge count={ cartItemsCount } />
+				<QuantityBadge
+					count={ cartItemsCount }
+					{ ...{ colorClassNames, style } }
```
Aljullu: Instead of passing the `style` object which uses a Gutenberg-specific format, I would already convert it to the CSS equivalent like you did for the `button` above:

```JS
			style={ {
					backgroundColor: style?.color?.background,
					color: style?.color?.text,
				} }
```

This way, the API of `QuantityBadge` is easier to understand because it uses the typical format for the `style` prop.

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -1,31 +1,107 @@
...
+							} )
+						}
+					/>
+				</PanelBody>
+			</InspectorControls>
+			<button
+				className={ classnames( 'wc-block-mini-cart__button', {
+					[ backgroundClass ]: backgroundClass,
+					[ textColorClass ]: textColorClass,
+				} ) }
```
Aljullu: Nit, I think you can rewrite this like:
```suggestion
				className={ classnames(
					'wc-block-mini-cart__button',
					backgroundClass,
					textColorClass
				) }
```
Keeps the code a bit cleaner and gets rid of a couple of TypeScript warnings.

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -1,31 +1,107 @@
...
-				<QuantityBadge count={ productCount } />
+				<QuantityBadge
+					count={ productCount }
+					colorClassNames={ classnames( {
+						[ backgroundClass ]: backgroundClass,
+						[ textColorClass ]: textColorClass,
+						'has-background':
+							backgroundClass || style?.color?.background,
+						'has-text-color': textColorClass || style?.color?.text,
+					} ) }
```
Aljullu: And this too:
```suggestion
					colorClassNames={ classnames(
						backgroundClass,
						textColorClass,
						{
							'has-background':
								backgroundClass || style?.color?.background,
							'has-text-color':
								textColorClass || style?.color?.text,
						}
					) }
```

src/BlockTypes/MiniCart.php
```diff
@@ -219,6 +226,34 @@ protected function get_markup() {
...
+
+		if ( ! empty( $attributes['textColor'] ) ) {
+			$classes .= sprintf(
+				' has-%s-color has-text-color',
+				esc_attr( $attributes['textColor'] )
+			);
+		} elseif ( ! empty( $attributes['style']['color']['text'] ) ) {
+			$style   .= 'color: ' . esc_attr( $attributes['style']['color']['text'] ) . ';';
+			$classes .= ' has-text-color';
+		}
```
Aljullu, line 248: Will it be possible to use some of the methods introduced in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5133 `StyleAttributesUtils` here?

No need to do it now, but might be a good follow-up issue.

cc @gigitux

Aljullu (approved):
Good job in this PR @dinhtungdu! Thanks for investigating and dealing with all the challenges of adding global styles to a block such as the Mini Cart.

I'm pre-approving but left some comments below.

src/BlockTypes/MiniCart.php
dinhtungdu: Sure, I added a todo to address your comment once the 5133 is merged.

src/BlockTypes/MiniCart.php
```diff
@@ -222,6 +224,39 @@ protected function get_markup() {
 			$cart_contents_total += $cart->get_subtotal_tax();
 		}
 
+		$wrapper_classes = 'wc-block-mini-cart wp-block-woocommerce-mini-cart';
```
dinhtungdu: Because we use `__experimentalSelector`, we don't need `wp-block-woocommerce-mini-cart` here anymore.

Aljullu (approved):

PR #5104, opened 2021-11-09
Fix PHP fatal error on Product tag pages due to wrong function name

Aljullu:
Small issue introduced in #5062. We were calling a function that was renamed so no longer existed with that name.

### Testing

Using a block theme:
1. Edit a product and assign a tag to it (ie: `Recommended`).
2. Go to the tag page (ie: `/product-tag/recommended/`).
3. Verify it renders correctly and there isn't any PHP error.


dinhtungdu (approved):
LGMT 👍🏽

PR #5105, opened 2021-11-09
Update taxonomy templates so Legacy Template blocks don't take all available width

### LINKED ISSUE #5092: Legacy Template block is always rendered full width

Aljullu:
Fixes #5092.

Same as #5093 but for taxonomy templates.

### Screenshots

_Before:_

Product category template

_After:_

Product category template

### Testing

1. Make sure your theme doesn't have any WC template and in Appearance > Templates you don't have any modified WC template either.
2. Go to the product category page (ie: `/product-category/clothing/`) and verify the templates no longer take all available width, but instead they are constrained to the theme width (see screenshots above).
3. If you want to test a product tag page you will need to assign some products to a specific tag (ie: _Recommended_) and then go to that tag page (ie: `/product-tag/recommended/`).

### DIFF
--- a/templates/block-templates/taxonomy-product_cat.html
+++ b/templates/block-templates/taxonomy-product_cat.html
@@ -1,4 +1,5 @@
 <!-- wp:template-part {"slug":"header"} /-->
 <!-- wp:group {"layout":{"inherit":true}} -->
 <div class="wp-block-group"><!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_cat"} /--></div>
+<!-- /wp:group -->
 <!-- wp:template-part {"slug":"footer"} /-->
--- a/templates/block-templates/taxonomy-product_cat.html
+++ b/templates/block-templates/taxonomy-product_cat.html
@@ -1,3 +1,4 @@
 <!-- wp:template-part {"slug":"header"} /-->
-<!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_cat"} /-->
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_cat"} /--></div>
 <!-- wp:template-part {"slug":"footer"} /-->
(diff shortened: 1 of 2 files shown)


templates/block-templates/taxonomy-product_cat.html
```diff
@@ -1,3 +1,4 @@
 <!-- wp:template-part {"slug":"header"} /-->
-<!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_cat"} /-->
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_cat"} /--></div>
```
frontdevde, line 3: Don't we need a closing comment here?
```suggestion
<div class="wp-block-group"><!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_cat"} /--></div>
<!-- /wp:group -->
```

templates/block-templates/taxonomy-product_tag.html
```diff
@@ -1,3 +1,4 @@
 <!-- wp:template-part {"slug":"header"} /-->
-<!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_tag"} /-->
+<!-- wp:group {"layout":{"inherit":true}} -->
+<div class="wp-block-group"><!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_tag"} /--></div>
```
frontdevde, line 3:
```suggestion
<div class="wp-block-group"><!-- wp:woocommerce/legacy-template {"template":"taxonomy-product_tag"} /--></div>
<!-- /wp:group -->
```

templates/block-templates/taxonomy-product_cat.html
Aljullu: Ah, good catch! You're right!

Aljullu:
Thanks for reviewing @frontdevde! I updated this PR with your suggestions, so it's ready for another look.

frontdevde (approved):
Nice change, LGTM! 👍

PR #5107, opened 2021-11-09
Add hydration for product data for Single Product block

### LINKED ISSUE #2698: Hydrate Product Data for the Single Product Block
We should be able to hydrate single product API Data on page load to speed up the initial render, like we have for other blocks.

gigitux:
Add hydration for product data for Single Product Block

Fixes #2698 

### Manual Testing

Check out this branch

1. Make sure you have Woo Blocks plugins installed
2. Create a new page and add the Single Product block
3. Save
4. Open the saved page
5. Check that browser doesn't do any request to `wc/store/products/${id}` endpoint

### Changelog

> Improve rendering performance for Single Product block.

### DIFF
--- a/src/BlockTypes/SingleProduct.php
+++ b/src/BlockTypes/SingleProduct.php
@@ -13,6 +13,17 @@ class SingleProduct extends AbstractBlock {
 	protected $block_name = 'single-product';
 
 	/**
+	 * Enqueue frontend assets for this block, just in time for rendering.
+	 *
+	 * @param array $attributes  Any attributes that currently are available from the block.
+	 */
+	protected function enqueue_assets( array $attributes ) {
+		parent::enqueue_assets( $attributes );
+		$product_id = $attributes['productId'];
+		$this->hydrate_from_api( $product_id );
+	}
+
+	/**
 	 * Get the editor script handle for this block type.
 	 *
 	 * @param string $key Data to get, or default to everything.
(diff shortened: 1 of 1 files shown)


src/BlockTypes/SingleProduct.php
```diff
@@ -12,6 +12,17 @@ class SingleProduct extends AbstractBlock {
...
 	protected $block_name = 'single-product';
 
+	/**
+	 * Enqueue frontend assets for this block, just in time for rendering.
+	 *
+	 * @param array $attributes  Any attributes that currently are available from the block.
+	 */
+	protected function enqueue_assets( array $attributes ) {
+		parent::enqueue_assets( $attributes );
+		$product_id = $attributes['productId'];
```
Aljullu: How do you feel about adding an extra protection here?

```suggestion
		$product_id = intval( $attributes['productId'] );
```

It's not really necessary because you already cast it to an int inside `hydrate_from_api()`, but I think it's a good practice to cast to the correct type the first time you read an external variable just in case in the future somebody refactors the code and uses that variable somewhere else assuming the type had been verified.

Aljullu (approved):
Looking great, it makes the block to render much faster! I left a comment below about adding an extra type check, but besides that, LGTM. :shipit:

src/BlockTypes/SingleProduct.php
gigitux: I agree with you! Thanks for the feedback 👍

PR #5115, opened 2021-11-09
Fix 'Country is required' error on the Cart block when updating shipping address

### LINKED ISSUE #5070: Getting 'Country is Required' even when entered to calculate shipping in basket
The latest WooCommerce Blocks is giving me the 'Country is Required' when calculating the shipping in the basket even when the full address is entered. 

Version 6.2.0

I have tested with other themes (Twenty Twenty One and Twenty Seventeen) and all other plugins off bar Woocommerce and problem remains.

I am guessing this is a bug unless as I have reproduced it loads of times my end.

alexflorisca:
This fixes an error on the cart page that was causing a 'Country is required' error to show up once a customer updates their shipping address. 

The main issue is that the default state of wc/store/cart has an empty billing address. This then gets hydrated by the client with JSON data that is injected onto the page by the API. This contains the country code set in WooCommerce settings. So initially, the billing address is empty, then gets populated with a country code. Between these two steps, the 
[use-customer-data](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/hooks/use-customer-data.ts) hook takes the value from wc/store/cart and assigns it to a [local state variable](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/hooks/use-customer-data.ts#L90). This means this local state variable `customerData` is out of sync with the wc/store/cart value. When a user updates their shipping address, an empty billing address is sent through the API call, and the server returns a "Country is required" error.

To solve this, we've used a [`useEffect()`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/hooks/use-customer-data.ts#L90) function to set the local state variable `customerData` to the shipping and billing addresses from wc/store/cart whenever they change. This only needs to run once, as any subsequent time will be the result of this effect or client side changes.

Fixes #5070 

#### Other Checks

### Screenshots
Screenshot 2021-11-09 at 17 08 46

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Open up the cart page
2. Change the shipping address in the shipping calculator and click Update
3. You should NOT see an error and the address should update to match what you entered
4. Go to Checkout and make sure the address that you entered has persisted in the checkout form
5. Complete checkout and make sure the address is correct on the order

### User Facing Testing
* [x] Same as above, or

### Changelog

> Fixed an issue that was causing an error when updating the address in the Cart block's shipping calculator.

### DIFF
--- a/assets/js/base/context/hooks/use-customer-data.ts
+++ b/assets/js/base/context/hooks/use-customer-data.ts
@@ -82,8 +82,10 @@ export const useCustomerData = (): {
 	const {
 		billingAddress: initialBillingAddress,
 		shippingAddress: initialShippingAddress,
+		cartIsHydrated,
 	}: Omit< CustomerData, 'billingData' > & {
 		billingAddress: CartResponseBillingAddress;
+		cartIsHydrated: boolean;
 	} = useStoreCart();
 
 	// State of customer data is tracked here from this point, using the initial values from the useStoreCart hook.
@@ -103,29 +105,18 @@ export const useCustomerData = (): {
 	>( false );
 
 	useEffect( () => {
-		if (
-			! hasCustomerDataSynced &&
-			( ! isShallowEqual(
-				customerData.billingData,
-				initialBillingAddress
-			) ||
-				! isShallowEqual(
-					customerData.shippingAddress,
-					initialShippingAddress
-				) )
-		) {
-			setHasCustomerDataSynced( true );
+		if ( cartIsHydrated && ! hasCustomerDataSynced ) {
 			const newCustomerData = {
 				shippingAddress: initialShippingAddress,
 				billingData: initialBillingAddress,
 			};
-			previousCustomerData.current = newCustomerData;
 			setCustomerData( newCustomerData );
+			setHasCustomerDataSynced( true );
 		}
 	}, [
+		cartIsHydrated,
 		initialBillingAddress,
 		initialShippingAddress,
-		customerData,
 		hasCustomerDataSynced,
 	] );
 	// Debounce updates to the customerData state so it's not triggered excessively.
--- a/assets/js/base/context/hooks/use-customer-data.ts
+++ b/assets/js/base/context/hooks/use-customer-data.ts
@@ -95,6 +95,39 @@ export const useCustomerData = (): {
 	// Store values last sent to the server in a ref to avoid requests unless important fields are changed.
 	const previousCustomerData = useRef< CustomerData >( customerData );
 
+	// Need to sync shipping and billing address from wp/store/cart to local state, because `customerData` is
+	// populated before the wc/store/cart is hydrated. We only need to run this the first time they're out of sync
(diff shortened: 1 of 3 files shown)


assets/js/base/context/hooks/use-customer-data.ts
```diff
@@ -95,6 +95,39 @@ export const useCustomerData = (): {
...
+	// Need to sync shipping and billing address from wp/store/cart to local state, because `customerData` is
+	// populated before the wc/store/cart is hydrated. We only need to run this the first time they're out of sync
+	// because subsequent times will be the result of this effect or client side changes.
+	const [ hasCustomerDataSynced, setHasCustomerDataSynced ] = useState<
+		boolean
+	>( false );
+
+	useEffect( () => {
+		if (
+			! hasCustomerDataSynced &&
```
mikejolley: This might be more readble if we returned early after checking just `hasCustomerDataSynced` before the rest of the conditional.

assets/js/base/context/hooks/use-customer-data.ts
```diff
@@ -95,6 +95,39 @@ export const useCustomerData = (): {
...
+			};
+			previousCustomerData.current = newCustomerData;
+			setCustomerData( newCustomerData );
+		}
+	}, [
+		initialBillingAddress,
+		initialShippingAddress,
+		customerData,
+		hasCustomerDataSynced,
+	] );
```
mikejolley, line 121: 👍🏻

mikejolley (commented):
I tested this case and this worked as described, good job!

This has me thinking--are there other consumers of `useStoreCart` which could be hit by this issue, or is it just in our implementation? I wonder if a hydration flag could instead be added to `useStoreCart` and exposed, so hooks like this one can react to changes.

e.g.
 
```
useEffect( () => {
// If cartIsHydrated became true, now we can update our values
}, [ cartIsHydrated ] );
```

Thoughts? cc @opr

alexflorisca:
Thanks for the review Mike! We've implemented your suggestion of creating a `cartIsHydrated` variable that we are now checking instead in `useCustomerData` hook. Can you take another look see if it's what you had in mind?

mikejolley (approved):
Thanks for making those changes, looks good to me! :shipit:

senadir:
This is a bit late but I'd like to leave my opinion anyway, feel free to ignore it.

It seems we're solving the problem at the wrong level? Instead of having another useState and useEffect there, can't we invalidate the local state if `use-customer-data` change? I had an issue like this before when local data doesn't get invalidated if the source change.

alexflorisca:
Closed in favour of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5129. See [this Slack thread](https://a8c.slack.com/archives/C8X6Q7XQU/p1636627461015900) for discussion

PR #5122, opened 2021-11-10
Add a placeholder to the product search block editor

### LINKED ISSUE #5078: Product Search (editor): the block's input can be edited
## Describe the bug
The input from the Product search block has a value(`Search products...`) instead of a placeholder in the editor. The user can edit this value.

## To reproduce

1. Go to a page and add the Product search block
2. Click on the input and see that you can remove the `Search products...` text

## Expected behavior
The user should not be able to edit the input, as we don't allow this in other blocks
The "Search products..." should be the input's placeholder instead of the input's value

## Screenshots
Screenshot 2021-11-05 at 12 40 42

alexflorisca:
The product search block, when added to a page in the editor, has a Search field with the initial value "Search products...". This represents the placeholder that will appear on the front end, but is not immediately clear to the user. We've added a placeholder to this search input to state this is the case, if the user deletes the initial value.

Related to #5078

#### Accessibility

### Screenshots
image

### Manual Testing

1. Add the Product Search block to any page in the editor
2. Delete the "Search Products..." value from the Search input
3. "Enter search placeholder text" should be displayed in the input as a placeholder
4. The text should disappear as you start typing

### Changelog

> Add placeholder text when modifying product search input in the editor.

### DIFF
--- a/assets/js/blocks/product-search/edit.js
+++ b/assets/js/blocks/product-search/edit.js
@@ -104,6 +104,7 @@ const Edit = ( {
 					<TextControl
 						className="wc-block-product-search__field input-control"
 						value={ placeholder }
+						placeholder="Enter search placeholder text"
 						onChange={ ( value ) =>
 							setAttributes( { placeholder: value } )
 						}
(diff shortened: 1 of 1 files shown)


mikejolley (approved):
This is clearer thanks 👍🏻 

The e2e test fail looks to be a timeout unrelated to this change so I restarted the job.

PR #5123, opened 2021-11-10
Filter Products By Price block: don't allow to insert negative values on inputs

### LINKED ISSUE #2695: [GlobalStep] User is able to set negative values in "Filter Products by Price" block.
**Environment:**
Windows 10: Chrome(v83.0.4103.61)(64-bit), Firefox(v77.0.1)(64-bit)

**Steps to Reproduce:**
1. Open a test site.
2. Login with valid credentials.
3. From dashboard go to Pages.
4. Add new page.
5. Add "All Products" block.
6. Add "Filter Products by Price" block
7. Add "Active Product filters" block
8. Click on Preview or Publish the page.
9. Go to the page having all the above block added.
10. Enter min price 10 and max price 100 in filter.
11. Now change max price to 0. 
12. Observe that negative values are displayed in min price field.

**Note:** Also All products show random products irrespective of the filter applied.

**Actual Result:**
User is able to set negative price  in "Filter by Price" block.

**Expected Result:**
User should not be able to set negative price as product price cannot be negative.

**Refer Artifact:** 
[#2695.zip](https://github.com/woocommerce/woocommerce-gutenberg-products-block/files/4765455/2695.zip)

gigitux:
This PR fixes some not unexpected behavior for the Filter Products By Price block:

### The user can insert a negative value in the inputs

Now, with this PR, the user can't insert a negative value. This is possible thanks to the prop `allowNegative`setted to `true`.

### The user can "break" the slider when inserting in the right input a value that is lower than the value on the left input
Now, with this PR, when the user inserts in the right input a value that is lower than the value on the left input, the latter is reset to 0.
This check is done thanks to `isValidMinValue`and `isValidMaxValue` and [this condition](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5123/files#diff-615d49ba906702d5afd8c7c4e079b5984633a74a1cd6eee0c1a74ef56c6a7d5bR223).

Fixes #2695 

### Manual Testing

Check out this branch

1. Make sure you have Woo Blocks plugins installed.
2. Add "All Products" block.
3. Add "Filter Products by Price" block
4. Add "Active Product filters" block
5. Click on Preview or Publish the page.
5. Go to the page having all the above block added.

Check that:

* the user can't insert in both inputs a negative number
* the user can't insert on input left a number that is greater than input on the right
* if the user inserts on the input on the right a number that is lower than input on the left, the component sets to 0 the minimum price


assets/js/base/components/formatted-monetary-amount/index.tsx
```diff
@@ -17,6 +17,8 @@ import './style.scss';
 interface FormattedMonetaryAmountProps {
 	className?: string;
 	displayType?: NumberFormatProps[ 'displayType' ];
+	allowNegative?: boolean;
+	isAllowed?: ( formattedValue: NumberFormatValues ) => boolean;
```
senadir, line 22: isAllowed is rather confusing, what are we supposed to allow here? what I understood from the PR is that it doesn’t allow negative numbers, for me `allowNegative` should be sufficient.

Also, do we need to pass a prop to `FormattedMonetaryAmount` just to support negative numbers? shouldn't just support them by default from what we pass to `value`?

assets/js/base/components/price-slider/index.js
```diff
@@ -212,9 +213,27 @@ const PriceSlider = ( {
 			) {
 				return;
 			}
+
 			const isMin = event.target.classList.contains(
 				'wc-block-price-filter__amount--min'
 			);
+
+			// When user insert in the minimum price input a value greater than the max price,
+			// we update the value of the slider to 0 - max.
```
senadir: Can't we just reject the min inserted and default to the previous min instead of reseting max as well? this is more a UX thing and not code related but if something changes, just reset it instead of reseting two things.

assets/js/base/components/price-slider/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+import { NumberFormatValues } from 'react-number-format';
+
+export const withMaxValueLimit = ( {
+	maxConstraint,
+	minorUnit,
+}: {
+	maxConstraint: number;
+	minorUnit: number;
+} ) => ( { floatValue }: NumberFormatValues ): boolean => {
+	const maxPrice = maxConstraint / 10 ** minorUnit;
```
senadir, line 16: We probably don't want to use this logic here given that we had issues with it, preferably we just deal with raw versions `1000` and let the final UI element handle formatting.

assets/js/base/components/price-slider/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+
+	return floatValue !== undefined && floatValue <= maxPrice;
+};
+
+export const withMinValueLimit = ( {
```
senadir: A doc block explaining the goal of this function would be great.
`with` prefix is usually used for HOCs, I find it rather confusing here. The logic in the return is also confusing from a first look and probably deserves to be dumped down or explained.

assets/js/base/components/formatted-monetary-amount/index.tsx
gigitux: >isAllowed is rather confusing, what are we supposed to allow here? what I understood from the PR is that it doesn’t allow negative numbers, for me allowNegative should be sufficient.

I used `isAllowed` because with the function `withMaxValueLimit` and `withMinValueLimit` we can validate the input when the user changes the value in the input. About the naming, I chose `isAllowed` to be consistent with the props of the library `react-number-format`.

> Also, do we need to pass a prop to FormattedMonetaryAmount just to support negative numbers? shouldn't just support them by default from what we pass to value?

Do you mean that we should implement some custom logic to check if the value is negative? I preferred to use a ready solution that `react-number-format` give

assets/js/base/components/price-slider/index.js
gigitux: > Can't we just reject the min inserted and default to the previous min instead of reseting max as well? this is more a UX thing and not code related but if something changes, just reset it instead of reseting two things.

My comment is probably unclear. For `max`, I mean the current max price. So what my code does is just reset to 0 the minimum price. 

What do you think about this: 

```suggestion
			// When the user inserts in the minimum price input a value greater than the max price,
			// we set to 0 the minimum price.
```

assets/js/base/components/price-slider/utils.ts
gigitux: > let the final UI element handle formatting.

This is code is not about formatting the number, but it is just about validation.

The code checks that on the right input the user can't insert a number that is minor then the number that there is in the left input. About the naming of the function, what do you think about `checkMaxValueLimit`?

assets/js/base/components/price-slider/utils.ts
gigitux:
```suggestion
// Check if that the value is greater than the lowest available price and minor than the current max price.
export const checkMinValueLimit = ( {
```

What do you think?

assets/js/base/components/formatted-monetary-amount/index.tsx
senadir: I still don't understand the goal of `isAllowed`, can't `allowNegative` do that instead? My comment is mostly why we need two.

assets/js/base/components/price-slider/index.js
senadir: Yes, this is great!

assets/js/base/components/formatted-monetary-amount/index.tsx
gigitux: The goal `isAllowed` is to do an extra check on the input field:
If on the right input the user write 30, on the left input the user can't write 31 (and this is the goal of the functions `withMinValueLimit` and `withMaxValueLimit`)

assets/js/base/components/price-slider/index.js
gigitux: I rephrased it again because I just re-checked the logic better.
```
// When the user inserts in the max price input a value less or equal than the current minimum price,
// we set to 0 the minimum price.
 ```

assets/js/base/components/price-slider/utils.ts
senadir: Given that `withMaxValueLimit` is a boolean function, anything with `has_` or `is_` should work. Maybe `isOutsideBound`? Naming is hard so don't stress it.

assets/js/base/components/price-slider/utils.ts
senadir: Yes! Much better, would also add a comment to `withMaxValueLimit`.

assets/js/base/components/formatted-monetary-amount/index.tsx
senadir: It took me sometime to understand `isAllowed`. I initially thought this is a prop you introduced, I see it's part of `react-format-money`, Not a great naming from their side really, but I see its value now.

senadir (approved):
I haven't done a feature review yet, but the code is clean! What's left is a changelog entry and a PR body description.

senadir (approved):
I've retested this and it's working fine, thank you for updating the code Luigi!

PR #5096, opened 2021-11-08
Fix e2e tests welcome guide/m1 mac support

mikejolley:
I had some issues running tests on m1 mac and needed to bump some dependencies to fix it. I also noticed test fails from the checkout block compatibility notice - debugging revealed the notice was being replaced by the WordPress Welcome Guide feature.

Several hours later, after fixing test fails in view switcher and cart, fingers crossed it's working. This PR has the following changes to improve e2e handling:

1. Copied the code from https://github.com/WordPress/gutenberg/blob/3da717b8d0ac7d7821fc6d0475695ccf3ae2829f/packages/e2e-test-utils/src/create-new-post.js to make the dismissal of the welcome notice more reliable. 
2. Removed unused test configs for e2e
3. Removed `<Columns>` which was unused and creating extra duplicate block wrappers which impact selectors
4. Added `span` element to view switcher to more easily select the buttons
5. Improved view switcher tests so click would fire correcrly.

**Testing**

- `npm install` to get latest dependencies
- Create a fresh env with `npm run wp-env destroy && npm run wp-env start`
- Run all e2e locally and ensure checkout tests pass. `npm run test:e2e`
- You can do a visual test with `npm run test:e2e-dev`
- You can run specific tests with `npm run test:e2e-dev tests/e2e/specs/backend/checkout.test.js`.

PR #5124, opened 2021-11-10
Downgrade @wordpress/scripts to 17.1.0 to fix dependency mistmach issue

Aljullu:
After the changes introduced in #5096, I wasn't able to run `npm run package-plugin` successfully. Instead, I was getting a dependency mismatch error. That was caused because `@wordpress/scripts` 19.1.0 depends on Webpack 5 while we use Webpack 4. For now, I downgraded `@wordpress/scripts` to 17.1.0, which is the last version that depended on Webpack 4.

### Testing

1. Run `npm run package-plugin` and verify it generates a ZIP instead of displaying an error.


mikejolley:
Need to test for regressions running tests on m1 macs cc @senadir

mikejolley (approved):
Tested `npm run package-plugin` and `npm run test:e2e` 👍🏻

PR #5125, opened 2021-11-11
Show decimals in mini cart regular price would not have decimals

### LINKED ISSUE #5018: `formatPrice` doesn't work with custom decimal separator
## To reproduce

1. Go to WooCommerce > Settings.
2. Change decimal separator to `,`.
3. Change decimal separator to `.`.
4. Save changes.
5. Add a block using `formatPrice` such as Mini Cart block to the page.
6. Add some products to cart, see the price was formatted using `.` as the decimal separator.

## Expected behavior
The formatted price should respect the store settings.

## Screenshots

image

* Core version: 5.8.1
* Gutenberg version (if installed): [2d6a363](https://github.com/WordPress/gutenberg/commit/2d6a363d55adeb50f382acd1a44af5ea85603b56J)
* WooCommerce version: 5.7.1 
* WooCommerce Blocks version: 6.2.0-dev
* Site language: English
* Any other plugins installed: WooCommerce Stripe Gateway 5.5.0

### LINKED ISSUE #5019: `formatPrice` strips decimals
## To reproduce

1. Go to WooCommerce > Settings.
2. Set the number of decimals separator to `2`.
4. Save changes.
5. Add a block using `formatPrice` such as Mini Cart block to the page.
6. Remove all products from the cart if any.
7. See the error

## Expected behavior
The formatted price should always contain decimals as set in the WooCommerce settings.

## Screenshots

**Storefront mini cart with formatted by `wc_price` (above) vs Mini Cart block using `formatPrice` (below):**

image

* Core version: 5.8.1
* Gutenberg version (if installed): [2d6a363](https://github.com/WordPress/gutenberg/commit/2d6a363d55adeb50f382acd1a44af5ea85603b56J)
* WooCommerce version: 5.7.1 
* WooCommerce Blocks version: 6.2.0-dev
* Site language: English
* Any other plugins installed: WooCommerce Stripe Gateway 5.5.0

nielslange:
Fixes #5018 & #5019

### Problem

In #5019, @dinhtungdu reported that the decimals in the mini cart block are not visible, when the price is 0. After viewing the unit tests of this fix, I noticed that this problem not only affects a $0 price, but also a price that has no decimals, such as $15.00. This price will then appear as $15 in the mini cart.

### Reason

To format the price, we convert the regular price into a decimal price by multiplying the regular price with 10 to the power of the number of decimals. Let's say the merchant sets the number of decimals to 2 and the cart contains products with a value of $23.05. We'd then convert the regular price to a decimal price as follows:

```
23.05 * 10 ^ 2 = 2305
```

Now, let's take a look what happens when we convert a $0 price:

```
0 * 10 ^ 2 = 0
```

To keep the decimal zeros, we'd expect the decimal price `000`, but we get the decimal price `0`. 

Let's also take a quick look what happens when we convert a $15 price:

```
15.00 * 10 ^ 2 = 1500
```

So far, this is ok. However, when formatting the price in https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/6cb995c49f11ee59f03951a57ab8a93b6de01753/packages/prices/utils/price.ts#L137, the decimals get lost:

```
1500 / 10 ^ 2 = 15
```

### Solution

To solve this problem, I'm splitting the price into an integer and a decimal part. That way, I can ensure that the decimals do not get lost independent if the cart holds a $0 price or a higher price, but without decimals, e.g. $15.

#### Automated testing

1. Run the following command in the terminal:
```
npm run test -i packages/prices/utils/test/price.js
```

##### Test $0.00 value

1. Go to `WP Admin » WooCommerce » Setting » General » Currency options`.
2. Set the number of decimals to `2`.
3. Create a test page with the mini cart block and save it.
4. Open the mini cart in the frontend.
5. See that the initial price shows as `$0.00`.
6. Hover over the price and see that it remains `$0.00`.

##### Test number of decimals

1. Go to `WP Admin » WooCommerce » Setting » General » Currency options`.
2. Set the number of decimals to `3`.
3. Open the mini cart in the frontend.
4. See that the initial price shows as `$0.000`.
5. Hover over the price and see that it remains `$0.000`.

##### Test separators

1. Go to `WP Admin » WooCommerce » Setting » General » Currency options`.
2. Set the thousand separator to `.` and the decimal separator to `,`.
3. Go to the frontend and add a product to the cart.
4. Increase the order quantity so that the price is above $1.000.
5. See that the thousand separator as `.` and the decimal separator as `,`.
6. Hover over the price and see that the thousand separator and the decimal separator remain the same.

##### Test currency and currency position

1. Go to `WP Admin » WooCommerce » Setting » General » Currency options`.
2. Set the currency to `Euro (€)` and the currency position to `Right with space`.
3. Go to the frontend and add a product to the cart.
4. See that the currency is `Euro (€)` and that it's place on the right-hand side, including a space.
5. Hover over the price and the currency and the currency position remain the same.

### Changelog

> Ensure that formated price shows the expected number of decimals, thousand separator and decimal separator.


packages/prices/utils/price.ts
```diff
@@ -134,7 +180,23 @@ export const formatPrice = (
...
+	);
+
+	const decimalValue: string | undefined = getDecimalValue(
+		priceInt,
+		currency.minorUnit
+	);
+
+	const formattedPrice: string =
+		currency.minorUnit > 0
+			? integerValue + currency.decimalSeparator + decimalValue
```
dinhtungdu, line 199: `decimalValue` can be `undefined`, so I think we should check for its value before appending it (and the decimal separator) for extra safety.

packages/prices/utils/price.ts
```diff
@@ -113,6 +115,50 @@ export const getCurrency = (
 	};
 };
 
+/**
+ * Get the integer value from the decimal price.
+ *
+ * @param {number} priceInt The price in decimals.
```
opr: What does "The price in decimals" mean? It is unclear to me what's being passed in.

packages/prices/utils/price.ts
```diff
@@ -113,6 +115,50 @@ export const getCurrency = (
...
+): string => {
+	return Math.floor( priceInt / 10 ** minorUnit )
+		.toString()
+		.replace( /\B(?=(\d{3})+(?!\d))/g, thousandSeparator );
+};
+
+/**
+ * Get the decimal value from the decimal price.
+ *
+ * @param {number} priceInt The price in decimals.
```
opr: What does "The price in decimals" mean? It is unclear to me what's being passed in.

opr (approved):
Works well for me, cool use of regex and I love the tests you added 👍🏼 

Besides what Tung has already mentioned, I've added a couple of comments relating to the documentation of functions, but these are only minor edits so I'll approve.

Please reconsider the testing instructions, I don't think there's any value getting people to experience the bug in trunk first before testing. Especially not for user-facing testing.

mikejolley:
@nielslange I only just noticed this PR and wanted to note that these issues were very similar:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5018
https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5019

I fixed the formatting in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5188

This is going to impact this PR. I also needed to separate the before/after decimal part and implemented a `splitDecimal` function based on `react-number-format` which is a package we use elsewhere. Might be worth a look?

packages/prices/utils/price.ts
nielslange: Thanks for pointing this out, @dinhtungdu. I added checks both for `currency.decimalSeparator` and `decimalValue` before using them.

packages/prices/utils/price.ts
nielslange: I changed `The price in decimals.` to `The price in minor unit value, e.g. 100 for $1.00.` to ensure that it's clear what is passed to the function.

packages/prices/utils/price.ts
nielslange: See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5125#discussion_r754783144

nielslange:
> @nielslange I only just noticed this PR and wanted to note that these issues were very similar:
> 
> #5018 #5019
> 
> I fixed the formatting in #5188
> 

Thanks for letting me know, @mikejolley. I just checked #5188 and noticed that it would break this PR, which is pretty much done. As this PR fixes both #5018 and #5019, I reverted your changes for now. 

> This is going to impact this PR. I also needed to separate the before/after decimal part and implemented a `splitDecimal` function based on `react-number-format` which is a package we use elsewhere. Might be worth a look?

In p1636638621033700/1636637781.030600-slack-C8X6Q7XQU, @senadir mentioned that `react-number-format` should be replaced in the future. That was one of the reasons, why I spin up this PR. I should have mentioned though that this PR not only fixes #5018, but also #5019. That was my bad.

nielslange:
> Please reconsider the testing instructions, I don't think there's any value getting people to experience the bug in trunk first before testing. Especially not for user-facing testing.

I'll adjust the testing instructions accordingly, @opr.

---

@dinhtungdu & @opr Do you two want to do a last review of this PR, or shall I go ahead and merge it?

opr (approved):
From my point of view it works and the code looks fine, so I can approve! I would like to cc @mikejolley since this affects work you did (Niels has reverted one of your PRs) please can you check over it too?

mikejolley:
>  @senadir mentioned that react-number-format should be replaced in the future. That was one of the reasons, why I spin up this PR.

I did not use `react-number-format` as an import, I copied the approach/function used to split the first and last part of the decimal (`splitDecimal`) which resulted in much less code that the splitting you original went with.

I think you've also reverted a fix where I added:

```
if ( ! currencyData?.currency_code ) {
```

I found that on the first load, `currencyData` can be empty (default state) with no `currency_code` which results in a flash of the wrong price. I tested this using console logs and noted the prices were not formatted correctly on first render. If you console.log the currencyData at that point you can see for yourself.

packages/prices/utils/test/price.js
```diff
@@ -53,13 +107,13 @@ describe( 'formatPrice', () => {
...
 		value          | expected
-		${ 1000 }      | ${ '$10' }
-		${ 0 }         | ${ '$0' }
+		${ 1000 }      | ${ '$10.00' }
+		${ 0 }         | ${ '$0.00' }
 		${ '' }        | ${ '' }
 		${ null }      | ${ '' }
 		${ undefined } | ${ '' }
 	`(
-		'correctly formats price given "$value" only as "$expected"',
```
mikejolley, line 62: Having $expected was useful when viewing the output of the tests.

packages/prices/utils/price.ts
```diff
@@ -76,7 +76,7 @@ export const getCurrencyFromPriceResponse = (
 		| Record< string, never >
 		| CartShippingPackageShippingRate
 ): Currency => {
-	if ( ! currencyData?.currency_code ) {
```
mikejolley, line 79: This prevents invalid (default state) `currencyData` taking precedence over the default `siteCurrencySettings`.

packages/prices/utils/price.ts
```diff
@@ -115,26 +115,48 @@ export const getCurrency = (
...
-		afterDecimal,
-	};
+/**
+ * Get the decimal value from the minor unit value.
+ *
+ * @param {number} priceInt The price in minor unit value, e.g. 100 for $1.00.
+ * @param {number} minorUnit The number of decimals to display.
+ * @return {string} The extracted decimal value.
+ */
+export const getDecimalValue = (
```
mikejolley, line 143: These functions seem to be parsing the decimal value, **and** formatting it, is that correct? The `splitDecimal` I added seemed more easy to understand. Can we not do the formatting part (adding 0's etc) later using a separate function? cc @senadir

nielslange:
Closing this PR in benefit of #5219.

PR #5129, opened 2021-11-11
Fix 'Country is required' error on the Cart block when updating shipping address

### LINKED ISSUE #5070: Getting 'Country is Required' even when entered to calculate shipping in basket

alexflorisca:
This fixes an error on the cart page that was causing a 'Country is required' error to show up once a customer updates their shipping address. 

The main issue is that the default state of wc/store/cart has an empty billing address. This then gets hydrated by the client with JSON data that is injected onto the page by the API. This contains the country code set in WooCommerce settings. So initially, the billing address is empty, then gets populated with a country code. Between these two steps, the 
[use-customer-data](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/hooks/use-customer-data.ts) hook takes the value from wc/store/cart and assigns it to a [local state variable](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/hooks/use-customer-data.ts#L90). This means this local state variable `customerData` is out of sync with the wc/store/cart value. When a user updates their shipping address, an empty billing address is sent through the API call, and the server returns a "Country is required" error.

To solve this, we've used a couple of `Refs` to store the shipping and billing addresses, which will update their values when the cart store has hydrated.  We then set the local state variable `customerData` to the shipping and billing addresses from wc/store/cart whenever they change. This only needs to run once, as any subsequent time will be the result of this effect or client side changes.

Fixes #5070 

#### Other Checks

### Screenshots
Screenshot 2021-11-09 at 17 08 46

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Open up the cart page
2. Change the shipping address in the shipping calculator and click Update
3. You should NOT see an error and the address should update to match what you entered
4. Go to Checkout and make sure the address that you entered has persisted in the checkout form
5. Complete checkout and make sure the address is correct on the order

### User Facing Testing
* [x] Same as above, or

### Changelog

> Fix an issue that was causing an error when updating the address in the Cart block's shipping calculator.

### DIFF
--- a/assets/js/base/context/hooks/use-customer-data.ts
+++ b/assets/js/base/context/hooks/use-customer-data.ts
@@ -14,7 +14,6 @@ import {
 import {
 	CartResponseBillingAddress,
 	CartResponseShippingAddress,
-	objectHasProp,
 	BillingAddressShippingAddress,
 	CartBillingAddress,
 	CartShippingAddress,
@@ -208,12 +207,6 @@ export const useCustomerData = (): {
 				debouncedCustomerData.shippingAddress;
 		}
 
-		if (
-			! objectHasProp( customerDataToUpdate, 'billing_address' ) &&
-			! objectHasProp( customerDataToUpdate, 'shipping_address' )
-		) {
-			return;
-		}
 		previousCustomerData.current = debouncedCustomerData;
 		updateCustomerData(
 			customerDataToUpdate as Partial< BillingAddressShippingAddress >
--- a/assets/js/base/context/hooks/use-customer-data.ts
+++ b/assets/js/base/context/hooks/use-customer-data.ts
@@ -82,11 +82,17 @@ export const useCustomerData = (): {
 	const {
 		billingAddress: initialBillingAddress,
 		shippingAddress: initialShippingAddress,
-		cartIsHydrated,
 	}: Omit< CustomerData, 'billingData' > & {
 		billingAddress: CartResponseBillingAddress;
-		cartIsHydrated: boolean;
 	} = useStoreCart();
+
+	// We need to store a ref to the shipping and billing addresses in order to re-populate the local state
+	// here with the hydrated values once they change. The refs keep track of the changing values of the addresses
+	// as they are hydrated.
+	const billingDataRef = useRef( {} );
+	const shippingAddressRef = useRef( {} );
+	billingDataRef.current = initialBillingAddress;
+	shippingAddressRef.current = initialShippingAddress;
 
 	// State of customer data is tracked here from this point, using the initial values from the useStoreCart hook.
 	const [ customerData, setCustomerData ] = useState< CustomerData >( {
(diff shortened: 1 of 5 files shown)


assets/js/base/context/hooks/use-customer-data.ts
```diff
@@ -146,23 +178,46 @@ export const useCustomerData = (): {
...
 		if (
-			! shouldUpdateAddressStore(
-				previousCustomerData.current.billingData,
-				debouncedCustomerData.billingData
-			) &&
-			! shouldUpdateAddressStore(
-				previousCustomerData.current.shippingAddress,
-				debouncedCustomerData.shippingAddress
-			)
+			! objectHasProp( customerDataToUpdate, 'billing_address' ) &&
```
mikejolley: Is this check necessary? It duplicates the check on line 191. We could remove either of them.

alexflorisca: Good spot, I've removed it 👍

assets/js/base/context/hooks/use-customer-data.ts
```diff
@@ -86,12 +89,40 @@ export const useCustomerData = (): {
 		billingAddress: CartResponseBillingAddress;
 	} = useStoreCart();
 
+	// We need to store a ref to the shipping and billing addresses in order to re-populate the local state
+	// here with the hydrated values once they change. The refs keep track of the changing values of the addresses
+	// as they are hydrated.
+	const billingDataRef = useRef( initialBillingAddress );
+	const shippingAddressRef = useRef( initialShippingAddress );
+	billingDataRef.current = initialBillingAddress;
+	shippingAddressRef.current = initialShippingAddress;
```
senadir: Why are doing this here if we have `hasCustomerDataSynced` and the condition below?

senadir (approved):
This is testing well! I'd be honest that the implementation is going a bit over my head as I was struggling to follow all the `shouldUpdates` but as long as the issue is fixed I'm okay with it right now.

assets/js/base/context/hooks/use-customer-data.ts
alexflorisca: It's because `initialBillingAddress` and `initialShippingAddress` will update once the cart store has been hydrated. So we are storing a ref to it to reflect the change

alexflorisca:
The shouldUpdateBillingAddress and shouldUpdateShippingAddress are just comparing when the 2 addresses have changed separately, so that we know which ones to send to the API request. It's a bunch of logic to work out weather to send either one of them or both based on what's actually changed.

assets/js/base/context/hooks/use-customer-data.ts
senadir: Shouldn't the initial value passed to `useRef` serve as that? I know that this line will execute each time the hook runs so are you updating the ref each time?
```
billingDataRef.current = initialBillingAddress;
```

PR #5132, opened 2021-11-12
Fix state validation to compare state codes, and only validate if a country is given

### LINKED ISSUE #5131: StoreAPI wc/store/cart returns address with empty state
## Describe the bug
The country state validation introduced in pr #4844 is seemingly always incorrect, causing the state to be empty. At least in my case, the value of the state address field is the state code, but `wc()->countries->get_states` returns an array of state names with the state codes as keys. Changing it to use `isset` fixed it in my case.

image

## To reproduce

1. Log in with a user that has a US billing address
2. Go to cart page
3. Enter the following in the console: `wp.data.select('wc/store/cart').getCartData().billingAddress`
4. And you can see the state is empty

image
image

## Environment
* WooCommerce version: 5.9
* WooCommerce Blocks version: 6.2
* Site language: EN
* Store location: USA

mikejolley:
Reported in #5131, and introduced in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4844:

- if the country is not set for whatever reason, rather than just getting the subset of valid states for a country, `wc()->countries->get_states()` returns the states of all countries keyed by country code. 
- the validation logic was using state names rather than state codes

This avoids the issue by checking if there is a country before getting an array of states, and then doing the validation based on state code.

Closes #5131

### Testing

#5131 has testing steps but these only apply due to a separate bug where billing data is not populated (due to hydration). I tested this only by logging data in the route to ensure correct states were populated.

The validation changes can also be tested by making requests to the API to verify validation is working. First disable nonce checks with:

```php
add_filter( 'woocommerce_store_api_disable_nonce_check', '__return_true' );
```

Then add something to cart in your rest api client, e.g. POST to `https://store.local/wp-json/wc/store/cart/add-item` with:

```json
{
	"id": 23,
	"quantity": 1
}
```

Then POST to checkout `https://store.local/wp-json/wc/store/checkout`:

```json
{
	"billing_address": {
		"first_name": "test",
		"last_name": "test",
		"company": "",
		"address_1": "test",
		"address_2": "",
		"city": "test",
		"state": "AL",
		"postcode": "90210",
		"country": "US",
		"phone": ""
	},
	"shipping_address": {
		"first_name": "test",
		"last_name": "test",
		"company": "",
		"address_1": "test",
		"address_2": "",
		"city": "test",
		"state": "INVALID-STATE",
		"postcode": "90210",
		"country": "US",
		"phone": ""
	},
	"payment_method": "bacs",
}
```

Try changing the state and ensure the response shows the parameter error.

### Changelog

> Fix state validation by only validating states in the API if a country is provided, and validating based on the state code.

### DIFF
--- a/src/StoreApi/Schemas/AbstractAddressSchema.php
+++ b/src/StoreApi/Schemas/AbstractAddressSchema.php
@@ -107,18 +107,47 @@ abstract class AbstractAddressSchema extends AbstractSchema {
 	}
 
 	/**
-	 * Format a state based on the country. If country has defined states, will return an upper case state code.
+	 * Get list of states for a country.
+	 *
+	 * @param string $country Country code.
+	 * @return array Array of state names indexed by state keys.
+	 */
+	protected function get_states_for_country( $country ) {
+		return $country ? array_filter( (array) \wc()->countries->get_states( $country ) ) : [];
+	}
+
+	/**
+	 * Validate provided state against a countries list of defined states.
+	 *
+	 * If there are no defined states for a country, any given state is valid.
+	 *
+	 * @param string $state State name or code (sanitized).
+	 * @param string $country Country code.
+	 * @return boolean Valid or not valid.
+	 */
+	protected function validate_state( $state, $country ) {
+		$states = $this->get_states_for_country( $country );
+
+		if ( count( $states ) && ! in_array( \wc_strtoupper( $state ), array_map( '\wc_strtoupper', array_keys( $states ) ), true ) ) {
+			return false;
+		}
+
+		return true;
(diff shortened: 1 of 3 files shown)


mikejolley:
Pause on this. Might be some confusion between names and codes: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5131#issuecomment-967209455

mikejolley:
Pushed some changes to correcrly match against the state code rather than the value. Should be ok now.

senadir (approved):
Code changes look wonderful and this is testing as expected!

PR #5133, opened 2021-11-12
Add global style for Product Categories List block

gigitux:
This is a part of #4965.

This PR:

- Adds text color support
- Adds link color support (WIP)
- Adds option to change font size (works but there is a problem)
- Adds option to change line-height

#### Link color support

~The link color support doesn't work with theme colors. We are investigating this problem (internal discussion `p1636652209321000-slack-C02FL3X7KR6`)~

With `useBlockProps` hook, I managed to make link color work 👍 

#### Font size 

~When the user changes the font size, the updated block is not rendered. We are investigating this problem (internal discussion `p1636643690311900-slack-C02FL3X7KR6`)~

~When the user changes the font size from the post/page editor everything works correctly. It seems that it doesn't work when the user changes the font size from the site editor (the font size is not applied to any element on the page)~

With the last version of Gutenberg, I'm not able to reproduce this problem 👍 

image
*before - editor*
image
*before - page*

image
*current implementation - new options in the editor*
image
*current implementation - new options in the editor*

image
*current implementation - block with custom styles*

### Manual Testing

How to test the changes in this PR:

Check out this branch:

1. Install and enable `Gutenberg` plugin
2. Install and enable `TT1 Blocks` theme
3. Add the All Product Categories list block to a post
4. Verify you can change the text color, font size, and line-height
5. Save
6. Go on the page and check if there are changes
7. Reset to default
8. Go to admin panel and click `Site Editor` 
9. Click on the Global styles icon 
10. Verify the Product Categories List block is shown and you can tweak its styles
11. Save 
12. Go on the page created earlier and check if all styles are applied correctly
10. Edit previous post/page again
11. Change again text color, font size, and line-height
12. Save
13. Check if these styles have priority over the styles from the Site editor.

### Changelog

> Added global styles (text color, link color, line height, and font size) to the Product Title block.

### DIFF
--- a/assets/js/blocks/product-categories/index.js
+++ b/assets/js/blocks/product-categories/index.js
@@ -13,6 +13,7 @@ import './style.scss';
 import Block from './block.js';
 
 registerBlockType( 'woocommerce/product-categories', {
+	apiVersion: 2,
 	title: __( 'Product Categories List', 'woo-gutenberg-products-block' ),
 	icon: {
 		src: <Icon srcElement={ list } />,
@@ -27,6 +28,14 @@ registerBlockType( 'woocommerce/product-categories', {
 	supports: {
 		align: [ 'wide', 'full' ],
 		html: false,
+		color: {
+			background: false,
+			link: true,
+		},
+		typography: {
+			fontSize: true,
+			lineHeight: true,
+		},
 	},
 	example: {
 		attributes: {
--- a/assets/js/blocks/product-categories/index.js
+++ b/assets/js/blocks/product-categories/index.js
@@ -5,6 +5,7 @@ import { __ } from '@wordpress/i18n';
 import { createBlock, registerBlockType } from '@wordpress/blocks';
 import { Icon, list } from '@woocommerce/icons';
 
+import { isFeaturePluginBuild } from '@woocommerce/block-settings';
 /**
  * Internal dependencies
  */
@@ -28,14 +29,16 @@ registerBlockType( 'woocommerce/product-categories', {
 	supports: {
 		align: [ 'wide', 'full' ],
 		html: false,
-		color: {
-			background: false,
-			link: true,
-		},
-		typography: {
-			fontSize: true,
-			lineHeight: true,
-		},
+		...( isFeaturePluginBuild() && {
+			color: {
+				background: false,
+				link: true,
+			},
+			typography: {
+				fontSize: true,
+				lineHeight: true,
+			},
+		} ),
 	},
 	example: {
 		attributes: {
--- a/src/BlockTypes/ProductCategories.php
+++ b/src/BlockTypes/ProductCategories.php
@@ -84,9 +84,13 @@ class ProductCategories extends AbstractDynamicBlock {
 			}
 		}
 
-		$classes_and_styles = $this->get_container_classes_and_styles( $attributes );
-		$classes            = $classes_and_styles[0];
-		$styles             = $classes_and_styles[1];
+		$classes_and_styles = StyleAttributesUtils::get_classes_and_styles_by_attributes(
+			$attributes,
+			array( 'line_height', 'text_color', 'font_size' )
+		);
+
+		$classes = $this->get_container_classes( $attributes ) . ' ' . $classes_and_styles['classes'];
+		$styles  = $classes_and_styles['styles'];
 
 		$output  = '<div class="wp-block-woocommerce-product-categories ' . esc_attr( $classes ) . '" style="' . esc_attr( $styles ) . '">';
 		$output .= ! empty( $attributes['isDropdown'] ) ? $this->renderDropdown( $categories, $attributes, $uid ) : $this->renderList( $categories, $attributes, $uid );
@@ -96,12 +100,12 @@ class ProductCategories extends AbstractDynamicBlock {
 	}
 
 	/**
-	 * Get the list of classes and styles to apply to this block.
+	 * Get the list of classes to apply to this block.
 	 *
 	 * @param array $attributes Block attributes. Default empty array.
-	 * @return array space-separated list of classes and space-separated list of inline styles.
+	 * @return string space-separated list of classes.
 	 */
-	protected function get_container_classes_and_styles( $attributes = array() ) {
+	protected function get_container_classes( $attributes = array() ) {
 
 		$classes = array( 'wc-block-product-categories' );
 
@@ -119,30 +123,7 @@ class ProductCategories extends AbstractDynamicBlock {
 			$classes[] = 'is-list';
 		}
 
-		$line_height_class_and_style = StyleAttributesUtils::get_line_height_class_and_style( $attributes );
-		$text_color_class_and_style  = StyleAttributesUtils::get_text_color_class_and_style( $attributes );
-		$font_size_class_and_style   = StyleAttributesUtils::get_font_size_class_and_style( $attributes );
-
-		$line_height_class = isset( $line_height_class_and_style['class'] ) ? $line_height_class_and_style['class'] : null;
-		$text_color_class  = isset( $text_color_class_and_style['class'] ) ? $text_color_class_and_style['class'] : null;
-		$font_size_class   = isset( $font_size_class_and_style['class'] ) ? $font_size_class_and_style['class'] : null;
-
-		$classes = array_filter(
-			array( $line_height_class, $text_color_class, $font_size_class )
-		);
-
-		$line_height_style = isset( $line_height_class_and_style['style'] ) ? $line_height_class_and_style['style'] : null;
-		$text_color_style  = isset( $text_color_class_and_style['style'] ) ? $text_color_class_and_style['style'] : null;
-		$font_size_style   = isset( $font_size_class_and_style['style'] ) ? $font_size_class_and_style['style'] : null;
-
-		$styles = array_filter(
-			array( $line_height_style, $text_color_style, $font_size_style )
-		);
-
-		$string_classes = implode( ' ', $classes );
-		$string_styles  = implode( ' ', $styles );
-
-		return array( $string_classes, $string_styles );
+		return implode( ' ', $classes );
 	}
 
 	/**
--- a/src/BlockTypes/ProductCategories.php
+++ b/src/BlockTypes/ProductCategories.php
@@ -99,7 +99,7 @@ class ProductCategories extends AbstractDynamicBlock {
 	 * Get the list of classes to apply to this block.
 	 *
 	 * @param array $attributes Block attributes. Default empty array.
-	 * @return array space-separated list of classes.
+	 * @return array space-separated list of classes and space-separated list of inline styles.
 	 */
 	protected function get_container_classes_and_styles( $attributes = array() ) {
 
--- a/src/BlockTypes/ProductCategories.php
+++ b/src/BlockTypes/ProductCategories.php
@@ -3,7 +3,6 @@ namespace Automattic\WooCommerce\Blocks\BlockTypes;
 
 use Automattic\WooCommerce\Blocks\Utils\StyleAttributesUtils;
 
-
 /**
  * ProductCategories class.
  */
@@ -50,7 +49,6 @@ class ProductCategories extends AbstractDynamicBlock {
 				'fontSize'       => $this->get_schema_string(),
 				'lineHeight'     => $this->get_schema_string(),
 				'style'          => array( 'type' => 'object' ),
-
 			)
 		);
 	}
@@ -86,7 +84,7 @@ class ProductCategories extends AbstractDynamicBlock {
 			}
 		}
 
-		$classes_and_styles = $this->get_container_classes( $attributes );
+		$classes_and_styles = $this->get_container_classes_and_styles( $attributes );
 		$classes            = $classes_and_styles[0];
 		$styles             = $classes_and_styles[1];
 
@@ -101,9 +99,9 @@ class ProductCategories extends AbstractDynamicBlock {
 	 * Get the list of classes to apply to this block.
 	 *
 	 * @param array $attributes Block attributes. Default empty array.
-	 * @return string space-separated list of classes.
+	 * @return array space-separated list of classes.
 	 */
-	protected function get_container_classes( $attributes = array() ) {
+	protected function get_container_classes_and_styles( $attributes = array() ) {
 
 		$classes = array( 'wc-block-product-categories' );
 
@@ -130,7 +128,7 @@ class ProductCategories extends AbstractDynamicBlock {
 		$font_size_class   = isset( $font_size_class_and_style['class'] ) ? $font_size_class_and_style['class'] : null;
 
 		$classes = array_filter(
-			[ $line_height_class, $text_color_class, $font_size_class ]
+			array( $line_height_class, $text_color_class, $font_size_class )
 		);
 
 		$line_height_style = isset( $line_height_class_and_style['style'] ) ? $line_height_class_and_style['style'] : null;
@@ -138,7 +136,7 @@ class ProductCategories extends AbstractDynamicBlock {
 		$font_size_style   = isset( $font_size_class_and_style['style'] ) ? $font_size_class_and_style['style'] : null;
 
 		$styles = array_filter(
-			[ $line_height_style, $text_color_style, $font_size_style ]
+			array( $line_height_style, $text_color_style, $font_size_style )
 		);
 
 		$string_classes = implode( ' ', $classes );
@@ -146,8 +144,6 @@ class ProductCategories extends AbstractDynamicBlock {
 
 		return array( $string_classes, $string_styles );
 	}
-
-
 
 	/**
 	 * Get categories (terms) from the db.
@@ -340,12 +336,12 @@ class ProductCategories extends AbstractDynamicBlock {
 
 		$link_color_class_and_style = StyleAttributesUtils::get_link_color_class_and_style( $attributes );
 
-		$link_color_style = $link_color_class_and_style['style'];
+		$link_color_style = isset( $link_color_class_and_style['style'] ) ? $link_color_class_and_style['style'] : '';
 
 		foreach ( $categories as $category ) {
 			$output .= '
 				<li class="wc-block-product-categories-list-item">
-				<a style="' . esc_attr( $link_color_style ) . '" href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
+					<a style="' . esc_attr( $link_color_style ) . '" href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
 				' . $this->getCount( $category, $attributes ) . '
 					' . ( ! empty( $category->children ) ? $this->renderList( $category->children, $attributes, $uid, $depth + 1 ) : '' ) . '
 				</li>
--- a/src/BlockTypes/ProductCategories.php
+++ b/src/BlockTypes/ProductCategories.php
@@ -90,7 +90,7 @@ class ProductCategories extends AbstractDynamicBlock {
 		$classes            = $classes_and_styles[0];
 		$styles             = $classes_and_styles[1];
 
-		$output  = '<div class="' . esc_attr( $classes ) . '" style="' . esc_attr( $styles ) . '">';
+		$output  = '<div class="wp-block-woocommerce-product-categories ' . esc_attr( $classes ) . '" style="' . esc_attr( $styles ) . '">';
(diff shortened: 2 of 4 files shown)


src/Utils/AttributesUtils.php
```diff
@@ -0,0 +1,145 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\Utils;
+
+/**
+ * AttributesUtils class used for getting class and style from attributes.
+ */
+class AttributesUtils {
```
gigitux: I'm not sure about this name. Maybe is it better `GlobalStyleUtils`?

assets/js/blocks/product-categories/index.js
```diff
@@ -13,6 +13,7 @@ import './style.scss';
 import Block from './block.js';
 
 registerBlockType( 'woocommerce/product-categories', {
+	apiVersion: 2,
```
Aljullu, line 17: With `apiVersion: 2`, you will need to pass `useBlockProps()` to the [edit component](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/6cb995c49f11ee59f03951a57ab8a93b6de01753/assets/js/blocks/product-categories/block.js#L187). Similarly to how it's done here:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/reviews/editor-container-block.js#L58

(I think you will need to replace the fragment `<>` with a wrapping div)

src/BlockTypes/ProductCategories.php
```diff
@@ -77,9 +86,11 @@ protected function render( $attributes, $content ) {
 			}
 		}
 
-		$classes = $this->get_container_classes( $attributes );
+		$classes_and_styles = $this->get_container_classes( $attributes );
+		$classes            = $classes_and_styles[0];
+		$styles             = $classes_and_styles[1];
 
-		$output  = '<div class="' . esc_attr( $classes ) . '">';
+		$output  = '<div class="' . esc_attr( $classes ) . '" style="' . $styles . '">';
```
Aljullu: We should escape every variable added to the output:
```suggestion
		$output  = '<div class="' . esc_attr( $classes ) . '" style="' . esc_attr( $styles ) . '">';
```

src/Utils/AttributesUtils.php
Aljullu: Agree, or maybe a mix between the two: `StyleAttributesUtil`? :thinking:

src/Utils/AttributesUtils.php
```diff
@@ -0,0 +1,145 @@
...
+	 * Get class and style for font-size from attributes.
+	 *
+	 * @param array $attributes Block attributes.
+	 *
+	 * @return string
+	 */
+	public static function get_font_size_class_and_style( $attributes ) {
+
+		$font_size        = $attributes['fontSize'];
+		$custom_font_size = $attributes['style']['typography']['fontSize'];
```
Aljullu: If you set `define( 'WP_DEBUG', 'True' );` in your site's `wp-config.php`, you will probably see errors appear on the screen because of this line. That's because `typography` might be missing inside the `style` object.

I think you will need to use `array_key_exists()` (or something similar?) every time you try to access an attribute inside `$attributes['style']`.

src/Utils/AttributesUtils.php
```diff
@@ -0,0 +1,145 @@
...
+			return array(
+				'class' => null,
+				'style' => sprintf( 'font-size: %s;', $custom_font_size ),
+			);
+		}
+
+		return array(
+			'class' => null,
+			'style' => null,
+		);
```
Aljullu: In this case, should we just return `null`?

src/BlockTypes/ProductCategories.php
```diff
@@ -301,10 +335,12 @@ protected function renderList( $categories, $attributes, $uid, $depth = 0 ) {
 	protected function renderListItems( $categories, $attributes, $uid, $depth = 0 ) {
 		$output = '';
 
+		$link_color_class_and_style = AttributesUtils::get_link_color_class_and_style( $attributes );
+
 		foreach ( $categories as $category ) {
 			$output .= '
 				<li class="wc-block-product-categories-list-item">
-					<a href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
+					<a style="' . $link_color_class_and_style['style'] . '" class="' . $link_color_class_and_style['class'] . '" href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
```
Aljullu: Here too, we should escape all attributes:
```suggestion
					<a style="' . esc_attr( $link_color_class_and_style['style'] ) . '" class="' . esc_attr( $link_color_class_and_style['class'] ) . '" href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
```

src/BlockTypes/ProductCategories.php
```diff
@@ -109,9 +121,31 @@ protected function get_container_classes( $attributes = array() ) {
 			$classes[] = 'is-list';
 		}
 
-		return implode( ' ', $classes );
+		$line_height_class_and_style = AttributesUtils::get_line_height_class_and_style( $attributes );
+		$text_color_class_and_style  = AttributesUtils::get_text_color_class_and_style( $attributes );
+		$font_size_class_and_style   = AttributesUtils::get_font_size_class_and_style( $attributes );
+
+		$classes = array_filter(
+			[ $line_height_class_and_style['class'], $text_color_class_and_style['class'], $font_size_class_and_style['class'] ],
```
Aljullu: `$line_height_class_and_style` might be null (according to [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7655afcc8a5b4867698cc7d7deb8e8985e3b949b/src/Utils/AttributesUtils.php#L108-L110)), so here you will also need to check if `class` exists before using it.

Same for the other objects.

Aljullu (commented):
Good job @gigitux adding global styles to the Product Categories List block. The utils class you created can be very useful for other PHP-rendered blocks. :clap: 

> The link color support doesn't work with theme colors.

I think that's because we need to add this class to the block: `wp-block-woocommerce-product-categories`. By default, Gutenberg appends a class like this to all blocks, but because we are rendering it in PHP, we need to do it manually. I think once you add it, the link styles will apply correctly.

> When the user changes the font size, the updated block is not rendered.

I'm not able to reproduce this issue (tested with Gutenberg `trunk` and theme Quadrat).

I added some other comments below, but overall the approach looks good to me. :ok_hand:

src/Utils/AttributesUtils.php
```diff
@@ -0,0 +1,145 @@
...
+ */
+class AttributesUtils {
+
+
+	/**
+	 * Get class and style for font-size from attributes.
+	 *
+	 * @param array $attributes Block attributes.
+	 *
+	 * @return string
```
dinhtungdu: We should update this comment to match the actual return value types.

src/Utils/AttributesUtils.php
```diff
@@ -0,0 +1,145 @@
...
+			font_size   => self::get_font_size_class_and_style( $attributes ),
+			link_color  => self::get_link_color_class_and_style( $attributes ),
+		);
+
+		return array_filter(
+			$classes_and_styles,
+			function ( $var ) {
+				return ! is_null( $var );
+			}
+		);
```
dinhtungdu: `array_filter` default callbackup already removes all falsy items, so we can use it without the callback.
```suggestion
		return array_filter( $classes_and_styles );
```

src/Utils/AttributesUtils.php
dinhtungdu: `StyleAttributesUtils` makes more sense to me, because the global part is automatically generated.

gigitux:
Thanks for your review. I fix the code.
I edited the description because:
- With `useBlockProps hook`, I managed to make link color work 👍
- When the user changes the font size from the post/page editor everything works correctly. It seems that it doesn't work when the user changes the font size from the site editor (the font size is not applied to any element on the page)

dinhtungdu (commented):
@gigitux Great work on this! The block style is working on my end, both editor and the front end. But the global style only works in the editor. We have this issue because the wrapper element of the block on the frontend doesn't have `wp-block-woocommerce-product-categories` class.

src/BlockTypes/ProductCategories.php
```diff
@@ -338,10 +338,14 @@ protected function renderList( $categories, $attributes, $uid, $depth = 0 ) {
...
 
+		$link_color_class_and_style = StyleAttributesUtils::get_link_color_class_and_style( $attributes );
+
+		$link_color_style = $link_color_class_and_style['style'];
+
 		foreach ( $categories as $category ) {
 			$output .= '
 				<li class="wc-block-product-categories-list-item">
-				<a href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
+				<a style="' . esc_attr( $link_color_style ) . '" href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
```
gigitux: Same here

src/BlockTypes/ProductCategories.php
```diff
@@ -77,9 +86,11 @@ protected function render( $attributes, $content ) {
 			}
 		}
 
-		$classes = $this->get_container_classes( $attributes );
+		$classes_and_styles = $this->get_container_classes( $attributes );
+		$classes            = $classes_and_styles[0];
+		$styles             = $classes_and_styles[1];
 
-		$output  = '<div class="' . esc_attr( $classes ) . '">';
+		$output  = '<div class="wp-block-woocommerce-product-categories ' . esc_attr( $classes ) . '" style="' . esc_attr( $styles ) . '">';
```
gigitux, line 95: There is a possibility that the style property will be empty. Do you think that is it necessary to exclude style property when the variable $styles is empty?
The same concept for class property and variable `$classes`

dinhtungdu: @gigitux  Empty style or class attribute is fine to me. But if you prefer a cleaner HTML output, go for it. I prefer empty attributes because of the readability. We will need some `if` condition checks to exclude the empty attributes, which makes the code harder to read IMO.

gigitux: >We will need some if condition checks to exclude the empty attributes, which makes the code harder to read IMO.

I agree with you! In any case, I preferred to point it out to you

src/BlockTypes/ProductCategories.php
```diff
@@ -42,6 +46,11 @@ protected function get_block_type_attributes() {
 				'hasEmpty'       => $this->get_schema_boolean( false ),
 				'isDropdown'     => $this->get_schema_boolean( false ),
 				'isHierarchical' => $this->get_schema_boolean( true ),
+				'textColor'      => $this->get_schema_string(),
+				'fontSize'       => $this->get_schema_string(),
+				'lineHeight'     => $this->get_schema_string(),
+				'style'          => array( 'type' => 'object' ),
+
```
dinhtungdu: This empty line should be removed.

src/BlockTypes/ProductCategories.php
```diff
@@ -109,9 +121,34 @@ protected function get_container_classes( $attributes = array() ) {
...
+		);
+
+		$string_classes = implode( ' ', $classes );
+		$string_styles  = implode( ' ', $styles );
+
+		return array( $string_classes, $string_styles );
 	}
 
+
+
```
dinhtungdu: Extra empty lines here should be removed too.

src/BlockTypes/ProductCategories.php
```diff
@@ -109,9 +121,34 @@ protected function get_container_classes( $attributes = array() ) {
...
+		$font_size_style   = isset( $font_size_class_and_style['style'] ) ? $font_size_class_and_style['style'] : null;
+
+		$styles = array_filter(
+			[ $line_height_style, $text_color_style, $font_size_style ]
+		);
+
+		$string_classes = implode( ' ', $classes );
+		$string_styles  = implode( ' ', $styles );
+
+		return array( $string_classes, $string_styles );
```
dinhtungdu: You're changing the return data type of this function, so you should update the comment docs block as well.

By the way, this method now returns an array of classes and styles, then the method name should be updated to reflect the return data.

src/BlockTypes/ProductCategories.php
```diff
@@ -109,9 +121,34 @@ protected function get_container_classes( $attributes = array() ) {
...
+		$line_height_class_and_style = StyleAttributesUtils::get_line_height_class_and_style( $attributes );
+		$text_color_class_and_style  = StyleAttributesUtils::get_text_color_class_and_style( $attributes );
+		$font_size_class_and_style   = StyleAttributesUtils::get_font_size_class_and_style( $attributes );
+
+		$line_height_class = isset( $line_height_class_and_style['class'] ) ? $line_height_class_and_style['class'] : null;
+		$text_color_class  = isset( $text_color_class_and_style['class'] ) ? $text_color_class_and_style['class'] : null;
+		$font_size_class   = isset( $font_size_class_and_style['class'] ) ? $font_size_class_and_style['class'] : null;
+
+		$classes = array_filter(
+			[ $line_height_class, $text_color_class, $font_size_class ]
```
dinhtungdu: We should use only one array style. Same for line 141 below.

src/BlockTypes/ProductCategories.php
```diff
@@ -301,11 +338,15 @@ protected function renderList( $categories, $attributes, $uid, $depth = 0 ) {
...
+
+		$link_color_style = $link_color_class_and_style['style'];
+
 		foreach ( $categories as $category ) {
 			$output .= '
 				<li class="wc-block-product-categories-list-item">
-					<a href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
-					' . $this->getCount( $category, $attributes ) . '
+				<a style="' . esc_attr( $link_color_style ) . '" href="' . esc_attr( get_term_link( $category->term_id, 'product_cat' ) ) . '">' . $this->get_image_html( $category, $attributes ) . esc_html( $category->name ) . '</a>
+				' . $this->getCount( $category, $attributes ) . '
```
dinhtungdu: I'm not sure why the indentation of these lines needs to change. They look correct.

src/BlockTypes/ProductCategories.php
```diff
@@ -1,11 +1,15 @@
 <?php
 namespace Automattic\WooCommerce\Blocks\BlockTypes;
 
+use Automattic\WooCommerce\Blocks\Utils\StyleAttributesUtils;
+
+
```
dinhtungdu: Another extra line.

src/BlockTypes/ProductCategories.php
```diff
@@ -301,11 +338,15 @@ protected function renderList( $categories, $attributes, $uid, $depth = 0 ) {
 	protected function renderListItems( $categories, $attributes, $uid, $depth = 0 ) {
 		$output = '';
 
+		$link_color_class_and_style = StyleAttributesUtils::get_link_color_class_and_style( $attributes );
+
+		$link_color_style = $link_color_class_and_style['style'];
```
dinhtungdu: This throws a warning on my end: `Warning: Trying to access array offset on value of type null in /wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTypes/ProductCategories.php on line 343`.

dinhtungdu (commented):
@gigitux This is working great on my end, both local and global block styles are working. I left some comments about coding standards and minor issues, please take a look.

src/BlockTypes/ProductCategories.php
gigitux: What do you mean with `use only one array style`. Maybe do you mean that I should use `array`?

```suggestion
			array($line_height_class, $text_color_class, $font_size_class)
```

dinhtungdu: @gigitux  Yeah, I mean the syntax, we should use one style to define arrays.

src/BlockTypes/ProductCategories.php
```diff
@@ -90,9 +99,10 @@ protected function render( $attributes, $content ) {
 	 * Get the list of classes to apply to this block.
```
dinhtungdu: This comment needs to be updated as well.

src/Utils/StyleAttributesUtils.php
```diff
@@ -0,0 +1,160 @@
...
+			link_color  => self::get_link_color_class_and_style( $attributes ),
+		);
+
+		return array_filter(
+			$classes_and_styles
+		);
+
+	}
+
+
```
dinhtungdu: Consecutive empty lines here.

src/Utils/StyleAttributesUtils.php
```diff
@@ -0,0 +1,160 @@
...
+	 * @param array $attributes Block attributes.
+	 *
+	 * @return array
+	 */
+	public static function get_classes_and_styles_by_attributes( $attributes ) {
+		$classes_and_styles = array(
+			line_height => self::get_line_height_class_and_style( $attributes ),
+			text_color  => self::get_text_color_class_and_style( $attributes ),
+			font_size   => self::get_font_size_class_and_style( $attributes ),
+			link_color  => self::get_link_color_class_and_style( $attributes ),
```
dinhtungdu: The array keys need to be string

src/Utils/StyleAttributesUtils.php
```diff
@@ -0,0 +1,160 @@
...
+		} else {
+			return array(
+				'class' => null,
+				'style' => sprintf( 'color: %s;', $link_color ),
+			);
+		}
+
+	}
+
+
```
dinhtungdu: Consecutive empty lines here.

src/Utils/StyleAttributesUtils.php
```diff
@@ -0,0 +1,160 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\Utils;
+
+/**
+ * StyleAttributesUtils class used for getting class and style from attributes.
+ */
+class StyleAttributesUtils {
+
+
```
dinhtungdu: Consecutive empty lines here.

src/BlockTypes/ProductCategories.php
```diff
@@ -109,7 +119,30 @@ protected function get_container_classes( $attributes = array() ) {
...
+			array( $line_height_class, $text_color_class, $font_size_class )
+		);
+
+		$line_height_style = isset( $line_height_class_and_style['style'] ) ? $line_height_class_and_style['style'] : null;
+		$text_color_style  = isset( $text_color_class_and_style['style'] ) ? $text_color_class_and_style['style'] : null;
+		$font_size_style   = isset( $font_size_class_and_style['style'] ) ? $font_size_class_and_style['style'] : null;
+
+		$styles = array_filter(
+			array( $line_height_style, $text_color_style, $font_size_style )
+		);
```
dinhtungdu: I'm wondering if we can use `StyleAttributesUtils:get_classes_and_styles_by_attributes` here? Something like this:
```suggestion
		foreach( StyleAttributesUtils::get_classes_and_styles_by_attributes( $attributes ) as $key => $item ) {
			// We skip link_color for wrapper element.
			if ( 'link_color' === $key ) {
				continue;
			}

			if ( ! empty( $item['class'] ) ) {
				$classes[] = $item['class'];
			}

			if ( ! empty( $item['style'] ) ) {
				$styles[] = $item['style'];
			}
		}
```

dinhtungdu (commented):
Hi @gigitux I'm not sure if you finished updating the PR but I left some more comments on the latest changes.

src/BlockTypes/ProductCategories.php
gigitux: >I'm wondering if we can use StyleAttributesUtils:get_classes_and_styles_by_attributes here? Something like this:

Maybe in the future, we could add more attributes and if this happens we should update this `foreach`. I would prefer to use an immutable approach. What do you think?

dinhtungdu: @gigitux I agree with an immutable approach. But I think we still need too much code to calculate classes/styles string to use in markup. I have an idea to refactor the util in #5277, please take a look!

gigitux: > I agree with an immutable approach. But I think we still need too much code to calculate classes/styles string to use in markup. I have an idea to refactor the util in #5277, please take a look!

Done and approved! Thanks for your great feedback!

dinhtungdu (approved):
LGTM 🚢

PR #5134, opened 2021-11-12
Add version to doing it wrong calls

### LINKED ISSUE #5074: Too few arguments to function _doing_it_wrong()
## Describe the bug
The call to the function `_doing_it_wrong()` [at line 67 inside src/Integrations/IntegrationRegistry.php](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/bdf7ab73889d2ec12af5db82ecc7334c202f0152/src/Integrations/IntegrationRegistry.php#L67) has a missing parameter: $name.

## Expected behavior
A correct function call with 3 parameters is expected otherwise a PHP Fatal error is produced. 

* Core version: [e.g. 5.8.0]
* WooCommerce Blocks version: [e.g. 6.2.0]

mikejolley:
Some environments report an error when `_doing_it_wrong` is called without a `$version`. This PR adds a version to each call to mitigate. 

Fixes #5074

I used `4.6.0` for the missing versions which was the version [this PR was part of which last touched them](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3829).

### Testing

These are only triggered by integrations that are doing something wrong, so there is nothing user-facing to test here and just requires code review.

### DIFF
--- a/src/BlockTypes/AbstractBlock.php
+++ b/src/BlockTypes/AbstractBlock.php
@@ -107,7 +107,7 @@ abstract class AbstractBlock {
 	 */
 	protected function initialize() {
 		if ( empty( $this->block_name ) ) {
-			_doing_it_wrong( __METHOD__, esc_html( __( 'Block name is required.', 'woo-gutenberg-products-block' ) ), '4.5.0' );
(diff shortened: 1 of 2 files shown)


senadir (approved):
Checks out!

PR #5135, opened 2021-11-12
Pass to payment methods a wrapper component that handles the loading state

### LINKED ISSUE #5013: Pass a loading component for payment methods to use via the payment method interface
In order to have a consistent UI for loading state presented to shoppers in the C&C blocks, we need to provide registered payment methods a component that they could use while they don't want the payment method to be available (because of internal reasons: a call, etc..). 

More details on pca54o-2lM-p2#comment-2923

ralucaStan:
In order to have a consistent UI for loading state presented to shoppers in the C&C blocks, we need to provide registered payment methods a component that they could use while they don't want the payment method to be available.

This PR:
-  exposes the LoadingMask component via the payment method interface, under the components key
- updates the docs listing the data that we pass to payment methods

Fixes #5013 

#### Other Checks

- [x] I've updated [this doc](../docs/blocks/feature-flags-and-experimental-interfaces.md) for any feature flags or experimental interfaces implemented in this pull request.

### Screenshots

Images don't reflect the actual size in the page, they are made to support comparing the changes. 
| Before   | After |
| Screenshot 2021-11-12 at 18 00 12 | Screenshot 2021-11-12 at 17 59 28 Screenshot 2021-11-12 at 17 59 52 |

### Testing
1. Register a payment method
2. Make sure you have access to the LoadingMask component under the `components` key.

### Changelog

> Pass to payment methods a wrapper component that handles the loading state


mikejolley (approved):
Looks good to me. Just to note, there is some overlap with https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5177 where I've added proper typescript definitions. Should we combine documentation/descriptions and refer to the TS rather than repeating in a doc, or have both?

ralucaStan:
> Looks good to me. Just to note, there is some overlap with #5177 where I've added proper typescript definitions. Should we combine documentation/descriptions and refer to the TS rather than repeating in a doc, or have both?

I think there is some extra info in the docs and it might be worth keeping both. Also, as the payment gateways are a topic of interest it might help having this overview in the docs for somebody who just wants to learn more. We can reconsider later.

PR #5139, opened 2021-11-15
Make order note block removable

senadir:
Order Note block was introduced as locked in Checkout i2. Originally, it was shown via a settings, this PR makes the block removable so it can be removed just like the settings based option.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5121

### Manual Testing

How to test the changes in this Pull Request:

1. Insert Checkout block.
2. Remove Order note block. It should not be added back, save.
3. On frontend, you should not see teh block.

### Changelog

> Fix an issue in which merchants couldn't remove Order Note block.

PR #5140, opened 2021-11-15
Move Checkout i1 -> i2 migration code to registerBlockType 

senadir:
This issue was spun from #5139

We had some Checkout i1 -> i2 related logic in edit.js, this could be a confusing given it's migration related only, so I moved the code to `registerBlockType`, `migrate` function allows us to do just that.

I also respected some attributes in frontend so Order Note and Terms blocks are not force shown on frontend if the settings are false.

### Testing

1. Checkout an old tag `git checkout v5.9.1-dev`
2. Insert two pages of Checkout:
    1. "Checkout", don't change any attributes, leave as is.
    2. "Checkout - modified" change attributes like show order note (to false) and show policy (to false). Save both.
3. Checkout this branch and build.
4. Visit "Checkout" on frontend, you should see the terms and condition block, the order note block.
5. Visit "Checkout - modified" on frontend, you should not see the terms and order note blocks.
6. Visit "Checkout" on the editor, it should be a Checkout i2 now, with terms and order note blocks present.
7. Visit "Checkout - modified" on the editor, it should be a Checkout i2 now, with terms and order note removed, you can insert them.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-fields-block/edit.tsx
+++ b/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-fields-block/edit.tsx
@@ -10,10 +10,7 @@ import type { TemplateArray } from '@wordpress/blocks';
 /**
  * Internal dependencies
  */
-import {
-	useCheckoutBlockControlsContext,
-	useCheckoutBlockContext,
-} from '../../context';
+import { useCheckoutBlockControlsContext } from '../../context';
 import { useForcedLayout, getAllowedBlocks } from '../../../shared';
 import './style.scss';
 
@@ -32,12 +29,6 @@ export const Edit = ( {
 			attributes?.className
 		),
 	} );
-	const {
-		showOrderNotes,
-		showPolicyLinks,
-		showReturnToCart,
-		cartPageId,
-	} = useCheckoutBlockContext();
 	const allowedBlocks = getAllowedBlocks( innerBlockAreas.CHECKOUT_FIELDS );
 
 	const {
@@ -51,20 +42,9 @@ export const Edit = ( {
 		[ 'woocommerce/checkout-billing-address-block', {}, [] ],
 		[ 'woocommerce/checkout-shipping-methods-block', {}, [] ],
 		[ 'woocommerce/checkout-payment-block', {}, [] ],
-		showOrderNotes
-			? [ 'woocommerce/checkout-order-note-block', {}, [] ]
-			: false,
-		showPolicyLinks
-			? [ 'woocommerce/checkout-terms-block', {}, [] ]
-			: false,
-		[
-			'woocommerce/checkout-actions-block',
-			{
-				showReturnToCart,
-				cartPageId,
-			},
-			[],
-		],
+		[ 'woocommerce/checkout-order-note-block', {}, [] ],
+		[ 'woocommerce/checkout-terms-block', {}, [] ],
+		[ 'woocommerce/checkout-actions-block', {}, [] ],
 	].filter( Boolean ) as unknown ) as TemplateArray;
 
 	useForcedLayout( {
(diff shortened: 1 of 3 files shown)


assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-fields-block/edit.tsx
```diff
@@ -51,20 +42,9 @@ export const Edit = ( {
...
-			? [ 'woocommerce/checkout-terms-block', {}, [] ]
-			: false,
-		[
-			'woocommerce/checkout-actions-block',
-			{
-				showReturnToCart,
-				cartPageId,
-			},
-			[],
-		],
```
senadir, line 67: Those should be handled by:
1. Migration code if we're migrating.
2. default attributes if we're inserting fresh.

src/BlockTypes/Checkout.php
```diff
@@ -96,10 +96,10 @@ protected function render( $attributes, $content ) {
...
 					<div data-block-name="woocommerce/checkout-billing-address-block" class="wp-block-woocommerce-checkout-billing-address-block"></div>
 					<div data-block-name="woocommerce/checkout-shipping-methods-block" class="wp-block-woocommerce-checkout-shipping-methods-block"></div>
-					<div data-block-name="woocommerce/checkout-payment-block" class="wp-block-woocommerce-checkout-payment-block"></div>
-					<div data-block-name="woocommerce/checkout-order-note-block" class="wp-block-woocommerce-checkout-order-note-block"></div>
-					<div data-block-name="woocommerce/checkout-terms-block" class="wp-block-woocommerce-checkout-terms-block"></div>
-					<div data-block-name="woocommerce/checkout-actions-block" class="wp-block-woocommerce-checkout-actions-block"></div>
+					<div data-block-name="woocommerce/checkout-payment-block" class="wp-block-woocommerce-checkout-payment-block"></div>' .
+					( false === $attributes['showOrderNotes'] ? '' : '<div data-block-name="woocommerce/checkout-order-note-block" class="wp-block-woocommerce-checkout-order-note-block"></div>' ) .
```
senadir, line 100: We don't register our attributes on PHP, so the default value of `showOrderNotes` === `true` won't be saved, and `showOrderNotes` won't be set. we only care if `showOrderNotes` is `false` and so would look for it, return nothing, or return the actual html.

src/BlockTypes/Checkout.php
```diff
@@ -96,10 +96,10 @@ protected function render( $attributes, $content ) {
...
-					<div data-block-name="woocommerce/checkout-payment-block" class="wp-block-woocommerce-checkout-payment-block"></div>
-					<div data-block-name="woocommerce/checkout-order-note-block" class="wp-block-woocommerce-checkout-order-note-block"></div>
-					<div data-block-name="woocommerce/checkout-terms-block" class="wp-block-woocommerce-checkout-terms-block"></div>
-					<div data-block-name="woocommerce/checkout-actions-block" class="wp-block-woocommerce-checkout-actions-block"></div>
+					<div data-block-name="woocommerce/checkout-payment-block" class="wp-block-woocommerce-checkout-payment-block"></div>' .
+					( isset( $attributes['showOrderNotes'] ) && false === $attributes['showOrderNotes'] ? '' : '<div data-block-name="woocommerce/checkout-order-note-block" class="wp-block-woocommerce-checkout-order-note-block"></div>' ) .
+					( isset( $attributes['showPolicyLinks'] ) && false === $attributes['showPolicyLinks'] ? '' : '<div data-block-name="woocommerce/checkout-terms-block" class="wp-block-woocommerce-checkout-terms-block"></div>' ) .
+					'<div data-block-name="woocommerce/checkout-actions-block" class="wp-block-woocommerce-checkout-actions-block"></div>
```
senadir, line 102: We don't register our attributes on PHP, so the default value of `showOrderNotes` === `true` won't be saved, and `showOrderNotes` won't be set. we only care if `showOrderNotes` is  set and `false` and so would look for it, return nothing, or return the actual html.

opr (approved):
Tests well for me and the code looks fine. 🚢  it

PR #5142, opened 2021-11-15
Mini Cart block fatal error caused by append_script_and_deps_src

### LINKED ISSUE #5073: Mini Cart block breaks Navigation block.
## Describe the bug
When adding the Navigation block to an existing page that contains Mini Cart Block, there is a fatal error that breaks the Navigation block.

> PHP Fatal error: Uncaught TypeError: count(): Argument woocommerce/woocommerce-blocks#12058 ($value) must be of type Countable|array, null given in /wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTypes/MiniCart.php:169

## To reproduce

1. Create a new page and add the Mini Cart block to that page. Save it and see no error.
2. Edit the newly added page.
4. Open the Developer inspector > Network Tab
3. Add Navigation Block.
4. See the Navigation block broken.
5. See the failed request.
6. See the error in the console log.

## Expected behavior
Mini Cart shouldn't break other blocks.

mikejolley:
Fixes an error in `append_script_and_deps_src` due to using object properties that may not be available. `get_script_from_handle` may not return a WP dependency object, it can return an empty string. I changed this to `null` and fixed the doc blocks.

While in the file I also fixed a few other docblocks/return types and added some additional output escaping I noticed was missing.

Fixes #5073

### Testing

1. Add Mini Cart block to page and save
2. Install a gateway like Stripe which has block integration
3. Edit the page with mini cart block and add navigation block
4. Before this PR, the request would time out and a fatal error would be logged.

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -100,7 +100,10 @@ class MiniCart extends AbstractBlock {
 
 		foreach ( $payment_methods as $payment_method ) {
 			$payment_method_script = $this->get_script_from_handle( $payment_method );
-			$this->append_script_and_deps_src( $payment_method_script );
+
+			if ( ! is_null( $payment_method_script ) ) {
+				$this->append_script_and_deps_src( $payment_method_script );
+			}
 		}
 
 		$this->scripts_to_lazy_load['wc-block-mini-cart-component-frontend'] = array(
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
LGTM

> While in the file I also fixed a few other docblocks/return types and added some additional output escaping I noticed was missing.

Thanks for these fixes @mikejolley!

nielslange:
Merging this PR as @Aljullu approved it and all tests are passing.

Issue #5000, opened 2021-10-26
Review `__experimentalRegisterCheckoutFilters` and `__experimentalApplyCheckoutFilter` JS filters


nerrad:
These two JavaScript filters are exposed on the `@woocommerce/blocks-checkout` exposed global and allow registering and applying a filter to certain parts of the Cart and Checkout Block. 

While these have been in place for a while (nearly a year) there may still be [some concerns around performance](https://href.li/?https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4570). So the purpose of this issue is to

- Review the actual performance impact of these filters.
- Evaluate if there is work that can be done to either improve the performance of the filters or mitigate the peripheral impact (such as a containing component re-rendering more than it needs to).
- It might be good to see if we can add some sort of executing time limit on callbacks executed by this filter and bail the execution if it's over ms?

Things to do if it's decided we can graduate:

* [ ] Introduce a new action with the experimental prefix removed.
* [ ] Deprecate the old action (so it will still fire, but have it throw a deprecated developer notice pointing to the new name.
* [ ] Update inline code documentation.
* [ ] Update any other documentation.
* [ ] Check this action off in the list

ralucaStan:
I did not test the performance impact of rerunning the filter registered functions but I've reviewed @opr's [PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5143#pullrequestreview-890045327) that prevents the functions from rerunning. It does not go to the root of the problem, aka why are we calling the functions when the parameters don't change, but it does prevent unnecessary reruns.

I say it's a good workaround and fix to this issue.

ralucaStan:
> I did not test the performance impact of rerunning the filter registered functions but I've reviewed @opr's [PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5143#pullrequestreview-890045327) that prevents the functions from rerunning. It does not go to the root of the problem, aka why are we calling the functions when the parameters don't change, but it does prevent unnecessary reruns.
> 
> I say it's a good workaround and fix to this issue.

Considering the work done here I am going to close this issue for now.

PR #5143, opened 2021-11-15
[Experiment] Try to memoize/cache filter results so that we don't call third party filters too often

opr:
This PR was created while investigating #5000. 
Our code runs third party filters every time the cart rerenders. We do memoize this, but unfortunately we often get a cache miss due to the arguments being passed to the filters being objects (`cart`, `cartItem` and any arbitrary data added by extensions to `extensions`) - I think this causes an equality check to fail which means memoization doesn't work perfectly.

This PR ensures the functions registered for a filter only run once if the previous values passed to it are the same.

I took a more 'general' approach to solving this problem as I feel that we should not be prescriptive about what data will be passed to the filter in `arg`. This is because, depending on the filter, different data will be passed. Keeping track of what data we pass to every filter will be difficult to maintain, so a solution that works with any data in `arg` was preferred.

To summarise the changes:
- I added a `checkMembersShallowEqual` function which checks an object against another to see if their members are shallowly equal. It's a deep equality function that stops at depth 1 basically.
- Each time a filter is run, the values it was called with: `filterName`, `arg`, and `extensions` are cached. The next time the same filter is run, a check is made against the current values and previous values is done. If they are the same then the third party filters don't run. This check is done in `shouldReRunFilters`.
- I added `updatePreviousFilterRun` to prevent duplication of code.
- Cached the _result_ of each filter in a ref, this is to keep the variable in scope of the `__experimentalApplyCheckoutFilter` function.

Relates to #5000

#### Other Checks

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

**IN TRUNK**
1. Check out trunk, add this code somewhere that will be executed
```js
let nameCalls = 0;
let priceCalls = 0;
__experimentalRegisterCheckoutFilters( 'my-text-extension', {
		itemName: ( name, arg, extensions ) => {
			nameCalls++;
			console.log( 'itemName filter:', nameCalls );
			return `${ name } + extra data!`;
		},
		cartItemPrice: ( name, arg, extensions ) => {
			priceCalls++;
			console.log( 'cartItemPrice filter:', priceCalls );
			return `${ name } + extra data!`;
		},
	} );
```
2. Add items to cart, and visit the Cart block.
3. Change the quantity of an item, notice that the console logs for both filters get called 22 times!
4. Go to the Checkout block, fill in some details. (Keep note of what you fill in so you can compare properly) - note how many times the filters get executed.

**ON THIS BRANCH**
1. Do the same steps, but notice how many fewer times the filters get executed.

### User Facing Testing
* [ ] Same as above, or
* [x] See steps below.

These instructions use WooCommerce Subscriptions as this is known to use several filters in different places.

1. Install WooCommerce Subscriptions and create a subscription product.
2. Add it to the cart.
3. Go to the Cart block and ensure it shows the details of the subscription product correctly:
image
4. Ensure you can still successfully check out and you are charged the amount you expect to be based on the item price in the database, what is displayed in the Cart block and what is displayed in the Checkout block.

### Performance Impact

This will improve performance when a store has an extension that uses filters active.

### Changelog

>Improve performance of the Checkout and Cart blocks for stores using extensions make use of checkout filters.

### DIFF
--- a/packages/checkout/filter-registry/index.ts
+++ b/packages/checkout/filter-registry/index.ts
@@ -6,7 +6,7 @@ import { __, sprintf } from '@wordpress/i18n';
 import { CURRENT_USER_IS_ADMIN } from '@woocommerce/settings';
 import deprecated from '@wordpress/deprecated';
 import isShallowEqual, { ComparableObject } from '@wordpress/is-shallow-equal';
-import { isObject, objectHasProp } from '@woocommerce/types';
+import { isNull, isObject, objectHasProp } from '@woocommerce/types';
 
 /**
  * A function that always return true.
@@ -90,7 +90,10 @@ const getCheckoutFilters = ( filterName: string ): CheckoutFilterFunction[] => {
 
 const cachedFilterRuns: Record<
 	string,
-	Record< 'arg' | 'extensions', unknown >
+	{
+		arg?: CheckoutFilterArguments;
+		extensions?: Record< string, unknown > | null;
+	} & Record< string, unknown >
 > = {};
 
 const updatePreviousFilterRun = (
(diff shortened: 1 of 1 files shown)


packages/checkout/filter-registry/index.ts
```diff
@@ -107,9 +199,16 @@ export const __experimentalApplyCheckoutFilter = < T >( {
 	/** Function that needs to return true when the filtered value is passed in order for the filter to be applied. */
 	validation?: ( value: T ) => true | Error;
 } ): T => {
+	const cachedValues = useRef< Record< string, T > >( {} );
+
 	return useMemo( () => {
+		if (
+			! shouldReRunFilters( filterName, arg, extensions ) &&
+			cachedValues.current[ filterName ]
```
ralucaStan: tricky one, but falsy values for `cachedValues.current[ filterName ]` would actually make the filter run again even if `! shouldReRunFilters( filterName, arg, extensions )`  evaluates to true.

ralucaStan:
```suggestion
			cachedValues.current[ filterName ] !== undefined
```

ralucaStan (changes requested):
This PR does indeed help with preventing the filtered registered functions too often. 
It does not fix the issue of calling them when there are no changes to the parameters, but it does keep track of whether the functions should be rerun or not by manually checking the parameter values. 

@opr with the code changes that I've requested this should be good to go.

packages/checkout/filter-registry/index.ts
opr: Ooh, good find! Nice one Raluca 😁

ralucaStan (approved):
Good to merge from my side, nice work.

packages/checkout/filter-registry/index.ts
```diff
@@ -107,9 +199,16 @@ export const __experimentalApplyCheckoutFilter = < T >( {
 	/** Function that needs to return true when the filtered value is passed in order for the filter to be applied. */
 	validation?: ( value: T ) => true | Error;
 } ): T => {
+	const cachedValues = useRef< Record< string, T > >( {} );
```
senadir, line 202: Any reason why `cachedValues` is a ref and not a global variable?

opr: No specific reason, feel free to change it to a global variable.

PR #5149, opened 2021-11-16
Re-add experimental discounts meta slot to checkout

opr:
It seems the  `<ExperimentalDiscountsMeta.Slot { ...discountsSlotFillProps } />` was removed from the Checkout Order Summary.

This breaks the Points and Rewards integration.

cc @layoutd thanks for reporting this :)

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Install WooCommerce Points and Rewards, go to `WooCommerce > Points and Rewards` and add points to your user.
2. Go to `WooCommerce > Points and Rewards > Settings` and enable "Allow partial redemption"
4. Go to Checkout and see that the same input shows in the order summary sidebar.

### User Facing Testing
* [x] Same as above

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-order-summary-block/block.tsx
+++ b/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-order-summary-block/block.tsx
@@ -14,6 +14,7 @@ import {
 	TotalsTaxes,
 	ExperimentalOrderMeta,
 	TotalsWrapper,
+	ExperimentalDiscountsMeta,
 } from '@woocommerce/blocks-checkout';
 
 import { getCurrencyFromPriceResponse } from '@woocommerce/price-format';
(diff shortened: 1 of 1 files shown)


ralucaStan (approved):
This testes out as decribed

ralucaStan:
I noticed a small visual bug in the Cart block and I opened [another issue](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5152) for it.

opr:
Thanks Raluca, maybe we should move that to the Points and Rewards repo since this isn't an issue with Blocks but with P&R's css :)

ralucaStan:
> Thanks Raluca, maybe we should move that to the Points and Rewards repo since this isn't an issue with Blocks but with P&R's css :)

@opr I am not sure it should be moved because I went back to an older commit in our repo and the form looked well.

opr:
> > Thanks Raluca, maybe we should move that to the Points and Rewards repo since this isn't an issue with Blocks but with P&R's css :)
> 
> @opr I am not sure it should be moved because I went back to an older commit in our repo and the form looked well.

Hm OK! Let's look into it then, thanks Raluca :)

PR #5151, opened 2021-11-16
Replace do_action_deprecated with the WC equivalent

opr:
This PR will stop deprecated warnings being shown in API requests, previously having the warning print would cause the API request to be malformed and unreadable by the client.

Replacing `do_action_deprecated` with `wc_do_deprecated_action` the error still gets printed to the debug log, just not in the response body.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Install WooCommerce Subscriptions and add a subscription product to your cart.
2. Ensure `WP_DEBUG` and `WP_DEBUG_LOG` are true in `wp-config.php`, **or** install https://wordpress.org/plugins/wp-debugging/
2. Test cart and checkout functionality.
3. Check out and then check the debug log (if using wp-debugging plugin then there's a log viewer in the admin bar)
4. Ensure a deprecation warning for `__experimental_woocommerce_blocks_checkout_update_order_meta` and `__experimental_woocommerce_blocks_checkout_order_processed` is shown.

### User Facing Testing
* [x] Same as above

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -248,7 +248,7 @@ class Checkout extends AbstractCartRoute {
 		 * @param \WC_Order $order Order object.
 		 * @deprecated 6.3.0 Use woocommerce_blocks_checkout_order_processed instead.
 		 */
-		do_action_deprecated(
+		wc_do_deprecated_action(
 			'__experimental_woocommerce_blocks_checkout_order_processed',
 			array(
 				$this->order,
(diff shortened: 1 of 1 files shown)


senadir (approved):

mikejolley (approved):
👍🏻 Just what I had in mind. Tested with and without subs active and confirmed the logs appear. Confirmed also that `is_rest_request` in woo core works for StoreAPI too. 🚢

nielslange:
Merging this PR as @senadir and @mikejolley approved it and all tests are passing.

PR #5153, opened 2021-11-16
Add `totalDescription` filter to cart/checkout totals component

opr:
This PR will add a new filter, 'totalDescription' to the `TotalsFooterItem` component - this will allow extensions to alter the value of the `description` prop passed to the `TotalsItem` which renders the price.

Two questions came up when creating this PR:
1. Do we want to enable a filter for every `TotalsItem` that gets rendered, e.g. the ones in Shipping, Fees, and Tax, or only the "grand total" one?
2. Is the name `totalDescription` ok or is there a better one?

**Todo: Add docs for this filter if the approach here is approved**
#### Other Checks

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Add this code somewhere that'll be executed
```js
__experimentalRegisterCheckoutFilters( 'my-test-extension', {
		totalDescription: ( value ) => {
			return `${ value } Charged October 25th`;
		},
	} );
```
2. Go to the Cart and Checkout blocks, verify the `Charged October 25th` text appears under the grand total
3. Enable taxes and set it so they appear under the total too, (WooCommerce > Settings > Tax > `Yes, I will enter prices inclusive of tax` and `Display prices during basket and checkout: including tax`.
4. Ensure the `Charged October 25th` appears below the tax total.
5. Change the filter code above to
```js
__experimentalRegisterCheckoutFilters( 'my-test-extension', {
		totalDescription: ( value ) => {
			return `Charged October 25th${ value }`;
		},
	} );
```
6. Ensure the tax comes _after_ the Charged October 25th text now.

### User Facing Testing
* [ ] Same as above, or
* [x] See steps below.

1. Enable taxes and set it so they appear under the total too, (WooCommerce > Settings > Tax > `Yes, I will enter prices inclusive of tax` and `Display prices during basket and checkout: including tax`.
2. Add items to your cart and go through the Cart -> Checkout flow.
3. Ensure the totals and the tax description display correctly in Cart and Checkout.

### Changelog

> Add a filter to allow modification of the description label of the cart total in the Cart and Checkout blocks.

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/footer-item/index.js
+++ b/assets/js/base/components/cart-checkout/totals/footer-item/index.js
@@ -8,6 +8,7 @@ import PropTypes from 'prop-types';
 import {
 	__experimentalApplyCheckoutFilter,
 	TotalsItem,
+	mustContain,
 } from '@woocommerce/blocks-checkout';
 import { useStoreCart } from '@woocommerce/base-context/hooks';
 import { getSetting } from '@woocommerce/settings';
@@ -37,34 +38,46 @@ const TotalsFooterItem = ( { currency, values } ) => {
 
 	const parsedTaxValue = parseInt( totalTax, 10 );
 
+	const TaxLabel = ( SHOW_TAXES && parsedTaxValue !== 0 && (
+		<p className="wc-block-components-totals-footer-item-tax">
+			{ createInterpolateElement(
+				__(
+					'Including <TaxAmount/> in taxes',
+					'woo-gutenberg-products-block'
+				),
+				{
+					TaxAmount: (
+						<FormattedMonetaryAmount
+							className="wc-block-components-totals-footer-item-tax-value"
+							currency={ currency }
+							value={ parsedTaxValue }
+						/>
+					),
+				}
+			) }
+		</p>
+	) ) || <></>;
+
+	const filteredTotalDescription = createInterpolateElement(
+		__experimentalApplyCheckoutFilter( {
+			filterName: 'totalDescription',
+			defaultValue: __( '<TaxLabel/>', 'woo-gutenberg-products-block' ),
+			extensions: cart.extensions,
+			arg: { cart },
+			validation: ( value ) => mustContain( value, '<TaxLabel/>' ),
+		} ),
+		{
+			TaxLabel,
+		}
+	);
+
 	return (
 		<TotalsItem
 			className="wc-block-components-totals-footer-item"
 			currency={ currency }
 			label={ label }
 			value={ parseInt( totalPrice, 10 ) }
-			description={
-				SHOW_TAXES &&
-				parsedTaxValue !== 0 && (
-					<p className="wc-block-components-totals-footer-item-tax">
-						{ createInterpolateElement(
-							__(
-								'Including <TaxAmount/> in taxes',
-								'woo-gutenberg-products-block'
-							),
-							{
-								TaxAmount: (
-									<FormattedMonetaryAmount
-										className="wc-block-components-totals-footer-item-tax-value"
-										currency={ currency }
-										value={ parsedTaxValue }
-									/>
-								),
-							}
-						) }
-					</p>
-				)
-			}
+			description={ filteredTotalDescription }
 		/>
 	);
 };
--- a/packages/checkout/utils/validation/index.ts
+++ b/packages/checkout/utils/validation/index.ts
@@ -18,8 +18,8 @@ export const mustContain = (
 					'Returned value must include %1$s, you passed "%2$s"',
 					'woo-gutenberg-products-block'
 				),
-				value,
-				requiredValue
+				requiredValue,
+				value
 			)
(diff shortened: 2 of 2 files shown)


assets/js/base/components/cart-checkout/totals/footer-item/index.js
```diff
@@ -37,34 +38,46 @@ const TotalsFooterItem = ( { currency, values } ) => {
...
+		}
+	);
+
 	return (
 		<TotalsItem
 			className="wc-block-components-totals-footer-item"
 			currency={ currency }
 			label={ label }
 			value={ parseInt( totalPrice, 10 ) }
-			description={
```
ralucaStan, line 46: It looks like the description field was used only for displaying the taxes. What comest to mind is to separate the two sections:
- rename the old description container (`.wc-block-components-totals-item__description`) to a taxes container (`.wc-block-components-totals-item__taxes`)
- add a separe section where the content comes from the filter you added. It could as well keep the name `.wc-block-components-totals-item__description`.

I think we should double check with design how the two sections should look (when both or only one is present)

ralucaStan: I've added some text in the DOM where the description wrapper currently is and there is also a visual issue for mobile, so we should definatelly bring in some designer input to this.
Screenshot 2021-11-16 at 14 30 19

assets/js/base/components/cart-checkout/totals/footer-item/index.js
```diff
@@ -37,34 +38,46 @@ const TotalsFooterItem = ( { currency, values } ) => {
...
+			filterName: 'totalDescription',
+			defaultValue: __( '<TaxLabel/>', 'woo-gutenberg-products-block' ),
+			extensions: cart.extensions,
+			arg: { cart },
+			validation: ( value ) => mustContain( value, '<TaxLabel/>' ),
+		} ),
+		{
+			TaxLabel,
+		}
+	);
```
nerrad, line 72: This is an interesting way of doing this. The only concern I have here is that what happens if in future designs we want to have another component represented in this area that "must" be contained in the returned string from the filter? There's a tight coupling to the current requirements of the `totalDescription` having a `<TaxLabel />` string in it for interpolation.

What about having a completely separate filtered string (default to empty string) that is appended after the description and then passed along? The `description` prop would then be composed by the filtered string and `<TaxLabel />` component:

```jsx
const totalsDescription = (
	<>
		<TaxLabel />
		{ filteredDescription }
	</>
);
```

nerrad: Of course, this might suggest a slot/fill is the better option here too. I'm on the fence.

assets/js/base/components/cart-checkout/totals/footer-item/index.js
nerrad: I missed Raluca's reply here while I was typing up my own comment. Her proposal sounds pretty good too. I like the idea of _explicitly_ calling out an area reserved for taxes and leaving any other descriptions to a separate area.

assets/js/base/components/cart-checkout/totals/footer-item/index.js
ralucaStan: As mentioned in the [comment above](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5153#:~:text=ralucaStan%20reviewed-,13%20minutes%20ago,-View%20changes), I think it's better to separate the taxes all together, as they will be forever present and will less likely change with time.

I was also thinking about wether a Slot / Fill would be a better implementation, considering that with a filter, the newly added 'description' zone could just contain a string, and a Slot would offer more flexibility (HTML content or so).

senadir:
What's this PR solving? do we have a request for such filter?

ralucaStan:
> What's this PR solving? do we have a request for such filter?

@senadir Have a look at pddDDg-77-p2#comment-240, it's supposed to help with the issue discovered in the Pre-orders extension.

assets/js/base/components/cart-checkout/totals/footer-item/index.js
opr: If we only want to allow modification to the "main" `TotalsItem` label after the description and be strict about the position, then maybe Slot/Fill will be better. The filter in its present state allows for the extension to choose whether to output before or after the label.

pluginslab: Extensions that offer the ability to pay in instalments will benefit from this being a slot fill for the HTML content a potential linking to conditions, terms or other content.

Also, we would like to have control over CSS separately, as this extensions would be wanting to highlight an important content in it separately from the tax.

senadir:
Hey Thomas! Is this still valid, if not, can we close it?

opr:
Hey @senadir I think given what's being discussed in #5189 then we can close it

PR #5154, opened 2021-11-16
In `@woocommerce/blocks-checkout` package documentation - reference WCDEWP instead of WPDEWP

nerrad:
The current documentation for the checkout package references the WordPress Dependency Extraction Webpack plugin. However, there's less configuration needed if folks use the WooCommerce Dependency Extraction Webpack plugin instead, so this PR modifies the documentation accordingly.

There is no impact to released code in this PR.

Reviews mostly just need to ensure the syntax of the example provided is correct and there are no other grammar/spelling errors in the documentation.

### DIFF
--- a/packages/checkout/README.md
+++ b/packages/checkout/README.md
@@ -6,7 +6,7 @@ Components and utilities making it possible to integrate with the WooCommerce Ca
 
 - [Installation](#installation)
 - [Usage](#usage)
-	- [Aliased imports](#aliased-imports)
+  - [Aliased imports](#aliased-imports)
 - [Folder Structure Overview](#folder-structure-overview)
 
 ## Installation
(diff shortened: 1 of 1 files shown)


packages/checkout/README.md
```diff
@@ -23,36 +23,22 @@ const { ... } = wc.blocksCheckout;
...
-	'@woocommerce/blocks-checkout': [ 'wc', 'blocksCheckout' ],
-};
+Now, you can include this plugin in your Webpack configuration:
 
-const handleMap = {
-	'@woocommerce/blocks-checkout': 'wc-blocks-checkout',
-};
+```js
+// webpack.config.js
+const WooCommerceDependencyExtractionWebpackPlugin = require( '@wordpress/dependency-extraction-webpack-plugin' );
```
opr:
```suggestion
const WooCommerceDependencyExtractionWebpackPlugin = require( '@woocommerce/dependency-extraction-webpack-plugin' );
```

packages/checkout/README.md
```diff
@@ -23,36 +23,22 @@ const { ... } = wc.blocksCheckout;
...
-					return dependencyMap[ request ];
-				}
-			},
-			requestToHandle( request ) {
-				if ( handleMap[ request ] ) {
-					return handleMap[ request ];
-				}
-			},
-		} ),
+		new WooCommerceDependencyExtractionWebpackPlugin,
```
opr:
```suggestion
		new WooCommerceDependencyExtractionWebpackPlugin(),
```

opr (changes requested):
Tested this out on my example plugin and had to make some changes to get it to work, I've added them as suggestions

packages/checkout/README.md
nerrad: doh 🤦🏻

PR #5156, opened 2021-11-16
Fix Featured Product Search not working for large stores

### LINKED ISSUE #4983: Featured Product Block Search is Broken
## Describe the bug
On a store with a large number (+100) products, search appears to be broken when using the Featured Product Block. On a smaller catalog you can simulate this by editing the values in:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/5f34c86072936ac908a93c49881775d21df69a6a/assets/js/editor-components/utils/index.js#L19

## To reproduce

1. Have a large catalog, or edit the function above so it does requests to the API
2. Insert Featured Product Block
3. Enter a search term for a product not visible in the list
4. No results. Query has `search=&` in it.

## Expected behaviour
It should do the query and return the filtered results.

This was reported here: https://wordpress.org/support/topic/feature-product-not-wordking-correcly/ I reproduced it locally.

alexflorisca:
On the editor, after inserting a featured product block, the search does not return any results for large stores (with more than 100 products currently). This is due to the behaviour of the [`WithSearchedProducts`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/hocs/with-searched-products.tsx) higher order component. This generates a request to fetch products, both on a user searching, and every time the user changes a selection.  

What's happening is [this useEffect](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/hocs/with-searched-products.tsx#L44) is ran on (almost) every render, because `selected` is an array that is checked by reference and it will have a different array instance each time the component is invoked, even if the value of the array hasn't changed.

The solution proposed here simply removes `selected` from the dependency array so that `getProducts` will only run once on the first render. From what I can see, there is no need to `getProducts` after each selection, and there is a [separate call to `getProducts` in the `onSearch` callback](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/hocs/with-searched-products.tsx#L53). This looks to me like the intended behaviour but please point out if I'm missing something.

I've also explored memoizing the `selected` array with `useMemo()` but this still doesn't work, presumably because it's a different array each time.

Yet another solution would be to use `JSON.stringify(selected)` to check for the change in value of the `selected` array as described [here](https://github.com/facebook/react/issues/14476#issuecomment-471199055). We could do this if this `useEffect` needs to run on every product selection.

Also eslint is throwing a warning about `React Hook useEffect has a missing dependency: 'selected'. Either include it or remove the dependency array`. But for this situation I think this is the right setting so I would disable the warning for this line but wanted to check first. Open to suggestions on this.

Fixes #4983

### Screenshots

Before: 
image

After: 
image

### Manual Testing

1. Have a large catalog, or edit [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/editor-components/utils/index.js#L24) and [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/editor-components/utils/index.js#L24) and set 100 to 5 say
2. Insert Featured Product Block in the editor
3. Enter a search term for a product not visible in the list
4. Should see relevant search results based on search

### Changelog

> Fixed Featured Product Block search not working for large stores

### DIFF
--- a/assets/js/hocs/with-searched-products.tsx
+++ b/assets/js/hocs/with-searched-products.tsx
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { useEffect, useState, useCallback } from '@wordpress/element';
+import { useEffect, useState, useCallback, useRef } from '@wordpress/element';
 import { blocksConfig } from '@woocommerce/block-settings';
 import { getProducts } from '@woocommerce/editor-components/utils';
 import { useDebouncedCallback } from 'use-debounce';
@@ -41,14 +41,16 @@ const withSearchedProducts = (
 			setIsLoading( false );
 		};
 
+		const selectedRef = useRef( selected );
+
 		useEffect( () => {
-			getProducts( { selected } )
+			getProducts( { selected: selectedRef.current } )
 				.then( ( results ) => {
 					setProductsList( results as ProductResponseItem[] );
 					setIsLoading( false );
 				} )
 				.catch( setErrorState );
-		}, [] );
+		}, [ selectedRef ] );
 
 		const [ debouncedSearch ] = useDebouncedCallback(
 			( search: string ) => {
--- a/assets/js/hocs/with-searched-products.tsx
+++ b/assets/js/hocs/with-searched-products.tsx
@@ -48,7 +48,7 @@ const withSearchedProducts = (
 					setIsLoading( false );
 				} )
 				.catch( setErrorState );
-		}, [ selected ] );
+		}, [] );
 
 		const [ debouncedSearch ] = useDebouncedCallback(
 			( search: string ) => {


assets/js/hocs/with-searched-products.tsx
```diff
@@ -48,7 +48,7 @@ const withSearchedProducts = (
 					setIsLoading( false );
 				} )
 				.catch( setErrorState );
-		}, [ selected ] );
+		}, [] );
```
senadir: This is an anti pattern that we fought a bit to get rid off and refactor away from, because it causes hard to debug bugs.
If you feel like `selected` is not going to change much here, I'd suggest you have it a ref then?

```js
const selectedRef = useRef( selected );

selectedRef.current;
```

This should achieve the same goal, and would not cause rerenders.

alexflorisca: Thanks for pointing that out @senadir, I'd be curious to know more about the types of bugs this causes! I've changed the implementation to use a ref as you suggested.

senadir: Hey! This causes out of sync bugs, things running when not needed, and generally hard to find bugs, examples:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3314
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/3285

alexflorisca:
Thanks, that’s interesting, thanks for sharing. I’m going to read up more about it. I was under the impression that if the argument list is empty, it will actually avoid re-renders (only once) so it’s a good way to say “I only want this to run once”.

opr (commented):
Is it expected to get two (debounced) requests every time you type into the search box? I get the following two:

- `/wp-json/wc/store/products?per_page=5&catalog_visibility=any&search=beanie&orderby=title&order=asc&_locale=user`
- `/wp-json/wc/store/products?catalog_visibility=any&include%5B0%5D=0&per_page=0&_locale=user `

alexflorisca:
> Is it expected to get two (debounced) requests every time you type into the search box? I get the following two:
> 
> * `/wp-json/wc/store/products?per_page=5&catalog_visibility=any&search=beanie&orderby=title&order=asc&_locale=user`
> * `/wp-json/wc/store/products?catalog_visibility=any&include%5B0%5D=0&per_page=0&_locale=user `

It looks like the `getProductsRequest` method ads another request [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/editor-components/utils/index.js#L37) - it makes another request to get the currently selected item, because searching again on a large store will change the list of results based on search and may remove the currently selected item from the dom. So it looks like it's meant to be there :)

senadir:
>  I was under the impression that if the argument list is empty, it will actually avoid re-renders (only once) so it’s a good way to say “I only want this to run once”.

Yes, that's still the case, however, we want to avoid diverging from that eslint rule, not adding things to the deps rule to achieve that indicates a code smell and an issue to fix higher up, or with an alternate solution.

opr (approved):
> it looks like the getProductsRequest method ads another request here - it makes another request to get the currently selected item, because searching again on a large store will change the list of results based on search and may remove the currently selected item from the dom. So it looks like it's meant to be there :)

OK thanks! If this second request is necessary then I'm happy to approve. 🚢 👍🏼

PR #5157, opened 2021-11-16
Introduce `<Noninteractive>` component to disable form elements non-visually

### LINKED ISSUE #2487: Explore ways so form elements in blocks don't look disabled in the editor

mikejolley:
In order to prevent form elements looking disabled in the editor (due to usage of the `<Disabled>` component, this PR introduces a new `<Noninteractive>` component. What this does is as follows:

- Adds styling to turn off user and pointer events
- Prevents the cursor changing from default
- Implements a negative tabindex to prevent keyboard navigation to disabled form elements

I updated the `use-debounce` library which is used in `<Noninteractive>`.

In #2487, there was a suggestion to use the HTML attribute `inert` and a polyfill, but since there is currently zero browser support for this, and the polyfill is fairly old, I decided to instead handle it in the component, following the pattern of `<Disabled>` in Gutenberg.

Fixes #2487

### Testing

Test using the Checkout Block, or any other block which uses `<Noninteractive>` now in place of `<Disabled>`.

1. Try to click on the address fields. You cannot.
2. The cursor should remain default whilst hovering address fields.
3. Try to click the place order button. It should not respond.
4. Click on the form step description in the contact block, then try to tab to the next field. It should skip over the form field.
5. Check console for errors.

### Changelog

> Fix display of form elements in the editor so they do not look disabled

### DIFF
--- a/assets/js/base/components/noninteractive/index.tsx
+++ b/assets/js/base/components/noninteractive/index.tsx
@@ -30,8 +30,11 @@ const INPUT_FIELD_NODE_NAMES = [
  */
 const Noninteractive = ( {
 	children,
+	style = {},
+	...props
 }: {
 	children: React.ReactChildren;
+	style?: Record< string, string >;
 } ): JSX.Element => {
 	const node = useRef( null );
 
@@ -85,7 +88,9 @@ const Noninteractive = ( {
 			style={ {
 				userSelect: 'none',
 				pointerEvents: 'none',
+				...style,
 			} }
+			{ ...props }
 		>
 			{ children }
 		</div>
--- a/assets/js/base/components/noninteractive/index.tsx
+++ b/assets/js/base/components/noninteractive/index.tsx
@@ -1,14 +1,14 @@
 /**
  * External dependencies
  */
-import { useRef, useLayoutEffect, useCallback } from '@wordpress/element';
-import { debounce } from 'lodash';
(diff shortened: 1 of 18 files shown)


nielslange (approved):
LGTM! 🚢 

Just one additional thought. I noticed that we're still using `<Disabled>` on various other spots. It might be good to replace all occurrences of `<Disabled>` with `<Noninteractive>` for maintainability reasons. What do you think about that?

mikejolley:
I think we should do it case by case @nielslange. If an input really should be disabled and knowingly disabled from use, the `Disabled` component I think is still more semantic. This should be reserved for places where we want something to appear non-disabled visually.

Do you think we should add `aria-disabled="true"` to this component as well?

nielslange:
> Do you think we should add aria-disabled="true" to this component as well?

I think that's a great idea, @mikejolley.

mikejolley:
@nielslange fixed that up now

nielslange:
> @nielslange fixed that up now

Looking great! Let's 🚢 it. 🙌

PR #5158, opened 2021-11-16
Remove feature gate condition for Legacy Template block

Aljullu:
Discovered while testing 6.3.0: the Legacy Template block was not available.

### Manual Testing

1. Run `npm run package-plugin`.
2. Install the generated ZIP into a site with Gutenberg 11.8.3 and a block theme (like [Quadrat](https://wordpress.org/themes/quadrat/)).
2. Go to Site Editor > General templates > Single Product Page and verify the block renders correctly.

### DIFF
--- a/assets/js/blocks/legacy-template/index.tsx
+++ b/assets/js/blocks/legacy-template/index.tsx
@@ -1,10 +1,8 @@
 /**
  * External dependencies
  */
-import {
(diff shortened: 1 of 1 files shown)


frontdevde (approved):
Can confirm this change fixes the issue we were seeing, LGTM 👍

Screenshot 2021-11-16 at 17 44 43

PR #5164, opened 2021-11-17
Checkbox Component Styling Fix

mikejolley:
I noticed 2 small issues with the checkbox component.

1. It was possible to accidentally click the label due to extra padding when it's the last element in a form step
2. Clicking the input lagged due to the SVG accepting pointer events

Fixed with CSS and a wrapping div.

### Testing

1. Checkout https://github.com/mailpoet/mailpoet/pull/3778 and enable mailpoet
2. See checkbox added to contact step in checkout
3. Go to frontend. Check the spacing between the label and the checkbox, and then click below the label in the gap between form steps. The checkbox should not be triggered
4. Confirm you can still toggle the checkbox input

### Changelog

> Reduce the size of the checkbox component label to prevent accidental input

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/editor.scss
+++ b/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/editor.scss
@@ -12,10 +12,6 @@
 	}
 }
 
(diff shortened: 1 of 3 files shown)


opr (approved):
👍🏼 Nice, works well code looks fine.

PR #5165, opened 2021-11-17
Product details component accepts `key` or `name` as props

### LINKED ISSUE #4400: Respect `key` attributes in Cart Items meta over `name`
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/196a68a781b5ade93d6d42140f2a2165ed6cfae5/assets/js/base/components/cart-checkout/product-details/index.tsx#L46

Here it can see using "name" filed, whereas, in default cart shortcode (PHP), it is given preference for "key", instead of name

https://github.com/woocommerce/woocommerce/blob/d1f5784fa479078a375438adc961726158f7fafb/includes/wc-template-functions.php#L3669

alexflorisca:
The PHP shortcode version of the product details metadata uses `key` instead of `name`. This PR allows the use of both, prioritising `key` as a default in order to keep consistent with the PHP shortcode and WooCommerce core.

Fixes #4400

## Screenshots
Screenshot 2021-11-17 at 11 48 06

### Manual Testing

1. Add a product with a variant to your basket
2. Go to Cart page
3. Make sure the variant name is displayed
3. Repeat for Checkout page

### Changelog

> Respect `key` attribute in product-details component over `name`

### DIFF
--- a/assets/js/base/components/cart-checkout/product-details/index.tsx
+++ b/assets/js/base/components/cart-checkout/product-details/index.tsx
@@ -30,20 +30,22 @@ const ProductDetails = ( {
 	return (
 		<ul className="wc-block-components-product-details">
 			{ details.map( ( detail ) => {
-				const className = detail.name
+				// Support both `key` and `name` props
+				detail.key = detail.key ? detail.key : detail.name || '';
+				const className = detail.key
 					? `wc-block-components-product-details__${ kebabCase(
-							detail.name
+							detail.key
 					  ) }`
 					: '';
 				return (
 					<li
-						key={ detail.name + ( detail.display || detail.value ) }
+						key={ detail.key + ( detail.display || detail.value ) }
 						className={ className }
 					>
-						{ detail.name && (
+						{ detail.key && (
 							<>
 								<span className="wc-block-components-product-details__name">
-									{ decodeEntities( detail.name ) }:
+									{ decodeEntities( detail.key ) }:
 								</span>{ ' ' }
 							</>
 						) }
--- a/assets/js/base/components/cart-checkout/product-details/index.tsx
+++ b/assets/js/base/components/cart-checkout/product-details/index.tsx
@@ -31,21 +31,21 @@ const ProductDetails = ( {
 		<ul className="wc-block-components-product-details">
 			{ details.map( ( detail ) => {
 				// Support both `key` and `name` props
-				detail.key = detail.key ? detail.key : detail.name || '';
-				const className = detail.key
+				const name = detail?.key || detail.name || '';
+				const className = name
 					? `wc-block-components-product-details__${ kebabCase(
-							detail.key
+							name
 					  ) }`
 					: '';
 				return (
 					<li
-						key={ detail.key + ( detail.display || detail.value ) }
+						key={ name + ( detail.display || detail.value ) }
 						className={ className }
 					>
-						{ detail.key && (
+						{ name && (
 							<>
 								<span className="wc-block-components-product-details__name">
-									{ decodeEntities( detail.key ) }:
+									{ decodeEntities( name ) }:
 								</span>{ ' ' }
 							</>
 						) }
--- a/assets/js/types/type-defs/product-response.ts
+++ b/assets/js/types/type-defs/product-response.ts
@@ -16,13 +16,14 @@ export interface ProductResponseItemPrices extends CurrencyResponse {
 	};
 }
 
-export interface ProductResponseItemData {
-	name?: string;
-	key?: string;
+export interface ProductResponseItemBaseData {
 	value: string;
 	display?: string;
 	hidden?: boolean;
 }
+
+export type ProductResponseItemData = ProductResponseItemBaseData &
+	( { key: string; name?: never } | { key?: never; name: string } );
 
 export interface ProductResponseImageItem {
 	id: number;
--- a/assets/js/types/type-defs/product-response.ts
+++ b/assets/js/types/type-defs/product-response.ts
@@ -17,7 +17,8 @@ export interface ProductResponseItemPrices extends CurrencyResponse {
 }
 
 export interface ProductResponseItemData {
-	name: string;
+	name?: string;
+	key?: string;
 	value: string;
 	display?: string;
 	hidden?: boolean;
--- a/assets/js/base/components/cart-checkout/product-details/test/__snapshots__/index.js.snap
+++ b/assets/js/base/components/cart-checkout/product-details/test/__snapshots__/index.js.snap
@@ -66,6 +66,7 @@ exports[`ProductDetails should render details 1`] = `
   <li
     className=""
   >
+    
     <span
       className="wc-block-components-product-details__value"
     >
--- a/assets/js/base/components/cart-checkout/product-metadata/index.tsx
+++ b/assets/js/base/components/cart-checkout/product-metadata/index.tsx
@@ -34,7 +34,7 @@ const ProductMetadata = ( {
 			<ProductDetails details={ itemData } />
 			<ProductDetails
 				details={ variation.map( ( { attribute = '', value } ) => ( {
-					name: attribute,
+					key: attribute,
 					value,
 				} ) ) }
 			/>


assets/js/base/components/cart-checkout/product-details/index.tsx
```diff
@@ -30,20 +30,22 @@ const ProductDetails = ( {
 	return (
 		<ul className="wc-block-components-product-details">
 			{ details.map( ( detail ) => {
-				const className = detail.name
+				// Support both `key` and `name` props
+				detail.key = detail.key ? detail.key : detail.name || '';
```
sunyatasattva: I would suggest using the nullish coalescing operator here, both for brevity and also to make sure intent is respected. I'ld also suggest not reassigning an object property when possible, to avoid side-effects. Perhaps not so important in this case, though.

So I'd probably go with something like:

```
const key = detail.key ?? detail.name ?? ''
```

assets/js/types/type-defs/product-response.ts
```diff
@@ -17,7 +17,8 @@ export interface ProductResponseItemPrices extends CurrencyResponse {
 }
 
 export interface ProductResponseItemData {
-	name: string;
+	name?: string;
+	key?: string;
```
sunyatasattva: I'm wondering here… 🤔  this changes a bit the intent of the type. Previously, `name` was required. Now you could in theory have an object without either `name` or `key` and respect this interface; or, other way around, both could be defined.

Given the code above, it doesn't seem like a big deal as we treat the case in which both are `undefined` by making it into an empty string (and, also, we pass an empty string by default to the component anyways 😂 ), and we are giving priority to `key`; but I can't help but wonder if the intention of this interface should instead be that one of those two should be required, but not both or neither at the same time.

What are your thoughts?

assets/js/base/components/cart-checkout/product-details/index.tsx
mikejolley: @sunyatasattva Using a nullish coalescing operator would use name if key was null, but what if it's empty? The behaviour in core we're copying uses the name if the key is empty (`null`, `false`, `0`, `''`).

assets/js/base/components/cart-checkout/product-details/index.tsx
```diff
@@ -30,20 +30,22 @@ const ProductDetails = ( {
...
-						key={ detail.name + ( detail.display || detail.value ) }
+						key={ detail.key + ( detail.display || detail.value ) }
 						className={ className }
 					>
-						{ detail.name && (
+						{ detail.key && (
 							<>
 								<span className="wc-block-components-product-details__name">
-									{ decodeEntities( detail.name ) }:
+									{ decodeEntities( detail.key ) }:
```
mikejolley: Just pointing out, if `name` makes more sense in this implementation (looking at the CSS classname), we could still respect `key` e.g. `const name = detail?.key || detail.name || '';`

assets/js/types/type-defs/product-response.ts
mikejolley: I agree. Whatever we're choosing as the correct property should be required. This will highlight issues in the codebase using the wrong property and allow future deprecation of the wrong one.

mikejolley (commented):
Added some feedback, nothing major.

assets/js/base/components/cart-checkout/product-details/index.tsx
sunyatasattva: @mikejolley I see, thanks for the clarification. Indeed you are 100% spot on: the nullish coalescing would work for `null` and `undefined` only, which I thought was the intention here, sorry.

Given the interface below, however, it seems that `detail.key` and `detail.name` could only be `string` or `undefined`, so shouldn't TS save us from `false` and `0`?

assets/js/types/type-defs/product-response.ts
sunyatasattva: Or, if both are supposed to be valid, but only one at the time, that should be also reflected in the type, IMHO.

assets/js/base/components/cart-checkout/product-details/index.tsx
mikejolley: Yeah possibly, if respected. It won't save us from `''` or '`0`' or anything else which is considered empty.

assets/js/types/type-defs/product-response.ts
alexflorisca: Thanks for the comments. I agree with you and I did initially investigate what the solution would be to make one of `key` or `name` required, but in my opinion it seems a bit convoluted, adding more complexity than necessary - see solution here:

https://stackoverflow.com/questions/40510611/typescript-interface-require-one-of-two-properties-to-exist

I'm fairly new to typescript so I can't think of a better way to do this but I'm open to suggestions. As an alternative, I think what Mike suggested makes sense of requiring `key` prop and having `name` as optional and marked as deprecated? What do you think @mikejolley @sunyatasattva ?

```
	key: string;
	name?: string; // deprecated

File: assets/js/base/components/cart-checkout/product-details/index.tsx (hunk 3 above)
Comment (alexflorisca) [reply assets/js/base/components/cart-checkout/product-details/index.tsx]: @sunyatasattva @mikejolley So based on the above discussion, I've updated the PR with @mikejolley's suggestion here, which removes overwriting the object key, renames the const to `name` for better readability in the code here, and respects the intent of using the name if the key is empty.

File: assets/js/types/type-defs/product-response.ts (hunk 2 above)
Comment (sunyatasattva) [reply assets/js/types/type-defs/product-response.ts]: Hey @alexflorisca, I agree TypeScript might feel convoluted at times, but in my experience it often pays off. I have experimented a bit and came up with a generic type that's hopefully a bit less convoluted. However, if this is still too much, I provide a non-generic solution below that will solve your problem while being fairly simple to read:

```
type EitherProperty<
    T, A extends keyof T, B extends keyof T
> = (Omit<T, B> & { [K in A]-?: T[A]; } & { [K in B]?: never} )
| (Omit<T, A> & { [K in B]-?: T[B] } & { [K in A]?: never });
```

See [<kbd>TypeScript Playground</kbd>](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgApQPYBMCuCwBKEAzgA4YjEQCSkAtgEJxUAicYcyA3gFDL-IA1hACeAfgBcyYmCigA5gG4+AkHDoRJ02QuUBIAG5wANjghSZckEp56swMsbjiLO6-oAWwLFgggtAEYYGMYQcCDKAL48PGAipCgAosBgHtDoGAlQcQA8KvwAKgA0yACCyBAAHpAgWMRCohgwyMXIDBXVfnUNIk0tPAB8yAC8yAAUAPJ0KTmtDEMAZNzIANoA0sigZQC6ALRaBSul24rIkchLXKsbWwzbWiAQBtDnAJQ8AD7jUzOtpYvLdabEBtPYHFZ3M4XQE3EHHB5PaBnV7KHgICgyTb0ADqwGMxkSUEwUCkyVS6UwWVyGVw+CIZAxNHoTFY7DgJQARMIRBzOWoNByhqNeAIelIOTBgrz8sh+eZkByAnAoNLDCYzOKjMZVfZHM5xbrSKqvD4-FJ4MYqEUeNE0RiwFiIHRcfiAHIYMCE4lrUSklJpKAZKkiHI0vCEEjkShMp0siBsDic7m8hVywUjbgy7niyUYVVajUKrU6hykJwiA2l43eXwgc0mK02mLoygOlJOl3Gd2eokYKCu9TyskBoPQamYWkRhnR2ix5jxtlJ0QpjlpoWZ0VynNS61q0zyjnF3eG8uVsjV0115AWxvRIA) on how to use this.

Alternatively, if this feels too complex, what about this?

```
interface ProductResponseItemBaseData {
	value: string;
	display?: string;
	hidden?: boolean;
}

type ProductResponseItem = ProductResponseItemBaseData & ({ key: string, name?: never } | { key?: never, name: string });
```

See [<kbd>TypeScript Playground</kbd>](https://www.typescriptlang.org/play?ssl=8&ssc=1&pln=1&pc=1#code/JYOwLgpgTgZghgYwgAgApQPYBMCuCwBKEAzgA4YjEQCSkAtgEJxUAicYcyA3gFACQANzgAbHBABcyYmCigA5gG5+WYGWFwAngH5J02SEX8AFsCxYIIHcgBGGDMIhwQSgL48eYDaRTpseQiTklDT0yAC8aJi4+ERkFFS0EIzMEGwcyABkyAAUXMgA1hAaujLyADTIIHB0EFYgEALQyC7IAD7cBUV1DdAVVTUl+nLNAJRKPAjxYMjA9ADqwMLCAKJQmFCSvtEBccGJdOHcPMgnncXIAEQwdhdlx6f9EpfEt-cnQqJPFx+vp8gqak0kguWF+pxMZgskngwioPDcEymM3mi2EADkMGBVusANJFTZRfyxIIJUIRXh-QrnK43O5-D5iYE-OmnAGkdTU0Esk4Q8wgaEiOEIyaUaazJILJYYrFrDBQNHVJ5bImBeIhJKHCkPRXAl7c5AMr7Mt7-VTsoGXLkm3lQ5AwoU8IA) on how to use this.

Basically, how this works is that you are defining your basic data in your interface and completely excluding the name/key keys, and then you create the full object type by telling TS it is either one shape or the other.

alexflorisca: Thanks for this suggestions @sunyatasattva, this somehow fell off my radar so sorry for taking so long to update. I've updated the PR with your second suggestion of defining a base interface and a type that makes one of `name` or `key` mandatory. I think it reads nicely and is friendly to non typescript experts such as myself :). Appreciate your help on this :)

sunyatasattva (approved):

assets/js/types/type-defs/product-response.ts
sunyatasattva: Nice! I'm glad it was helpful!

PR #5176, opened 2021-11-17
Legacy Template Block: Remove the deletion lock (#5166)

### LINKED ISSUE #5109: It's not possible to move the Legacy Template Block outside of its Group block parent
This is probably a bug in Gutenberg and not in our codebase, but still it's affecting WC Blocks so I'm logging it here: it's not possible to move the Legacy Template Block out of the group block where it's contained.

## To reproduce

1. Edit any of the WC block templates.
2. Try dragging the WooCommerce Legacy Template out of the Group block.
3. Notice it can't be moved.

## Expected behavior
I should be able to move the block outside of its parent, even if it's not possible to remove it.

## Screenshots

https://user-images.githubusercontent.com/3616980/140926063-f2354f0c-4310-49e0-bf0c-3cced4632237.mp4

### LINKED ISSUE #5163: The `WooCommerce Legacy Template` block can be removed from the template if the user deletes the entire group
## Describe the bug
`WooCommerce Legacy Template` block can be removed from the template if the user deletes the entire group.

## To reproduce

1. Install and activate [Gutenberg plugin](https://wordpress.org/plugins/gutenberg/).
2. Install and activate a FSE enabled theme, such as [the TT1 Blocks theme](https://wordpress.org/themes/tt1-blocks/).
3. Load the Site Editor (Appearance → Editor). Open the left editor sidebar navigation by clicking the WordPress logo in the top left. In the Templates > General Templates section select the `Single Product Page`, `Product Tag Page`, `Product Archive Page`, or `Product Category Page` template.
4. Open the `List view` and find `WooCommerce Legacy Template` block.
5. Remove the group that wraps `WooCommerce Legacy Template` block.

## Expected behavior
The user shouldn't be able to remove `WooCommerce Legacy Template` block

## Screenshots

https://user-images.githubusercontent.com/4463174/142165992-24acf233-d20a-4799-a1b9-f17900f8e6ad.mp4

### LINKED ISSUE #5166: Legacy Template block: remove block locking
We originally introduced block locking to the Legacy Template block because we wanted to protect users from accidentally deleting the block from their template. While prototyping a solution for https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5163, we discovered that locking the block severely limits the creative freedom users have in customizing the templates to their heart's content i.e. it can't be moved inside Columns, Group, etc. We thus decided to allow the block to be (re)moved again. 

We're currently hiding the block from the inserter as it's only relevant to the Woo Templates. If folks accidentally remove the block from their template, they can delete their custom template and the default template will reappear with the block present. We will also open a different issue to explore if it's possible to limit the block to only show in the block inserter on our templates.

For more details see the discussion in slack-archives/C02FL3X7KR6/p1637149371397200.

Closes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5163
Closes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5109

sunyatasattva:
Previously, our legacy template block was locked to prevent the users from removing it. The block could however be removed by erasing the parent, so this change allows the user to directly remove the block itself.

The standing issue is that if the user removes the block, they can't add it back through the block inserter. However, they can revert their change by deleting the locally saved template (from Appearance → Templates). 

For the entire context: see #5166
Closes #5166
Closes #5163
Closes #5109

### Screenshots

can-remove

remove-template

### Manual Testing

1. Go to Appearance → Editor
2. Open the sidebar and select General templates → Product Archive Page
3. Select the Legacy Template Block and click “Remove WooCommerce Legacy Template” (screenshot above)
4. Confirm that the block gets deleted without errors in the console

### User Facing Testing

See steps below.

1. Go to Appearance → Editor
2. Open the sidebar and select General templates → Product Archive Page
3. Select the Legacy Template Block and click “Remove WooCommerce Legacy Template” (screenshot above)
4. Confirm that the block gets deleted in the editor

### Changelog

> Legacy Template Block: allow users to delete the block.

### DIFF
--- a/assets/js/blocks/legacy-template/index.tsx
+++ b/assets/js/blocks/legacy-template/index.tsx
@@ -84,12 +84,6 @@ registerBlockType( 'woocommerce/legacy-template', {
 			type: 'string',
 			default: 'any',
 		},
-		lock: {
-			type: 'object',
-			default: {
-				remove: true,
-			},
-		},
 	},
 	edit: Edit,
 	save: () => null,


tjcafferkey (approved):
Works as described and LGTM! Just waiting for tests to pass

PR #5177, opened 2021-11-17
Typedefs and cartData for `usePaymentMethodInterface`

### LINKED ISSUE #5042: Pass extra payment method data from `ExtendRestAPI` response via the payment method interface API.
With the [`ExtendRestAPI`](https://href.li/?https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/extend-rest-api-add-data.md), extensions are able to register additional data to pass along with the response for any Cart and/or Checkout endpoint responses. However, this is not exposed to registered payment method components which means they will have to make extra requests for anything they need to do on the server as a part of their logic (example Stripe, normalizing cart data for use with any Stripe API use).

This issue is for ensuring that any extra payment method data coming through the `ExtendRestAPI` response is passed along to registered payment method components as a part of the payment method interface API. One thing I'd like to aim for as a part of this work:

- If possible, limit the data exposed. In other words, instead of passing along all of the data on the `extensions` property of the response, just return the payment method specific data.

mikejolley:
Adds extra cart data within `usePaymentMethodInterface`. This includes items, fees, and extension data. Other props are already included in `shippingData` and `billing`.

I've typed this interface now within `PaymentMethodInterface` and included the new properties.

Fixes #5042

### Screenshots

Console log of extension data accessed via `usePaymentMethodInterface`:

Screenshot 2021-11-17 at 16 24 33

### Changelog

> Expose extension data via `usePaymentMethodInterface` for Payment Integrations.

### Dev Note

> We've updated `usePaymentMethodInterface` to include improved typescript definitions, so if you were importing our typedefs, you'll need to modify where you pull them from.


mikejolley:
@ralucaStan rebased this with your doc changes.

assets/js/types/type-defs/payment-method-interface.ts
```diff
@@ -0,0 +1,176 @@
...
+	PaymentMethodLabel: typeof PaymentMethodLabel;
+	// A container for holding validation errors
+	ValidationInputError: () => JSX.Element | null;
+}
+
+export interface EmitResponseProps {
+	// Response types that can be returned from emitter observers.
+	responseTypes: responseTypes;
+	// Available contexts that can be returned as the value for the messageContext property on the object  returned from an emitter observer.
+	NoticeContexts: noticeContexts;
```
ralucaStan: this should be `noticeContexts` instead of `NoticeContexts`. Changing this will result in an TS warning for `noticeContexts` in usePaymentMethodInterface.

assets/js/base/context/hooks/payment-methods/use-payment-method-interface.ts
```diff
@@ -124,38 +130,38 @@ export const usePaymentMethodInterface = (): Record< string, unknown > => {
...
-			ValidationInputError,
+			LoadingMask,
 			PaymentMethodIcons,
 			PaymentMethodLabel,
-			LoadingMask,
+			ValidationInputError,
 		},
 		emitResponse: {
 			noticeContexts,
 			responseTypes,
```
ralucaStan, line 140: I see this TS warning
```
type 'typeof responseTypes' is not assignable to type 'responseTypes'.ts(2322)
payment-method-interface.ts(83, 2): The expected type comes from property 'responseTypes' which is declared here on type 'EmitResponseProps'
(property) EmitResponseProps.responseTypes: responseTypes
```
Might have something to do with enums, as they are not types (only enum members can be types) and responseTypes is actually an enum.

the same warning appears for `noticeContexts` after the typo I've mentioned in my other comment is fixed.

ralucaStan (changes requested):
I've added some comments regarding the types for `emitResponse`.

I've also  saw in #5042 that there is this request, `If possible, limit the data exposed. In other words, instead of passing along all of the data on the extensions property of the response, just return the payment method specific data.` Would it be possible to have for extensions a something like a getter that returns a certain key for the extensions object?

mikejolley:
@ralucaStan I queried that here: p1637149910386300-slack-C8X6Q7XQU

> I was mostly wanting to avoid passing in data that potentially frequently updates, but doesn't look like we can avoid that anyways here. As long as we just pass in read-only data from the hook I'm good with that.

A getter would solve immediately exposing properties, but all properties would still be exposed right? We'd just be introducing an extra layer of friction for extensions to retrieve the data. Are there other benefits I'm overlooking?

ralucaStan:
> ralucaStan [I queried that here.](p1637149910386300-slack-C8X6Q7XQU)

the link does not work

ralucaStan:
@mikejolley I was thinking of having something in place that allows payment gateways to access only their namespaced  data from `extensions`, so that the payment method does not have access to other extensions' data. 

getExtensionsData('my-payment-namespace');

This is what I understood was the goal. But from I also see in the discussion this was not the case.

ralucaStan:
@mikejolley Should we mention in the description this PR comes with breaking-changes? I see Stripe gateways uses [ this import ](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/8f181abe9ad8321ebdd09a494ed0e570b2f2c448/client/blocks/payment-request/payment-request-express.js#L21)
`@typedef {import('@woocommerce/type-defs/registered-payment-method-props').RegisteredPaymentMethodProps} RegisteredPaymentMethodProps`

mikejolley:
@ralucaStan I think given an API like `getExtensionsData('my-payment-namespace');` extensions could still read other properties, but this just adds extra steps. @nerrad can you chime in here please?

> Should we mention in the description this PR comes with breaking-changes? I see Stripe gateways uses this import 

I think this would just cause an error in the IDE right? It wouldn't impact builds.

senadir:
> @ralucaStan I think given an API like getExtensionsData('my-payment-namespace'); extensions could still read other properties, but this just adds extra steps. @nerrad can you chime in here please?

Even if we don't do that, and just match the payment namespace <-> extension namespace and only pass that, we would still be limiting things for no reason. I don't expect that extensions would be accessing other extensions data, so for 95% use cases, we would be creating an extra layer of abstraction that doesn't match the rest of our API.

ralucaStan:
> I think this would just cause an error in the IDE right?

What I saw happening is that for example, in this case, `@typedef {import('@woocommerce/type-defs/registered-payment-method-props').RegisteredPaymentMethodProps} RegisteredPaymentMethodProps` RegisteredPaymentMethodProps will have type any, because `.RegisteredPaymentMethodProps` will fail

mikejolley:
`RegisteredPaymentMethodProps will have type any, because .RegisteredPaymentMethodProps will fail`

That sounds reasonable. I wouldn't call this a breaking change, but I will add the dev notes label so it's highlighted in the release post.

nerrad:
In response to

> I've also saw in #5042 that there is this request, If possible, limit the data exposed. In other words, instead of passing along all of the data on the extensions property of the response, just return the payment method specific data. Would it be possible to have for extensions a something like a getter that returns a certain key for the extensions object?

and:

> @ralucaStan I queried that here: p1637149910386300-slack-C8X6Q7XQU
>
> >I was mostly wanting to avoid passing in data that potentially frequently updates, but doesn't look like we can avoid that anyways here. As long as we just pass in read-only data from the hook I'm good with that.
>
> A getter would solve immediately exposing properties, but all properties would still be exposed right? We'd just be introducing an extra layer of friction for extensions to retrieve the data. Are there other benefits I'm overlooking?

In the original issue, I was mostly wanting to make sure we considered limiting the data payment method extensions were exposed to given the risk a payment method could mutate any passed in data causing hard to troubleshoot bugs in for any other integrations relying on that data. I think it's a valid risk given we could see folks developing integrations who are more familiar with the PHP context where mutation by reference isn't a big deal. They might see this as an opportunity to try to control other extension behaviour via this mutation.

With that said, we already have that risk right now with the objects we're currently passing along to payment methods via this interface so I'm fine with rolling this out for now and if this risk becomes a reality in the real world, there are ways we can address that which could help (object copy or making the data immutable). Doing that now could introduce performance issues - so let's address it if it surfaces as a problem. 

However, I do wonder, for the sake of folks unfamiliar with JS working with this interface, if we should call out in the docs and/or inline code blocks that the data provided via this interface should be treated as immutable and not mutated (giving examples of what that means)?

nerrad:
> @mikejolley Should we mention in the description this PR comes with breaking-changes? I see Stripe gateways uses this import
> > @typedef {import('@woocommerce/type-defs/registered-payment-method-props').RegisteredPaymentMethodProps} RegisteredPaymentMethodProps

Unless the Stripe extension is already aliasing `@woocommerce/type-defs` somehow to the WooBlocks source, or something internal in their extension, this is already "broken" for them I think. We currently don't export a package for the `@woocommerce/type-defs` namespace. So, I don't think a dev note calling attention to this is necessary.

mikejolley:
I added an inline note. @ralucaStan can you give final ✅ if it can be merged?

ralucaStan:
> I added an inline note. @ralucaStan can you give final ✅ if it can be merged?

Looks good. One small note, I've noticed that `cartData` is not added [in this table](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/payment-method-integration.md#props-fed-to-payment-method-nodes), so as the last change to this PR let's add it along with the immutability comment.

ralucaStan (approved):
👏

PR #5184, opened 2021-11-18
Mini Cart block: Add label for tax when it should be visible in accordance with WC settings

### LINKED ISSUE #4849: Mini Cart block: add `(incl. VAT)` text on some settings combinations
## To reproduce
1. Go to WooCommerce > Settings and check this option: 'Enable tax rates and calculations'.
2. In the Taxes tab, select 'No, I will enter prices exclusive of tax' and set 'Display prices during cart and checkout' to 'Including tax'.
3. You will notice the Storefront mini cart displays `(incl. VAT)` text in the Mini Cart.

imatge

4. The Mini Cart block should behave in the same way.

gigitux:
Mini Cart block: add a label for tax when it should be visible in accordance with WC settings

Fixes #4849

### Screenshots

image
*When `Display prices during cart and checkout` is set to `Including tax`*

image
*When `Display prices during cart and checkout` is set to `Excluding tax`*

### Manual Testing

Check out this branch.

1. Make sure you have Woo Blocks plugins installed.
2. Go to WooCommerce > Settings and check this option: 'Enable tax rates and calculations'.
3. In the Taxes tab, select 'No, I will enter prices exclusive of tax' and set 'Display prices during cart and checkout' to 'Including tax'.
4. Create a page with the `Mini Cart` block and save it.
5. Check that the `Mini Cart` block has no label when there isn't a product in the cart.
6. Check that the `Mini Cart` block, when there is at least one product in the cart, has `(incl. VAT)` or `(incl. tax)` label (it depends by the country).


Aljullu (commented):
Code looks good and overall testing well, but I found a couple of discrepancies on the way it works compared to Storefront mini cart (which we should use as a reference):

* If you go to WooCommerce > Tax > Standard rates and remove all rates (or set the _Rate %_ column to 0), you will notice Storefront's mini cart doesn't display `(incl. VAT)` while the Mini Cart block does.
* If you go to WooCommerce > Tax > Tax options and set _Prices entered with tax_ to _Yes..._ and _Display prices..._ to _Excluding..._ you will notice SF's mini cart has a `(excl. VAT)` string while the block, doesn't.

gigitux:
@Aljullu Thanks Albert for pointing me these problems. Actually, I focused only on the use case you indicated in the issue.
Instead, after your comment, I noticed that many cases were still not being handled (it just lacked the logic).

In my tests, sometimes the storefront cart did not update properly after changing the settings: so my suggestion if this happens, it is to clear the cache or change the browser.

Aljullu:
> @Aljullu Thanks Albert for pointing me these problems. Actually, I focused only on the use case you indicated in the issue.

Ah, good point. I'm fine leaving those cases for a follow-up if you prefer.

gigitux:
> Ah, good point. I'm fine leaving those cases for a follow-up if you prefer.

No, I updated the PR and everything should be okay.
I'm waiting for your review and after that, if I don't miss any case, I will update the description of the PR, and I will add these repro steps 👍

src/BlockTypes/MiniCart.php
```diff
@@ -338,4 +403,53 @@ protected function get_markup( $attributes ) {
...
+	 * Get array with data for handle the tax label.
+	 * the entire logic of this function is was taken from:
+	 * https://github.com/woocommerce/woocommerce/blob/e730f7463c25b50258e97bf56e31e9d7d3bc7ae7/includes/class-wc-cart.php#L1582
+	 *
+	 * @return array;
+	 */
+	protected function get_tax_label() {
+		$cart = WC()->cart;
+
+		if ( $cart->display_prices_including_tax() ) {
```
Aljullu, line 443: I'm getting this error in the editor in this line:

> Fatal error: Uncaught Error: Call to a member function display_prices_including_tax() on null
in /wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTypes/MiniCart.php on line 426

src/BlockTypes/MiniCart.php
```diff
@@ -74,6 +104,26 @@ protected function enqueue_data( array $attributes = [] ) {
...
+			'labelIncludingTax',
+			$this->label_including_tax,
+			''
+		);
+
+		$this->asset_data_registry->add(
+			'displayCartPricesIncludingTax',
+			$this->display_cart_prices_including_tax,
+			false
+		);
```
Aljullu: This is also producing an error for me:

> Fatal error: Uncaught Exception: Overriding existing data with an already registered key is not allowed
in /wp-content/plugins/woocommerce-gutenberg-products-block/src/Assets/AssetDataRegistry.php on line 380

src/BlockTypes/MiniCart.php
```diff
@@ -338,4 +403,53 @@ protected function get_markup( $attributes ) {
...
+	 * the entire logic of this function is was taken from:
+	 * https://github.com/woocommerce/woocommerce/blob/e730f7463c25b50258e97bf56e31e9d7d3bc7ae7/includes/class-wc-cart.php#L1582
+	 *
+	 * @return array;
+	 */
+	protected function get_tax_label() {
+		$cart = WC()->cart;
+
+		if ( $cart->display_prices_including_tax() ) {
+			if ( $cart->get_subtotal_tax() > 0 && ! wc_prices_include_tax() ) {
```
Aljullu: One case that is not working for me is that, if you start with an empty cart, now it correctly doesn't include the `(incl. VAT)` string, but when you add a product to your Cart, the string doesn't appear when it should. Instead, you need to refresh the page.

Aljullu (commented):
Thanks for updating the PR and fixing the conflicts! All cases are testing great now, but I found a couple of errors in the editor, and there is one case which doesn't work right yet: when the cart is empty and you add the first product.

src/BlockTypes/MiniCart.php
gigitux: Now it should be fine!

src/BlockTypes/MiniCart.php
gigitux: Now it should be fine!

src/BlockTypes/MiniCart.php
gigitux: Oh, great catch, thanks!

The problem was the check `$cart->get_subtotal_tax() > 0`.

The BE always has to return the information for handling the tax label. The client display or not the tax label based on whether there are products in the shopping cart

assets/js/blocks/cart-checkout/mini-cart/block.tsx
```diff
@@ -113,7 +114,19 @@ const MiniCartBlock = ( {
...
 
-	const subTotal = getSetting( 'displayCartPricesIncludingTax', false )
+	const showIncludingTax = getSettingWithCoercion(
+		'displayCartPricesIncludingTax',
+		false,
+		isBoolean
+	);
+
+	const labelTaxName = getSettingWithCoercion(
+		'labelIncludingTax',
```
Aljullu: I feel this setting name might be confusing, given that the value might be _(**excl.** VAT)_. What about just calling it `taxLabel`?

assets/js/blocks/cart-checkout/mini-cart/block.tsx
```diff
@@ -154,6 +167,11 @@ const MiniCartBlock = ( {
 						getCurrencyFromPriceResponse( cartTotals )
 					) }
 				</span>
+				{ labelTaxName !== '' && subTotal !== 0 && (
+					<small className="wc-block-mini-cart__tax_label">
```
Aljullu: [We try to use BEM for CSS class names](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/contributors/coding-guidelines.md#naming), so `_` should only be used to separate parts of the selector. When it's the same part but different words, we use `-`:
```suggestion
					<small className="wc-block-mini-cart__tax-label">
```

assets/js/settings/shared/utils.ts
```diff
@@ -25,6 +25,15 @@ export const getSetting = < T >(
 	return filter( value, fallback ) as T;
 };
 
+export const getSettingWithCoercion = < T >(
```
Aljullu, line 28: Could we add some tests to this function? Similar to what we have for `getSetting()`:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a180f22ecf0de60547cdc881bed1df6b4847a4a1/assets/js/settings/shared/test/get-setting.js

src/BlockTypes/MiniCart.php
```diff
@@ -338,4 +403,53 @@ protected function get_markup( $attributes ) {
...
+				'display_cart_prices_including_tax' => true,
+			);
+		}
+
+		if ( wc_prices_include_tax() ) {
+			$label_including_tax               = WC()->countries->ex_tax_or_vat();
+			$display_cart_prices_including_tax = true;
+			return array(
+				'label_including_tax'               => $label_including_tax,
+				'display_cart_prices_including_tax' => false,
```
Aljullu: This is a bit confusing, we set `$display_cart_prices_including_tax` to true, but never use it again inside this method. But then, it returns `'display_cart_prices_including_tax' => false`. Was that an overlook or am I missing something?

src/BlockTypes/MiniCart.php
```diff
@@ -189,6 +239,19 @@ protected function append_script_and_deps_src( $script ) {
...
+	 * Returns the markup for render the tax label.
+	 *
+	 * @return string
+	 */
+	protected function get_include_tax_label_markup() {
+		$cart_controller     = $this->get_cart_controller();
+		$cart                = $cart_controller->get_cart_instance();
+		$cart_contents_total = $cart->get_subtotal();
+
+		return ( ! empty( $this->label_including_tax ) && 0 !== $cart_contents_total ) ? ( "<small class='wc-block-mini-cart__tax_label'>" . $this->label_including_tax . '</small>' ) : '';
```
Aljullu: We should escape this label:
```suggestion
		return ( ! empty( $this->label_including_tax ) && 0 !== $cart_contents_total ) ? ( "<small class='wc-block-mini-cart__tax_label'>" . esc_html( $this->label_including_tax ) . '</small>' ) : '';
```

src/BlockTypes/MiniCart.php
```diff
@@ -77,6 +107,26 @@ protected function enqueue_data( array $attributes = [] ) {
...
 		if ( ! is_admin() && ! WC()->is_rest_api_request() ) {
 			$this->hydrate_from_api();
+
+			$label_info = $this->get_tax_label();
+
+			$label_including_tax               = $label_info['label_including_tax'];
+			$display_cart_prices_including_tax = $label_info['display_cart_prices_including_tax'];
+
+			$this->label_including_tax               = $label_including_tax;
+			$this->display_cart_prices_including_tax = $display_cart_prices_including_tax;
```
Aljullu: Could this be compressed in two lines?

```suggestion
			$this->label_including_tax               = $label_info['label_including_tax'];
			$this->display_cart_prices_including_tax = $label_info['display_cart_prices_including_tax']
```

Aljullu (commented):
Thanks for updating this PR @gigitux! Confirming the issues in the editor are fixed now. :ok_hand: 

I left some more comments below, but overall it's looking good.

src/BlockTypes/MiniCart.php
gigitux: I forgot to remove it

assets/js/settings/shared/test/get-setting-with-coercion.ts
```diff
@@ -0,0 +1,28 @@
...
+	it( 'returns expected value for existing setting', () => {
+		expect(
+			getSettingWithCoercion( 'adminUrl', 'not this', isString )
+		).toEqual( ADMIN_URL );
+	} );
+	it( "returns expected value for existing setting even if it's a falsy value", () => {
+		expect(
+			getSettingWithCoercion( 'currentUserIsAdmin', true, isBoolean )
+		).toBe( false );
+	} );
```
Aljullu, line 32: Thanks for adding the tests! Should we add an additional test to verify that the fallback is returned if the typeguard returns false? I think none of the tests above covers that case.

Aljullu (approved):
This is looking great. The only last thing I would change is the alignment between the tax label and the surrounding elements.

imatge

I would try to align them vertically: ideally `32,72€` and `(excl. VAT)` would have the same baseline. I would also add some extra right margin to the tax label so there is more space between it and the icon, but leaving it up to you.

I'm pre-approving given that these last changes are quite small. Good job!

gigitux:
> This is looking great. The only last thing I would change is the alignment between the tax label and the surrounding elements.
> 
> imatge
> 
> I would try to align them vertically: ideally `32,72€` and `(excl. VAT)` would have the same baseline. I would also add some extra right margin to the tax label so there is more space between it and the icon, but leaving it up to you.
> 
> I'm pre-approving given that these last changes are quite small. Good job!

image

Mmm, it is very strange. Above, you can find an image of how my browser renders the element. Can you check again?


In any case, I added more space between the tax label and icon:
image

Aljullu (approved):
Weird, I'm not able to reproduce it anymore. Let's get this merged! :shipit:

PR #5197, opened 2021-11-19
Fix manual entry within Quantity Inputs

### LINKED ISSUE #5187: Unable to type in Quantity Inputs
## Describe the bug
Typing a value is not possible in a quantity input. It remains at 1.

If the quantity input has a value higher than one, typing resets it to 1.

## To reproduce

1. Go to the cart page or open the mini cart block
2. Click on the plus/minus buttons - qty changes ok
3. Click within the qty input and try to type a numeric value. Nothing happens.

## Expected behaviour
I should be able to enter a valid numeric qty manually.

mikejolley:
Typing into quantity inputs is currently not possible because the `isNumber` type check does not convert string-based numbers into actual numbers.

This fixes the problem by using `Number.isNaN` instead.
 
Fixes #5187

### Testing

Same as below.

### User Facing Testing

- Go to the cart page with an item in the cart
- Try typing a quantity into the line item
- Totals should update
- Try typing a letter. Input should be ignored and reset to `1`

### Changelog

> Fix manual entry within Quantity Inputs in Cart block

### DIFF
--- a/assets/js/base/components/quantity-selector/index.tsx
+++ b/assets/js/base/components/quantity-selector/index.tsx
@@ -6,7 +6,6 @@ import { speak } from '@wordpress/a11y';
 import classNames from 'classnames';
 import { useCallback } from '@wordpress/element';
 import { DOWN, UP } from '@wordpress/keycodes';
-import { isNumber } from '@woocommerce/types';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
Confirming it fixes the issue, thanks for investigating and fixing this, @mikejolley!

WC Blocks 6.3.2 seems to be affected by this. For now, I'm marking this as a blocker of WC Blocks 6.4.0, but I think we might need to include this in a WC Blocks 6.3.3 release as well. cc @nielslange

Issue #5195, opened 2021-11-19
Mini Cart: allow selecting a specific template part for the Mini Cart block


Aljullu:
Given that it will be possible to have more than one Mini Cart template part (see #5194), we want to allow users to select which Mini Cart template part is used for each Mini Cart block they add to their site.

This issue is about adding some kind of control so users can select from a list of template parts when adding a Mini Cart block.

Aljullu:
We decided not to allow users to create additional template parts in the Mini Cart template part area. Because of that, this issue is no longer needed.

PR #5201, opened 2021-11-19
Mini Cart Block: add link to edit template part from sidebar

### LINKED ISSUE #5186: Mini Cart: add link to edit template part from Mini Cart block sidebar
With #5025, the Mini Cart drawer contents are inside a template part. In order to offer the users an easy way to modify the drawer contents when inserting the Mini Cart block, we would like to offer a link in the sidebar of the Mini Cart block directing to the Mini Cart contents template part.

## To reproduce
1. Create a post or page and add the Mini Cart block.
2. Notice there is no link to edit the drawers contents of the Mini Cart block.

## Expected behavior
We would like to have a link to the template part editor.

## Screenshots
mini-cart-link

We don't have a design, but I think for now we can use a similar layout as the Payment Options block (inside the Checkout block), which redirects to Managing payment methods:

gigitux:
This PR renders a link on the right sidebar to allow the user to land on the page to edit the `Mini Cart` template part.

Fixes #5186 

### Screenshots

image

### Manual Testing

Check out this branch 

1. Make sure you have Woo Blocks plugins installed and a theme that supports FSE.
2. Create a page.
3. Add the `Mini Cart` block and click on it.
4. On the right sidebar, in the block section, check if there is a `Template Editor` menu.
5. Check that there is a link and the link structure is `/wp-admin/themes.php?page=gutenberg-edit-site&postId=$SLUG_THEME%2F%2Fmini-cart&postType=wp_template_part`.
6. Change the theme and use a theme that doesn't support FSE.
7. Create a page.
8. Add the `Mini Cart` block and click on it.
9. On the right sidebar, in the block section, check if there isn't the `Template Editor` menu.


assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -87,6 +94,32 @@ const MiniCartBlock = ( {
 						}
 					/>
 				</PanelBody>
+				{ isSiteEditorAvailable &&
+					isString( themeSlug ) &&
```
gigitux, line 142: I don't like this check, but the type of themeSlug is `unknown`. Maybe I can refactor this when [`getSettingWithCoercion` function ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5184/files#diff-0d3952ed1d1f5ca7da0fbccaaab5bb07976920377492db9ce119d91f80a5b5d3R28)will be merged

dinhtungdu (commented):
@gigitux This is working for the themes that have `mini-cart` template. For ones that don't have that template and use `mini-cart` template from the plugin, we get this error:
image

dinhtungdu:
By the way, because this PR is under Mini Cart Epic, so we should skip Change log generation for this.

gigitux:
> @gigitux This is working for the themes that have `mini-cart` template. For ones that don't have that template and use `mini-cart` template from the plugin, we get this error: image

In accordance with internal discussion (`p1637661565445900-slack-C02FL3X7KR6`) for now, we can skip this

assets/js/blocks/cart-checkout/mini-cart/edit.tsx
```diff
@@ -52,6 +55,10 @@ const MiniCartBlock = ( {
 		} ),
 	} );
 
+	const themeSlug = getSetting( 'themeSlug', '' );
```
dinhtungdu, line 58: I don't have a strong opinion here but we can get the theme data using Gutenberg selector, like the following snippet. Both are working for me but wondering if we have a preferred way to get the core data? cc @Aljullu 

```
	const currentTheme = useSelect( ( select ) =>
		select( coreStore ).getCurrentTheme()
	);
```

dinhtungdu: Note that `useSelect` dispatch a REST request, so the theme data isn't available right after loaded. This is a disadvantage compared to the current way.

Aljullu: > Note that `useSelect` dispatch a REST request, so the theme data isn't available right after loaded. This is a disadvantage compared to the current way.

That's a good point! I think we can go now with the solution @gigitux implemented and refactor it if needed in the future. How does it sound? In #5195 we will need to make the link dynamic, so we might need to do further adjustments to the logic introduced in this PR.

dinhtungdu: @Aljullu Sounds good to me!

dinhtungdu (approved):

gigitux:
@dinhtungdu I updated the PR because there was an error with E2E. It would be great if you can review the last commit. Thanks!

dinhtungdu (approved):

Issue #5192, opened 2021-11-18
Checkboxes appear squished if paired with longer text


sunyatasattva:
## Describe the bug
Within certain themes (e.g. TT1), checkboxes paired with a label, will appear squished if the text wraps to a second line. This only happens in the front-end.

## To reproduce

1. Add a checkout block
2. Select Terms and Conditions block
3. Enable `Require checkbox` attribute
4. Preview your page
5. Compare the checkbox next to “Add a note to your order” and the one next to “Terms and conditions”.

*Optional:*

1. Remove text from the “Terms and conditions” until it no longer wraps
2. See the difference

## Expected behavior
Checkboxes should maintain a square aspect-ratio.

## Screenshots

![Screen Shot 2021-11-18 at 23 46 40](https://user-images.githubusercontent.com/1847066/142509470-72e3f025-42b4-4346-abf1-12de890c6316.png)

sunyatasattva:
Fixed by #5206

PR #5206, opened 2021-11-21
Checkbox Control: Fix squished checkboxes when label text wraps (#5192)

sunyatasattva:
On the front-end, checkboxes with labels would look squished when the labels wrapped on a second line. That was due to the flex container taking too much space for the label as the checkbox minimum dimensions was set in `px` rather than `em`, making it work differently depending on the relative font size.

Fixes # 5192

### Screenshots

*Before:*

before

*After:*

Screen Shot 2021-11-21 at 17 59 38

### Manual Testing

1. Add a checkout block
2. Select Terms and Conditions block
3. Enable Require checkbox attribute
4. Preview your page
5. Make sure one label of the checkbox wraps to two lines on your screen.
6. Confirm square aspect ratio

### User Facing Testing
* [X] Same as above


Aljullu (approved):
LGTM :shipit:

Issue #5180, opened 2021-11-17
UX improvement on Legacy Blocks deletion


sunyatasattva:
**For full context and history, cfr:** #5166, #5176

While we are in the process of phasing out the legacy PHP blocks, an interesting UX question arises in the transition period. With the issues linked above, we have decided to revert our previous judgement to make those blocks locked and un-erasable for the user.

The reasoning was that erasing those blocks was probably not something the users wanted to do, and possibly accidental (or they didn't know the consequences thereof). The side-effect was that these blocks could not be moved either.

With #5176 we decided that disempowering the user to move the blocks just to protect them from themselves was defeating the purpose of FSE. However, legacy blocks do not appear normally on our Block Inserter (because they might not be compatible with the page the user is trying to add them to).

This means that, after an accidental deletion, the user **cannot intuitively** add the block back.

A workaround currently is to delete the template the user has been working on, and thus reverting to the default override. This workaround has two drawbacks:

1. It's hard to discover for the user.
2. The user will lose any other change they have done to the template prior to the accidental deletion of the legacy block.

This ticket is meant to spark a discussion on what the solutions to this problem might be.

## Proposed solutions

1. Ignore the issue as it's going to go away when blocks get phased out. This is not what I would recommend, but indeed a possibility worth discussing. It is currently unclear to me what the timing of this phasing out can be, and this issue potentially can bring many headaches to users and HEs alike.
2. Check if we can feasibly identify which pages are compatible with which blocks and restrict block insertion to those pages. This is probably a longer-term/higher-effort solution.
3. Figure out if there is any way we can lock deletion but not moving.
4. Add some copy to the block preview that warns the user about the situation, and the effects of deleting the block. This is a UX pattern I see relatively often when, e.g., certain more side-effects heavy settings are hidden or grouped below a *“HERE BE DRAGONS”*-kind of section (GitHub, for example, does this for certain irreversible actions). 
  * I believe this to be the lowest hanging fruit. It's a short-term solution, and the drawback is that, if the notice is in the copy, once you delete the block you don't have access to the notice anymore (which was perhaps the place where instructions to reverse would be). But, nonetheless, I think it's something quick that would add value.
[...]

PR #5208, opened 2021-11-21
Legacy Template: Add warning about block removal (#5207)

### LINKED ISSUE #5207: Add note to make it clear that removing a legacy block is dangerous
**Follow-up issue from: #5180**

Following another round of testing of the 6.3.1 release, we decided to revert our previous decision to prevent the block deletion of the Legacy Template block, however if the user deletes the block it is hard for them to know (a) how to revert the change and (b) what consequences does this deletion have.

As such, we are trying to make it more clear for the user the consequences of this action by adding a notice to the block itself, and at least some kind of troubleshooting FAQ as suggested in the discussion of the linked issue.

sunyatasattva:
**For full context, read: #5180**

Legacy blocks were previous locked for removal to avoid unintended consequences. However, this would lock the ability to move the block within other e.g. layout blocks and unnecessarily limit merchant customization ability.

Now that we have [reverted this decision](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5176), merchants could delete this block, which is likely **not** what they want to do. While we investigate other, more long-term, solutions, we are adding a warning notice.

Refs #5180.

Fixes #5207

## Discussion

> ⚠️ This PR is a draft, we need to come up with the definitive copy

@vivialice I have consulted with @frontdevde and he mentioned you would be the best person to come up with some good copy for this block. I took some of your inputs from the conversation at #5180 and added them as a draft. But I'd be super thankful if you could come up with the best version of this :)

Also, [you have mentioned](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5180#issuecomment-972729497) adding some sort of link to troubleshooting docs. Me and @frontdevde tried to look for the best place for it in WooCommerce.com dashboard, but were not sure, so I'd love to ask your guidance here as well 🙏 

### Screenshots

*Before:*
Screen Shot 2021-11-21 at 18 24 50

*After:*

Screen Shot 2021-11-21 at 18 20 08

### Manual Testing

1. Go to `Appearence → Editor`
2. Select `General templates → Product Archive Page`
3. See longer warning notice on the legacy block

### User Facing Testing
* [X] Same as above


vivialice:
Hey @sunyatasattva ! The mockup looks great, I have made a few tweaks to the copy: 

> Attention: Do not remove this block! Removal will cause unintended effects on your store. We are working on a better editing experience that will replace legacy blocks. Keep an eye out for updates! 
> 
> This is a placeholder for the WooCommerce Product Archive Template. It will be displayed on your store as a grid of your product image(s), title, price, etc. You can move this placeholder around and add blocks around it to extend the template.

I think we may have pulled the notice UI from this at some point but I have mocked up a version with it as I think it works a bit better to balance out the messaging here. 

Screenshot 2021-11-23 at 15 01 33

As for placement for the link to troubleshooting docs, I have a couple ideas I'd be keen to hear everyones thoughts on:

This is the 'Help' section on the WooCommerce plugin. We could include a top-level link in this list, "WooCommerce Templates" or something like that. 
 
Screenshot 2021-11-23 at 15 19 22

We could also use this feedback component to direct to docs. We could include it on all Woo templates. 

Troubleshooting 2

vivialice:
Hey there 👋  

Just a quick follow up thought. I discussed these with my design team and they brought up a good point. The naming of these are unclear to a general user on their intended purpose. Also, using the word "template" is unclear as the page itself is a template and it is replacing a block. 

I suggest renaming the Product Archive block to: WooCommerce Product Grid block and the others to Product Category Grid block. Single Product block is clear enough. 

Example:
Edit Suggestion

sunyatasattva:
Excellent point @vivialice ! Exactly why I asked you as a designer, given that sometimes developers tend to get stuck into devspeech and won't notice these sorts of details.

As no one else has provided more feedback, I'll proceed with this ASAP (once my current pending task is done). Would you be able to update the troubleshooting as you proposed,  or could you show me around perhaps?

vivialice:
@sunyatasattva Awesome! 

Re: Troubleshooting suggestions. I am not too sure how to add something to the Help menu but I will find out after I am back from my AFK on Wed, Dec 01. Do we already have documentation somewhere? Happy to help set up a page if we don't. 

For the second suggestion, we use something like this as a feedback component on some of our newer blocks. I think Cart & Checkout should have them. I was thinking we could re-purpose these.

sunyatasattva:
@vivialice Indulge me, I'm super new and this is still my second week, so please forgive my excessive questions.

> Re: Troubleshooting suggestions. I am not too sure how to add something to the Help menu but I will find out after I am back from my AFK on Wed, Dec 01. Do we already have documentation somewhere? Happy to help set up a page if we don't.

I do not know 😬 We were looking at this with @frontdevde but we weren't 100% sure.

> For the second suggestion, we use something like this as a feedback component on some of our newer blocks. I think Cart & Checkout should have them. I was thinking we could re-purpose these.

Could you elaborate a bit on that if you don't mind? What do you think we could repurpose?

nerrad:
> Could you elaborate a bit on that if you don't mind? What do you think we could repurpose?

I think Viv is referring to [this component](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/editor-components/feedback-prompt/index.js). However, it would likely need to be tweaked to have the flexibility of the copy Viv suggested.

sunyatasattva:
Oh, I see! Thank you @nerrad! I've looked around to this component and let me see if I understand: you are suggesting that we add both the little notice and the feedback block, but dividing the copy in two: leaving the notice in the block itself, and the part saying:

> We are working on a better editing experience that will replace legacy blocks. Keep an eye out for updates!

In the feedback part on the right sidebar as shown below. Is that correct?

Screen Shot 2021-11-29 at 23 36 47

If that's the case, currently, clicking on the *feedback* link brings the user to a specific GitHub new issue template form. I have got two questions here:

1. For @nerrad : Don't you think that given our proposal to start using GitHub discussions, this would be a good opportunity to bring the user to that place, instead of the Issues?
2. For @vivialice : Since you are the Copy Master 🙇  would you be able to provide some meaningful template for the feedback to be useful to us? You can check [this template](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/.github/ISSUE_TEMPLATE/--cart-checkout-feedback.md) for inspiration.

---

> I suggest renaming the Product Archive block to: WooCommerce Product Grid block and the others to Product Category Grid block. Single Product block is clear enough.

Just so I get it right, do you suggest renaming both the Category and Taxonomy templates to “Category Block” or just changing the “Template” word with “Block”? 

---

Here is a screenshot implementing all these things as I have understood them:

Screen Shot 2021-11-30 at 00 08 38

vivialice:
Hey @sunyatasattva 👋 I imagined we could link the user to a documentation page with some details on WooCommerce templates rather than Github, so perhaps that suggestion doesn't work. For now, we can make sure Woo support understands how to help users who have accidentally removed the placeholders from the page since this is a temporary solution. I am not sure how many users will be working with the templates at this point. 

Moving forward, it would be great to have some documentation on Woo FSE for both merchants and site builders - I will look into this. 

> Just so I get it right, do you suggest renaming both the Category and Taxonomy templates to “Category Block” or just changing the “Template” word with “Block”?

I would replace the word "Template" with "Block" 😸

frontdevde:
To my mind, the Feedback section on the right feels quite disconnected from what's going on in the editor canvas. By the time a person would benefit most from reading support documentation on how to add the block back, they would have likely already deleted the block and the legacy feedback prompt wouldn't be present anymore.

It probably would be most valuable to include a link to documentation in the notice that warns the user about deleting the block.

Either way, given that we don't have documentation to link out to just yet, can we limit the changes here to updating the header wording and adding the warning notice? We could then address the documentation link as a follow-up issue. This would unblock this for now. Just having the notice seems like a good improvement to go in asap.

sunyatasattva:
@frontdevde Done. Please check that this is satisfying those requirements and let me know if there is anything else to do. 🙏

tjcafferkey (approved):
The changes LGTM based on the discussion on this PR. I've also rebased this branch so it's good to be merged.

PR #5213, opened 2021-11-22
Allow saved HTML to render

senadir:
This doesn't fix any bug I'm aware of, but we had a limitation in which passed HTML (saved via save function) would be stripped if it had tags in it, this fixes it.

It also fixes an issue in Checkbox in which spaces are removed on those type of strings.

This PR tried to update the Terms and Conditions block to use this, but it's not possible to migrate an attribute source for some reason. For any new block that has inline text, this is the recommended way of doing things.

### Testing steps

- Load this PR https://github.com/opr/newsletter-test/pull/6/
- Insert the block into checkout, save.
- Visit frontend, it should render fine.
- Edit Checkout again, edit the text and add formatting, save.
- Visit Checkout, it should be respected.

### DIFF
--- a/assets/js/atomic/utils/render-parent-block.tsx
+++ b/assets/js/atomic/utils/render-parent-block.tsx
@@ -7,6 +7,7 @@ import {
 	Suspense,
 	isValidElement,
 	cloneElement,
+	Children,
 } from '@wordpress/element';
 import parse from 'html-react-parser';
 import {
@@ -153,8 +154,7 @@ const renderInnerBlocks = ( {
 		 */
 		if ( ! InnerBlockComponent ) {
 			const parsedElement = parse(
-				element?.outerHTML || element?.textContent || '',
-				{ trim: false }
+				element?.outerHTML || element?.textContent || ''
 			);
 
 			// Returns text nodes without manipulation.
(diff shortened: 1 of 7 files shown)


assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -137,7 +138,7 @@ const renderInnerBlocks = ( {
 		const { blockName = '', ...componentProps } = {
 			key: `${ block }_${ depth }_${ index }`,
 			...( element instanceof HTMLElement ? element.dataset : {} ),
-			className: element.className || '',
+			className: element.className || undefined,
```
senadir, line 141: Having `''` caused `class` to be appended anyway on elements, this caused some elements like `<b>` to have an empty class attribute.

assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -190,42 +191,49 @@ const renderInnerBlocks = ( {
...
-						}
+						{ Children.toArray( [
+							innerChildren,
+							innerForcedChildren,
+						] ).length
+							? Children.toArray( [
+									innerChildren,
+									innerForcedChildren,
+							  ] )
+							: null }
```
senadir, line 236: `Children.toArray` handles filtering out null.

packages/checkout/components/checkbox-control/index.tsx
```diff
@@ -44,23 +44,23 @@ const CheckboxControl = ( {
...
+				{ ...rest }
+			/>
+			<svg
+				className="wc-block-components-checkbox__mark"
+				aria-hidden="true"
+				xmlns="http://www.w3.org/2000/svg"
+				viewBox="0 0 24 20"
+			>
+				<path d="M9 16.2L4.8 12l-1.4 1.4L9 19 21 7l-1.4-1.4L9 16.2z" />
+			</svg>
```
senadir, line 62: This moves input and svg outside label and into div, now everyone is on the same level, but only text is inside label.

src/BlockTypesController.php
```diff
@@ -110,6 +110,12 @@ public function add_data_attributes( $content, $block ) {
 			if ( in_array( $key, $exclude_attributes, true ) ) {
 				continue;
 			}
+
+			// Attributes already persisted in HTML should be skipped.
+			if ( is_string( $value ) && is_int( strrpos( $content, $value ) ) ) {
+				continue;
+			}
+
```
senadir, line 118: This is probably the trickiest part that I'm not sure about, basically, we don't append attributes that are already present on the HTML, this includes strings, but would also skip classNames, I'm not sure what this might omit as well, but I can't think of anything.

mikejolley: I think we can just leave the duplication if it's harmless. It's less risky and could be removed in the future if we find a way to grab block attributes via JSON or some other way.

mikejolley (changes requested):
I've tested against cart, checkout, and single product block. Looks to be working.

- I left one comment.
- [I had to fix newsletters](https://github.com/opr/newsletter-test/pull/6/commits/49487f001221e4ca83232554e89c448f737074d6)
- There is extra margin in the editor between the checkbox and the text

Thats all I spoted.

PR #5215, opened 2021-11-23
Change source value on WooCommerce block templates to be plugin

tjcafferkey:
Related to https://github.com/WordPress/gutenberg/issues/36597#issuecomment-976232909

Change the `source` attribute of the templates to be consistent with its intended use, where in this case the value should be `plugin` instead of `woocommerce`

### Manual Testing

1. Smoke test the Store Editing Templates project:
2. Follow prerequisite tests from the [epic](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4926).
3. Ensure templates load in the Site Editor.
4. Create templates inside the active themes `block-templates` directory to ensure the theme's template supersedes the plugins by loading it in the Site Editor and on the frontend
5. Ensure when editing and saving these templates they save correctly and are visible in Appearance > Templates. Check these correctly load in the Site Editor and render on the frontend.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

1.
2.
3.

### DIFF
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -348,7 +348,7 @@ class BlockTemplatesController {
 				'path'        => $template_file,
 				'type'        => $template_type,
 				'theme'       => 'woocommerce',
-				'source'      => 'woocommerce',
+				'source'      => 'plugin',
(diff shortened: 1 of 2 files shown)


frontdevde (approved):

opr (approved):
LGTM! 🚢

PR #5216, opened 2021-11-23
Refactor Mini Cart contents block unmount to fix ESLint warning

Aljullu:
It turns out we had this ESLint warning:

> [react-hooks/exhaustive-deps] The ref value 'contentsRef.current' will likely have changed by the time this effect cleanup function runs. If this ref points to a node rendered by React, copy 'contentsRef.current' to a variable inside the effect, and use that variable in the cleanup function.

in [assets/js/blocks/cart-checkout/mini-cart/block.tsx#L77](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/abc7b14dda5e34e2b386b7f18e7d4ec52601d3e6/assets/js/blocks/cart-checkout/mini-cart/block.tsx#L77).

While the current implementation seems to work, I refactored it using this [post from the React docs](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node). In short: instead of using `useRef()` it uses `useCallback()`.

### Manual Testing

1. Add the Mini Cart block to a post or page.
2. In the frontend, verify you can click on the button and it opens the drawer.
3. Clicking outside (or on the close button) should close the drawer.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart/block.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart/block.tsx
@@ -7,7 +7,7 @@ import {
 	RawHTML,
 	useState,
 	useEffect,
-	useRef,
+	useCallback,
 	unmountComponentAtNode,
 } from '@wordpress/element';
 import {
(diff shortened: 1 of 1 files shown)


tjcafferkey (approved):
LGTM and tests well 🚢

PR #5219, opened 2021-11-23
Display decimals if intended by the merchant

### LINKED ISSUE #5019: `formatPrice` strips decimals

nielslange:
Fixes #5019 

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests
  
Run the following command in the terminal:
```
npm run test -i packages/prices/utils/test/price.js
```
  
#### Test $0 price

1. Create a test page, add the mini cart block and save it.
2. Open the mini cart page in the frontend.
3. See that the price shows $0.00.
4. Hover over the price and verify that it remains $0.00.

#### Test $50 price

1. Add product(s) to the cart with a value of $50.
2. Open the mini cart page in the frontend.
3. See that the price shows $50.00.
4. Hover over the price and verify that it remains $50.00.

### User Facing Testing

* [x] Same as above
* [ ] See steps below

### Changelog

> Ensure that expected numbers of decimals are visible, if wanted my the merchant, even if the price is $0 or has no decimals, such as $50.

### DIFF
--- a/packages/prices/utils/price.ts
+++ b/packages/prices/utils/price.ts
@@ -137,6 +137,25 @@ const splitDecimal = (
 	};
 };
 
+const applyDecimal = (
+	afterDecimal: string,
+	decimalSeparator: string,
+	minorUnit: number
+): string => {
+	if ( afterDecimal ) {
+		return `${ decimalSeparator }${ afterDecimal.padEnd(
+			minorUnit,
+			'0'
+		) }`;
+	}
+
+	if ( minorUnit > 0 ) {
+		return `${ decimalSeparator }${ '0'.repeat( minorUnit ) }`;
+	}
+
+	return '';
+};
+
 /**
  * Format a price, provided using the smallest unit of the currency, as a
  * decimal complete with currency symbols using current store settings.
@@ -176,9 +195,11 @@ export const formatPrice = (
 	const formattedValue = `${ prefix }${ applyThousandSeparator(
 		beforeDecimal,
 		thousandSeparator
-	) }${
-		afterDecimal ? `${ decimalSeparator }${ afterDecimal }` : ''
-	}${ suffix }`;
+	) }${ applyDecimal(
+		afterDecimal,
+		decimalSeparator,
+		minorUnit
+	) }${ suffix }`;
 
 	// This uses a textarea to magically decode HTML currency symbols.
 	const txt = document.createElement( 'textarea' );
--- a/packages/prices/utils/test/price.js
+++ b/packages/prices/utils/test/price.js
@@ -3,48 +3,46 @@
  */
 import { formatPrice, getCurrency } from '../price';
 
-describe( 'formatPrice', () => {
+describe( 'The function formatPrice()', () => {
 	test.each`
-		value           | prefix   | suffix   | expected
-		${ 1000 }       | ${ '€' } | ${ '' }  | ${ '€10' }
-		${ 1000 }       | ${ '' }  | ${ '€' } | ${ '10€' }
-		${ 1000 }       | ${ '' }  | ${ '$' } | ${ '10$' }
-		${ '1000' }     | ${ '€' } | ${ '' }  | ${ '€10' }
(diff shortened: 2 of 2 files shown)


packages/prices/utils/test/price.js
```diff
@@ -3,48 +3,45 @@
  */
 import { formatPrice, getCurrency } from '../price';
 
-describe( 'formatPrice', () => {
+describe( 'The function formatPrice()', () => {
```
mikejolley, line 6: 👍🏻

packages/prices/utils/price.ts
```diff
@@ -177,7 +193,9 @@ export const formatPrice = (
 		beforeDecimal,
 		thousandSeparator
 	) }${
-		afterDecimal ? `${ decimalSeparator }${ afterDecimal }` : ''
+		afterDecimal || minorUnit > 0
```
mikejolley: Since we have a value for `afterDecimal`, and we are passing `minorUnit`, how about we remove the conditional here and just format it in the other util you wrote? We don't need to check twice.

packages/prices/utils/price.ts
```diff
@@ -137,6 +137,22 @@ const splitDecimal = (
 	};
 };
 
+const applyDecimal = (
+	afterDecimal: string,
+	decimalSeparator: string,
+	minorUnit: number
+): string => {
+	if ( afterDecimal ) {
```
mikejolley, line 145: Do we need to pad decimals here regardless?

Create a test case or 2dp currency, with a given value of `10.2`. It should result in `10.20`

packages/prices/utils/price.ts
```diff
@@ -137,6 +137,22 @@ const splitDecimal = (
...
+		return `${ decimalSeparator }${ afterDecimal }`;
+	}
+
+	if ( minorUnit > 0 ) {
+		return `${ decimalSeparator }${ '0'.repeat( minorUnit ) }`;
```
mikejolley, line 153: Concerning my above note, we could use `minorUnit` minus `afterDecimal` strlen.

mikejolley (changes requested):

packages/prices/utils/price.ts
nielslange: Great suggestion! I've updated the code accordingly.

packages/prices/utils/test/price.js
nielslange: 😊

packages/prices/utils/price.ts
nielslange: After adding the additional test case, I realized that padding the decimals was needed. I add this accordingly.

packages/prices/utils/price.ts
nielslange: I'm not sure if that's applicable here, as this part can only be reached if `afterDecimal` is empty. Therefore, `afterDecimal.strlen` should always result in `0` in this part. That's why I'm repeating the zeros based on the `Number of decimals` selected by the merchant. 

Was I missing something in this implementation?

mikejolley: Ok should be good. The minor units we're converting from mitigate this.

mikejolley (approved):

nielslange:
Thanks for your review, @mikejolley! 🙌

PR #5220, opened 2021-11-23
Fix Mini Cart block when Gutenberg is not enabled

Aljullu:
Fixes a fatal in WP 5.8 (or lower) with Gutenberg disabled.

It also wraps Mini Cart contents with `wp_kses_post()` so contents are sanitized.

### Manual Testing

1. Make sure you are testing in a WP 5.8.x site with Gutenberg disabled.
2. Add the Mini Cart block to a post or page.
3. Visit it in the frontend.
4. Click on the Mini Cart button and verify the drawer opens and it contains the default template part.

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -308,8 +308,7 @@ class MiniCart extends AbstractBlock {
 
 		$template_part_contents = '';
 		if ( function_exists( 'gutenberg_get_block_template' ) ) {
-			$part          = 'mini-cart';
-			$template_part = gutenberg_get_block_template( get_stylesheet() . '//' . $part, 'wp_template_part' );
+			$template_part = gutenberg_get_block_template( get_stylesheet() . '//mini-cart', 'wp_template_part' );
 			if ( $template_part && ! empty( $template_part->content ) ) {
 				$template_part_contents = do_blocks( $template_part->content );
 			}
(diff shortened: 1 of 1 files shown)


src/BlockTypes/MiniCart.php
```diff
@@ -306,11 +306,19 @@ protected function get_markup( $attributes ) {
...
 		}
 
-		$part                   = 'mini-cart';
-		$template_part          = gutenberg_get_block_template( get_stylesheet() . '//' . $part, 'wp_template_part' );
 		$template_part_contents = '';
-		if ( $template_part && ! empty( $template_part->content ) ) {
-			$template_part_contents = do_blocks( $template_part->content );
+		if ( function_exists( 'gutenberg_get_block_template' ) ) {
+			$part          = 'mini-cart';
+			$template_part = gutenberg_get_block_template( get_stylesheet() . '//' . $part, 'wp_template_part' );
```
dinhtungdu: Because we only use `$part` one time, I think we should use the string directly.
```suggestion
			$template_part = gutenberg_get_block_template( get_stylesheet() . '//mini-cart', 'wp_template_part' );
```

dinhtungdu (approved):
This is looking good to me. I left an optional comment about readability, otherwise, this is ready to be merged.

src/BlockTypes/MiniCart.php
Aljullu: Makes sense!

PR #5223, opened 2021-11-23
Update WordPress Monorepo packages with React 17x support

### LINKED ISSUE #4533: Update WordPress Packages to Latest versions that support React 17x
During cooldown/porter week I looked into why some [dependency updates were failing tests](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4527). 

It looks like some packages (packages from WordPress and types) are shifting to React 17x compatibility, but some of our existing packages only support React 16x. When there is a mix of React 17x and 16x, the JS unit tests fail because multiple versions of React are present, which breaks rules of hooks checks.

Running `npm ls react` lists what dependencies the packages have. The ones which causes issues were:

- `@automattic/data-stores`
- `@woocommerce/components`

We should have those packages widen their peer dependencies to include React 17x. This will allow us to bump all `@wordpress/` package versions (I think).

The workaround I've employed for now is to bump those packages to the latest version supporting React 16x, then checking with `npm ls react` to ensure only 16x is listed. This can be found in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4532

mikejolley:
This PR exists to update all WordPress dependencies to the latest and greatest versions, and since they require React 17x, fixes conflicts with other dependencies we do not directly manage.

Fixes #4533
Closes #4904

Work involved:

- Updated all `@wordpress/` dependencies, including those raised in #4904 which have been Blocked by the React 17x requirement
- Audited all dependencies for React support using ` npm ls react`
- Moved `mapped-types` back from `@automattic/data-stores` since having that as a dependency for 3 functions seems too much. `@automattic/data-stores` has not been updated for 8 months and does not support React 17x. There are some other versions of `@automattic/data-stores` within the org with React 17x support but these are not deployed to NPM. It's unclear where the single source of truth is, so to unblock this work I've just moved the types across.
- Moved `SearchListControl` back into our `editor-components`, from `@woocommerce/components` to remove that dependency. `@woocommerce/components` is another package that does not support React 17x (I will log this), but again, to unblock us, and prevent future conflicts, I spoke to @senadir and moved this back into our repo. I don't think we can just widen the React version requirements in `@woocommerce/components` to fix this because there are other dependencies (from the WordPress monorepo) that also need bumping, similar to the work in this PR.
- `SearchListControl` uses now-deprecated HOCs such as `withState`, so I converted this to Typescript and functional components with hooks to resolve. This is again something I could contribute back to `@woocommerce/components`, but until the dependency issues are resolved and monorepos updated, I think this will be problematic. It's not clear to me who is managing this type of work (for us it's porter and we seem more active with dependency updates).

### Testing

Developer Smoke Testing:

1. Npm install successfully
2. `npm ls react` should only list versions of React 17x, not 16x
3. Successful Build
4. Successful Tests
5. Go to the editor and smoke test block insertion. Check for console errors.
6. Go to frontend and smoke test block rendering. Check for console errors.

`SearchListControl`:

1. Insert a product block - test Products by Attribute and Handpicked Products
2. Confirm you can toggle and check items in the search list
3. Confirm you can remove selected items using the tag inputs above the list
4. Confirm you can search for an item and the list is filtered accordingly
5. Confirm you can save and results persist

Aside from smoke testing, there are no user facing changes to test here.

### Performance Impact

Should not be impacted. With fewer dependencies, if anything,  builds and install should be slightly faster. App performance should not be impacted.


senadir:
I tested it it builds and works fine, will head to code, but it seems we lost some weight here and there, but gained some weight in some scripts:

Filename | Size | Change
-- | -- | --
build/checkout-frontend.js | 76.7 kB | +25.7 kB (+50%)
build/cart-frontend.js | 74.5 kB | +25.6 kB (+52%)
build/single-product-frontend.js | 32.6 kB | +7.13 kB (+28%) 
build/single-product.js | 10.4 kB | +1.94 kB (+23%)
build/blocks-checkout.js | 28.1 kB | +7.31 kB (+35%)
build/all-products-frontend.js | 29.7 kB | +7.59 kB (+34%)

Can we investigate this?

mikejolley:
@senadir did you have more thoughts on the package increase? Are we accepting it?

senadir:
Last time I checked I thought we wanted to downgrade to `@wordpress/components` 14.x.x unless something changed?
I prefer we resolve or leave the `@wordpress/components` part until that issue is resolved upstream? I'm still planning to ping the `@wordpress/components` about this but had a lot of other stuff last week to actually get to it.

mikejolley:
@senadir I tried but realised it doesn't include one of the patches I made for the country select. If we do downgrade, I need to create another repo-level patch file to fix it. The newer versions in this PR include the fix, and support react 17x.

senadir (approved):
All gains!

It seems we have a lot of outdated dependencies, I will make sure to reach out to all of them to see this sorted, ideally, most of what we did here is temporary to unblock us.

mikejolley:
Looks like this is fixed now. Package size is back down with components 14x, and React 17x is being used throughout. Tests are passing. The light is green, the trap is clean.

PR #5234, opened 2021-11-24
Remove obsolete files and folders from deploy zip

### LINKED ISSUE #4928: Remove extra files/folders in built zip
As of WooCommerce Blocks 6.1.0, those files are in the final zip:

- [x] patches (only needed when building)
- [x] packages (is a source folder, not needed on final build)
- [ ] inbox-notifier (seems like an empty folder, cc @opr)
- [x] global.d.ts (dev file)
- [x] tsconfig.base.json (dev file)

We should skip them on our build zips.

nielslange:
Fixes #4928

### Notes

The current deploy zip, that is generated for testing purposes, contains various build files and folders such as `patches/` and `global.d.ts`. This PR aims to generate a deploy zip file that only contains files and folders that are needed for production. It shall not contain any build files and folders.

In addition to the files and folders mentioned in #4928, I also excluded the `reports/` folder which contains the e2e report file and is therefore not relevant for production/testing.

### Testing

1. Check out this PR.
2. Run `npm run package-plugin:deploy`.
3. Ensure that the created deploy zip does not contain any build files and folders.
4. Smoke test the deploy zip to ensure that all functionality still works as expected.

### DIFF
--- a/bin/copy-plugin-files.sh
+++ b/bin/copy-plugin-files.sh
@@ -23,8 +23,12 @@ rsync ./ "$2"/ --recursive --delete --delete-excluded \
 	--exclude=blocks.ini \
 	--exclude=docker-compose.yml \
 	--exclude=tsconfig.json \
+	--exclude=tsconfig.base.json \
 	--exclude=woocommerce-gutenberg-products-block.zip \
 	--exclude="zip-file/" \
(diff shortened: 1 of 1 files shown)


opr (approved):
Works fine for me when I smoke tested it, thanks Niels! Let's 🚢 it.

PR #5238, opened 2021-11-24
 Convert renderFrontend to TypeScript 

Aljullu:
Refactors `render-frontend.js` to TypeScript.

### Manual Testing

Any regressions should be found by automated tested, but just in case:

1. Add the Cart or Checkout blocks to a post or page. 
2. Go to the frontend and verify they render correctly.


assets/js/base/utils/render-frontend.tsx
```diff
@@ -0,0 +1,206 @@
...
+import BlockErrorBoundary from '@woocommerce/base-components/block-error-boundary';
+
+// Some blocks take care of rendering their inner blocks automatically. For
+// example, the empty cart. In those cases, we don't want to trigger the render
+// function of inner components on load. Instead, the wrapper block can trigger
+// the event `wc-blocks_render_blocks_frontend` to render its inner blocks.
+const selectorsToSkipOnLoad = [ '.wp-block-woocommerce-cart' ];
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type blockType = ( props: any ) => JSX.Element;
```
Aljullu: I couldn't find any way to replace this `any` with something more specific. Each block passed as a param to this function can have different required props and I couldn't find any way to let TypeScript ignore them. :confused:

gigitux: I tried with `Record<string, unknown>` and  I didn't see any error. I think that it can be a good fit, because we are sure that we will pass an object

gigitux: I think that for the name type it is better if we use PascalCase. So in this case `BlockType`. What do you think?

gigitux (changes requested):
I think that we should adopt the PascalCase for types and interface. What do you think?

There is just a suggestion, for the rest LGTM 👍

assets/js/base/utils/render-frontend.tsx
Aljullu: > I tried with `Record<string, unknown>` and  I didn't see any error. I think that it can be a good fit, because we are sure that we will pass an object

The error doesn't appear right there, but in consumers of `renderFrontend` and `renderBlock` (ie, [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/95ac903535b45674058148dacd255edc302fba39/assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx#L30)). Let me know if you can reproduce.

> I think that for the name type it is better if we use PascalCase. So in this case `BlockType`. What do you think?

Oh, right, I have been doing it wrong for a long time. :man_facepalming: I updated it in this file and a couple more in 3029677cfba7ac725a27111b2cfae67a81b3fc4d.

gigitux:
@Aljullu I pushed the changes for introducing generics. There is still room for improvement, but for now, I believe it is not necessary.

My concern is more to see how these changes work on the "field" (if the inheritance works well), It may be that there is something else to improve/fix. I will await your feedback!

Aljullu (commented):
Thanks for helping me with this tricky TypeScript issue and finding a solution, @gigitux! It seems to be working fine. I can't approve the PR because I'm the author, but LGTM. :ok_hand:

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
```diff
@@ -28,7 +28,7 @@ const renderMiniCartFrontend = () => {
 	renderFrontend( {
 		selector: '.wc-block-mini-cart',
 		Block: MiniCartBlock,
-		getProps: ( el: HTMLElement ) => {
+		getProps: ( el ) => {
```
tjcafferkey, line 31: Any reason you've removed `HTMLElement` type from `el`?

assets/js/base/utils/render-frontend.tsx
```diff
@@ -0,0 +1,256 @@
...
+>( {
+	Block,
+	getProps,
+	getErrorBoundaryProps,
+	selector,
+	wrappers,
+}: RenderBlockOutsideWrappersParams< TProps, TAttributes > ): void => {
+	const containers = document.body.querySelectorAll( selector );
+	// Filter out blocks inside the wrappers.
+	if ( wrappers && wrappers.length > 0 ) {
```
tjcafferkey, line 159: Can we simplify to `wrappers?.length > 0`?

assets/js/base/utils/render-frontend.tsx
```diff
@@ -0,0 +1,256 @@
...
+ * External dependencies
+ */
+import { render, Suspense } from '@wordpress/element';
+import BlockErrorBoundary from '@woocommerce/base-components/block-error-boundary';
+
+// Some blocks take care of rendering their inner blocks automatically. For
+// example, the empty cart. In those cases, we don't want to trigger the render
+// function of inner components on load. Instead, the wrapper block can trigger
+// the event `wc-blocks_render_blocks_frontend` to render its inner blocks.
+const selectorsToSkipOnLoad = [ '.wp-block-woocommerce-cart' ];
```
tjcafferkey, line 11: Can we type this `string[]`

assets/js/blocks/cart-checkout/mini-cart/block.tsx
```diff
@@ -29,13 +29,13 @@ import QuantityBadge from './quantity-badge';
 import MiniCartContentsBlock from '../mini-cart-contents/block';
 import './style.scss';
 
-interface Props {
+type Props = {
```
tjcafferkey: Just curious, what is the reason for the change?

tjcafferkey (commented):
Just some minor comments & questions

tjcafferkey (approved):
Just left a few comments but nothing blocking.

assets/js/blocks/cart-checkout/mini-cart/component-frontend.tsx
gigitux: Yes! Now with generics, Typescript can correctly understand what type it is.
It is better not to specify types when possible

assets/js/base/utils/render-frontend.tsx
gigitux: Typescript doesn't like it :'( 
image

assets/js/base/utils/render-frontend.tsx
gigitux: Same here. Typescript can automatically figure out what type it is

assets/js/blocks/cart-checkout/mini-cart/block.tsx
gigitux: Not really. I am used to working with `type` and while experimenting, I made this change. Restored it

gigitux:
> Just left a few comments but nothing blocking.

I replied to the comments because they were changes I made

Aljullu:
@tjcafferkey @gigitux just to confirm, can we already merge this PR?

gigitux:
> @tjcafferkey @gigitux just to confirm, can we already merge this PR?

For me, we can merge the PR

gigitux (approved):

Aljullu:
Thanks! Merged, given that it already had two approvals.

PR #5241, opened 2021-11-24
Convert components to Typescript

alexflorisca:
Converted the following components (and their respective sub components) to Typescript:

- [x] ProductPrice
- [x] OrderSummary
- [x] FormStep
- [x] ReturnToCartButton

Some small changes to types in 
- [x] FormattedMonetaryAmount

Removed  redundant `propTypes` in 
- [x] AddressForm (see [this](https://a8c.slack.com/archives/C8X6Q7XQU/p1637758785345400) and [this](https://a8c.slack.com/archives/C8X6Q7XQU/p1637764581366600) slack conversation)

I'm new to TypeScript so I'm taking this as a learning opportunity. Please give me as much feedback as you can, I'm sure I've screwed up a thing or two :). More specifically, I'm interested to know how other people decide if the props for a component are required or not. I went for a mix of what made the most sense and the current implementation. As in, if a component has 10 props, and it's called with only 3 of them somewhere in the code at a minimum, then those 3 are required.

### DIFF
--- a/assets/js/base/components/cart-checkout/form-step/index.tsx
+++ b/assets/js/base/components/cart-checkout/form-step/index.tsx
@@ -37,7 +37,7 @@ interface FormStepProps {
 	title?: string;
 	legend?: string;
 	description?: string;
-	children?: JSX.Element | JSX.Element[] | string | string[];
+	children?: React.ReactNode;
 	disabled?: boolean;
 	showStepNumber?: boolean;
 	stepHeadingContent?: () => JSX.Element | undefined;
--- a/assets/js/base/components/cart-checkout/form-step/index.tsx
+++ b/assets/js/base/components/cart-checkout/form-step/index.tsx
@@ -10,7 +10,12 @@ import Title from '@woocommerce/base-components/title';
  */
 import './style.scss';
 
-const StepHeading = ( { title, stepHeadingContent } ) => (
+interface StepHeadingProps {
+	title: string;
+	stepHeadingContent?: JSX.Element;
+}
+
+const StepHeading = ( { title, stepHeadingContent }: StepHeadingProps ) => (
 	<div className="wc-block-components-checkout-step__heading">
 		<Title
 			aria-hidden="true"
@@ -27,6 +32,18 @@ const StepHeading = ( { title, stepHeadingContent } ) => (
 	</div>
 );
 
+interface FormStepProps {
+	id: string;
+	className: string;
+	title?: string;
+	legend?: string;
+	description: string;
+	children: JSX.Element | JSX.Element[] | string | string[];
+	disabled: boolean;
+	showStepNumber: boolean;
+	stepHeadingContent: () => JSX.Element | undefined;
+}
+
 const FormStep = ( {
 	id,
 	className,
@@ -36,8 +53,8 @@ const FormStep = ( {
 	children,
 	disabled = false,
 	showStepNumber = true,
-	stepHeadingContent = () => {},
-} ) => {
+	stepHeadingContent = () => undefined,
+}: FormStepProps ): JSX.Element => {
 	// If the form step doesn't have a legend or title, render a <div> instead
 	// of a <fieldset>.
 	const Element = legend || title ? 'fieldset' : 'div';
--- a/assets/js/base/components/cart-checkout/form-step/index.tsx
+++ b/assets/js/base/components/cart-checkout/form-step/index.tsx
@@ -2,7 +2,6 @@
  * External dependencies
  */
 import classnames from 'classnames';
-import PropTypes from 'prop-types';
 import Title from '@woocommerce/base-components/title';
 
 /**
@@ -33,15 +32,15 @@ const StepHeading = ( { title, stepHeadingContent }: StepHeadingProps ) => (
 );
 
 interface FormStepProps {
-	id: string;
-	className: string;
+	id?: string;
+	className?: string;
 	title?: string;
 	legend?: string;
-	description: string;
-	children: JSX.Element | JSX.Element[] | string | string[];
-	disabled: boolean;
-	showStepNumber: boolean;
-	stepHeadingContent: () => JSX.Element | undefined;
+	description?: string;
+	children?: JSX.Element | JSX.Element[] | string | string[];
+	disabled?: boolean;
+	showStepNumber?: boolean;
+	stepHeadingContent?: () => JSX.Element | undefined;
 }
 
 const FormStep = ( {
@@ -95,18 +94,6 @@ const FormStep = ( {
 			</div>
 		</Element>
 	);
-};
-
-FormStep.propTypes = {
-	id: PropTypes.string,
-	className: PropTypes.string,
-	title: PropTypes.string,
-	description: PropTypes.string,
-	children: PropTypes.node,
-	showStepNumber: PropTypes.bool,
-	stepHeadingContent: PropTypes.func,
-	disabled: PropTypes.bool,
-	legend: PropTypes.string,
 };
 
 export default FormStep;
--- a/assets/js/base/components/cart-checkout/order-summary/index.tsx
+++ b/assets/js/base/components/cart-checkout/order-summary/index.tsx
@@ -2,17 +2,23 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import PropTypes from 'prop-types';
 import { useContainerWidthContext } from '@woocommerce/base-context';
 import { Panel } from '@woocommerce/blocks-checkout';
 
 /**
  * Internal dependencies
  */
-import OrderSummaryItem from './order-summary-item.js';
+import OrderSummaryItem from './order-summary-item';
 import './style.scss';
+import { CartItem } from '../../../../types/type-defs/cart';
 
-const OrderSummary = ( { cartItems = [] } ) => {
+interface OrderSummaryProps {
+	cartItems: CartItem[];
+}
+
+const OrderSummary = ( {
+	cartItems = [],
+}: OrderSummaryProps ): null | JSX.Element => {
 	const { isLarge, hasContainerWidth } = useContainerWidthContext();
 
 	if ( ! hasContainerWidth ) {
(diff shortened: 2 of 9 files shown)


assets/js/base/components/cart-checkout/form-step/index.tsx
```diff
@@ -27,6 +31,18 @@ const StepHeading = ( { title, stepHeadingContent } ) => (
 	</div>
 );
 
+interface FormStepProps {
+	id?: string;
+	className?: string;
+	title?: string;
+	legend?: string;
+	description?: string;
+	children?: JSX.Element | JSX.Element[] | string | string[];
```
opr: Does using `React.ReactNode` here work instead?

assets/js/base/components/cart-checkout/order-summary/index.tsx
```diff
@@ -2,17 +2,23 @@
...
 import { useContainerWidthContext } from '@woocommerce/base-context';
 import { Panel } from '@woocommerce/blocks-checkout';
 
 /**
  * Internal dependencies
  */
-import OrderSummaryItem from './order-summary-item.js';
+import OrderSummaryItem from './order-summary-item';
 import './style.scss';
+import { CartItem } from '../../../../types/type-defs/cart';
```
opr: You can use `import type` if you're only importing a type, also we have `@woocommerce/types` aliased for this

assets/js/base/components/cart-checkout/order-summary/order-summary-item.tsx
```diff
@@ -27,10 +26,16 @@ import ProductBackorderBadge from '../product-backorder-badge';
 import ProductImage from '../product-image';
 import ProductLowStockBadge from '../product-low-stock-badge';
 import ProductMetadata from '../product-metadata';
+import { CartItem } from '../../../../types/type-defs/cart';
```
opr: Same as above

assets/js/base/components/cart-checkout/order-summary/order-summary-item.tsx
```diff
@@ -27,10 +26,16 @@ import ProductBackorderBadge from '../product-backorder-badge';
...
 
-const productPriceValidation = ( value ) => mustContain( value, '<price/>' );
+const productPriceValidation = ( value: string ): true | Error =>
```
opr: This actually seems like `mustContain` was typed wrong, it never _returns_ an error, it throws.

The definition for `mustContain` should be:

```typescript
export const mustContain = (
	value: string,
	requiredValue: string
): true | never
```

and so the return type of `productPriceValidation` should be changed to match. `never` means that the function won't return (because it throws).

assets/js/base/components/cart-checkout/order-summary/order-summary-item.tsx
```diff
@@ -72,13 +77,13 @@ const OrderSummaryItem = ( { cartItem } ) => {
 
 	const regularPriceSingle = Dinero( {
 		amount: parseInt( prices.raw_prices.regular_price, 10 ),
-		precision: parseInt( prices.raw_prices.precision, 10 ),
+		precision: prices.raw_prices.precision,
```
opr: Are you certain this will always be an int?

assets/js/base/components/cart-checkout/return-to-cart-button/index.tsx
```diff
@@ -10,7 +10,7 @@ import { Icon, arrowBack } from '@woocommerce/icons';
  */
 import './style.scss';
 
-const ReturnToCartButton = ( { link } ) => {
+const ReturnToCartButton = ( { link }: { link?: string } ): JSX.Element => {
```
opr: Annoying I know, but I think it's preferred to create an interface for this. `interface ReturnToCartButtonProps` - I think it has a minor performance benefit, and also lets us refer to something by name, if we have to in future.

assets/js/base/components/formatted-monetary-amount/index.tsx
```diff
@@ -6,12 +6,12 @@ import NumberFormat, {
 	NumberFormatProps,
 } from 'react-number-format';
 import classNames from 'classnames';
-import type { Currency } from '@woocommerce/price-format';
 import type { ReactElement } from 'react';
 
 /**
  * Internal dependencies
  */
+import type { Currency } from '../../../types/type-defs/currency';
```
opr: You can import from `@woocommerce/types` here :)

assets/js/base/components/formatted-monetary-amount/index.tsx
```diff
@@ -20,6 +20,8 @@ interface FormattedMonetaryAmountProps {
 	value: number | string; // Value of money amount.
 	currency: Currency | Record< string, never >; // Currency configuration object.
 	onValueChange?: ( unit: number ) => void; // Function to call when value changes.
+	style?: Record< string, unknown >;
```
opr: You can use `React.CSSProperties` as a type here

assets/js/base/components/product-price/index.tsx
```diff
@@ -4,22 +4,30 @@
 import { __, sprintf } from '@wordpress/i18n';
 import FormattedMonetaryAmount from '@woocommerce/base-components/formatted-monetary-amount';
 import classNames from 'classnames';
-import PropTypes from 'prop-types';
 import { formatPrice } from '@woocommerce/price-format';
 import { createInterpolateElement } from '@wordpress/element';
+import { Currency } from '@woocommerce/type-defs/currency';
```
opr: This works, but so does `@woocommerce/types`

assets/js/base/components/product-price/index.tsx
```diff
@@ -4,22 +4,30 @@
...
  * Internal dependencies
  */
 import './style.scss';
 
+interface PriceRangeProps {
+	currency: Currency | Record< string, never >; // Currency configuration object;
+	maxPrice: string | number;
+	minPrice: string | number;
+	priceClassName?: string;
+	priceStyle?: Record< string, string >;
```
opr: Can we use `React.CSSProperties` again?

assets/js/base/components/product-price/index.tsx
```diff
@@ -161,7 +194,7 @@ const ProductPrice = ( {
 				regularPriceStyle={ regularPriceStyle }
 			/>
 		);
-	} else if ( minPrice !== null && maxPrice !== null ) {
+	} else if ( minPrice && maxPrice ) {
```
opr: This will evaluate to true if either of these values are `0` - is this going to cause problems? Can you test it and see what happens?

opr (changes requested):
Great work Alex! This is looking pretty solid, but I've left a few questions and pointers that we should address/check before merging.

assets/js/base/components/cart-checkout/form-step/index.tsx
alexflorisca: It does indeed 👍

assets/js/base/components/cart-checkout/order-summary/order-summary-item.tsx
alexflorisca: Yes, this is typed for a `number`[ here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/types/type-defs/cart.ts#L97) so `parseInt()` actually complains that it's receiving a `number` instead of `string`. I thought safe to remove as it should get caught out by type error if anything other than a number is passed to it

assets/js/base/components/product-price/index.tsx
alexflorisca: A number `0` evaluates to false so this expression would be false in that case. ~~The intent here is that if either the minPrice or maxPrice is either an empty string `''` or number `0`, this will evaluate to false and not run~~. I've just made that up. You should be able to have a product be £0.00 and still display the price. I've updated the logic here to check if `minPrice` or `maxPrice` are undefined here instead

assets/js/base/components/cart-checkout/order-summary/order-summary-item.tsx
opr: Sorry I actually meant: are you sure that _at runtime_ it will always be an `number`, TypeScript only works at compile time. Do we need a check to cater for the chance that it might be a string or is it fine?

assets/js/base/components/product-price/index.tsx
opr: Sorry, I got my logic mixed up! But thanks for understanding what I meant and fixing it anyway :)

assets/js/base/components/cart-checkout/order-summary/order-summary-item.tsx
alexflorisca: You're right, I've changed this now to check for a string :)

assets/js/base/components/cart-checkout/order-summary/order-summary-item.tsx
```diff
@@ -72,13 +77,15 @@ const OrderSummaryItem = ( { cartItem } ) => {
...
+		precision: isString( prices.raw_prices.precision )
+			? parseInt( prices.raw_prices.precision, 10 )
+			: prices.raw_prices.precision,
 	} )
 		.convertPrecision( priceCurrency.minorUnit )
 		.getAmount();
 	const priceSingle = Dinero( {
 		amount: parseInt( prices.raw_prices.price, 10 ),
-		precision: parseInt( prices.raw_prices.precision, 10 ),
+		precision: prices.raw_prices.precision,
```
opr: You will need to do the `isString` check here too

opr (changes requested):
Just one last thing that needs changing, then I can approve 🙌🏼

assets/js/base/components/cart-checkout/order-summary/order-summary-item.tsx
alexflorisca: Oops, done :)

opr (approved):
LGTM now, great work 👏🏼 
Let's 🚢 it!

PR #5247, opened 2021-11-25
Remove extra padding for cart sidebar to fix visual regression bug

### LINKED ISSUE #5152: Cart block: Visual bug for WooCommerce Points and Rewards
## Describe the bug
The form for WooCommerce Points and Rewards  in the Cart block does not look as before.

## To reproduce

1. Install WooCommerce Points and Rewards, go to WooCommerce > Points and Rewards and add points to your user.
2. Go to WooCommerce > Points and Rewards > Settings and enable "Allow partial redemption"
4. Notice that it does not look ok

## Expected behaviour
See screenshots for difference. 

## Screenshots
**Before**
Screenshot 2021-11-16 at 11 53 51
**Now**
Screenshot 2021-11-16 at 12 07 05

alexflorisca:
When upgrading the cart, there was an extra left and right margin added to the Cart page sidebar. I'm not sure if this was intentional, however it causes a [visual regression bug with the points and rewards plugin](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5152) where the input box and the button don't line up properly (see issue for screenshots). 

This fixes #5152 

### Testing

1. Install WooCommerce Points and Rewards, go to WooCommerce > Points and Rewards and add points to your user.
2. Go to WooCommerce > Points and Rewards > Settings and enable "Allow partial redemption"
4. The text input should be inline with the "Apply Discount" button

### Changelog

> Fixed a visual bug (#5152) with the points and rewards plugin

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/style.scss
@@ -238,13 +238,6 @@ table.wc-block-cart-items {
 		}
 	}
 
-	.wc-block-cart__sidebar {
-		& > div:not(.wc-block-components-totals-wrapper) {
-			margin-left: $gap;
-			margin-right: $gap;
-		}
-	}
-
 	.wc-block-components-radio-control__input {
 		left: 0;
 	}


senadir:
Took me a bit of hunting, but I limited that change to this PR https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4974 so we need to make sure it won't regress it as well.

senadir (commented):
I tested this and failed to see what's trying to fix. A before and after screenshot would help.

alexflorisca:
Hi @senadir, thought I'd reply here for posterity and with some more useful details following our [slack convo](p1638450416174100-slack-C8X6Q7XQU). Feel free to ping me on slack in that thread in response!

Screenshots of the bug can be seen [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5152).

This happens mainly when the cart page has a sidebar.

So it looks like when the cart block sidebar is a certain width (236px), the browser decides there isn't enough real estate and it drops the "Apply Discount" button on the second line. The styles that control the input/button come from the Woocommerce Points And Rewards plugin so we have not control over making the button smaller for example. 

My initial approach was to remove the extra margin added in the [PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4974) you mentioned. I realise this causes changes to the design and may not be the best solution

As an alternative, we could reduce the margin to `8px` each side, or set a min-width of `236px` on the sidebar and that would make sure those buttons are on the same line.

However this seems a bit arbitrary to me and i'm aware that we are trying to fix the layout for a third party plugin at a specific width. It's difficult to test the impact this might have on other plugins that add content to the sidebar and there will likely always be a case of misalignment if we change the width of the sidebar. 

I guess reverting the extra `16px` margin would ensure any third party plugins that were designed before that PR was merged would still display correctly. Otherwise if we want to keep the current design with the extra margin, we could reach out to the developers of the Woocommerce points and rewards plugin, point out the issue and get them to update their plugin. 

I'm looking for a bit of guidance here as I'm not sure what the best approach is.

opr (approved):
LGTM! :)

PR #5260, opened 2021-11-26
Don't run enqueue_assets on REST requests

### LINKED ISSUE #5173: Avoid running enqueue_assets on REST requests
Initially discussed in this comment: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5051#pullrequestreview-797624158

Every time a post is saved, the `enqueue_assets()` method is called for each block contained in the post. It's not needed to call those methods given that saving a post is a REST request, so there is nothing to enqueue.

This issue is about investigating if it would be possible to not call `enqueue_assets()` when making a REST request. Currently, we are already doing this when `is_admin()` is true:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e76f5694855329e1082893db049563acce4d4180/src/BlockTypes/AbstractBlock.php#L84-L86

## To reproduce

1. Go to the post editor and add a Cart block.
2. Edit [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e76f5694855329e1082893db049563acce4d4180/src/BlockTypes/Cart.php#L57) of the Cart block as follows:
```diff
	protected function enqueue_assets( array $attributes ) {
+		die( 'Cart block enqueue_assets called' );
```
3. Try saving the post and notice it fails because the `enqueue_assets()` function from Cart was called.

Aljullu:
Fixes #5173.

### Manual Testing

Smoke testing:

1. Copy the contents of this [gist](https://gist.github.com/Aljullu/d9c76ed510ae6973bd41dcc4184e8eb2), which includes all WC Blocks, and paste it into a post or page (via <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>V</kbd>).
2. Save it and verify there is no error.
3. Visit the post or page in the frontend and verify blocks are rendered correctly, except:
3.1 Product Search: you will need to remove it and add it again because it has the URL of the site hard-coded.
3.2 Filter Products by Attribute: there is an open issue for it: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5245.

### Changelog

> Avoid enqueueing block assets and data when not needed in REST requests (ie, when saving a post or page).

### DIFF
--- a/src/BlockTypes/AbstractBlock.php
+++ b/src/BlockTypes/AbstractBlock.php
@@ -81,7 +81,7 @@ abstract class AbstractBlock {
 	 */
 	public function render_callback( $attributes = [], $content = '' ) {
 		$render_callback_attributes = $this->parse_render_callback_attributes( $attributes );
-		if ( ! is_admin() ) {
+		if ( ! is_admin() && ! WC()->is_rest_api_request() ) {
 			$this->enqueue_assets( $render_callback_attributes );
 		}
 		return $this->render( $render_callback_attributes, $content );


dinhtungdu (approved):
Tested and this is working great on my end.

PR #5261, opened 2021-11-26
Create dev packages with a blocks.ini file

### LINKED ISSUE #5126: npm run package-plugin:dev should copy blocks.ini
## To reproduce
1. Run `npm run package-plugin:dev` in the WC Blocks directory.
2. Install the resulting zip in an external site without WC Blocks.
3. Notice some of the gated features are not available (at the time of writing this, the Mini Cart block isn't).

## Expected behavior
I would expect all gated features to be available when running `package-plugin:dev`.

## Describe the bug
This happens because `package-plugin:dev` builds the files as a dev environment, but PHP files still think they are in a production evironment. To solve this, we would need to include a `blocks.ini` file with these contents:

```
woocommerce_blocks_phase = 3
woocommerce_blocks_env = development
```

## Work-around

For now, the only work-around is to generate the zip, unzip it, copy the blocks.ini file inside and compress it again. But ideally, this shouldn't be necessary.

Aljullu:
Fixes #5126.

### Manual Testing

1. Run `npm run package-plugin:dev` in the WC Blocks directory.
2. Verify the resulting zip contains a `blocks.ini` file.
3. Install the resulting zip in an external site without WC Blocks.
4. Notice all gated features are available (ie, the Mini Cart block).

### DIFF
--- a/bin/build-plugin-zip.sh
+++ b/bin/build-plugin-zip.sh
@@ -129,6 +129,10 @@ mkdir zip-file
 mkdir zip-file/build
 sh "$SOURCE_PATH/bin/copy-plugin-files.sh" "$SOURCE_PATH" "$SOURCE_PATH/zip-file"
 cd "$(pwd)/zip-file"
+if [ $TYPE = 'DEV' ]; then
+	touch blocks.ini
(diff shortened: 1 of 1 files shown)


gigitux (approved):

gigitux:
LGTM 👍

PR #5263, opened 2021-11-26
More ts conversions

alexflorisca:
Converted the following components (and their respective sub components) to Typescript:

- [x] BlockError
- [x] CheckboxListOptions
- [x] Chip
- [x] RemovableChip


assets/js/base/components/block-error-boundary/block-error.tsx
```diff
@@ -2,9 +2,17 @@
...
 
+interface BlockErrorProps {
+	imageUrl: string;
+	header: string;
+	text: React.ReactNode;
+	errorMessage: string | null;
+	errorMessagePrefix: string;
+	button: React.ReactNode;
+}
+
```
sunyatasattva: I would suggest copying the JSDoc associated with the properties, they are still useful! They appear in Intellisense and also will appear in the Storybook docs.

Also, reading the `propTypes` it appears that `header, imageUrl, text` can be: `string | null | undefined` from the docs. In your case they are required strings.

assets/js/base/components/block-error-boundary/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+	status: string;
+	statusText: string;
+	message: string;
+}
+
+type RenderError = ( props: RenderErrorProps ) => React.ReactNode | null;
+
+interface RenderErrorProps {
+	errorMessage: React.ReactNode;
+}
```
sunyatasattva: Super extra nitpick here: I'd define `RenderErrorProps` before `RenderError` so that they are used in order and are easier to read and parse. Not a big deal if you don't want to change it, it's totally a matter of opinion.

assets/js/base/components/block-error-boundary/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+interface BlockErrorBoundaryProps {
```
sunyatasattva: This seems like it has a lot in common with `BlockErrorProps`, what about extending that?

assets/js/base/components/block-error-boundary/index.tsx
```diff
@@ -0,0 +1,93 @@
...
+ */
+import BlockError from './block-error';
+import './style.scss';
+
+interface DerivedStateReturn {
+	errorMessage: JSX.Element | string;
+	hasError: boolean;
+}
+
+interface Error {
```
sunyatasattva: Since `Error` is a default constructor type, I'd avoid overriding it. I'd call this a slightly different name perhaps?

assets/js/base/components/checkbox-list/index.tsx
```diff
@@ -25,13 +39,13 @@ import './style.scss';
  */
 const CheckboxList = ( {
 	className,
-	onChange = () => {},
+	onChange = () => void null,
```
sunyatasattva: Just to keep things consistent, what about `void 0`? Like discussed in this thread: https://a8c.slack.com/archives/C8X6Q7XQU/p1638213487071400

I don't mind `void null` either, but we should choose one :)

assets/js/base/components/checkbox-list/index.tsx
```diff
@@ -11,6 +10,21 @@ import classNames from 'classnames';
  */
 import './style.scss';
 
+interface CheckboxListOptions {
+	label: React.ReactNode;
+	value: string;
+}
+
+interface CheckboxListProps {
```
sunyatasattva, line 18: I'd check which props are required and which are optional. This type makes everything required, while the previous type had almost everything as optional.

I'd definitely check the code to see what's really required by the component. Because things like `className` are often not among them.

assets/js/base/components/chip/chip.tsx
```diff
@@ -1,32 +1,30 @@
...
-import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
 /**
  * Internal dependencies
  */
 import './style.scss';
 
-/** @typedef {import('react')} React */
+export interface ChipProps {
```
sunyatasattva, line 11: Here again perhaps it's worth it to preserve the docs? Especially for things that are not super clear immediately like `element`.

assets/js/base/components/chip/removable-chip.tsx
```diff
@@ -1,15 +1,24 @@
...
 import { __, sprintf } from '@wordpress/i18n';
 import { Icon, noAlt } from '@woocommerce/icons';
 
 /**
  * Internal dependencies
  */
-import Chip from './chip.js';
+import Chip, { ChipProps } from './chip';
+
+interface RemovableChipProps extends ChipProps {
```
sunyatasattva, line 13: Same as above regarding the docs? If it doesn't take too long…

sunyatasattva (changes requested):

assets/js/base/components/block-error-boundary/index.tsx
alexflorisca: You're right here and I realised that we are defining a lot of types in the index.tsx file so I've moved them all into a `types.ts` file.

alexflorisca:
Thanks for your review @sunyatasattva. I've made the changes you requested and corrected some errors I've spotted in the process as well. When you get the chance, have another look and let me know what you think :)

sunyatasattva (approved):
LGTM! 🐿️

PR #5264, opened 2021-11-26
Migrate Storybook to 6.4.0 and enable auto-detection of controls (#5249)

### LINKED ISSUE #5249: Migrate to latest Storybook best-practices
Storybook 6 has introduced [various improvements and deprecations](https://medium.com/storybookjs/storybook-6-migration-guide-200346241bb5). The goal of this issue is to migrate our current implementation to take advantage of everything that has been added and streamline the process of component/stories creation.

In particular:

* Deprecation of various `addons` we are currently using in favor of no-config, default addons (e.g. `actions`, `docs` etc.)
* Deprecation of manual configuration for `knobs` in favor of mainly automatic inference of components control. Migration guide here: https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-do-i-migrate-from-addon-knobs
* Out of the box TypeScript support: https://github.com/storybookjs/storybook/blob/next/addons/controls/README.md#how-do-i-migrate-from-addon-knobs

…and more.

sunyatasattva:
Previously our Babel configuration conflicted with the one provided by default by Storybook, making us unable to leverage the automatic detection of properties from both JS and TS components to create controls (formerly known as knobs).

Upgrading to 6.4.0 fixes a bug that makes it easier to work with Babel configurations. Also done here:

* Removed a few deprecated addons in favor of the replacement ones bundled within Storybook Essentials.
* Add auto-detection of certain control types depending on the property names
* Add support for TS stories

More context: https://github.com/storybookjs/storybook/issues/12292

Fixes #5249

### Screenshots

Screen Shot 2021-11-26 at 20 21 51

#### Manual Testing

1. Run `npm run storybook`
2. …be patient 😬 
3. Open Storybook at localhost:6006, in particular, you can look at [this component](http://localhost:6006/?path=/story/woocommerce-blocks-base-components-priceslider--default).
4. Notice there is a new `Controls` tab, where the properties of the components are auto-detected and documented along with the description

> ⚠️ &nbsp;**Please note:** This PR is only addressing the migration and the detection and not touching any stories. As such, I am temporarily keeping the legacy knobs, until each story is properly migrated to the controls paradigm.

### User Facing Testing

N/A (internal)

### Changelog

> Migrated to Storybook 6.4.0 with auto-detection of Controls


Aljullu (commented):
This is working and testing great, and the new features of Storybook are really cool.

The only issue I noticed is that there is quite a big increase in bundle sizes. You can see it if, after building, you compare a file from the `build` folder between `trunk` and this branch. You can also see it in the [github action results](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5264#issuecomment-980355262) in this PR.

I took a quick look and I think it's caused because of docgen. Is there any way to avoid it being applied to production builds?

sunyatasattva:
Yep, this is indeed a problem I noticed and I don't have a great answer for it @Aljullu. Formerly, it was possible to have a Babel configuration that was specific for Storybook, but this is apparently not supported anymore, according to [the docs](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#babel-mode-v7):

> In 6.x, Storybook supported a .storybook/babelrc configuration option. This is no longer supported and it's up to you to reconcile this with your project babelrc.

I had in fact asked clarification on this on the thread I had made on the Storybook repo specifically because of this bundle size increase: https://github.com/storybookjs/storybook/issues/16789#issuecomment-980410263

Any suggestion on this would be appreciated. Maybe I can force the NODE_ENV to be development and use the docgen plugin only then, but perhaps building storybook would have further differences? 🤔 Again, any suggestion appreciated.

sunyatasattva:
@Aljullu problem fixed. 0% increase in production bundle size. I have enabled doc generation only on `development` environment. My only issue with that as mentioned in the previous comment is that I didn't want to change `NODE_ENV`. Good news is that I didn't have to, as I only changed `BABEL_ENV` and provided Babel with an extra config for development transpilation and forced Storybook to get that when building and deploying itself.

Let me know what you think about this solution.

Aljullu (approved):
Thanks for investigating this, @sunyatasattva! LGTM. :ship:

PR #5273, opened 2021-11-29
Add dry run and simulate release mode

### LINKED ISSUE #3059: Refactor deploy tooling so dangerous/destructive outputs can be disabled, to aid debugging and iterating on scripts
Our build scripts are really powerful and automate a lot of stuff for us, including committing version changes, and pushing releases up to WPORG and GitHub. This power makes it tricky to iterate on the scripts and test changes without doing a real release (or hacking the scripts to comment out some sections).

This came up in testing #3027.

Open to any ideas for how to separate the safe vs impactful sections of the scripts. Note - this issue is potentially wide in scope - however small improvements / changes are probably the best way forward. Here's two ideas that could be part of a solution, or something to work towards:

- Add a --dry-run param to potentially dangerous scripts. This way, when we need to test a script that deploys, or commits, or pushes (etc), we can do so without having to hack code.
- Refactor scripts so high-impact and "safe" functionality can be run separately. (This could help power a --dry-run option.)

Note there's no silver bullet here – the only way to really ensure these scripts are reliable and robust is to use them for releases regularly. Mostly I'm advocating for changes here to help us all understand the scripts more, and to make testing/review easier as a secondary benefit.

gigitux:
This PR adds two modes that allow us to work with safety with deploy scripts.

They have been added two variables. Each variable if it is set to True corresponds to a specific mode.
- `DRY_RUN_MODE`: the commands are just printed, not executed.
- `SIMULATE_RELEASE_MODE`: the commands that affect the local env are executed (e.g. `git commit`), while the commands that affect the remote env are not executed but just printed (e.g. `git push`).

Fixes #3059

### Manual Testing

Check out this branch and disconnect the computer from the Internet to avoid unintentional deployment 🚨 

1. Set the variable `DRY_RUN_MODE` to `true` for each script file.
2. Run the command `npm run deploy`.
3. Check that no command is executed.
4. Set the variable `DRY_RUN_MODE` to `false`.
5. Set the variable `SIMULATE_RELEASE` to `true`.
6. Check that only the commands that affect the local env are executed.

Should we add specific commands in the `package.json` to run the deployment script with these different modes?


bin/github-deploy.sh
```diff
@@ -3,6 +3,11 @@
 RELEASER_PATH="$(pwd)"
 IS_PRE_RELEASE=false
 
+# When it is set to True, the commands are just printed but not executed.
+DRY_RUN_MODE=False
```
nerrad: Use `false` lower case :)

nerrad:
Sorry Luigi it looks like this escaped my attention 🤦🏻 once it's rebased, is this something that should still be usable?

gigitux:
> Sorry Luigi it looks like this escaped my attention 🤦🏻 once it's rebased, is this something that should still be usable?

Hey Darren, no worries. I rebased it and tested it and it is usable. Now, that I have more visibility, I see more space for improvements, first of all, having a global  `DRY_RUN_MODE` and `SIMULATE_RELEASE` variables.

bin/github-deploy.sh
```diff
@@ -128,63 +153,64 @@ fi
...
 # remove composer.json version bump after autoload regen (we don't commit it)
-git checkout -- composer.json
+run_command "git checkout -- composer.json" true
 
 output 2 "Committing version change..."
 echo
 
-git commit -am "Bumping version strings to new version." --no-verify
-git push origin $CURRENTBRANCH
+run_command "git commit -am 'Bumping version strings to new bbb.' --no-verify" true
```
nerrad: Where did `bbb` come from here :)

package.json
```diff
@@ -40,7 +40,6 @@
 		"changelog": "node ./bin/changelog",
 		"changelog:zenhub": "node ./bin/changelog --changelogSrcType='ZENHUB_RELEASE'",
 		"deploy": "npm run build:deploy && sh ./bin/github-deploy.sh",
-		"postdeploy": "npm run storybook:deploy",
```
nerrad, line 43: Should this be removed here?

nerrad (commented):
A couple minor things I noticed in here that I've left review comments on.

The only other thing I want to bring up is whether we'd want to show any output when simulate is executed warning the user that local file changes exist and they may want to revert those before executing the full commands? I didn't look in detail but are there any concerns that running the command again after simulating might result in some unexpected behaviour?

package.json
gigitux: Yes, the storybook deployment will be run by `storybook-deploy.sh` https://github.com/woocommerce/woocommerce-blocks/blob/5c7c3e49d94eab0337a35791cf7ff8157944799c/bin/storybook-deploy.sh/#L46-L49

gigitux:
> A couple minor things I noticed in here that I've left review comments on.

Thanks for the review. I addressed your comments!

> The only other thing I want to bring up is whether we'd want to show any output when simulate is executed warning the user that local file changes exist and they may want to revert those before executing the full commands? I didn't look in detail but are there any concerns that running the command again after simulating might result in some unexpected behaviour?

The only "unexpected" behavior is that there will be added commit and the tag in the working branch. BTW, this behavior is documented [in the comment](https://github.com/woocommerce/woocommerce-blocks/blob/5c7c3e49d94eab0337a35791cf7ff8157944799c/bin/storybook-deploy.sh).

nerrad (approved):
[Version updates](https://github.com/woocommerce/woocommerce-blocks/pull/5273/files#diff-8eb635d31e3f3345dee9ee5eb1362bb1247a9baee37f69e5c560b0369c4ce263R156) are done too in simulations right? I guess this wouldn't be a problem assuming the user uses the same version string in the live run.

I'm going to approve this because code-wise things look okay but note I **have not tested.** Thanks for taking on this issue Luigi, it'll be a helpful addition for folks trying out the release process for the first time and nervous about running the commands (also good for testing changes to the script non-destructively)

gigitux:
> [Version updates](https://github.com/woocommerce/woocommerce-blocks/pull/5273/files#diff-8eb635d31e3f3345dee9ee5eb1362bb1247a9baee37f69e5c560b0369c4ce263R156) are done too in simulations right?

To be sure I tested again and the answer is yes! Thanks for the review, I'm going to merge the PR 🥳

PR #5277, opened 2021-11-30
refactor StyleAttributesUtils

dinhtungdu:
This PR refactor the `StyleAttributesUtils` introduced in #5133. The issue with the original implementation to me is, we still need a lot of code to get the space-separated classes/styles - which is the main use-case of the utils. This PR refactor it so we can get desired properties in an immutable way without complex calculation in the markup.


gigitux (approved):
Thanks for improving `StyleAttributesUtils`! It makes a lot of sense!

PR #5287, opened 2021-12-01
[PHP 8.1] Fix code that causes a deprecation notice in BlockTypesController

Konamiman:
This pull request just fixes one small part of the code that causes a deprecation notice under PHP 8.1 (caused by passing `null` to `strtok`) when running the WooCommerce core unit tests. See [the fix pull request in WooCommerce core](https://github.com/woocommerce/woocommerce/pull/31333).


alexflorisca (approved):
Looks good, thanks for the PR!

PR #5297, opened 2021-12-02
Prevent third party blocks from breaking the entire checkout/cart block

### LINKED ISSUE #5101: Third party blocks can break cart/checkout
I noticed this when working on Mailpoet/AutomateWoo integration. Any errors or exception thrown in a third party block would break the Cart block, which is dangerous for us to allow.

### How to test
- Insert a third party block in Checkout/Cart like MailPoet, AutomateWoo.
- Load the block in incognito, it should load fine.
- Throw an error in your block.
- the whole cart/checkout should be down.

alexflorisca:
If a third party block registers itself on the checkout/cart block (or one of its' inner blocks) it will cause the entire Checkout/Cart block to error. See #5101 for more details.

This PR ads an empty `<BlockErrorBoundary>` component around each of the inner blocks rendered within the checkout/cart components. This will prevent the entire checkout/cart block from erroring if one of the sub components throws and error.

There is a risk here that because the error boundary doesn't show any errors and fails silently, that it may cause some difficulty in debugging. I think from a user perspective this is the best solution, as we don't want to show any cryptic error messages at the critical stage of a customer's journey. From a dev perspective, the error is thrown in the console and If a whole block goes missing from your checkout, you will know something is wrong! Would love to hear other's thoughts on this.

Further context in this slack p1638436874157200-slack-C8X6Q7XQU

Fixes #5101 

### Manual Testing

1. You will need two plugins that add a checkbox within the Contact Information inner block on the checkout. You can Install the [Newsletter test](https://github.com/woocommerce/newsletter-test) plugin and either duplicate it or install Mailpoet.
2. Visit the Checkout page. Expect to see two checkboxes under the email input field added by the newsletter-test and mailpoet plugins. They are rendered here as forced blocks, as the HTML hasn't been saved to the checkout yet.
3. Throw an error anywhere in this [Block](https://github.com/woocommerce/newsletter-test/blob/main/assets/js/checkout-newsletter-subscription-block/block.js#L10)
4. Reload the Chekout page. 
5. The newsletter checkbox should not be visible and the rest of the checkout block should display and function as normal. Specifically, note that the sibling checkbox is not affected. If you are logged in as an admin, you should see an error message in place of the erroring checkbox.
6. Now go to the Block based Checkout page. Make a small change here and click "Update".
7. Repeat steps 2-5. This time, the markup of the two checkbox blocks will be saved in the database, and the blocks will not be "forced" anymore.

### Edit: Additional Bug fix
While reviewing the PR, @opr pointed out a bug where the label of the checkbox from the `newsletter-subscription` plugin would disappear. Turns out this was to do with text nodes not being rendered properly within the `renderInnerBlocks()` function, which was expecting children only of type `Element`. I've changed the `renderInnerBlocks()` and `renderForcedBlocks()` functions to work with any `Node` rather than `Element`, and added the necessary types and type guards to prevent text nodes from accessing any properties only available on elements. I learned the difference between nodes and elements doing this which I wasn't aware of:

> An element is a specific type of node, one that can be directly specified in the HTML with an HTML tag and can have properties like an id or a class. can have children, etc. Nodes vs Elements: Nodes are all the different components that a webpage is made up of and elements are one type of node 

### Changelog

> Prevent third party blocks from breaking the entire checkout/cart blocks


opr:
Hey @alexflorisca, I believe we talked about this briefly, there's an issue with the implementation because any third party block erroring will cause the section it's rendered in to disappear. For example, using the newsletter plugin, making the checkbox block error causes the "Contact Information" section to leave the page. We'll need to explore more ways to prevent third party blocks breaking the checkout experience while also ensuring that the experience remains fully functional.

senadir (approved):
Thank you for working on this Alex! Are we sure that for two blocks on the same level, one of them breaking won't break the other one? I tested locally and it didn't happen to me so I think we're all good.

Another issue is that we want to show errors for admin and store managers, but not everyone else. Currently, logged in as a admin, I didn't see any errors whatsoever?

Apart from that, this is good to go!

assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -90,9 +92,14 @@ const renderForcedBlocks = (
 						? component
 						: getBlockComponentFromMap( blockName, blockMap );
 					return ForcedComponent ? (
-						<ForcedComponent
-							key={ `${ blockName }_forced_${ index }` }
-						/>
+						<BlockErrorBoundary
+							text={ `Unexpected error in: ${ blockName } }` }
```
opr: There is an extra `}` here

assets/js/atomic/utils/render-parent-block.tsx
```diff
@@ -195,39 +202,45 @@ const renderInnerBlocks = ( {
...
-								blockMap,
-								element.children,
-								blockWrapper
-							)
-						}
-					</InnerBlockComponent>
-				</InnerBlockComponentWrapper>
+				{ /* Prevent third party components from breaking the entire checkout */ }
+				<BlockErrorBoundary
+					text={ `Unexpected error in: ${ blockName } }` }
```
opr: And here

assets/js/base/components/block-error-boundary/types.ts
```diff
@@ -34,15 +38,16 @@ export interface BlockErrorProps extends BlockErrorBase {
...
 	errorMessage: React.ReactNode;
 };
 
 export interface BlockErrorBoundaryProps extends BlockErrorBase {
 	/**
 	 * Override the default error with a function that takes the error message and returns a React component
 	 */
-	renderError: ( props: RenderErrorProps ) => React.ReactNode;
+	renderError?: ( props: RenderErrorProps ) => React.ReactNode;
+	showErrorMessage?: boolean;
```
opr, line 50: Nice 😎  thanks for adding these types!

opr (changes requested):
Hey Alex, thanks for working on this, I noticed something that doesn't work quite right. The children of `InnerBlockComponentWrapper` receive the `children` and `checkoutExtensionData` props, but since we're wrapping each inner block within a `BlockErrorBoundary` the props don't get passed down to the children.

You could make `renderForcedBlocks` return this to fix it, let me know what you think.

```typescript
	return forcedBlocks.map(
		( { blockName, component }, index: number ): JSX.Element | null => {
			const ForcedComponent = component
				? component
				: getBlockComponentFromMap( blockName, blockMap );
			return ForcedComponent ? (
				<BlockErrorBoundary
					text={ `Unexpected error in: ${ blockName } }` }
					showErrorBlock={ CURRENT_USER_IS_ADMIN as boolean }
				>
					<InnerBlockComponentWrapper>
						<ForcedComponent
							key={ `${ blockName }_forced_${ index }` }
						/>
					</InnerBlockComponentWrapper>
				</BlockErrorBoundary>
			) : null;
		}
	);
```

I think there's also an issue with `renderInnerBlocks` it looks like it's only rendering one part of the block, and `children` is null, do you see this too?

assets/js/atomic/utils/render-parent-block.tsx
alexflorisca: Good spot, thanks 👍

opr (approved):
Woo! This is working for me now, both forced and unforced inner blocks are rendering correctly, nice one Alex!

Issue #5299, opened 2021-12-03
ReadMore component missing reactivity


sunyatasattva:
## Describe the bug
The `ReadMore` component is not reacting correctly to its props being changed, as some of them are transformed only once in `componentDidMount` but no update logic is there.

## To reproduce

1. Run `npm run storybook` (at the time of writing, you'll need to use branch on #5300)
2. Go to [ReadMore storybook entry](http://localhost:6006/?path=/story/woocommerce-blocks-base-components-readmore--default)
2. Try changing, e.g., `maxLines`
3. Nothing will update
4. Reload the page
5. The component will behave correctly

> ℹ️  Note: this is because Storybook saves the updated prop in the URL, so of course you can also just change the prop from there: e.g. http://localhost:6006/?path=/story/woocommerce-blocks-base-components-readmore--default&args=maxLines:10

## Next steps
Component was recently migrated to TS (see #5300), possibly the best next step is migrate it to hooks, since the component is still a class component.

sunnyjain15699:
Hi @sunyatasattva . I would like to work on this issue. Can you guide me through it ?

sunyatasattva:
Hello there @sunnyjain15699 ! Cool that you want to work on this issue.

In a nutshell, if you see the `ReadMore` component, you'll see that the clamping calculation is only done on `componentDidMount`:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/9a8f27fcd72ddfce7c75eaa681ac271732da5e94/assets/js/base/components/read-more/index.tsx#L113-L144

As such, whenever the properties get updated, the clamping will not be reflected in the actual component unless it is unmounted and mounted again. An example of this can be seen in the Storybook entry, by playing around with properties such as `maxLines`. However, this means that also such properties can't be changed dynamically by a user of this component.

That's what should be fixed basically.

The best path to do so, would be to migrate this Class component to React Hooks, if you are familiar with them. Otherwise, you'll have to use things like [`componentDidUpdate`](https://reactjs.org/docs/react-component.html#componentdidupdate) and such.

We'd really appreciate if you moved this to hooks, however.

---

Does this make sense? Any further questions?

nielslange:
@sunnyjain15699 Are you still looking into this issue?

tomasztunik:
Will be looking into this 🕵️ 

Noticed a couple of additional issues that I'll document here as I go and verify if they have been reported already, else will report them when done.

[...]

PR #5300, opened 2021-12-03
Storybook and TS migration of ReadMore component

### LINKED ISSUE #5254: Revise storybook entry for `ReadMore` component

In particular, for the `ReadMore` component:

### Controls

- [x] maxLines
- [x] ellipsis
- [x] more/lessText

sunyatasattva:
This PR adds stories for the `ReadMore` component and migrates it to TypeScript. It also migrates a bunch of associated functions and components.

Fixes #5254

### Manual Testing

1. Run `npm run storybook`
2. Explore the [ReadMore stories](http://localhost:6006/?path=/story/woocommerce-blocks-base-components-readmore--default). You should see one default story with the appropriate controls.

> ⚠️  **Please note:** Component will not properly react with control changes. This is because of the way component was written. Fixing this is not part of this PR. See #5299

### User Facing Testing
* [x] Same as above


assets/js/base/components/read-more/index.tsx
```diff
@@ -38,18 +104,22 @@ class ReadMore extends Component {
...
 		this.getButton = this.getButton.bind( this );
 		this.onClick = this.onClick.bind( this );
 	}
 
-	componentDidMount() {
+	componentDidMount(): void {
 		if ( this.props.children ) {
 			const { maxLines, ellipsis } = this.props;
 
+			if ( ! this.reviewSummary.current || ! this.reviewContent.current )
```
tjcafferkey: This seems inconsistent with the syntax in the rest of this file, and codebase. Can we open this with curly braces?

assets/js/base/components/read-more/index.tsx
```diff
@@ -70,7 +140,7 @@ class ReadMore extends Component {
 		}
 	}
 
-	getButton() {
+	getButton(): JSX.Element | undefined {
```
tjcafferkey, line 146: Can we return `null` instead of `undefined` so its consistent with other methods that render JSX e.g. the `render()`

assets/js/base/components/read-more/index.tsx
```diff
@@ -38,18 +104,22 @@ class ReadMore extends Component {
...
+	componentDidMount(): void {
 		if ( this.props.children ) {
 			const { maxLines, ellipsis } = this.props;
 
+			if ( ! this.reviewSummary.current || ! this.reviewContent.current )
+				return;
+
 			const lineHeight = this.reviewSummary.current.clientHeight + 1;
 			const reviewHeight = this.reviewContent.current.clientHeight + 1;
+
```
tjcafferkey: Added space accidentally?

assets/js/base/components/read-more/utils.ts
```diff
@@ -20,35 +30,39 @@ export const truncateHtml = ( html, length, ellipsis = '...' ) => {
...
-export const clampLines = (
-	originalContent,
-	targetElement,
-	maxHeight,
-	ellipsis
+const moveMarkers = (
+	markers: Markers,
+	currentHeight: number,
+	maxHeight: number
 ) => {
```
tjcafferkey: Can we type the return value here as `Markers`?

assets/js/base/components/read-more/utils.ts
```diff
@@ -20,35 +30,39 @@ export const truncateHtml = ( html, length, ellipsis = '...' ) => {
...
+ * @param {HTMLElement} targetElement Element which will contain the clamped content.
  * @param {number} maxHeight Max height of the clamped content.
  */
-const calculateLength = ( originalContent, targetElement, maxHeight ) => {
-	let markers = {
+const calculateLength = (
+	originalContent: string,
+	targetElement: HTMLElement,
+	maxHeight: number
+) => {
```
tjcafferkey: Can we type the return value here? Im assuming `number`?

assets/js/base/components/read-more/stories/index.tsx
```diff
@@ -1,15 +1,24 @@
...
-};
+	args: defaultProps,
+	argTypes: {
+		children: { control: { disable: true } },
+	},
+} as Meta< ReadMoreProps >;
 
-export const Default = () => (
-	<ReadMore maxLines={ 2 }>
+const LongText = (
```
tjcafferkey, line 20: Can we type this as `JSX.Element`?

tjcafferkey (changes requested):

assets/js/base/components/read-more/index.tsx
sunyatasattva: Mm, sure though I was just typing the existing situation here. As I'm not sure about the current test coverage, this could potentially have side effects. Though likely people will check for falsy values and not specifically for `undefined` and, even more likely, nobody outside of the component is consuming this function which should be `private`. 

I was just super cautious I guess 😬

assets/js/base/components/read-more/index.tsx
sunyatasattva: Good catch, thanks! Yea, I had a console.log there 😬

assets/js/base/components/read-more/index.tsx
tjcafferkey: Yeah also I don't think they will be checking the explicit value of a method that renders markup but worth checking if you think so

sunyatasattva:
@tjcafferkey Addressed the feedback! Thanks for the pointers. I have one doubt though about the whole explicit return types, but especially about `const LongText: JSX.Element` thing: wouldn't you say that's a bit unnecessary? TypeScript already knows that and it being a `const` you can't really reassign it to anything different. What's the benefit there? (I fixed it though)

tjcafferkey:
>wouldn't you say that's a bit unnecessary? TypeScript already knows that

@sunyatasattva yeah you're right I think about this. Theres no need for it, my bad.

sunyatasattva:
@tjcafferkey Reverted the explicit typing on the LongText. Let me know if there is anything else 🙏

tjcafferkey (approved):

PR #5231, opened 2021-11-24
Fix: When creating a template part for the Mini Cart area, it should be created already containing the Mini Cart Contents block

dinhtungdu:
Fixes #5196 
Depends on #5221

This PR adds the Mini Cart content to the template parts created for the Mini Cart area. Template content is taken dynamically from the template part file, and theme has a higher priority than plugin.

### Screenshots

Screen after adding a new Mini Cart template part:
image

### Manual Testing

How to test the changes in this Pull Request:

1. Check out the latest `trunk` of `woocommerce` or install the [nightly build](https://github.com/woocommerce/woocommerce/releases/tag/nightly).
2. Check out the latest `trunk` of Gutenberg.
3. Remove mini-cart template part file from the active theme if exists.
3. Go to Appearance > Editor (Beta).
4. Click on the W icon > Template Parts.
5. Click the Add new button on the top right corner.
6. Select the Mini Cart area, and give the template part a random name.
7. Click `Create`, after the editor has been loaded, the template content should contain the Mini Cart Contents block.
8. Update the content of the `woocommerce-gutenberg-products-block/templates/block-template-parts/mini-cart.html` with the one below.
9. Repeat steps 4-7. See the content of the newly created template part is the updated one in the step 9.
10. Add mini-cart template to the active theme with the content below.
11. Repeat steps 4-7. See the content of the newly created template part is the updated one in the step 11, the teampate content is taken from theme now, not the plugin.
12. Repeat steps 4-5 then Edit the default Mini Cart template part.
13. Repeat steps 4-7. See the content of the newly created template part is the updated one in the step 11, the template content is taken from template file in theme, not the customized one in the database.
---
**woocommerce-gutenberg-products-block/templates/block-template-parts/mini-cart.html:**
```
<!-- wp:paragraph --> <p>WC Mini Cart content template</p> <!-- /wp:paragraph -->
<!-- wp:woocommerce/mini-cart-contents /-->
```
---
**theme-slug/block-template-parts/mini-cart.html:**
```
<!-- wp:paragraph --> <p>Theme Mini Cart content template</p> <!-- /wp:paragraph -->
<!-- wp:woocommerce/mini-cart-contents /-->
```

PR #5221, opened 2021-11-23
Add: Mini Cart template part area

dinhtungdu:
Fixes #5194

This PR adds the Mini Cart area so users can create multiple templates for Mini Cart content.

Some limitations:

- We can't set the custom area icon because it [isn't supported](https://github.com/WordPress/gutenberg/blob/c3c3466ef6cc04d590e03fba58f3997efc653cf6/packages/editor/src/utils/get-template-part-icon.js) yet. I'm using the sidebar icon for the area. We can also think about adding custom icon support to Gutenberg later.
- The area is available for any page. It means users can add it to the page without the Mini Cart block. Ideally, that area should be available for the Mini Cart view only. Hiding it from the blocks list is an option IMO. FYI, Header/Footer/Mini Cart is a variation of the Template Part block, which are registered [here](https://github.com/WordPress/gutenberg/blob/c3c3466ef6cc04d590e03fba58f3997efc653cf6/packages/block-library/src/template-part/index.php#L149-L158).

For now, we still should merge this PR to unblock other issues. We can go back and fix these limitations later.

### Screenshots
image

#### Automated Tests
* [ ] Changes in this PR are covered by Automated Tests.
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

How to test the changes in this Pull Request:

1. Check out the latest `trunk` of `woocommerce` or install the [nightly build](https://github.com/woocommerce/woocommerce/releases/tag/nightly).
2. Check out the latest `trunk` of Gutenberg.
3. Go to Appearance > Editor (Beta).
4. Click on the W icon > Template Parts.
5. Click the Add new button on the top right corner.
6. See the Mini Cart area in the list.

### User Facing Testing
These are steps for user testing (where "user" is someone interacting with this change that is not editing any code).
* [x] Same as above, or
* [ ] See steps below.

1.
2.
3.

PR #5306, opened 2021-12-04
Fix: Mini Cart: don't allow users to create extra template parts in the Mini Cart template part area

### LINKED ISSUE #5232: The Mini Cart template part block should be available inside Mini Cart view only.
## To reproduce

1. Check out the latest `trunk` of `woocommerce` or install the [nightly build](https://github.com/woocommerce/woocommerce/releases/tag/nightly).
2. Check out the latest `trunk` of Gutenberg.
3. Go to Appearance > Editor (Beta).
4. Add new block, search for Mini Cart template.
5. See it can be added to the page template.

## Expected behavior
Mini cart template part should be available inside Mini Cart block view only.

## Screenshots
image

## Additional context
This issue comes from #5221.

The Mini Cart template part block is a variation of the Template Part block. From my initial research, I think this should be fixed in Gutenberg, but adding this issue for discussion ​to see if we can do anything on our end.

### LINKED ISSUE #5284: Mini Cart: don't allow users to create extra template parts in the Mini Cart template part area
Having more than one template part registered in the Mini Cart template part area might be confusing by users. Because of that, we won't allow the creation of extra template parts in the Mini Cart template part area.

In order to achieve this we need to:

* Revert the changes included in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5231.
* Find a way so the Mini Cart template part area is not available when creating a template part (this was introduced in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5221). I think there are two options to achieve this:
  * reverting #5221 so there isn't a Mini Cart template part area at all
  * finding some other solution so the Mini Cart template part area exists, but it's not shown in the modal when creating a new template part

dinhtungdu:
Fixes #5284.

This PR reverts changes in #5231 and #5221 to remove the template part area completely. So users can only modify the mini cart template part from the plugin/theme. This also closes #5232.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Check out the latest `trunk` of `woocommerce` or install the [nightly build](https://github.com/woocommerce/woocommerce/releases/tag/nightly).
2. Check out the latest `trunk` of Gutenberg.
3. Go to Appearance > Editor (Beta).
4. Click on the W icon > Template Parts.
5. Click the Add new button on the top right corner.
6. **Don't see** the Mini Cart area in the list.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -57,7 +57,6 @@ class BlockTemplatesController {
 		add_action( 'template_redirect', array( $this, 'render_block_template' ) );
 		add_filter( 'pre_get_block_template', array( $this, 'maybe_return_blocks_template' ), 10, 3 );
 		add_filter( 'get_block_templates', array( $this, 'add_block_templates' ), 10, 3 );
-		add_filter( 'default_wp_template_part_areas', array( $this, 'add_template_part_areas' ) );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
LGTM

@dinhtungdu, do you think we should also remove this `if` clause that assigns the Mini Cart template part to that template part area?

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/dfe4c46fe7c5b7bfd5ab7bd81cfb38834f84c98b/src/Utils/BlockTemplateUtils.php#L147-L149

dinhtungdu:
> do you think we should also remove this `if` clause that assigns the Mini Cart template part to that template part area?

Great catch, I missed that part, thanks so much!

PR #5312, opened 2021-12-05
Storybook migration of `CountryInput` component

### LINKED ISSUE #5252: Revise storybook entry for `CountryInput` component

In particular, for the `CountryInput` component:

### Controls

- [x] Error messages are not shown anywhere

sunyatasattva:
This PR fixes stories for the `CountryInput` component and migrates it to the new Storybook version. It also adds a stricter option to our TS Config: `exactOptionalPropertyTypes`.

This option will create more strict types out of optional properties. For example, in our case we had a `countries` object which looked like this:

```
const countries = {
   AF: "Afghanistan",
   AL: "Albania",
   DZ: "Algeria",
   // ...
}
```

But this component can accept a subset of countries. However, without this strict check, we would have to jump through a lot of hoops or explicitly define the subset precisely. That is:

```ts

interface CountryInputProps {
   countries: Partial<typeof countries> // ← This means we can have a subset of countries
}

const props: CountryInputProps = {
   countries: {
      AF: "Afghanistan"
   }
}

```

Without the strict check, TypeScript would be confused at this, and we'd have to define all the possible countries, with `undefined` as a value instead of their country name. With the strict check, TypeScript knows how to interpret the optional property types.

Read more: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4-beta/#exact-optional-property-types

Fixes #5252

### Manual Testing

1. Run `npm run storybook`
2. Explore the [CountryInput stories](http://localhost:6006/?path=/story/woocommerce-blocks-base-components-countryinput--default). You should see two stories:
   * A default story with all the appropriate controls inherited from the `Combobox` (note that documentation was added only for `CountryInput` specific props, as `Combobox` was not in the scope of this PR).
   * A story showing the error message when the input is required.

### User Facing Testing
* [x] Same as above


assets/js/base/components/country-input/country-input.tsx
```diff
@@ -13,7 +13,7 @@ import Combobox from '../combobox';
 import './style.scss';
 import type { CountryInputWithCountriesProps } from './CountryInputProps';
 
-const CountryInput = ( {
```
dinhtungdu, line 16: We have already exported this component as the default export. Can you explain why do we need to export it here?

dinhtungdu (commented):
Tested and this is working great! I have a question related to `CountryInput` component, otherwise, this is LGTM 👍🏽

assets/js/base/components/country-input/country-input.tsx
sunyatasattva: Hey there! Thanks for the question, I understand it might look odd.

It's for two reasons:

1. https://github.com/storybookjs/storybook/issues/13408 default exports are not correctly read by react-docgen currently. I think there is also a paragraph in the docs that mentions that but couldn't find it right now. So, that's the main issue…
2. We were discussing to move away from `default` exports to named exports anyways.

However, I didn't want to remove the `default` export at this moment as it might create bugs elsewhere. As I am very new to the repo, I'm not 100% sure on what's covered by tests and what is not, but I know we don't really stop the build on TS errors, so I thought I'd err on the safe side by not removing the default export.

dinhtungdu: @sunyatasattva Thanks for the explanation, that makes sense to me. Can you add a comment to the component to prevent confusion for other folks? The PR is LGTM now 🚢

dinhtungdu (approved):

assets/js/base/components/country-input/country-input.tsx
sunyatasattva: Good idea! I'm merging this for now, and I'll add something to explain that in a README file or something like that, to avoid having to add the same comment over and over in each component.

PR #5315, opened 2021-12-06
Remove full-width mode from cart and checkout block

### LINKED ISSUE #5010: Missing margin when using cart and checkout blocks in full width mode
## Describe the bug

* When using the cart block in full width mode, the block touches the left border of the browser.
* When using the checkout block in full width mode, the block touches the left and the right border of the browser.

## To reproduce

1. Create a test page and add the cart block.
2. Set the block to full width and save the test page.
3. Look up the cart block in the frontend and see that there's no spacing on the left-hand side.
4. Create another test page and add the checkout block.
5. Set the block to full width and save the test page.
6. Look up the checkout block in the frontend and see that there's no spacing on the left-hand and the right-hand side.

## Expected behavior

I expected that both the cart and the checkout block have some spacing between the block and the browser border.

## Screenshots

<table>
<tr>
<td>Cart:
<br><br>

cart-without-margin
</td>
<td>Checkout:
<br><br>

checkout-without-margin</td>
</tr>
</table>

nielslange:
Fixes #5010

Until now, it was possible to show the cart and the checkout block in full-width. However, there the cart block had no margin on the left hand-side margin and the checkout block had no margin on both sides. The screenshots on https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5010#:~:text=the%20browser%20border.-,Screenshots,-Cart%3A demonstrate this problem.

With this PR, I removed the full-width option so that these blocks can only use the wide option. When using the wide option, the margin problem no longer appears.

### Testing

1. Create a test page with the cart block. 
2. Verify that both the `Cart`, the `Filled Cart` and the `Empty Cart` only show the alignment options `None` and `Wide width`.
3. Create a test page with the checkout block.
4. Verify that only the alignment options `None` and `Wide width` are available.

### User Facing Testing

* [x] Same as above

### Changelog

> Fix cart and checkout margin problem by removing the full-width option.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/index.js
+++ b/assets/js/blocks/cart-checkout/cart/index.js
@@ -28,7 +28,7 @@ const settings = {
 	keywords: [ __( 'WooCommerce', 'woo-gutenberg-products-block' ) ],
 	description: __( 'Shopping cart.', 'woo-gutenberg-products-block' ),
 	supports: {
-		align: [ 'wide', 'full' ],
+		align: [ 'wide' ],
 		html: false,
 		multiple: false,
 		__experimentalExposeControlsToChildren: true,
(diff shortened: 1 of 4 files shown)


alexflorisca (approved):
Looks good, tested for both filled and empty cart and the "full width" option is not there any longer 👍

PR #5317, opened 2021-12-06
Update icon of legacy template block

### LINKED ISSUE #5269: Legacy Template block: use box icon
## To reproduce

1. With WC 6.0, Gutenberg and a block theme installed, go to Appearance > Editor.
2. Select the Single Product template.
3. Click on the Legacy Template block and notice it has the generic block icon.

## Expected behavior
It should use the `box` icon.

## Screenshots
Screenshot 2021-11-26 at 10 58 41

Change suggested by @jameskoster.

nielslange:
Currently, the Legacy Template Blocks have been using the `default` and the `page` icon. This PR replaces the current icons with a `box` icon.

Fixes #5269

### Screenshots

<table>
<tr>
<td>Before:
<br><br>

#5269-before
</td>
<td>After:
<br><br>

#5269-after
</td>
</tr>
</table>

### Testing

To test this PR, you need WooCommerce 6.0 or higher, Gutenberg and a FSE-theme, e.g. TT1 Blocks.

1. Go to `Appearance » Editor`.
2. Click on the `W` icon in the upper-left corner.
3. Go to `General templates`.
4. Open one of the product templates.
5. Click on the `Legacy Template Block` and open the `List View`.
6. Verify that the correct icon is visible in the `List View`, the block toolbar and the block itself.

### DIFF
--- a/assets/css/editor.scss
+++ b/assets/css/editor.scss
@@ -48,6 +48,15 @@
 	}
 }
 
+// Display white icon when block is selected in list view.
+.block-editor-list-view-leaf.is-selected {
+	.block-editor-list-view-block-contents {
+		svg {
+			fill: #fff;
+		}
+	}
+}
+
 // Selectors with extra specificity to override some editor styles.
 .woocommerce-search-list__list.woocommerce-search-list__list {
 	box-sizing: border-box;
--- a/assets/js/blocks/legacy-template/index.tsx
+++ b/assets/js/blocks/legacy-template/index.tsx
@@ -6,7 +6,7 @@ import { WC_BLOCKS_IMAGE_URL } from '@woocommerce/block-settings';
 import { useBlockProps } from '@wordpress/block-editor';
 import { Placeholder } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
-import { box } from '@wordpress/icons';
+import { box, Icon } from '@wordpress/icons';
 
 /**
  * Internal dependencies
@@ -57,7 +57,7 @@ const Edit = ( { attributes }: Props ) => {
 
 registerBlockType( 'woocommerce/legacy-template', {
 	title: __( 'WooCommerce Legacy Template', 'woo-gutenberg-products-block' ),
-	icon: box,
+	icon: <Icon icon={ box } color="#7f54b3" />,
 	category: 'woocommerce',
 	apiVersion: 2,
 	keywords: [ __( 'WooCommerce', 'woo-gutenberg-products-block' ) ],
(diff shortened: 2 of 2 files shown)


Aljullu (approved):
LGTM, thanks @nielslange!

mikejolley:
I was trying to test this too. I think `General templates` is just `Templates` now. I was going to comment on icon colour, but since it's not in the inpspector it probably doesn't matter.

nielslange:
> I was trying to test this too. I think `General templates` is just `Templates` now.

When using the TT1 theme, I see the following structure: `Editor » General templates » Templates`

<table>
<tr>
<td>

#5317-Editor
</td>
<td>

#5317-Templates
</td>
<td>

#5317-General-Templates
</td>
</tr>
</table>

> I was going to comment on icon colour, but since it's not in the inpspector it probably doesn't matter.

Good catch, @mikejolley! I've added the color additionally. And, I noticed that when selecting our blocks, the icon in the sidebar remained purple, which makes the icon very hard to spot. Gutenberg inverts the colors for selected blocks and displays the icon in white. I went ahead and applied the same logic for our blocks, resulting in this:

<table>
<tr>
<td>Before:
<br><br>

#5317-colored-icon-before
</td>
<td>After:
<br><br>

#5317-colored-icon-after
</td>
</tr>
</table>

Would you mind double-checking this PR?

assets/css/editor.scss
```diff
@@ -48,6 +48,15 @@
...
 }
 
+// Display white icon when block is selected in list view.
+.block-editor-list-view-leaf.is-selected {
+	.block-editor-list-view-block-contents {
+		svg {
+			fill: #fff;
+		}
+	}
+}
```
Aljullu, line 58: Oh, this looks much better. Good catch and thanks for fixing it! :clap:

Aljullu (approved):
Good point about the color, @mikejolley!

I already approved this PR, but tested it again (after rebasing with `trunk`) so giving another approval. :slightly_smiling_face:

assets/css/editor.scss
Aljullu: Taking a closer look, I think we should do a couple of changes:

* Ideally, we should make sure these styles only apply to WooCommerce icons. I think the only way of doing so is adding a class to each `Icon` instance used as the block icon. So `<Icon icon={ box } color="#7f54b3" />` would become something like `<Icon className="wc-block-editor-components-block-icon" icon={ box } color="#7f54b3" />`. Then, we could use that class in the CSS.
At this point, it might even make sense creating a `<BlockIcon />` component that encapsulates the class name and the `color` attribute so we don't need to repeat it for each block.
* We should use [`currentColor`](https://css-tricks.com/currentcolor/) instead of `#fff`, this is what Gutenberg is currently doing and it's more future-proof if they ever change the colors.

What do you think about these proposals @nielslange? I know they increase the scope of this PR so I'm ok with leaving it for a follow-up, but I think they would make our code more resilient for the future.

nielslange: > What do you think about these proposals @nielslange? I know they increase the scope of this PR so I'm ok with leaving it for a follow-up, but I think they would make our code more resilient for the future.

That's a great idea, @Aljullu! To keep this PR lean, I created https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5355. I believe adding your suggestions to this PR will shift the scope away from the initial case. Nevertheless, it's a great idea. I'll merge this PR for now and will implement the following-up PR later today.

PR #5318, opened 2021-12-06
Rollback checkout status to started instead of pristine in case of error

### LINKED ISSUE #5296: Checkout: Selected payment method token gets out of sync with paymentMethodData
If the payment data context gets reset to pristine status after selecting a saved payment method option, the saved payment method option stays selected but the paymentMethodData sent to the server when submitting the checkout again remains empty.

## To reproduce

Note that I'm using the stripe payment method in this case, but it is the same with any gateway that saves the payment method token.

1. Log in with a user that has a saved credit card that is set as the default payment method
2. Go to checkout and the default saved payment method should be selected. All good so far, react state is in sync
3. Clear a required filed like the address post code
4. Click the place order button, which will run validation and highlight the required field that is empty. Fill in the required field.
5. The saved payment method should still be selected and all information is filled out now, so click place order again
6. API returns with an error processing payment because no payment method was included in the request

## Expected behavior
The saved payment method should be included in the place order request when it is selected.

## Screenshots
image

## Environment
* Core version: 5.8.2
* WooCommerce version: 5.9
* WooCommerce Blocks version: trunk bf282a5c
* Any other plugins installed: Stripe
* Browser: Chrome 95

## Additional information
I have tracked this down to the fact that the selected saved payment method react state only sets the paymentMethodData on change: [blocks/cart-checkout/payment-methods/saved-payment-method-options.js#L53](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/cart-checkout/payment-methods/saved-payment-method-options.js#L53). The paymentMethodData gets cleared when set to pristine status [base/context/providers/cart-checkout/payment-methods/reducer.ts#L90](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/providers/cart-checkout/payment-methods/reducer.ts#L90), but saved payment methods are not part of the regular payment method processing here and it does not get set again: [payment-method-data-context.tsx#L252](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx#L252)

senadir:
When checkouts runs into an error (local validation for example), its status is rolled back to pristine, causing payment saved data to unload. For some reasoning, fixing errors and hitting submit again doesn't reset saved payment methods, @alexflorisca and I had two solutions:

- Make sure going to pristine preserve saved tokens (not a great solution given pristine should be empty).
- Don't roll back to pristine, roll back to started (the solution we did).
- Make sure saved payment methods are still picked up again later when you hit Place Order (not sure how to do this).
- 

Fixes #5296

### Testing

1. Make sure you have a saved payment.
2. On Checkout, don't check terms block, or clear postcode, just make sure there's an error.
3. Place order, you have an error.
4. Fix the error, place again, it should work.
5. The original code has something to do with express payment methods (not sure what is it @mikejolley) so make sure they're also working fine.

### Changelog

> Fix a certain use case error with saved payment methods.

### DIFF
--- a/assets/js/base/context/providers/cart-checkout/payment-methods/constants.ts
+++ b/assets/js/base/context/providers/cart-checkout/payment-methods/constants.ts
@@ -8,6 +8,7 @@ import type {
 
 export enum STATUS {
 	PRISTINE = 'pristine',
+	IDLE = 'idle',
 	STARTED = 'started',
 	PROCESSING = 'processing',
 	ERROR = 'has_error',
@@ -38,6 +39,7 @@ export const DEFAULT_PAYMENT_DATA_CONTEXT_STATE: PaymentMethodDataContextState =
 export const DEFAULT_PAYMENT_METHOD_DATA: PaymentMethodDataContextType = {
 	setPaymentStatus: () => ( {
 		pristine: () => void null,
+		idle: () => void null,
 		started: () => void null,
 		processing: () => void null,
 		completed: () => void null,
@@ -50,6 +52,7 @@ export const DEFAULT_PAYMENT_METHOD_DATA: PaymentMethodDataContextType = {
 	currentStatus: {
 		isPristine: true,
 		isStarted: false,
+		isIdle: false,
 		isProcessing: false,
 		isFinished: false,
 		hasError: false,
--- a/assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
+++ b/assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
@@ -222,7 +222,7 @@ export const PaymentMethodDataProvider = ( {
 	// When checkout is returned to idle, set payment status to pristine but only if payment status is already not finished.
 	useEffect( () => {
 		if ( checkoutIsIdle && ! currentStatus.isSuccessful ) {
-			setPaymentStatus().pristine();
+			setPaymentStatus().started();
 		}
 	}, [ checkoutIsIdle, currentStatus.isSuccessful, setPaymentStatus ] );
 
--- a/assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
+++ b/assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
@@ -222,7 +222,7 @@ export const PaymentMethodDataProvider = ( {
 	// When checkout is returned to idle, set payment status to pristine but only if payment status is already not finished.
 	useEffect( () => {
 		if ( checkoutIsIdle && ! currentStatus.isSuccessful ) {
-			setPaymentStatus().started();
+			setPaymentStatus().idle();
 		}
 	}, [ checkoutIsIdle, currentStatus.isSuccessful, setPaymentStatus ] );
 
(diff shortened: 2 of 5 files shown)


senadir:
cc @peterlama you did a lot of debugging in #5296 do you think this solutions cover all of its basis, do you happen to have another approach in mind?

@mikejolley you wrote the original code so it's worth giving this another look?

peterlama:
@senadir Setting the status to started seems like a good solution to me! Although, I wonder if the status should only be started if a saved payment method is selected? I noticed a check for that here: [payment-method-data-context.tsx#L234](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx#L234)

senadir:
After much consideration, I'm not sure `started` is the best approach here, maybe `pristine` or a new status `idle` would work best, we're still looking at this but would welcome any suggestions.

mikejolley:
@senadir As long as we have a way to distinguish `started` (when payment should be happening) from `idle` or `pristine` state, we should cover all bases. Since you also need to distinguish `pristine` from a checkout that is not yet being paid for, but has data, I think the new `idle` status will make the most sense.

assets/js/base/context/providers/cart-checkout/payment-methods/constants.ts
```diff
@@ -8,6 +8,7 @@ import type {
 
 export enum STATUS {
 	PRISTINE = 'pristine',
+	IDLE = 'idle',
```
ralucaStan, line 11: I suggest documenting these statuses for a better understanding in the future. Comments above each of them should do it.

assets/js/base/context/providers/cart-checkout/payment-methods/types.ts
```diff
@@ -66,6 +67,8 @@ export interface PaymentMethodDataContextState {
 export type PaymentMethodCurrentStatusType = {
 	// If true then the payment method state in checkout is pristine.
 	isPristine: boolean;
+	// If true then the payment method state in checkout is modified, but isn't doing anyting.
```
ralucaStan, line 70: I find this comment confusing, especially the ` but isn't doing anyting.` part. It's unclear to me what part is not doing anything? Does this idle state for a payment menthod happen when at the same time with this[ idle status](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/224c5f06ccf819d1d70e946af8445f5d7104a7bf/assets/js/types/type-defs/contexts.js#L176-L177)?

assets/js/base/context/providers/cart-checkout/payment-methods/payment-method-data-context.tsx
```diff
@@ -222,7 +222,7 @@ export const PaymentMethodDataProvider = ( {
 	// When checkout is returned to idle, set payment status to pristine but only if payment status is already not finished.
```
ralucaStan, line 222: This comment also needs updated, just replace pristine with idle here `set payment status to pristine but only`. It might help to leave some comment with the reason why this is necessary

PR #5319, opened 2021-12-06
Rest API - Namespace error handling

### LINKED ISSUE #4683: Add error handling for `Automattic\WooCommerceBlocks\RestApi::get_routes_from_namespace()`.
In various places in WordPress.org forums and at least one instance in our repo (see #4647) I've noticed reports of this error show up:

> Fatal error: Uncaught Error: Call to undefined method WP_Error::get_data()

The callstack for that usually points to this line:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/RestApi.php#L64

While at least one report (#4647) is triggered in the context of Elementor being used on the site, I suspect the underlying cause is that the function in question gets called before the `WP_REST_Server` is setup and/or before the namespaced being called is registered.

At a minimum, we should add some `WP_Error` handling in this method and return null if it's present. This will prevent the fatal error from happening in those contexts and remove it from being a potential red herring for the real issue that may be present in this scenarios.

mikejolley:
Prevents an error calling `get_data()` on the wrong object, if a WP_Error object is returned for whatever reason e.g. the endpoints not being registered.

Also fixes warnings from returned types.

Fixes #4683

### Testing

This error is not reproducible without 3rd party code. I forced an error by swapping the `wc/store` when calling the function to something invalid to ensure no fatal error occurred. Nothing a user can test here.

### Changelog

> Add error handling to `get_routes_from_namespace` method.

### DIFF
--- a/src/RestApi.php
+++ b/src/RestApi.php
@@ -78,7 +78,7 @@ class RestApi {
 	 */
 	public function store_api_authentication( $result ) {
 		// Pass through errors from other authentication methods used before this one.
-		if ( ! empty( $result ) || ! self::is_request_to_store_api() ) {
+		if ( ! empty( $result ) || ! $this->is_request_to_store_api() ) {
 			return $result;
 		}
 		return true;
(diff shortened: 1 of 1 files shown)


src/RestApi.php
```diff
@@ -61,9 +61,13 @@ public function get_routes_from_namespace( $namespace ) {
...
 		);
 
+		if ( is_wp_error( $namespace_index ) ) {
+			return [];
+		}
+
 		$response_data = $namespace_index->get_data();
 
-		return isset( $response_data['routes'] ) ? $response_data['routes'] : null;
+		return $response_data['routes'] ?? [];
```
nielslange, line 70: Is it intended that we return an empty array instead of `null` here?

nielslange (approved):
LGTM! 🚢

PR #5322, opened 2021-12-06
Chip component stories for storybook

### LINKED ISSUE #5251: Revise storybook entry for `Chip` component

In particular, for the `Chip` component:

### Controls

- [ ] Missing radius
- [ ] Fix removable chip

alexflorisca:
Refactoring the stories for the Chip component for compatibility with the new Storybook.

Fixes #5251 

1. Open up storybook with `npm run storybook`
2. You should be able to see the `Chip` and `RemovableChip` components.
3. Play around!


assets/js/base/components/chip/stories/chip.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+		},
+		className: {
+			control: 'text',
+		},
+		radius: {
+			control: 'select',
+			options: availableRadius,
+		},
+		children: {
+			control: 'text',
```
sunyatasattva: Are chips really supposed to have children? I thought they were self-closing elements with the `text` attribute. In fact, when you edit this control as text, it makes things pretty weird. What do you think?

assets/js/base/components/chip/stories/removable-chip.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+const Template: Story< RemovableChipProps > = ( args ) => (
+	<RemovableChip { ...args } />
+);
+
+export const Default = Template.bind( {} );
+Default.args = {
+	text: 'Take me to the casino',
+	screenReaderText: "I'm a removable chip, me",
+	onRemove: () => {
+		window.alert( 'Chip Removed!' ); //eslint-disable-line
```
sunyatasattva: There shouldn't be any need for this. See: https://storybook.js.org/docs/react/essentials/actions/

I'm not sure how removing the chip actually works, but if it's easy to mock and it has some interesting features (like animations and whatnot) it'd be nice to have it in the Storybook. Otherwise as long as it appears in the Actions panel, that's basically the purpose of it, to avoid those alerts or console logs.

assets/js/base/components/chip/stories/removable-chip.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+	component: RemovableChip,
+} as Meta< RemovableChipProps >;
+
+const Template: Story< RemovableChipProps > = ( args ) => (
+	<RemovableChip { ...args } />
+);
+
+export const Default = Template.bind( {} );
+Default.args = {
+	text: 'Take me to the casino',
```
sunyatasattva, line 31: You should provide a default element as an arg as well. Try to add an element string in the Storybook, when you click on the button it breaks.

You can provide non-story dependent args in the default export above. You basically do:

```
export default {
	title: 'WooCommerce Blocks/@base-components/RemovableChip',
	component: RemovableChip,
	args: {
	  element: 'li'
	}
} as Meta< RemovableChipProps >;
```

And all your stories will have that. It's pretty handy!

assets/js/base/components/chip/stories/chip.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+ * External dependencies
+ */
+import { Story, Meta } from '@storybook/react';
+
+/**
+ * Internal dependencies
+ */
+import Chip, { ChipProps } from '../chip';
+const availableElements = [ 'li', 'div', 'span' ];
+const availableRadius = [ 'none', 'small', 'medium', 'large' ];
```
sunyatasattva: As a descendant of the ancient Romans, I feel compelled to say that the plural of `radius` is `radii` 😂 

Just kidding of course, I just couldn't stop myself. It's like when Italians see people cooking pasta, they always *have to* say something.

assets/js/base/components/chip/stories/removable-chip.stories.tsx
```diff
@@ -0,0 +1,27 @@
...
+ */
+import { Story, Meta } from '@storybook/react';
+
+/**
+ * Internal dependencies
+ */
+import { RemovableChip, RemovableChipProps } from '../removable-chip';
+
+export default {
+	title: 'WooCommerce Blocks/@base-components/RemovableChip',
```
sunyatasattva: I'm thinking whether it makes sense to group the chips together like they were in the old Storybook implementation. I think it's probably nicer, especially if we think about how many components we are going to eventually have.

sunyatasattva (changes requested):
Code mostly LGTM, besides those few suggestions.

Also, is it normal that the cursor doesn't change on hover for the RemovableChip? If it is normal, it seems a small UX oversight, maybe if the team agrees we could fix it?

assets/js/base/components/chip/stories/chip.stories.tsx
alexflorisca: I know what you mean, setting children here is not really possible, whatever the type of control. I think I added this just to try to parse a html string `<div>child</div>` but gave up on the idea. The `Chip` component does render its' children and it is a valid prop, but I will remove this control here and just leave the default in place

assets/js/base/components/chip/stories/removable-chip.stories.tsx
alexflorisca: Aha nice! It looks like to remove the chip it just relies on the `onRemove` function that it takes as a prop. So nothing is hardcoded in, it's up to the consumer to define how the chip is removed.

assets/js/base/components/chip/stories/chip.stories.tsx
alexflorisca: Haha I too am a descendent of the Romans, and I bow my head in shame at the atrocity committed here. I will correct this injustice at once.

assets/js/base/components/chip/stories/removable-chip.stories.tsx
alexflorisca: I hear you and I initially did it that way. The reason I separated them into different files is because I think it's easier to separate concerns and scan the smaller files this way. I know it won't make things super complicated in this scenario, but if we later have a component directory that includes 5-10 subcomponents that all need stories, I think it'll be easier to find the relevant storybook entry if we follow the one component per file approach as much as possible.

sunyatasattva: Oh, definitely agree with you there. But stories don't have to be in the same file to be grouped in the same category on the front-end. It's all a matter of naming this title path.

See [these stories](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/assets/js/editor-components/error-placeholder/stories) and how they look on the frontend Storybook.

alexflorisca:
I've made most of the changes you requested here. Had a conversation on slack about setting the cursor to a pointer for the `x` on the RemovableChip and decided not to do it (p1638959717387500-slack-C8X6Q7XQU)

Let me know if there's anything else :)

sunyatasattva (approved):
LGTM! Thanks for bearing with me and the Romans!

https://www.youtube.com/watch?v=0lczHvB3Y9s

PR #5324, opened 2021-12-07
Storybook and TS migration of some cart checkout components

### LINKED ISSUE #5311: Storybook and TS migration of cart checkout components

As scope of this issue, we should migrate the following small components which already exist in our Storybook:

- [x] `TotalsCoupon`
- [x] `TotalsDiscount`
- [x] `TotalsFooterItem`
- [x] `ProductName`

sunyatasattva:
This PR adds stories for a few components related to the cart and checkout and migrates them to TypeScript.

The following components are within the scope of this PR:

* `totals/Coupon`
* `totals/Discount`
* `totals/FooterItem`
* `ProductName`

Fixes #5311

### Manual Testing

1. Run `npm run storybook`
2. Explore the [cart-checkout category stories](http://localhost:6006/?path=/story/woocommerce-blocks-base-components-cart-checkout-totals-coupon--default). Upon opening the category, you should see a sub-category called `totals` and overall the four components mentioned above should be available.

Specifically, these stories

* `totals/Coupon`:
    *  A Default story where, when clicking `Apply`, a loading action is simulated.
    * A story showing the loading state
    * A story showing the error state
* `totals/Discount`:
    *  A Default story where you can simulate the deletion of the coupon code by clicking on the `X`. It will simulate the loading state, an `action` will appear in the `Actions` tab, and after that the coupon will be removed and the discount total will be updated.
    * A story showing the loading state
* `totals/FooterItem`:
   * A default story
   * Story showing the item including taxes
* `ProductName`:
   * A default story with the link
   * A story showing the disabled state.

### User Facing Testing
* [x] Same as above


assets/js/base/components/cart-checkout/totals/discount/index.tsx
```diff
@@ -4,18 +4,37 @@
...
 
 /**
  * Internal dependencies
  */
 import './style.scss';
 
+export interface TotalsDiscountProps {
+	cartCoupons: LooselyMustHave<
+		CartResponseCouponItemWithLabel,
+		'code' | 'label' | 'totals'
```
alexflorisca, line 27: By looking at the old proptypes and the code, it looks like only `code` is needed here.

assets/js/base/components/cart-checkout/totals/coupon/stories/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+};
+
+const Template: Story< TotalsCouponProps > = ( args ) => {
+	const [ {}, setArgs ] = useArgs();
+
+	const onSubmit = ( code: string ) => {
+		args.onSubmit?.( code );
+		setArgs( { isLoading: true } );
+
+		setTimeout( () => setArgs( { isLoading: false } ), 3500 );
```
alexflorisca: I'd suggest a shorter timeout here, maybe 1500, otherwise it looks like the UI hangs, it's unclear that things will change back to the initial state.

assets/js/base/components/cart-checkout/totals/discount/stories/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+		const cartCoupons = args.cartCoupons.filter(
+			( coupon ) => coupon.code !== code
+		);
+
+		const values = extractValuesFromCoupons( cartCoupons );
+
+		setTimeout(
+			() => setArgs( { cartCoupons, values, isRemovingCoupon: false } ),
+			3500
```
alexflorisca: I suggest we make the timeout here a bit shorter as in my previous comment, maybe 1500, otherwise it looks like the UI hangs

assets/js/base/components/cart-checkout/totals/discount/stories/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+			'total_discount' | 'total_discount_tax'
+		>
+	);
+}
+
+export default {
+	title: 'WooCommerce Blocks/@base-components/cart-checkout/totals/Discount',
+	component: Discount,
+	argTypes: {
+		currency: currencyControl,
```
alexflorisca, line 73: I noticed that changing the currency between USD to EUR adds the € sign _after_ the price, and the $ before the price. Maybe a quick fix that can be done as part of this PR?

assets/js/base/components/cart-checkout/totals/discount/stories/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+		isRemovingCoupon: false,
+		values: extractValuesFromCoupons( EXAMPLE_COUPONS ),
+	},
+} as Meta< TotalsDiscountProps >;
+
+const Template: Story< TotalsDiscountProps > = ( args ) => {
+	const [ {}, setArgs ] = useArgs();
+
+	const removeCoupon = ( code: string ) => {
+		args.removeCoupon( code );
```
alexflorisca, line 87: You used `args.removeCoupon?.( code );` in previous components for safety. Consider doing the same here for consistency.

assets/js/base/components/product-name/index.tsx
```diff
@@ -10,10 +10,25 @@ import { AnchorHTMLAttributes, HTMLAttributes } from 'react';
...
+	/**
+	 * If `true` renders a `span` element instead of a link
+	 */
 	disabled?: boolean;
+	/**
+	 * The product name
+	 *
+	 * Note: can be an HTML string
+	 */
 	name: string;
```
alexflorisca, line 24: This is really pedantic so feel free to ignore, but I like to have all the required properties first as it's easier to scan :)

storybook/custom-controls/currency.ts
```diff
@@ -0,0 +1,43 @@
...
+		minorUnit: 2,
+		prefix: '$',
+		suffix: '',
+	},
+} as const;
+
+export const currenciesAPIShape: Record<
+	string,
+	CurrencyResponse
+> = Object.fromEntries(
```
alexflorisca, line 31: I like the use of fromEntries here to convert back to object, neat 👍

alexflorisca (commented):
Nice work, I've learnt a couple of things reviewing your PR so thanks! I left a few comments and suggestions but mostly it looks fantastic, good to see more components on Storybook 💪

assets/js/base/components/cart-checkout/totals/discount/index.tsx
sunyatasattva: `label` is used, e.g., on L 93 and 94. It's basically using to create the text for the chip :)

As for `totals` it is not technically required by the component code itself, but I noticed that's the way it is used in practice, as you need to calculate the total discount by summing up each coupon. An example of it can be seen in the story I wrote.

assets/js/base/components/cart-checkout/totals/coupon/stories/index.tsx
sunyatasattva: Good point. I'll also refactor that time into a const since I'm using in different places.

assets/js/base/components/cart-checkout/totals/discount/stories/index.tsx
sunyatasattva: Oh, I guess this is the point of the two currency configurations, look at currency.ts:

```

export const currencies: Record< string, Currency > = {
	EUR: {
		code: 'EUR',
		symbol: '€',
		thousandSeparator: '.',
		decimalSeparator: ',',
		minorUnit: 2,
		prefix: '',
		suffix: '€',
	},
	USD: {
		code: 'USD',
		symbol: '$',
		thousandSeparator: ',',
		decimalSeparator: '.',
		minorUnit: 2,
		prefix: '$',
		suffix: '',
	},
} as const;
```

Currencies have a few different configuration props, two of them are `prefix` and `suffix` and they are generally mutually exclusive. These configuration are set up to show the difference in `prefix` and `suffix`.

assets/js/base/components/cart-checkout/totals/discount/stories/index.tsx
sunyatasattva: Well, the difference here is that `removeCoupon` is not an optional prop, so it technically can't be `undefined`. In this case I also explicitly define it as an action on L74, and the user can't change event handlers through Storybook, so it's safe :)

assets/js/base/components/product-name/index.tsx
sunyatasattva: I love pedantic! And I love these kind of things. And this is something I generally want to bring up in teams, because consistency is key, I think. I tend to do alphabetical order (in my own project I even use the eslint rule for that), because I think it makes it easier to scan for people who don't know the code.

Though it sometimes creates unintuitive prop orders such as `height` before `width` and also you could have a number of props between these two.

I consider this an acceptable drawback, however.

Maybe alphabetical required, alphabetical optional could be a good ordering principle. I'll write a brief post about it and we can discuss, what say ye?

assets/js/base/components/cart-checkout/totals/discount/index.tsx
alexflorisca: Aha right you are, my bad I missed that

assets/js/base/components/cart-checkout/totals/discount/stories/index.tsx
alexflorisca: Got it 👍

assets/js/base/components/cart-checkout/totals/discount/stories/index.tsx
alexflorisca: Ah I thought `argTypes` only define the controls or type of argument, and `args` provides initial values for them. 

Either case you're right that `removeCoupon` is a required property here and there's no need 👍

sunyatasattva: Ah yes, they do, it's a bit confusing, but actions are slightly different. I'm writing a README to clarify some Storybook gotchas!

assets/js/base/components/product-name/index.tsx
alexflorisca: Sounds good, leave it as is for now 👍

storybook/custom-controls/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './currency';
+
+export const INTERACTION_TIMEOUT = 3500;
```
alexflorisca: I think you may have forgotten to change this to something lower here ;)

sunyatasattva: Thanks for catching that! 🤦‍♀️

alexflorisca (approved):
Looks good, thanks for addressing my changes, again I learn something every time :)

sunyatasattva:
> Looks good, thanks for addressing my changes, again I learn something every time :)

Likewise :)

PR #5330, opened 2021-12-07
Add missing Exception use statement in RoutesController

mikejolley:
A small fix. We're missing a `use` statement for Exception which would lead to the wrong error being thrown. I don't think this requires special testing because it's just a syntax issue.


nielslange (approved):
LGTM! 🚢

PR #5337, opened 2021-12-08
Wait until carts load before calling `canMakePayment`

senadir:
This PR fixes a new introduced issue with payment methods.

Recently, we updated the list of things we pass to payment methods `canMakePayment`, this resulted in us calling `canMakePayment` too many times so we debounced all calls to `canMakePayment` in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pul/4776

Debouncing resulted in `canMakePayment` initial call being delayed by 500ms, something we didn't wish to do. Having that delay broke our unit tests around payment methods. That was only visible once we updated our `react-testing-library` versions.

We changed that by calling the debounced `canMakePayment` directly (via `leading: true`), so it's not called at 0ms and at each 500ms max.

This fixed the broken tests, but introduced a new issue with actual payment methods that were reaching to an empty `cart` to initiate, and would produce an error because `cart` was missing currency data.

image

In this PR, we delay the initial call for the debounced `canMakePayment` until cart is done loading, we also only call it when cart isn't loading.

I also updated the cart data in tests to use previewData, not defaultData. `defaultData` is meant to be used as a placeholder until we either load `previewData` or real data, therefore, it would have `isCartLoading` as true.

### Testing steps

1. Enable stripe express payment methods.
2. Load cart/checkout, express payments should load without an error.

### DIFF
--- a/assets/js/base/context/hooks/cart/use-store-cart.ts
+++ b/assets/js/base/context/hooks/cart/use-store-cart.ts
@@ -106,7 +106,7 @@ export const defaultCartData: StoreCart = {
 	cartNeedsShipping: true,
 	cartItemErrors: EMPTY_CART_ITEM_ERRORS,
 	cartTotals: defaultCartTotals,
-	cartIsLoading: true,
+	cartIsLoading: false,
 	cartErrors: EMPTY_CART_ERRORS,
 	billingAddress: defaultBillingAddress,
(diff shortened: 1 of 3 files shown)


mikejolley (approved):
Resolves the error and tests passing, so 👍🏻

PR #5340, opened 2021-12-08
ignore eslint-plugin-woocommerce build files

alexflorisca:
After pulling latest trunk and `npm install`, I had a bunch of  build files in `bin/eslint-plugin-woocommerce/node_modules` that wanted to be added to the git repo. I'm guessing this isn't required so this PR just ignores them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ tests/cli/vendor
 # Built files
 /build/
 bin/languages
+bin/eslint-plugin-woocommerce/node_modules/
 woocommerce-gutenberg-products-block.zip
 storybook-static/
 blocks.ini


sunyatasattva (approved):
I mean, this is ok to me, but wouldn't the issue be to check why we are adding extra files in the `bin` directory via `npm i`? 🤔

alexflorisca:
The reason is multiple packages requiring different versions of `eslint-plugin-woocommerce`. The actual npm package is installed in `./node_modules`. We also have a custom version of it in `./bin/eslint-plugin-woocommerce`, and so the way npm handles the conflict with another package with the same name, is to install it in a localised `node_modules` folder. I don't think there's much we can do about this so at least this way we don't have to delete those node_modules every time

sunyatasattva:
Thanks for the explanation!

PR #5342, opened 2021-12-08
Fix reverting WC templates

### LINKED ISSUE #5272: WooCommerce templates: it's not possible to revert them
## To reproduce

1. With WC 6.0, Gutenberg and a block theme installed, go to Appearance > Editor.
2. Go to the Templates page and edit one of the WooCommerce templates (ie: Single Product Page).
3. Go back to the Templates page and try to _Clear Customizations_ of the template you just edited.
4. See error: _No templates with that id exist._

## Expected behavior
The user-edited template should be removed.

## Screenshots

https://user-images.githubusercontent.com/3616980/143886428-1283a1ca-1eb9-4b24-a0d6-fca8ce435ca6.mp4

Aljullu:
Fixes #5272.

We were hooking into the [`get_block_template()` function](https://github.com/WordPress/wordpress-develop/blob/8697aee5f9d834c1d7135524c7f13dc3c96f4d85/src/wp-includes/block-template-utils.php#L750), but when reverting a template, the function that is called is [`get_block_file_template()`](https://github.com/WordPress/wordpress-develop/blob/8697aee5f9d834c1d7135524c7f13dc3c96f4d85/src/wp-includes/block-template-utils.php#L821). In fact, `get_block_template` internally calls `get_block_file_template`, so moving our hooks into the latter would fix the issue and it shouldn't introduce any regressions. This PR implements that.

### Manual Testing

1. With WC 6.0 or later, Gutenberg and a block theme installed, go to Appearance > Editor.
2. Go to the Templates page and edit one of the WooCommerce templates (ie: Single Product Page).
3. In the frontend, verify the changes you just saved are applied.
4. Go back to the Templates page and press on _Clear Customizations_ of the template you just edited.
5. Verify there is no error and the changes have been reverted in the frontend.
6. Repeat the steps above with WP 5.9 beta without Gutenberg enabled.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Changelog

> Fix error when reverting WooCommerce templates.

### DIFF
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -55,7 +55,7 @@ class BlockTemplatesController {
 	 */
 	protected function init() {
 		add_action( 'template_redirect', array( $this, 'render_block_template' ) );
-		add_filter( 'pre_get_block_template', array( $this, 'maybe_return_blocks_template' ), 10, 3 );
+		add_filter( 'pre_get_block_file_template', array( $this, 'maybe_return_blocks_template' ), 10, 3 );
 		add_filter( 'get_block_templates', array( $this, 'add_block_templates' ), 10, 3 );
 	}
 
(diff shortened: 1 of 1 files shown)


tjcafferkey (approved):
Nice! This works as expected. Thanks for fixing!

PR #5345, opened 2021-12-09
Introduce 2 new hooks in Customer Update Route

### LINKED ISSUE #5202: `woocommerce_blocks_checkout_update_order_from_request` doesn't trigger for customer email.
I'm not sure if this was planned or not, but `woocommerce_blocks_checkout_update_order_from_request` only triggers when order updates, this means all hits to `store/checkout` would trigger it, however, `store/update-customer` (what we use when persisting the customer email) isn't part of that, and so `woocommerce_blocks_checkout_update_order_from_request` won't run.

A workaround for that is to use `woocommerce_before_order_object_save` hook.

See https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5179#issuecomment-974270622

@mikejolley is this expected behavior? I would probably move email updates to happen in `store/checkout` to solve this maybe.

mikejolley:
Adds 2 new action hooks to allow vendors to edit order/customer data when customers are updated via the API.

| Hook      | Description | Params | 
| ----------- | ----------- | ----------- |
| `woocommerce_blocks_cart_update_customer_from_request`      | Fires when the Checkout Block/Store API updates a customer from the API request data. | `\WC_Customer $customer`, `\WP_REST_Request $request`       |
| `woocommerce_blocks_cart_update_order_from_customer_request`   | Fires when the Checkout Block/Store API updates an existing draft order from customer data.        | `\WC_Order $draft_order`, `\WC_Customer $customer`, `\WP_REST_Request $request`       |

Fixes #5202

### Testing

No user facing tests. Unit test coverage added to confirm hooks fire.

### Changelog

> StoreAPI - Introduce `woocommerce_blocks_cart_update_customer_from_request` and `woocommerce_blocks_cart_update_order_from_customer_request` hooks in the StoreAPI.

### DIFF
--- a/bin/hook-docs/data/actions.json
+++ b/bin/hook-docs/data/actions.json
@@ -62,6 +62,54 @@
             }
         },
         {
+            "name": "woocommerce_after_main_content",
+            "file": "BlockTypes/LegacyTemplate.php",
+            "type": "action",
+            "doc": {
+                "description": "Hook: woocommerce_after_main_content.",
+                "long_description": "",
+                "tags": [
+                    {
+                        "name": "hooked",
+                        "content": "woocommerce_output_content_wrapper_end - 10 (outputs closing divs for the content)"
+                    }
+                ],
+                "long_description_html": ""
+            }
+        },
+        {
+            "name": "woocommerce_after_main_content",
+            "file": "BlockTypes/LegacyTemplate.php",
+            "type": "action",
+            "doc": {
+                "description": "Woocommerce_after_main_content hook.",
+                "long_description": "",
+                "tags": [
+                    {
+                        "name": "hooked",
+                        "content": "woocommerce_output_content_wrapper_end - 10 (outputs closing divs for the content)"
+                    }
+                ],
+                "long_description_html": ""
+            }
+        },
+        {
+            "name": "woocommerce_after_shop_loop",
+            "file": "BlockTypes/LegacyTemplate.php",
+            "type": "action",
+            "doc": {
+                "description": "Hook: woocommerce_after_shop_loop.",
+                "long_description": "",
+                "tags": [
+                    {
+                        "name": "hooked",
+                        "content": "woocommerce_pagination - 10"
+                    }
+                ],
+                "long_description_html": ""
+            }
+        },
+        {
             "name": "woocommerce_applied_coupon",
             "file": "StoreApi/Utilities/CartController.php",
             "type": "action",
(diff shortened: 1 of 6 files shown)


alexflorisca:
Is there any documentation we should update with these new hooks? Also is there a way to test that these fire correctly?

Code looks good, although with a caveat that I haven't done any PHP for 10 years!

mikejolley:
@alexflorisca I've added unit tests to confirm the hooks run, and updated hook docs (`npm run build:docs`). Should be good now 👍🏻

bin/hook-docs/data/actions.json
```diff
@@ -61,6 +61,54 @@
...
+                    {
+                        "name": "hooked",
+                        "content": "woocommerce_output_content_wrapper_end - 10 (outputs closing divs for the content)"
+                    }
+                ],
+                "long_description_html": ""
+            }
+        },
+        {
+            "name": "woocommerce_after_main_content",
```
alexflorisca, line 81: This is a duplicate of the above. It's also unrelated to the hooks you added in this PR. Is this something that exists but was missing from the docs?

bin/hook-docs/data/actions.json
```diff
@@ -81,9 +129,97 @@
...
+                    {
+                        "name": "hooked",
+                        "content": "woocommerce_breadcrumb - 20"
+                    }
+                ],
+                "long_description_html": ""
+            }
+        },
+        {
+            "name": "woocommerce_before_main_content",
```
alexflorisca, line 173: Again this seems to be the same as the one above, with a minor difference in the description and an extra tag. Just checking this is intended?

bin/hook-docs/data/actions.json
```diff
@@ -81,9 +129,97 @@
...
+                        "content": "woocommerce_catalog_ordering - 30"
+                    }
+                ],
+                "long_description_html": ""
+            }
+        },
         {
             "name": "woocommerce_blocks_cart_enqueue_data",
-            "file": "BlockTypes/Cart.php",
+            "file": "BlockTypes/MiniCart.php",
```
alexflorisca, line 222: Why has the Cart and MiniCart swapped round here and on the action below?

alexflorisca (commented):
Mostly looks good, but there seem to be a bunch of duplicated entries and a lot of changes made to the docs unrelated to the 2 hooks you added here. Can you just have a scan through and double check that everything is good doc wise, and if you're happy, I'm happy to approve.

alexflorisca (approved):
Discussed in slack (p1639399554056600-slack-C8X6Q7XQU) and decided to ignore the duplicated actions for now (bug reported in #5374)

PR #5348, opened 2021-12-09
Fix Mini Cart template part link on WP 5.9

Aljullu:
Originally mentioned in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5347#pullrequestreview-827560158.

The Mini Cart block link to the template part was not working on WP 5.9 beta 2 when Gutenberg was disabled. This PR fixes that:

* 7d481a58a5eab971f7e1f00723137614da88a390 fixes the link. Instead of `themes.php`, now it's in `site-editor.php`.
* 5a2095553193efbd0bd1254ddf782125028ddb7a replaces `gutenberg_experimental_is_site_editor_available` with `wp_is_block_theme`. Notice they [are equivalent](https://github.com/WordPress/gutenberg/blob/7762302a2aacd1da09d7c848ce009598aca21ef9/lib/global-styles.php#L161-L163), but only the latter is in WP 5.9.
* a5920eef059f50e7e51074d083aade90008a4ad0 adds [protection](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/6bd222805ba1fecab71a520d4da2f087827f8473/src/Assets/AssetDataRegistry.php#L276-L277) against overriding `isSiteEditorAvailable`. I don't think there was any need to remove that protection.
* 2bcc3d0a0ea411de00923022cb258185c6f945d3 renames `isSiteEditorAvailable` to `isBlockTheme` to match the WP function name.

### Manual Testing

1. With the last version of Gutenberg installed.
2. Add Mini Cart block to a page.
3. Select the block, click on the `Edit template part` link in the sidebar.
4. See site editor loaded with the plugin template part.
5. Disable Gutenberg and make sure you are running WP 5.9 beta 2.
6. Repeat steps 2-4.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart/edit.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart/edit.tsx
@@ -150,9 +150,8 @@ const MiniCartBlock = ( {
 						>
 							<ExternalLink
 								href={ addQueryArgs(
-									`${ ADMIN_URL }themes.php`,
+									`${ ADMIN_URL }site-editor.php`,
 									{
-										page: 'gutenberg-edit-site',
 										postId: `${ themeSlug }//mini-cart`,
 										postType: 'wp_template_part',
 									}
--- a/assets/js/blocks/cart-checkout/mini-cart/edit.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart/edit.tsx
@@ -58,7 +58,7 @@ const MiniCartBlock = ( {
 
 	const themeSlug = getSetting( 'themeSlug', '' );
 
-	const isSiteEditorAvailable = getSetting( 'isSiteEditorAvailable', false );
+	const isBlockTheme = getSetting( 'isBlockTheme', false );
 
 	/**
 	 * @todo Replace `getColorClassName` and manual style manipulation with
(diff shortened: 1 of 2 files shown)


dinhtungdu (approved):
Initially, I'm a bit concerned about this change because it breaks the Mini Cart template part editing on WP 5.8. But we haven't shipped the Mini Cart and we're closer to 5.9 release so this is LGTM! :shipit:

Aljullu:
Thanks for the review @dinhtungdu!

> I'm a bit concerned about this change because it breaks the Mini Cart template part editing on WP 5.8.

I think our goal should be to make the Mini Cart block available on WP 5.8, but the template editing part will need to be limited to WP 5.9, which is the version that includes the Site Editor.

@dinhtungdu just to confirm, did you test with WP 5.8 + Gutenberg 12.1 (or `trunk`) enabled?

dinhtungdu:
@Aljullu Tested with WP 5.8.2, Gutenberg `trunk` and Quadrat theme. The edit template part link is still visible in the sidebar and I got a fatal error clicking on that link:

image

Aljullu:
Thanks for checking it @dinhtungdu, I created a follow-up issue to fix that (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5357), so I will merge this PR as-is.

PR #5349, opened 2021-12-09
Downgrade @wordpress/scripts to 17.1.0 to fix dependency mistmach issue

Aljullu:
Same as #5124.

Fixes `npm run package-plugin` not working because of a dependency mismatch issue.

### Testing

1. Run `npm run package-plugin` and verify it generates a ZIP instead of displaying an error.


ralucaStan:
I've tested with a fresh `npm ci` and the zip gets created. 
However, after running `npm run package-plugin`  I see some unstaged changes in package-lock.json file. 
Does that happen to you as well @Aljullu ?

sunyatasattva (changes requested):
@ralucaStan I also get the changes in the `package-lock.json`. However, if I run `npm i` before running the scripts and commit the new lockfile, then running the scripts don't generate any extra modification.

So, perhaps, the lockfile committed here is out of date? @Aljullu maybe you can run `npm i` and see if you get a different lockfile you can commit with this PR?

Aljullu:
Yes, for some reason I'm getting a different lock file when running:

1. `npm ci && npm i` (that's equivalent to `npm ci & npm run package-plugin`, which @ralucaStan reported)
2. `rm -rf node_modules/ package-lock.json && npm i` (this is what I originally did for the PR)

I pushed the updated `package-lock.json` with the steps from 1.

sunyatasattva (approved):
LGTM! No more `package-lock.json` created!

PR #5350, opened 2021-12-09
Revise checkout payment statuses to avoid data loss on error

### LINKED ISSUE #5296: Checkout: Selected payment method token gets out of sync with paymentMethodData

mikejolley:
We're seeing errors and inconsistencies with payment method data being lost when errors cause a status change (to pristine) on checkout. On reviewing this, we've also noticed some discrepancies with how statuses (in particular the `STARTED` status) are used in a non-documented way. It's also confusing how some data is stored in the context provider, some in hooks, and yet it all needs to be kept in sync (currently with useEffect calls).

To solve this, I've taken the following approach:

- Existing payment statuses are left in place ([unlike this PR which added an IDLE status](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5318)). The main difference here is, like specified in the docs, STARTED is reserved only for Express Payment Methods. Before this PR, saved methods were also using this status to add extra data (tokens).
- Fixed PRISTINE status so that it only removes errors, not payment method data.
- Removed the `useActivePaymentMethod` hook which had methods to set active method, and active token. This has been replaced with data in the context provider.
- `setActivePaymentMethod` now accepts payment method data. By consolidating it within this function, saved methods no longer need to use the `STARTED` status to store data.
- Consolidated active payment method with active token. The active token can be derived from the active payment method's defined data. Removed `setActiveToken`.
- When a different method is made active, all payment data is cleared so that it doesn't linger for other methods.
- The logic for selecting a default gateway was handled in two places. Saved methods, and the main context provider. I've combined this into one `useEffect` hook to avoid conflicts.

Fixes #5296
Closes #5318 in favor of this.

### Testing

Please smoke test the following:

1. Payment using a core gateway (BACS, Check)
2. Payment via Stripe (using a new test card)
3. Payment via a saved card (with Stripe)
4. Payment via an express payment method (also Stripe)
5. Start an express payment and cancel it. Confirm the previously selected gateway is selected after cancellation.

Also, test the case from #5296.

1. Go to checkout and select a saved method for payment
2. Leave a required field blank and place order
3. See error
4. Correct mistake and place order again. Order should go through.
 
### Changelog

> Fixed a case where payments could fail after validation errors when using saved cards.


assets/js/blocks/cart-checkout/payment-methods/saved-payment-method-options.js
```diff
@@ -25,139 +22,91 @@ import { getPaymentMethods } from '@woocommerce/blocks-registry';
...
+						value: paymentMethod.tokenId.toString(),
+						onChange: ( token ) => {
+							const savedTokenKey = `wc-${ paymentMethodSlug }-payment-token`;
+							setActivePaymentMethod( paymentMethodSlug, {
+								token,
+								payment_method: paymentMethodSlug,
+								[ savedTokenKey ]: token.toString(),
+								isSavedToken: true,
+							} );
+							removeNotice(
```
mikejolley, line 89: This was previously handled in the context provider via useEffect, but because we want the token and method ID it seemed best to move it all to the onClick handler.

assets/js/blocks/cart-checkout/payment-methods/express-payment-methods.js
```diff
@@ -79,10 +80,10 @@ const ExpressPaymentMethods = () => {
 		( errorMessage ) => {
 			setPaymentStatus().error( errorMessage );
 			setExpressPaymentError( errorMessage );
-			setActivePaymentMethod( previousActivePaymentMethod.current );
-			if ( previousPaymentMethodData.current.isSavedToken ) {
```
mikejolley, line 83: This code appears to have been added as a workaround for started/pristine data resetting. Saved methods should not be using `STARTED` so I've removed it.

assets/js/base/context/providers/cart-checkout/payment-methods/types.ts
```diff
@@ -27,17 +27,21 @@ export type CustomerPaymentMethods =
...
+export interface PaymentMethodDispatchers {
 	setRegisteredPaymentMethods: ( paymentMethods: PaymentMethods ) => void;
 	setRegisteredExpressPaymentMethods: (
 		paymentMethods: ExpressPaymentMethods
 	) => void;
 	setShouldSavePayment: ( shouldSave: boolean ) => void;
-};
+	setActivePaymentMethod: (
+		paymentMethod: string,
+		paymentMethodData?: ObjectType | EmptyObjectType
```
mikejolley, line 38: Data can now be set at the same time as setting an active payment method. This removes the need to use STARTED status to set data.

assets/js/base/context/providers/cart-checkout/payment-methods/actions.ts
```diff
@@ -14,72 +14,68 @@ import { ACTION, STATUS } from './constants';
...
 	success: ( {
 		paymentMethodData,
 	}: {
 		paymentMethodData?: Record< string, unknown >;
-	} ): ActionType =>
-		( {
-			type: STATUS.SUCCESS,
-			paymentMethodData,
-		} as const ),
-	started: ( {
```
senadir, line 53: So we can no longer dispatch `started`?

mikejolley: No, it's dispatched as `statusOnly` now like pristine. That's all express methods need. The data addition was added for saved payment methods - with the new `setActivePaymentMethod` method that's no longer a requirement. They can stay pristine.

assets/js/base/context/providers/cart-checkout/payment-methods/reducer.ts
```diff
@@ -28,17 +25,30 @@ const reducer = (
...
+				paymentMethods: {
+					...state.paymentMethods,
+				},
+				expressPaymentMethods: {
+					...state.expressPaymentMethods,
+				},
+				shouldSavePaymentMethod: state.shouldSavePaymentMethod,
+				activePaymentMethod: state.activePaymentMethod,
+				paymentMethodData: state.paymentMethodData,
+			};
```
senadir: Instead of doing this, might as well
```suggestion
			return {
			// This keeps payment method registration state and any set data. This effectively just resets the
				// status and any error messages.
				...DEFAULT_PAYMENT_DATA_CONTEXT_STATE,
				...state,
				errorMessage: '',
				currentStatus: STATUS.PRISTINE,
			};
```

senadir: This would make sure we don't forget to add a state here later.

assets/js/blocks/cart-checkout/payment-methods/saved-payment-method-options.js
```diff
@@ -25,139 +22,91 @@ import { getPaymentMethods } from '@woocommerce/blocks-registry';
...
-							  )
-							: getDefaultPaymentMethodOptions(
-									paymentMethod,
-									setActivePaymentMethod,
-									setPaymentStatus
-							  );
-					if (
-						! activePaymentMethod &&
-						paymentMethod.is_default &&
-						activeSavedToken === ''
```
mikejolley, line 143: Not sure why we were handling defaults here. I moved it to context with the other default setter.

senadir (approved):
Tests and works very well!

chrism245:
Issue reported here as as well 4580573-zen.

PR #5352, opened 2021-12-09
Update item quantity if server result is different from client

### LINKED ISSUE #5239: Cart item quantities not re-rendered when new values are passed down via the API
## Describe the bug

One issue I attempted to tackle [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4241), is this:

When custom code is used to keep the quantities of two (or more) items in sync, the quantities of "dependent" items fail to render correctly.

## To reproduce

Here's some code you can use to play around with this state-related problem:

```
add_action( 'woocommerce_after_cart_item_quantity_update', function( $cart_item_key, $quantity = 0 ) {

	if ( isset( WC()->cart->is_syncing ) && WC()->cart->is_syncing ) {
		return;
	}

	if ( ! empty( WC()->cart->cart_contents[ $cart_item_key ] ) ) {

		$cart_item = WC()->cart->cart_contents[ $cart_item_key ];

		if ( ! $quantity || $quantity < 0 ) {
			$quantity = 0;
		} else {
			$quantity = $cart_item[ 'quantity' ];
		}

		if ( in_array( $cart_item[ 'data' ]->get_name(), array( 'Hoodie', 'Belt' ) ) ) {

			foreach ( WC()->cart->cart_contents as $search_item_key => $search_item ) {

				if ( 'Belt' === $search_item[ 'data' ]->get_name() ) {

					if ( 'Hoodie' !== $cart_item[ 'data' ]->get_name() ) {
						continue;
					}

				} elseif ( 'Hoodie' === $search_item[ 'data' ]->get_name() ) {

					if ( 'Belt' !== $cart_item[ 'data' ]->get_name() ) {
						continue;
					}

				} else {
					continue;
				}

				WC()->cart->is_syncing = true;
				WC()->cart->set_quantity( $search_item_key, $quantity, false );
				WC()->cart->is_syncing = false;
			}

		}
	}
}, 10, 2 );
```

What this code does, in a nutshell: Whenever you change the quantity of an item called "Hoodie", it searches for an item called "Belt", and tries to keep the quantity of the latter in sync with the former. Here's how this works in the legacy cart:

https://user-images.githubusercontent.com/1783726/143051125-e9eff4d8-1897-45e2-ac7a-5c629f864b1b.mp4

And here's what's happening in the block cart:

https://user-images.githubusercontent.com/1783726/143051615-1b3f9cf0-6c88-468d-9d6a-d3afe71e487d.mp4

Note that I need to refresh to get the qty of the "synced" item to update, although it's coming back just fine through the API. 
This is a roadblock that we'll need to overcome to make the new Cart/Checkout Blocks compatible with Force Sells, Product Bundles, Composite Products, Mix and Match Products, and likely other extensions that rely on the `woocommerce_after_cart_item_quantity_update` filter to keep cart item quantities in sync.

## Expected behavior
Cart item quantities should always render the values passed down via the API.

## Screenshots
See above.

* Core version: [e.g. 5.8.2]
* WooCommerce version: [e.g. 5.9.0]
* WooCommerce Blocks version: [add/change-quantity-with-steps](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/add/change-quantity-with-steps)

senadir:
Our cart's quantity input (as well as our checkout inputs) have a data direction issues, they only bubble up updates, but don't take updates from the server, meaning once the input mounts, its internal state will not reset, even if the server returns a different result.

Such systemic issues are fixed in two ways:
- Having a useEffect that watch the server state and update the local state accordingly (what we did here).
- Having the local state and server state be the same thing (what `react-query` uses) so you update the cache you have while you wait for the server to respond. This was complicated to do in this PR, and would probably need an additional action. It's worth exploring nevertheless.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5239

You can test this with two methods, `recieveCart` and server updates.

1. Copy the code in #5239 into the bottom of `woocommerce-gutenberg-products-block.php`
2. Add two products, Hoodie and Belt to your cart.
3. Updating once quantity should update the other when the server responds.

1. Add an item to your cart, go to cart.
2. In console, run this code:
```js
const cartData = wp.data.select('wc/store/cart').getCartData()
wp.data.dispatch('wc/store/cart').changeCartItemQuantity(cartData.items[0].key, 5)
```
It should update your item quantity to 5.

### Changelog

> Respect implicit quantity updates coming from server or directly from data stores.

### DIFF
--- a/assets/js/base/context/hooks/cart/use-store-cart-item-quantity.ts
+++ b/assets/js/base/context/hooks/cart/use-store-cart-item-quantity.ts
@@ -66,6 +66,9 @@ export const useStoreCartItemQuantity = (
 		storeKey
 	);
 
+	// Update local state when server updates.
+	useEffect( () => setQuantity( cartItemQuantity ), [ cartItemQuantity ] );
+
 	// Track when things are already pending updates.
 	const isPending = useSelect(
 		( select ) => {
--- a/assets/js/blocks/cart-checkout/cart/test/block.js
+++ b/assets/js/blocks/cart-checkout/cart/test/block.js
@@ -178,4 +178,31 @@ describe( 'Testing cart', () => {
 		await waitFor( () => expect( fetchMock ).toHaveBeenCalled() );
 		expect( screen.getAllByRole( 'cell' )[ 1 ] ).toHaveTextContent( '16€' );
 	} );
+
+	it( 'updates quantity when changed in server', async () => {
+		const cart = {
+			...previewCart,
+			// Make it so there is only one item to simplify things.
+			items: [
+				{
+					...previewCart.items[ 0 ],
+					quantity: 5,
+				},
+			],
+		};
+		const itemName = cart.items[ 0 ].name;
+		render( <CartBlock /> );
+
+		await waitFor( () => expect( fetchMock ).toHaveBeenCalled() );
+		const quantityInput = screen.getByLabelText(
+			`Quantity of ${ itemName } in your cart.`
+		);
+		expect( quantityInput.value ).toBe( '2' );
+
+		act( () => {
+			dispatch( storeKey ).receiveCart( cart );
+		} );
+
+		expect( quantityInput.value ).toBe( '5' );
+	} );
 } );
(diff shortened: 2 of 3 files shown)


nielslange (approved):
LGTM! 🚢

PR #5354, opened 2021-12-09
Fix conflict between Storybook and `@emotion/styled@11.x`

### LINKED ISSUE #5353: `npm run storybook` is broken
## Describe the bug
On the latest [trunk](ffc42ecb748b932ad67c47be0258699985e03233), `npm run storybook` is broken due to a dependency conflict with the `@emotion/styled` package. A [recent PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5223) upgraded a lot of packages, including the `@wordpress/components` package, which now relies on `@emotion/styled@11.6.0`. Storybook needs `@emotion/styled@10.0.27`.

There is a module resolution error where storybook is trying to resolve a new path (`@emotion/styled/base`) introduced in the latest version, using the 10.0.27 version. 

[We can't downgrade `@wordpress/components`](p1639055253464400/1639053028.462100-slack-C8X6Q7XQU)

I checked out the storybook code and it doesn't seem like the `@emotion/styled` package is used in too many places so I'm going to try and submit a PR to see if we can get them to upgrade. 🤞 

## To reproduce

1. Pull latest trunk.
2. Run `npm run storybook`
3. You should see a really really long cryptic webpack error

## Expected behavior
No error, should run storybook

sunyatasattva:
Resolve the alias for `@emotion/styled` to the root of the module in Storybook WebPack config.

Fixes #5353

### Manual Testing

1. Run `npm run storybook`
2. Confirm that the Storybook server gets run and that you can access the stories at [http://localhost:6006](http://localhost:6006)

### User Facing Testing
* [x] Same as above, or

### DIFF
--- a/storybook/webpack.config.js
+++ b/storybook/webpack.config.js
@@ -26,6 +26,8 @@ const aliases = Object.keys( tsConfig.compilerOptions.paths ).reduce(
 	{}
 );
 
+const pathToModule = ( module ) => path.join( __dirname, module );
+
 module.exports = ( { config: storybookConfig } ) => {
 	const wooBlocksConfig = getMainConfig( { alias: getAlias() } );
 	const wooStylingConfig = getStylingConfig();
@@ -42,6 +44,7 @@ module.exports = ( { config: storybookConfig } ) => {
 		'wordpress-components': require.resolve(
 			'../node_modules/wordpress-components'
 		),
+		'@emotion/styled': pathToModule( '../node_modules/@emotion/styled' ),
 	};
 	storybookConfig.module.rules.push(
 		{


storybook/webpack.config.js
```diff
@@ -42,6 +44,7 @@ module.exports = ( { config: storybookConfig } ) => {
 		'wordpress-components': require.resolve(
 			'../node_modules/wordpress-components'
 		),
+		'@emotion/styled': pathToModule( '../node_modules/@emotion/styled' ),
```
alexflorisca, line 47: I'm curious to know what the difference is between this line and 

```
'@emotion/styled': require.resolve('../node_modules/@emotion/styled'),
```

alexflorisca (approved):
Nice, great work, storybook has been resurrected! just one curious comment from me

storybook/webpack.config.js
sunyatasattva: Resurrected once again! 😬 

The difference is that `require.resolve` resolves the path to the CJS module which is exposed by the package, which would still cause problems. This is basically just outputting the string to the actual directory.

alexflorisca: Hmm, but if we pass the same path (`'../node_modules/@emotion/styled'`) to `require.resolve`, I would expect it to output the path to THAT particular node module, not the one resolved by default (i.e. the one inside the `storybook/@emotion/styled` folder)? What am I missing? 

From the docs, 

> `require.resolve` Use(s) the internal require() machinery to look up the location of a module, but rather than loading the module, just return the resolved filename.

sunyatasattva: Yea the module is the one you would have expected with `require.resolve`, but I believe it parses the package.json for the entry point or something like that. At any rate, it was linking me to the cjs bundle, but I literally just needed the path.

PR #5358, opened 2021-12-10
Switch to correct view if inner block is selected

### LINKED ISSUE #5351: Editor: On the sidebar - list view, when the user clicks on a block related to a different status, the editor doesn't render the block with the new status
## Describe the bug
For an improved UX the editor should render the block with the new status.

## To reproduce

1. Open template parts.
2. Edit the Mini Cart template.
3. Be sure that in the editor the block state is set to `Filled`
4. Open the list view and click on `Empty Mini Cart`.
5. Check that in the editor the block state doesn't change

https://user-images.githubusercontent.com/4463174/145421682-c77f754b-2d1f-4dac-b5d3-17b2642b3fb2.mp4

## Expected behavior
For an improved UX the editor should render the block with the new status after that the user clicks on `Empty Mini Cart`.

dinhtungdu:
Fixes #5351

This PR watches for the change of current selected block then switch the current view of the view switcher if an inner block is selected.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Insert or edit the Cart block.
2. Open the block navigation list panel.
3. Given the current view is the Filled Cart.
4. Select an inner block of Empty Cart.
5. See the view changes to Empty.
6. Repeat all steps above to test Mini Cart Contents block.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Changelog

> Cart block: switch to correct view if inner block is selected.

### DIFF
--- a/assets/js/blocks/cart-checkout/shared/use-view-switcher.tsx
+++ b/assets/js/blocks/cart-checkout/shared/use-view-switcher.tsx
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useState } from '@wordpress/element';
+import { useState, useEffect } from '@wordpress/element';
 import { useDispatch, select } from '@wordpress/data';
 import { ToolbarGroup, ToolbarDropdownMenu } from '@wordpress/components';
 import { Icon, eye } from '@woocommerce/icons';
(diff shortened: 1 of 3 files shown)


assets/js/blocks/cart-checkout/shared/use-view-switcher.tsx
```diff
@@ -23,8 +23,47 @@ export const useViewSwitcher = (
 } => {
 	const initialView = views[ 0 ];
 	const [ currentView, setCurrentView ] = useState( initialView );
+	const viewNames = views.map( ( { view } ) => view );
```
Aljullu: This line can be moved inside the `useEffect()`, `viewNames` doesn't seem to be used outside of the effect.

Aljullu (commented):
Works like a charm! :clap: 

Code looks good but I have a performance improvement proposal: can we move the [`views` definition](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx#L39-L50) from Cart and Mini Cart blocks outside of the component? This way the array is persisted across renders and that would avoid that `useEffect()` running on every render.

dinhtungdu:
@Aljullu thanks for the suggestion, I moved those definitions outside of components in bdf9748, can you please take another look at this PR?

Aljullu (approved):
Thanks for updating the PR @dinhtungdu, this is working like a charm. :clap:

PR #5360, opened 2021-12-10
Remove lazy loading on core components

### LINKED ISSUE #5011: Cart/Checkout blocks' loading placeholders disappear for a few seconds before the block fully loads
## Describe the bug
The Cart and Checkout blocks' loading placeholders disappear before the blocks are fully loaded.

## To reproduce

1. Go to the Cart block with an item in your cart.
2. Notice the loading placeholder disappears well before the actual block loads in.
3. Refresh the page and notice it again.

## Expected behavior
There should be a seamless transition between the loading state and the actual block.

## Screenshots
https://www.loom.com/share/64554bccbbde46b1b239fe0870d93eb0

mikejolley:
The Cart and Checkout blocks' loading placeholders disappear before the blocks are fully loaded. This is in part due to lazy loading of components which can add a visible delay while the script is loaded from the server.

To mitigate this problem we should ensure that the core components of the checkout **which always render** are not lazy-loaded, reserving lazy loading for optional components, or components that don't need to be visible straight away.

Fixes #5011

This was split from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5026

Components that should not lazy load:

- Wrapper components that contain other blocks (cart and checkout main area/sidebar)
- Mandatory interactive components (address fields and form steps)
- Other required components (cart items)

#### Testing

To test the skeleton before DOM and react is finished loading:

- Open dev tools network panel
- Set the speed to 3g or something slow
- View a page containing the cart and checkout blocks

You will see the skeleton loading view, and once loaded, the components will replace them.

#### Performance Impact

The bundle size may increase due to removing some lazy loading components, but the perceived speed should be improved because the cart and checkout blocks will render faster without an obvious delay.

#### Changelog

> Improve the cart and checkout loading states.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/inner-blocks/register-components.ts
+++ b/assets/js/blocks/cart-checkout/cart/inner-blocks/register-components.ts
@@ -21,77 +21,41 @@ import cartOrderSummaryMetadata from './cart-order-summary-block/block.json';
 import cartTotalsMetadata from './cart-totals-block/block.json';
 import cartProceedToCheckoutMetadata from './proceed-to-checkout-block/block.json';
 import cartAcceptedPaymentMethodsMetadata from './cart-accepted-payment-methods-block/block.json';
-
-registerCheckoutBlock( {
-	metadata: filledCartMetadata,
-	component: lazy( () =>
-		import(
-			/* webpackChunkName: "cart-blocks/filled-cart" */ './filled-cart-block/frontend'
-		)
-	),
-} );
+import emptyCartComponent from './empty-cart-block/frontend';
+import filledCartComponent from './filled-cart-block/frontend';
+import cartItemsComponent from './cart-items-block/frontend';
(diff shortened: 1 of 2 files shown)


senadir (changes requested):
Pulling in the conclusion of our discussion in slack.

Directly loading core blocks with instead of lazy loading would prevent us in the future from efficiently replacing them. This is not a huge concern right now given we don't allow replacing blocks yet, and the size of core blocks isn't that big. But switching now would just regress translations again.

The alternative solution we decided on is to:
- Revisit our loading placeholders/animation so that it still works fine, even with lazy loaded blocks.
- If that's not possible, break core blocks to explicit files loaded by WordPress, this would get the benefit of loading before rendering, and would us to dequeue in the future if we need to. There's still a risk of dequeueing.
- If all fails, the current approach is what we go with.

mikejolley:
I'm going to close this so it can be looked at with fresh eyes @senadir

PR #5362, opened 2021-12-10
Loading animation: change from pulse to swipe

mikejolley:
Pulled from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5026

I updated the pulsing animation to a 'loading' swipe type animation, based on a Skeleton component I've been using. The pulse was a little too subtle and the swipe seems to better represent that loading is progressing. No issue for this one, just something I thought looked better :)

### Screenshots

| Before      | After |
| ----------- | ----------- |
|       |        |

### Changelog

> Update loading skeleton animations

### DIFF
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -12,32 +12,56 @@ $fontSizes: (
 	font-size: map-get($fontSizes, $sizeName) * $multiplier;
 }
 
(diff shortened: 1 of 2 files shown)


alexflorisca (approved):
Looks good, both by throttling the network request and the sliding animation is a bit more prominent you're right 👍

PR #5363, opened 2021-12-10
Limit CSS styles to WooCommerce Block icons only

### LINKED ISSUE #5355: Limit CSS styles to WooCommerce Block icons only
## Describe the bug

In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5317, I added the following CSS definition that ensures that the icon remains readable when being selected in the list view:

```css
// Display white icon when block is selected in list view.
.block-editor-list-view-leaf.is-selected {
	.block-editor-list-view-block-contents {
		svg {
			fill: #fff;
		}
	}
}
```

In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5317#discussion_r765857807, @Aljullu pointed out that this CSS definition should only affect the icons of the WooCommerce Blocks plugin. In addition, he suggested using `currentColor` instead of `#fff`.

## To reproduce

1. Create a test page.
2. Add one block to the page, that comes from the WooCommerce Blocks plugin.
3. Add one block to the page, that does not come from the WooCommerce Blocks plugin.
4. Open the list view.
5. Select the block that comes from the WooCommerce Blocks plugin.
6. Verify that the icon color, when the block is selected, comes from the the WooCommerce Blocks plugin.
7. Select the block that does not come from the WooCommerce Blocks plugin.
8. Verify that the icon color, when the block is selected, also comes from the the WooCommerce Blocks plugin.

## Expected behavior

The CSS definition for the icon should only affect blocks of the WooCommerce Blocks plugin.

nielslange:
Fixes #5355

### Notes

At the moment, when selecting a block within the List View (Mac: `control + option + o`), the icon remains purple (`#7F54B3`). With the blue background color (`#007CBA`), the contrast ratio of the icon drops to 1.2:1 (see https://webaim.org/resources/contrastchecker/ or https://color.adobe.com/create/color-contrast-analyzer). This PR is doing the following things:

1. Remove the hard-coded foreground color.
2. Add a CSS class to all icons.
3. Add CSS styles for the unselected and the selected state. 

---

Initially, I thought about adding the CSS class to the `<Icon>` component itself rather than the individual elements. However, we're also using icons without a defined color, such as the `eye` icon within the `ViewSwitcherComponent`. Therefore, I decided to add the CSS classes to the affected icons itself.

### Testing

1. Create a test page.
2. Add all blocks of this plugin to the test page.
3. Open the List View (Mac: `control + option + o`).
4. Click on each block.
5. Verify that, before selecting a block, the icon color is purple, and the text color is black.
6. Verify that, when selecting a block, both the icon color and the text color is white.

### User Facing Testing

* [x] Same as above

### DIFF
--- a/assets/css/editor.scss
+++ b/assets/css/editor.scss
@@ -45,6 +45,18 @@
 	}
 }
 
+svg.wc-block-editor-components-block-icon {
+	color: rgb(127, 84, 179);
+}
+
+.block-editor-list-view-leaf.is-selected {
+	.block-editor-list-view-block-contents {
+		svg.wc-block-editor-components-block-icon {
+			color: currentColor;
+		}
+	}
+}
+
 // Selectors with extra specificity to override some editor styles.
 .woocommerce-search-list__list.woocommerce-search-list__list {
 	box-sizing: border-box;
--- a/assets/css/editor.scss
+++ b/assets/css/editor.scss
@@ -46,7 +46,7 @@
 }
 
 svg.wc-block-editor-components-block-icon {
-	color: rgb(127, 84, 179);
+	color: $studio-woocommerce-purple-50;
 }
 
 .block-editor-list-view-leaf.is-selected {
--- a/assets/js/atomic/blocks/product-elements/add-to-cart/constants.js
+++ b/assets/js/atomic/blocks/product-elements/add-to-cart/constants.js
@@ -5,7 +5,12 @@ import { __ } from '@wordpress/i18n';
 import { cart, Icon } from '@woocommerce/icons';
 
 export const BLOCK_TITLE = __( 'Add to Cart', 'woo-gutenberg-products-block' );
-export const BLOCK_ICON = <Icon srcElement={ cart } />;
+export const BLOCK_ICON = (
+	<Icon
+		srcElement={ cart }
+		className="wc-block-editor-components-block-icon"
+	/>
+);
 export const BLOCK_DESCRIPTION = __(
 	'Displays an add to cart button. Optionally displays other add to cart form elements.',
 	'woo-gutenberg-products-block'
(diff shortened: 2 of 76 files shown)


assets/css/editor.scss
```diff
@@ -45,6 +45,18 @@
 	}
 }
 
+svg.wc-block-editor-components-block-icon {
+	color: rgb(127, 84, 179);
```
Aljullu: One of the benefits of moving this to the CSS is that we can use the [CSS variable](https://github.com/Automattic/color-studio/blob/master/dist/color-variables.scss#L166), now: :slightly_smiling_face: 
```suggestion
	color: $studio-woocommerce-purple-50;
```

assets/css/editor.scss
```diff
@@ -45,6 +45,18 @@
...
+}
+
+.block-editor-list-view-leaf.is-selected {
+	.block-editor-list-view-block-contents {
+		svg.wc-block-editor-components-block-icon {
+			color: currentColor;
+		}
+	}
+}
```
Aljullu: Nit: this CSS selector can be simplified:
* `.block-editor-list-view-block-contents` isn't really needed.
* `svg` isn't needed either because now that we use our own class, we know it's always applied to an svg.

```CSS
.block-editor-list-view-leaf.is-selected .wc-block-editor-components-block-icon {
	color: currentColor;
}
```

Aljullu (approved):
Looks good, thanks for following-up from the previous PR @nielslange! I left a couple of suggestions to simplify the CSS, but pre-approving. :ship:

assets/css/editor.scss
nielslange: Nice! Thanks for pointing this out, @Aljullu. I should have checked that. 🙈

assets/css/editor.scss
nielslange: I really appreciate `nit` comments! You're absolute correct in pointing them out, and I applied the suggested changes right away. I'm now going to merge this PR so that everyone can benefit from these changes. 

Thanks again for reviewing this PR so quickly, @Aljullu! 🙏

PR #5364, opened 2021-12-10
Move is-loading` class removal to render callback

mikejolley:
This is a minor change pulled from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5026/files.

Currently, the `is-loading` class is removed when we're traversing the DOM and replacing HTML elements with React Components. While not noticeable, this is really too early. We should remove the `is-loading` class once the component has been rendered, and to do this, we just need to add a render callback function. 

[Render docs are here.](https://reactjs.org/docs/react-dom.html#render)

> If the optional callback is provided, it will be executed after the component is rendered or updated.

#### Testing

Just smoke test that blocks (cart and checkout) render on the frontend.


nielslange (approved):
Works as expected! 🚢

PR #5370, opened 2021-12-10
Migrate storybook entries of QuantitySelector component

### LINKED ISSUE #5367: Storybook migration of `QuantitySelector` component

In particular, for the `QuantitySelector` component:

### Controls

- Missing all controls

sunyatasattva:
This PR migrates stories for the `QuantitySelector` component and adds documentation for its props.

It also reverts a workaround introduced in #5354 in lieu of the official workaround recently published on the Storybook docs (see: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#emotion11-quasi-compatibility).

Fixes #5367

### Manual Testing

1. Run `npm run storybook`
2. Explore the [QuantitySelector stories](http://localhost:6006/?path=/story/woocommerce-blocks-base-components-quantityselector--default). There should be two stories: a default one with the controlled component (i.e. can't set the values from the controls, but only from the action), a second one showing the component in its disabled state.

### User Facing Testing
* [x] Same as above


assets/js/base/components/quantity-selector/stories/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+import QuantitySelector, { QuantitySelectorProps } from '..';
+
+export default {
+	title: 'WooCommerce Blocks/@base-components/QuantitySelector',
+	component: QuantitySelector,
+	args: {
+		itemName: 'widgets',
+	},
+	argTypes: {
+		quantity: { control: { disable: true } },
```
alexflorisca: Why is the control for quantity disabled here? I suggest it might be useful to let a user vary the initial quantity with a control here

assets/js/base/components/quantity-selector/stories/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+	},
+} as Meta< QuantitySelectorProps >;
+
+const Template: Story< QuantitySelectorProps > = ( args ) => {
+	const [ quantity, setQuantity ] = useState( 1 );
+
+	const onChange = ( newVal: number ) => {
+		args.onChange?.( newVal );
```
alexflorisca, line 25: I didn't know the optional chaining can be used as a check for a function here, that's cool!

alexflorisca (commented):
Nice work, just left a minor question/suggestion

assets/js/base/components/quantity-selector/stories/index.tsx
sunyatasattva: @alexflorisca I've updated this, but because the way this component is supposed to be used is that the state is updated by its own plus/minus, though it still technically is a controlled component, it doesn't play in the best way with Storybook, and I had to use some magic to make it work.

Notice that, however, because of [this bug](https://github.com/storybookjs/storybook/issues/15924) you can change the number from the control and it will reflected in the component, but not viceversa.

alexflorisca (approved):
Looks good 👍

Issue #5357, opened 2021-12-10
Mini Cart: make it so the link to the template part works on WP 5.8 + GB and WP 5.9


Aljullu:
Issue surfaced by @dinhtungdu in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5348#pullrequestreview-828545463. The template editor URL is different in WP 5.8 + GB and WP 5.9, so we can't reliably link to it without doing a WP version check.

Ideally, instead of adding the logic to check the WP version in WC Blocks, we can contribute upstream in Gutenberg and create a `get_site_editor_url()` function that we, and other plugin devs, can use to link to the site editor. See: https://github.com/WordPress/gutenberg/issues/37277.

gigitux:
In accordance with [this comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5372#issuecomment-992298108), I think that we can close this issue. What do you think? @Aljullu

Aljullu:
Right!

PR #5372, opened 2021-12-13
Mini Cart: only display the editing link to users with site editing permission

### LINKED ISSUE #5359: Mini Cart: investigate if 'Edit template part' link should be available only to users with certain capabilities
The Site editor is not available to all users, only those with `current_user_can( 'edit_theme_options' )` can access it, if I'm not wrong. Because of that, the link to edit the template part should only be visible to users that can edit it.

This issue is about:

* Researching which permission level is needed to edit template parts.
* Update the Mini Cart block code so the link to edit the template part is only available to users with those permissions.

## To reproduce
1. Go to Users > Add new and create a user with role _Author_.
2. Log in with that user.
3. Add a Mini Cart block and click on the Edit template part link:
imatge
4. Notice you are redirected to a page which you don't have permission to see:
imatge

dinhtungdu:
Fixes #5359

Instead of passing the theme slug and the theme type as we're doing, this PR prepares the full edit URI for the Mini Cart template part in the editor then passes it to the block editor. By doing so, we can check for the current user permission and WP version then pass the appropriate value to the editor. On the editor site, we just need to check for one variable, which makes our code more readable.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Switch to a user with the Editor role.
2. Edit a page containing Mini Cart block.
3. Select the Mini Cart block, don't see the edit template part link.
4. Switch to an admin account.
5. Edit a page containing Mini Cart block.
6. Select the Mini Cart block, see the edit template part link.
7. Switch to a non-block theme.
8. Edit a page containing Mini Cart block.
9. Select the Mini Cart block, don't see the edit template part link.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart/edit.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart/edit.tsx
@@ -12,12 +12,10 @@ import type { ReactElement } from 'react';
 import { formatPrice } from '@woocommerce/price-format';
 import { CartCheckoutCompatibilityNotice } from '@woocommerce/editor-components/compatibility-notices';
 import { PanelBody, ExternalLink, ToggleControl } from '@wordpress/components';
-import { addQueryArgs } from '@wordpress/url';
-import { ADMIN_URL, getSetting } from '@woocommerce/settings';
+import { getSetting } from '@woocommerce/settings';
 import { __ } from '@wordpress/i18n';
 import { positionCenter, positionRight, positionLeft } from '@wordpress/icons';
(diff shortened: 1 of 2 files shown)


dinhtungdu:
This PR also closes #5357. The `site-editor.php` is only used when 5.9 is released. Beta and lower versions still use the Gutenberg edit site page.

sunyatasattva (approved):
LGTM! :)

PR #5373, opened 2021-12-13
Trigger validation on change if the input is not focussed 

### LINKED ISSUE #5130: Autofill not working in Cart shipping form
Like https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4495, Cart form is also broken for autofill. If you autofill a form, your fields errors won't clear
## To reproduce

1. Go to Cart block
2. Open shipping calculator.
3. Autofill an address.
4. Except State and the field you selected, other fields won't actually clear errors.

## Expected behavior
Errors should clear.

## Screenshots

https://user-images.githubusercontent.com/6165348/141306538-a66e0ef2-6808-49db-b937-d323443c3247.mov

mikejolley:
This PR fixes #5130 and replaces the fix for #4495.

When browser autofill triggers, from what I can tell with logging, onChange events are fired. However, validation occurs onBlur and this is not fired for autofilled inputs.

Because we can track the onChange event, we can ensure that validation runs (in the background) on non-focused fields.

@senadir this replaces your event handling fix. Does it make more sense to you?

### Testing

1. In a clean browser session with no address, add an item to the cart
2. Go to Cart block
3. Open shipping calculator.
4. Autofill an address.
5. Click update
6. You should see no inline validation errors

### Changelog

> Fix validation error handling after using browser autofill

### DIFF
--- a/assets/js/base/components/button/index.tsx
+++ b/assets/js/base/components/button/index.tsx
@@ -15,6 +15,9 @@ interface ButtonProps extends WPButton.ButtonProps {
 	className?: string;
 	showSpinner?: boolean;
 	children?: ReactNode;
+	disabled?: boolean;
+	onClick?: ( e: React.MouseEvent< HTMLButtonElement, MouseEvent > ) => void;
+	type?: 'input' | 'submit';
 }
 
 /**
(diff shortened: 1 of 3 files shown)


senadir (approved):
This is an elegant solution that fixes the issue and decouples inputs from the checkout processor.

PR #5375, opened 2021-12-13
Set the origin of custom templates to 'plugin' and fix plugin name on REST requests

### LINKED ISSUE #5327: WooCommerce templates: two 404 requests when loading the templates list
WP makes two 404 requests when loading the Templates list (in Appearance > Editor > Templates). Those requests are:

* `/wp-json/wp/v2/themes/woocommerce`
* `/wp-json/wp/v2/plugins/WooCommerce`

While nothing seems to break because of that, we should try to avoid making unnecessary requests.

## To reproduce

1. Go to Appearance > Editor > Templates with your browser devtools Network tab open.
2. Notice there were two 404 requests:

imatge

## Expected behavior
We should investigate why those requests return 404 and avoid them.

Aljullu:
Fixes #5327.

Templates that had been edited by the user were considered as being theme templates even though they are plugin templates. This PR fixes that setting the `origin` to `plugin` in those cases.

This PR also fixes the plugin name, after some investigation I found it must be in the form `woocommerce/woocommerce`.

### Manual Testing

1. Go to Appearance > Site Editor > Templates.
2. If you haven't edited any of the WooCommerce templates, edit one and go back to the Templates list.
3. Verify the WooCommerce edited template has the plugin icon (plug) instead of the theme icon (icon of a webpage).
4. Verify it shows the plugin name as `WooCommerce` instead of `woocommerce` (notice the capital letters).
5. Verify there were no 404 calls made when loading the page.

| Before | After |
| --- | --- |
| imatge | imatge |

Note: This PR introduces a regression, the 'blue' circle on top of the icon indicating that it has been customized by the user does not appear when the origin is a plugin. I think that's a bug in Gutenberg and I created a PR in that repo: https://github.com/WordPress/gutenberg/pull/37329.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Changelog

> Site Editor template list: Fix wrong icon displayed on WooCommerce templates after they have been edited.

### DIFF
--- a/src/Utils/BlockTemplateUtils.php
+++ b/src/Utils/BlockTemplateUtils.php
@@ -117,6 +117,7 @@ class BlockTemplateUtils {
 		}
 
 		if ( 'woocommerce' === $theme ) {
+			$template->theme  = 'woocommerce/woocommerce';
 			$template->origin = 'plugin';
 		}
 
@@ -137,7 +138,7 @@ class BlockTemplateUtils {
 		$template_content         = file_get_contents( $template_file->path );
 		$template                 = new \WP_Block_Template();
 		$template->id             = 'woocommerce//' . $template_file->slug;
-		$template->theme          = 'WooCommerce';
+		$template->theme          = 'woocommerce/woocommerce';
 		$template->content        = self::gutenberg_inject_theme_attribute_in_content( $template_content );
 		$template->source         = 'plugin';
 		$template->slug           = $template_file->slug;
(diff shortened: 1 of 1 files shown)


gigitux (approved):
LGTM! 

The only thing that is not clear to me: what does the theme variable represent?

Aljullu:
Thanks for the review @gigitux!

> The only thing that is not clear to me: what does the theme variable represent?

That's what Gutenberg uses to know what's the theme/plugin the template belongs to. I agree it's confusing using `theme` when the template belongs to a plugin. I opened an issue in Gutenberg to know their thoughts on that: https://github.com/WordPress/gutenberg/issues/37401.

PR #5379, opened 2021-12-14
Sync order data with cart data when cart is updated from any route

### LINKED ISSUE #5366: Updating shipping method at checkout doesn't update the draft order that's been created.
## Describe the bug
During checkout, changing the selected shipping method updates the data in the session, but does not update the draft order that has been created.

## To reproduce

1. Make sure at least 2 shipping methods are enabled for your store.
2. Add products to your cart.
2. Go to the block checkout page.
3. In a new window/tab, open the 'Edit' screen for the draft order created on `wp-admin/edit.php?post_status=wc-checkout-draft&post_type=shop_order`
4. Observe shipping method.
5. Go back to the checkout page and update the shipping method.
6. Go back to the draft order screen, refresh and see that the shipping method has not changed to match the checkout UI.

## Expected behavior
I would expect that updating the shipping method on the checkout UI would cause that shipping method to be updated anywhere that my cart/order/session data is stored.

## Screenshots
Here's a screen recording of me going through the above steps and then also updating the shipping address in checkout to demonstrate that doing so updates the draft order.

https://user-images.githubusercontent.com/1558827/145601832-64d3fffe-de36-4308-aff1-0dc6559a2d75.mp4

* Core version: 5.8.2
* WooCommerce version: 5.8.0
* WooCommerce Blocks version: 6.2.0

* OS: Mac
* Browser Brave
* Version 1.32.115

## Additional context
cc @senadir

mikejolley:
Syncs cart to draft order (if one exists) on all POST PUT DELETE requests so draft orders are always current, even on the cart page.

Fixes #5366

### Testing

Ensure you have a store with products and at least 2 shipping methods.

1. Add an item to the cart.
2. Visit the checkout page.
3. In another tab, open admin and view the draft order. Note the shipping method.
4. Leave checkout and go to the cart page. Change the shipping method.
5. In your admin tab, refresh the order. Confirm the shipping method + totals all updated.
6. Smoke test checkout/cart to ensure totals, taxes, items, and fee totals all make sense and change when the cart changes.

### Performance Impact

Does write to orders more often so API requests may slow down due to this.

### Changelog

> Sync draft orders whenever cart data changes.


src/StoreApi/Routes/CartUpdateCustomer.php
```diff
@@ -146,56 +145,4 @@ protected function get_customer_billing_address( \WC_Customer $customer ) {
...
-		);
-
-		/**
-		 * Fires when the Checkout Block/Store API updates an existing draft order from customer data.
-		 *
-		 * @param \WC_Order $draft_order Order object.
-		 * @param \WC_Customer $customer Customer object.
-		 * @param \WP_REST_Request $request Full details about the request.
-		 */
-		do_action( 'woocommerce_blocks_cart_update_order_from_customer_request', $draft_order, $customer, $request );
```
mikejolley, line 197: @senadir this didn't make sense after moving all updates to cart...

ralucaStan:
Heads up: The PHP unit tests are falling, I've rerun them

ralucaStan (commented):
I tested and this is what I've noticed:
- changing the address using the shipping calculator does not change the address on the order draft. Should it?
- updating the nr of items is reflected in the draft order. But if I have one item, and I remove it from the cart and I see the empty cart message, the draft order does not get removed. I would expect that this order gets deleted

ralucaStan:
Also, might be worth having another pair of eyes on the PHP code.

src/StoreApi/Routes/CartUpdateCustomer.php
opr: Do we need to deprecate this, it doesn't look like we made this hook public but is it possible someone is using it out there? We should also remove these lines from the tests since we're removing this hook. https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/cda1bae2353e6d072ab1891cb4d1decfa59c40fc/tests/php/StoreApi/Routes/Cart.php#L194
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/cda1bae2353e6d072ab1891cb4d1decfa59c40fc/tests/php/StoreApi/Routes/Cart.php#L191
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/cda1bae2353e6d072ab1891cb4d1decfa59c40fc/tests/php/StoreApi/Routes/Cart.php#L199

src/StoreApi/Routes/AbstractCartRoute.php
```diff
@@ -102,6 +116,29 @@ protected function requires_nonce( \WP_REST_Request $request ) {
...
+			$this->order_controller->update_order_from_cart( $draft_order );
+
+			/**
+			 * Fires when the order is synced with cart data from a cart route.
+			 *
+			 * @param \WC_Order $draft_order Order object.
+			 * @param \WC_Customer $customer Customer object.
+			 * @param \WP_REST_Request $request Full details about the request.
+			 */
+			do_action( 'woocommerce_blocks_cart_update_order_from_request', $draft_order, $request );
```
opr, line 138: Shall we add a deprecated call to `woocommerce_blocks_cart_update_order_from_customer_request` here?

src/StoreApi/Routes/AbstractCartRoute.php
```diff
@@ -102,6 +116,29 @@ protected function requires_nonce( \WP_REST_Request $request ) {
 		return 'GET' !== $request->get_method();
 	}
 
+	/**
+	 * Triggered after an update to cart data. Re-calculates totals and updates draft orders (if they already exist) to
+	 * keep all data in sync.
+	 *
+	 * @param \WP_REST_Request $request Request object.
+	 */
+	protected function cart_updated( \WP_REST_Request $request ) {
```
opr, line 125: Not saying it's wrong because I'm not sure - but why is this in the past tense?

opr (changes requested):
Functionality-wise this works great thanks Mike, but we removed a hook (`woocommerce_blocks_cart_update_order_from_customer_request`) as mentioned in my comment, I don't think we ever publicly documented this but should we deprecate it anyway in case someone used it? We also need to update tests so we're no longer testing for it

src/StoreApi/Routes/AbstractCartRoute.php
mikejolley: In the docblock I wrote "Triggered after an update to cart data", so I guess because the cart has already updated by this point.

src/StoreApi/Routes/AbstractCartRoute.php
mikejolley: > is it possible someone is using it out there

Not yet because it hasn't released. Was merged for 6.6

src/StoreApi/Routes/CartUpdateCustomer.php
mikejolley: Since this was unreleased in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5345 we can avoid deprecation if we ship this in 6.6.0 instead.

mikejolley:
> updating the nr of items is reflected in the draft order. But if I have one item, and I remove it from the cart and I see the empty cart message, the draft order does not get removed. I would expect that this order gets deleted

I don’t think we should delete the order because orders are cleaned up automatically after a certain time elapses. Doing so would also not free up the order ID, so if they added items to the cart it would cause another order to be created with another ID. I think it's best to leave it around.

mikejolley:
> changing the address using the shipping calculator does not change the address on the order draft. Should it?

A sync is indeed triggered but the update is hidden by core because the address is not complete. Shipping calculator only updates the City, Country, Postcode fields so it's not visible. If you edit the address in admin, you'll see the populated fields:

Screenshot 2021-12-21 at 13 02 50
Screenshot 2021-12-21 at 13 03 04

opr (approved):

PR #5380, opened 2021-12-14
Check if the current theme has a product-archive block template in the absence of a product taxonomy block template.

### LINKED ISSUE #5230: Legacy template block: themes can't override the archive-product.html only
## Describe the bug
It seems like a theme can only override the archive product template by overriding the taxonomy category/tag templates.

## To reproduce

1. Install Blockbase
2. Create a template inside `block-templates/archive-product.html` with this content:

```
<!-- wp:template-part {"slug":"header","tagName":"header"} /-->

<!-- wp:group {"tagName":"main","layout":{"inherit":true}} -->
<main class="wp-block-group">

	<!-- wp:group {"align":"wide"} -->
	<div class="wp-block-group alignwide">
		<!-- wp:woocommerce/legacy-template {"template":"archive-product"} /-->
	</div>
	<!-- /wp:group -->

</main>
<!-- /wp:group -->

<!-- wp:template-part {"slug":"footer","tagName":"footer","layout":{"inherit":true},"className":"site-footer-container"} /-->
```

3. Go to the frontend and check an archive page, it's loading either taxonomy-product_cat.html or taxonomy-product_tag.html from the plugin instead of the theme template we just created (the one we created is wide aligned).

## Expected behavior
I expected `archive-product.html` being defined in the theme to override the more specific templates from the plugin

* Core version: 5.8.2
* Gutenberg version (if installed): trunk
* WooCommerce version: 6.1.0
* WooCommerce Blocks version: 6.5.0

tjcafferkey:
This change means that if a theme has a `product-archive.html` block template, but not more taxonomy specific ones we will use the `archive-product.html` template for taxonomy frontend _instead of_ the Woo Blocks `taxonomy-product_cat`/`taxonomy-product_tag` files.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5230

### Manual Testing

To make this easier to test, add the following block markup to the following templates so you know which one is getting rendered:

```
<!-- wp:paragraph -->
<p>Woo Blocks: [template-filename].html</p>
<!-- /wp:paragraph -->
```

* `woo-blocks/archive-product.html`
* `woo-blocks/taxonomy-product_tag.html`
* `woo-blocks/taxonomy-product_cat.html`

1. Install Gutenberg plugin and Blockbase theme to activate Site Editor and FSE.
2. Go to a product category page and check that `woo-blocks/taxonomy-product_cat.html` is the template being used.
3. Duplicate `woo-blocks/archive-product.html` into `blockbase/block-templates/` and change the markup you added above to say "Blockbase" instead of "Woo Blocks" so we can differentiate.
4. Reload the product category page, product tag page, and main shop page/product archive page and check that the `archive-product.html` from Blockbase is being loaded.
5. Go into Site Editor > All Templates and check that "Product Archive Page", "Product Category Page" and "Product Tag Page" are all in the list and it states they're being loaded from the Blockbase theme
6. Customise the Product Category Page template and save your changes
7. Reload the product category page on the frontend and check the customised version of this template is being rendered.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

1.
2.
3.

### Changelog

> Use the themes product archive block template for product category & product tag pages if the theme does not have more specific templates for those.


src/Utils/BlockTemplateUtils.php
```diff
@@ -134,25 +134,59 @@ public static function gutenberg_build_template_result_from_post( $post ) {
...
 		// phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
 		$template_content         = file_get_contents( $template_file->path );
 		$template                 = new \WP_Block_Template();
-		$template->id             = 'woocommerce//' . $template_file->slug;
-		$template->theme          = 'woocommerce/woocommerce';
+		$template->id             = $template_is_from_theme ? strtolower( $theme_name ) . '//' . $template_file->slug : 'woocommerce//' . $template_file->slug;
+		$template->theme          = $template_is_from_theme ? $theme_name : 'woocommerce/woocommerce';
 		$template->content        = self::gutenberg_inject_theme_attribute_in_content( $template_content );
-		$template->source         = 'plugin';
+		$template->source         = $template_file->source || 'plugin';
```
Aljullu: According to the [inline docs](https://github.com/WordPress/gutenberg/blob/8eb1f341285362f6a131fcba81a8d054b6b5b5c5/lib/compat/wordpress-5.9/class-gutenberg-block-template.php#L63-L68), `source` can only have two values: `theme` or `custom`. I guess for plugins we should simply use `theme`.

src/Utils/BlockTemplateUtils.php
```diff
@@ -134,25 +134,59 @@ public static function gutenberg_build_template_result_from_post( $post ) {
...
+	public static function create_new_block_template_object( $template_file, $template_type, $template_slug, $template_is_from_theme = false ) {
+		$theme_name = strtolower( wp_get_theme()->get( 'Name' ) );
+
+		$new_template_item = array(
+			'slug'        => $template_slug,
+			'id'          => $template_is_from_theme ? $theme_name . '//' . $template_slug : 'woocommerce//' . $template_slug,
+			'path'        => $template_file,
+			'type'        => $template_type,
+			'theme'       => $template_is_from_theme ? $theme_name : 'woocommerce',
+			'source'      => $template_is_from_theme ? 'theme' : 'plugin',
```
Aljullu, line 182: Same as the previous comment. :slightly_smiling_face:

Aljullu (commented):
Overall this is testing well, thanks for investigating this issue @tjcafferkey!

One thing I noticed is that if I delete the Product Category and Product Tag templates from the theme, they still appear in the Template list as being owned by the theme:

imatge

If I try to edit them, the editor opens the Product Archive template. Can you reproduce it too? Maybe we need to limit the logic from this PR to only happen in the frontend?

tjcafferkey:
>One thing I noticed is that if I delete the Product Category and Product Tag templates from the theme, they still appear in the Template list as being owned by the theme:

>If I try to edit them, the editor opens the Product Archive template. Can you reproduce it too? Maybe we need to limit the logic from this PR to only happen in the frontend?

I did this intentionally. If we limit this PR to on the frontend then:

* It will load the Woo Blocks templates into the Site Editor templates list, but will be loading the Themes Product Archive template on the frontend. This could be very confusing
* Presumably if the user wants to edit their Product Category/Tag templates they will want it to be a variation of the themes Product Archive template not the one from Woo Blocks

I hope this makes sense.

src/Utils/BlockTemplateUtils.php
```diff
@@ -134,25 +134,59 @@ public static function gutenberg_build_template_result_from_post( $post ) {
...
+		$template_is_from_theme = 'theme' === $template_file->source ? true : false;
+		$theme_name             = wp_get_theme()->get( 'Name' );
+
 		// phpcs:ignore WordPress.WP.AlternativeFunctions.file_get_contents_file_get_contents
 		$template_content         = file_get_contents( $template_file->path );
 		$template                 = new \WP_Block_Template();
-		$template->id             = 'woocommerce//' . $template_file->slug;
-		$template->theme          = 'woocommerce/woocommerce';
+		$template->id             = $template_is_from_theme ? strtolower( $theme_name ) . '//' . $template_file->slug : 'woocommerce//' . $template_file->slug;
+		$template->theme          = $template_is_from_theme ? $theme_name : 'woocommerce/woocommerce';
```
Aljullu: I think `theme` should be the theme slug, not the name. Ie, when using Twenty Twenty-Two it makes a Rest API request to an endpoint with a space and capital letters:

imatge

src/Utils/BlockTemplateUtils.php
```diff
@@ -134,25 +134,59 @@ public static function gutenberg_build_template_result_from_post( $post ) {
...
+	 */
+	public static function create_new_block_template_object( $template_file, $template_type, $template_slug, $template_is_from_theme = false ) {
+		$theme_name = strtolower( wp_get_theme()->get( 'Name' ) );
+
+		$new_template_item = array(
+			'slug'        => $template_slug,
+			'id'          => $template_is_from_theme ? $theme_name . '//' . $template_slug : 'woocommerce//' . $template_slug,
+			'path'        => $template_file,
+			'type'        => $template_type,
+			'theme'       => $template_is_from_theme ? $theme_name : 'woocommerce',
```
Aljullu, line 180: It would be the same here.

Aljullu (commented):
> I hope this makes sense.

Ah, right, I tested it again and it makes a lot of sense. Following my previous example, if I edit the Product Tag template, in the editor I see the Product Archive template, but any edit I do will only impact the Product Tag template. I missed that in my first review!

Btw, I left a couple of extra comments that I noticed during this second round of testing. :point_down:

src/Utils/BlockTemplateUtils.php
tjcafferkey: In preparation for templates to be included for plugins we agreed we could use the `plugin` value for `source`. Please see this PR (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5215) and it's related issue where this was discussed and agreed (https://github.com/WordPress/gutenberg/issues/36597#issuecomment-976232909)

src/Utils/BlockTemplateUtils.php
tjcafferkey: In preparation for templates to be included for plugins we agreed we could use the `plugin` value for `source`. Please see this PR (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5215) and it's related issue where this was discussed and agreed (https://github.com/WordPress/gutenberg/issues/36597#issuecomment-976232909)

src/Utils/BlockTemplateUtils.php
tjcafferkey: Good catch Ive used `TextDomain` instead of `Name` instead and this fixes this.

gigitux (approved):
LGTM 🚢

src/Utils/BlockTemplateUtils.php
```diff
@@ -133,24 +133,61 @@ public static function gutenberg_build_template_result_from_post( $post ) {
...
-		$template->id             = 'woocommerce//' . $template_file->slug;
-		$template->theme          = 'WooCommerce';
-		$template->content        = self::gutenberg_inject_theme_attribute_in_content( $template_content );
+		$template_content  = file_get_contents( $template_file->path );
+		$template          = new \WP_Block_Template();
+		$template->id      = $template_is_from_theme ? $theme_name . '//' . $template_file->slug : 'woocommerce//' . $template_file->slug;
+		$template->theme   = $template_is_from_theme ? $theme_name : 'WooCommerce';
+		$template->content = self::gutenberg_inject_theme_attribute_in_content( $template_content );
+		// Plugin was agreed as a valid source value despite existing inline docs at the time of creating: https://github.com/WordPress/gutenberg/issues/36597#issuecomment-976232909.
+		$template->source         = $template_file->source || 'plugin';
```
mikejolley: As noted in Slack, this syntax does not work with PHP and needs to use `?:` or an if statement/ternary.

src/Utils/BlockTemplateUtils.php
```diff
@@ -133,24 +133,61 @@ public static function gutenberg_build_template_result_from_post( $post ) {
 	 */
 	public static function gutenberg_build_template_result_from_file( $template_file, $template_type ) {
 		$template_file = (object) $template_file;
+
+		// If the theme has an archive-products.html template but does not have product taxonomy templates
+		// then we will load in the archive-product.html template from the theme to use for product taxonomies on the frontend.
+		$template_is_from_theme = 'theme' === $template_file->source ? true : false;
```
mikejolley: nit: `'theme' === $template_file->source` evaluates to boolean, so the `true : false` is not needed here.

src/Utils/BlockTemplateUtils.php
Aljullu: Oh, I wasn't aware of that discussion, thanks for linking to it. :+1:

PR #5381, opened 2021-12-14
Remove validateOnMount in favour of validation on state change

mikejolley:
Fixes an issue where the input validation may break if there is a value on page load. Caused in part by a conflict with existing code, state, and https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5373 

To replicate:

- Whilst logged in with customer data,
- Go to checkout. Note that some address fields should be pre-filled.
- Hit place order without interacting with fields
- Errors will show below populated fields

### Testing

- Whilst logged in go to checkout.
- Note that some address fields should be pre-filled with customer data. If they are not, go to admin and fill out profile data for your address then repeat the test.
- Hit place order without interacting with address fields
- No errors should be shown below filled fields.

### DIFF
--- a/assets/js/base/components/text-input/validated-text-input.tsx
+++ b/assets/js/base/components/text-input/validated-text-input.tsx
@@ -2,7 +2,13 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import { useCallback, useRef, useEffect, useState } from 'react';
+import {
+	useCallback,
+	useRef,
+	useEffect,
+	useLayoutEffect,
+	useState,
+} from 'react';
 import classnames from 'classnames';
 import {
 	ValidationInputError,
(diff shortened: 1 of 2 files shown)


mikejolley:
I'm adding blocker label since the issue exists in trunk.

assets/js/base/components/text-input/validated-text-input.tsx
```diff
@@ -96,34 +96,30 @@ const ValidatedTextInput = ( {
...
-				validateInput();
-			}
-			setIsPristine( false );
+		if (
+			inputRef.current?.ownerDocument?.activeElement !== inputRef.current
+		) {
+			validateInput( true );
 		}
-	}, [ isPristine, setIsPristine, validateOnMount, validateInput ] );
+	}, [ value, validateInput ] );
```
senadir, line 123: This might get removed by mistake given it's not used in the use effect, a comment would help protect against that.
```suggestion
		// We need to track value even if it's directly used.
	}, [ value, validateInput ] );
```

senadir (approved):
This is looking and testing good, thank you for digging this one up Mike!

PR #5385, opened 2021-12-15
Make it so WooCommerce template names are not editable

### LINKED ISSUE #5281: FSE: WooCommerce template names are editable after changes
## Describe the bug
The core wordpress template names are not editable. The woocommerce templates shouldn't be either. However if one is to edit a template, save it and reload it, the title of the template becomes editable.

## To reproduce

1. With WC 6.0, Gutenberg (latest trunk) and a block theme installed, go to Appearance > Editor
2. Go to the Templates page and edit one of the WooCommerce templates (ie: Single Product Page).
3. Save and refresh the page
4. The title of the templates is now editable

## Expected behavior
The title of the template should not be editable

## Screenshots
Screenshot 2021-11-30 at 18 08 53

* Core version: 5.8.2
* Gutenberg version (if installed): lates trunk - tested at commit [26556937b70bbf4aaef8c112321703b61a740284](https://github.com/WordPress/gutenberg/commit/26556937b70bbf4aaef8c112321703b61a740284)
* WooCommerce version: 6.0.0
* WooCommerce Blocks version: 6.5.0-dev
* Site language:

## Additional context
Slack conversation here: p1638279389105200-slack-C8X6Q7XQU

Aljullu:
Fixes #5281.

It turns out that [`is_custom`](https://github.com/WordPress/gutenberg/blob/8eb1f341285362f6a131fcba81a8d054b6b5b5c5/lib/compat/wordpress-5.9/class-gutenberg-block-template.php#L100-L105) in a template object means that it's a template created by the user, not a template from the theme/plugin edited by the user. So in WooCommerce template we always have to set it to `false`.

In this PR I also removed this line:
```PHP
$template->source = 'plugin';
```
according to the [comment docs](https://github.com/WordPress/gutenberg/blob/8eb1f341285362f6a131fcba81a8d054b6b5b5c5/lib/compat/wordpress-5.9/class-gutenberg-block-template.php#L63-L68), `$source` can only have two values: `theme` or `custom`. So I think it's better to leave it blank, so it uses the default value (`theme`).

### Manual Testing

1. With WC 6.0 beta 3 and a block theme installed, go to Appearance > Editor.
2. Go to the Templates page and edit one of the WooCommerce templates (ie: Single Product Page).
3. Save and refresh the page.
4. Verify the template name is not editable.

| Before | After |
| --- | --- |
| imatge | imatge |

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Changelog

> Avoid WooCommerce template names being editable.

### DIFF
--- a/src/Utils/BlockTemplateUtils.php
+++ b/src/Utils/BlockTemplateUtils.php
@@ -107,7 +107,7 @@ class BlockTemplateUtils {
 		$template->title          = $post->post_title;
 		$template->status         = $post->post_status;
 		$template->has_theme_file = $has_theme_file;
-		$template->is_custom      = true;
+		$template->is_custom      = false;
 		$template->post_types     = array(); // Don't appear in any Edit Post template selector dropdown.
 		if ( 'wp_template_part' === $post->post_type ) {
 			$type_terms = get_the_terms( $post, 'wp_template_part_area' );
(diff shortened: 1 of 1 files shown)


opr (approved):
Works well for me! 🚢  it

PR #5390, opened 2021-12-15
POC save cart state in data store.

senadir:
POC. 
This saves cart state in data store instead of local state, saves us handling state updates in two places.


alexflorisca:
@opr Wana close this, since you're already steam rolling through this in other PRs?

opr:
> @opr Wana close this, since you're already steam rolling through this in other PRs?

This is interesting, let's leave it open until this project has run a bit more, it may be useful :) Thanks

opr:
I would suggest we focus on this in the "refactor/improve performance" part of the data store migration. The work done here is a good starting point, and I don't think we need to keep local state in the hook either, moving this logic to the data store is great.

cc @ralucaStan - I think this is a good item for potential next cycle work

PR #5394, opened 2021-12-15
Fix extra network request on cart

alexflorisca:
[PR5129](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5129) introduced an extra unnecessary network request to update customer data when the cart is mounted. Initially, the local state `customerData` was out of sync with the cart store. The code removed in this PR was added in the previous PR to keep the local state and the cart in sync. While this would still be nice, it's not necessary and introduces an extra network request.

### Manual Testing

1. Load the cart page
2. There should be no requests (batch or otherwise) to update customer data once the cart is mounted

### Changelog

> Fixed a performance issue with the cart by preventing an extra network request on mount.

### DIFF
--- a/assets/js/base/context/hooks/use-customer-data.ts
+++ b/assets/js/base/context/hooks/use-customer-data.ts
@@ -95,26 +95,6 @@ export const useCustomerData = (): {
 		shippingAddress: initialShippingAddress,
 	} );
 
-	// We only want to update the local state once, otherwise the data on the checkout page gets overwritten
-	// with the initial state of the addresses here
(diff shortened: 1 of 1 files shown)


opr (approved):
This works for me and I've confirmed there's no regression for #5070

🚢 it!

mikejolley:
This PR seems to have regressed the population of the address fields on the first mount (see https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5456#issuecomment-1002825206). I'm investigating now.

PR #5400, opened 2021-12-16
Add correct class to Legacy Template block icon

Aljullu:
We forgot to add the correct class name to this icon. :slightly_smiling_face: 

### Manual Testing

1. With a block theme, go to Appearance > Editor.
2. Modify one of the WooCommerce templates.
3. In the editor, open the List view and select the WooCommerce Legacy Template block.
4. Verify the icon changes to white.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.


tjcafferkey (approved):
Works great, good catch!

PR #5406, opened 2021-12-16
Add minimum quantity, maximum quantity, and step (multiple_of) to the Cart Block and Store API

### LINKED ISSUE #3888: Hide quantity selector if product is sold individually
## What do you like about the Cart & Checkout blocks?

I like that I don't have to code them myself!

## What do you think is missing from the Cart & Checkout blocks?

I'd like to be able to remove elements that aren't relevant to my store. For example, everything we sell is a unique item. The quantity is always 1 and can't be increased, so I'd like to be able to remove the quantity selector.

mikejolley:
Closes https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5037

Implements the following changes:

1. `quantity_limits` returned from the Store API `cart` endpints, which contains `minimum`, `maximum`, and `multiple_of`. There is also an `editable` prop which can be set to false to disable inputs.
2. For products, which are not yet in the cart remember, the product route returns `minimum`, `maximum`, and `multiple_of` as part of the `add_to_cart` param. I've updated the `QuantityInput` component to match and support step, in the same way @senadir added support to `QuantitySelector`.
3. `quantity_limits` are validated by the cart controller when adding an item to the cart, or when updating an item already in the cart. The API returns error 400 if the provided value is out of bounds.
4. ~~Styles the quantity input for stepped (`multiple_of`) quantities so the +/- buttons clearly show what value comes next.~~
5. Hides the quantity selector if the product is sold individually or if `editable` is false.

Fixes #3888

#### Caveats:

This functionality is implemented in the Store API. Therefore, minimum, maximum, and step attributes can be circumvented via core functionality e.g. Shop page. Core offers its own filters to control this.

#### List of hooks

- `woocommerce_store_api_product_quantity_multiple_of`
- `woocommerce_store_api_product_quantity_minimum`
- `woocommerce_store_api_product_quantity_maximum`
- `woocommerce_store_api_product_quantity_editable` (cart only)

#### Developer testing steps:

Smoke test adding items to cart and changing quantity, either via the + and - buttons, or directly via the input.

**Disable Selectors**

Add the following code to disable quantity inputs for cart items:

```php
add_filter( 'woocommerce_store_api_product_quantity_editable', function( $quantity, $product, $cart_item )  {
	return false;
}, 10, 3 );
```

Test that no selectors are shown on the cart page. If you attempt to update a qty via the API for an existing cart item, you will get a 400 error.

**Step Property**

Add the following code to change the multiple/step property for all cart items:

```php
add_filter( 'woocommerce_store_api_product_quantity_multiple_of', function( $quantity, $product, $cart_item )  {
	return 4;
}, 10, 3 );
```

- Refresh your cart, change values, quantity should up and down by 4 (or whatever value you set above).
- If the initial value isn't a multiplier of 4, after changing the quantity, your new quantity would be a multiplier of 4.
- Try to type a number inside the quantity selector, trying something that isn't a multiplier of 4 would change to the nearest one, for example, typing 14 would yield back 12.

**Min/Max for Cart Items**

Add the following code to change the minimum and maximum for all cart items:

```php
add_filter( 'woocommerce_store_api_product_quantity_minimum', function( $quantity, $product, $cart_item )  {
	return 5;
}, 10, 3 );

add_filter( 'woocommerce_store_api_product_quantity_maximum', function( $quantity, $product, $cart_item )  {
	return 30;
}, 10, 3 );
```

- Refresh your cart, change values. You'll be restricted to a qty between 5 and 30.

**Min/Max for Cart Products**

Add the following code to change the minimum qty for products (this affects add to cart events):

```php
add_filter( 'woocommerce_store_api_product_quantity_minimum', function( $quantity, $product, $cart_item )  {
	return 5;
}, 10, 3 );
```

- Use the All Products block to add a product to the cart.
- You should see 5 in cart after adding to the cart the first time.

#### User facing testing steps:

1. Smoke test adding items to cart and changing quantity, either via the + and - buttons, or directly via the input.
2. Edit a product and make it "sold individually"
3. Confirm that no quantity box shows on the cart page for this product
4. Edit a product and set stock to 6, no backorders.
5. Confirm that you can only have 6 maximum for this product.

#### Screenshots

Qty vs qty disabled:

Screenshot 2021-12-21 at 14 39 20
Screenshot 2021-12-21 at 14 39 16

#### Changelog

> Store API and Cart block now support defining a quantity stepper and a minimum quantity.


assets/js/atomic/blocks/product-elements/add-to-cart/shared/quantity-input.js
```diff
@@ -1,25 +1,71 @@
...
+
+		// We round off the value to our steps.
+		newValue = Math.floor( newValue / step ) * step;
+
+		// Only commit if the value has changed
+		if ( newValue !== initialValue ) {
+			onChange( newValue );
+		}
+	}, 300 );
+
```
senadir, line 55: I think we used this in a lot of places that it might be worth abstracting up?

assets/js/atomic/blocks/product-elements/add-to-cart/product-types/simple.js
```diff
@@ -43,6 +44,7 @@ const Simple = () => {
 				value={ quantity }
 				min={ minQuantity }
 				max={ maxQuantity }
+				step={ multipleOf }
```
senadir, line 47: Thank you for thinking of the ~~children~~ element blocks!

assets/js/blocks/cart-checkout/cart/cart-line-items-table/cart-line-item-row.tsx
```diff
@@ -70,7 +70,13 @@ const CartLineItemRow = forwardRef< HTMLTableRowElement, CartLineItemRowProps >(
...
 			low_stock_remaining: lowStockRemaining = null,
 			show_backorder_badge: showBackorderBadge = false,
-			quantity_limit: quantityLimit = 99,
+			quantity_limits: quantityLimits = {
+				minimum: 1,
+				maximum: 99,
+				multiple_of: 1,
+				editable: true,
+			},
+			sold_individually: soldIndividually = false,
```
senadir, line 79: I guess we can't remove `sold_individually` yet? its job can be fullfiled by what's in `editable`.

src/StoreApi/Schemas/CartItemSchema.php
```diff
@@ -363,25 +391,6 @@ protected function prepare_product_price_response( \WC_Product $product, $tax_di
...
-		if ( is_null( $product->get_stock_quantity() ) ) {
-			return null;
-		}
-
-		$reserve_stock  = new ReserveStock();
-		$reserved_stock = $reserve_stock->get_reserved_stock( $product, $this->get_draft_order_id() );
-
-		return $product->get_stock_quantity() - $reserved_stock;
-	}
-
```
senadir, line 384: Was this just sitting here not used?

src/StoreApi/Utilities/QuantityLimits.php
```diff
@@ -0,0 +1,211 @@
...
+		/**
+		 * Filters the quantity limit for a product being added to the cart via the Store API.
+		 *
+		 * Filters the variation option name for custom option slugs.
+		 *
+		 * @param integer $quantity_limit Quantity limit which defaults to 99 unless sold individually.
+		 * @param \WC_Product $product Product instance.
+		 * @return integer
+		 */
+		return apply_filters( 'woocommerce_store_api_product_quantity_limit', max( min( array_filter( $limits ) ), 1 ), $product );
```
senadir, line 163: I don't get `min( array_filter( $limits ) )`. Are you casting the filter value?

senadir (approved):
Thank you for working on this Mike! absolutely great effort. I have some comments, UX wise only, @vivialice is also going to add a comment here later regarding design. Feel free to merge after you handle my review, even if don't come to the same conclusion, I trust you have a better vision than me.

- When an item quantity doesn't match its step, you get this weird edge case (limit is 4, quantity was 2):
image.
Once I clicked 6, the new quantity was 4, bound by 0 and 8. 0 didn't make a lot of sense to me
image

It's hard to know if the number is disabled or enabled.
Should we try to correct quantity to multipleOf so in the case above, the item quantity is increased to 4 on load.
I know you mentioned that
> If the initial value isn't a multiplier of 4, after changing the quantity, your new quantity would be a multiplier of 4.

But I think we should correct it nevertheless? I managed to checkout no problem with that quantity.

The same is for limits, should the quantity be corrected to min or max in store API before returning? or would this complicate things?


For adding to cart, for items that already existed in cart and their quantity was less the minimum, I got an error:
Min is 5. I'd assume adding to cart would correct that to 5?
image
image

I know that most of those limitation are only showing given I'm adding filters to the fly while I have items in cart. I believe we should support this nevertheless, given a lot of limits are going to be conditional which mean they would kick in even if an item is already in cart.

mikejolley:
Since this has some follow up work Im punting to next release.

vivialice:
Heya @mikejolley ! I just have a comment about the stepper UI. It seems like it should be a selector instead to me, like this: 

selector
 
Is this stepped solution displayed on the product page as well?

mikejolley:
@vivialice no, it's just on the cart right now but I think they will be unified eventually. 

A dropdown would force us to populate all possible options. That's fine for small numbers, but flawed for lots of steps or when there is no upper limit. The step attribute we're using comes from number inputs [ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/step).

I'm fine with putting the + / - back, but this will need a design solution because it's not clear visually what the + and - do when step is greater than 1.

vivialice:
Ahh yes that's true. The original + / - is probably the better solution overall. 

> this will need a design solution because it's not clear visually what the + and - do when step is greater than 1.

Do you mean it might be unclear it increases in multiples?

mikejolley:
> Do you mean it might be unclear it increases in multiples?

@vivialice yes, because the UI for all + - buttons look the same, but the qty increment or decrement may differ.

src/StoreApi/Utilities/QuantityLimits.php
mikejolley: I didn't change this https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7329935b1dce1c8bab77b213ffb2742e088d3b13/src/StoreApi/Schemas/ProductSchema.php#L539 It's getting the min value from all possible limits, and then using max() to ensure it's greater than 1.

src/StoreApi/Schemas/CartItemSchema.php
mikejolley: It was used, but some other code moved and this moved with it.

assets/js/blocks/cart-checkout/cart/cart-line-items-table/cart-line-item-row.tsx
mikejolley: I didn't want to remove existing props from the API in case flaws were found.

mikejolley:
I've [removed the custom step indicator ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5406/commits/23a866981c6d19cec451256c154b2a434be5052c)so now all increments use + / - with no indication of qty. 



> When an item quantity doesn't match its step, you get this weird edge case (limit is 4, quantity was 2):

[I've mitigated this by running normalize on mount. ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5406/commits/1fae9a9a468c1f045f9c53c8c1a1d8276a07c3a4)

> The same is for limits, should the quantity be corrected to min or max in store API before returning? or would this complicate things?

More difficult server side because we'd have to update cart items, but I think it's unlikely to change randomly like we're doing in testing, so we can leave for the normalization function in the client.

> For adding to cart, for items that already existed in cart and their quantity was less the minimum, I got an error:
Min is 5. I'd assume adding to cart would correct that to 5?

I didn't account for the case of min changing unexpectedly. For items not in the cart, 5 would be the inital number added. I think it might be useful to support but as it's edge case I won't work on this yet.

assets/js/atomic/blocks/product-elements/add-to-cart/shared/quantity-input.js
mikejolley: I think we should use the same component in both places (follow up) rather than abstracting this one function.

ralucaStan:
does `woocommerce_store_api_product_quantity_editable` work for the Mini cart block as well?

mikejolley:
@ralucaStan Yes, I just confirmed so. The Mini Cart Block uses the same qty component, and same endpoints, so yes, these new props are respected there too.

Issue #5399, opened 2021-12-16
FSE - Template: Template changes are not saved


gigitux:
## Describe the bug
If you save a `WooCommerce` template and refresh the page, you don't see any changes.

## To reproduce

Be sure that you have: `WordPress 5.9`, `Gutenberg trunk version`, `WooCommerce 6.0.0`, and `WooCommerce Blocks trunk` 

1. Go to `Template Page` on `Full Site Editor`.
2. Click on one of the `WooCommerce template` like `Single Product Page`.
3. Add a new block (be sure that you don't add it on the header or footer group).
4. Save.
5. Refresh, and you can check that your changes are missing.

## Expected behavior
The changes should be visible

## Info

It seems that the system saves a new post in the DB and doesn't update the old one. This bug appears after #5375 merged.

## Screenshots

https://user-images.githubusercontent.com/4463174/146355581-98f54bd8-ddf4-4889-9624-cb418d743419.mp4

### WordPress (please complete the following information):
* Core version: `5.9-beta-3`
* Gutenberg version (if installed): `trunk`
* WooCommerce version: [e.g. 3.5.3]:` 6.0.0`
* WooCommerce Blocks version: [e.g. 1.3.1]: `trunk`

gigitux:
This issue is fixed with #5412

PR #5408, opened 2021-12-17
Fix saving WooCommerce templates in WP 5.9 beta 3

### LINKED ISSUE #5402: Templates can't be saved in WP 5.9 beta 3 with Gutenberg disabled
## To reproduce
1. With WP 5.9 beta 3 and Gutenberg disabled, go to Appearance > Editor.
2. Edit one of the WooCommerce templates and try to save it.
3. Notice you get a 404 error: `No templates exist with that id.`.

The problem seems to be that we [check for the existence of `gutenberg_get_block_template()`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/64246a7a651a84b25fa4ed649db8371be7fb8e42/src/BlockTemplatesController.php#L75), however that function was renamed to `get_block_template()` when included to WP core, so if Gutenberg is not enabled that function always returns early.

Aljullu:
Fixes #5402.

### Manual Testing

1. With WP 5.9 beta 3 and Gutenberg disabled, go to Appearance > Editor.
2. Edit one of the WooCommerce templates and try to save it.
3. Verify the template is saved without errors.
4. Repeat the process with Gutenberg enabled and verify it works as well.

Currently, step 2 only works for templates that you have already edited because of #5399.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -72,7 +72,13 @@ class BlockTemplatesController {
 	 * @return mixed|\WP_Block_Template|\WP_Error
 	 */
 	public function maybe_return_blocks_template( $template, $id, $template_type ) {
-		if ( ! function_exists( 'gutenberg_get_block_template' ) ) {
+		// 'get_block_template' was introduced in WP 5.9. We need to support
+		// 'gutenberg_get_block_template' for previous versions of WP with
+		// Gutenberg enabled.
+		if (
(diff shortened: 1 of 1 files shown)


gigitux (approved):
LGTM 🚢

gigitux:
@Aljullu I'm merging this because it is necessary for the 6.5.1 release

PR #5421, opened 2021-12-20
Fix: Mini Cart: Default content for empty state

### LINKED ISSUE #5377: Mini Cart: Default content for empty state
The content of the empty mini cart view is different from the design.

The current design for the empty view:
Screen Shot 2021-12-14 at 10 01 04

dinhtungdu:
Fixes #5377 

This PR updates the default empty view of the Mini Cart Contents blocks and fixes some minor styling issues.

### Screenshots

image

### Manual Testing

1. Make sure the active block theme doesn't have a `mini-cart` template part.
2. Delete all customizations of the Mini Cart template part.
3. On the front-end, remove all products from the cart.
4. Open the Mini Cart, see the empty view that matches with the screenshot above.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss
@@ -1,6 +1,8 @@
 .wp-block-woocommerce-mini-cart-contents {
 	max-width: 480px;
-	margin: 0 auto;
+	/* We need to override the margin top here to simulate the layout of
+	the mini cart contents on the front end. */
+	margin: 0 auto !important;
 }
 
 .wp-block-woocommerce-filled-mini-cart-contents-block {
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss
@@ -1,13 +1,13 @@
 .wp-block-woocommerce-mini-cart-contents {
 	max-width: 480px;
-	margin: 0 auto;
+	margin: 0 auto !important;
 }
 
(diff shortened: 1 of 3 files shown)


assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss
```diff
@@ -1,13 +1,13 @@
 .wp-block-woocommerce-mini-cart-contents {
 	max-width: 480px;
-	margin: 0 auto;
+	margin: 0 auto !important;
```
gigitux, line 5: Maybe is it better to add a comment that explains why we are using `!important`?

gigitux (approved):
Everything works! 

just a small consideration, but LGTM 👍

assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss
dinhtungdu: Actually we don't need this with changes introduced in 5d6818e

dinhtungdu: Turned out that we need it. I added the comment to clarify the reason.

gigitux: Mmm, are you sure about this?

In my local env, I tried and I see a different result with and without `!important`

image

*with !important

image
*without !important

Check the scrollbar on the right

gigitux: Oh, I see that you noticed the difference and add a comment. For me, now it's perfect 🚢

PR #5433, opened 2021-12-21
Add wide and full alignment support for legacy template block

### LINKED ISSUE #5229: Legacy template Block: add alignment support
To avoid needing extra wrappers, it would be useful to have alignment supports on the legacy templates, so the themes could implement something like this:

`<!-- wp:woocommerce/legacy-template {"template":"single-product","align":"wide"} /-->`

Wide and full-width support would be enough for this block, I believe

gigitux:
This PR adds support for the wide and full alignment for the legacy template block.

Fixes #5229 

### Screenshots

image
*editor side
image
*frontend side

### Manual Testing

Check out this branch

1. Open FSE editor.
2. Edit a legacy template (for example `Product Category Page`).
3. Check if you can change the alignment. 
4. Change it.
5. Check if the changes are also reflected on the frontend side.

You should test at least one between `Product Category Page`, `Product Archive Page` and `Product Tag Page`. 
You should test `Single Product Page`.

### Changelog

> Add support for the wide and full alignment for the legacy template block.


src/Utils/StyleAttributesUtils.php
```diff
@@ -154,6 +154,59 @@ public static function get_background_color_class_and_style( $attributes ) {
...
+	/**
+	 * Get class and style for align from attributes.
+	 *
+	 * @param array $attributes Block attributes.
+	 *
+	 * @return (array | null)
+	 */
+	public static function get_align_class_and_style( $attributes ) {
+
+		$align_attribute = $attributes['align'];
```
dinhtungdu: `align` may not exist in `$attributes` so we should check for its existence first.

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +184,22 @@ protected function render_archive_product() {
...
+	 * @return string Rendered block type output.
+	 */
+	public function get_markup_with_classes_by_attributes( string $content, array $block ) {
+		$pattern     = '/(?<=class=\")[^"]+(?=\")/';
+		$attributes  = (array) $block['attrs'];
+		$align_class = StyleAttributesUtils::get_align_class_and_style( $attributes );
+		$matches     = array();
+		preg_match( $pattern, $content, $matches );
+
+		return preg_replace( $pattern, $matches[0] . ' ' . $align_class['class'], $content, 1 );
```
dinhtungdu: `$align_class` may be `null`, so we need another check before using it as an array.

src/BlockTypes/LegacyTemplate.php
```diff
@@ -99,6 +102,7 @@ protected function render_single_product() {
 	 * @return string Rendered block type output.
 	 */
 	protected function render_archive_product() {
+		add_filter( 'render_block', array( $this, 'get_markup_with_classes_by_attributes' ), 10, 2 );
```
dinhtungdu: What do you think about calling this `add_filter` once in the `initialize()` method? I mean overriding the `AbstractBlock::initialize()` and adding this filter.

src/Utils/StyleAttributesUtils.php
```diff
@@ -154,6 +154,59 @@ public static function get_background_color_class_and_style( $attributes ) {
...
 	}
 
+	/**
+	 * Get class and style for align from attributes.
+	 *
+	 * @param array $attributes Block attributes.
+	 *
+	 * @return (array | null)
+	 */
+	public static function get_align_class_and_style( $attributes ) {
```
dinhtungdu, line 164: I don't think that we will need style for alignment. And even if we need it, we can add a new utility in the future. For now, looks like we only need the class, so I think we should create the utility to return only the class as a string.

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +184,22 @@ protected function render_archive_product() {
...
+
+	/**
+	 * Get HTML markup with the right classes by attributes.
+	 *
+	 * @param string $content Block content.
+	 * @param array  $block Parsed block data.
+	 * @return string Rendered block type output.
+	 */
+	public function get_markup_with_classes_by_attributes( string $content, array $block ) {
+		$pattern     = '/(?<=class=\")[^"]+(?=\")/';
```
dinhtungdu: The `global/wrapper-start.php` can be overridden in the theme, so the wrapper may not have the `class` attribute. Although I never see any wrapper element that doesn't have class, it's still a valid case. We don't need to cover that case in our logic, but a note for the theme developer will be nice here.

dinhtungdu: For example, I modified the markup to simulate a `wrapper-start.php` template overridden by the theme.

Screen Shot 2021-12-22 at 11 25 37

dinhtungdu (commented):
@gigitux The alignment is working on my test 👏🏽 . I got several warnings testing this PR due to the new utility added in this PR, please check my comments for more detail.

src/BlockTypes/LegacyTemplate.php
gigitux: > The global/wrapper-start.php can be overridden in the theme, so the wrapper may not have the class attribute. 

Great catch! I don't think about this case.

>but a note for the theme developer will be nice here.

Do you mean to add a comment or what else?

src/Utils/StyleAttributesUtils.php
gigitux: My idea for this class is to return always an array with class and style because I would really to have only one function for each style. This way we can easily extend the functionality in the future. What do you think?

gigitux:
Thanks for your great comments! I just update the PR, but I don't understand what do you mean with
> but a note for the theme developer will be nice here.

src/BlockTypes/LegacyTemplate.php
dinhtungdu: > We don't need to cover that case in our logic

I want to clarify on this. I meant that we may not need to address that issue in our logic if it takes too much effort because it's a rare case. Personally, I never see any wrapper that doesn't have class. But if you can solve it, go ahead ; ).

> Do you mean to add a comment or what else?

I was thinking about adding a comment to your callback `get_markup_with_classes_by_attributes`.

src/Utils/StyleAttributesUtils.php
dinhtungdu: Make sense!

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +191,26 @@ protected function render_archive_product() {
...
 	}
+
+	/**
+	 * Get HTML markup with the right classes by attributes.
+	 *
+	 * @param string $content Block content.
+	 * @param array  $block Parsed block data.
+	 * @return string Rendered block type output.
+	 */
+	public function get_markup_with_classes_by_attributes( string $content, array $block ) {
```
dinhtungdu: This callback should run only for the `LegacyTemplate` block. Currently, it runs for every block.

src/BlockTypes/LegacyTemplate.php
dinhtungdu: I'm still getting a warning on this line:

```
Warning: Trying to access array offset on value of type null in /Users/tenup/Sites/woofse/app/public/wp-content/plugins/woocommerce-gutenberg-products-block/src/BlockTypes/LegacyTemplate.php on line 213
```

src/BlockTypes/LegacyTemplate.php
gigitux: So do you suggest adding an if condition to check if is it a `LegacyTemplate`?

src/BlockTypes/LegacyTemplate.php
gigitux: Mmm, it is very strange, I can't reproduce the warning

src/BlockTypes/LegacyTemplate.php
dinhtungdu: Yes.

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +195,43 @@ protected function render_archive_product() {
...
+	 * Based on the experience, all the wrapper elements have a class attribute.
+	 *
+	 * @param string $content Block content.
+	 * @param array  $block Parsed block data.
+	 * @return string Rendered block type output.
+	 */
+	public function get_markup_with_classes_by_attributes( string $content, array $block ) {
+		if ( ! $this->is_legacy_template( $block ) ) {
+			return $content;
+		}
```
dinhtungdu: I think we can check for `blockName` key of the `$block` here instead of creating a new helper method (which checks for the `template` attribute).

Screen Shot 2021-12-24 at 11 42 29

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +195,43 @@ protected function render_archive_product() {
...
+	/**
+	 * Get HTML markup with the right classes by attributes.
+	 * This function appends the classname at the first element that have the class attribute.
+	 * Based on the experience, all the wrapper elements have a class attribute.
+	 *
+	 * @param string $content Block content.
+	 * @param array  $block Parsed block data.
+	 * @return string Rendered block type output.
+	 */
+	public function get_markup_with_classes_by_attributes( string $content, array $block ) {
```
dinhtungdu: I think this should be renamed to `add_alignment_class_to_wrapper` for better expressing what the method does.

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +195,43 @@ protected function render_archive_product() {
...
+		if ( ! $this->is_legacy_template( $block ) ) {
+			return $content;
+		}
+
+		$pattern               = '/(?<=class=\")[^"]+(?=\")/';
```
dinhtungdu: Looking at the `BlockTypesController::add_data_attributes`, I notice the ways we target the element there and in this method are different:

- In `BlockTypesController::add_data_attributes`, we target only the openning `<div>`. This may not work for `LegacyTemplate` with themes that use `<section>` as the wrapper tag.
- Here, we target the first element (any element, not limited to `<div>`) that contain the `class` attribute.

IMO, we need to use the same mechanism to target wrapper elements in both places. I'm tagging @mikejolley here for more insights because he originally created `BlockTypesController::add_data_attributes`.

gigitux: Ok, it makes sense!
I agree with you. This regex is not great :D
Let's wait for @mikejolley feedback!

mikejolley: In the case of `add_data_attributes` we know we're dealing with a `div` element because thats how the block markup is saved to the post. If the opening element is unknown, your regex is probably better, but I wonder if you can still target only the opening element rather than the first class found which could be deeper? I am not going to pretend to be regex expert, because I am not :D

gigitux: > but I wonder if you can still target only the opening element rather than the first class found which could be deeper? I am not going to pretend to be regex expert, because I am not :D

It can be a good idea, but I'm pretty sure that the regex will be hard to write and read. I'm not a fan of complex regex, but if you think that it makes sense to work on it, I can do it!

mikejolley: It might not be desirable to add that class deeper in case it breaks other parts of the layout 👍🏻 

Are there any alternatives considered for this additional class inserted by regex? Would it be a disaster to wrap in another div with the class, or does that not work here?

gigitux: > Are there any alternatives considered for this additional class inserted by regex? Would it be a disaster to wrap in another div with the class, or does that not work here?

It works without problem, but I don't know if we can wrap a block in another div. If we can follow this path, I think that it is the best solution.

image

dinhtungdu: > works without problem, but I don't know if we can wrap a block in another div. If we can follow this path, I think that it is the best solution.

I think wrapping inside another element can make it harder (and confusing) for themes developers to style the page because the opening element now is not the first element in the `wrapper-start.php` but the alignment `div`.

gigitux: With [f4689e6](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5433/commits/f4689e61b78baea2d005b697431e732c811a77ed), I implemented a series of regexes.

First of all, I pick the first tag HTML, after that:
- if the first tag already has the class attribute, I just add the new class.
- if the first tag doesn't have the class attribute, I add the class attribute + the new class

Let me know what do you think :D

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +190,47 @@ protected function render_archive_product() {
...
+			return $content;
+		}
+
+		// Find the first tag.
+		$first_tag = '<[^<>]+>';
+		$matches   = array();
+		preg_match( $first_tag, $content, $matches );
+
+		// If there is a tag, but it doesn't have a class attribute, add the class attribute.
+		if ( isset( $matches[0] ) && strpos( $matches[0], 'class' ) === false ) {
```
dinhtungdu: We should check for the exact class attribute here. Elements can have `class` as a part of attribute name or attribute value, like:  `data-class`, `data-type="class"`, `id="fist-class"`.
```suggestion
		if ( isset( $matches[0] ) && strpos( $matches[0], ' class=' ) === false ) {
```

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +190,47 @@ protected function render_archive_product() {
...
+			$pattern_before_tag_closing = '/.+?(?=>)/';
+			$matches                    = array();
+			preg_match( $pattern_before_tag_closing, $content, $matches );
+			return preg_replace( $pattern_before_tag_closing, $matches[0] . ' class="' . $align_class_and_style['class'] . '"', $content, 1 );
```
dinhtungdu: Can we use backreferences in [`preg_replace`](https://www.php.net/manual/en/function.preg-replace.php)? If doing so, we can remove the `preg_match` above.

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +190,47 @@ protected function render_archive_product() {
...
+		}
+
+		// If there is a tag, and it has a class already, add the class attribute.
+		$pattern_get_class = '/(?<=class=\")[^"]+(?=\")/';
```
dinhtungdu: If the wrapper element uses single quotes for class, this regex will fail.

dinhtungdu (commented):
@gigitux The updates look good. I have some concerns about the regex, please check the review comments for more detail!

src/BlockTypes/LegacyTemplate.php
gigitux: I don't think with the current regex.
Because the value of `match[0]` is a tag element without `<` and `>`, for example:
```
div data-block-name="woocommerce/legacy-template" data-template="taxonomy-product_cat" id="primary" 
```

But the regex `/.+?(?=>)/` match the text before a tag closing `>`.

Potentially we can use `str_replace`, in this way:

```			
return str_replace( $matches[0], $matches[0] . ' class="' . $align_class_and_style['class'] . '"', $content );
```
instead to use regex.

I don't have a strong opinion, what do you think?

gigitux:
Thanks for your great feedback!

I updated the regex. There is just a question, let me know, and I will update the PR!

src/BlockTypes/LegacyTemplate.php
dinhtungdu: > Because the value of `match[0]` is a tag element without `<` and `>`, for example:

@gigitux I think we can use the fourth argument of `preg_replace` to only replace one time, you can check my example using your regex below:

```php
<?php
$re = '/.+?(?=>)/m';
$str = '<div data-block-name="woocommerce/legacy-template" data-template="taxonomy-product_cat" id="primary" ><div class="inner">';
$subst = '$0 class="alignwide"';

$result = preg_replace($re, $subst, $str, 1);

echo "The result of the substitution is <pre>" . htmlspecialchars( $result ) . "</pre>";
```

gigitux: Thanks for sharing this!

src/BlockTypes/LegacyTemplate.php
```diff
@@ -180,4 +190,45 @@ protected function render_archive_product() {
...
+		if ( isset( $matches[0] ) && strpos( $matches[0], ' class=' ) === false ) {
+			$pattern_before_tag_closing = '/.+?(?=>)/';
+			return preg_replace( $pattern_before_tag_closing, '$0 class="' . $align_class_and_style['class'] . '"', $content, 1 );
+		}
+
+		// If there is a tag, and it has a class already, add the class attribute.
+		$pattern_get_class = '/(?<=class=\"|\')[^"|\']+(?=\"|\')/';
+		$matches_class     = array();
+		preg_match( $pattern_get_class, $content, $matches_class );
+		return preg_replace( $pattern_get_class, $matches_class[0] . ' ' . $align_class_and_style['class'], $content, 1 );
```
dinhtungdu: I think we can follow the same approach as we use at line 223rd above.

dinhtungdu (approved):
@gigitux This is working like a charm 🎉 ! I've just left one comment for coding style consistency, otherwise, this is ready to be merged!

Issue #5388, opened 2021-12-15
Checkout Block ignores woocommerce_available_payment_gateways hook


chrismademe:
## Describe the bug
When using the WooCommerce Checkout block, the `woocommerce_available_payment_gateways` hook is not respected and so any payment gateways added/removed here are not reflected at checkout.

## To reproduce

1. Enable a payment gateway in WooCommerce settings, bank transfer for example.
2. Add this code to your theme's `functions.php`:

```php
<?php
add_filter( 'woocommerce_available_payment_gateways', function( $methods ) {

    // Remove BACS for normal customers
    if ( ! current_user_can( 'manage_woocommerce' ) ) {
        unset($methods['bacs']);
    }

    return $methods;
} );
```

3. Make sure you're logged out, add a product to your cart and head to the checkout
4. BACS will be available, dispite being removed in a filter

## Expected behavior
The available gateways should be the same as those that appear in the classic shortcode checkout, having been filtered.

### WordPress (please complete the following information):
* Core version: 5.8.2
* WooCommerce version: 5.9.0
* WooCommerce Blocks version: 6.5.0
* Site language: en_US

### Desktop (please complete the following information):
* OS: macOS
* Browser: Chrome
* Version: Latest

Thanks!
Chris

mikejolley:
For anyone reviewing this issue, this is not a straightforward swap to `get_available_payment_gateways`. The example above is simplistic and doesn't factor in more advanced use cases where payment method availability could be based on cart items, customer, address, etc etc. Therefore the ideal solution would be to return available methods via the API, potentially on both the `cart` and `checkout` routes. This can then be used to render payment options on checkout.

chrismademe:
Thanks for the response. I appreciate it's probably not as straight forward as it seems.
With that said, is there currently a way to filter available methods on the frontend?

ralucaStan:
> is there currently a way to filter available methods on the frontend? 

@chrismademe take a look at [the filtering payment gateways section](https://developer.woocommerce.com/2021/11/09/available-extensibility-interfaces-for-the-cart-and-checkout-blocks#filtering-payment-methods) in the _Available Extensibility Interfaces for The Cart and Checkout Blocks_ post, you might find what you need.

chrismademe:
@ralucaStan I will do, thank you! :)

PR #5442, opened 2021-12-22
Experiment: Add available_payment_methods to the API and cart data

mikejolley:
This attempts to solve the problem in #5388 by adding `available_payment_methods` to the cart response data with the current list of available payment method IDs.

Payment methods are registered if `enabled` in settings. This enqueues the scripts for Blocks. 

This script is not aware of the `woocommerce_available_payment_gateways` filter, so to handle this we need to check if a method is available via its `canMakePayment` callback. 

So `canMakePayment` knows which methods are available, I've updated the cart data to include a list of available payment method IDs which are surfaced to gateways.

### Testing

Use a snippet like this:

```
add_filter(
	'woocommerce_available_payment_gateways',
	function( $methods ) {
		if ( wc()->cart->get_total( 'edit' ) < 100 ) {
			unset( $methods['bacs'] );
		}
		return $methods;
	}
);
```

Make a free shipping and flat rate shipping method costing $50. Add $60 of items to the cart. Toggle between free shipping and flat rate shipping so the totals change. BACS should only be available when free shipping is checked.

### DIFF
--- a/assets/js/base/context/hooks/cart/use-store-cart.ts
+++ b/assets/js/base/context/hooks/cart/use-store-cart.ts
@@ -14,6 +14,7 @@ import {
 	EMPTY_CART_ERRORS,
 	EMPTY_SHIPPING_RATES,
 	EMPTY_TAX_LINES,
+	EMPTY_PAYMENT_METHODS,
 	EMPTY_PAYMENT_REQUIREMENTS,
 	EMPTY_EXTENSIONS,
 } from '@woocommerce/block-data';
@@ -113,6 +114,7 @@ export const defaultCartData: StoreCart = {
 	shippingRates: EMPTY_SHIPPING_RATES,
 	shippingRatesLoading: false,
 	cartHasCalculatedShipping: false,
+	availablePaymentMethods: EMPTY_PAYMENT_METHODS,
 	paymentRequirements: EMPTY_PAYMENT_REQUIREMENTS,
 	receiveCart: () => undefined,
 	extensions: EMPTY_EXTENSIONS,
@@ -169,6 +171,8 @@ export const useStoreCart = (
 					shippingRatesLoading: false,
 					cartHasCalculatedShipping:
 						previewCart.has_calculated_shipping,
+					availablePaymentMethods:
+						previewCart.availablePaymentMethods,
 					paymentRequirements: previewCart.paymentRequirements,
 					receiveCart:
 						typeof previewCart?.receiveCart === 'function'
@@ -229,6 +233,7 @@ export const useStoreCart = (
 				shippingRates: cartData.shippingRates,
 				shippingRatesLoading,
 				cartHasCalculatedShipping: cartData.hasCalculatedShipping,
+				availablePaymentMethods: cartData.availablePaymentMethods,
 				paymentRequirements: cartData.paymentRequirements,
 				receiveCart,
 			};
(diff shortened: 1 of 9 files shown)


senadir:
I actually wrote a reply for `5442` that got lost :/ the gist is that:
- we want to fix adjacent issues like `woocommerce_available_payment_gateways` not being respected in Store API (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5405, https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5404)...
- In the client, filtering should be done via JS and available extensibility methods that we have, I'm afraid having two API (PHP and JS) doing the same thing might cause confusion and a subpar experience. We already have client side filtering https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/filtering-payment-methods.md

mikejolley:
@senadir just to be clear, are you saying we need to remove support for the filtering on the PHP side (that this PR fixes)?

senadir:
I'm saying that this PR (along side `availablePaymentMethods`) is not needed.
Developers who don't want to adopt JS filtering can still continue to filter out using. PHP, the payment method would still be visible but the order cannot be placed.

mikejolley:
@senadir 

> I'm afraid having two API (PHP and JS) doing the same thing might cause confusion and a subpar experience.

For us yes.

If we don't support `woocommerce_available_payment_gateways` in the client, developers who need to filter things out would need to do the same thing in PHP and JS, which might be tricky if the data used on the PHP side is not available in the client. Do you see any issues with this?

Where does this leave us with issue 5388?

senadir:
> If we don't support woocommerce_available_payment_gateways in the client, developers who need to filter things out would need to do the same thing in PHP and JS, which might be tricky if the data used on the PHP side is not available in the client. Do you see any issues with this?

I also see the issue of filtering using data that's only available in the frontend, which might not be a big issue since the data will get synced back to the client.

You raise a good point, which is probably a good time for us to reconsider our filtering approach, assuming we go with this, you now can filter payments using:
- `canMakePayment` JS filter.
- `woocommerce_available_payment_gateways` PHP filter.
- `payment_requirements` PHP filter.

Which is a bit excessive and at least 1 method too much.

@ralucaStan how do you feel about the `canMakePayment`, how does it fit within this system, and if we now expose `woocommerce_available_payment_gateways` on the frontend, can we deprecate the `canMakePayment` filter?

@mikejolley the main thing that's preventing me from going with `availablePaymentMethods` is that we're overpopulating the cart response again, with checkout related data.

> Where does this leave us with issue 5388?

Assuming we continue with not exposing and filtering available payment methods on the frontend, we can just close that with a response that we use frontend filters.

ralucaStan:
> @ralucaStan how do you feel about the canMakePayment, how does it fit within this system, and if we now expose woocommerce_available_payment_gateways on the frontend, can we deprecate the canMakePayment filter?

canMakePayment is actually the place where we call the callbacks that extensions might [register with registerPaymentMethodExtensionCallbacks](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/filtering-payment-methods.md#filtering-payment-methods-in-the-checkout-block) to filter payment methods, so removing it is not really an option.

What do you thing that instead of adding a new `availablePaymentMethods` field in the response we add a `isAvailable` field per payment gateways. This would reflect the fact that a PHP filter was used to remove that specific payment gateway. 

If this availability is something we want to take into consideration we could also do a general check in the [canMakePayment function](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/e84577d0065a16d1af79e9bdee07011839deb07f/assets/js/blocks-registry/payment-methods/payment-method-config.tsx#L66-L72) just by checking that `paymentGateways.isAvailable === true`, instead in doing the check per payment gateway (like [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5442/files#diff-58537b6deb139574a60eae45a0f87465586d226792faed08ff115034f7be1a8bR48-R50)).

PR #5444, opened 2021-12-22
Don't show shipping price when no options are available

### LINKED ISSUE #5378: Cart block: confusing info when no shipping options are available
## Describe the bug
When no shipping options are available for the given address, WooCommerce Cart Block still displays shipping cost as $0. This adds visual noise and can be confusing to some customers.

## To reproduce

1. Go to **WooCommerce > Settings > General**
2. Under **Shipping locations** select **Ship to specific countries only** and choose US underneath.
3. Add a shippable product to cart.
4. Proceed to checkout.
5. On the checkout page, enter an address outside of the US (for example: test street, Warszawa, 00-000 Poland)
6. Go back to the Cart page and observe it says `Shipping $0 via` above the details section where the error message can be found.

## Expected behavior
Only display an error message without price or the word `via` if no shipping options are available, similar to how it works in WooCommerce core cart shortcode.

## Screenshots
Settings:
image

Cart block:
image

Cart shortcode for comparison:

image

* Core version: 5.8.2
* Gutenberg version (if installed): ---
* WooCommerce version: 5.9.0, 6.0.0 beta
* WooCommerce Blocks version: 6.5.0
* Site language: English
* Any other plugins installed: ---

* OS: OS X 10.15.7
* Browser Chrome 
* Version 96

opr:
This PR will add a prop called `isCheckout` to the `NoShippingPlaceholder` and `TotalsShipping` components. This will allow us to determine whether the component is being rendered as part of the Checkout block or the Cart block.

It also adds a check to the `TotalsShipping` component to check if there are any rates available for the order. If there are no rates then the price should not be displayed.

Following the change to hide the price if there are no rates, the component would show `Calculated at checkout` if no rates were available which is confusing when the customer is actually on the Checkout block. Instead I added a new error message to be shown instead.

Fixes #5378

### Screenshots

| Before | After |
|---|---|
| image | image |

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Go to WooCommerce > Settings > Shipping and remove all shipping methods, including shipping methods for `Locations not covered by your other zones`.
2. Add a shipping zone for a specific country, for example USA.
3. Add some methods to this zone, free shipping and flat rate are fine.
4. Add some items to the cart.
5. Go to the Cart.
6. Check you can still use the shipping calculator.
7. Check that the shipping rates show when using a USA address.
8. Check the error shows when using an address from another country, e.g. UK.
9. Go to the Checkout block.
10. Notice the shipping subtotal does not contain a price if your address is in an invalid country.
11. Change country on the Checkout form.
12. Notice that a shipping price is only shown when the address is in a valid country.
13. Try changing country back and forth and selecting different shipping rates.

### User Facing Testing
* [x] Same as above

### Changelog

> Prevent a 0 value shipping price being shown in the Checkout if no shipping methods are available.

### DIFF
--- a/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
+++ b/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
@@ -90,10 +90,15 @@ const NoShippingPlaceholder = ( {
 	if ( ! showCalculator ) {
 		return (
 			<em>
-				{ __(
-					'Calculated during checkout',
-					'woo-gutenberg-products-block'
-				) }
+				{ isCheckout
+					? __(
+							'No shipping options available',
+							'woo-gutenberg-products-block'
+					  )
+					: __(
+							'Calculated during checkout',
+							'woo-gutenberg-products-block'
+					  ) }
 			</em>
 		);
 	}
--- a/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
+++ b/assets/js/base/components/cart-checkout/totals/shipping/index.tsx
@@ -77,6 +77,7 @@ const ShippingAddress = ( {
 interface NoShippingPlaceholderProps {
 	showCalculator: boolean;
 	isShippingCalculatorOpen: boolean;
+	isCheckout?: boolean;
 	setIsShippingCalculatorOpen: CalculatorButtonProps[ 'setIsShippingCalculatorOpen' ];
 }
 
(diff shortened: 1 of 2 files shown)


mikejolley (approved):
Code looks good and testing steps worked 👍🏻 LGTM

PR #5446, opened 2021-12-23
Mini Cart Contents block improvements

### LINKED ISSUE #5445: Mini Cart Contents block improvements
There are some enhancements we should address to make the Mini Cart Contents block more usable:

- [ ] With the Block Navigation opened, the view doesn't change if users select the view block (not inner blocks). This applies to both Cart and Mini Cart.
- [ ] The empty view is using forced layout, which is unnecessary. Because of that, users can't delete the default heading block as well.
- [ ] On the empty view, users can add unwanted blocks like Mini Cart, Submenu. We should limit blocks can be added to the empty view of Mini Cart Contents.

dinhtungdu:
Fixes #5445

This PR:

- Limits blocks can be added to the empty view of the Mini Cart Contents block.
- Removes the forced layout of the empty view.
- Switchs to the correct view if the view block is selected.

### Testing

1. Edit the Mini Cart template part using the Site Editor.
2. Open the Block Navigation.
3. Select the Empty Mini Cart Contents block.
4. See the view is updated accordingly.
5. Try removing all blocks inside the empty view.
6. See the all blocks can be removed.
7. Try adding a block to empty view.
8. See Mini Cart block isn't in the blocks list.

### User Facing Testing
* [x] Same as above, or

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/edit.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/edit.tsx
@@ -1,40 +1,39 @@
 /**
  * External dependencies
  */
-import { __ } from '@wordpress/i18n';
 import { useBlockProps, InnerBlocks } from '@wordpress/block-editor';
-import { innerBlockAreas } from '@woocommerce/blocks-checkout';
-import type { TemplateArray } from '@wordpress/blocks';
 import { useEditorContext } from '@woocommerce/base-context';
+import { getBlockTypes } from '@wordpress/blocks';
 
 /**
  * Internal dependencies
  */
-import { useForcedLayout, getAllowedBlocks } from '../../../shared';
 
-export const Edit = ( { clientId }: { clientId: string } ): JSX.Element => {
+const excludedBlocks = [
+	'woocommerce/mini-cart',
+	'core/template-part',
+	'core/post-template',
+	'core/comment-template',
+];
+
+export const Edit = (): JSX.Element => {
 	const blockProps = useBlockProps();
-	const allowedBlocks = getAllowedBlocks( innerBlockAreas.EMPTY_MINI_CART );
 	const { currentView } = useEditorContext();
+	const allowedBlocks = [
+		...getBlockTypes()
+			.filter( ( block ) => {
+				if ( excludedBlocks.includes( block.name ) ) {
+					return false;
+				}
 
-	const defaultTemplate = ( [
-		[
-			'core/heading',
-			{
-				content: __(
-					'Empty mini cart content',
-					'woo-gutenberg-products-block'
-				),
-				level: 2,
-			},
-		],
-	].filter( Boolean ) as unknown ) as TemplateArray;
+				if ( block.parent && block.parent.length > 0 ) {
+					return false;
+				}
 
-	useForcedLayout( {
-		clientId,
-		registeredBlocks: allowedBlocks,
-		defaultTemplate,
-	} );
+				return true;
+			} )
+			.map( ( { name } ) => name ),
+	];
 
 	return (
 		<div
@@ -44,8 +43,7 @@ export const Edit = ( { clientId }: { clientId: string } ): JSX.Element => {
 			}
 		>
 			<InnerBlocks
-				template={ defaultTemplate }
-				templateLock={ false }
+				allowedBlocks={ allowedBlocks }
 				renderAppender={ InnerBlocks.ButtonBlockAppender }
 			/>
 		</div>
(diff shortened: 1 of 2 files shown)


sunyatasattva:
In my tests, I can remove and edit blocks from the Empty Mini Cart contents, but not from the filled one. I suppose that's expected, right?

Also the button to add blocks is on the right of the content, instead of center and full width. Is this also expected?

Screen Shot 2021-12-27 at 20 17 33

Otherwise, everything else described in the testing steps is fine.

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/edit.tsx
```diff
@@ -1,40 +1,39 @@
...
  */
-import { useForcedLayout, getAllowedBlocks } from '../../../shared';
 
-export const Edit = ( { clientId }: { clientId: string } ): JSX.Element => {
+const excludedBlocks = [
+	'woocommerce/mini-cart',
+	'core/template-part',
+	'core/post-template',
+	'core/comment-template',
+];
```
sunyatasattva, line 24: My suggestion here is to cast this `as const`. This will let TypeScript know that this is a readonly array.

Also, perhaps, worth changing this to CONSTANT_CASE

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/edit.tsx
```diff
@@ -1,40 +1,39 @@
...
+
+export const Edit = (): JSX.Element => {
 	const blockProps = useBlockProps();
-	const allowedBlocks = getAllowedBlocks( innerBlockAreas.EMPTY_MINI_CART );
 	const { currentView } = useEditorContext();
+	const allowedBlocks = [
+		...getBlockTypes()
```
sunyatasattva: It doesn't seem to be required to spread this array, since there is nothing else within here. Can we just do:

```ts
const allowedBlocks = getBlockTypes().filter().map()
```

?

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/edit.tsx
```diff
@@ -1,40 +1,39 @@
...
-			{
-				content: __(
-					'Empty mini cart content',
-					'woo-gutenberg-products-block'
-				),
-				level: 2,
-			},
-		],
-	].filter( Boolean ) as unknown ) as TemplateArray;
+				if ( block.parent && block.parent.length > 0 ) {
```
sunyatasattva: I'm a newcomer, so sorry about that: but I'm not sure what's the intent here. Would you be so kind as to explain why blocks with parents are to be excluded? It would make me understand our intentions better 🙏 

Also, we could perhaps replace this with:

```
if ( block.parent?.length )
```

sunyatasattva (changes requested):
Small suggestions :)

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/edit.tsx
dinhtungdu: Great point! I changed this to a read-only array of strings.

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/edit.tsx
dinhtungdu: Ah, you're right. We don't need `...` here. Thanks!

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/edit.tsx
dinhtungdu: Initially, I want to remove the blocks that have a parent from the allowed blocks to limit unwanted theme components like the previous post link. But doing so also removes other blocks that actually make sense to be included in the block list. I remove it for now and we should find another way to exclude blocks.

dinhtungdu (commented):
@sunyatasattva Thanks for your great review! I updated this PR to address your suggestions. Please take another look!

> In my tests, I can remove and edit blocks from the Empty Mini Cart contents, but not from the filled one. I suppose that's expected, right?

Yeah. Currently, we only allow customizing the empty mini cart contents view.

> Also the button to add blocks is on the right of the content, instead of center and full width. Is this also expected?

This is the behavior of WordPress's inner block appender. If there isn't any inner block, the appender is full-width. If there is more than one block, the appender button position change to the bottom right.

sunyatasattva (approved):
Thanks a lot for addressing my feedback! LGTM now! 🚀

PR #5447, opened 2021-12-23
Fix custom templates with fallback being incorrectly attributed

### LINKED ISSUE #5441: Store Editing Templates: Added By template information incorrect for customised template based off product-archive.html
## Describe the bug

This appears to be a fairly niche bug and is related to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5380 however this wasn't an issue at the time of merging. 

If we are using the `product-archive.html` template from the theme also for the product categories/tags (because templates for these dont exist in the theme), when we customise one of the product category template (or product tag template) and save it. It will set the template icon as your Gravatar and the added by value will be "demo"

## To reproduce

1.  Duplicate the Woo Blocks `archive-product.html` template into your FSE enabled theme.
2. Go to the templates list within the Site Editor
3. You will see now that "Product Archive", "Product Category" and "Product Tag" are added by the theme. This is because we are using the themes `archive-product.html` file for the category and tag templates as they're typically the same.
4. Customize the Product Category template and save it.
5. Navigate back to the templates list within the Site Editor. You will notice under the "Added By" column it should say your theme along with a theme icon. However it will display your Gravatar icon along with the text "demo".

## Expected behavior
I expect the "Added By" value to be the theme

## Screenshots
Screenshot 2021-12-22 at 14 20 41

sunyatasattva:
Category and tags templates can fallback to the generic archive if, e.g., the theme provides one for the latter but not for the former (#5380). However, since Gutenberg is not aware of this fallback mechanism, it would incorrectly attribute the custom template
to the user instead of the theme.

Here we are explicitly setting the `has_theme_file` to make sure Gutenberg knows we do, in fact, have a theme fail (if not what it expects).

Fixes #5441

#### How to test the changes in this Pull Request:

1. Install and enable the `Gutenberg` plugin.
2. Install and enable the `TT1 Blocks` theme.
3. Add an `archive-product.html`. This is because we are using the themes archive-product.html file for the category and tag templates as they're typically the same.
4. Go to the templates list within the Site Editor
5. You will see now that `Product Archive`, `Product Category` and `Product Tag` are added by the theme.
6. Customize the `Product Category` template and save it.
7. Navigate back to the templates list within the Site Editor. You will notice under the "Added By" column it should say your theme along with a theme icon.
8. Notice that these templates render in the Site Editor, and on the front-end as expected with and without customizations.

#### Regression testing

**Note:** The logic of all these steps is a bit convoluted, so doing some regression tests is advised for the following scenarios:

1. If you put product taxonomy template(s) in the theme, please check that these are used over the themes `product-archive.html` on both the frontend and in the site editor. Please also check these are customizable and render as expected.
2. If no Woo related templates are in the theme, check it uses the templates from Woo Blocks as expected, on both the frontend and in the site editor. Please also check these are customizable and render as expected.
4. If you put a taxonomy template(s) in the theme, it doesn't appear twice in the template list.

### User Facing Testing
* [x] Same as above

### Changelog

> Fix custom templates with fallback to archive being incorrectly attributed to the user in the editor instead of the parent theme.

### DIFF
--- a/src/Utils/BlockTemplateUtils.php
+++ b/src/Utils/BlockTemplateUtils.php
@@ -317,7 +317,7 @@ class BlockTemplateUtils {
 	 * @return boolean
 	 */
 	public static function set_has_theme_file_if_fallback_is_available( $query_result, $template ) {
-		foreach ( $query_result as $query_result_template ) {
+		foreach ( $query_result as &$query_result_template ) {
 			if (
 				$query_result_template->slug === $template->slug
 				&& $query_result_template->theme === $template->theme
--- a/src/Utils/BlockTemplateUtils.php
+++ b/src/Utils/BlockTemplateUtils.php
@@ -281,4 +281,68 @@ class BlockTemplateUtils {
 
 		return true;
 	}
+
+	/**
+	 * Checks if we can fallback to the `archive-product` template for a given slug
+	 *
+	 * `taxonomy-product_cat` and `taxonomy-product_tag` templates can generally use the
+	 * `archive-product` as a fallback if there are no specific overrides.
+	 *
+	 * @param string $template_slug Slug to check for fallbacks.
+	 * @return boolean
+	 */
+	public static function template_is_eligible_for_product_archive_fallback( $template_slug ) {
+		$eligible_for_fallbacks = array( 'taxonomy-product_cat', 'taxonomy-product_tag' );
+
+		if (
+			in_array( $template_slug, $eligible_for_fallbacks, true )
+			&& ! self::theme_has_template( $template_slug )
+			&& self::theme_has_template( 'archive-product' )
+		) {
+			return true;
+		}
+
+		return false;
+	}
+
+	/**
+	 * Sets the `has_theme_file` to `true` for templates with fallbacks
+	 *
+	 * There are cases (such as tags and categories) in which fallback templates
+	 * can be used; so, while *technically* the theme doesn't have a specific file
+	 * for them, it is important that we tell Gutenberg that we do, in fact,
+	 * have a theme file (i.e. the fallback one).
+	 *
+	 * **Note:** this function changes the array that has been passed.
+	 *
+	 * It returns `true` if anything was changed, `false` otherwise.
+	 *
+	 * @param array $query_result Array of template objects.
+	 * @param array $template A specific template object which could have a fallback.
+	 *
+	 * @return boolean
+	 */
+	public static function confirm_theme_file_when_fallback_is_available( $query_result, $template ) {
+		$template_with_fallback_idx = null;
+
+		array_walk(
+			$query_result,
+			function( $query_result_template, $idx ) use ( $template, &$template_with_fallback_idx ) {
+				if (
+					$query_result_template->slug === $template->slug
+					&& $query_result_template->theme === $template->theme
+					&& self::template_is_eligible_for_product_archive_fallback( $template->slug )
+				) {
+						$template_with_fallback_idx = $idx;
+				}
+			}
+		);
+
+		if ( is_int( $template_with_fallback_idx ) ) {
+			$query_result[ $template_with_fallback_idx ]->has_theme_file = true;
+			return true;
+		}
+
+		return false;
+	}
 }
--- a/src/Utils/BlockTemplateUtils.php
+++ b/src/Utils/BlockTemplateUtils.php
@@ -317,7 +317,7 @@ class BlockTemplateUtils {
 	 * @return boolean
 	 */
 	public static function set_has_theme_file_if_fallback_is_available( $query_result, $template ) {
-		foreach ( $query_result as $i => &$query_result_template ) {
+		foreach ( $query_result as $query_result_template ) {
 			if (
 				$query_result_template->slug === $template->slug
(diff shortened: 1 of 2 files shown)


src/Utils/BlockTemplateUtils.php
```diff
@@ -281,4 +281,75 @@ public static function supports_block_templates() {
...
+	 *
+	 * It returns `true` if anything was changed, `false` otherwise.
+	 *
+	 * @param array $query_result Array of template objects.
+	 * @param array $template A specific template object which could have a fallback.
+	 *
+	 * @return boolean
+	 */
+	public static function confirm_theme_file_when_fallback_is_available( $query_result, $template ) {
+		$template_with_fallback_idx = null;
```
opr: What is `idx`? If it means index can we just use the full word instead?

src/Utils/BlockTemplateUtils.php
```diff
@@ -281,4 +281,75 @@ public static function supports_block_templates() {
...
+		$is_duplicate               = false;
+
+		array_walk(
```
opr: Any reason why you chose to use this over a `foreach`? `foreach` is quicker :)

src/Utils/BlockTemplateUtils.php
```diff
@@ -281,4 +281,75 @@ public static function supports_block_templates() {
...
+
+					if ( self::template_is_eligible_for_product_archive_fallback( $template->slug ) ) {
+						$template_with_fallback_idx = $idx;
+					}
+				}
+			}
+		);
+
+		if ( $is_duplicate ) {
+			if ( is_int( $template_with_fallback_idx ) ) {
```
opr: Could we add a comment here explaining why we need to check `is_int`? This will stop someone "refactoring" it in future 🤓

src/Utils/BlockTemplateUtils.php
```diff
@@ -281,4 +281,75 @@ public static function supports_block_templates() {
...
+	 * **Note:** this function changes the array that has been passed.
+	 *
+	 * It returns `true` if anything was changed, `false` otherwise.
+	 *
+	 * @param array $query_result Array of template objects.
+	 * @param array $template A specific template object which could have a fallback.
+	 *
+	 * @return boolean
+	 */
+	public static function confirm_theme_file_when_fallback_is_available( $query_result, $template ) {
```
opr: Do you think something like `set_has_theme_file_if_fallback_is_available` would be a slightly better name? At a glance it is hard to tell what this function does imo.

src/Utils/BlockTemplateUtils.php
```diff
@@ -281,4 +281,75 @@ public static function supports_block_templates() {
...
+
+		if (
+			in_array( $template_slug, $eligible_for_fallbacks, true )
+			&& ! self::theme_has_template( $template_slug )
+			&& self::theme_has_template( 'archive-product' )
+		) {
+			return true;
+		}
+
+		return false;
```
opr: Could we just return the result of the evaluation instead of using an `if` here?

opr (changes requested):
Works well! Thanks for doing this, I just have a couple of nitpicky changes though, let me know what you think :)

👏🏼

src/Utils/BlockTemplateUtils.php
sunyatasattva: No reason but my massive PHP ignorance. I didn't know you could get the index with `foreach` but it looks like you can! Thanks for pointing that out.

sunyatasattva:
@opr Thanks for the feedback, especially for the `foreach` ! My code looks much cleaner as I refactored the whole need to have an index there. 🙏

src/Utils/BlockTemplateUtils.php
```diff
@@ -281,4 +281,55 @@ public static function supports_block_templates() {
...
+	 *
+	 * It returns `true` if anything was changed, `false` otherwise.
+	 *
+	 * @param array $query_result Array of template objects.
+	 * @param array $template A specific template object which could have a fallback.
+	 *
+	 * @return boolean
+	 */
+	public static function set_has_theme_file_if_fallback_is_available( $query_result, $template ) {
+		foreach ( $query_result as $i => &$query_result_template ) {
```
opr: Looks much better but `$i` is not used

```suggestion
		foreach ( $query_result as $query_result_template ) {
```

opr (approved):
🙌🏼  Thanks for addressing the changes! There's just one tiny thing but I'll pre-approve.

src/Utils/BlockTemplateUtils.php
sunyatasattva: 🤦‍♀️  I had used the index, but then I realized I didn't need that. I think I still need to pass the query_result_template as reference because I edit it, right? Again, my lackluster PHP knowledge her, but I thought the `&` was required here.

opr: Oh right, yeah you do sorry I missed that! I didn't notice the `&` there.

sunyatasattva: At least that's what I understand from [the docs](https://www.php.net/manual/en/control-structures.foreach.php):

> In order to be able to directly modify array elements within the loop precede `$value` with `&`

PR #5448, opened 2021-12-23
Mini Cart Contents: Fix missing custom background

### LINKED ISSUE #5416: Mini Cart Contents: background doesn't work on the frontend with Gutenberg disabled
## Describe the bug
Using WordPress 5.9 beta, with the Gutenberg disabled, admins can still edit and save the background color of the Mini Cart Contents block in the site editor. But the color doesn't apply on the front-end.

## To reproduce

1. Update to WordPress 5.9 beta.
2. Disable Gutenberg.
3. Go to Appearance > Site Editor > Template Parts > Mini Cart.
4. Select the Mini Cart Contents block, change the background.
5. See the background changed in the site editor.
6. Save the block, see the background doesn't work on the front end.

## Expected behavior
The background should work with WordPress 5.9.

## Additional context
The background color works with Gutenberg Site Editor: `/wp-admin/themes.php?page=gutenberg-edit-site`. So the issue is already fixed in upstream Gutenberg.

[Original comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5386#issuecomment-996642892) by @Aljullu:
> Something I noticed while doing some more testing is that in WP 5.9 beta 3 with Gutenberg disabled, I can check the color in the editor, but the new color is not applied in the frontend. @dinhtungdu can you reproduce that too? If so, it would be great to create an issue about that so we can investigate it further.

gigitux:
With this PR using WordPress 5.9 beta, with the Gutenberg disabled, is visible the custom background if it is set. 

Fixes #5416

### Manual Testing

Check out this branch

1. Update to WordPress 5.9 beta.
2. Disable Gutenberg.
3. Go to Appearance > Site Editor > Template Parts > Mini Cart.
4. Select the Mini Cart Contents block, change the background.
5. See the background changed in the site editor.
6. Save the block, check on the fronted side that the custom background is visible


src/Utils/BlockTemplateUtils.php
```diff
@@ -282,6 +282,28 @@ public static function supports_block_templates() {
...
+	/**
+	 * Retrieves a single unified template object using its id.
+	 *
+	 * @param string $id            Template unique identifier (example: theme_slug//template_slug).
+	 * @param string $template_type Optional. Template type: `'wp_template'` or '`wp_template_part'`.
+	 *                             Default `'wp_template'`.
+	 *
+	 * @return WP_Block_Template|null Template.
+	 */
+	public static function fse_get_block_template( $id, $template_type ) {
```
Aljullu: I'm not sure about the `fse` prefix... :thinking: I would just call it `get_block_template` like in WP core or `wc_get_block_template` if we want to make it clear it's WC-specific.

Aljullu (commented):
Good investigation with this issue! I like the util function you created, do you think we could use it in other instances (like [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8d16ff2f62ad57e6bb3b766d7304270667e1c06e/src/BlockTemplatesController.php#L95-L97) and [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8d16ff2f62ad57e6bb3b766d7304270667e1c06e/src/BlockTemplatesController.php#L107-L109))?

I also left a minor suggestion about renaming the function.

src/Utils/BlockTemplateUtils.php
gigitux: Yes, it makes sense!

gigitux:
> Good investigation with this issue! I like the util function you created, do you think we could use it in other instances (like here and here)?

As we already discussed in our internal channel, maybe before adopting this utility globally it is better to investigate why there is this different behavior even if both functions should do the same thing. This can require a little bit of time :(

Aljullu:
> As we already discussed in our internal channel, maybe before adopting this utility globally it is better to investigate why there is this different behavior even if both functions should do the same thing. This can require a little bit of time :(

Sounds good! Do you know if we have an issue for that? It would be good adding it to the Template Editing v2 epic so we don't forget.

gigitux:
> Sounds good! Do you know if we have an issue for that? It would be good adding it to the Template Editing v2 epic so we don't forget.

No, it was my intention to create it after checking with you. 
In the next hours, I will create a detailed issue 👍

src/Utils/BlockTemplateUtils.php
dinhtungdu: I would call it `get_block_template` as well, we already have `BlockTemplateUtils` as the prefix.

gigitux:
@Aljullu I updated the PR. Should we merge this PR now or should we investigate issue #5474 and after that merge it?

Aljullu (approved):
> @Aljullu I updated the PR. Should we merge this PR now or should we investigate issue #5474 and after that merge it?

Given that the Mini Cart block hasn't been released yet, I think it's safe merging this PR already. So :shipit: from my side. :slightly_smiling_face:

PR #5449, opened 2021-12-23
Remove Built-in Stripe Integration in favour of Stripe Extension

### LINKED ISSUE #5108: Remove built-in Stripe integration for Cart and Checkout Blocks
When we first built the Cart and Checkout blocks, the first payment method integrated was Stripe which we built alongside developing the payment method integration API. To help expediate the validation process of the API, we built the integration directly in this plugin rather than in the Stripe extension.

Later on, the Stripe team added support for the Blocks in the extension itself and at the time we left the built-in integration in place to make the transition transparent for users of both Stripe and the new Checkout flow. In the meantime, the current built-in Stripe integration has not received any updates and all improvements have been done directly in the Stripe extension.  We always knew that some point, once enough time had passed, and users have updated both the Stripe extension and WooCommerce blocks, we'd be able to remove the built-in integration.

That time is now.

I've done some investigation through the data we have on store usage of Woo Blocks and the Stripe extension (stores that have opted in to this collection) and am able to determine that the number of Stores that would be potentially impacted by us removing the built-in integration is around 1.2% of all stores using the Stripe extension. This 1.2% are stores that:

- Are using an older version of the Stripe extension that does not have the integration with the Cart and Checkout Blocks.
- Are actively using the Cart and Checkout blocks (so the built-in integration is in use).

Of those stores, all of them are on older versions of both Stripe _and_ Woo Blocks so they have not been keeping their sites up to date anyways.

The impact is thus very limited and will only pop up for stores that:

- Are in the list above.
- Update WooCommerce Blocks to a version with the built-in Stripe integration removed _without_ updating the Stripe extension to the latest version.

### Steps

Whoever takes this issue on, here's some things that must be done:

* [ ] Remove all stripe integration code from the Woo Blocks extension from the JavaScript implementation.
* [ ] Remove all stripe integration related code from the Woo Blocks extension from the PHP implementation.
* [ ] Verify after removing that above code that the integration with Cart and Checkout blocks in the Stripe extension  still works as expected.
* [ ] In your PR, make sure you indicate that this should be highlighted in the release notes. Ideally, craft what the release lead can use in the release post.
* [ ] When your PR is merged and the release version is known, share with the WooCommerce Happiness team about the change, when it will surface, and what to do if Stores report Stripe not working with the Cart and Checkout blocks (the solution is to simply update their Stripe extension to the latest). I anticipate any reports should be rare.

mikejolley:
This PR removes the Stripe integration code (server and client) and Stripe dependencies. This has been included with the [Stripe Payment Method Extension](https://github.com/woocommerce/woocommerce-gateway-stripe/) since 5.2.0 (2021-05-19).
 
Closes #5108

### Testing

1. No build errors
2. Other payment methods still function (do a test checkout)

### User Facing Testing

Install the [Stripe Payment Method Extension](https://github.com/woocommerce/woocommerce-gateway-stripe/), activate it, enable it and test it works by:

1. Adding items to your cart
2. Go to checkout
3. Select Stripe Credit Card extension and make a payment. Do so while logged in and save the card to your account.
4. Repeat checkout with a saved card.

### Dev Note

The built-in [Stripe Payment Method Extension](https://github.com/woocommerce/woocommerce-gateway-stripe/) integration has been removed. This has been part of the extension itself since Stripe version 5.2.0 (2021-05-19) and is no longer necessary. If you're using an older version of Stripe (lower than 5.2.0) you'll need to update to have the payment method still shown during Block Checkout.

### Changelog

> Remove Stripe Payment Method Integration (which is now part of the Stripe Payment Method extension itself)


opr (approved):
Tested and it's working, code changes look good too. Cheers Mike!

Issue #5343, opened 2021-12-09
Update the path of WooCommerce templates to match the new Gutenberg folders


Aljullu:
In https://github.com/WordPress/gutenberg/pull/36647, Gutenberg changed the folder names of block themes to use `templates` and `parts` instead of `block-templates` and `block-template-parts`. In WooCommerce, [we are using the old ones](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/de2dd8e9b5584ea9e7d79972b9ff3ec13d166556/templates), it's not a big deal because we are not a theme and are loading them via hooks, but still, it would be great to update our directory structure to use the new one.

tjcafferkey:
Closing as this is a duplicate of https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5450 and I believe has now become more important as we are checking theme files for the existence of templates before loading our own.

tjcafferkey:
>Closing as this is a duplicate of #5450 and I believe has now become more important as we are checking theme files for the existence of templates before loading our own.

Reopening this issue. It's not quite the same as #5450 where we need to update the paths of the theme file locations we are checking. This PR should just be scoped to renaming the Woo Block directories where we are storing Core block templates

PR #5455, opened 2021-12-24
Check block templates and parts directories to the Gutenberg 12.1.0 convention

### LINKED ISSUE #5450: Store Editing Templates: Consider new block templates & block-template-parts directories
## Describe the bug
The new naming conventions set for FSE enabled themes now look for the templates html files in `templates` and `parts`. [Tove](https://en-gb.wordpress.org/themes/tove/) is an example of a theme that uses these now.

We need to also check these directories in [our code](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Utils/BlockTemplateUtils.php#L247-L267) as well as the current `block-templates` and `block-template-parts` directories.

Note: There may be instances elsewhere in our BlockTemplating code where we look for these directories so please check.

sunyatasattva:
With WordPress/gutenberg#36647, Gutenberg 12.1.0 has changed the convention for the directory paths from `block-templates` and `block-template-parts` to `templates` and `parts` respectively.

We are also moving to that convention but making sure that we remain backwards-compatible.

Fixes #5450

#### Manual Testing

> ⚠️ **Prerequisite:** you'll need to have [these changes](https://github.com/woocommerce/woocommerce/pull/31522) applied to WooCommerce Core for this PR to work as intended

1. Make sure your existing theme with the old convention (e.g. TT1 Blocks, using `block-templates`) still shows the templates correctly.
2. Change the directory names to `templates` and `parts` respectively, or download a theme which uses this new convention (e.g. [Tove](https://en-gb.wordpress.org/themes/tove/)), and make sure the templates still show correctly. 
3. Have both the old convention and the new convention living in the same theme and make sure that the templates under the newer names get prioritized instead of the deprecated ones.

### User Facing Testing
* [x] Same as above

### Changelog

> Update the template retrieving logic to allow for older Gutenberg convention and newer one (`block-templates`/`block-template-parts` vs. `templates`/`parts`)


tjcafferkey (changes requested):
I have a few thoughts on this:

* We shouldn't be importing the `BlockTemplatesController` into the `BlockTemplatesUtils`.
* The consts defined at the top of the `BlockTemplatesController` are for the Woo Blocks templates/parts, _not_ the themes. If we are going to update the Woo Block template locations (https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5343) we should do this in a separate PR
* All we need to do is to add `templates` and `parts` to the checks we do when looking for themes template files whilst leaving the `block-templates` and `block-template-parts` directories already in place

I feel like we are doing too much in this PR. Let's limit the scope to just consider the new theme directories. We can move _our_ templates in a separate PR. Let me know what you think.

Edit: The reason for me suggesting we should be limiting scope in this PR is because this is quite an important issue we need to solve. Limiting scope will make it easier to test it works, avoid any negative side effects and easily trace back to this PR if there are. Given that we don't have much test coverage for this feature I am just trying to be a bit more cautious.

sunyatasattva:
Understood, I'll separate the two things!

sunyatasattva:
Alright, as requested, I have split the original changes from this PR within here and #5464. I've also updated the OP to reflect just these changes in the changelog, references and testing notes.

@Aljullu I took the liberty to tag you as a reviewer as @tjcafferkey is AFK for the next week and you have context on what's this about since you had originally opened #5343.

src/Utils/BlockTemplateUtils.php
```diff
@@ -6,6 +6,42 @@
...
+	 * @var string
+	 */
+	const TEMPLATES_DIR_NAME = 'templates';
+
+	/**
+	 * Directory name of the block template parts directory.
+	 *
+	 * @var string
+	 */
+	const TEMPLATE_PARTS_DIR_NAME = 'parts';
```
tjcafferkey: Can we not put these values in an [associative array](https://www.php.net/manual/en/language.types.array.php)?

sunyatasattva: We 100% can. I was just following the pre-existing convention. Also, in the spirit of trying to avoid as many side-effects as possible, not changing that convention makes the PR slightly safer.

But I can convert those values in an associative array, if you so prefer.

tjcafferkey: I think it might make sense to put these in an associative array, I don't think it'll be too risky given the convention hasn't been set in this file.

src/Utils/BlockTemplateUtils.php
```diff
@@ -244,15 +280,65 @@ public static function generate_template_slug_from_path( $path, $directory_name
...
+			self::TEMPLATES_DIR_NAME,
+			self::DEPRECATED_TEMPLATES_DIR_NAME,
+		) : array(
+			self::TEMPLATE_PARTS_DIR_NAME,
+			self::DEPRECATED_TEMPLATE_PARTS_DIR_NAME,
+		);
+
+		// Combine the possible root directory names with either the template directory
+		// or the stylesheet directory for child themes.
+		$possible_paths = array_reduce(
```
tjcafferkey, line 296: In my tests, at this point the array has 4 entries when it should have 2 I think? There are duplicates.

tjcafferkey:
If I put `archive-product.html` in `tove/templates` it does not render on the frontend for me? Does the same happen for you?

src/Utils/BlockTemplateUtils.php
tjcafferkey: Ah, I see why now. This will also consider child themes. We should probably check the value doesnt exist in the array before adding it.

sunyatasattva: `$possible_paths` here should be 4 entries, because it combines the two possible conventions (newer and deprecated) and the two possible roots (`template_directory` and `stylesheet_directory`).

For most themes, I suppose, these would be duplicates as `get_template_directory` and `get_stylesheet_directory` return the same path. However, this would allow for child themes, where, AFAIU, these two functions return different paths (the former returns the parent theme's path, the latter the child theme's one)

sunyatasattva: > Ah, I see why now. This will also consider child themes. We should probably check the value doesnt exist in the array before adding it.

I see your point, but I'm not sure this is required. The next bit of logic iterates through the array and returns the first available path. There is basically no harm done (as the array is also super small), and it keeps the code a bit more readable, IMHO.

But if you feel strongly about this, I can add the check.

sunyatasattva:
> If I put `archive-product.html` in `tove/templates` it does not render on the frontend for me? Does the same happen for you?

Oh! Thanks for spotting it! I see why: I moved everything I had changed in the Controller to the other PR, to separate the changes as requested. But they are somewhat tightly coupled so, this change was moved as well, going to revert that at once.

sunyatasattva:
@tjcafferkey I've addressed the issues:

1. Directory names are now in an associative array
2. Fallback templates should appear correctly

Did not change the `reduce` logic, yet. As I mentioned, I can do it if you feel strongly about it no problem: I feel it causes no harm and reduces cognitive load on the reader of the code.

tjcafferkey:
>@tjcafferkey I've addressed the issues:

Thanks!

>Oh! Thanks for spotting it! I see why

The themes `product-archive.html` template loads for the product category page on the frontend (e.g. `/product-category/clothing/`) if there isn't a taxonomy template (which is correct) but does not load for the main shop archive page (e.g. `/shop`).

src/Utils/BlockTemplateUtils.php
tjcafferkey: This is fine then, I can see now why it works that way. I agree lets leave it 🙂

sunyatasattva:
> The themes `product-archive.html` template loads for the product category page on the frontend (e.g. `/product-category/clothing/`) if there isn't a taxonomy template (which is correct) but does not load for the main shop archive page (e.g. `/shop`).

It took me a long time, but I found the culprit. Turns out that this is a WooCommerce problem, as they don't support this convention yet. I have opened a bug on their repo here: https://github.com/woocommerce/woocommerce/issues/31518 (also a [Slack thread](p1640646806110100-slack-C0E1AV8T0))

I could also open a PR to fix that.

In the meantime, I added a fix on our side by hooking into the `woocommerce_has_block_template` filter and adding our own checking logic with all the possible directories.

tjcafferkey:
Nice catch @sunyatasattva!

>I added a fix on our side by hooking into the woocommerce_has_block_template filter and adding our own checking logic with all the possible directories.

I think we should revert the temporary fix on our side which you just added, and open a PR to amend the following code directly in the [WC Core codebase](https://github.com/woocommerce/woocommerce/blob/baadffe3a669878ee93b9e0fdc43bb610333d61f/plugins/woocommerce/includes/class-wc-template-loader.php#L115). Then in the testing instructions for this PR just detail that the tester should have the other WC PR checked out locally too, to test this. Feel free to add me as a reviewer too.

```
$has_template = is_readable( get_stylesheet_directory() . '/block-templates/' . $template_name . '.html' ) ||  is_readable( get_stylesheet_directory() . '/templates/' . $template_name . '.html' );
```

As a side note we should also add to the testing instructions anyway for the reviewer to check the frontend templating is working as expected.

Let me know what you think.

sunyatasattva:
Hey @tjcafferkey !

First of all, thanks a lot for keeping an eye on this despite being AFK.

To the subject matter: I did submit [a PR to Woo Core](https://github.com/woocommerce/woocommerce/pull/31522) with the permission of @vedanshujain on [Slack](p1640646806110100-slack-C0E1AV8T0). I also did remove the temporary patch and also added the caveat in the testing note on the OP.

However, I have a few remarks I'd like to bounce to you:

1. Perhaps the temporary patch was ok to keep for a while, as now we'll need to make sure both the repos are aligned in various installations from our users and also developers. This might complicate stuff. Maybe we could have left the filter as a safety net and removed that on the next release or other milestone like that.
2. There are these lines which use a similar patchy trick, which we might want to address in Core as well, maybe 🤔 : https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/9a8f27fcd72ddfce7c75eaa681ac271732da5e94/src/BlockTemplatesController.php#L414-L439

I'm mostly worried about (1.), I don't think it's much harm to leave the patch in as long as we remember to remove it when everything is aligned and stable.

As for now, I have reverted as requested. Let me know your thoughts.

tjcafferkey:
>First of all, thanks a lot for keeping an eye on this despite being AFK.

No problem, thank you for doing such a thorough job on this issue I really appreciate it!

>Perhaps the temporary patch was ok to keep for a while, as now we'll need to make sure both the repos are aligned in various installations from our users and also developers. This might complicate stuff. Maybe we could have left the filter as a safety net and removed that on the next release or other milestone like that.

It depends on the issue I think. In this case we don't need the temp fix for themes using `block-templates/` so it's not really too critical, and we can specify (until the Woo Core PR is merged) in the Epic (like we did in V1 [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4926)) that if reviewers are going to use the new directories they need said PR to be checked out locally for it to work.

If this was a time sensitive issue that was needed for an imminent release I would totally be onboard with the fix, however I don't think that it's the case here (unless I am wrong?). Also by putting in a temp fix this means we need to review and test:

* The temporary fix
* The actual fix (Woo Core)
* The removal of the temporary fix

This I think would be unnecessary unless it was absolutely needed.

>There are these lines which use a similar patchy trick, which we might want to address in Core as well

Could you please elaborate on this I am not sure what you mean? Sorry.

Let me know your thoughts.

tjcafferkey (approved):
Great job @sunyatasattva thanks for addressing all the feedback. I've tested this and it seems to be working as expected. Thank you for fixing.

sunyatasattva:
@tjcafferkey Alright, everything you said sounds good to me.

I thought it was a bit more time sensitive because there already are themes (like Tove) which use the new convention. Just that.

> > There are these lines which use a similar patchy trick, which we might want to address in Core as well
> 
> Could you please elaborate on this I am not sure what you mean? Sorry.

I see that in that piece of code I linked, we are basically forcing Woo to know that, in specific circumstances, there are indeed block templates. So I thought that perhaps that logic should be made available to Woo Core itself, if possible.

Not sure if that made any sense.

In any case, nothing that should take more of your AFK attention :) Thanks again for sticking with it and approving!

src/Utils/BlockTemplateUtils.php
```diff
@@ -244,15 +265,65 @@ public static function generate_template_slug_from_path( $path, $directory_name
...
+			$possible_templates_dir,
+			function( $carry, $item ) use ( $template_filename ) {
+				$filepath = DIRECTORY_SEPARATOR . $item . DIRECTORY_SEPARATOR . $template_filename;
+
+				$carry[] = get_template_directory() . $filepath;
+				$carry[] = get_stylesheet_directory() . $filepath;
```
Aljullu, line 302: Sorry for being late on this one, but if I'm not wrong this is searching templates in this order:

1. Parent theme `templates` folder.
2. Child theme `templates` folder.
3. Parent theme `block-templates` folder.
4. Child theme `block-templates` folder.

But I think the child theme should always have priority. Something like this:

1. Child theme `templates` folder.
2. Child theme `block-templates` folder.
3. Parent theme `templates` folder.
4. Parent theme `block-templates` folder.

What do you think @sunyatasattva?

tjcafferkey:
>I see that in that piece of code I linked, we are basically forcing Woo to know that, in specific circumstances, there are indeed block templates. So I thought that perhaps that logic should be made available to Woo Core itself, if possible.

Yeah you are correct that is how that code works, it gives us the ability to override Core and load templates from the plugin.

I'm not so sure how this would fit in Core currently? Core checks the active theme for block templates, this is our way of loading them from our plugin Woo Blocks (which Core would have no concept of, although we do include the Woo Blocks package in Core every month). Maybe I'm missing something or misunderstanding but I'm open to discussing/seeing if what you mean could work 🙂

src/Utils/BlockTemplateUtils.php
sunyatasattva: @Aljullu Indeed this is correct! I'll open a PR to amend this! Thanks for spotting that.

sunyatasattva:
Ok I see now! I think you got my meaning, it was I who did not fully understand the relationship between Core and Blocks and themes and plugins. Thanks for taking the time to clarify.

ralucaStan:
I've added skip-changelog label as the  prerequisite PR has not been merged by the time of our release

PR #5464, opened 2021-12-27
Move Woo Blocks template directories to latest Gutenberg convention

### LINKED ISSUE #5343: Update the path of WooCommerce templates to match the new Gutenberg folders

sunyatasattva:
With WordPress/gutenberg#36647, Gutenberg 12.1.0 has changed the convention for the directory paths from `block-templates` and `block-template-parts` to `templates` and `parts` respectively.

We are also moving our repo to that convention.

Fixes #5343

### Note

Tagging also @tjcafferkey as a reviewer, just to page him (as he had requested this PR to be split in two); however, he's going to be AFK for the next week, so it's really just a CC.

#### Manual Testing

1. Open Appearance → Editor → Templates
2. Make sure all the templates and parts appear correctly in line with what's available within our directories
3. Make sure the Mini Cart renders correctly (as it is a template part). Try with and without Gutenberg enabled.

### User Facing Testing
* [x] Same as above

### Changelog

> Update the block theme folders to latest Gutenberg convention (i.e. `templates` and `parts`).


sunyatasattva:
@Aljullu As #5455 was merged, I rebased this onto `trunk` and removed the warning on the top.

src/Utils/BlockTemplateUtils.php
```diff
@@ -254,15 +254,12 @@ public static function convert_slug_to_title( $template_slug ) {
...
-	public static function generate_template_slug_from_path( $path, $directory_name = 'block-templates' ) {
-		return substr(
-			$path,
-			strpos( $path, $directory_name . DIRECTORY_SEPARATOR ) + 1 + strlen( $directory_name ),
-			-5
-		);
+	public static function generate_template_slug_from_path( $path ) {
+		$template_extension = '.html';
+
+		return basename( $path, $template_extension );
```
Aljullu, line 262: Much better. :clap:

Aljullu (approved):
This is testing well and code looks good, thanks for working on this @sunyatasattva!

Heads-up that we will need to update the path in this line:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c2444b68dfc658d49dfd5ac41df50a70dc66b523/src/BlockTypes/MiniCart.php#L390

Is that something we can do as part of this PR?

Besides that, everything looks good to me, so pre-approving.

src/Utils/BlockTemplateUtils.php
sunyatasattva: Thanks! It was also necessary, by the way. I didn't simplify it just for the sake of it. But because our dir structure became `templates/templates`, the original `substr` code would not work anymore.

sunyatasattva:
> Is that something we can do as part of this PR?

@Aljullu  I updated that code, but I'm not 100% sure how to test it. Would you mind testing the mini cart part and perhaps add testing instructions to the OP for when the release is made?

I also snuck in the change of order between child and parent theme priority.

Lastly, I did a search on `block-template` and I found this function: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2635445c5d303eba4dfef12efbde0f29667f2673/src/BlockTemplatesController.php#L62-L120

It mentions that it looks for the `block-templates` directory, but I'm not sure where it does that. Perhaps one of the functions: `gutenberg_get_block_template` or `get_block_template` which I'm not sure where they come from.

Would this mean we might have to change those somewhere?

Aljullu (approved):
> @Aljullu I updated that code, but I'm not 100% sure how to test it. Would you mind testing the mini cart part and perhaps add testing instructions to the OP for when the release is made?

Done! :slightly_smiling_face: Right now, it could be reproduced with Gutenberg disabled, but once https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5448 is merged, I think it will require testing with WP 5.8 and Gutenberg disabled.

> I also snuck in the change of order between child and parent theme priority.

Thanks! That still loads the "Parent theme `templates` folder" before "Child theme `block-templates` folder", if I'm reading it correctly. I don't have a strong opinion on this one, but I feel it should be the other way around (child theme wins even if it's using the old directory name). In any case, not blocking this PR.

> Lastly, I did a search on `block-template` and I found this function:

Good catch! I think that comment is outdated and needs to be updated.

> It mentions that it looks for the `block-templates` directory, but I'm not sure where it does that. Perhaps one of the functions: `gutenberg_get_block_template` or `get_block_template` which I'm not sure where they come from.

I think that happens when `get_single_block_template` is hooked into `get_block_file_template`. If you keep looking, `get_single_block_template()` calls `get_block_templates()` which, in its turn, calls `get_block_templates_from_woocommerce()`, where templates are loaded from files, if I'm not wrong.

`gutenberg_get_block_template()` calls `get_block_template()` are Gutenberg and WordPress globals, but I don't think we need to modify them here.

sunyatasattva:
> > I also snuck in the change of order between child and parent theme priority.
> 
> Thanks! That still loads the "Parent theme `templates` folder" before "Child theme `block-templates` folder", if I'm reading it correctly. I don't have a strong opinion on this one, but I feel it should be the other way around (child theme wins even if it's using the old directory name). In any case, not blocking this PR.

🤔  I see what you mean. You are probably right, that should likely be the order… though possibly unlikely that a parent theme has the new convention and the child the old one. I mean, that's possible, but I get the feeling that this is a relatively edge case. Should I create a follow-up issue so we don't forget?

> 
> > Lastly, I did a search on `block-template` and I found this function:
> 
> Good catch! I think that comment is outdated and needs to be updated.
> 

Updated the comment!

> > It mentions that it looks for the `block-templates` directory, but I'm not sure where it does that. Perhaps one of the functions: `gutenberg_get_block_template` or `get_block_template` which I'm not sure where they come from.
> 
> I think that happens when `get_single_block_template` is hooked into `get_block_file_template`. If you keep looking, `get_single_block_template()` calls `get_block_templates()` which, in its turn, calls `get_block_templates_from_woocommerce()`, where templates are loaded from files, if I'm not wrong.
> 
> `gutenberg_get_block_template()` calls `get_block_template()` are Gutenberg and WordPress globals, but I don't think we need to modify them here.

Thanks a lot for taking the time to clarify!

Aljullu:
> 🤔  I see what you mean. You are probably right, that should likely be the order… though possibly unlikely that a parent theme has the new convention and the child the old one. I mean, that's possible, but I get the feeling that this is a relatively edge case. Should I create a follow-up issue so we don't forget?

I wonder if we could do here the same [we did in WC core](https://github.com/woocommerce/woocommerce/pull/31522/files#diff-b9feeabfae8d54fb2aa16ee00af817a4d66331b13d4a555e8a97340c28a96701R124-R134). This way, we can easily change the order templates are loaded.

src/BlockTemplatesController.php
```diff
@@ -25,27 +25,22 @@ class BlockTemplatesController {
...
 
 	/**
 	 * Constructor.
 	 */
 	public function __construct() {
 		// This feature is gated for WooCommerce versions 6.0.0 and above.
 		if ( defined( 'WC_VERSION' ) && version_compare( WC_VERSION, '6.0.0', '>=' ) ) {
-			$this->templates_directory      = plugin_dir_path( __DIR__ ) . 'templates/' . self::TEMPLATES_DIR_NAME;
-			$this->template_parts_directory = plugin_dir_path( __DIR__ ) . 'templates/' . self::TEMPLATE_PARTS_DIR_NAME;
+			$root_path = plugin_dir_path( __DIR__ ) . self::TEMPLATES_ROOT_DIR . DIRECTORY_SEPARATOR;
```
tjcafferkey, line 40: Can we not replace `self::TEMPLATES_ROOT_DIR` with `BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATES']` ?

sunyatasattva: Technically, we can because they have the same value. However, the reason why I added `TEMPLATES_ROOT_DIR` in the first place is because they represent two different things:

1. `BlockTemplateUtils::DIRECTORY_NAMES['TEMPLATES']` represents the `block-templates` directory name in themes.
2. `BlockTemplatesController::TEMPLATES_ROOT_DIR` represents the root directory of all templates in our own repo. For example, we have email templates.

The just happen to be both named `templates` now, so we have the awkward path `templates/templates` (which makes me sad :'(). But technically they are separate entities that could also be changed independently at any point.

Does that make any sense?

tjcafferkey:
>Does that make any sense?

Thought that would be the case, makes sense. Thanks!

PR #5465, opened 2021-12-28
Add global style for Active Filters block

gigitux:
This is a part of #4965.

This PR:

- Adds text color support

image
*editor side*

image
*frontend side*

### Manual Testing

How to test the changes in this PR:

Check out this branch:

1. Install and enable the `Gutenberg` plugin.
2. Install and enable the `TT1 Blocks` theme.
3. Add the Active Product Filters block block to a post. (you need to add a block like `Filters Product By Attribute`)
4. Go to Color section & verify you can change the text color.
5. Save your changes.
7. Reset to default using the `Clear` button from Styles > Text color window.
8. Go to Dashboard and select Appearance > Editor (beta).
10. Verify that the Active Product Filters block is shown Under the `Blocks` section .
11. Save your changes.
14. Change again the text color.
15. Save your changes.

### Changelog

> Added global styles (text color) to the Active Filters block.


assets/js/blocks/active-filters/block.js
```diff
@@ -117,19 +125,34 @@ const ActiveFiltersBlock = ( {
...
 	return (
 		<>
 			{ ! isEditor && blockAttributes.heading && (
-				<TagName className="wc-block-active-filters__title">
+				<TagName
+					className={ classnames(
+						'wc-block-active-filters__title',
+						textColorClass
+					) }
+					style={ textColorStyle }
```
Aljullu: Is it necessary to pass the `style` and the `textColorClass`? I would expect the heading element to simply inherit the color of the parent without us needing to do anything special.

assets/js/blocks/active-filters/block.js
```diff
@@ -117,19 +125,34 @@ const ActiveFiltersBlock = ( {
...
+				>
 					{ blockAttributes.heading }
 				</TagName>
 			) }
-			<div className="wc-block-active-filters">
+			<div
+				className={ classnames(
+					'wc-block-active-filters',
+					textColorClass
```
Aljullu: Same here, if we remove this class, would it just work? Color will be inherited from the parent div wrapper.

assets/js/blocks/active-filters/block.js
```diff
@@ -174,6 +197,8 @@ const ActiveFiltersBlock = ( {
 					} }
 				>
 					<Label
+						className={ textColorClass }
+						style={ textColorStyle }
```
Aljullu: And the same here. :slightly_smiling_face:

assets/js/blocks/active-filters/edit.js
```diff
@@ -2,20 +2,30 @@
...
 const Edit = ( { attributes, setAttributes } ) => {
 	const { className, displayStyle, heading, headingLevel } = attributes;
 
+	const blockProps = useBlockProps( {
+		className,
+	} );
+
+	const textColorClassAndStyle = getTextColorClassAndStyleFromAttributeObject(
+		attributes
+	);
```
Aljullu: Same here, I think we can remove those two variables and all styles will be applied by default.

Aljullu (commented):
This is working great! However, I think we can simplify the code. In blocks that save some elements in the save function, like this case, I think using `useBlockProps()` is enough, and then there is no need to parse the attributes to generate class names and styles.

gigitux:
> This is working great! However, I think we can simplify the code. In blocks that save some elements in the save function, like this case, I think using useBlockProps() is enough, and then there is no need to parse the attributes to generate class names and styles.

From your comments, I understand that you prefer that the child components inherit the style from the parent component and not that we pass the class or style to each child?

gigitux:
Thanks for letting me know.
I added `useBlockProps` this morning to fix the tests. Now I understand even better what they are for! 👍

assets/js/editor-components/block-title/index.js
```diff
@@ -17,10 +17,11 @@ const BlockTitle = ( {
 	onChange,
 	heading,
 	instanceId,
+	style,
 } ) => {
 	const TagName = `h${ headingLevel }`;
 	return (
-		<TagName className={ className }>
+		<TagName style={ style } className={ className }>
```
Aljullu: Nit: given that this prop is no longer used, I think we should remove it?

src/BlockTypesController.php
```diff
@@ -191,7 +191,7 @@ protected function get_block_types() {
 		/**
 		 * This disables specific blocks in Widget Areas by not registering them.
 		 */
-		if ( in_array( $pagenow, [ 'widgets.php', 'themes.php', 'customize.php' ], true ) ) {
+		if ( in_array( $pagenow, [ 'widgets.php', 'themes.php', 'customize.php' ], true ) && ! empty( 'gutenberg-edit-site' !== $_GET['page'] ) ) { // phpcs:ignore WordPress.Security.NonceVerification
```
Aljullu: Just to clarify, is this `empty()` to account for the case that `$_GET['page']` is not defined? In that case, I think you should check first that `$_GET['page']` is defined and then do the comparison. Something like this:
```PHP
&& ( empty( $_GET['page'] ) || 'gutenberg-edit-site' !== $_GET['page'] )
```

Aljullu (approved):
Thanks for updating this PR @gigitux!

> From your comments, I understand that you prefer that the child components inherit the style from the parent component and not that we pass the class or style to each child?

Right, from my understanding that's how global styles are expected to be used in most cases. The code becomes much simpler because we don't need to apply any 'custom solutions', all the magic is handled by Gutenberg. :slightly_smiling_face: 

I left a couple of comments but I'm pre-approving this PR because everything else is looking good.

PR #5471, opened 2021-12-29
Remove wpdb prepare statement in `get_category_review_count` method

### LINKED ISSUE #5159: Not passing args to `$wpdb->prepare` causes PHP notice to show which breaks API requests
## Describe the bug
When you have error reporting turned on, calls to `$wpdb->prepare` causes notices to print which breaks API requests, for example:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/a4602f8be133bcc6072cd691693f2ea0fffca0d1/src/StoreApi/Schemas/ProductCategorySchema.php#L115

is missing 2nd...nth arguments, which causes this error: `PHP Notice:  Undefined offset: 0 in /var/www/html/wp-includes/wp-db.php on line 1323`

We should check usages of this function and make it so we pass additional args to it.

## To reproduce

1. Ensure your site is set up to report notices
2. Try to add a Featured Category block to the editor
3. Notice the API errors
4. Check the network log and see the error `PHP Notice:  Undefined offset: 0 in /var/www/html/wp-includes/wp-db.php on line 1323` printed in the API response.

## Expected behavior
No notices should be printed in the API Response.

## Environment
Happens in both PHP 8.0 and 7.4

* Core version: [e.g. 5.0.3] 5.8.2
* Gutenberg version (if installed): [e.g. 4.8.0] n/a
* WooCommerce version: [e.g. 3.5.3] 5.9
* WooCommerce Blocks version: [e.g. 1.3.1] 6.3.0-dev
* Site language: en
* Any other plugins installed: nothing unusual

* OS: [e.g. macOS, Windows]
* Version [e.g. 22]

* Device: [e.g. iPhone6]
* OS: [e.g. iOS8.1]
* Version [e.g. 22]

mikejolley:
It's not valid to call `$wpdb->prepare()` without args. This PR removes it's usage in this way and replaces it with a single `esc_sql` call. Note, the data here is safe anyway because it's cast with `absint`.

Fixes #5159

### Testing

- Insert a featured category block
- Confirm the counts shown in the block are still functioning

### Changelog

> Remove invalid `$wpdb->prepare()` statement in Featured Category Block.

### DIFF
--- a/src/StoreApi/Schemas/ProductCategorySchema.php
+++ b/src/StoreApi/Schemas/ProductCategorySchema.php
@@ -111,14 +111,12 @@ class ProductCategorySchema extends TermSchema {
 			$terms_to_count_str = implode( ',', $terms_to_count );
 		}
 
-		// phpcs:disable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
-		$products_of_category_sql = $wpdb->prepare(
-			"SELECT SUM(comment_count) as review_count
+		$products_of_category_sql = "
+			SELECT SUM(comment_count) as review_count
 			FROM {$wpdb->posts} AS posts
 			INNER JOIN {$wpdb->term_relationships} AS term_relationships ON posts.ID = term_relationships.object_id
-			WHERE term_relationships.term_taxonomy_id IN ({$terms_to_count_str})"
-		);
-		// phpcs:enable WordPress.DB.PreparedSQL.InterpolatedNotPrepared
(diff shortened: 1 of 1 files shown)


opr (approved):
Works great, and the SQL looks OK - going to let @ralucaStan approve this too since it's a change to DB related code so we need two reviewers.

mikejolley:
@ralucaStan just needs your 👁️

ralucaStan (approved):
from what I can tell it looks 👍

PR #5472, opened 2021-12-30
Prevent duplicate hooks in hook docs

### LINKED ISSUE #5374: Duplicate actions and filters generate duplicate hook docs
## Describe the bug
When running the hook-docs script, `npm run build:docs`, if there are two actions in the same file (e.g. `woocommerce_after_main_content` action in [this file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/LegacyTemplate.php), the docs will output the same action twice.

## To reproduce

1. run `npm run build:docs`
2. [bin/hook-docs/data/actions.json](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/bin/hook-docs/data/actions.json) should have 2 entries for `woocommerce_after_main_content`.

## Expected behavior
Only 1 entry per action or filter hook expected in the docs.

mikejolley:
Prevents duplicate hooks by combining them by name.

- Updates the hook documentation script
- Combines hooks by name. If hooks are called in multiple files, it collects all file names and lists all files in the doc itself.
- Updates some docblocks in LegacyTemplates class so they are consistent 

Fixes #5374

### Testing

Run `npm run build:docs` and check the output. I've updated docs with this PR.

### DIFF
--- a/bin/hook-docs/actions/index.js
+++ b/bin/hook-docs/actions/index.js
@@ -15,6 +15,7 @@ const {
 	returns,
 	example,
 	related,
+	files,
 } = require( '../format-hook-doc' );
 const {
 	createDocs,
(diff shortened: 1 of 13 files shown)


opr (approved):
Thanks Mike, looks great! I've updated a file in this PR because it was causing ESLint to fail - when tests go green I'll merge 👌🏼 

For anyone else testing this you need to `composer install` to get the latest hooks-generator package.

PR #5473, opened 2021-12-30
Fix initial population of address data in `useCustomerData` hook

### LINKED ISSUE #5456: Checkout Block fields are not pre-filled via Customer payment page link for Subscription Renewal Orders
## Describe the bug
WC Block's Checkout block does not pre-fill the Checkout Block fields upon sharing the "Customer payment page" link. 

## To reproduce

Make sure the Checkout page has a Checkout block. 
Make sure to have WC SUB and SUB product.
Make sure to have a Subscriber to assign. 

1. Create an order with a subscription product (WooCommerce > Order > Create an order)
2. Create a subscription (WooCommerce > Subscriptions > Add Subscription)
3. Select the order created in step 1 as a Parent Order
4. Once the Subscription is created, create a renewal order "Create a pending renewal order"
5. Copy "Custoer payment page" link
6. Log in as the Subscriber or use plugin such as User Switching - https://wordpress.org/plugins/user-switching/
7. When pasting the link you copied, you will not see any pre-filled fields

## Expected behavior

The Checkout fields should be pre-filled for all the information such as First Last, Address, and Email (and Phone)

## Screenshots

Comparison of Checkout page via shortcode vs. Checkout Block

**WC Core Checkout page ` [woocommerce_checkout] `**
https://d.pr/i/VinEr5
Image link: https://d.pr/i/VinEr5

**WC Block Checkout block**
https://d.pr/i/ducYmU
Image link: https://d.pr/i/ducYmU

mikejolley:
Fixes the issue [described here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5456#issuecomment-1002825206) where address fields are blank on mount. Traced back to this change: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5394

While we don't want to update the server on the initial mount, we do need to sync initial address data with the cart once that's finished initialising. I believe this PR fixes both cases.

1. Changes `useCustomerData` so it tracks when the cart is finished loading, and then initialized itself with the updated address.
2. Added inline documentation to `useCustomerData` to make it easier to maintain.
3. Updated `useCustomerData` to use `useDebounceCallback` which is more easy to follow than debouncing several values.
2. Fixes a small issue in `useCheckoutAddress` which was also tracking the initial empty address.

Fixes #5456

### Testing

To avoid further regressions we need to test against all related issues.

1. While logged in, with a customer than has checked out before, visit the checkout page
2. Confirm shipping address fields are populated
3. Toggle the "different billing address' box. Confirm billing fields are populated.
4. Open the network inspector in browser tools
5. Repeat the test above. Open the checkout page. Confirm there are no extra requests to the `batch/` endpoint once the address data is populated.
6. Change an address field such as state or country. Confirm the cart sends a request to the server after a small 1000ms delay.
7. Smoke test the full checkout flow

### Changelog

> Fixed an issue where the checkout address fields would be blank for logged in customers

### DIFF
--- a/assets/js/base/context/hooks/use-checkout-address.js
+++ b/assets/js/base/context/hooks/use-checkout-address.js
@@ -27,7 +27,7 @@ export const useCheckoutAddress = () => {
 	} = useCustomerDataContext();
 
 	const currentShippingAsBilling = useRef( shippingAsBilling );
-	const previousBillingData = useRef( billingData );
+	const previousBillingData = useRef();
 
 	/**
 	 * Sets shipping address data, and also billing if using the same address.
@@ -71,7 +71,7 @@ export const useCheckoutAddress = () => {
 					email,
 					/* eslint-enable no-unused-vars */
 					...billingAddress
-				} = previousBillingData.current;
+				} = previousBillingData.current || billingData;
 
 				setBillingData( {
 					...billingAddress,
--- a/assets/js/base/context/hooks/use-customer-data.ts
+++ b/assets/js/base/context/hooks/use-customer-data.ts
@@ -4,7 +4,7 @@
 import { useDispatch } from '@wordpress/data';
 import { useEffect, useState, useCallback, useRef } from '@wordpress/element';
 import { CART_STORE_KEY as storeKey } from '@woocommerce/block-data';
-import { useDebounce } from 'use-debounce';
+import { useDebouncedCallback } from 'use-debounce';
 import isShallowEqual from '@wordpress/is-shallow-equal';
 import {
 	formatStoreApiErrorMessage,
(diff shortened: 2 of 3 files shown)


assets/js/base/context/hooks/use-checkout-address.js
```diff
@@ -27,7 +27,7 @@ export const useCheckoutAddress = () => {
 	} = useCustomerDataContext();
 
 	const currentShippingAsBilling = useRef( shippingAsBilling );
-	const previousBillingData = useRef( billingData );
+	const previousBillingData = useRef();
```
mikejolley, line 30: This prevents it storing the empty address data which will exist on mount.

ralucaStan (approved):

PR #5476, opened 2021-12-30
Fix duplicated error notices

### LINKED ISSUE #5475: Duplicate error notices in checkout when payment fails
Following issue was originally reported https://github.com/Automattic/woocommerce-payments/issues/3373
and as far as I can see is a result of `StoreNoticesProvider` wrappers in [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-payment-block/frontend.tsx#L51) and [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-payment-block/block.tsx#L16) which ends up duplicating the error notices. This is also reproducible in [WooCommerce Stripe Gateway](https://github.com/woocommerce/woocommerce-gateway-stripe) when used with blocks

### Describe the bug
When using the checkout block and UPE payment methods, two notices show up with the same message when the payment fails. 

### To Reproduce

1. Enable [WooCommerce blocks](https://woocommerce.com/products/woocommerce-gutenberg-products-block/)
2. Add the Checkout block on a page
3. In WCPay -> Settings, enable UPE
4. Go to the store and add a product to the cart
5. Go to the page with the Checkout block
6. Pay with a card that will be declined, like `4000 0000 0000 0002`
7. Notice that two notices show up with the same message
image

### Actual behavior

Two error notices show up when a UPE payment fails.

### Expected behavior

Only one notice should show up.

### Additional context
Mentioned in [this PR](https://github.com/Automattic/woocommerce-payments/pull/3260#issuecomment-964485530)

malithsen:
`StoreNoticeProvider` is already being used within `Block` component. This PR proposes its removal from `FrontendBlock` to prevent duplicate error notices mentioned in  #5475 . I am not quite sure if it should rather be removed from the (child) `Block` component instead.  

Fixes #5475 

### Screenshots

**Before**
image

**After**
Screen Shot 2021-12-30 at 12 15 43 PM

### Testing

1. Make sure you're using the Blocks checkout and WCPay.
2. Enable UPE from WCPay Dev Tools
2. Add 1 or more products to your cart.
3. Go to the blocks checkout page
4. Pay through the credit card form with a card that will be declined, e.g. 4000 0000 0000 0002.
5. Ensure that only one error notice is displayed 

### User Facing Testing
* [x] Same as above

### Changelog

> Fix duplicated checkout error notices.

### DIFF
--- a/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-payment-block/block.tsx
+++ b/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-payment-block/block.tsx
@@ -1,8 +1,6 @@
 /**
  * External dependencies
  */
-import { StoreNoticesProvider } from '@woocommerce/base-context';
-import { useEmitResponse } from '@woocommerce/base-context/hooks';
 
 /**
  * Internal dependencies
@@ -10,13 +8,7 @@ import { useEmitResponse } from '@woocommerce/base-context/hooks';
 import { PaymentMethods } from '../../../payment-methods';
 
 const Block = (): JSX.Element | null => {
-	const { noticeContexts } = useEmitResponse();
-
-	return (
-		<StoreNoticesProvider context={ noticeContexts.PAYMENTS }>
-			<PaymentMethods />
-		</StoreNoticesProvider>
-	);
+	return <PaymentMethods />;
 };
 
 export default Block;
--- a/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-payment-block/frontend.tsx
+++ b/assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-payment-block/frontend.tsx
@@ -2,13 +2,10 @@
  * External dependencies
  */
 import classnames from 'classnames';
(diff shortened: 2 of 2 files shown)


opr (changes requested):
I think there's a _slight_ performance benefit for removing it from the child block as you pointed out and leaving it in the frontend block.

Doing it this way `StoreNoticesProvider` won't load in the editor. It's not needed there as we don't show notices in the editor.

I also had to rebase this on trunk to fix an issue with the country code being passed through to WC pay which was fixed earlier.

malithsen:
Thanks for pointing out the performance improvement @opr

- Rebased on trunk
- Moved `StoreNoticeProvider` to frontend-block from block https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5476/commits/149461f985832b964e084eaf5755a287b651a3c5

opr (approved):
Great, thank you for making these changes this seems to be working well!

PR #5515, opened 2022-01-04
Product Title block: add support global style

### LINKED ISSUE #4965: Add global styles to all WC Blocks

gigitux:
This PR adds support for global style for the `Product Title` block.

- line-height
- font family
- font-weight
- text transform
- background-color
- margin
- font-size

Fixes #4965 

### Screenshots

image
*without global style*

image
*with global style*

### Manual Testing

Check out this branch:

1. Install and enable the `Gutenberg` plugin on WordPress 5.9 site.
3. Add the `Single Product block` (this block contains `Product Title Block`) to a post.
4. Get the focus on `Product Title block`.
8. Go to Dashboard and select Appearance > Editor (beta).
10. Verify that the `Product Title block` is shown under the `Blocks` section. Personalize again the block.
11. Save your changes.
14. Change again the styles.
15. Save your changes.

### Changelog

> Add support for the global style for the Product Title block.


assets/js/atomic/blocks/product-elements/title/index.ts
```diff
@@ -24,17 +24,25 @@ const blockConfig: BlockConfiguration = {
...
+			fontSize: true,
+			lineHeight: true,
+			...( isFeaturePluginBuild() && {
+				__experimentalFontWeight: true,
+				__experimentalTextTransform: true,
+				__experimentalFontFamily: true,
+			} ),
+		},
+		color: {
+			text: false,
```
dinhtungdu: Can you explain why text color is disabled here? Given that we allow changing the background, it makes more sense to me to allow changing text color too.

assets/js/utils/style-attributes-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+};
+
+const parseSpacingStyle = (
+	spacing: Record< string, unknown >
+): Record< string, unknown > => {
+	const keys = [ 'margin' ];
+
+	const getValueOrDefault = ( value: unknown ) => {
+		return isString( value ) && value.length > 0 ? value : '0';
+	};
```
dinhtungdu: I don't think we should return `0` as the default value here, it should be `undefined` instead. The element may have spacing rules defined in CSS, using `0` will override its styling.

assets/js/utils/style-attributes-utils.ts
```diff
@@ -0,0 +1,103 @@
...
+			return {
+				...acc,
+				[ key ]: `${ getValueOrDefault(
+					spacingProperty.top
+				) } ${ getValueOrDefault(
+					spacingProperty.right
+				) } ${ getValueOrDefault(
+					spacingProperty.bottom
+				) } ${ getValueOrDefault( spacingProperty.left ) }`,
+			};
```
dinhtungdu: Same as above, we should only use shorthand when all four sides are set, otherwise, we should use single side rules.

dinhtungdu (commented):
The block style is working for me 🎉 , both color and typography are working. But for the global style, the background color didn't work in my test. Please check the screencast for me detail.

https://user-images.githubusercontent.com/5423135/148203207-059f13e3-c74c-42a5-b569-d091ee277950.mov

assets/js/atomic/blocks/product-elements/title/index.ts
gigitux: > The block style is working for me 🎉 , both color and typography are working. But for the global style, the background color didn't work in my test. Please check the screencast for me detail.
> 
>  Screen.Recording.2022-01-05.at.16.49.06.mov

Mmm, it seems that it is a GB problem. What version are you using?

gigitux:
> The block style is working for me 🎉 , both color and typography are working. But for the global style, the background color didn't work in my test. Please check the screencast for me detail.
> 
>  Screen.Recording.2022-01-05.at.16.49.06.mov

Mmm, it seems that it is a GB problem. What version are you using?

dinhtungdu:
> Mmm, it seems that it is a GB problem. What version are you using?

@gigitux I used the [latest `trunk`](https://github.com/WordPress/gutenberg/tree/744c61a1fc57d80bc6630ca44f6c217ef43fa638) (I've just tested again).

assets/js/utils/style-attributes-utils.ts
gigitux: Thanks for this! I updated the function!

assets/js/atomic/blocks/product-elements/title/index.ts
gigitux: By #4965, the color for the text should be disabled, but I agree with you. 
I added the support for text color, too, now I will add a comment on the main issue.

gigitux:
> > Mmm, it seems that it is a GB problem. What version are you using?
> 
> @gigitux I used the [latest `trunk`](https://github.com/WordPress/gutenberg/tree/744c61a1fc57d80bc6630ca44f6c217ef43fa638) (I've just tested again).

Mmm, it is a very strange problem. 
https://user-images.githubusercontent.com/4463174/148233284-f73cd082-0da4-4fc1-b5ee-64e74c1ddd7a.mp4
If you see my video, I don't have this problem. Same with TT1 theme.

I updated the PR with your feedback. I also modified the CSS, because before I hadn't edited the save function to be compliant with apiVersion 2: now everything should be ok 👍

dinhtungdu:
> Mmm, it is a very strange problem.
> [user-images.githubusercontent.com/4463174/148233284-f73cd082-0da4-4fc1-b5ee-64e74c1ddd7a.mp4](https://user-images.githubusercontent.com/4463174/148233284-f73cd082-0da4-4fc1-b5ee-64e74c1ddd7a.mp4)
> If you see my video, I don't have this problem. Same with TT1 theme.

@gigitux Which WP version are you using?

gigitux:
> @gigitux Which WP version are you using?

WordPress 5.9 - beta 4. What about you?

assets/js/utils/style-attributes-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+
+		if ( keys.includes( key ) ) {
+			return {
+				...acc,
+				...getSpacingStyleInline( key, {
+					top: getValueOrDefault( spacingProperty.top ),
+					right: getValueOrDefault( spacingProperty.right ),
+					bottom: getValueOrDefault( spacingProperty.bottom ),
+					left: getValueOrDefault( spacingProperty.left ),
+				} ),
```
dinhtungdu:
```suggestion
				...getSpacingStyleInline( key, spacingProperty ),
```

dinhtungdu: I think we can pass spacingProperty as the second argument then handle sides there instead. It makes the callback look cleaner.

assets/js/utils/style-attributes-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+) => {
+	return {
+		...( isString( values.top ) && { [ `${ key }Top` ]: values.top } ),
+		...( isString( values.right ) && {
+			[ `${ key }Right` ]: values.right,
+		} ),
+		...( isString( values.bottom ) && {
+			[ `${ key }Bottom` ]: values.bottom,
+		} ),
+		...( isString( values.left ) && { [ `${ key }Left` ]: values.left } ),
```
dinhtungdu: Can we use key: value here then remove the null properties? We can even return style object with null properties because they will be omitted during render. Using spread here is quite hard to read IMO.

assets/js/utils/style-attributes-utils.ts
```diff
@@ -0,0 +1,123 @@
...
+			fontSize: attributesObject.fontSize || typography.fontSize,
+			lineHeight: typography.lineHeight,
+			fontWeight: typography.fontWeight,
+			textTransform: typography.textTransform,
+			fontFamily: attributesObject.fontFamily,
+		},
+	};
+};
+
+export const useColorProps = ( attributes: unknown ): WithStyle & WithClass => {
```
dinhtungdu: Can we use the correct type of `attributes` here instead of `unknow`? I don't see we're passing string to this hook anywhere. By doing so we don't need this line:

```
const attributesObject = isObject( attributes ) ? attributes : {};
```

dinhtungdu (commented):
@gigitux With a fresh installation, the global style works as expected now. I left some comments about the utilities, please take a look.

assets/js/utils/style-attributes-utils.ts
gigitux: The problem is that this code is also used on javascript code (so without types), so the check should be done on the js side. I preferred to do it here, so as not to repeat the same check in all the blocks.

assets/js/utils/style-attributes-utils.ts
gigitux: > I think we can pass spacingProperty as the second argument then handle sides there instead. It makes the callback look cleaner.

It can be an idea, but to be honest I prefer to create a more strict typed code. 

image


With this code (so without type), we lose some information about how the function `getSpacingStyleInline` works.
I mean without an object with `top`, `right`, `bottom`, and `left` key, it doesn't make sense invoke getSpacingStyleInline` function. With the type `Record<string, unknown>` we lose this information and the developer is forced to read the whole function to understand what kind of data to pass.

But this is just my opinion, if you think that it is better to change it, I will do it :D

dinhtungdu: @gigitux With your example, I don't see the need for `getSpacingStyleInline`, we can add the inline style directly like this. Note that we can use `undefined` properties just fine because they will get omitted during render.

```
		if ( keys.includes( key ) ) {
			return {
				...acc,
				[ `${ key }Left` ]: spacingProperty.left,
				[ `${ key }Right` ]: spacingProperty.right,
				[ `${ key }Top` ]: spacingProperty.top,
				[ `${ key }Bottom` ]: spacingProperty.bottom,
			};
		}
```

dinhtungdu:
@gigitux Looking at https://github.com/WordPress/gutenberg/blob/0666dee9235aedd3d1fa51c733d0c1e146227452/packages/block-library/src/button/edit.js#L26, I'm wondering if we can reuse the spacing handler in Gutenberg instead of creating our own utilities? Like what we do with `useColorProps`.

Edit: By the way, we're creating custom hooks, so `style-attributes-utils.ts` may not be the best place for those hooks.

gigitux:
> @gigitux Looking at https://github.com/WordPress/gutenberg/blob/0666dee9235aedd3d1fa51c733d0c1e146227452/packages/block-library/src/button/edit.js#L26, I'm wondering if we can reuse the spacing handler in Gutenberg instead of creating our own utilities? Like what we do with `useColorProps`.
> 
> Edit: By the way, we're creating custom hooks, so `style-attributes-utils.ts` may not be the best place for those hooks.

Thanks for pointing this out. With my first experiments, it returned to me an empty object. Now I have managed to get it to work correctly.

I moved the hooks under the path `/js/style-attributes.ts`. Let me know if what do you think 👍 

Update: Unfortunately the API `__experimentalGetSpacingClassesAndStyles` is not available on WP 5.8. What should we do? Should we maintain the previous code for WordPress 5.8 and use the new hook for WordPress 5.9?

dinhtungdu (approved):
@gigitux Great work on this PR 👏🏽 !

> Update: Unfortunately the API __experimentalGetSpacingClassesAndStyles is not available on WP 5.8. What should we do? Should we maintain the previous code for WordPress 5.8 and use the new hook for WordPress 5.9?

As long as our custom hook is feature-gated and no fatal error throws on WP 5.8, I think we can just focus on 5.9, it's not worth the effort to support 5.8 given that most of the global style functionalities are available from 5.9. cc @Aljullu for more insights : ).

gigitux:
> As long as our custom hook is feature-gated and no fatal error throws on WP 5.8, I think we can just focus on 5.9, it's not worth the effort to support 5.8 given that most of the global style functionalities are available from 5.9. cc @Aljullu for more insights : ).

Mmm, we have got a fatal error on WP 5.8 (without GB)
image

The error:
image


This bug is caused by `__experimentalGetSpacingClassesAndStyles`

Aljullu:
> As long as our custom hook is feature-gated and no fatal error throws on WP 5.8, I think we can just focus on 5.9, it's not worth the effort to support 5.8 given that most of the global style functionalities are available from 5.9. cc @Aljullu for more insights : ).

Agree with this. :slightly_smiling_face: 

> This bug is caused by `__experimentalGetSpacingClassesAndStyles`

Would it be possible to check if `__experimentalGetSpacingClassesAndStyles` exists at [this point](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5515/files#diff-50fcc3107b21b78495f2049855c06d3f7624590b712a9174078cbd90004b70e9R37), and return early if it doesn't?

Then, we might probably need to do the same check [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5515/files#diff-bc6b772c8d9baf82eae8ac09948b4e392e5b526b1d9550c96d18d493a9e17738R30), so controls don't appear in the editor if the function doesn't exist.

gigitux:
> Would it be possible to check if `__experimentalGetSpacingClassesAndStyles` exists at [this point](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5515/files#diff-50fcc3107b21b78495f2049855c06d3f7624590b712a9174078cbd90004b70e9R37), and return early if it doesn't?
> 
> Then, we might probably need to do the same check [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5515/files#diff-bc6b772c8d9baf82eae8ac09948b4e392e5b526b1d9550c96d18d493a9e17738R30), so controls don't appear in the editor if the function doesn't exist.

Yes! Sure! I will add this check! Thanks for the feedback!

gigitux:
@Aljullu @dinhtungdu I just added a utility function [18642e6](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5515/commits/18642e60becfec59e289471ccae74275e6bc3de9) to check if there is support for the spacing style support.

Let me know, what do you think!

assets/js/atomic/blocks/product-elements/title/index.ts
```diff
@@ -36,15 +37,16 @@ const blockConfig: BlockConfiguration = {
...
-			color: {
-				text: true,
-				background: true,
-				link: false,
-				gradients: true,
-				__experimentalSkipSerialization: true,
-			},
-		} ),
+		...( isFeaturePluginBuild() &&
+			hasSpacingStyleSupport() && {
```
dinhtungdu: @gigitux Isn't that check should be used for spacing below?

gigitux: Update! Thanks for the double-check!

assets/js/atomic/blocks/product-elements/title/index.ts
```diff
@@ -24,17 +26,34 @@ const blockConfig: BlockConfiguration = {
...
+				spacing: {
+					margin: true,
+					__experimentalSkipSerialization: true,
 				},
-				typography: {
-					fontSize: true,
-				},
-		  }
-		: sharedConfig.supports,
+			} ),
```
dinhtungdu: Did `html: false` get removed on purpose? 

By the way, this part can be rewritten to:

```
		...sharedConfig.supports,
		...( isFeaturePluginBuild() && {
			typography: {
				fontSize: true,
				lineHeight: true,
				__experimentalFontWeight: true,
				__experimentalTextTransform: true,
				__experimentalFontFamily: true,
			},
			color: {
				text: true,
				background: true,
				link: false,
				gradients: true,
				__experimentalSkipSerialization: true,
			},
			...( hasSpacingStyleSupport() && {
				spacing: {
					margin: true,
					__experimentalSkipSerialization: true,
				},
			} ),
		} ),
```

dinhtungdu (commented):
@gigitux In my test with the latest `trunk` of Gutenberg, WP 5.9 beta 4, and WC 6.1.0-dev, the text color works for block style but doesn't work for global style.

https://user-images.githubusercontent.com/5423135/148975227-d0362ea4-8160-4606-b969-40bc3d384abb.mov

gigitux:
> @gigitux In my test with the latest `trunk` of Gutenberg, WP 5.9 beta 4, and WC 6.1.0-dev, the text color works for block style but doesn't work for global style.
> 
>  Screen.Recording.2022-01-11.at.22.45.42.mov

Thanks for your feedback. If you save the custom color and you check the frontend, you will be able to see the custom color.

The problem, in this case, is that Gutenberg injects custom style for all anchor elements on global style
image

We are not sure that this element is an `anchor` element because the user can enable or disable the link feature: if we want to fix the problem, we should enable both `link` and `text` color. To be honest, I think that it is not a great UX for the user, but from what I understand in this weeks there isn't an alternative.

dinhtungdu:
> The problem, in this case, is that Gutenberg injects custom style for all anchor elements on global style

@gigitux I opened #5548 to fix this issue by overriding the style for the anchor of the title in the editor. Please take a look!

> We are not sure that this element is an `anchor` element because the user can enable or disable the link feature: if we want to fix the problem, we should enable both `link` and `text` color. To be honest, I think that it is not a great UX for the user, but from what I understand in this weeks there isn't an alternative.

I think the title should use the same color setting regardless it's a link or not. Enable `link` color makes the block more complex.

gigitux:
> I opened #5548 to fix this issue by overriding the style for the anchor of the title in the editor. Please take a look!

Thanks for the help! I didn't know that the `editor.scss` is loaded on the FSE editor page too. Thanks for the help!

Let me know if I can merge this PR!

dinhtungdu (approved):
:shipit:

PR #5519, opened 2022-01-05
Update WooCommerce plugin slug for Block Templates

### LINKED ISSUE #5327: WooCommerce templates: two 404 requests when loading the templates list

### LINKED ISSUE #5423: Store Editing Templates: Currently using the incorrect slug for WooCommerce
## Describe the bug

Amongst the code for the WooCommerce Store Editing Templates we are using the id/slug `woocommerce` or `WooCommerce` for our block templates. We've recently discovered the correct value should possibly be `woocommerce/woocommerce`and have noticed a number of issues because of this:

* On the frontend when the component tries to [fetch the WooCommerce plugin data by slug](https://github.com/WordPress/gutenberg/blob/trunk/packages/edit-site/src/components/list/added-by.js#L61-L64), it fails to get anything.
* XHR requests are being sent to incorrect endpoints such as `/wp-json/wp/v2/plugins/WooCommerce` and getting a 404 response, when the should be going to `/wp-json/wp/v2/plugins/woocommerce/woocommerce`.
* In the database under the table `wp_terms` we're saving this incorrect slug as `woocommerce` (under `name` and `slug` columns) when it should be `woocommerce/woocommerce` (name) and `woocommerce-woocommerce` (slug).

As a result of this, if we were to simply make some code amendments and update our usages to use the correct value, this means that users who have already modified their WooCommerce templates would no longer see them listed in the Templates list view, and they would also not be found when fetching them to render on the frontend because we'd be looking for the new value (`woocommerce/woocommerce`) when searching the DB for templates saved against the old value (`woocommerce`)

Related https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5420
Related https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5412
Related https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5375
Related https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5327

tjcafferkey:
#### Description

Use the correct WooCommerce plugin slug `woocommerce/woocommerce` instead of the current incorrect one `woocommerce`. However, we are maintaining the old slug until we deploy a migration script to update the DB entires.

#### Consequences of this approach

**If the site has DB entries using the old slug _and_ the new slug:**
We will have to find the `term_id` for the old slug and update all entries using it to use the _new_ slugs `term_id`.

**If the site has DB entries just using the old slug:**
We can simply update the old slug row in `wp_terms` 

From:
name: `WooCommerce` 
slug: `woocommerce`

To:
name:`woocommerce/woocommerce` 
slug: `woocommerce-woocommerce`

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5423
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5327

### Testing

Note: This is mostly regression testing.

1. **Before checking out this PR**, customise the Single Product template so you have a DB entry with the old slug.
2. Checkout this PR, ensure on the Site Editor > Templates list you are still seeing the customised template in this list and it renders correctly on the frontend, and in site editor.
3. Now customise the Product Archive template and ensure on the Site Editor > Templates list you are still seeing the customised template in this list and it renders correctly on the frontend, and in site editor.
4. Clear the customisations of the Product Archive template to ensure this functionality also works correctly.
5. Create a `archive-product.html` template in your theme and ensure this overrides the default one provided by WooCommerce. Check this also renders correctly on the frontend and within the site editor.
6. Customise the newly created `archive-product.html` provided by your theme,  ensure on the Site Editor > Templates list you are still seeing the customised template in this list and it renders correctly on the frontend, and in site editor.
7. Clear the customisations of the themes `archive-product.html` template to ensure this works correctly.
8. Check that we're now not receiving 404 network requests for WooCommerce templates on the Site Editor templates list page as described here https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5327

### Performance Impact

We are querying both the old and the new slugs for customised templates which have been saved in the database. This will be resolved later using a migration script described above in this PR.

### DIFF
--- a/src/BlockTemplatesController.php
+++ b/src/BlockTemplatesController.php
@@ -105,8 +105,8 @@ class BlockTemplatesController {
 		// been unhooked so won't run again.
 		add_filter( 'get_block_file_template', array( $this, 'get_single_block_template' ), 10, 3 );
 		$maybe_template = function_exists( 'gutenberg_get_block_template' ) ?
-			gutenberg_get_block_template( 'woocommerce//' . $slug, $template_type ) :
-			get_block_template( 'woocommerce//' . $slug, $template_type );
+			gutenberg_get_block_template( BlockTemplateUtils::PLUGIN_SLUG . '//' . $slug, $template_type ) :
+			get_block_template( BlockTemplateUtils::PLUGIN_SLUG . '//' . $slug, $template_type );
 
 		// Re-hook this function, it was only unhooked to stop recursion.
 		add_filter( 'pre_get_block_file_template', array( $this, 'maybe_return_blocks_template' ), 10, 3 );
@@ -270,6 +270,11 @@ class BlockTemplatesController {
 	 * @return int[]|\WP_Post[] An array of found templates.
 	 */
 	public function get_block_templates_from_db( $slugs = array(), $template_type = 'wp_template' ) {
+		// This was the previously incorrect slug used to save DB templates against.
+		// To maintain compatibility with users sites who have already customised WooCommerce block templates using this slug we have to still use it to query those.
+		// More context found here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5423.
+		$invalid_plugin_slug = 'woocommerce';
+
 		$check_query_args = array(
 			'post_type'      => $template_type,
 			'posts_per_page' => -1,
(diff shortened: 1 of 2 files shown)


Aljullu (approved):
This is testing well and code looks good, thanks for investigating @tjcafferkey!

>**If the site has DB entries using the old slug _and_ the new slug:**
>We will have to find the `term_id` for the old slug and update all entries using it to use the _new_ slugs `term_id`.
>
>**If the site has DB entries just using the old slug:**
>We can simply update the old slug row in `wp_terms` 

To help me understand the urgency of this: with your fix in this PR, we are reading templates from the old and the new slugs, right? If that's the case, my understanding is that the database migration is not be really urgent, correct? It would just be a clean-up so we don't need to account for the old slug in new code we write, but that's all. Am I right?

tjcafferkey:
>Am I right?

Yeah that is correct @Aljullu. It's no longer as urgent and likely won't effect that many sites assuming the feature isn't widely adopted yet.

I will merge this, and create a new issue about this migration and reference this PR and the original issue for context.

Thanks for testing!

PR #5525, opened 2022-01-06
Stock Indicator block: Add support for global style

gigitux:
~This PR is blocked by #5515~

This PR adds support for global style for the `Summary Product block`.

- text-color
- font-size

#4965 

### Screenshots

image
*without global style*

image
*with global style*

### Manual Testing

Check out this branch:

1.  On `WordPress 5.9`, install and enable the `Gutenberg` plugin.
3. Add the `Single Product block` (this block contains `Stock Indicator block`) to a post.
4. Get the focus on `Stock Indicator block`.
10. Verify that the `Stock Indicator block` is shown under the `Blocks` section. Personalize again the block.
11. Save your changes.
14. Change again the styles.
15. Save your changes.

### Changelog

> Add support for the global style for the Stock Indicator block.


assets/js/atomic/blocks/product-elements/sale-badge/save.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const Save = ( { attributes }: Props ): JSX.Element => {
+	return (
+		<div
+			{ ...useBlockProps.save( {
+				className: classnames( 'is-loading', attributes.className ),
+			} ) }
+		/>
+	);
+};
```
dinhtungdu: This code is for the Sale badge block. Was it added by accident?

assets/js/atomic/blocks/product-elements/stock-indicator/style.scss
```diff
@@ -3,15 +3,4 @@
...
-	&--in-stock {
-		color: $in-stock-color;
-	}
-	&--out-of-stock {
-		color: $no-stock-color;
-	}
-	&--low-stock,
-	&--available-on-backorder {
-		color: $low-stock-color;
-	}
```
dinhtungdu, line 16: We are removing the feature that store owners actually need here. The current Supports API doesn't allow multiple colors, so I don't think it's a good idea to remove styles for stock statuses. Instead, we should accept the limitation of the Supports API. In the other words, ignoring the color, only add global style support for typography.

We can also create custom inspector settings to control colors for each state, and it will be local block style only, not global style.

assets/js/atomic/blocks/product-elements/stock-indicator/supports.js
```diff
@@ -0,0 +1,17 @@
...
+export const supports = {
+	...( isFeaturePluginBuild() && {
+		color: {
+			text: true,
+			background: false,
+			link: false,
+		},
+	} ),
+	typography: {
+		fontSize: true,
```
dinhtungdu, line 15: It makes sense to me to support all typography properties for this block instead of only font size. At least, font-weight and text-transform should be supported.

dinhtungdu (commented):
Block and global styles were both working in my test. But I have some concerns about the color based on stock statuses. Please check the review comments for more detail.

assets/js/atomic/blocks/product-elements/stock-indicator/style.scss
gigitux: > We are removing the feature that store owners actually need here. The current Supports API doesn't allow multiple colors, so I don't think it's a good idea to remove styles for stock statuses. 

Before proceeding, I asked the rest of the team. Please check this discussion (p1641461102003800-slack-C02FL3X7KR6), maybe we can continue the discussion there.

assets/js/atomic/blocks/product-elements/stock-indicator/supports.js
gigitux: According to main issue #4965, we should enable only font-size, but yeah we can enable them easily. @Aljullu, what do you think?

assets/js/atomic/blocks/product-elements/sale-badge/save.tsx
gigitux: Fixed, thanks!

dinhtungdu (approved):
> According to main issue #4965, we should enable only font-size, but yeah we can enable them easily.

I agree to follow the requirements. We can merge this PR as it is for now and add more properties later. By doing so, we don't block the release of the Global Style project.

gigitux:
Thanks for the review!

assets/js/atomic/blocks/product-elements/stock-indicator/supports.js
Aljullu: Yes, makes sense to me. I tested the Post Categories block from Gutenberg and it offers many more typography options than what we offer in Stock indicator.

I guess the same could be said about the Product Category List, Product Summary and Product Tag List.

@vivialice what are your thoughts on this? Should we enable all typography properties in those blocks?

PR #5527, opened 2022-01-06
Allow inserting blocks to filled mini cart

### LINKED ISSUE #4856: Mini Cart block: allow extensions to insert blocks in certain areas
In #4855, the Mini Cart block contents will be converted into individual blocks. They will be locked so it won't be possible to move them around neither to delete them, but we will allow extensions to insert blocks in between.

These mockups show some of the things that should be possible for extensions to achieve:

drawer-mini-cart-_-product-recommendation-01-2

drawer-mini-cart-_-discount-code-01-2

That means we will need to add at least two areas where extensions can insert blocks: before the product list and before the proceed to Cart/Checkout buttons.

AFAIK some work was done in that direction in Checkout i2, so we should use that as a reference.

dinhtungdu:
Fixes #4856 

This PR allows inserting blocks to the filled mini cart content view.
- The Title and Footer blocks are locked to the top and bottom of the filled view. 
- A new block called Mini Cart Items was created. The Products Table block now becomes an inner block of the Mini Cart Items.
- Users can add more blocks to the Mini Cart Items.
- Inner blocks of the Mini Cart Items can be moved, including the Products Table block.
- The Products Table block always takes all of the remaining vertical space to match the design.
- The Mini Cart Title font size now is the same between the editor and the front-end.

This PR also fixes the block toolbar position issue by removing `__experimentalExposeControlsToChildren` support.

### Screenshots

image

### Manual Testing

1. Clear customization for the Mini Cart template part.
2. Edit the Mini Cart template parts.
3. See the Block Appender of the Mini Cart Items work as expected.
4. See the inner blocks of the Mini Cart Items block can be moved up and down.
5. See the Products Table block that can't be removed.
6. Save the template part and see it works correctly on the front end.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.


sunyatasattva (approved):
LGTM! And know that I see and appreciate you putting CSS props in alphabetical order
:feelsgood:

PR #5528, opened 2022-01-06
Tag List block: Add support for global style 

gigitux:
~~This PR is blocked by #5515~~

This PR adds support for global style for the `Tag List block`.

- text-color
- link-color
- font-size

#4965 

### Screenshots

image
*without global style*

image
*with global style*

### Manual Testing

Check out this branch:

3. Add the `Single Product block` (this block contains `Tag List block`) to a post.
4. Get the focus on `Tag List block`.
10. Verify that the `Tag List block` is shown under the `Blocks` section. Personalize again the block.
11. Save your changes.
14. Change again the styles.
15. Save your changes.

### Changelog

> Add support for the global style for the Tag List block.


assets/js/atomic/blocks/product-elements/tag-list/index.js
```diff
@@ -14,13 +14,18 @@ import {
 	BLOCK_ICON as icon,
 	BLOCK_DESCRIPTION as description,
 } from './constants';
+import { Save } from './save';
+import { supports } from './supports';
```
dinhtungdu, line 18: I'm not sure why should we use a separated file instead of using `index.ts`?

dinhtungdu (approved):
Working great! Left an option comment but this is ready to be merged!

assets/js/atomic/blocks/product-elements/tag-list/index.js
gigitux: Just a stylistic choice. I preferred to keep the `index.js` short

PR #5529, opened 2022-01-06
Add margin below shipping packages in checkout

### LINKED ISSUE #5279: Checkout block: Spacing issue when having different shipping options
## Describe the bug
The space between the different shipping options is too small

## To reproduce

1. Install Woo Blocks and Woo Subscriptions
2. Add Simple subscription product with a subscription price and a signup fee
3. Go to the front-end and add the subscription price to the cart
4. Go to the Checkout block and look at the shipping options, notice that there isn't any visual separation between the two

## Expected behavior
There should vertical spacing

## Screenshots

Screenshot 2021-11-30 at 11 18 11

opr:
This PR adds margin below the `wc-block-components-shipping-rates-control__package` class, but not on the last of type to prevent this from affecting the parent component's spacing.

Fixes #5279

### Screenshots

| Before | After |
|---|---|
| image | image |

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Install Woo Blocks and Woo Subscriptions.
2. Add Simple subscription product with a subscription price and a signup fee.
3. Go to the front-end and add the subscription price to the cart.
4. Go to the Checkout block and look at the shipping options, notice that there is a gap between the two shipping options.

### User Facing Testing
* [x] Same as above

### Changelog

>  Fixed a styling issue in the Checkout block when an order has multiple shipping packages.

### DIFF
--- a/assets/js/base/components/cart-checkout/shipping-rates-control-package/style.scss
+++ b/assets/js/base/components/cart-checkout/shipping-rates-control-package/style.scss
@@ -1,9 +1,4 @@
 .wc-block-components-shipping-rates-control__package {
-	margin-bottom: em($gap-large);
-
-	&:last-of-type {
-		margin-bottom: 0;
-	}
 
 	.wc-block-components-panel__button {
 		margin-bottom: 0;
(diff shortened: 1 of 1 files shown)


alexflorisca (changes requested):
This looks ok on the checkout page, but it also introduces a disproportionate gap on the cart page. The space between the 3 titles should be equal (see screenshot)
Screenshot 2022-01-07 at 11 57 02

Could we add some extra specificity here to pin this to the checkout block only?

opr:
> Could we add some extra specificity here to pin this to the checkout block only?

Absolutely, cheers for the review, I've updated it now :)

alexflorisca (approved):
Nice one, looks good!

PR #5531, opened 2022-01-07
Fix Search Product placeholder's contrast

### LINKED ISSUE #1683: Product Search (editor): placeholder doesn't have enough contrast
imatge

Placeholder is `#828B96` while background is white, that combination doesn't have enough contrast:

https://webaim.org/resources/contrastchecker/?fcolor=828B96&bcolor=FFFFFF

We should replace it with another color. Ideally, one in our `_colors.scss` file:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/css/abstracts/_colors.scss

Or one of the colors we consume upstream:

https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/_colors.scss
https://github.com/Automattic/color-studio/blob/master/dist/color-variables.scss

tarhi-saad:
It was reported in this issue: #1683, that the placeholder of the Product Search block doesn't have enough contrast.

After looking further into it, I discovered the following:

- In the editor mode, we don't have the contrast issue anymore, but we are relying on the colors set by the WordPress core for our placeholders (e.g. we are getting our placeholder colors from `/wp-includes/css/dist/components/style.css` & `/wp-admin/css/forms.css`).
- In the preview or the live version mode of our Product Search block, the default placeholder color of the browser is applied instead. The reason for that is the class name change of the input element.

### Issue

The issue here is that we shouldn't rely on the CSS applied by the WordPress core to style our blocks on the editor mode, and we shouldn't rely on the browser default styles either.
For example, in this case, the latest version of the Safari browser is applying the following placeholder color: `#A9A9A9` while the background is white, and this combination is failing every contrast test:
[https://webaim.org/resources/contrastchecker/?fcolor=A9A9A9&bcolor=FFFFFF](https://webaim.org/resources/contrastchecker/?fcolor=A9A9A9&bcolor=FFFFFF)

image

### Fix

Set a placeholder color that passes the contrast test for the Product Search block.

### Another approach

While inspecting the codebase of the WooCommerce Blocks it seems that we aren't setting any default colors for our placeholders in all of our blocks. 
A better solution would have been setting a default placeholder color globally instead of fixing this issue for the Product Search alone. But, since I'm not yet familiar with this product and to be safe I choose the latter option.

### Manual Testing
The theme used for this test: "Twenty Twenty-One".

1. Checkout the branch
2. Create a test page.
3. Add the "Search Product" block.
4. Publish the page
5. Check the live version of this page on different browsers (e.g. Chrome, Safari, etc.)
6. Inspect the search input and check that our WooCommerce Blocks plugin is applying its own placeholder color (in wc-blocks-style.css).
7. Test contrast on this [contrast checker](https://webaim.org/resources/contrastchecker/) tool by providing the placeholder's color. It should pass on white background.

### Changelog

> Fix the default placeholder color contrast for the Product Search block

### DIFF
--- a/assets/js/blocks/product-search/style.scss
+++ b/assets/js/blocks/product-search/style.scss
@@ -1,8 +1,4 @@
 .wc-block-product-search {
-	@include placeholder-auto-prefixer {
-		color: $placeholder-color;
-	}
-
 	.wc-block-product-search__fields {
 		display: flex;
 	}
@@ -10,6 +6,11 @@
 		padding: 6px 8px;
 		line-height: 1.8;
 		flex-grow: 1;
+
+		&::placeholder,
+		input::placeholder {
+			color: $placeholder-color;
+		}
 	}
 	.wc-block-product-search__button {
 		display: flex;
--- a/assets/js/blocks/product-search/style.scss
+++ b/assets/js/blocks/product-search/style.scss
@@ -1,4 +1,8 @@
 .wc-block-product-search {
+	@include placeholder-auto-prefixer {
+		color: $placeholder-color;
+	}
+
 	.wc-block-product-search__fields {
 		display: flex;
 	}
--- a/assets/css/abstracts/_colors.scss
+++ b/assets/css/abstracts/_colors.scss
@@ -7,7 +7,7 @@ $low-stock-color: $alert-yellow;
 $in-stock-color: $alert-green;
 $discount-color: $alert-green;
 
-$placeholder-color: var(--global--color-primary, $gray-200);
+$placeholder-color: var(--global--color-primary, $gray-700);
 $input-border-gray: #50575e;
 $input-border-dark: rgba(255, 255, 255, 0.4);
 $input-disabled-dark: rgba(255, 255, 255, 0.3);
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -6,6 +6,10 @@ $fontSizes: (
 	"larger": 2,
 );
 
+$placeholders: "-webkit-input-placeholder",
+	"-moz-placeholder",
+	"-ms-input-placeholder";
+
 // Maps a named font-size to its predefined size. Units default to em, but can
 // be changed using the multiplier parameter.
 @mixin font-size($sizeName, $multiplier: 1em) {
@@ -215,4 +219,30 @@ $fontSizes: (
 
 	$hex: str-slice(ie-hex-str($color), 4);
 	@return "%23" + unquote("#{$hex}");
+}
+
+@mixin placeholder-auto-prefixer {
+	@each $placeholder in $placeholders {
+		@if $placeholder == "-webkit-input-placeholder" {
+			::#{$placeholder} {
+				@content;
+			}
+		}
+		@else if $placeholder == "-moz-placeholder" {
+			// FF 18-
+			:#{$placeholder} {
+				@content;
+			}
+
+			// FF 19+
+			::#{$placeholder} {
+				@content;
+			}
+		}
+		@else {
+			:#{$placeholder} {
+				@content;
+			}
+		}
+	}
 }
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -6,10 +6,6 @@ $fontSizes: (
 	"larger": 2,
 );
 
-$placeholders: "-webkit-input-placeholder",
-	"-moz-placeholder",
-	"-ms-input-placeholder";
-
 // Maps a named font-size to its predefined size. Units default to em, but can
 // be changed using the multiplier parameter.
 @mixin font-size($sizeName, $multiplier: 1em) {
@@ -219,30 +215,4 @@ $placeholders: "-webkit-input-placeholder",
 
 	$hex: str-slice(ie-hex-str($color), 4);
 	@return "%23" + unquote("#{$hex}");
-}
-
-@mixin placeholder-auto-prefixer {
-	@each $placeholder in $placeholders {
-		@if $placeholder == "-webkit-input-placeholder" {
-			::#{$placeholder} {
-				@content;
-			}
-		}
-		@else if $placeholder == "-moz-placeholder" {
-			// FF 18-
-			:#{$placeholder} {
-				@content;
-			}
-
-			// FF 19+
-			::#{$placeholder} {
-				@content;
-			}
-		}
-		@else {
-			:#{$placeholder} {
-				@content;
-			}
-		}
-	}
 }


assets/js/blocks/product-search/style.scss
```diff
@@ -1,4 +1,8 @@
 .wc-block-product-search {
+	@include placeholder-auto-prefixer {
```
nielslange: Kudos for going above and beyond in this case! 👏 It's great to see that you took all the vendor prefixes into consideration. That said, as we're using the [autoprefixer dependency](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2cb9a483762cd0d642ac5583ccaa4184baf25374/package.json#L133), there's no need to create a placeholder mixin. Instead, you can simply write:

```css
.wc-block-product-search::placeholder {
  color:green;
}
```

This then gets compiled to:

```css
.wc-block-product-search::-webkit-input-placeholder {
  color:green;
}
.wc-block-product-search::-moz-placeholder {
  color:green;
}
.wc-block-product-search:-ms-input-placeholder {
  color:green;
}
.wc-block-product-search::-ms-input-placeholder {
  color:green;
}
.wc-block-product-search::placeholder {
  color:green;
}
```

⚠️ Please note that I used `color:green;` only for demonstration purposes. Of course, you'd use your color definition there.

To keep this PR lean and the affected section maintainable, I suggest using the Autoprefixer instead of your manual mixin. And in case you're interested to see the Autoprefixer in action, you can find a playground on https://autoprefixer.github.io/.

nielslange (changes requested):
Besides the feedback, I added regarding the Autoprefixer, I also noticed that two sections are missing from the PR description:

1. Testing
2. Changelog

The testing steps allow both a reviewer and an external tester to simply test this PR without thinking too much on how to test it.

The changelog is relevant for the release notes. There might be cases in which the change should not be mentioned in the changelog. In this case, we do not add a changelog description and we add the label `skip-changelog`.

On https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5541, you can see a PR with both testing steps and the label `skip-changelog`. And on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5542, you can find an example of a changelog description. Finally, on https://developer.woocommerce.com/2022/01/04/woocommerce-blocks-6-7-0-release-notes/, you can find a release post, which shows all the changelog entries from the PR. 😀

assets/js/blocks/product-search/style.scss
tarhi-saad: @nielslange, thank you for your great review! And for all the information you provided. 👍

tarhi-saad:
@nielslange I made the changes suggested above & updated the first PR message by including the missing section (e.g. Testing) and added the "skip-changelog" tag.

Applied the placeholder color differently in this last commit:

```SCSS
.wc-block-product-search__field {
  ... // Some code above

  &::placeholder,
  input::placeholder {
    color: $placeholder-color;
  }
}
```

Our Search Product block has different HTML structures in the editor and the live version modes. The HTML element with the class name `wc-block-product-search__field` is an `input` element in the live version, but in the editor mode, it's a `div` element. So, to target our `input` in both modes the above selector was needed.

nielslange (changes requested):
Thanks for the change, @tarhi-saad. The change looks much cleaner now. I also checked the compatibility of the `::placeholder` pseudo-element via https://caniuse.com/css-placeholder and based on https://make.wordpress.org/core/handbook/best-practices/browser-support/, we can use this pseudo-element without any vendor prefix. 

I also checked your PR against various themes and noticed that your change works fine with the Storefront and the Twenty Twenty-One themes, but it does not work with core themes before the Twenty Twenty-One theme. 

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b72efda06f68d1caa5ab225bfb057489cf92ecd8/assets/css/abstracts/_colors.scss#L1-L2

Here, I see that we're consuming colors both from `@wordpress/base-styles/colors` and `@automattic/color-studio/dist/color-variables`. The corresponding color definitions can be found on https://color-studio.blog/ and https://github.com/WordPress/gutenberg/blob/trunk/packages/base-styles/_colors.scss.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/b72efda06f68d1caa5ab225bfb057489cf92ecd8/assets/css/abstracts/_colors.scss#L10

This is where we're defining the color for the placeholder text. If `--global--color-primary` is available, we're using that color, and we're using `$gray-200` as a fallback. Looking at https://github.com/WordPress/gutenberg/blob/3a99ec3c2dbb9cc0b9a023597960459817475b48/packages/base-styles/_colors.scss#L13, I see that `$gray-200` has the value `#e0e0e0`. Using this fallback color on a white background, leads to the contrast ratio 1.32:1.  [WordPress.org follows the AA standard](https://make.wordpress.org/accessibility/handbook/design/use-of-color/#about-color-contrast) which requires a contrast ratio of 4.5:1 for normal text. Therefore, I suggest changing the fallback color from `$gray-200` to `$gray-700`. This then leads to a contrast ratio of 4.61:1.

Finally, I see that you added the `skip-changelog` label. I believe that this PR is beneficial for our customers. Thus, I'd mention it in the release notes. Therefore, I'd remove the `skip-changelog` label and add a changelog description instead. In general, I'd only add the `skip-changelog` label when a change is not relevant for customers, such as updating Node or Composer dependencies.

nielslange:
> Our Search Product block has different HTML structures in the editor and the live version modes. The HTML element with the class name `wc-block-product-search__field` is an `input` element in the live version, but in the editor mode, it's a `div` element. So, to target our `input` in both modes the above selector was needed.

Thanks for explaining the reason behind that change. Without your additional information, I would have wondered by you added both `&::placeholder` and `input::placeholder`. So, thanks for pointing this out. 🙌

senadir:
Setting a placeholder color assumes the input background is also set (to white). What would happen if the theme changes the input colors, would it still meet the contrast check?

I think we would need to set both the input background and the placeholder, or set none?

tarhi-saad:
> Setting a placeholder color assumes the input background is also set (to white). What would happen if the theme changes the input colors, would it still meet the contrast check?
> 
> I think we would need to set both the input background and the placeholder, or set none?

@senadir, thank you for the comment! This placeholder color is intended to be applied when the used theme didn't set any input background and placeholder colors. And since the default input background set by browsers is white we are trying to fix the contrast issue for this particular case.

We have a problem with the default placeholder color set by the browsers. For example, the latest Safari browser is setting a placeholder color of `#A9A9A9` with a white background which is failing every contrast test: [https://webaim.org/resources/contrastchecker/?fcolor=A9A9A9&bcolor=FFFFFF](https://webaim.org/resources/contrastchecker/?fcolor=A9A9A9&bcolor=FFFFFF). So, not setting a placeholder color is an issue in this case. We can see that using the Twenty Twenty-One theme.

On the other hand, if the theme changed the input background they will be expected to change the placeholder color accordingly to preserve a correct contrast.

PR #5532, opened 2022-01-07
Fix: Category block products display when `All selected categories` is selected

### LINKED ISSUE #3085: Products by Category shows different results when a single category is selected depending on the 'Display products matching' attribute
## To reproduce

1. Add a Products by Category block.
2. Select a category (_Clothing_, for example).
3. Press Done and notice several products appear.
4. Now, edit the block (pencil icon).
5. Select another category (_Accessories_, for example).
6. Set the option to 'Display products matching: All selected categories'.
7. Unselect the second category and click on Done.
8. Notice now there are fewer products than before.

## Expected behavior
If only one category is selected and, because of that, the `Display products matching` option is hidden, that option should have no effect on what products are shown.

## Screenshots
Peek 2020-08-31 10-51

### LINKED ISSUE #42676: Fix Products by Category Block showing different results when a single category is selected depending on the 'Display products matching' attribute
## To reproduce

1. Add a Products by Category block.
2. Select a category (_Clothing_, for example).
3. Press Done and notice several products appear.
4. Now, edit the block (pencil icon).
5. Select another category (_Accessories_, for example).
6. Set the option to 'Display products matching: All selected categories'.
7. Unselect the second category and click on Done.
8. Notice now there are fewer products than before.

## Expected behavior
If only one category is selected and, because of that, the `Display products matching` option is hidden, that option should have no effect on what products are shown.

## Screenshots
Peek 2020-08-31 10-51

ovidiul:
This PR will try and fix https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3085 where, when more than one category is selected and `All selected categories` option is set, we get an incorrect number of products displayed.

The proposed fix looks at the operator taxonomy query for filtering the product results:
1. The `AND` operator means that the matching products need to be in all the categories selected
2. The `IN` operator means that the matching products can be in any categories selected

I would think that 1. is not desired and that can be replaced with `IN`, and since the default operator value is `IN`, then that can be removed altogether.

Fixes woocommerce/woocommerce#42676

#### Accessibility

There are no changes at the accessibility level since the PR does not affect any stylings.

#### Other Checks

### Screenshots

Screenshot 2022-01-07 at 10 00 12

Screenshot 2022-01-07 at 10 00 55

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Install WooCommerce Blocks and WooCommerce core
2. Install WooCommerce [sample data](https://woocommerce.com/document/importing-woocommerce-sample-data/)
3. In my setup, it seems the `woocommerce/packages/woocommerce-blocks/src/BlockTypes/AbstractProductGrid.php` took precedence on load, so copy also the original file from `woocommerce-gutenberg-products-block/src/BlockTypes/AbstractProductGrid.php` to that one
4. Create a new Post
5. Select Products by Category block
6. Select 2 or more categories
7. Select the `All selected categories`
8. Check the frontend results display count, the result should only shows products assigned to those specific categories with child categories excluded

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Performance Impact

No performance impact was estimated.

### Changelog

> Fix the Products by Category block showing the incorrect number of products when `All selected categories` is selected

### DIFF
--- a/src/BlockTypes/AbstractProductGrid.php
+++ b/src/BlockTypes/AbstractProductGrid.php
@@ -240,7 +240,6 @@ abstract class AbstractProductGrid extends AbstractDynamicBlock {
 				'taxonomy'         => 'product_cat',
 				'terms'            => $categories,
 				'field'            => 'term_id',
-				'operator'         => 'all' === $this->attributes['catOperator'] ? 'AND' : 'IN',
 
 				/*
 				 * When cat_operator is AND, the children categories should be excluded,


Aljullu (commented):
Thanks for opening this PR and looking into this issue @ovidiul!

> I would think that 1. is not desired and that can be replaced with IN, and since the default operator value is IN, then that can be removed altogether.

I'm not sure to understand why you think 1 is not desired. Imagine we have two categories:

1. Clothing
2. Branded products

And three products:

* Hoodie (belongs to `Clothing`)
* Hoodie with logo (belonds to `Clothing` and `Branded products`)
* Mug with logo (belongs to `Branded products`)

Then, if the user adds the Products by Category block and selects the `Clothing` and `Branded products` categories, they should be able to choose the operator:

* `AND`: only displays Hoodie with logo
* `IN`: displays all three products

I think the issue here is that `include_children` is inferred from the `catOperator`, but it shouldn't because it's a different setting with a different meaning. I guess we would need to introduce a new UI element to allow the user to choose if categories should include children or not.

ovidiul:
> I think the issue here is that include_children is inferred from the operator, but it shouldn't because it's a different setting with a different meaning. I guess we would need to introduce a new UI element to allow the user to choose if categories should include children or not.

Adding the new UI element makes sense to me as it would give the users more flexibility, and maybe changing the default logic behavior from `AND` to `IN` might help address the initial issue.

Aljullu:
Given that this change will require some design work and that at some point we are planning to refactor those blocks to use the Query Loop block underneath, I discussed the next steps with @ovidiul and we decided to leave this PR on hold for now.

Thanks @ovidiul for working on this!

PR #5534, opened 2022-01-07
Update PR assignment automation to match new Team Kirigami name

Aljullu:
The name of the `rubik-fse-squad` GitHub team has been changed to `rubik`, so we need to update the PR review assignment automation too.


Aljullu:
I'm going to merge this PR directly. If we notice it doesn't work, we can fix it afterwards.

PR #5542, opened 2022-01-10
Featured Category block: add support for global style

gigitux:
~This PR is blocked by #5515~

This PR adds support for global style for the `Featured Category block`.

- text-color
- background-color
- font-size
- border-color
- border-radius
- border-width

### Duotone

At this stage, the `Duotone` feature is implemented with the `filter` CSS property. The problem is that on this block, the image is implemented like `background-image`, so we can't use the `filter` property.

### Padding

At this stage, with the current global style API, the padding is applied to the entire wrapper, so I guess that it doesn't make sense enable it.

#4965 

### Screenshots

image
*without global style*

image
*with global style*

### Manual Testing

Check out this branch:

3. Add the `Featured Category Block` to a post.
9. Verify that the `Featured Category block` is shown under the `Blocks` section. Personalize again the block.
10. Save your changes.
13. Change again the styles.
14. Save your changes.

### Changelog

> Add support for the global style for the Featured Category block.


assets/js/blocks/featured-category/block.js
```diff
@@ -325,7 +308,10 @@ const FeaturedCategory = ( {
 		return (
 			<ResizableBox
 				className={ classes }
-				size={ { height } }
+				size={ {
```
gigitux, line 305: I think that we should not limit the height in the editor.
Now the user can change the `font-size`. If the `font-size` is high, the block should become bigger otherwise the user can't see the entire text.

dinhtungdu (commented):
@gigitux Overall, the block style and global style are working as expected. But there are some things we can improve:

---

The background is different between editors (both Blocks and Site Editors) and the front-end. In the Blocks Editor, changing the background Opacity doesn't update the style of the preview block.

<details>
<summary>Screenshot</summary>

Screen Shot 2022-01-19 at 21 58 07

</details>

---

When the border-radius is set, there are small curves at four corners of the block. This happens because we set the border-radius for both wrapper and overlay elements.

<details>
<summary>Screenshot</summary>

Screen Shot 2022-01-19 at 21 55 56

</details>

---

The preset font sizes in the editors don't match corresponding ones on the front-end. Look like this issue occurs in other blocks, if so we can address it in a separated PR.

<details>
<summary>Screenshot</summary>

Screen Shot 2022-01-20 at 00 00 43

</details>

gigitux:
Thanks for the review!

I fixed the issue about the `opacity` and `border-radius` 👍 

For the font-size, I have always the same value (I'm using Edge):

image
*frontend*

image
*FSE Editor*

assets/js/blocks/featured-category/index.js
```diff
@@ -37,6 +39,21 @@ registerBlockType( 'woocommerce/featured-category', {
 	supports: {
 		align: [ 'wide', 'full' ],
 		html: false,
+		color: {
+			text: true,
+			background: true,
+		},
```
dinhtungdu: Background and text color are enabled by default so this can be shortened like this:
```suggestion
		color: true,
```

assets/js/blocks/featured-category/style.scss
```diff
@@ -1,10 +1,16 @@
+.wp-block-woocommerce-featured-category {
+	border-color: transparent;
+	overflow: hidden;
+	color: #fff;
+	background-color: #000;
```
dinhtungdu: Can we order CSS properties alphabetically? It's easier to maintain.

dinhtungdu (approved):
This is LGTM now! 🎉  I left some minor comments but pre-approving this PR so you can merge it after addressing those comments.

PR #5545, opened 2022-01-11
Allow LegacyTemplate block to be reinserted, only on WooCommerce block templates.

### LINKED ISSUE #5193: Explore if we can limit the Legacy Template block to only show in the block inserter on Woo block templates
For full context of the overall issue please take a look at https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5180 first.

Following another round of testing of the 6.3.1 release, we decided to [revert](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5176) our previous decision to prevent the block deletion of the Legacy Template block, as this had the unintended side effect of limiting the user's creative ability to create compelling templates.

The goal of this issue is to explore (research spike) if there's a way to limit a block to only appear in the block inserter on specific block templates. It'd be the preferred solution for only showing the Legacy Template block in the block inserter on Woo block templates.

If that's not possible, it might potentially be easier to implement hiding the block from the inserter in the post/page editor but show it in the site editor.

tjcafferkey:
#### Description

Allow the WooCommece Legacy Template block to be reinserted if it's deleted, and also only make it eligible to be inserted on the WooCommerce templates: `archive-product.html`, `single-product`, `taxonomy-product_cat`, `taxonomy-product_tag`

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5193

### Manual Testing

1. Ensure you have an FSE theme enabled such as TT1 blocks
2. Load each WooCommerce block template, remove the Legacy Template block and reinsert it.
3. Ensure this loads correctly on the frontend for each template.
4. Load a block template in the Site Editor which _is not_ from WooCommerce
5. Add a category specific template to your theme such as `taxonomy-product_cat-clothing.html` and test you can load, remove and re-add the block to this.
6. Check that you are **not** able to insert the WooCommerce Legacy Template block

### User Facing Testing
* [x] Same as above

### Changelog

> Allow users to reinsert the WooCommerce Legacy Template block in their block template if it is a WooCommerce block template.

### DIFF
--- a/assets/js/blocks/legacy-template/test/utils.ts
+++ b/assets/js/blocks/legacy-template/test/utils.ts
@@ -1,0 +1,86 @@
+/**
+ * Internal dependencies
+ */
+import { beginsWith, getMatchingTemplateData } from '../utils';
+
+export const TEMPLATES = {
+	'single-product': {
+		title: 'WooCommerce Single Product Block',
+		placeholder: 'single-product',
+	},
+	'archive-product': {
+		title: 'WooCommerce Product Grid Block',
+		placeholder: 'archive-product',
+	},
+	'taxonomy-product_cat': {
+		title: 'WooCommerce Product Taxonomy Block',
+		placeholder: 'archive-product',
+	},
+	'taxonomy-product_tag': {
+		title: 'WooCommerce Product Tag Block',
+		placeholder: 'archive-product',
+	},
+};
+
+describe( 'beginsWith', () => {
+	it( 'should return true for matching first part of a given string', () => {
+		expect(
+			beginsWith(
+				'taxonomy-product_tag',
+				'taxonomy-product_tag-clothing'
+			)
+		).toBe( true );
+		expect(
+			beginsWith(
+				'taxonomy-product_cat',
+				'taxonomy-product_cat-winter-collection'
+			)
+		).toBe( true );
+	} );
+
+	it( 'should return false if first part of a given string does not match', () => {
+		expect(
+			beginsWith(
+				'taxonomy-product_tag',
+				'taxonomies-product_tag-clothing'
+			)
+		).toBe( false );
+		expect(
+			beginsWith(
+				'taxonomy-product_cat',
+				'taxonomy-products_cat-winter-collection'
+			)
+		).toBe( false );
+		expect(
+			beginsWith(
+				'taxonomy-product_tag',
+				'taxonomy-product_cat-winter-collection'
+			)
+		).toBe( false );
+	} );
+} );
+
+describe( 'getMatchingTemplateData', () => {
+	it( 'should return template data if a correct match has been found', () => {
+		expect(
+			getMatchingTemplateData(
+				TEMPLATES,
+				'taxonomy-product_cat-winter-collection'
+			)
+		).toBe( TEMPLATES[ 'taxonomy-product_cat' ] );
+
+		expect( getMatchingTemplateData( TEMPLATES, 'single-product' ) ).toBe(
+			TEMPLATES[ 'single-product' ]
+		);
+
+		expect(
+			getMatchingTemplateData( TEMPLATES, 'taxonomy-product_tag' )
+		).toBe( TEMPLATES[ 'taxonomy-product_tag' ] );
+	} );
+
+	it( 'should return null if given template slug does not match any of the expected options', () => {
+		expect(
+			getMatchingTemplateData( TEMPLATES, 'slug-does-not-match' )
+		).toBe( null );
+	} );
+} );
--- a/assets/js/blocks/legacy-template/test/utils.ts
+++ b/assets/js/blocks/legacy-template/test/utils.ts
@@ -2,25 +2,7 @@
  * Internal dependencies
  */
 import { getMatchingTemplateData } from '../utils';
-
-export const TEMPLATES = {
-	'single-product': {
-		title: 'WooCommerce Single Product Block',
-		placeholder: 'single-product',
-	},
-	'archive-product': {
-		title: 'WooCommerce Product Grid Block',
-		placeholder: 'archive-product',
-	},
-	'taxonomy-product_cat': {
-		title: 'WooCommerce Product Taxonomy Block',
-		placeholder: 'archive-product',
-	},
-	'taxonomy-product_tag': {
-		title: 'WooCommerce Product Tag Block',
-		placeholder: 'archive-product',
-	},
-};
+import { TEMPLATES } from '../constants';
 
 describe( 'getMatchingTemplateData', () => {
 	it( 'should return template data if a correct match has been found', () => {
--- a/assets/js/blocks/legacy-template/test/utils.ts
+++ b/assets/js/blocks/legacy-template/test/utils.ts
@@ -1,7 +1,7 @@
 /**
  * Internal dependencies
  */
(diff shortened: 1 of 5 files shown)


sunyatasattva:
First of all: it all looked and worked great! As discussed on Slack, I've submitted my small change proposal.

What I did:

1. I removed the whole state management within the component. Since we are doing some template retrieving logic outside of the component anyways (to check whether it is eligible for the inserter), I felt doing it again inside the component was adding a bunch of complexity (calling the data store, and the whole `useEffect` hook, etc.) *and* it was repeating logic after all.
2. So, I thought, why don't we do everything we need in one place? Everything happens in the function you had called `isEligibleForInserter`, but instead of returning a `boolean`, it returns the actual template. In this way, we can actually populate the properties right in the inserter. The user can now see the correct title of the block within the inserter itself, and there is less confusion when he deletes the block and wants to re-add it.
3. I refactored the default properties within the `constants.ts` file, for consistency, and also fixed a very generic type there.

I **did not** change the `URLSearchParams` approach to `@wordpress/url` as I had mentioned on Slack, just to keep the changes as minimal as I could and keep your code mostly the same. `URLSearchParams` is not available on IE, if we care about that, however, so this won't work there. `@wordpress/url` comes with its own set of hassles because of some type definitions, but nothing huge, just a few more guard clauses are needed.

Thoughts?

tjcafferkey:
>I removed the whole state management within the component. Since we are doing some template retrieving logic outside of the component anyways (to check whether it is eligible for the inserter), I felt doing it again inside the component was adding a bunch of complexity (calling the data store, and the whole useEffect hook, etc.) and it was repeating logic after all.

I'd prefer to keep that in but we can get a separate opinion on that. The reason being is that is the correct approach to retrieve that data within the component so we're following best practices and it's a more robust solution, also because it's only a selector we're using performance-wise it's not a problem either. Furthermore the logic being performed outside of the component for the `inserter` value isn't the standard way to do things so if/when we have to remove that, or refactor it we are limiting the scope of that future work and potential side effects knowing we wont be affecting the logic inside the component.

sunyatasattva:
Yea, performance wasn't really my issue. It was more about code complexity and intentionality. I feel that when we have to refactor as we move on, we'd just have to remove a small piece of code instead of two pieces which do slightly different but connected things.

I feel that the use of `useEffect` would be best suited if we were expecting the `templateId` to change. But, really, we only allow one single template on each page, so that's never going to happen. So, I feel that the responsibilities are not correct: it's not the responsibility of the component to react to changes in `templateId`. Rather, it is the responsibility of the inserter to show and insert the correct block.

I feel the confusion might come from the fact that the different blocks have been refactored in the `constants.ts`, making it feel as if the `templateId` had a different role than it actually has. Really, what it is doing, is rendering a different component every time.


With that said, I'm also not married to this solution, so I'm totally willing to sack it ☺️ 

Maybe @frontdevde could chime in, since he was part of the previous discussion? Or @tjcafferkey feel free to make a call here :)

tjcafferkey:
I understand why you'd rather the data come from one source, however we've gone from using data we're accessing from an unreliable/not-recommended source in a single place, to now using it in multiple places which is my main concern.

>I feel that the use of useEffect would be best suited if we were expecting the templateId to change. 

`useEffect` is officially the recommended method of performing this type of logic even if data doesn't change. Unless I am wrong?

>It was more about code complexity

If the code was significantly complex and hard to follow I'd definitely agree. However it seemed fairly straight forward to follow and its consistent with how we get this type of data throughout our codebase, if it wasn't easy to understand perhaps we can make this easier by using better variable names and comments rather than removing this reliable method of retrieving data? After-all the whole point of a store is for tracking a single source of truth, if we can access it for this data we probably should.

tjcafferkey:
@sunyatasattva I've made some changes so that we subscribe to the store _outside_ of the component where we register the block. Once we have the value we then unsubscribe to store changes as they're no longer necessary.

I guess the only thing to note here is we are delaying the registration of the block until we have the necessary values. I've not come across this yet in our codebase but having tested this locally everything seems to work as expected.

Let me know what you think, I feel this covers both of our concerns which we mentioned above.

sunyatasattva (approved):
Oh man! That's awesome, and solves all my concerns with the previous solution. One source of truth, like you said, and we don't have two different ways to pass data around, nor convoluted workarounds.

This is the solution I wanted to try to come up with at the beginning and just didn't know about `subscribe` on `@wordpress/data`.

Amazing work and implementation! I'm really glad we iterated on it!

tjcafferkey:
@sunyatasattva are you able to review my latest commit and test again with a more specific template (see testing step 5) as we did not accommodate this scenario in our initial round of testing.

assets/js/blocks/legacy-template/utils.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Internal dependencies
+ */
+import { TemplateAttributes } from './types';
+
+export function beginsWith( needle: string, haystack: string ): boolean {
+	return haystack.substr( 0, needle.length ) === needle;
+}
```
sunyatasattva, line 8: 🤔 Why not using [`.startsWith()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith)?

assets/js/blocks/legacy-template/test/utils.ts
```diff
@@ -0,0 +1,86 @@
+/**
+ * Internal dependencies
+ */
+import { beginsWith, getMatchingTemplateData } from '../utils';
+
+export const TEMPLATES = {
```
sunyatasattva: Isn't this duplicated? Why don't we import it from `constants`? Is it because of the localization?

sunyatasattva (approved):
I'm approving with some optional comments :) Feel free to ignore them

assets/js/blocks/legacy-template/test/utils.ts
tjcafferkey: Yeah its duplicated, I wanted to isolate this as test data only. I did consider importing the actual data but if the production data changes for whatever reason it might cause a false negative if its causes the unit tests to fail.

assets/js/blocks/legacy-template/utils.ts
tjcafferkey: I started off doing, then considered compatibility with IE, and then remembered we dropped support for IE so I could have used it after all. I will replace, and if the unit tests for `getMatchingTemplateData` still pass I'll merge. Thanks!

assets/js/blocks/legacy-template/test/utils.ts
sunyatasattva: 🤔  Mm, ok. But wouldn't we want it to fail if the production data changes? Like, wouldn't that mean that our matching logic doesn't work and we need to take another look at it?

tjcafferkey: >But wouldn't we want it to fail if the production data changes?

I'm in two minds about this, because I get understand your point but in that scenario we are testing the data not the function. Let me try and update and see how this goes.

tjcafferkey: Updated 👌🏻

sunyatasattva: Now I see what you mean and I understand your conundrum. Thanks for elaborating.

PR #5546, opened 2022-01-11
When processing payments, use the woocommerce_hold_stock_minutes duration instead of 10 mins

### LINKED ISSUE #4658: Increase duration of reserve stock holds once an order changes from draft to pending_payment
- Block Checkout reserves stock for 10 mins.
- Core checkout reserves stock for pending payment orders for 60 mins.

But what reserves stock if a draft order becomes pending payment?

WooCommerce core reserves stock via the `woocommerce_checkout_order_created` hook, but this is not run as part of the Block Checkout flow. This means draft orders which change to pending_payment have 10 minutes to resolve, or the stock is freed up. This is not necessarily a bug but is different to the usual behaviour of 60 mins. 

One possible solution would be to use the woocommerce_order_status_draft hook instead which would run regardless of where the order was created. Another would be to target the transition only. We can add this code on the blocks side and move it to core when the Checkout Block is released.

mikejolley:
When we create a `draft` order during checkout, we put a hold on stock for all items in the cart to prevent things going out of stock while the customer enters their details. This is set at 10 minutes.

WC core (which I remind you does not use draft orders), sets a timeout of 60 (user configurable) when the order is created with a `pending-payment` status. [This happens here](https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/wc-stock-functions.php#L357).

There is no mechanism in Blocks/Store API which currently increases the timeout from 10 to 60 when the draft becomes pending.

This PR works around this, but rather than messing with order status, it simply switches between 10 and 60 depending on if the customer is entering checkout (draft is created), or attempting to pay (draft is updated to pending status).

Fixes #4658

### Testing

For developers:

- Edit an item and ensure it has inventory enabled
- Add the item to the cart and visit checkout page
- View DB. In local WP that would be here: http://localhost:10005/?username=root&db=local&table=wp_wc_reserved_stock
- Stock is held 10 mins (see expires column)
- Comment out `do_action_ref_array( 'woocommerce_rest_checkout_process_payment_with_context', [ $context, &$payment_result ] );` in blocks repo so payment is disabled.
- Place order again
- View database table. Stock expires should be 60 mins rather than 10.

For end-users:

- Edit an item and ensure it has inventory enabled
- Successfully place an order

### Changelog

> Hold stock for 60mins if the order is pending payment.

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -119,7 +119,7 @@ class Checkout extends AbstractCartRoute {
 	 * @return \WP_REST_Response
 	 */
 	protected function get_route_response( \WP_REST_Request $request ) {
-		$this->create_or_update_draft_order();
+		$this->create_or_update_draft_order( $request );
 
 		return $this->prepare_item_for_response(
 			(object) [
@@ -161,7 +161,7 @@ class Checkout extends AbstractCartRoute {
 		 * uses the up to date customer address.
 		 */
 		$this->update_customer_from_request( $request );
-		$this->create_or_update_draft_order();
+		$this->create_or_update_draft_order( $request );
 		$this->update_order_from_request( $request );
 
 		/**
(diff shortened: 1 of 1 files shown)


nielslange (approved):
LGTM! 🚢

PR #5553, opened 2022-01-12
Prefetch cart and checkout block assets when cart has contents

### LINKED ISSUE #2207: Implement resource hinting between Cart and Checkout blocks
A feature in modern browser is the ability to hint the resources you might need next, this includes the ability to prefetch a script file, or pre-render a page, connect a handshake, load a dns, all before the user need it, collectively those are called browser resource hint, and I think there is a chance to implement some of those in Cart and Checkout blocks, like preloading Checkout when you're in Cart or when you intend to click "continue to checkout", the browser controls if it should pre-render or not, depending on the network activity for the current panel.

More read, from [Addy Osmani](https://medium.com/reloading/preload-prefetch-and-priorities-in-chrome-776165961bbf)
and a [full guide from 3perf](https://3perf.com/blog/link-rels/)

mikejolley:
This PR implements resource hinting for cart and checkout scripts when we can predict the customer will be visiting those pages during the current session. It essentially pre-caches block assets in the background when the page is idle so that when the customer lands on the cart and checkout pages, the block scripts can be loaded from cache.

Fixes #2207

cc @senadir 

See:
- https://3perf.com/blog/link-rels/
- https://medium.com/reloading/preload-prefetch-and-priorities-in-chrome-776165961bbf

The new rules include:

1. Cart has contents, not on the cart page = Load cart block assets
2. Cart has contents, not on checkout page = Load checkout assets
3. Cart has contents, viewing the cart page = Prerender the checkout page

Case 3 uses the `prerender` attribute which is not supported in all browsers, but is by Chrome. It does not actually render the page or run JS, but it does pre-load all assets and resources on the page. See https://developers.google.com/web/updates/2018/07/nostate-prefetch

### Manual Testing

1. Clear website cache.
2. Go to the homepage with nothing in your cart.
3. View source. Look for `rel='prefetch'` rules. Cart and checkout will not be there.
4. Go to the shop page and add something to the cart.
5. Go back to the homepage.
6. View source. Look for `rel='prefetch'` rules:

Screenshot 2022-01-12 at 13 17 57

If you open network tools at this point and then visit the cart page, you should see the majority of scripts are loaded from cache.

### User Facing Testing

Users cannot see these changes in an obvious way, so smoke testing should be performed only of the checkout flow.

### Performance Impact

For first time visitors to the cart and checkout pages, this should increase the page loading times due to assets being retrieved from cache rather than the server.

### Changelog

> Add resource hinting for cart and checkout blocks to improve first time performance.

### DIFF
--- a/src/AssetsController.php
+++ b/src/AssetsController.php
@@ -35,6 +35,7 @@ final class AssetsController {
 	 */
 	protected function init() {
 		add_action( 'init', array( $this, 'register_assets' ) );
+		add_filter( 'wp_resource_hints', array( $this, 'add_resource_hints' ), 10, 2 );
 		add_action( 'body_class', array( $this, 'add_theme_body_class' ), 1 );
 		add_action( 'admin_body_class', array( $this, 'add_theme_body_class' ), 1 );
 		add_action( 'admin_enqueue_scripts', array( $this, 'update_block_style_dependencies' ), 20 );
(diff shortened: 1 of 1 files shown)


alexflorisca (changes requested):
Nice., hopefully this will speed up the checkout process a fair bit! Would be interested to see by how much and if we can measure it somehow. I'll have a think.

The code looks good, again with the caveat that I'm not great at PHP. Wonder how long I can get away with that for haha

When I tried running it in the browser, it looks like the cart isn't loading properly. I can see the fetch to the cart store (`wc/store/v1/cart?_locale=user`) fails with a 404. That's even when the resources are loaded normally, not from prefetch cache, so I'm not sure why...

alexflorisca (approved):
Nice, thanks for fixing the 404, looks good :)

PR #5554, opened 2022-01-12
Move DropdownSelect to `@woocommerce/blocks-checkout` package

opr:
This PR will move the `DropdownSelect` component to the `@woocommerce/blocks-checkout` package. This will make it easier for third parties to re-use some of the UI we have built.

This is especially relevant for 75-gh-woocommerce/woocommerce-cart-add-ons

Helps 75-gh-woocommerce/woocommerce-cart-add-ons

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

1. Add a `Filter by Attribute` block and a `All Products` block and ensure they both load in the editor.
2. Choose an attribute and then in the Block Settings, set the display style to `Dropdpown`

3. View the block on the front end and ensure the dropdown works correctly and filters the products shown in the `All Products` block.

### User Facing Testing
* [x] Same as above

### Changelog


senadir:
As stated in slack, I don't think we should be exporting this component, and increasing our API surface, for this one use case. If the component is really needed, they can just copy it.
If we found a compelling use case for exporting it, then it should happen on the `@woocommerce/components` package level, not on `checkout/components` package, as it's not strictly limited to checkout (and to my knowledge, not even used there).

But again, I think we should keep it as is, we had plan to remove it in favor of something from Gutenberg if I remenber correctly.

opr:
Yup agreed, this isn't necessary I don't think. Closing

ralucaStan:
There is actually a ticket to replace the DropdownSelector with Gutenberg's FormTokenField https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/3762

PR #5555, opened 2022-01-12
Featured Product: Enable global style

gigitux:
This PR is blocked by #5542

This PR adds support for global style for the `Featured Product block`.

- text-color
- background-color
- font-size
- border-color
- border-radius
- border-width

### Duotone

At this stage, the `Duotone` feature is implemented with the `filter` CSS property. The problem is that on this block, the image is implemented like `background-image`, so we can't use the `filter` property.

### Padding

At this stage, with the current global style API, the padding is applied to the entire wrapper, so I guess that it doesn't make sense enable it.

#4965 

### Screenshots

image
*without global style*

image
*with global style*

### Manual Testing

Check out this branch:

1.  On `WordPress 5.9`, install and enable the `Gutenberg` plugin.
3. Add the `Featured Product block` to a post.
9. Verify that the `Featured Product block` is shown under the `Blocks` section. Personalize again the block.
10. Save your changes.
13. Change again the styles.
14. Save your changes.

### Changelog

> Add support for the global style for the Featured Product block.


assets/js/blocks/featured-product/style.scss
```diff
@@ -1,10 +1,16 @@
...
+}
+
 .wc-block-featured-product {
 	position: relative;
-	background-color: $gray-900;
 	background-size: cover;
 	background-position: center center;
 	width: 100%;
-	margin: 0 0 1.5em 0;
+	margin: 0 0 0 0;
```
Aljullu: Should we use the shorthand here?
```suggestion
	margin: 0;
```

assets/js/blocks/featured-product/block.js
```diff
@@ -301,7 +282,10 @@ const FeaturedProduct = ( {
 		return (
 			<ResizableBox
 				className={ classes }
-				size={ { height } }
+				size={ {
+					height: '',
+					width: '',
+				} }
```
Aljullu: How does that work? I seem to be able to change the height of the block, but not the width. Is that the expected behavior?

assets/js/blocks/featured-category/style.scss
```diff
@@ -1,10 +1,16 @@
...
+}
+
 .wc-block-featured-category {
 	position: relative;
-	background-color: $gray-900;
 	background-size: cover;
 	background-position: center center;
 	width: 100%;
-	margin: 0 0 1.5em 0;
+	margin: 0 0 0 0;
```
Aljullu: Here too:
```suggestion
	margin: 0;
```

assets/js/blocks/featured-product/style.scss
```diff
@@ -49,10 +55,11 @@
 	.wc-block-featured-product__variation,
 	.wc-block-featured-product__description,
 	.wc-block-featured-product__price {
-		color: $white;
 		line-height: 1.25;
 		margin-bottom: 0;
 		text-align: center;
+		color: inherit;
+		font-size: inherit;
```
Aljullu: Personally, I think the `.wc-block-featured-product__title` shouldn't have `font-size: inherit;`. That changes the default design:

| `trunk` | This branch |
| --- | --- |
| imatge | imatge |

What do you think?

Aljullu (commented):
This is working great and code looks good, I just left some minor comments below. Heads-up that I skipped reviewing all Featured Categories files, I guess after rebasing the PR they will be removed from the diff.

assets/js/blocks/featured-product/block.js
gigitux: I think that we should not limit the height of the editor.
Now the user can change the font size. If the font size is high, the block should become bigger otherwise the user can't see the entire text.

In accordance with TS type, we need to specify either `height` and `width`

assets/js/blocks/featured-product/style.scss
gigitux: You are right! 
This is a bug that affects `Featured Category` too :(  (already merged in the trunk), but this bug is replicable only when the site has a block theme enabled.

The Global Styles API doesn't support different font-size options, so we can:
- disable the font size option.
- font size option will impact only the description (it doesn't make sense IMHO).

gigitux: In the end, I followed the same approach that we adopted in #5682: the user can't customize font-size

assets/js/blocks/featured-product/block.js
Aljullu: Now that the font size can't be modified anymore, should we undo this change?

Aljullu (approved):
I left [one last comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5555#discussion_r796428459), but besides that, everything looks good to me, so pre-approving.

gigitux:
Thanks for the review!
I restored the height for the Featured Category block too (d565abf) since we disabled the font size on the latter.

PR #5557, opened 2022-01-13
Attribute Filter block: Enable global style

gigitux:
This PR is blocked by #5542

This PR adds support for global style for the `Attribute Filter block`.

- text-color
- border (color and radius) on the input fields

## Note
Unfortunately at the current stage, global style API doesn't support the customization of the inner element instead of block wrapper, so this PR contains a hacky CSS for accomplishing the goal 

https://github.com/WordPress/gutenberg/issues/33255
https://github.com/WordPress/gutenberg/issues/32417

#4965 

#### Mode: Display List

image
*without global style*

image
*with global style*

#### Mode: Dropdown List

image
*without global style*

image
*with global style*

### Manual Testing

Check out this branch:

1.  Upgrade to `WordPress 5.9`.
3. Add the `Attribute Filter block` to a post.
9. Verify that the `Attribute Filter block` is shown under the `Blocks` section. Personalize again the block.
10. Save your changes.
13. Change again the styles.
14. Save your changes.

### Changelog

> Add support for the global style for the Attribute Filter block.


gigitux:
Do you think it makes sense to add some comments about this hacky approach in the CSS? (maybe where I use `!important`)

assets/js/base/components/dropdown-selector/index.js
```diff
@@ -23,6 +23,7 @@ import './style.scss';
  * @param {Object} props Incoming props for the component.
  * @param {string} props.attributeLabel Label for the attributes.
  * @param {string} props.className CSS class used.
+ * @param {Object} props.style CSS style object used.
```
sunyatasattva: What about using `CSSProperties`?

assets/js/blocks/featured-category/block.js
```diff
@@ -148,21 +142,14 @@ const FeaturedCategory = ( {
...
 				</PanelBody>
-				<PanelColorSettings
-					title={ __( 'Overlay', 'woo-gutenberg-products-block' ) }
-					colorSettings={ [
-						{
-							value: overlayColor.color,
-							onChange: setOverlayColor,
-							label: __(
-								'Overlay Color',
+				{ !! url && (
```
sunyatasattva: I understand you want to cast this as a boolean, but I have to say I've always found the `!!` confusing and I'm not sure there is any clearer advantage to just having `url &&`. What are your thoughts?

assets/js/blocks/featured-category/edit.tsx
```diff
@@ -0,0 +1,23 @@
...
+ * External dependencies
+ */
+import { useBlockProps } from '@wordpress/block-editor';
+
+/**
+ * Internal dependencies
+ */
+import Block from './block';
+
+export const Edit = ( props: unknown ): JSX.Element => {
```
sunyatasattva: I haven't worked with blocks much, so forgive me for asking: isn't there a way to know this type at all?

assets/js/blocks/featured-category/index.js
```diff
@@ -146,14 +163,12 @@ registerBlockType( 'woocommerce/featured-category', {
...
-	edit( props ) {
-		return <Block { ...props } />;
-	},
+	edit: Edit,
 
 	/**
 	 * Block content is rendered in PHP, not via save function.
 	 */
-	save() {
+	save: () => {
```
sunyatasattva: Why was this changed? Was this automatically done by Prettier? I find 13 instances of `save() {` but only this one of `save: () =>`, so perhaps the former is more in line with the repo conventions?

sunyatasattva (changes requested):
Yes please do leave some comments about the hacky solution. Every time I see an `!important` it'd be good to have the context on why it was added. 🙏

assets/js/blocks/featured-category/block.js
gigitux: I just moved this code, I didn't edit it. I agree with you that it is not very clear this cast! I will change it!

assets/js/blocks/featured-category/edit.tsx
gigitux: > I haven't worked with blocks much, so forgive me for asking: isn't there a way to know this type at all?

Yes, we can try to type it, but without the right typeguards, there is the risk that the code will be less safe than now.

sunyatasattva (approved):
Thanks for adding the comments!

PR #5559, opened 2022-01-13
Price Filter block: enable global style 

gigitux:
This PR adds support for global style for the `Price Filter block`.

- text-color
- border (color and radius) on the input fields

## Note
Unfortunately at the current stage, global style API doesn't support the customization of the inner element instead of block wrapper, so this PR contains a hacky CSS for accomplishing the goal 

https://github.com/WordPress/gutenberg/issues/33255
https://github.com/WordPress/gutenberg/issues/32417

#4965 

### Screenshots

image
*without global style*

image
*with global style*

### Manual Testing

Check out this branch:

3. Add the `Price Filter block` to a post.
9. Verify that the `Price Filter block` is shown under the `Blocks` section. Personalize again the block.
10. Save your changes.
13. Change again the styles.
14. Save your changes.

### Changelog

> Add support for the global style for the Price Filter block.

### DIFF
--- a/assets/js/blocks/price-filter/block.js
+++ b/assets/js/blocks/price-filter/block.js
@@ -17,6 +17,7 @@ import { getCurrencyFromPriceResponse } from '@woocommerce/price-format';
  * Internal dependencies
  */
 import usePriceConstraints from './use-price-constraints.js';
+import './style.scss';
 
 /**
  * Component displaying a price filter.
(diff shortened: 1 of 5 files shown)


gigitux:
Do you think it makes sense to add some comments about this hacky approach in the CSS? (maybe where I use `!important`)

dinhtungdu (approved):
> Do you think it makes sense to add some comments about this hacky approach in the CSS?

Yes! I don't think that CSS is hacky. Having the child components inherit CSS from the parent is normal usage IMO.

gigitux:
Thanks for the review!

> Yes! I don't think that CSS is hacky. Having the child components inherit CSS from the parent is normal usage IMO.
Yeah, I agree with you! My concerns are more about `!important` usage.

Anyway, I will merge the PR!

PR #5560, opened 2022-01-13
Prevent Redirect URL leaking between payment gateways

### LINKED ISSUE #4224: Investigate why redirectURL could be non-truthy and wheth...
Investigate why redirectURL could be non-truthy and whether this would cause a bug if multiple gateways were used for payment e.g. 1st set the redirect URL but failed, and then the 2nd did not provide a redirect URL and succeeded.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/42b89b49820d1d92d98024face0d723c782f6030/assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts#L69-L80
---
###### :rocket: This issue was generated by the [automations bot](https://github.com/woocommerce/automations) based on a `todo` comment in 42b89b49820d1d92d98024face0d723c782f6030 when #4200 was merged. cc @mikejolley

mikejolley:
This PR prevents (due to implementation error) redirect URLs being shared between different payment gateways.

Fixes #4224

### Testing

You'll need to modify 2 gateways to test this. 

*plugins/woocommerce/includes/gateways/cheque/class-wc-gateway-cheque.php*

Replace the return statement to be:

```
return [
			'result'   => 'success',
		];
```

*plugins/woocommerce/includes/gateways/bacs/class-wc-gateway-bacs.php*

Change the `process_payment` method to contain only:

```
return array(
			'result'   => 'fail',
			'redirect' => $this->get_return_url( $order ),
		);
```

Now if you add something to cart, and go to checkout you can test this case.

1. Try first paying with BACS. Payment will fail. Notice this returned a redirect.
2. Now switch payment method and pay with Cheque.
3. Before this patch, Cheque would redirect to the URL provided by BACS. After this patch, checkout will not redirect.

### User Facing Testing

Smoke test payments on the checkout page (using BACS should suffice).

### Changelog

> Prevent redirect URLs being incorrectly shared between payment gateways.

### DIFF
--- a/assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
+++ b/assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
@@ -148,12 +148,9 @@ export const CheckoutStateProvider = ( {
 				const paymentResult = getPaymentResultFromCheckoutResponse(
 					response
 				);
-
-				if ( paymentResult.redirectUrl ) {
-					dispatch(
-						actions.setRedirectUrl( paymentResult.redirectUrl )
-					);
-				}
+				dispatch(
+					actions.setRedirectUrl( paymentResult?.redirectUrl || '' )
+				);
 				dispatch( actions.setProcessingResponse( paymentResult ) );
 				dispatch( actions.setAfterProcessing() );
 			},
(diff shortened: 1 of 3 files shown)


nielslange (approved):
LGTM! 🚢

PR #5565, opened 2022-01-14
Sale Product Block: Enable global style

gigitux:
~This PR is blocked by #5542~

This PR adds support for global style for the `Sale Product block`. This code registers the block on the PHP side. This is necessary for using the `__experimentalSelector` from Global Styles API.

- border-radius
- font-size
- text color
- background-color (including gradients)
- padding
- color 
- radius

#4965 

### Screenshots

image
*without global style*

image

*with global style*

### Manual Testing

Check out this branch:

1. Install and enable the `Twenty Twenty-Two` theme.
2. Add the `All Product Block` to a post.
3. Go to Dashboard and select Appearance > Editor (beta).
4. On the Editor page click on the `Styles` icon on the right-top corner.
5. Verify that the `Sale Product block` is shown under the `Blocks` section. Click on the label. Personalize the block.
6. Save your changes.
7. Go on the page created earlier and check if all styles are applied correctly.

### Changelog

> Add support for the global style for the Sale Product block.


src/BlockTypesController.php
```diff
@@ -178,6 +178,7 @@ protected function get_block_types() {
 			'ProductTitle',
 			'ProductSummary',
 			'ProductStockIndicator',
+			'ProductSaleBadge',
```
dinhtungdu, line 182: I think a comment here explaining why we register an atomic block here will be helpful.

dinhtungdu (approved):
This is look good to me. Even the block registration looks a little bit confusing, I think we can always address that issue later when we have another atomic block that needs `__experimentalSelector` support. `__experimentalSelector`, by its name, is experimental, so it can be changed in the future Gutenberg release, so it isn't worth addressing it now.

src/BlockTypesController.php
gigitux: Good point! I added a `@todo comment`. In this way, in the end, I will re-organize the order of the elements in the array (maybe we can move down the atomic blocks and add a comment that explains why)

PR #5443, opened 2021-12-22
Mini Cart Contents Block: Add text and link color support 

gigitux:
This PR adds text and link color support for the Mini Cart Contents Block.

Fixes #5413

### Screenshots

image
*editor side*

image
*frontend side*

### Manual Testing

How to test the changes in this Pull Request:

Check out this branch

1. Open the FSE editor.
2. Pick the Mini Cart Content templates.
3. Be sure to focus on the `Mini Cart Contents`.
4. On the right sidebar, set different colors.
5. Check if the new colors are visible on the editor and frontend side.

### Question
I'm not sure that this is PR is complete.

I have some questions about both statuses.

#### Filled Cart

image

- Should the color of the text into the input be editable?
- Should `Remove item` text have the color of the text or the color of the link?

#### Empty Cart

image

- Should the text into the button have the color of the text or the color of the link?

### Changelog

> Add text and link color support for Mini Cart Contents Block

PR #5566, opened 2022-01-14
Mini Cart Contents: Support link and text colors

### LINKED ISSUE #5413: Add text color and link support to Mini Cart Contents block
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5386 added background support to the Mini Cart Contents block. To make the block more usable, we should allow the admin to control the text and link color as well.

The admin only needs to set the colors for the Mini Cart Contents block. The inner blocks should inherit those colors and adjust to fit their need.

dinhtungdu:
Fixes #5413. Replaces #5443 

This PR adds text and link color support to the Mini Cart Contents block. Users only need to change three colors of the Mini Cart Contents block then inner blocks get styled correctly based on those colors.

#### Limitations
- The global style doesn't work. Because we only have one Mini Cart template part, its block style is basically the global style. AFAIK, we can't opt-out a block from global style, so Mini Cart Contents still appears in the Global Style blocks list.

### Screenshots

<details>
<summary>Details</summary>

Screen Shot 2022-01-14 at 21 08 10
Screen Shot 2022-01-14 at 20 27 22
Screen Shot 2022-01-14 at 20 27 31

</details>

### Manual Testing

1. Edit the Mini Cart template part.
2. Select Mini Cart Contents block.
3. Change background, text, and link colors.
4. See the color of inner blocks change accordingly.
5. Save the template part. See the colors applied on the front end.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/js/atomic/blocks/product-elements/title/editor.scss
+++ b/assets/js/atomic/blocks/product-elements/title/editor.scss
@@ -1,3 +1,3 @@
-.editor-styles-wrapper a.wc-block-components-product-name {
+.editor-styles-wrapper .wc-block-components-product-title a.wc-block-components-product-name {
 	color: inherit;
 }
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss
@@ -36,4 +36,9 @@
 	table.wc-block-cart-items {
 		color: inherit;
 	}
+
+	.block-editor-button-block-appender {
+		box-shadow: inset 0 0 0 1px;
+		color: inherit;
+	}
 }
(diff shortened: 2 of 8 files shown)


gigitux (approved):
Thanks for helping me with this issue!

Is it correct that we change the style of the button that allows the user to add new components? 
image


> AFAIK, we can't opt-out a block from global style, so Mini Cart Contents still appears in the Global Style blocks list.

There isn't an official method, but we could opt-out a block [from here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/d08e9cef3fdf4f522133d89fed5040bf51c78122/src/BlockTypesController.php#L194). Maybe it makes sense to create a dedicated issue to improve that logic because it could be useful in the future. What do you think?

Besides these two questions, LGTM 👍

dinhtungdu:
> Is it correct that we change the style of the button that allows the user to add new components?

@gigitux Yes. Because the background color can be the same as the color of the appender, I apply the text color to it.

> There isn't an official method, but we could opt out a block from here. Maybe it makes sense to create a dedicated issue to improve that logic because it could be useful in the future. What do you think?

I think a separate issue is better.

PR #5574, opened 2022-01-17
Remove deprecated data controls in favour of those from the `@wordpress/data` package

### LINKED ISSUE #5339: Handle select from @wordpress/data-controls deprecation warning
The `select`, `syncSelect` and `dispatch` exports from @wordpress/data-controls [have been deprecated](https://github.com/WordPress/gutenberg/pull/26509) and we see this type of warnings:

```
["`select` control in `@wordpress/data-controls` is deprecated. Please use built-in `resolveSelect` control in `@wordpress/data` instead."]
```

We should fix this by switching to the corresponding controls. (For example: built-in `resolveSelect` control in `@wordpress/data`)

Things to take into consideration: see p1634809299040100-slack-C8X6Q7XQU
- What version of WordPress is the replacement available?
- Is the deprecated implementation in Woo Blocks in code that is exposed to users via Woo Core?
- What version of WordPress does Woo Core still require as minimum support and does that fall in the appropriate range?

mikejolley:
Removes usage of the deprecated functions `select` and `dispatch` imported from `@wordpress/data-controls` ([deprecated here](https://github.com/WordPress/gutenberg/pull/26509)) and replaces them with `controls.select` and `controls.dispatch` from the `@wordpress/data` package.  

Fixes #5339

Questions from the issue:

> What version of WordPress is the replacement available?

The switch to `@wordpress/data` happened in Gutenberg 9.2, which was part of [WordPress 5.6](https://developer.wordpress.org/block-editor/contributors/versions-in-wordpress/), but controls existed before then. 

> Is the deprecated implementation in Woo Blocks in code that is exposed to users via Woo Core?

Yes, it's part of schema and collections data-stores, used by All Products block.

> What version of WordPress does Woo Core still require as minimum support and does that fall in the appropriate range?

The `controls` export existed in WordPress 5.6 ([see here](https://github.com/WordPress/WordPress/blob/5.6/wp-includes/js/dist/data.js)). This is the min supported version of Woo core.

Blocks min support is 5.8. Both are satisfied.

### Testing

1. Smoke test changing quantities of items in the cart
2. Smoke test All Products block and add to cart buttons within that block
3. Smoke test All Products block with Filter Blocks

JS tests must also pass. I removed the deprecation assertions from those.

### Changelog

> Remove deprecated data controls in favour of those from the `@wordpress/data` package

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/test/block.js
+++ b/assets/js/blocks/cart-checkout/cart/test/block.js
@@ -83,7 +83,9 @@ describe( 'Testing cart', () => {
 	} );
 
 	it( 'renders cart if there are items in the cart', async () => {
-		render( <CartBlock /> );
+		act( () => {
+			render( <CartBlock /> );
+		} );
 		await waitFor( () => expect( fetchMock ).toHaveBeenCalled() );
 		expect(
 			screen.getByText( /Proceed to Checkout/i )
@@ -96,7 +98,9 @@ describe( 'Testing cart', () => {
 		allSettings.displayCartPricesIncludingTax = false;
 		// The criteria for showing the Taxes section is:
 		// Display prices during basket and checkout: 'Excluding tax'.
-		render( <CartBlock /> );
+		act( () => {
+			render( <CartBlock /> );
+		} );
 
 		await waitFor( () => expect( fetchMock ).toHaveBeenCalled() );
 		expect( screen.getByText( /Tax/i ) ).toBeInTheDocument();
@@ -108,7 +112,9 @@ describe( 'Testing cart', () => {
 		// The criteria for showing the lines in the Taxes section is:
 		// Display prices during basket and checkout: 'Excluding tax'.
 		// Display tax totals: 'Itemized';
-		render( <CartBlock /> );
+		act( () => {
+			render( <CartBlock /> );
+		} );
 		await waitFor( () => expect( fetchMock ).toHaveBeenCalled() );
 		expect( screen.getByText( /Sales tax/i ) ).toBeInTheDocument();
 	} );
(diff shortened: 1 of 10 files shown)


assets/js/blocks/cart-checkout/cart/test/block.js
```diff
@@ -133,13 +131,15 @@ describe( 'Testing cart', () => {
...
+			fetchMock.mockResponse( ( req ) => {
+				if ( req.url.match( /wc\/store\/cart/ ) ) {
+					return Promise.resolve(
+						JSON.stringify( defaultCartState.cartData )
+					);
+				}
+				return Promise.resolve( '' );
+			} );
 		} );
 		render( <CartBlock /> );
```
alexflorisca: I think the render method should go inside the `act()` function from the [react docs](https://reactjs.org/docs/test-utils.html#act)

alexflorisca (changes requested):
Hey @mikejolley, ran into an issue when testing this

The all products block seems to think there are no products (even though there definitely are)

Screenshot 2022-02-03 at 16 41 56

Filter blocks alongside it also seem to be broken:

Screenshot 2022-02-03 at 16 42 52

I don't know enough about how `@wordpress/data` works to suggest why this might be but I'm happy to help you figure it out!

The code looks good though and the Cart quantity selector seems to work fine ok and the tests pass 👍

mikejolley:
I see the products issue too. I can't figure out why though 🤔  I don't see anything obvious in the diff which would break. I do see `const route = yield controls.select` is returning blank for some reason.

mikejolley:
Found the issue. `select` does not wait for the resolution. `resolveSelect` does. See https://github.com/WordPress/gutenberg/blob/trunk/packages/data/src/controls.js#L40-L84

alexflorisca (approved):
Ah good spot. I wonder why there's 2 methods, from what I know `select` ran the resolver automatically if there was one specified and waited for it. Anyway good to go 👍

PR #5585, opened 2022-01-17
Add ensureCleanAttributes back and run only if E2E tests are running locally

### LINKED ISSUE #5410: E2E tests: Add a manual way of tearing down the attributes
See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4562#discussion_r688413063 for details

opr:
This PR will add the `ensureCleanAttributes` function back, but this time it will only run locally and not in CI.

It was removed from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/4562#discussion_r688413063 because of concerns about running this code unnecessarily.

While it is still a useful function for testing locally, we can keep it but just skip running it if the tests are running in CI. To determine whether they are we can check an environment variable that is set by GitHub actions. I chose `GITHUB_RUN_ID`.

Fixes #5410

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

1. Force a commit that includes a console log in this conditional
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7301aac85852e91ff3296480096167d01c5f89a9/tests/e2e/config/setup.js#L37-L39
2. Check the output of the E2E tests action, ensure you can't see your console logged text.

### User Facing Testing
* [ ] Same as above, or
* [ ] See steps below.

This does not need user-facing testing to be carried out.

### Performance Impact
When running locally, there will be a (very) minor performance decrease while the attributes are deleted, when running in CI there should be no change.

### Changelog

### DIFF
--- a/tests/e2e/config/setup.js
+++ b/tests/e2e/config/setup.js
@@ -35,6 +35,7 @@ module.exports = async ( globalConfig ) => {
 		// We only want to run ensureCleanAttributes when running locally, and not in CI.
 		// This env variable should not exist on someone's local machine
 		if ( ! process.env.GITHUB_RUN_ID ) {
+			console.log( 'cleaning attributes' );
 			await ensureCleanAttributes();
 		}
 		/**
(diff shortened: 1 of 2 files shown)


mikejolley (approved):
I'm going to approve this based on the content and the fact I trust you tested (I can see the commits with the console log), but I wasn't able to find the log for that test.

Issue #5103, opened 2021-11-09
Unify/merge attributes in Cart/Checkout inner blocks.


senadir:
Several of Cart/Checkout inner blocks define attributes both in `block.json` and an `attributes.js` file, this causes two source of truths and two places to update each time.

We should either:
- Remove `attributes.js` file in favor of block.json. For blocks using `formStepAttributes`, I think we can remove that abstraction in favor of a single source of truth, this:
```js
export default {
	...formStepAttributes( {
		defaultTitle: __( 'Shipping address', 'woo-gutenberg-products-block' ),
		defaultDescription: __(
			'Enter the address where you want your order delivered.',
			'woo-gutenberg-products-block'
		),
	} ),
};
```

Can be replaced with this:
```json
"attributes": {
	"title": {
		"type": "string",
		"default": "Shipping address"
	},
	"description": {
		"type": "string",
		"default": "Enter the address where you want your order delivered."
	},
	"showStepNumber": {
		"type": "boolean",
		"default": true
	}
},
```

- For blocks in which you need a dynamic default attribute (using getSettings), `attributes.js` should only contain those dynamic attributes, and in `index.js`, you merge them:
```js
registerFeaturePluginBlockType( metadata, {
	icon: {
		src: <Icon srcElement={ address } />,
		foreground: '#7f54b3',
	},
	attributes: {
		...blockAttributes,
		...metadata.attributes,
	},
	edit: Edit,
	save: Save,
} );
```

PR #5594, opened 2022-01-19
Move Checkout block attributes into metadata (`block.json`)

opr:
This PR will move the attributes for the Checkout block out of `attributes.ts` and into `block.json` this is done to unify the block attributes which is part of #5103.

I'm making this PR now before proceeding to make sure that the approach is correct before addressing all the other blocks.

### Changes
- Added a `get_block_metadata_path` function to the `Assets\Api` class, this allows us to get the path to a `block.json` file (which will be specified in the block's class).
- Removes `attributes` and `supports` from the `$block_settings` array in `AbstractBlock`, this is required because when registering a block with metadata, the values in the `$block_settings` override it. We want the metadata to be the source of truth to avoid duplicating config in PHP and JS. We need to keep `$block_settings` around though to pass `render_callback`, `editor_script`, `editor_style`, and `style` to `register_block_type`. These
- Moved all static attributes from `attributes.ts` into `block.json` (I left the ones that were derived from `getSetting`).
- Moved all static data from `index.ts` into `block.json`

#### Automated Tests
  * [x] Unit tests
  * [x] E2E tests

### Manual Testing
1. Smoke test the checkout block, try placing an order, using different payment methods, make sure validation etc. is all working as intended

### User Facing Testing
* [x] Same as above, or


assets/js/blocks/cart-checkout/checkout/attributes.ts
```diff
@@ -5,56 +5,10 @@ import { getSetting } from '@woocommerce/settings';
...
-	},
-	showPhoneField: {
-		type: 'boolean',
-		default: true,
-	},
-	requirePhoneField: {
-		type: 'boolean',
-		default: false,
-	},
-	// Deprecated - here for v1 migration support
```
mikejolley, line 41: Should we leave these out of metadata and possible put them with the block registration function so it's easier to remove them in future?

assets/js/blocks/cart-checkout/checkout/block.json
```diff
@@ -0,0 +1,62 @@
...
+  "description": "Display a checkout form so your customers can submit orders.",
+  "category": "woocommerce",
+  "keywords": [ "WooCommerce" ],
+  "supports": {
+	"align": [ "wide" ],
+	"html": false,
+	"multiple": false
+  },
+  "attributes": {
+	"isPreview": {
```
mikejolley: Not sure if this is just the preview, but it seems this JSON file lacks indenting.

src/BlockTypes/AbstractBlock.php
```diff
@@ -181,14 +188,26 @@ protected function register_block_type() {
...
-			'attributes'      => $this->get_block_type_attributes(),
-			'supports'        => $this->get_block_type_supports(),
 		];
 
 		if ( isset( $this->api_version ) && '2' === $this->api_version ) {
 			$block_settings['api_version'] = 2;
 		}
 
+		// Prefer to register with metadata if the path is set in the block's class.
+		if ( isset( $this->metadata_path ) && '' !== $this->metadata_path ) {
```
mikejolley: You can combine these 2 conditionals into a single `! is_empty()` call.

src/BlockTypes/AbstractBlock.php
```diff
@@ -181,14 +188,26 @@ protected function register_block_type() {
...
+			register_block_type(
+				$this->asset_api->get_block_metadata_path( $this->metadata_path ),
+				$block_settings
+			);
+			return;
+		}
+
+		// Insert attributes and supports if we're not registering the block using metadata.
```
mikejolley: nit: I think it's ok to use `/* */` for multiline comments.

src/BlockTypes/AbstractBlock.php
```diff
@@ -181,14 +188,26 @@ protected function register_block_type() {
...
+		// These are added after because registering with metadata caused attributes and supports to be empty arrays.
```
mikejolley: I don't understand this comment - we're registering with metadata further up and returning, so why would it affect this?

src/BlockTypes/Checkout.php
```diff
@@ -14,6 +14,13 @@ class Checkout extends AbstractBlock {
 	 */
 	protected $block_name = 'checkout';
 
+	/**
+	 * Metadata path
+	 *
+	 * @var string;
+	 */
+	protected $metadata_path = 'cart-checkout/checkout/block.json';
```
mikejolley: Do you think internally, we could restructure things a bit so it's always derived from `block-name/block.json` to avoid manually setting paths per block class? Then we can just look to see if it exists when registering the block.

assets/js/blocks/cart-checkout/checkout/block.json
opr: Mine indented with spaces instead of tabs, I've updated it now

assets/js/blocks/cart-checkout/checkout/attributes.ts
opr: I'm not sure, would it really make it much easier? I see we lose the comment when moving it to metadata, but besides that what would be easier about leaving them in the block registration function?

src/BlockTypes/AbstractBlock.php
opr: Oops, I made a typo, the comment makes more sense now I've reworded it.

Essentially, if `$block_settings['supports']` or `$block_settings['attributes']` is set, then it overrides what is in metadata so we don't want to pass them to `register_block_type` if we're also using metadata.

src/BlockTypes/Checkout.php
opr: I moved checkout into its own directory and updated the function in `AssetsApi` to find metadata based on the block name, is this what you had in mind?

mikejolley: Yup 👍🏻 thanks

assets/js/blocks/cart-checkout/checkout/attributes.ts
mikejolley: The comment as you say, with a `deprecated` docblock. Some are now split between this file and metadata, so it's going to be hard to remove them in the future unless it's documented somewhere. Not major.

opr: Right, I've moved it to `deprecatedAttributes` in `attributes.ts` this seems to make more sense and makes it more obvious they're deprecated.

assets/js/blocks/checkout/index.tsx
```diff
@@ -24,25 +23,20 @@ const settings = {
...
-	apiVersion: 2,
 	edit: Edit,
 	save: Save,
 	// Migrates v1 to v2 checkout.
 	deprecated: [
 		{
-			attributes: blockAttributes,
+			attributes: {
+				...metadata.attributes,
+				...blockAttributes,
```
mikejolley, line 38: Just checking; we missing deprecatedAttributes here?

mikejolley (commented):
Smoke tests checkout and it works 👍🏻 I made a comment on something that may be missing.

assets/js/blocks/checkout/index.tsx
opr: Yup, good catch! Thanks Mike 😓

assets/js/blocks/checkout/index.tsx
```diff
@@ -24,25 +23,21 @@ const settings = {
...
-	attributes: blockAttributes,
-	apiVersion: 2,
 	edit: Edit,
 	save: Save,
 	// Migrates v1 to v2 checkout.
 	deprecated: [
 		{
-			attributes: blockAttributes,
+			attributes: {
+				...metadata.attributes,
```
alexflorisca, line 37: I would expect only `...deprecatedAttributes` here. Why are we destructuring `metadata.attributes` and `blockAttributes` here?

assets/js/blocks/checkout/index.tsx
```diff
@@ -24,25 +23,21 @@ const settings = {
...
-		'woo-gutenberg-products-block'
-	),
-	supports: {
-		align: [ 'wide' ],
-		html: false,
-		multiple: false,
+	attributes: {
+		...metadata.attributes,
+		...blockAttributes,
+		...deprecatedAttributes,
```
alexflorisca, line 29: Shouldn't `...deprecatedAttributes` only be desctructed under the `deprecated` property below?

alexflorisca (commented):
I'm not gona lie, when I saw 106 files changed I did cry a bit! Good to see most of it was just renaming. Speaking of which, how come you decided to move the checkout block out from cart-checkout? If we're going down this route, does it make sense to do the same for the cart folder?

Code looks good and I can see how this simplifies things. However I just wanted to throw my 2 cents in here as well. I notice that with this approach, most of the attributes have moved to `block.json`...but not all! This make sense as some need to use the `getSetting()` function and can't live in block.json. But this still leaves two sources where attributes are / could be defined. Wouldn't it be clearer to have `attributes.js` as the single source of truth and always import them from there? Considering that we are trying to make it easier for newbies (like myself!) to work on the repository, I think it's more straightforward to remember `all attributes live in `attributes.js` than it is to remember `all attributes live in block.json, unless you need to use any javascript, then put them in attributes.js, and don't forget to merge the two sources in index.js`. I know this might be a bigger change to implement, and may seem unnecessary for some simpler blocks that would otherwise just have block.json, but it seems a bit neater in the grand scheme of things. Has there been any discussion about this? What do you think?

assets/js/blocks/checkout/index.tsx
opr: I don't think so, since `deprecated` needs to take all attributes from the previous iteration of the block, not only the deprecated ones

opr:
> I'm not gona lie, when I saw 106 files changed I did cry a bit! Good to see most of it was just renaming. Speaking of which, how come you decided to move the checkout block out from cart-checkout? If we're going down this route, does it make sense to do the same for the cart folder?
> 
> Code looks good and I can see how this simplifies things. However I just wanted to throw my 2 cents in here as well. I notice that with this approach, most of the attributes have moved to `block.json`...but not all! This make sense as some need to use the `getSetting()` function and can't live in block.json. But this still leaves two sources where attributes are / could be defined. Wouldn't it be clearer to have `attributes.js` as the single source of truth and always import them from there? Considering that we are trying to make it easier for newbies (like myself!) to work on the repository, I think it's more straightforward to remember `all attributes live in `attributes.js`than it is to remember`all attributes live in block.json, unless you need to use any javascript, then put them in attributes.js, and don't forget to merge the two sources in index.js`. I know this might be a bigger change to implement, and may seem unnecessary for some simpler blocks that would otherwise just have block.json, but it seems a bit neater in the grand scheme of things. Has there been any discussion about this? What do you think?

Thanks Alex, yes I agree it would be good to move them all, but for the sake of keeping this PR focused, I don't want to move cart yet, I'll follow that up in another PR

I see where you're coming from with the "mixed" attribute sources, and the goal of this PR is to "unify" the sources of attributes across the cart/checkout inner blocks, I think this is the way it has to be done to keep it consistent with other blocks, they should all be defined in `block.json` where they can be, and fall back to `attributes.ts` where they can't. I guess it's better to do it this way, and have consistently applied this way rather than to have some blocks using `block.json` and some blocks not using it at all. @senadir and @mikejolley  what do you think?

mikejolley:
@opr The problem seems to be those attribute with dynamic values. If that's problematic we could set server defaults in another way, for example, if we deliberately used a `null` or `global` value during registration, when the block loads, replace `null` with the setting from the server. Would that make it more consistent or worse?

opr:
> @opr The problem seems to be those attribute with dynamic values. If that's problematic we could set server defaults in another way, for example, if we deliberately used a `null` or `global` value during registration, when the block loads, replace `null` with the setting from the server. Would that make it more consistent or worse?

Thanks for the suggestion but I'm not super keen on it, while it'd make it a bit more consistent in terms of what files blocks have, you'd introduce complexity and make it harder to reason about where values are coming from, to someone who doesn't know about replacing `null` with something from the server it could be confusing, whereas seeing the attributes spread from `metadata` and imported from `attributes.ts` is more straightforward.

mikejolley:
@opr What about a string with the value `global`? I guess it depends on if we fix the value at time of block saving, or if it inherits the current value of the global setting...

opr:
> @opr What about a string with the value `global`? I guess it depends on if we fix the value at time of block saving, or if it inherits the current value of the global setting...

I still think it adds a bit of complexity, and I don't really see how it's preferable to the alternative (having an `attributes.ts` file, which by the way we probably want to keep too so we can easily see deprecated  attributes, since JSON won't let us have comments to clearly distinguish these)

With that in mind, if you think it's really worth pursuing, I can implement this but I'd prefer to prefix it something like `woo-gutenberg-products-block__global` just in case 😛

mikejolley:
@opr afk right now, but to be clear I meant setting the value, so `“default”: “global”`.  This isn’t major though, only if it’s important that all settings are defined at JSON level. Can be follow up or not at all.

senadir:
Thank you for raising this concern Alex!

Ideally (I word I say a lot tbh), we want all attributes to live in block.json. Why? Because block.json is the standard way of accessing attributes from any medium, whether be PHP or JS, or even other sources in the future.

We need attributes in PHP to assign them when server side rendering the block, or any possible optimization we might need.

Having attributes in block.json is also the standard in Gutenberg, so for a newbie who have some experience, block.json is the first place they will look at.

I agree that having an attribute.js and a block.json is a bit too much. We can delete the attibute.js file in favor of having those dynamic values in index.js.

In the future, we want to have our blocks be more resilient and their definition less reliant on settings, a solution like the one suggested above seems to be favorable for some instances, when we have a sensible default that doesn't rely on settings, and we read from settings in our block.js, which might make more sense.

alexflorisca (approved):
Great comments, thanks for clarifying everyone, it makes sense to me. I didn't consider the PHP side or that it's the standard gutenberg way of doing things so it makes sense to rely on block.json as much as possible. This is a great move towards that 👍

opr:
Thanks for your comments everyone, Nadir and I discussed this and agreed it can be merged as is, but #5652 is now open to track a solution for having attributes defined in two places.

opr:
On the basis that my E2E tests are passing locally, and smoke testing is passing I am going to merge this. Please keep an eye on trunk and report anything you think doesn't seem right with the checkout block!

PR #5595, opened 2022-01-19
Fix global style for Product Summary block, Product Stock Indicator block, and Product Title block

gigitux:
This PR fixes the global style for `Product Summary block`, `Product Stock Indicator block`, and `Product Title block` when these blocks are used in the `All Products` block.

This code register the `Product Summary block`, `Product Stock Indicator block`, and `Product Title` block on the PHP side. This is necessary for using the `__experimentalSelector` from global style API.

### Screenshots

image
*with global style*

image
*without global style*

### Manual Testing

1.  Upgrade to `WordPress 5.9`.
2.  Install and enable the `Twenty Twenty-Two` theme.
3.  Add the `All Product Block` (this block contains `Stock Indicator block`) to a post.
4.  Click on the pencil to edit the block, add the `Stock Indicator Block` and get the focus on the `Stock Indicator block`.
5.  On the right sidebar, personalize the styles of the block.
6.  Go on the page and check if there are changes.
7.  Reset to default using the `Reset` button from the different sections.
8.  Go to Dashboard and select Appearance > Editor (beta). On top of the screen, select Home > Browser all templates > Single Post. When the page is loaded, add the block to the page. Click on the pencil to edit the block, add the `Stock Indicator Block`.
9.  On the Editor page click on the `Styles` icon on the right-top corner.
10. Verify that the `Stock Indicator block` is shown under the `Blocks` section. Personalize the block.
11. Save your changes.
14. Change again the styles.
15. Save your changes.

Repeat from 4th point for `Product Summary Block` and `Product Title Block`.

### User Facing Testing
* [x] Same as above

### DIFF
--- a/src/BlockTypes/ProductStockIndicator.php
+++ b/src/BlockTypes/ProductStockIndicator.php
@@ -1,0 +1,46 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\BlockTypes;
+
+/**
+ * ProductStockIndicator class.
+ */
+class ProductStockIndicator extends AbstractBlock {
+
+	/**
+	 * Block name.
+	 *
+	 * @var string
+	 */
+	protected $block_name = 'product-stock-indicator';
+
+	/**
+	 * API version name.
+	 *
+	 * @var string
+	 */
+	protected $api_version = '2';
+
+	/**
+	 * Get block supports. Shared with the frontend.
+	 * IMPORTANT: If you change anything here, make sure to update the JS file too.
+	 *
+	 * @return array
+	 */
+	protected function get_block_type_supports() {
+		return array(
+			'color'                  =>
+			array(
+				'link'  => false,
+				'background' => false,
+				'text' => true,
+
+			),
+			'typography'             =>
+			array(
+				'fontSize'   => true,
+			),
+			'__experimentalSelector' => '.wc-block-components-product-stock-indicator',
+		);
+	}
+
+}
--- a/src/BlockTypes/ProductStockIndicator.php
+++ b/src/BlockTypes/ProductStockIndicator.php
@@ -30,14 +30,13 @@ class ProductStockIndicator extends AbstractBlock {
 		return array(
 			'color'                  =>
 			array(
-				'link'  => false,
+				'link'       => false,
 				'background' => false,
-				'text' => true,
-
+				'text'       => true,
 			),
 			'typography'             =>
 			array(
-				'fontSize'   => true,
+				'fontSize' => true,
 			),
 			'__experimentalSelector' => '.wc-block-components-product-stock-indicator',
 		);
--- a/src/BlockTypes/ProductSummary.php
+++ b/src/BlockTypes/ProductSummary.php
@@ -1,0 +1,46 @@
+<?php
+namespace Automattic\WooCommerce\Blocks\BlockTypes;
+
+/**
+ * ProductSummary class.
+ */
+class ProductSummary extends AbstractBlock {
+
+	/**
+	 * Block name.
+	 *
+	 * @var string
+	 */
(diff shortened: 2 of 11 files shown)


Aljullu (approved):
Looks good, thanks for fixing this @gigitux!

I noticed there are some global styles that are not applied in the All Products frontend. Ie, testing the Product Title block:

| Style editor | Frontend |
| --- | --- |
| imatge | imatge |

Those seem to be:
* Text color
* Font family
* Font size
* Line height
* Font weight
* Margin

Can you reproduce it too? I don't think it's blocking release, but we should create an issue and fix it in a future release.

I also found another two issues (#5596 and #5597), but they only affect dev builds so not blocking the release at all. However, if you think they might be regressions after the work done in the global styles project, it might make sense adding them to the epic.

src/BlockTypes/ProductStockIndicator.php
```diff
@@ -0,0 +1,46 @@
...
+	 *
+	 * @return array
+	 */
+	protected function get_block_type_supports() {
+		return array(
+			'color'                  =>
+			array(
+				'link'  => false,
+				'background' => false,
+				'text' => true,
```
nielslange: Based on the PHP Coding Standard, we align `=` and `=>` under each other to increase the readability. You can see an example of that on https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation. 

Don't worry, there's no need to manually do that. `npm run lint:php-fix` will do that job for you. 😉

src/BlockTypes/ProductSummary.php
```diff
@@ -0,0 +1,46 @@
...
+	 *
+	 * @return array
+	 */
+	protected function get_block_type_supports() {
+		return array(
+			'color'                  =>
+			array(
+				'link'  => true,
+				'background' => false,
+				'text' => true,
```
nielslange: Same as the comment before.

nielslange (commented):
Functionality-wise, the PR works as expected. 👏 I only noticed two minor points:

1. The `Product Title` cannot be aligned due to https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/345bfcd0b83c6cb5c71a66fc3e402352a23a8e1e/assets/js/atomic/blocks/product-elements/title/style.scss#L1-L3
2. In PHP, `=` and `=>` of a code block should be aligned according to https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#indentation. I left a note that explains how to fix that.

nielslange (approved):
LGTM! 🚢

manospsyx:
@gigitux I'm seeing some 404s related with these blocks when editing widget areas -- is it likely that they are [registered there](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/557cbbc18a1b0db1ad6482abb695996a642bf3af/src/BlockTypesController.php#L200-L208)?

Screen Shot 2022-01-23 at 12 29 06 AM

gigitux:
@manospsyx do you still be able to reproduce this issue? I'm not able to see these errors. I'm on WordPress 5.9.
I installed [Classic Widgets plugin](https://wordpress.org/plugins/classic-widgets/) and navigated to `/wp-admin/widgets.php`, but I didn't see any HTTP error

manospsyx:
Hey @gigitux -- I still see those when navigating to `wp-admin/widgets.php` (WP 5.9, Blocks built from `trunk`). Not sure if it helps but I'm not running the Classic Widgets Plugin. @jimjasson were you also seeing those?

gigitux:
@manospsyx Ok, without `Classic Widgets` I'm able to see the errors. I'm already working on the fix, thanks!

gigitux:
The fix is ready! #5649 

@manospsyx, thanks for reporting it!

PR #5602, opened 2022-01-20
Manually update dependencies

mikejolley:
Just some manual dependency updates to:

- Bump some minor/fix versions of WP packages
- Use newer versions of follow-redirects and markdown-it
- Remove some unused deps
- Moved label sync to dev dependencies

Once tests are green this should be merged to prevent conflicts (this will probably go stale quick)


opr (approved):
Tests are green and code change look good. Cheers Mike.

PR #5614, opened 2022-01-23
Add Shopping Button Inner block

### LINKED ISSUE #5582: Mini Cart: Start shopping button should link to the store
## To reproduce
1. Add the Mini Cart block somewhere.
2. In the frontend, open the Mini Cart drawer and remove all items from your cart until the `Start shopping` button appears.
3. Click on it and notice it redirects to the front page of the store.

## Expected behavior
It should redirect to the shop page, similar to the `Browse store` link from the Cart block.

## Screenshots
imatge

dinhtungdu:
Fixes #5582 

Previously, the `Start shopping` button in the Mini Cart template part is a Button block with the link set to the root domain. The reason for this is: we use the template part file as the starter template for the Mini Cart Drawer, so the root domain (`/`) makes the most sense to be the default URL of the shopping button.

This PR links the correct shop page to the shopping button by creating a new inner block that pulls the `SHOP_URL` from the WC settings. The Shopping Button block can be moved and deleted just like a normal block.

The block is styled as an outline button. The button style and alignment can't be changed. If store owners need more styling options, they should replace the block with a `Button` block.

### Screenshots
N/A. No visual change was introduced.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Install and activate Gutenberg plugin.
2. Clear any customizations to the Mini Cart block.
3. Add Mini Cart block to a page.
4. On the front end, ensure the cart is empty.
5. Open the Mini Cart, see the Shopping button link to the correct Shop page.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/templates/parts/mini-cart.html
+++ b/templates/parts/mini-cart.html
@@ -29,15 +29,9 @@
 		</p>
 		<!-- /wp:paragraph -->
 
-		<!-- wp:buttons -->
-		<div class="wp-block-buttons">
-			<!-- wp:button {"align":"center","className":"is-style-outline"} -->
-			<div class="wp-block-button aligncenter is-style-outline">
-				<a href="/" class="wp-block-button__link">Start shopping</a>
-			</div>
-			<!-- /wp:button -->
-		</div>
-		<!-- /wp:buttons -->
+		<!-- wp:woocommerce/mini-cart-shopping-button-block -->
+		<div class="wp-block-woocommerce-mini-cart-shopping-button-block"></div>
+		<!-- /wp:woocommerce/mini-cart-shopping-button-block -->
 	</div>
 	<!-- /wp:woocommerce/empty-mini-cart-contents-block -->
 </div>
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/index.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/index.tsx
@@ -7,3 +7,4 @@ import './mini-cart-title-block';
 import './mini-cart-items-block';
 import './mini-cart-products-table-block';
 import './mini-cart-footer-block';
+import './mini-cart-shopping-button-block';
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/mini-cart-shopping-button-block/block.json
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/mini-cart-shopping-button-block/block.json
@@ -1,0 +1,26 @@
+{
+	"name": "woocommerce/mini-cart-shopping-button-block",
+	"version": "1.0.0",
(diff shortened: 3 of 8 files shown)


gigitux:
The PR looks great! 

I noticed that now the UI of the button is changed:
image
*with the current PR*

image
*previous implementation*

I will try to restore the previous design

gigitux (changes requested):
The PR looks great! 

I noticed that now the UI of the button is changed:
image
*with the current PR*

image
*previous implementation*

I will try to restore the previous design

gigitux:
After some investigation, I discovered that it is necessary to add this code in the template:
``` 
<!-- wp:buttons -->
	<div class="wp-block-buttons">
		<!-- wp:button -->
		<!-- /wp:button -->
<!-- /wp:buttons -->
```
It is necessary because, in this way, Gutenberg loads the style of buttons from theme settings.

templates/parts/mini-cart.html
```diff
@@ -31,13 +31,15 @@
 
 		<!-- wp:buttons -->
 		<div class="wp-block-buttons">
-			<!-- wp:button {"align":"center","className":"is-style-outline"} -->
-			<div class="wp-block-button aligncenter is-style-outline">
-				<a href="/" class="wp-block-button__link">Start shopping</a>
-			</div>
+			<!-- wp:button -->
```
dinhtungdu: @gigitux Thanks for taking care of this! I think instead of using a fake button to load the core button style, we should add dedicated style for the Shopping Button block.

gigitux: The problem that I see is that every theme has a different style for the button element. In this way, we are able to have the theme style for the button.

dinhtungdu: @gigitux If the `Shopping Button` block is removed then added back, the button style will be broken again because the button block markup is removed.

dinhtungdu: This is also breaking the Site Editor:

https://user-images.githubusercontent.com/5423135/153164658-cf690320-f883-4b58-9e62-87fc7393df8c.mov

gigitux: @dinhtungdu Yes! :( 

How can we style the button to make it consistent with most themes? Do you have any ideas?

dinhtungdu: I think we should keep this block simple with a basic style (square outlined button with color inherit from the parent). If users need more customization, they can replace this block by a button block.

dinhtungdu: I can continue working on this PR if you don't mind : )

gigitux: Yes, sure!

dinhtungdu: I updated the PR, please take another look!

gigitux (approved):
LGTM! 🚢

PR #5618, opened 2022-01-24
Fix: Mini Cart global style is not applied

### LINKED ISSUE #5578: Mini Cart: background global style is not applied
## To reproduce
0. With WP 5.9 and with a block theme (ie: Twenty Twenty Two).
1. Go to the Site Editor and add the Mini Cart block somewhere.
2. Go to global styles, and change the background color of the Mini Cart block.
3. Go to the frontend of your site and notice the background color is not applied.

| Editor | Frontend |
| --- | --- |
| imatge | imatge |

## Expected behavior
If in the editor I chose a green background color, I would expect the green background color to be applied in the frontend too.

* Core version: 5.9-RC2
* Gutenberg version (if installed): disabled
* WooCommerce version: 6.3.0-dev
* WooCommerce Blocks version: `trunk`

dinhtungdu:
Fixes #5578 

The mini cart global style isn't applied on the front-end. This PR fixes that issue by registering the `__experimentalSelector` support on the PHP side. By doing so, the global style for Mini Cart is generated correctly.

Notes:

- The style works as expected at the time it was merged, but it doesn't with the latest version of Gutenberg. I'm not aware of which change in Gutenberg or our plugin caused this issue.
- Using Xdebug, I notice that the blocks data [here](https://github.com/WordPress/gutenberg/blob/8c3b613754540afc52d0c8ea8d8b7c75bf86de74/lib/compat/wordpress-5.9/class-wp-theme-json-gutenberg.php#L518) shows all `woocommerce` blocks have `apiVersion` set to `1`. It means that we aren't registering correct metadata on the PHP side.

### Screenshots

Screen Shot 2022-01-24 at 17 56 53

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

0. Disable the Gutenberg plugin.
1. Make sure you're using WP 5.9 and a block theme (ie Twenty Twenty Two).
2. Add the Mini Cart block to header.
3. Use global style to change the background and text color of the mini cart icon.
4. See style applied on the front end.
5. Enable the Gutenberg plugin and repeat steps above.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -455,4 +455,16 @@ class MiniCart extends AbstractBlock {
 			'display_cart_prices_including_tax' => false,
 		);
 	}
+
+	/**
+	 * Get the supports array for this block type.
+	 *
+	 * @see $this->register_block_type()
+	 * @return string;
+	 */
+	protected function get_block_type_supports() {
+		return [
+			'__experimentalSelector' => '.wc-block-mini-cart__button, .wc-block-mini-cart__badge',
+		];
+	}
 }


sunyatasattva (approved):
This looks good to me. There is a failing bundle size thing, but it's probably just a false positive.

PR #5191, opened 2021-11-18
Checkout Terms Block: Fix Terms and Conditions checkbox position in editor (#5150)

sunyatasattva:
Within the editor, the checkbox appeared aligned at the bottom of the paragraph, while on the front-end it would appear aligned at the top. The alignment was actually correct but a top margin was assigned to checkboxes in the backend.

Instead of removing that globally, this commit resets it for checkboxes within the checkout block.

Fixes #5150

### Screenshots

*Editor (before):*

Screen Shot 2021-11-18 at 23 34 28

*Editor (after):*

Screen Shot 2021-11-18 at 23 34 05

*Reference (front-end):*

fe

### Manual Testing

How to test the changes in this Pull Request:

1. Go to the editor
2. Add checkout block
3. Select **Terms and Conditions** child block
4. Activate `Require checkbox` display option
5. Confirm that the checkbox is aligned to the top of the paragraph

### User Facing Testing
Same as above

### Changelog

> Correctly align Terms and Conditions block checkbox in Checkout block

PR #5633, opened 2022-01-25
Keep top margin for create account section in editor

nielslange:
In #5191, we've adjusted the T&C checkbox position. This change lead to the fact that the top margin of the create account section disappeared. This PR aims to correct this regression.

### Screenshots

<table>
<tr>
<td>Before:
<br><br>

fix-before
</td>
<td>After:
<br><br>

fix-after
</td>
</tr>
</table>

### Testing

1. Create a test page and add the checkout block.
2. Click on email address field.
3. Open settings sidebar.
4. Activate `Account Options » Allow shoppers to sign up for a user account during checkout`.
5. Verify that the `Create an account?` section has sufficient top margin.

### User Facing Testing

* [x] Same as above

### Changelog

> Fix alignment issue with the "create account" section on the checkout block in the editor

### DIFF
--- a/assets/js/blocks/checkout/inner-blocks/checkout-terms-block/editor.scss
+++ b/assets/js/blocks/checkout/inner-blocks/checkout-terms-block/editor.scss
@@ -14,6 +14,10 @@
 
 .wc-block-components-checkbox {
 	margin-top: 0;
+
+	&.wc-block-checkout__create-account {
+		margin-top: 1em;
+	}
 }
 
 .wc-block-checkout__terms_notice .components-notice__action {


nielslange:
Hello @sunyatasattva 👋

Given that you worked on #5191, I manually added you as a reviewer. If you don't have the capacity to do the review, feel free to unassign you. 😀

sunyatasattva (approved):
This LGTM @nielslange thanks for spotting that. We did some regression testing and we didn't see that one.

I wonder if, long term, the best solution is adding exceptions to the margin rule instead of maybe taking a more generic approach. For now, I think this is totally fine! 👍

nielslange:
> I wonder if, long term, the best solution is adding exceptions to the margin rule instead of maybe taking a more generic approach. For now, I think this is totally fine! 👍

I had the same thoughts, @sunyatasattva. What do you think about changing my fix from

```js
.wc-block-components-checkbox {
	margin-top: 0;

	&.wc-block-checkout__create-account {
		margin-top: 1em;
	}
}
```

to 

```js
.wp-block-woocommerce-checkout-terms-block .wc-block-components-checkbox {
	margin-top: 0;
}
```

That way, we only target the element that you inted to affect in #5191?

assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/editor.scss
```diff
@@ -14,8 +14,13 @@
 
 .wc-block-components-checkbox {
```
nielslange, line 15: > I wonder if, long term, the best solution is adding exceptions to the margin rule instead of maybe taking a more generic approach. For now, I think this is totally fine! 👍

I had the same thoughts, @sunyatasattva. What do you think about changing it to:

```suggestion
.wp-block-woocommerce-checkout-terms-block .wc-block-components-checkbox {
	margin-top: 0;
}
```

That way, the CSS definition only targets the element that you intend to change in #5191.

assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/editor.scss
```diff
@@ -14,8 +14,13 @@
...
 	margin-top: 0;
+
+	&.wc-block-checkout__create-account {
+		margin-top: 1em;
+	}
 }
```
nielslange: > I wonder if, long term, the best solution is adding exceptions to the margin rule instead of maybe taking a more generic approach. For now, I think this is totally fine! 👍

I had the same thoughts, @sunyatasattva. What do you think about changing it to:

```suggestion
.wp-block-woocommerce-checkout-terms-block .wc-block-components-checkbox {
	margin-top: 0;
}
```

That way, the CSS definition only targets the element that you intend to change in #5191.

sunyatasattva: I'm actually not sure that the best way is to target the checkbox. The problem is that the checkbox is often associated with a label. If the label wraps to 2 lines, the alignment gets out of whack.

I think we are thinking it backwards. It's probably not the checkbox which needs any margin, but the input that needs a bottom margin. Or perhaps it's the ToS block which uses a strange and unorthodox HTML structure (the label is a `div` next to the checkbox container).

In any case, my point here is that if we keep targeting the margin of the checkbox we might go on playing a whack-a-mole situation.

nielslange:
> In any case, my point here is that if we keep targeting the margin of the checkbox we might go on playing a whack-a-mole situation.

That's true. How would you suggest proceeding in this case, though, to solve the regression bug?

sunyatasattva:
What about adding bottom margin to the text input containers? Can we try that and see if that doesn't create any disaster? Because of how [margin collapsing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) works, that shouldn't be a disaster.

If that makes it a disaster (perhaps because they are floating), then we can just divide the current input margin by two between top and bottom.

What do you think?

assets/js/blocks/cart-checkout/checkout/inner-blocks/checkout-terms-block/editor.scss
nielslange: > What about adding bottom margin to the text input containers? Can we try that and see if that doesn't create any disaster? Because of how [margin collapsing](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing) works, that shouldn't be a disaster.

I looked up this option and currently, all input fields, except the input fields of the first row of each section, use a `margin-top`. Adding a general `margin-bottom` affects all input fields. 

I see two potential solutions:

1. Add `margin-bottom` to the email address input field.
2. Add `margin-top` to the create an account div.

Funny enough, the second option, is my initial fix. Do you want to take another look, @sunyatasattva, or shall we go ahead and merge my initial solution?

sunyatasattva: I don't want to stall you forever on this. I had seen about the `margin-top`: splitting it by two between top and bottom did the trick in my testing.

But, again, I don't know if this is worth stalling this issue even more: if we find any problem later, we'll refer to it and discuss a more robust solution, maybe. For now, I'm going to approve it :)

PR #5634, opened 2022-01-25
Compare assets from builds

### LINKED ISSUE #4399: Add an action that checks for changes to any `*.asset.php` files and highlights them in the conversation of the PR
Following an incident report (pca54o-1Hq-p2) ~~I propose we add a GitHub action that checks for any changes to the `*.asset.php` files generated in the `build` directory when a PR is made.~~

Edited to include Darren's suggestion:

---
**So what this automation would have to do is:**

- Build a branch of current trunk and cache all *.asset.php files (or read from a cache artifact that is automatically created on every merge to trunk).
- Build a branch of the PR and compare the contents of each *.assets.php file generated from the PR build with the trunk build and make note of any changes to dependencies within the php file.
Note, the dependency extraction plugin has the ability to write dependencies to json files instead of PHP, so we could potentially create a Webpack config that does that on a special build for this check (to make the comparison a bit easier to do within a node script).

---

Based on the incident report, we should ensure that context is provided in the comment, explaining _why_ the change is being highlighted. Reviewers should be made aware that changes to dependencies should be discussed and find out why they're being made, the potential performance impact is discussed and weighed up, and any potential compatibility impact, especially in the case of the changes being moved into WooCommerce Core is well considered.

~~Right now I am not sure what actions (if any) we can use to make this easy, but it's essentially going to be a check to see if any files matching `*.asset.php` are present in the diff.~~

mikejolley:
This PR adds a new workflow/action to detect changes to built `*.asset.php files` included with WooCommerce Blocks. The asset files list external script dependencies that WordPress will enqueue. If we introduce new dependencies accidentally (pca54o-1Hq-p2) this can have a negative effect on performance.

Screenshot 2022-01-26 at 18 16 28

The action works as follows:

1. When a new PR is submitted, the workflow runs
2. It builds `trunk` and builds a single `asset.json` file which lists all external dependencies for all script handles. This is uploaded as an artifact.
3. A second build of this branch does the same.
4. The branch `asset.json` is then compared against the asset artifact for trunk using a custom JS script.
5. If there are changes, a report is generated, and a comment is left on the PR.

You can see some example reports below which I generated by including fake data. When this is merged to core, all new PRs will generate reports against `trunk`.

**Testing**

Unfortunetely because this targets `trunk`, which does not yet contain the action/workflow, this won't work or be testable. I have however made it generate some sample reports which can be seen below (testing against a different target branch). When this is merged, new PRs will use the new action.

Fixes #4399


tarhi-saad (commented):
Thank you, @mikejolley, for building this workflow!

It seems challenging to test the changes since the workflow doesn't exist on the `trunk` branch yet. But, based on the actions steps described in this PR and the sample reports generated, the changes align with the requirements from this issue #4399. 🙌

> 1. When a new PR is submitted, the workflow runs
> 2. It builds trunk and builds a single asset.json file which lists all external dependencies for all script handles. This is uploaded as an artifact.
> 3. A second build of this branch does the same.

Does this mean that two builds will be running on each new submitted PR, one on the `trunk` branch and the other on the current one?
Couldn't we skip the trunk build at this step and use the last `asset.json` file generated after the previous merge to `trunk`?

The action would go something like this:

1. When a new PR is submitted, the workflow runs
2. A build of this branch generates a single `asset.json` file.
3. This branch's `asset.json` is then compared against the asset artifact generated after the latest merge to `trunk` using a custom JS script.
4. ....

The other aspect of this action would be to build the `trunk` branch and generate a new `asset.json` file following every merge instead.

Regarding the code, I'm not very knowledgeable about GitHub actions, so if you think a more in-depth review of the code is necessary, please feel free to request a second reviewer for this task! 👍

mikejolley:
Thanks @tarhi-saad 

[Going by this comment, an artifact may cause some edge case issues](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/4399#issuecomment-868706097):

> Perhaps a build of trunk each time would be better than a cache artefact, that way any PRs created immediately after a merge into trunk but before the creation of this artefact would be checked properly. It's an edge case but worth considering.

I'm not sure how much of an issue this would be. But to answer your question, yes, in the current state it does two builds; one for trunk, and one for the branch. This takes around 1.5mins, but runs in tandem with the branch build.

I could try the trunk artifact version, but not sure how yet or how I can test that. No workflows can run on trunk until the workflow is added in trunk.

tarhi-saad (approved):
Thank you, @mikejolley, for the explanation. I do understand better the issue at hand! Indeed, running a `trunk` build each time would be the safest route to avoid this edge case. And we can't do further tests until this workflow is added to `trunk`.

The current changes look good to me! Thank you, you've done a great job! 👍

mikejolley:
Rebased and ready. Will merge this now and keep an eye on builds.

PR #5637, opened 2022-01-25
Fix Legacy Template block being stuck with stale data when changing templates

### LINKED ISSUE #5636: Full Store Editing: Block templates are showing wrong information in the Site Editor.
## Describe the bug
If you load the `single-product` block template in the Site Editor and then switch to load the `archive-product` it will load the Single Product block in the Editor. This happens with other WooCommerce block templates also, not just the Single Product one.

## To reproduce

1. Go to Site Editor > Templates
2. Click on Single Product
3. Navigate back to Templates
4. Click on Archive Product
5. The editor block should show "Single Product Block"

## Expected behavior
It should show the correct block for archive product.

## Screenshots
Screenshot 2022-01-25 at 14 54 53

tomasztunik:
Legacy Template block when navigating between different page templates would be stuck in a stale state representing [block definition](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/legacy-template/constants.ts) for the first page template that user have visited it with. This would possibly cause the block to [not be insertable](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/legacy-template/index.tsx#L95) with block editor again if the user deleted Legacy Template block and the first template they visited was not one of the page template types that [this block supports](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/legacy-template/constants.ts). It would not show up in block selector anymore.

This would get fixed on page reload but only for current page and when navigating user would be stuck with the block state representing the view on which the page was refreshed.

This fix ensures that we remain subscribing to changes to the current edited page template type and update the Legacy Template block's definition to properly represent the context in which it is being displayed. By properly unregistering and registering the block with an updated definition we ensure that it always displays the correct state for the active page template type.

A note that might be useful to others - registered blocks are not mutable and to change the block's definition we have to explicitly unregister the block first before registering a new definition for an already existing block type.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5636

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

This PR includes regression testing around the feature due to the nature of the bug and its fix.

_"Start again*" - Start your journey outside of the Site Editor e.g. on the Appearance > Themes page._

1. First confirm you can replicate the bug in the linked issue.
2. Checkout this branch and start outside of the Site Editor e.g. Appearance > Themes
3. Load Appearance > Site Editor and select one of the WooCommerce templates. Confirm it loads the correct Legacy Template block in the Editor for the given block template.
4. Remove this block and confirm you are able to reinsert it.
5. Start again* and then load the Appearance > Site Editor. Select the "Single Post" template, confirm you're **not able** to insert the Legacy Template block into this template.
6. Navigate to the Site Editor > Templates list and edit a WooCommerce template (e.g. Single Product). Confirm again that the correct Legacy Template block gets rendered in the Site Editor and that you're able to remove and reinsert it.
7. Refresh the page and confirm the correct Legacy Template block still gets rendered in the Site Editor and that you're able to remove and reinsert it.
8. Ensure all the WooCommerce templates are rendering correctly on the frontend and any customisations are correctly reflected on the frontend.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Performance Impact

We are now subscribing to store changes and unregister/reregister the Legacy Template block if the template ID changes.

### Changelog

> Fix: WooCommerce block templates show correct information in the Site Editor.

### DIFF
--- a/assets/js/blocks/legacy-template/index.tsx
+++ b/assets/js/blocks/legacy-template/index.tsx
@@ -2,7 +2,7 @@
  * External dependencies
  */
 import { select, subscribe } from '@wordpress/data';
-import { registerBlockType } from '@wordpress/blocks';
+import { registerBlockType, unregisterBlockType } from '@wordpress/blocks';
 import { WC_BLOCKS_IMAGE_URL } from '@woocommerce/block-settings';
 import { useBlockProps } from '@wordpress/block-editor';
 import { Placeholder } from '@wordpress/components';
@@ -72,29 +72,38 @@ const Edit = ( { attributes }: Props ) => {
 };
 
 let templateId: string | undefined;
+let prevTemplateId: string | undefined;
 
-const unsubscribe = subscribe( () => {
+subscribe( () => {
 	const store = select( 'core/edit-site' );
 
 	if ( ! store ) {
-		// The store will only exist in the Site Editor so we need to unsubscribe and early return for Posts / Pages.
-		unsubscribe();
+		// The store will only exist in the Site Editor so we need to early return for Posts / Pages.
 		return;
 	}
 
 	templateId = store?.getEditedPostId();
 
-	if ( templateId ) {
-		unsubscribe();
+	// Legacy template block changes depending on the context of currently edited template.
+	// To avoid infinite loop on store updates we need to put a guard here to update the registerd Block Type
+	// only if the current edited post template id has changed.
+	if ( templateId && templateId !== prevTemplateId ) {
 		const currentTemplateSlug = templateId?.split( '//' )[ 1 ];
 		const templateData = getMatchingTemplateData(
 			TEMPLATES,
 			currentTemplateSlug
 		);
+
+		prevTemplateId = templateId;
+
 		// We only want this block to be available for use in specified WooCommerce templates.
 		const eligibleForInserter = templateData !== null;
 		const title = templateData?.title ?? currentTemplateSlug;
 		const placeholder = templateData?.placeholder ?? 'fallback';
+
+		// We need to unregister the block before we can register its new variant again
+		// (ie. Single Product, Product Archive).
+		unregisterBlockType( 'woocommerce/legacy-template' );
 
 		registerBlockType( 'woocommerce/legacy-template', {
 			title,


tomasztunik:
We’ve noticed that the current implementation on the legacy-template block depending on execution order/initialisation of Gutenberg and WooCommerce-Gutenberg-Products-Block plugins could cause the legacy template block to never register because of the early unsubscribe at https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/legacy-template/index.tsx#L81.

It would then show this warning in the editor:

Screenshot 2022-01-26 at 10 42 29

tjcafferkey:
**tl;dr: Let's revert the [PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5545) that introduced this issue so it's included in the release on Monday the 31st of Jan, whilst we continue trying to fix the problems we have uncovered in this PR.**

So the original issue this was trying to fix was only allowing users to insert the Legacy Template block to WooCommerce templates only. Prior to [this PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5545) if you removed the Legacy Template block from your template you could not reinsert it and thus had to reset the customizations on the template to force it to be loaded from the [filesystem](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/templates/templates) again.

With this approach in its current state we are:
* Subscribing to store changes before determining whether we want to register the block depending on the current template ID.
* Unregistering the block, and then reregistering it.

As it stands these are the problems I’ve seen with our approach:
* If we customise a template and then save it, it does not save the attributes for that template (saved templates get stored in the DB, this is important to note because we do not want to release a feature to users which could store incorrect data to their database as thats hard to fix for us) and as a result of this, customized WooCommerce templates dont appear on the frontend.
* When we load a saved template in the Site Editor, we get the markup from the DB before we register the block so Gutenberg complains that the block its trying to load is not registered yet. When the subscribe callback runs this rerenders and fixes the editor but still not ideal.

We _could_ look at fixing those two points in our current approach but a few things we need to consider which is making me think twice is:

* Gutenberg was not meant to work like this, and being an Automattic owned product I think it’s important for us to adhere to Gutenberg best practices
* If Gutenberg changes how it handles its routing in the Site Editor (we've seen the Site Editor change before), there is a chance we could end up subscribing to the store multiple times and end up with a memory leak which would break the feature for our users until we were able to get a release out.

So, reflecting on all of that I believe that the original problem of “if you removed the Legacy Template block from your template you could not reinsert it” seems like a better problem to have at this point. And I propose we undo our changes and go back to what it was prior to [this PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5545) and then find a different way to fix that original problem, if it’s possible. If it’s not we should open an issue in Gutenberg and get this feature looked into upstream

tomasztunik:
closing in favour of another path to solving underlying issue

PR #5653, opened 2022-01-27
Fix: only allow Mini Cart block in Widgets and Site Editor

### LINKED ISSUE #5576: Mini Cart: only allow adding the Mini Cart block inside the Site Editor and the widgets screen, but not in posts or pages
## Describe the bug
We discussed restricting the Mini Cart block to only be available inside the Site Editor and the widgets screen, so it couldn't be added to a post or page; however, I think we never created an issue for it.

## To reproduce

1. Create a post or page.
2. Try to add the Mini Cart block and notice you are allowed to.

## Expected behavior
You shouldn't be allowed to add the Mini Cart block inside a post or page, it should only be available inside the Site Editor and the widgets screen.

dinhtungdu:
Fixes #5576 

This PR prevents adding the Mini Cart block to posts, pages, or any custom post types by checking the `$pagenow`.

### Screenshots

image

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Edit a page
2. Try to add Mini Cart block, see no block found in the Inserter.
3. Go to Widgets (I used Storefront to test Widgets), see you can search and add Mini Cart to widget areas.
4. Using a block theme with Gutenberg enabled, go to the Site Editor, see you can search and add the Mini Cart block.
5. Using WP 5.9 with Gutenberg disabled, go to the Site Editor, see you can search and add the Mini Cart block.
### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.


src/BlockTypesController.php
```diff
@@ -187,8 +187,14 @@ protected function get_block_types() {
...
-			$block_types[] = 'MiniCart';
-			$block_types[] = 'MiniCartContents';
+
+			/**
+			 * Mini Cart blocks shouldn't be available in the Blocks Editor.
+			 */
+			if ( 'post.php' !== $pagenow ) {
+				$block_types[] = 'MiniCart';
+				$block_types[] = 'MiniCartContents';
+			}
```
dinhtungdu: Currently, this check does a bit different from the original issue requirement. We're excluding the Mini Cart block from the Blocks Editor, while #5576 wants to **only allow adding the Mini Cart block inside the Site Editor and Widgets**. In the future, we may have other new editors that we may not want the Mini Cart to be available there.

The following code does exactly what #5576 requires:

```suggestion
			/**
			 * Mini Cart blocks should be available in Site Editor and Widgets only.
			 */
			if (
				'widgets.php' === $pagenow ||
				'site-editor.php' === $pagenow ||
				! empty( $_GET['page'] ) && 'gutenberg-edit-site' === $_GET['page'] // phpcs:ignore WordPress.Security.NonceVerification
			) {
				$block_types[] = 'MiniCart';
				$block_types[] = 'MiniCartContents';
			}
```

dinhtungdu:
Update: I decided to explicitly register the Mini Cart blocks only for Site Editor and Widgets screen. This is more future-proof IMO.

gigitux (changes requested):
With the implemented check the block is not loaded on FE side. I tried with a classic theme and with a block theme. 

Maybe should we add this check too?

```
'index.php' === $pagenow
```

gigitux:
I updated the PR. I preferred to use the function `is_admin` (not great naming. [In accordance with the docs](https://developer.wordpress.org/reference/functions/is_admin/), this function check what page is loaded, not the user role) instead to this check:

```
'index.php' === $pagenow
```

Aljullu (commented):
Works perfectly! :ok_hand: 

The only thing that needs to be fixed is that this change breaks the Mini Cart e2e tests we have in place because they are run in the post editor. We will probably need to refactor them so they run in the Widgets Screen (we have a [`openWidgetsEditorBlockInserter()` util](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/cd5c5dedf731aa5a2d6938ecd60a4f55b9b55181/tests/e2e/utils.js#L78-L82) that might be handy) or the Site Editor (@sunyatasattva is [working on exposing some utils](https://github.com/WordPress/gutenberg/pull/38463) from `@worpdress/e2e-utils`).

gigitux:
@Aljullu I added the tests for the widget area. For Site Editor, I guess that we should wait until we upgrade `@wordpress/e2e-test-utils`: this is the reason why I added the @todo comment

tests/e2e/specs/backend/mini-cart.test.js
```diff
@@ -20,52 +20,72 @@ if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 3 ) {
...
 			const compatibilityNoticeTitle = await page.$x(
 				`//h1[contains(text(), 'Compatibility notice')]`
 			);
 			expect( compatibilityNoticeTitle.length ).toBe( 1 );
 		} );
-	} );
 
-	describe( 'after compatibility notice is dismissed', () => {
-		beforeAll( async () => {
+		it( "after the compatibility notice is dismissed, this latter doesn't appear anymore", async () => {
```
Aljullu: Nitpicking, but I feel this test title can be simpler:
```suggestion
		it( "after the compatibility notice is dismissed, it doesn't appear again", async () => {
```

tests/e2e/specs/backend/mini-cart.test.js
```diff
@@ -20,52 +20,72 @@ if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 3 ) {
...
+				`//button[@aria-disabled]//span[text()='${ block.name }']`
+			);
 
-		it( 'renders without crashing', async () => {
-			await expect( page ).toRenderBlock( block );
+			expect( miniCartButton ).toHaveLength( 1 );
 		} );
 	} );
+
+	// @todo Add tests for the block for FSE editor
```
Aljullu: We have a bot that converts all `@todo` comments into issues, so if you specify the name of the block it will be easier to understand what block the issue refers to.
```suggestion
	// @todo Add tests for the Mini Cart block in FSE editor
```

Aljullu (approved):
Good job updating the tests, @gigitux!

> For Site Editor, I guess that we should wait until we upgrade `@wordpress/e2e-test-utils`: this is the reason why I added the `@todo` comment

Yes, makes sense. I think we can merge this PR and work on the Site Editor tests in a follow-up.

gigitux:
Thanks for the suggestion! I updated the test title 👍

PR #5668, opened 2022-01-28
Ts blocks

alexflorisca:
This PR converts the following components to Typescript
- `base/components/ProductList`
- `base/components/FilterElementLabel`
- `base/components/FilterSubmitButton`
- `base/components/Form`
- `base/components/LoadMoreButton`
- `base/components/LoadingMask`
- `base/components/Pagination`
- `base/components/SortSelect`

The ProductList component was quite tricky to get completely error free for a couple of reasons:

- Couldn't find the right way to type `withScrollToTop` so that it accepts `ProductList` as a component. The HOC expects a constructor, yet `ProductList` returns a `JSX.Element`. If anyone has any bright ideas here, I'd love to hear them. I've tried to type the `OriginalComponent` prop in the `withScrollToTop` HOC as `React.ComponentType`, as suggested [here](p1643129926106600-slack-CLCNELZE1)
- I'm casting an `unknown[]` as `ProductResponseItem[]` as a temporary shortcut. I've left an inline comment with more explanation about this.

#### Automated Tests
  * [x] Unit tests
  * [x] E2E tests
### Manual Testing

1. Smoke test the All Products block at the least, adding extra inner blocks and making sure they render correctly. 

### User Facing Testing
* [x] Same as above


assets/js/base/context/hooks/use-store-products.ts
```diff
@@ -34,8 +44,8 @@ export const useStoreProducts = ( query ) => {
 		query,
 	} );
 	return {
-		products,
-		totalProducts: parseInt( totalProducts, 10 ),
+		products: products as ProductResponseItem[], // TODO: Remove this once getCollection selector and resolver is converted to TS.
```
alexflorisca, line 47: On line [38 above](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5668/files#diff-badb52abe6c3543682e0a447fe296a3bbbee37d80b05d5e6e79685963e76b860R38) we call the `useCollection` hook, which looks up some data from a store and returns it. We need to convert quite a few other things to TS in order to get the correct types here (`useCollection` hook, some selectors and resolvers in the data folder, and it's not an easy task as most of these return generic nested objects so it needs a little more thought. I didn't have the time to convert everything needed by the  `useStoreProducts` hook in order to return the correct types from here, so I cast it to an array of `ProductResponseItem` which we are expecting here until we get to convert all other relevant hooks and state related code to TS.

assets/js/base/components/product-list/product-sort-select/index.tsx
```diff
@@ -54,7 +55,6 @@ const ProductSortSelect = ( { onChange, readOnly, value } ) => {
 					),
 				},
 			] }
-			readOnly={ readOnly }
```
alexflorisca, line 57: removed the readOnly property from the SortSelect component as it was never used, and was being applied to a <select> HTML Element which is not valid

assets/js/base/components/product-list/product-sort-select/index.tsx
```diff
@@ -2,19 +2,20 @@
...
 
-const ProductSortSelect = ( { onChange, readOnly, value } ) => {
+const ProductSortSelect = ( {
+	onChange,
+	value,
+}: ProductSortSelectProps ): JSX.Element => {
 	return (
 		<SortSelect
 			className="wc-block-product-sort-select wc-block-components-product-sort-select"
-			name="orderby"
```
alexflorisca, line 17: removed invalid `name` property being passed to SortSelect component. This is not defined or used in `SortSelect`

tomasztunik:
Hi Alex, got a question regarding typing the components. 

Is there a reason we were not using generic `React.FC` and `React.VFC` functional component types but rather than that we type children when present and return type on every component definition?

Ie. for `React Void Functional Component / VFC`, children-less
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8e707a8931d9518b8d364ccc49ba9e9363943a1c/assets/js/base/components/filter-element-label/index.tsx#L23-L26

```typescript
interface FilterElementLabelProps {
	name: string;
	count: number;
}

const FilterElementLabel: React.VFC< FilterElementLabelProps > = ( {
	name,
	count,
} ) => {
```

and for `React Functional Component / FC`, children included
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8e707a8931d9518b8d364ccc49ba9e9363943a1c/assets/js/base/components/form/index.tsx#L7-L17

```typescript
 interface FormProps { 
 	className: string; 
 	onSubmit: ( event: FormEvent ) => void; 
 } 
  
 const Form: React.FC< FormProps > = ( { 
 	className, 
 	children, 
 	onSubmit = ( event ) => void event, 
 } ) => { 
    // ...
 }
```

Using `React.FC` and `React.VFC` we can be explicit about if the component accepts children or not without having to type them every time or having to inspect props to learn what type of component it was. On top of it, we would be future compatible if the base functional component interface evolved.

I've noticed it was used only once in the [Chip](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/dfe4c46fe7c5b7bfd5ab7bd81cfb38834f84c98b/assets/js/base/components/chip/chip.tsx#L45) component but I believe this to be the most explicit way to type functional components in React and would love us to follow this typing pattern.

What do you think?

alexflorisca:
Hi @tomasztunik,

Thanks for your comment. I agree that using `React.FC` is neater, however there is a lot of evidence that it's not actually best practice, and it's better to be explicit defining the types of the props and return type. There's been some discussion around this and I'll point you to a couple of P2 articles, (pca54o-2ym-p2 and pca54o-2WY-p2) for further reading and some links to a bunch of articles & prs that explain in more detail why using `React.FC` is generally a bad idea.

assets/js/atomic/utils/get-block-map.ts
```diff
@@ -13,5 +13,7 @@ import '../blocks/component-init';
  *
  * @param {string} blockName Name of the parent block.
  */
-export const getBlockMap = ( blockName ) =>
+export const getBlockMap = (
+	blockName: string
+): Record< string, React.ComponentType > =>
```
mikejolley: I see some types in our checkout block registry which look related and use:

```
component:
		| LazyExoticComponent< React.ComponentType< unknown > >
		| ( () => JSX.Element | null )
		| null;
```

Is this typing the same thing? https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8475bb8c47499ad535aceb4391d9bafa30ec3228/packages/checkout/blocks-registry/types.ts#L36

assets/js/base/components/filter-element-label/index.tsx
```diff
@@ -9,20 +9,27 @@ import Label from '@woocommerce/base-components/label';
...
+	name,
+	count,
+}: FilterElementLabelProps ): JSX.Element => {
 	return (
 		<>
 			{ name }
 			{ Number.isFinite( count ) && (
 				<Label
-					label={ count }
+					label={ count.toString() }
```
mikejolley, line 32: I know this fixes the type, but I see some usage of this component like:

```
count={ blockAttributes.showCounts ? count : null }
```

I think this may error? We should fix the cases which consume this component, or make it optional.

assets/js/base/components/form/index.tsx
```diff
@@ -2,14 +2,20 @@
  * External dependencies
  */
 import classnames from 'classnames';
-import PropTypes from 'prop-types';
+import { FormEvent } from 'react';
```
mikejolley: Should this be `import type`?

assets/js/base/components/load-more-button/index.tsx
```diff
@@ -2,15 +2,25 @@
  * External dependencies
  */
 import { __ } from '@wordpress/i18n';
-import PropTypes from 'prop-types';
 import Label from '@woocommerce/base-components/label';
+import { MouseEventHandler } from 'react';
```
mikejolley: `import type`?

assets/js/base/components/loading-mask/index.tsx
```diff
@@ -11,14 +10,21 @@ import classNames from 'classnames';
 import './style.scss';
 import Spinner from '../spinner';
 
+interface LoadingMaskProps {
+	children?: React.ReactChildren;
```
mikejolley: I see in other places in the codebase we are using `ReactNode` or `ReactNode[]` rather than `React.ReactChildren; `. Do you know which is best practice, and can we decide on a standard to use throughout?

assets/js/base/components/pagination/index.tsx
```diff
@@ -2,36 +2,66 @@
...
 		currentPage,
 		totalPages
 	);
+
 	const showFirstPage = displayFirstAndLastPages && Boolean( minIndex !== 1 );
 	const showLastPage =
 		displayFirstAndLastPages && Boolean( maxIndex !== totalPages );
 	const showFirstPageEllipsis =
-		displayFirstAndLastPages && Boolean( minIndex > 3 );
+		displayFirstAndLastPages && Boolean( minIndex && minIndex > 3 );
```
mikejolley, line 61: Would this logic be simplier if we provided the default for `minIndex`?

assets/js/base/components/product-list/container.tsx
```diff
@@ -3,24 +3,28 @@
  */
 import { useState, useEffect } from '@wordpress/element';
 import PropTypes from 'prop-types';
+import { HTMLElementEvent } from '@woocommerce/type-defs/utils';
```
mikejolley:
```suggestion
import type { HTMLElementEvent } from '@woocommerce/type-defs/utils';
```

assets/js/base/components/product-list/no-matching-products.tsx
```diff
@@ -5,7 +5,14 @@ import { __ } from '@wordpress/i18n';
...
 import { Icon, search } from '@woocommerce/icons';
 
-const NoMatchingProducts = ( { resetCallback = () => {} } ) => {
+/**
+ * Internal dependencies
+ */
+import { NoMatchingProductsProps } from './types';
+
+const NoMatchingProducts = ( {
+	resetCallback = () => void 0,
```
mikejolley, line 14: Noob question, what is 0 in the `void 0` for?

assets/js/base/components/product-list/product-list-item/utils.tsx
```diff
@@ -13,21 +19,22 @@ import { Suspense } from '@wordpress/element';
...
 	}
 
 	const blockMap = getBlockMap( blockName );
-
 	return layoutConfig.map( ( [ name, props = {} ], index ) => {
-		let children = [];
+		let children = [] as ( JSX.Element | null )[] | undefined;
 
 		if ( !! props.children && props.children.length > 0 ) {
+			// Qprops.children here refers to the children stored in the block attributes. which
```
mikejolley: `Qprops` is a type?

assets/js/base/components/product-list/product-list.tsx
```diff
@@ -195,7 +207,7 @@ const ProductList = ( {
 		! Number.isFinite( totalProducts ) &&
 		Number.isFinite( previousQueryTotals?.totalProducts ) &&
 		isEqual( totalQuery, previousQueryTotals?.totalQuery )
-			? Math.ceil( previousQueryTotals.totalProducts / perPage )
+			? Math.ceil( ( previousQueryTotals?.totalProducts || 0 ) / perPage )
```
mikejolley, line 210: Can we force this shape earlier to avoid checking for existance of totalProducts? Just for readability sake.

assets/js/base/components/sort-select/index.tsx
```diff
@@ -1,40 +1,63 @@
...
+	 */
+	label?: string;
+	/**
+	 * Function to call on the change event.
+	 */
+	onChange: ChangeEventHandler;
+	/**
+	 * Option values for the select.
+	 */
+	options?: {
```
mikejolley: This should probably be required otherwise the sort select has no function.

mikejolley (commented):
Hi @alexflorisca 

Overall looks good, I added some inline feedback and questions.

assets/js/atomic/utils/get-block-map.ts
alexflorisca: Yep, thanks for pointing this out. I've extracted a type `RegisteredBlockComponent` and re-used it instead

assets/js/base/components/filter-element-label/index.tsx
alexflorisca: I see what you mean. Those occurrences should also pass `count.toString()` but they're not typed yet. The `Label` component expects a string as the `count`, but it doesn't do any number logic with the count, it just outputs it as inside a HTML element, so should be safe.

assets/js/base/components/form/index.tsx
alexflorisca: It [doesn't matter too much](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html), but it's a little safer and more declarative to use type so I've changed it 👍

assets/js/base/components/loading-mask/index.tsx
alexflorisca: You're right, there are multiple ways to type children and [this article](https://itnext.io/using-react-children-prop-with-typescript-37f6821351bb) sums them up pretty well. I suggest we stick with ReactNode or ReactNode[] for now as our codebase is so complex, that narrowing children down further will require changes at multiple levels where components are nested within each other. It will be quite time consuming to trace down every component that the `children` pass through and modify the types, I don't think it's worth the effort. It's something to bare in mind going forward though: being more specific with the types of children that a component is allowed is the better way to go and will catch more errors

assets/js/base/components/product-list/container.tsx
alexflorisca: 👍  I've also changed `@woocommerce/type-defs/utils` to `@woocommerce/types` and exported `utils.js` in the main `types/index.ts` file.

assets/js/base/components/product-list/no-matching-products.tsx
alexflorisca: `void` is an operator in JS, which means it needs an expression after it. Much like you can't just write `const` without defining a value after it. You can read up about it more [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void) if you want. There was also a [discussion on slack](p1638213487071400-slack-C8X6Q7XQU) about the standard way to type empty functions and we decided to use `() => void 0`

assets/js/base/components/product-list/product-list-item/utils.tsx
alexflorisca: That's a typo, sorry, it's meant to say `props.children`, good catch 👍

assets/js/base/components/product-list/product-list.tsx
alexflorisca: The problem here is that `previousQueryTotals` could be `undefined`. I think this is the cleanest way using the optional chaining (`?.`) operator and defining a default value inline if the prop isn't defined. The alternative I see would be to wrap that entire block in an `if(previousQueryTotals) {}` statement. I don't think we can force `previousQueryTotals` to never be undefined. If you have any other ideas I'm open to suggestions

alexflorisca:
Thanks for the review @mikejolley, I've addressed most of what you rightly pointed out and answered some of your questions inline. I've also fixed the unit tests so fingers crossed everything should be green. Let me know if you spot anything else!

mikejolley (approved):
Thanks for replying to feedback.

PR #5675, opened 2022-01-28
Add Checkout Order Summary inner block tests

### LINKED ISSUE #5625: Add tests for the Checkout Order Summary block
This issue is to track the addition of tests for Checkout inner blocks. The Checkout Order Summary block needs tests to be written for it.

opr:
This PR adds tests for the Order Summary inner block.

There was no need to add tests for this block in the editor context due to the only control changing the `showRateAfterTaxName` which should be tested further down the stack in the `TotalsTaxes` component.

I added two utility functions to the `tests/utils` folder which allow you to search for text across elements.

The `textContentMatcher` function will allow you to find text when it is structured like this:

```
<div>Some <span>nested text</span></div>
```

and `textContentMatcherAcrossSiblings` will let you find text when it is structured like this:
```
<div>
    <span>Text</span>
    <span>nested</span>
    <span>in sibling</span>
    <span>elements.</span>
</div>
```

Fixes #5625

#### Accessibility

#### Other Checks

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests
### Manual Testing

1. `npm run test` and ensure the tests pass

There are no user-facing testing instructions required.

### Changelog

> Skipping


nielslange (approved):
Pre-approving this PR as it looks great! Also props for introducing the following functions:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2a2f7c51bd0740a54406fdfb11647793312cffb1/tests/utils/find-by-text.ts#L13

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2a2f7c51bd0740a54406fdfb11647793312cffb1/tests/utils/find-by-text.ts#L36

**Nitpic** As for mocking the shipping rates, I wonder if it wouldn't be better to extract them, so that other tests can also use them. Or would you say it's better that every test creates their own specific mocks? Again, this is just nitpicking and feel free to merge this PR. I was just wondering if extracting these mocks could be beneficial in the future.

opr:
> As for mocking the shipping rates, I wonder if it wouldn't be better to extract them

Yes it would be, but I'm not sure (right now) if that's doable because of the way jest handles the mocks. I'll make a follow up issue to check if it's possible, but for now I will merge the PR.

Thanks for reviewing Niels 👏🏼 

https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5677 opened to track

PR #5691, opened 2022-02-01
fix: accessibility of search products block

### LINKED ISSUE #1671: Button in Product Search block is missing accessible text
Axe returns an issue when testing the _Product Search_ block:

>  Buttons must have discernible text

It refers to the arrow (`>`) button.

**To Reproduce**
1. Install the Axe extension ([Firefox](https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/), [Chrome](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd)).
2. Create a post with a _Product Search_ block and run Axe analysis.
3. Notice the `Buttons must have discernible text` error appears.

**Screenshots**
imatge

This issue is a bit tricky because directly changing the markup invalidates the block.

tomasztunik:
When output with php the block had correct accessibility on
the frontend but in the editor or when rendered as block
the correct aria label was missing because of misspelled
aria-label attribute.

Also fixed wrong value type provided for tabIndex in editor
while at it.

Fixes #1671

#### Accessibility

- [✔️  ] I've tested using only a keyboard (no mouse)
- [✔️ ] I've tested using a screen reader

### Manual Testing

In editor and on the frontend:

1. Verify using Axe extension ([Firefox](https://addons.mozilla.org/en-US/firefox/addon/axe-devtools/), [Chrome](https://chrome.google.com/webstore/detail/axe-web-accessibility-tes/lhdoppojpmngadmnindnejefpokejbdd)).
2. There should be no `Buttons must have discernible text` error in accessibility report

### User Facing Testing
* [ ] Same as above, or
* [ ] See steps below.

With screenreader (CMD+F5 on OSX)

1. enable screen reader
2. focus on the search, tab to the button
3. the button should respond with `Search button` audio description

### Changelog

> ~Fixed accessibility label for submit Search button in Product Search block~ reverted in #5717


assets/js/blocks/product-search/edit.js
```diff
@@ -115,9 +115,12 @@ const Edit = ( {
...
 						type="submit"
 						className="wc-block-product-search__button"
-						label={ __( 'Search', 'woo-gutenberg-products-block' ) }
+						aria-label={ __(
+							'Search',
+							'woo-gutenberg-products-block'
+						) }
 						onClick={ ( e ) => e.preventDefault() }
-						tabIndex="-1"
+						tabIndex={ -1 }
```
tomasztunik, line 123: grabbed it along the way, valid types here are `number | undefined` only

danieldudzic:
I have tested the fix using Axe and the error is resolved ✅  - also the update doesn't invalidate the block on my test site:


Edit_Post_‹_kirigami_—_WordPress

gigitux:
LGTM! 
Thanks for updating the type error too! 👍

gigitux (approved):
LGTM! 
Thanks for updating the type error too! 👍

PR #5708, opened 2022-02-02
Add to cart button ignored redirect to cart setting

### LINKED ISSUE #1766: All Products: Ignores Setting - Redirect to the Cart Page on Successful Addition
**Describe the bug**
When you click on "Add to Cart" from a product in the All Products block, it ignores Add to cart behavior when set to Redirect to the cart page after successful addition, and stays on the page. 

**To Reproduce**

1. WooCommerce » Settings » Products » General
2. Make sure "Redirect to the cart page after successful addition" is checked
3. On the page set as the Shop Page, add an All Products bloc and update.
4. Now, on the Shop page add a product from the All Products block. 
5. Notice it adds the product, but stays on the same page. 
6. Do the same form a product rendered natively to the Shop page. 
7. Notice it adds the product and redirects to the cart page as expected. 

**Expected behavior**
Expected to be redirected to the cart page, given the settings was checked.

**Screenshots**
n/a

* Core version: 5.3.2
* WooCommerce version: 3.9.2
* WooCommerce Blocks version: 2.5.11 (but not required for this to work)
* Site language: English
* Any other plugins installed: No

* OS: MacOS 10.15.2 (19C57)
* Browser Firefox 72.0.2

tomasztunik:
This fixes issue where All Products (and all other blocks) where Add to Cart button was present were ignoring the *Redirect to Cart after item added* setting was enabled.

We expose WooCommerce products settings in `wcSettings`

For the time being we expose only what is used by the blocks
which is `cartRedirectAfterAdd`. In the future more can be added
as needed. Setting is accessible via `getSetting( 'productsSettings' )`.

We namespace the settings under productsSettigns to reflect
the domain and how settings are organised in Woo admin and to
inform that this is an object with more settings within.

This setting normally was available **only** if AJAX add to cart was set
as a JS global `wc_add_to_cart_params.cart_redirect_after_add`.
By accessing the option directly we ensure it’s exposed
to blocks regardless of if AJAX option is enabled. 

This adds the redirect directly on the AddToCartButton after successful
add to cart action and it follows convention that redirects or other side
effects shouldn’t happen as part of the action but rather be part of the
control that triggers such flow.

Fixes #1766 

### Manual Testing

1. Go to WooCommerce Settings > Products > General and toggle *Redirect to the cart page after successful addition*
2. Add All Blocks and/or other products blocks. 
3. Hit the *Add to Cart* button
4. If setting is enabled you should be automatically redirected to the cart

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Changelog

> Add to Cart button on Products listing blocks will respect the "Redirect to the cart page after successful addition" setting

### DIFF
--- a/assets/js/atomic/blocks/product-elements/button/block.js
+++ b/assets/js/atomic/blocks/product-elements/button/block.js
@@ -9,6 +9,8 @@ import {
 	useStoreAddToCart,
 } from '@woocommerce/base-context/hooks';
 import { decodeEntities } from '@wordpress/html-entities';
+import { CART_URL } from '@woocommerce/block-settings';
+import { getSetting } from '@woocommerce/settings';
 import {
 	useInnerBlockLayoutContext,
 	useProductDataContext,
@@ -101,6 +103,12 @@ const AddToCartButton = ( { product } ) => {
 			dispatchStoreEvent( 'cart-add-item', {
 				product,
 			} );
+			// redirect to cart if the setting to redirect to the cart page
+			// on cart add item is enabled
+			const { cartRedirectAfterAdd } = getSetting( 'productsSettings' );
+			if ( cartRedirectAfterAdd ) {
+				window.location.href = CART_URL;
+			}
 		};
 	}
 
--- a/src/Assets/AssetDataRegistry.php
+++ b/src/Assets/AssetDataRegistry.php
@@ -89,6 +89,7 @@ class AssetDataRegistry {
 			'locale'             => $this->get_locale_data(),
 			'orderStatuses'      => $this->get_order_statuses(),
 			'placeholderImgSrc'  => wc_placeholder_img_src(),
+			'productsSettings'   => $this->get_products_settings(),
 			'siteTitle'          => get_bloginfo( 'name' ),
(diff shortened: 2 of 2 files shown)


tomasztunik:
I've explored hooking it up in `datra/cart/actions.ts` and looked into using `addAction` client hook but ultimately I felt that keeping this redirect side effect as part of the control that triggers it when setting is enabled is most explicit and ensures this redirect doesn't happen in unexpected situations or with controls where it wasn't meant to.

(And thanks @opr for the support and pointers!)

sunyatasattva (approved):
Great job. I love the well-thought-through PR description with all your thoughts and all that was done, and the comments in the code.

I'm just wondering… are we (or Gutenberg) not using any form of router? Redirecting via hard-coding the `window.location.href` _feels_ a bit crude.

But if we don't have any routing set up, then no problem!

tomasztunik:
Was curious about it as well when working on this @sunyatasattva. Woo is not a SPA and out of the box it doesn't have any routers or custom history handler contexts it seems. WooCommerce extension where it does redirect does it this way as well.

Issue #5647, opened 2022-01-27
Cart/Mini Cart buttons missing hover and focus styles


Aljullu:
### Cart block
1. Install the theme [Twenty Twenty Two](https://wordpress.org/themes/twentytwentytwo/) (you might need to update to WP 5.9 or install Gutenberg to get it working).
2. Go to the Cart page with the Cart block.
3. Hover/focus the _Proceed to Checkout_ button and notice there is no hover/focus style.

https://user-images.githubusercontent.com/3616980/151334365-69312873-5079-451c-839c-92461de4701f.mp4

### Mini Cart block
1. Install the theme Twenty Twenty Two (you might need to update to WP 5.9 or install Gutenberg to get it working).
2. Go to Apperance > Editor and add the Mini Cart block to the header.
3. Go to the frontend and open the Mini Cart.
4. Notice the View my cart and Checkout blocks don't have hover/focus styles.

https://user-images.githubusercontent.com/3616980/151334446-97059859-d5ba-46e5-848a-dab0fe69b5b6.mp4

## Expected behavior
All buttons should have hover/focus styles.

tarhi-saad:
Hi @Aljullu! Thank you for reporting this issue!

Indeed when following the steps provided above, we can see that the hover and focus styles are missing from the Cart & mini cart buttons. But, I think this isn't specific to the Twenty Twenty Two theme. I've tested the following themes:

- Twenty Twenty One
- Twenty Twenty
- Twenty Nineteen
- Blockbase (a non-core theme)

Each one of the above themes has the same styling issues. And the only one that has the correct style is the Storefront theme.

Could you share with me the themes you've tested where the focus & hover styles were showing correctly?

Aljullu:
> Could you share with me the themes you've tested where the focus & hover styles were showing correctly?

You're probably right. I only tested Storefront and Twenty Twenty Two. Hover/focus styles were present in Storefront, but they weren't in Twenty Twenty Two. I didn't test any other themes, so the problem might be broader than what the issue describes, sorry for the confusion!

PR #5712, opened 2022-02-02
Fix/cart button focus hover states

tarhi-saad:
This PR is intended to fix this issue #5647. The cart and mini-cart buttons didn't have hover & focus styles.

Our WC Block Button component is based on [Gutenberg's button](https://wordpress.github.io/gutenberg/?path=/story/components-button--buttons). The package name is [@wordpress/components](https://www.npmjs.com/package/@wordpress/components). This package's latest version displays a correct hover & focus style on the button component, but we are using an old version. We faced some issues with the bundle size of the newer versions because they include some libraries we don't need.

So, the solution is to set our focus and hover style for now. 

The latest Gutenberg's Button component introduced the `variant` attribute for "text", "contained" and "outlined" buttons. And since we are using an older version of this component that doesn't have this attribute, and we are already using the "contained" and "outlined" buttons in our plugin, implementing it on our end seems to be a good idea.

### Screenshots

Default state (filled button)
:-------------------------:
image 

Hover state (filled button)
:-------------------------:
image

Focus state (filled button)
:-------------------------:
image

Default state (outlined button)
:-------------------------:
image

Hover state (outlined button)
:-------------------------:
image

Focus state (outlined button)
:-------------------------:
image

### TODOS

- [X] Introduce the `outlined button` and replace the mini-cart custom styles

### Manual Testing

1. Checkout the branch
2. Create test pages for cart, mini cart, and checkout blocks.
3. In each page check the hover and focus styles for the `View my cart`, `Got to checkout`, `Proceed to Checkout`, and `Place Order` buttons.
4. You can do the same test using the available default themes and the Strofront theme as well.
### Changelog

> Fixed hover and focus states for button components


mikejolley:
@tarhi-saad Not sure if this is ready for review but if so, please convert from draft to PR and add the "Needs Review" label. Cheers!

tarhi-saad:
> @tarhi-saad Not sure if this is ready for review but if so, please convert from draft to PR and add the "Needs Review" label. Cheers!

Hi @mikejolley, this PR is now ready for review!

mikejolley (commented):
Hover states work like a charm, good job. I tested cart, checkout, mini-cart.

The only issue I spotted which might be nice to fix our side, the outlined buttons are ever so slightly taller due to the border:

Screenshot 2022-02-08 at 15 41 49

Can we maybe fix this with an inner shadow instead of a border, or some padding changes?

mikejolley (approved):
Thanks, looks good now.

PR #5717, opened 2022-02-04
Revert "fix: accessibility of search products block (#5691)"

tomasztunik:
This reverts commit cd5c5dedf731aa5a2d6938ecd60a4f55b9b55181.

Reasoning is it has minor impact but could cause blocks invalidation for already existing blocks as caught by E2E and missed in testing. As we have plans for bigger rewrite of this component in the future this change is not worth it. The underlaying issue it was addressing is partially covered as the currently used php rendering of the block has correct aria-label set and block should be accessible on the frontend.

At the same time I believe #1671 can remain closed as it was tested and confirmed that PHP rendering on the frontend does produce correct aria-lable as mentioned but this was fixed/correct before the change that #5691 introdcued.

Reverts #5691

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. E2E tests should pass and there should be no errors for product-search block

### User Facing Testing

Follow the steps from #5691 

### Changelog

> ⚠️ Removes changelog entry from #5691


gigitux (approved):
Thanks for working on this!
CI pass again. LGTM! :shipit:

PR #5722, opened 2022-02-04
E2E tests for Single Product Template

### LINKED ISSUE #5656: Store Editing Templates: Add E2E tests for Single Product block template
**Scenario to cover:**

* Should a user navigate to Site Editor > All Templates with an FSE theme active they should see the WooCommerce provided Single Product block template.
* Should a user load the Single Product  template, the editor will have the “WooCommerce Single Product Block” Legacy Template block present.
* Should a user load the product page it will render product details and allow the user to add the product to their basket.
* Should a user customise the Single Product template and save it. The user should see a "Clear customizations" option on the All Templates view (clearing customisations functionality should be covered by Gutenberg)
* Should a user load  the customised Single Product template the Site Editor should render these customisations
* Should a user have a customised Single Product template, these customisations would be present on the frontend.

sunyatasattva:
This PR adds E2E tests for the Single Product Template, and a bunch of test utils that we are going to be able to use in the next tests as well. It also adds an empty theme which supports block editing to the repo to be used with E2E tests.

Specifically regarding #5656, it covers the following test cases:

- [x] Should a user navigate to Site Editor > All Templates with an FSE theme active they should see the WooCommerce provided Single Product block template.
- [x] Should a user load the Single Product template, the editor will have the “WooCommerce Single Product Block” Legacy Template block present.
- [ ] Should a user load the product page it will render product details and allow the user to add the product to their basket. _(more details on why this wasn't added yet below)_
- [x] Should a user customise the Single Product template and save it. The user should see a "Clear customizations" option on the All Templates view (clearing customisations functionality should be covered by Gutenberg)
- [x] Should a user load the customised Single Product template the Site Editor should render these customisations
- [x] Should a user have a customised Single Product template, these customisations would be present on the frontend.

Closes #5656

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

1. Pull the latest `trunk` on the `gutenberg` monorepo, as this PR requires https://github.com/WordPress/gutenberg/pull/38463 to be there and the changes in `e2e-test-utils`, while merged on trunk, have not yet been published to npm.
2. Run `npm run test:e2e -- -i "tests/e2e/specs/backend/site-editing-templates.test.js"`
3. Assess that the included test cases pass.

### User Facing Testing
* [x] Same as above

---

@tjcafferkey As I was writing tests for this, I have been thinking more about [the conversation in #5656](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5656#issuecomment-1025506771), and I understood a bit more why it felt a bit out of place in this issue to test for the functionality mentioned in point 3.

Basically, I feel that this issue and PR should be more about the Single Product Template, while the adding to basket and so on, is more of a test of the Legacy Single Product Block.

After reading https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5346 which you mentioned, I am 100% convinced that it is important that we test certain functionality separately from Woo Core. So, I'm not saying we should add the test case 3 above, but I feel that in the interest of trying to keep PRs small and focused, and iterate fast, it would be best to exclude that from this PR and instead open a new issue with tests for the Legacy Block itself.

What are your thoughts?

If you disagree, I'm totally ok with adding those tests here, but they will definitely require much more work, and perhaps it would be best to spec them in a bit more detailed way and decide what to test within a legacy block.


tjcafferkey:
>So, I'm not saying we should add the test case 3 above, but I feel that in the interest of trying to keep PRs small and focused, and iterate fast, it would be best to exclude that from this PR and instead open a new issue with tests for the Legacy Block itself.

>What are your thoughts?

I think this is a good idea @sunyatasattva, thank you for proposing it. If it reduces the scope of this PR and can be done in a follow-up issue I think that is totally fine. At the moment we have zero tests, so if this means we get _some_ tests sooner with a view to iterate then I am all for that! 👍🏻

tests/e2e/utils.js
```diff
@@ -92,3 +125,140 @@ export const isBlockInsertedInWidgetsArea = async ( blockName ) => {
...
+}
+
+/**
+ * Waits for the Gutenberg canvas to be available
+ *
+ * @param {number} [timeout=DEFAULT_TIMEOUT] The amount of ms to wait for the element
+ */
+export async function waitForCanvas( timeout = DEFAULT_TIMEOUT ) {
+	await page.waitForSelector( SELECTORS.canvas, { timeout } );
+}
```
tomasztunik: Should we return the waitForSelector promise? it returns elementHandle so we could wait and act on it. Or do you want it to be explicitly about waiting only? :) 

I've noticed Gutenberg e2e tests could benefit from this. Maybe worth moving there? There are couple different ways of waiting for canvas present in Gutenberg e2e tests that could be unified if this one was added there. 

Some of the approaches/usages there that could be unified for reference but might be more :) 

```
    await page.waitForFunction( () =>
	Array.from( window.frames ).find(
            ( { name } ) => name === 'editor-canvas'
	)
    );
```
```
    await page.waitForSelector( 'iframe[name="editor-canvas"]' );
```
```
    await page.waitForSelector( '.edit-site-visual-editor', {
        timeout: 120000,
    } );
```

tests/e2e/utils.js
```diff
@@ -92,3 +125,140 @@ export const isBlockInsertedInWidgetsArea = async ( blockName ) => {
...
+/**
+ * Visits the Site Editor main page in Core WordPress
+ *
+ * There are two different possible site editor pages:
+ *
+ * 1. `themes.php?page=gutenberg-edit-site` is the one used and available if the Gutenberg plugin is enabled.
+ * 2. `site-editor.php` is the one available in WP Core.
+ *
+ * @param {string} query String to be serialized as query portion of URL.
+ * @param {'core' | 'gutenberg'} [editorContext='core'] Whether to go to the Gutenberg URL or the Core one.
```
tomasztunik, line 142: TIL - thanks for all the included insights!

tests/e2e/specs/backend/site-editing-templates.test.js
```diff
@@ -0,0 +1,180 @@
...
+};
+
+const SELECTORS = {
+	blocks: {
+		paragraph: blockSelector( 'core/paragraph' ),
+		singleProduct: legacyBlockSelector(
+			'WooCommerce Single Product Block'
+		),
+	},
+};
```
tomasztunik, line 58: Would it make sense for universal selectors and selector functions above to be part of utils?
We already seem to have some SELECTORS there.

I could imagine this could grow one day to earn its own file :)

tests/e2e/utils.js
tomasztunik: I was curious and had a look at the util on gutenberg side and the query params seems to be used as JSON object but I've noticed we are using it as url `search` string. 

There is a wrong typing information on the [visitSiteEditor](https://github.com/WordPress/gutenberg/blob/a4bf9b68b81f78b477b9476021f2ddffeddabbfd/packages/e2e-test-utils/src/site-editor.js#L200-L213) util. if we passed query as a string to visitSiteEditor it would fail or at least the query part can produce unexpected results, no? The [addQueryArgs](https://github.com/WordPress/gutenberg/blob/trunk/packages/url/src/add-query-args.js) used underneath seems to expect an object of query params but then gets mutated into a string with `addQueryArgs`

[visitAdminPage](https://github.com/WordPress/gutenberg/blob/a4bf9b68b81f78b477b9476021f2ddffeddabbfd/packages/e2e-test-utils/src/visit-admin-page.js#L20) does indeed expect a string. I think we should handle the query to ensure it's compatible with both paths if we want to make both of them available.

tests/e2e/utils.js
```diff
@@ -92,3 +125,140 @@ export const isBlockInsertedInWidgetsArea = async ( blockName ) => {
...
+ * Checks whether an element exists under a certain context
+ *
+ * @param {string} selector The selector for the desired element
+ * @param {Page | ElementHandle} [root=page] The root from which to search for the selector
+ *
+ * @return {Promise<boolean>} Whether the element exists or not
+ */
+export async function elementExists( selector, root = page ) {
+	return !! ( await root.$( selector ) );
+}
```
tomasztunik, line 164: would it make sense to split into "general" utils and user flow utils? Ie. repeated user actions related to particular domain.

I liked how you laid it out in Gutenberg repo 🎉  where different domains (ie. site-editor, particular block type etc.) would have their files.

tests/mocks/emptytheme/theme.json
```diff
@@ -0,0 +1,10 @@
+{
+	"version": 2,
+	"settings": {
+		"appearanceTools": true,
+		"layout": {
+			"contentSize": "840px",
+			"wideSize": "1100px"
+		}
+	}
+}
```
tomasztunik: there seems to be bunch of files with empty end line missing

tests/e2e/utils.js
sunyatasattva: > Should we return the waitForSelector promise? it returns elementHandle so we could wait and act on it. Or do you want it to be explicitly about waiting only? :)

I guess there is no harm in returning it. Maybe someone in the posterity would like it :)

> I've noticed Gutenberg e2e tests could benefit from this. Maybe worth moving there? There are couple different ways of waiting for canvas present in Gutenberg e2e tests that could be unified if this one was added there.

Alas in my discussion with Gutenberg test people, I've been discouraged from adding new utils there for now. We have a sort of agreement about adding new things when they are ripe and discussed.

So, for the scope of these series of PRs, for now I'm focusing on demonstrating their values here, and then perhaps continuing the discussion at a later point.

tests/e2e/utils.js
sunyatasattva: > There is a wrong typing information on the [visitSiteEditor](https://github.com/WordPress/gutenberg/blob/a4bf9b68b81f78b477b9476021f2ddffeddabbfd/packages/e2e-test-utils/src/site-editor.js#L200-L213) util. if we passed query as a string to visitSiteEditor it would fail or at least the query part can produce unexpected results, no? The [addQueryArgs](https://github.com/WordPress/gutenberg/blob/trunk/packages/url/src/add-query-args.js) used underneath seems to expect an object of query params but then gets mutated into a string with addQueryArgs

Good catch! I must have typed it early when I didn't use the query contructor util and then forgot to change it. The woes of not having TS!

> [visitAdminPage](https://github.com/WordPress/gutenberg/blob/a4bf9b68b81f78b477b9476021f2ddffeddabbfd/packages/e2e-test-utils/src/visit-admin-page.js#L20) does indeed expect a string. I think we should handle the query to ensure it's compatible with both paths if we want to make both of them available.

> I was curious and had a look at the util on gutenberg side and the query params seems to be used as JSON object but I've noticed we are using it as url search string.

Not sure I understand what you mean in those two, could you elaborate?

tests/e2e/specs/backend/site-editing-templates.test.js
sunyatasattva: In my opinion and experience with Puppeteer, it's not recommended to have one huge selectors object. In fact, the correct way would be to use the Page Object Pattern, which I am discussing with the Gutenberg fellas (they are actually also agreeing, it's just a matter of a huge refactor).

In this case, my intention is to separate selectors which are universally useful, and selectors which are specific to a test case/test file.

tests/mocks/emptytheme/theme.json
sunyatasattva: Thanks for catching that!

tests/e2e/utils.js
tomasztunik: good to know!

tests/e2e/utils.js
sunyatasattva: It does make sense, I agree. Will edit!

tests/e2e/utils.js
tomasztunik: Sorry placed the comment marker above implementation because I referred to the comment first. It's about the `goToSiteEditor` method right under and the error in typing I mentioned above.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/3963eabb3efdd1820f2f8c323d01c9a9b2ba2c8d/tests/e2e/utils.js#L139-L146

We accept query as string but visitSiteEditor if context is gutenberg expects object.

tests/e2e/specs/backend/site-editing-templates.test.js
tomasztunik: Had a quick look at  Page Object Pattern, looks really interesting. If you had a link to the discussions on this I'd love to learn more!

tomasztunik (approved):
👌

tests/e2e/page-utils.js
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Checks whether an element exists under a certain context
+ *
+ * @param {string} selector The selector for the desired element
+ * @param {Page | ElementHandle} [root=page] The root from which to search for the selector
+ *
+ * @return {Promise<boolean>} Whether the element exists or not
+ */
+export async function elementExists( selector, root = page ) {
```
ralucaStan, line 19: We could use instead of this the [toMatchElement](https://www.npmjs.com/package/expect-puppeteer#toMatchElement) utility that we already have via expect-puppeteer.

```
Expect an element be present in the page or element.
```

@sunyatasattva @tomasztunik

sunyatasattva: @ralucaStan Oh but they are not mutually exclusive. `toMatchElement` is an assertion, this is not supposed to be an assertion. It can be used for example as a guard clause. Or you can see how it is used [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/035838f750c89bc3045bd723e53ed61cac283de3/tests/e2e/utils.js#L200).

ralucaStan: I see, thank you for explaining it to me. The example helped. Maybe it would be good to add a few words about the purpose, so that we don't end up using it in places where [toMatchElement](https://www.npmjs.com/package/expect-puppeteer#toMatchElement) might be better. What do you think?

sunyatasattva: Yes! I agree, I didn't think about it, I'll add it to my notes.

PR #5735, opened 2022-02-07
Reformatted prettier files

alexflorisca:
A lot of people have been having problems with `prettier` errors on the pre-commit hook. I ran `npm run reformat-files` and it looks like a few files weren't formatted properly so I've committed the formatting changes here. 

It would  be useful if a few people could check this to make sure i doesn't cause any problems for anyone. 

To test: 
1. Checkout the branch
2. Run `npm install`
3. Make a commit anywhere in `js/assets` folder.
4. You should not get any prettier or es-lint errors.

### DIFF
--- a/assets/js/base/hooks/tsconfig.json
+++ b/assets/js/base/hooks/tsconfig.json
@@ -1,6 +1,6 @@
 {
-  "extends": "../../../../tsconfig.base.json",
-  "compilerOptions": {},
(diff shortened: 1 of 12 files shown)


tarhi-saad (commented):
Thank you, @alexflorisca, for reformatting the files. The test was successful, and I checked the reformatted files as well, and they look good too.👍

tarhi-saad (approved):

PR #5739, opened 2022-02-07
Mini Cart Block contents: Remove item border bottom

### LINKED ISSUE #5679: Remove dividing lines (Design)
Could we remove the horizontal dividing lines between products to align more with the designs and to keep the layout clean.

## Additional context
Just for clarity, here's an image indicating which line I'm referring to:

Screenshot 2022-01-31 at 14 42 55

danieldudzic:
Remove the item border-bottom, inside of the Mini Cart block contents.

I'm not a huge fan of using `!important` but didn't see a clean way to overwrite the core styles.

Fixes #5679 

### Screenshots

Before:
Products_–_kirigami

After:
Products_–_kirigami

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Active `Storefront` theme and add the Mini Cart Block as a widget.
2. Go on `/shop` page and add a product.
3. Confirm that the flyout mini cart items don't have a border

### User Facing Testing

* [x] Same as above

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/style.scss
@@ -179,7 +179,6 @@ table.wc-block-cart-items {
 			display: none;
 		}
 		.wc-block-cart-items__row {
-			@include with-translucent-border( 0 0 1px );
 			display: grid;
 			grid-template-columns: 80px 132px;
 			padding: $gap 0;
@@ -262,6 +261,18 @@ table.wc-block-cart-items {
 
 	.wc-block-cart__payment-options {
 		padding: $gap 0 0;
+	}
+}
+
+// Remove the border from cart items inside the Mini Cart drawer
+// https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5679
+.is-medium,
+.is-small,
+.is-mobile:not(.wc-block-mini-cart__drawer) {
+	table.wc-block-cart-items {
+		.wc-block-cart-items__row {
+			@include with-translucent-border( 0 0 1px );
+		}
 	}
 }
 
--- a/assets/js/blocks/cart-checkout/cart/style.scss
+++ b/assets/js/blocks/cart-checkout/cart/style.scss
@@ -178,6 +178,11 @@ table.wc-block-cart-items {
 		.wc-block-cart-item__remove-link {
 			display: none;
 		}
+		&:not(.wc-block-mini-cart-items) {
+			.wc-block-cart-items__row {
+				@include with-translucent-border( 0 0 1px );
+			}
+		}
 		.wc-block-cart-items__row {
 			display: grid;
 			grid-template-columns: 80px 132px;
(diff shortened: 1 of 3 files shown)


assets/js/blocks/cart-checkout/mini-cart/style.scss
```diff
@@ -100,16 +100,25 @@ h2.wc-block-mini-cart__title {
...
 	flex-grow: 1;
 	overflow-y: hidden;
-	padding: 0 $gap;
+	padding: $gap $gap 0;
 
 	.wc-block-mini-cart__products-table {
 		margin-bottom: auto;
 		margin-right: -$gap;
 		overflow-y: auto;
 		padding-right: $gap;
```
gigitux, line 110: Is there a particular reason for adding this style?

I just tried with this code and it seems to work

```suggestion
.wc-block-cart-items__row::after {
    border: none !important
}
```

gigitux (commented):
Thanks for working on this! I have just a question about the style that you added.

assets/js/blocks/cart-checkout/mini-cart/style.scss
danieldudzic: Thanks for the review!

The additional code adds a bit of space on top of the mini cart contents further improving it visually.

gigitux: Ok! It makes sense!

gigitux (approved):
LGTM! :shipit: 
NIT: I think that it is a good thing add a comment about why we are using `!important`

danieldudzic:
The initial fix was overwriting the default border (coming from the cart styles) using `!important`. To remove the bottom border from the mini cart items (without potentially breaking any future CSS overwrites) we have refactored the cart styles (and removed the `!important` overwrite). Can you please have a look?

gigitux:
> The initial fix was overwriting the default border (coming from the cart styles) using `!important`. To remove the bottom border from the mini cart items (without potentially breaking any future CSS overwrites) we have refactored the cart styles (and removed the `!important` overwrite). Can you please have a look?

Thanks for your refactor! LGTM! 

Let's wait if Rubik folks have feedback! 👍

assets/js/blocks/cart-checkout/cart/style.scss
```diff
@@ -265,6 +264,18 @@ table.wc-block-cart-items {
 	}
 }
 
+// Remove the border from cart items inside the Mini Cart drawer
+// https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5679
+.is-medium,
+.is-small,
+.is-mobile:not(.wc-block-mini-cart__drawer) {
```
ralucaStan: This solution works but it has the disadvantage of not using a targeted class for the Mini Cart (is-mobile is a class that is also used in other places), and it's also quite far from the line that it is overwriting. 

My recommendation is to add a class to the table used in the MiniCart so we can better target its rows.
- Add this under [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8475bb8c47499ad535aceb4391d9bafa30ec3228/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/mini-cart-products-table-block/block.tsx#L17) `className="wc-block-mini-cart-items"`
- Add the selected rule from the image below and remove the original line with the border
Screenshot 2022-02-14 at 18 25 32

Notice that this way you don't need the comments explaining what is happening.

```
		&:not(.wc-block-mini-cart-items) {
			.wc-block-cart-items__row {
			}
		}
```

ralucaStan (changes requested):
I've made a suggestion, let me know what you think.

ralucaStan (approved):
Works as expected. Ready to ship. 🚢

danieldudzic:
> Works as expected. Ready to ship. 🚢

Thank you for the suggestion (valid points) and the review!

PR #5741, opened 2022-02-08
Fix Mini Cart doesn't update data 

### LINKED ISSUE #5729: Mini Cart block and `shop` endpoint: data is not updated when adding the first product to the cart
## Describe the bug
If the user adds a product from ´/shop´ endpoint, the mini cart will not update. 
Similiar issue #5266 

## To reproduce

(with classic theme)
1. Active `Storefront` theme and add the Mini Cart Block as a widget.
2. Go on `/shop` page and add a product.
3. Check that the cart (`/cart/`), while the Mini Cart block is not updated.

(with block theme)
1. Active `TT1` theme and add the Mini Cart Block in the footer.
2. Go on `/shop` page and add a product.
3. Check that the cart (`/cart/`), while the Mini Cart block is not updated.

## Expected behavior
The Mini Cart should be updated.

## Screenshots

https://user-images.githubusercontent.com/4463174/152780949-a87ddade-4834-400c-b71f-6fd9d4384e8e.mp4

* Core version: 5.9

gigitux:
This PR removes API hydration for the Mini Cart block. This is because when the user opens the mini cart should have always the last data from the backend.

I noticed that removing the API hydration even if the Mini Cart block didn't open, the client did fetch requests to `/wc/store/cart`. This happened when I did some tests with `Single Product` and `All Product` blocks. Obviously, this is the wrong behavior because we wanted to prevent doing requests if we could avoid them.

I discovered that the provider `StoreNoticesProvider` triggers the fetch request to `/wc/store/cart` due to `useStoreEvent` hook that call `useStoreCart` hook (https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/51d7eaef4ed7f122698466249e47ab72ddeccc01/assets/js/base/context/providers/store-notices/context.js/#L66)

`StoreNoticesProvider` is used by `Single Product` and `All Product` blocks. So I added API hydration for these two blocks.

I guess that we should refactor `useStoreEvent` hook to avoid this problem for the next blocks.

Fixes #5729

### Manual Testing

Check out this branch

1. Active `StoreFront` theme and add `Mini Cart` block as a widget.
2. Create two pages/posts with `Single Product` block and the other one with `All Product` block. Save them.
3. Check that the cart is empty.
4. Go on `/shop` page and add a product.
5. Check that the Mini Cart updates correctly.
6. Empty the cart. 
7. Open the network tab from Dev Tools.
8. Go on the created pages/post and check that the browser doesn't fetch `/wc/store/cart`
9. Add a product for both pages and check that the Mini Cart updates correctly.

### DIFF
--- a/assets/js/base/context/providers/add-to-cart-form/form/submit/index.js
+++ b/assets/js/base/context/providers/add-to-cart-form/form/submit/index.js
@@ -13,6 +13,7 @@ import { useAddToCartFormContext } from '../../form-state';
 import { useValidationContext } from '../../../validation';
 import { useStoreCart } from '../../../../hooks/cart/use-store-cart';
 import { useStoreNotices } from '../../../../hooks/use-store-notices';
+import { triggerAddedToCartEvent } from '@woocommerce/base-utils';
 
 /**
  * FormSubmit.
@@ -107,6 +108,7 @@ const FormSubmit = () => {
 					} else {
 						receiveCart( response );
 					}
+					triggerAddedToCartEvent( { preserveCartData: true } );
 					dispatchActions.setAfterProcessing( response );
 					setIsSubmitting( false );
 				} );
--- a/assets/js/base/context/providers/add-to-cart-form/form/submit/index.js
+++ b/assets/js/base/context/providers/add-to-cart-form/form/submit/index.js
@@ -5,6 +5,7 @@ import { __ } from '@wordpress/i18n';
 import triggerFetch from '@wordpress/api-fetch';
 import { useEffect, useCallback, useState } from '@wordpress/element';
 import { decodeEntities } from '@wordpress/html-entities';
+import { triggerAddedToCartEvent } from '@woocommerce/base-utils';
 
 /**
  * Internal dependencies
@@ -13,7 +14,6 @@ import { useAddToCartFormContext } from '../../form-state';
 import { useValidationContext } from '../../../validation';
 import { useStoreCart } from '../../../../hooks/cart/use-store-cart';
 import { useStoreNotices } from '../../../../hooks/use-store-notices';
-import { triggerAddedToCartEvent } from '@woocommerce/base-utils';
 
 /**
  * FormSubmit.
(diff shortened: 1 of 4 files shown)


dinhtungdu (commented):
> I noticed that removing the API hydration even if the Mini Cart block didn't open, the client did fetch requests to /wc/store/cart.

Can you explain this? I don't totally get your point.

With this fix, the Mini Cart is working on Shop page, adding product to cart from the `All Products` block is working as well. But adding product from Single Product block doesn't work for me, both Storefront mini cart and the Mini Cart block didn't update.

gigitux:
> > I noticed that removing the API hydration even if the Mini Cart block didn't open, the client did fetch requests to /wc/store/cart.
> 
> Can you explain this? I don't totally get your point.
> 
> With this fix, the Mini Cart is working on Shop page, adding product to cart from the `All Products` block is working as well. But adding product from Single Product block doesn't work for me, both Storefront mini cart and the Mini Cart block didn't update.

First of all thanks for testing the PR!

If you start from `trunk` branch and remove API hydration for the Mini Cart Block, you will see that when you load `Single Product` and `All Product` blocks, the browser does a fetch request to `/wc/store/cart`.

This is wrong because we want to fetch that endpoint only when the Mini Cart Block is opened. (it doesn't make sense that these two blocks fetch that endpoint).

This is caused by `StoreNoticesProvider`. I hope that now it is more clear.


> But adding product from Single Product block doesn't work for me

This is another problem. Now the problem is that the `Add to Cart Button` block didn't send the event `wc-blocks_added_to_cart` and so the javascript Mini Cart Block is never loaded. I fixed this in 6860a4149940ccdccf9717d86fc9337feef49ca8

dinhtungdu (approved):
Thanks for the update and explaination!

There is a minor issue with the `Single Product` block and the Storefront Mini Cart. When you add a product to cart from the Single Product block, the Mini Cart drawer is open and updated, but the Storefront mini cart doesn't update. However, I think we can create another issue then fix that issue later. The current PR is LGTM!

gigitux:
@dinhtungdu Thanks for the review! In the next hours, I will check the issue that you bring up and I will open an issue 👍 

EDIT: Created! #5812

PR #5744, opened 2022-02-08
Set focus on Empty Mini Cart Contents Block

### LINKED ISSUE #4618: Mini Cart block: after removing a product, drawer can't be closed clicking outside

gigitux:
This PR set the focus on the Empty Mini Cart Contents element: in this way when the user clicks outside the drawer, this latter will close.

Without this PR when the Empty Mini Cart Contents Block is rendered after removing the last product, the active element is the body. This causes that [these callbacks](https://github.com/WordPress/gutenberg/blob/c86a198597fa6d237699b7883bd9de8d1663bbd0/packages/components/src/modal/index.js#L74) aren't being called.

Fixes #4618

### Manual Testing

Check out this branch

2. Add a Product to the cart.
3. Visit that post or page in the front and click on the Mini Cart button.
4. Remove the product.
5. Click on the overlay and check that the drawer will close.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
@@ -2,6 +2,7 @@
  * External dependencies
  */
 import { useStoreCart } from '@woocommerce/base-context/hooks';
+import { useEffect, useRef } from 'react';
 
 /**
  * Internal dependencies
@@ -16,12 +17,22 @@ const EmptyMiniCartContentsBlock = ( {
 }: EmptyMiniCartContentsBlockProps ): JSX.Element | null => {
 	const { cartItems, cartIsLoading } = useStoreCart();
 
+	const elementRef = useRef< HTMLDivElement >( null );
+
+	useEffect( () => {
+		elementRef.current?.focus();
+	} );
+
 	if ( cartIsLoading || cartItems.length > 0 ) {
 		return null;
 	}
 
 	return (
-		<div className="wp-block-woocommerce-empty-mini-cart-contents-block">
+		<div
+			tabIndex={ 0 }
+			ref={ elementRef }
+			className="wp-block-woocommerce-empty-mini-cart-contents-block"
+		>
 			{ children }
 		</div>
 	);
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
@@ -20,8 +20,10 @@ const EmptyMiniCartContentsBlock = ( {
 	const elementRef = useRef< HTMLDivElement >( null );
 
 	useEffect( () => {
-		elementRef.current?.focus();
-	} );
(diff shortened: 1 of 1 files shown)


dinhtungdu (dismissed):
Tested and this is working great for me!

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
```diff
@@ -16,12 +17,22 @@ const EmptyMiniCartContentsBlock = ( {
 }: EmptyMiniCartContentsBlockProps ): JSX.Element | null => {
 	const { cartItems, cartIsLoading } = useStoreCart();
 
+	const elementRef = useRef< HTMLDivElement >( null );
+
+	useEffect( () => {
+		elementRef.current?.focus();
+	} );
```
tomasztunik: This would trigger it on every re-render/update of the component and we want to focus it only on initial mount/render right? Then empty dependency array is required.

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
```diff
@@ -16,12 +17,22 @@ const EmptyMiniCartContentsBlock = ( {
...
+
 	if ( cartIsLoading || cartItems.length > 0 ) {
 		return null;
 	}
 
 	return (
-		<div className="wp-block-woocommerce-empty-mini-cart-contents-block">
+		<div
+			tabIndex={ 0 }
```
tomasztunik: We should consider using `-1` here - It allows focusing with JavaScript but wouldn't allow to focus using TAB and would not interrupt tabbing flow for accessibility.

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
tomasztunik: or if it's based on if cart is loaded then we should add loaded state as dependency and focus when loaded and ref is set :)

gigitux: > This would trigger it on every re-render/update of the component and we want to focus it only on initial mount/render right? Then empty dependency array is required.

Great catch! Thanks for bringing this up!

tomasztunik (approved):
Thanks for the tweaks! Looks good 🎉

PR #5745, opened 2022-02-08
Disable quantity selector and remove item link if item is a synced force sell

alexflorisca:
This solves an [integration issue with the WooCommerce Force Sells plugin](57-gh-woocommerce/woocommerce-force-sells). When a product is forced to sync to another product, the user should not be able to change the quantity or remove the linked product. This PR disables those options for a sync forced product. 

Fixes  57-gh-woocommerce/woocommerce-force-sells

### Screenshots

| Before | After |
| ----- | ------ |
| Screenshot 2022-02-08 at 17 22 23 | Screenshot 2022-02-08 at 17 21 54 |

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Add the [WooCommerce Blocks plugin](https://wordpress.org/plugins/woo-gutenberg-products-block/).
2. Create a test page with the cart block.
3. Go to `product A` and add `product B` as `Synced Force Sells` product.
4. Go to the frontend and add `product A` to the cart.
5. Go to the test page with the cart block.
6. You should not be able to adjust the quantity or remove `product B`

### User Facing Testing
* [x] Same as above

### Changelog

> Fix an integration issue with WooCommerce Force Sells plugin where a user could modify a sync forced product in the Cart block.

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/cart-line-items-table/cart-line-item-row.tsx
+++ b/assets/js/blocks/cart-checkout/cart/cart-line-items-table/cart-line-item-row.tsx
@@ -200,6 +200,12 @@ const CartLineItemRow = forwardRef< HTMLTableRowElement, CartLineItemRowProps >(
 			validation: productPriceValidation,
 		} );
 
+		// If this item is set as a Sync Force Sell (by the Woocommerce Force Sells plugin).
+		// we want to disable the quantity selector and remove item link
+		const isSyncForcedItem = !! itemData.find(
+			( item ) => item?.name === 'Linked to'
+		);
+
 		return (
 			<tr
 				className={ classnames(
@@ -288,7 +294,9 @@ const CartLineItemRow = forwardRef< HTMLTableRowElement, CartLineItemRowProps >(
 							{ ! soldIndividually &&
 								!! quantityLimits.editable && (
 									<QuantitySelector
-										disabled={ isPendingDelete }
+										disabled={
+											isPendingDelete || isSyncForcedItem
+										}
 										quantity={ quantity }
 										minimum={ quantityLimits.minimum }
 										maximum={ quantityLimits.maximum }
@@ -326,7 +334,7 @@ const CartLineItemRow = forwardRef< HTMLTableRowElement, CartLineItemRowProps >(
 										)
 									);
 								} }
-								disabled={ isPendingDelete }
+								disabled={ isPendingDelete || isSyncForcedItem }
 							>
 								{ __(
 									'Remove item',


manospsyx:
Might be a good idea to address these issues in an integration on the Force Sells side.

In addition to helping keep concerns separated, doing so will also help surface extensibility needs, which need to be served using publicly accessible APIs.

alexflorisca:
Yeah you might be right there, it's not the neatest solution. I couldn't see a solution on the Force Sells side so I'd love to hear your thoughts on how we can make that work

manospsyx:
Hey @alexflorisca 👋 

Check out our work [here](https://somewherewarmattic.wordpress.com/2022/02/01/product-bundles-cart-and-checkout-blocks-integration-phase-2/). One of the extensibility themes we have identified as part of that work is:

> Cart/Checkout Item Grouping and Syncing

Force Sells relies primarily on the second part (the syncing side of things), but there are many similarities between the two products (Force Sells / Product Bundles) that make it possible to borrow a lot [from here](https://github.com/woocommerce/woocommerce-product-bundles/compare/master...issue-758) (WIP).

I'd be happy to help with integrating Force Sells once we're done with the integrations lined up ahead of us now! The 3 integrations we have prioritized (Bundles, Gift Cards, Conditional Shipping and Payments) have a very large surface area that can help inform other integrations (both internal and by 3PDs).

alexflorisca:
Nice one, I'll check out your work soon and make a decision on where to take this PR. Thanks for sharing!

opr (commented):
Hey Alex, thanks for working on this, seems like a simple fix, but I don't think we should be adding plugin-specific code to our repository, instead we should create extensibility points to allow other plugins to decide the behaviour.

I think we should maybe look into adding a filter like `allowCartItemQuantityChange` or something, or `allowCartItemRemoval`. What do you think?

ralucaStan:
I agree this should be fixed on the plugin's side. There is an open [PR for Box office]( 328-gh-woocommerce/woocommerce-box-office/files#diff-dead2a50f34c2dd942a56ac83114b942dd9d07941f386ffe63fac95fa8d67d4fR24) fixing a similar issue, where the quantity selector is disabled using the `woocommerce_store_api_product_quantity_editable` PHP filter.

This filter was introduced by [this PR ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5406) and it could also be used by WC Force Sells

alexflorisca:
Thanks for all your points everyone - nice work on the integration work @manospsyx, look forward to diving deeper into that and applying those principles here. I'll close this PR and will open a new PR on the Force Sells side with the integration work. At the very least, it looks like we can disable the quantity selectors, even if we might not be able to remove the "Remove product" button

PR #5801, opened 2022-02-09
Fix wide appender buttons overlap in Cart & Checkout blocks in the Editor

### LINKED ISSUE #5742: "Add Block" button overlaps content on the Checkout block in the editor
## Describe the bug
The "Add Block" button overlaps the other inner blocks on the checkout block (see screenshot)

## Screenshots
| Place Order | Totals |
| ------------ | ------ |
| Screenshot 2022-02-08 at 11 16 59 | Screenshot 2022-02-08 at 11 16 51 |

## To reproduce

1. Open up a page with the Checkout block in the editor
2. Confirm that the "Place Order" button and the "Add Block" button at the bottom of the block overlap
3. Confirm that in the right hand side bar, the total price overlaps with the "Add block" button

## Expected behaviour
The "Add block" button should sit neatly underneath the "Place order" button. 

* Core version: 5.9
* Gutenberg version (if installed): 12.5.3
* WooCommerce version: 6.3.0-dev
* WooCommerce Blocks version:  7.0.0-dev

ralucaStan:
Since this [Gutenberg change](https://github.com/WordPress/gutenberg/pull/36605) all appender buttons have  `position:absolute` resulting in layout changes for the wide appender button.

Initially I've tried to remove the renderAppender prop from places like [here ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/446fade219c3f31058947d7560a203f834446dd4/assets/js/blocks/checkout/inner-blocks/checkout-totals-block/edit.tsx#L60) so that the default black Add block button should be shown instead of the white one with black borders. That worked, but the button was visible only when the block was selected. We needed the appender buttons to be visible all the time.

I've decided to fix this in place via CSS, so that we have the old buttons working as before.

I’m doing some research, and it looks like [the GB change ](https://github.com/WordPress/gutenberg/pull/36605)made people unhappy, as it ruined their layouts as well. I’ve also added a [comment](https://github.com/WordPress/gutenberg/pull/36605), I hope they will only make the small black buttons absolute and let the others as before so that we can revert this change.

Fixes #5742 

### Screenshots

| Before | After |
| ------------ | ------ |
| Screenshot 2022-02-08 at 11 16 59 | Screenshot 2022-02-09 at 16 23 02 |
|Screenshot 2022-02-08 at 11 16 51 | Screenshot 2022-02-09 at 16 23 37 |

### Manual Testing

1. Open up a page with the Checkout block in the editor
2. Confirm that the "Place Order" button and the "Add Block" button at the bottom of the block don't overlap
3. Confirm that in the right hand side bar, the total price doesn't overlaps with the "Add block" button
4. Also check the Cart block that the insert inner block buttons don't overlap

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Changelog

> Fix wide appender buttons overlap in Cart & Checkout blocks in the Editor

### DIFF
--- a/assets/js/blocks/cart-checkout/cart/editor.scss
+++ b/assets/js/blocks/cart-checkout/cart/editor.scss
@@ -20,3 +20,12 @@ body.wc-lock-selected-block--remove {
 	}
 }
 
+.wp-block-woocommerce-cart-items-block,
+.wp-block-woocommerce-cart-totals-block,
+.wp-block-woocommerce-empty-cart-block {
+	// Temporary fix after the appender button was positioned absolute
+	// See https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5742#issuecomment-1032804168
+	.block-list-appender {
+		position: relative;
+	}
+}
--- a/assets/js/blocks/cart-checkout/cart/inner-blocks/cart-items-block/edit.tsx
+++ b/assets/js/blocks/cart-checkout/cart/inner-blocks/cart-items-block/edit.tsx
@@ -28,7 +28,6 @@ export const Edit = ( { clientId }: { clientId: string } ): JSX.Element => {
 				allowedBlocks={ allowedBlocks }
 				template={ defaultTemplate }
 				templateLock={ false }
-				renderAppender={ InnerBlocks.ButtonBlockAppender }
 			/>
 		</Main>
 	);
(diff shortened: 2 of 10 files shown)


ralucaStan:
@gigitux I've also assigned you as this was affecting the Mini Cart as well and I've applied the fix there as well. It would be great if you could also take a look.

alexflorisca (approved):
Look good, thanks @ralucaStan, let's keep an eye out on Gutenberg and see if they revert the absolute positioning in the future.

gigitux (approved):
Thanks for fixing Mini Cart too! 

LGTM :shipit:

PR #5802, opened 2022-02-09
Remove `useSelectShippingRate` hook and adjust how local state works for `PackageRates`

### LINKED ISSUE #5734: Selected shipping rate shouldn't have local state
Currently, selected shipping rate is persisted in a local state

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/cbbd9d7dcc30a2c4224aa83e62486501658a4050/assets/js/base/context/hooks/shipping/use-select-shipping-rate.ts#L49-L53

While the source of truth lives in `wc/store/cart`, this means that in the time between selecting a shipping rate, and the server responding, you get different results. Anyone reacting to this (via the `set-selected-shipping-rate` event for example) will get a stale value.

While not a huge issue by its own (the current value is still passed to `set-selected-shipping-rate`), having a local state and state living in data store is starting to be an anti-pattern and causing us a lot of boilerplate, so I'd we move this state back to the store.

mikejolley:
While looking at #5734 I've established that we **do** need local state, but we can make some improvements so it's more clear as to why.

If the radio buttons are controlled inputs, without a local state you click on shipping options and get no UI feedback until the request completes and the shipping methods update. This is poor UX.

if the radio buttons are uncontrolled inputs, you get instant feedback when selecting a shipping option, but if the shipping method changes elsewhere (on the server, or on another tab) it becomes out of sync with the data stores.

Therefore, the best solution is to store the checked option in local state, and update it both on click, and when the data stores update post-update.

You can see in this PR I was able to remove `useSelectShippingRate`, remove it's local state, and move the logic to `useSelectShippingRates`.

I've moved the Radio button local state into `PackageRates`. There it just stores the value of the selected option. This also reacts to updates from the server using `useEffect`.

`RadioControl` has also been changed to TypeScript.

Closes #5734

### Testing

1. Ensure your store has multiple shipping methods setup
2. Add something to your cart that requires shipping
2. Go to the cart block
3. Select a shipping option. Confirm the UI updates instantly, and the option remains checked once totals are updated
4. Open 2 tabs with a cart block in each
5. In tab 1, select a shipping method
6. In tab 2, change an item qty. Confirm that once the update completes, the correct shipping method is selected.


assets/js/base/components/radio-control/types.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+import type { ReactElement } from 'react';
+
+export interface RadioControlProps {
+	// Class name for control.
+	className?: string;
+	// ID for the control.
+	id?: string;
+	// Selected should be passed if using this as a controlled component.
+	selected: string;
```
opr: Should this be optional?

assets/js/base/components/radio-control/option.tsx
```diff
@@ -7,16 +7,24 @@ import classnames from 'classnames';
  * Internal dependencies
  */
 import OptionLayout from './option-layout';
+import type { RadioControlOptionProps } from './types';
 
-const Option = ( { checked, name, onChange, option } ) => {
+const Option = ( {
+	checked,
+	defaultChecked,
```
opr: Looks like this is missing from the `RadioControlOptionProps` type.

assets/js/base/context/hooks/shipping/use-select-shipping-rates.ts
```diff
@@ -49,7 +64,7 @@ export const useSelectShippingRates = (): {
 	}, [] );
 
 	return {
-		selectShippingRate: setRate,
+		selectShippingRate,
```
opr: I'm just wondering if `useSelectShippingRate` (no s) would be a better name for this hook, I just felt a bit weird about this line elsewhere in the code:

```ts
const { selectShippingRate } = useSelectShippingRates();
``` 

I think 
```ts
const { selectShippingRate } = useSelectShippingRate();
```
might look better. My basis for this is the action the user is taking when they change the radio button is they are "selecting a shipping rate". 

Feel free to ignore this if you don't agree.

opr (approved):
This works well, thanks for taking care of this Mike! I left a few TS comments and one about naming, but those are non-blocking.

mikejolley:
There was an idea to use loading states from @tarhi-saad which could also work, but I think we should avoid this for now until https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5365 has had a proper review. I don't think we should introduce more loading states until existing have been reviewed.

assets/js/base/components/radio-control/types.ts
mikejolley: No, I messed up the comments when I moved back to controlled. I've updated the inline comments.

assets/js/base/context/hooks/shipping/use-select-shipping-rates.ts
mikejolley: Done, I do agree it makes more sense now.

PR #5816, opened 2022-02-11
Add Global Styles support for the Add To Cart Button block #5666

### LINKED ISSUE #5666: Migrate `Add to cart button` block to global style API
In accordance with the documentation, the `Add to cart button` block should support:
- text color,
- background color (including gradients), 
- padding, 
- border (color and radius) and width

gigitux:
This PR adds support for global style for the `Add To Cart Button` block.

- text-color
- background-color
- border-radius
- font-size

#4965 

### Screenshots

image
*without global style*

image
*with global style*

### Manual Testing

1. Be sure that you are on `WordPress 5.9`.
3. Add the `All Product` block (this block contains `Add To Cart Button Block`) to a post.
4. Get the focus on `Add To Cart Button`.
8. Go to Dashboard and select Appearance > Editor (beta). At the top of the screen, select Home > Browser all templates > Single Post. When the page is loaded, add the block to the page.
10. Verify that the `Add To Cart Button` is shown under the `Blocks` section. Personalize again the block.
11. Save your changes.
14. Change again the styles.
15. Save your changes.

### Changelog

> Add Global Styles support to the Add To Cart Button block.


assets/js/atomic/blocks/product-elements/button/block.js
```diff
@@ -116,14 +151,22 @@ const AddToCartButton = ( { product } ) => {
 		<ButtonTag
 			aria-label={ buttonAriaLabel }
 			className={ classnames(
+				'wc-block-components-product-button__button--global-style',
```
dinhtungdu: Why do we need another class here? Scanning this PR, I see it's for Global Style. Is there any issue with the existing class (`wc-block-components-product-button__button`)?

dinhtungdu (commented):
> The problem is that this block uses padding that comes from the wp-button-link class: if we override padding property we lose the default padding value. In this case, the style of the button would be not great.

Can you explain this further? I tested adding padding to the block and didn't see issues related to `.wp-button-link` style.

gigitux:
> > The problem is that this block uses padding that comes from the wp-button-link class: if we override padding property we lose the default padding value. In this case, the style of the button would be not great.
> 
> Can you explain this further? I tested adding padding to the block and didn't see issues related to `.wp-button-link` style.

Did you try with the `Global Styles`? 
image
 
With the current implementation, the current style will have always greater priority respect than the Global Style settings.
 
 If I remove the padding, the user can see the padding that inserts, but at this point, the default padding will be 0 and the button doesn't look great.

I wasn't able to find a good workaround, but maybe you have another idea to fix this
 
 
 
 Unfortunately, I noticed that we have the same problem with the font size :'(

assets/js/atomic/blocks/product-elements/button/block.js
gigitux: I was convinced that this resolved the problem about the font size :(

I did another cycle of tests and I noticed that this is not true :/ 
I removed this new class and thanks for bringing up this!

dinhtungdu (commented):
> With the current implementation, the current style will have always greater priority respect than the Global Style settings.

@gigitux https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5843 is my workaround to fix the style priority, can you please take a look?

gigitux:
> > With the current implementation, the current style will have always greater priority respect than the Global Style settings.
> 
> @gigitux #5843 is my workaround to fix the style priority, can you please take a look?

Thanks for your help! 🙏

assets/js/atomic/blocks/product-elements/button/supports.ts
```diff
@@ -11,15 +11,10 @@ export const supports = {
 			link: false,
 			__experimentalSkipSerialization: true,
 		},
-		typography: {
-			fontSize: true,
-			__experimentalSkipSerialization: true,
-		},
```
dinhtungdu: @gigitux Should we add the typo support back?

dinhtungdu (commented):
The font-weight of the Button is different between the Blocks Editor and the front end. Otherwise, this is LGTM

image

gigitux:
Fixed with https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5816/commits/abcd96a31ee72c618a8379344e5395f4dfdca16d!

image
*frontend side*

image
*editor side*

dinhtungdu (commented):
@gigitux Because we added font-size support, what do you think about adding font-weight support as well? Using 400 as the font weight for the add to cart button is a valid use case to me.

gigitux:
> @gigitux Because we added font-size support, what do you think about adding font-weight support as well? Using 400 as the font weight for the add to cart button is a valid use case to me.

I added the font-weight support. 
I added [this line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5816/files#diff-95a248c54e90940a2c3a4595a96d039f8a70d8fb593258000c17c3e658cd2ea5R10) because I noticed that the button element used the Arial font.

## Before my changes 

image
*button element (third button)*
image
*anchor element (two first buttons)*

assets/js/atomic/blocks/product-elements/button/supports.ts
```diff
@@ -0,0 +1,30 @@
...
+			padding: true,
+			__experimentalSkipSerialization: true,
+		},
+		typography: {
+			fontSize: true,
+			__experimentalFontWeight: true,
+			__experimentalSkipSerialization: true,
+		},
+		__experimentalSelector:
+			'.wp-block-button.wc-block-components-product-button .wc-block-components-product-button__button',
```
dinhtungdu, line 28: Just FYI, on the server side, Global Style only cares about `__experimentalSelector` to generate CSS.

dinhtungdu (approved):
LGTM 🎉  :shipit:

Issue #5711, opened 2022-02-02
Cart & Checkout “I3”: decoupling data flows - Epic issue


opr:
This epic will track the progress of investigating whether each context we're using for the Cart and Checkout blocks can be converted to a data store

PR #5817, opened 2022-02-11
Customer Data Store and revised handling for Shipping As Billing

### LINKED ISSUE #5737: `useCheckoutAddress` causes multiple updates to `setBillingData` when the 'Use same address for billing' checkbox is changed
When checking the `Use same address for billing` checkbox, the effect here: 

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/7a8daa3eed1f1f095d1c142bd46432cff64990b0/assets/js/base/context/hooks/use-checkout-address.js#L62-L82

runs multiple times. This is caused by multiple subscribers to the hook, causing the effect to be registered multiple times.

This is going to cause unnecessary actions to be dispatched to the store if we switch to using `@wordpress/data` stores instead of context. At present the billing data on the context is being updated multiple times already.

To resolve this we could potentially move this effect to a new hook that is consumed only in one place.

### LINKED ISSUE #5747: Find a way to keep the address up to date in CheckoutShippingAddress and CheckoutBillingAddress
Check out the `try/decouple-customer-data-context` branch and do the following:

1. Uncheck `Use same address for billing`.
2. Change postcode on either address.
3.  While the block is submitting the new data to the server, change the postcode again (you have to be quick!).
4. Observe the postcode change to the old one, then back to the new one.

The reason this is happening is the `wc/data/store` is the new source of truth for the address data. The data flow works like so:

1. User updates address in form, the `wc/store/cart` is updated to reflect the new address information.
2. If necessary, `UPDATE_CUSTOMER_DATA` is dispatched which sends the new address info to the server.
2a. Making further updates to the address form at this point causes step 2 to happen again.
3. When this completes, the new cart is returned by the server. We then dispatch `RECEIVE_CART` onto the store (remember, this contains the address data we sent along with `UPDATE_CUSTOMER_DATA`).
4. The `wc/store/cart` store is restored to the same data sent in step 2.

We need to find a way to allow the server to be updated, but prevent the returned value overwriting the data in the store if it is fresher.

### LINKED ISSUE #5803: Remove local state from `useCheckoutAddress` related to `Use same address for billing` checkbox
In this comment, pca54o-2Zx-p2#comment-3612 we discuss the current behaviour of the `Use same address for billing` checkbox. It uses local state so that values entered into the billing can be restored if the user toggles the checkbox. We can simplify the logic here and change the behaviour a little to make things easier.

New behaviour should be:

1. The `Use same address for billing` checkbox is unchecked, both the shipping and billing forms are visible.
2. The user enters two different addresses in the forms.
3. The user checks the checkbox.
4. Behind the scenes `billingAddress` gets set to the same value as `shippingAddress`.
5. The user unchecks the checkbox again.
6. The Billing form is now the same as the shipping form.

mikejolley:
This is a combination of:

- https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5810
- https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5809

Part of #5711
Closes #5810
Closes #5809
Closes #5737 
Closes #5747 
Closes #5803

This PR:

- Moves _shipping as billing_ to the checkout state provider rather than context and moves all customer data from context to WP data stores.
- Changes the behaviour of _shipping as billing_ so it no longer relies on local state. See the testing steps below.
- Adds customer data to the cart data store. This is used as the source of truth in existing context providers.

### Testing

The following functionality requires smoke testing:

1. Checking out - edit shipping and billing sections, confirm things persist, confirm orders place with correct data
2. Paying via express payment methods. Confirm they can still set and update addresses.
3. Confirm updates to the server are debounced (500ms intervals) when editing address data

**User testing**

1. In a new window logged out, add something to cart and go to checkout, open the network tab of your dev tools.
2. Enter a shipping address - watch for XHR requests to the batch endpoint. You should only get one request one second after you last made changes. If you make multiple changes all within one second of each other, there should still only be one request. Ensure multiple requests for the same change don't appear.
3. Uncheck "use shipping as billing"
4. Enter a different billing address -
5. Place order
6. Add another item to cart and checkout again
7. "use shipping as billing" should be unchecked and you should see the different shipping and billing addresses
8. Check "use shipping as billing"
9. Uncheck "use shipping as billing"
10. Billing should be the same as shipping (previous values removed)

After testing, check the orders in WordPress admin and confirm the correct addresses were persisted.

### Changelog

> Moved customer data from context provider into wp data store.
> Toggling the "use shipping as billing" checkbox will now copy values from shipping rather than preserve a past entered billing address.


opr (approved):
My review is pretty biased as just under half the commits were mine and I worked on this as well, but it's working well for me, I've been unable to reproduce the double request bug since pulling the latest changes.

Express payments work OK, and the code changes look clean, but let's get a couple more reviewers on this.

mikejolley:
@ralucaStan this is ready for review now. It's a combination of 2 PRs. Testing instructions above. Thanks!

assets/js/base/context/hooks/use-checkout-address.ts
```diff
@@ -0,0 +1,88 @@
...
+
+interface CheckoutAddress {
+	shippingAddress: ShippingAddress;
+	billingData: BillingAddress;
+	setShippingAddress: ( data: Partial< EnteredAddress > ) => void;
+	setBillingData: ( data: Partial< EnteredAddress > ) => void;
+	setEmail: ( value: string ) => void;
+	setPhone: ( value: string ) => void;
+	setShippingPhone: ( value: string ) => void;
+	shippingAsBilling: boolean;
```
ralucaStan: Unless I am missing something, I think that the technically correct name for this setting is `billingAsShipping`

ralucaStan:
There is a change to `package-lock.json` but no change to `package.json`

assets/js/base/context/providers/cart-checkout/checkout-state/types.ts
```diff
@@ -104,6 +107,8 @@ export type CheckoutStateContextType = {
 	orderNotes: CheckoutStateContextState[ 'orderNotes' ];
 	// This is the ID of the customer the draft order belongs to.
 	customerId: CheckoutStateContextState[ 'customerId' ];
+	// Should the billing form be hidden and inherit the shipping address?
+	shippingAsBilling: CheckoutStateContextState[ 'shippingAsBilling' ];
```
ralucaStan: Another case to rename the variable to `billingAsShipping`

assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
```diff
@@ -68,19 +79,26 @@ const Block = ( {
...
+						) as ( keyof AddressFields )[]
+					}
 					fieldConfig={ addressFieldsConfig }
 				/>
 				{ showPhoneField && (
 					<PhoneNumber
 						id="shipping-phone"
 						isRequired={ requirePhoneField }
-						value={ shippingFields.phone }
+						value={ shippingAddress?.phone || '' }
```
ralucaStan: Wouldn't it be better to use `defaultShippingAddress.phone` here?

assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
```diff
@@ -97,9 +115,14 @@ const Block = ( {
...
 				) }
 				checked={ shippingAsBilling }
-				onChange={ ( checked: boolean ) =>
-					setShippingAsBilling( checked )
-				}
+				onChange={ ( checked: boolean ) => {
+					setShippingAsBilling( checked );
+					if ( checked ) {
+						setBillingData(
+							( shippingAddress || {} ) as BillingAddress
```
ralucaStan: Shouldn't we pass `defaultShippingAddress` instead of {}?

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
```diff
@@ -152,6 +153,17 @@ export const reducer = (
 					  }
 					: state;
 			break;
+		case ACTION.SET_SHIPPING_ADDRESS_AS_BILLING_ADDRESS:
+			if (
+				shippingAsBilling !== undefined &&
```
ralucaStan: Can shippingAsBilling  be undefined? The type definition only states `shippingAsBilling: boolean;`?

mikejolley:
@ralucaStan I just ran `nvm use && npm install` and it didn't change. Maybe trunk is wrong?

assets/js/base/context/hooks/use-checkout-address.ts
```diff
@@ -0,0 +1,88 @@
...
+
+	const setEmail = useCallback(
+		( value ) =>
+			void setBillingData( {
+				email: value,
+			} ),
+		[ setBillingData ]
+	);
+
+	const setPhone = useCallback(
```
ralucaStan: maybe for more clarity, as we have `setShippingPhone` bellow?
```suggestion
	const setBillingPhone = useCallback(
```

assets/js/data/cart/actions.ts
```diff
@@ -41,6 +42,26 @@ export const receiveCart = (
 	};
 };
 
+/**
+ * Returns an action object used in updating the store with the provided cart. This omits the customer addresses.
```
ralucaStan: Maybe a few words why the addresses are omitted would be beneficial.

assets/js/data/cart/push-changes.ts
```diff
@@ -0,0 +1,149 @@
...
+	shippingAddress: false,
+};
+
+/**
+ * Function to dispatch an update to the server. This is debounced.
+ */
+const updateCustomerData = debounce( (): void => {
+	const customerDataToUpdate = {} as Partial< BillingAddressShippingAddress >;
+
+	if ( dirtyProps.billingData && customerData?.billingData ) {
```
ralucaStan: would there ever be a missing billingData under customerData? Won't that be equal to the default address? Or is this because of customerData being null? Could that be treated separately?

assets/js/data/cart/push-changes.ts
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Local cache of customerData used for comparisons.
+ */
+let customerData = <CustomerData>null;
+
+/**
+ * Tracks which props have changed so the correct data gets pushed to the server.
+ */
+const dirtyProps = {
```
ralucaStan, line 79: maybe `const DIRTY_PROPS = {,,,}` for consistency ?

ralucaStan (changes requested):
👏 👏  for all the changes brought in
From what I've tested it works as expected. I've made some comments.

assets/js/base/context/hooks/use-checkout-address.ts
mikejolley: @ralucaStan The longer form would be "Use shipping address as billing address", or useShippingAsBilling. Is the short form too confusing?

assets/js/blocks/checkout/inner-blocks/checkout-shipping-address-block/block.tsx
mikejolley: I think we might be able to get rid of it since it's always set.

assets/js/base/context/hooks/use-checkout-address.ts
opr: It is named like this because on the form you enter your shipping address first, and the checkbox controls whether this shipping address should be used as the billing address.

assets/js/data/cart/push-changes.ts
mikejolley: Consistency with what? I see caps used for server data constants, but this is just an object tracking changes. It does change.

assets/js/data/cart/push-changes.ts
mikejolley: This is because of the initial `null` value for customerData and is needed to appease TypeScript. The `null` is needed so we know to populate the customer address from `null` to the received value on first mount.

assets/js/base/context/providers/cart-checkout/checkout-state/types.ts
mikejolley: See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5817#discussion_r811807361

assets/js/base/context/providers/cart-checkout/checkout-state/reducer.ts
mikejolley: Not for this action but the reducer requires the check to keep TS happy. Same for `shouldCreateAccount` and other props used in this reducer, so Ill leave this one in place following that pattern.

assets/js/data/cart/push-changes.ts
mikejolley: I'm happy to leave this as-is unless I'm missing something :)

ralucaStan: I thought we have a all caps for constants approach, but looks like that's not the case.

assets/js/base/context/hooks/use-checkout-address.ts
ralucaStan: > @ralucaStan The longer form would be "Use shipping address as billing address", or useShippingAsBilling. Is the short form too confusing?

@mikejolley For me it was confusing. Initially I thought it is inherited from the shortcode block, where the billing is added first.  With `useShippingAsBilling` it makes more sense.

assets/js/data/cart/actions.ts
```diff
@@ -43,7 +43,10 @@ export const receiveCart = (
 };
 
 /**
- * Returns an action object used in updating the store with the provided cart. This omits the customer addresses.
+ * Returns an action object used in updating the store with the provided cart.
+ *
+ * This omits the customer addresses so that only updates to cart items and totals are received. This is useful when
+ * currently editing address information to prevent it being overwritten from the server.
```
ralucaStan, line 49: 👌

assets/js/data/cart/push-changes.ts
ralucaStan: @mikejolley I recommend we add a comment with 
`The null is needed so we know to populate the customer address from null to the received value on first mount.`,
maybe above the declaration of customerData?

Also, I am thinking we should call `updateCustomerData();` only when we have dirty props.
 ```
if (dirtyProps.billingData || dirtyProps.shippingData) {
updateCustomerData();
}
```

and so we can be more intentional in the code.

```
		const { billingData, shippingAddress } = customerData;
		if ( dirtyProps.billingData ) {
			customerDataToUpdate.billing_address = billingData;
			dirtyProps.billingData = false;
		}

		if ( dirtyProps.shippingAddress ) {
			customerDataToUpdate.shipping_address = shippingAddress;
			dirtyProps.shippingAddress = false;
		}
```

ralucaStan (changes requested):
I left some comments. Let me know what you think.

assets/js/data/cart/push-changes.ts
mikejolley: Fixed in [Tweak customerData to avoid null](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5817/commits/f3109acb2cf90c5dee5889d41a8f6caadd516e1e)

assets/js/base/context/hooks/use-checkout-address.ts
mikejolley: See https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5817/commits/9bb57dbdc906ee29db2a98ff0f54e0aeb04bee6f

assets/js/data/cart/push-changes.ts
ralucaStan: Way better 👏

ralucaStan (approved):
@opr @mikejolley huge congrats on the big improvement you brought with this PR

Approving, I've retested after the latest changes and it works as expected. 
@mikejolley there is just the package.json change that will probably cause a conflict

mikejolley:
Tried another rebuild and the package-lock didn't change. Must be current ¯\_(ツ)_/¯ 

Thanks for the review, and thanks @opr for your changes. Merging this now!

PR #5818, opened 2022-02-11
Update version numbers in bug report template

alexflorisca:
I noticed that the version numbers used as examples in the Bug report template are really old, so  just updating to the latest(ish) ones. #nitpick


tarhi-saad:
Thank you @alexflorisca! It looks good to me! 🚀

tarhi-saad (approved):
Thank you @alexflorisca! It looks good to me! 🚀

PR #5825, opened 2022-02-14
Fix: Mini Cart template part editor height

### LINKED ISSUE #5813: Mini Cart template part editor height
## Describe the bug
Hey folks. Not sure if this is just something I am experiencing, but the editing area of the Mini Cart in the Template Part editor is squashed.

## To reproduce

1. Go to Mini Cart Template part

## Expected behavior
I expected it to fill the height of the editor view or proportionally be closer to what it looks like on the site. 

## Screenshots
Screenshot 2022-02-10 at 18 02 08

* Core version: 5.9
* Gutenberg version (if installed): 12.5.4
* WooCommerce version: 6.2.0
* WooCommerce Blocks version: 7.0.0-dev
* Site language: Eng

MacOS 12.0.1
Firefox 96.0.2

dinhtungdu:
Fixes #5813 

When a template part is loaded, Gutenberg set the height of the editor area using `iframe.contentDocument.documentElement.scrollHeight`, at this point, the editor CSS hasn't been loaded yet, which causes the issue in #5813. If we switch to edit another template, then switch back to edit the Mini Cart template part, the height is correct because, at that point, the CSS had been loaded.

This PR uses the inline style to set the height of the Mini Cart Contents block so Gutenberg can calculate the correct height when loading the template part.

### Screenshots
image

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Go to Site Editor.
2. From the W icon, select Template Parts, then select Mini Cart.
3. See the height of the editor area expanded to the remaining browser height.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
@@ -44,6 +44,12 @@ interface Props {
 
 const Edit = ( { clientId }: Props ): ReactElement => {
 	const blockProps = useBlockProps( {
+		/**
+		 * This is a workaround for the Site Editor to calculate the
+		 * correct height of the Mini Cart template part on the first load.
+		 *
+		 * @see https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5825
+		 */
 		style: {
 			minHeight: '100vh',
 		},
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
@@ -43,7 +43,11 @@ interface Props {
 }
 
 const Edit = ( { clientId }: Props ): ReactElement => {
-	const blockProps = useBlockProps();
+	const blockProps = useBlockProps( {
+		style: {
+			minHeight: '100vh',
+		},
+	} );
 
 	const defaultTemplate = [
 		[ 'woocommerce/filled-mini-cart-contents-block', {}, [] ],


assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
```diff
@@ -43,7 +43,11 @@ interface Props {
 }
 
 const Edit = ( { clientId }: Props ): ReactElement => {
-	const blockProps = useBlockProps();
+	const blockProps = useBlockProps( {
+		style: {
+			minHeight: '100vh',
+		},
+	} );
```
tomasztunik: Can we move it to css at https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/d67e5191f0265a9ea8eb02aa274b9d3f459335ff/assets/js/blocks/cart-checkout/mini-cart-contents/editor.scss#L1-L6

We already have the `max-width` styles and other editor specific styles living there.

tomasztunik (changes requested):
The fix works great 🎉  but it would be great if we could keep it in scss along other overrides.

assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
dinhtungdu: @tomasztunik Hey there 👋🏽 , https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5813 is caused by a race condition related to loading CSS files. When we switch to edit the Mini Cart template part, the Site Editor calculates the height of the editor based on the height of template content, the editor style of the Mini Cart Contents block hasn't been loaded yet, which causes the issue we're fixing in this PR. I have to use the inline style to avoid the race condition so the Site Editor can calculate the correct height every time the template part is loaded.

dinhtungdu: On `trunk`, if you switch back and forth editing the template part, the editor height is correct the second time.

tomasztunik: Ah, I see. I didn't understand the full context properly. Would it be worth to document it with a comment so whoever stumbles upon it doesn't move it to styles in the future?

dinhtungdu: > Would it be worth to document it with a comment so whoever stumbles upon it doesn't move it to styles in the future?

@tomasztunik I added the comment explaining the usage of the inline style.

tomasztunik: thanks!

tomasztunik (approved):

PR #5839, opened 2022-02-14
Shopper → Checkout → Can choose payment option

### LINKED ISSUE #5782: Critical flows: Shopper → Checkout → Can choose payment option
### Story

As a shopper, I want to be able to choose a payment option.

### File

`tests/e2e/specs/shopper/checkout-payments.test.js`

#### Direct bank transfer

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Select the payment option `Direct bank transfer`.
5. Place the order.
6. Verify that the shipping options `Direct bank transfer` is visible on the order confirmation page.

#### Cash on delivery

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Select the payment option `Cash on delivery`.
5. Place the order.
6. Verify that the shipping options `Cash on delivery` is visible on the order confirmation page.

#### Check payments

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Select the payment option `Check payments`.
5. Place the order.
6. Verify that the shipping options `Check payments` is visible on the order confirmation page.

nielslange:
Fixes #5782

### Notes

This PR adds e2e tests to verify that a shopper can choose the default payment options (`Direct bank transfer`, `Cash on delivery` and `Check payments`) on the checkout block. 

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/checkout-payments.test.js` to run only this test case.
5. Run `npm run test:e2e -- tests/e2e/specs/shopper/*` to run all shopper-related test cases.
6. Run `npm run test:e2e` to run the whole test suite.


tests/e2e/specs/shopper/checkout-payments.test.js
```diff
@@ -0,0 +1,86 @@
+/**
+ * External dependencies
+ */
+import {
+	merchant,
+	setCheckbox,
+	settingsPageSaveChanges,
+	verifyCheckboxIsSet,
```
tarhi-saad: Hi @nielslange! The following functions aren't used in the code. I think it's safe to remove them:

- `setCheckbox`
- `settingsPageSaveChanges`
- `verifyCheckboxIsSet`

tests/e2e/specs/shopper/checkout-payments.test.js
```diff
@@ -0,0 +1,86 @@
...
+		await shopper.removeFromCart( simpleProductName );
+		await page.evaluate( () => {
+			localStorage.removeItem(
+				'wc-blocks_dismissed_compatibility_notices'
+			);
+		} );
+	} );
+
+	it( 'allows customer to choose available payment methods', async () => {
+		await shopper.goToCheckoutBlock();
```
tarhi-saad: This line doesn't seem necessary! We go to the Checkout Block after adding an item to our cart. So, instead of:

```js
await shopper.goToCheckoutBlock();
await page.goto( productPermalink );
await shopper.addToCart();
await shopper.goToCheckoutBlock();
```

We can have:

```js
await page.goto( productPermalink );
await shopper.addToCart();
await shopper.goToCheckoutBlock();
```

tests/e2e/specs/shopper/checkout-payments.test.js
```diff
@@ -0,0 +1,86 @@
...
+		await page.goto( productPermalink );
+		await shopper.addToCart();
+		await shopper.goToCheckoutBlock();
+
+		await expect( page ).toClick(
```
tarhi-saad: I wonder if we shouldn't ensure that the correct payment method gets selected after clicking on it. The following code only ensures that an element with the class `wc-block-components-payment-method-label` and inner text `Direct bank transfer` gets clicked, but we should follow up with a code to test if it did get indeed selected. Sometimes, an element gets clicked without getting the proper behavior, and the test will still pass.

```js
await expect( page ).toClick(
	'.wc-block-components-payment-method-label',
	{
		text: 'Direct bank transfer',
	}
);
```

For example, I have tested the following code and the test passed! I clicked on an H1 element:

```js
await expect( page ).toClick( '.entry-title', {
	text: 'Checkout Block',
} );
```

tarhi-saad (changes requested):
Hi @nielslange! Thank you for the changes! I have added some comments on the code below.

tests/e2e/specs/shopper/checkout-payments.test.js
nielslange: Good catch! I just removed them.

tests/e2e/specs/shopper/checkout-payments.test.js
nielslange: I cannot explain what's going on, but when I remove the first `await shopper.goToCheckoutBlock();` call, the test keeps failing. I only kept the first call, to avoid flakiness. 😬

nielslange: Might have been a problem due to repeatable reads. When spinning up a fresh e2e environment,  the tests runs expected. 😀

tests/e2e/specs/shopper/checkout-payments.test.js
nielslange: Great suggestion! I updated the test in a way that I check if the payment option had been checked after it had been selected.

nielslange:
> Hi @nielslange! Thank you for the changes! I have added some comments on the code below.

Thanks for your review, @tarhi-saad. I applied all suggestions. Feel free to do another review.

tarhi-saad (approved):
Excellent work, @nielslange! Thank you! 👍 🚀

nielslange:
> Excellent work, @nielslange! Thank you! 👍 🚀

Thanks a lot, @tarhi-saad! Would you mind doing another review, as I applied some changes based on feedback I received for another test? I outsourced some functionality and added another step to test is the selected payment option appears on the order confirmation page. That way, we know for sure that the selected payment method had been successfully stored.

tests/utils/shopper.js
```diff
@@ -46,4 +46,22 @@ export const shopper = {
...
+				text: payment,
+			}
+		);
+	},
+
+	placeOrder: async () => {
+		await expect( page ).toClick( '.wc-block-components-button__text', {
+			text: 'Place Order',
+		} );
+		page.waitForNavigation( {
```
tarhi-saad: Hi @nielslange!

The `page.waitForNavigation()` function returns a Promise, so it should be preceded by the `await` keyword to get resolved. And since it resolves when the page navigates to a new URL or reloads, we should use the `Promise.all()` method to work properly, like so:

```js
placeOrder: async () => {
  await Promise.all( [
	  page.waitForNavigation( {
		  waitUntil: 'networkidle0',
	  } ),
	  expect( page ).toClick( '.wc-block-components-button__text', {
		  text: 'Place Order',
	  } ),
  ] );
},
```

This way `page.waitForNavigation()` & `expect( page ).toClick()` will execute at the same time. The latter will resolve when the element is clicked, and the first one will resolve when the page navigates. So, executing at the same time is important in this particular case.

Otherwise, if we have this code instead:

```js
placeOrder: async () => {
  await page.waitForNavigation( {
    waitUntil: 'networkidle0',
  } ),
    text: 'Place Order',
  } ),
},
```

Or, this one:

```js
placeOrder: async () => {
    text: 'Place Order',
  } ),
  await page.waitForNavigation( {
    waitUntil: 'networkidle0',
  } ),
},
```

Both functions will have to wait for each promise to resolve sequentially, which will result in indefinitely waiting for a page to navigate to a new URL when the `page.waitForNavigation()` is executed.

You can test that using the following code:

This one fails with: `TimeoutError: Navigation timeout of 30000 ms exceeded` error ❌
```js
await page.waitForNavigation(); // The promise resolves after navigation has finished
await page.click( 'li.page-item-6 a' ); // Clicking the link will indirectly cause a navigation
```

And this one succeed ✅

```js
await Promise.all( [
	page.waitForNavigation(), // The promise resolves after navigation has finished
	page.click( 'li.page-item-6 a' ), // Clicking the link will indirectly cause a navigation
	// 'li.page-item-6 a' is the 'cart' page link on the navbar
] );
```

tarhi-saad (changes requested):
Hi @nielslange! Thank you for the changes!

As a follow-up to [my comment in the code below](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5839#discussion_r809595830), the changes suggested will still result in the `TimeoutError: Navigation timeout of 30000 ms exceeded` error. This means that clicking on the `Place Order` button doesn't trigger navigation to a new URL (i.e., The order received page). So, either the payment fails or takes too long to redirect (i.e., <30000 ms).

tarhi-saad:
Hi @nielslange! I have done some debugging, and I would like to share some findings that could help us!

## Form warnings bug

As I mentioned on slack, the reason why we are still getting the `TimeoutError: Navigation timeout of 30000 ms exceeded` error even after the suggested changes is getting form warnings. Some fields aren't prefilled:

- First name
- Last name
- Address
- City
- ZIP Code

**Note:** We can use the interactive mode when running our E2E tests to see exactly what's happening using the following command: `npm run test:e2e-dev -- tests/e2e/specs/shopper/checkout-payments.test.js `. This opens up a Chrome window and starts a live testing session. 

So, the solution for this first bug is to fill the required fields before placing our order.

We can add a `fillBillingDetails` method to our `shopper` object like so:

```js
// tests/utils/shopper.js file

export const shopper = {
      ... // Some code above

      fillBillingDetails: async ( customerBillingDetails ) => {
            await expect( page ).toFill(
	            '#billing-first_name',
	            customerBillingDetails.firstname
            );
            await expect( page ).toFill(
	            '#billing-last_name',
	            customerBillingDetails.lastname
            );
            await expect( page ).toFill(
	            '#components-form-token-input-0',
	            customerBillingDetails.country
            );
            await expect( page ).toFill(
	            '#billing-address_1',
	            customerBillingDetails.addressfirstline
            );
            await expect( page ).toFill(
	            '#billing-address_2',
	            customerBillingDetails.addresssecondline
            );
            await expect( page ).toFill(
	            '#billing-city',
	            customerBillingDetails.city
            );
            await expect( page ).toFill(
	            '#components-form-token-input-2',
	            customerBillingDetails.state
            );
            await expect( page ).toFill(
	            '#billing-postcode',
	            customerBillingDetails.postcode
            );
            await expect( page ).toFill( '#phone', customerBillingDetails.phone );
            await expect( page ).toFill( '#email', customerBillingDetails.email );
      },
};
```

Then, in our test file, we can import our default JSON file to get our billing info:

```js
// tests/e2e/specs/shopper/checkout-payments.test.js file

... // Some code above

import config from '../../config/default.json';

const billing = config.addresses.customer.billing;

... // Some code

it( 'allows customer to pay using Direct bank transfer', async () => {
    await page.goto( productPermalink );
    await shopper.addToCart();
    await shopper.goToCheckoutBlock();

    await shopper.fillBillingDetails( billing ); // ====> The added code. Fill details before placing the order.

    await shopper.placeOrder();
});
```

## The payment method isn't mentioned on the `order-received` page

Even after the above changes, the following check fails:

```js
// tests/e2e/specs/shopper/checkout-payments.test.js file

await expect( page ).toMatch( bacs );
``` 

This means we can't find any mention of the selected payment on this page. The only thing that we can verify at this stage is if we did actually get redirected to the `order-received` page. One way to do this check is:

```js
// tests/e2e/specs/shopper/checkout-payments.test.js file

const orderReceived = 'Thank you. Your order has been received.';

it( 'allows customer to pay using Direct bank transfer', async () => {
    ... // Some code above

    await shopper.placeOrder();
    await expect( page ).toMatch( orderReceived ); // Instead of the payment method name
} );
```

## Check if the desired payment is selected

Until now, we didn't check if our payment was actually selected. Maybe we should run this check before placing our order!

## Error on the built-in `shopper.removeFromCart()` method

With all our last changes, we will still get another error on the following part of the code:

```js
// tests/e2e/specs/shopper/checkout-payments.test.js file

afterAll( async () => {
    // empty cart from shortcode page
    await shopper.goToCart();
    await shopper.removeFromCart( simpleProductName ); // ====> This is the line causing this error
    await merchant.reactivateCompatibilityNotice();
    } );
```

This error is displaying: `TypeError: Cannot read properties of undefined (reading 'click')`. After doing some investigation, I found that the `shopper.removeFromCart()` method doesn't check if the cart is already empty. And since our purchase is now successful, the cart gets emptied automatically, but when running this method on an empty cart, we get this error.

To fix that, we can use another built-in method that checks if the cart is empty: `shopper.emptyCart()`. So, the above code will look like this:

```js
afterAll( async () => {
  // empty cart from shortcode page
  await shopper.emptyCart();
  await merchant.reactivateCompatibilityNotice();
} );
```

You'll notice that we have removed the `shopper.goToCart()` as well. This is because our new `shopper.emptyCart()` method does go to the cart, then remove products if they exist, and remove coupons if they exist as well.

nielslange:
@tarhi-saad Could you do another review of this PR? Please note that when running the complete test suite, many test cases fail. I assume that's related to https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/6071.

tarhi-saad (approved):
Great job, @nielslange! 

I'm not sure if it's worth emptying the cart before each test and after the end of the test suite, even though we place an order. This will be fine if our tests are successful, but if one of them fails for some reason, our product will remain in the cart, which may cause other tests to fail.

nielslange:
> I'm not sure if it's worth emptying the cart before each test and after the end of the test suite, even though we place an order. This will be fine if our tests are successful, but if one of them fails for some reason, our product will remain in the cart, which may cause other tests to fail.

Good call! To be on the safe side, I'm emptying the cart before and after each test.

PR #5844, opened 2022-02-15
Shopper → Checkout → Can choose shipping option

### LINKED ISSUE #5781: Critical flows: Shopper → Checkout → Can choose shipping option
### Story

As a shopper, I want to be able to choose a shipping option.

### File

`tests/e2e/specs/shopper/checkout-shipping.test.js`

#### Free Shipping

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Select shipping option `Free Shipping`.
5. Place the order.
6. Verify that the shipping options `Free Shipping` is visible on the order confirmation page.

#### Flat Rate Shipping

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Select shipping option `Normal Shipping`.
5. Place the order.
6. Verify that the shipping options `Normal Shipping` is visible on the order confirmation page.

nielslange:
Fixes #5781

### Notes

This PR adds e2e tests to verify that a shopper can choose the shipping options (`Normal Shipping ` and `Free Shipping`) on the checkout block.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/checkout-shipping.test.js` to run only this test case.

### DIFF
--- a/tests/e2e/specs/shopper/checkout-shipping.test.js
+++ b/tests/e2e/specs/shopper/checkout-shipping.test.js
@@ -1,0 +1,57 @@
+/**
+ * External dependencies
+ */
+import { merchant } from '@woocommerce/e2e-utils';
+
+/**
+ * Internal dependencies
+ */
+import { shopper } from '../../../utils';
+
+const productName = '128GB USB Stick';
+const freeShippingName = 'Free Shipping';
+const freeShippingPrice = '$0.00';
+const normalShippingName = 'Normal Shipping';
+const normalShippingPrice = '$20.00';
+
+if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( 'Skipping Checkout tests', () => {} );
+
+describe( `Shopper → Checkout → Can choose shipping option`, () => {
+	beforeAll( async () => {
+		await merchant.login();
+	} );
+
+	afterAll( async () => {
+		await merchant.logout();
+	} );
+
+	it( 'allows customer to choose free shipping', async () => {
+		await shopper.goToShop();
+		await shopper.addToCartFromShopPage( productName );
+		await shopper.block.goToCheckout();
+		await shopper.block.selectAndVerifyShippingOption(
+			freeShippingName,
+			freeShippingPrice
+		);
+		await shopper.block.placeOrder();
+		await page.waitForTimeout( 2000 );
+		await expect( page ).toMatch( 'Order received' );
+		await expect( page ).toMatch( freeShippingName );
+	} );
+
+	it( 'allows customer to choose flat rate shipping', async () => {
+		await shopper.goToShop();
+		await shopper.addToCartFromShopPage( productName );
+		await shopper.block.goToCheckout();
+		await shopper.block.selectAndVerifyShippingOption(
+			normalShippingName,
+			normalShippingPrice
+		);
+		await shopper.block.placeOrder();
+		await page.waitForTimeout( 2000 );
+		await expect( page ).toMatch( 'Order received' );
+		await expect( page ).toMatch( normalShippingName );
+	} );
+} );
--- a/bin/wp-env-config.sh
+++ b/bin/wp-env-config.sh
@@ -18,5 +18,6 @@ wp rewrite flush
 wp core version --extra
 wp plugin list
 wp theme activate storefront
+wp wc customer update 1 --user=1 --billing='{"first_name":"John","last_name":"Doe","company":"Automattic","country":"US","address_1":"addr 1","address_2":"addr 2","city":"San Francisco","state":"CA","postcode":"94107","phone":"123456789"}' --shipping='{"first_name":"John","last_name":"Doe","company":"Automattic","country":"US","address_1":"addr 1","address_2":"addr 2","city":"San Francisco","state":"CA","postcode":"94107","phone":"123456789"}'
 
 exit $EXIT_CODE
--- a/tests/e2e/fixtures/fixture-data.js
+++ b/tests/e2e/fixtures/fixture-data.js
@@ -344,6 +344,30 @@ const Shipping = () => [
 			},
 		],
 	},
+	{
+		name: 'US',
+		locations: [
+			{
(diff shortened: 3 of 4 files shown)


tests/e2e/specs/shopper/checkout-shipping.test.js
```diff
@@ -0,0 +1,92 @@
...
+			}
+		);
+
+		await expect( page ).toClick(
+			'.wc-block-components-radio-control__label',
+			{
+				text: normalShippingName,
+			}
+		);
+		await page.waitForTimeout( 1000 );
```
mikejolley: How reliable is this? Is 1000 a guess? Will it fail if it takes longer to select?

tests/e2e/specs/shopper/checkout-shipping.test.js
```diff
@@ -0,0 +1,92 @@
...
+if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( `skipping ${ block.name } tests`, () => {} );
+
+describe( `${ block.name } Block (frontend)`, () => {
+	let productPermalink;
+
+	beforeAll( async () => {
+		// prevent CartCheckoutCompatibilityNotice from appearing
+		await page.evaluate( () => {
```
mikejolley: I wonder if we can more this to a util since it's used in many of the e2e tests now.

tests/e2e/specs/shopper/checkout-shipping.test.js
```diff
@@ -0,0 +1,92 @@
...
+		} );
+	} );
+
+	it( 'allows customer to choose available shipping methods', async () => {
+		await page.goto( productPermalink );
+		await shopper.addToCart();
+		await shopper.goToCheckoutBlock();
+
+		await expect( page ).toClick(
+			'.wc-block-components-radio-control__label',
```
mikejolley: If we're going to be doing similar things across all of our critical paths, it might make sense to make some helper functions for doing common actions such as selecting shipping methods and waiting for selection to complete. Thoughts?

mikejolley (commented):
This passed, but I do see many "Failed to load resource: the server responded with a status of 404 (Not Found)" for some reason, just like the CI gets. I'm not sure if it's related to this change or a general issue.

nielslange:
> This passed, but I do see many "Failed to load resource: the server responded with a status of 404 (Not Found)" for some reason, just like the CI gets. I'm not sure if it's related to this change or a general issue.

I also noticed this and this is a general problem triggered by this section:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1c478e60acb805733be04c5ca4c9f2d2443f7551/tests/e2e/config/jest.setup.js#L126-L130
I haven't figured out how to solve it, though. It's something I need to investigate closer.

tests/e2e/specs/shopper/checkout-shipping.test.js
nielslange: > How reliable is this? Is 1000 a guess?

When running the tests initially, verifying free shipping worked as expected, as that option is preselected. However, verifying flat rate shipping failed, as the shipping price is not immediately visible after selecting that option. I looked up the e2e tests in WooCommerce core and noticed that they're using a timeout of `500ms`. When testing the timeout of `500ms`, the tests sometimes passed, but also sometimes failed. When setting the value to `1000ms`, they pass in more than 90% of the runs. 

> Will it fail if it takes longer to select?

Yes. If the price is not visible `100ms` after selecting the corresponding shipping option, the test case fails.

Do you have any thoughts on how to handle this differently?

tests/e2e/specs/shopper/checkout-shipping.test.js
nielslange: Thanks for pointing this out. I created the file `tests/utils/merchant.js` and added the following two functions:

```js
preventCompatibilityNotice: async () => {
	await page.evaluate( () => {
		localStorage.setItem(
			'wc-blocks_dismissed_compatibility_notices',
			'["checkout"]'
		);
	} );
},

reactivateCompatibilityNotice: async () => {
	await page.evaluate( () => {
		localStorage.removeItem(
			'wc-blocks_dismissed_compatibility_notices'
		);
	} );
},
```

This way, I can call them in `beforeAll()` and `afterAll()` as `await merchant.preventCompatibilityNotice();` and `await merchant.reactivateCompatibilityNotice();`. Is that what you had in mind?

tests/e2e/specs/shopper/checkout-shipping.test.js
nielslange: Yes. That makes indeed sense. I added the following two functions to `tests/utils/shopper.js`:

```js
selectAndVerifyShippingOption: async ( shippingName, shippingPrice ) => {
	await expect( page ).toClick(
		'.wc-block-components-radio-control__label',
		{
			text: shippingName,
		}
	);
	await page.waitForTimeout( 500 );
	await expect( page ).toMatchElement(
		'.wc-block-components-totals-shipping .wc-block-formatted-money-amount',
		{
			text: shippingPrice,
		}
	);
},

placeOrder: async () => {
		text: 'Place Order',
	} );
	page.waitForNavigation( {
		waitUntil: 'networkidle0',
	} );
},
```

Do you think `selectAndVerifyShippingOption()` is in a good shape, or would you suggest creating the two sub-functions `selectShippingOption()` and `verifyShippingOption()`?

tests/e2e/specs/shopper/checkout-shipping.test.js
mikejolley: wait for selector?

tests/e2e/specs/shopper/checkout-shipping.test.js
mikejolley: This looks good to me 👍🏻

mikejolley (approved):
Tests passed and the requested changes look good. Thanks for that.

Just a merge conflict and it can merge.

PR #5848, opened 2022-02-15
Limit API response based on what is updated

### LINKED ISSUE #5747: Find a way to keep the address up to date in CheckoutShippingAddress and CheckoutBillingAddress

opr:
This PR will add a utility function to remove items from arrays based on their key (in `ArrayUtils.php`) and will also specify some keys to be omitted on the `CartUpdateCustomer (/wc/store/cart/update-customer)` route. Specifically it will omit `billing_address` and `shipping_address` from the response.

The reason we want to do this is because: now we've removed local state, and we're now using the data store to drive the data in the checkout form, when the user is typing, the `UPDATE_CUSTOMER_DATA` action pushes the new address data to the server. The server then responds with the cart object, which includes `shipping_address` and `billing_address` - when the new cart is received, in the `RECEIVE_CART` action, any changes to the address data since _sending_ the request will be overwritten by the response.

Fixes #5747

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Type into either address field on the Checkout block, update the city/postcode/country.
2. Ensure the address does not revert to a previous one while typing.
3. Keep typing, but slow your rate down to 1 character per second, or similar, ensure the form contents do not switch back and forth while you type.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/js/data/cart/reducers.ts
+++ b/assets/js/data/cart/reducers.ts
@@ -70,6 +70,7 @@ const reducer: Reducer< CartState > = (
 					...state,
 					errors: EMPTY_CART_ERRORS,
 					cartData: {
+						...state.cartData,
 						...action.response,
 					},
 				};
--- a/src/StoreApi/Routes/CartUpdateCustomer.php
+++ b/src/StoreApi/Routes/CartUpdateCustomer.php
@@ -122,7 +122,7 @@ class CartUpdateCustomer extends AbstractCartRoute {
 
 		$this->calculate_totals();
 
-		return rest_ensure_response( $this->schema->get_item_response( $cart ) );
+		return rest_ensure_response( $this->schema->get_item_response( $cart, [ 'shipping_address', 'billing_address' ] ) );
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


mikejolley:
Closing in favour of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5817

To note my feedback via Slack, we need to avoid creating the ability for endpoints to change like this - they must always match the route schema.

PR #5851, opened 2022-02-15
fix regression for Twenty Twenty-Two with sale prices being underlined

### LINKED ISSUE #5286: Twenty Twenty Two: discounted prices are underlined
## To reproduce
1. With the [Twenty Twenty Two theme](https://wordpress.org/themes/twentytwentytwo/), add the Products by Category block to a product or page.
2. In the frontend, check how discounted prices are shown.
3. Notice the discounted prices is underlined.

## Expected behavior
There shouldn't be underline under the discounted price.

## Screenshots
imatge

tomasztunik:
class name on price container in Product by * blocks changed from `wc-block-components-product-price` to `wc-block-grid__product-price`.

This affects trunk only so shouldn't require changelog entry.

Fixes #5286

### Manual Testing

1. On Twenty Twenty-Two theme 
2. On a page add Products by (something) block
3. Make sure it has a product on sale
4. Make sure product sale price is not underlined

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -329,7 +329,8 @@
 		}
 	}
 
-	.wc-block-components-product-price {
+	.wc-block-components-product-price,
+	.wc-block-grid__product-price {
 		ins {
 			text-decoration: none;
 		}


tomasztunik:
I've kept previous css class name as well - it seems like it was done this way on twenty-twenty theme overrides so imagine both classes are possible?

sunyatasattva (approved):
I'm approving this and cherry picking into the next release!

Thanks @tomasztunik for fixing it.

PR #5852, opened 2022-02-15
Fix Global Styles for Product Image, Product Category List and Product Tag List Blocks

gigitux:
This PR fixes Global Styles for Product Image, Product Category List, and Product Tag List Blocks when they are used on `All Products` block.

The fix is pretty easy, we need to register the blocks on PHP side, in this way we can use the `__experimentalSelector` and attach the generated Global Style to a specific element in the DOM.

## Product Image
3. Add the `All Products` block  (this block contains `Product Image`) to a post.
4. Get the focus on `Product Image` block.
10. Verify that the `Product Image Block` is shown under the `Blocks` section. Personalize again the block.
11. Save your changes.
14. Change again the styles.
15. Save your changes.

## Product Category List
3. Add the `All Products` block  (this block contains `Product Category List`) to a post.
4. Get the focus on `Product Category List` block.
10. Verify that the `Product Category List` is shown under the `Blocks` section. Personalize again the block.
11. Save your changes.
14. Change again the styles.
15. Save your changes.
16. Check if these styles have priority over the styles from the Site Editor (except for link color).

**Due to the limitations of the global styles API, the global style link color will have a higher priority than the block style :(**

The generated global style CSS is more specific than the generated block style CSS and there is no way to change this behavior.   

image

## Product Tag List
3. Add the `All Products` block  (this block contains `Product Tag List`) to a post.
4. Get the focus on the `Product Tag List` block.
10. Verify that the `Product Tag List` is shown under the `Blocks` section. Personalize again the block.
11. Save your changes.

### DIFF
--- a/assets/js/atomic/blocks/product-elements/category-list/index.ts
+++ b/assets/js/atomic/blocks/product-elements/category-list/index.ts
@@ -33,9 +33,11 @@ const blockConfig: BlockConfiguration = {
 				text: true,
 				link: true,
 				background: false,
+				__experimentalSkipSerialization: true,
 			},
 			typography: {
 				fontSize: true,
+				__experimentalSkipSerialization: true,
 			},
 			__experimentalSelector:
 				'.wc-block-components-product-category-list',
--- a/assets/js/atomic/blocks/product-elements/category-list/index.ts
+++ b/assets/js/atomic/blocks/product-elements/category-list/index.ts
@@ -34,10 +34,12 @@ const blockConfig: BlockConfiguration = {
 				link: true,
 				background: false,
 			},
+			typography: {
+				fontSize: true,
+			},
+			__experimentalSelector:
+				'.wc-block-components-product-category-list',
 		} ),
-		typography: {
-			fontSize: true,
-		},
 	},
 	save: Save,
 	edit,
--- a/assets/js/atomic/blocks/product-elements/image/supports.ts
+++ b/assets/js/atomic/blocks/product-elements/image/supports.ts
@@ -8,8 +8,9 @@ export const supports = {
 		__experimentalBorder: {
 			radius: true,
 		},
+		typography: {
+			fontSize: true,
+		},
+		__experimentalSelector: '.wc-block-components-product-image',
 	} ),
-	typography: {
-		fontSize: true,
-	},
 };
--- a/assets/js/atomic/blocks/product-elements/tag-list/supports.js
+++ b/assets/js/atomic/blocks/product-elements/tag-list/supports.js
@@ -10,8 +10,9 @@ export const supports = {
 			background: false,
 			link: true,
 		},
+		typography: {
+			fontSize: true,
+		},
+		__experimentalSelector: '.wc-block-components-product-tag-list',
 	} ),
-	typography: {
-		fontSize: true,
-	},
 };
--- a/assets/js/atomic/blocks/product-elements/image/edit.js
+++ b/assets/js/atomic/blocks/product-elements/image/edit.js
@@ -14,7 +14,6 @@ import { getAdminLink } from '@woocommerce/settings';
 import Block from './block';
 import withProductSelector from '../shared/with-product-selector';
 import { BLOCK_TITLE, BLOCK_ICON } from './constants';
-import './editor.scss';
 
 const Edit = ( { attributes, setAttributes } ) => {
 	const {
--- a/assets/js/atomic/blocks/product-elements/image/editor.scss
+++ b/assets/js/atomic/blocks/product-elements/image/editor.scss
@@ -1,5 +1,0 @@
-.wp-block-woocommerce-product-image {
-	.components-disabled {
-		border-radius: inherit;
-	}
-}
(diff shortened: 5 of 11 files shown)


assets/js/atomic/blocks/product-elements/category-list/index.ts
```diff
@@ -34,10 +34,12 @@ const blockConfig: BlockConfiguration = {
...
+			typography: {
+				fontSize: true,
+			},
+			__experimentalSelector:
+				'.wc-block-components-product-category-list',
 		} ),
-		typography: {
-			fontSize: true,
-		},
 	},
```
tomasztunik, line 45: Just want to confirm - we do not want to expose typography unless it's feature build? Was it available in core with typography enabled before? If so it would mean people who used it would lose the control unless they installed stand alone plugin and it wouldn't be obvious. Should we move only the experimentalSelector inside of featurePluginBuild check?

I imagine if CSS was changed keeping it here would make CSS not work right?

assets/js/atomic/blocks/product-elements/image/supports.ts
```diff
@@ -8,8 +8,9 @@ export const supports = {
 		__experimentalBorder: {
 			radius: true,
 		},
+		typography: {
+			fontSize: true,
+		},
```
tomasztunik, line 13: What is the function of typography on image block? I couldn't find it affecting anything.

assets/js/atomic/blocks/product-elements/tag-list/supports.js
```diff
@@ -10,8 +10,9 @@ export const supports = {
...
 			link: true,
 		},
+		typography: {
+			fontSize: true,
+		},
+		__experimentalSelector: '.wc-block-components-product-tag-list',
 	} ),
-	typography: {
-		fontSize: true,
-	},
```
tomasztunik, line 16: same question here regarding existing users if we want to hide it behind the flag

tomasztunik (commented):
Looks good just couple questions :) 

- global styles work
- block styles work
- default styles work

all settings override styles correctly

Editor and front-end seems to render typography and buttons differently. Do we consider this part of the issue or separate? 

| editor | front-end |
| - | - |
| Screenshot 2022-02-22 at 10 59 53 |  Screenshot 2022-02-22 at 10 59 46 |

Also important to note this affects not only All Products block but also Single Product block

assets/js/atomic/blocks/product-elements/category-list/index.ts
gigitux: > Was it available in core with typography enabled before? 

Yes, but these blocks are available on the `All Products` block (but the global style doesn't work and this PR fixes it), and the `Single Product` block that it is an experimental block. So it is visible only on the dev build.

So I guess that this change is pretty safe. What do you think?

assets/js/atomic/blocks/product-elements/image/supports.ts
gigitux: > What is the function of typography on image block? I couldn't find it affecting anything.

For the sale label (visible only when the product is in sale).

https://user-images.githubusercontent.com/4463174/155153605-b870790e-4b0e-4d22-83bb-a954e24abefe.mp4

assets/js/atomic/blocks/product-elements/tag-list/supports.js
gigitux: Same here

assets/js/atomic/blocks/product-elements/category-list/index.ts
tomasztunik: I think usage should be small enough that we can always recommend to install standalone blocks to have it enabled if people request it restored :)

assets/js/atomic/blocks/product-elements/image/supports.ts
tomasztunik: Ah! True, didn't have it visible on single product. 

Thinking if having control over the sale badge as a separate sale badge, or generic badge block wouldn't be better maybe. This way one could control how it looks (including border, typography, color etc) and apply across the board.

What do you think?

gigitux: > Thinking if having control over the sale badge as a separate sale badge, or generic badge block wouldn't be better maybe. This way one could control how it looks (including border, typography, color etc) and apply across the board.

I agree, but we need to create another block. I guess that it is outside of the scope of this project.

assets/js/atomic/blocks/product-elements/category-list/index.ts
gigitux: Yes, exactly!

tomasztunik (approved):
(my commits from pairing session)

🙌⚡️

gigitux:
Thanks @tomasztunik for the pair programming session about the global styles! With this PR we introducing a patch for `useTypographyProps` https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5852/commits/613addf9e99c2b3373db5a82d73c1de3ba15e269 that is already in pending on https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5816 PR

PR #5855, opened 2022-02-15
Update testing instructions for 6.8.0 and 6.9.0 releases

alexflorisca:
Some minor changes to the testing instructions for the 6.8.0 and 6.9.0 releases to enable WC peeps to discern what to test a bit better for the WC 6.3.0 release.

### To test:
- Check that everything listed under "WC Core" is work that will be included in core, and anything under "Feature Plugin only" will only be included for users with this plugin active


nielslange (approved):
LGTM! 🚢

PR #5860, opened 2022-02-16
Shopper → Checkout → Can have different shipping and billing addresses

### LINKED ISSUE #5778: Critical flows: Shopper → Checkout → Can have different shipping and billing addresses
### Story

As a shopper, I want to be able to provide shipping and billing addresses.

### File

`tests/e2e/specs/shopper/checkout-different-addresses.test.js`

### Test

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Uncheck the checkbox `Use same address for billing`.
5. Provide different shipping and billing addresses.
6. Place the order.
7. Verify that the shipping and billing addresses are correct.

nielslange:
Fixes #5778

### Notes

This PR adds e2e tests to verify that a shopper can choose different shipping and billing addresses on the checkout block.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/checkout-different-addresses.test.js` to run only this test case.
5. Run `npm run test:e2e -- tests/e2e/specs/shopper/*` to run all shopper-related test cases.
6. Run `npm run test:e2e` to run the whole test suite.


ralucaStan:
For tests we have already two folders: `backend`, for tests related to the editor and `frontend`, for tests related to the website rendering. This PR introduces an additional `shopper` folder which has the same purpose as the `frontend` one.  

I recommend removing this `shopper` folder and moving this test to the `frontend` test, ideally under a `/checkout` folder.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,101 @@
...
+		await shopper.removeFromCart( simpleProductName );
+		await page.evaluate( () => {
+			localStorage.removeItem(
+				'wc-blocks_dismissed_compatibility_notices'
+			);
+		} );
+	} );
+
+	it( 'allows customer to have different shipping and billing addresses', async () => {
+		await shopper.goToCheckoutBlock();
```
ralucaStan: Is this needed?

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,101 @@
...
+		await page.goto( productPermalink );
+		await shopper.addToCart();
+		await shopper.goToCheckoutBlock();
+
+		await page.evaluate( () => {
+			const checkbox = document.querySelector(
+				'.wc-block-checkout__use-address-for-billing .wc-block-components-checkbox__input'
+			);
+			if ( checkbox.checked ) {
```
ralucaStan: Here we need to test if this checkbox is checked, as this is the default behaviour and we need to make sure this is respected.

After that a click needs to be made and then the fields checked.

nielslange:
> For tests we have already two folders: `backend`, for tests related to the editor and `frontend`, for tests related to the website rendering. This PR introduces an additional `shopper` folder which has the same purpose as the `frontend` one.
> 
> I recommend removing this `shopper` folder and moving this test to the `frontend` test, ideally under a `/checkout` folder.

I introduced the `shopper` and `merchant` folders as this is the way WooCommerce core is structuring their e2e tests. Given that the WooCommerce Blocks plugin will become part of the WooCommerce core over time, I decided to follow their lead instead of shifting away from their known structure. The `shopper` and `merchant` folders of the WooCommerce core e2e tests can be found on https://github.com/woocommerce/woocommerce/tree/trunk/packages/js/e2e-core-tests/src/specs.

Based on this information, would you still suggest sticking to the `frontend` and `backend` folders, risking that we might have to rename them later if WooCommerce Blocks gets finally merged into core?

tests/e2e/specs/shopper/checkout-different-addresses.test.js
ralucaStan: I am curious if we can test that this data is actually handled correctly when submitting the form. Because filling the form fields is a good thing, because it tests that all the fields are there. But we don't have an assertion that the data is used.

Would you mind trying to submit the form, and trying to see if:
- the form is successfully submitted 
- don't know if it would be possible to use `await page.waitForNavigation();` and search on the thank you page for the data under shipping address to match the data you introduced
Screenshot 2022-02-16 at 13 44 06
.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,101 @@
...
+};
+
+const simpleProductName = '128GB USB Stick';
+const pageName = 'Checkout Block';
+
+if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( `skipping ${ block.name } tests`, () => {} );
+
+describe( `${ block.name } Block (frontend)`, () => {
```
ralucaStan: Let's change this to something specific to what is being tested. 
`Shopper → Checkout → Can have different shipping and billing addresses` could work here

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+	afterAll( async () => {
+		// empty cart from shortcode page
+		await shopper.goToCart();
+		await shopper.removeFromCart( simpleProductName );
+		await page.evaluate( () => {
+			localStorage.removeItem(
+				'wc-blocks_dismissed_compatibility_notices'
+			);
+		} );
```
ralucaStan: You will need to also undo the setting of the fields you are specifically enabling in the before all block, because they might interfere with other test suites.

```
await expect( page ).toClick( 'label', { text: 'Company' } );
		await expect( page ).toClick( 'label', {
			text: 'Apartment, suite, etc.',
		} );
		await expect( page ).toClick( 'label', { text: 'Phone' } );
```

ralucaStan (changes requested):
Huge congrats on this first tests PR. 

I've added some comments. Let me know what you think.

ralucaStan:
> Based on this information, would you still suggest sticking to the `frontend` and `backend` folders, risking that we might have to rename them later if WooCommerce Blocks gets finally merged into core?

I think we could do the renaming in this PR so that we have a shared understanding of the structure.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
ralucaStan: > > Based on this information, would you still suggest sticking to the `frontend` and `backend` folders, risking that we might have to rename them later if WooCommerce Blocks gets finally merged into core?
> 
> I think we could do the renaming in this PR so that we have a shared understanding of the structure.

I thought better about this and I propose to do the renaming of the folders in another PR. This way we keep this PR clean and we have a separate refactoring PR.

So `frontend` -> `shopper` and `backend` -> `merchant`. We can make a separate ticket for it for better visibility, and also explaining the reasoning behind it.

tests/utils/shopper.js
```diff
@@ -47,6 +47,71 @@ export const shopper = {
 		);
 	},
 
+	placeOrder: async () => {
+		await expect( page ).toClick( '.wc-block-components-button__text', {
+			text: 'Place Order',
+		} );
+		page.waitForNavigation( {
```
ralucaStan: await is missing here :)

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Great suggestion. I changed it accordingly.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Good point. The test should indeed check if all provided data appear on the order confirmation page as expected.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Nope, it's not needed and I removed it.

tests/utils/shopper.js
ralucaStan:
```
await page.waitForNavigation( {
			waitUntil: 'networkidle0',
		} );
```

ralucaStan:
```suggestion
		await page.waitForNavigation( {
```

tests/e2e/specs/shopper/checkout-different-addresses.test.js
ralucaStan:
```
		const checkbox = await page.$(
			'.wc-block-checkout__use-address-for-billing .wc-block-components-checkbox__input'
		);
		const isCheckBoxChecked = await (
			await checkbox.getProperty( 'checked' )
		 ).jsonValue();
		expect( isCheckBoxChecked ).toBe( true );
		await expect( page ).toClick(
			'.wc-block-checkout__use-address-for-billing .wc-block-components-checkbox__input'
		);
```

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,111 @@
...
+		);
+		await expect( page ).toClick( 'label', { text: 'Company' } );
+		await expect( page ).toClick( 'label', {
+			text: 'Apartment, suite, etc.',
+		} );
+		await expect( page ).toClick( 'label', { text: 'Phone' } );
+		await saveOrPublish();
+
+		// Get product page permalink.
+		await visitPostOfType( simpleProductName, 'product' );
```
ralucaStan: WC has this [utility](https://www.npmjs.com/package/@woocommerce/e2e-utils#shopper-shopper) `goToProduct`, it's more straight forward and maybe faster to use that instead

tests/utils/shopper.js
ralucaStan: actually let's do it like [WC does it](https://github.com/woocommerce/woocommerce-monorepo/blob/27009cdc0a2af1cca88b88b5ee3600416bab6b7a/plugins/woocommerce/tests/e2e/utils/src/flows/shopper.js#L71-L76), using `await Promise.all`

```
	placeOrder: async () => {
		await Promise.all( [
			text: 'Place Order',
		} ),
			page.waitForNavigation( { waitUntil: 'networkidle0' } ),
		] );
	},
```

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,92 @@
...
+				checkbox.click();
+			}
+		} );
+		await uiUnblocked();
+		// await page.waitForTimeout( 1000 );
+		// await jestPuppeteer.debug();
+		await shopper.block.fillShippingDetails( SHIPPING_DETAILS );
+		await shopper.block.fillBillingDetails( BILLING_DETAILS );
+		await shopper.block.placeOrder();
+		await page.waitForTimeout( 1000 );
```
tarhi-saad: @nielslange, did you experience any flakiness at this step? On my end, the test passed successfully without this line:

```js
await page.waitForTimeout( 1000 );
```

I think `page.waitForNavigation( { waitUntil: 'networkidle0' } )` in the `placeOrder` function is already handling this case.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,92 @@
...
+import {
+	shopper,
+	preventCompatibilityNotice,
+	reactivateCompatibilityNotice,
+} from '../../../utils';
+
+import { BILLING_DETAILS, SHIPPING_DETAILS } from '../../../utils/constants';
+const SIMPLE_PRODUCT_NAME = '128GB USB Stick';
+
+const config = require( 'config' );
```
tarhi-saad: We can remove this line. The `config` variable isn't used by this file.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,92 @@
...
+		await visitBlockPage( 'Checkout Block' );
+		await openDocumentSettingsSidebar();
+		await selectBlockByName(
+			'woocommerce/checkout-shipping-address-block'
+		);
+		await expect( page ).toClick( 'label', { text: 'Company' } );
+		await expect( page ).toClick( 'label', {
+			text: 'Apartment, suite, etc.',
+		} );
+		await expect( page ).toClick( 'label', { text: 'Phone' } );
```
tarhi-saad: If I understand correctly, we want to activate all shipping address fields by toggling on the checkboxes in the image below:

image

And to do that, we are looking for the corresponding label by name then clicking on it. There are two issues with this approach:

- We aren't checking if a field is already active. Initially, only the "Company" field is disabled. The other two fields (i.e., phone & apartment) are already active.
- The **Shipping address** form also contains a label with the same name. (See image below)
image
image
image

This means, that if a field is already active/displayed we are actually clicking on the form instead of the checkboxes.

Since the **"Phone"** and **"Apartment"** fields are active by default, I suggest we only target the **"Company"** field. And, instead of looking for the label, we can use the `setCheckbox` utility function from Woo:

```js
/**
 * External dependencies
 */
import { setCheckbox } from '@woocommerce/e2e-utils';

// Activate the "Company" field
await setCheckbox('#inspector-toggle-control-1');
```

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,92 @@
...
```
tarhi-saad: The same thing as in the `beforeAll` block! The **"Company"** field is not being disabled here. This time let's use the `unsetCheckbox` Woo's utility function:

```js
/**
 * External dependencies
 */
import { unsetCheckbox } from '@woocommerce/e2e-utils';

// Deactivate the "Company" field
await unsetCheckbox('#inspector-toggle-control-1');
```

And no need to deactivate the remaining fields since they were active by default.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,92 @@
...
+		await merchant.logout();
+		await reactivateCompatibilityNotice();
+	} );
+
+	it( 'allows customer to have different shipping and billing addresses', async () => {
+		await shopper.goToShop();
+		await shopper.addToCartFromShopPage( SIMPLE_PRODUCT_NAME );
+		await shopper.block.goToCheckout();
+
+		await page.evaluate( () => {
```
tarhi-saad: We can use the `unsetCheckbox` utility function here as well. So, instead of:

```js
await page.evaluate( () => {
	const checkbox = document.querySelector( '#checkbox-control-0' );
	if ( checkbox.checked ) {
		checkbox.click();
	}
} );
```

We can have:

```js
/**
 * External dependencies
 */
import { unsetCheckbox } from '@woocommerce/e2e-utils';

// Activate the "Company" field
await unsetCheckbox('#checkbox-control-0');
```

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,92 @@
...
+const SIMPLE_PRODUCT_NAME = '128GB USB Stick';
+
+const config = require( 'config' );
+
+if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( 'Skipping Checkout tests', () => {} );
+
+describe( 'Shopper → Checkout → Can have different shipping and billing addresses', () => {
+	beforeAll( async () => {
```
tarhi-saad, line 36: We'll be adding a product to the cart, so let's do the necessary reset before our tests:

```js
beforeAll( async () => {
  await shopper.block.emptyCart();
}
```

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,92 @@
...
+		);
+		await expect( page ).toClick( 'label', { text: 'Company' } );
+		await expect( page ).toClick( 'label', {
+			text: 'Apartment, suite, etc.',
+		} );
+		await expect( page ).toClick( 'label', { text: 'Phone' } );
+		await saveOrPublish();
+	} );
+
+	afterAll( async () => {
```
tarhi-saad, line 59: We've added a product to the cart. To avoid leaving any side effects from our tests, let's remove it:

```js
afterAll( async () => {
  await shopper.block.emptyCart();
}
```

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,92 @@
...
+		await shopper.addToCartFromShopPage( SIMPLE_PRODUCT_NAME );
+		await shopper.block.goToCheckout();
+
+		await page.evaluate( () => {
+			const checkbox = document.querySelector( '#checkbox-control-0' );
+			if ( checkbox.checked ) {
+				checkbox.click();
+			}
+		} );
+		await uiUnblocked();
```
tarhi-saad: We won't need the `await uiUnblocked();` line. We are already awaiting for the checkbox to be unchecked before executing the next action.

tarhi-saad (changes requested):
Thank you, @nielslange, for your excellent work! I left some suggestions below to apply, and we will be good to go! 🙌

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: I did experience some flakiness, but removed that line given that the test passes on your end.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Oh, I must have missed that. I just removed it.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Changed! 😀

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Changed! 😀

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Isn't the cart empty given that we place the order? I decided not to run the empty cart command to save time.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Removed! 😀

tests/e2e/specs/shopper/checkout-different-addresses.test.js
tarhi-saad: @nielslange, this is true if our test is successful. But, if our test fails (e.g., the order didn't go through), our product will not get removed from the cart, which may affect the following tests. Please let me know if you think this is something worth considering. Otherwise, feel free to ignore this suggestion.

nielslange: > @nielslange, this is true if our test is successful. But, if our test fails (e.g., the order didn't go through), our product will not get removed from the cart, which may affect the following tests. Please let me know if you think this is something worth considering. Otherwise, feel free to ignore this suggestion.

I see your point and just added `await shopper.block.emptyCart();` both to `beforeAll()` and `afterAll()`.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Changed! 😀

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Changed! 😀

nielslange:
@tarhi-saad I applied all your suggestions. Do you want to take a final look at this PR?

tarhi-saad (approved):
Wonderfull, @nielslange! And thank you for the great work! Ok, let's 🚀.

tarhi-saad (commented):
@nielslange, the CI has a failing test, but it passes when I run it locally. Here is the [error message](https://github.com/woocommerce/woocommerce-gutenberg-products-block/runs/5662164473?check_suite_focus=true):

```
Summary of all failing tests
FAIL tests/e2e/specs/shopper/checkout-different-addresses.test.js (26.725 s)
  ● Shopper → Checkout → Can have different shipping and billing addresses › allows customer to have different shipping and billing addresses

    TimeoutError: Text not found "Automattic"

      waiting for function failed: timeout 500ms exceeded


Test Suites: 1 failed, 37 passed, 38 total
Tests:       1 failed, 133 passed, 134 total
```

Do you have any idea what could have caused this issue?

tests/e2e/specs/shopper/checkout-different-addresses.test.js
```diff
@@ -0,0 +1,82 @@
+/**
+ * External dependencies
+ */
+import {
+	merchant,
+	openDocumentSettingsSidebar,
+	setCheckbox,
+	unsetCheckbox,
+	uiUnblocked,
```
tarhi-saad: We can remove `uiUnblocked ` since it's unused.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
tarhi-saad: I think we missed this one. I can't see the change. See [comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5860#discussion_r823962062).

tests/utils/shopper.js
```diff
@@ -188,85 +188,78 @@ export const shopper = {
...
-				'#phone',
-				customerBillingDetails.phone
-			);
-			await expect( page ).toFill(
-				'#email',
-				customerBillingDetails.email
-			);
+			await expect( page ).toFill( '#billing-first_name', customerBillingDetails.firstname );
+			await expect( page ).toFill( '#billing-last_name', customerBillingDetails.lastname );
+			await expect( page ).toFill( '#billing-country input', customerBillingDetails.country );
```
tarhi-saad, line 216: @nielslange, this part is causing the test to fail in CI. In the `shopper.verifyBillingDetails` function, we verify if `customerBillingDetails.company` exists on the page, but we didn't fill in the `company` name on the input field in this function. So we should add bellow this line the following code:

```js
await expect( page ).toFill( '#billing-company', customerBillingDetails.company );
```

Note: The test is successful locally and not in CI because the `company` input is already prefilled with the correct name, but in CI, it seems to be empty by default.

tests/utils/shopper.js
```diff
@@ -188,85 +188,78 @@ export const shopper = {
...
+			await expect( page ).toFill( '#billing-postcode', customerBillingDetails.postcode );
+			await expect( page ).toFill( '#phone', customerBillingDetails.phone );
+			await expect( page ).toFill( '#email', customerBillingDetails.email );
+		},
+
+		// prettier-ignore
+		fillShippingDetails: async ( customerShippingDetails ) => {
+			await expect( page ).toFill( '#shipping-first_name', customerShippingDetails.firstname );
+			await expect( page ).toFill( '#shipping-last_name', customerShippingDetails.lastname );
+			await expect( page ).toFill( '#shipping-country input', customerShippingDetails.country );
```
tarhi-saad, line 236: The same thing here:

This part is causing the test to fail in CI. In the `shopper. verifyShippingDetails` function, we verify if `customerShippingDetails.company` exists on the page, but we didn't fill in the `company` name on the input field in this function. So we should add bellow this line the following code:

```js
await expect( page ).toFill( '#shipping-company', customerShippingDetails.company );
```

Note: The test is successful locally and not in CI because the `company` input is already prefilled with the correct name, but in CI, it seems to be empty by default.

tarhi-saad (changes requested):
Hi @nielslange!

Our test was successful locally but failed in CI. After some investigation, it seems to be an issue with verifying the value of the `company` field in those functions `verifyBillingDetails` and `verifyShippingDetails`. You can find more info in the comments below, and you'll also find some additional suggestions as well.

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Done!

tests/e2e/specs/shopper/checkout-different-addresses.test.js
nielslange: Ah, good catch. I changed a similar part in `beforeAll()` and missed the one in `it()`. Just changed this one as well.

tests/utils/shopper.js
nielslange: Done!

tests/utils/shopper.js
nielslange: Done!

nielslange:
> Our test was successful locally but failed in CI. After some investigation, it seems to be an issue with verifying the value of the `company` field in those functions `verifyBillingDetails` and `verifyShippingDetails`. You can find more info in the comments below, and you'll also find some additional suggestions as well.

Good catch, @tarhi-saad. I just applied all changes, but the tests keep failing with the error message that `#billing-company` hasn't been found. Do you have any other explanation what might be causing this?

tarhi-saad:
Hi, @nielslange! Thank you for the changes!

> I just applied all changes, but the tests keep failing with the error message that #billing-company hasn't been found. Do you have any other explanation what might be causing this?

I've done some tests and debugging, and it's getting late. So, I'll continue tomorrow. But, here are some findings:

- The `ID` of our checkbox (i.e., the company checkbox) isn't stable. Its value changes when we refresh the page, toggle it On/Off, etc. I didn't check the code to see when and why it changes precisely, but I've done the test on my local site. So, we should select our input using anything else but the `ID`.
- Adding the following line to the `fillBillingDetails` did break another test that is using this same function (i.e., This is our other failing test: [tests/e2e/specs/shopper/checkout-place-order.test.js](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/e2e/specs/shopper/checkout-place-order.test.js#L17)):

```
await expect( page ).toFill( '#billing-company', customerBillingDetails.company );
``` 

We get an error since it's invoking the same function, and we didn't enable the `company` field in this [test](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/e2e/specs/shopper/checkout-place-order.test.js#L17).

tarhi-saad:
Hi @nielslange! I have pushed two commits to test my findings. Fortunately, the test is now passing!

In this commit, [Fill company if field exist](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5860/commits/ff9e9f67ea8daf39fdb36d58fedd5121964a1b36), I added a condition to fill the company field only when it's enabled.

And the [Get checkbox ID from label](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5860/commits/29e95494b94d3754262edc7ed0b625d7f0421ec8) commit, is to get the correct checkbox ID even if it's unstable (changing on page refresh, on checkbox toggle, etc.).

The failing test is about this file: `tests/e2e/specs/shopper/cart-checkout-translations.test.js`. So, it's not related to our changes. Please, review the latest changes and feel free to 🚢 it!

Hi @alexflorisca, it seems even with the increased timeout, the test of `tests/e2e/specs/shopper/cart-checkout-translations.test.js` does fail in [some cases](https://github.com/woocommerce/woocommerce-gutenberg-products-block/runs/5747598751?check_suite_focus=true):

```
FAIL tests/e2e/specs/shopper/cart-checkout-translations.test.js (33.737 s)
  ● Shopper → Cart → Can view translated cart & checkout blocks › should be able to view translated Cart block 

    TimeoutError: Text not found "Produit"

      waiting for function failed: timeout 2000ms exceeded

  ● Shopper → Cart → Can view translated cart & checkout blocks › should be able to view translated Checkout block

    TimeoutError: Text not found "Coordonnées"

      waiting for function failed: timeout 2000ms exceeded
```

tarhi-saad (approved):

nielslange:
Thanks for your review and the additional changes, @tarhi-saad! I'll merge this PR for now and will keep an eye on the translation e2e tests. If they keep failing, we should definitely look into them in the future.

PR #5867, opened 2022-02-16
Update innerBlock template for Featured Product and Featured Category blocks

### LINKED ISSUE #5676: Featured Product/Category: some button styles don't get applied
Width settings and font size styles don't take effect in the Featured Product/Category inner button.

## To reproduce

1. Add a Featured Product or Featured Category block to a post or page.
2. Click on the inner button.
3. Try changing its Width settings and Typography size.
4. Notice nothing changes. This also affects the frontend.

## Expected behavior
The user should be able to change the width and font size of the button.

## Screenshots

https://user-images.githubusercontent.com/3616980/151771093-c2785b3d-7879-490a-81c9-b211e712e459.mp4

gigitux:
This PR updates the inner block template for Featured Product and Featured Category. This is necessary because a button block should have as the parent the buttons block. In this way, the user is able to customize again:
- font-size
- width settings

**NOTE:** The width is customizable only when the alignment is set to `none`. This is the default behavior on Gutenberg.

Fixes #5676

### Manual Testing

1. Add a Featured Product or Featured Category block to a post or page.
2. Click on the inner button.
3. Try changing the Typography size.
4. Set alignment to none and change the width.
5. Check that the changes are visible on the editor side and frontend side.
### User Facing Testing
* [x] Same as above


tomasztunik:
Nice! This also fixed a bug where `align: center` was ignored in the Editor and the block was misaligned as seen on the screenshot here #5886. 

~While trying to see if this would fix it I've noticed that setting button width (25, 50, 75, 100%) doesn't work and it's now either full width or centred regardless of the option selected~ seen the note

tomasztunik (approved):
👌 looks good

To get the align and width to work we have to set align to `none` and use justify setting to control `left | center | right` position

PR #5870, opened 2022-02-16
Decode HTML entities when formatting Store API error messages

### LINKED ISSUE #5869: HTML entities not decoded when handling failed order payment responses
## Describe the bug

`formatStoreApiErrorMessage` does not decode HTML entities, resulting in inconsistencies like this:

### Cart

Screen Shot 2022-02-16 at 11 28 10 PM

### Checkout

Screen Shot 2022-02-16 at 11 29 08 PM

## To reproduce

1. Add a callback to the `wooocommerce_store_api_validate_cart_item` action and throw a `RouteException` that includes HTML entities.
2. Open the Cart Block.
3. You should see a notice with the message you added to your Exception.
4. Go to the Checkout Block.
5. Attempt to place the order.
6. The HTML entities are not decoded.

## Expected behavior

When I attempt to place the order, I would expect the HTML entities in my exception message to be decoded

## Screenshots

Provided.

manospsyx:
Fixes #5869 

#### Accessibility

#### Other Checks

### Screenshots

See #5869 

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Add a callback to the wooocommerce_store_api_validate_cart_item action and throw a RouteException that includes HTML entities.
2. Open the Cart Block.
3. You should see a notice with the message you added to your Exception.
4. Go to the Checkout Block.
5. Attempt to place the order.
6. You should see the same message, with entities decoded.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### Changelog

> Decode HTML entities when formatting Store API error messages.


assets/js/base/context/providers/store-notices/components/store-notices-container.js
```diff
@@ -50,7 +51,7 @@ const StoreNoticesContainer = ( { className, notices, removeNotice } ) => {
 						}
 					} }
 				>
-					{ props.content }
+					{ decodeEntities( props.content ) }
```
PanosSynetos: Hey @manospsyx - I discussed briefly with @xristos3490 and he suggested using `decodeEntities` in one place (here - the store notices container) rather than everywhere else we call `createNotice`

What do you think?

I tested it locally and it works as expected (with `decodeEntities` in `assets/js/base/utils/errors.js` removed). If you agree, we can revert `errors.js` and just keep the change in the container.

manospsyx: I remember there's at least one instance where `decodeEntities` is used outside -- I wasn't sure if that's intentional. I agree with Chris, but let's make sure that we're not double-decoding in other places. Can you check usage of `decodeEntities` in other locations to make sure?

PanosSynetos: As discussed with @senadir over slack

> My personal preference is to not call something unless needed seeing that you can't opt out of it in a clean way, I think decoding all server returned stings makes more sense than decoding at the createNotice level.

senadir (approved):
:shipit:

PR #5874, opened 2022-02-17
Introduce new cart/order validation hooks

manospsyx:
While diving deeper into Cart/Checkout Error Handling, we found that

- The Store API provides an <code>errors</code> field as the "canonical" way to pass cart errors to front-end consumers.
- The presence of <code>errors</code> does not prevent access to the contents of the Checkout Block. To prevent access to the Checkout Block, developers must throw <a href="https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1c478e60acb805733be04c5ca4c9f2d2443f7551/src/StoreApi/Routes/Checkout.php#L122">[an Exception here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1c478e60acb805733be04c5ca4c9f2d2443f7551/src/StoreApi/Routes/Checkout.php#L122)</a>. 
- If <code>errors</code> have been added to the cart response via the <code>wooocommerce_store_api_validate_cart_item</code> <a href="https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1c478e60acb805733be04c5ca4c9f2d2443f7551/src/StoreApi/Utilities/CartController.php#L577">[action](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1c478e60acb805733be04c5ca4c9f2d2443f7551/src/StoreApi/Utilities/CartController.php#L577)</a>, they are not rendered in the Checkout block.

So far so good! This is a great architecture that gives developers a lot of flexibility to:

- determine if their own errors should prevent access to the checkout block;
- decide how to display/consume `errors` in the checkout block.

To better serve extensibility needs:

- Developers will need a way to add cart `errors` that replaces the legacy `woocommerce_check_cart_items` hook. Such a way does not seem to exist right now.
- We need to make it clear to developers that the presence of `errors` does not automatically prevent access to the Checkout block. Going forward, they are responsible for determining if their errors should trigger this behavior in the front-end. On our side, we need to make this clearer in the code, and ideally also provide a hook for this purpose.

This PR:

1. Introduces a new `wooocommerce_store_api_validate_cart` action, which can be used by third party developers to add extra 'errors' in cart route responses. This is meant to function as a replacement of the legacy `woocommerce_check_cart_items` hook.
2. Introduces a new `wooocommerce_blocks_draft_order_updated` action. This new action can be used to modify the finalized draft order object. Additionally, developers can use it to throw a RouteException in order to prevent access to the checkout block, instead of the `woocommerce_blocks_checkout_update_order_meta` (naming more aligned with intent!).

Finally the PR also shuffles some existing code that checks the order instance and sets the draft order ID before the `woocommerce_blocks_checkout_update_order_meta` gets fired.

#### Accessibility

#### Other Checks

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests

### User Facing Testing

None.

### Changelog

> Introduced `wooocommerce_store_api_validate_cart` and `wooocommerce_blocks_draft_order_updated` actions.

### DIFF
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -310,18 +310,6 @@ class Checkout extends AbstractCartRoute {
 			$this->order_controller->update_order_from_cart( $this->order );
 		}
 
-		// Confirm order is valid before proceeding further.
-		if ( ! $this->order instanceof \WC_Order ) {
-			throw new RouteException(
-				'woocommerce_rest_checkout_missing_order',
-				__( 'Unable to create order', 'woo-gutenberg-products-block' ),
-				500
-			);
-		}
-
-		// Store order ID to session.
-		$this->set_draft_order_id( $this->order->get_id() );
-
 		/**
 		 * Fires when the Checkout Block/Store API updates an order's meta data.
 		 *
@@ -365,6 +353,18 @@ class Checkout extends AbstractCartRoute {
 		 */
 		do_action( 'woocommerce_blocks_checkout_update_order_meta', $this->order );
 
+		// Confirm order is valid before proceeding further.
+		if ( ! $this->order instanceof \WC_Order ) {
+			throw new RouteException(
+				'woocommerce_rest_checkout_missing_order',
+				__( 'Unable to create order', 'woo-gutenberg-products-block' ),
+				500
+			);
+		}
+
+		// Store order ID to session.
+		$this->set_draft_order_id( $this->order->get_id() );
+
 		/**
 		 * Try to reserve stock for the order.
 		 *
@@ -382,14 +382,6 @@ class Checkout extends AbstractCartRoute {
 				$e->getCode()
 			);
 		}
-
-		/**
-		 * Fire action after draft order creation/update.
-		 * Functions hooking into this can throw a \RouteException to force the checkout block into an error state that prevents access to its contents.
-		 *
-		 * @param \WC_Order $order Order object.
-		 */
-		do_action( 'wooocommerce_blocks_draft_order_updated', $this->order );
 	}
 
 	/**
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -310,6 +310,18 @@ class Checkout extends AbstractCartRoute {
 			$this->order_controller->update_order_from_cart( $this->order );
 		}
 
+		// Confirm order is valid before proceeding further.
+		if ( ! $this->order instanceof \WC_Order ) {
+			throw new RouteException(
+				'woocommerce_rest_checkout_missing_order',
+				__( 'Unable to create order', 'woo-gutenberg-products-block' ),
+				500
+			);
+		}
+
+		// Store order ID to session.
+		$this->set_draft_order_id( $this->order->get_id() );
+
 		/**
 		 * Fires when the Checkout Block/Store API updates an order's meta data.
 		 *
@@ -353,18 +365,6 @@ class Checkout extends AbstractCartRoute {
 		 */
 		do_action( 'woocommerce_blocks_checkout_update_order_meta', $this->order );
 
-		// Confirm order is valid before proceeding further.
-		if ( ! $this->order instanceof \WC_Order ) {
-			throw new RouteException(
-				'woocommerce_rest_checkout_missing_order',
-				__( 'Unable to create order', 'woo-gutenberg-products-block' ),
-				500
-			);
-		}
-
-		// Store order ID to session.
-		$this->set_draft_order_id( $this->order->get_id() );
-
 		/**
 		 * Try to reserve stock for the order.
 		 *
@@ -382,6 +382,14 @@ class Checkout extends AbstractCartRoute {
 				$e->getCode()
 			);
 		}
+
+		/**
+		 * Fire action after draft order creation/update.
+		 * Functions hooking into this can throw a \RouteException to force the checkout block into an error state that prevents access to its contents.
+		 *
+		 * @param \WC_Order $order Order object.
+		 */
+		do_action( 'wooocommerce_blocks_draft_order_updated', $this->order );
 	}
 
 	/**
--- a/src/StoreApi/Routes/Checkout.php
+++ b/src/StoreApi/Routes/Checkout.php
@@ -341,6 +341,7 @@ class Checkout extends AbstractCartRoute {
 		 * Fires when the Checkout Block/Store API updates an order's meta data.
 		 *
 		 * This hook gives extensions the chance to add or update meta data on the $order.
+		 * Throwing an exception from a callback attached to this action will make the Checkout Block render in a warning state, effectively preventing checkout.
 		 *
 		 * This is similar to existing core hook woocommerce_checkout_update_order_meta.
 		 * We're using a new action:
--- a/src/StoreApi/Schemas/CartSchema.php
+++ b/src/StoreApi/Schemas/CartSchema.php
@@ -411,6 +411,7 @@ class CartSchema extends AbstractSchema {
 		try {
 			/**
 			 * Fire action to validate cart. Functions hooking into this should throw a \RouteException.
+			 * @example See docs/examples/validate-cart.md
 			 *
 			 * @param \WC_Cart $cart Cart object.
 			 */
--- a/src/StoreApi/Schemas/CartSchema.php
+++ b/src/StoreApi/Schemas/CartSchema.php
@@ -419,7 +419,8 @@ class CartSchema extends AbstractSchema {
 			$cart_errors[] = new WP_Error( $error->getErrorCode(), $error->getMessage() );
 		}
 
-		$cart_errors = array_filter( array_merge( $cart_errors, $item_errors, $coupon_errors ),
+		$cart_errors = array_filter(
+			array_merge( $cart_errors, $item_errors, $coupon_errors ),
 			function ( WP_Error $error ) {
 				return $error->has_errors();
 			}
--- a/src/StoreApi/Schemas/CartSchema.php
+++ b/src/StoreApi/Schemas/CartSchema.php
@@ -414,7 +414,7 @@ class CartSchema extends AbstractSchema {
 			 *
 			 * @param \WC_Cart $cart Cart object.
 			 */
-			do_action( 'wooocommerce_store_api_validate_cart', $cart );
+			do_action( '__experimental_wooocommerce_store_api_validate_cart', $cart );
 		} catch ( RouteException $error ) {
 			$cart_errors[] = new WP_Error( $error->getErrorCode(), $error->getMessage() );
 		}
--- a/src/StoreApi/Schemas/CartSchema.php
+++ b/src/StoreApi/Schemas/CartSchema.php
@@ -414,7 +414,7 @@ class CartSchema extends AbstractSchema {
 			 *
 			 * @param \WC_Cart $cart Cart object.
 			 */
-			do_action( '__experimental_wooocommerce_store_api_validate_cart', $cart );
+			do_action( '__experimental_woocommerce_store_api_validate_cart', $cart );
 		} catch ( RouteException $error ) {
 			$cart_errors[] = new WP_Error( $error->getErrorCode(), $error->getMessage() );
 		}
--- a/src/StoreApi/Schemas/CartSchema.php
+++ b/src/StoreApi/Schemas/CartSchema.php
@@ -404,14 +404,27 @@ class CartSchema extends AbstractSchema {
 	 */
 	protected function get_cart_errors( $cart ) {
 		$controller    = new CartController();
-		$item_errors   = array_filter(
-			$controller->get_cart_item_errors(),
+		$cart_errors   = [];
+		$item_errors   = $controller->get_cart_item_errors();
+		$coupon_errors = $controller->get_cart_coupon_errors();
+
+		try {
+			/**
+			 * Fire action to validate cart. Functions hooking into this should throw a \RouteException.
+			 *
+			 * @param \WC_Cart $cart Cart object.
+			 */
+			do_action( 'wooocommerce_store_api_validate_cart', $cart );
+		} catch ( RouteException $error ) {
+			$cart_errors[] = new WP_Error( $error->getErrorCode(), $error->getMessage() );
+		}
+
+		$cart_errors = array_filter( array_merge( $cart_errors, $item_errors, $coupon_errors ),
 			function ( WP_Error $error ) {
 				return $error->has_errors();
 			}
 		);
-		$coupon_errors = $controller->get_cart_coupon_errors();
 
-		return array_values( array_map( [ $this->error_schema, 'get_item_response' ], array_merge( $item_errors, $coupon_errors ) ) );
+		return array_values( array_map( [ $this->error_schema, 'get_item_response' ], $cart_errors ) );
 	}
 }
--- a/docs/examples/validate-cart.md
+++ b/docs/examples/validate-cart.md
@@ -1,0 +1,19 @@
+```php
+// The action callback function.
+function my_function_callback( $cart ) {
+  // Validate the $cart object and throw a RouteException. For example, to create an error if the cart contains more than 10 items:
+  if ( $cart->get_cart_contents_count() > 10 ) {
+    throw new Automattic\WooCommerce\Blocks\StoreApi\Routes\RouteException( 'my_exception_code', 'Too many cart items!' );
+  }
+}
+
+add_action( '__experimental_woocommerce_store_api_validate_cart', 'my_function_callback', 10 );
+```
+<!-- FEEDBACK -->
+---
+
+[We're hiring!](https://woocommerce.com/careers/) Come work with us!
+
+🐞 Found a mistake, or have a suggestion? [Leave feedback about this document here.](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/new?assignees=&labels=type%3A+documentation&template=--doc-feedback.md&title=Feedback%20on%20./docs/examples/checkout-order-processed.md)
+<!-- /FEEDBACK -->
+


mikejolley:
> To prevent access to the Checkout Block, developers must throw an Exception

This is likely just a side effect of pre-loading the draft order on checkout entry, rather than a way specifically provided to prevent access to the block. 

Checkout validation is handled when the order is placed. Not before the block is rendered. That's likely the main issue here. The big question is, why is the cart in an erroneous state? Can the cart be "fixed" before entry to checkout? Doing all validation during order creation is probably preferred if that's not possible.

> The Store API provides an errors field as the "canonical" way to pass cart errors to front-end consumers.

Yes, as a response to an event. e.g. adding to cart, or placing an order.

> If errors have been added to the cart response via the wooocommerce_store_api_validate_cart_item [action](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1c478e60acb805733be04c5ca4c9f2d2443f7551/src/StoreApi/Utilities/CartController.php#L577), they are not rendered in the Checkout block.

Sounds legit. Errors are surfaced on events and I think the checkout only handles events such as placing order, updating customer data, and selecting shipping methods.

src/StoreApi/Routes/Checkout.php
```diff
@@ -310,6 +310,18 @@ private function create_or_update_draft_order( \WP_REST_Request $request ) {
 			$this->order_controller->update_order_from_cart( $this->order );
 		}
 
+		// Confirm order is valid before proceeding further.
```
mikejolley: Handling for no order should be done by the `woocommerce_blocks_checkout_update_order_meta`. Swapping these conditionals around could mean that someone replaces the order object via `woocommerce_blocks_checkout_update_order_meta` leaving the API in an error state.

src/StoreApi/Routes/Checkout.php
```diff
@@ -382,6 +382,14 @@ private function create_or_update_draft_order( \WP_REST_Request $request ) {
...
 			);
 		}
+
+		/**
+		 * Fire action after draft order creation/update.
+		 * Functions hooking into this can throw a \RouteException to force the checkout block into an error state that prevents access to its contents.
+		 *
+		 * @param \WC_Order $order Order object.
+		 */
+		do_action( 'wooocommerce_blocks_draft_order_updated', $this->order );
```
mikejolley: Can you explain why this is needed specifically? Is the hook above not useful?

If we do add a new hook it needs to be added with `_experimental` prefix to validate its usefulness before committing to it.

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -404,14 +404,28 @@ protected function get_tax_lines( $cart ) {
...
+		$item_errors   = $controller->get_cart_item_errors();
+		$coupon_errors = $controller->get_cart_coupon_errors();
+
+		try {
+			/**
+			 * Fire action to validate cart. Functions hooking into this should throw a \RouteException.
+			 *
+			 * @param \WC_Cart $cart Cart object.
+			 */
+			do_action( 'wooocommerce_store_api_validate_cart', $cart );
```
mikejolley: I see checkout route calls  `woocommerce_check_cart_items` which can also be used to add notices and have them converted to exceptions. Is that more suitable?

I also see we already have a `wooocommerce_store_api_validate_cart_item` action which puts error messages into this array. Why is a new hook needed which does the same thing?

manospsyx:
Hey @mikejolley 

> Checkout validation is handled when the order is placed. Not before the block is rendered. That's likely the main issue here. The big question is, why is the cart in an erroneous state? Can the cart be "fixed" before entry to checkout? Doing all validation during order creation is probably preferred if that's not possible.

From an extensibility point of view, we'll want to provide a way to third party code to:

**(A) Introduce `errors` in cart responses, before an attempt is made to place an order**

The rationale behind this is that, from a UX point of view, it makes sense to display notices related with the cart state/session as early as possible, rather than to expect users to click the Place Order button to receive that feedback. 

Currently, the cart block consumes these `errors` and converts them directly to notices, which means that, at least in cart block context, we have a working mechanism to convert errors to notices.

> I see checkout route calls `woocommerce_check_cart_items` which can also be used to add notices and have them converted to exceptions. Is that more suitable?

There is currently no way to inject `errors` in the cart route response by hooking into this action. The checkout route is only used:

- once when hydrating the checkout block (and creating a draft order);
- then, only when clicking the place order button.

All actions that happen inside the checkout block trigger a response that follows the cart schema. Therefore, to provide an always-up-to-date set of `errors`, extension code will need to hook in somewhere downstream from `get_cart_errors` [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/ddf77b02d7d39f814e9c13c6464845d676ec4647/src/StoreApi/Schemas/CartSchema.php#L405).

The `woocommerce_check_cart_items` action is not fired there.

> I also see we already have a `wooocommerce_store_api_validate_cart_item` action which puts error messages into this array. Why is a new hook needed which does the same thing?

Validation needs go beyond cart items -- core has coupons, for example, but extension code may have other requirements. For example, extensions may need to validate the selected shipping address against some business logic, or validate the selected shipping method in relation to something else in the session.

One _could_ hook into this action once to do these things, but it might be a better idea to provide a formal, suitably named hook for these common validation needs, instead of encouraging third parties to hook into an action meant for cart item validation.

Then:

> This is likely just a side effect of pre-loading the draft order on checkout entry, rather than a way specifically provided to prevent access to the block.

**(B) Prevent entry to the checkout block if certain errors are present**

> Errors are surfaced on events and I think the checkout only handles events such as placing order, updating customer data, and selecting shipping methods.

Errors in the API are associated with cart/session state, not to isolated events.

How these errors are displayed in the front-end is up for the front end to decide. The cart block renders them at the top of the page, for instance. 

The checkout block does not render them at all, and leaves it up to developers to decide how to consume the `errors` in the cart response. I like that approach, especially given that the checkout block/flow may support multiple error contexts soon, or eventually be broken up into multiple steps. 

At the same time, we need to encourage the use of a specific pattern for blocking access to the checkout block. Some types of errors may need to be resolved before users are allowed to enter the checkout page. We can't know exactly which ones; this is up for developers to determine. What we need to do is to provide that mechanism.

Right now, the only way to do this is to throw an Exception from `woocommerce_blocks_checkout_update_order_meta`.

I generally believe that naming should go hand in hand with intended use, not only to properly guide developers now, but also to prevent future changes that introduce regressions ("oh, but our intention with this hook was never to do X/Y").

If `woocommerce_blocks_checkout_update_order_meta` is the suggested way to prevent entry to the checkout block, we may want to document that "third party developers can throw an Exception here to prevent access to the checkout block".

Finally to complete the picture when it comes to extensibility needs that fall under the "Cart/Checkout Error Handling" umbrella, we'll want to:

**(C) Make it possible for third party code to render notices in the checkout block, at different contexts**

Assuming that we don't want to respond to the presence of `errors` (in the cart response) inside the Checkout Block, we need to provide a mechanism to 3PDs to intercept them + render them at different notice contexts (e.g. "top", "billing details", "shipping details", "shipping options", "payment options").

This problem is beyond the scope of this PR, so I will not go into details here.

and

**(D) Validate (and respond to) attempts to place an order**

Ideally, primarily for performance reasons, the checkout block should be able to understand if notices are rendered in a context, and prevent payment attempts (perhaps scroll the viewport to the first populated notice area).

If this is impossible, third party code should be able to hook into the lifecycle of "placing an order" (evenEmitters?), prevent payment, and "manually" handle things from there.

The validation logic that exists on the Store API side (checkout route) is much-needed, but should be the last resort. For those extensibility needs, we already have `woocommerce_blocks_checkout_order_processed`.

src/StoreApi/Routes/Checkout.php
manospsyx: Thanks Mike! 

This should be covered [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5874#issuecomment-1042955544).

src/StoreApi/Schemas/CartSchema.php
manospsyx: Covered [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5874#issuecomment-1042955544) as well.

`woocommerce_check_cart_items` is only fired during a payment attempt -- it is not used to populate the `errors` field in the cart route response. For more context, check out the comment above.

src/StoreApi/Routes/Checkout.php
manospsyx: >  could mean that someone replaces the order object

Can you share an example? Could a callback replace the object passed into it as an argument and have that change persist (passing by reference?).

If yes, then I see your point here!

mikejolley:
> The rationale behind this is that, from a UX point of view, it makes sense to display notices related with the cart state/session as early as possible, rather than to expect users to click the Place Order button to receive that feedback

That's fine. My question is, why is the cart in an erroneous state? Why were the errors not handled during the previous add to cart attempts, or when loading cart from the session? If there are valid cases where this may happen, yeah, sure we can add some mechanism before checkout block render to surface the errors. But that's still not ideal because a customer has to go back to the cart page to fix them. 

> Validation needs go beyond cart items -- core has coupons, for example, but extension code may have other requirements. For example, extensions may need to validate the selected shipping address against some business logic, or validate the selected shipping method in relation to something else in the session.
> 
> One could hook into this action once to do these things, but it might be a better idea to provide a formal, suitably named hook for these common validation needs, instead of encouraging third parties to hook into an action meant for cart item validation.

and

> I generally believe that naming should go hand in hand with the intended use, not only to properly guide developers now, but also to prevent future changes that introduce regressions ("oh, but our intention with this hook was never to do X/Y").

The premise of this PR is that this validation is currently not possible? It seems like it may be possible, just using an action/filter hook with a non-ideal name. It would be better to use existing actions to validate approaches are fit for purpose before introducing more hooks.

If you can use existing actions first to verify the approach, then we can follow up with **_experimental** action hooks in other places afterwards. This gives a little more time to verify everything that is needed is in place, naming is sufficient, and hooks are working before committing to a public hook which is much harder to remove once in place. 

As for hook naming, rather than basing them on "intended use", we should probably base them on what Store API is doing or make them more generic because we'll never be able to predict all possible use-case. So a hook name around validating items is fine if that's what Store API is currently doing.

I won't touch on points C and D because we're going off-topic. Those would be client-side emitters out of scope for API.

mikejolley:
FYI docs for experimental interfaces are here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/1db29ebd2d5cf541b547af5c60b6a00cc51bdb52/docs/blocks/feature-flags-and-experimental-interfaces.md#usages-of-__experimental-prefix So any new actions will probably start with the prefix before being graduated to stable.

manospsyx:
> My question is, why is the cart in an erroneous state? Why were the errors not handled during the previous add to cart attempts, or when loading cart from the session?

There are a few cases I can think of:

1. Something changed after an item was added to the cart, e.g. stock may have gone down, or perhaps store admins may change a product setting that brings a cart item to an error state, like min/max quantity.
2. Different types of business logic may depend on conditions that introduce errors after an add-to-cart action, for example time/date might be such an independent condition.
3. Validation is often used as a means of exposing what's possible to customers, in a way that encourages certain action/behavior.

An example of (3):

"We currently do not ship Batteries to non-continental US states"; or
"Free Shipping is only available to Gold Members".

Conditional Shipping and Payments is a classic example of a plugin that implements this type of validation/messaging for 

- Shipping Destinations
- Shipping Method selections
- Payment Method selections

In cart/checkout pages, validation messages may have an informational role -- but they always have to prevent order payment.

> The premise of this PR is that this validation is currently not possible? It seems like it may be possible, just using an action/filter hook with a non-ideal name.

This PR focuses on the Store API side of the equation -- it aims to slightly improve things for developers by introducing **and documenting**:

- a hook that extensions can use to create `errors` -- `wooocommerce_store_api_validate_cart`;
- a hook that extensions can use to prevent the checkout block from loading -- we have `woocommerce_blocks_checkout_update_order_meta` but its name does not encourage developers to use it for a purpose other than updating order meta, and it doesn't help us maintain a stable mechanism for fulfilling that need.

src/StoreApi/Routes/Checkout.php
mikejolley: Yeah I believe these objects are passed by reference, so this guard is in place to prevent the route trying to format an invalid order object.

manospsyx:
@mikejolley I have reverted changes to the Checkout route and added some extra bits of information in the `woocommerce_blocks_checkout_update_order_meta` docblock.

I also added an `__experimental` prefix to the `wooocommerce_store_api_validate_cart` action. 

Edit: ...and fixed the triple `ooo`.

manospsyx:
Some more context on:

> "We currently do not ship Batteries to non-continental US states"; or
> "Free Shipping is only available to Gold Members".

From an API point of view, these are all cart states with `errors`. Whoever the consumer of the API might be, they need to be aware that the specific cart state is erroneous.

> I won't touch on points C and D because we're going off-topic. Those would be client-side emitters out of scope for API.

Definitely! The block-based front-end always renders these errors in the cart block, but also gives developers a lot of flexibility to handle them in different ways in the checkout block, without necessarily blocking access to it 👌  We are in the process of exploring that flexibility :)

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -404,14 +404,28 @@ protected function get_tax_lines( $cart ) {
...
+		$item_errors   = $controller->get_cart_item_errors();
+		$coupon_errors = $controller->get_cart_coupon_errors();
+
+		try {
+			/**
+			 * Fire action to validate cart. Functions hooking into this should throw a \RouteException.
+			 *
+			 * @param \WC_Cart $cart Cart object.
+			 */
+			do_action( '__experimental_wooocommerce_store_api_validate_cart', $cart );
```
mikejolley: Do you think it's better to use exceptions here, or should we look at a WP_Error based approach?

If we use exceptions, I wonder if it would be worth passing the exception class/namespace down so you can throw something without knowing the full RouteException namespace. Not something I've tried before, thinking out loud.

src/StoreApi/Schemas/CartSchema.php
```diff
@@ -404,14 +404,28 @@ protected function get_tax_lines( $cart ) {
...
 		$controller    = new CartController();
-		$item_errors   = array_filter(
-			$controller->get_cart_item_errors(),
+		$cart_errors   = [];
+		$item_errors   = $controller->get_cart_item_errors();
+		$coupon_errors = $controller->get_cart_coupon_errors();
+
+		try {
+			/**
+			 * Fire action to validate cart. Functions hooking into this should throw a \RouteException.
```
mikejolley, line 413: We could also put an example here. This will be pulled into the [docs here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/actions.md). There is special docblock syntax for defining examples, see https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/actions.md#example

src/StoreApi/Schemas/CartSchema.php
manospsyx: I like exceptions, but I have the impression that there's no fundamental difference between a `RouteException` and a plain old `Exception` from a third-party developer's point of view. 

In places where we catch exceptions and convert them to `WP_Error` objects, it probably makes little sense to require folks to throw a `RouteException`. 

My intent here was to be consistent with how `CartController::validate_cart_item` throws `RouteException`. There, you can see that when `validate_cart_item` is called via `CartSchema::get_cart_errors`, exceptions are converted to `WP_Errors` again, so again we could catch plain Exceptions -- if it wasn't for the fact that `CartController::validate_cart_item` is also called via `CartController::validate_cart_items`, which does not catch `RouteException`.

I can see that:

- When we populate the cart `errors` field via `CartSchema::get_cart_errors`, all exceptions are caught and converted to `WP_Errors`, and therefore a plain `Exception` would do, for both actions -- `__experimental_wooocommerce_store_api_validate_cart` and `wooocommerce_store_api_validate_cart_item`.
- However, the `validate_cart_item` method is also called when [validating order payment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/add/cart-order-validation-hooks/src/StoreApi/Utilities/CartController.php#L374) -- and there, we expect that any `RouteException` thrown via the `wooocommerce_store_api_validate_cart_item` action will "bubble up". What I don't like about it is that order payment will halt when the first exception is thrown. Ideally, I would probably catch all exceptions thrown via `CartController::validate_cart_item`, convert them to `WP_Error`, and include all of them inside an `InvalidCartException` instead of the current `InvalidStockLevelsInCartException` (too specific) -- then catch that downstream.

I would also argue that `__experimental_wooocommerce_store_api_validate_cart` needs to fire when [validating order payment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/796bce91d0192ec18c3b430a5683654010240111/src/StoreApi/Routes/Checkout.php#L154), too.

Initially, before submitting the PR, I had created a `validate_cart` method and included `validate_cart_items` and `validate_cart_coupons` in it, along with the `do_action( '__experimental_wooocommerce_store_api_validate_cart' )`. However, I reverted all of that work as I was worried that it would widen the scope of the PR too much.

src/StoreApi/Schemas/CartSchema.php
manospsyx: Done! Let me know if I did it right 👍 

PS: Love `npm run build:docs`.

src/StoreApi/Schemas/CartSchema.php
xristos3490: Hey folks,

I have been thinking about this error-handling architecture. 

There might be a case that a cart has more than _one_ error. This case would probably require a different process than throwing a fairly generic exception like the `RouteException`, which can only include a single string as the error message. Introducing a new `InvalidCartException` that will replace the `InvalidStockLevelsInCartException`, allows us to handle it through this new experimental filter, too. 

For example, when someone uses the `__experimental_woocommerce_store_api_validate_cart`, they could gather all errors produced in the cart and add them in a single `WP_Error`. Then throw this error object using this newly created `InvalidCartException`.

A generic Exception such as the `InvalidCartException` will allow us to control all errors produced in the cart by a single extension. For example, CSP may face more than one error in the same cart based on potential complex business restrictions. I think this is the best time to deal with multiple errors in a single cart state and that this consideration will further expand the maturity of the codebase. 

Lastly, It seems that the only errors considered "critical" and will stop the checking out process are all stock-related ones, which makes sense, but we could follow a more comprehensive approach.

I could make a commit to demonstrate this architecture with example usage.

Thoughts?

manospsyx: @xristos3490 let's see what Mike thinks -- I also think that the code could benefit from a slight refactor in part for the reasons mentioned by Mike, but also to introduce parity between error handling for populating the `errors` field vs validating order payment. It's possible that the code here could be made a bit more dry, too.

xristos3490: Sounds good, @manospsyx! I'll work on a separate branch based on this one and refactor some bits. I will update this thread.

manospsyx:
@mikejolley [this builds upon our discussion](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5904) in this PR. It should give you a more complete picture of the potential improvements we've discussed with @xristos3490 around error handling and validation.

mikejolley:
Closing in favour of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5904#pullrequestreview-891004850

PR #5719, opened 2022-02-04
Add a Remove Image option to the Featured Category block

tjcafferkey:
Adds a "Remove" image option to the Featured Category block to remove a previously selected image.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5309

#### Accessibility

- [x] I've tested using only a keyboard (no mouse)
- [ ] I've tested using a screen reader
- [ ] All animations respect [`prefers-reduced-motion`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion)
- [x] All text has [at least a 4.5 color contrast with its background](https://webaim.org/resources/contrastchecker/)

#### Other Checks

- [ ] I've updated [this doc](../docs/blocks/feature-flags-and-experimental-interfaces.md) for any feature flags or experimental interfaces implemented in this pull request.
- [ ] I tagged two reviewers because this PR makes queries to the database or I think it might have some security impact.

### Screenshots

https://user-images.githubusercontent.com/8639742/152528706-0d978094-9d41-4127-854c-dc8ad84cb86d.mp4

#### Automated Tests
* [ ] Changes in this PR are covered by Automated Tests.
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

How to test the changes in this Pull Request:

1. Go to post or page
2. Add Featured Category block and add an image to it
2. Check the "Remove" button that appears in the toolbar
3. Click the "Remove" button and check it removes the image, save block and check frontend.

### User Facing Testing
These are steps for user testing (where "user" is someone interacting with this change that is not editing any code).
* [x] Same as above, or
* [ ] See steps below.

1.
2.
3.

### Changelog

> Add a remove image button to the WooCommerce Feature Category block

PR #5886, opened 2022-02-17
Add to Featured Product block option to remove custom image

### LINKED ISSUE #5725: Image on Featured Product block cannot be removed
## Describe the bug
While it's possible to change an image when using the Featured Product block, it's not possible to remove once added and leave the default background.

## To reproduce
1. Add a Featured Product block.
2. Add an image to the block.
3. See how there's not an option to remove it.

## Expected behavior
There's no button to remove it, just to switch it. We'd want an option to remove the image.

## Rellevant links
* Same issue for the Featured Category block:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5309
* PR that fixed it in the Featured Category block (can be used as a reference for this issue fix):
https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5719

tomasztunik:
This adds ability to remove custom image from Featured Product block. Removing custom image will reset it back to the default product image if available. This follows solution from #5719 which added this option to Featured Category block.

Fixes #5725

#### Accessibility

**UPDATE**: "Remove" label replaced with "Reset"
### Screenshots
Screenshot 2022-02-18 at 00 02 37

### Manual Testing

1. Add a Featured Product block to a page, post, or template.
2. Add custom background media.
3. The button to remove the image should be available.
4. If clicked it should remove the custom media.
5. If the product had an image it should display it again.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.
### Note

misaligned "Shop Now" is an unrelated issue

### Changelog

> Featured Product block: add the ability to reset to a previously set custom background image.

### DIFF
--- a/assets/js/blocks/featured-product/block.js
+++ b/assets/js/blocks/featured-product/block.js
@@ -31,6 +31,7 @@ import PropTypes from 'prop-types';
 import { getSetting } from '@woocommerce/settings';
 import ProductControl from '@woocommerce/editor-components/product-control';
 import ErrorPlaceholder from '@woocommerce/editor-components/error-placeholder';
+import TextToolbarButton from '@woocommerce/editor-components/text-toolbar-button';
 import { withProduct } from '@woocommerce/block-hocs';
 import { Icon, starEmpty } from '@wordpress/icons';
 
@@ -138,19 +139,29 @@ const FeaturedProduct = ( {
 						setAttributes( { contentAlign: nextAlign } );
 					} }
 				/>
-				<MediaReplaceFlow
-					mediaId={ mediaId }
-					mediaURL={ mediaSrc }
-					accept="image/*"
-					onSelect={ ( media ) => {
-						setAttributes( {
-							mediaId: media.id,
-							mediaSrc: media.url,
-						} );
-					} }
-					allowedTypes={ [ 'image' ] }
-				/>
-
+				<ToolbarGroup>
+					<MediaReplaceFlow
+						mediaId={ mediaId }
+						mediaURL={ mediaSrc }
+						accept="image/*"
+						onSelect={ ( media ) => {
+							setAttributes( {
+								mediaId: media.id,
+								mediaSrc: media.url,
+							} );
+						} }
+						allowedTypes={ [ 'image' ] }
+					/>
+					{ mediaId && mediaSrc ? (
+						<TextToolbarButton
+							onClick={ () =>
+								setAttributes( { mediaId: 0, mediaSrc: '' } )
+							}
+						>
+							{ __( 'Remove', 'woo-gutenberg-products-block' ) }
+						</TextToolbarButton>
+					) : null }
+				</ToolbarGroup>
 				<ToolbarGroup
 					controls={ [
 						{
--- a/assets/js/blocks/featured-product/block.js
+++ b/assets/js/blocks/featured-product/block.js
@@ -158,7 +158,7 @@ const FeaturedProduct = ( {
 								setAttributes( { mediaId: 0, mediaSrc: '' } )
 							}
 						>
-							{ __( 'Remove', 'woo-gutenberg-products-block' ) }
+							{ __( 'Reset', 'woo-gutenberg-products-block' ) }
 						</TextToolbarButton>
 					) : null }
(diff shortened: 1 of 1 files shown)


dinhtungdu (commented):
@tomasztunik This is working great on my end. I have two minor issues, otherwise, this is looking good to me.

- For the button label, I think `Reset` is better to describe the function of that button than `Remove`. When users click that button, if the product has a featured image, the button doesn't really "remove" the image from the users' perspective. If the label of the button is `Remove custom featured image`, then using `Remove` makes sense, but it'd be too long.
- For the buttons group containing only icons, we don't need a separator between buttons, but for buttons with text labels, I think a separator should be used to distinguish buttons.

image

tomasztunik:
Valid points @dinhtungdu. Tried to just implement existing flow but maybe we could improve this and Featured Category blocks.

With the separator I think it wasn't added there to highlight that these two actions are related to the same background image selector context. Not sure what is the definitive way to go here.

In general I feel like the dominating "Replace" button especially if category on Featured Category block or Product is selected these seem to refer to the context of a block so feature or product rather than the image. 

Maybe we should use icons there? Went with undo as trash felt too destructive and "remove" like a stop sign :) 

Screenshot 2022-02-18 at 12 35 09

Screenshot 2022-02-18 at 12 33 51

VS current with text labels

Screenshot 2022-02-18 at 12 38 43

@vivialice what do you think? Should we stick to current solution? go with icons? else?

(Used Featured Category block as ref as I had it at hand, but they follow the same patterns)

vivialice:
Hey team, I like these solutions and I think it makes sense to make that portion of the toolbar more useful. I like the icon solution but I think it could be a safer bet to follow what Gutenberg have in image blocks and keep `replace` rather than the picture icon for consistency. I like the `reset` arrow icon but since we are adding something "new" we might need to check with Gutenberg if that is okay. I agree we should group these two functions in the toolbar. 

Another idea:
Screenshot 2022-02-18 at 12 34 15
This is from the Cover block. There is a `clear media` CTA in the sidebar settings but we could change the label to `reset changes`.

tomasztunik:
Updated with "Reset" label and retained no buttons separation to keep the context of working with background media. I've seen this mentioned elsewhere that it would be nice if these were more unified with Cover block and I think any bigger changes for Categories and Product blocks should probably include this consideration. The UI on editing cover block is more clear and more user friendly and provides more options in the sidebar than we have available here.

dinhtungdu (approved):
LGTM :shipit:

Aljullu:
@tomasztunik could you create a good-first-issue (or directly a PR) to make sure the button label is aligned between both blocks? Currently it's [_Reset_ in Featured Product](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/82ff784f25a78b1d3557a73c23eeee40dfb417cb/assets/js/blocks/featured-product/block.js#L161) but [_Remove_ in Featured Category](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/82ff784f25a78b1d3557a73c23eeee40dfb417cb/assets/js/blocks/featured-category/block.js#L113), ideally I think they should be the same. Do you agree?

tomasztunik:
@Aljullu was planning to but the alignment issue got fixed by https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5867 already!

Regarding the reset/remove - here it resets to product image - so we still have an image, there it removes the image completely and uses a background color as selected with styles. Should we unify these despite slightly different behaviour?

Aljullu:
> there it removes the image completely and uses a background color as selected with styles

Does it? It seems to reset the image to whatever the category image is for me:

https://user-images.githubusercontent.com/3616980/155307005-ac5f6cd9-08eb-4b5e-9b8b-b6ad7c5d02d4.mp4

(Notice that if the category doesn't have an image, then yes, a background color is used)

tomasztunik:
Ah I see, had a category without image when testing. Will update it to reflect that if image is present it will say reset else remove? or always reset?

Aljullu:
> Ah I see, had a category without image when testing. Will update it to reflect that if image is present it will say reset else remove? or always reset?

IMO always reset is good enough. To me, it makes sense in both contexts.

PR #5892, opened 2022-02-18
Pre-hydrate Mini Cart Block #5882

### LINKED ISSUE #5882: Mini Cart block: hydrating JS block before showing it
## Describe the bug

When a filled empty cart is focused for the first time it is loaded the JS block. For a time, the JS block doesn't have data and it shows 0 as the price for a second. We need to hydrate the JS block before showing it.

## To reproduce

1. Open the FSE editor.
2. Add the Mini Cart block.
3. Save it.
4. On the fronted side add a product.
5. Refresh the page.
6. Get the focus on the Mini Cart block.
7.  Notice that for a moment the price is 0.

## Expected behavior

The price should be always the same even when the JS block is loaded.

## Screenshots

https://user-images.githubusercontent.com/4463174/154497920-abbf2142-9d03-449e-9829-6001a2d56a82.mp4

gigitux:
This PR resolves the hydration problem for the Mini Cart Block.

Fixes #5882

### Manual Testing

Check out this branch.

1. Open the FSE editor.
2. Add the Mini Cart block.
3. Save it.
4. On the fronted side add a product.
5. Refresh the page.
6. Get the focus on the Mini Cart block.
7.  Check that the price is always the same. (prior to this PR, for a moment the price returned to 0)

### DIFF
--- a/assets/js/types/type-guards/cart-response-totals.ts
+++ b/assets/js/types/type-guards/cart-response-totals.ts
@@ -1,0 +1,42 @@
+/**
+ * Internal dependencies
+ */
+import { CartResponseTotals } from '../type-defs';
+import { isObject } from './object';
+
+// It is the only way to create a type that contains all the object's keys and gets type-checking.
+// This is useful because we want to check that the keys object ALWAYS contains all the object's keys.
+// https://stackoverflow.com/questions/52028791/make-a-generic-type-arraykeyof-t-require-all-keys-of-t
+
+type CartResponseTotalsKeys = Record< keyof CartResponseTotals, 0 >;
+
+export const isCartResponseTotals = (
+	value: unknown
+): value is CartResponseTotals => {
+	if ( ! isObject( value ) ) {
+		return false;
+	}
+
+	const keys: CartResponseTotalsKeys = {
+		total_items: 0,
+		total_items_tax: 0,
+		total_fees: 0,
+		total_fees_tax: 0,
+		total_discount: 0,
+		total_discount_tax: 0,
+		total_shipping: 0,
+		total_shipping_tax: 0,
+		total_price: 0,
+		total_tax: 0,
+		tax_lines: 0,
+		currency_code: 0,
+		currency_symbol: 0,
+		currency_minor_unit: 0,
+		currency_decimal_separator: 0,
+		currency_thousand_separator: 0,
+		currency_prefix: 0,
+		currency_suffix: 0,
+	};
+
+	return Object.keys( keys ).every( ( key ) => key in value );
+};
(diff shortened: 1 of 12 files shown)


src/BlockTypes/MiniCart.php
```diff
@@ -118,10 +118,18 @@ protected function enqueue_data( array $attributes = [] ) {
...
+			$cart_payload = $this->get_cart_payload();
+
+			$this->asset_data_registry->add(
+				'cartTotals',
+				isset( $cart_payload['totals'] ) ? $cart_payload['totals'] : null,
+				null
+			);
+
 			$this->asset_data_registry->add(
-				'displayCartPricesIncludingTax',
```
gigitux, line 122: Removed this, because it is already defined at line 174 (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5892/files#diff-4c8bd3d0cfbd21fce3d29b5b50c28a5172490edbdabe90867fbf4f36dd5d3cb4R174)

assets/js/types/type-guards/cart-response-totals.ts
```diff
@@ -0,0 +1,42 @@
...
+		total_tax: 0,
+		tax_lines: 0,
+		currency_code: 0,
+		currency_symbol: 0,
+		currency_minor_unit: 0,
+		currency_decimal_separator: 0,
+		currency_thousand_separator: 0,
+		currency_prefix: 0,
+		currency_suffix: 0,
+	};
```
Aljullu, line 39: Is there any advantage of this being an object instead of an array of the keys? From my understanding, we don't read the values at any moment, right?

Aljullu (approved):
Looks good to me, and thanks for investigating that tricky issue with the test!

assets/js/types/type-guards/cart-response-totals.ts
gigitux: Yes! I will attach some videos that explain better.

If we use an array implementation, we can't say to typescript that all the keys from the type should stay in the array.

If you see this video, you can see that TS returns an error only if we add a value that is not equal to one of the keys of the type.

https://user-images.githubusercontent.com/4463174/154692817-794e5e22-2f0d-4eac-a994-944044de43ff.mp4

Instead, we are creating a type guard and we want to be sure at compilation time that we check ALL the keys. This is possible only with an object. So, if the object has a key that is not included in the type or missing a key we will have an error (this case is not covered with an array implementation).

https://user-images.githubusercontent.com/4463174/154692872-f1a90235-5803-482a-bdcb-b6e7a1e17b75.mp4



This approach is future proof: if another developer will add a new key to the type and does not update the type guard we will have an error at compile time 👍

PR #5894, opened 2022-02-18
Shopper → Mini Cart → Can open/close the drawer

### LINKED ISSUE #5829: Critical flows: Shopper → Mini Cart → Can open/close the drawer
### Note

Create e2e tests to verify that a shopper can open and close the Mini cart drawer

### File

`tests/e2e/specs/shopper/mini-cart-drawer.test.js`

Part of #5814

dinhtungdu:
Fixes #5829

### Notes

This PR adds e2e tests to verify that a shopper can open and close the Mini Cart drawer.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/mini-cart.test.js -t "Drawer"` to run only this test case.


tests/e2e/specs/shopper/mini-cart-drawer.test.js
```diff
@@ -0,0 +1,64 @@
+/**
+ * Internal dependencies
+ */
+import { shopper } from '../../../utils';
+
+const block = {
+	name: 'Mini Cart Block',
+};
+
+if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 3 )
```
sunyatasattva: What about doing something like this (and perhaps putting it in our utils):

```js
function describeIf(condition, ...args) {
  return condition ? describe(...args) : describe.skip(...args);
}
```

So then we can do:

```js
describeIf(
  process.env.WOOCOMMERCE_BLOCKS_PHASE > 2,
  'Shopper → Mini Cart → Can open/close the drawer',
  () => /* ... */
);
```

tests/e2e/specs/shopper/mini-cart-drawer.test.js
```diff
@@ -0,0 +1,64 @@
...
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( `skipping ${ block.name } tests`, () => {} );
+
+describe( 'Shopper → Mini Cart → Can open/close the drawer', () => {
```
sunyatasattva: I'm wondering… 🤔  would it make sense structurally to have a high level `describe` block for the whole `Mini Cart` thing and then nest the functionalities within other `describe`s?

So, it would look something like this:

```js
describe( 'Mini Cart', () => {
  describe( 'Drawer', () => {
    it( 'should open when shopper clicks outside', () => { /* ... */ } )
  } );

  describe( 'Icon', () => {
    it( 'should be visible with a badge on the front-end', () => { /* ... */ } )
  } );
} );
```

Notice how this reads much more like smooth English and already organizes the tests for you. And in the console it would be neatly printed as well, which is a plus.

What do you think?

dinhtungdu: @sunyatasattva I think we have the winner here:

image
	
image

dinhtungdu: I thought splitting the test suites into separated files is easier to maintain and run (Eg: running a single test case using a file filter). But I don't think we need to run a single test suite very option. And we can still run a single test suite using filters.

dinhtungdu: The problem with this approach is the test file can be very long. Combining Icon and Drawer tests, the test file is 100+ lines. It's not a problem for me but I'm not sure about others' preferences.

sunyatasattva (approved):
Approving as discussed on Slack :)

tests/e2e/specs/shopper/mini-cart-drawer.test.js
dinhtungdu: As discussed, I will work on this on a separate issue/PR.

sunyatasattva: Sounds good to me :)

PR #5896, opened 2022-02-18
Try move shipping related data to a `@wordpress/data` store

### LINKED ISSUE #5898: Investigate `@wordpress/data` store for shipping specific data and a new `useShippingData` hook to get this data without relying on context
As part of the untangling work, the `useShippingDataContext` hook draws information from a few different places, we should look into whether we can remove this context provider and instead derive these data from a data store.

opr:
This PR depends on #5981 - the changes from that are included in this PR.

This PR will create a new hook called `useShippingData` it is very similar to the extant `useShippingDataContext` but instead of getting data from the context provider, the data is derived from the `wc/store/v1/cart` data store. You can get this hook from `@woocommerce/base-context/hooks`.

At the moment, the `shippingErrorStatus` `dispatchErrorStatus`, `shippingErrorTypes` and the event observers are still kept in context as it doesn't make sense to move them outside of it.

- Updated tests to use `useShippingData` rather than `useShippingDatacontext`.
- Create a hook `useShippingData` that gets shipping data out of the `wc/store/v1/cart` data store.
- Update `assets/js/base/context/hooks/payment-methods/use-payment-method-interface.ts` because it was returning incorrect methods. There is no such thing as `setSelectedRates`, it is `selectShippingRate`.
- Update types in `assets/js/types/type-defs/payment-method-interface.ts`.
- Create `SelectShippingRateType` which describes the shape of the value returned by the `useSelectShippingRate` hook.
- Convert `assets/js/base/utils/derive-selected-shipping-rates.ts` to TypeScript.
- Add the following selectors to the `wc/store/v1/cart` store:
   - `getShippingRates`
   - `getNeedsShipping`
   - `getHasCalculatedShipping`

Fixes #5898 

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Set up multiple shipping methods, ensure you can change between them. Ensure the price updates when selecting different ones.
2. Check out using a few different payment methods, make sure to try express payment methods and ensure the correct address is submitted.
3. Ensure CI tests pass.

### User Facing Testing
* [x] Same as above

### Changelog


mikejolley:
@opr I think this now needs a rebase before I can review the new changes.

opr:
@mikejolley this is still a draft, but I'll rebase it before I next start working on it. If you want to work on anything to do with it feel free to rebase now though.

opr:
@mikejolley this is now ready for review

assets/js/base/context/hooks/shipping/use-shipping-data.ts
```diff
@@ -0,0 +1,66 @@
...
+		needsShipping,
+		hasCalculatedShipping,
+		shippingRatesLoading,
+	} = useSelect( ( select ) => {
+		const store = select( storeKey );
+		return {
+			shippingRates: store.getShippingRates(),
+			needsShipping: store.getNeedsShipping(),
+			hasCalculatedShipping: store.getHasCalculatedShipping(),
+			shippingRatesLoading: store.isCustomerDataUpdating(),
```
mikejolley, line 38: Would now be a good time to change this property name to `isSomething`? `isLoading` or `isUpdating` maybe? To better reflect it's a boolean.

assets/js/base/context/providers/cart-checkout/shipping/index.js
```diff
@@ -25,9 +23,9 @@ import {
 	emitEvent,
 } from './event-emit';
 import { useCheckoutContext } from '../checkout-state';
-import { useCustomerDataContext } from '../customer';
 import { useStoreCart } from '../../../hooks/cart/use-store-cart';
 import { useSelectShippingRate } from '../../../hooks/shipping/use-select-shipping-rate';
+import { useShippingData } from '../../../hooks/shipping/use-shipping-data';
 
 /**
  * @typedef {import('@woocommerce/type-defs/contexts').ShippingDataContext} ShippingDataContext
```
mikejolley, line 31: Are we keeping `useShippingDataContext` around?

assets/js/base/context/hooks/cart/use-store-cart.ts
```diff
@@ -85,7 +82,7 @@ const defaultCartTotals: CartResponseTotals = {
 const decodeValues = (
 	object: Record< string, unknown >
 ): Record< string, unknown > =>
-	fromEntriesPolyfill(
+	Object.fromEntries(
```
mikejolley: I know you have a separate PR for this, but I think it's a good change now 👍🏻  Very little share for ie11 and we don't test it anyway.

assets/js/base/context/hooks/shipping/use-shipping-data.ts
```diff
@@ -0,0 +1,66 @@
...
+import { deriveSelectedShippingRates } from '@woocommerce/base-utils';
+import isShallowEqual from '@wordpress/is-shallow-equal';
+import { isObject } from '@woocommerce/types';
+
+/**
+ * Internal dependencies
+ */
+import { useSelectShippingRate } from './use-select-shipping-rate';
+
+type ShippingData = Pick<
```
mikejolley: Is `Pick` prefered over redeclaring the props and/or importing the typedef per prop via `typeof`? e.g.

```
shippingRates: typeof Cart['shippingRates'];
```

Asking because I've done it the other way but not considered Pick.

mikejolley (approved):
Tested the updating of shipping methods in cart and checkout, and also with multiple methods from Subscriptions. All seems to work and I can see updates via redux dev tools. Good job :)

assets/js/base/context/providers/cart-checkout/shipping/index.js
opr: Yes we need to for now because it contains information relating to the error states and emits events. I didn't think it made sense to move it anywhere else for the time being, maybe when validation is moved it could go there.

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/123511d7ed892fb4622991f076140daf1d9df137/assets/js/base/context/providers/cart-checkout/shipping/index.js#L62

assets/js/base/context/hooks/shipping/use-shipping-data.ts
opr: Good idea I'll change it

assets/js/base/context/hooks/shipping/use-shipping-data.ts
opr: I'm not sure if it's preferred, but I think your way is actually a bit tidier, imo it makes it easier to read.

PR #5899, opened 2022-02-18
Fix New Block Inserter Overlaps

### LINKED ISSUE #5849: Checkout: new block inserter overlaps shipping options if Shipping Address block doesn't have step number
## To reproduce
1. Add the Checkout block.
2. Click on the Shipping address inner block and uncheck the Show step number attribute.
3. Notice the new block inserter at the bottom of the block overlaps with the Shipping options title.
imatge

## Expected behavior
Blocks shouldn't overlap.

tarunvijwani:
When the Show step number attribute is unchecked, the new block inserter will not get overlapped with the title of the next block at the bottom.

Fixes #5849

### Screenshots

| Before | After |
|---|---|
| image | image |

### Manual Testing

1. Add the Checkout block.
2. Click on any inner block (Shipping address) and uncheck the Show step number attribute.
3. The new block inserter at the bottom of the block should not overlap with the title of the next block (Shipping options).
### User Facing Testing
* [x] Same as above

### Changelog

> Fix new block inserter overlaps

### DIFF
--- a/assets/js/blocks/checkout/form-step/editor.scss
+++ b/assets/js/blocks/checkout/form-step/editor.scss
@@ -1,6 +1,6 @@
 
 .wc-block-checkout__additional_fields {
-	margin: 1.5em 0 -1.5em;
+	margin: 1.5em 0 0;
 }
 .wc-block-components-checkout-step__description-placeholder {
 	opacity: 0.5;


nielslange (approved):
Works like charm! 👏 I'd say, let's 🚢 it!

Issue #5754, opened 2022-02-09
Critical flows: Shopper → Cart → Can see correct tax


nielslange:
### Story

As a shopper, I want to see the correct tax being applied on the cart page.

### File

`tests/e2e/specs/shopper/cart-tax.test.js`

#### Tax free

1. Go to the product page.
2. Add a tax-free product to the cart.
3. Verify that the correct tax is visible on the cart page.
4. Go to the checkout page.
5. Place the order.
6. Verify that the correct tax is visible on the order confirmation page.

#### 20% tax

1. Go to the product page.
2. Add a product, with 20% tax, to the cart.
3. Verify that the correct tax is visible on the cart page.
4. Go to the checkout page.
5. Place the order.
6. Verify that the correct tax is visible on the order confirmation page.

PR #5900, opened 2022-02-18
Critical flow: Can see correct tax

### LINKED ISSUE #5772: Critical flows: Shopper → Checkout → Can see correct tax
### Story

As a shopper, I want to see the correct tax being applied on the checkout page.

### File

`tests/e2e/specs/shopper/checkout-tax.test.js`

#### Tax free

1. Go to the product page.
2. Add a tax-free product to the cart.
3. Go to the checkout page.
4. Verify that the correct tax is visible on the checkout page.
5. Place the order.
6. Verify that the correct tax is visible on the order confirmation page.

#### 20% tax

1. Go to the product page.
2. Add a product, with 20% tax, to the cart.
3. Go to the checkout page.
4. Verify that the correct tax is visible on the checkout page.
5. Place the order.
6. Verify that the correct tax is visible on the order confirmation page.

alexflorisca:
This PR ads an e2e test to check that the right tax is being displayed, both in Cart and Checkout blocks. It also builds on the WC e2e-utils `shopper` functions and adds some useful ones related to blocks.

Fixes #5772 and #5754 

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. `npm run wp-env`
2. Either run `npm run test:e2e -- Tax` 
3. Make sure all tests pass
4. Make sure e2e tests pass in CI


nielslange (dismissed):

bin/wp-env-config.sh
```diff
@@ -18,5 +18,12 @@ wp rewrite flush
 wp core version --extra
 wp plugin list
 wp theme activate storefront
+wp plugin is-installed Basic-Auth
+if [ $? -eq 1 ]
+then
+  wp plugin install https://github.com/WP-API/Basic-Auth/archive/master.zip --activate
+else 
+  wp plugin activate Basic-Auth
+fi
```
nielslange: After approving this PR before, I noticed the following section. This plugin should already be installed and activated by default due to the following entry:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/c8a5041817eb672a8490453731bd3dccc5dfa58a/.wp-env.json#L3-L7

Therefore, I'd say the following section is not necessary, isn't it?

package.json
```diff
@@ -65,6 +65,7 @@
 		"storybook:deploy": "rimraf ./storybook/dist/* && npm run storybook:build && gh-pages -d ./storybook/dist",
 		"test": "wp-scripts test-unit-js --config tests/js/jest.config.json",
 		"test:e2e": "npm run wp-env:config && cross-env NODE_CONFIG_DIR=tests/e2e/config wp-scripts test-e2e --config tests/e2e/config/jest.config.js",
+		"test:e2e-single": "npm run wp-env:config && cross-env NODE_CONFIG_DIR=tests/e2e/config wp-scripts test-e2e --config tests/e2e/config/jest.config.js -t '<NAME OF YOUR TEST>' --watch --puppeteer-interactive",
```
nielslange: I'm uncertain if we need this command, as we can easily run an individual test by specifying the test path. E.g. `npm run test:e2e-single Tax` could also be run with `npm run test:e2e-dev-watch -- tests/e2e/specs/shopper/tax.test.js`. 

I'd rather use the current commands and pipe the test URL into these commands, when required. Personally, I'd only introduce a new command if the desired action cannot be achieved with the existing commands. What are your thoughts on this?

nielslange (commented):
Overall, your PR looks fantastic! 👋 I just added two minor questions to it.

package.json
alexflorisca: Yeah you're right here, I didn't realise you could pass params to the underlying jest script through the current npm script. I'll remove this and use your way of doing it, much neater, thanks 👍

bin/wp-env-config.sh
alexflorisca: Yep, good spot, I wasn't aware this even existed, thanks for pointing it out 👍

alexflorisca:
Thanks for review @nielslange , I've made the change you requested. I also noticed that I haven't checked the order confirmation page as your issue describes, so I'll add that in and re-request the review in a bit

tests/e2e/specs/shopper/tax.test.js
```diff
@@ -0,0 +1,80 @@
...
+	getTaxesFromOrderSummaryPage,
+	showTaxes,
+} from '../../../utils/taxes';
+
+const taxRates = Taxes();
+const productWooSingle1 = Products().find(
+	( prod ) => prod.name === 'Woo Single #1'
+);
+
+jest.setTimeout( 120000 );
```
ralucaStan: why is this needed?

tests/utils/shopper.js
```diff
@@ -47,6 +58,148 @@ export const shopper = {
...
+			address.state
+		);
+		await shopper.clearAndTypeInput(
+			`#${ shippingOrBilling }-postcode`,
+			address.postcode
+		);
+	},
+
+	placeOrder: async () => {
+		await sleep( 1 );
```
ralucaStan: Could [uiUnblocked](https://www.npmjs.com/package/@woocommerce/e2e-utils) be used instead?

tests/utils/shopper.js
```diff
@@ -47,6 +58,148 @@ export const shopper = {
...
+		await page.click( '.wc-block-components-checkout-place-order-button' );
+	},
+
+	clearAndTypeInput: async ( selector, text ) => {
```
ralucaStan: We don't need this if we use the `await expect(page).toFill('input[name="firstName"]', 'James')` from [expect-puppeteer](https://www.npmjs.com/package/expect-puppeteer), which automatically clears the input

tests/e2e/specs/shopper/tax.test.js
```diff
@@ -0,0 +1,80 @@
...
+			await page.waitForNavigation();
+			await page.waitForSelector( 'h1.entry-title' );
+			const orderSummaryTaxes = await getTaxesFromOrderSummaryPage(
+				taxRates.filter( ( taxRate ) => taxRate.country === 'US' )
+			);
+			expect( orderSummaryTaxes.sort() ).toEqual( expectedTaxes.sort() );
+		} );
+	} );
+} );
+jest.setTimeout( 60000 );
```
ralucaStan: Same question, I wonder why this is needed.

tests/utils/shopper.js
```diff
@@ -2,15 +2,26 @@
...
 
+	goToCartBlock: async () => {
+		const checkoutBlockPermalink = await getBlockPagePermalink(
+			`Cart Block`
+		);
+
+		await page.goto( checkoutBlockPermalink, {
+			waitUntil: 'networkidle0',
+		} );
+		await page.waitForSelector( 'h1', { text: 'Cart' } );
```
ralucaStan: I know this is something present in other tests, but the API for waitForSelector does not allow to pass text here. https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagewaitforselectorselector-options

Better to use `await expect(page).toMatchElement('h1', { text: 'Cart' })` from expect-pupetter.

tests/utils/shopper.js
```diff
@@ -47,6 +58,148 @@ export const shopper = {
...
+		);
+		if ( ! applyButton ) {
+			await page.click( '.wc-block-components-panel__button' );
+		}
+		await page.type(
+			'.wc-block-components-totals-coupon__input input',
+			couponCode
+		);
+		await page.click( '.wc-block-components-totals-coupon__button' );
+		await page.waitForSelector( '.wc-block-components-chip__text', {
```
ralucaStan: again, this is incorrect, we need to use `await expect(page).toMatchElement('.wc-block-components-chip__text', { text: couponCode })`

ralucaStan: This also does some waiting until the selector is in the page.

tests/utils/shopper.js
```diff
@@ -47,6 +58,148 @@ export const shopper = {
 		);
 	},
 
+	/* We need to overwrite this function from wcShopper because clicking through to the
+		product doesn't work. There is a fix in https://github.com/woocommerce/woocommerce/pull/31915
+		We can delete this function once the PR is merged
+		*/
+	searchForProduct: async ( productname ) => {
+		const searchFieldSelector = 'input.wp-block-search__input';
+		await page.waitForSelector( searchFieldSelector, {
```
ralucaStan: `await expect(page).toMatchElement()` will wait for the element

ralucaStan (dismissed):
@alexflorisca Really nice work on testing taxes, nobody likes doing taxes :)

I've left some comments

tests/e2e/specs/shopper/tax.test.js
alexflorisca: These tests are timing out, but only on CI. This is just me trying to increase the timeout to see if it fixes it (it didn't, it doesn't seem to increase the timeout at all). Weirdly this doesn't work inside a test or inside any of the `before` hooks which is why I've tried putting it out here

tests/utils/shopper.js
alexflorisca: That was my first try, I'm not sure how it works in woocommerce, when the `.blockUI` class is added on which the check is based, but it didn't work in this case. I suspect because we have to wait until the data store is updated, at which time there might be some UI changes and things can become unresponsive for some milliseconds. I know this isn't a neat solution. We could maybe try to write something more clever but I thought this was a good first pass. Happy to consider alternatives though

tests/e2e/specs/shopper/tax.test.js
ralucaStan: I had a test timing out for me locally. We need to increase the general setting, as we will have a lot of _longer_ tests. Can you try to only add this new value [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2bfecbcf4ac26ea3af89c5929864ab119bd3263b/tests/e2e/config/jest.setup.js#L13-L14) and see the result.

tests/utils/shopper.js
ralucaStan: I've made this suggestion below based on the WC implementation. @nielslange  also added a similar utility

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5860#discussion_r813802506

ralucaStan: I've made this suggestion based on the WC implementation. @niels also added a similar utility

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5860#discussion_r813802506

ralucaStan: I've made this suggestion based on the WC implementation. @niels also added a similar utility

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5860#discussion_r813802506

senadir:
> I've added the [Basic-Auth](https://github.com/WP-API/Basic-Auth) plugin in order to be able to use the WC Rest API helper withRestApi from the [e2e-utils](https://www.npmjs.com/package/@woocommerce/e2e-utils) package to toggle WC settings.

This should already be added as we use it in our fixtures.

alexflorisca:
> > I've added the [Basic-Auth](https://github.com/WP-API/Basic-Auth) plugin in order to be able to use the WC Rest API helper withRestApi from the [e2e-utils](https://www.npmjs.com/package/@woocommerce/e2e-utils) package to toggle WC settings.
> 
> This should already be added as we use it in our fixtures.

Yeah @nielslange pointed this out as well, thanks @senadir , I've removed the extra invocation of the plugin

tests/e2e/specs/frontend/legacy-template-blocks.test.js
```diff
@@ -43,7 +43,7 @@ describe( 'Legacy Template blocks', () => {
 	} );
 
 	afterAll( async () => {
-		await activateTheme( 'twentytwentyone' );
+		await activateTheme( 'storefront' );
```
alexflorisca, line 46: Switching theme to twentytwentyone makes the following tests fail, as most tests expect storefront as the theme. I will update this in a future PR to use the `useTheme` hook as @tomasztunik pointed out

nielslange (approved):
Well done, @alexflorisca! Let's 🚢 it! 🙌

tests/e2e/specs/shopper/tax.test.js
alexflorisca: Yep this worked, thanks :)

PR #5904, opened 2022-02-21
Store API: Refactored validation handling and introduced `woocommerce_store_api_cart_errors` hook

xristos3490:
This refactor is based on how the system handles the `InvalidStockLevelsInCartException` and aims to create a more _DRY_ error validation architecture.

Currently, when there are multiple _(stock-related)_errors raised from the cart contents, the system will group them in a single `WP_Error` instance with a "409 Conflict" code. During the REST response, these errors are deconstructed into multiple error keys. One error/key.

This PR replaces the `InvalidStockLevelsInCartException` with a more generic `InvalidCartException` that should handle every validation-related error from the core, stock, or 3PD. 

In addition, the `InvalidCartException` is designed to handle multiple errors in its own `WP_Error`. 

When a `validate_xxxx` method throws an `InvalidCartException` containing multiple errors:
- The related cart errors getter will catch it and deconstruct its errors into multiple `WP_Error` objects; the final REST response will be the same.
- The checkout route will catch it and continue with the same approach; the difference from the `InvalidStockLevelsInCartException` is that the system will have the original error codes instead of "409 Conflict".

In a nutshell:
- Error getters are _try...catch_ wrappers of the validate-related methods (DRY).
- Introduces a new generic `InvalidCartException` class to be used in all errors related to the cart validity.
- The new exception type supports multiple errors on a single cart check.
- Uses as a primary source of validation all the `validate_xxxx` methods.
- You can still break the checkout block if you return an Exception from the POST response context; this probably needs a mechanism to prevent it.
- The `__experimental_woocommerce_store_api_validate_cart` and the `woocommerce_check_cart_items` are connected. It seems like a good path for deprecating the latter in the API. See Next Steps # 2
- The newly created `__experimental_woocommerce_store_api_validate_cart` hook is called in the cart and checkout context to create consistency.

cart_validation_refactor

**What's changed**

- When multiple errors are thrown from the "Place Order" flow, each will retain a readable string as error-code instead of the generic _409_ error.
- `RouteException` is now handled in both the cart and checkout flows, as previously, were only held in the cart flow.

**Example 3PD cart validation usage:**

```php

add_filter( '__experimental_woocommerce_store_api_validate_cart', 'test_cart_errors' );
function test_cart_errors( $cart ) {

    $error = new \WP_Error();
    // First error.
    $error->add(
        'test_error_1',
        'Cart error #1'
    );
    // Second error.
    $error->add(
        'test_error_2',
        'Cart error #2'
    );

    throw new \Automattic\WooCommerce\Blocks\StoreApi\Utilities\InvalidCartException(
        'test_plugin_error_code',
        $error,
        409
    );
}
```

The snippet above adds multiple errors, and the result is handled accordingly: https://d.pr/i/vPircs (checkout) and https://d.pr/i/alWAW9 (cart).

**A present use-case:**

_Conditional Shipping and Payments_ extension is a condition-based restriction system that could potentially produce complex business restrictions. Throwing a `WP_Error` object containing multiple errors is crucial for correctly handling all situations. 

**Next steps:**

- Add a`context` argument in `validate_cart()` and control whether the `__experimental_woocommerce_store_api_validate_cart` or the `woocommerce_check_cart_items` should fire. Perhaps, for performance reasons, we need to only run the `woocommerce_check_cart_items` legacy hook in the checkout context to avoid handling WC notices to exceptions all the time.
- We could avoid adding a `context` argument by moving the `woocommerce_check_cart_items` directly inside the `Checkout::get_route_post_response`; it will be deprecated, so it shouldn't be a _clean-code_ issue.

#### Accessibility

#### Other Checks

- [x] I've updated [this doc](../docs/blocks/feature-flags-and-experimental-interfaces.md) for any feature flags or experimental interfaces implemented in this pull request.

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests

### User Facing Testing

None.

### Changelog

> Introduce the `InvalidCartException` for handling cart validation.

### DIFF
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -1,12 +1,12 @@
 <?php
 namespace Automattic\WooCommerce\Blocks\StoreApi\Utilities;
 
-use Automattic\WooCommerce\Blocks\StoreApi\Exceptions\InvalidStockLevelsInCartException;
 use Automattic\WooCommerce\Blocks\StoreApi\Exceptions\NotPurchasableException;
 use Automattic\WooCommerce\Blocks\StoreApi\Exceptions\OutOfStockException;
 use Automattic\WooCommerce\Blocks\StoreApi\Exceptions\PartialOutOfStockException;
 use Automattic\WooCommerce\Blocks\StoreApi\Exceptions\TooManyInCartException;
 use Automattic\WooCommerce\Blocks\StoreApi\Exceptions\RouteException;
+use Automattic\WooCommerce\Blocks\StoreApi\Exceptions\InvalidCartException;
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\DraftOrderTrait;
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\NoticeHandler;
 use Automattic\WooCommerce\Blocks\StoreApi\Utilities\QuantityLimits;
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -396,9 +396,9 @@ class CartController {
 		 * @example See docs/examples/validate-cart.md
 		 *
 		 * @param WP_Error $errors  WP_Error object.
-		 * @param WC_Cart $cart     Cart object.
+		 * @param WC_Cart  $cart    Cart object.
 		 *
-		 * @return WP_Error  The WP_Error object returned.
+		 * @return WP_Error         The WP_Error object returned.
 		 */
 		$cart_errors = apply_filters( '__experimental_woocommerce_store_api_cart_errors', new \WP_Error(), $cart );
 
@@ -425,7 +425,7 @@ class CartController {
 		 *
 		 * Allow 3rd parties to validate cart items. This is a legacy hook from Woo core.
 		 * This filter will be deprecated because it encourages usage of wc_add_notice. For the API we need to capture
-		 * notices and convert to exceptions instead.
+		 * notices and convert to wp errors instead.
 		 */
 		do_action( 'woocommerce_check_cart_items' );
 
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -668,7 +668,7 @@ class CartController {
 	/**
 	 * Validate the cart and get a list of errors.
 	 *
-	 * @return WP_Error[] An array of WP_Errors describing the cart's error state.
+	 * @return WP_Error A WP_Error instance containing the cart's errors.
 	 */
 	public function get_cart_errors() {
 		$errors = [];
@@ -676,73 +676,18 @@ class CartController {
 		try {
 			$this->validate_cart();
 		} catch ( RouteException $error ) {
-			$errors[] = new WP_Error( $error->getErrorCode(), $error->getMessage(), $error->getAdditionalData() );
+			$errors->add( $error->getErrorCode(), $error->getMessage(), $error->getAdditionalData() );
 		} catch ( InvalidCartException $error ) {
 
 			$wp_error = $error->getError();
 			foreach ( (array) $wp_error->errors as $code => $messages ) {
 				foreach ( (array) $messages as $message ) {
 					$additional_data = $wp_error->get_error_data( $code );
-					$errors[]        = new WP_Error( $code, $message, $additional_data );
+					$errors->add( $code, $message, $additional_data );
 				}
 			}
 		} catch ( \Exception $error ) {
-			$errors[] = new WP_Error( $error->getCode(), $error->getMessage() );
-		}
-
-		$cart_errors = array_filter(
-			$errors,
-			function ( WP_Error $error ) {
-				return $error->has_errors();
-			}
-		);
-
-		return $cart_errors;
-	}
-
-	/**
-	 * Validate all items in the cart and get a list of errors.
-	 *
-	 * @return WP_Error[] An array of WP_Errors describing the cart's error state.
-	 */
-	public function get_cart_item_errors() {
-		$errors = [];
-
-		try {
-			$this->validate_cart_items();
-		} catch ( InvalidCartException $error ) {
-
-			$wp_error = $error->getError();
-			foreach ( (array) $wp_error->errors as $code => $messages ) {
-				foreach ( (array) $messages as $message ) {
-					$additional_data = $wp_error->get_error_data( $code );
-					$errors[]        = new \WP_Error( $code, $message, $additional_data );
-				}
-			}
-		}
-
-		return $errors;
-	}
-
-	/**
-	 * Validate all coupons in the cart and get a list of errors.
-	 *
-	 * @return WP_Error[] An array of WP_Errors describing coupons error state.
-	 */
-	public function get_cart_coupon_errors() {
-		$errors = [];
-
-		try {
-			$this->validate_cart_coupons();
-		} catch ( InvalidCartException $error ) {
-
-			$wp_error = $error->getError();
-			foreach ( (array) $wp_error->errors as $code => $messages ) {
-				foreach ( (array) $messages as $message ) {
-					$additional_data = $wp_error->get_error_data( $code );
-					$errors[]        = new \WP_Error( $code, $message, $additional_data );
-				}
-			}
+			$errors->add( $error->getCode(), $error->getMessage() );
 		}
 
 		return $errors;
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -671,21 +671,14 @@ class CartController {
 	 * @return WP_Error A WP_Error instance containing the cart's errors.
 	 */
 	public function get_cart_errors() {
-		$errors = [];
+		$errors = new WP_Error();
 
 		try {
 			$this->validate_cart();
 		} catch ( RouteException $error ) {
 			$errors->add( $error->getErrorCode(), $error->getMessage(), $error->getAdditionalData() );
 		} catch ( InvalidCartException $error ) {
-
-			$wp_error = $error->getError();
-			foreach ( (array) $wp_error->errors as $code => $messages ) {
-				foreach ( (array) $messages as $message ) {
-					$additional_data = $wp_error->get_error_data( $code );
-					$errors->add( $code, $message, $additional_data );
-				}
-			}
+			$errors->merge_from( $error->getError() );
 		} catch ( \Exception $error ) {
 			$errors->add( $error->getCode(), $error->getMessage() );
 		}
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -384,23 +384,25 @@ class CartController {
 	 * @throws InvalidCartException Exception if invalid data is detected in the cart.
 	 */
 	public function validate_cart() {
-
-		$cart = $this->get_cart_instance();
 		$this->validate_cart_items();
 		$this->validate_cart_coupons();
 
+		$cart        = $this->get_cart_instance();
+		$cart_errors = new WP_Error();
+
 		/**
-		 * Fire action to validate cart. Functions hooking into this should add custom errors into the WP_Error instance.
+		 * Fires an action to validate the cart.
+		 *
+		 * Functions hooking into this should add custom errors using the provided WP_Error instance.
 		 *
 		 * @example See docs/examples/validate-cart.md
 		 *
-		 * @param WP_Error $errors  WP_Error object.
-		 * @param WC_Cart  $cart    Cart object.
+		 * @param \WP_Error $errors  WP_Error object.
+		 * @param \WC_Cart  $cart    Cart object.
 		 */
-		$cart_errors = new WP_Error();
 		do_action( '__experimental_woocommerce_store_api_cart_errors', $cart_errors, $cart );
-		if ( $cart_errors->has_errors() ) {
 
+		if ( $cart_errors->has_errors() ) {
 			throw new InvalidCartException(
 				'woocommerce_cart_error',
 				$cart_errors,
@@ -418,15 +420,17 @@ class CartController {
 		 * Allow 3rd parties to validate cart items. This is a legacy hook from Woo core.
 		 * This filter will be deprecated because it encourages usage of wc_add_notice. For the API we need to capture
 		 * notices and convert to wp errors instead.
+		 *
+		 * @deprecated
 		 */
 		do_action( 'woocommerce_check_cart_items' );
 
-		$error_notices = NoticeHandler::convert_notices_to_wp_errors( 'woocommerce_rest_cart_item_error' );
-		if ( $error_notices->has_errors() ) {
+		$cart_errors = NoticeHandler::convert_notices_to_wp_errors( 'woocommerce_rest_cart_item_error' );
 
+		if ( $cart_errors->has_errors() ) {
 			throw new InvalidCartException(
 				'woocommerce_cart_error',
-				$error_notices,
+				$cart_errors,
 				409
 			);
 		}
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -428,7 +428,16 @@ class CartController {
 		 * notices and convert to exceptions instead.
 		 */
 		do_action( 'woocommerce_check_cart_items' );
-		NoticeHandler::convert_notices_to_exceptions( 'woocommerce_rest_cart_item_error' );
+
+		$error_notices = NoticeHandler::convert_notices_to_wp_errors( 'woocommerce_rest_cart_item_error' );
+		if ( $error_notices->has_errors() ) {
+
+			throw new InvalidCartException(
+				'woocommerce_cart_error',
+				$error_notices,
+				409
+			);
+		}
 	}
 
 	/**
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -395,10 +395,10 @@ class CartController {
 		 *
 		 * @example See docs/examples/validate-cart.md
 		 *
-		 * @param \WP_Error $errors  WP_Error object.
-		 * @param \WC_Cart $cart     Cart object.
+		 * @param WP_Error $errors  WP_Error object.
+		 * @param WC_Cart $cart     Cart object.
 		 *
-		 * @return \WP_Error  The WP_Error object returned.
+		 * @return WP_Error  The WP_Error object returned.
 		 */
 		$cart_errors = apply_filters( '__experimental_woocommerce_store_api_cart_errors', new \WP_Error(), $cart );
 
--- a/src/StoreApi/Utilities/CartController.php
+++ b/src/StoreApi/Utilities/CartController.php
@@ -463,18 +463,8 @@ class CartController {
 			);
 		}
 
-		$errors = $this->stock_exceptions_to_wp_errors( $too_many_in_cart_products, $not_purchasable_products, $partial_out_of_stock_products, $out_of_stock_products );
-
-		if ( ! empty( $errors ) ) {
-
-			$error = new WP_Error();
-			foreach ( $errors as $wp_error ) {
-				$error->add(
-					$wp_error->get_error_code(),
-					$wp_error->get_error_message(),
-					$wp_error->get_error_data()
-				);
-			}
+		$error = $this->stock_exceptions_to_wp_errors( $too_many_in_cart_products, $not_purchasable_products, $partial_out_of_stock_products, $out_of_stock_products );
+		if ( $error->has_errors() ) {
 
 			throw new InvalidCartException(
 				'woocommerce_stock_availability_error',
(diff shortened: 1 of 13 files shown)


manospsyx:
This pretty much supersedes #5874 (it is a bit broader in scope as it is meant to add some extra clarity around ongoing discussions on extensibility, consistency, and code re-use).

src/StoreApi/Routes/AbstractRoute.php
```diff
@@ -2,7 +2,7 @@
 namespace Automattic\WooCommerce\Blocks\StoreApi\Routes;
 
 use Automattic\WooCommerce\Blocks\StoreApi\Schemas\AbstractSchema;
-use Automattic\WooCommerce\Blocks\StoreApi\Utilities\InvalidStockLevelsInCartException;
+use Automattic\WooCommerce\Blocks\StoreApi\Utilities\InvalidCartException;
```
mikejolley: This will need an updated namespace now [v1 has merged](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5911).

src/StoreApi/Utilities/InvalidCartException.php
```diff
@@ -0,0 +1,74 @@
...
+	public $error;
+
+	/**
+	 * Setup exception.
+	 *
+	 * @param string   $error_code      Machine-readable error code, e.g `woocommerce_invalid_product_id`.
+	 * @param WP_Error $error           The WP_Error object containing all errors relating to stock availability.
+	 * @param array    $additional_data Extra data (key value pairs) to expose in the error response.
+	 */
+	public function __construct( $error_code, $error, $additional_data = [] ) {
```
mikejolley: We can add type hinting here if the `$error` must be a `WP_Error` object.

src/StoreApi/Utilities/CartController.php
```diff
@@ -355,15 +355,52 @@ private function get_error_message_for_stock_exception_type( $exception_type, $s
...
+		$this->validate_cart_coupons();
+
+		/**
+		 * Fire action to validate cart. Functions hooking into this should throw a \InvalidCartException.
+		 *
+		 * @example See docs/examples/validate-cart.md
+		 *
+		 * @param \WC_Cart $cart Cart object.
+		 */
+		do_action( '__experimental_woocommerce_store_api_validate_cart', $cart );
```
mikejolley: I can see why you've done this, but this does force a 3PD to have knowledge of both `WP_Error` handling, as well as `InvalidCartException`. 

At the very least, we could provide the `InvalidCartException` class/namespace via this hook so that a 3PD can throw the error without needing to know the correct namespace. This also guards against future updates if a namespace were to change.

The alternative that we should perhaps consider would be to rather than force 3PD to throw the exception with a `WP_Error` object, make this a filter hook which passes through a `WP_Error` object. 3PD could then use `$error->add()` to add errors to the shared instance.

This method (`validate_cart`) could also be further enhanced to replace `convert_notices_to_exceptions` to a new `convert_notices_to_errors` method which again, appends more errors to the shared `WP_Error` object.

Finally, once all this has completed, the generic `InvalidCartException` can be thrown internally.

The advantage of this approach would be:

1. 3PD only need knowledge of `WP_Error`, not the internals of our exception handling
2. 3PD can add/return multiple error messages
3. All messages + notices would be converted to a single Exception with the full list of errors which will be more useful for the response.

We also don't block the usage of Exceptions should there be a critical issue that would require no other logic or errors to be thrown/added. We leave that down to the 3PD.

Thoughts?

src/StoreApi/Utilities/CartController.php
```diff
@@ -580,69 +594,117 @@ public function validate_cart_item( $cart_item ) {
...
 		}
 	}
 
+	/**
+	 * Validate the cart and get a list of errors.
+	 *
+	 * @return WP_Error[] An array of WP_Errors describing the cart's error state.
+	 */
+	public function get_cart_errors() {
+		$errors = [];
```
mikejolley: Rather than an array of `WP_Error`, why don't we use a single `WP_Error` instance?

mikejolley (commented):
Added some initial feedback. I think the main thing we need to do is clean up the approach to WP_Error/Exception because there is a bunch of to-and-fro between the two types.

src/StoreApi/Utilities/CartController.php
xristos3490: I wanted to be consistent with the other error getters like:
- `get_cart_item_errors`
- `get_cart_coupon_errors`

At this point, we don't use any of them 😅 

Should we consider deprecating them? Or we could refactor them all in case we want to _catch_ errors from a specific context.

xristos3490: Also, I refactored the private `stock_exceptions_to_wp_errors` based on this idea.

src/StoreApi/Utilities/CartController.php
xristos3490: It took me a while to wrap my head around this. 

> [...] this force a 3PD to have knowledge of both WP_Error handling, as well as InvalidCartException.

Excellent point; having to handle both seems a bit overwhelming. Also, the namespace-"parade" is a commitment. 

> make this a filter hook which passes through a WP_Error object. 3PD could then use $error->add() to add errors to the shared instance.

Suppose we go in this direction, which seems like a promising alternative and, I'd say, much cleaner; we need to re-think the whole validation sequence.

**Introducing a shared wp error instance**

Keeping all the exceptions internally sounds reasonable if we _green-light_ the following:
- The `validate_xxx` methods will be responsible for _throwing_ all the exceptions.
- The `validate_xxx` methods should be clean from _try...catch_ blocks to further highlight this architecture.
- Exceptions will be caught from (a) the error getters and (b) the general `AbstractCartRoute::get_response` method as our last resort.

From a 3PD perspective, I would say that dealing with a filter can sometimes lead to unwanted effects:
- If you return nothing/null, the system will need to handle this type of validation.
- If you return a new `WP_Error` instance, that will "replace" all previous messages.

These seem acceptable as long as we only group all the 3PDs messages and the legacy notices.

As far as I understand, the broad idea behind the StoreAPI system is that someone can throw a `RouteException` that will end up in some errors array or UI context at any moment of new or existing legacy code. 

Definitely agree that we need to _draw the line_ on where and which exceptions should be publicly considered best practice.

This approach can simplify things for 3PDs by leaving them the following options:
- Throw a `RouteException` if things are going bad at anywhere, or;
- Return a `WP_Error` instance for multiple errors handling in validation-specific filters' context.

That said, we probably need to decide which points/hooks will be "_packing_" the "multi-errors" feature versus the single `RouteException` error.

We could introduce the shared `WP_Error` handling for more hooks than the `__experimental_woocommerce_store_api_validate_cart` to keep things consistent. For example:
- `wooocommerce_store_api_validate_cart_item`
- maybe more?

Moreover, the system will use the legacy(?) `woocommerce_coupon_is_valid` to validate coupons, so in order to be aligned, it should make sense to introduce a new `wooocommerce_store_api_validate_cart_coupon`. This could also leverage the shared wp error approach. Not sure if this is actually needed, though.

Thinking ahead probably, so I'll sum up:
- I agree that exposing more exceptions than the `RouteException` seems too _technical_ for 3PDs; the shared wp error keeps the technical barrier a bit lower, which is always a good thing.
- Handling multi-errors in specific only contexts using `WP_Error` instances sounds enjoyable.
- Having the ability to streamline error handling by throwing any known exception is a plus.
- If we go there, we probably need to commit to this approach in all related places in the code. 

Thoughts?

src/StoreApi/Utilities/InvalidCartException.php
xristos3490: Done!

xristos3490:
I tried to merge with `trunk` in order to fix the versioning namespace, but I got stuck on some unrelated PHPCS errors after the merge -- will look into this tomorrow.

**Update**: This branch is in line with `trunk`. Also updated the namespaces.

src/StoreApi/Utilities/CartController.php
mikejolley: Sounds like a fine proposal on the surface. Given the CartController classes are marked as `@internal` I think it will be fine to change the `validate_x` functions if it makes sense. 

I'd probably stick to one action or filter hook for the 3PD aspect of this - surfacing a WP_Error Object which they can append to. We cannot get around the usage of legacy hooks, so in those instances I'd just work with whatever data type they return and convert to exceptions or WP_Error. The docblock can explain the use of those filters is discouraged.

Then we can, finally, consolidate all errors into the single response with appropriate response code.

> From a 3PD perspective, I would say that dealing with a filter can sometimes lead to unwanted effects

The alternative which might work better is a registration system where you can register a callback. This callback function is then passed the cart and returns a WP_Error object. The advantage of this would be we can process each callback in sequence and validate that each returns a valid response.

src/StoreApi/Utilities/CartController.php
mikejolley: If they are not needed but let's finalise the new system first.

src/StoreApi/Utilities/CartController.php
xristos3490: > I'd probably stick to one action or filter hook for the 3PD aspect of this - surfacing a WP_Error Object which they can append to.

One filter sounds like a good plan. We are also introducing this new error-handling wrapped in an 'experimental__x' filter, making even more sense.

> We cannot get around the usage of legacy hooks, so in those instances I'd just work with whatever data type they return and convert to exceptions or WP_Error

Sounds good! Given that we can now handle multiple errors, a part of integration work for several plugins could be _plug-n-play_ with this approach.

> The alternative which might work better is a registration system where you can register a callback. 

As far as I understand, you are suggesting a custom `apply_filters` mechanism? Only this would be strict-typed, or is there perhaps an easier way to accomplish this?

I like this approach!

Dealing with recursive functions _(filters-in-filters, etc.)_ could require creating a new Utility class that will extend the `WP_Hook` class and introduce a method e.g., `apply_filters_to_wp_error()`. 

This pathway:
- Offers a simple way for 3PDs to use the `add_filter()` function.
- Protects the system and ensures that everybody will get a `WP_Error` argument no matter what they return.
- Covers recursive behavior and keeps the global `$wp_filters` intact.
- Has low maintenance due to this class's low frequency of changes -- but not _zero_, unfortunately.
- It leaves the ability for 3PDs to throw a `RouteException` to avoid the error-handling group.

I hope I'm not missing an obvious alternative here.

> It leaves the ability for 3PDs to throw a `RouteException` to avoid the error-handling group.

If we start gathering all errors from third parties, we may end up with a _not aesthetic_ number of error notices in the UI.

However, someone could throw a `RouteException` in the validate-filter context; this is a feature/bug that can be used to bypass the error grouping UX issue and respond with a _single_ error string. 

Trying to group and display all 3PDs errors/notices, **without the one interrupting the other** seems like a solid trade-off for this possible UX issue, and a nice plan to promote. Having the exception _shortcut_ is also always a plus.

mikejolley: We can probably leave the error handling to the client. As long as we can distinguish errors by their error codes. Critical errors should probably just throw Exceptions however, so there is only one shown.

> As far as I understand, you are suggesting a custom apply_filters mechanism? Only this would be strict-typed, or is there perhaps an easier way to accomplish this?

To clarify what I meant, you'd have a single `apply_filter` named something like `cart_validation_callbacks`, and the consumers would return functions there. We'd have an array of callbacks to process.

So if we want with the callback filter idea, a callback would be added something like this:

```
add_filter( 'cart_validation_callbacks', function( $callbacks ) {
     $callbacks[] = 'some_callback_function';
}

function some_callback_function( $cart ) {
   // Do something here...
   return new WP_Error();
}
```

Advantage would be, in the Store API side we can ensure each callback returns a valid `WP_Error` object.

Disadvantage here is extra work to add a callback, and callbacks could be misused or broken. Just like before, so it might not be worth the complication.

We could avoid filters completely with a registration function, similar to the ExtendRestAPI class approach: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Domain/Services/ExtendRestApi.php#L291 but again, this is extra complication which may not be worth it.

What are your thoughts?

xristos3490: > Overcomplicated?

Yes and no. In my opinion, having the opportunity to start over and reduce the frustration is a big plus for us.

On the plus side, if we can introduce a way that everyone would get a `WP_Error` to work with, without needing to deep dive into the hooks sequence to search for the _one_ piece of code that ruins everything, I'd say let's go.

On the other hand, all this is starting to feel like we are moving forward from the typical `apply_filters` system, which isn't necessarily bad. Some thoughts on this approach, though:
- 3PDs can no longer use a _priority_ in their callbacks; this could lead to a race to who comes first.
- Requires a new way of _hooking-in_; instead of traditional filters and actions.
- Could lead to a technical barrier that is equally frustrating as we started this analysis (having to deal with exceptions).

We have to decide on this trade-off here. We could also develop a global Utility class for handling "callback queues" and reuse it. But the main question is about _convenience_ and _robustness_ of extensibility needs.

In my opinion, this custom payments callback queue is an excellent piece of work when handling complex data/schema operations; I'm just not sure if we need this in a more straightforward case like this. Also, the need for a validation process may be more complex when dealing with multiple business rules. 

For example, someone might need to wait for "Points and Rewards"-discounts validation to kick in before validating his own thing in an _assumed_ final total. This example should highlight more the need for having the _priority_ feature.

Thoughts? 

**Update:** ~~A system that will inherit the `WP_Hook` class could work here, but I definitely understand that this might also be "too complicated."~~
The `WP_Hook` class is a `final` class, so this isn't an option after all.

xristos3490: To sum up our options here:
- Introduce a validation callback queue and perhaps revisit later to introduce _priority_ or even a Utility _singleton_.
- Work with traditional filters and leave the type handling to everyone's need to _follow the rules_.

At this point, both should work with or without revisits, but the first one is a bit _radical_.

mikejolley: @xristos3490 I toyed with a callback registry for this (fire action passing registry class, others use `$registry->register()` to add a callback, then callbacks are processed) and it worked but I think it is a bit convoluted...

I think all things considered (trying to balance robustness with each of use), let's do a **filter** hook with the following caveats:

1. The filter hook provides a `WP_Error` object. Extensions can add errors to this object for later handling.
2. The filter hook is wrapped in try/catch to catch any exceptions.
3. The filter hook must return `WP_Error` - if it doesn't we need an exception.

Along these lines:

```php
$cart_errors = apply_filters( '__experimental_woocommerce_store_api_cart_errors', new \WP_Error() );

if ( ! is_wp_error( $cart_errors ) ) {
	// Throw exception.
}

if ( $cart_errors->has_errors() ) {
	throw new RouteException( 'cart_error', $cart_errors->get_error_message() );
}
```

We'll need handling for multiple errors (I think we already have work in place for this), and it would be consumed like:

```php
add_action(
				'__experimental_woocommerce_store_api_cart_errors',
				function( \WP_Error $errors ) {
					$errors->add( 'some_error', 'I am broken.' );
					return $errors;
				}
			);
```

src/StoreApi/Utilities/CartController.php
xristos3490: Should we deprecate the:
- `get_cart_item_errors` and;
- `get_cart_coupon_errors` methods?

Currently, the only ref on these methods is a single Unit Test that uses the `get_cart_item_errors` and will be changed to `get_cart_errors()`.

After deprecating these, we can go ahead and return a single `WP_Error` instance in the `get_cart_errors()` as well.

src/StoreApi/Utilities/CartController.php
xristos3490: I also experimented with a simple queue runner logic, and I concluded to the same judgment. Here's the main code part for reference:
```php
// Given a WP_Error $global_error.
$queue = (array) apply_filters( 'api_cart_validation_callbacks', array() );
foreach ( $queue as $cb ) {
    if ( ! is_callable( $cb ) ) {
        continue;
    }

    $error = call_user_func_array( $cb, array( $cart ) );
    if ( is_a( $error, 'WP_Error' ) ) {
        $global_error->merge_from( $error );
    }
}
```
One could potentially find their way into "hacking" this system with the same _severity_ as the legacy hooks system. So, I definitely agree with your spottings! 

I did a [commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5904/commits/792623392d9c06bd32049fd96580c812b0ec4779) that adds the `__experimental_woocommerce_store_api_cart_errors
` filter.

> The filter hook is wrapped in try/catch to catch any exceptions.

Can you please elaborate on this? I think that we could leave the filter unwrapped. This way, the corresponding wrapper _(either the cart getter or the route response handler)_ will catch all the exceptions.

Other notes:
- Notices are grouped in a single `WP_Error` instance and handled accordingly. @ [commit](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5904/commits/501132bdde0e3624f5e36d136c1e5192ab9201f0)
- The previous `NoticeHandler::convert_notices_to_exceptions` is removed -- perhaps we need to add a deprecation warning?
- We could deprecate the previous error getters -- see this [comment](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5904/files#r816006972)

Lastly, what are your thoughts about the existing `wooocommerce_store_api_validate_cart_item` hook?

Every purpose of this action could probably be handled using the new  `__experimental_woocommerce_store_api_cart_errors`, and we could safely remove/deprecate it to avoid confusion.

At this point, the code in this PR should be aligned with our discussion.

src/StoreApi/Utilities/NoticeHandler.php
```diff
@@ -1,31 +1,33 @@
...
-	 * This method will find the first error message and thrown an exception instead. Discards notices once complete.
-	 *
-	 * @throws RouteException If an error notice is detected, Exception is thrown.
+	 * This method will discards notices once complete.
 	 *
 	 * @param string $error_code Error code for the thrown exceptions.
+	 *
+	 * @return \WP_Error The WP_Error object containing all error notices.
 	 */
-	public static function convert_notices_to_exceptions( $error_code = 'unknown_server_error' ) {
```
mikejolley: There are other consumers of this such as legacy payments API. This needs to stay in place (tests too), with the wp_error version alongside.

mikejolley: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/8c7c9009a1a7cd52165d6d1a381c7c0607649f5b/src/Payments/Api.php#L152

src/StoreApi/Utilities/CartController.php
```diff
@@ -360,15 +360,78 @@ private function get_error_message_for_stock_exception_type( $exception_type, $s
...
+		 * @param WP_Error $errors  WP_Error object.
+		 * @param WC_Cart  $cart    Cart object.
+		 *
+		 * @return WP_Error         The WP_Error object returned.
+		 */
+		$cart_errors = apply_filters( '__experimental_woocommerce_store_api_cart_errors', new WP_Error(), $cart );
+
+		if ( ! is_wp_error( $cart_errors ) ) {
+
+			throw new RouteException( 'missing_cart_validation_server_error', __( 'There was an error with your cart.', 'woo-gutenberg-products-block' ) );
```
mikejolley: Maybe we should switch to a developer facing error here, along these lines: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/bbe5b3f9c909139b59c3bea57abd52fb77e1b7ca/src/StoreApi/Schemas/ExtendSchema.php#L311-L319

That way customers won't see strange issues if a developer makes a mistake in a filter, and we can specifically state that consumers of the `__experimental_woocommerce_store_api_cart_errors` filter must return the original `WP_Error` object.

src/StoreApi/Utilities/CartController.php
```diff
@@ -360,15 +360,78 @@ private function get_error_message_for_stock_exception_type( $exception_type, $s
...
+		}
+
+		if ( $cart_errors->has_errors() ) {
```
mikejolley, line 405: Based on the above comment, this would need to be wrapped part of the `is_wp_error` conditional.

src/StoreApi/Utilities/CartController.php
```diff
@@ -388,92 +453,54 @@ public function validate_cart_items() {
...
-				409,
-				$this->add_product_names_to_message( $singular_error, $plural_error, $too_many_in_cart_products )
-			);
-		}
+		if ( count( $errors ) > 0 ) {
 
-		if ( count( $partial_out_of_stock_products ) > 0 ) {
-			$singular_error = $this->get_error_message_for_stock_exception_type( 'partial_out_of_stock', 'singular' );
-			$plural_error   = $this->get_error_message_for_stock_exception_type( 'partial_out_of_stock', 'plural' );
+			$error = new WP_Error();
```
mikejolley, line 472: Might this be useful here? https://developer.wordpress.org/reference/classes/wp_error/merge_from/

src/StoreApi/Utilities/CartController.php
mikejolley: yes if unused now.

src/StoreApi/Utilities/CartController.php
xristos3490: It won't make _much_ of a difference because we have a `WP_Error[]` type. But it's a lot _cleaner_! :)

src/StoreApi/Utilities/NoticeHandler.php
xristos3490: Oops! Missed that! Reverting...

src/StoreApi/Utilities/CartController.php
xristos3490: In this error case:
- We are dealing with potentially more error messages that were _lost_ in the process, and;
- An invalid cart state could be purchased without issues.

On the other hand, _throwing_ an exception here would disable the store altogether -- no orders can be made either you have errors or don't. 

We might need to consider this as a developer _bug_ that should be handled. I feel like it doesn't worth the risk. Thoughts? 

> we can specifically state that consumers of the __experimental_woocommerce_store_api_cart_errors filter must return the original WP_Error object.

Love this feedback! Regardless of the final message, I thought we could check the original object using the [spl_object_id](https://www.php.net/manual/en/function.spl-object-id.php) lifetime object id. This way, we can always be sure that no other messages are lost!

**Update**: We might want to use the [spl_object_hash](https://www.php.net/manual/en/function.spl-object-hash.php) instead, if we want to support PHP5. The `spl_object_id` is only available for PHP 7+.

mikejolley: @xristos3490 Just thought, would it work to create an WP_Error instance and pass it via action hook? That would work as a reference right? Sorry to backtrack.

xristos3490: > would it work to create an WP_Error instance and pass it via action hook? That would work as a reference right?

Yes! Calling a simple `do_action` should work by the definition of how PHP handles objects. As long as we use a `WP_Error` instance, every callback in the action's queue will have the same object to handle/add errors in.

This way we can be sure that no one could replace the main iterated error object, but; anyone could _replace/remove/update_ all previous errors.

Although, messing with the error codes can be done with both the `apply_filters` and the `do_action`, approach. 

To sum up, having a single `do_action` that:
- Consolidates all errors,
- Doesn't require anything for the `return` statement, so it can't be broken, and;
- Depends on how PHP works with objects in memory.

But:
- Can feel a bit weird when changing something but never returning it.
- Everyone could still mess with the all errors.

should work perfectly as well!

PS. Nice thinking out of the box! 🙇‍♂️

mikejolley: Should be safer, and we'll have inline documentation. Let's go that route. It means we won't have to do any fancy comparison with the WP_Error once the hook has finished running.

mikejolley (approved):
Changes look good now, up to date with trunk, smoke tested the new (and old) hooks, 🚢

PR #5909, opened 2022-02-21
Remove hover effect from disabled Mini Cart button

Aljullu:
Removes the `:hover` effect from the Mini Cart button when it's disabled.

### Manual Testing

1. Go to Appearance > Site Editor and add the Mini Cart block to the header.
2. Go to the Cart or Checkout page and hover the Mini Cart block.
3. Verify the button has no `:hover` effect.

Before | After
--- | ---
Kooha-02-21-2022-16-27-38 | Kooha-02-21-2022-16-27-00


tjcafferkey (approved):
LGTM

PR #5910, opened 2022-02-21
Correctly load JS translations from WC core

Aljullu:
While doing some testing, I noticed translations of JS files were not correctly loaded in some locales. The locale needed to have the strings translated in WC core but not in WC Blocks.

### Manual Testing

1. Go to Settings > General > Language (`/wp-admin/options-general.php`) and select a language which has WC core translation at 100% but WC Blocks' at 0%. (I tested with `Català`).
2. Go to the updates screen (`/wp-admin/update-core.php`), scroll down and update the language translations.
3. Visit the Cart or Checkout block in the frontend and verify the texts are translated.

Before | After
--- | ---
imatge | imatge

4. Switch to a language where WC core translations already exist (ie: `Deutsch`, `Español` or `Français`) and verify translations are still loaded correctly (you will need to go to the updates screen again).

### Changelog

> Fix loading WC core translations in locales where WC Blocks is not localized for some strings.

### DIFF
--- a/woocommerce-gutenberg-products-block.php
+++ b/woocommerce-gutenberg-products-block.php
@@ -205,7 +205,7 @@ function woocommerce_blocks_get_i18n_data_json( $translations, $file, $handle, $
 JS;
 
 	// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.NoExplicitVersion
-	wp_register_script( $handle_filename, '', array(), false, true );
+	wp_register_script( $handle_filename, '', array( 'wp-i18n' ), false, true );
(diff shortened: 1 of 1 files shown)


Aljullu:
Since this is Cart & Checkout related, I assigned it to you (@mikejolley) for review. Hope you don't mind. :slightly_smiling_face:

mikejolley (approved):
Hey @Aljullu Code looks good but there was an issue causing translations not to work because the scripts were output too early. I added the `wp-i18n` dependency which fixed it.

Issue #5707, opened 2022-02-02
Stabilize the Store API for core inclusion


mikejolley:
This is an Epic to track issues relating to the Store API and its move from experimental to stable.

PR #5911, opened 2022-02-21
Store API: Introduce API Versioning System

mikejolley:
This is a large part of #5707 which introduces versioning to the Store API, meaning endpoints can now be reached via `wc/store/v1/` as well as `wc/store/`. This does not break the existing structure which continues to map to `v1`.

What this PR does:

1. Moves all endpoints (routes and schema) under `V1` directories and PHP namespaces. Note, to maintain backwards compatibility I've maintained a list of previous namespaces which are autoloader by composer (see `deprecated.php`). This structure change was needed so that in the future, if a new version is introduced it can go under `V2` namespace and so on.
2. Registers all REST API routes (using `register_rest_route`) using `wc/store/v1` namespace. This code also registers the original `wc/store` namespace as an alias, to avoid breaking existing routes.
4. Updates unit tests to test against `v1` routes and to confirm all routes are indeed registered.
5. Updates documentation to include guiding principles and to include all updated endpoints. 

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests

### Manual Testing

1. Smoke test checkout process to highlight any problems with cart/checkout endpoints
2. Smoke test all products block to highlight any problems with product endpoints
3. Try accessing old endpoints, for example, `https://store.local/wp-json/wc/store/cart`. These should still work.
4. Try accessing new endpoints, for example, `https://store.local/wp-json/wc/store/v1/cart`

Unit tests cover schema and route registration.

### Changelog

> Store API - Introduced `wc/store/v1` namespace.


senadir:
If we're going to have `v2` would this require us to duplicate all existing code, or inherent and overwrite classes?

mikejolley:
@senadir either, it doesn’t matter. There is no way around it. For minor changes extending would work fine. 

That’s why we must always encourage evolution to avoid new versions and avoid breaking changes.

src/StoreApi/RoutesController.php
```diff
@@ -40,70 +39,92 @@ public function __construct( SchemaController $schemas ) {
...
+	 * @return int
 	 */
-	protected function initialize() {
-		global $wp_version;
+	protected function get_route_version() {
+		preg_match( '/\/v(?P<version>\d)\//', untrailingslashit( $GLOBALS['wp']->query_vars['rest_route'] ), $matches );
 
-		$cart_controller  = new CartController();
-		$order_controller = new OrderController();
+		return (int) $matches['version'] ?? 1;
```
alexflorisca: Is it worth extracting the default version into a variable somewhere in the top scope so that it can be changed in one place later down the line when we want to make `v2` default. I've seen similar logic to this in a few places.

tests/php/StoreApi/Routes/Batch.php
```diff
@@ -88,32 +84,32 @@ public function test_success_cart_route_batch() {
 	 * Test for a mixture of successful and non-successful requests in a batch.
 	 */
 	public function test_mix_cart_route_batch() {
-		$request  = new \WP_REST_Request( 'POST', '/wc/store/batch' );
+		$request = new \WP_REST_Request( 'POST', '/wc/store/v1/batch' );
```
alexflorisca, line 87: If we're calling the versioned api in these tests, should we not have `v1` folder for the tests as well? How would one test a new `v2` endpoint?

alexflorisca (commented):
Nice work @mikejolley, there's a lot in this PR! The code looks good although with the caveat that my PHP isn't that strong, so I'd strongly encourage someone else to review as well. That's why i'll leave my review as a comment. I've also added a couple of inline comments

I've smoke tested the All Products block, Cart and Checkout and all seems ok. I'm not seeing any failed requests when filtering products, adding/removing, changing quantity, updating addresses, etc and everything works as expected 👍 

I'm assuming that we're not updating all the client side requests to use the new endpoints in this PR?

src/StoreApi/RoutesController.php
```diff
@@ -40,70 +39,92 @@ public function __construct( SchemaController $schemas ) {
...
 
 	/**
-	 * Load route class instances.
+	 * Get the requested version of the Store API from the route request.
+	 *
+	 * @return int
 	 */
-	protected function initialize() {
-		global $wp_version;
+	protected function get_route_version() {
```
opr: This is duplicated in `src/RestApi.php` is it possible to move this to a central place to avoid duplication?

src/StoreApi/RoutesController.php
mikejolley: We won't be making `v2` the default because that would break existing requests. The unversioned endpoints should always map to v1.

alexflorisca: Makes sense 👍

src/StoreApi/docs/guiding-principles.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Error Handling
+
+Errors, including validation errors, should return an error response code (4xx or 5xx) and a [`WP_Error` object](https://developer.wordpress.org/reference/classes/wp_error/). The `AbstractRoute` class will handle the conversion of the `WP_Error` object into a valid JSON response.
+
+Error messages should be localized, but do not need to be written with language aimed at customers (clients should use the given error code to create customer-facing notices as needed).
+
+### Cart Operations
+
+Some endpoints are designed around operations to avoid client’s needing to make multiple round trips to the API. This is purely for convenience.
```
opr:
```suggestion
Some endpoints are designed around operations to avoid clients needing to make multiple round trips to the API. This is purely for convenience.
```

src/StoreApi/docs/cart-items.md
```diff
@@ -6,6 +6,7 @@
 - [Edit Single Cart Item](#edit-single-cart-item)
 - [Delete Single Cart Item](#delete-single-cart-item)
 - [Delete All Cart Items](#delete-all-cart-items)
+- [](#)
```
opr:
```suggestion

```

tests/php/StoreApi/Routes/Cart.php
```diff
@@ -63,120 +63,135 @@ public function setUp() {
...
+					'total_price'                 => '2900',
+					'tax_lines'                   => array(),
+				),
+			)
+		);
 	}
 
 	/**
-	 * Test removing a nonexistent cart item.
+	 * Test removing a nonexistent cart item. This should return 404.
```
opr: This comment does not match the test on line 118. The tests pass, so it's the comment that should be updated. 404 should be returned when the _route_ doesn't exist, right?

tests/php/StoreApi/Routes/Cart.php
```diff
@@ -63,120 +63,135 @@ public function setUp() {
...
 
 		// Test removing same item again - should return 404 (item is already removed).
-		$response = rest_get_server()->dispatch( $request );
-		$data     = $response->get_data();
-
-		$this->assertEquals( 409, $response->get_status() );
-		$this->assertEquals( 'woocommerce_rest_cart_invalid_key', $data['code'] );
+		$this->assertAPIResponse(
+			$request,
+			409,
```
opr, line 154: Same comment/code mismatch

tests/php/StoreApi/Routes/ControllerTestCase.php
```diff
@@ -46,4 +54,106 @@ public function tearDown() {
...
+	 * @return \WP_Rest_Response
+	 */
+	public function getApiResponse( $endpoint ) {
+		return rest_get_server()->dispatch( new \WP_REST_Request( 'GET', $endpoint ) );
+	}
+
+	/**
+	 * Asserts that the given fields are present in the given object.
+	 *
+	 * @since UT (3.7.0)
```
opr: What is this?

src/StoreApi/RoutesController.php
```diff
@@ -40,70 +39,92 @@ public function __construct( SchemaController $schemas ) {
 	/**
 	 * Get a route class instance.
 	 *
-	 * @throws \Exception If the schema does not exist.
+	 * Each route class is instantized with the SchemaController instance, and it's main Schema Type.
```
opr:
```suggestion
	 * Each route class is instantized with the SchemaController instance, and its main Schema Type.
```

opr (changes requested):
Nice work! 😎 Functionality-wise, everything works fine on my test site, I've tried using the site itself and postman to mess with the API and everything I tried works OK.

As for the code, I am seeing a lot of remaining instances of unversioned API routes - I know this isn't a problem as they map to v1, but do you think we should be mixing v1 and unversioned in our own codebase? We should keep it consistent imo. For testing, it's fine, but in the actual working code maybe not?

Also, sorry but I added a couple of comments about apostrophes 🙈

tests/php/StoreApi/Routes/Batch.php
mikejolley: In this case I didn't see a reason to re-namespace the tests. The code is not public, so if there is ever a v2 there can be a decision either to add directories and separate the tests, or add v2 tests inline alongside existing v1 tests.

src/StoreApi/RoutesController.php
mikejolley: I've added a note about this in the code.

src/StoreApi/RoutesController.php
mikejolley: I'll remove both. I was using this to register only the requested version, but in the end I registered everything instead.

mikejolley:
@opr I've gone ahead and switched over usage to `v1` now and addressed feedback. Thanks for the corrections.

opr:
> @opr I've gone ahead and switched over usage to `v1` now and addressed feedback. Thanks for the corrections.

Thanks @mikejolley - I found these too

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/607bd25bcbcbef25920a88cf0e50b2377ce11c4e/assets/js/editor-components/utils/index.js#L125-L136

do they need to change to v1?

mikejolley:
@opr ah, I missed the ones without a slash at the start.

mikejolley:
Got some new test fails to resolve :( Will look tomorrow.

opr (approved):
Thanks for fixing those unversioned routes and my other feedback. The tests are 🟢 now so I am approving! 🙌🏼

pluginslab:
When using `wp.data.select( 'wc/store/cart' );` or `useSelect()` for example, how could we support backwards (without version number) and forwards (when v2 is coming) compatibility with this change? Getting information about the cart using this method avoids a new network call. Should we add code for that support?

mikejolley:
@senadir what are your thoughts on the above? Is going direct to wp.data something we need to support?

Issue #5913, opened 2022-02-22
E2E test: Storefront isn't always active.


dinhtungdu:
## Describe the bug

The E2E test is failing on the `trunk` after merging #5877. The error belongs to a utility introduced in #5877, the `shopper.goToBlockPage` helper.

#5877 was passed before merging to trunk. Digging this issue deeper, I found that the Storefront theme isn't always active in the test environment, please see the screenshots for more details. 

`shopper.goToBlockPage` was created with an assumption that we always use Storefront for testing. But in fact, on the trunk, the test environment uses Twenty Twenty One. The Storefront was installed, but not activated.

I can't reproduce this issue on localhost.

Note: `shopper.goToBlockPage` needs to be restored to the [previous version](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5877#pullrequestreview-887416788) to ensure it works with any theme.

## Expected behavior

Storefront should be active for test environment. In the other words, the test environment on local and CI should be the same.

## Screenshots

The following screenshots were taken on GitHub Actions.

![cart](https://user-images.githubusercontent.com/5423135/155059359-7f06db2d-6f02-4581-8b35-864623a071a4.png)
![checkout](https://user-images.githubusercontent.com/5423135/155059366-273b4d71-2894-46bc-ad80-932105f84596.png)
![test](https://user-images.githubusercontent.com/5423135/155059368-48fc0d9c-b9df-400a-b358-e1de68856a5a.png)
![themes](https://user-images.githubusercontent.com/5423135/155059373-eda5a649-2515-40de-95ea-965a4d8e7f53.png)

dinhtungdu:
Closing in favor of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5914#pullrequestreview-889277707

PR #5914, opened 2022-02-22
Fix: Ensure `shopper.goToBlockPage` works with any theme

dinhtungdu:
Relates #5913 

The `shopper.goToBlockPage` was created with the assumption that the test environment is always using the Storefront theme, we use that assumption to [speed up the test](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5877#pullrequestreview-887416788). #5913 shows that the test environment doesn't always use Storefront.

While we need to investigate why the Storefront theme doesn't active for the test environment, I think it's better for a utility to work with any theme. 

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

See the GH Actions passing (Running the E2E tests locally doesn't have any issue. Only tests on GH Actions are failing. )

### DIFF
--- a/tests/utils/shopper.js
+++ b/tests/utils/shopper.js
@@ -2,7 +2,6 @@
  * External dependencies
  */
 import { shopper as wcShopper } from '@woocommerce/e2e-utils';
-import { createURL } from '@wordpress/e2e-test-utils';
 
 /**
  * Internal dependencies
@@ -49,11 +48,9 @@ export const shopper = {
 	},
 
 	goToBlockPage: async ( title ) => {
-		await page.goto( createURL( '/' ), { waitUntil: 'networkidle0' } );
-
-		await expect( page ).toClick( '.nav-menu a', { text: title } );
-
-		await page.waitForNavigation();
+		await page.goto( await getBlockPagePermalink( title ), {
+			waitUntil: 'networkidle0',
+		} );
 
 		await expect( page ).toMatchElement( 'h1', { text: title } );
 	},


sunyatasattva (approved):
The changes look good to me and, as you say, more solid.

> While we need to investigate why the Storefront theme doesn't active for the test environment, I think it's better for a utility to work with any theme.

Regarding this, actually, we shouldn't assume that Storefront is active in test environment. In fact, as Storefront is not a block theme, Storefront _should not_ be the default theme going forwards for the test environment, as we won't be able to test the Site Editor otherwise.

Moreover, because we need to [test more complex scenarios on the FSE front](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5660), we'll probably need more themes or variations.

So, in my opinion, tests shouldn't make any assumption on what the theme is as they run, but they should declare their own dependencies, so to speak.

As a matter of fact, what I do here is the following:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/bdc299a2b2a1ac59ed2bdc30488c84fd5f98c2de/tests/e2e/specs/backend/site-editing-templates.test.js#L109-L117

Make sure that I have the theme I need (in this case "Empty theme" before I run all my tests), and then I switched the theme back to TT1 after them. Perhaps that's what broke the tests assuming Storefront.

There are two ways I see in this:

1. I could change the `afterAll` code there to make sure it switches back to the theme which was active previously.
2. We could remove that `afterAll` hook which was meant as a sort of “teardown” step to reset the state of the site, and make sure each test actually declare its own “theme dependencies”, so to speak, in the `beforeAll` hook.

Well, a third way is we do both of these things.

dinhtungdu:
> I could change the afterAll code there to make sure it switches back to the theme which was active previously.

@sunyatasattva This makes more sense to me. By doing so, we don't affect other test cases.

Issue #5881, opened 2022-02-17
Mini Cart block: additional classes don't add on the frontend side


gigitux:
## Describe the bug

The addition classes don't render on the frontend side.

## To reproduce

1. Open the FSE editor.
2. Add the Mini Cart block.
3. Get focused on it.
4. On the right sidebar, on the block -> advanced section, add custom classes.
5. Save.
6. On the frontend page, inspect the Mini Cart Block.
7. The custom classes aren't added.

## Expected behavior

The custom classes should be added on the frontend side

## Screenshots
![image](https://user-images.githubusercontent.com/4463174/154496384-74f4d913-0922-4353-aa5d-78cd13f2b240.png)

dashkevych:
Will this fix be included in WooCommerce plugin itself (without active WooCommerce Blocks plugin)?

It looks like the issue is present in WooCommerce (6.7.0)

gigitux:
> Will this fix be included in WooCommerce plugin itself (without active WooCommerce Blocks plugin)?
> 
> It looks like the issue is present in WooCommerce (6.7.0)

Thanks for this comment. Unfortunately, there was a regression. I fixed it in #6716. For now, I will reopen this issue.

PR #5916, opened 2022-02-22
Load MiniCart template part using the correct template ID

### LINKED ISSUE #5883: Mini Cart Contents template: the blocks added by the user are not visible on frontend side
## Describe the bug

Mini Cart Contents template: the blocks added by the user are not visible on the frontend side 

## To reproduce

1. Open the FSE editor.
2. Open the Mini Cart template.
3. Add a block in the template (add the blocks to the `Mini Cart Items` or `Empty Mini Cart Contents` .
4. Save it.
5. Go on the frontend side.
6. The added block is not visible.

## Expected behavior

The added block should be visible.

tjcafferkey:
Previously we were loading the MiniCart template part using `get_stylesheet() . '//mini-cart'` which would be `twentytwentytwo//mini-cart`. This is wrong as the Mini Cart should have the WooCommerce slug. I've updated this now so it should evaluate `woocommerce/woocommerce//mini-cart` like the rest of the templates.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5883

### Manual Testing

1. Before checking out this PR clear all Mini Cart customizations
2. Checkout this PR
3. Check Mini Cart renders correctly inside the Site Editor and on the frontend
4. Make some customizations and save them.
5. Check Mini Cart renders the customizations inside the Site Editor and on the frontend correctly.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

1.
2.
3.

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -377,7 +377,7 @@ class MiniCart extends AbstractBlock {
 		}
 
 		$template_part_contents = '';
-		$template_part          = BlockTemplateUtils::get_block_template( get_stylesheet() . '//mini-cart', 'wp_template_part' );
+		$template_part          = BlockTemplateUtils::get_block_template( BlockTemplateUtils::PLUGIN_SLUG . '//mini-cart', 'wp_template_part' );
 
 		if ( $template_part && ! empty( $template_part->content ) ) {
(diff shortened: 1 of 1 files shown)


gigitux (approved):
Thanks for working on this!
It works correctly! 

This PR fixes #5883, but it doesn't fix #5881 

:shipit:

PR #5919, opened 2022-02-22
Ensure screenshots can be attached in jest html report

opr:
The signature for attaching images via `addAttach` in `jest-html-reporters` was changed here https://github.com/Hazyzh/jest-html-reporters/commit/ad7dd8c2486114e09b78781a38b2e51903ea19e8 and now accepts  an object.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

1. Push a commit here that will cause the e2e tests to error, for example change the state on this line to one that doesn't exist.
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/35c5161a3e95901a1668f4360d5ae0227986a117/tests/e2e/specs/frontend/checkout.test.js#L50-L53
2. When e2e tests finish running, go to the artifacts section of the job and download the zip file.
3. Open the report, and ensure there's an image of the screen captures on the failing test.
4. Revert the bad commit.

Alternatively, view the artifacts from here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/actions/runs/1886571707 and see the report

### User Facing Testing
No user-facing testing required.

### DIFF
--- a/tests/e2e/config/environment.js
+++ b/tests/e2e/config/environment.js
@@ -8,15 +8,14 @@ class E2EEnvironment extends PuppeteerEnvironment {
 	async handleTestEvent( event ) {
 		if (
 			event.name === 'test_fn_failure' ||
-			event.asyncError !== undefined
+			event.name === 'hook_failure'
 		) {
 			const attach = await this.global.page.screenshot();
-			console.log( 'taking screenshot' );
 			await addAttach( {
 				attach,
 				description: 'Full Page Screenshot',
 				context: this.global,
-				bufferFormat: 'utf8',
+				bufferFormat: 'png',
 			} );
 		}
 	}
--- a/tests/e2e/config/environment.js
+++ b/tests/e2e/config/environment.js
@@ -6,7 +6,10 @@ const { addAttach } = require( 'jest-html-reporters/helper' );
 
 class E2EEnvironment extends PuppeteerEnvironment {
 	async handleTestEvent( event ) {
-		if ( event.name === 'test_fn_failure' ) {
+		if (
+			event.name === 'test_fn_failure' ||
+			event.asyncError !== undefined
+		) {
 			const attach = await this.global.page.screenshot();
 			console.log( 'taking screenshot' );
 			await addAttach( {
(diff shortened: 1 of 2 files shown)


tests/e2e/config/environment.js
```diff
@@ -6,9 +6,17 @@ const { addAttach } = require( 'jest-html-reporters/helper' );
...
 class E2EEnvironment extends PuppeteerEnvironment {
 	async handleTestEvent( event ) {
-		if ( event.name === 'test_fn_failure' ) {
-			const data = await this.global.page.screenshot();
-			await addAttach( data, 'Full Page Screenshot', this.global );
+		if (
+			event.name === 'test_fn_failure' ||
+			event.name === 'hook_failure'
+		) {
+			const attach = await this.global.page.screenshot();
```
dinhtungdu: I think a full-page screenshot here will me more useful:
```suggestion
			const attach = await this.global.page.screenshot( {
				fullPage: true,
			} );
```

dinhtungdu (commented):
@opr The screenshots were taken but some of them weren't attached to the HTML report. Testing locally with a failed test case, I see the image in the `jest-html-reporters-attach` folder but I don't see it in the HTML report.

The artifact you mentioned [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/actions/runs/1886571707) has the same issue. There are two failed tested cases, but one has a screenshot attached, one doesn't.

tests/e2e/config/environment.js
opr: Using `fullPage: true` stops it from capturing the right page for me, it should be capturing the back-end WC settings page for the case I tested with, but `fullPage: true` makes it capture the front-end cart page?

Also, I think in the artifact with missing images, this might be caused by the error happening in the `beforeAll` hook, it seems like the error is happening before it even gets into a test case which is why the image doesn't attach.

I made it so it only takes a full page screenshot when the event name is not `hook_failure`. This seems to show us the right page in `beforeAll`, `before`, `after`, and `afterAll` errors, and the right page in regular test failures.

opr: Also, what are you changing locally to get this:
>Testing locally with a failed test case, I see the image in the jest-html-reporters-attach folder but I don't see it in the HTML report.

Thanks!

dinhtungdu (approved):
Works like a charm! 🎉  :shipit:

PR #5888, opened 2022-02-18
E2E tests for Product Tag Template and Legacy Block

sunyatasattva:
This PR adds E2E tests for the Product Tag Template

Specifically, it covers the following test cases:

- [x] Should a user navigate to Site Editor > All Templates with an FSE theme active they should see the WooCommerce provided Product Tag block template.
- [x] Should a user load the Product Tag template, the editor will have the “WooCommerce Product Grid Block” Legacy Template block present.
- [x] Should a user load a product tag page on the frontend, it will render a list of products.
- [x] Should a user customise the Product Tag template and save it. The user should see a "Clear customizations" option on the All Templates view (clearing customisations functionality should be covered by Gutenberg)
- [x] Should a user load the customised Product Tag template the Site Editor should render these customisations
- [x] Should a user have a customised Product Tag template, these customisations would be present on the frontend.

Closes #5659

---

Note that, as opposed to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5722, where I opted not to test the functionality of **both** the template and the block (I tested only the template-related functionality, see discussion there), here I also tested the frontend part for the block itself, due to the fact that the test was fairly simply and straightforward to write.

#### Automated Tests
* [x] Changes in this PR are covered by Automated Tests.
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

How to test the changes in this Pull Request:

1. Remember to `npm i` when you pull, as there is a requirement for the newest `e2e-test-utils`.
2. Run `npm run test:e2e -- -i "tests/e2e/specs/backend/site-editing-templates.test.js" -t "Product Tag block template"` and `npm run test:e2e -- -i "tests/e2e/specs/frontend/legacy-template-blocks.test.js" -t "Product Tag block"`.
3. Assess that the included test cases pass.

### User Facing Testing
These are steps for user testing (where "user" is someone interacting with this change that is not editing any code).
* [x] Same as above

PR #5820, opened 2022-02-11
E2E tests for Product Category Template and Legacy Block

sunyatasattva:
This PR adds E2E tests for the Product Category Template

Specifically, it covers the following test cases:

- [x] Should a user navigate to Site Editor > All Templates with an FSE theme active they should see the WooCommerce provided Product Category block template.
- [x] Should a user load the Product Category template, the editor will have the “WooCommerce Product Grid Block” Legacy Template block present.
- [x] Should a user load a product category page on the frontend, it will render a list of products.
- [x] Should a user customise the Product Category template and save it. The user should see a "Clear customizations" option on the All Templates view (clearing customisations functionality should be covered by Gutenberg)
- [x] Should a user load the customised Product Category template the Site Editor should render these customisations
- [x] Should a user have a customised Product Category template, these customisations would be present on the frontend.

Closes #5657

---

Note that, as opposed to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5722, where I opted not to test the functionality of **both** the template and the block (I tested only the template-related functionality, see discussion there), here I also tested the frontend part for the block itself, due to the fact that the test was fairly simply and straightforward to write.

#### Automated Tests
* [x] Changes in this PR are covered by Automated Tests.
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

How to test the changes in this Pull Request:

1. Remember to `npm i` when you pull, as there is a requirement for the newest `e2e-test-utils`.
2. Run `npm run test:e2e -- -i "tests/e2e/specs/backend/site-editing-templates.test.js"` and `npm run test:e2e -- -i "tests/e2e/specs/frontend/legacy-template-blocks.test.js"`.
3. Assess that the included test cases pass.

### User Facing Testing
These are steps for user testing (where "user" is someone interacting with this change that is not editing any code).
* [x] Same as above

PR #5748, opened 2022-02-08
E2E tests for Product Archive Template and Legacy Block

sunyatasattva:
This PR adds E2E tests for the Product Archive Template, and refactors a few things used in the Single Product Template which are useful for other Template tests.

Specifically regarding #5657, it covers the following test cases:

- [x] Should a user navigate to Site Editor > All Templates with an FSE theme active they should see the WooCommerce provided Product Archive block template.
- [x] Should a user load the Product Archive template, the editor will have the “WooCommerce Product Grid Block” Legacy Template block present.
- [x] Should a user load the Shop page on the frontend, it will render a list of products.
- [x] Should a user customise the Product Archive template and save it. The user should see a "Clear customizations" option on the All Templates view (clearing customisations functionality should be covered by Gutenberg)
- [x] Should a user load the customised Product Archive template the Site Editor should render these customisations
- [x] Should a user have a customised Product Archive template, these customisations would be present on the frontend.

Closes #5657

---

Note that, as opposed to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5722, where I opted not to test the functionality of **both** the template and the block (I tested only the template-related functionality, see discussion there), here I also tested the frontend part for the block itself, due to the fact that the test was fairly simply and straightforward to write.

#### Automated Tests
* [x] Changes in this PR are covered by Automated Tests.
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

How to test the changes in this Pull Request:

1. Remember to `npm i` when you pull, as there is a requirement for the newest `e2e-test-utils`.
2. Run `npm run test:e2e -- -i "tests/e2e/specs/backend/site-editing-templates.test.js"` and `npm run test:e2e -- -i "tests/e2e/specs/frontend/legacy-template-blocks.test.js"`.
3. Assess that the included test cases pass.

### User Facing Testing
These are steps for user testing (where "user" is someone interacting with this change that is not editing any code).
* [x] Same as above

PR #5920, opened 2022-02-22
Add mock theme with Woo Templates for E2E tests

### LINKED ISSUE #5660: Store Editing Templates: Explore if we can cover scenarios of themes overriding block templates with E2E tests
### Description

The FSE feature for WooCommerce exhibits specific behaviours based on what the theme has available. For example:

* If the theme has a `single-product.html` template it will override the one provided by Woo Blocks
* If the theme has an `archive-product.html` template it will override the Product Archive, Product Category and Product Tag templates provided by Woo Blocks.
* If a theme has a `taxonomy-product_cat.html` or `taxonomy-product_tag.html` template file it will override the one(s) provided by Woo Blocks.

We need to explore if in our testing environment we can cover these scenarios, as we will need to have the ability to add/remove certain template files from the theme in order to do this (or have different themes available to activate which covers each scenario)

sunyatasattva:
This PR adds a block-enabled theme with templates for WooCommerce which is intended to be used when testing the more complex scenarios described in #5660. In order to test our logic, we need two templates:

1. One block-enabled but Woo-unaware template to test default behaviours. (Such template was added in #5722)
2. Another block-enabled theme with Woo templates to test theme overrides.

Both themes are required. Tests for scenario 1 where added in #5722, #5748, #5820 and #5888. This PR does not add any test, but merely the basis for further E2E test scenarios.

Additionally, since theme-dependent logic seems to be becoming more and more useful (see #5913), this PR also adds a utility function used to explicitly declare theme dependencies of a test. Said function will activate the theme before running a given suite and then reactivate the previous theme, as to make sure to not have any side-effects (such as the ones described in #5913 above).

Closes #5660 

---

@dinhtungdu I refactored my previous code to not activate TT1 once done, but to reactivate the previously active theme ([as discussed here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5914#pullrequestreview-889277707)).

If you come across tests which are dependent on a specific theme activation, I suggest you refactor / create those using the code introduced in this PR. It's a simple test "hook", just using it at the top level of your `describe` closure should do the trick of activating the required theme and reactivating the previous one on teardown, like so:

```js
describe( 'My test suite', () => {
  useTheme( 'storefront' );

  it( 'tests something', async () => { /* ... */ } );
} );
```


tjcafferkey (approved):
Nice work @sunyatasattva, I really like the `useTheme` hook and how you've considered it enough to only activate the specified theme _just_ for the suite of tests it's being used in to prevent the chances of its usage breaking any subsequent test suites.

All the checks seem to be passing, and the code LGTM so I'm going to approve this.

PR #5923, opened 2022-02-23
Shopper → Mini Cart → Can view filled mini cart contents block

### LINKED ISSUE #5830: Critical flows: Shopper → Mini Cart → Can view filled mini cart contents block
### Note

Create e2e tests to verify that a shopper can see the filled Mini cart contents.

### File

`tests/e2e/specs/shopper/mini-cart-filled.test.js`

Part of #5814

dinhtungdu:
Fixes #5830

### Notes

This PR adds e2e tests to verify that a shopper can see the filled Mini cart contents.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/mini-cart.test.js -t "Filled mini cart"` to run only this test case.


tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -7,12 +12,33 @@ const block = {
...
+
 if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 3 ) {
 	// eslint-disable-next-line jest/no-focused-tests
 	test.only( `skipping ${ block.name } tests`, () => {} );
 }
 
 describe( 'Shopper → Mini Cart', () => {
+	beforeAll( async () => {
+		setDefaultOptions( { ...options, timeout: 5000 } );
+	} );
```
tomasztunik: let's add a comment to highlight the details about how this is required because of the script being lazy loaded on hover action over the Mini Cart icon

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -7,12 +12,33 @@ const block = {
...
+
+const clickMiniCartButton = async () => {
+	await page.hover( '.wc-block-mini-cart__button' );
+
+	await expect( page ).not.toMatchElement(
+		'.wc-block-mini-cart__drawer.is-loading'
+	);
+
+	await page.click( '.wc-block-mini-cart__button' );
+};
```
tomasztunik: Can we rename this to represent outcome more than user direct user action ie. `openMiniCartDrawer` so that it's easier to understand the state in which the test is?

Also thinking if we should have `expect` in here or if it should be part of a separate test ie. `Mini Cart drawer is initially closed` and then we could use `page.waitForXPath` option `hidden: true` to wait for loading class to go away and have this as a pure util.

then we could return `page.waitForXPath` from this action helper and use its return value in expect to ensure drawer is open. 

This way we wouldn't have to test for drawer being closed and open repeatedly and just ensure it in initial tests focused on these two states. 

What do you think?

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -94,4 +123,89 @@ describe( 'Shopper → Mini Cart', () => {
...
+			} );
+		} );
+
+		it( 'The Mini Cart products table show added products', async () => {
+			const products = await page.$$(
+				'.wc-block-all-products .wc-block-grid__product'
+			);
+			if ( products.length === 0 ) {
+				return;
+			}
```
tomasztunik: wouldn't this mask potential issue in test setup where there would be no products even though there should be?

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -94,4 +123,89 @@ describe( 'Shopper → Mini Cart', () => {
...
+			// Get a random product to ensure the test isn't false positive.
+			const product =
+				products[ Math.floor( Math.random() * products.length ) ];
+			const productTitle = await page.evaluate( ( el ) => {
+				return el.querySelector( '.wc-block-components-product-name' )
+					.textContent;
+			}, product );
+			await page.evaluate( ( el ) => {
+				el.querySelector( '.add_to_cart_button' ).click();
+			}, product );
```
tomasztunik: is there a reason for why we sometimes use it like so and other times just via `await page.click( '.add_to_cart_button' );` ?

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -94,4 +123,89 @@ describe( 'Shopper → Mini Cart', () => {
...
+				return;
+			}
+
+			// Get a random product to ensure the test isn't false positive.
+			const product =
+				products[ Math.floor( Math.random() * products.length ) ];
+			const productTitle = await page.evaluate( ( el ) => {
+				return el.querySelector( '.wc-block-components-product-name' )
+					.textContent;
+			}, product );
```
tomasztunik: Could we use ElementHandle getProperty here -> `product.getProperty( 'textContent' )` for clarity?

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -94,4 +123,89 @@ describe( 'Shopper → Mini Cart', () => {
...
+			const products = await page.$$(
+				'.wc-block-all-products .wc-block-grid__product'
+			);
+			if ( products.length === 0 ) {
+				return;
+			}
+
+			// Get a random product to ensure the test isn't false positive.
+			const product =
+				products[ Math.floor( Math.random() * products.length ) ];
```
tomasztunik: Does it have to be non deterministic here? Had random factor in tests bite me in the past :) How could this be a false postiive? Would it be because of cart not being properly cleared up in `beforeAll`?

tests/utils/shopper.js
```diff
@@ -56,4 +60,38 @@ export const shopper = {
 
 		await expect( page ).toMatchElement( 'h1', { text: title } );
 	},
+
+	/**
+	 * Override the @woocommerce/e2e-utils `emptyCart` method to fix the
+	 * ReferenceError issue and remove the cart items.
+	 *
+	 * @todo Remove shopper.emptyCart overload once the upstream  is fixed
```
tomasztunik, line 68: is this something we should contribute to WooCommerce or just local to us?

tomasztunik (commented):
Looks good, like this approach for setting the timeouts more than what I proposed in #5960. 

Dropping couple of questions for consideration!

tests/e2e/specs/shopper/mini-cart.test.js
tomasztunik: got it, we are clicking on the product related addToCart 😅

tests/e2e/specs/shopper/mini-cart.test.js
tomasztunik: Ah same here... :) 

Would pattern like this make it clearer? 
```
const productTitleEl = await product.$( '.wc-block-components-product-name' );
const productTitle = await productTitleEl.getProperty( 'textContent' ); 
```

I'm pondering usage of evaluate vs more declarative use of the page/frame/elementHandle APIs from perspective of readability and representing user action more than direct on page manipulation with JS.

tomasztunik: or in case of clicking the cart 

```
const productAddToCart = await product.$( '.add_to_cart_button' );
await productAddToCart.click();
```

tests/utils/shopper.js
dinhtungdu: Yes. I submitted a PR and it was merged last week: https://github.com/woocommerce/woocommerce/pull/31977.

tests/e2e/specs/shopper/mini-cart.test.js
dinhtungdu: > Does it have to be non deterministic here?

It isn't required but I want to replicate closer the real user behavior.

> How could this be a false positive?

`false positive` here may not exactly. But initially, I hard-coded the product title for the test. Lately, I get the product title dynamically, it can't be false positive anymore, I will update the comment to prevent confusion.

tests/utils/shopper.js
tomasztunik: Nice! Should we add a @todo to remove it when next WooCommerce e2e utils get released?

tests/e2e/specs/shopper/mini-cart.test.js
dinhtungdu: > Can we rename this to represent outcome more than user direct user action ie. `openMiniCartDrawer` so that it's easier to understand the state in which the test is?

The drawer can be opened by clicking the mini cart button or adding a product to the cart, so I don't think `openMiniCartDrawer` is fully precise here. What do you think?

> Also thinking if we should have `expect` in here or if it should be part of a separate test ie. `Mini Cart drawer is initially closed` and then we could use `page.waitForXPath` option `hidden: true` to wait for loading class to go away and have this as a pure util.

I removed `expect` from `clickMiniCartButton`.

> then we could return `page.waitForXPath` from this action helper and use its return value in expect to ensure drawer is open.
> 
> This way we wouldn't have to test for drawer being closed and open repeatedly and just ensure it in initial tests focused on these two states.

I do not fully understand your idea, could you please explain it in more detail?

tests/e2e/specs/shopper/mini-cart.test.js
dinhtungdu: Thrown an error instead. Thanks for the head-up!

tests/utils/shopper.js
dinhtungdu: I added the todo already. Look like I should change `the upstream  is fixed` to `the fix is released`? 😄

tomasztunik: 🤦 me being me... sorry for the oversight! :)

tests/e2e/specs/shopper/mini-cart.test.js
dinhtungdu: Applied your suggestion! Thanks!

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -17,8 +17,8 @@ const options = getDefaultOptions();
 const clickMiniCartButton = async () => {
 	await page.hover( '.wc-block-mini-cart__button' );
 
-	await expect( page ).not.toMatchElement(
-		'.wc-block-mini-cart__drawer.is-loading'
+	await page.waitForFunction(
+		'document.querySelector(".wc-block-mini-cart__drawer.is-loading") === null'
 	);
```
tomasztunik: we can use 
```
page.waitForSelector( '.wc-block-mini-cart__drawer.is-loading', { hidden: true } )
```
`hidden: true` inverts selector from waiting for presence to waiting for removal/non-presence - what do you think?

tests/e2e/specs/shopper/mini-cart.test.js
tomasztunik: I guess it makes sense to keep the name considering alternative paths - or make it more explicit about source of how the drawer was opened.

Regarding the idea I always liked for simplicity sake - to have tests build up and never repeat expects if given case is covered/ensured by previous test as this duplicates the concern over single area of code. Having this in mind I thought  clickMiniCartButton could return an elementHandle that we could test for presence of a drawer. With keepeing the name this doesn't make sense any more I guess!

Though I'm still thinking if we couldn't simplify the tests by not having to repeat test for or (not) variant if they don't represent alternative paths for opening these.

```
			await expect( page ).toMatchElement(
				'.wc-block-mini-cart__drawer',
				{
					text: 'Start shopping',
				}
			);
```

With this in mind I thought `openMiniCartDrawer` could also wait for/try to ensure drawer is open so that any further interaction can be made without having to wait for presence of the drawer with MatchElement or other selectors.

tests/e2e/specs/shopper/mini-cart.test.js
dinhtungdu: I'm a bit worried about `hidden` at first, as the docs says

> hidden <[boolean](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type)> wait for element to not be found in the DOM or to be hidden, i.e. have display: none or visibility: hidden CSS properties. Defaults to false.

Because the drawer is hidden by default, I thought I can return immediately, which leads to clicking the Mini Cart button while it's still loading (which causes the flaky tests we're trying to solve).

But testing your suggestion locally several times, looks like it works as expected. Committing yours because I also prefer the declarative approach. Thanks!

tomasztunik: Cheers!

tomasztunik (approved):
👍

PR #5925, opened 2022-02-23
Add margin support on the Product Image block

### LINKED ISSUE #5671: Product Image: Improve Global Styles API integration
The `Product Image` block, with the Global Styles API, should support:

- margin

gigitux:
This PR adds support for margin customization on the Global Style.

Fixes #5671 

### Manual Testing

This PR adds support for global style for the `Product Image` block.

- margin

#5671 

### Screenshots

image
*without global style*

image
*with global style*

### Manual Testing

Check out this branch:

1. Use `WordPress 5.9`.
3. Add the `All Products` block  (this block contains `Product Image`) to a post.
4. Get the focus on the `Product Image` block.
5. On the right sidebar, personalize the margin style.
9. Verify that the `Product Image Block` is shown under the `Blocks` section. Personalize again the margin style.
10. Save your changes.
13. Change again the styles.
14. Save your changes.


assets/css/style.scss
```diff
@@ -167,7 +167,6 @@
 
 // Element spacing.
 .wc-block-grid__product {
-	.wc-block-grid__product-image,
```
Aljullu, line 170: This seems to affect other blocks. This is Top Rated Products, for example:

`trunk` | This branch
--- | ---
imatge | imatge

assets/js/atomic/blocks/product-elements/image/style.scss
```diff
@@ -57,3 +55,13 @@
 .is-loading .wc-block-components-product-image {
 	@include placeholder();
 }
+
+.wc-block-components-product-image {
+	margin: 0 0 $gap-small;
+}
+
+
+// Don't use global style while skeleton is visible.
```
Aljullu: I was wondering if we should honor global styles also while the block is loading. Otherwise there is a visual jump:

https://user-images.githubusercontent.com/3616980/155564392-32ca4096-50c0-4768-875f-df21616a00b6.mp4

Aljullu (commented):
Left a couple of small comments, but overall looking good. Good job adding even more styles to the blocks! :clap:

assets/js/atomic/blocks/product-elements/image/style.scss
gigitux: If I comment this code when I edit the margin with the Global Style, we have this UI (I think that it is worse):

image
image

GIF
Screen Capture on 2022-02-24 at 18-01-53

assets/css/style.scss
gigitux: Just fixed in [this way](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5925/files#diff-cd886530abfa495dc6782189ca779f1304bb73bc1758e5d3f0968448dfb67abeR171). What do you think?

assets/js/atomic/blocks/product-elements/image/style.scss
Aljullu: Oh, interesting. I think the issue here comes from the `placeholder()` styles, which have `width: 100%`:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f4a11f158c5efd45f2caeeb9e67a0bbc61de0f94/assets/js/atomic/blocks/product-elements/image/style.scss#L58

Could we just add a `width: auto;` declaration after it? Besides that, you would also need to pass `spacingProps.style` to the [placeholder div](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/f4a11f158c5efd45f2caeeb9e67a0bbc61de0f94/assets/js/atomic/blocks/product-elements/image/block.js#L56-L65). It seemed to work fine for me:

https://user-images.githubusercontent.com/3616980/155696490-686165e2-8754-4522-92c0-1a18a5dfca4f.mp4

You could also pass `borderProps.style` to get the border radius to be applied as well.

gigitux: Oh! Thanks for the hint! 

I update the mixin too because the borderProps works only for block-level style.
With the global style, the margin value of the mixin has more priority than the global style value.
I added an optional parameter that allows to not include the border-radius value.


image
*before updating mixin function*

assets/js/atomic/blocks/product-elements/image/style.scss
```diff
@@ -55,5 +53,10 @@
 }
 
 .is-loading .wc-block-components-product-image {
-	@include placeholder();
+	@include placeholder($include-border-radius: true);
```
Aljullu: Should this be `false`?
```suggestion
	@include placeholder($include-border-radius: false);
```

assets/js/atomic/blocks/product-elements/image/style.scss
Aljullu: > With the global style, the margin value of the mixin has more priority than the global style value.

Good catch!

Aljullu (approved):
Left one last comment, but besides that it LGTM. :ship:

Aljullu:
@gigitux was this PR included in 7.1.0? Otherwise, should we change the milestone to 7.2.0?

gigitux:
I updated the milestone! Thanks for catching this :D

PR #5938, opened 2022-02-24
Add constants for C & C pages in our E2E tests

ralucaStan:
Inspired by [WC E2E way](https://github.com/woocommerce/woocommerce-monorepo/blob/27009cdc0a2af1cca88b88b5ee3600416bab6b7a/plugins/woocommerce/tests/e2e/utils/src/flows/shopper.js#L53-L57) of navigating to the C & C pages, we are also saving the path to the C & C block pages and accessing that directly.

This will avoid going to admin to find out the links for these pages, which improves the speed of the test.
BEFORE:
Screenshot 2022-02-24 at 14 42 33
AFTER:
Screenshot 2022-02-24 at 14 44 40


alexflorisca (approved):
Great idea, simplifies and speeds up loading of pages significantly! I'm wondering if we can do the same not just for cart and checkout, but for all the other block pages as well? This would simplify the [getBlockPagePermalink](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/utils/get-block-page-permalink.js#L20) function as well. What do you think?

opr (approved):
Looks good, will make things quicker too!

ralucaStan:
I've decided to undo navigating to the C & C block pages by navigating directly to their hardcoded URLS, even if it is a bit faster, see image below.
Screenshot 2022-02-24 at 14 44 40

But the `goToBlockPage` function was introduced by @dinhtungdu in [this PR ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5877/files#diff-ca4a8416e3651adb2e4de52a40d6ce7bdae89544db7f77fd09d9455e26c7a159R51) and I think it's better to be consistent.

@alexflorisca / @opr ready for a re-review

tests/utils/shopper.js
```diff
@@ -8,19 +8,24 @@ import { shopper as wcShopper } from '@woocommerce/e2e-utils';
...
-			`Checkout Block`
-		);
+	goToBlockPage,
 
-		await page.goto( checkoutBlockPermalink, {
-			waitUntil: 'networkidle0',
-		} );
-		await expect( page ).toMatchElement( 'h1', { text: 'Checkout' } );
-	},
+	goToCheckoutBlock: goToBlockPage( 'Checkout Block' ),
```
ralucaStan: I will switch this to `block.name` in a separate PR coming soon, where I will collect all the data about blocks in a file so that we can reuse across tests.

dinhtungdu:
> But the `goToBlockPage` function was introduced by @dinhtungdu in [this PR ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5877/files#diff-ca4a8416e3651adb2e4de52a40d6ce7bdae89544db7f77fd09d9455e26c7a159R51) and I think it's better to be consistent.

@ralucaStan please note that I refactored the `goToBlockPage` to make it more reliable and work with any themes in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5914.

ralucaStan:
> > But the `goToBlockPage` function was introduced by @dinhtungdu in [this PR ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5877/files#diff-ca4a8416e3651adb2e4de52a40d6ce7bdae89544db7f77fd09d9455e26c7a159R51) and I think it's better to be consistent.
> 
> @ralucaStan please note that I refactored the `goToBlockPage` to make it more reliable and work with any themes in #5914.

thank you @dinhtungdu, I didn't actually notice the difference in the implementation that no longer clicked on the main navigation links. That explains why there wasn't a bigger time improvement anymore

ralucaStan:
I've reverted to the initial changes that used constants, that weren't based on `goToBlockPage`

alexflorisca (commented):
~~Hey @ralucaStan thanks for the changes. Sorry if I misunderstood, but I thought we were going with the approach mentioned [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5877#discussion_r810096719) - of clicking on the nav link item. It seems the `goToBlockPage` function here in unchanged here?~~ I missed the messages above, my bad, lots of related work happening in different PRs! Looks good :)

ralucaStan:
> I missed the messages above, my bad, lots of related work happening in different PRs! Looks good :)
But the goToBlockPage function was introduced by @dinhtungdu inand I think it's better to be consistent.

Thank you @alexflorisca , let's stick with the initial approach of the PR as we need some time improvements. We can reiterate our approach.

To reiterate, initially the `goToBlockPage` was clicking on navigation links  [see this PR ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5877/files#diff-ca4a8416e3651adb2e4de52a40d6ce7bdae89544db7f77fd09d9455e26c7a159R51). But this didn't work with other themes, it worked on Storefront, so [this PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5914) changed the implementation to use `getBlockPagePermalink`, which is really slow.

dinhtungdu:
>  But this didn't work with other themes, it worked on Storefront, so https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5914 changed the implementation to use getBlockPagePermalink, which is really slow.

Yeah, it's really really slow. I'll try a different approach to speed it up.

ralucaStan:
Take a look at this PR changes where we hardcode the URLs. It is something inspired by how WC does navigation between pages. I have an idea about how this could be applied to other blocks, I'll put out another PR today to showcase it and move the conversation there

PR #5939, opened 2022-02-24
Full Store Editing: Technical documentation

### LINKED ISSUE #5661: Store Editing Templates: Technical Documentation
## Description

The WooCommerce Store Editing Templates project has been built to provide block templates via the WooCommerce (or WooCommerce Blocks) plugin. Currently WordPress FSE does not account for plugins providing block templates, only themes. In order to achieve this we had to build this solution from scratch based on the functionality that WordPress and Gutenberg provided. As a result of this we have an undocumented solution.

Some things we need to document:

* How does the WooCommerce Store Editing Templates logic work.
    * How the Legacy Template block works together with [`LegacyTemplate.php`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/LegacyTemplate.php)
    * Compatibility with WP and WC
    * An overview of the logic involved in [`BlockTemplateController.php`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTemplatesController.php) and what its responsibility is.
    * This should include context around any significant technical decision we had to make along with how it interacts with themes that provide WooCommerce block template files.
* How can other WooCommerce plugins utilise the functionality we have created to add new templates of their own?
* Any limitations of the WooCommerce Store Editing Templates feature.

These will likely live in the [docs directory](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/docs) in the most appropriate place.

For some inspiration, feel free to look in the repos docs directory to see how documentation is structured and presented but here are some examples:
* [Stock Reservation during Checkout](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/blocks/stock-reservation.md)
* [Extensibility in WooCommerce Blocks](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/README.md)

**If it makes sense to split this issue into multiple issues for different parts of the documentation. Please do so.**

tjcafferkey:
This is some technical documentation and related imagery to detail the purpose and usage of the BlockTemplateController and the WooCommerce block template feature overall.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5661


assets/js/blocks/legacy-template/README.md
```diff
@@ -0,0 +1,18 @@
+# Legacy Template Block
```
sunyatasattva: [Since we are renaming those to Classic Template Blocks](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5861), even though that issue is not closed yet, I'd suggest we already call them like that in the docs, since these docs are not going to be out before the rename anyways.

assets/js/blocks/legacy-template/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+The Legacy Template block is a placeholder block for specific WooCommerce block templates which are rendered on the server-side when a block theme is active.
+
+Given a specific template attribute prop, the block will render a specific template both on the front-end and in the Site Editor in the form of a placeholder.
```
sunyatasattva: I feel this sentence is a bit ambiguous, but always keep in mind that English is not my native language. I find it ambiguous on two fronts:

> Given a specific template attribute prop, the block will render a specific template

This is a bit unclear more than ambiguous. What about something like:

_By assigning a template identifier to the attribute prop, the block will render that specific template_

> both on the front-end and in the Site Editor in the form of a placeholder.

Here I feel that it sounds like the placeholder will be rendered both on the front-end and in the Site Editor. Perhaps it's because of “both”? Maybe a comma between “front-end” and “and” could help? Or perhaps, the entire sentence could be something like this:

_By assigning a template identifier to the attribute prop, the block will render that specific template on the front-end, and a placeholder for said template in the Site Editor_

assets/js/blocks/legacy-template/README.md
```diff
@@ -0,0 +1,18 @@
...
+
```
sunyatasattva, line 8: I'm wondering if here we should add a paragraph explaining that the placeholder does not completely represent what is rendered on the front-end.

[Considered that a suggestion](pdnLyh-tW-p2) by @nerrad  was to apply some styles to the template placeholder eventually, this might be useful so that people are aware that this is not going to happen yet and don't get surprised when it doesn't.

What do you think?

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+  * [Requirements](#requirements) 
+* [Technical Overview](#technical-overview)
+  * [The Problem](#the-problem)
+  * [The Solution](#the-solution)
+* [Some things to be aware of](#some-things-to-be-aware-of)
+* [Related Files](#related-files)
+
+## Overview
+
+At the highest level, WooCommerce Block Templates are a collection of WooCommere core templates that are available in the WordPress Full Site Editing feature introduced as part of WordPress 5.9. Enabling you to customize these templates with blocks via the Site Editor.
```
sunyatasattva:
```suggestion
At the highest level, WooCommerce Block Templates are a collection of WooCommerce core templates that are made available in the WordPress Full Site Editing experience introduced as part of WordPress 5.9, enabling you to customize these templates with blocks via the Site Editor. 
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+You can read more about the Full Site Editing (FSE) feature [here](https://developer.wordpress.org/block-editor/getting-started/full-site-editing/).
+
+### Requirements
+
+| Software        | Minimum Version  |
+|-----------------|------------------|
+| WordPress       | 5.9              |
+| WooCommerce     | 6.0              |
+| Block theme     | -                |
```
sunyatasattva: Perhaps _“Any block theme”_? Is that the official name, by the way? Or perhaps, “block-enabled theme” or “FSE-enabled theme”?

Am not sure what the official name is. But adding “any” would help by not confusing people that “Block theme” is a specific thing.

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Technical Overview
+
+### The Problem
+
+The FSE feature does not accommodate for loading templates from plugins such as WooCommerce (or WooCommerce Blocks). Instead all template files loaded natively with this feature are to be present within the active themes files. This is a problem because we would like WooCommerce specific templates to be compatible with every theme, and to do this we have to provide themes that don't have WooCommerce specific templates with them from our plugin.
+
+### The Solution
+
+We created a custom solution which involves a handful of files (listed in the below table) that are responsible for both making the templates available as a block template, and rendering it on the front-end.
```
sunyatasattva, line 37:
```suggestion
We created a custom solution which involves a handful of files (listed in the table below) that are responsible for both making the templates available as block templates, and rendering them on the front-end.
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+The BlockTemplateController class is primarily responsible for hooking into both WordPress core, and WooCommerce core filters to load WooCommerce [templates](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/templates/templates) in the Site Editor and on the front-end. It is in this class where the majority of the logic is handled.
```
sunyatasattva, line 39:
```suggestion
The `BlockTemplateController` class is primarily responsible for hooking into both WordPress core, and WooCommerce core filters to load [WooCommerce templates](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/templates/templates) in the Site Editor and on the front-end. It is in this class where the majority of the business logic is handled.
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Some things to be aware of
+
+* Individual templates are represented by a placeholder block within the Site Editor until we can 'block-ify' these. This is the long-term goal. Until then users will be able to customize templates by adding blocks above and below the placeholder.
```
sunyatasattva, line 43:
```suggestion
* Individual templates are represented in their entirety by a placeholder block within the Site Editor until we can cut them up into discrete blocks. This is our long-term goal. However, until then, users will be able to customize templates by adding blocks above, below and around the placeholder.
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+| templates/templates/\*      | Location in the filesystem where WooCommerce block template HTML files are stored.                                                                                              | [Source files](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/templates/templates)              |                                                           |
+| legacy-template/\*          | The JavaScript block rendered in the Site Editor. This is a server-side rendered component which is handled by LegacyTemplate.php                                               | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/assets/js/blocks/legacy-template)  | [README](../../assets/js/blocks/legacy-template/README.md)|
```
sunyatasattva:
```suggestion
| legacy-template/\*          | The JavaScript block rendered in the Site Editor. This is a server-side rendered component which is handled by `LegacyTemplate.php`                                             | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/assets/js/blocks/legacy-template)  | [README](../../assets/js/blocks/legacy-template/README.md)|
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+| BlockTemplateController.php | Class which contains all the business logic which loads the templates into the Site Editor or on the front-end through various hooks available in WordPress & WooCommerce core. | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTemplatesController.php)  | [README](./block-template-controller.md)                  |
+| BlockTemplateUtils.php      | Class containing a collection of useful utility methods.                                                                                                                        | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/Utils/BlockTemplateUtils.php)  |                                                           |
+
```
sunyatasattva, line 57: Should all the elements within the first column be surrounded by backticks perhaps?

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,77 @@
+# BlockTemplateController
+[Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTemplatesController.php)
+
+The BlockTemplateController class contains all the business logic which loads the templates into the Site Editor or on the front-end through various hooks available in WordPress & WooCommerce core. Without documenting every method individually, I will look to provide some insight into key functionality.
```
sunyatasattva, line 4:
```suggestion
The `BlockTemplateController` class contains all the business logic which loads the templates into the Site Editor or on the front-end through various hooks available in WordPress and WooCommerce core. Without documenting every method individually, I will look to provide some insight into (its?) key functionality. 
```

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,77 @@
+# BlockTemplateController
+[Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTemplatesController.php)
+
+The BlockTemplateController class contains all the business logic which loads the templates into the Site Editor or on the front-end through various hooks available in WordPress & WooCommerce core. Without documenting every method individually, I will look to provide some insight into key functionality. 
+
+## Table of Contents
+
+* [Overview](#overview)
+* [add_block_templates( $query_result, $query, $template_type )](#add_block_templates-query_result-query-template_type-)
+* [get_block_file_template( $template, $id, $template_type )](#get_block_file_template-template-id-template_type-)
```
sunyatasattva, line 10: Just double checking, is the ending `-` correct as an anchor?

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,77 @@
...
+* Loading one of the templates on the frontend where the query would build a list of relevant templates based on a hierarchy (for example, the product page hierarchy could be an array containing  `single-product-[product-name].html`, `single-product.html`, `single.html`).
+
+**This method is responsible for:**
+
+* Giving our templates a user-friendly title (e.g. turning "single-product" into "Product Page").
+* It collects all the WooCommerce templates from both the filesystem and the database (customized templates are stored in the database as posts) and adds them to the returned list.
+* In the event the theme has a `archive-product.html` template file, but not category/tag template files, it is eligible to use the `archive-product.html` file in their place. So we trick Gutenberg in thinking some templates (e.g. category/tag) have a theme file available if it is using the `archive-product.html` template, even though _technically_ the theme does not have a specific file for them.
+* If the query is targeting templates for a specific post type (e.g. on the Edit Post view), we do not add ineligible WooCommerce templates to the returned list.
```
sunyatasattva: I'm reading this docs without looking at the source code for added realism 😁 , and I am not sure I understand what this means. Is there any way we can make this clearer?

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Return value
+
+This method will return an array of `WP_Block_Template` values
+
+## get_block_file_template( $template, $id, $template_type )
+
+This method is applied to the filter `pre_get_block_file_template` inside the WordPress core function `get_block_file_template` (not to be confused with this method from the BlockTemplateController, of which has the same name).
```
sunyatasattva:
```suggestion
This method is applied to the filter `pre_get_block_file_template` inside the WordPress core function `get_block_file_template` (not to be confused with this method from the `BlockTemplateController`, of which has the same name). 
```

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,77 @@
...
+This method will return an array of `WP_Block_Template` values
+
+## get_block_file_template( $template, $id, $template_type )
+
+This method is applied to the filter `pre_get_block_file_template` inside the WordPress core function `get_block_file_template` (not to be confused with this method from the BlockTemplateController, of which has the same name). 
+
+The order of execution is as follows:
+
+1. `get_block_template()` from WordPress core will execute, and attempt to retrieve a customized version of the template from the database.
+2. If it fails to retrieve one, it will execute the `get_block_file_template()` from WordPress core which will apply the filter `pre_get_block_file_template` which is what we hook into to return our template file, which will trigger an early return to prevent WordPress from continuing its query.
```
sunyatasattva: Any way we can contain the “whiches”? So many relative clauses are a bit hard to parse, especially as a non native speaker, I suppose.

One way I can think of would be to just stop the sentence and resume another one, something like:

_If it fails to retrieve one, it will execute the `get_block_file_template()` from WordPress core which will apply the filter `pre_get_block_file_template`. This is were we hook into to return our template file, and trigger an early return to prevent WordPress from continuing its query_

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,77 @@
...
+This method is applied to the filter `template_redirect` and executed before WordPress determines which template to load.
+
+This allows us to hook into WooCommerce core through the filter `woocommerce_has_block_template` where we can determine if a specific block template exists and should be loaded.
+
+**Typically executed when:**
+* A user loads a page on the front-end.
+
+**This method is responsible for:**
+* Determining if the current page has an appropriate WooCommerce block template available to render.
+* Checking if the currently loaded page is from WooCommerce. It then checks if the theme has an appropriate template to use, if it does not then it finally checks if WooCommerce has a default block template available. If so, we override the value through `woocommerce_has_block_template` to resolve true.
```
sunyatasattva:
```suggestion
* Checking if the currently loaded page is from WooCommerce. It then checks if the theme has an appropriate template to use: if it does not, then it finally checks if WooCommerce has a default block template available. If so, we override the value through `woocommerce_has_block_template` to resolve `true`.
```

docs/templates/legacy-template.md
```diff
@@ -0,0 +1,13 @@
+# LegacyTemplate
+
+[Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/LegacyTemplate.php)
+
+The LegacyTemplate is a class used to set up the Legacy Template block on the server-side, and render the correct template.
```
sunyatasattva:
```suggestion
The `LegacyTemplate` is a class used to set up the Legacy Template block on the server-side, and render the correct template.
```

Here again we might want to use the Classic Template name. This is a good point to talk about whether renaming the file and the class itself is safe as discussed [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5861#issuecomment-1049627578)

docs/templates/legacy-template.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Overview
+
+It is from this file we enqueue the frontend scripts necessary for the product gallery, add to basket etc. functionality to work on the frontend.
```
sunyatasattva:
```suggestion
It is from this file we enqueue the front-end scripts necessary for the product gallery, add to basket etc. functionality to work on the front-end.
```

So far we used the hyphenated form of "front-end", so I am suggesting that. However, this sentence sounds somewhat a bit strange. I'm not sure how I would reword it, maybe:

_It is from this file we enqueue the front-end scripts necessary to enable dynamic functionality, such as the product gallery, add to basket etc._

What do you think?

assets/js/blocks/legacy-template/README.md
tjcafferkey: Yeah, that is fine. I was unsure on whether to use the term Legacy or Classic when writing these docs so I kept it inline with what was already there (which happened to be Legacy). I will update all references to say Classic.

assets/js/blocks/legacy-template/README.md
tjcafferkey: >I'm wondering if here we should add a paragraph explaining that the placeholder does not completely represent what is rendered on the front-end.

It cannot hurt to add a paragraph explaining that so no problem, it's probably more useful to belong in the user/merchant facing docs but either way not a problem.

>this might be useful so that people are aware that this is not going to happen yet and don't get surprised when it doesn't.

Personally, I don't think this really belongs in these docs. Once a technical decision & change is made we should also update the documentation accordingly. I'd prefer to avoid adding "X or Y might be coming in the future" as it can lead to some confusion and muddy the waters for those who are looking for what is implemented right now.

We can however refer to other documentation, or a roadmap where this might be more applicable?

docs/templates/README.md
tjcafferkey: >But adding “any” would help by not confusing people 

Yep, I agree

>Am not sure what the official name is.

More info [here](https://developer.wordpress.org/block-editor/getting-started/full-site-editing/#block-themes)

assets/js/blocks/legacy-template/README.md
sunyatasattva: Yea, I just mean we should warn people that styles and other customizations won't be applied to the placeholder. I'm not suggesting we necessarily write in the docs that it's our intention to make that happen eventually.

(Though, to be fair, we have some reference to future plans [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5939/files#diff-162dd54970925f7e59d7d2b0fdda4785f6b4ed538b628459e4a89bf2f7767952R43), though I think it's different in this case since that's an over-arching goal and not a detail such as styles applied to the placeholder, so I'd be for leaving that reference to the long-term blockification into the docs, personally)

tjcafferkey: >Yea, I just mean we should warn people that styles and other customizations won't be applied to the placeholder.

I see what you mean. At the moment the block does not have any customizable options anyway, but I'll make a note of it anyway in the docs.

docs/templates/block-template-controller.md
tjcafferkey: Nice! Reads much better

docs/templates/legacy-template.md
tjcafferkey: Ive renamed this to `ClassicTemplate` and even updated the source file link to `ClassicTemplate` in anticipation that is what the file will also be called.

docs/templates/legacy-template.md
tjcafferkey: I agree!

docs/templates/block-template-controller.md
tjcafferkey: Yep, this is correct.

docs/templates/block-template-controller.md
tjcafferkey: Hmm yes, let me work on rewording this.

tjcafferkey: Hopefully this is an improvement https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5939/commits/c0bf7a0cd668d45986dc11310f6d1f8a3ce1a235

assets/js/blocks/legacy-template/README.md
```diff
@@ -0,0 +1,22 @@
+# Classic Template Block
+
+The Classic Template block is a placeholder block for specific WooCommerce block templates which are rendered on the server-side when a block theme is active.
+
+By assigning a template identifier to the attribute prop, the block will render that specific template on the front-end, and a placeholder for said template in the Site Editor.
+
+It's worth noting that in the Site Editor where a placeholder merely represents a front-end view, this representation will not always be completely accurate.
```
frontdevde:
```suggestion
It's worth noting that the placeholder in the Site Editor is merely an approximate representation of a template's front-end view.
```

docs/readme.md
```diff
@@ -10,6 +10,7 @@ The WooCommerce Blocks Handbook provides documentation for designers and develop
...
 | [Extensibility](extensibility/README.md)          | These documents cover extensibility of WooCommerce Blocks.                                                                                                                                                                                                                      |
 | [Theming](theming/README.md)                      | These documents cover theming for blocks, styles, CSS classnames and other theming best practices.                                                                                                                                                                              |
+| [Templates](templates/README.md)                  | These documents cover WooCommerce block template (and parts) functionality and technical overview and related information.                                                                                                                                                      |
```
frontdevde:
```suggestion
| [Templates](templates/README.md)                  | These documents provide a technical overview of WooCommerce block template (parts) functionality.                                                                                                                                                      |
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+  * [Requirements](#requirements) 
+* [Technical Overview](#technical-overview)
+  * [The Problem](#the-problem)
+  * [The Solution](#the-solution)
+* [Some things to be aware of](#some-things-to-be-aware-of)
+* [Related Files](#related-files)
+
+## Overview
+
+At the highest level, WooCommerce Block Templates are a collection of WooCommerce core templates that are made available in the WordPress Full Site Editing experience introduced as part of WordPress 5.9, enabling you to customize these templates with blocks via the Site Editor.
```
frontdevde:
```suggestion
WooCommerce Block Templates are a collection of WooCommerce Core templates for the WordPress Full Site Editing experience introduced in WordPress 5.9. You can customize these templates in the Site Editor. 
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+You can read more about the Full Site Editing (FSE) experience [here](https://developer.wordpress.org/block-editor/getting-started/full-site-editing/).
+
+### Requirements
+
+| Software        | Minimum Version  |
+|-----------------|------------------|
+| WordPress       | 5.9              |
+| WooCommerce     | 6.0              |
+| Any block theme | -                |
```
frontdevde:
```suggestion
| Theme | Any [block theme](https://developer.wordpress.org/block-editor/how-to-guides/themes/block-theme-overview/#what-is-a-block-theme)               |
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Technical Overview
+
+### The Problem
+
+The FSE feature does not accommodate for loading templates from plugins such as WooCommerce (or WooCommerce Blocks). Instead all template files loaded natively with this feature are to be present within the active themes files. This is a problem because we would like WooCommerce specific templates to be compatible with every theme, and to do this we have to provide themes that don't have WooCommerce specific templates with them from our plugin.
```
frontdevde:
```suggestion
Currently, the FSE feature does not accommodate loading block templates from plugins such as WooCommerce (or WooCommerce Blocks). Instead, all template files loaded natively must be present within the active theme's templates folder. When the active block theme doesn't come with WooCommerce specific templates, we want to be able to provide sensible defaults, though.```

File: docs/templates/README.md (hunk 6 above)
Comment (frontdevde) [reply docs/templates/README.md:37]:
```suggestion
To solve the aforementioned problem, we created a custom solution that involves the files listed below. They are responsible for making the templates available as block templates and rendering them on the front-end.
```

docs/templates/README.md
frontdevde: Agree that 'block-ify' is unclear! That said, maybe `divide them up` instead of `cut them up` here? Might just be me but `cut them up` does feel like it could have a bit of a negative connotation. 😅

Also, I'd leave it at above and below, as around isn't really that much of an option in the current state.

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+The BlockTemplateController class is primarily responsible for hooking into both WordPress core, and WooCommerce core filters to load WooCommerce [templates](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/templates/templates) in the Site Editor and on the front-end. It is in this class where the majority of the logic is handled.
+
+### Some things to be aware of
+
+* Individual templates are represented by a placeholder block within the Site Editor until we can 'block-ify' these. This is the long-term goal. Until then users will be able to customize templates by adding blocks above and below the placeholder.
+* At the beginning of the project, we unintentionally used the incorrect WooCommerce plugin slug. This has resulted in us maintaining both the incorrect and correct slugs. We reference these via `BlockTemplateUtils::DEPRECATED_PLUGIN_SLUG` and `BlockTemplateUtils::PLUGIN_SLUG`. More information on that [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5423).
```
frontdevde, line 44:
```suggestion
* At the beginning of the project, we unintentionally used an incorrect WooCommerce plugin slug. We thus maintain both the incorrect and correct slugs (referenced as `BlockTemplateUtils::DEPRECATED_PLUGIN_SLUG` and `BlockTemplateUtils::PLUGIN_SLUG`). More information on that [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5423).
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+* If a theme has a `archive-product.html` template file, but does not have any taxonomy related files. We will automatically duplicate their themes `archive-product.html` file in place of these taxonomy files within `BlockTemplateController->get_block_file_template()`.
```
frontdevde, line 45:
```suggestion
* If a theme has an `archive-product.html` template file, but does not have any other taxonomy-related files, we will automatically duplicate their theme's `archive-product.html` file in place of these taxonomy files within `BlockTemplateController->get_block_file_template()`.
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+| templates/templates/\*      | Location in the filesystem where WooCommerce block template HTML files are stored.                                                                                              | [Source files](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/templates/templates)              |                                                           |
+| classic-template/\*         | The JavaScript block rendered in the Site Editor. This is a server-side rendered component which is handled by ClassicTemplate.php                                              | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/assets/js/blocks/classic-template)  | [README](../../assets/js/blocks/classic-template/README.md)|
```
frontdevde, line 53:
```suggestion
| classic-template/\*         | The JavaScript block rendered in the Site Editor.                                              | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/assets/js/blocks/classic-template)  | [README](../../assets/js/blocks/classic-template/README.md)|
```

frontdevde: This feels redundant as we're addressing ClassicTemplate.php below.

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+| ClassicTemplate.php         | Class used to setup the block on the server-side and render the correct template                                                                                                | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/ClassicTemplate.php) | [README](./classic-template.md)                            |
```
frontdevde, line 54:
```suggestion
| ClassicTemplate.php         | Class used to setup the block on the server-side and render the correct template.                                                                                                | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/ClassicTemplate.php) | [README](./classic-template.md)                            |
```

docs/templates/README.md
```diff
@@ -0,0 +1,57 @@
...
+| BlockTemplateController.php | Class which contains all the business logic which loads the templates into the Site Editor or on the front-end through various hooks available in WordPress & WooCommerce core. | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTemplatesController.php)  | [README](./block-template-controller.md)                  |
```
frontdevde, line 55:
```suggestion
| BlockTemplateController.php | Class containing all the business logic loading the templates into the Site Editor or on the front-end through various WordPress & WooCommerce core hooks. | [Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTemplatesController.php)  | [README](./block-template-controller.md)                  |
```

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,78 @@
...
+
+**Typically executed when:**
+* Loading the "All Templates" view in the Site Editor
+* Loading one of the templates on the front-end where the query would build a list of relevant templates based on a hierarchy (for example, the product page hierarchy could be an array containing  `single-product-[product-name].html`, `single-product.html`, `single.html`).
+* Loading the "Edit Product" view.
+
+**This method is responsible for:**
+
+* Giving our templates a user-friendly title (e.g. turning "single-product" into "Product Page").
+* It collects all the WooCommerce templates from both the filesystem and the database (customized templates are stored in the database as posts) and adds them to the returned list.
```
frontdevde, line 35:
```suggestion
* Collecting all the WooCommerce templates from both the filesystem and the database (customized templates are stored in the database as posts) and adding them to the returned list.
```

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,78 @@
...
+* In the event the theme has a `archive-product.html` template file, but not category/tag template files, it is eligible to use the `archive-product.html` file in their place. So we trick Gutenberg in thinking some templates (e.g. category/tag) have a theme file available if it is using the `archive-product.html` template, even though _technically_ the theme does not have a specific file for them.
+* Ensuring we do not add irrelevant WooCommerce templates in the returned list. For example, if  `$query['post_type']` has a value (e.g. `product`) this means the query is requesting templates related to that specific post type, so we filter out any irrelevant ones. This _could_ be used to show/hide templates from the template dropdown on the "Edit Product" screen in WP Admin.
+
+### Return value
+
+This method will return an array of `WP_Block_Template` values
```
frontdevde, line 41:
```suggestion
This method will return an array of `WP_Block_Template` values.
```

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,78 @@
...
+
+## get_block_file_template( $template, $id, $template_type )
+
+This method is applied to the filter `pre_get_block_file_template` inside the WordPress core function `get_block_file_template` (not to be confused with this method from the `BlockTemplateController` class, of which has the same name).
```
frontdevde:
```suggestion
This method is applied to the filter `pre_get_block_file_template` inside the WordPress core function `get_block_file_template` (not to be confused with this method from the `BlockTemplateController` class, which has the same name). 
```

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,78 @@
...
+This method will return an array of `WP_Block_Template` values
+
+## get_block_file_template( $template, $id, $template_type )
+
+This method is applied to the filter `pre_get_block_file_template` inside the WordPress core function `get_block_file_template` (not to be confused with this method from the `BlockTemplateController` class, of which has the same name). 
+
+The order of execution is as follows:
+
+1. `get_block_template()` from WordPress core will execute, and attempt to retrieve a customized version of the template from the database.
+2. If it fails to retrieve one, it will execute the `get_block_file_template()` from WordPress core which will apply the filter `pre_get_block_file_template`. This is where we hook into to return our template file, and trigger an early return to prevent WordPress from continuing its query.
```
frontdevde:
```suggestion
2. If it fails to retrieve one, it will execute the `get_block_file_template()` function from WordPress core which will apply the filter `pre_get_block_file_template`. This is where we hook in to return our template file, and trigger an early return to prevent WordPress from continuing its query.
```

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,78 @@
...
+
+During step 2 it's important we hook into the `pre_get_block_file_template` because if we don't, the function will check if the first part of the template ID (e.g. `woocommerce/woocommerce`) is the same as the current themes ID (e.g. `twentytwentytwo`), which will resolve `false` and return `null` instead of the expected `WP_Block_Template` object.
```
frontdevde:
```suggestion
During step 2 it's important we hook into the `pre_get_block_file_template`. If we don't, the function will check if the first part of the template ID (e.g. `woocommerce/woocommerce`) is the same as the current themes ID (e.g. `twentytwentytwo`), which will resolve `false` and return `null` instead of the expected `WP_Block_Template` object.
```

docs/templates/block-template-controller.md
```diff
@@ -0,0 +1,78 @@
...
+
+**Typically executed when:**
+* A user clears the customizations of a WooCommerce template.
+
+**This method is responsible for:**
+* Loading the template files from the filesystem, and building a `WP_Block_Template` version of it.
+
+### Return value
+
+This method will return `WP_Block_Template` or `null`
```
frontdevde:
```suggestion
This method will return `WP_Block_Template` or `null`.
```

docs/templates/classic-template.md
```diff
@@ -0,0 +1,13 @@
+# ClassicTemplate
+
+[Source file](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/ClassicTemplate.php)
+
+The `ClassicTemplate` is a class used to set up the Classic Template block on the server-side, and render the correct template.
+
+## Overview
+
+It is from this file we enqueue the front-end scripts necessary to enable dynamic functionality, such as the product gallery, add to basket etc.
```
frontdevde:
```suggestion
From this file, we enqueue the front-end scripts necessary to enable dynamic functionality, such as the product gallery, add to basket, etc.
```

frontdevde (approved):
Great work on writing this well-written & easy to follow documentation, @tjcafferkey!
LGTM! 👍
Please consider the comments in this review as optional suggestions for minor subjective improvements.

sunyatasattva (approved):
Thanks for listening to the feedback. Great work, it must have been quite a feat to write it!

PR #5941, opened 2022-02-24
Refactor ExtendRestApi as ExtendSchema and expose `register_endpoint_data` as a public function

### LINKED ISSUE #3477: Evaluate if we should expose `register_endpoint_data` in a public function.
Currently, extensions can extend our StoreAPI by fetching our `ExtendRestApi` from our DI.

```php
<?php

$extend = Package::container()->get( ExtendRestApi::class );
$extend->register_endpoint_data( ... );
```

We should consider if this is the right approach.
If we want to make this available in a global function or some other solution.

mikejolley:
This PR makes some changes to the `ExtendRestApi` class to prepare it for core.

First, we needed to move `ExtendRestApi` from the services directory, directly under the StoreAPI directory and namespace. The reason for this is that if Store API codebase moves to WooCommerce core, `ExtendRestApi` would need to come too.

Second, when moving `ExtendRestApi` to StoreAPI, it became apparent this only extends Schema. So to better reflect this, I've renamed the class to `ExtendSchema` and moved it under the `StoreApi/Schema` directory.

Finally, to make this more easy to consume for 3rd parties, and avoid them having to use the Package container as well as know the namespace for the `ExtendSchema` class, I've added some global function helpers:

- `woocommerce_store_api_register_endpoint_data( $args )`
- `woocommerce_store_api_register_update_callback( $args )`
- `woocommerce_store_api_register_payment_requirements( $args )`
- `woocommerce_store_api_get_formatter( $name )`

These functions just act as wrappers for the `ExtendSchema` class methods.

Fixes #3477

To prevent a backwards compatibility break, I had to keep `ExtendRestApi` in our container (it just returns the `ExtendSchema` instance instead, and I setup an alias between the old and new namespaces:

```php
class_alias( Automattic\WooCommerce\Blocks\StoreApi\Schemas\ExtendSchema::class, Automattic\WooCommerce\Blocks\Domain\Services\ExtendRestApi::class );
```

### Testing

This change is invisible to users. Please smoke test, ensure tests pass.

1. Please smoke test checkout
2. Ensure unit tests pass
3. Test an extension (Subscriptions and Pre-orders) and confirm the integration is still function. There will be deprecation notices from the old namespace if they use `ExtendRestApi` 

### Changelog

> Renamed `ExtendRestApi` to `ExtendSchema` and added helper functions for accessing class methods.


mikejolley:
@senadir can I get your approval on this too? You originally authored the ExtendRestAPI class.

docs/contributors/documentation-guidelines.md
```diff
@@ -17,6 +17,7 @@
   - [Sort releases descending](#sort-releases-descending)
   - [Structure “How to” instructions](#structure-how-to-instructions)
   - [Link references](#link-references)
+  - [](#)
```
opr:
```suggestion
```

docs/extensibility/README.md
```diff
@@ -8,6 +8,7 @@ These documents are all dealing with extensibility in the various WooCommerce Bl
 - [REST API](#rest-api)
 - [Checkout Payment Methods](#checkout-payment-methods)
 - [Checkout Block](#checkout-block)
+- [](#)
```
opr:
```suggestion
```

docs/extensibility/available-slot-fills.md
```diff
@@ -2,9 +2,14 @@
...
--   [ExperimentalOrderShippingPackages](#experimentalordershippingpackages)
--   [ExperimentalDiscountsMeta](#experimentaldiscountsmeta)
+- [Available Slots](#available-slots)
+  - [ExperimentalOrderMeta](#experimentalordermeta)
+    - [Passed parameters](#passed-parameters)
+  - [ExperimentalOrderShippingPackages](#experimentalordershippingpackages)
+    - [Passed parameters](#passed-parameters-1)
+  - [ExperimentalDiscountsMeta](#experimentaldiscountsmeta)
+    - [Passed paramters](#passed-paramters)
+  - [](#)
```
opr:
```suggestion
```

docs/extensibility/filtering-payment-methods.md
```diff
@@ -11,6 +11,7 @@
   - [The solution](#the-solution-1)
   - [Basic usage](#basic-usage)
   - [Putting it all together](#putting-it-all-together)
+  - [](#)
```
opr:
```suggestion
```

tests/php/StoreApi/Utilities/OrderController.php
```diff
@@ -27,5 +27,6 @@ public function test_validate_selected_shipping_methods() {
 		$class->validate_selected_shipping_methods( true, array( 'free-shipping' ) );
 		$class->validate_selected_shipping_methods( false, array( 'free-shipping' ) );
 		$class->validate_selected_shipping_methods( true, null );
+		$this->assertTrue( true );
```
opr, line 32: Is there a reason for this being here? If so could you comment it please :)

mikejolley:
I'm going to give this another pass. I think we might see cases where we need a system like this at route level, not just schema. Investigating.

mikejolley:
Looks like we'd have to create an ExtendRoute if that need arose. Thats fine I think. Ready for review.

opr (approved):
I tested this and it all works fine, left a couple of comments but nothing huge. The extensibility points are still working OK - I tested with WooCommerce Points and Rewards and WooCommerce Subscriptions.

tests/php/StoreApi/Utilities/OrderController.php
mikejolley: Done. 

```
// The above methods throw Exception on error, but this is classed as a risky test because there are no 
// assertions. Assert true to work around this warning.
```

senadir (approved):
I did a code scan, this is mainly a renaming/moving PR so it's safe to approve. I like the new name, love the alias deprecation option (it should just be tested with a third party plugin), and dig the new functions!

PR #5944, opened 2022-02-24
Change `curl` example calls to use `sh` code block

kaznovac:
Change curl examples 'code block's language' from `http` to `sh` to solve github marking the examples as syntactically incorrect


mikejolley (approved):
👍🏻  Thanks makes sense.

PR #5946, opened 2022-02-25
Determine if we need to load the Mini Cart from the theme or WooCommerce

### LINKED ISSUE #5936: Mini Cart doesn't load template part from theme
## Describe the bug

After #5746 and #5916, the Mini Cart no longer loads the custom template part from themes.

## To reproduce

1. Clear any customization made to the Mini Cart template part from the Site Editor.
2. Copy the `woocommerce-gutenberg-products-block/templates/parts/mini-cart.html` to the active block theme template part folder.
3. Edit the copied template part to distinguish between it and the original. Here I change the empty cart message from `Your cart is currently empty!` to `From the theme: Your cart is currently empty!`.
4. Add the mini cart block to header.
5. Go to the frontend, ensure cart doesn't contain any product.
6. Open the Mini Cart.
7. See the original message: `Your cart is currently empty!`. Which means that the template is load from plugin.
8. Go to Site Editor, edit the Mini Cart template part, notice that it shows the theme in the `Added by` column.
9. Make changes to the template part.
10. See front-end unchanged.

## Expected behavior

`mini-cart.html` template part from the theme should be loaded if exists.

tjcafferkey:
Check if we should load the template part from the theme files or WooCommerce (in that order).

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5936

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Clear any customization made to the Mini Cart template part from the Site Editor.
2. Copy the `woocommerce-gutenberg-products-block/templates/parts/mini-cart.html` to the active block theme template part folder.
3. Edit the copied template part to distinguish between it and the original. Here I change the empty cart message from `Your cart is currently empty!` to `From the theme: Your cart is currently empty!`.
4. Add the mini cart block to header.
5. Go to the frontend, ensure cart doesn't contain any product.
6. Open the Mini Cart.
7. See the original message: `From the theme: Your cart is currently empty!`. Which means that the template is loaded from the theme.
8. Customize the template part and check that the customizations are loaded.
9. Clear the customizations and then delete the mini cart file from the theme and complete the steps again checking that when the theme file does not exist we can load and customize the WooCommerce version of the Mini Cart.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

1.
2.
3.

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -377,7 +377,11 @@ class MiniCart extends AbstractBlock {
 		}
 
 		$template_part_contents = '';
-		$template_part          = BlockTemplateUtils::get_block_template( BlockTemplateUtils::PLUGIN_SLUG . '//mini-cart', 'wp_template_part' );
+
+		// Determine if we need to load the template part from the theme, or WooCommerce in that order.
+		$theme_has_mini_cart   = BlockTemplateUtils::theme_has_template_part( 'mini-cart' );
+		$template_slug_to_load = $theme_has_mini_cart ? get_stylesheet() : BlockTemplateUtils::PLUGIN_SLUG;
+		$template_part         = BlockTemplateUtils::get_block_template( $template_slug_to_load . '//mini-cart', 'wp_template_part' );
 
 		if ( $template_part && ! empty( $template_part->content ) ) {
 			$template_part_contents = do_blocks( $template_part->content );


gigitux (approved):
Thanks for working on this! Everything works correctly!

I just noticed that the name on the editor is different:

image
*template loaded from theme*

image
*template loaded from plugin*

Is it normal or can we fix this?

I'm approving the PR, but if you think that we can fix this, we should do.

tjcafferkey:
>Is it normal or can we fix this?

It is the responsibility of the theme to define the template and template parts title, they can (and would) typically do so via the `theme.json` file as noted [here](https://developer.wordpress.org/themes/advanced-topics/theme-json/#assigning-template-parts) so the behaviour you've pointed out is actually expected seeing as in the testing instructions I didnt instruct you to update your themes `theme.json` file.

PR #5949, opened 2022-02-25
Add a blocks file that gathers data about block to reuse in our tests

ralucaStan:
I've noticed that the block related data we use in our tests is not aggregated, so that we can reuse it with certainty in our tests. We don't have a visible and single source of truth for blocks, which is safely reusable.

I've created `blocks.js` in order to have one source of truth about the block data used in tests.
Each block is defined independently and reused in corresponding tests.

Available block data: 
 - title: fixture data, used for defining the block page URL when the test site is created
 - name: used as page title
 - slug: not really used in tests
 - class: CSS class present on the block's wrapper

With this structure in place we can have a more structured approach when dealing with blocks, especially navigating to shopper block pages.

With the utility below we can access any block's page url without navigating to the admin to find out the permalink. This will result in speed improvements.

```
export const getBlockPageURL = ( block ) =>
	baseUrl + block.title.toLowerCase().split( ' ' ).join( '-' );
```

As an example I've implemented two methods for Cart & Checkout, available under `shopper`.
```
import { cart, checkout } from '../e2e/blocks';
import { getBlockPageURL } from '../e2e/utils';

goToCheckoutBlock: async () => {
		await page.goto( getBlockPageURL( checkout ), {
			waitUntil: 'networkidle0',
		} );
	},
```


dinhtungdu:
@ralucaStan This approach definitely cut down many seconds running the tests, but I have some concerns:
- This approach won't work if the pretty permalink is disabled.
- If for any reason, the test data wasn't cleared correctly, the slug of block pages will be different from our assumption. I think we shouldn't know about the slug of the pages.

Those make the approach a bit less reliable IMO. Even though the approach works in most cases, this can make us confused when the edge cases come in, we can take #5914 as an example.

In https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5958, I tried to refactor the `goToBlockPage()` to speed it up. Can you please take a look?

ralucaStan:
@dinhtungdu you make some really good points in your answer, thank you for that. Indeed, the approach is dependent on some settings, and that is ideal. I would not go forward with it. I will add my comments on the PR you mention

What do you think about the blocks.js file? I would like to keep it as a collection of the blocks we use in the tests because:
- it's a good overview
- it avoids duplication, as we now redeclare our block data in each test, depending on the need

dinhtungdu:
> What do you think about the blocks.js file?

@ralucaStan I'm sorry for the late reply. I'm thinking about auto-generate that file instead of something hardcoded and committed to the repo. In the setup phase, we create block pages using fixture data. We can update the "blocks.js" file with the information we need such as the permalink and post ID then use it to navigate between block pages.

> it avoids duplication, as we now redeclare our block data in each test, depending on the need

Can you explain more about the duplication? AFAIK, the test data is removed after running the test, so I don't know where the duplication happens.

PR #5950, opened 2022-02-25
Global Styles support for Product Price block

### LINKED ISSUE #5672: Product Price: Improve Global Styles API integration
The `Product Price` block, with the Global Styles API, should support:

- font-style
- font-size
- font-weight 
- specific text color for sale price -> it impossible to implement with the current Global Styles API
- specific text color for regular price ->  it impossible to implement with the current Global Styles API

tomasztunik:
Adds Global Styles support for Product Price element block.

- ⚠️ **comes sat a cost of losing control over regular vs sale price**
- adds font style and weight controls
- enables new ui controls via `supports`
- includes refactor of attribute from align to textAlign as align
   is a reserved supports feature attribute and textAlign is
   used across the Gutenberg project when text-align css is used
- includes removal of attributes that are controlled by supports 
   
I’ve noticed that we are using align attribute in couple of places in a wrong way. Since introduction of supports: { align } Gutenberg uses context based align attributes for anything that is not align. IE. css text-align should be `textAlign` and so on. Apart of that it seemed like it might be causing conflicts when using these “reserved” attributes in other contexts (ie. product title align which doesn't use the supports property stopped working on front-end and works just fine if changed to another name). 
   

Fixes #5672

#### Accessibility

### Screenshots

|before|after|
|-|-|
|  Screenshot 2022-02-25 at 15 56 16 |  Screenshot 2022-02-25 at 15 43 09 |

### Manual Testing

1. Edit Global Styles for the Product Price block (color, font-size).
2. Add Single Product and All Products components to a page.
3. The Price component should reflect global styles.
4. Add another set of Single Product and All Products components.
5. For these override Global Styles with local styles.
6. Ensure both sets work as expected on the Front-End of your store.

### Changelog

> Add Global Styles support to the Product Price block.

### DIFF
--- a/assets/js/atomic/blocks/product-elements/price/block.js
+++ b/assets/js/atomic/blocks/product-elements/price/block.js
@@ -80,6 +80,9 @@ const Block = ( props ) => {
 			maxPrice={ prices?.price_range?.max_amount }
 			// This is the regular or original price when the `price` value is a sale price.
 			regularPrice={ prices.regular_price }
+			regularPriceClassName={ classnames( {
+				[ `${ parentClassName }__product-price__regular` ]: parentClassName,
+			} ) }
 		/>
 	);
 };
--- a/assets/js/atomic/blocks/product-elements/price/block.js
+++ b/assets/js/atomic/blocks/product-elements/price/block.js
@@ -9,111 +9,72 @@ import {
 	useInnerBlockLayoutContext,
 	useProductDataContext,
 } from '@woocommerce/shared-context';
-import { getColorClassName, getFontSizeClass } from '@wordpress/block-editor';
-import { isFeaturePluginBuild } from '@woocommerce/block-settings';
 import { withProductDataContext } from '@woocommerce/shared-hocs';
+
+/**
+ * Internal dependencies
+ */
+import {
+	useColorProps,
+	useTypographyProps,
+} from '../../../../hooks/style-attributes';
 
 /**
  * Product Price Block Component.
  *
  * @param {Object} props                          Incoming props.
  * @param {string} [props.className]              CSS Class name for the component.
- * @param {string} [props.align]                  Text alignment.
+ * @param {string} [props.textAlign]              Text alignment.
  * @param {string} [props.fontSize]               Normal Price font size name.
- * @param {number} [props.customFontSize]         Normal Price custom font size.
- * @param {string} [props.saleFontSize]           Original Price font size name.
- * @param {number} [props.customSaleFontSize]     Original Price custom font size.
  * @param {string} [props.color]                  Normal Price text color.
- * @param {string} [props.customColor]            Normal Price custom text color.
- * @param {string} [props.saleColor]              Original Price text color.
- * @param {string} [props.customSaleColor]        Original Price custom text color.
  * context will be used if this is not provided.
  * @return {*} The component.
  */
-const Block = ( {
-	className,
-	align,
-	fontSize,
-	customFontSize,
-	saleFontSize,
-	customSaleFontSize,
-	color,
-	customColor,
-	saleColor,
-	customSaleColor,
-} ) => {
+const Block = ( props ) => {
+	const { className, textAlign } = props;
 	const { parentClassName } = useInnerBlockLayoutContext();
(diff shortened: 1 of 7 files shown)


assets/js/atomic/blocks/product-elements/price/block.js
```diff
@@ -9,127 +9,89 @@ import {
...
 			price={ prices.price }
-			priceClassName={ priceClassName }
-			priceStyle={ isFeaturePluginBuild() ? priceStyle : {} }
 			// Range price props
 			minPrice={ prices?.price_range?.min_amount }
 			maxPrice={ prices?.price_range?.max_amount }
 			// This is the regular or original price when the `price` value is a sale price.
 			regularPrice={ prices.regular_price }
-			regularPriceClassName={ classnames( {
-				[ `${ parentClassName }__product-price__regular` ]: parentClassName,
```
gigitux, line 113: I guess that it is better to keep this class. If a user has a custom CSS that targets this class, the user will lose the customization. What do you think?

assets/js/atomic/blocks/product-elements/price/edit.js
```diff
@@ -1,133 +1,51 @@
...
-				description: __(
-					'Choose a product to display its price.',
-					'woo-gutenberg-products-block'
-				),
-			} ),
-	  ] )( PriceEdit )
-	: PriceEdit;
-
-export default Price;
+export default withProductSelector( {
```
gigitux, line 44: I think that we should use the HOC `withProductSelector` only when `isFeaturePluginBuild()` is true

assets/js/atomic/blocks/product-elements/price/index.js
```diff
@@ -16,14 +18,32 @@ import {
...
+	apiVersion: 2,
 	title,
 	description,
 	icon: { src: icon },
 	attributes,
 	edit,
+	save: Save,
+	supports: {
+		...( isFeaturePluginBuild() && {
+			...sharedConfig.supports,
```
gigitux: We should move `...sharedConfig.supports` outside of the check

gigitux (changes requested):
Thanks for working on this block  🚀 
I saw a PHP error: 
```
Block type "woocommerce/product-price" is already registered. 
```

We need to remove `product-price` from [this array](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5950/files#diff-eeb3ba10500aa641680d1400a0a55d65fca9eb1f086c176ac7e41960ba7756c4R240)

I left a couple of small comments, but overall looking good 💪

assets/js/atomic/blocks/product-elements/price/edit.js
tomasztunik: I think this should be always available as this block does nothing without product present.

gigitux: Mmm, but in this case, we are introducing a new block in Core. I guess that this is not the scope of this PR.

If I search `Product Price` is available as a block with our plugin, but this is not true with WooCommerce Core only.

image
*with the plugin*

image
*without plugin*

tomasztunik: Ah, I thought it was responsible only for adding product selector to the block toolbar hmm. How is it that it controls if block is selectable or not? Not sure If I follow how this HOC  in Edit controls works 😢 

(also your first screenshot is showing a bug with icons for blocks eek! see filter on 1st vs 2nd)

gigitux: After our investigation, we found that this check is not done here, but [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/d03169059e73b6fa8813ab48003bb6a11f40b1c9/assets/js/atomic/blocks/product-elements/shared/config.tsx#L32).

So it is a safe change even because other blocks don't have this check (for example `Product Image` and  `Product Rating`)

gigitux (approved):
Great job! LGTM :shipit:

tomasztunik:
Confirmed regression of possible user customisation via controls with @vivialice.

PR #5951, opened 2022-02-25
Unify copy on Featured Category with Featured Product block

tomasztunik:
Updates media remove label to read "Reset". 

This is to reflect how we agreed it would be best 
when working on Featured Product block

Related to #5886


frontdevde (approved):
LGTM

Issue #18, opened 2018-02-14
Flow: I want to change the layout of the products I've inserted


jameskoster:
The user should be able to specify how many columns / rows products should be arranged in to.

Current prototype: todo

Sketch: todo

Issue #17, opened 2018-02-14
Flow: I want the products I've inserted to be paginated


jameskoster:
Current prototype: todo

Sketch: todo

warrendholmes:
Requested here https://wordpress.org/support/topic/product-pages-3/

Issue #16, opened 2018-02-14
Flow: I want customers to be able to add all my products to their cart in a post/page


jameskoster:
Current prototype: https://automattic.invisionapp.com/share/P4FL7W6VDT8#/screens/276308942_01e
Select `All products` in scope selector.

Sketch: https://www.dropbox.com/s/dvp2x1poq1yukes/Add%20Products%20Block.sketch?dl=0

Issue #15, opened 2018-02-14
Flow: I want customers to be able to add products with a specific attribute & term to their cart in a post/page


jameskoster:
Current prototype: https://automattic.invisionapp.com/share/P4FL7W6VDT8#/screens/276308942_01e
Select `Filter products > Attribute` in scope selector.

Sketch: https://www.dropbox.com/s/dvp2x1poq1yukes/Add%20Products%20Block.sketch?dl=0

PR #5953, opened 2022-02-25
Store API - Cart JWT tokens/session handling

### LINKED ISSUE #5683: Load cart via key for headless and mobile apps
The Store API cart cannot be used in a headless app because it relies on cookies. Cookies cannot be accessed in cross-domain or headless environments.

What happens is you can add an item to the cart, but when you make another request, the cart is always empty. This is using the proper nonce authentication.

To make the cart work, you would need to send a cart or session key in the API response, not in a cookie header. Then allow loading a cart with that key with a header or url parameter. CoCart takes this approach as an example: https://docs.cocart.xyz/#cart-get-cart-for-guest-customers

Would you be open to supporting this?

mikejolley:
POC for woocommerce/woocommerce-blocks#5683 Includes changes from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6020

Returns a header containing a `Cart-Token` valid for 48h, which can be added as a request header to load a specific cart.

Works by extending the WC_Session class to intercept requests with a valid header.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests

#### User Facing Testing

1. Using a REST API client
2. Disabling basic auth
3. Adding something to the cart via POST `/wc/store/v1/cart/add-item`. Note down the value of the `Cart-Token` header.
4. Deleting or disabling cookies before requesting GET `/wc/store/v1/cart`
5. No items
6. Repeat this request with a header called `Cart-Token` and the value you noted earlier.
7. Cart response has items!

* [ ] Do not include in the Testing Notes 

### WooCommerce Visibility

* [x] WooCommerce Core
* [ ] Feature plugin
* [ ] Experimental

### Changelog

> StoreApi  requests will return a `Cart-Token` header that can be used to retrieve the cart from the corresponding session via **GET** `/wc/store/v1/cart`


src/StoreApi/Routes/V1/AbstractCartRoute.php
```diff
@@ -99,13 +102,65 @@ public function get_response( \WP_REST_Request $request ) {
...
+	/**
+	 * Generates a cart token for the response headers.
+	 *
+	 * Current namespace is used as the token Issuer.
+	 *
+	 * @param string $customer_id Session (customer ID) for the token.
+	 * @return string
+	 */
+	protected function get_cart_token( $customer_id ) {
+		return Token::create( $customer_id, $this->get_cart_token_secret(), $this->get_cart_token_expiration(), $this->namespace );
```
ant-magic42: Is it strictly necessary to use a JWT library to create this useful `Cart-Token` header?

Could the existing approach from [set_customer_session_cookie](https://github.com/woocommerce/woocommerce/blob/9db2c452905ebf796a8e8b4d63db64b57e169dbe/plugins/woocommerce/includes/class-wc-session-handler.php#L160-L161) be used?

Given the session cookie is already present in the API response headers, providing the same data in the `Cart-Token` would be consistent and avoid having to add the JWT library.

scottopolis:
This is great Mike. In a mobile app, you'd want the Cart-Token to stay valid for longer than 48 hours, could there be a filter on that?

If not, you'd need a way to save the cart locally in the app and restore it via API.

pluginslab:
Indeed, this is wonderful stuff. I'd say @scottopolis we'd save the cart, nonetheless. On communication failures or while on the run, going in and out of service, we'd like the cart to be updated locally, and then sync?

src/StoreApi/Routes/V1/AbstractCartRoute.php
wavvves: Hi Anthony, I picked up this PR. I agree we should get rid of this library if we can. Also, we're using it with `wp_salt()` as the secret and it enforces a "secretValidation" (🙄) causing it to throw an error if our provided secret does not obey certain security rules (either default or provided by us). And nowhere it is certain that `wp_salt()` will, I don't believe we enforce salt keys rules right?

I'm having trouble locating the header you mentioned in API responses, can you send me an example of the response header you are talking about, or point out where is it inserted?

wavvves: I left [some feedback](https://github.com/woocommerce/woocommerce-blocks/issues/5683#issuecomment-1158771475) on the Issue thread about this.

albarin:
This is looking good!
Just one question, would it be possible to add a test that does basically what the PR testing instructions say? (adding a product, getting the cart-token and retrieve the cart using it).

albarin (approved):

pluginslab:
This is looking great. I have a few questions if you don't mind, as I understand we won't rely on the session cookie anymore. 

Let's say I add an item, and get the `Cart-Token`, what happens if I then add another item? Do I need to use the same `Cart-Token` on that add-item request?

Also, what if we empty the cart, is the `Cart-Token` the same or will the add-item provide a new one?

Thanks! 🙏🏻

senadir:
> Let's say I add an item, and get the Cart-Token, what happens if I then add another item? Do I need to use the same Cart-Token on that add-item request?

Yes, you will need to send your current token, and you will receive a new token. This is because tokens reflect an actual cart content, so changing its content will generate a new token.

You will need to manage the token the same way you manage a nonce, by intercepting incoming and outgoing requests.

> Also, what if we empty the cart, is the Cart-Token the same or will the add-item provide a new one?

An empty cart will have no token, adding an item will get you a new token, this token might be the same if you added the same item again, because tokens are a hash of the cart content.

wavvves:
> This is looking good! Just one question, would it be possible to add a test that does basically what the PR testing instructions say? (adding a product, getting the cart-token and retrieve the cart using it).

Hey @albarin, thanks for reviewing this. 
Actually, I tried to [add those tests](https://github.com/woocommerce/woocommerce-blocks/pull/5953/commits/4f650e961ff714a9089bc924d57df87969725eab) without success, testing headers is not very easy, specially when we use filters to replace session handling etc.

schmitzoide:
Writing a test using Javascript would possibly be a way to have more control over headers and cookies, since it runs on the "client side"?

wavvves:
> Writing a test using Javascript would possibly be a way to have more control over headers and cookies, since it runs on the "client-side"?

Initially, I started out with E2E for that with @opr's help, but it was turning out not to be as straightforward as expected and I went back into PHP unit testing as I had [previous success](https://github.com/woocommerce/woocommerce-blocks/pull/5962/files#diff-b4cca344144eb7420dad7bd616bcbb192cdc331b8e5924b07b31e94eb5f4d0b6) with it.

I'll give JS another go 😉

EDIT: Due to including this in the next 8.5.0 release next week, the test will be added in a separate PR.

composer.json
```diff
@@ -12,9 +12,11 @@
 	"prefer-stable": true,
 	"minimum-stability": "dev",
 	"require": {
+		"ext-hash": "*",
+		"ext-json": "*",
```
senadir, line 16: Are those still needed? I can't find a reference for them?

src/StoreApi/Routes/V1/AbstractCartRoute.php
```diff
@@ -81,7 +93,7 @@ public function get_response( \WP_REST_Request $request ) {
 		} catch ( RouteException $error ) {
 			$response = $this->get_route_error_response( $error->getErrorCode(), $error->getMessage(), $error->getCode(), $error->getAdditionalData() );
 		} catch ( \Exception $error ) {
-			$response = $this->get_route_error_response( 'woocommerce_rest_unknown_server_error', $error->getMessage(), 500 );
+			$response = $this->get_route_error_response( 'woocommerce_rest_unknown_server_error', $error->getMessage() );
```
senadir, line 96: Why are we omitting the error code here?

src/StoreApi/Routes/V1/AbstractCartRoute.php
```diff
@@ -181,7 +255,7 @@ protected function check_nonce( \WP_REST_Request $request ) {
 		if ( $request->get_header( 'Nonce' ) ) {
 			$nonce = $request->get_header( 'Nonce' );
 		} elseif ( $request->get_header( 'X-WC-Store-API-Nonce' ) ) {
-			$nonce = $request->get_header( 'X-WC-Store-API-Nonce' );
+			$nonce = $request->get_header( 'Nonce' );
```
senadir: This should be kept the same for backward compatibility.

src/StoreApi/SessionHandler.php
```diff
@@ -0,0 +1,107 @@
...
+	 * @var string
+	 */
+	protected $token;
+
+	/**
+	 * Table name for session data.
+	 *
+	 * @var string Custom session table name
+	 */
+	protected $_table; // phpcs:ignore PSR2.Classes.PropertyDeclaration.Underscore
```
mikejolley: I don't think we need to use the _ prefix here? The property is already declared as protected.

src/StoreApi/SessionHandler.php
```diff
@@ -0,0 +1,107 @@
...
+
+	/**
+	 * Expiration timestamp.
+	 *
+	 * @var int
+	 */
+	protected $_session_expiration; // phpcs:ignore PSR2.Classes.PropertyDeclaration.Underscore
```
mikejolley: This prefix has to stay because it's used in WC_Session. We don't need to redeclare it but maybe this was done to prevent linting errors?

src/StoreApi/SessionHandler.php
```diff
@@ -0,0 +1,107 @@
...
+	 * @return string|array|bool
+	 */
+	public function get_session( $customer_id, $default = false ) {
+		global $wpdb;
+
+		if ( Constants::is_defined( 'WP_SETUP_CONFIG' ) ) {
+			return false;
+		}
+
+		$value = $wpdb->get_var( $wpdb->prepare( "SELECT session_value FROM $this->_table WHERE session_key = %s", $customer_id ) ); // @codingStandardsIgnoreLine.
```
mikejolley: We should turn off specific linting rules if possible rather than use the blanket `@codingStandardsIgnoreLine` because that can hide legitimate issues.

src/StoreApi/SessionHandler.php
```diff
@@ -0,0 +1,115 @@
...
+	 */
+	public function get_session( $customer_id, $default = false ) {
+		global $wpdb;
+
+		if ( Constants::is_defined( 'WP_SETUP_CONFIG' ) ) {
+			return false;
+		}
+
+		$value = $wpdb->get_var(
+			$wpdb->prepare(
```
mikejolley, line 80: This is causing an SQL error. If error reporting/logging is enabled you should see it when you make the request. The table name cannot be prepared because WordPress will wrap it in quotes.

composer.json
```diff
@@ -12,9 +12,11 @@
...
 		"composer/installers": "^1.7.0",
 		"automattic/jetpack-autoloader": "^2.9.1"
-	},
+    },
```
mikejolley: Indents are wrong here I think? Should be tabs.

docs/internal-developers/testing/releases/440.md
```diff
@@ -96,9 +96,7 @@ This is a regression test due to the changes made for Subscriptions Integration.
 2. Enter an address that has no rates. Ensure an error notice is shown in the cart shipping package "No shipping options were found.".
 3. Place an order. Confirm details persist.
 
-<!-- FEEDBACK -->
-
----
+## <!-- FEEDBACK -->
```
mikejolley: I think this has borked. It shouldn't be a heading. Can we revert these lines?

src/StoreApi/SessionHandler.php
```diff
@@ -0,0 +1,115 @@
...
+	 *
+	 * @param string $customer_id Customer ID.
+	 * @param mixed  $default Default session value.
+	 *
+	 * @return string|array|bool
+	 */
+	public function get_session( $customer_id, $default = false ) {
+		global $wpdb;
+
+		if ( Constants::is_defined( 'WP_SETUP_CONFIG' ) ) {
```
mikejolley, line 75: Should probably add a comment about why we're doing this.

src/StoreApi/SessionHandler.php
```diff
@@ -0,0 +1,115 @@
...
+	 * Save data and delete user session.
+	 */
+	public function save_data() {
+		// Dirty if something changed - prevents saving nothing new.
+		if ( $this->_dirty ) {
+			global $wpdb;
+
+			$wpdb->query(
+				$wpdb->prepare(
+					'INSERT INTO %s (`session_key`, `session_value`, `session_expiry`) VALUES (%s, %s, %d)
```
mikejolley: The `%s` for table name will break the query.

src/StoreApi/Utilities/JsonWebToken.php
```diff
@@ -0,0 +1,194 @@
...
+
+		$parts = self::get_parts( $token );
+
+		/**
+		 * Check if header declares a supported JWT by this class.
+		 */
+		if (
+			! is_object( $parts->header ) ||
+			! property_exists( $parts->header, 'typ' ) ||
+			! property_exists( $parts->header, 'alg' ) ||
```
mikejolley, line 69: Alternatively could combine these into a single isset if cleaner. e.g. `! isset( $parts->header->typ, $parts->header->alg )`

src/StoreApi/Utilities/JsonWebToken.php
```diff
@@ -0,0 +1,194 @@
...
+		 * Check if token is expired.
+		 */
+		if ( ! property_exists( $parts->payload, 'exp' ) || time() > (int) $parts->payload->exp ) {
+			return false;
+		}
+
+		/**
+		 * Check if the token is based on our secret.
+		 */
+		return self::to_base_64_url(
```
mikejolley: Is `hash_equals` more appropriate?

src/StoreApi/docs/cart.md
```diff
@@ -469,9 +469,7 @@ curl --header "Nonce: 12345" --request POST /cart/select-shipping-rate?package_i
 
 Returns the full [Cart Response](#cart-response) on success, or an [Error Response](#error-response) on failure.
 
-<!-- FEEDBACK -->
-
----
+## <!-- FEEDBACK -->
```
mikejolley: Needs reverting.

src/StoreApi/docs/checkout.md
```diff
@@ -135,9 +135,7 @@ curl --header "Nonce: 12345" --request POST https://example-store.com/wp-json/wc
 }
 ```
 
-<!-- FEEDBACK -->
-
----
+## <!-- FEEDBACK -->
```
mikejolley: Needs reverting.

src/StoreApi/docs/nonce-tokens.md
```diff
@@ -47,9 +47,7 @@ Nonce checks will be bypassed if `woocommerce_store_api_disable_nonce_check` eva
 
 NOTE: This should only be done on development sites where security is not important. Do not enable this in production.
 
-<!-- FEEDBACK -->
-
----
+## <!-- FEEDBACK -->
```
mikejolley: Needs reverting.

mikejolley (commented):
Added some comments and suggestions. I'm not sure why the documentation broke in this PR but those changes can be reversed. Cheers!

senadir (commented):
Those are some of my initial review comments that I forget to submit.

src/StoreApi/SessionHandler.php
wavvves: It is used on WC_Session_Handler, not the abstract WC_Session, we need to declare it. What we don't need is keeping up with the underscore prefix like the $_table example. I'll remove that for those, as they were only done to keep WC_Session_Handler "tradition" 👍🏼

src/StoreApi/Routes/V1/AbstractCartRoute.php
mikejolley: `X-WC-Store-API-Nonce` is deprecated, we use both in places.

composer.json
wavvves: good catch, seems like .editorconfig has conflicting settings, fixed that too

```
[*.{json,yml}]
trim_trailing_whitespace = false
indent_style = space
indent_size = 2

[*.json]
indent_style = tab
```

src/StoreApi/SessionHandler.php
wavvves: Thanks for noticing this, reverted!

src/StoreApi/Utilities/JsonWebToken.php
wavvves: I usually tend to use `isset()` for particular cases since it also returns `false` if the property existed but had `null` value. We couldn't compare it to `null` later (won't probably ever be the case here) and might make the code prone to errors. Personally, I also find these lines more legible and semantic "an object possessing these properties carrying these values".

What do you think?

src/StoreApi/Utilities/JsonWebToken.php
wavvves: Awesome catch Mike 😃. Timing attack possibility mitigated 🔫

composer.json
wavvves: `ext-json` was already needed as it is a [WordPress requirement](https://make.wordpress.org/hosting/handbook/server-environment/#php-extensions). Since I added more json_*() function usage I also updated composer.json 😃 

`ext-hash` is an optional requirement (listed on the same page) and we are using `hash_hmac()` to generate the `sha256` signature required for an HS256 JWT

src/StoreApi/Routes/V1/AbstractCartRoute.php
wavvves: It's the default value on the function signature

```
protected function get_route_error_response( $error_code, $error_message, $http_status_code = 500, $additional_data = [] ) { ... }
```

src/StoreApi/Routes/V1/AbstractCartRoute.php
wavvves: That was wrongly changed, I already changed it back, saw it didn't make sense with the if/else 👍🏼

src/StoreApi/SessionHandler.php
wavvves: Fixed 👍🏼

tests/php/StoreApi/Routes/ControllerTestCase.php
```diff
@@ -95,9 +95,9 @@ public function assertEqualSets( $expected, $actual ) {
 	/**
 	 * Custom assertion of an API response to confirm data and response code is expected.
 	 *
-	 * @param string $endpoint_or_request Route endpoint to get.
-	 * @param int    $expected_response_code Expected response code.
-	 * @param array  $expected_response_data Expected response data.
+	 * @param mixed $endpoint_or_request Route endpoint to get.
```
mikejolley: Should this be `string|\WP_Rest_Request`?

src/StoreApi/docs/cart.md
```diff
@@ -471,8 +471,6 @@ Returns the full [Cart Response](#cart-response) on success, or an [Error Respon
 
 <!-- FEEDBACK -->
```
mikejolley: These files still have some unrelated changes for some reason.

mikejolley (commented):
Thanks for working on this some more. I made 2 small comments.

Does this PR need to be rebased with trunk? Github is complaining about conflicts. I'm happy to do some more in depth testing after that to ensure there are no regressions.

src/StoreApi/docs/cart.md
wavvves: Sorry about this, seems like some rogue *.md changes spilled into this PR, not sure what happened. Reverted all of them 👍🏼

tests/php/StoreApi/Routes/ControllerTestCase.php
wavvves: Yes it should, fixed!

wavvves:
> Thanks for working on this some more. I made 2 small comments.
> 
> Does this PR need to be rebased with trunk? Github is complaining about conflicts. I'm happy to do some more in depth testing after that to ensure there are no regressions.

Where is the conflict warning 🤔? I have been updating it with trunk changes, and so far I haven't dealt with any conflicts.

Can I mark the other comments as resolved?

mikejolley:
> Can I mark the other comments as resolved?

👍🏻 Yup

> Where is the conflict warning 

I just see a warning thats the branch is out of date. I'm not sure if this has changed since my afk, but I think it's preferable to rebase with trunk rather than merge in — maybe thats why? @senadir

senadir:
> I just see a warning thats the branch is out of date. I'm not sure if this has changed since my afk, but I think it's preferable to rebase with trunk rather than merge in — maybe thats why? @senadir

It could be that, we enabled this check recently but I'm not sure why it's not rebasing.

senadir:
So I reviewed this and while it works fine and all, it seems to return a new Cart Token on each GET/POST request, I was under the impression that the token would be somehow stable? I will investigate more if it's a sign of time or we're actually creating a new cart each time.

senadir:
Okay so I checked the session table and we still have a single session even if the token is changing, this should be fine, I honestly don't understand why, because old tokens still work fine :/

senadir (approved):
LGTM, :shipit:

wavvves:
> Okay so I checked the session table and we still have a single session even if the token is changing, this should be fine, I honestly don't understand why, because old tokens still work fine :/

@senadir this is expected behavior, all of those previous tokens point to the same session. The tokens change because with each request the expiration gets renewed, so the token hash is different. Still, all cart changes are made to the same session. We don't need to save tokens, as, being JWT, the signature is verified always.

asumaran:
@mikejolley @wavvves With these changes, would it be possible to create and manage multiple WooCommerce Carts passing a token ID only? Context: p91TBi-5FD-p2#comment-5690

wavvves:
@asumaran It is possible to manage multiple carts, from different users. The Cart-Token, upon signature verification, loads up the corresponding session for it. The nonce must still be supplied for now though.

asumaran:
> It is possible to manage multiple carts, from different users.

@wavvves For my use case we need to be able to manage multiple WC Carts wether the user is logged in or not. Do you think this can be possible with these changes or adding additional code on top of it?

wavvves:
Only the current Cart for each session can be accessed, but as long as Cart-Token header is provided along a valid nonce for that particular session, you can manipulate multiple Carts.

An example, access GET `/wc/store/v1/cart` on your browser  (logged on or not) and get the Cart-Token and Nonce values from the response headers. Now, with a separate rest client (make sure cookies are off), update or add items supplying those two values in the request headers. If you go back into your browser and refresh the cart endpoint, you should see all changes you made via the rest client.

ant-magic42:
Well done @senadir @mikejolley @wavvves this works well in woocommerce 7.2.2 🥳 

```
curl -v \
--header "Nonce: <nonce goes here>" \
--header "cart-token: <cart token goes here>" \
--request POST \
"https://<your store here>/wp-json/wc/store/v1/cart/add-item?id=<id goes here>&quantity=<quantity goes here>"
```
with full API docs here https://github.com/woocommerce/woocommerce-blocks/tree/trunk/src/StoreApi

Endless headless possibilities!

pierre-dekode:
This is great, thank you! Is there any official documentation mentioning this `Cart-Token` header somewhere?

wavvves:
Hi @pierre-dekode, no official documentation has been released yet about Cart-Token.  Still here is some information to help you get started:

### How to obtain a Cart-Token
Cart endpoints will now return a Cart-Token header in the response headers. This contains a [JSON Web Token (JWT)](https://href.li/?https://jwt.io/), which can later be sent as a request header to the Store API Cart and Checkout endpoints.

The quickest way to obtain one is to request GET /wp-json/wc/store/v1/cart and observe the response headers. You should see the Cart-Token header there.

image

### How to use a Cart-Token
Include it in your request for GET /wp-json/wc/store/v1/cart, and the response will contain the current cart state from the session associated with the Cart-Token.
image

Tip: Add things to a cart in the browser on a logged-in standard session. Take note of the Cart-Token value returned in the cart endpoints response. Supply that token as a header in a request made through a Rest client app or Curl to receive the cart contents from your previous browser session.

Furthermore, you can manipulate cart contents (eg: POST /wp-json/wc/store/v1/cart/add-item) by submitting a valid [Nonce](https://href.li/?https://developer.woocommerce.com/2022/03/25/store-api-is-now-considered-stable/#:~:text=We%20also%20need%20a%20special%20header%20(Nonce)%20to%20confirm%20the%20user%20actually%20wants%20to%20checkout.%20This%20header%20is%20returned%20via%20other%20requests%20so%20can%20be%20stored%20in%20your%20client.) request header along Cart-Token.

The same method will allow you to checkout using JWT and Nonce via /wp-json/wc/store/v1/checkout

pierre-dekode:
Thank you @wavvves. That's what I ended up doing, but your message will surely help many other people :)

jayostwalapporio:
@wavvves I am using the same method of 'cart-token' for cart apis but its giving error.
image

I have checked the logs and it gives following error

PHP Fatal error:  Uncaught Error: Call to undefined method Automattic\WooCommerce\StoreApi\SessionHandler::has_session() in /wordpress/plugins/woocommerce-gift-cards/1.15.5/includes/class-wc-gc-cart.php:194
Stack trace:
#0 /wordpress/core/6.2.1/wp-includes/class-wp-hook.php(308): WC_GC_Cart->after_calculate_totals(Object(WC_Cart))
woocommerce/woocommerce-blocks#12058 /wordpress/core/6.2.1/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters(NULL, Array)
woocommerce/woocommerce-blocks#2 /wordpress/core/6.2.1/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
woocommerce/woocommerce-blocks#3 /wordpress/plugins/woocommerce/7.7.0/includes/class-wc-cart.php(1403): do_action('woocommerce_aft...', Object(WC_Cart))
woocommerce/woocommerce-blocks#4 /wordpress/plugins/woocommerce/7.7.0/packages/woocommerce-blocks/src/StoreApi/Utilities/CartController.php(43): WC_Cart->calculate_totals()
woocommerce/woocommerce-blocks#5 /wordpress/plugins/woocommerce/7.7.0/packages/woocommerce-blocks/src/StoreApi/Routes/V1/AbstractCartRoute.php(98): Automattic\WooCommerce\StoreApi\Utilities\CartController->calculate_totals()
woocommerce/woocommerce-blocks#6 /wordpress/core/6.2.1/wp-includes/rest-api/class-wp-rest-server.php(1181): Automattic\WooCommerce\StoreApi\Routes\V1\AbstractCartRoute->get_response(Object(WP_REST_Request))
woocommerce/woocommerce-blocks#7 /wordpress/core/6.2.1/wp-includes/rest-api/class-wp-rest-server.php(1028): WP_REST_Server->respond_to_request(Object(WP_REST_Request), '/wc/store/v1/ca...', Array, NULL)
woocommerce/woocommerce-blocks#8 /wordpress/core/6.2.1/wp-includes/rest-api/class-wp-rest-server.php(442): WP_REST_Server->dispatch(Object(WP_REST_Request))
woocommerce/woocommerce-blocks#9 /wordpress/core/6.2.1/wp-includes/rest-api.php(410): WP_REST_Server->serve_request('/wc/store/v1/ca...')
woocommerce/woocommerce-blocks#10 /wordpress/core/6.2.1/wp-includes/class-wp-hook.php(308): rest_api_loaded(Object(WP))
woocommerce/woocommerce-blocks#11 /wordpress/core/6.2.1/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters(NULL, Array)
woocommerce/woocommerce-blocks#12 /wordpress/core/6.2.1/wp-includes/plugin.php(565): WP_Hook->do_action(Array)
woocommerce/woocommerce-blocks#13 /wordpress/core/6.2.1/wp-includes/class-wp.php(399): do_action_ref_array('parse_request', Array)
woocommerce/woocommerce-blocks#14 /wordpress/core/6.2.1/wp-includes/class-wp.php(780): WP->parse_request('')
woocommerce/woocommerce-blocks#15 /wordpress/core/6.2.1/wp-includes/functions.php(1334): WP->main('')
woocommerce/woocommerce-blocks#16 /wordpress/core/6.2.1/wp-blog-header.php(16): wp()
woocommerce/woocommerce-blocks#17 /wordpress/core/6.2.1/index.php(17): require('/wordpress/core...')
woocommerce/woocommerce-blocks#18 {main}
  thrown in /wordpress/plugins/woocommerce-gift-cards/1.15.5/includes/class-wc-gc-cart.php on line 194

Please help me I am really stuck in this and I am new to woocommerce

wavvves:
@jayostwalapporio thanks for reporting that. There is currently [work in progress](https://github.com/woocommerce/woocommerce-blocks/pull/9163) trying to fix that problem. Please [create an issue](https://github.com/woocommerce/woocommerce-blocks/issues/new/choose) for that so I can link it to the PR

PR #5958, opened 2022-02-28
Refactor: improve goToBlockPage performance

dinhtungdu:
Relates #5949 

This PR speeds up the `shopper.goToBlockPage()`.

Currently, `shopper.goToBlockPage()` uses `getBlockPagePermalink()`, which is very slow because of how `getBlockPagePermalink()` works:

1. Search for the block page in the page list table.
2. Go to the edit page of the page.
3. Open the sidebar if needed.
4. Toggle the Permalink panel.
5. Get the link to the page on the front end.
6. Navigate to the page using `page.goto()`.

This PR refactors `shopper.goToBlockPage()` so it takes fewer steps to navigate to a block page:
1. Search for the block page in the page list table.
2. Get the edit link and build the page permalink using the page id (which is extracted from the edit link): `http://localhost:8889/?p=1234`.
3. Navigate to the block page permalink.

Notes: 
- Initially, I want to refactor the `getBlockPagePermalink()`, however, that utility doesn't only get the permalink of an existing page but also create the page if the page with the given title doesn't exist.
- Go to the admin page won't slow us down because by default, before every test, we go to Admin Dashboard > Products.

### DIFF
--- a/tests/utils/visit-block-page.js
+++ b/tests/utils/visit-block-page.js
@@ -33,16 +33,7 @@ async function visitPage( link ) {
 	}
 }
 
-/**
- *
- * @param {string} title the page title, written as `BLOCK_NAME block`
- *
- * This function will attempt to search for a page with the `title`
- * if that block is found, it will open it, if it's not found, it will open
- * a new page, insert the block, save the page content and title as a fixture file.
- * In both cases, this page will end up with a page open with the block inserted.
- */
-export async function visitBlockPage( title ) {
+export async function getBlockPageEditLink( title ) {
 	let link = '';
 	// Visit Import Products page.
 	await visitAdminPage( 'edit.php', 'post_type=page' );
@@ -61,6 +52,22 @@ export async function visitBlockPage( title ) {
 			);
 		}
 	}
+
+	return link;
+}
+
+/**
+ *
+ * @param {string} title the page title, written as `BLOCK_NAME block`
+ *
+ * This function will attempt to search for a page with the `title`
+ * if that block is found, it will open it, if it's not found, it will open
+ * a new page, insert the block, save the page content and title as a fixture file.
+ * In both cases, this page will end up with a page open with the block inserted.
+ */
+export async function visitBlockPage( title ) {
+	const link = await getBlockPageEditLink( title );
+
 	if ( link ) {
 		await visitPage( link );
 	} else {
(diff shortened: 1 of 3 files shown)


tests/utils/visit-block-page.js
```diff
@@ -61,6 +52,22 @@ export async function visitBlockPage( title ) {
...
+/**
+ *
+ * @param {string} title the page title, written as `BLOCK_NAME block`
+ *
+ * This function will attempt to search for a page with the `title`
+ * if that block is found, it will open it, if it's not found, it will open
+ * a new page, insert the block, save the page content and title as a fixture file.
+ * In both cases, this page will end up with a page open with the block inserted.
+ */
+export async function visitBlockPage( title ) {
```
ralucaStan, line 68: could we use `blockName` instead of `title`, considering the use case is `await shopper.goToBlockPage( block.name );`?

ralucaStan (commented):
This looks ok with me. How is the time gain with this approach, compared to your initial approach?

I am ok with merging this, but I would also still like to keep the goToCartBlock and goToCheckoutBlock utilities we have now

ralucaStan (approved):

dinhtungdu:
> How is the time gain with this approach, compared to your initial approach?

@ralucaStan Interesting, there isn't a big difference between the two approaches.

```
~/Sites/wc/app/public/wp-content/plugins/woocommerce-gutenberg-products-block on refactor/goToBlockPage [$]
Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        8.314 s

~/Sites/wc/app/public/wp-content/plugins/woocommerce-gutenberg-products-block on trunk [$]
Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        8.843 s
```

More interestingly, @Aljullu added a simple but excellent cache to `getBlockPagePermalink()` which solve entirely the performance issue from the second time running this method. With this minimal performance gain, I'm closing this PR without merging it.

tests/utils/visit-block-page.js
dinhtungdu: I think `blockName` and `title` are the same things for block pages, we use the block name as the title for each block page. Using `title` is more correct to me in this case considering how the function works.

Take the Mini Cart one as an example:

```
{"title":"Mini Cart Block","pageContent":"<!-- wp:woocommerce/mini-cart /-->..."}
```

PR #5962, opened 2022-02-28
Experiment: Add Rate Limits to Store API

### LINKED ISSUE #5728: Add rate limiting to Store API
See https://github.com/woocommerce/woocommerce/blob/d12717211a7cb97b6eef426d48219c3cba804d30/plugins/woocommerce/includes/class-wc-rate-limiter.php

We should be able to add basic rate limiting to cart and checkout endpoints to avoid excessive updates which can hurt performance. We only need to target endpoints with POST/PUT support, rather than GET.

mikejolley:
This is an attempt at adding rate limiting to Store API endpoints. This uses a modified wc_rate_limit table with an additional remaining column for tracking the request count in any given request window. This allows us to limit X requests within X seconds. I've set this up as 5 requests in 60 seconds for now which should help prevent abuse of endpoints from excessive calls.

Since this is unauthenticated, rate limits are keyed by either USER ID (logged in) or IP ADDRESS (guest user). There is a [discussion](https://github.com/woocommerce/woocommerce-blocks/pull/pcShBQ-i4-p2) about IP rate limiting behind a proxy, load balancer, etc.

A filter is available for setting options for rate limiting:

```
add_filter( 'woocommerce_store_api_rate_limit_options', function() {
	return [
		'enabled' => RateLimits::ENABLED, // enables/disables Rate Limiting. Default: false
		'proxy_support' => RateLimits::PROXY_SUPPORT, // enables/disables Proxy support. Default: false
		'limit' => RateLimits::LIMIT, // limit of request per timeframe. Default: 25
		'seconds' => RateLimits::SECONDS, // timeframe in seconds. Default: 10
	];
} );
```

Closes https://github.com/woocommerce/woocommerce-blocks/issues/5728

#### Accessibility

#### Other Checks

- [ ] This PR adds/removes a feature flag & I've updated [this doc](../docs/blocks/feature-flags-and-experimental-interfaces.md) .
- [ ] This PR adds/removes an experimental interfaces and I've updated [this doc](../docs/blocks/feature-flags-and-experimental-interfaces.md)
- [X] I tagged two reviewers because this PR makes queries to the database or I think it might have some security impact.

#### Automated Tests
* [X] Changes in this PR are covered by Automated Tests.
  * [X] Unit tests
  * [ ] E2E tests
  
#### Developer Facing Testing (easier and more in-depth than regular user facing instructions)

**_Without proxy support:_**

1. Enable Rate Limiting
2. In a short window, keep making API requests.
3. Check that RateLimit-xxx headers change on each request.
4. Check that once you've hit the limit, an error response is returned. You can modify the limits in this PR to make this easier to test.

**_With proxy support, do the same as before and:_**

1. Enable proxy support
2. Make your requests with one of the following request headers containing always the same IP address:
    - X-Real-IP or Client-IP
    - [X-Forwarded-For](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)
    - [Fowarded](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded)

#### User Facing Testing

1. Enable Rate Limiting by using the options filter
2. Try to apply a coupon or access `/wp-json/wc/store/cart` beyond current limits (currently 25 requests under 10 seconds)
3. Ensure you get presented with an error "_Too many requests. Please wait xx seconds before trying again_."

* [x] Do not include in the Testing Notes 

### WooCommerce Visibility

* [x] WooCommerce Core
* [x] Feature plugin
* [ ] Experimental

### Performance Impact

When enabled, this feature updates a new field on wc_rate_limits table to maintain track of requests by user/IP. This is a minor negative performance impact, regarding the feature is designed to prevent malicious attacks with a tremendous negative performance impact.

### Changelog

>  Adding rate limiting functionality to Store API endpoints


senadir:
Would this also work if the user has no session? I'm thinking out loud here but if someone was just to hit the same `/store/cart/add-item` with a new session each request would escape this.

To me, this would only protect the checkout order process endpoint, which isn't reached unless you have a valid session and cart, I think this goal by itself and worth pursuing.

I also understand that rate limiting shouldn't be the script responsiblity, but the server engine (nginx or apache). Would it be worthwhile to add an action that is triggered each time an API is added to the rate limting pool? hosts can (along side a mu-plugin) pick up on that and propagate it into their own edge firewall.

src/Installer.php
```diff
@@ -59,6 +60,22 @@ public function maybe_create_tables() {
 			"
 		);
 
+		require_once ABSPATH . 'wp-admin/includes/upgrade.php';
+
+		// Modify wc_rate_limits to include extra column.
+		dbDelta(
```
vedanshujain: We definitely need to do this in core as well, we run table definitions via dbDelta on WC updates. Although dbDelta does not drop a column, only alter or create, so there won't be conflicts even on old versions.

src/StoreApi/Utilities/RateLimits.php
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		return $current_limit;
+	}
+
+	/**
+	 * If exceeded, seconds until reset.
+	 *
+	 * @param string $action_id Identifier of the action.
+	 * @return bool
```
vedanshujain: nitpick: `@return bool|int`

src/StoreApi/Utilities/RateLimits.php
```diff
@@ -0,0 +1,125 @@
...
+	 */
+	protected static function get_rate_limit_row( $action_id ) {
+		global $wpdb;
+
+		$row = $wpdb->get_row(
+			$wpdb->prepare(
+				"
+					SELECT rate_limit_expiry as reset, rate_limit_remaining as remaining
+					FROM {$wpdb->prefix}wc_rate_limits
+					WHERE rate_limit_key = %s
```
vedanshujain, line 70: we need to add another clause for `rate_limit_expiry > now()`? otherwise `$wpdb` may return expired rows even if there are is a valid rate-limiting row present.

vedanshujain (commented):
Great work! Left a minor comment and a nitpick, otherwise looks good to me. I see that we fixed our window of attempts at the first attempt before reset which seems like a good approach.

I tested directly with SQL queries and works as expected. Although, we also probably need a cleanup routine to delete old data.

mikejolley:
Thanks @vedanshujain. I've taken care of feedback and raised a core PR here: https://github.com/woocommerce/woocommerce/pull/32041

Once I have an idea of what version this will ship with, I can either remove our install handling or lock it down to certain DB versions.

Thanks

vedanshujain (approved):
LGTM!

src/Installer.php
senadir: @vedanshujain if I understand correctly, now with https://github.com/woocommerce/woocommerce/pull/32041/ merged, we can delete this section?

senadir (approved):
Rebased and cleaned up, now for a round of testing and passing tests.

src/StoreApi/Routes/V1/AbstractCartRoute.php
```diff
@@ -86,11 +97,11 @@ public function get_response( \WP_REST_Request $request ) {
 
 		if ( is_wp_error( $response ) ) {
 			$response = $this->error_to_response( $response );
-		} elseif ( in_array( $request->get_method(), [ 'POST', 'PUT', 'PATCH', 'DELETE' ], true ) ) {
+		} elseif ( $this->is_update_request( $request ) ) {
 			$this->cart_updated( $request );
 		}
```
wavvves: Just a really tiny nitpick, we should do like before and return early on wp_error and ditch the elseif. Having the success flow on the function's main branch will never require else/ifelse, and those can always be avoided.
```
if ( is_wp_error( $response ) ) {
    return $this->error_to_response( $response );
}

if ( $this->is_update_request( $request ) ) {
   $this->cart_updated( $request );
}
```

src/StoreApi/Authentication.php
```diff
@@ -55,4 +93,18 @@ protected function is_request_to_store_api() {
...
+	/**
+	 * Get current user IP Address.
+	 *
+	 * Note for HTTP_X_FORWARDED_FOR, Proxy servers can send through this header like this: X-Forwarded-For: client1, proxy1, proxy2.
+	 * Make sure we always only send through the first IP in the list which should always be the client IP.
+	 *
+	 * @return string
+	 */
+	protected static function get_ip_address() {
+		$ip_address = trim( current( explode( ',', sanitize_text_field( wp_unslash( $_SERVER['HTTP_X_REAL_IP'] ?? $_SERVER['HTTP_X_FORWARDED_FOR'] ?? $_SERVER['REMOTE_ADDR'] ?? '' ) ) ) ) );
```
wavvves: We might bump into some problems here and break functionality for some users. `$_SERVER['HTTP_X_REAL_IP']` and `$_SERVER['HTTP_X_FORWARDED_FOR']` can very easily get spoofed as they mirror the respective request headers, and `$_SERVER['REMOTE_ADDR']` will report wrongly if behind some proxy, or some caching system like Varnish or Cloudflare. This means a real attacker will bypass this very easily, and a real user might get blocked easily too.
`$_SERVER['REMOTE_ADDR']` can also be spoofed if the attacker would just want to abuse a certain endpoint, not caring about getting a response.

Also, I am not sure returning a default value (empty string here) would be the way to go, as it would group all users in that condition (well they were up to no good probably). I wonder if it wouldn't make sense to error out the request.

All this being said, one actionable thing we can do, is to expand this function, remove the confusing one-liner and increase readability as there is a fair chance someone has to tweak this in the future.

One more thing, do we have data on how many users resort to Cloudflare? Because if I remember correctly, Cloudflare passes the proxies through some specific headers, if it justifies, we should address that case here too.

wavvves (commented):
Looks good to me, apart from a tiny nitpick, I left comments about getting the non-logged key. This is something where we can only make life harder for an attacker, as it will probably be always exploitable 🙄

I also agree about the cleanup routine to delete old data.

src/StoreApi/Authentication.php
```diff
@@ -20,15 +22,51 @@ public function init() {
...
-			return true;
+		/**
+		 * Filters the Store API rate limit check.
+		 *
+		 * This can be used to disable the rate limit check when testing API endpoints via a REST API client.
+		 *
+		 * @param boolean $disable_rate_limit_check If true, checks will be disabled.
+		 * @return boolean
+		 */
+		if ( ! apply_filters( 'woocommerce_store_api_disable_rate_limit_check', false ) ) {
```
senadir: Switch this to off by default.

```suggestion
		if ( apply_filters( 'woocommerce_store_api_enable_rate_limit_check', false ) ) {
```

src/StoreApi/Authentication.php
```diff
@@ -20,15 +22,51 @@ public function init() {
...
+			$action_id          = 'store_api_request_' . ( is_user_logged_in() ? get_current_user_id() : md5( $this->get_ip_address() ) );
+			$rate_limit_limit   = 25;
```
senadir: Reduce this to 5
```suggestion
			$rate_limit_limit   = 5;
```

src/StoreApi/Authentication.php
```diff
@@ -20,15 +22,51 @@ public function init() {
...
+					'rate_limit_exceeded',
+					sprintf(
+						'Too many requests. Please wait %d seconds before trying again.',
+						$retry
+					),
+					[
+						'status' => 429,
+					]
+				);
+			}
```
senadir: Add an action here that carries over the IP address being blocked.

src/StoreApi/Authentication.php
```diff
@@ -20,15 +22,51 @@ public function init() {
...
+		 *
+		 * This can be used to disable the rate limit check when testing API endpoints via a REST API client.
+		 *
+		 * @param boolean $disable_rate_limit_check If true, checks will be disabled.
+		 * @return boolean
+		 */
+		if ( ! apply_filters( 'woocommerce_store_api_disable_rate_limit_check', false ) ) {
+			$action_id          = 'store_api_request_' . ( is_user_logged_in() ? get_current_user_id() : md5( $this->get_ip_address() ) );
+			$rate_limit_limit   = 25;
+			$rate_limit_seconds = 10;
```
senadir: Increase this to 60
```suggestion
			$rate_limit_seconds = 60;
```

src/StoreApi/Authentication.php
```diff
@@ -55,4 +93,84 @@ protected function is_request_to_store_api() {
...
+		if ( array_key_exists( 'HTTP_X_FORWARDED_FOR', $_SERVER ) ) {
+			$ips = explode( ',', sanitize_text_field( wp_unslash( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) );
+			if ( is_array( $ips ) && ! empty( $ips ) ) {
+				return self::validate_ip( trim( $ips[0] ) );
+			}
+		}
+
+		if ( array_key_exists( 'HTTP_X_FORWARDED', $_SERVER ) ) {
+			preg_match( // using regex instead of explode() for a smaller code footprint.
+				'/(?<=for\=)[^;,]*/i', // We catch everything on the first "for" entry, and validate later.
```
senadir, line 140: Maybe a comment here about what we're trying to match against.

src/StoreApi/Authentication.php
```diff
@@ -1,6 +1,8 @@
 <?php
```
senadir, line 1: Let's add a doc page to `src/StoreApi/docs` explaining how this works, the filters in action, and the logic we have.

src/StoreApi/Authentication.php
wavvves: By that, you mean a `do_action()` from where someone can tap into with `add_action()` and perform logic with the IP (eg logging etc) right? See [changes](https://github.com/woocommerce/woocommerce-blocks/pull/5962/commits/9588864b7d1349077c0654c088419520a5541c23)

src/Installer.php
wavvves: Resolving this, as section was deleted.

src/StoreApi/Authentication.php
senadir: yes, probably rename it to `woocommerce_store_api_rate_limit_exceeded`.

src/StoreApi/Authentication.php
wavvves: [Doc](https://github.com/woocommerce/woocommerce-blocks/pull/5962/commits/a8bd4fc22e3884b325571739829a116933b249ce)

src/StoreApi/Authentication.php
```diff
@@ -55,4 +109,98 @@ protected function is_request_to_store_api() {
...
+		if ( array_key_exists( 'HTTP_X_FORWARDED_FOR', $_SERVER ) ) {
+			$ips = explode( ',', sanitize_text_field( wp_unslash( $_SERVER['HTTP_X_FORWARDED_FOR'] ) ) );
+			if ( is_array( $ips ) && ! empty( $ips ) ) {
+				return self::validate_ip( trim( $ips[0] ) );
+			}
+		}
+
+		if ( array_key_exists( 'HTTP_X_FORWARDED', $_SERVER ) ) {
+			// Using regex instead of explode() for a smaller code footprint.
+			// Expected format: Forwarded: for=192.0.2.60;proto=http;by=203.0.113.43,for="[2001:db8:cafe::17]:4711"...
```
opr, line 138: Sorry to be picky, but I am wondering if the expected format should be `X-Forwarded-For: ...` rather than `Forwarded: ...` since we're examining `HTTP_X_FORWARDED`?

src/StoreApi/Authentication.php
```diff
@@ -20,15 +22,67 @@ public function init() {
...
+		 */
+		if ( apply_filters( 'woocommerce_store_api_enable_rate_limit_check', false ) ) {
+			$action_id = 'store_api_request_';
+
+			$ip_address = self::get_ip_address();
+
+			if ( is_user_logged_in() ) {
+				$action_id .= get_current_user_id();
+			} else {
+				if ( ! $ip_address ) {
```
opr: Do we only care about this if the user is not logged in? What if the user is logged in but `$ip_address` is `false`.

src/StoreApi/Authentication.php
```diff
@@ -20,15 +22,67 @@ public function init() {
...
+			$retry  = RateLimits::is_exceeded_retry_after( $action_id );
+			$server = rest_get_server();
+			$server->send_header( 'RateLimit-Limit', RateLimits::LIMIT );
+
+			if ( false !== $retry ) {
+				$server->send_header( 'RateLimit-Retry-After', $retry );
+				$server->send_header( 'RateLimit-Remaining', 0 );
+				$server->send_header( 'RateLimit-Reset', time() + $retry );
+
+				if ( $ip_address ) {
```
opr: See my comment above, I'm wondering if this could be refactored so we don't need to check `$ip_address` again here, since we're checking it above if the user is not logged in. 

If the user is not logged in, and `$ip_address` is `false` this code is unreachable.

src/StoreApi/Authentication.php
```diff
@@ -55,4 +109,98 @@ protected function is_request_to_store_api() {
...
+	 * Forwarded request header contains information that may be added by reverse proxy servers (load balancers, CDNs, and so on).
+	 * Documentation at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded
+	 * Full RFC at https://datatracker.ietf.org/doc/html/rfc7239
+	 *
+	 * @return string|false
+	 */
+	protected static function get_ip_address() {
+
+		/**
+		 * Filters the Store API rate limit check proxy support, which is disabled by default.
```
opr: This is a little tough to read imo. Maybe consider: `Filters whether proxy support is enabled for the Store API rate limit check. This is disabled by default.`

src/StoreApi/Authentication.php
```diff
@@ -55,4 +109,98 @@ protected function is_request_to_store_api() {
...
+		 *
+		 * If the store is behind a proxy, load balancer, CDN etc, the user can enable this for proper obtaining
+		 * of client's ip address through standard transport headers.
```
opr: Just a suggestion to improve the wording here.

```suggestion
		 * If the store is behind a proxy, load balancer, CDN etc. the user can enable this to properly obtain
		 * the client's IP address through standard transport headers.
```

src/StoreApi/Authentication.php
```diff
@@ -55,4 +109,98 @@ protected function is_request_to_store_api() {
...
+					'/(?<=\[).*(?=\])/i', // We catch only the ipv6 and overwrite $matches.
+					$matches[0],
+					$matches
+				);
+			}
+
+			if ( ! empty( $matches ) ) {
+				return self::validate_ip( trim( $matches[0] ) );
+			}
+		}
```
opr, line 156: Should we add some unit tests to verify this is working correctly? Maybe unit tests for the whole `get_ip_address` function would be beneficial here?

src/StoreApi/Authentication.php
```diff
@@ -55,4 +109,98 @@ protected function is_request_to_store_api() {
...
+	}
+
+	/**
+	 * Uses filter_var() to validate and return ipv4 and ipv6 addresses
+	 *
+	 * @param string $ip ipv4 or ipv6 ip string.
+	 *
+	 * @return string|bool
+	 */
+	protected static function validate_ip( $ip ) {
```
opr, line 169: could this be simplified to just return the result of the `filter_var` call?

opr (commented):
Hi this looks good so far! I've left a few in line comments.

Also, I'm wondering if we could improve the error message returned by the server, when I try updating my address I see this, which could frustrate or confuse a user, since they won't know why they can't proceed.

image

src/StoreApi/Authentication.php
wavvves: Thanks a lot for reviewing this @opr 😎.

So we need to tie the requests to a unique key, it will either be a hashed version of IP or User ID depending if the user is logged in or not. To defeat this while using the same IP an attacker would have to use a lot of accounts which would take some effort as well as defeating the IP key, hence why we don't care for IP if the user is logged in.

src/StoreApi/Authentication.php
wavvves: I moved the get IP out because it was getting called multiple times. This code will be unreachable in those not because of the conditions but because every request will return an error if the user is not logged in and the `$ip_address` is false.

But your suggestion makes sense since we can remove the possibly unnecessary `get_ip_address()` call by altering those if conditions and using the first one to call and check at the same time, and the second one to check&call/check if it wasn't obtained previously. See this [commit](https://github.com/woocommerce/woocommerce-blocks/pull/5962/commits/826669279bf0bd00a8606e95ccc719dfb09845da).

EDIT: Our linter does not allow assigning vars inside conditions checks, moved those outside.

src/StoreApi/Authentication.php
wavvves: 👍🏼 done.

src/StoreApi/Authentication.php
wavvves: yeah, it can! good catch! Changed

src/StoreApi/Authentication.php
wavvves: Oh, this is so wrong. This is not being picky 😂, this was a major mess-up on my side 🤐! Even the check and the header used inside don't match. I should be using HTTP_FORWARDED on both! Thanks a lot for catching this! Fixed.

wavvves:
> Hi this looks good so far! I've left a few in line comments.
> 
> Also, I'm wondering if we could improve the error message returned by the server, when I try updating my address I see this, which could frustrate or confuse a user, since they won't know why they can't proceed.
> 
> image

This is related to our discussion on how client-side code handles error responses, it seems we can only handle batch error responses (since we're abusing batch). Maybe we should improve on that on a separate issue, should be a quick fix to unblock this and would make our client-side code more resilient to errors. Do you want me to open an issue right away, or talk a bit about this before opening?

src/StoreApi/Authentication.php
wavvves: Good point, [added](https://github.com/woocommerce/woocommerce-blocks/pull/5962/commits/2f1976e253a7818ef988810b7282f58bd8975a58) 😎

opr:
> Do you want me to open an issue right away

Actually, I think we can just change this line: https://github.com/woocommerce/woocommerce-blocks/blob/ddf75cfb87b780b4cde9add83381c3b8a0f4a09d/assets/js/data/shared-controls.ts#L191 to include `errorResponse` in the `reject`. I tried it on my site and it works.

image

Maybe you could explore this solution and make sure it doesn't cause any other problems?

tests/php/StoreApi/RateLimitsTests.php
```diff
@@ -0,0 +1,135 @@
...
+			array(
+				'REMOTE_ADDR'          => '76.45.67.100',
+				'HTTP_X_REAL_IP'       => '76.45.67.101',
+				'HTTP_CLIENT_IP'       => '76.45.67.102',
+				'HTTP_X_FORWARDED_FOR' => '76.45.67.103,2001:db8:85a3:8d3:1319:8a2e:370:7348,150.172.238.178',
+				'HTTP_FORWARDED'       => 'for="[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:4711";proto=http;by=203.0.113.43,for=192.0.2.60;proto=https;by=203.0.113.43',
+			)
+		);
+
+		$authentication = new ReflectionClass( Authentication::class );
```
opr, line 110: Is there a specific reason we're using this `ReflectionClass`, rather than just instantiating `Authentication` and saving the result of `get_ip_address`.

The reason I am asking is because this isn't done anywhere else in our codebase, and this way of testing methods is new to me.

wavvves: I'm using `ReflectionClass` because it is a protected method which means I cannot call it from the outside for the purpose of unit testing. So the standard way is using that and marking it as accessible. PHPUnit has a way of doing this, but I find it less readable for someone understanding what is happening.

opr: Ok sounds good!

wavvves:
> Actually, I think we can just change this line:
> 
> https://github.com/woocommerce/woocommerce-blocks/blob/ddf75cfb87b780b4cde9add83381c3b8a0f4a09d/assets/js/data/shared-controls.ts#L191
> 
> to include `errorResponse` in the `reject`. I tried it on my site and it works.

[Added](https://github.com/woocommerce/woocommerce-blocks/pull/5962/commits/170789426b0bf527c6d237b789d03161e36d27d0), abused cart&checkout in every way I could think of and didn't find any issue so far. What a sleek suggestion @opr ✨

src/StoreApi/Utilities/RateLimits.php
```diff
@@ -0,0 +1,141 @@
...
+	 * @var int
+	 */
+	const LIMIT = 25;
+
+	/**
+	 * Time in seconds before rate limits are reset.
+	 *
+	 * @var int
+	 */
+	const SECONDS = 10;
```
opr: Should these be filtered so the merchant can modify them at will?

wavvves: Might be interesting, since this is not a feature for the average merchant, that might want more control over it.
@senadir what do you think about @opr 's suggestion?

albarin (approved):
Hey, all looking good to me! Great work! 💪

src/StoreApi/Utilities/RateLimits.php
wavvves: Ok, so these will be filtered and condensed into having only one filter for options (proxy support, limits..) 😎

opr: I don't understand, sorry! Are you suggesting we'd have only one filter to control three different variables? 🤔

senadir: Yes, instead of individual filter for each value, a single filter that for all of them should work.

src/StoreApi/Authentication.php
```diff
@@ -55,4 +104,84 @@ protected function is_request_to_store_api() {
...
+	 * Forwarded request header contains information that may be added by reverse proxy servers (load balancers, CDNs, and so on).
+	 * Documentation at https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded
+	 * Full RFC at https://datatracker.ietf.org/doc/html/rfc7239
+	 *
+	 * @param boolean $proxy_support Enables/disables proxy support.
+	 * @return string|false
+	 */
+	protected static function get_ip_address( $proxy_support = false ) {
+
+		if ( ! $proxy_support ) {
```
mikejolley, line 117: Question; Why is proxy support optional? Is there something about it that may break in certain environments? Just wondering why it wouldn't be on by default.

src/StoreApi/Authentication.php
```diff
@@ -55,4 +104,84 @@ protected function is_request_to_store_api() {
...
+			if ( array_key_exists( 'REMOTE_ADDR', $_SERVER ) ) {
+				return self::validate_ip( sanitize_text_field( wp_unslash( $_SERVER['REMOTE_ADDR'] ) ) );
```
mikejolley: If validation fails, do we want to return at this point, or continue to run through the other headers below?

src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -92,11 +92,11 @@ public function get_args() {
 							],
 						],
 					],
-					$this->schema->get_endpoint_args_for_item_schema( \WP_REST_Server::CREATABLE )
```
mikejolley: What is this change for?

src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -92,11 +92,11 @@ public function get_args() {
...
+					$this->schema->get_endpoint_args_for_item_schema()
 				),
 			],
 			'schema'      => [ $this->schema, 'get_public_item_schema' ],
-			'allow_batch' => [ 'v1' => true ],
```
mikejolley: Is this a breaking change? Can we implement rate limiting for batch requests?

src/StoreApi/docs/rate-limiting.md
```diff
@@ -0,0 +1,90 @@
+# Rate Limiting for Store API endpoints  <!-- omit in toc -->
```
mikejolley, line 1: 👍🏻

assets/js/data/shared-controls.ts
```diff
@@ -188,7 +188,7 @@ export const controls = {
 						if ( errorResponse.body ) {
 							reject( errorResponse.body );
 						} else {
-							reject();
+							reject( errorResponse );
```
mikejolley, line 191: What did this fix if you don't mind me asking?

src/StoreApi/Authentication.php
```diff
@@ -20,15 +22,62 @@ public function init() {
...
+				$action_id .= get_current_user_id();
+			} else {
+				$ip_address = self::get_ip_address( $rate_limiting_options->proxy_support );
+
+				if ( ! $ip_address ) {
+					return new \WP_Error(
+						'ip_address_cannot_be_determined',
+						'Bad request. Client IP address cannot be determined.',
+						array( 'status' => 400 )
+					);
```
mikejolley: Just throwing ideas here, but I was thinking if we treat cases with no IP address as a group... so any request that comes from a logged-out user where the IP cannot be determined for any reason shares the same rate limiting, we could allow more requests through—that would unblock add to cart requests somewhat.

Additionally, I wonder if in preference to IP address, could we use the cart session ID/customer ID to identify users making requests?

So in order, we'd use User ID, Customer Session ID, Customer IP Address.

cc @opr any thoughts on this too?

src/StoreApi/Authentication.php
```diff
@@ -20,15 +22,62 @@ public function init() {
...
+			$server->send_header( 'RateLimit-Limit', $rate_limiting_options->limit );
+
+			if ( false !== $retry ) {
+				$server->send_header( 'RateLimit-Retry-After', $retry );
+				$server->send_header( 'RateLimit-Remaining', 0 );
+				$server->send_header( 'RateLimit-Reset', time() + $retry );
+
+				$ip_address = $ip_address ?? self::get_ip_address( $rate_limiting_options->proxy_support );
+				if ( $ip_address ) {
+					do_action( 'woocommerce_store_api_rate_limit_exceeded', $ip_address );
```
mikejolley: Why don't we fire this regardless of IP?

mikejolley (commented):
Thanks for taking over this. I added some questions and suggestions to make this more robust.

src/StoreApi/Authentication.php
wavvves: If proxy support is ON by default, activating Rate Limiting without being behind another machine will leave this vulnerable to IP spoofing. An attacker could just make requests providing different values for IP address carrying headers.

src/StoreApi/Authentication.php
wavvves: `REMOTE_ADDR` will always be set if we're not using CLI or the merchant has some odd server conf. Still, this makes sense, it is a loose end.

Maybe the perfect place to implement your suggestion of grouping requests where we cannot determine IP.

src/StoreApi/Routes/V1/Checkout.php
wavvves: I don't remember exactly. Most definitely a mistake since it does not relate to the scope of this task. I'll revert it, thanks for catching it.
Btw, it was redundant since it is the default value, and we already state the method type on line 74, guess it got flagged by the IDE.

src/StoreApi/Routes/V1/Checkout.php
wavvves: We don't use batch for Checkout, but clearly, I wasn't thinking straight disabling this, as others might. Reverting it.

I discussed this with Nadir a while back, releasing rate limiting in steps, first for the main request, gathering some feedback, and then with batch support. It is a scary feature when generalized to work in the wild, and it can break the most sensitive action: actual purchase.

src/StoreApi/Authentication.php
wavvves: Good call@mikejolley, the action should always be triggered regardless it is possible or not to resolve an IP to go with it :high-five:
Removing the `if`.

assets/js/data/shared-controls.ts
wavvves: [This](https://github.com/woocommerce/woocommerce-blocks/pull/5962#pullrequestreview-1037191967) was an issue raised by @opr , the next two comments are related.

src/StoreApi/Authentication.php
senadir: In my understanding, we're rejecting requests without an IP, I think it's safe to assume all requests will have an IP unless explicitly malicious.

I very much like the User ID, Session ID, IP flow, but it can be followed up with.

src/StoreApi/Authentication.php
senadir: We can just assign 0.0.0.0 for requests without an IP

src/StoreApi/docs/rate-limiting.md
```diff
@@ -0,0 +1,90 @@
...
+- [Proxy standard support](#proxy-standard-support)
+- [Limit usage information observability](#limit-usage-information-observability)
+    - [Response headers example](#response-headers-example)
+- [Tracking limit abuses](#tracking-limit-abuses)
+    - [Custom tracking usage example](#custom-tracking-usage-example)
+
+[Rate Limiting](https://github.com/woocommerce/woocommerce-blocks/pull/5962) can now be enabled for Store API endpoints.
+This was designed to prevent abuse of endpoints from excessive calls and performance degradation on the machine running the store.
+
+It is unauthenticated, rate limits are keyed by either `USER ID` (logged in) or `IP ADDRESS` (guest user), and standard support for running behind a proxy, load balancer, etc. for unauthenticated users can be enabled.
```
senadir: I'm struggling to read this line.

src/StoreApi/docs/rate-limiting.md
```diff
@@ -0,0 +1,90 @@
...
+- [Limit information](#limit-information)
+- [Methods restricted by Rate Limiting](#methods-restricted-by-rate-limiting)
+- [Rate Limiting options filter](#rate-limiting-options-filter)
+- [Proxy standard support](#proxy-standard-support)
+- [Limit usage information observability](#limit-usage-information-observability)
+    - [Response headers example](#response-headers-example)
+- [Tracking limit abuses](#tracking-limit-abuses)
+    - [Custom tracking usage example](#custom-tracking-usage-example)
+
+[Rate Limiting](https://github.com/woocommerce/woocommerce-blocks/pull/5962) can now be enabled for Store API endpoints.
```
senadir: We should probably rewrite this in present tense, "can now be enabled for Store API endpoints." feels more like an announcement and not a doc page.

src/StoreApi/Authentication.php
opr: > I think it's safe to assume all requests will have an IP unless explicitly malicious.

I also think this. If a request somehow doesn't have an IP, then there could be other issues further down the line, e.g. processing payment might not work, for example Stripe logs the IP of the customer when paying. I don't know how to spoof not having an IP so can't test it, but I don't imagine Stripe would allow it to go through? 

> So in order, we'd use User ID, Customer Session ID, Customer IP Address.

I like this idea!

src/StoreApi/docs/rate-limiting.md
senadir: We should mention that it's opt in by default.

senadir (approved):
:shipit:

src/StoreApi/Authentication.php
wavvves: Can't someone delete cookies and reset the session with each request?
We already use User ID if available as the $action_id, before resorting to resolving an IP address 👍🏼

senadir: good point, a session id wouldn't catch multiple attacks, it would just slightly change an attacker's flow but wouldn't block him.

wavvves: Marking this as resolved then, User Id -> IP Address is our order preference for tracking rate limits 🙂

src/StoreApi/Authentication.php
wavvves: I [added a return false](https://github.com/woocommerce/woocommerce-blocks/pull/5962/commits/7772169e3ee3c2bc70c5ca53e2b5df363c64eaaf) for Mike's comment.

Currently, we return an error response if the IP can't be resolved, my line of thought was that something shady should happen if this is the case.

This is currently the only blocker merging this, @mikejolley @senadir do you think grouping those cases to share the same limits instead of returning an error response would be the best solution?

If so I'll make the change, if not things will be merged this way 👍🏼

mikejolley: I vote to group them—we're only really trying to rate limit here, not prevent users hitting the endpoints without IPs, so if we use the 0.0.0.0 fallback @senadir suggested that will be fine. 

Happy for this to be in a follow up if you want to merge this while it has approval.

wavvves: [Done](https://github.com/woocommerce/woocommerce-blocks/pull/5962/commits/dbe8c687afd1b03a55199ed06e33cac867a3f4fc) 👍🏼

senadir (approved):
Thank you for the new changes around caching Paulo!


I also like grouping bad requests for 0.0.0.0, while this is unlikely to happen, it's always better to allow them and rate limit them then to completely deny them access.

:shipit: away

tobinfekkes:
Thank you all for this update supporting rate-limiting. A few stores have been getting hammered with fraud orders on the StoreAPI endpoints after switching from the legacy checkout shortcodes. This feature is a breath of fresh air.

So far, it's been working as expected. Thanks again.

PR #5965, opened 2022-03-01
Remove lodash forEach usage

mikejolley:
Got fed up seeing:

Screenshot 2022-03-01 at 11 02 54

Removed lodash `forEach` to resolve it.

### Testing

1. Insert products by category block
2. Check console for errors once the search list component has rendered

### DIFF
--- a/assets/js/editor-components/search-list-control/utils.tsx
+++ b/assets/js/editor-components/search-list-control/utils.tsx
@@ -1,7 +1,7 @@
 /**
  * External dependencies
  */
-import { groupBy, keyBy, forEach } from 'lodash';
+import { groupBy, keyBy } from 'lodash';
 import { __, _n, sprintf } from '@wordpress/i18n';
 import { Fragment } from '@wordpress/element';
 
@@ -77,7 +77,7 @@ export const buildTermsTree = (
 	delete termsByParent[ '0' ];
 
 	// anything left in termsByParent has no visible parent
-	forEach( termsByParent, ( terms ) => {
+	Object.values( termsByParent ).forEach( ( terms ) => {
 		tree.push( ...fillWithChildren( terms || [] ) );
 	} );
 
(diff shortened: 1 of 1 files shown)


assets/js/editor-components/search-list-control/utils.tsx
```diff
@@ -77,7 +77,7 @@ export const buildTermsTree = (
 	delete termsByParent[ '0' ];
 
 	// anything left in termsByParent has no visible parent
-	forEach( termsByParent, ( terms ) => {
+	Object.values( termsByParent ).forEach( ( terms ) => {
```
tomasztunik: Had a quick look at this one before and unfortunately the forEach from lodash works a bit different than  native array forEach on `Object.values`. This will have unit tests failing.

This is because we rely here on the fact that lodash operates on a mutable `termsByParent` object that changes as `fillWithChildren` is called and when we use `Object.values( termsByParent ).forEach` it ignores mutation of `termsByParent` and iterates over sealed a new enumerable it gets from Object.values. 

This causes duplicate/false values to be added to the tree.

This will require rewrite of the whole method :(

mikejolley: O I didn't see duplication testing locally. Ill take another look thanks.

mikejolley: I pushed a change which made tests pass. Does that code make more sense to you?

tomasztunik: looks good! I think it's clearer what it does and how. Was super surprised when I've noticed that it relies on the collection being mutated. Had a look at how it's implemented in lodash and... was quite a trip, also they didn't aim for "simple" :)

tomasztunik (approved):
thanks for cleaning up our PRs 🙆‍♂️!

PR #5967, opened 2022-03-01
Remove the ToggleButtonControl in favor of ToggleGroupControl

### LINKED ISSUE #5873: Replace ToggleButtonControl with ToggleGroupControl
WC Blocks has a custom component named [`ToggleButtonControl`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/assets/js/editor-components/toggle-button-control) which generates this kind of UI:

ToggleButtonControl screenshot

That's very similar to Gutenberg's [`ToggleGroupControl`](https://github.com/WordPress/gutenberg/tree/trunk/packages/components/src/toggle-group-control):

ToggleGroupControl screenshot

This issue is about completely replacing all instances of `ToggleButtonControl` with `ToggleGroupControl`. This way we will more closely follow Gutenberg UI's and we won't need to maintain this specific component.

danieldudzic:
This PR replaces the `ToggleButtonControl` Product Blocks component in favor of the core `ToggleGroupControl`.

Fixes #5873 

### Screenshots
|Before|After|
|-|-|
|OLD|new|

### Manual Testing

2. Activate a **block theme**, like Twenty Twenty Two
3. Create a new page, and add the following blocks:
- Active Filters
- Attribute Filter
- Price Filter
- Product Categories
- Reviews
4. Ensure that options are using the new component and are working correctly

### User Facing Testing

- [x] Same as above

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests


gigitux:
Thanks for working on this! LGTM! 

Are we migrating to experimental components, are we sure that it is safe? @Aljullu 

If we agree that it is safe, the thing to do is fix ESLint errors and after that, we can merge this PR 🥳

Aljullu (commented):
LGTM as well, good job @danieldudzic!

> Are we migrating to experimental components, are we sure that it is safe? @Aljullu

That's a good point, thanks for raising it. I do think that in general we should avoid using experimental features in WC core. However, in this specific case, I think we can do an exception given that it seems a low-risk change and we can revert back if needed.

However, one thing that we should make sure to test is whether this works with WP 5.7, which is the oldest version supported by WC core. (Currently, it's set to [5.6](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/woocommerce.php#L11), but I think that needs to be updated)

Aljullu:
> (Currently, it's set to [5.6](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/woocommerce.php#L11), but I think that needs to be updated)

Just found that there is a PR already prepared to update it: https://github.com/woocommerce/woocommerce/pull/32013 :slightly_smiling_face:

gigitux:
Thanks for updating the PR! It seems that there are other eslint errors. Could you check?

If you want, you can try locally with:
```
npm run lint:js
```

gigitux (approved):
Thanks for working on this and fix the linter errors. Great job! LGTM :shipit:

tomasztunik:
@danieldudzic is anything blocking this from being merged?

PR #5973, opened 2022-03-01
Fix sort-select frontend CSS

### LINKED ISSUE #2506: All Products: <select> looks different in the editor vs frontend
Several blocks, like _All Products_, have native `<select>` elements that inherit some styles in the editor (via the selector `.wp-core-ui select`). Those styles are not loaded in the frontend so they look different editor vs frontend.

**Screenshots**
_Editor:_
imatge

_Frontend:_
imatge

_Note: this issue is not about the `<select>` being disabled (see #2487), but about it having different padding, background, border, etc._

tarunvijwani:
This PR fixes the CSS for the `<select>`  element of the `All Products` block to provide consistent style across the frontend and page editor.

Fixes #2506 

### Screenshots

| Before | After |
|---|---|
| image | image |

### Manual Testing

2. Create a page with the `All Products` block. 
3. Publish the page. 
4. The frontend `<select>` dropdown style should match the page editor's `<select>` dropdown style.

### User Facing Testing
* [x] Same as above,

### Changelog

> Fixed sort-select frontend CSS

### DIFF
--- a/assets/js/base/components/sort-select/style.scss
+++ b/assets/js/base/components/sort-select/style.scss
@@ -11,4 +11,14 @@
 .wc-block-components-sort-select__select {
 	font-size: inherit;
 	width: max-content;
+	border-color: #8c8f94;
+	-webkit-appearance: none;
+	-moz-appearance: none;
+	border-radius: 3px;
+	padding: 0 24px 0 8px;
+	min-height: 30px;
+	max-width: 25rem;
+	font-weight: normal;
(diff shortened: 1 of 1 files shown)


Aljullu (commented):
Thanks for working on this @tarunvijwani! Confirming the select looks the same between the editor and the frontend with this PR! However, I'm not 100% sure that we should be adding these new CSS declarations.

The issue seems to be that some wp-admin styles are leaking into the block. That's true in the page/post editor in desktop view, but if you switch to tablet or mobile preview, that no longer happens. The issue can't be reproduced either in the Site Editor. That's because Site editor, tablet and mobile previews have the editor iframed (so styles from wp-admin don't leak into the editor).

The effort to completely iframe the page/post editor can be followed here: https://github.com/WordPress/gutenberg/issues/20797.

So, thinking about it a bit more, I consider we shouldn't fix #2506 on our end. Instead, we should wait for the editor to be iframed everywhere (https://github.com/WordPress/gutenberg/issues/20797), that will automatically solve our issue without us needing to change the CSS in our end.

Thoughts @tarunvijwani? Sorry for not keeping that issue up to date. :pray:

tarunvijwani:
@Aljullu No worries! Thank you for sharing the details! I agree with you that we should wait for [#20797](https://github.com/WordPress/gutenberg/issues/20797) to get resolved and revisit  #2506 to check if the issue still persists.

PR #5974, opened 2022-03-01
Remove not used attributes

### LINKED ISSUE #5952: Check attributes for each block that now implement Global Styles
@tomasztunik noticed that after the migration of the Global Styles, some attributes (related to the style) are not necessary anymore because Global Styles handles all automatically.

We should review all the blocks and remove the attributes not necessary anymore.

gigitux:
After the migration to the Global Styles, some attributes (related to the style) are not necessary anymore because Global Styles handles all automatically. This PR clean up the attributes that are not necessary anymore.

Fixes #5952 

### Manual Testing

Nothing to test. This PR doesn't add/remove features.


tomasztunik:
As we were looking into 404 errors generated by atomic components I've noticed that we often use `withFilteredAttributes` HOC that ensures only valid props are passed to the component. I'm afraid it is not aware of the attributes that are added automatically when using supports features. Will try to verify that tomorrow and ensure it doesn't break any blocks that happened to use it

tomasztunik (approved):
👌

PR #5979, opened 2022-03-02
Shopper → Mini Cart → Can see correct tax

### LINKED ISSUE #5833: Critical flows: Shopper → Mini Cart → Can see correct tax
### Note

Create e2e tests to verify that a shopper can see the correct tax for products in cart.

### File

`tests/e2e/specs/shopper/mini-cart.test.js`

Part of #5814

dinhtungdu:
Fixes #5833

### Notes

This PR adds e2e tests to verify that a shopper can see the correct tax for products in cart.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/mini-cart.test.js -t "Tax included"` to run only this test case.


tests/e2e/page-utils.js
```diff
@@ -21,20 +21,37 @@ export async function elementExists( selector, root = page ) {
...
  * @return {Promise<string[]>} An array of text contained in those selected elements
  */
-export function getTextContent( selector, root = page ) {
+export function getTextContentAll( selector, root = page ) {
 	return root.$$eval( selector, ( $elements ) => {
 		return $elements.map(
 			( $element ) => $element.value || $element.textContent
 		);
 	} );
 }
```
dinhtungdu: @sunyatasattva it's common to get the text content from a single element, so I think it's handier to have two versions of this function, one for multiple elements, and one for a single element. What do you think?

sunyatasattva: I think a nice trick you can do to avoid having these two very similar functions is the following, change this:

```js
const priceInCart = await getTextContent(
  '.wc-block-cart-item__prices'
);
```

to this:

```js
const [ priceInCart ] = await getTextContent(
  '.wc-block-cart-item__prices'
);
```

Very readable, you avoid the ugly `( await getTextContent( '.wc-block-cart-item__prices' ) )[0]`, and we don't have to have two utility functions which do almost the same thing.

What do you think?

dinhtungdu: @sunyatasattva It's a great trick! I updated my PR and added some examples to `getTextContent` including your destructuring trick above. Can you please take another look?

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -26,6 +28,19 @@ const clickMiniCartButton = async () => {
 	await page.click( '.wc-block-mini-cart__button' );
 };
 
+const WooCommerce = new WooCommerceRestApi( {
+	url: `${ process.env.WORDPRESS_BASE_URL }/`,
+	consumerKey: 'consumer_key', // Your consumer key
+	consumerSecret: 'consumer_secret', // Your consumer secret
```
sunyatasattva: Shouldn't these be `env` variables?

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -3,6 +3,8 @@
  */
 import { setDefaultOptions, getDefaultOptions } from 'expect-puppeteer';
 import { SHOP_CART_PAGE, SHOP_CHECKOUT_PAGE } from '@woocommerce/e2e-utils';
+const WooCommerceRestApi = require( '@woocommerce/woocommerce-rest-api' )
```
sunyatasattva: Why not

```suggestion
import WooCommerceRestApi from '@woocommerce/woocommerce-rest-api';
```

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -173,11 +188,9 @@ describe( 'Shopper → Mini Cart', () => {
 			// Get a random product to better replicate human behavior.
 			const product =
 				products[ Math.floor( Math.random() * products.length ) ];
-			const productTitleEl = await product.$(
-				'.wc-block-components-product-name'
-			);
-			const productTitle = await productTitleEl.getProperty(
-				'textContent'
+			const productTitle = await getTextContent(
```
sunyatasattva: Shouldn't this be:

```suggestion
			const [ productTitle ] = await getTextContent(
```

tests/e2e/specs/shopper/mini-cart.test.js
```diff
@@ -221,6 +234,90 @@ describe( 'Shopper → Mini Cart', () => {
...
+			} );
+
+			await page.waitForSelector( '.wc-block-cart-item__prices' );
+			const [ priceInCart ] = await getTextContent(
+				'.wc-block-cart-item__prices'
+			);
+
+			expect( priceInLoop ).not.toMatch( priceInCart );
+
+			await page.mouse.click( 50, 200 );
```
sunyatasattva: What's the purpose of this? To close the mini cart? In general, I'd advise against such low level implementation details in a test file. It might be tempting since a method would be a one liner, but `await miniCart.close()` reads much better than `await page.mouse.click( 50, 200 )` and will also allow us to more easily switch implementation (for example when we'll switch to another library or framework).

sunyatasattva (changes requested):

tests/e2e/specs/shopper/mini-cart.test.js
dinhtungdu: This makes a lot of sense to me. I addressed this in [ff3ece3](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5979/commits/ff3ece3d3702e9e700cf5fcc75952d178ade3701)

tests/e2e/specs/shopper/mini-cart.test.js
dinhtungdu: Fixed in [ff3ece3](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5979/commits/ff3ece3d3702e9e700cf5fcc75952d178ade3701)

tests/e2e/specs/shopper/mini-cart.test.js
dinhtungdu: I added a comment explaining why we use `consumerKey` and `consumerSecret` that way.

dinhtungdu:
@sunyatasattva Thanks so much for the thoughtful review! I addressed your comments, please take another look when you have time : ) !

tests/e2e/specs/shopper/mini-cart.test.js
sunyatasattva: Nice, thank you for clarifying! Perhaps removing those comments (`// Your consumer key`) which seem they might have been copy+pasted from docs and are perhaps a bit confusing would be the last suggestion I've got :)

sunyatasattva (approved):
Great work @dinhtungdu ! And thanks for the patience with the review. I have added one last thing, but I'm already approving the PR :)

sunyatasattva:
Hey @dinhtungdu ! Is this waiting for anything else from me before merging? :)

dinhtungdu:
@sunyatasattva I was updating the branch then I forgot to merge it :D. Thanks for the reminder!

PR #5981, opened 2022-03-02
Remove `fromEntriesPolyfill`

opr:
This PR will remove the `fromEntriesPolyfill` we have - this is no longer required as `Object.fromEntries` is [supported by all modern browsers](https://caniuse.com/mdn-javascript_builtins_object_fromentries), and given we no longer support IE11 there is no need to keep it around.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Ensure nothing breaks when checking out. Change shipping rates several times.
2. Use WooCommerce Subscriptions to check out with multiple shipping packages and ensure you can select different shipping rates without issues.

### User Facing Testing
* [x] Same as above


ralucaStan (approved):
looks good, nice catch. yey for browser compatibility

PR #5982, opened 2022-03-02
StoreAPI: Rename namespace to remove `\Blocks`

### LINKED ISSUE #5948: Update Store API Namespace
Currently the Store API is under the `Automattic/WooCommerce/Blocks/StoreAPI` namespace.

To prepare the codebase for any potential moves to core, we should probably remove the `Blocks` part since technically the API is not part of Blocks at all. 

I think we should update the namespace to `Automattic/WooCommerce/StoreAPI`. We can update the autoloader config in `composer.json` to support this, and handle any changes with `class_alias` functions.

mikejolley:
This PR changes the namespace of StoreAPI classes from `Automattic\WooCommerce\Blocks\StoreApi` to `Automattic\WooCommerce\StoreApi` to better indicate it is separate to Blocks functionality and to make any future migration of codebase more straightforward. 

Fixes #5948

After merge, you'll need to do a `composer dump-autoload`.

There are some follow-ups for later:

- Untangle any `\Blocks` usage from inside Store API and move it within.
- Move bootstrap code from `src/RestAPI.php` inside Store API if applicable.

### Testing

This is only a rename, so issues should be flagged by unit tests if an incorrect namespace is used anywhere.


opr (approved):
Works OK for me, thanks for fixing the issue with Subscriptions.

PR #5985, opened 2022-03-03
Refactor: Use an alternative mini cart icon to fix the Mini Cart styling issues

### LINKED ISSUE #5930: Twenty Twenty Two theme and Mini Cart Block: the number of items is not visible when the Mini Cart is added on the header on the front page
## Describe the bug

On the `Twenty Twenty Two` template, the number of items is not visible when the `Mini Cart` block is added on the header on the front page. This is caused by the white background of the bubble.

## To reproduce

1. Install and active Twenty Twenty Two theme.
2. Open the FSE editor and add the `Mini Cart` block in the header area.
3. Check that the 0 number is not visible due to the white background of the bubble.

## Expected behavior

The counter should be always visible.

## Screenshots

image

dinhtungdu:
Fixes #5930

The current Mini Cart icon leads to some difficulties and style issues:
- We need the background for the counter (badge) => an extra setting which isn't really necessary.
- The theme block doesn't work great with dark themes (at least, out of the box).
  - For a special case - Twenty Twenty Two - its large header page template uses a dark header, while other pages use the normal white header. The Mini Cart doesn't work with it.
  - Increase code complexity: because we need to style an inner element of the Mini Cart block, we need `__experimentalSkipSerialization` to target inner elements for global style. We also need to calculate manually color classes and style inside the block component.

This PR tries to solve all of the above issues by using the icon as proposed [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5930#issuecomment-1054942322). By using that icon, we can:
- Remove the background setting which is useless now.
- Remove the transparent button option, which doesn't make sense anymore without the background setting.
- Remove `__experimentalSkipSerialization`, let Blocks Editor and Site Editor target the Mini Cart block using `.wp-block-woocommerce-mini-cart`.
- Remove color and class manipulation for the counter.
- Icon inherits the color from the parent by default, so it works with both dark and light backgrounds, and at the same time.

#### Accessibility

#### Other Checks

### Screenshots

image

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Add the block to the header.
2. Make sure it works by default with Twenty Twenty Two. On both dark and light headers.
3. Test the predefined color and custom color for block style and global style.
4. Ensure the block style priority is higher than global style.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

## Notes

- Look like the Site Editor is having an issue clearing the global style cache. Changing the global text color for the Mini Cart icon works after load. But it doesn't work if we keep the editor open:
  - Go to Site Editor > Style (Beta) > Blocks Styles > Mini Cart.
  - Change the text color > Save it.
  - Keep the current tab open, in a new tab, go to the site front-end, see the text color has been updated correctly.
  - Switch to the site editor tab, change the text color of the mini cart to another, see it work on the Editor > Save.
  - Switch to the front end tab, reload the editor, see the color still the previous value.


dinhtungdu:
Requested additional reviews from Albert and Luigi because this changed how we style the Mini Cart block.

Aljullu (commented):
I like this approach and it's really cool to see how much code could be removed. :clap: My only concern is that it removes the background color setting. Was that intentional? I think we should try to keep it.

Ie, in `trunk` it's possible to change the styles of the Mini Cart button to something like this:

imatge

dinhtungdu:
@Aljullu I added the background back in [5a11609](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5985/commits/5a11609eb28fdbc76347f607f5e1fe6a92d638e9)

> My only concern is that it removes the background color setting. Was that intentional?

Yes, I removed the background because I don't want to use experimental features and the background setting was initially created for the badge background. But I agree with you that having the ability to set the background for the button is actually useful.

Aljullu (commented):
Works perfectly now, thanks @dinhtungdu! I just have one last question: why do we need to use `__experimentalSelector` and `__experimentalSkipSerialization`? Can't we apply the background styles directly into the wrapper element (`.wc-block-mini-cart`)?

dinhtungdu:
> why do we need to use `__experimentalSelector` and `__experimentalSkipSerialization`? Can't we apply the background styles directly into the wrapper element (`.wc-block-mini-cart`)?

@Aljullu If we apply the style to the wrapper element, this will happen.

image

At glance, we can solve this problem by styling the wrapper as an inline block. But we will have an issue with alignment because we need the wrapper full width (to its container) for the alignment to work.

By default, the Global style target the wrapper element (`.wp-block-woocommerce-mini-cart` in our case). It works fine with text color, but not with background color. That's why we need `__experimentalSelector`.

About `__experimentalSkipSerialization`, we need that flag to prevent Site Editor generate inline style for Mini Cart block, which is injected into the wrapper element.

### Alternative options

Actually, there is another way to solve this problem, to make styling work without experimental features

`<Button>` block is an inline element. It often comes with `<Buttons>` as the wrapper for multiple `<Button>` blocks. The `<Buttons>` block provides alignment for inner blocks.

We can apply the approach of the `<Button>` block to Mini Cart block by removing the alignment and styling it as an inline element. If we need alignment, we can use a `<Row>` block and add `<MiniCart>` as an inner block. By doing so, we don't need experimental flags and we can reduce even more lines of code. But it makes the block harder to use at the same time IMO.

Aljullu (approved):
Oh, I missed that, thanks @dinhtungdu! I was in fact testing it inside a Row block, that's why I missed it. :sweat_smile: 

Let's merge this PR as-is. However, I would like us to try to get rid of those experimental APIs. Otherwise, we can only enable background color controls in the Feature Plugin (but not in WC core), and that would be a missed opportunity. :disappointed: 

> We can apply the approach of the <Button> block to Mini Cart block by removing the alignment and styling it as an inline element. If we need alignment, we can use a <Row> block and add <MiniCart> as an inner block. By doing so, we don't need experimental flags and we can reduce even more lines of code. But it makes the block harder to use at the same time IMO.

I like this solution. How do you feel about creating a PR to experiment with it? Honestly, I think most usages of the Mini Cart block will come once we start adding it to patterns, so we will be able to control where it renders. Another thing to consider is that at some point we would like to make the Mini Cart block available inside the Navigation block, that would remove another instance of needing it to have alignment for itself.

cc'ing @gigitux here too for his thoughts on this.

gigitux:
> Oh, I missed that, thanks @dinhtungdu! I was in fact testing it inside a Row block, that's why I missed it. 😅
> 
> Let's merge this PR as-is. However, I would like us to try to get rid of those experimental APIs. Otherwise, we can only enable background color controls in the Feature Plugin (but not in WC core), and that would be a missed opportunity. 😞
> 
> > We can apply the approach of the  block to Mini Cart block by removing the alignment and styling it as an inline element. If we need alignment, we can use a  block and add  as an inner block. By doing so, we don't need experimental flags and we can reduce even more lines of code. But it makes the block harder to use at the same time IMO.
> 
> I like this solution. How do you feel about creating a PR to experiment with it? Honestly, I think most usages of the Mini Cart block will come once we start adding it to patterns, so we will be able to control where it renders. Another thing to consider is that at some point we would like to make the Mini Cart block available inside the Navigation block, that would remove another instance of needing it to have alignment for itself.
> 
> cc'ing @gigitux here too for his thoughts on this.

I agree with @dinhtungdu's thoughts. I guess, in accordance with our goal, for the next block we should follow the alternative solution. We should avoid using custom style and use the blocks that Gutenberg makes available for positioning as much as possible.
I don't know how difficult this approach is, but if it doesn't take long, could we try it, so we can avoid a future block migration?

dinhtungdu:
> How do you feel about creating a PR to experiment with it?

https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6002 is the attempt to remove the experimental flags and simplify the codebase.

Aljullu (approved):

PR #5986, opened 2022-03-03
Store API: Add dedicated container and make Store API code self-contained

mikejolley:
**This contains changes from https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5982 too so please review that first so this can be rebased.**

This PR ensures that all code within the `src/StoreApi` directory is self-contained and does not reference classes from Blocks. It introduces a new container class specifically for the `StoreApi` which houses the Store API class instances, whilst maintaining backwards compatibility with the old class and namespace for [existing implementations such as this one](https://github.com/Automattic/woocommerce-subscriptions-stripe-billing/blob/6b7da7ce7073140214f922e48da79429df29f3b2/subscriptions-base/includes/class-wc-subscriptions-extend-store-endpoint.php#L12-L50). 

Some functionality only used by Store API was moved. This includes account creation handling, Array Utilities, and the main Store API bootstrap classes.

Part of #5707

#### Automated Tests

Existing Store API tests must pass. There are new tests covering account creation on checkout.

### Manual Testing

1. Smoke test checkout and product blocks.
2. Smoke test direct calls to the API endpoints e.g. `/wc/store/cart`

This PR (and the previous) also impacts extensions. The deprecation notices are expected in Subscriptions, but you should test:

1. Adding a subscription to cart
2. Going through the checkout process
3. Ensuring all subscriptions meta data was added to the order

Notices (with debug on) such as this will be expected until the extension is updated to match:

> Deprecated: __experimental_woocommerce_blocks_checkout_update_order_meta is deprecated since version 6.3.0! Use woocommerce_store_api_checkout_update_order_meta instead. This action was deprecated in WooCommerce Blocks version 6.3.0. Please use woocommerce_store_api_checkout_update_order_meta instead. in /Users/mikejolley/Developer/local/app/public/wp-includes/functions.php on line 5691


opr (approved):
Thanks Mike, this is working fine for me, I tested with WC Subscriptions and with WC Points and Rewards, both were fine. Direct calls to the API worked too.

mikejolley:
Thanks @opr I'll get them merged together :)

PR #5987, opened 2022-03-03
Remove v1 string from Store Keys

mikejolley:
With the versioning change in the API (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5911) `/v1` was appended in bulk to all `wc/store` keys and endpoints. This change was not intended to affect the Store Keys of wp.data stores. This PR reverts that change.

### Testing

1. Smoke test the products block and confirm products are still queried
2. Smoke test cart/checkout
3. Open redux dev tools and confirm the stores are named without a `v1` for example, `wc/store/cart` should be there and have data within.

### DIFF
--- a/assets/js/base/context/hooks/cart/use-store-cart.ts
+++ b/assets/js/base/context/hooks/cart/use-store-cart.ts
@@ -119,7 +119,7 @@ export const defaultCartData: StoreCart = {
 };
 
 /**
- * This is a custom hook that is wired up to the `wc/store/v1/cart` data
+ * This is a custom hook that is wired up to the `wc/store/cart` data
  * store.
  *
  * @param {Object} options                An object declaring the various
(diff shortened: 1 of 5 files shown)


ralucaStan (commented):
@mikejolley Changes look good. I've tested and things work as expected.

Only one thing, there was this v1 in the store when navigating to the shop page. I suppose that is not expected?
Screenshot 2022-03-04 at 13 26 12

mikejolley:
As discussed, this is to be expected due to how collections work. The namespace (with v1) is used to construct the route/endpoint from the API here https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/343f70d7af57f751c327349dec864f9e176d0f70/assets/js/data/schema/resolvers.js#L20

In this case, we want to query a v1 route. So it seems to be correct.

ralucaStan (approved):
Thank you for quickly jumping on this. Approving, the namespace `wc/store/v1` from the Redux toolbar maps the the Rest API endpoint namespace, which should be v1

PR #5989, opened 2022-03-03
Fix 404 and 500 errors in E2E test logs, editor and front-end.

### LINKED ISSUE #5945: All Product Block and atomic blocks: 404 errors
## Describe the bug

When it is loaded the `All Product Block`, there are a lot of 404 errors related to the Atomic Blocks.

I don't understand if there is a regression due to the Global Styles epic. I already resolved a similar issue #5649.

## To reproduce

1. Create a post.
2. Add the `All Product Block`.
3. Save it
4. Open the post on the front end with the dev tools opened.
5. Check that there are a lot of 404 errors related to the Atomic Blocks.

## Expected behavior
There should be no 404 errors.

## Screenshots

image

### LINKED ISSUE #5957: 404 error on product-sale-badge.js
## Describe the bug

When we try to edit any post/page (post.php/post-new.php) or try to access customizer it throws a 404 error for product-sale-badge.js.

It seems like the plugin is trying to load this js file from the following path:

`/wp-content/plugins/woo-gutenberg-products-block/build/product-sale-badge.js?ver=7.0.0
`

## To reproduce

1. Go to Posts > Add New
2. Check the console and we will find the website is trying to load product-sale-badge.js.

## Expected behavior

Not to throw this warning in the console.

-   Core version: 5.9.1
-   WooCommerce version: 6.1
-   WooCommerce Blocks version: 7.0
-   Site language: En
-   Any other plugins installed: N/A

-   OS: macOS
-   Browser chrome
-   Version: 22

tomasztunik:
## Motivation

Running the E2E tests and debugging their failures was becoming unwieldy. Our logs were polluted by large amounts of errors, most of which wouldn't make the tests fail, but were in fact captures of the browser's console logs. This made it hard to identify the actual culprits of a failed test run and to parse the logs output.

This PR aims to clean up all those false negatives, and output cleaner logs.

## Changes

- **Change atomic blocks to not use custom webpack build paths**:

Originally, we had treated atomic blocks as different, since they were only supposed to be registered in case a parent component was consuming them. With the advent of Global Styles, and the move towards FSE where the atomic blocks are going to be used as product elements, this distinction made less sense.

Specifically, we had [outputted the built components as custom-named webpack chunks](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/2777), which meant that classes inheriting from `AbstractBlock` needed to hardcode this custom path to be compatible.

This meant that we previously had to enqueue scripts in the editor to make sure they were loaded correctly (see, e.g., https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5816), but the class handling of the assets were still trying to fetch them by their default path, causing a slew of `404` errors which weren't really impacting the functionality of the application, but were triggering unnecessary network requests and were caught by Jest during our E2E tests, polluting the logs.

Additionally, on the front-end it also caused more `404` errors, [due to blocks being registered normally as a requirement for enabling Global Styles](https://teamkirigami.wordpress.com/2022/01/28/global-style-what-is-and-how-it-works).

A side-effect of this PR is simplifying this configuration, as now default paths work and don't have to be hard-coded.

This change moves build output for atomic blocks from  `build/atomic-block-components/_name_` to `build/product-_name_` reflecting the actual block name and always prefixed with context namespace (product) to avoid possible name clashes.

- **Ensure all atomic blocks are now registered via php**

Currently AddToCart and SKU blocks were missing as the only blocks that did not support global styles yet, this adds basic setup so that we can unify how 

- **Ensure all lazy loaded blocks (using chunks) register their chunk_translations by themselves via php**

Before AllProducts.php took the responsibility for registering all chunks which was easy to miss or not understand implications. Registered chunks in AllProducts were making them available for usage outside of AllProducts as well.

- **Add missing emptytheme footer.html template part**

The theme `emptytheme` was added as the most minimal block theme possible, in order to cleanly test Site Editor related features. As its purpose was supposed to be getting away with the most minimal of themes, the `footer` template part had not originally been included.

This caused 404 error as this template part is always fetched when user navigates in the site template editor.

- **Replaced trashAllPosts with new REST API based deleteAllTemplates util**

Previously, we used a util called `trashAllPosts` which navigated to the post UI and deleted all the posts to tear down any side-effects of template editing tests.

However, with a [recent change introduced in WordPress 5.9.1](https://github.com/WordPress/wordpress-develop/commit/14e20f72b568cfb5a85f57c77eaa538c17a5f302), Core removed the UI for those and that made our tests meet a 500 error.

Using the REST API should also make all the tests faster.

- **Remove deprecated usage of Puppeteer `waitFor`**

Puppeteer has deprecated the use of the [generic `waitFor`](https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagewaitforselectororfunctionortimeout-options-args) method for more specific methods to avoid confusion.

Deprecation warnings were polluting the logs, so we switched to more specific methods (i.e. mostly changed `waitFor` with `waitForTimeout`). 
  

Fixes #5945
Fixes #5957

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

1. Run E2E tests 
2. there should be no 404 or 500 errors in the test logs

### User Facing Testing
* [ ] Same as above, or
* [ ] See steps below.

1. Visit the editor
2. Add Single Page and All Products blocks 
3. There should be no 404 errors in the editor
4. Visit the created page in the front-end
5. There should be no 404 errors in the page


package.json
```diff
@@ -131,7 +131,7 @@
 		"@wordpress/data-controls": "2.2.7",
 		"@wordpress/dependency-extraction-webpack-plugin": "3.2.1",
 		"@wordpress/dom": "3.2.7",
-		"@wordpress/e2e-test-utils": "6.0.1",
+		"@wordpress/e2e-test-utils": "^6.0.2",
```
gigitux: We should lock the dep
```suggestion
		"@wordpress/e2e-test-utils": "6.0.2",
```

package.json
```diff
@@ -182,7 +182,7 @@
 		"patch-package": "6.4.7",
 		"postcss": "8.2.13",
 		"postcss-loader": "4.2.0",
-		"prettier": "npm:wp-prettier@2.0.5",
+		"prettier": "npm:wp-prettier@^2.0.5",
```
gigitux: Same here

```suggestion
		"prettier": "npm:wp-prettier@2.0.5",
```

gigitux (commented):
Thanks for working on this! 💪 

404 errors disappear, but for the tests, I have some errors in my local env, but on CI, everything looks great. I guess it is a problem with my env.

I added two comments about the package.json. My suggestion is to do another `npm install` to generate a new package-lock.json.

src/BlockTypes/AllProducts.php
```diff
@@ -46,18 +46,18 @@ protected function register_block_type_assets() {
...
+				'product-summary',
+				'product-sale-badge',
+				'product-sku',
+				'product-category-list',
+				'product-tag-list',
+				'product-stock-indicator',
+				'product-add-to-cart',
 			]
 		);
 	}
```
tomasztunik: While reviewing this PR we've noticed that changes to the names of the packages unfortunately affected ability to read translation files for the blocks. 

The change to how atomic blocks were registered for translation chunks was controlled via the `register_block_type_assets` method on the `All Products`. After the blocks package chunk name changed from `atomic-block-components/price` pattern to `product-price` pattern the translations are no longer recognised as they were referenced in the .po files via their old chunk name. 

Before this can be merged we need to learn if we can update the translation files easily to make it work or if we need to adopt the whole code to support custom names/paths for atomic blocks family with old names until the translation files can be migrated to new names that would highlight these blocks moving to be first party blocks.

Apart of that now that all atomic blocks have their own php registration we should have them register the translation chunk by themselves rather than have All Blocks control this responsibility.

tomasztunik: Also we should probably cover i18n with E2E - we've caught this one by chance 😨

tomasztunik:
With @gigitux we've wanted to understand more the possible impact of changing the translations and we've learned a little bit more how the process works. Changing the chunk names will cause the translations to stop working as the loading of translation files is based on the md5 checksum of the translated filename path. Once new translation files are generated when deployed to w.org existing translations should be reused and new translation json file will be generated with md5 part reflecting changed block entrypoint filename.

This release will require downloading of new translation files if it doesn't happen automatically :) 

So 🤞 but looks like it should be good.

tomasztunik:
We should verify after realease if we need to do any manual trigger to regenerate translation files for new version

gigitux (approved):
Great job! Thanks for working on this @tomasztunik @sunyatasattva 

> We should verify after realease if we need to do any manual trigger to regenerate translation files for new version

Good point, but from what we discovered, the process should be pretty smooth 🤞

PR #5991, opened 2022-03-03
Mini Cart Block and related blocks: now the additional classes are visible on frontend page

### LINKED ISSUE #5881: Mini Cart block: additional classes don't add on the frontend side

gigitux:
With this PR, the additional classes added on the editor side are now rendered on the frontend side.
This PR fixes the scrollbar issue when there are too many products.

Fixes #5881 

### Manual Testing

Check out this branch:

#### Classname problem

1. Open the FSE editor.
2. Add the Mini Cart block.
3. Get focused on it.
4. On the right sidebar, on the block settings -> advanced section, add custom classes.
5. Save.
6. On the frontend page, inspect the Mini Cart Block element.
7. Check that the custom classes are added.
8. Open the FSE editor.
9. Open Mini Cart template editor.
10. Repeat 3-7 steps for each block in the template ( `Mini Cart Contents`, `Filled Mini Cart Contents`, `Mini Cart Title`, `Mini Cart Items`, `Mini Cart Products Table`, `Mini Cart Footer`,  `Empty Mini Cart Contents`, `Mini Cart Shopping Button`)

#### Scrollbar issue
1. Open the FSE editor.
2. Add the Mini Cart block.
3. Save.
4. On the frontend side, add a lot of products to the cart.
5. Open the Mini Cart Block and check that the scroll works well (check the attached gif to see the potential issue).

Screen Capture on 2022-03-03 at 12-08-40

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/filled-mini-cart-contents-block/frontend.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/filled-mini-cart-contents-block/frontend.tsx
@@ -3,18 +3,22 @@
  */
 import { useStoreCart } from '@woocommerce/base-context/hooks';
 
+type FilledMiniCartContentsBlockProps = {
+	children: JSX.Element;
+	className: string;
+};
+
 const FilledMiniCartContentsBlock = ( {
 	children,
-}: {
-	children: JSX.Element | JSX.Element[];
-} ): JSX.Element | null => {
+	className,
+}: FilledMiniCartContentsBlockProps ): JSX.Element | null => {
 	const { cartItems } = useStoreCart();
 
 	if ( cartItems.length === 0 ) {
 		return null;
 	}
 
-	return <>{ children }</>;
+	return <div className={ className }>{ children }</div>;
 };
 
 export default FilledMiniCartContentsBlock;
--- a/assets/js/base/components/drawer/style.scss
+++ b/assets/js/base/components/drawer/style.scss
@@ -69,7 +69,6 @@ $drawer-width-mobile: 100vw;
 	display: block;
 	height: 100%;
 	left: 100%;
-	overflow: auto;
 	position: fixed;
 	right: 0;
 	top: 0;
--- a/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/empty-mini-cart-contents-block/frontend.tsx
@@ -10,10 +10,12 @@ import { useEffect, useRef } from 'react';
 
 type EmptyMiniCartContentsBlockProps = {
 	children: JSX.Element | JSX.Element[];
+	className: string;
 };
 
 const EmptyMiniCartContentsBlock = ( {
 	children,
+	className,
 }: EmptyMiniCartContentsBlockProps ): JSX.Element | null => {
 	const { cartItems, cartIsLoading } = useStoreCart();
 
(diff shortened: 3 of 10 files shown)


assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/filled-mini-cart-contents-block/frontend.tsx
```diff
@@ -3,18 +3,22 @@
...
+	className,
+}: FilledMiniCartContentsBlockProps ): JSX.Element | null => {
 	const { cartItems } = useStoreCart();
 
 	if ( cartItems.length === 0 ) {
 		return null;
 	}
 
-	return <>{ children }</>;
+	return <div className={ className }>{ children }</div>;
```
Aljullu, line 21: For blocks that didn't have any kind of markup like this one, I wonder if instead of adding support for custom class names, maybe we should just set [`supports: { customClassName: false }`](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-supports/#customclassname). If users want to add custom class names to be able to add styles, they can still add them to the block which render some markup. What do you think?

Aljullu (approved):
LGTM, good job adding support for custom class names in all those blocks. :clap: 

I left one question, but besides that, pre-approving.

assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/filled-mini-cart-contents-block/frontend.tsx
gigitux: Mmmm, good point!


The problem that I see is that the HTML (attached image) generated is different from [the declared template](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/87ccbddfa06bcd5bd0d7fd5398bf201ac6705c55/templates/parts/mini-cart.html#L4). From my experience it seems that every block should have a wrapper.

image

dinhtungdu: > If users want to add custom class names to be able to add styles, they can still add them to the block which render some markup.

This approach sounds good to me too. Filled and Empty Mini Cart Contents blocks are always inside the Mini Cart Contents block. Users can add custom classes there to target inner blocks.

gigitux: > The problem that I see is that the HTML (attached image) generated is different from [the declared template](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/87ccbddfa06bcd5bd0d7fd5398bf201ac6705c55/templates/parts/mini-cart.html#L4). From my experience it seems that every block should have a wrapper.

What about this? I guess that now, the Gutenberg team is [adding more APIs for container block](https://github.com/WordPress/gutenberg/issues/27324), I prefer to keep the wrapper, in this way we:
- are consistent with the template.
- are future proof, since if we want to add support for the new API from Gutenberg we already have the wrapper (I guess no migration).
- follow the standard (I'm not sure about this point, but as I said above, it seems that every block has a dedicated wrapper).

Aljullu: Good points. Honestly, I don't have a strong opinion on this, but agree that it makes sense keeping it consistent with other blocks that also have a wrapper div and allow a custom class name. So :shipit: from my side.

PR #5992, opened 2022-03-03
Store API: Consistent error code and hook naming

### LINKED ISSUE #5980: Update Error Codes in `stock_exceptions_to_wp_errors`
These use a `woocommerce-blocks-` prefix which is not consistent with other error codes and need some modifications.

Found when reviewing https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5904/files

mikejolley:
This is a PR for hook name and error code consistency across the Store API codebase. I've updated docs + usage, and deprecated the previous names inline.

 - **For error codes:** Codes should have the prefix `woocommerce_rest_`. This is inline with the majority of existing error codes, but there were a few with either no prefix, or `blocks` prefix. Additionally, codes should use underscores rather than hyphens to separate words.
 - **For action and filter hooks:** Hook names from WooCommerce should be marked as such. I've added inline `@internal` documentation to state this. For new Store API hooks, they should have the `woocommerce_store_api_` prefix. Newer hooks are doing this, but we had some older ones using `woocommerce_blocks_`. I've deprecated these.

We also had some instances where we had different codes for the same type of error in the cart controller, so I cleared this up also.

Fixes #5980

### Testing

Nothing should be user facing however, placing an order via Block checkout should be smoke tested, as well as one of the extensions consuming the deprecated hooks such as pre-orders (https://github.com/woocommerce/woocommerce-pre-orders/blob/bd1169c93ed3ccd8c2f33f3a54b244aefe63b0c2/includes/class-wc-pre-orders-checkout.php#L39  or subscriptions (https://github.com/woocommerce/woocommerce-subscriptions/blob/0704f35c00ea6df2821b9c1592b77205a96e2937/includes/early-renewal/class-wcs-cart-early-renewal.php#L34-L38)

### Changelog

> Deprecated `woocommerce_blocks_checkout_order_processed` in favour of `woocommerce_store_api_checkout_order_processed`
> Deprecated `woocommerce_blocks_checkout_update_order_meta` in favour of `woocommerce_store_api_checkout_update_order_meta`
> Deprecated `woocommerce_blocks_checkout_update_order_from_request` in favour of `woocommerce_store_api_checkout_update_order_from_request`
> Normalised Store API error codes

### DIFF
--- a/assets/js/blocks/checkout/checkout-order-error/constants.js
+++ b/assets/js/blocks/checkout/checkout-order-error/constants.js
@@ -1,8 +1,8 @@
-export const PRODUCT_OUT_OF_STOCK = 'woocommerce_product_out_of_stock';
+export const PRODUCT_OUT_OF_STOCK = 'woocommerce_rest_product_out_of_stock';
 export const PRODUCT_NOT_PURCHASABLE =
-	'woocommerce_rest_cart_product_is_not_purchasable';
+	'woocommerce_rest_product_not_purchasable';
 export const PRODUCT_NOT_ENOUGH_STOCK =
-	'woocommerce_rest_cart_product_no_stock';
+	'woocommerce_rest_product_partially_out_of_stock';
 export const PRODUCT_SOLD_INDIVIDUALLY =
-	'woocommerce_rest_cart_product_sold_individually';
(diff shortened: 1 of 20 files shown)


nielslange (approved):
LGTM 🚢

PR #5996, opened 2022-03-04
Hide Mini Cart Block on Cart and Checkout page 

### LINKED ISSUE #5931: Mini Cart Block: if the user updates the number of items from WC Core cart page, the Mini Cart block doesn't update
## Describe the bug

On the WC core cart page, if the user updates the number of items, the Mini Cart block doesn't update

## To reproduce

1. Open the FSE editor.
2. Add the `Mini Cart` block on the footer or header area.
3. Save.
6. On the frontend page, add a product to the cart
7. Go on the WC core cart page (by default is `/cart/`)
8. Update the item number.
9. Check that the Mini Cart block counter doesn't update.

## Expected behavior

The Mini Cart should be updated.

## Screenshots

image

gigitux:
This PR hides the Mini Cart Block on Cart and Checkout page because it is not necessary to load this specific block on these two pages.

Fixes #5931 

### Manual Testing

Check out this branch

1. Open the FSE editor.
2. Add the Mini Cart Block in the header or footer section.
3. Open the cart page and check that the Mini Cart Block is not visible.
4. Open the checkout page and check that the Mini Cart Block is not visible.

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -371,7 +371,10 @@ class MiniCart extends AbstractBlock {
 		</span>';
 
 		if ( is_cart() || is_checkout() ) {
-			return;
+			// It is not necessary to load the Mini Cart Block on Cart and Checkout page.
+				return '<div class="' . $wrapper_classes . '" style="visibility:hidden" aria-hidden="true">
+				<button class="wc-block-mini-cart__button ' . $classes . '" aria-label="' . esc_attr( $aria_label ) . '" style="' . $style . '" disabled>' . $button_html . '</button>
+			</div>';
 		}
 
 		$template_part_contents = '';
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -371,9 +371,7 @@ class MiniCart extends AbstractBlock {
 		</span>';
 
 		if ( is_cart() || is_checkout() ) {
-			return '<div class="' . $wrapper_classes . '">
-				<button class="wc-block-mini-cart__button ' . $classes . '" aria-label="' . esc_attr( $aria_label ) . '" style="' . $style . '" disabled>' . $button_html . '</button>
-			</div>';
+			return;
 		}
 
 		$template_part_contents = '';


src/BlockTypes/MiniCart.php
```diff
@@ -371,9 +371,7 @@ protected function get_markup( $attributes ) {
 		</span>';
 
 		if ( is_cart() || is_checkout() ) {
-			return '<div class="' . $wrapper_classes . '">
-				<button class="wc-block-mini-cart__button ' . $classes . '" aria-label="' . esc_attr( $aria_label ) . '" style="' . $style . '" disabled>' . $button_html . '</button>
-			</div>';
+			return;
 		}
```
Aljullu: If we no longer need to render real cart data here, we can move this check to the top of the `get_markup()` method so we return earlier.

Aljullu (commented):
One issue of this approach that I didn't consider initially is that if the Mini Cart block is added to a Row block, then it doesn't occupy any space when it renders and empty markup. For example, see how the menu is right-aligned in the Checkout page but centered in the My Account page.

Checkout page:

imatge

My Account page:

imatge

I wonder if instead of returning empty markup, we should render the same markup but set its `visibility` to hidden and `aria-hidden` to true.

Thoughts?

src/BlockTypes/MiniCart.php
gigitux: > One issue of this approach that I didn't consider initially is that if the Mini Cart block is added to a Row block, then it doesn't occupy any space when it renders and empty markup. For example, see how the menu is right-aligned in the Checkout page but centered in the My Account page.
> 
> Checkout page:
> 
> imatge
> 
> My Account page:
> 
> imatge
> 
> I wonder if instead of returning empty markup, we should render the same markup but set its `visibility` to hidden and `aria-hidden` to true.
> 
> Thoughts?

Great catch! I never tested the Mini Cart Block in the row block. Thanks for pointing this out! 🙏 

> If we no longer need to render real cart data here, we can move this check to the top of the get_markup() method so we return earlier.

About this, I think that if we want to use this solution it is better to continue to render the entire block (with the classes that come from Global Styles too). If there is a class/style that increases the size of the Mini Cart, and then we don't use them on these two pages we might have a similar problem.

Aljullu: > About this, I think that if we want to use this solution it is better to continue to render the entire block (with the classes that come from Global Styles too). If there is a class/style that increases the size of the Mini Cart, and then we don't use them on these two pages we might have a similar problem.

Makes sense. :ok_hand:

Aljullu (approved):
LGTM

src/BlockTypes/MiniCart.php
```diff
@@ -371,7 +371,8 @@ protected function get_markup( $attributes ) {
 		</span>';
 
 		if ( is_cart() || is_checkout() ) {
-			return '<div class="' . $wrapper_classes . '">
+			// It is not necessary to load the Mini Cart Block on Cart and Checkout page.
+				return '<div class="' . $wrapper_classes . '" style="visibility:hidden" aria-hidden="true">
 				<button class="wc-block-mini-cart__button ' . $classes . '" aria-label="' . esc_attr( $aria_label ) . '" style="' . $style . '" disabled>' . $button_html . '</button>
 			</div>';
 		}
```
senadir, line 378: I'm honestly more on the camp of not rendering anything because the mini cart would be occupying limited real estate and the layout would look a bit strange.

I think this highly depends on the layout of the header and I wonder if it can be solved via CSS only.

I don't see it as a priority bug because it can be fixed with CSS.
IMG_0449
IMG_0450

gigitux: `visibility:hidden` was added for [this reason](https://github.com/woocommerce/woocommerce-blocks/pull/5996#pullrequestreview-900342669). As you said, it depends on the layout. I'm not sure that we will be able to find a solution that has a good fit for all layouts.

senadir: Yes, I was just raising what I consider an equally valid issue and that you can't have it all :/ somehow a layout will be broken and we should optimize for the most popular one (default) which is the one Albert showed.

PR #5999, opened 2022-03-04
Change `shippingRatesLoading` to `isLoadingRates` 

opr:
Since whether the shipping rates are loading is a boolean value, the name should be updated to reflect this. I have changed every instance of `shippingRatesLoading` to `isLoadingRates` - the context it is used in is always related to shipping so it's not going to be confusing.

This is a follow up to https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5896#discussion_r819531870 
#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Smoke test cart and checkout and ensure you can change shipping method.
2. Change your country and ensure the shipping method selector is disabled while the new rates load.
3. Try with express payment methods, ensure checking out works fine.

### User Facing Testing
* [x] Same as above


mikejolley (changes requested):
Looks like we won't be able to do this without keeping `shippingRatesLoading` in the exposed payment method interface. This API is public so this would be a breaking change.

Can we just update the naming internally?

opr:
> Looks like we won't be able to do this without keeping `shippingRatesLoading` in the exposed payment method interface. This API is public so this would be a breaking change.
> 
> Can we just update the naming internally?

😖 Yup, I also noticed a TS mistake too so I've updated that in `assets/js/types/type-defs/payment-method-interface.ts`

assets/js/base/context/hooks/payment-methods/use-payment-method-interface.ts
```diff
@@ -160,11 +160,11 @@ export const usePaymentMethodInterface = (): PaymentMethodInterface => {
 			isSelectingRate,
 			needsShipping,
 			selectedRates,
-			selectShippingRate,
+			setSelectedRates: selectShippingRate,
```
mikejolley, line 163: Why did this change?

docs/extensibility/payment-method-integration.md
```diff
@@ -166,7 +166,7 @@ A big part of the payment method integration is the interface that is exposed fo
...
...                                                                            | `shippingRates`, `isLoadingRates`, `selectedRates`, `setSelectedRates`, `isSelectingRate`, `shippingAddress`, `setShippingAddress`, and `needsShipping`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
```
mikejolley: Is this change still true?

docs/block-client-apis/checkout/checkout-api.md
```diff
@@ -47,21 +47,13 @@ The customer data context exposes the api interfaces for the following things vi
...
--   `setSelectedRates`: Function for setting new selected rates.
--   `isSelectingRate`: True when shipping rate selection is persisting to the server.
--   `shippingAddress`: The current set shipping address.
--   `setShippingAddress`: A function for setting the shipping address. This will trigger shipping rates updates.
--   `onShippingRateSuccess`: This is a function for registering a callback to be invoked when shipping rates are retrieved successfully. Callbacks will receive the new rates as an argument.
--   `onShippingRateFail`: This is a function for registering a callback to be invoked when shipping rates fail to be retrieved. Callbacks will receive the error status as an argument.
--   `onShippingRateSelectSuccess`: This is a function for registering a callback to be invoked when shipping rate selection is successful.
--   `onShippingRateSelectFail`: This is a function for registering a callback to be invoked when shipping rates selection is unsuccessful.
--   `needsShipping`: Boolean indicating whether the contents of the order needs shipping (`true`) or not (`false`).
+- `shippingErrorStatus`: The current error status for the context.
```
mikejolley, line 50: Are these changes still correct?

mikejolley (commented):
Added a few questions; some changes might need reverting if the interface is not being changed.

assets/js/base/context/hooks/payment-methods/use-payment-method-interface.ts
opr: ~~Sorry this is my bad. The TS definition and docs needs to change, not the value here.~~

#5896 brought in breaking changes so I have updated this so the code is backward compatible, but the new method name is used internally.

docs/extensibility/payment-method-integration.md
opr: ~~No, but a new change to this file is, there is no such method as `setSelectedRates` on `usePaymentMethodInterface` so the TS definition and docs have been updated.~~

In #5896 I changed this - I will revert it back to how it was because this is incorrect.

docs/block-client-apis/checkout/checkout-api.md
opr: Yes, if you look at what's returned by `ShippingDataProvider` it matches this.

mikejolley (approved):
Seems to be working. I tested google pay successfully also with shipping rate changes.

PR #6002, opened 2022-03-07
style Mini Cart block as inline to remove the experimental features usage.

dinhtungdu:
As discussed in https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5985#issuecomment-1059039656, https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5985#pullrequestreview-900156457 and https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5985#issuecomment-1059199933, in this PR, I try to remove the usage of the experimental features:

- Styling Mini Cart block as an `inline-block` element.
- Removing the alignment support.
- Removing unnecessary style manipulation required by the former approach.

The pros: simpler codebase, simpler styling.

The cons: Lack of alignments (which can be provided by a `<Group>` or `<Row>` block).

My take: I prefer this approach. The button is an inline element by its nature. The Mini Cart button is also 'likely' to be put inside a row.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Add the Mini Cart block to the header using Site Editor.
2. Don't see the alignment option in the block toolbar.
3. See the Mini Cart button is on the left.
4. Add a Row block and move the Mini Cart block there.
5. Change the alignment of the Row block.
6. See the Mini Cart position updated accordingly.
7. Save changes, see it works on the front end.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart/edit.tsx
+++ b/assets/js/blocks/cart-checkout/mini-cart/edit.tsx
@@ -1,19 +1,13 @@
 /**
  * External dependencies
  */
-import {
-	AlignmentControl,
-	BlockControls,
-	InspectorControls,
-	useBlockProps,
-} from '@wordpress/block-editor';
+import { InspectorControls, useBlockProps } from '@wordpress/block-editor';
 import type { ReactElement } from 'react';
 import { formatPrice } from '@woocommerce/price-format';
 import { CartCheckoutCompatibilityNotice } from '@woocommerce/editor-components/compatibility-notices';
 import { PanelBody, ExternalLink } from '@wordpress/components';
 import { getSetting } from '@woocommerce/settings';
 import { __ } from '@wordpress/i18n';
-import { positionCenter, positionRight, positionLeft } from '@wordpress/icons';
 import Noninteractive from '@woocommerce/base-components/noninteractive';
 
 /**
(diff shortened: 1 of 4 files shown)


tomasztunik:
I agree with the conclusion that Mini Cart Block is an element to be used within Layout/Container Blocks that should be responsible for positioning and spacing of elements, could be group, row, columns or some absolute/fixed layers.

If one day Gutenberg added context aware control for align/justify-self for when parent container/block supports it this would make sense but the way it was now I like the proposed approach more.

gigitux:
> I agree with the conclusion that Mini Cart Block is an element to be used within Layout/Container Blocks that should be responsible for positioning and spacing of elements, could be group, row, columns or some absolute/fixed layers.

I agree with you! 

> If one day Gutenberg added context aware control for align/justify-self for when parent container/block supports it this would make sense but the way it was now I like the proposed approach more.

Yes! They are already [working on this](https://github.com/WordPress/gutenberg/issues/27324)!

gigitux (approved):
Thanks for working on this!

LGTM 🚢

Aljullu (approved):
Looks good to me as well. Good job! :shipit:

PR #6005, opened 2022-03-07
Remove `internal` comments from Store API file comments

### LINKED ISSUE #5998: Remove all "internal" notices in Store API
The final step before Store API is considered stable is to remove the internal comments:

> @internal This API is used internally by Blocks.

Within the `src/StoreAPI` directory.

mikejolley:
Removes all `internal` notices stating that Store API is a private API so it can be considered stable.

Nothing to test here; only touches docblocks. Check PHP Unit passes in case of syntax error.

Closes #5998


alexflorisca (approved):
Looks good 👍

PR #6008, opened 2022-03-07
Add feature gate for enabling the global styles for Product Sale Badge block only for the feature plugin

### LINKED ISSUE #6007: Product Sale Badge - add Global Styles with the feature gate
@tomasztunik and I noticed that the global style support for the Product Sale Badge block is not enabled only for the feature plugin

gigitux:
This PR adds the feature gate for enabling the global styles support for the Product Sale Badge block only for the feature plugin. We don't introduce any breaks because Product Sale Badge Global Styles support is introduced in the 7.0.0 version that is not integrated into WooCommerce Core.

It also adds additional controls to add spacing support to several blocks:  Button, Image, and Sale Badge.

Fixes #6007

### DIFF
--- a/assets/js/atomic/blocks/product-elements/button/supports.ts
+++ b/assets/js/atomic/blocks/product-elements/button/supports.ts
@@ -3,6 +3,11 @@
  */
 import { isFeaturePluginBuild } from '@woocommerce/block-settings';
 
+/**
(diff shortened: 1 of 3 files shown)


tomasztunik (approved):
👌

PR #6011, opened 2022-03-08
Move __experimentalSelector outside of hasSpacingStyleSupport check

tomasztunik:
Minor regression - we've overlooked __experimentalSelector ending up inside of hasSpacingStyleSupport check.


gigitux (approved):
Thanks for catching this!

PR #6013, opened 2022-03-08
E2E test improvements and consolidatng

alexflorisca:
There's been a few PRs on E2E tests and as such there's been a bit of cross over. This PR aims to simplify things a bit. Specifically:

- Use `useTheme()` helper to make sure we are remembering and switching back to the original theme after activating a new theme within a test
- Move all block related functions under the `shopper.block` object for clarity
- Simplify the `goToBlockPage()` function to derive the block page permalink directly, rather than visiting wp-admin
- Remove now unnecessary constants

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. Make sure all E2E tests pass

### DIFF
--- a/tests/utils/shopper.js
+++ b/tests/utils/shopper.js
@@ -7,71 +7,60 @@ import {
 	SHOP_CART_PAGE,
 } from '@woocommerce/e2e-utils';
 
-/**
- * Internal dependencies
- */
-import { getBlockPagePermalink } from './get-block-page-permalink';
-import { SHOP_CART_BLOCK_PAGE, SHOP_CHECKOUT_BLOCK_PAGE } from './constants';
-
 export const shopper = {
 	...wcShopper,
 
-	goToCheckoutBlock: async () => {
-		await page.goto( SHOP_CHECKOUT_BLOCK_PAGE, {
-			waitUntil: 'networkidle0',
-		} );
-	},
-
-	goToCartBlock: async () => {
-		await page.goto( SHOP_CART_BLOCK_PAGE, {
-			waitUntil: 'networkidle0',
-		} );
-	},
-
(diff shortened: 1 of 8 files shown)


tests/utils/shopper.js
```diff
@@ -7,71 +7,60 @@ import {
...
+	// This is important as we might one day merge this into core WC.
 	block: {
-		goToCart: async () => {
-			await page.goto( SHOP_CART_BLOCK_PAGE, {
+		// All block pages have a title composed of the block name followed by "Block".
+		// E.g. "Checkout Block or Mini Cart Block". The permalinks are generated from
+		// the page title so we can derive them directly
+		goToBlockPage: async ( blockName ) => {
+			const config = require( 'config' );
+			const baseUrl = config.get( 'url' );
```
tarhi-saad: We can import the `BASE_URL` from `tests/e2e/utils.js` and replace:

```js
const config = require( 'config' );
const baseUrl = config.get( 'url' );
...
const url = baseUrl + pageTitle.toLowerCase().replace( / /g, '-' );
```

with

```js
/**
 * Internal dependencies
 */
import { BASE_URL } from '../utils';
...
const url = BASE_URL + pageTitle.toLowerCase().replace( / /g, '-' );
```

tarhi-saad (approved):
Thank you, @alexflorisca, for this great PR! Everything looks good. There is just a tiny suggestion I made below. And I also pushed some commits where I implemented some suggestions you made in my latest PR.

I've used `beforeEach` as you suggested in `cart-proceed-to-checkout.test.js`, but kept `afterAll` to reset the last test in this file to avoid any left side effects at the end. Please let me know your thoughts on this.

PR #6014, opened 2022-03-08
Update debug script name

alexflorisca:
I've noticed the `debug` script is in the middle of the test scripts. Thought it'd be sensible to update the name.


tarhi-saad (commented):
Hi @alexflorisca! Thank you for the update! I also thought about this when adding this script, but the `ndb` command seems to be working with any kind of `node.js` running script (e.g. unit tests), so I figured we may use it for something else other than E2E tests debugging in the future. Please let me know if you think this is a scenario that is likely to happen or if it should be used only while debugging our E2E tests.

alexflorisca:
@tarhi-saad I think we should be specific here for our use case, and if that changes we can rename the script accordingly. "debug" is quite a vague name and doesn't really suggest what we are debugging. We could name it `test:debug` instead, that way it can be used for unit & integration tests as well?

tarhi-saad (approved):
> @tarhi-saad I think we should be specific here for our use case, and if that changes we can rename the script accordingly. "debug" is quite a vague name and doesn't really suggest what we are debugging. We could name it `test:debug` instead, that way it can be used for unit & integration tests as well?

It makes sense! Thank you for the clarification, @alexflorisca! Then it looks good to me.🚢

PR #6019, opened 2022-03-09
Refactor BlockTemplateUtils

tjcafferkey:
Slight tidy up of the `BlockTemplateUtils.php` file.

* Moved `remove_theme_templates_with_custom_alternative` from `BlockTemplatesController` to `BlockTemplateUtils`.
* Dropped `gutenberg_` prefix from some methods as this was accidentally carried over from duplicate functions from the Gutenberg codebase.
* Updated some comments around _almost_ duplicate methods.
* Denoted the `BlockTemplateUtils` class as `@internal` to indicate it's not for public use.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests

### Manual Testing

1. Ensure you have a block theme active and WordPress 5.9
2. Load Appearance > Site Editor and check that Woo templates are being loaded in the All Templates view.
3. Load one in the Site Editor and ensure this is loading on the frontend without any customizations
4. Make customizations to the template, and ensure theyre reflected on the frontend.
5. Clear customizations and ensure the frontend template is reset to its original uncustomized state.

Now copy some of the [Woo template files](https://github.com/woocommerce/woocommerce-gutenberg-products-block/tree/trunk/templates/templates) into your block theme and repeat steps 1-5 again ensuring your themes templates are overriding the Woo default themes.


albarin (approved):
I don't have a lot of context, but the refactor looks good. 👍

src/Utils/BlockTemplateUtils.php
```diff
@@ -3,10 +3,9 @@
 
 use Automattic\WooCommerce\Blocks\Domain\Services\FeatureGating;
 
-
 /**
- * BlockTemplateUtils class used for serving block templates from Woo Blocks.
- * IMPORTANT: These methods have been duplicated from Gutenberg/lib/full-site-editing/block-templates.php as those functions are not for public usage.
+ * Utility methods used for serving block templates from WooCommerce Blocks.
+ * {@internal This class and its methods should only be used within the BlockTemplateController.php and is not intended for public use. }}
```
sunyatasattva: There are two closing `}}` instead of one. I suppose this is a mistake. I think in general it looks ok without brackets and it's still valid: https://docs.phpdoc.org/guide/references/phpdoc/tags/internal.html

src/Utils/BlockTemplateUtils.php
```diff
@@ -119,12 +118,13 @@ public static function gutenberg_inject_theme_attribute_in_content( $template_co
 
 	/**
 	 * Build a unified template object based a post Object.
+	 * Important: This method is an almost identical duplicate from wp-includes/block-template-utils.php as it was not intended for public use.
```
sunyatasattva: Since it was not intended for public use, perhaps let's add `@internal` here as well? Also, what about mentioning why do we need this if it is “almost” identical to the other one? What do you think?

src/Utils/BlockTemplateUtils.php
```diff
@@ -172,6 +172,7 @@ public static function gutenberg_build_template_result_from_post( $post ) {
 
 	/**
 	 * Build a unified template object based on a theme file.
+	 * Important: This method is an almost identical duplicate from wp-includes/block-template-utils.php as it was not intended for public use.
```
sunyatasattva: Same as above

sunyatasattva (dismissed):
Great first steps, love it!

Aside from the notes I have mentioned, given our recent conversation about release process pain points, perhaps we could improve the manual testing steps to provide the exact specific step-by-step instruction?

As far as I understood, the release manager happens to do a lot of filling of the gaps. So, for example, if I had no context on this PR, I would probably be missing a few steps.

Hopefully I'm not too pedantic, what are your thoughts?

src/Utils/BlockTemplateUtils.php
tjcafferkey: It was intentional, I've seen both used in PHP and its documentation. Perhaps there is a reason I'm not aware of for double closing `}}` rather than single. Either way, happy to remove one of the two!

src/Utils/BlockTemplateUtils.php
tjcafferkey: The entire class is marked as `@internal` at the top of the file. I think it would add confusion if we then add it to various other methods as that would insinuate non-marked `@internal` methods are OK for public use?

As for the difference between this and its original counterpart, I will write something non-specific to avoid us having to update this comment with every future change. I think as long as the reader knows its not identical that should be good too.

src/Utils/BlockTemplateUtils.php
sunyatasattva: Ok so I found it, in [this old reference](https://docs.phpdoc.org/2.9/references/phpdoc/tags/internal.html):

> The inline version of this tag may, contrary to other inline tags, contain text but also other inline tags. To increase readability and ease parsing should the tag be terminated with a double closing brace, instead of a single one.

Note that this is only for the inline version of the tag. Also note that in the most updated version of the docs ([linked above](https://docs.phpdoc.org/guide/references/phpdoc/tags/internal.html)), basically the same page, this sentence is not present.

It is also quite ironic for me that the reasoning given was to “increase readability”. While searching for this, I found this funny thread on `r/lolphp`: https://www.reddit.com/r/lolphp/comments/4i27ud/to_increase_readability_should_the_tag_be/

I would perhaps consider removing the braces altogether. It's valid syntax both in the old and new version. What do you think?

sunyatasattva (approved):
I replied to the comment regarding the strange braces rule. But in the meantime I am approving this, as to not block you when you make a decision in that regard :)

src/Utils/BlockTemplateUtils.php
tjcafferkey: I removed one of the two braces yesterday, like you suggested. Looking at other usages in the WC codebase that looks more consistent with what is there. Ill merge for now, and we can always remove them in another PR if we feel the need to

tjcafferkey:
I'm going to go ahead and merge this. The failing E2E tests are unrelated due to a regression with the latest Gutenberg. Potential fix https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6080

PR #6020, opened 2022-03-09
Remove X- Prefixes from custom headers

mikejolley:
I've read in a few places that `X-` prefixed HTTP headers are no longer recommended. Before we officially launch Store API v1, I thought it might be a good time to rename them.

Ref: https://www.keycdn.com/support/custom-http-headers#naming-conventions

Renames:

- `X-WC-Store-Api-Nonce` to `Nonce`
- `X-WC-Store-Api-Nonce-Timestamp` to `Nonce-Timestamp`
- `X-WC-Store-User` to `User-ID`

### Testing

Covered by unit tests.

### Changelog

> Renamed Store API custom headers to remove `X-WC-Store-API` prefixes.

### Dev Note

As part of finalising the Store API, we've removed all `X-` prefixes from headers since this is no longer a recommended practice. The `X-WC-Store-API-Nonce` has been renamed to just `Nonce` going forward. If you've been consuming the experimental Store API, this will need to be renamed in your client. This also includes `X-WC-Store-Api-Nonce-Timestamp` to `Nonce-Timestamp` and `X-WC-Store-User` to `User-ID`.

### DIFF
--- a/assets/js/middleware/store-api-nonce.js
+++ b/assets/js/middleware/store-api-nonce.js
@@ -39,12 +39,12 @@ const isStoreApiRequest = ( options ) => {
 const setNonce = ( headers ) => {
 	const nonce =
 		typeof headers?.get === 'function'
-			? headers.get( 'X-WC-Store-API-Nonce' )
-			: headers[ 'X-WC-Store-API-Nonce' ];
+			? headers.get( 'Nonce' )
+			: headers.Nonce;
 	const timestamp =
 		typeof headers?.get === 'function'
-			? headers.get( 'X-WC-Store-API-Nonce-Timestamp' )
-			: headers[ 'X-WC-Store-API-Nonce-Timestamp' ];
+			? headers.get( 'Nonce-Timestamp' )
+			: headers[ 'Nonce-Timestamp' ];
 
 	if ( nonce ) {
 		updateNonce( nonce, timestamp );
@@ -85,7 +85,7 @@ const appendNonceHeader = ( request ) => {
 	const headers = request.headers || {};
 	request.headers = {
 		...headers,
-		'X-WC-Store-API-Nonce': currentNonce,
+		Nonce: currentNonce,
 	};
 	return request;
 };
--- a/assets/js/base/context/providers/cart-checkout/utils.ts
+++ b/assets/js/base/context/providers/cart-checkout/utils.ts
@@ -55,9 +55,9 @@ export const processCheckoutResponseHeaders = (
 	}
 
 	// Update user using headers.
-	if ( headers?.get( 'X-WC-Store-API-User' ) ) {
+	if ( headers?.get( 'User-ID' ) ) {
 		dispatchActions.setCustomerId(
-			parseInt( headers.get( 'X-WC-Store-API-User' ) || '0', 10 )
+			parseInt( headers.get( 'User-ID' ) || '0', 10 )
 		);
 	}
 };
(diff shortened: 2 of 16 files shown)


assets/js/middleware/store-api-nonce.js
```diff
@@ -39,12 +39,12 @@ const isStoreApiRequest = ( options ) => {
 const setNonce = ( headers ) => {
 	const nonce =
 		typeof headers?.get === 'function'
-			? headers.get( 'X-WC-Store-API-Nonce' )
-			: headers[ 'X-WC-Store-API-Nonce' ];
+			? headers.get( 'Nonce' )
+			: headers.Nonce;
```
nielslange, line 43: Shouldn't this be `headers[ 'Nonce' ];`?

nielslange (commented):
Given your testing steps, I checked the unit tests and looked at the code changes, but didn't run a smoke test. The code changes look good. I just spotted one case in which you converted an array into an object. I added a note, as I'm unsure if that change will work or not.

assets/js/middleware/store-api-nonce.js
mikejolley: Ill double check this but I think it's the same (this will be an object). I'm pretty sure key=>value pairs are always objects. 

FYI it was prettier which changed this.

nielslange: In that case, let's 🚢 the change!

mikejolley: > JavaScript arrays are not associative arrays and so, [array elements cannot be accessed using strings as indexes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#notes), but must be accessed using integers as indexes.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array

I was correct, this is an object.

senadir (approved):
This is looking good. I'm not sure if we should provide any backward compatibility or not? maybe for `X-WC-Store-API-Nonce`? Would that overcomplicate our store API code?

mikejolley:
@senadir It would overcomplicate slightly because both would need to be sent and received (client side). Seeing as we're the only/main consumer of this right now, and it's "internal" status (at least until next week!) It's now or never really. Otherwise we'd also have to worry about deprecation of those headers too in the future.

Dev note should sufffice?

senadir:
I think a dev note would be fine for now. I don't think we need any client implementation, but just because I'm glancing over the code, mainly something like:

```php
protected function add_nonce_headers( \WP_REST_Response $response ) {
		$response->header( 'Nonce', wp_create_nonce( 'wc_store_api' ) );
		$response->header( 'X-WC-Store-API-Nonce', wp_create_nonce( 'wc_store_api' ) );
		$response->header( 'Nonce-Timestamp', time() );
		$response->header( 'User-ID', get_current_user_id() );
		return $response;
	}

protected function check_nonce( \WP_REST_Request $request ) {
		if ( $request->get_header( 'X-WC-Store-API-Nonce' ) ) {
			$nonce = $request->get_header( 'X-WC-Store-API-Nonce' );
			// deprication logic.
		} else {
			$nonce = $request->get_header( 'Nonce' );
```

mikejolley:
Ok, we have back compatibility in place and deprecation warnings. Todo has been added to remove this by Blocks 7.5. cc @senadir

PR #6021, opened 2022-03-09
Rename Legacy Template block to Classic Template block

### LINKED ISSUE #5861: Rename Legacy Template block to Classic Template block
Suggested by @nerrad, we want to rename the Legacy Template block to Classic Template block.

Some relevant files:

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/legacy-template/index.tsx

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/LegacyTemplate.php

albarin:
This PR addresses the suggestion to rename "Legacy Template" block to "Classic Template" block.

With @sunyatasattva we decided to change every occurrence and get feedback on it to see if anyone sees any issue with the changes.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5861

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### User Facing Testing

1. With a block theme active open the site editor.
2. Go to Appearance > Site Editor and select one of the WooCommerce templates (for example Single Product Page).
4. Open the List view and check that `WooCommerce Legacy Template` is now `WooCommerce Classic Template`.
5. Click on it and check that in the right sidebar it also appears as Classic template: `WooCommerce Classic Template
Renders classic WooCommerce PHP templates.`


albarin:
Hey @tjcafferkey, could you have a look?
With @sunyatasattva we decided to change everything `legacy` to `classic` and we wanted to discuss it with you since you may know what potential implications this could have and what could break.

I just saw that changing the internal `$block_name` of the `LegacyTemplate.php` breaks things and I was getting this:
Screenshot 2022-03-09 at 15 12 58
So I reverted the change.

tjcafferkey:
The main thing I think we need to be wary of changing is the block name/ID. For example [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/legacy-template/index.tsx#L72) it would need to remain the same. The same thing applies in the template markup, we won't be able to change `<!-- wp:woocommerce/legacy-template {"template":"single-product"} /-->` to `<!-- wp:woocommerce/classic-template {"template":"single-product"} /-->`.

This is because users who have customized templates using this blocks old ID (e.g. `legacy-template`) will have saved this in their database, and it would result in a corrupt blocks when loaded, as seen in your screenshot.

Filenames, PHP class names etc though should be fine as far as I know but I may have overlooked other parts.

tjcafferkey:
A few thoughts:
* I think we should rename the blocks directory in `assets/js/blocks/` from `legacy-template` > `classic-template` so its consistent with other filenames
* I think we should also add a comment [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/assets/js/blocks/legacy-template/index.tsx#L72) explaining why the block is registered under `legacy-template` rather than `classic-template` and potentially link to the original issue for context.
* We can also add some regression testing steps to this PR like [these here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6019)

sunyatasattva (approved):
Yuhuu! I'm approving this and also merging it in for you by skipping the required CI steps. The failed E2E is due to something unrelated to this and to the newest Gutenberg version causing a regression. A fix might be available here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6080

PR #6025, opened 2022-03-09
Shopper → Cart → Can update product quantity

### LINKED ISSUE #5755: Critical flows: Shopper → Cart → Can update product quantity
### Story

As a shopper, I want to be able to adjust the product quantity (in case I added too much or as little before) so that I receive the desired quantity.

### File

`tests/e2e/specs/shopper/cart-products.test.js`

### Test

1. Go to the product page.
2. Add a product to the cart.
3. Go to the cart page.
4. Verify that the product quantity can be increased.
5. Verify that the total price increased.
6. Verify that the product quantity can be decreased.
7. Verify that the total price decreased.

tarhi-saad:
Fixes #5755

### Notes

This PR adds an e2e test to verify that shoppers can update product quantity.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/cart-update-product-quantity.test.js` to run only this test case.

### TODO

- [X] Find why test passes only in interactive mode, otherwise, we have the following error: `TimeoutError: Element a.wc-block-cart__submit-button not found`
- [X] Add qty using the "add" button
- [X] decrease qty using the "decrease" button


tests/e2e/specs/shopper/cart-update-product-quantity.test.js
```diff
@@ -0,0 +1,66 @@
...
+		await shopper.addToCartFromShopPage( SIMPLE_PRODUCT_NAME );
+		await shopper.block.goToCart();
+
+		await shopper.block.setCartQuantity( SIMPLE_PRODUCT_NAME, 4 );
+		await expect( page ).toMatchElement(
+			'button.wc-block-cart__submit-button[disabled]'
+		);
+
+		// To avoid flakiness: The default "idleTime: 500" fails in headless mode
+		await page.waitForNetworkIdle( { idleTime: 1000 } );
```
alexflorisca, line 35: I think you can use the `uiUnblocked()` function from the [woo e2e-utils](https://github.com/woocommerce/woocommerce/blob/5af5be510032f9de1b1c6c6dacbbe24e53f8b691/packages/js/e2e-utils/src/page-utils.js#L95) here. I'm not sure it works with the block, but it's the standard way of waiting for UI updates in Woo

tests/e2e/specs/shopper/cart-update-product-quantity.test.js
```diff
@@ -0,0 +1,66 @@
...
+	} );
+
+	it( 'allows customer to increase product quantity via the plus button', async () => {
+		await shopper.block.increaseCartQuantityByOne( SIMPLE_PRODUCT_NAME );
+		await expect( page ).toMatchElement(
+			'button.wc-block-cart__submit-button[disabled]'
+		);
+
+		// To avoid flakiness: The default "idleTime: 500" fails in headless mode
+		await page.waitForNetworkIdle( { idleTime: 1000 } );
```
alexflorisca, line 48: Try using `uiUnblocked` here too

tests/e2e/specs/shopper/cart-update-product-quantity.test.js
```diff
@@ -0,0 +1,66 @@
...
+	} );
+
+	it( 'allows customer to decrease product quantity via the minus button', async () => {
+		await shopper.block.decreaseCartQuantityByOne( SIMPLE_PRODUCT_NAME );
+		await expect( page ).toMatchElement(
+			'button.wc-block-cart__submit-button[disabled]'
+		);
+
+		// To avoid flakiness: The default "idleTime: 500" fails in headless mode
+		await page.waitForNetworkIdle( { idleTime: 1000 } );
```
alexflorisca, line 61: uiUnblocked here too

tests/utils/expressions.js
```diff
@@ -0,0 +1,37 @@
...
+
+export const getQtyInputExpression = ( args = {} ) => {
+	let qtyValue = '';
+
+	if ( args.qty ) {
+		qtyValue = ` and @value="${ args.qty }"`;
+	}
+
+	return (
+		'input[contains(@class, "wc-block-components-quantity-selector__input")' +
```
alexflorisca: noob question: I've never used x path selectors, what are the benefits over normal selectors?

tests/utils/expressions.js
```diff
@@ -0,0 +1,37 @@
...
+	return 'button[contains(@class, "wc-block-components-quantity-selector__button--plus")]';
+};
+
+export const getQtyMinusButtonExpression = () => {
+	return 'button[contains(@class, "wc-block-components-quantity-selector__button--minus")]';
+};
+
+export const getCartItemExpression = ( productTitle, args ) =>
+	'//tr[contains(@class, "wc-block-cart-items__row") and ' +
+	'td[@class="wc-block-cart-item__product" and div[@class="wc-block-cart-item__wrap" and ' +
```
alexflorisca: What do you think about making these selectors a little less specific? We should be able to refactor the html here to not use tables, without breaking this test.

tests/utils/expressions.js
```diff
@@ -0,0 +1,37 @@
...
+export const getQtyPlusButtonExpression = () => {
+	return 'button[contains(@class, "wc-block-components-quantity-selector__button--plus")]';
+};
+
+export const getQtyMinusButtonExpression = () => {
+	return 'button[contains(@class, "wc-block-components-quantity-selector__button--minus")]';
+};
+
+export const getCartItemExpression = ( productTitle, args ) =>
+	'//tr[contains(@class, "wc-block-cart-items__row") and ' +
```
alexflorisca: I would suggest changing the name of any of these expression functions to include XPath instead (for example `getCartItemExpression` -> `getCartItemXPath`, to give it more context as to what the syntax is. (I'd never seen it before for example, and I thought you were just commenting out that line!)

tests/utils/shopper.js
```diff
@@ -293,5 +300,50 @@ export const shopper = {
 				}
 			);
 		},
+
+		setCartQuantity: async ( productTitle, quantityValue ) => {
+			const cartItemXPath = getCartItemExpression( productTitle );
+			const quantityInputXPath =
+				cartItemXPath + '//' + getQtyInputExpression();
```
alexflorisca: I noticed that `getCartItemExpression` returns the x path with `//` at the start, but `getQtyInputExpression` doesn't include the `//`. This was a bit misleading to me, I would suggest consistency with the expression functions here for clarity

tests/utils/shopper.js
```diff
@@ -293,5 +300,50 @@ export const shopper = {
...
+
+			const [ quantityInput ] = await page.$x( quantityInputXPath );
+			await quantityInput.focus();
+			await pressKeyWithModifier( 'primary', 'a' );
```
alexflorisca, line 320: You can use `await expect( page ).toFill()` which will clear the input before typing in it.

alexflorisca (commented):
Nice work @tarhi-saad, I didn't know about x path selectors so I've learned something today! I left a few comments inline

tests/utils/expressions.js
tarhi-saad: @alexflorisca, I never used it before either. But after doing some research, it appears `XPath` is more powerful than `CSS Selector` when it comes to using complex conditional logic to get a specific element. In my case, what made me ditch `CSS Selector` in this instance is that it can't target an element by the content of its text node.

tests/utils/expressions.js
tarhi-saad: `XPath` is an expression language. And the text returned by each of the above functions is called `expression`. So, it's more accurate to name the functions by their return type instead of the language itself. But, we may change it to `PathExpression` instead to keep an accurate naming and help anyone who isn't familiar with `XPath` to follow the code (e.g., `getCartItemPathExpression `).

tests/utils/shopper.js
tarhi-saad: `getQtyInputExpression ` is also reused in the `getQtyItemExpression` function in `tests/utils/expressions.js`:

```js
export const getQtyItemExpression = ( args ) =>
	'div[@class="wc-block-cart-item__quantity" and div[@class="wc-block-components-quantity-selector" and ' +
	getQtyInputExpression( args ) + // =====> Here
	']]';
```

In the code snippet above, if `getQtyInputExpression` starts with the axis `//` this will return unwanted result.

tests/utils/expressions.js
tarhi-saad: Removed table elements from the query, and kept only: `//div[@class="wc-block-cart-item__wrap"...`

tests/utils/expressions.js
tarhi-saad: - Renamed functions by adding: `PathExpression` at the end.
- Renamed the expression file: `path-expressions`

tests/utils/shopper.js
tarhi-saad: We can't use `toFill()` function in this case. We will need to use `CSS Selectors` to target the right input. And we can't do that when we have multiple elements in the cart, so we use `XPath` instead.

tests/utils/shopper.js
alexflorisca: Ok cool, let's leave it as is then

tests/utils/shopper.js
alexflorisca: Ok understood 👍

alexflorisca (approved):
Great work @tarhi-saad, ship it! I notice, the E2E tests with Gutenberg installed fail, but that happens for unrelated reasons and on my PR too. I'll react out to Kirigami as it seems store editing related

PR #6033, opened 2022-03-10
Ignore storybook dist folder when running eslint

senadir:
Recently we saw that running `npm run lint:js` yielded nothing, after running `npm run lint:js -- --debug` (h/t @tomasztunik ) we found that Storybook dist folder is being included and it's blocking the linter from returning.

I excluded that folder and the command is working fine now.

### Steps to reproduce:

- `npm run storybook:build` to generate the dist folder.
- `npm run lint:js` should work just fine and return.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -7,3 +7,4 @@ node_modules
 vendor
 legacy
 tests/e2e/specs
+storybook/dist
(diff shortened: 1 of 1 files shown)


alexflorisca (approved):
I've added the `/reports` folder to `.eslintignore` as well as it was hanging after a failed linting atempt. Nice find 👍

PR #6035, opened 2022-03-10
Add `CurrencyCode` type for the Currency `code` property

### LINKED ISSUE #5397: Improve Currency type documentation
> This issue is an umbrella issue for the automatically generated issues via the `@todo` comments by @github-actions. This replaces and closes #5391, #5392, #5393

As part of TypeScript migration and Storybook docs, we have made and documented a few TS interfaces. Currently, the currency interface accepts any `string` as currency code, which would not be exact. A stricter type would mean allowing only available currency codes.

The same, arguably, should be done for allowed currency symbols, though perhaps it would be less important.

Also, during the documentation, the [meaning of `minorUnit`](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/509bede1c950d3426ab3819fed81c5349642c106/assets/js/types/type-defs/currency.ts#L13) it wasn't understood and hence not documented. Documentation should be added there.

---
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/509bede1c950d3426ab3819fed81c5349642c106/assets/js/types/type-defs/currency.ts#L5-L18
---

See #5293

albarin:
Currently, the currency interface accepts any string as currency code, which would allow entering incorrect values.

This PR adds a new `type CurrencyCode` used as the type for the `code` and `currency_code` properties. That way the possible string values are restricted only to the allowed _ISO 4217_ codes.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5397

#### Automated Tests

### Manual Testing
1. Create a new page with the "All Products" and the "Filter Products by Price" blocks.
2. Go to the page on the front
3. Check that the price/currency is showing ok on the slider inputs and the product prices.

-

1. Add a product to the cart
2. Go to the checkout
3. Check the price and currency are ok


packages/prices/utils/price.ts
```diff
@@ -90,7 +90,7 @@ export const getCurrencyFromPriceResponse = (
 		currency_suffix: suffix,
 	} = currencyData;
 
-	return {
+	return <Currency>{
```
tomasztunik: Casting is not required here, it will validate returned object against the return type provided in function definition at line 78

albarin: Oh okay, the IDE was complaining about it 🤔

sunyatasattva:
Link to the Slack thread where it was decided to go for union types on this: p1646913422664749-slack-C02SGH7JBGS

sunyatasattva:
I've added the label `skip-changelog` and removed the changelog entry. This is an internal change and should probably not be listed in the changelog which is exposed to our users.

sunyatasattva (approved):
I also checked that WooCommerce has uppercase Currency Codes, so that's nice. I'll approve this and also merging it in for you by skipping the required CI steps. The failed E2E is due to something unrelated to this and to the newest Gutenberg version causing a regression. A fix might be available here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6080

However, I'd probably amend the manual testing steps in the PR description itself, to make sure to test a few more components which consume the `Currency` type.

So, nothing code related.

Issue #6031, opened 2022-03-10
Translated String still shows in english


masteradhoc:
## Describe the bug
The String "shipping method" seems translated but still shows in english.
Although everything else seems to be shown correctly in german

## To reproduce
1) install woocommerce blocks
2) change site language to de_DE
3) check cart with regular block

## Expected behavior
String should show in german

## Screenshots
https://translate.wordpress.org/projects/wp-plugins/woo-gutenberg-products-block/stable/de/default/?filters%5Bterm%5D=shipping+method&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Apply+Filters&sort%5Bby%5D=priority&sort%5Bhow%5D=desc

![image](https://user-images.githubusercontent.com/6242098/157615363-31ed7350-e92a-41a9-b519-c7c84d2e8fba.png)
![image](https://user-images.githubusercontent.com/6242098/157615380-a2e51567-456f-4b81-99f2-ac4b17bd1039.png)

### WordPress (please complete the following information):

-   Core version: 5.9.1
-   WooCommerce version: 6.3
-   WooCommerce Blocks version: 7.1.0
-   Site language: de_DE

masteradhoc:
@Aljullu @mikejolley Thanks for the fast fix and release here! Can confirm this works now on our side. (also your detected phone issue)

Closing this issue as its resolved now.

PR #6036, opened 2022-03-10
Correctly load JS translations from WC core (II)

Aljullu:
Might be related to #6031, but I'm not 100% sure because I wasn't able to reproduce that one.

The reason of the issue is that in #5910 we changed how translation scripts are loaded. Instead of directly printing them to the page, we now enqueue them. That works well for most of them, except the ones that are being rendered in the footer. At that point, enqueuing a new script won't work because after scripts are printed in the footer, they aren't printed again, that means they will miss the last opportunity to be printed.

What this PR does is that it checks if scripts have already been printed into the page or not. If they haven't, it enqueues the script, if they have, if directly prints the script.

### Manual Testing

1. Go to Settings > General > Language (`/wp-admin/options-general.php`) and select a language which has WC core translation at 100% but WC Blocks' at 0%. (I tested with `Català`).
2. Go to the updates screen (`/wp-admin/update-core.php`), scroll down and update the language translations.
3. Visit the Cart or Checkout block in the frontend and verify **all texts** are translated.

Before #6031| `trunk` | With this fix
--- | --- | ---
imatge | imatge | imatge

_(notice some input fields aren't translated in `trunk`)_

There are other blocks affected, like Filter Products by Attribute.

@mikejolley given that this can be considered a regression introduced in WC Blocks 7.1.0, should we include it in 7.2.0?

### Changelog

> Fix loading more WC core translations in locales where WC Blocks is not localized for some strings.

### DIFF
--- a/woocommerce-gutenberg-products-block.php
+++ b/woocommerce-gutenberg-products-block.php
@@ -204,13 +204,22 @@ function woocommerce_blocks_get_i18n_data_json( $translations, $file, $handle, $
 	} )( "{$domain}", {$json_translations} );
 JS;
 
-	// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.NoExplicitVersion
-	wp_register_script( $handle_filename, '', array( 'wp-i18n' ), false, true );
-	wp_enqueue_script( $handle_filename );
-	wp_add_inline_script(
-		$handle_filename,
-		$output
-	);
+	if ( count( $wp_scripts->done ) === 0 ) {
+		// If we hadn't printed any script into the page, let's enqueue the translations.
+		// phpcs:ignore WordPress.WP.EnqueuedResourceParameters.NoExplicitVersion
+		wp_register_script( $handle_filename, '', array( 'wp-i18n' ), false, true );
+		wp_enqueue_script( $handle_filename );
+		wp_add_inline_script(
+			$handle_filename,
+			$output
+		);
+	} else {
+		// If we have already printed scripts into the page, there is the change that
+		// scripts have finished being printed. That means that if we enqueued them here,
+		// they would never be printed. Instead of enqueuing, then, let's print directly
+		// the script tag.
+		printf( "<script type='text/javascript'>\n%s\n</script>\n", $output ); // phpcs:ignore
+	}
 
 	// Finally, short circuit the pre_load_script_translations hook by returning
 	// the translation JSON from the feature plugin, if it exists so this hook
--- a/woocommerce-gutenberg-products-block.php
+++ b/woocommerce-gutenberg-products-block.php
@@ -214,7 +214,7 @@ JS;
 			$output
 		);
 	} else {
-		// If we have already printed scripts into the page, there is the change that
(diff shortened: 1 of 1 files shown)


dinhtungdu (commented):
I followed the testing steps and generally, it worked as expected.

But I noticed that, in the checkout form, `Phone (optional)` wasn't translated. That string has a [translation](https://translate.wordpress.org/projects/wp-plugins/woocommerce/dev/ca/default/?filters%5Bterm%5D=Phone+%28optional%29&filters%5Bterm_scope%5D=scope_any&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Apply+Filters&sort%5Bby%5D=priority&sort%5Bhow%5D=desc) on translate.wordpress.org. Searched through the `wp-content/languages/plugins/` folder, I found a json file containing translation for `Phone (optional)` as well.

Screen Shot 2022-03-11 at 09 35 38

<details>
<summary>wp-content/languages/plugins/woocommerce-ca-4721ee46d333c429f595e88bd73399ba.json</summary>

```
{"translation-revision-date":"2022-03-04 11:52:05+0000","generator":"GlotPress\/3.0.0-alpha.2","domain":"messages","locale_data":{"messages":{"":{"domain":"messages","plural-forms":"nplurals=2; plural=n != 1;","lang":"ca"},"Step description text.":["Text de descripci\u00f3 del pas."],"Step":["Pas"],"Please select a country to calculate rates.":["Seleccioneu un pa\u00eds per calcular les tarifes."],"Please select a state.":["Seleccioneu un estat."],"Please select a country.":["Seleccioneu un pa\u00eds."],"Please select a value.":["Seleccioneu un valor."],"Phone (optional)":["Tel\u00e8fon (opcional)"],"Invalid value.":["Valor no v\u00e0lid."],"Phone":["Tel\u00e8fon"]}},"comment":{"reference":"packages\/woocommerce-blocks\/build\/checkout-blocks\/billing-address--checkout-blocks\/shipping-address-frontend.js"}}
```

</details>

Aljullu:
Good catch with that @dinhtungdu! I have been investigating this a bit more, and couldn't figure out what's going on. For some reason the translation is included in the editor file, but not the frontend one. I opened an issue here: #6044. However, given that this seems to be a different issue and it's not a regression, how do you feel about not blocking this PR because of it?

mikejolley:
I think we can push ahead with this fix for 7.2. I found that the translation file with the string in core right now uses this file reference:

```
{"reference":"packages\/woocommerce-blocks\/build\/checkout-blocks\/billing-address--checkout-blocks\/shipping-address-frontend.js"}
```

This is not the path to the build file, so the md5 hash of the file path does not match meaning this translation is never loaded. It should fix itself once the paths and translations are updated. Maybe :)

mikejolley (approved):

mikejolley:
@Aljullu I made a small change just to ensure ->done is defined. Ill cherry pick this into 7.2.0.

PR #6049, opened 2022-03-14
Revert "Remove .gitattributes file"

mikejolley:
Reverts woocommerce/woocommerce-gutenberg-products-block#5242

While reverting this does "break" the source zip for releases, the larger issue is that this broke the composer packages from https://packagist.org/packages/woocommerce/woocommerce-blocks#v6.9.0, meaning WooCommerce core would bundle `/bin` and other development directories after installing the blocks package.

We need to revert this and decide on a better solution e.g. cleanup on the core side after installing the package.

cc @senadir

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,32 @@
+# Ignore all hidden files and directories
+/.* export-ignore
+
+# Ignore all markdown files
+/*.md export-ignore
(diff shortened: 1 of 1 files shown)


senadir (approved):
LGTM

PR #6050, opened 2022-03-14
Checkout: Missing country error for virtual orders (edge case)

mikejolley:
Prevents an error when placing a virtual order without a shipping address.

### Testing

Requires the following setup:

- WC settings > General - set default customer location to "no location by default"
- Create a downloadable, virtual product

Testing steps:

1. Go to the store as a logged out guest in incognito mode
2. Add the virtual/downloadable product to your cart
3. Go to checkout
4. Fill out billing address
5. Place order. It should succeed.
6. Check order details. Shipping should match billing address.

### Changelog

> Ensure shipping address is set for virtual orders to prevent missing country errors.

### DIFF
--- a/src/StoreApi/Routes/V1/Checkout.php
+++ b/src/StoreApi/Routes/V1/Checkout.php
@@ -393,21 +393,21 @@ class Checkout extends AbstractCartRoute {
 	private function update_customer_from_request( \WP_REST_Request $request ) {
 		$customer = wc()->customer;
 
-		if ( isset( $request['billing_address'] ) ) {
-			foreach ( $request['billing_address'] as $key => $value ) {
-				if ( is_callable( [ $customer, "set_billing_$key" ] ) ) {
-					$customer->{"set_billing_$key"}( $value );
-				}
+		// Billing address is a required field.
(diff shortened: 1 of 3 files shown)


src/StoreApi/Routes/V1/Checkout.php
```diff
@@ -393,21 +393,21 @@ private function create_or_update_draft_order( \WP_REST_Request $request ) {
...
-					$customer->update_meta_data( 'shipping_phone', $value );
-				}
+		// If shipping address (optional field) was not provided, set it to the given billing address (required field).
+		$shipping_address_values = $request['shipping_address'] ?? $request['billing_address'];
+
+		foreach ( $shipping_address_values as $key => $value ) {
+			if ( is_callable( [ $customer, "set_shipping_$key" ] ) ) {
+				$customer->{"set_shipping_$key"}( $value );
+			} elseif ( 'phone' === $key ) {
+				$customer->update_meta_data( 'shipping_phone', $value );
```
senadir, line 410: This seems to be a shared logic between this and `update-customer` router, so we might abstract at some point to avoid bugs like this.

senadir (approved):
LGTM

PR #6052, opened 2022-03-15
Fix: Mini Cart: Ability to choose the drawer behavior when user add products to cart

### LINKED ISSUE #6026: Mini Cart: Ability to choose the drawer behavior when user add products to cart
Currently, whenever users add a product to their cart, the Mini Cart drawer opens immediately. One disadvantage of the current behavior is it interrupts shopper flow if they might want to add multiple items to the cart.

An option allowing users to choose the behavior of Mini Cart when a product is added to the cart.

## Additional context
This was raised in a Slack discussion by @tomasztunik. It's also recently raised by @nerrad in a Monthly Walkthrough.

dinhtungdu:
Fixes #6026 

This PR:

- Adds a new attribute named ~`openDrawerOnAddedToCart` as a toggle~. I changed this option to `addedToCartBehavior` using Select Control in 2a842b1c40e86ec70e143a827c5f7a52c629f766.
- Only opens the drawer when the added to cart behavior is set to `Open the drawer`.
- Changes default behavior to **not** open the drawer when a product is added to the cart.

To the reviewers: '
- I'm wondering which default behavior makes more sense here. The current behavior is opening the drawer every time a product is added to the cart and it interrupts the shopping flow.
- The copy for options of the new attribute may be not so good, any feedback on that is more than welcome.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Add Mini Cart to the Header template part.
2. Open the Block Setting sidebar, see the new setting pane: `Mini Cart Settings`.
3. See it set to `Do not show anything` as default.
4. Save changes.
5. On a new tab, go to the site front end.
6. Add a product to the cart from the `All products` block. Don't see the drawer open.
7. Back to the Site Editor, change the behavior to `Open drawer`. Save changes.
8. On a new tab, go to the site front end.
9. Add a product to cart, see the drawer open.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.


assets/js/blocks/cart-checkout/mini-cart/block.tsx
```diff
@@ -108,8 +110,10 @@ const MiniCartBlock = ( {
 
 	useEffect( () => {
```
Aljullu, line 111: It looks like `addedToCartBehavior` needs to be added as a dependency of this `useEffect()`.

Aljullu (approved):
I'm very happy to get this before release, good job @dinhtungdu!

Code looks good and it's testing well, but we will need @vivialice's feedback in these two elements:

> I'm wondering which default behavior makes more sense here. The current behavior is opening the drawer every time a product is added to the cart and it interrupts the shopping flow.

Personally, I prefer not opening it. That's how it worked in Storefront. What are your thoughts, @vivialice?

> The copy for options of the new attribute may be not so good, any feedback on that is more than welcome.

Right, I will leave the final decision also to @vivialice. I would probably make the title of the panel a bit more specific (ie: "Mini Cart settings" → "Add to cart behavior"), and try to make the sentence above the select a bit shorter so it fits on the screen.

In case it helps, here is a screenshot:

dinhtungdu:
The failed E2E test is flaky and unrelated to this PR.

dinhtungdu:
@Aljullu updated the setting copy per your suggestion. Thanks a lot! I also change the attribute name from `addedToCartBehavior` to `addToCartBehavior` to match with the setting title.

> Personally, I prefer not opening it. That's how it worked in Storefront.

As mentioned in the PR description, the default behavior of the Mini Cart drawer changed to **NOT** open when a product is added to the cart.

vivialice:
Hi @dinhtungdu ! This looks great. 😄 

> Personally, I prefer not opening it. That's how it worked in Storefront. What are your thoughts, @vivialice?

I agree to keep the default as closed, however in Storefront there was confirmation visually with a banner. This still might be something we keep in mind for future iterations.

> The copy for options of the new attribute may be not so good, any feedback on that is more than welcome.

> Right, I will leave the final decision also to @vivialice. I would probably make the title of the panel a bit more specific (ie: "Mini Cart settings" → "Add to cart behavior"), and try to make the sentence above the select a bit shorter so it fits on the screen.

This is what I recommend for the copy: 

_Label Title:_ Add-to-Cart behaviour 
_Settings description:_ Select what happens when a customer adds a product to the cart.
_Options:_ Do nothing | Open cart drawer

Also it would be interesting to track which option is more popular, or how often users switch to the open drawer option. Are we planning on doing this already?

dinhtungdu:
@vivialice Thanks so much! I updated the copy using your suggestion.

> I agree to keep the default as closed, however in Storefront there was confirmation visually with a banner. This still might be something we keep in mind for future iterations.

I saw your design of the pop-up mini cart (which is similar to the Storefront mini cart style). I think we can change the default behavior to show the popup cart when it's introduced. That's also the reason why I change the `addToCartBehaviour` from a toggle to a dropdown to support more behaviors in the future.

> Also it would be interesting to track which option is more popular, or how often users switch to the open drawer option. Are we planning on doing this already?

I don't have any information about this. @Aljullu may have more insights.

Also, look like we got an agreement about the default behavior. So I will merge this PR when all workflows are passed.

PR #6065, opened 2022-03-21
Add Inner blocks to order summary

### LINKED ISSUE #6059: Convert Order Totals to Inner Blocks
Ref: https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5001#issuecomment-972845928

The blocks `woocommerce/cart-order-summary-block` and `woocommerce/checkout-order-summary-block` should be refactored to use Inner Blocks. 

Right now, they contain multiple TotalsWrapper components for each section, those sections being:

- Subtotal
- Total Fees
- Total Discount
- Coupons (and form)
- ExperimentalDiscountsMeta Slot
- Shipping totals (contains ExperimentalOrderShippingPackages slot)
- Tax totals
- Grand total
- ExperimentalOrderMeta

There is also a heading in the cart order summary block, and a list of items in the checkout order summary block.

Converting to Inner Blocks opens up several nice use cases such as:

- Moving coupon forms, or disabling them
- Rearranging how totals are displayed
- Hiding shipping from the cart

Some potential pitfalls:

- Removing a total could make the grand total confusing, if the other parts which made up this total were to be hidden.

## Inner block structure
Each totals block could have the same structure, similar to form steps, so you have:

- Heading (editable) and potentially sub-heading (like shipping row which shows the rate)
- Value
- An inner block area where extra content could be added

Going back to the Slot Fills and where they would fit:

- ExperimentalOrderShippingPackages would be deprecated and replaced with the inner block area within Shipping Totals. This area needs to receive similar props so extra shipping rows could be added. Styling/collapsing rates may be a problem that we need to workaround or remove.
- ExperimentalDiscountsMeta would be deprecated - an inner block can just be inserted between the other totals.
- ExperimentalOrderMeta is deprecated - an inner block can just be inserted below the other totals.

## Next steps
I don't believe these slot fills need to graduate. Let's instead work on inner blocks.

 - [x] Investigate new inner blocks per total and create POC
 - [ ] Export reusable totals block for others to consume
 - [ ] Update subs/points and rewards with new structure to validate the approach
 - [ ] Deprecate Slot Fills with no replacement.

mikejolley:
The WIP prefix is a leftover, the work on this branch was completed.

This PR converts the Cart Block and Checkout Block "order summary" areas into an inner block structure. 

Cart Block:

- Heading `woocommerce/cart-order-summary-heading`
- Subtotal `woocommerce/cart-order-summary-subtotal`
- Fee `woocommerce/cart-order-summary-fee`
- Shipping `woocommerce/cart-order-summary-shipping`
- Discount `woocommerce/cart-order-summary-discount`
- Coupon Form `woocommerce/cart-order-summary-coupon-form`
- Taxes `woocommerce/cart-order-summary-taxes`

Checkout Block

- Cart Items `woocommerce/checkout-order-summary-cart-items`
- Subtotal `woocommerce/checkout-order-summary-subtotal`
- Fee `woocommerce/checkout-order-summary-fee`
- Shipping `woocommerce/checkout-order-summary-shipping`
- Discount `woocommerce/checkout-order-summary-discount`
- Coupon Form `woocommerce/checkout-order-summary-coupon-form`
- Taxes `woocommerce/cart-order-summary-taxes`

**NOTE:** Yes there are many similarities between the cart and checkout inner blocks, however, they cannot be shared due to the way we use context providers. If they were shared, both Cart and Checkout Blocks would import the above inner blocks, and those inner blocks would then receive the wrong values from context depending on what was imported first. To work around this we have a unique set of inner blocks per each parent block.

The new inner blocks implement existing slotfills, making this change backwards compatible. Subscriptions extension can be installed to verify this since it extends the order totals area including its own shipping and total row.

Fixes #6059

#### Cart block before

**The Order summary block**

Screenshot 2022-03-29 at 15 47 30

#### Cart block after

**The Order summary block**
Screenshot 2022-03-29 at 15 48 20

**Selection of an Order summary Inner block**
Screenshot 2022-03-29 at 15 48 32

#### Automated Tests

Existing block tests cover these inner blocks since the functionality has just moved down a level.

  * [x] Unit tests
  * [x] E2E tests
  
### Manual Testing

1. Go to a page and add the Checkout block.
2. Go to a page and add the Cart block
3. In the editor, inside each block, select the Order summary and notice that you can select the following inner blocks:

| Cart | Checkout |
| --------------- | --------------- | 
|  Heading, Subtotal, Cart Items, Fees, Coupon Form, Shipping, Taxes | Subtotal, Cart Items, Fees, Coupon Form, Shipping, Taxes | 
4. Try to move some of the inner blocks and save. Confirm that in the Cart block only Order Summary Heading and the Coupon form inner blocks can be removed, and for the Checkout block only the Coupon form.
5. Check on the website that your changes are reflected. Note that Taxes, Fees and Coupons will appear on the website only if the shop supports them.
6. For the Cart > Shipping inner block make sure you can enable disable the shipping calculator and that the change is visibile in the website
7. Enable Taxes on your website (`/wp-admin/admin.php?page=wc-settings`) by clicking  `Enable tax rates and calculations`
8. Go to the C & C blocks and select the Taxes inner block. 
9. In the inner block's setting toggle on / of `Show rate after tax name` and make sure that is reflected in the website
10. Try to place an order and make sure it's successful.

### User Facing Testing

* [x] Same as above, or
* [ ] See steps below.

### Changelog

> The order summary area for the Cart and Checkout Blocks is now powered by Inner Blocks allowing for more customizations and extensibility.


assets/js/blocks/cart/inner-blocks/index.tsx
```diff
@@ -5,8 +5,15 @@ import './filled-cart-block';
...
 import './empty-cart-block';
 import './cart-accepted-payment-methods-block';
+import './cart-order-summary-block';
+import './cart-order-summary-subtotal';
+import './cart-order-summary-fee';
+import './cart-order-summary-discount';
+import './cart-order-summary-shipping';
+import './cart-order-summary-coupon-form';
+import './cart-order-summary-taxes';
+import './cart-order-summary-heading';
```
ralucaStan, line 19: Minor, but for consistency reasons these should also have `-block` suffix. `./cart-order-summary-heading-block` instead of `./cart-order-summary-heading`

assets/js/blocks/cart/editor.scss
```diff
@@ -29,3 +29,15 @@ body.wc-lock-selected-block--remove {
...
+
+.wp-block-woocommerce-cart-order-summary-block {
+	.block-editor-block-list__layout > div {
+		margin: 0 !important;
+	}
+	.wc-block-components-totals-item,
+	.wc-block-components-totals-coupon,
+	.wc-block-components-order-summary,
+	.wc-block-components-totals-shipping {
+		box-sizing: border-box;
```
ralucaStan, line 39: Before we had the `.wc-block-components-totals-wrapper` class for each item of the summary.
Now:
- we use the `.wc-block-components-totals-wrapper` class only for slots
- we have `.wc-block-components-totals-item` for some of the items 
- summary, coupon and shipping don't have the `.wc-block-components-totals-item` and it needs to be styled separately.

I think it was good to have one class for each of this summary items. I wonder why:

- we didn't keep the `.wc-block-components-totals-wrapper`? 
- we don't add `.wc-block-components-totals-item` to coupon, summary and shipping as well?

assets/js/base/components/cart-checkout/totals/footer-item/index.tsx
```diff
@@ -30,6 +31,8 @@ export interface TotalsFooterItemProps {
 	 * convenience, but will use only these two properties.
 	 */
 	values: LooselyMustHave< CartResponseTotals, 'total_price' | 'total_tax' >;
+	// CSS classname.
```
ralucaStan: let's remove this

assets/js/blocks/cart/inner-blocks/cart-order-summary-coupon-form/block.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+	const { applyCoupon, isApplyingCoupon } = useStoreCartCoupons();
+
+	if ( ! couponsEnabled ) {
+		return null;
+	}
+
+	return (
+		<TotalsCoupon
+			className={ className }
```
ralucaStan: The `Additional CSS class(es)` attribute doesn't work for this block.

assets/js/blocks/cart/inner-blocks/cart-order-summary-heading/attributes.ts
```diff
@@ -0,0 +1,18 @@
...
+export default {
+	content: {
+		type: 'string',
+		default: __( 'Heading', 'woo-gutenberg-products-block' ),
+	},
+	lock: {
+		type: 'object',
+		default: {
+			remove: false,
+			move: false,
```
ralucaStan, line 15: Should we allow to move/remove this considering the corresponding title in the left side is not an inner block? Maybe we can enable this once that becomes an inner title as well?
Screenshot 2022-03-29 at 19 12 12
I

assets/js/blocks/cart/inner-blocks/cart-order-summary-coupon-form/block.tsx
ralucaStan: I also wonder why we don't show the `.wp-block-woocommerce-cart-order-summary-coupon-form-block` class like for the other inner blocks `.wp-block-woocommerce-cart-order-summary-NAME-block`.

assets/js/blocks/cart/inner-blocks/register-components.ts
```diff
@@ -0,0 +1,185 @@
...
+	component: lazy( () =>
+		import(
+			/* webpackChunkName: "cart-blocks/empty-cart" */
+			'./empty-cart-block/frontend'
+		)
+	),
+} );
+
+registerCheckoutBlock( {
+	metadata: metadata.FILLED_CART,
```
ralucaStan: this is registered a second time, it should be removed.

ralucaStan (commented):
Posting a partial review comment for the Cart Inner blocks, I'll still need to check the Checkout inner blocks

assets/js/blocks/checkout/inner-blocks/index.tsx
```diff
@@ -13,3 +13,10 @@ import './checkout-order-summary-block';
 import './checkout-payment-block';
 import './checkout-express-payment-block';
 import './checkout-shipping-methods-block';
+import './checkout-order-summary-subtotal';
+import './checkout-order-summary-fee';
+import './checkout-order-summary-discount';
+import './checkout-order-summary-shipping';
+import './checkout-order-summary-coupon-form';
+import './checkout-order-summary-taxes';
+import './checkout-order-summary-cart-items';
```
ralucaStan, line 22: these should also end with `-block`

assets/js/blocks/checkout/styles/editor.scss
```diff
@@ -72,3 +72,15 @@ body.wc-lock-selected-block--remove {
...
+.wp-block-woocommerce-checkout-order-summary-block {
+	.block-editor-block-list__layout > div {
+		margin: 0 !important;
+	}
+	.wc-block-components-totals-item,
+	.wc-block-components-totals-coupon,
+	.wc-block-components-order-summary,
+	.wc-block-components-totals-shipping {
+		box-sizing: border-box;
+	}
```
ralucaStan, line 82: same comment about adding one common class for all section wrappers from the Order summary column

assets/js/blocks/checkout/inner-blocks/checkout-order-summary-coupon-form/block.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+	const { applyCoupon, isApplyingCoupon } = useStoreCartCoupons();
+
+	if ( ! couponsEnabled ) {
+		return null;
+	}
+
+	return (
+		<TotalsCoupon
+			className={ className }
```
ralucaStan: The coupon form wrapper does not apply the additional class provided by the user.

ralucaStan (changes requested):
I've also added my comments about the Checkout inner blocks.

All and all this works as expected. There are some layout shifts when selecting the inner blocks, but I will create a separate issue about this.

I've added some comments that need to be addressed.

ralucaStan:
It was a real effort reviewing this as I imagine was the case for writing this. 
It would have helped with testing if we would have had separate PRs for Cart and Checkout innerblocks. In this case the decisions made for the first block would have been applied to the 2nd block.

assets/js/blocks/cart/inner-blocks/cart-order-summary-coupon-form/block.tsx
ralucaStan: both fixed

assets/js/blocks/cart/inner-blocks/cart-order-summary-heading/attributes.ts
ralucaStan: After internal discussions it was decided that  we can leave it since it’ll be up to the user to remove it. And he can add a custom CSS to add some margin to fix the alignment issue.

ralucaStan:
I've added two commits to address the changes requests [10670393a00ebd18d12b86755b5278cdfa49ace6](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6065/commits/10670393a00ebd18d12b86755b5278cdfa49ace6) and https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6065/commits/d654f5b074932b68a218ae5fc319d1b4444133a4:

### Add back the wc-block-components-totals-wrapper class that was used for each segment in the totals Order summary

Because, removing them was:
-  a breaking change for the old structure
- was making it harder to target the inner blocks. Before the class was used to target each segment
- it was making the wc-block-components-totals-item behave as a child or parent depending on the inner block, inconsitency

### Reuse the TotalsWrapper component for C& C blocks inner blocks
This component was removed in this PR, but  we wrap components in the Cart and Checkout sidebar in a TotalsWrapper. This will ensure consistent spacing and borders are applied to items in the sidebar.

PR #6069, opened 2022-03-21
Product Query: allow adding more arguments by applying a filter.

cpapazoglou:
While working on 12894-gh-Automattic/woocommerce.com I have noticed that there aren't any hooks allowing us to extend what arguments can be passed to the `wc/store/products/` API. This PR aims at allowing code outside of this repo to hook into `$query_args` before it gets passed to `WP_Query();`

https://github.com/Automattic/woocommerce.com/blob/476e2a09a2b6e9ff175b9230054ce6ad6dd871a6/plugins/woo-gutenberg-products-block/src/StoreApi/Utilities/ProductQuery.php#L242-L243

#### Accessibility

#### Other Checks

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

Perhaps it will be easier testing this 12894-gh-Automattic/woocommerce.com

### User Facing Testing
* [ ] Same as above, or
* [ ] See steps below.

1.
2.
3.

### Changelog

> Adds `store_api_product_query_args` filter to allow extending the `$query_args` before it gets passed to `WP_Query();`

### DIFF
--- a/bin/hook-docs/data/filters.json
+++ b/bin/hook-docs/data/filters.json
@@ -952,6 +952,42 @@
             "args": 3
         },
         {
+            "name": "woocommerce_store_api_product_query_args",
+            "file": "StoreApi/Utilities/ProductQuery.php",
+            "type": "filter",
+            "doc": {
+                "description": "Filters the arguments for the products WP_Query.",
+                "long_description": "",
+                "tags": [
+                    {
+                        "name": "param",
+                        "content": "WP_Query args.",
+                        "types": [
+                            "array"
+                        ],
+                        "variable": "$args"
+                    },
+                    {
+                        "name": "param",
+                        "content": "Request object.",
+                        "types": [
+                            "\\WP_REST_Request"
+                        ],
+                        "variable": "$request"
+                    },
+                    {
+                        "name": "return",
+                        "content": "",
+                        "types": [
+                            "array"
+                        ]
+                    }
+                ],
+                "long_description_html": ""
+            },
+            "args": 2
+        },
+        {
             "name": "woocommerce_variation_option_name",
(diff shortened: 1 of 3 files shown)


senadir:
Hey @cpapazoglou! We don't use filters for Store API and prefer to use extensibility methods that only append data to the response, this is to ensure stability in our requests and ensure a consistent, predictable, schema, something we can't do with filters.

Can you walk me through your case, we might be able to provide alternative solutions that still fulfil that. If not, you can follow this file to add `ExtendSchema` to that route:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/extend-rest-api-new-endpoint.md

cpapazoglou:
> Hey @cpapazoglou! We don't use filters for Store API and prefer to use extensibility methods that only append data to the response, this is to ensure stability in our requests and ensure a consistent, predictable, schema, something we can't do with filters.
> 
> Can you walk me through your case, we might be able to provide alternative solutions that still fulfil that. If not, you can follow this file to add `ExtendSchema` to that route: https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/docs/extensibility/extend-rest-api-new-endpoint.md

👋 @senadir!
  In our case, we want to filter the `/wc/store/products` results based on a new taxonomy. How could we extend that endpoint through the `ExtendSchema` so that it only returns products with a specific term of that new taxonomy? An example request would be `https://woocommerce.test/wp-json/wc/store/products?wpcom_marketplace_categories=[that-category-id]`

cpapazoglou:
@senadir I will be closing this in favour of https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6152

PR #6078, opened 2022-03-22
Stop showing the `PricesSlider` skeleton when moving the slider handles

### LINKED ISSUE #5395: `PriceSlider` turns into a skeleton when loading
The `PriceSlider` component disappears and turns into its skeleton form in its loading state. Specifically, in the context of the `Filter by price` block, the loading state is whenever the change event is fired as the user releases one of the handles.

## Video

https://user-images.githubusercontent.com/1847066/146248331-a04fb31b-01ce-4270-8d63-19429f5d9bc1.mov

## Proposal

This issue was [discussed on Slack](p1639426466072100-slack-C8X6Q7XQU) and it was agreed that it is an unexpected behavior. The proposed solutions are:

1. On change, while the connected block is loading, disable the input.
2. This, as pointed out by @vivialice, might be a problem when it takes long to load the products. And, I might add, perhaps the user just released the slider by mistake and now has to wait. So, perhaps, a better solution would be to not do anything at all to the input when the handle is released. The loading state is clearly shown by the fact that the linked block will get the loading state, and in order to reduce network requests, we can just debounce or throttle them.

albarin:
The changes on this PR make the `PriceSlider` component only turn into its skeleton form when it is actually loading for the first time (loading & disabled). Before, it was turning into the skeleton also when the user moved and released one of the sliders.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5395

Screen Capture on 2022-03-22 at 15-30-22

### Testing

1. Create a new page with the "All Products" and the "Filter Products by Price" blocks.
2. Go to the page on the front and check that when the `PriceSlider` loads it shows the skeleton like: Screenshot 2022-03-22 at 15 35 15
3. After it finishes loading, move the slider handles and check it does not show the skeleton on the slider again.

### Changelog

> Stop showing the price slider skeleton when moving the slider handles.

### DIFF
--- a/assets/js/base/components/price-slider/style.scss
+++ b/assets/js/base/components/price-slider/style.scss
@@ -63,8 +63,7 @@
 		}
 	}
 
-	&.is-loading,
-	&.is-disabled {
+	&.is-loading.is-disabled {
 		.wc-block-components-price-slider__range-input-wrapper,
 		.wc-block-components-price-slider__amount,
 		.wc-block-components-price-slider__button {
@@ -267,8 +266,7 @@
 	}
 
 	.wc-block-components-price-slider {
-		&.is-loading,
-		&.is-disabled {
+		&.is-loading.is-disabled {
 			.wc-block-components-price-slider__range-input-wrapper {
 				@include placeholder();
 				box-shadow: none;


sunyatasattva (approved):
Hahah, after all the digging we did on this issue, I kinda thought it was more than this change, but I was so surprised to just see this. Funny! But great job and we learned a lot!

Approving and also merging it in for you by skipping the required CI steps. The failed E2E is due to something unrelated to this and to the newest Gutenberg version causing a regression. A fix might be available here: https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6080

PR #6080, opened 2022-03-22
Fix site editor E2E tests for Gutenberg latest

### LINKED ISSUE #6071: E2E tests failing with the last version of Gutenberg
With the last version of Gutenberg, some tests related to `site-editing-templates.js` file failed. This is the error:

```
TimeoutError: waiting for selector `.edit-site-list-table` failed: timeout 30000ms exceeded
```
image

When the browser visited `wp-admin/site-editor.php?postType=wp_template`, Gutenbergs shows this error:

```
The editor is unable to find a block template for the homepage.
``` 

image

tomasztunik:
This makes the E2E tests on latest Gutenberg to pass again.

- Add handling of editor context via `GUTENBERG_EDITOR_CONTEXT` env var. It's set on CI to 'gutenberg' to ensure we are using `themes.php` admin page to interface with templates as because of the #6071 regression it's not possible to use site-editor.php for installed gutenberg and core gutenberg using the same URL. We already had `themes.php` and `site-editor.php` handling in place but it was not used and had a bug in it. Adding the context now makes use of this and fixed the bug that was hidden there.
- Ensure `saveTemplate` util completes by waiting for successful response before handing over control to follow up steps
- ~Updates wordpress/e2e-test-utils package~ ⚠️ rolled back the update to 7.0.2 - we are stuck on 6.0.x until further testing. It is causing test flakiness around usage of `disableSiteEditorWelcomeGuide`

TODO for future after WordPress v6 gets released we should review if we can run all the tests just using `site-editor.php` and drop the  `GUTENBERG_EDITOR_CONTEXT` usage.

Fixes #6071

#### Automated Tests
  * [x] E2E tests
  
### Manual Testing

1. E2E tests should be passing on CI for latest Gutenberg and Gutenberg included in WP core


tests/e2e/specs/backend/site-editing-templates.test.js
```diff
@@ -167,6 +170,8 @@ describe( 'Store Editing Templates', () => {
 			await visitTemplateAndAddCustomParagraph( 'single-product' );
 
 			await goToSiteEditor( '?postType=wp_template' );
+			// we need to wait for the selctor to show up, sometimes the loading is delayed and test becomes falky
```
sunyatasattva: There are a few typos in this comment:

`selctor` and `falky`

tests/e2e/specs/backend/site-editing-templates.test.js
```diff
@@ -167,6 +170,8 @@ describe( 'Store Editing Templates', () => {
...
+			await page.waitForSelector( SELECTORS.templates.templateActions );
```
sunyatasattva, line 174: In the interest of separating business logic from testing (like I mentioned in the E2E workshop), it'd be better to move this line within `goToSiteEditor`. Basically, it's `goToSiteEditor` responsibility to make sure we are in the right place and give back control to the program only when we are (and handle errors when we aren't).

tests/e2e/specs/backend/site-editing-templates.test.js
```diff
@@ -264,6 +268,7 @@ describe( 'Store Editing Templates', () => {
 			await visitTemplateAndAddCustomParagraph( 'archive-product' );
 
 			await goToSiteEditor( '?postType=wp_template' );
+			await page.waitForSelector( SELECTORS.templates.templateActions );
```
sunyatasattva, line 271: Same as above

tests/e2e/specs/backend/site-editing-templates.test.js
```diff
@@ -365,6 +369,7 @@ describe( 'Store Editing Templates', () => {
 			await visitTemplateAndAddCustomParagraph( 'taxonomy-product_cat' );
 
 			await goToSiteEditor( '?postType=wp_template' );
+			await page.waitForSelector( SELECTORS.templates.templateActions );
```
sunyatasattva, line 372: Hehe, another reason to include this in `goToSiteEditor` → Code will be DRYer

tests/e2e/specs/backend/site-editing-templates.test.js
```diff
@@ -460,6 +464,7 @@ describe( 'Store Editing Templates', () => {
 			await visitTemplateAndAddCustomParagraph( 'taxonomy-product_tag' );
 
 			await goToSiteEditor( '?postType=wp_template' );
+			await page.waitForSelector( SELECTORS.templates.templateActions );
```
sunyatasattva, line 467: 😬

tests/e2e/utils.js
```diff
@@ -143,21 +144,34 @@ export const isBlockInsertedInWidgetsArea = async ( blockName ) => {
  *
  * There are two different possible site editor pages:
  *
- * 1. `themes.php?page=gutenberg-edit-site` is the one used and available if the Gutenberg plugin is enabled.
- * 2. `site-editor.php` is the one available in WP Core.
+ * 1. `themes.php?page=gutenberg-edit-site` this is legacy editor access used for wp <=5.8 compatibility
+ * 	  themes.php is not accessible in 5.9+ when local Gutenberg is active.
```
sunyatasattva: Missing some backticks for consistency around `themes.php`. But, more importantly, I am not sure I understand this comment, could you clarify it a bit?

So `themes.php?page=gutenberg-edit-site` seems to be the way to go whenever Gutenberg is installed on <6.0, right? At least that's what I get by reading your code. The comment says `themes` is not accessible in 5.9+ when local Gutenberg is active, and I'm not sure I understand.

tests/e2e/utils.js
```diff
@@ -179,8 +193,13 @@ export async function saveTemplate() {
 
 	await page.click( saveButton );
 	await page.waitForSelector( savePrompt );
+
 	await page.click( confirmSave );
 	await page.waitForSelector( `${ saveButton }[aria-disabled="true"]` );
+	await page.waitForResponse( ( res ) => {
```
sunyatasattva, line 198: Very excellent solution! 👌 +100 for this!

sunyatasattva (changes requested):
Nothing major, code looks excellent, just some small refactoring and you're good to go :)

tests/e2e/specs/backend/site-editing-templates.test.js
tomasztunik: But these wouldn't be there always. It's not a generic check. These would appear only if there were any edited templates available. Should we provide "states" along the path to distinguish what to wait for? We'd probably need another helper ie. goToSiteEditorTemplates(target: 'templates' | 'parts', hasEdits: boolean) or something - maybe then it would make sense. 

Should we make it a separate issue to extract these as a separate helpers and think bit more about what utils we actually need there? These page objects would be great for this!

(resolved others for clarity)

tests/e2e/utils.js
tomasztunik: Argh, was tweaking text and made error - should be site-editor is not accessible! 

Will reword it :)

tests/e2e/specs/backend/site-editing-templates.test.js
tomasztunik: falky selctors, wasn't this a thing?

tests/e2e/specs/backend/site-editing-templates.test.js
tomasztunik: Or maybe if we were extracting this to a separate hellper util we could use the `waitForRespponse` to ensure the templates data was loaded like I did it with the `saveTemplate`? Templates data is loaded async after the page was loaded first so we could improve it like so - should make it much more stable and maybe allow us to get rid of the try-catch?

sunyatasattva: Alright, because I know this is a blocker right now for many people, let's merge it. I'd be thankful if you created a new issue so we won't forget about this. Maybe link this thread. 🙏

sunyatasattva (approved):
Approved (with reservation 😬 ). It is a blocker for many people so this solution is ok for now. I'm not fond of the repetition and the mixing of business logic and test logic, but we'll take care of it in another issue.

PR #6086, opened 2022-03-23
Mini Cart: Fix footer buttons hover style

### LINKED ISSUE #6085: Mini Cart: Footer buttons: colors setting doesn't apply to hover state
## Describe the bug

The style for hover state of Cart and Checkout buttons still use the default colors regardless of the colors set in the Mini Cart Contents block.

## To reproduce

1. Edit the Mini Cart Content Blocks.
2. Change the text and background colors.
3. Save. Go to the front end.
4. Add a product to open the Mini Cart Drawer.
5. Hover on the Cart and Checkout buttons to see they're still using the default color for the hover state.

## Expected behavior

Custom color settings should apply to both the normal and hover states of the buttons.

## Screenshots

image

dinhtungdu:
Fixes #6085

This PR applies the custom color settings to the hover state of the footer buttons.

### Screenshots

image

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Edit the Mini Cart Content Blocks.
2. Change the text and background colors.
3. Save. Go to the front end.
4. Add a product to open the Mini Cart Drawer.
5. Hover on the Cart and Checkout buttons to see the hover style uses new color settings.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.


src/BlockTypes/MiniCartContents.php
```diff
@@ -95,13 +95,17 @@ protected function enqueue_assets( array $attributes ) {
...
+						'value'    => $text_color ? "inset 0 0 0 1px {$text_color['value']}" : false,
 					),
 				),
 			),
 			array(
-				'selector'   => '.wc-block-mini-cart__footer .wc-block-mini-cart__footer-actions .wc-block-mini-cart__footer-checkout',
+				'selector'   => array(
+					'.wc-block-mini-cart__footer .wc-block-mini-cart__footer-actions .wc-block-mini-cart__footer-checkout',
+					'.wc-block-mini-cart__footer .wc-block-mini-cart__footer-actions .wc-block-mini-cart__footer-checkout:hover',
+					'.wc-block-mini-cart__footer .wc-block-mini-cart__footer-actions .wc-block-mini-cart__footer-cart.wc-block-components-button:hover',
```
Aljullu: Is the extra class (`.wc-block-components-button`) for specificity reasons or it can be removed?

src/BlockTypes/MiniCartContents.php
```diff
@@ -122,6 +126,8 @@ protected function enqueue_assets( array $attributes ) {
 		$parsed_style = '';
 
 		foreach ( $styles as $style ) {
+			$selector = is_array( $style['selector'] ) ? implode( ', ', $style['selector'] ) : $style['selector'];
```
Aljullu: Nitpicking, but if we want to save some characters, we can remove the space here...
```suggestion
			$selector = is_array( $style['selector'] ) ? implode( ',', $style['selector'] ) : $style['selector'];
```

src/BlockTypes/MiniCartContents.php
```diff
@@ -130,7 +136,7 @@ function( $property ) {
 			);
 
 			if ( ! empty( $properties ) ) {
-				$parsed_style .= $style['selector'] . '{' . PHP_EOL;
+				$parsed_style .= $selector . '{' . PHP_EOL;
```
Aljullu: ...and remove `PHP_EOL` here and the lines below.

I think we don't need to nice-format CSS if we are printing it directly into the page and instead should prioritize making it shorter. But not a big deal, so leaving the final decision to you.

Aljullu (commented):
Nice fix! I think besides the `:hover` selectors, we should also add `:focus`, so styles are applied when navigating with the keyboard.

src/BlockTypes/MiniCartContents.php
dinhtungdu: I added it to override the core style. But we don't need it this time because we're using internal style which has a higher priority. I removed that class. Thanks so much for the head-up!

Aljullu: Nice, thanks! Can it be removed it also from line 91?

https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2036eb6adeafd878a79ba306bdc17d1ff30698bf/src/BlockTypes/MiniCartContents.php#L91

Aljullu (approved):
LGTM, I left one last comment, but shouldn't be blocking.

PR #6087, opened 2022-03-23
Mini Cart: Style doesn't apply to quantity input

### LINKED ISSUE #6029: Mini Cart: Style doesn't apply to quantity input
## To reproduce

1. Edit Mini Cart Template part.
2. Set the text, link, and background color for the Mini Cart Contents block.
3. See the color doesn't apply to the quantity input in the editor.
4. Save. See the issue on the front end.

## Expected behavior

Quantity input should be styled as other elements.

## Screenshots

image

## Additional context

Thank @nerrad to report this issue!

dinhtungdu:
Fixes #6029

This PR styles the quantity input using the text color set in the Mini Cart Contents block. The style in the editor and frontend are handled differently. I added a comment explaining the reason for the difference in code.

### Screenshots

**Editor:**
image

**Frontend:**
image

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Edit Mini Cart Contents block.
2. Set the text and background color.
3. See the input style updated accordingly.
4. Save changes.
5. See the style applied on the front-end as well.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.


assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
```diff
@@ -81,6 +82,26 @@ const Edit = ( { clientId }: Props ): ReactElement => {
...
+				{ `
+				.wc-block-mini-cart__products-table .wc-block-components-quantity-selector {
+					border-color: ${ blockProps.style.color };
+				}
+				.wc-block-mini-cart__products-table input.wc-block-components-quantity-selector__input,
+				.wc-block-mini-cart__products-table .wc-block-components-quantity-selector .wc-block-components-quantity-selector__button {
+					color: ${ blockProps.style.color };
+				}
+			` }
+			</style>
```
dinhtungdu: Using `<style>` here is not the best approach. If anyone has a better idea, please let me know!

dinhtungdu:
[0fcff01](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6087/commits/0fcff01262b4f8f1ca78da3e6d50497e35c48fe9) refactor the footer button styling by using new `<style>` tag introduced in this PR.

Aljullu:
Thanks for working on this, @dinhtungdu! I noticed the input field border color is now more opaque than before. You can compare it between Mini Cart and Cart (I set the background to `#fff` and the text color to `#000`, this is how it looks):

Mini Cart | Cart
--- | ---
imatge | imatge

Could we make it so it looks the same? Maybe changing the border with a `box-shadow` might help?

Regarding this comment in the code, I have mixed feelings:

> This is a workaround to style inner blocks using the color settings of the Mini Cart Contents block. This is needed because we reuse the <CartLineItemsTable/> component from the Cart block and we don't want to change anything to that component.

Personally, I think if we do this changes to the Mini Cart block, we should try to do the same to the Cart block. I know that block doesn't have global styles so it's a bit different, but I wonder if we could play a bit with `inherit` and `currentColor` in CSS. (I didn't experiment it, but it _might_ work?).

We might even be able to get rid of the `Dark mode inputs` setting in the Cart block?

cc'ing @senadir and @ralucaStan for your thoughts on this. Your team maintains the Cart block so you have the last word, but I think it would be better to keep Mini Cart & Cart blocks aligned when possible.

dinhtungdu:
> I noticed the input field border color is now more opaque than before. You can compare it between Mini Cart and Cart (I set the background to #fff and the text color to #000, this is how it looks):

@Aljullu Thanks so much for the review! I should have documented this behavior in the PR description. This is expected behavior and I decided to style the input that way because:
- (Note that dynamic style is only applied when the text color is set).
- We have text, background, and link color settings. The border is still experimental. For the quantity input, we use text and background color for styling.
- The quantity input uses `#dddddd` for the border color. I decided to use the text color for the border. Ideally, we should use the text color with opacity, but the color can be a hex color but also can be a CSS variable. So it isn't possible to add opacity for the text color.
- We can set the opacity for the quantity input, but it will apply to the amount and buttons inside.
- When the input is focused, originally, we apply a `#f0f0f0` background. I disable the focus state background when the text color is set.
- Minus and plus buttons use different colors for normal and hover states. When text color is set, I use opacity to replicate the visual look.

dinhtungdu:
@Aljullu I also updated the comment to better explain the reason for `<style>`:

```js
			{ /**
			 * This is a workaround to style inner blocks using the color
			 * settings of the Mini Cart Contents block. It's possible to get
			 * the Mini Cart Contents block's attributes inside the inner blocks
			 * components, but we have 4/7 inner blocks that inherit style from
			 * the Mini Cart Contents block, so we need to apply the styles here
			 * to avoid duplication.
			 *
			 * Another reason for this is we reuse the <CartLineItemsTable/>
			 * component from the Cart block and we don't want to change
			 * anything to that component.
			 *
			 * We only use this hack for the Site Editor. On the frontend, we
			 * manipulate the style using block attributes and inject the CSS
			 * via `wp_add_inline_style()` function.
			 */ }
```

dinhtungdu:
> Could we make it so it looks the same? Maybe changing the border with a `box-shadow` might help?

@Aljullu Thanks so much for your suggestion! In 11a2470ad92e4a311dbbbe0a726b111d6cc89518, I use `box-shadow` to 'reduce' the opacity of the border color to replicate the visual look of the original style.

Update: They don't look the same. :(

image

Update 2: My best result:

image

Note: The original input is on the right.

dinhtungdu:
The failed E2E test is flaky and unrelated to this PR.

Aljullu (commented):
Thanks for updating the PR, @dinhtungdu!

There are still a couple of things that I think we could do:

1. I wonder if in all cases where we use `${ blockProps.style.color }` (JS-side) and `{$text_color['value']}` (PHP-side), we could simply use [`currentColor`](https://css-tricks.com/currentcolor/). If we do that, we can then move that code to CSS files. That means it will always be applied instead of only when global styles are defined, but IMO that's ok.
2. And if we go that way, then I would update the base component so these new styles apply to the Cart block as well. This way we won't have to maintain two different styles for Mini Cart and Cart block.

How does it sound?

dinhtungdu:
> How does it sound?

@Aljullu It sounds briliant! I haven't thought about using `currentColor` that way before (I'm sorry I missed its mention in your previous review). 

For components that only use text color and doesn't have `color` property, this approach works surprisingly well. 

For some that have `color` value, it doesn't work. Take the checkout button as an example:

Screen Shot 2022-03-25 at 23 42 15

In d9ea7492ea10eee388fd439ddb3ec16474012c0d, I refactored the style using `currentColor` when possible.

dinhtungdu:
In dbce312138eddad7d6c9c194d1444618a005fbc5 I added style for shopping button. It's a small change so I included it in this PR.

dinhtungdu:
> That means it will always be applied instead of only when global styles are defined, but IMO that's ok.

For the Mini Cart Contents, the global style won't work with our current approach on styling the inner blocks. This can be solved using a custom selector, but it's still an experimental feature. There is only one Mini Cart template part so the block style of Mini Cart Contents is the global style too.

> And if we go that way, then I would update the base component so these new styles apply to the Cart block as well. This way we won't have to maintain two different styles for Mini Cart and Cart block.

I agree to update the style of the Cart inner blocks to make them dynamic by nature. We may want to do that in a dedicated PR.

dinhtungdu:
Converting this PR to draft to work on refactoring the `<CartLineItemsTable/>` instead of overriding its style in Mini Cart Contents.

dinhtungdu:
Close in favor of #6167

PR #6096, opened 2022-03-23
Filter Products by Attribute: Fix dropdown search case sensitivity

### LINKED ISSUE #5718: Filter Products by Attribute block in Dropdown display mode fails search because of input case sensitivity handling
## Describe the bug

Searching on a mobile browser (Chrome and Safari) in the **Filter by Attribute** block the selections disappear as soon as you start typing.

## To reproduce

1. Visit [a page with a **Search by Attribute**](https://luanwootesting.mystagingwebsite.com/custom-shop/) block on an mobile device with the Chrome or Safari browser.
2. Start typing in the search field.

## Expected behavior

The search results should display and be clickable.

## Screen recordings of the issue

Safari (iPhone): https://d.pr/v/JROW4M
Chrome (Android): https://d.pr/v/jSdrRb

* Core version:  5.9
* WooCommerce version: 6.0.0
* WooCommerce Blocks version:  6.3.3

## Additional context

This issue isn't present in a mobile Firefox browser.

Video: https://d.pr/v/mYCucQ

danieldudzic:
Fix dropdown search case-sensitivity handling for the Filter Products by Attribute block.

Fixes #5718 

### Screenshots
|Before|After|
|-|-|
|before|after|

### Manual Testing

2. Activate a **block theme**, like Twenty Twenty Two
3. Create a new page, and add the **Filter Products by Attribute** block
4. Change the **Display Style** to **Dropdown**
5. In the **Filter Products by Attribute** section select the attribute you want to be filtering by (in my case 'Size')
6. Test the page on the front-end by searching for product attributes
7. Make sure that relevant values are being found for both Uppercase and Lowercase input

### User Facing Testing

- [x] Same as above

### Changelog

> Filter Products by Attribute: Make dropdown search case sensitive.

### DIFF
--- a/assets/js/base/components/dropdown-selector/index.js
+++ b/assets/js/base/components/dropdown-selector/index.js
@@ -31,6 +31,7 @@ import './style.scss';
  * @param {boolean} props.multiple Whether multi-select is allowed.
  * @param {function():any} props.onChange Function to be called when onChange event fires.
  * @param {Array} props.options The option values to show in the select.
+ * @param {boolean} props.isCaseSensitive Whether the dropdown search should be case-sensitive.
  */
 const DropdownSelector = ( {
 	attributeLabel = '',
@@ -43,6 +44,7 @@ const DropdownSelector = ( {
 	multiple = false,
 	onChange = () => {},
 	options = [],
+	isCaseSensitive = false,
 } ) => {
 	const inputRef = useRef( null );
 
@@ -182,13 +184,20 @@ const DropdownSelector = ( {
 							getItemProps={ getItemProps }
 							getMenuProps={ getMenuProps }
 							highlightedIndex={ highlightedIndex }
-							options={ options.filter(
-								( option ) =>
+							options={ options.filter( ( option ) => {
+								if ( isCaseSensitive ) {
+									return (
+										! inputValue ||
+										option.value.includes( inputValue )
+									);
+								}
+								return (
 									! inputValue ||
-									option.value.startsWith(
-										inputValue.toLowerCase()
-									)
-							) }
+									option.value
+										.toLowerCase()
+										.includes( inputValue.toLowerCase() )
+								);
+							} ) }
 						/>
 					) }
 				</div>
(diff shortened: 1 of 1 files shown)


danieldudzic (commented):
By default, the search will be case-insensitive, but setting `isCaseSensitive` to `true` changes it to case-sensitive.

This refactor also fixes the original issue.

assets/js/base/components/dropdown-selector/index.js
```diff
@@ -31,6 +31,7 @@ import './style.scss';
  * @param {boolean} props.multiple Whether multi-select is allowed.
  * @param {function():any} props.onChange Function to be called when onChange event fires.
  * @param {Array} props.options The option values to show in the select.
+ * @param {boolean} props.isCaseSensitive Whether the dropdown search should be case-sensitive.
```
tomasztunik: FYI - optional params can be denoted with [] and we can provide default value here as well.

ie. `@param {boolean} [props.isCaseSensitive=false] ...`

assets/js/base/components/dropdown-selector/index.js
```diff
@@ -182,11 +184,20 @@ const DropdownSelector = ( {
...
+								}
+								return (
 									! inputValue ||
-									option.value.startsWith( inputValue )
-							) }
+									option.name
+										.toLowerCase()
+										.includes( inputValue.toLowerCase() )
+								);
+							} ) }
```
tomasztunik, line 198: Could we dry this bit a bit? we could simplify it to normalize the inputValue and optionName only when needed ie.

we could also add trim to the inputValue

ie.

```
const optionName = isCaseSensitive ? option.name : option.name.toLowerCase();
const nameQuery = isCaseSensitive ? inputValue...

return ! inputValue || optionName.includes( nameQuery.trim() )
```

This way we'll have input/data normalisation separated and all business logic in one place

tomasztunik (changes requested):
Good idea including the option for case sensitivity control!

Leaving small change request regarding drying the code a bit and will be good to go!

tomasztunik (approved):

tomasztunik:
Looks good! :)

PR #6102, opened 2022-03-23
Allow memoized filters to re-run if `defaultArg` changes

### LINKED ISSUE #6066: Coupons applied in cart, cannot be removed
## Describe the bug

When applying a coupon in the cart block, the coupon is applied but cannot be removed.

## To reproduce

1. Create a coupon and add a product to the cart
2. Visit a page that contains the cart block
3. Apply the coupon
4. See that the coupon is applied 
5. The `pill` to remove the coupon does not exist
6. Reload the page, the `pill` will be shown

## Expected behavior

The `pill` to remove the coupon should be visible, the moment the coupon is applied. The checkout block works as expected

## Screenshots

https://user-images.githubusercontent.com/2484390/159268676-2aff53b0-a9c0-4a9f-877a-54d8f559f3e5.mp4

-   Core version: 5.9.2
-   WooCommerce version: 6.3.1
-   WooCommerce Blocks version: 7.3.0-dev (built from trunk)

-   OS: macOS 12.2.1
-   Browser: Chrome v99

### LINKED ISSUE #6097: Snackbar notices aren't showing
## Describe the bug

It seems that snack bar notices are not getting rendered since v7.2.0

## To reproduce

1. Go to the Cart page
2. Add a test coupon
3. The coupon is applied, but the snack bar notice never appears.

## Expected behavior

When I apply a coupon, I want to see a snack bar notice stating that my coupon was successfully applied. Same for removing a coupon.

## Screencast

https://d.pr/v/2XfSfe

### WordPress:

-   Core version: 5.9.2
-   WooCommerce version: 6.4.0-dev 
-   WooCommerce Blocks version: 7.2.1 _(Probably introduced in v7.2.0)_ 
-   Any other plugins installed: No.

opr:
In #5143 memoization was added to prevent third party filter functions running unnecessarily. There was no check in place to see if the default value differed between runs. Usually this would be fine because the value is usually static, but in the case of snackbar notices, it is a list that changes based on what notices are in the `core/notices` data store.

This PR will add a check to see if the `defaultValue` has been seen before in previous runs, and if not the filter functions are allowed to run. 

Fixes #6097

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Create a filter, you can add this code somewhere that will be executed in any JS file.
```
__experimentalRegisterCheckoutFilters('custom-extension', {
  itemName: (name) => {
           console.log('rerunning filter for', name);
	  return `${name} + extra data!`;
  },
});
```
2. Open the browser console and go to the Cart block. Observe the console log added above, ensure it only gets output 1-3 times for each item in the cart. Change the quantity and expect only 1-3 logs for each item again.
3. Create a coupon in your store.
4. Add this coupon to your order in the Cart block, ensure a snackbar notice appears when adding and removing the coupon.
5. Do this multiple times and ensure the snackbar notice appears each time it changes.
6. Repeat step 4 and 5 in the Checkout block.

### User Facing Testing
* [ ] Same as above, or
* [x] See steps below.

1. Create a coupon in your store.
2. Add this coupon to your order in the Cart block, ensure a snackbar notice appears when adding and removing the coupon.
3. Do this multiple times and ensure the snackbar notice appears each time it changes.
4. Repeat step 2 and 3 in the Checkout block.

### Changelog

> Allow memoized checkout filters to re-run if the default value changes between runs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 # Editors
+
 project.xml
 project.properties
 /nbproject/private/
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 # Editors
-
 project.xml
 project.properties
 /nbproject/private/
--- a/packages/checkout/filter-registry/index.ts
+++ b/packages/checkout/filter-registry/index.ts
@@ -93,17 +93,20 @@ const cachedFilterRuns: Record<
 	{
 		arg?: CheckoutFilterArguments;
 		extensions?: Record< string, unknown > | null;
+		defaultValue: unknown;
 	} & Record< string, unknown >
 > = {};
 
-const updatePreviousFilterRun = (
+const updatePreviousFilterRun = < T >(
 	filterName: string,
 	arg: CheckoutFilterArguments,
-	extensions: Record< string, unknown > | null
+	extensions: Record< string, unknown > | null,
+	defaultValue: T
 ): void => {
 	cachedFilterRuns[ filterName ] = {
 		arg,
 		extensions,
+		defaultValue,
 	};
 };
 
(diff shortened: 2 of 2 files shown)


xristos3490:
Thanks for the detailed explanation, @opr! 

Snackbar notices are back! 🎉 

Although, the number of logs in the console seems to be the same in both `trunk` and this branch.
Filters run _once_ on the first-load and _twice_ on every qty update. Here are some screencasts demonstrating this behavior:
- `trunk`: https://d.pr/v/GauoJS and;
- this branch: https://d.pr/v/mT6C0W

I'm not entirely sure that I'm missing anything important here. I'm using WC 6.3.0

opr:
> Thanks for the detailed explanation, @opr! 
> 
> Snackbar notices are back! 🎉 
> 
> Although, the number of logs in the console seems to be the same in both `trunk` and this branch.
> Filters run _once_ on the first-load and _twice_ on every qty update. Here are some screencasts demonstrating this behavior:
> - `trunk`: https://d.pr/v/GauoJS and;
> - this branch: https://d.pr/v/mT6C0W
> 
> I'm not entirely sure that I'm missing anything important here. I'm using WC 6.3.0

Great stuff, yep the number of logs should be the same as trunk. I think two is fine, because the point of the #5143 PR was to stop _many_ logs (sometimes 100+) so two is ok, this testing step was to ensure no regression. Thanks for testing 😌

nielslange (approved):
Works as expected! Let's 🚢 it!

xristos3490 (approved):
> the point of the https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5143 PR was to stop many logs (sometimes 100+) so two is ok

Great stuff indeed! In that case, everything looks good!

opr:
@xristos3490 : @tjcafferkey will add this to the release today 🥳

PR #6106, opened 2022-03-23
Add CI for monitoring TypeScript errors

### LINKED ISSUE #6068: Tracking TypeScript errors
To be successful in Typescript migration, we should start tracking TypeScript errors during CI.

The build has to fail if the number of errors increases compared with the last successful build.

gigitux:
This PR adds an action for monitoring the TypeScript errors in our codebase.

Fixes #6068

The action use `typescript-checkstyle` package that generates a valid [checkstyle file](https://checkstyle.sourceforge.io/) that contains all the errors and the relative files.

After a developer creates a PR, the action will:
- generate a new `checkstyle.xml`, and if there are changes, automatically will push it into the branch.
- add a comment on Github with the number of errors and the number of files that contain errors.  
- check the generated `checkstyle.xml` with the trunk version and compare the two files: if the report of the current branch has more errors than the trunk version, the action will update the comment and will add the path of the files that have the new errors.

Also, at each push in the trunk (every time that a PR is merged in the trunk) the action will update a table on AirTable (not publicly visible), where a linked chart shows the migration status.

## Note 

At this stage, the diff between the `checkstyle` in the current branch and the `checkstyle` in the trunk branch is handled by checking the number of errors for each file. So, for instance, on a specific file if the developer fixes an error, but introduces another one the current implementation will not catch the error, so the action will not add any warning in the comment.

I removed the nested `tsconfig.json` configurations because they did not extend any rules and tsc gave me a lot of errors as this:
```
File 'xxx/xxx/xxx.ts' is not listed within the file list of project 'xxx/xxx/tsconfig.json'. Projects must list all files or use an 'include' pattern.
```

### Testing instructions

Now, the action is failing because the branch to compare is the trunk (that doesn't have a `checkstyle.xml`). I prepared a branch for testing that is equal to this branch.

Check out the branch and update the code following the next instructions.
It is necessary to update [these line](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/19264a4bcf3bb2f210883c3278fd9c5374e55ae7/.github/monitor-typescript-errors/utils/github.js/#L9-L13) and replace with:
``` javascript
return await octokit.rest.repos.getContent( {
    owner,
    repo,
    ref: "try/typescript-monitor-second-branch"
    path: fileName,
    mediaType: {
      format: 'raw',
      },
} );
```

Please, remember that you need to create a PR to trigger the action.

You should check that:

### Action pushes automatically a new checkstyle.xml is something in the codebase change

You can add a new TS error in any file in the codebase. The action should be to push the updated checkstyle automatically.

### Action adds a warning if you introduce a new error in the codebase

You can add a new TS error in any file in the codebase. The actions should add a comment with a list of the file that has new errors.

### Action updates the AirTable table

For this, you need update [this check](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6106/files#diff-0142a71a92636a1f9867130ef62411824f63a59f3adcaaaa4d2625464edccf96R56) and replace `trunk` with try/typescript-monitor-second-branch.

If you visit our AirTable and click on `TypeScript migration`, you should see that there is a new row.

## NOTE

Now, the Monitor TypeScript errors fail in this branch because the job compares the `checkstyle.xml` with the trunk branch, but at this stage, the `checkstyle.xml` is missing in the trunk. This is the reason it is necessary to tweak the file `github.js`.


.github/workflows/typescript-monitoring.yml
```diff
@@ -0,0 +1,62 @@
...
+                  # npm cache files are stored in `~/.npm` on Linux/macOS
+                  path: ~/.npm
+                  key: ${{ runner.OS }}-build-${{ secrets.CACHE_VERSION }}-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+                  restore-keys: |
+                      ${{ runner.OS }}-build-${{ secrets.CACHE_VERSION }}-${{ env.cache-name }}-
+                      ${{ runner.OS }}-build-${{ secrets.CACHE_VERSION }}-
+            - name: Use Node.js 16.13.2
+              uses: actions/setup-node@v1
+              with:
+                  node-version: 16.13.2
```
tomasztunik: Let's use setup-node@v3, caching and version controlled by .nvmrc

```
              uses: actions/setup-node@v3
              with:
                  node-version-file: '.nvmrc'
                  cache: 'npm'
```

We will be able to remove the whole `- name: Cache node modules` step as setup-node@v2+ includes caching module via the simple `cache` setting.

.github/monitor-typescript-errors/action.yml
```diff
@@ -0,0 +1,15 @@
...
+        required: true
+    compare:
+        description: 'Path checkstyle.xml file'
+        required: true
+    trunk_branch_folder:
+        description: 'Path trunk folder'
+        required: true
+runs:
+    using: 'node16'
+    main: 'index.js'
```
tomasztunik, line 13: Is it possible to use nvmrc here like for workflows (see other comment)?

.github/monitor-typescript-errors/index.js
```diff
@@ -0,0 +1,61 @@
...
+	const message =
+		header +
+		'\n' +
+		( filesWithNewErrors.length > 0
+			? `⚠️ ⚠️ This PR introduces new TS errors on ${ filesWithNewErrors.length } files: \n` +
+			  '<details> \n' +
+			  filesWithNewErrors.join( '\n' ) +
+			  '\n' +
+			  '</details>'
+			: '' );
```
tomasztunik: should we provide a feedback message that would say something like
`New errors: 0` and `New errors: X` in the header? This way we'd have positive feedback as well. Currently, it would only show the total which can get someone confused and stressed with the number we get :)

.github/monitor-typescript-errors/utils/airtable.js
```diff
@@ -0,0 +1,38 @@
...
+// https://community.airtable.com/t/datetime-date-field-woes/32121
+const generateDateValueForAirtable = () => {
+	const today = new Date();
+	const string = today.toLocaleDateString();
+
+	return new Date( string );
+};
+
+exports.addRecord = async ( errorsNumber ) =>
+	axios.post(
```
tomasztunik, line 20: We could add error handling here so that if column names changes or a table gets deleted we get friendlier feedback in the logs. Could we test that? Or should airtable be treated as an "effect" and be non breaking for the CI? Ie. show warning about airtable not being accessible but do not crash the step.

.github/monitor-typescript-errors/utils/file.js
```diff
@@ -0,0 +1,13 @@
...
+	return fs.existsSync( path ) && fs.lstatSync( path ).isFile();
+};
+
+exports.openFile = ( path ) => {
+	return fs.readFileSync( path );
+};
+
+exports.saveFile = ( path, content ) => {
+	fs.writeFileSync( path, content );
+};
```
tomasztunik: `isValidPath` and `saveFile` don't seem to be used - do we need these? and in spite of that do we need `openFile` at all? If we don't plan to provide here any custom handling I'd rather use fs directly in the index file to not hide these implementation details behind pass through utils

package.json
```diff
@@ -77,6 +77,7 @@
 		"test:update": "wp-scripts test-unit-js --updateSnapshot --config tests/js/jest.config.json",
 		"test:watch": "npm run test -- --watch",
 		"ts:check": "tsc --build",
+		"ts:monitor": "npm --silent run ts:check | npx -y @bartekbp/typescript-checkstyle > checkstyle.xml",
```
tomasztunik: Also the name is bit misleading I think, monitor seems to imply it continuously monitors something and what it does is it logs TS errors - `ts:log-errors` maybe?

Actually though, I'm thinking if it's needed in package.json at all and if it shouldn't be called in the action itself so that it has access to the name of the checkstyle.xml file that we configure in the workflow. Now we pass it as param but also hardcode in the package.json script.

if we decide to keep it as script we should probably include the script to generate checkstyle formatted log in the dependencies.

.github/workflows/typescript-monitoring.yml
```diff
@@ -0,0 +1,62 @@
...
+            - name: npm install
+              run: |
+                  npm ci
+                  npm run ts:monitor
+            - name: Clone trunk branch
+              uses: actions/checkout@v2
+              with:
+                  ref: trunk
+                  path: trunk
```
tomasztunik: We could move this one into monitor-typescript-errors action script I believe.

We can use octokit request API to fetch contents of a single file from another branch so we wouldn't have to clone whole repo and encapsulate monitor related logic within single action :) ref: https://docs.github.com/en/rest/repos/contents#get-contents

.github/workflows/typescript-monitoring.yml
```diff
@@ -0,0 +1,62 @@
...
+            - name: Get branch name
+              id: branch-name
+              uses: tj-actions/branch-names@v5.2
+
+            - name: Monitor TypeScript errors
+              uses: ./.github/monitor-typescript-errors
+              with:
+                  repo-token: '${{ secrets.GITHUB_TOKEN }}'
+                  compare: checkstyle.xml
+                  trunk_branch_folder: trunk
```
tomasztunik: if we used the `octokit.request` we wouldn't need this input

.github/monitor-typescript-errors/action.yml
```diff
@@ -0,0 +1,15 @@
+name: 'Typescript Monitor'
+description: 'Check TypesScript errors'
+inputs:
+    repo-token:
```
tomasztunik, line 4: we use snake case in other places

.github/monitor-typescript-errors/action.yml
```diff
@@ -0,0 +1,15 @@
...
+        description: 'GitHub token'
+        required: true
+    compare:
+        description: 'Path checkstyle.xml file'
+        required: true
+    trunk_branch_folder:
+        description: 'Path trunk folder'
+        required: true
```
tomasztunik: `trunk_branch_folder` might not be needed as mentioned further

.github/monitor-typescript-errors/action.yml
tomasztunik: I've checked and it is not possible to do that

.github/monitor-typescript-errors/index.js
```diff
@@ -0,0 +1,61 @@
...
+		currentCheckStyleFileParsed
+	);
+
+	const message =
+		header +
+		'\n' +
+		( filesWithNewErrors.length > 0
+			? `⚠️ ⚠️ This PR introduces new TS errors on ${ filesWithNewErrors.length } files: \n` +
+			  '<details> \n' +
+			  filesWithNewErrors.join( '\n' ) +
```
tomasztunik: we'd need to join it with `<br>` maybe? new line char doesn't seem to put file paths on separate lines. 

Also it should be possible to actually display new errors on the files like eslint errors show up now at the end of the PR files review I believe but it could be next steps :)

tomasztunik (changes requested):
Dropped some comments - it's pretty cool and was nice to learn more about how setting up custom actions looks like! Some really nice things in here.

As I mentioned I think I'd hide more of the functionality and logic within the action. 

I'm wondering if we need to generate and store artefacts in the repo (checkstyle.xml) to compare vs trunk and if we decide it's most efficient maybe we should as mentioned use octokit.request to fetch the file directly instead of cloning and simplify this step a bit.

Also as the checkstyle.xml is internal file for the github action I'd maybe create `artefacts` folder in `.github` and save the current branch file in there to not pollute the root of the repo.

We could remove old status comments automatically if follow up commits are pushed.

The language on descriptions could be maybe improved but let's keep it for the final review :) 

Also when there is less erorrs than before we should totally share some kind of WOOHOO GOOD JOB! message :) 

Great stuff!

(ps ran my tests here -> https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6297)

.github/monitor-typescript-errors/utils/airtable.js
gigitux: Yes, my idea is to break the CI because it is important that we monitor our progress. The error handling is done [here](https://github.com/woocommerce/woocommerce-blocks/blob/490fab10ad4092a5a705fa6b5a221613172311c2/.github/monitor-typescript-errors/index.js#L52-L58)

.github/monitor-typescript-errors/action.yml
gigitux: Only for the env variable, we use the snake case. In other GitHub CI Jobs,  as `bundle-size.yml` [we are using `repo-token` ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/490fab10ad4092a5a705fa6b5a221613172311c2/.github/workflows/check-modified-assets.yml/#L73-L74)

gigitux:
Thanks for your great review, @tomasztunik!

> Also as the checkstyle.xml is internal file for the github action I'd maybe create artefacts folder in .github and save the current branch file in there to not pollute the root of the repo.

I don't completely agree. It can be useful to have it at the root of the repo. Potentially for some cases, in `trunk` branch, we can tweak the file to avoid some false positives or some strange cases (I don't know).  Another point, technically speaking, it could be hard picking the last artifacts related to the trunk branch.

> We could remove old status comments automatically if follow up commits are pushed.
I prefer to not add complexity. Keep it simple, at least, for now.

> The language on descriptions could be maybe improved but let's keep it for the final review :)

I'm not greatest on this. Open to suggestions :D

.github/workflows/typescript-monitoring.yml
gigitux: Great suggestion, thanks!

.github/monitor-typescript-errors/utils/xml.js
```diff
@@ -0,0 +1,66 @@
...
+exports.getFilesWithNewErrors = (
+	newCheckStyleFileParsed,
+	currentCheckStyleFileParsed
+) => {
+	const newFilesReport = newCheckStyleFileParsed.files;
+	const currentFilesReport = currentCheckStyleFileParsed.files;
+
+	return Object.keys( newFilesReport ).reduce(
+		( acc, pathfile ) =>
+			currentFilesReport[ pathfile ] === undefined ||
```
Aljullu: Should we intentionally compare the type, here? Just in case somewhere in the code `undefined` was assigned to a value.

```suggestion
			typeof currentFilesReport[ pathfile ] === 'undefined' ||
```

Aljullu (commented):
I'm still testing in #6722, but wanted to leave some early feedback: code looks good and thanks for the detailed testing instructions! I'm going to ask review from Rubik as well, given that this affects the entire repo.

I left a couple of minor comments below. :point_down: 

Should we add `checklist.xml` to [`.distignore`](https://github.com/woocommerce/woocommerce-blocks/blob/trunk/.distignore)?

.github/monitor-typescript-errors/utils/xml.js
gigitux: > Should we intentionally compare the type, here? Just in case somewhere in the code, `undefined` was assigned to a value.
It is impossible that undefined was assigned to a value because everything is generated by the `@bartekbp/typescript-checkstyle` package.

I prefer to keep this version for readability reasons, but if you think the change is necessary, I will do it.

.github/monitor-typescript-errors/utils/xml.js
```diff
@@ -0,0 +1,66 @@
...
+};
+
+exports.getFilesWithNewErrors = (
+	newCheckStyleFileParsed,
+	currentCheckStyleFileParsed
+) => {
+	const newFilesReport = newCheckStyleFileParsed.files;
+	const currentFilesReport = currentCheckStyleFileParsed.files;
+
+	return Object.keys( newFilesReport ).reduce(
```
wavvves: Would it be a good idea to sort these alphabetically so we don't mix folders?

Also instead of displaying a list of just files, could it be more useful to display a list of errors, eg file + line:
`folder/filename:<error_line_number>`

While we're at it, comparing files&lines combos instead could be a way of getting rid of the "solving one, introduce another" problem you mentioned. I'm out of ideas for dealing with file renaming though 🤔

.github/monitor-typescript-errors/utils/xml.js
wavvves: Personally, I don't find this more readable (first time I saw this type of check being done without using `typeof` 😉), and although chances are minimal, `undefined` [is not a reserved word and it is writable outside global scope](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined).

EDIT: Just noted that you were referring to the fact that the object could come with `undefined` primitive value assigned to _pathfile_ property, while @Aljullu mentioned creating and assigning something to a var named "undefined" in this scope. My comment was with @Aljullu suggestion in mind.

.github/monitor-typescript-errors/utils/xml.js
gigitux: > Would it be a good idea to sort these alphabetically, so we don't mix folders?

Agreed. I added the sort.

> Also instead of displaying a list of just files, could it be more useful to display a list of errors, eg file + line:
folder/filename:<error_line_number>


I agree, but it is not so easy to implement, and I don't see that it makes sense to make this effort. As the code is structured now, if I add the error_line_number, it will be added all the errors of that file, not only the errors that the developer introduced in the PR. Furthermore, if I implement an algorithm that checks only the difference between the old `checkstyle.xml` and the new one with ` files&lines combos` when the developers add a space line, the algorithm will fail because all the lines are changing, so for this latter, new errors are introduced even if it isn't true.

.github/monitor-typescript-errors/utils/xml.js
```diff
@@ -0,0 +1,75 @@
...
+
+exports.getFilesWithNewErrors = (
+	newCheckStyleFileParsed,
+	currentCheckStyleFileParsed
+) => {
+	const newFilesReport = newCheckStyleFileParsed.files;
+	const currentFilesReport = currentCheckStyleFileParsed.files;
+
+	return Object.keys( newFilesReport )
+		.sort( ( a, b ) => a.localeCompare( b.firstname ) )
```
Aljullu: I might be missing something, but is it expected that we are comparing `a` with `b.firstname`?

.github/monitor-typescript-errors/utils/xml.js
wavvves: ~~I see. Since we're dealing with tech debt here, can this job not be merge-blocking for now (although you can force merge stuff)?~~

~~Maybe produce a warning about typescript errors in the code, or to keep it failing, analyze only changed files (as the git hook does for phpcs).~~
~~We're suffering a bit with flaky tests; another red mark on the pipeline would desensitize us even more and cause blind force merges 🤔~~

Well that comment didn't made much sense, I now see the reason for the failing CI job:

```
Run ./.github/monitor-typescript-errors
Error: Not Found
/home/runner/work/woocommerce-blocks/woocommerce-blocks/.github/monitor-typescript-errors/index.js:2[8](https://github.com/woocommerce/woocommerce-blocks/actions/runs/3330304374/jobs/5508617034#step:6:9)
	if ( ! currentCheckStyleFile.data ) {
	                             ^

TypeError: Cannot read properties of undefined (reading 'data')
    at runner (/home/runner/work/woocommerce-blocks/woocommerce-blocks/.github/monitor-typescript-errors/index.js:28:3***)
    at processTicksAndRejections (node:internal/process/task_queues:[9](https://github.com/woocommerce/woocommerce-blocks/actions/runs/3330304374/jobs/5508617034#step:6:10)6:5)
```

gigitux: It is expected behavior that the CI fails on this branch:

```
Now, the action is failing because the branch to compare is the trunk (that doesn't have a checkstyle.xml).
```

.github/monitor-typescript-errors/index.js
```diff
@@ -0,0 +1,70 @@
...
+	const { header } = generateMarkdownMessage( newCheckStyleFileParsed );
+	const filesWithNewErrors = getFilesWithNewErrors(
+		newCheckStyleFileParsed,
+		currentCheckStyleFileContentParsed
+	);
+
+	const message =
+		header +
+		'\n' +
+		( filesWithNewErrors.length > 0
```
Aljullu, line 46: If `filesWithNewErrors.length > 0`, should we make the CI job fail to make it more evident? I feel the comment might be easy to miss if the merge button "is green":

imatge

Aljullu (approved):
Exciting to see this moving forward. I did some basic testing in #7488 and everything seems to be working fine, so approving from my side (I only left a couple of minor comments below :point_down: ).

In any case, I think it would be good to get an approval from Rubik before merging. cc @wavvves

.github/monitor-typescript-errors/index.js
gigitux: I agree with this strikethrough text from @wavvves's comment https://github.com/woocommerce/woocommerce-blocks/pull/6106#discussion_r1005989244.

wavvves:
> In any case, I think it would be good to get an approval from Rubik before merging. cc @wavvves

I'll be testing this today

wavvves:
@gigitux I'm seeing errors being reported for the `node_module`s folder:

image

[Btw, this is my testing branch](https://github.com/woocommerce/woocommerce-blocks/pull/7512)

gigitux:
> @gigitux I'm seeing errors being reported for the `node_module`s folder:
> 
> image
> 
> [Btw, this is my testing branch](https://github.com/woocommerce/woocommerce-blocks/pull/7512)

This is a problem with our `tsconfig.json` configuration. I think that it makes sense that shows this error. Also, you see all these errors because the `checkstyle.xml` your branch isn't aligned with `try/typescript-monitor-second-branch`.

wavvves (approved):

PR #6109, opened 2022-03-23
Improve goToSiteEditor E2E utils

tomasztunik:
This is a follow up to #6080 as discussed [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6080#discussion_r833195402). 

The change cleans up `tests/e2e/specs/backend/site-editing-templates.test.js` E2E tests by hiding UX details of waiting for navigation to different contexts of Site Editor inside the goToSiteEditor E2E. 

It will: 
- wait for table being rendered when accessing template or template parts list - via `postType` param (defaults to `wp_template`)
- or waiting for Actions controls on template or template parts list when we want to ensure latest templates state is fetched and Actions are available on modified templates - via `waitForActions` param.
- wait for canvas when going to editing a template or template_part view - via `postId` param

By default the method if no params are provided will take us to the templates list view.

Fixes #6080 

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
### Manual Testing

1. E2E CI should pass

### DIFF
--- a/tests/e2e/specs/backend/site-editing-templates.test.js
+++ b/tests/e2e/specs/backend/site-editing-templates.test.js
@@ -15,7 +15,8 @@ import {
 	DEFAULT_TIMEOUT,
 	filterCurrentBlocks,
 	getAllTemplates,
-	goToSiteEditor,
+	goToTemplateEditor,
+	goToTemplatesList,
 	saveTemplate,
 	useTheme,
 } from '../../utils';
@@ -24,7 +25,7 @@ async function visitTemplateAndAddCustomParagraph(
 	templateSlug,
 	customText = CUSTOMIZED_STRING
 ) {
-	await goToSiteEditor( {
+	await goToTemplateEditor( {
 		postId: `woocommerce/woocommerce//${ templateSlug }`,
 	} );
 
@@ -118,7 +119,7 @@ describe( 'Store Editing Templates', () => {
 		it( 'default template from WooCommerce Blocks is available on an FSE theme', async () => {
 			const EXPECTED_TEMPLATE = defaultTemplateProps( 'Single Product' );
 
-			await goToSiteEditor();
+			await goToTemplatesList();
 
 			const templates = await getAllTemplates();
 
@@ -136,7 +137,7 @@ describe( 'Store Editing Templates', () => {
 		} );
 
 		it( 'should contain the "WooCommerce Single Product Block" classic template', async () => {
-			await goToSiteEditor( {
+			await goToTemplateEditor( {
 				postId: 'woocommerce/woocommerce//single-product',
 			} );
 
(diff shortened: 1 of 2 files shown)


sunyatasattva (commented):
I'm personally not a big fan of a function that does too many things depending on which parameters are passed to it.

Puppeteer itself had a function called `waitFor` which would behave differently depending on what arguments were passed. They have since deprecated that in favor of more specific functions.

I agree with this approach, so instead of a `goToSiteEditor` that behaves differently depending on where you want to go, I'd have a `goToTemplatesList` that does just that: you can still call `goToSiteEditor` under the hood and then basically just add the `waitForSelector`.

I'm not putting this as a “requesting change“, because I'd love to hear your opinion. If you are strongly married to this approach, I will approve it no problem.

tomasztunik:
I'd see it all as a PageObject and thought it could be improvement for the
time being. This method was already handling the same cases it just wasn't
taking full responsibility for hiding the UX/waiting details you mentioned.

If you think this will be an improvement for now I'd go with it and then do
a proper move to pageobjects once we are there and split the responsibility
better. At least it would be all now in one place making it much easier I
think.

A to-do comment inside?

pt., 25 mar 2022, 17:31 użytkownik Lucio Giannotta ***@***.***>
napisał:

sunyatasattva (approved):
I'm approving, for now, then. I don't think it's worth to add a todo comment: this is already a non blocking issue which sprang from a todo comment. The fix to this is fast enough; you know what they say, _if it takes less than 2 minutes, don't create a todo, do it now!_

So, this for me is an architecture decision: if we go with this current implementation, it is because we like it. If we think separation of concerns of this util is useful/better, we should just do it now because it takes very little time.

I'm ok with both, but I wouldn't go on spawning tickets.

tomasztunik:
Let's split it then - it's the solution we've originally discussed.

Should we split it by type or by context? Ie. `goToTemplatePart` vs `goToTemplate` with ability to go to list or item view by providing id and hiding complexity of having to understand what is `wp_template` vs `wp_template_part`. These could use single helper behind the scene ie. the one we have now. 

Or should we make it per view type and go with goToTemplateEditor({ id, type }) goToTemplatesList({ waitForActions, type }) with similar defaults we had now (`wp_template`)

Which one do you lean toward to more?

Something I know I'd like to keep is the ability to pass params/options as an object so that we can provide strong typing - passing by query string like we had before is too prone for errors.

sunyatasattva:
Totally fine with passing params via object. I'd differentiate per view type (second option) and perhaps instead of a `waitForActions` I'd have something like `waitFor: 'list' | 'anyEditedTemplate' = 'list'` or something along those lines, to make it more explicit what we are doing.

tomasztunik:
Sounds good - will update it quickly on Monday :)

sunyatasattva:
I'm loving it! Thanks for taking my feedback into account, Tomek! I'm now much more happy with this. Approved with flying colors! 🌈

PR #6114, opened 2022-03-24
Upgrade @wordpress/scripts and /env packages

### LINKED ISSUE #6076: Upgrade @wordpress/scripts to the latest version
### Update @wordpress/scripts to latest version (v22.2)

We were stuck on old versions because of difficulties udpating tests to jest 27 and some eslint issues.

ref: [Full changelog](https://github.com/WordPress/gutenberg/blob/trunk/packages/scripts/CHANGELOG.md)

Main changes:
- Jest 27
- current Puppeteer 
- node 12+
- webpack 5
- eslint 8

- wp-polyfil not included by default
- support for hot-reload (we might "steal" some of the setup for our puprose)
- for more check changelog

tomasztunik:
Package updates

- Upgraded @WordPress/scripts to latest (22.3.0) [[changelog](https://github.com/WordPress/gutenberg/blob/dba91fbd7cbbe7c49773fa19198853f8c164de6d/packages/scripts/CHANGELOG.md)]
- Upgraded @WordPress/env to latest (4.1.3) [[changelog](https://github.com/WordPress/gutenberg/blob/dba91fbd7cbbe7c49773fa19198853f8c164de6d/packages/env/CHANGELOG.md)]
- `bin/eslint-plugin-woocommerce` eslint dependency upgraded to 8
- removed `@types/wordpress__deprecated` and `@types/wordpress__keycodes` - these packages now provide their own types

Tests
- we are now using **Jest v27** [[changelog](https://github.com/facebook/jest/blob/main/CHANGELOG.md)] [[release post](https://jestjs.io/blog/2021/05/25/jest-27)]
- we had to upgrade how [jest is setup](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6114/files#diff-23b569516ff643e5844316b28124cb89e6c534d1cab8d3bbb76d4abd253ab395) - this is because them moving to ESM and not exporting things properly - fix is on its way in Jest 28
- we are now using latest jest-jsdom in place of temp package 
- 🧐 had to add [babel-plugin-explicit-exports-references](https://www.npmjs.com/package/babel-plugin-explicit-exports-references) to enable spying on external dependencies of other packages because of how updated babel/typescript changed how output is generated - before all methods used would be output as exports.methodName now they are output as just methodName making it impossible to mock. This plugin brings back the previous behaviour. This caused `assets/js/blocks-registry/payment-methods/test/payment-method-config.tsx` to fail.
- improved e2e stability of `tests/e2e/specs/shopper/cart-products.test.js`
- 💡 Jest 28 is [on its way](https://github.com/facebook/jest/milestone/14) 

JS/TS linting
- we are now using **ESlint v8** [[changelog](https://github.com/eslint/eslint/releases)] [[release posts](https://eslint.org/blog/)]
- fixed linting issues that emerged after upgrade

CSS linting
- we are now using **stylelint v14.x**
- updated our stylelintrc to use "@wordpress/stylelint-config/scss" to enable proper validation of scss files
- it is now recommended to use [SASS modules](https://sass-lang.com/documentation/modules) rather than use the utils directly (ie. `math.percentage` instead of `percentage`)
- in [bin/webpack-configs.js](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6114/files#diff-fe2a96dd06027bb736a2f6e1cbaae6d5174d6aefd9fc1058a2527f810648504a) exposed mroe sass modules so that we don't have to add them to all the files where they are used. This is also because of how we added math like that in the first places and @use can't be added after @import rules which follows the use in the mentioned webpack config mod.
- fixed css errors that appeared after upgrade
- removed unsupported rules from stylelintrc
- 🆚 added recommended stylelint extension for VSCode
- 🆚 added VSCode workspace settings to enable stylelint as CSS/SCSS validator

WP-Env
- we are now on latest released v4.1.3 [[changelog](https://github.com/WordPress/gutenberg/blob/0c0768f198f3cc72fad30d7c7b1498012dae2ea0/packages/env/CHANGELOG.md)]
- optimizations to how env is setup and how much of extra deps are fetched for setup

Fixes #6076 

#### Automated Tests
  * [x] Unit tests
  * [x] E2E tests
  * [x] js lint
  * [x] css lint

### Manual Testing

1. All CI workflows should pass


assets/css/editor.scss
```diff
@@ -40,7 +40,7 @@
 		margin: -$gap 0 $gap;
 	}
 
-	& + .wc-block-base-control-notice:last-child {
+	+ .wc-block-base-control-notice:last-child {
```
sunyatasattva, line 43: Why this change? It seems the intention is changed? 🤔

assets/css/style.scss
```diff
@@ -38,7 +38,7 @@
 .editor-styles-wrapper .wc-block-grid__product-title,
 .wc-block-grid__product-title {
 	font-family: inherit;
-	line-height: 1.2em;
+	line-height: 1.2;
```
sunyatasattva, line 41: Is this change also intentional? The meaning is slightly different.

assets/js/blocks/cart-checkout/payment-methods/style.scss
```diff
@@ -89,12 +89,12 @@
 		}
 		.wc-block-gateway-input.focused.empty,
 		.wc-block-gateway-input:not(.empty) {
-			& + label {
+			+ label {
```
sunyatasattva, line 92: Same question here and in the following lines. Has the meaning of `&` changed in some way I am unaware of, perhaps?

sunyatasattva (commented):
Just a few questions, not necessarily requested changes.

assets/css/editor.scss
tomasztunik: This is actually the same and was marked as an error by css lint.  ref. https://css-tricks.com/the-sass-ampersand/#aa-using-the-with-and

assets/css/style.scss
tomasztunik: It was marked as an error by updated stylelint. These mean the same, unless I missed something. em is relative unit to current font size, and unitless line-height is ratio of current font size. In context of line-height these produce the same value.

assets/js/blocks/cart-checkout/payment-methods/style.scss
tomasztunik: resolving in favour of first occurrence :)

assets/css/editor.scss
sunyatasattva: You're right!

assets/css/style.scss
sunyatasattva: It's just slightly different, and generally recommended to go without `em` so I'd agree with stylelint, but it might have changed stuff.

The difference is the following:

Imagine `.wc-block-grid__product-title` has `font-size: 10px`. Now imagine that we have a child element of `.wc-block-grid__product-title` called `.subtitle`, which has `font-size: 5px`.

Now, if `line-height` on `.wc-block-grid__product-title` is set to `1.2em`, the computed `line-height` on `.subtitle` will be `12px` (generally not what one wants). Otherwise, if set to `1.2`, it will be `6px`. In both cases, `line-height` on `.wc-block-grid__product-title` will be `12px`.

Does it make sense?

Again, generally it's more correct to go with the fix you made, I was just wandering if you tested for any differences.

tomasztunik: Ah, I see what you mean. Spit out a quick https://codepen.io/tomasztunik/pen/yLpbjgw to make it stuck :). I think I was always using it in a way and expecting the results but I didn't put much thought into it.

In this case this shouldn't be an issue as the title block doesn't have a child. Though this gets me thinking if this linter setting that comes from WordPress recommended is not going to bite us in the future somehow. In theory it shouldn't and if it reports error in an unexpected way it will resurface and we'd be able to override it with `.stylelintrc.json`. 

I think for now we should go with the default - what do you think?

sunyatasattva: I'm totally fine with using this, though I think there were more changes like that to line heights that might have children. So perhaps just do a quick before/after check of these elements in context?

I'll approve to not block you in the meantime.

sunyatasattva (approved):
Approving! Let's just be mindful of the changed units in line heights.

assets/css/style.scss
tomasztunik: confirmed that all of them should be good and were on content elements not on containers.

PR #6116, opened 2022-03-24
Shopper → Checkout → Can see warnings when form is incomplete

### LINKED ISSUE #5771: Critical flows: Shopper → Checkout → Can see warnings when form is incomplete
### Story

As a shopper, I want to see warnings when the checkout form is incomplete so that I know which of the fields need to be corrected.

### File

`tests/e2e/specs/shopper/checkout-warnings.test.js`

### Precondition

Make sure that all fields, that can be required, are set to required in the checkout block.

### Test

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Clear all form fields.
5. Try to place an order. 
6. Verify that all required fields show the correct warning.

tarhi-saad:
Fixes #5771

### Notes

This PR adds an e2e test to verify that shoppers can see warnings when the checkout form is incomplete.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

If E2E tests run by GitHub actions pass in this PR then it's all good.

As an extra, if you want to test it locally as well:

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/checkout-form-warnings.test.js` to run only this test case.

### Changelog

> Show warnings when form is incomplete on Checkout

### DIFF
--- a/tests/e2e/specs/shopper/checkout-form-warnings.test.js
+++ b/tests/e2e/specs/shopper/checkout-form-warnings.test.js
@@ -1,0 +1,91 @@
+/**
+ * External dependencies
+ */
+import { MY_ACCOUNT_ACCOUNT_DETAILS } from '@woocommerce/e2e-utils';
+
+/**
+ * Internal dependencies
+ */
+import { shopper } from '../../../utils';
+import { SIMPLE_PRODUCT_NAME } from '../../../utils/constants';
+
+const block = {
+	name: 'Checkout',
+};
+
+if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( `skipping ${ block.name } tests`, () => {} );
+
+describe( 'Shopper → Checkout → Can see warnings when form is incomplete', () => {
+	beforeAll( async () => {
+		const isShopperLoggedIn = await shopper.block.isLoggedIn();
+
+		// @todo Find a better way to reset the checkout form instead of login then logout
+		if ( isShopperLoggedIn ) await shopper.logout();
+		else {
+			await shopper.login();
+			await shopper.logout();
+		}
+	} );
+
+	beforeEach( async () => {
+		await shopper.block.emptyCart();
+	} );
+
+	afterAll( async () => {
+		await shopper.block.emptyCart();
+	} );
+
+	it( 'Shows warnings when form is incomplete', async () => {
+		await shopper.goToShop();
+		await shopper.addToCartFromShopPage( SIMPLE_PRODUCT_NAME );
+		await shopper.block.goToCheckout();
+
+		// Click on "Place Order" button
+		await expect( page ).toClick(
+			'.wc-block-components-checkout-place-order-button',
+			{
+				text: 'Place Order',
+			}
+		);
+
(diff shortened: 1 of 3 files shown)


tests/utils/shopper.js
```diff
@@ -354,5 +354,11 @@ export const shopper = {
 
 			await expect( page.$x( cartItemXPath ) ).resolves.toHaveLength( 1 );
 		},
+
+		isLoggedIn: async () => {
```
alexflorisca: This checks if user is logged in as an admin, so maybe we could change this to `isLoggedInAdmin`?

I would also move this function directly on the `shopper` object as it's not directly related to blocks

alexflorisca (approved):
One minor comment, not a blocker for merging this PR. Great work :)

tests/utils/shopper.js
tarhi-saad: Hi @alexflorisca! Thank you for the review. And thank you for pointing this out. My intention wasn't to check if a user is logged in as admin only, but other types of accounts too (as a customer). I changed the implementation to make the correct check and applied the above suggestions. Please could you have a second look at it! 🙌

alexflorisca (approved):
Nice one, thanks @tarhi-saad , good to go!

tests/e2e/specs/shopper/checkout-form-warnings.test.js
```diff
@@ -0,0 +1,100 @@
...
+			await shopper.loginFromMyAccountPage(
+				CUSTOMER_USERNAME,
+				CUSTOMER_PASSWORD
+			);
+
+			await await page.click(
+				'.woocommerce-MyAccount-navigation-link--customer-logout a'
+			);
+		}
+	} );
```
senadir, line 39: Why are you checking if a user is logged in, for you to log him out, or log him in again? What's the end goal here?

tarhi-saad: We log him out to reset the checkout form by triggering a new session. The other method available to us is to clear each input manually, but this seemed to be slower than simply logout.

The automation bot created an issue regarding this particular case: #6123

tarunvijwani: Thank you @senadir and @tarhi-saad for raising this question. In the recent [issue on which I'm working](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5780), I'm using the following code to logout user and navigate to checkout block to check if login option is available or not. 

We may also use the same code to logout and clear form fields:


```
await shopper.logout();
await shopper.goToShop();
await shopper.addToCartFromShopPage( SIMPLE_PRODUCT_NAME );
await shopper.block.goToCheckout();
```

tarhi-saad: Thank you, @tarunvijwani, for suggesting an alternative way to achieve the same behavior. 🙌 In this case, since we are already on the `My account` page before login out, using the `shopper.logout` function will navigate us to the same page a second time [see implementation](https://github.com/woocommerce/woocommerce/blob/trunk/packages/js/e2e-utils/src/flows/shopper.js#L324).
So, to avoid unnecessary actions, we are directly clicking on the `logout` link instead.

PR #6130, opened 2022-03-28
Mini Cart Contents: Refactor inner blocks styling using `currentColor`

dinhtungdu:
This PR refactors the style of Mini Cart Contents inner blocks to make them work better with color settings of Mini Cart Contents.

Note: This doesn't include style for quantity input which is shared between Cart and Mini Cart.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Edit the Mini Cart template part.
2. Set color and background color to custom values.
3. See the inner blocks applied correctly on both editor and front end.
### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.


assets/js/blocks/cart-checkout/mini-cart/style.scss
```diff
@@ -116,6 +116,7 @@ h2.wc-block-mini-cart__title {
 			}
 		}
 	}
+
```
Aljullu:
```suggestion
```

assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
```diff
@@ -81,6 +82,30 @@ const Edit = ( { clientId }: Props ): ReactElement => {
 					templateLock={ false }
 				/>
 			</EditorProvider>
+			{ /**
+			 * This is a workaround to style inner blocks using the color
+			 * settings of the Mini Cart Contents block. It's possible to get
+			 * the Mini Cart Contents block's attributes inside the inner blocks
+			 * components, but we have 4/7 inner blocks that inherit style from
```
Aljullu: What does `4/7` mean here?

Aljullu (commented):
I like how it simplifies the code, good job!

The only issue I noticed is that if I set the text color but not the background color, the Checkout button is not visible in the editor:

imatge

assets/js/blocks/cart-checkout/mini-cart-contents/edit.tsx
dinhtungdu: it's 4 out of 7, I will update the comment to make it clearer.

dinhtungdu:
> The only issue I noticed is that if I set the text color but not the background color, the Checkout button is not visible in the editor:

@Aljullu I can't reproduce this issue. I tried both Gutenberg and Core's Editor, the editor and front end.

image

I also tried setting only the background:

image

Aljullu (approved):
> @Aljullu I can't reproduce this issue.

Not sure what I did, I'm unable to reproduce again. Sorry for the false alarm. :sweat: 

This LGTM, so pre-approving. My only last concern is that it might print styles like `color: undefined` (when text color is not set). Do you think we could avoid this?

imatge

dinhtungdu:
> My only last concern is that it might print styles like color: undefined (when text color is not set). Do you think we could avoid this?

@Aljullu Yes, we can. In 9ed5ec0, I check for the value of the style before returning it. I decided to create another component to handle the style there to keep the main edit component easy to read. The dedicated component is more future-proof because we may introduce new inner blocks (I may overthink here).

Aljullu (approved):
I like this abstraction! My only suggestion would be to rename the component from `InnerBlocksStyle` to `MiniCartInnerBlocksStyle`. The rationale being that currently, I don't think it can't be used in any other blocks/components, so I think it's better to make it clear in the component name that it's specific to the Mini Cart.

Another option would be to refactor it so things like the `selector` or the mappings (`color` → `style.backgroundColor`) are passed from the component props. That would pave the way in case sometime in the future we decide to use that component somewhere else. However, I don't think it's worth it until we need to do it. Thoughts?

dinhtungdu:
> I like this abstraction! My only suggestion would be to rename the component from InnerBlocksStyle to MiniCartInnerBlocksStyle. 

@Aljullu I renamed the component per your suggestion.

> However, I don't think it's worth it until we need to do it. Thoughts?

I strongly agree. The first reason for the `MiniCartInnerBlocksStyle` is I want the edit component of the Mini Cart Contents block shorter and easier to read. When we actually need this for other blocks, we can refactor `MiniCartInnerBlocksStyle` at that point.

PR #6131, opened 2022-03-28
Update `Product Title` link settings to match `Post Title`

### LINKED ISSUE #6004: Update Product Title link settings to match Post Title
We would like to align the link settings from the Product Title block with the ones from the Post Title block from Gutenberg.

## To reproduce

1. Add a Product Title and Post Title blocks into the same page.
2. Compare their link settings:

Product Title | Post Title
--- | ---
imatge | imatge

## Expected behavior

Settings should look the same for both blocks, so users get the same UI in both of them.

## Specific actions

- Rename Panel from _Content_ to _Link settings_.
- Rename toggle _Link to Product Page_ to _Make title a link_.
- Add option to _Open in new tab_.
- Remove text _Links the image to the single product listing_.
- Add link rel input.

albarin:
This PR changes the settings on the `Product Title` block to be aligned with the `Post Title` block ones.
It introduces these changes:
- Rename Panel from `Content` to `Link settings`.
- Rename toggle `Link to Product Page` to `Make title a link`.
- Remove text `Links the image to the single product listing`.
- Add an option to `Open in new tab`.

Fixes #6004

### Screenshots

Before | After
--- | ---
before|Screenshot 2022-04-01 at 08 48 54

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests

### Manual Testing

1. Add a `Product title` block to a page
2. Go to the `Link settings` section on the block settings
3. Check it has the `Make title a link`, `Open in new tab` toggles
5. Make changes to those settings, click the Update button, make sure the settings are stored
6. Open the page frontend and check that the selected settings are applied when the product title block is rendered

### Changelog

> Update `Product Title` link settings content and add an option to `Open in new tab`.

### DIFF
--- a/assets/js/atomic/blocks/product-elements/title/attributes.ts
+++ b/assets/js/atomic/blocks/product-elements/title/attributes.ts
@@ -17,7 +17,6 @@ let blockAttributes: Record< string, Record< string, unknown > > = {
 	},
 	rel: {
 		type: 'string',
-		default: '',
 	},
 	productId: {
 		type: 'number',
--- a/assets/js/atomic/blocks/product-elements/title/attributes.ts
+++ b/assets/js/atomic/blocks/product-elements/title/attributes.ts
@@ -12,6 +12,14 @@ let blockAttributes: Record< string, Record< string, unknown > > = {
 		type: 'boolean',
 		default: true,
 	},
+	linkTarget: {
+		type: 'string',
+		default: '_self',
+	},
+	rel: {
+		type: 'string',
+		default: '',
+	},
 	productId: {
 		type: 'number',
 		default: 0,
--- a/assets/js/atomic/blocks/product-elements/title/attributes.ts
+++ b/assets/js/atomic/blocks/product-elements/title/attributes.ts
@@ -14,7 +14,6 @@ let blockAttributes: Record< string, Record< string, unknown > > = {
 	},
 	linkTarget: {
 		type: 'string',
-		default: '_self',
 	},
 	rel: {
 		type: 'string',
--- a/assets/js/atomic/blocks/product-elements/title/attributes.ts
+++ b/assets/js/atomic/blocks/product-elements/title/attributes.ts
@@ -15,9 +15,6 @@ let blockAttributes: Record< string, Record< string, unknown > > = {
 	linkTarget: {
 		type: 'string',
 	},
-	rel: {
-		type: 'string',
-	},
 	productId: {
 		type: 'number',
 		default: 0,
--- a/assets/js/atomic/blocks/product-elements/title/block.tsx
+++ b/assets/js/atomic/blocks/product-elements/title/block.tsx
@@ -124,7 +124,7 @@ export const Block = ( props: Props ): JSX.Element => {
 				disabled={ ! showProductLink }
 				name={ product.name }
 				permalink={ product.permalink }
-				rel={ rel }
+				rel={ showProductLink ? 'nofollow ' + rel : '' }
 				target={ linkTarget }
 				onClick={ () => {
 					dispatchStoreEvent( 'product-view-link', {
--- a/assets/js/atomic/blocks/product-elements/title/block.tsx
+++ b/assets/js/atomic/blocks/product-elements/title/block.tsx
@@ -47,7 +47,6 @@ const TagName = ( {
  * @param {number}  [props.headingLevel]    Heading level (h1, h2, etc.)
  * @param {boolean} [props.showProductLink] Whether or not to display a link to the product page.
  * @param {string}  [props.linkTarget]      Specifies where to open the linked URL.
- * @param {string}  [props.rel]             The relationship of the linked URL.
  * @param {string}  [props.align]           Title alignment.
  * will be used if this is not provided.
  * @return {*} The component.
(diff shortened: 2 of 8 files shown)


assets/js/atomic/blocks/product-elements/title/block.tsx
```diff
@@ -120,7 +124,8 @@ export const Block = ( props: Props ): JSX.Element => {
 				disabled={ ! showProductLink }
 				name={ product.name }
 				permalink={ product.permalink }
-				rel={ showProductLink ? 'nofollow' : '' }
+				rel={ 'nofollow' + ( rel && ' ' + rel ) }
```
sunyatasattva: I'd build the second part of this string separately and then assign it here. Also, did you know about [Template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)? They are cooler than ol' fashioned string concatenation! 😎

assets/js/atomic/blocks/product-elements/title/block.tsx
```diff
@@ -120,7 +124,8 @@ export const Block = ( props: Props ): JSX.Element => {
...
+				target={ linkTarget || undefined }
```
sunyatasattva: Since `linkTarget` is an optional prop, wouldn't it already be `string | undefined`?

assets/js/atomic/blocks/product-elements/title/test/block.test.js
```diff
@@ -0,0 +1,80 @@
...
+
+			const productName = component.getByText( product.name );
+			const anchor = productName.closest( 'a' );
+
+			expect( anchor.getAttribute( 'href' ) ).toBe( product.permalink );
+			expect( anchor.getAttribute( 'rel' ) ).toContain( 'some_rel' );
+			expect( anchor.getAttribute( 'rel' ) ).toContain( 'nofollow' );
+		} );
+	} );
+} );
```
sunyatasattva, line 63: +1 for tests!

sunyatasattva (changes requested):
Just minor things! Great job!

assets/js/atomic/blocks/product-elements/title/block.tsx
albarin: I've extracted this into a `const`, also changed slightly the condition since the result was sometimes `nofollowundefined` 😅

assets/js/atomic/blocks/product-elements/title/block.tsx
```diff
@@ -120,7 +126,8 @@ export const Block = ( props: Props ): JSX.Element => {
 				disabled={ ! showProductLink }
 				name={ product.name }
 				permalink={ product.permalink }
-				rel={ showProductLink ? 'nofollow' : '' }
+				rel={ 'nofollow' + linkRel }
```
tomasztunik: This could produce invalid value because of a missing space and if linkRel was provided.

Also do we know why we use `nofollow` here? This should be used only for external links afaik that we do not want google to follow. In case of a product title we want google to continue indexing/traversing the page or tracking navigation - right?

It shouldn't be used for internal pages and as per [google recommendations](https://developers.google.com/search/docs/advanced/guidelines/qualify-outbound-links) we should use [robots.txt disallow](https://developers.google.com/search/docs/advanced/robots/robots_txt#disallow) to control what we do not want indexed/tracked for internal pages

The more I'm thinking about this the more I think if we need `rel` on the product title at all. It is always an internal link to a product and not sure if we can have anything meaningful for `rel` and if we should expose it at all.

I think the post title maybe had an option to have a custom/external link set for it? Or maybe a sponsored post, UGC or something like that.

assets/js/atomic/blocks/product-elements/title/attributes.ts
```diff
@@ -12,6 +12,14 @@ let blockAttributes: Record< string, Record< string, unknown > > = {
 		type: 'boolean',
 		default: true,
 	},
+	linkTarget: {
+		type: 'string',
+		default: '_self',
```
tomasztunik: We shouldn't have to provide this as the default - this is a browser default and we should let browsers control that.

tomasztunik (changes requested):
Looks good! A note on target default value and small issue found around the `rel` attribute - and question if we need it at all.

assets/js/atomic/blocks/product-elements/title/block.tsx
albarin: I think the issue you mention about a missing space is controlled in [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6131/files#diff-e7199a70d4d8ad7ef850a7421730ef85eba7104248b9fb83cfcbd4d2e58a0abaR100) so it would not happen, since the space is added.

About the `nofollow`, I kept it because it was the behavior it had before the change, but I think your reasoning makes total sense, we could get rid of it.

tomasztunik: Ah, it was collapsed and did not notice it - my bad - but hey something good came out of it! :)

I think we should remove it to not overload people with options that do not make much sense.

ralucaStan: In WC the title, part of the first anchor in the image below, does not have a `rel` at all, only the Add to Cart link has this `rel="nofollow", which makes sense to me. So I think it's ok to remove the `rel` from the title.

Screenshot 2022-03-31 at 11 19 12

tomasztunik (approved):

sunyatasattva (approved):
LGTM!

Issue #6134, opened 2022-03-28
Shop catalog template with filters compatibility for Block themes


Aljullu:
Currently, it's not possible to use filters in the Shop page when a block theme is active. That's because WC filters are widgets so, unavailable in block themes.

We do have some block equivalents to those widgets; however, the blocks currently only work together with the All Products block, but don't work with the Shop catalog template.

This project is about updating our filter blocks so they can work with both, the All Products block and classic PHP templates from WC.

Aljullu:
Closing as this project has been completed.

PR #6146, opened 2022-03-28
Update Filter Products by Price block to work with PHP templates

### LINKED ISSUE #6135: Make Filter Products by Price work with PHP templates
Currently, it's not possible to use the Filter Products by Price block with PHP templates. That's because that block is designed to only work with the All Products block.

This issue is about updating it, so it can work with the Shop catalog template.

## To reproduce

1. With a block theme, go to Appearance > Editor > Template > Product Catalog.
2. Add the Filter Products by Price block above the product grid.
3. Change the values of the price slider.
4. Notice the products in the product grid don't update.

## Expected behavior

The products in the product grid should update based on the selected filters.

### LINKED ISSUE #6139: Provide a system for filter blocks to know if they need to interact with All Products or with a PHP template
As part of #6134, we need filter blocks to know if they need to interact with the All Products block or with a PHP template. This issue is about providing a system so filter blocks know how to behave.

I guess [this util](https://github.com/woocommerce/woocommerce/blob/2215c9ab3ee7759effde7a93ea01a95654b56b60/plugins/woocommerce/includes/blocks/class-wc-blocks-utils.php#L69) from WC core can be used to detect if the All Products blocks is in the page or not? Then we would need to pass that value from PHP to JS.

Aljullu:
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/6135
Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/6139

Filter Products by Price block will only work with the PHP rendered `Classic Template` block if the filters are applied to the URL and the page is re-rendered. This behaviour happens with the Classic Themes such as Storefront.

**Changes introduced:**
* Use `asset_data_registry` to pass data to the frontend when the PHP template is being used. This will let the JS components know how to behave when interacting with them as opposed to the All Products block.
* Refactor the Filter Products by Price block to get filter params, append them to the URL and reload the page.

### Manual Testing

1. With a block theme, go to Appearance > Editor > Template > Product Catalog.
2. Add the Filter Products by Price block above the product grid and set _Filter Button_ attribute to true.
3. Change the values of the price slider.
4. Notice the page reloads and the list of products updates accordingly.

Kooha-03-28-2022-16-01-09b

### Changelog

> Allow adding the Filter Products by Price block to Product Catalog templates to filter products.


assets/js/blocks/price-filter/block.js
```diff
@@ -12,13 +12,40 @@ import PriceSlider from '@woocommerce/base-components/price-slider';
...
 import { getCurrencyFromPriceResponse } from '@woocommerce/price-format';
+import { getSetting } from '@woocommerce/settings';
 
 /**
  * Internal dependencies
  */
 import usePriceConstraints from './use-price-constraints.js';
 import './style.scss';
 
+function findGetParameter( paramName ) {
```
sunyatasattva, line 29: I suggest we use the existing `@wordpress/url` package for these kind of utils. In this case[, `getQueryArg`](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-url/#getqueryarg) should fit the bill.

What do you think?

assets/js/blocks/price-filter/block.js
```diff
@@ -12,13 +12,40 @@ import PriceSlider from '@woocommerce/base-components/price-slider';
...
+	if ( ! window ) {
+		return;
+	}
+	const url = new URL( window.location );
+	const params = new URLSearchParams( url.search );
+
+	return params.get( paramName );
+}
+
+function formatParams( url, paramSettings ) {
```
sunyatasattva: I'd like some JSDoc here if possible. From the name of the function, I would expect this to only format parameters, but it actually also filters them according to a constraint. Also, with JSDoc it would be clearer how to use the `paramSettings` without having to look at the implementation.

It is also a bit unclear to me whether the filtering should be responsibility of this function and why would we not just filter the constraint within the components before parsing the new parameters, if the question makes sense?

Lastly, sorry if this is a stupid question: it seems that, in practice, we delete the param if the value equals the maximum/minimum value. What if it exceeds it in one direction or the other? Is that even possible?

assets/js/blocks/price-filter/block.js
```diff
@@ -12,13 +12,40 @@ import PriceSlider from '@woocommerce/base-components/price-slider';
...
+	const keys = Object.keys( paramSettings );
+
+	for ( let i = 0; keys.length > i; i++ ) {
+		const { value, constraint } = paramSettings[ keys[ i ] ];
+		if ( value === constraint ) {
+			params.delete( keys[ i ] );
+		} else {
+			params.set( keys[ i ], value / 100 );
+		}
+	}
```
sunyatasattva:
```suggestion
    for ( const [ key, setting ] of Object.entries( paramSettings ) ) {
        const { value, constraint } = setting;
 
		if ( value === constraint ) {
			params.delete( key );
		} else {
			params.set( keys, value / 100 );
		}
	}
```

or, of course, we could destructure that in the loop declaration itself:

```js
    for ( const [ key, { value, constraint } ] of Object.entries( paramSettings ) ) {
```

Either way, I think using `for ... of` is more idiomatic than a regular `for` loop.

assets/js/blocks/price-filter/block.js
```diff
@@ -27,22 +54,31 @@ import './style.scss';
...
 		queryState,
 	} );
 
-	const [ minPrice, setMinPrice ] = useState();
-	const [ maxPrice, setMaxPrice ] = useState();
+	const [ minPrice, setMinPrice ] = useState(
+		findGetParameter( 'min_price' ) * 100 || null
+	);
+	const [ maxPrice, setMaxPrice ] = useState(
+		findGetParameter( 'max_price' ) * 100 || null
```
sunyatasattva: Since these statements are used twice, what do you think of refactoring them above? Something like:

```js
const startingMinPrice = findGetParameter( 'min_price' ) * 100 || null;
// ...
```

assets/js/blocks/price-filter/block.js
```diff
@@ -59,14 +95,42 @@ const PriceFilterBlock = ( { attributes, isEditor = false } ) => {
...
-			setMinPriceQuery(
-				newMinPrice === minConstraint ? undefined : newMinPrice
-			);
-			setMaxPriceQuery(
-				newMaxPrice === maxConstraint ? undefined : newMaxPrice
-			);
+			if ( filteringForPhpTemplate && window ) {
+				// For block templates that render the PHP Classic Template block we need to add the filters as params and reload the page.
+				const url = new URL( window.location );
+				const currentParams = new URLSearchParams( url.search );
```
sunyatasattva: Perhaps here we could also use https://developer.wordpress.org/block-editor/reference-guides/packages/packages-url/#getqueryargs ?

assets/js/blocks/price-filter/block.js
```diff
@@ -59,14 +95,42 @@ const PriceFilterBlock = ( { attributes, isEditor = false } ) => {
...
+					window.location =
+						url.origin + url.pathname + '?' + newParams;
+				}
+			} else {
+				setMinPriceQuery(
+					newMinPrice === minConstraint ? undefined : newMinPrice
+				);
+				setMaxPriceQuery(
+					newMaxPrice === maxConstraint ? undefined : newMaxPrice
+				);
```
sunyatasattva: As I mentioned above, I am not sure the whole `value/constraint` thing is required. Since the constraint is always equality, it seems, and the evaluation is done already on L120 and 123, why not do that anyways before passing it to the `formatParams` function, like we do it before passing it to the `setMinPriceQuery`?

Something like:

```js
// Using snake_case, just for ease of assignment to URL params, but we can jump
// an extra hoop and assign it later.
const max_price = newMaxPrice === maxConstraint ? undefined : newMaxPrice;
// ...

const newParams = formatParams( url, { max_price, min_price } );

// ...else
setMinPriceQuery(min_price);
setMaxPriceQuery(max_price);
```

sunyatasattva (changes requested):

assets/js/blocks/price-filter/block.js
tjcafferkey: Nice observation, you're totally right.

assets/js/blocks/price-filter/block.js
tjcafferkey: >What if it exceeds it in one direction or the other? Is that even possible?

I've refactored some of the code to account for this

assets/js/blocks/price-filter/block.js
```diff
@@ -95,33 +110,30 @@ const PriceFilterBlock = ( { attributes, isEditor = false } ) => {
 	// Updates the query based on slider values.
 	const onSubmit = useCallback(
 		( newMinPrice, newMaxPrice ) => {
+			// Using snake_case, just for ease of assignment to URL params.
```
sunyatasattva: You didn't end up using `snake_case` though, so this comment might not be required :P

assets/js/blocks/price-filter/block.js
```diff
@@ -95,33 +110,30 @@ const PriceFilterBlock = ( { attributes, isEditor = false } ) => {
...
+			const finalMaxPrice =
+				newMaxPrice >= Number( maxConstraint )
+					? undefined
+					: newMaxPrice;
+			const finalMinPrice =
+				newMinPrice <= Number( minConstraint )
+					? undefined
+					: newMinPrice;
```
sunyatasattva, line 120: This is extremely minor and just a suggestion, so feel free to ignore.

I personally find it slightly harder to read when the truthy assignment of a ternary operator is `undefined`. It _feels_ a bit more right that `undefined` would sorta be the fallback, you know what I mean? So, I'd more easily read:

```suggestion
			const finalMaxPrice =
				newMaxPrice < Number( maxConstraint )
					? newMaxPrice
					: undefined;
			const finalMinPrice =
				newMinPrice > Number( minConstraint )
					? newMinPrice
					: undefined;
```

Again, totally personal preference. Just throwing it out there.

sunyatasattva (approved):
Really minor thing (comment removal, the other is just a suggestion). Approving as to not block you when you fix it :) 

Thanks for addressing my feedback so thoroughly!

assets/js/blocks/price-filter/block.js
tjcafferkey: Good catch, I decided against `snake_case` after using it in my first approach but obviously forgot to remove the comment!

assets/js/blocks/price-filter/block.js
tjcafferkey: I can understand that. I'll reverse the logic here, I think the cognitive load is less like you suggested.

tjcafferkey:
Failing tests seem unrelated (checkout) and are a result of this PR being merged supposedly https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6126 so going to go ahead and merge this.

dinhtungdu:
@tjcafferkey Not sure if this is happening on your end, but the price filter doesn't work for me. The min/max prices in the URL are wrong after changing the price slider:

image

Note: I'm using latest `trunk`.

Aljullu:
What are your currency settings, @dinhtungdu? You can find them in WooCommerce > Settings. It might be worth opening an issue and adding it to the #6134 epic.

dinhtungdu:
@Aljullu You're right. I'm setting `Number of decimals` to `0`. This PR only works if the Number of decimals is set to 2. I will create an issue and a PR to fix it.

PR #6152, opened 2022-03-29
Product Query: pass any product taxonomies existing in the URL parameters.

cpapazoglou:
While working on 12894-gh-Automattic/woocommerce.com I have noticed that there isn't a way to filter products based on custom product taxonomies. The code only allows a static list of `product_cat` and `product_tag`. This PR aims at using any product taxonomy present in the URL parameters to build the `$query_args` for the `WP_Query`.

Related convo: p1648465508181829-slack-C8X6Q7XQU

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

- On your local environment register a new taxonomy
```
$labels = array(
	'name'                       => 'Custom Category',
	'singular_name'              => 'Category',
	'menu_name'                  => 'Categories',
	'all_items'                  => 'All Categories',
	'parent_item'                => 'Parent Category',
	'parent_item_colon'          => 'Parent Category:',
	'new_item_name'              => 'New Category Name',
	'add_new_item'               => 'Add New Category',
	'edit_item'                  => 'Edit Category',
	'update_item'                => 'Update Category',
	'separate_items_with_commas' => 'Separate Category with commas',
	'search_items'               => 'Search Categories',
	'add_or_remove_items'        => 'Add or remove Categories',
	'choose_from_most_used'      => 'Choose from the most used Categories',
);
$taxonomy_args = array(
	'labels'       => $labels,
	'description'  => 'A new custom product taxonomy',
	'hierarchical' => true,
	'public'       => true,
	'show_ui'      => true,
);
register_taxonomy( 'custom_categories', 'product', $taxonomy_args );
```
- Edit a product and add to it a new term of the `custom_categories` taxonomy
- Go to `/wp-json/wc/store/products?_unstable_tax_custom_categories=[that-category-id]` and make sure you get only products where this term was added
- Go to `/wp-json/wc/store/products` and make sure that all products get returned as expected

### User Facing Testing
* [ ] Same as above, or
* [x] See steps below.

1. Add an All Products Block to a page, add the Filter Products by Attribute Block, and Filter by Price Block too.
2. Go to the page and use the filters, ensure they update the All Products Block correctly.

### Changelog

Enhancements
> Allow Store API to filter products by custom taxonomies.

Bug Fixes
> Fix `not_in` operator for category, category, and attribute params.

### Release Note
**Store API**
As part of enhancing the Store API, we added a way to filter products by custom defined product taxonomies. The are two new parameters that can be passed to the API and should follow the dynamic patters `_unstable_tax_[product-taxonomy]` and `_unstable_tax_[product-taxonomy]_operator`. They work the same way as `category` and `category_operator` but in this case you should replace `[product-taxonomy]` with any custom defined product taxonomy. eg
`?_unstable_tax_hot_categories=1,2,3&_unstable_tax_hot_categories_operator=in` 
will return all products appearing in `hot_categories` taxonomy for terms 1 or 2 or 3.


src/StoreApi/docs/products.md
```diff
@@ -35,34 +36,35 @@ GET /products?return_attribute_counts=pa_size,pa_color
...
+| `type`               | string  |    no    | Limit result set to products assigned a specific type.                                                                                                                     |
+| `sku`                | string  |    no    | Limit result set to products with specific SKU(s). Use commas to separate.                                                                                                 |
+| `featured`           | boolean |    no    | Limit result set to featured products.                                                                                                                                     |
+| `category`           | string  |    no    | Limit result set to products assigned a specific category ID.                                                                                                              |
+| `product-taxonomy`   | string  |    no    | Limit result set to products assigned to the term ID of that custom product taxonomy. `product-taxonomy` should be the key of the custom product taxonomy registered.      |
```
Aljullu: Nitpicking, but I feel it's odd to list `product-taxonomy` between `category` and `category_operator`.

Can we move `product-taxonomy` above `category` or, even better, after `category_operator`?

And two more questions:

* Should we also document `product-category_operator`?
* And should we make it more explicit that `product-category` is a placeholder? Maybe enclosing it in square brackets or something along these lines?

Aljullu (commented):
Thanks for working on this @cpapazoglou! I definitely agree that's a good addition. I will leave the final review for @opr, I see you already discussed this implementation with him.

I left a couple of questions regarding the docs part below. I also wonder if we could add some tests (maybe [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/php/StoreApi/Routes/Products.php))?

src/StoreApi/docs/products.md
cpapazoglou: Great ideas, thanks @Aljullu 

[0de1d1d](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6152/commits/0de1d1d812d0c2bc701a5d97dfbe2fd8386e3055) moves the `product-taxonomy` after category operator, encloses it in square brackets and also documents `[product-taxonomy]_operator`.

src/StoreApi/Utilities/ProductQuery.php
```diff
@@ -95,12 +95,16 @@ public function prepare_objects_query( $request ) {
 			'and'    => 'AND',
 		];
 
+		$all_product_taxonomies = get_taxonomies( array( 'object_type' => array( 'product' ) ), 'names' );
+
 		// Map between taxonomy name and arg key.
-		$taxonomies = [
+		$default_taxonomies = [
```
tjcafferkey, line 109: Don't these `$default_taxonomies` also get returned by `get_taxonomies()`? If so, is there any need for `$default_taxonomies` and `$taxonomies = array_merge( $default_taxonomies, $all_product_taxonomies );`

cpapazoglou: Good question @tjcafferkey. `get_taxonomies( array( 'object_type' => array( 'product' ) ), 'names' );` returns all product taxonomies including `product_cat` and `product_tag`. The thing is that the endpoint currently uses `category` instead of `product_cat` and `tag` instead of `product_tag` for parameters. So I have opted merging with `$default_taxonomies` for backwards compatibility. Would be preferable updating the docs and all code references (if any)?

tjcafferkey: >The thing is that the endpoint currently uses category instead of product_cat and tag instead of product_tag for parameters.

@cpapazoglou it looks like after the merge it uses `product_cat` and `product_tag` unless I am wrong?

Screenshot 2022-04-01 at 11 32 28

cpapazoglou: 🤦, sorry for this and good catch. I was merging with wrong order. Fixed here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6152/commits/6885fd266ed8bdefeba8931158475d016af1ba19

Works in my local machine 😁
CleanShot 2022-04-01 at 14 10 03@2x

cpapazoglou: @tjcafferkey is the following preferable?
```
$all_product_taxonomies = get_taxonomies( array( 'object_type' => array( 'product' ) ), 'names' );
$all_product_taxonomies['product_cat'] = 'category';
$all_product_taxonomies['product_tag'] = 'tag';
```

tjcafferkey: Thanks @cpapazoglou this seems to work as expected now. However like @Aljullu mentioned I will leave the final review for @opr

tjcafferkey: > is the following preferable?

I think as it currently is works fine.

opr (commented):
Hey thanks @Aljullu and @tjcafferkey for reviewing this for me while I was away.

@cpapazoglou it looks great, just wondering about the operator, the AND operator doesn't seem to work for me, could you check it for me? (in and not_in seem to be fine)

- Set two products up assign two custom categories to the first one, to the second one give only one custom category.
- Go to the following URL and expect to only see the product with two categories.
`/wp-json/wc/store/products?custom_categories=35,36&custom_categories_operator=and`

senadir (changes requested):
Thank you for working on this Harris! I feel this is a much safer option for now while we look into better options into how we filter things. I'd like to request some changes first if that's possible:

- This needs a friendlier changelog, possibly: Allow Store API to filter products by custom taxonomies, unless I'm getting this PR wrong.
- This needs a dev note as well, given it's a change to a developer related page.
- It doesn't seem like we have any protection against core params, what if someone names a taxonomy `featured` or `sku`, would that conflict? can we make sure all of our core args are prioritized above custom taxonomies.

cpapazoglou:
> Hey thanks @Aljullu and @tjcafferkey for reviewing this for me while I was away.
> 
> @cpapazoglou it looks great, just wondering about the operator, the AND operator doesn't seem to work for me, could you check it for me? (in and not_in seem to be fine)
> 
> * Set two products up assign two custom categories to the first one, to the second one give only one custom category.
> * Go to the following URL and expect to only see the product with two categories.
>   `/wp-json/wc/store/products?custom_categories=35,36&custom_categories_operator=and`

👋 @opr, I agree the behaviour is very weird and decided to check what happens with the `product_cat` since it seems like a broader issue. It seems that when using multiple ids by default all children are included 
https://github.com/WordPress/WordPress/blob/master/wp-includes/class-wp-tax-query.php#L109
this results in more results for `in` and less results for `not_in`, `and`. Especially for `and`, selecting even one parent category might return 0 results as there are very few chances that a product is assigned to all children terms of that parent 🤦. Anyway, selecting leaf nodes of an hierarchical taxonomy should be enough. 

BUT, I notice another bug with custom taxonomies and multiple term ids passed in the arguments. Passing for example `4201,4202` doesn't get transformed to an array of terms and so it doesn't work as expected. For the `category` param, this happens in the `sanitize_callback` (not sure why `sanitize_callback` changes the data structure..)
https://github.com/Automattic/woocommerce.com/blob/33bb013c94ed0838f06d95b94bae41d4ec9bf792/plugins/woo-gutenberg-products-block/src/StoreApi/Routes/V1/Products.php#L273
I will be working to fix that. 

👋  @senadir
> This needs a friendlier changelog, possibly: Allow Store API to filter products by custom taxonomies, unless I'm getting this PR wrong.

Yep, very good suggestion ✅

> This needs a dev note as well, given it's a change to a developer related page.

I am not familiar with dev notes, can you point me to relevant documentation or code?

> It doesn't seem like we have any protection against core params, what if someone names a taxonomy featured or sku, would that conflict? can we make sure all of our core args are prioritized above custom taxonomies.

Very good catch! Yes it would conflict. I am thinking of prefixing the query parameter with `tax_` or `taxonomy_` and the request will look like `/wp-json/wc/store/products?tax_custom_categories=[that-category-id]` this will make sure the `key` isn't clashing with any other argument this endpoint currently accepts. What's your take?

src/StoreApi/Routes/V1/Products.php
```diff
@@ -281,6 +281,28 @@ public function get_collection_params() {
...
+					'type'              => 'string',
+					'sanitize_callback' => 'wp_parse_id_list',
+					'validate_callback' => 'rest_validate_request_arg',
+				);
+			}
+			if ( str_starts_with( $param, 'tax_' ) && str_ends_with( $param, '_operator' ) ) {
+				$params[ $param ] = array(
+					'description'       => __( 'Operator to compare product category terms.', 'woo-gutenberg-products-block' ),
+					'type'              => 'string',
+					'enum'              => [ 'in', 'not in', 'and' ],
```
opr: I am wondering if this should be `not_in` with an underscore?

src/StoreApi/docs/products.md
```diff
@@ -35,34 +36,36 @@ GET /products?return_attribute_counts=pa_size,pa_color
...
+| `sku`                         | string  |    no    | Limit result set to products with specific SKU(s). Use commas to separate.                                                                                                 |
+| `featured`                    | boolean |    no    | Limit result set to featured products.                                                                                                                                     |
+| `category`                    | string  |    no    | Limit result set to products assigned a specific category ID.                                                                                                              |
+| `category_operator`           | string  |    no    | Operator to compare product category terms. Allowed values: `in`, `not_in`, `and`                                                                                          |
+| `[product-taxonomy]`          | string  |    no    | Limit result set to products assigned to the term ID of that custom product taxonomy. `[product-taxonomy]` should be the key of the custom product taxonomy registered.    |
```
opr: Does this need to be updated so it includes the `tax_` prefix?

src/StoreApi/Routes/V1/Products.php
```diff
@@ -281,6 +281,28 @@ public function get_collection_params() {
 			'validate_callback' => 'rest_validate_request_arg',
 		);
 
+		// If the $_REQUEST contains a taxonomy query, add it to the params and sanitize it.
+		foreach ( $_REQUEST as $param => $value ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
opr, line 285: Do you think we should check whether the slug after `tax_` is a real taxonomy on the site before adding this param?

opr (commented):
Hi, sorry for the delay!

I tested and it is working with `and`, but `in` and `not in` aren't working - I also left a comment about `not in`.

It's unclear from your last comment whether you expect `in` and `not in` to be working or if you still need to work on it?

> I am not familiar with dev notes, can you point me to relevant documentation or code?

It's something we include in the release notes to describe some changes at a higher level. If you could write one for these changes (after they are finalised of course) that would go into the post.

https://developer.woocommerce.com/2022/03/15/woocommerce-blocks-7-2-0-release-notes/

Also you may want to try rebasing on `trunk` to get the E2E tests passing

src/StoreApi/docs/products.md
cpapazoglou: Yes, I just waited for validation of the logic to be used. I updated here [9d6b1a7](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6152/commits/9d6b1a744a2c06365c67746c8815121dc0a9fc7a)

src/StoreApi/Routes/V1/Products.php
cpapazoglou: Good catch! Since the `$operator_mapping` has it as `not_in` I have updated category, tag, attribute, and taxonomy operator enum from `not in` to `not_in` 
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/5be9023bcbc1918f84e9951f0acca114dcca99d4/src/StoreApi/Utilities/ProductQuery.php#L92-L96

[604cb1a](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6152/commits/604cb1a61fa34f7468d11924d9a0b81c7d800e8a)
[38d6edd](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6152/commits/38d6edd312edc24dee0259336c997970ba8fc415)

src/StoreApi/Routes/V1/Products.php
cpapazoglou: I don't see that as a problem, the parameter will be collected anyway but will be disregarded here 
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/5be9023bcbc1918f84e9951f0acca114dcca99d4/src/StoreApi/Utilities/ProductQuery.php#L101-L127

I don't have strong opinion on this though, just want to avoid two more operation (get all taxonomies, check against the parameter passed).

cpapazoglou:
> Hi, sorry for the delay!
> 
> I tested and it is working with `and`, but `in` and `not in` aren't working - I also left a comment about `not in`.
> 
> It's unclear from your last comment whether you expect `in` and `not in` to be working or if you still need to work on it?
> 
> > I am not familiar with dev notes, can you point me to relevant documentation or code?
> 
> It's something we include in the release notes to describe some changes at a higher level. If you could write one for these changes (after they are finalised of course) that would go into the post.
> 
> https://developer.woocommerce.com/2022/03/15/woocommerce-blocks-7-2-0-release-notes/
> 
> Also you may want to try rebasing on `trunk` to get the E2E tests passing

Thanks @opr  for reviewing / testing this!

`in` is working as expected for me. Bear in mind, that if you select a parent term it will add the parent term and all children terms in the query.

I hadn't tested `not in` but after making the changes you suggested here https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6152#discussion_r852873027, `not_in` seems also working fine for me. This patch should also fix a bug with all other params `not_in` operator.

|Before | After|
|-------|------|
|CleanShot 2022-04-20 at 12 26 52@2x|CleanShot 2022-04-20 at 12 25 49@2x|

cpapazoglou:
Added a release note and updated the changelog.

senadir:
I'm not sure if a `tax` prefix makes sense for now, and it would hard to roll back, but I don't think we can prioritize core params without creating confusion, I'm happy to move this forward with an `_unstable_tax` prefix instead so that we have room to change this in the future if it doesn't make sense, and also to address any future bugs we might not be seeing right now.

cpapazoglou:
> I'm not sure if a `tax` prefix makes sense for now, and it would hard to roll back, but I don't think we can prioritize core params without creating confusion, I'm happy to move this forward with an `_unstable_tax` prefix instead so that we have room to change this in the future if it doesn't make sense, and also to address any future bugs we might not be seeing right now.

Good idea @senadir, I replaced `tax_` with `_unstable_tax_` in code, changelog, and release note. [ba4b819](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6152/commits/ba4b819d7b315d38fb0e87570a699eb5535375ad)

Here are the `query` args returned
CleanShot 2022-04-20 at 13 03 40@2x

senadir (approved):
Thank you for all the effort and back and forth with us @cpapazoglou 
Let's merge this!

PR #6154, opened 2022-03-29
Fix E2E Tests on trunk

alexflorisca:
Fixing E2E tests in CI. We merged a [PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6101) that has some failing e2e tests, this is to rectify that

If all is green, all is good :)


tarhi-saad (approved):
Thank you, @alexflorisca, for the fix! All the checks did pass, so we are good to go! 🚀

So, the problem was that the translated text didn't render right away?

alexflorisca:
Yeah it looks like some of the selectors took a bit longer to render. I thought the default timeout was 30s, not 500ms though so that's a bit confusing but I didn't dig into it too much this time. If they fail again due to timeouts, we can look into this further

PR #6157, opened 2022-03-29
Fix: Shopping Button default hover and focus style

dinhtungdu:
Follows up #6130

This PR adds the default hover and focus style for the Shopping Button. This fixes the issue that the button disappears on hover/focus when only the background color is set.

### Screenshots

<table>
<tr>
	<td> Before
	<td> Screen Shot 2022-03-29 at 21 39 27
<tr>
	<td> After
	<td> Screen Shot 2022-03-29 at 21 38 48

</table>

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Edit the Mini Cart Contents block.
2. Set only the background color.
3. On the front end, hover over the Start shopping button.
4. See the button is still visible, with black background and border.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.

### DIFF
--- a/assets/js/blocks/cart-checkout/mini-cart/style.scss
+++ b/assets/js/blocks/cart-checkout/mini-cart/style.scss
@@ -171,7 +171,15 @@ h2.wc-block-mini-cart__title {
 	a {
 		border: 2px solid;
 		color: currentColor;
+		font-weight: 600;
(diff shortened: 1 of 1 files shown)


Aljullu (approved):
LGTM

PR #6158, opened 2022-03-29
Mini Cart block - fix translations handling #6153

### LINKED ISSUE #6153: TypeError: wp.i18n is undefined in new Mini Cart Block
## Describe the bug
With the new MiniCart Block, its inline JavScript for translation purposes is loaded before wp object and i18n object, so a few TypeError are thrown and no translation is applied.

wp.i18n.setLocaleData( localeData, domain )

## To reproduce

1. Go to **Settings -> General** and change the language of the site to any language other than English (FYI I'm using Spanish)
2. Now go to **Appearance -> Widgets** and add **Mini Cart** widget to any Widget Area
3. Go to any page containing that Widget Area and see error in browser console

## Expected behavior

Widget string to be translated

### WordPress:

-   Core version: 5.9.2
-   WooCommerce version: 6.3.1
-   WooCommerce Blocks version: 7.3.0
-   Site language: Spanish
-   Any other plugins installed: None

### Desktop:

-   OS: Windows
-   Browser Firefox, Chrome

gigitux:
This PR fixes how the Mini Cart block loads the translations.

Fixes #6153 

### Technical details

I guess that the current method of how the Mini Cart block loads [the translations is not correct](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/src/BlockTypes/MiniCart.php#L267). It seems that all the blocks using the method `register_chunk_translations`.

@dinhtungdu noticed that the function `append_script_and_deps_src` is used to lazy load the dependencies of Mini Cart and payment method scripts. Are we sure that the line about the translation is necessary?

Another doubt that I have: what are the advantages to split the handling of the translations compared to the other scripts with the suffix `before`? ([here the code](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/ab91245382ef23705d4802f23809e22b482a8d22/assets/js/base/utils/lazy-load-script.ts#L99-L110))

### Manual Testing

Check out this branch

1. Go to **Settings -> General** and change the language of the site to any language other than English (my suggestion is to use Dutch because all the translations are available).
2. Enable a classic theme (for example Storefront)
3. Now go to **Appearance -> Widgets** and add `Mini Cart` widget to any Widget Area.
4. Go to any page containing that Widget Area and check that any error related to `wp.i18n`.
5. Check that the `Mini Cart` is translated (check with an empty cart and with a filled cart). Please, be sure that you are using a language that has translations for the Mini Cart Block. My suggestion is to use the Dutch language.
6. Enable a block theme (for example TT2).
7. Now go to **Appearance -> Editor** and add the `Mini Cart` block in the header. Save.
8. Check that the `Mini Cart` is translated (check with an empty cart and with a filled cart). Please, be sure that you are using a language that has translations for the Mini Cart Block. My suggestion is to use the Dutch language.

### User Facing Testing
* [x] Same as above

### Changelog

> Mini Cart block: Fix translations loading

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -161,8 +161,9 @@ class MiniCart extends AbstractBlock {
 		}
 
 		$this->scripts_to_lazy_load['wc-block-mini-cart-component-frontend'] = array(
-			'src'     => $script_data['src'],
-			'version' => $script_data['version'],
+			'src'          => $script_data['src'],
+			'version'      => $script_data['version'],
+			'translations' => $this->get_dependencies_translations(),
 		);
 
 		$this->asset_data_registry->add(
@@ -443,13 +444,12 @@ class MiniCart extends AbstractBlock {
 	}
 
 	/**
-	 * Register script and style assets for the block type before it is registered.
-	 *
-	 * This registers the scripts; it does not enqueue them.
-	 * The children blocks are register here because this block handles the loading of the frontend scripts.
+	 * Prepare translations for inner blocks and dependencies.
 	 */
-	protected function register_block_type_assets() {
-		parent::register_block_type_assets();
+	protected function get_dependencies_translations() {
+		$wp_scripts   = wp_scripts();
+		$translations = array();
+
 		$blocks = [
 			'mini-cart-contents-block/filled-cart',
 			'mini-cart-contents-block/empty-cart',
@@ -461,6 +461,15 @@ class MiniCart extends AbstractBlock {
 		];
 		$chunks = preg_filter( '/$/', '-frontend', $blocks );
 
-		$this->register_chunk_translations( $chunks );
+		foreach ( $chunks as $chunk ) {
+			$handle = 'wc-blocks-' . $chunk . '-chunk';
+			$this->asset_api->register_script( $handle, $this->asset_api->get_block_asset_build_path( $chunk ), [], true );
+			$translations[] = $wp_scripts->print_translations( $handle, false );
+			wp_deregister_script( $handle );
+		}
+
+		$translations = array_filter( $translations );
+
+		return implode( '', $translations );
 	}
 }
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -260,11 +260,10 @@ class MiniCart extends AbstractBlock {
 			return;
 		}
 		$this->scripts_to_lazy_load[ $script->handle ] = array(
-			'src'          => $script->src,
-			'version'      => $script->ver,
-			'before'       => $wp_scripts->print_inline_script( $script->handle, 'before', false ),
-			'after'        => $wp_scripts->print_inline_script( $script->handle, 'after', false ),
-			'translations' => $wp_scripts->print_translations( $script->handle, false ),
+			'src'     => $script->src,
+			'version' => $script->ver,
+			'before'  => $wp_scripts->print_inline_script( $script->handle, 'before', false ),
+			'after'   => $wp_scripts->print_inline_script( $script->handle, 'after', false ),
 		);
 	}
 
@@ -441,5 +440,27 @@ class MiniCart extends AbstractBlock {
 	 */
 	protected function get_cart_payload() {
 		return WC()->api->get_endpoint_data( '/wc/store/cart' );
+	}
+
+	/**
+	 * Register script and style assets for the block type before it is registered.
+	 *
+	 * This registers the scripts; it does not enqueue them.
+	 * The children blocks are register here because this block handles the loading of the frontend scripts.
+	 */
+	protected function register_block_type_assets() {
+		parent::register_block_type_assets();
+		$blocks = [
+			'mini-cart-contents-block/filled-cart',
+			'mini-cart-contents-block/empty-cart',
+			'mini-cart-contents-block/title',
+			'mini-cart-contents-block/items',
+			'mini-cart-contents-block/products-table',
+			'mini-cart-contents-block/footer',
+			'mini-cart-contents-block/shopping-button',
+		];
+		$chunks = preg_filter( '/$/', '-frontend', $blocks );
+
+		$this->register_chunk_translations( $chunks );
 	}
 }
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -447,6 +447,15 @@ class MiniCart extends AbstractBlock {
 	 * Prepare translations for inner blocks and dependencies.
 	 */
 	protected function get_dependencies_translations() {
+		/**
+		 * Temporary remove the this filter so $wp_scripts->print_translations
+		 * calls won't accident print the translations scripts for the block.
+		 *
+		 * $wp_scripts->print_translations() calls load_script_textdomain()
+		 * which calls load_script_translations() containing the below filter.
+		 */
+		remove_filter( 'pre_load_script_translations', 'woocommerce_blocks_get_i18n_data_json', 10, 4 );
+
 		$wp_scripts   = wp_scripts();
 		$translations = array();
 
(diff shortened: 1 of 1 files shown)


gigitux:
Added as reviewers @dinhtungdu and @Aljullu that have participated in the discussion

src/BlockTypes/MiniCart.php
```diff
@@ -260,11 +260,10 @@ protected function append_script_and_deps_src( $script ) {
 			return;
 		}
 		$this->scripts_to_lazy_load[ $script->handle ] = array(
-			'src'          => $script->src,
-			'version'      => $script->ver,
-			'before'       => $wp_scripts->print_inline_script( $script->handle, 'before', false ),
-			'after'        => $wp_scripts->print_inline_script( $script->handle, 'after', false ),
-			'translations' => $wp_scripts->print_translations( $script->handle, false ),
```
dinhtungdu: Looking closer, I notice that the `$wp_scripts->print_translations()` of enqueued scripts cause the JS errors; even we don't `echo` them, they are still printed.

dinhtungdu (commented):
Tested with `deutsch` and all strings are translated 🎉 . However, the translations are included in the server-rendered HTML with this approach. I think we should lazy load the translations instead. #6164 is my idea to lazy load the translations, can you please take a look?

src/BlockTypes/MiniCart.php
Aljullu: Not sure if related, but I remember being confused by something similar and it was because we hook into `pre_load_script_translations` to print the translations, which runs no matter if you `echo` them or not:
https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/2d0270cc23121f79a5d3f1759a7e7df43f1323a4/woocommerce-gutenberg-products-block.php#L239

dinhtungdu: You're the lifesaver @Aljullu! Thanks so much for the suggestion!

src/BlockTypes/MiniCart.php
```diff
@@ -442,4 +442,34 @@ protected function get_tax_label() {
...
+			$handle = 'wc-blocks-' . $chunk . '-chunk';
+			$this->asset_api->register_script( $handle, $this->asset_api->get_block_asset_build_path( $chunk ), [], true );
+			$translations[] = $wp_scripts->print_translations( $handle, false );
+			wp_deregister_script( $handle );
+		}
+
+		$translations = array_filter( $translations );
+
+		return implode( '', $translations );
+	}
```
dinhtungdu:
```suggestion
	/**
	 * Prepare translations for inner blocks and dependencies.
	 */
	protected function get_dependencies_translations() {
		/**
		 * Temporary remove the this filter so $wp_scripts->print_translations
		 * calls won't accident print the translations scripts for the block.
		 *
		 * $wp_scripts->print_translations() calls load_script_textdomain()
		 * which calls load_script_translations() containing the below filter.
		 */
		remove_filter( 'pre_load_script_translations', 'woocommerce_blocks_get_i18n_data_json', 10, 4 );

		$wp_scripts   = wp_scripts();
		$translations = array();

		$blocks = [
			'mini-cart-contents-block/filled-cart',
			'mini-cart-contents-block/empty-cart',
			'mini-cart-contents-block/title',
			'mini-cart-contents-block/items',
			'mini-cart-contents-block/products-table',
			'mini-cart-contents-block/footer',
			'mini-cart-contents-block/shopping-button',
		];
		$chunks = preg_filter( '/$/', '-frontend', $blocks );

		foreach ( $chunks as $chunk ) {
			$handle = 'wc-blocks-' . $chunk . '-chunk';
			$this->asset_api->register_script( $handle, $this->asset_api->get_block_asset_build_path( $chunk ), [], true );
			$translations[] = $wp_scripts->print_translations( $handle, false );
			wp_deregister_script( $handle );
		}

		foreach ( array_keys( $this->scripts_to_lazy_load ) as $script ) {
			$translations[] = $wp_scripts->print_translations( $script, false );
		}

		$translations = array_filter( $translations );

		// Re-add the filter.
		add_filter( 'pre_load_script_translations', 'woocommerce_blocks_get_i18n_data_json', 10, 4 );

		return implode( '', $translations );
	}
```

dinhtungdu: This code adds translations of Mini Cart dependencies back without throwing the JS error on the non-block themes. Huge props @Aljullu for the suggestion!

dinhtungdu: Also, we should review `woocommerce_blocks_get_i18n_data_json()`. That function should only return value, but it contains a `printf` call which causes the issue we're solving here.

dinhtungdu: @gigitux I corrected the suggestion in case you checked it.

dinhtungdu (approved):
LGTM!

src/BlockTypes/MiniCart.php
```diff
@@ -260,11 +261,10 @@ protected function append_script_and_deps_src( $script ) {
 			return;
 		}
 		$this->scripts_to_lazy_load[ $script->handle ] = array(
-			'src'          => $script->src,
-			'version'      => $script->ver,
-			'before'       => $wp_scripts->print_inline_script( $script->handle, 'before', false ),
-			'after'        => $wp_scripts->print_inline_script( $script->handle, 'after', false ),
-			'translations' => $wp_scripts->print_translations( $script->handle, false ),
```
Aljullu: Can you remind me why we needed to remove this line?

src/BlockTypes/MiniCart.php
```diff
@@ -442,4 +442,50 @@ protected function get_tax_label() {
...
+		foreach ( $chunks as $chunk ) {
+			$handle = 'wc-blocks-' . $chunk . '-chunk';
+			$this->asset_api->register_script( $handle, $this->asset_api->get_block_asset_build_path( $chunk ), [], true );
+			$translations[] = $wp_scripts->print_translations( $handle, false );
+			wp_deregister_script( $handle );
+		}
+
+		foreach ( array_keys( $this->scripts_to_lazy_load ) as $script ) {
+			$translations[] = $wp_scripts->print_translations( $script, false );
+		}
```
Aljullu: If we could add back the line I mentioned in my previous comment, we wouldn't need this `foreach`. Not a big deal, but I think it's better to load translations closer to the scripts they are associated with.

src/BlockTypes/MiniCart.php
```diff
@@ -442,4 +442,50 @@ protected function get_tax_label() {
...
+	 */
+	protected function get_dependencies_translations() {
+		/**
+		 * Temporary remove the this filter so $wp_scripts->print_translations
+		 * calls won't accident print the translations scripts for the block.
+		 *
+		 * $wp_scripts->print_translations() calls load_script_textdomain()
+		 * which calls load_script_translations() containing the below filter.
+		 */
+		remove_filter( 'pre_load_script_translations', 'woocommerce_blocks_get_i18n_data_json', 10, 4 );
```
Aljullu: Is this really needed? I tried removing this line and line 487 and everything seemed to keep working. But maybe I'm missing something?

Aljullu (approved):
Thanks for investigating this tricky issue, @gigitux and @dinhtungdu! I could verify this PR fixes the issue, and translations are loaded correctly, so I'm pre-approving. Besides that, I left a couple of questions of changes to help me understand the changes.

src/BlockTypes/MiniCart.php
dinhtungdu: Did you test it with Storefront? With the block themes, there is no issue. But with classic themes, when Mini Cart is inserted as a widget, there are some JS errors as mentioned in https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/6153. I've explained the reason for this temporary removal.

src/BlockTypes/MiniCart.php
dinhtungdu: In `woocommerce_blocks_get_i18n_data_json()`, we use `printf` that causes some translation scripts are always printed when we call `$wp_scripts->print_translations()` regardless of the value of the second argument (which is `false` in this case). We can keep this line but we will need to remove the filter temporarily as we did in `get_dependencies_translations()`. I want to remove it only one time so I combine the translation of dependencies in `get_dependencies_translations()`.

> I think it's better to load translations closer to the scripts they are associated with.

I agree with this. I've just thought about a better way to remove the filter (and add it back).

dinhtungdu:
@Aljullu I updated the PR to load translations close to the associated scripts. Please take another look! Thanks so much!

Aljullu (approved):
Thanks for the explanations @dinhtungdu!

:ship:

Issue #6155, opened 2022-03-29
Hide the `StoreNoticesContainer`s if `isExpressPaymentMethodActive` is `true`


opr:
As part of the data store decoupling project, we are removing `StoreNoticeContext`. Within this, a `StoreNoticeContainer` is rendered, but since we're removing it we will display `StoreNoticeContainer`s without the context, getting the notices from the data store instead of from context.

There is one blocker that is the `StoreNoticeContext` contains a variable: `isSuppressed` which is used to hide the `StoreNoticeContainer` - this seemingly depends on only one condition: whether the `isExpressPaymentMethodActive` is true - this is stored in the `PaymentMethodDataContext` which will also be moved to a data store during this cycle. When this is complete, the `StoreNoticesContext` and `StoreSnackbarNoticesContext` can be fully removed.

PR #6159, opened 2022-03-29
Remove `useStoreNotices` and interact directly with data store instead

### LINKED ISSUE #6100: Extract logic from `StoreNoticesProvider` into a hook and remove the context provider
We should take the current `StoreNoticesProvider` and turn it into a hook that can be used without requiring the consuming component to be within the context provider.

Internally this uses the `core/notices` data store anyway so we don't need to recreate any data stores or actions/selectors.

As a follow up (or as part of this issue) we could move the `StoreNoticesContainer` out of the `base/context/providers` folder and somewhere else that seems more sensible (maybe `base/components`) then the https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/89b5d3032aa9d8b7368ba4edd3da222d076fbcaa/assets/js/base/context/providers/store-notices folder should be able to be deleted.

opr:
This PR will remove a lot of the items within `StoreNoticesContext` (except one value - `isSuppressed` - which will be moved later when #6155 is closed). The notice system is provided by `@wordpress/notices` and a data store from `@wordpress/data` so we can interact directly with this store instead of confusing hooks and contexts.

- Remove `useStoreNotices` hook.
- Strip out everything in `StoreNoticesContext` besides `isSuppressed`.
- Change value in the context to only `isSuppressed` and `setIsSuppressed`.
- Add a util function (`hasNoticesOfType`) that will check if a notice for a given context and given type exists in the `core/notices` store.
- Add a util function (`removeNoticesByStatus`) which will remove all notices with a given status.
- Add tests for `hasNoticesOfType` and `removeNoticesByStatus`.
- Make `StoreNoticesContainer` a named export, so it can be imported from `@woocommerce/base-context`.
- Update all uses of `useStoreNotices` so that it gets data from the `core/notices` store directly.
- Add a `StoreNoticeContainer` to the All Products block and the Single Product block.
- Set contexts in the error notices generated by both of these blocks.

Fixes #6100

#### Automated Tests
  * [x] Unit tests
  * [ ] E2E tests
### Manual Testing

#### Testing scenario 1
1. Add a coupon to your store.
2. Add items to your cart.
3. Go to the Cart block, and apply the coupon. Ensure you see the snackbar notice for successful coupon application.
4. **In a new tab**, open the Cart block again and remove the coupon. Ensure you see the snackbar notice for successful coupon removal.
5. Go back to the original tab and try removing the coupon, ensure you see an error message above the Cart.
6. Repeat the same steps but for the Checkout block.

#### Testing scenario 2
1. Go to the Checkout block and enter an invalid address, for example make the postcode an incorrect format for the country you've selected.
2. Ensure an error appears above the block.
3. Correct the error and ensure the notice disappears.

### Testing scenario 3

1. Enable Express Payment Methods, this is easiest done using the stripe plugin.
2. Add items to your cart and go to the Checkout block.
3. Enter an invalid postcode (for example in the UK enter `123`).
4. When the error appears above the block, click the express payment button. The error should be hidden while the express payment method modal is open.
5. Close the modal and expect the errors to reappear.

### Testing scenario 4
1. Make **two items** in your store only be able to be sold individually by checking the `Enable this to only allow one of this item to be bought in a single order` checkbox in the inventory section of the Edit product page.
2. Go to the All Products block.
3. Add this item to your cart.
4. Expect no errors.
5. Add the same item.
6. Expect an error at the top of the page.
7. Add it again, while the request is processing the error should disappear.
8. Create a new page and add a Single Product block for _each_ product you set to be individually sold above.
9. Visit the page and try to add each product more than once. Ensure an error appears for each block, and they are independent of each other (can be cleared independently etc.)
10. Try to add one of the products to the cart while the error is showing, it should disappear during the request.

#### Testing scenario 5
1. This requires you to have a development copy of the Stripe plugin, go here and make it return an error message instead of an empty string.
https://github.com/woocommerce/woocommerce-gateway-stripe/blob/8f181abe9ad8321ebdd09a494ed0e570b2f2c448/client/blocks/payment-request/hooks.js#L111
2. Go to the Cart block and ensure you have express payments set up.
11. Click the express payment button, ensure no errors appear while the modal is displaying.
12. Close the modal without paying and ensure you see your error message in the express payment method section.

### User Facing Testing
* [x] Same as above, **EXCEPT SCENARIO 5**


assets/js/base/context/providers/cart-checkout/checkout-processor.js
```diff
@@ -58,7 +58,8 @@ const CheckoutProcessor = () => {
 		paymentMethods,
 		shouldSavePayment,
 	} = usePaymentMethodDataContext();
-	const { addErrorNotice, removeNotice, setIsSuppressed } = useStoreNotices();
+	const { setIsSuppressed } = useStoreNoticesContext();
```
alexflorisca, line 61: Why can't we move `setIsSurpressed` to the store as well?

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
```diff
@@ -38,10 +39,10 @@ import {
 	reducer as emitReducer,
 } from './event-emit';
 import { useValidationContext } from '../../validation';
-import { useStoreNotices } from '../../../hooks/use-store-notices';
 import { useStoreEvents } from '../../../hooks/use-store-events';
 import { useCheckoutNotices } from '../../../hooks/use-checkout-notices';
 import { useEmitResponse } from '../../../hooks/use-emit-response';
+import { removeNoticesByStatus } from '../../../../../utils/notices';
```
alexflorisca, line 45: Could this not be an action creator instead of a utils function?

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
```diff
@@ -311,11 +315,16 @@ export const CheckoutStateProvider = ( {
...
-								: undefined;
-							addErrorNotice(
+						if (
+							errorResponse.message &&
+							isString( errorResponse.message )
+						) {
+							const errorOptions =
+								errorResponse.messageContext &&
+								isString( errorResponse.messageContext )
+									? { context: errorResponse.messageContext }
```
alexflorisca, line 325: How come you cast this to `as string` above, but not here?

assets/js/base/context/providers/store-notices/components/store-notices-container.js
```diff
@@ -23,18 +25,31 @@ const getWooClassName = ( { status = 'default' } ) => {
...
+	const { isSuppressed } = useStoreNoticesContext();
+
+	const { notices } = useSelect( ( select ) => {
+		const store = select( 'core/notices' );
+		return {
+			notices: store.getNotices( context ),
+		};
+	} );
+	const { removeNotice } = useDispatch( 'core/notices' );
+	const regularNotices = notices
```
alexflorisca, line 42: Instead of getting all notices above and filtering here, could we not get the `snackbar` notices directly from the store?

assets/js/blocks/checkout/block.tsx
```diff
@@ -28,6 +30,8 @@ import CheckoutOrderError from './checkout-order-error';
 import { LOGIN_TO_CHECKOUT_URL, isLoginRequired, reloadPage } from './utils';
 import type { Attributes } from './types';
 import { CheckoutBlockContext } from './context';
+import { hasNoticesOfType } from '../../utils/notices';
```
alexflorisca, line 33: Could this also not be an action creator?

assets/js/blocks/shared/payment-methods/express-payment/cart-express-payment.js
```diff
@@ -57,11 +57,10 @@ const CartExpressPayment = () => {
 			>
 				<div className="wc-block-components-express-payment wc-block-components-express-payment--cart">
 					<div className="wc-block-components-express-payment__content">
-						<StoreNoticesProvider
+						<StoreNoticesContainer
```
alexflorisca: Just checking this doesn't need to be wrapped in `<StoreNoticesProvider>`? I'm assuming we don't need to wrap it if the notice isn't dismissible

assets/js/blocks/shared/payment-methods/express-payment/checkout-express-payment.js
```diff
@@ -81,17 +81,16 @@ const CheckoutExpressPayment = () => {
 						</Title>
 					</div>
 					<div className="wc-block-components-express-payment__content">
-						<StoreNoticesProvider
+						<StoreNoticesContainer
```
alexflorisca: Again, just checking this doesn't need to be wrapped in `<StoreNoticesProvider>`?

alexflorisca (commented):
Good work mate, hell of a PR! Left a few comments inline, mainly with questions or minor suggestions, but overall it looks ace.

Re the tests:

**Scenario 1** - the snackbar shows fine when I add and remove the coupon, but I don't see an error when trying to remove the coupon from the original tab. Both in Cart and Checkout. I'm in incognito in chrome, signed in as admin.

Scenario 2 - ✅
Scenario 3 - ✅ 
Scenario 4 - When on the All Products block page, the error doesn't disappear when the request is processing. On the page with 2 single products it does.
Scenario 5 - I spent 30mins trying to get stripe-payment-gateway to build lcoally but ran into all sorts of trouble so I'd like to avoid going down this rabbit hole if possible. Maybe someone else would be able to try without too much headache?

assets/js/base/context/providers/cart-checkout/checkout-processor.js
opr: The reason is:

1) This is not a value on the `core/notices` store. This is a store created in Gutenberg and we don't control what shape the state gets. I don't think this is something we could make a PR to Gutenberg for as it is something specific to our use-case and would not add value to Gutenberg.
2) Its value is `true` if the express payment method is active and `false` otherwise, so we can derive this value from the payment method data store when this the work to convert that is done. We need to create a follow up issue (but I think GitHub will do this automagically based on the TODO comment!)

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
opr: We don't get to choose what actions are on this store, but I've actually got a PR open (https://github.com/WordPress/gutenberg/pull/39940) on Gutenberg which will add a way to bulk remove notices. I'll open a PR (or update my one) to add an action to remove by status too, and if it gets approved we can return here and use it :)

assets/js/blocks/checkout/block.tsx
opr: This would actually be a selector in `@wordpress/data` land, but again we don't control the data store, but it could definitely be worth improving the data store and adding this to my PR (https://github.com/WordPress/gutenberg/pull/39940)

assets/js/base/context/providers/cart-checkout/checkout-state/index.tsx
opr: Because the `errorResponse` in the code above is given a type of `Record< string, unknown > | null` and the `Response` gets assigned to it, the one below is operating directly on the `Response`.

assets/js/base/context/providers/store-notices/components/store-notices-container.js
opr: Not sure I follow you here, we can only indiscriminately `getNotices` from the store, so filtering must be done. Again, maybe this is a selector that can be added to Gutenberg?

assets/js/blocks/shared/payment-methods/express-payment/checkout-express-payment.js
opr: It is already further up :)
image

assets/js/blocks/shared/payment-methods/express-payment/cart-express-payment.js
opr: Can't show a screenshot because it's too big, but this is inside one already.

alexflorisca (approved):
Nice, thanks for fixing the first scenario. Eveything looks good after our call, got all scenarios to pass. Let's merge it!

PR #6164, opened 2022-03-30
Mini Cart block - fix translations handling

dinhtungdu:
This PR lazy loads the translations of Mini Cart inner blocks instead of including them in the initial serverside rendered HTML.

### DIFF
--- a/src/BlockTypes/MiniCart.php
+++ b/src/BlockTypes/MiniCart.php
@@ -161,8 +161,9 @@ class MiniCart extends AbstractBlock {
 		}
 
 		$this->scripts_to_lazy_load['wc-block-mini-cart-component-frontend'] = array(
-			'src'     => $script_data['src'],
-			'version' => $script_data['version'],
+			'src'          => $script_data['src'],
+			'version'      => $script_data['version'],
+			'translations' => $this->get_dependencies_translations(),
 		);
 
 		$this->asset_data_registry->add(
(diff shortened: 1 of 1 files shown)


src/BlockTypes/MiniCart.php
```diff
@@ -462,5 +462,16 @@ protected function register_block_type_assets() {
 		$chunks = preg_filter( '/$/', '-frontend', $blocks );
 
 		$this->register_chunk_translations( $chunks );
```
gigitux: Can we remove this line, right?

gigitux (approved):
Thanks for showing me this code. Now I understand better the different ways to handle translations 👍 

I added a comment. I'm sure we can remove that line. What do you think?

I'm pre-approving the PR.

src/BlockTypes/MiniCart.php
dinhtungdu: My bad, it's my copy-pasta issue. Yes, we can, that's the purpose of this PR :D.

alexflorisca:
Hey guys, just stumbled across this and thought I'd mention that I found an easier way to load translations through the wp-env cli. This also means we can load them and clear them after a particular test as to not pollute the environment. You can check out how I did it for the Cart and Checkout blocks in the `beforeAll` [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/e2e/specs/shopper/cart-checkout-translations.test.js#L19).

gigitux:
> Hey guys, just stumbled across this and thought I'd mention that I found an easier way to load translations through the wp-env cli. This also means we can load them and clear them after a particular test as to not pollute the environment. You can check out how I did it for the Cart and Checkout blocks in the `beforeAll` [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/e2e/specs/shopper/cart-checkout-translations.test.js#L19).

Thanks for letting us know. This PR is not related to the tests, but to how we load the translations in the DOM for the Mini Cart block.

PR #6165, opened 2022-03-30
Fix/e2e tests timeout issue

alexflorisca:
Some of the tests were failing due to importing the wrong constant (`SIMPLE_PRODUCT_NAME` instead of `SIMPLE_VIRTUAL_PRODUCT_NAME`). 

Translation tests were failing randomly because the block hadn't finished loading when the assertions were being made. @gigitux added an extra `waitForSelector` [here](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6165/files#diff-54d86d13d8fb11ba1eedf513463e83ef15c33f3041f0a25b8b964f9b5bef58d4R50) to wait for the cart to load.

I also realised that the translation tests are failing for legitimate reasons, which are being addressed in this [PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6195)

### Testing

1. Make sure these are the only failing e2e tests:

```
Shopper → Cart → Can view translated cart & checkout blocks › should be able to view translated Cart block 

    TimeoutError: Text not found "Total panier"

      waiting for function failed: timeout 500ms exceeded

Shopper → Cart → Can view translated cart & checkout blocks › should be able to view translated Checkout block

    TimeoutError: Text not found "Coupon code"

      waiting for function failed: timeout 500ms exceeded
```

### DIFF
--- a/tests/e2e/specs/shopper/cart-checkout-payments.test.js
+++ b/tests/e2e/specs/shopper/cart-checkout-payments.test.js
@@ -2,7 +2,7 @@
  * Internal dependencies
  */
 import { shopper } from '../../../utils';
-import { SIMPLE_PRODUCT_NAME } from '../../../utils/constants';
+import { SIMPLE_VIRTUAL_PRODUCT_NAME } from '../../../utils/constants';
 
 if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 ) {
 	// eslint-disable-next-line jest/no-focused-tests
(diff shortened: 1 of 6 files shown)


alexflorisca:
Fixed flakey tests. The remaining failing test here is due to an actual bug on trunk where some of the cart inner blocks do not show up

opr (commented):
The network will be idle while the blocks are loading anyway won't it? The client-side rendering could take time, but won't necessarily be using the network. Moreover it doesn't look like this was actually added, so maybe we need to update the PR description?

alexflorisca:
Sorry @opr, there's been a lot of trial and error with this PR, I've updated the PR description now

opr (approved):
Looks good, thanks @alexflorisca!

PR #6166, opened 2022-03-30
Prevent `Featured Product` block from breaking when product is out of stock + hidden from catalog 

### LINKED ISSUE #5344: Featured Product block breaks when product is out of stock + hidden from catalog
This is a replica of https://github.com/Automattic/wp-calypso/issues/58595

----

### Quick summary

'Featured Product' block breaks in the editor when the product is out of stock _and_ the Inventory is set to hide out of stock items from the catalog. Selecting `edit` results in `This block has encountered an error and cannot be previewed.`

### Steps to reproduce

1. Open a WooCommerce Site
2. Add a Featured Product block to a page and save
3. Head to the products and set the product to **Out of Stock**
4. Under WooCommerce > Settings > Products > Inventory, check the box that says "Hide out of stock items from the catalog"
5. Return to the page with your Featured Product block, select it, and select "Edit" to choose a new product to feature

Markup 2021-11-29 at 10 04 32

Markup 2021-11-29 at 10 11 34

### What you expected to happen

The block can still be edited to choose a new product. 

- When "Hide out of stock items from the catalog" is _unchecked_, the block responds as expected. 

### What actually happened

Clicking edit results in a broken block

https://user-images.githubusercontent.com/27249804/143914004-d40f04aa-967b-4c61-9be7-57e00d209d06.mp4

### Context

Reported by a user here: 32966334-hc
Was able to recreate on my test site

### Operating System

macOS

### Browser

Mozilla Firefox

### Simple, Atomic or both?

Atomic

### Theme-specific issue?

_No response_

### Other notes

_No response_

### Reproducibility

Consistent

### Severity

Some (< 50%)

### Available workarounds?

Yes, easy to implement

### Workaround details

Removing the block and re-inserting it is a quick workaround

I can confirm this happens with WooCommerce Blocks 6.5.0

albarin:
When a product is `out of stock` and hidden from the catalog, it is not returned from the `/wc/store/v1/products` endpoint (which is used on the `Featured Product` block to display the list of available products). 
This causes the block to break when editing a `Feature Product` block configured with an `out of stock` product since it tries to find the selected product on the array returned from the endpoint and on the variations, but it's not there.

This PR prevents accessing to the `parentProduct` array when it's empty (which was causing the JS error) and also uses the `stock_status` query param on the `/products` endpoint to explicitly request all products regardless on their statuses.

Fixes https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/5344

### Manual Testing
1. Add a Featured Product block to a page and save
2. Head to the product and set the product to Out of Stock
3. Under WooCommerce > Settings > Products > Inventory, check the box that says "Hide out of stock items from the catalog"
4. Return to the page with your Featured Product block, select it, and select "Edit" to choose a new product to feature
5. Make sure the block can still be edited to choose a new product and the Out of Stock product is also available.

### Changelog

> Fix Featured Product: prevent the block from breaking for out of stock products hidden from catalog

### DIFF
--- a/assets/js/editor-components/utils/index.js
+++ b/assets/js/editor-components/utils/index.js
@@ -25,6 +25,7 @@ const getProductsRequests = ( {
 	const defaultArgs = {
 		per_page: isLargeCatalog ? 100 : 0,
 		catalog_visibility: 'any',
+		stock_status: [ 'instock', 'outofstock', 'onbackorder' ],
 		search,
 		orderby: 'title',
 		order: 'asc',
--- a/assets/js/hocs/with-product-variations.js
+++ b/assets/js/hocs/with-product-variations.js
@@ -124,7 +124,7 @@ const withProductVariations = createHigherOrderComponent(
 						p.variations &&
 						p.variations.find( ( { id } ) => id === variationId )
 				);
-				return parentProduct.length && parentProduct[ 0 ].id;
+				return parentProduct[ 0 ]?.id;
 			}
 
 			getExpandedProduct() {
--- a/assets/js/hocs/with-product-variations.js
+++ b/assets/js/hocs/with-product-variations.js
@@ -124,7 +124,7 @@ const withProductVariations = createHigherOrderComponent(
 						p.variations &&
 						p.variations.find( ( { id } ) => id === variationId )
 				);
-				return parentProduct[ 0 ].id;
+				return parentProduct.length && parentProduct[ 0 ].id;
 			}
 
 			getExpandedProduct() {
@@ -186,6 +186,7 @@ const withProductVariations = createHigherOrderComponent(
 				showVariations: false,
 			};
 		}
+
 		return WrappedComponent;
 	},
 	'withProductVariations'


assets/js/hocs/with-product-variations.js
```diff
@@ -124,7 +124,7 @@ const withProductVariations = createHigherOrderComponent(
 						p.variations &&
 						p.variations.find( ( { id } ) => id === variationId )
 				);
-				return parentProduct[ 0 ].id;
+				return parentProduct.length && parentProduct[ 0 ].id;
```
tomasztunik: We could simplify it with optional chaining operator

```suggestion
				return parentProduct[ 0 ]?.id;
```

assets/js/editor-components/utils/index.js
```diff
@@ -25,6 +25,7 @@ const getProductsRequests = ( {
 	const defaultArgs = {
 		per_page: isLargeCatalog ? 100 : 0,
 		catalog_visibility: 'any',
+		stock_status: [ 'instock', 'outofstock', 'onbackorder' ],
```
tomasztunik, line 28: This nicely pulls in out of stock items 👌.

But it doesn't exclude hidden items - now the question is do we want to show hidden items in spirit of what I mentioned where user should be able to design pages for future use - Or should we protect user from publishing things that are hidden? 

We do not have a system that would prevent such page/blocks to be published.

Also it seems if we create a block with a product and then set its visibility to `Hidden` the block with product will still render normally and won't be hidden on a published page. 

As a user I'd be confused that I marked something hidden and it's still showing up.

So in general this is an improvement I think question is if we want to tackle hidden as part of this or separate issue? I feel like handling hidden products might be a bigger thing!

tomasztunik (changes requested):
Works as advertised! :) 

Minor improvement opportunity and some extra thoughts on handling hidden products.

assets/js/editor-components/utils/index.js
albarin: Good question. Let's tackle the hidden product's issue separately, I'll follow up in a P2. 👍

tomasztunik (approved):

PR #6173, opened 2022-03-31
Ensure e2e tests work with TS

### LINKED ISSUE #6067: Check that Jest supports TS
To be consistent with our TypeScript migration, we should start to write tests in Typescript. 
We need to check that the current Jest configuration supports Typescript. If not, we have to update it.

tomasztunik:
Convert example E2E test to TypeScript and verify it runs ok.

Bonus: clean up unnecessary default value override in e2e jest config for [moduleFileExtensions](https://jestjs.io/docs/configuration#modulefileextensions-arraystring) to ensure also `tsx` and other valid modules are supported. 

Fixes #6067 

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests
  
  
 E2E should pass

### DIFF
--- a/tests/e2e/config/jest.config.js
+++ b/tests/e2e/config/jest.config.js
@@ -4,9 +4,6 @@ module.exports = {
 	// Automatically clear mock calls and instances between every test
 	clearMocks: true,
 
-	// An array of file extensions your modules use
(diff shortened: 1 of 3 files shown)


gigitux (approved):
Great! Thanks for checking! :)

PR #6174, opened 2022-03-31
Shopper → Checkout → Can apply single-use coupon once

### LINKED ISSUE #5773: Critical flows: Shopper → Checkout → Can apply single-use coupon once
### Story

As a shopper, I want to be able to apply a single-use coupon so that I have to pay less.

### File

`tests/e2e/specs/shopper/checkout-coupons.test.js`

### Precondition

To run the following tests, a single-use coupon is required.

#### Singe-use coupon can be applied once

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Apply the single-use coupon.
5. Verify that the discount had been applied correctly on the checkout page.
6. Place the order.
7. Verify that the discount had been applied correctly on the order confirmation page.

#### Singe-use coupon cannot be applied twice

1. Go to the product page.
2. Add a product to the cart.
3. Go to the checkout page.
4. Apply the same single-use coupon as in the first test case.
5. Verify that an error appears, showing that the coupon can no longer be used.

tarhi-saad:
Fixes #5773

### Notes

This PR adds an e2e test to verify that a "single-use" coupon can be used once.

#### Automated Tests
  * [ ] Unit tests
  * [x] E2E tests

### Manual Testing

If the E2E tests run by GitHub actions are successful in this PR then it's all good.

As an extra, if you want to test it locally as well:

1. Check out this PR.
2. Start Docker Desktop.
4. Run `npm run test:e2e -- tests/e2e/specs/shopper/checkout-single-use-coupon.test.js` to run only this test case.

### Rest API instead of Fixture

We are creating our "single-use" coupon using the Woo's Rest API instead of including it in the fixture data (`tests/e2e/fixtures/fixture-data.js`) to avoid using it in other tests.


tests/e2e/specs/shopper/checkout-single-use-coupon.test.js
```diff
@@ -0,0 +1,76 @@
...
+	it( 'allows checkout to apply single-use coupon once', async () => {
+		await shopper.goToShop();
+		await shopper.addToCartFromShopPage( SIMPLE_PRODUCT_NAME );
+		await shopper.block.goToCheckout();
+
+		// Apply the single-use coupon.
+		await shopper.block.applyCouponFromCheckout( couponCode );
+
+		// Verify that the discount had been applied correctly on the checkout page.
+		// Selectors
```
nielslange: I'd probably remove a few of these inline comments, as I don't see an added value by having them.

tests/e2e/utils.js
```diff
@@ -336,3 +342,46 @@ export function useTheme( themeSlug ) {
...
+
+/**
+ * Creates a basic coupon with the provided coupon amount. Returns the coupon code.
+ *
+ * @param {Object} [coupon] Coupon object. Default to fixed cart type and amount = 5.
+ * @param {string}  [coupon.amount] Amount to be applied. Defaults to 5.
+ * @param {string} [coupon.discountType] Type of a coupon. Defaults to Fixed cart discount.
+ * @param {number} [coupon.usageLimit] How many times the coupon can be used in total. Defaults to -1.
+ */
+export const createCoupon = async ( coupon ) => {
```
nielslange, line 370: I wonder if moving this function to https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/utils/merchant.js would make more sense, as it's something the merchant would do. Similar to having `applyCouponFromCheckout()` located in https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/utils/shopper.js.

nielslange (commented):
Nice PR! Works as expected! 👏 I just added a few comments of things I noticed.

tests/e2e/utils.js
tarhi-saad: I thought about that too! But, from what I understood based on our file structure and Woo's as well, the functions that we add to the `merchant` should describe its flow (i.e., Describing how a merchant would take a specific action on the admin page). And the `createCoupon` function uses Rest API instead. So, this isn't how a merchant would create a coupon.

tests/e2e/specs/shopper/checkout-single-use-coupon.test.js
tarhi-saad: Thank you! I've cleaned up the comments! 👍

tests/e2e/utils.js
nielslange: Ah, right. That makes sense. I haven't seen it from that angle. Thanks for explaining that, @tarhi-saad.

nielslange (approved):
LGTM! 🚢

tests/e2e/specs/shopper/checkout-single-use-coupon.test.js
```diff
@@ -0,0 +1,67 @@
...
+import { getOrderDetailsDiscountPathExpression } from '../../../utils/path-expressions';
+
+if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 )
+	// eslint-disable-next-line jest/no-focused-tests
+	test.only( `Skipping checkout tests`, () => {} );
+
+let couponCode;
+
+beforeAll( async () => {
+	couponCode = await createCoupon( { usageLimit: 1 } );
```
ralucaStan: We usually clear the setup in the afterAll branch, should we also remove this coupon that was created?

tests/utils/shopper.js
```diff
@@ -390,6 +390,21 @@ export const shopper = {
...
+		applyCouponFromCheckout: async ( couponCode ) => {
+			const couponInputSelector =
+				'#wc-block-components-totals-coupon__input-0';
+			const couponApplyButtonSelector =
+				'.wc-block-components-totals-coupon__button';
+			const couponExpandButtonSelector =
+				'.wc-block-components-totals-coupon button';
+
+			await expect( page ).toClick( couponExpandButtonSelector );
+			await page.waitForSelector( couponInputSelector );
```
ralucaStan: this shouldn't be needed, expect-puppeteer will wait for element before running an action

tests/utils/path-expressions.js
```diff
@@ -34,3 +34,7 @@ export const getCartItemPathExpression = ( productTitle, args ) =>
 	' and ' +
 	getQtyItemPathExpression( args ) +
 	']';
+
+export const getOrderDetailsDiscountPathExpression = ( amount ) => {
+	return `//tr[th[contains(text(), "Discount:")] and //span[contains(text(), "${ amount }")]]`;
+};
```
ralucaStan: What do you think about replacing this with a more human friendly check:
```
	await expect(page).toMatchElement(
				`th`,
				{ text: 'Discount' }
			);
       await expect(page).toMatchElement(
				`span.woocommerce-Price-amount`,
				{ text: amount }
			);
```

ralucaStan (commented):
👏 I left some comments, let me know what you think.

tests/e2e/specs/shopper/checkout-single-use-coupon.test.js
tarhi-saad: Thank you for pointing this out, @ralucaStan! Now, we have `deleteCoupon()` in the `afterAll` block.

tests/utils/shopper.js
tarhi-saad: Great, Thanks! I removed this line of code, and the test is still successful! 👍

tests/utils/path-expressions.js
tarhi-saad: This is indeed a better way to do the check, @ralucaStan! Thank you for this brilliant idea. It will be easier to refactor the code if we decide to change the HTML structure one day. 🙌

PR #6181, opened 2022-04-01
Enable more style options for the Featured Product block

### LINKED ISSUE #6150: Update the Featured Product block with more styles/options
For more information see pdnLyh-Jv-p2#comment-449

sunyatasattva:
This PR adds several style options to the Featured Product block. The current status of features supported and planned for each block can be seen in [this spreadsheet](https://docs.google.com/spreadsheets/d/1KsKoq3J86s4PRZoHufj_qshGl_tTWjfolJXsrqE2AsU/edit#gid=1253950386).

**This PR fixes the following problems:**

* Previously, the overlay of this block was actually rendered as a background. That meant inconsistent behavior in certain circumstances (like when an external padding was added). The overlay is now a separate DOM element.
* Previously, the overlay color was controlled within the `Color` section of the sidebar, while its opacity was controlled elsewhere. This PR consolidates the controls under the `Overlay` section.
* Previously, the featured product image was rendered as a background. This meant we could not correctly apply things like duotone filters (see #5555). The image is now rendered as a separate DOM element.
* Previously, the featured image was resized constraining it to the block size. This meant that the focal point could only be moved vertically. Now the focal point can be moved correctly (see below: **Image fit options**).
* The resizable container handle was not completely visible (it would only show half a circle). This is fixed.
* The resizable container handle was always visible regardless of the selected status of the block.
* Previously, if the padding was large enough, the content could be pushed out of the container. Now the block behaves more like the Cover block, in the sense that the padding will scale the container up. This also means that the resizable control is subject to this rule, and can't resize the container lower than a certain minimum (determined by content and size). This is the same behavior as the Cover block.

**This PR adds the following features:**

* **Support for Duotone:** users can now set a duotone filter on the featured product image.
* **Support for gradients:** users can now set the overlay color as a gradient.
* **Support for padding:** users can now set custom padding for the block. Which element to add the padding to was a bit unclear in the specs and was pending an answer [here](pdnLyh-Jv-p2#comment-494). While waiting for an answer, I decided to implement it in the way that made most sense to me after some experimentation, following the example of the *Cover* block. This allows the user to have creative solution like the one screenshotted below. Please @vivialice let me know if you are ok with this decision.
* **Image fit options:** The user can now decide how should the image behave on the resizing of the component; it can either scale to always cover the entire container, or remain its original size.

Closes #6150

### Screenshots

Screen Shot 2022-04-06 at 00 52 23

Notice how, thanks to the fixed focal point picker and how the padding was implemented, I can set my image on the top right and my text on the bottom left instead of being constricted to everything being centered.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests

### Manual Testing

1. Use a block theme (such as Twenty Twenty Two)
2. Add the ”Featured Product” block to a page and select a product.
3. Make sure that the resizable handle is a shown fully (it should be a full circle), and that is shown only when the block is selected.
4. Select the block and apply a Duotone filter.
5. Make sure the overlay color control is under the `Overlay` section and change the overlay color to a gradient.
6. Add a padding to the block and make sure the padding is added on the inside of the wrapper (i.e. no white space is created outside of the block).
7. Move around your image focal point and make sure you can actually move around the entire image and not only constrained on one dimension.
8. Apply “Image fit > Cover” and notice how the image will scale along with the container. As opposed to point 5 above, now the focal point picker is limited to one dimension.
9. Increase the `padding-top` of the block and notice that the content can't get pushed out of the container, but instead the container resizes.
10. When that's the case, try resizing the block through the handle to a lower height: the handle should move but the container should not resize. When the mouse is lifted, the handle should return to its original position.
11. For the steps 4–9 above, save the page and check that the styles are applied correctly on the front-end (as the block is rendered statically via PHP).

### User Facing Testing
* [x] Same as above

---

**❗  Note:** Since this PR includes several changes, [we agreed that it should have a multi-line changelog](p1649199937104809-slack-C02SGH7JBGS). I checked the automation script which creates the changelog, and it seems to be fine with that. There might be some formatting issues with newlines and such, so perhaps, _release lead_, you might have to adjust a few things before publishing.

### Changelog

>  Enhanced the *Featured Product block*:
> 
> * Implemented support for duotone.
> * Implemented support for gradients on the overlay.
> * Implemented support for custom inner padding.
> * Implemented image fit options: users can now decide how should the image behave on the resizing of the component; it can either scale to always cover the entire container, or remain its original size.
> * Fixed an inconsistency where the overlay color was controlled by the background color control. It is now moved to the correct section.
> * Fixed the focal point picker, it now works on both axis as long as the image fit (above) is set to `none`.
> * Fixed an issue with the visibility of the resizing handle.
> * Fixed an issue which would keep the resizing handle always active regardless of block selection status.
> * Changed the behavior of the resizing: the block can't be resized below a minimum height determined by its content plus the padding.

### DIFF
--- a/assets/js/blocks/featured-product/index.js
+++ b/assets/js/blocks/featured-product/index.js
@@ -120,6 +120,20 @@ registerBlockType( 'woocommerce/featured-product', {
 		},
 
 		/**
+		 * A minimum height for the block.
+		 *
+		 * Note: if padding is increased, this way the inner content will never
+		 * overflow, but instead will resize the container.
+		 *
+		 * It was decided to change this to make this block more in line with
+		 * the “Cover” block.
+		 */
+		minHeight: {
+			type: 'number',
+			default: getSetting( 'default_height', 500 ),
+		},
+
+		/**
 		 * Text for the product link.
 		 */
 		linkText: {
--- a/assets/js/blocks/featured-product/index.js
+++ b/assets/js/blocks/featured-product/index.js
@@ -38,7 +38,15 @@ registerBlockType( 'woocommerce/featured-product', {
 	supports: {
 		align: [ 'wide', 'full' ],
 		html: false,
-		color: true,
+		color: {
+			__experimentalDuotone:
+				'.wc-block-featured-product__background-image',
+			background: true,
+			text: true,
+		},
+		spacing: {
+			padding: true,
+		},
 		...( isFeaturePluginBuild() && {
 			__experimentalBorder: {
 				color: true,
--- a/assets/js/blocks/featured-product/index.js
+++ b/assets/js/blocks/featured-product/index.js
@@ -41,11 +41,12 @@ registerBlockType( 'woocommerce/featured-product', {
 		color: {
 			__experimentalDuotone:
 				'.wc-block-featured-product__background-image',
-			background: true,
+			background: false,
 			text: true,
 		},
 		spacing: {
 			padding: true,
+			__experimentalSkipSerialization: true,
 		},
 		...( isFeaturePluginBuild() && {
 			__experimentalBorder: {
@@ -122,10 +123,33 @@ registerBlockType( 'woocommerce/featured-product', {
 		},
 
 		/**
+		 * Color for the overlay layer on top of the product image.
+		 */
+		overlayColor: {
+			type: 'string',
+			default: '#000000',
+		},
+
+		/**
+		 * Gradient for the overlay layer on top of the product image.
+		 */
+		overlayGradient: {
+			type: 'string',
+		},
+
+		/**
 		 * The product ID to display.
 		 */
 		productId: {
 			type: 'number',
+		},
+
+		/**
+		 * Product preview.
+		 */
+		previewProduct: {
+			type: 'object',
+			default: null,
 		},
 
 		/**
@@ -142,14 +166,6 @@ registerBlockType( 'woocommerce/featured-product', {
 		showPrice: {
 			type: 'boolean',
 			default: true,
-		},
-
-		/**
-		 * Product preview.
-		 */
-		previewProduct: {
-			type: 'object',
-			default: null,
 		},
 	},
 
--- a/assets/js/blocks/featured-product/index.js
+++ b/assets/js/blocks/featured-product/index.js
@@ -96,11 +96,14 @@ registerBlockType( 'woocommerce/featured-product', {
 		},
 
 		/**
-		 * A fixed height for the block.
+		 * Whether the image should fit the container or not be resized
+		 *
+		 * Note: when the image is resized to fit the container, the user loses
+		 * the ability to have full control over the focus.
 		 */
-		height: {
-			type: 'number',
-			default: getSetting( 'default_height', 500 ),
+		imageFit: {
+			type: 'string',
+			default: 'none',
 		},
 
 		/**
--- a/assets/js/blocks/featured-product/index.js
+++ b/assets/js/blocks/featured-product/index.js
@@ -46,7 +46,12 @@ registerBlockType( 'woocommerce/featured-product', {
 		},
 		spacing: {
 			padding: true,
-			__experimentalSkipSerialization: true,
+			...( isFeaturePluginBuild() && {
+				__experimentalDefaultControls: {
+					padding: true,
+				},
+				__experimentalSkipSerialization: true,
+			} ),
 		},
 		...( isFeaturePluginBuild() && {
 			__experimentalBorder: {
--- a/assets/css/abstracts/_mixins.scss
+++ b/assets/css/abstracts/_mixins.scss
@@ -204,6 +204,19 @@ $fontSizes: (
 	}
 }
 
+// Positions an element absolutely and stretches it over the container
+@mixin absolute-stretch() {
+	margin: 0;
+	padding: 0;
+	position: absolute;
+	top: 0;
+	left: 0;
+	right: 0;
+	bottom: 0;
+	width: 100%;
+	height: 100%;
+}
+
 // Converts a px unit to em.
 @function em($size, $base: 16px) {
 	@return math.div($size, $base) * 1em;
--- a/assets/js/blocks/featured-product/block.js
+++ b/assets/js/blocks/featured-product/block.js
@@ -1,3 +1,5 @@
+/* eslint-disable @wordpress/no-unsafe-wp-apis */
+
 /**
  * External dependencies
  */
@@ -9,6 +11,9 @@ import {
 	InspectorControls,
 	MediaReplaceFlow,
 	RichText,
+	__experimentalGetSpacingClassesAndStyles as getSpacingClassesAndStyles,
+	__experimentalPanelColorGradientSettings as PanelColorGradientSettings,
+	__experimentalUseGradient as useGradient,
 } from '@wordpress/block-editor';
 import { withSelect } from '@wordpress/data';
 import {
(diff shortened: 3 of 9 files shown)


albarin:
@sunyatasattva While testing and comparing with _Cover_, I've noticed a couple of things:

1. In _Cover_, when you set a padding and hover over the padding size, it shows like an overlay with the size of the padding. In the _Featured Product_ this doesn't happen, do we want to keep the same behavior?

https://user-images.githubusercontent.com/186112/161951696-984c4fd7-b05e-4ebf-a348-35b650c3cb9b.mp4

2. The block settings "loses" the applied gradient settings when you open the menu again:

https://user-images.githubusercontent.com/186112/161952032-bdcad46f-8e9c-48d5-89e5-249e9f061bf7.mp4

assets/js/blocks/featured-product/index.js
```diff
@@ -38,7 +38,16 @@ registerBlockType( 'woocommerce/featured-product', {
...
-		color: true,
+		color: {
+			__experimentalDuotone:
+				'.wc-block-featured-product__background-image',
+			background: false,
+			text: true,
+		},
+		spacing: {
+			padding: true,
+			__experimentalSkipSerialization: true,
```
frontdevde: Given this is `__experimental` would we want the `spacing` to be wrapped in `isFeaturePluginBuild()` below?

vivialice:
👋 Heya! A few comments:

As @albarin mentioned above, it would be nice to keep that overlay behaviour once the padding is implemented!

The image won't fill the background by default with a smaller image. It's obviously not the best regarding image quality, but can we align with the cover block? 

Screenshot 2022-04-06 at 12 36 57

I noticed when you put a high number for padding on one side it can push the elements "off" the block. Here's a screenshot where I input the same values for the cover and the featured product block: 

Screenshot 2022-04-06 at 12 53 02

frontdevde:
Thinking out loud here: For both the Featured Product/Category blocks we want to add more advanced media control options. From what I hear @vivialice saying is that these capabilities should match the options we have on the Cover block. Instead of reinventing the wheel, could we potentially import/use the actual `core/cover` block in our blocks? 

In an ideal case would be neat if we didn't have to duplicate all the media-related functionality the Cover block already has. If we end up having to duplicate it, I'm wondering if we could achieve it in a way where we can reuse that functionality within Woo Blocks at least. It does seem like we could use it in a number of blocks if it "just works".

Screenshot 2022-04-06 at 16 20 58

assets/js/blocks/featured-product/index.js
sunyatasattva: Looks like we should! Thanks for the catch, I'm still a bit confused with the whole `__experimental` thing.

sunyatasattva:
@albarin 

> In Cover, when you set a padding and hover over the padding size, it shows like an overlay with the size of the padding. In the Featured Product this doesn't happen, do we want to keep the same behavior?

Yea, I noticed that. It's really nice and useful feature indeed. I tried to figure out how that's implemented but couldn't manage yet. I posted in #gutenberg to see if anybody can give me pointers. Otherwise I'll do some more research.

> The block settings "loses" the applied gradient settings when you open the menu again:

Nice catch! Will look into that!

---

@vivialice 

> As @albarin mentioned above, it would be nice to keep that overlay behaviour once the padding is implemented!

Yep, I agree. I'll try my best. Just a clarification: padding is already implemented the same way as in the cover block. It just doesn't show the overlay.

> The image won't fill the background by default with a smaller image. It's obviously not the best regarding image quality, but can we align with the cover block?

I've got a little qualm here. My problem is not the image quality, but rather the behavior of the focal point which I mentioned in the OP. If you see in the cover block you can't really set the focal point of your image. You can just scroll it vertically, because the image is contained to the borders of the block (so the UI feels weird and the users can't achieve what they want).

With my implementation, you can actually meaningfully change the focal point as the UI and the result actually match.

https://user-images.githubusercontent.com/1847066/162024065-7dd4d4fe-e14a-451a-ae81-12f6613b6dc5.mov

I think the solution here is to add an option for the user to constrain the image to the borders of the block (i.e. if it's too large, it will scale it down; otherwise, will scale it up). What are your thoughts?

> I noticed when you put a high number for padding on one side it can push the elements "off" the block. 

Interesting! So, the cover block scales up with the padding, I see. Good catch!

---

@frontdevde 

> For both the Featured Product/Category blocks we want to add more advanced media control options. From what I hear @vivialice saying is that these capabilities should match the options we have on the Cover block. Instead of reinventing the wheel, could we potentially import/use the actual core/cover block in our blocks?

Yea, I am trying to reuse as much logic as possible from the existing Gutenberg core APIs. Though pretty much none of them are documented, and I have to do some reverse engineering, if you look at the code, you can see a lot of stuff is imported and reused. The comments from @nerrad on the spreadsheet implied that we should look to move away from custom implementations, and that's what I am trying to do.

Just note, re: media controls, that they are not part of this MVP PR, as mentioned in [this comment](pdnLyh-Jv-p2#comment-449) by @Aljullu.

sunyatasattva:
So, I pushed a bunch of commits addressing the feedback I got. In the meantime, I also fixed other bugs I found along the way, I'll update the OP so that it's completely reflective of what this PR is currently doing. Regarding the single issues that have been brought up:

1. All `__experimental` features are now wrapped in such a way that are only available on feature builds and making sure that support for those feature is available.
2. As @vivialice requested, now the block behaves more like the Cover block when it comes to `padding` and height: if the padding is large enough that the content would be pushed down away from the container (and thus hidden), now instead the container resizes as to accommodate the content.

⚠️  *However, please note:* this means the resizeable controls aren't anymore exactly reflective of reality and _might_ confuse the user. In fact, a user can only resize the block down to the minimum allowed by the container (this is the same behavior as the Cover block).

3. Speaking of the resizable container, I noticed that the handle was showing only half of the circle (visible in all screenshots above) and was also showing when the block was not selected (which is not consistent with the Cover block). So I fixed that.
4. I fixed the problem of gradients not persisting in the UI.
5. Lastly, I added an option for the user to be able to decide how the image should scale and fit the container, as described above, and added some help text to not confuse the users: as pointed out above, if the image always scales up to the entire container, the “focal point picker” is not entirely usable anymore (you can often only shift *either* the X *or* Y axis, like in the Cover block). As this doesn't seem to me a very nice UX nor a common use-case, I added this option so that users can decide how should the block behave (I can personally more readily envision a user needing to switch the focal point of the product image to fix better in the design, or around the text, or to highlight a certain feature of the product, than to make extremely oversized featured product blocks).

As @frontdevde is off for a while, @albarin the onus of approving this PR will be on you 😸

albarin:
@sunyatasattva 
Hey, I'm testing locally and seeing some weird behavior, not sure if I have some issue on my end tbh. 🤔 

The **Focal Point picker** rectangle seems too big and there's nothing under the **Dimensions** section:
dimesions

Tried it also on the **Cover** block and the same happens on the Focal Point picker, also the **Minimum height of cover** input has the wrong height.
height

albarin (approved):

sunyatasattva:
Oh, I realize I forgot to mention a really important part of the feedback I had received: **the padding indicators**!

I had a [conversation](p1649262388321719-slack-C45SNKV4Z) with people over at Gutenberg, and I wanted to share with people who asked here and whoever comes across this PR. It turns out it's called [BoxControlVisualizer](https://github.com/WordPress/gutenberg/blob/f1f0cd7adceeecce546641dafdb2a1d0844f2bb2/packages/components/src/box-control/README.md#visualizer) and currently I have been heavily discouraged from using it.

cf. this PR: https://github.com/WordPress/gutenberg/issues/40057 where it is mentioned to be in an “unusable state”.

I have been told heavily rework is going on there and we shouldn't touch it until the rework is done. I am following the above mentioned issue so I can see when things are moving.

Cc @vivialice @albarin

PR #6191, opened 2022-04-05
Product search block as core/search variation

### LINKED ISSUE #6170: Update Search Block to register as Gutenberg Search Block variation

tomasztunik:
Implements Search Block as a variation of Gutenberg core/search block:

Compatibility:
- Ensures the legacy block work before users convert it to Core Search variation in the Editor.
- Even though WooCommerce Blocks has an L0 support policy, we need to care about backward compatibility if we want to merge https://github.com/woocommerce/woocommerce-blocks/pull/6191 to WC Core, so adding legacy block back is necessary.

Should be tested against Gutenberg ≥ 13.4

Limitations:
- Once the block is converted to `core/search` variation, it's out of our control.
- We can't transform the search widget to product search variation.

Fixes #6170

#### Accessibility

#### Other Checks

### Screenshots

https://user-images.githubusercontent.com/5423135/181692749-8ce8ea51-824a-49c3-97e3-fc1c05130fdb.mov

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests

### User Facing Testing

1. Ensure the Gutenberg version is ≥ 13.4 and it is deactivated.
2. With a block theme (2022 for example), add the Product Search block to the header.
3. See the Product Search is the independent block (as it is in `trunk`), not a variation of the Search block. See the block is editable in the editor and working as expected on the front end.
4.  Activate Gutenberg.
5. On the front end, see the block is still working as expected and is still a WC block (notice the classes, and search icon).
6. Edit the header, see the Deprecation notice, and a call to action button to upgrade the search block to search variation.
7. Click on the CTA button to see if the block is converted to WC variation of the core search block.
8. Save and see the block functions flawlessly on the front end.
9. Edit the header again, try adding the Product Search block, only variation is available in the inserter. The legacy can't be found and inserted from the inserter.

### Performance Impact
N/A

### Changelog

> Register product search as a core/search variation when available.


assets/js/blocks/product-search/index.tsx
```diff
@@ -0,0 +1,112 @@
...
+				icon={ search }
+				className="wc-block-editor-components-block-icon"
+			/>
+		),
+	},
+	// @ts-ignore waiting for @types/wordpress__blocks update
+	isActive: ( blockAttributes, variationAttributes ) => {
+		return (
+			blockAttributes.query?.post_type ===
+			variationAttributes.query.postType
```
Aljullu: I think `postType` here should be snake_case:

```suggestion
			variationAttributes.query.post_type
```

Aljullu (commented):
Leaving a small comment here as I was reviewing this PR as part of #6761.

assets/js/blocks/product-search/index.tsx
dinhtungdu: Great catch! Thank you so much!

assets/js/blocks/product-search/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+					PRODUCT_SEARCH_ATTRIBUTES.placeholder,
+				query: PRODUCT_SEARCH_ATTRIBUTES.query,
+			} )
+		);
+	};
+
+	const actions = [
+		<Button key="update" onClick={ updateBlock } variant="primary">
+			{ __(
+				'Update to new Search Block',
```
Aljullu: Can we make `Block` lowercase for consistency with the other string?

```suggestion
				'Update to new Search block',
```

assets/js/blocks/product-search/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+	const updateBlock = () => {
+		replaceBlocks(
+			clientId,
+			createBlock( 'core/search', {
+				label:
+					currentBlockAttributes?.label ||
+					PRODUCT_SEARCH_ATTRIBUTES.label,
+				buttonText:
+					currentBlockAttributes?.label ||
```
Aljullu: Is it expected that the label is also applied to the button text? I'm thinking in this case we can just use the default one in all cases.

assets/js/blocks/product-search/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+	placeholder: attributes.placeholder.default,
+	query: {
+		post_type: 'product',
+	},
+};
+
+/**
+ * editor.scss and style.scss are required
+ * to gracefully handle old block deprecation
+ */
```
Aljullu: I'm not 100% to understand this comment. You mean those files are needed for the `DeprecatedBlockEdit` component?

Aljullu (commented):
This looks fantastic, good job @tomasztunik and @dinhtungdu!

I left some minor comments below, but overall looking good:

Nit: the contents of the deprecation notice are not correctly vertically aligned (they are closer to the bottom than to the top):

imatge

assets/js/blocks/product-search/index.tsx
dinhtungdu: This is left over from the original work where we removed the block and edit component. I will remove it.

assets/js/blocks/product-search/index.tsx
dinhtungdu: Agree!

assets/js/blocks/product-search/index.tsx
dinhtungdu: I agree! It makes more sense than using the label. Sometimes labels can be long and not suitable for the submit button.

dinhtungdu (commented):
> Nit: the contents of the deprecation notice are not correctly vertically aligned (they are closer to the bottom than to the top)

I noticed this too. The reason I keep it that way is Gutenberg used a `1em` margin-top for the editor warning action button and I want to keep it aligned with other editor warning notices. But I don't have a strong opinion on this so I'm happy to change if it makes more sense.

@vivialice Can you please review the design of the deprecation notice and give me some feedback?

vivialice:
Apologies on the delay on feedback @dinhtungdu 🙏 

> Can you please review the design of the deprecation notice and give me some feedback?

Are we required to use the term `deprecated`? I feel like it's too technical for the average user. I'd suggest something like `This version of the Product Search block is outdated` > `Upgrade Block`.

Re: the design, I think there's more we could do for this block - I'm going to check if there's another pattern other teams are using. Otherwise I agree with @Aljullu 's comment on the vertical alignment.

dinhtungdu:
@vivialice @Aljullu I updated the Warning component copy and spacing in [c672048](https://github.com/woocommerce/woocommerce-blocks/pull/6191/commits/c672048a46d4b85c38e11ddc3c70e34c069b855b). Here is how it looks now:

image

vivialice:
Alignment looks good! 😄 I think I need to change the copy slightly. To confirm: the product search block will _not work_ if the user doesn't upgrade? Or will it still appear on the front end? 

If so, I think we should update to `This version of the Product Search block is outdated. Upgrade to continue using.` 

Also cc @sunyatasattva for copy alignment if required!

dinhtungdu:
> To confirm: the product search block will _not work_ if the user doesn't upgrade? Or will it still appear on the front end?

@vivialice The legacy block will continue working until users upgrade. So I think `Upgrade to continue using` is not really accurate as users are still able to use the (old) search block. What do you think?

Update: Even the (old) search block works without users' action, I think we still should use `Upgrade to continue using` to encourage users to upgrade the block. They have to do it anyway to edit the block.

Aljullu (approved):
LGTM :shipit: 

Thanks @vivialice for the design feedback and @dinhtungdu for your work moving this forward. :clap:

dinhtungdu:
@Aljullu I would like to pause merging this PR to think about the post-deprecation process.

Eventually, we want to remove the `DeprecatedBlockEdit` and the legacy Edit/Block component. So how will we do it?

My plan is we will just remove those components once WP 6.1 is released due to our L0 support policy. I'm wondering if there is any better alternative and if there is anything we can do now to make the transition easier.

Aljullu:
That's a good point, thanks for raising this, @dinhtungdu!

> Eventually, we want to remove the `DeprecatedBlockEdit` and the legacy Edit/Block component. So how will we do it?

I'm not sure if I agree to consider this a blocker, though. That component is only loaded in the editor, and its performance impact should be negligible, right?

However, I agree that it might be good to 'clean up the code at some point, I wonder if we could do something like what Gutenberg did with the Gallery block: https://github.com/WordPress/gutenberg/pull/36191. I didn't take a deep look, but they seem to have deprecated several past versions.

How do you feel about doing a 1-day research spike to see if there is anything from that PR (or from other GB blocks) that we can reuse? Based on that, we can decide whether we merge this PR as-is or not. (Even though, if possible, I would prefer to merge this PR and work in an auto-migration in another PR)

dinhtungdu:
> I'm not sure if I agree to consider this a blocker, though. That component is only loaded in the editor, and its performance impact should be negligible, right?

Yes, it's just a concern about code maintenance.

> (Even though, if possible, I would prefer to merge this PR and work in an auto-migration in another PR)

@Aljullu I agree. Let's merge this one and create an issue to track the research and development of auto-migration. It may be the solution for not only the search block but filter blocks as well.

PR #6192, opened 2022-04-05
Move mini cart outside cart-checkout folder

alexflorisca:
This is a second attempt at this, as the [previous PR](https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/6126) caused a [nasty merge conflict](p1649150426224029-slack-C02UBB1EPEF). This PR moves the `mini-cart` and  `shared` folders outside the `cart-checkout` folder and removes this. The `cart` and `checkout` blocks have already been moved outside of this folder, so it doesn't make sense anymore.

This also gets rid of some special webpack config needed to specify a custom directory instead of matching the regex used for all other blocks.

### Manual Testing

1. Smoke test the Cart and make sure everything shows up, including all inner blocks
2. Smoke tests the Mini Cart block in the same way


assets/js/base/context/providers/cart-checkout/payment-methods/test/payment-method-data-context.js
```diff
@@ -24,7 +24,7 @@ import {
 import {
 	CheckoutExpressPayment,
 	SavedPaymentMethodOptions,
-} from '../../../../../../blocks/cart-checkout/payment-methods';
+} from '../../../../../../blocks/shared/payment-methods';
```
ralucaStan: Having a `shared` folder under `/blocks` seems to me like a place that will gather common logic for a diferent blocks. Should we rename it to make it cart & checkout specific? `cart-checkout-shared` maybe?

alexflorisca: My thinking was that anything that needs to be shared between two or more blocks could live here.

ralucaStan: My comment was coming from the fact that checkout related blocks are scope wise separated from the rest, so having common parts with those won’t be the case. There is also no shared folder between the non-checkout blocks at the moment, so renaming would avoid the confusion around which blocks actually use shared functionality from the folder

alexflorisca: Ok that makes sense, happy to go with your suggestion and rename to `cart-checkout-shared`

ralucaStan (approved):
I've tested Cart, Checkout & Mini block and they work as expected. Maybe have somebody from Kirigami to take a look as well, as it touches the Mini Cart block.

ok from me!

assets/js/base/context/tsconfig.json
```diff
@@ -10,7 +10,8 @@
 		"../../base/utils/",
 		"../../data/",
 		"../../types/",
-		"../components"
+		"../components",
+		"../../blocks/shared/payment-methods"
```
dinhtungdu: Do you mean this instead?
```suggestion
		"../../blocks/cart-checkout-shared/payment-methods"
```

tsconfig.json
```diff
@@ -3,7 +3,7 @@
 	"include": [
 		"./assets/js/**/*",
 		"./assets/js/blocks/**/block.json",
-		"./assets/js/blocks/cart-checkout/mini-cart-contents/inner-blocks/**/block.json",
+		"assets/js/blocks/mini-cart/mini-cart-contents/inner-blocks/**/block.json",
```
dinhtungdu: Should we keep `./` at the beginning for uniformity?

docs/testing/releases/440.md
```diff
@@ -66,7 +66,7 @@ This test requires the browser dev tools to inspect rendered HTML.
 This test requires a code edit to force an error.
 
 1. Chose Storefront and set the background to black and text to white.
-2. Force an error to the Cart or Checkout block. Add on line 74 throw new Error( 'test' ); in `assets/js/blocks/cart-checkout/cart/block.js`
```
dinhtungdu: I don't think we need to update the release note of old versions.

dinhtungdu (approved):
Left some minor comments but this is LGTM! Thanks so much for this great work!

PR #6195, opened 2022-04-05
Register missing C&C inner blocks and update fallback template for older C& C versions

### LINKED ISSUE #6185: Not all Cart & Checkout inner blocks render when migrating to inner blocks for Order Summary
## Describe the bug

When migrating to the Cart & Checkout blocks that have Order summary with inner blocks not all content is loaded for the Order summary.

See internal conversation p1649074270986619/1649068099.800419-slack-C8X6Q7XQU
## To reproduce

1. Go to version v7.2.0
2. Build and add C & C blocks to pages
3. Check to see the content
4. Go to `git checkout trunk`
5. Build and see that some inner blocks are missing as follows:
**Cart**
- Order summary heading, Coupons & Taxes(?) inner blocks
**Checkout**
- Cart summary, Coupons & Taxes(?) inner blocks

## Expected behavior

All inner blocks should be rendered in the Order summary.

ralucaStan:
This will fix the issues with missing Order summary inner blocks: Coupons (both in C & C blocks) and the Cart header. 

The issue was happening because, for example, for Cart the coupons were registered on the on frontend, but it just wasn't forced in the attributes. Because it also wasn't added to the PHP fallback layout, the render function didn't include it. For the Checkout block the coupons inner block wasn't registered at all.

Fixes #6185 

#### Other Checks

### Screenshots

| Before  | After |
|   Screenshot 2022-04-05 at 11 53 18 |   Screenshot 2022-04-05 at 16 55 41 |

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

**Test from tag v7.2.0**

1. Run `git checkout tags/v7.2.0` 
2. Insert fresh Cart/Checkout blocks in different pages, save them
3. Notice that there is no inner blocks for Order summary
4. Switch to the PR branch & build
5. Test that Cart/Checkout blocks load fine in the frontend and the editor without crashing and the missing inner blocks are there

### User Facing Testing
* [ ] Same as above, or
* [ ] See steps below.
* [x] Not included in user facing testing instructions

### DIFF
--- a/src/BlockTypes/Cart.php
+++ b/src/BlockTypes/Cart.php
@@ -118,14 +118,13 @@ class Cart extends AbstractBlock {
 		 * Cart i3 added inner blocks for Order summary. We need to add them to Cart i2 templates.
 		 * The order needs to match the order in which these blocks were registered.
 		 */
-		$coupons_enabled                 = wc_coupons_enabled();
 		$order_summary_with_inner_blocks = '$0
 			<div data-block-name="woocommerce/cart-order-summary-heading-block" class="wp-block-woocommerce-cart-order-summary-heading-block"></div>
 			<div data-block-name="woocommerce/cart-order-summary-subtotal-block" class="wp-block-woocommerce-cart-order-summary-subtotal-block"></div>
 			<div data-block-name="woocommerce/cart-order-summary-fee-block" class="wp-block-woocommerce-cart-order-summary-fee-block"></div>
-			<div data-block-name="woocommerce/cart-order-summary-discount-block" class="wp-block-woocommerce-cart-order-summary-discount-block"></div>' .
-			( $coupons_enabled ? '<div data-block-name="woocommerce/cart-order-summary-coupon-form-block" class="wp-block-woocommerce-cart-order-summary-coupon-form-block"></div>' : '' ) .
-			'<div data-block-name="woocommerce/cart-order-summary-shipping-form-block" class="wp-block-woocommerce-cart-order-summary-shipping-block"></div>
+			<div data-block-name="woocommerce/cart-order-summary-discount-block" class="wp-block-woocommerce-cart-order-summary-discount-block"></div>
+			<div data-block-name="woocommerce/cart-order-summary-coupon-form-block" class="wp-block-woocommerce-cart-order-summary-coupon-form-block"></div>
+			<div data-block-name="woocommerce/cart-order-summary-shipping-form-block" class="wp-block-woocommerce-cart-order-summary-shipping-block"></div>
 			<div data-block-name="woocommerce/cart-order-summary-taxes-block" class="wp-block-woocommerce-cart-order-summary-taxes-block"></div>
 		';
 		// Order summary subtotal block was added in i3, so we search for it to see if we have a Cart i2 template.
--- a/src/BlockTypes/Cart.php
+++ b/src/BlockTypes/Cart.php
@@ -81,8 +81,9 @@ class Cart extends AbstractBlock {
 		wp_dequeue_style( 'select2' );
 
 		/**
-		 * We need to check if $content has any templates from prior iterations of the block, in order to update to the latest iteration
-		 * When testing for inner blocks presence, only test for blocks with the locked:{remove:true} attribute.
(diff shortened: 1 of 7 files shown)


src/BlockTypes/Checkout.php
```diff
@@ -109,6 +111,30 @@ protected function render( $attributes, $content ) {
...
+		 */
+		$coupons_enabled                 = wc_coupons_enabled();
+		$order_summary_with_inner_blocks = '$0
+			<div data-block-name="woocommerce/checkout-order-summary-cart-items-block" class="wp-block-woocommerce-checkout-order-summary-cart-items-block"></div>
+			<div data-block-name="woocommerce/checkout-order-summary-subtotal-block" class="wp-block-woocommerce-checkout-order-summary-subtotal-block"></div>
+			<div data-block-name="woocommerce/checkout-order-summary-fee-block" class="wp-block-woocommerce-checkout-order-summary-fee-block"></div>
+			<div data-block-name="woocommerce/checkout-order-summary-discount-block" class="wp-block-woocommerce-checkout-order-summary-discount-block"></div>' .
+			( $coupons_enabled ? '<div data-block-name="woocommerce/checkout-order-summary-coupon-form-block" class="wp-block-woocommerce-checkout-order-summary-coupon-form-block"></div>' : '' ) .
+			'<div data-block-name="woocommerce/checkout-order-summary-shipping-block" class="wp-block-woocommerce-checkout-order-summary-shipping-block"></div>
+			<div data-block-name="woocommerce/checkout-order-summary-taxes-block" class="wp-block-woocommerce-checkout-order-summary-taxes-block"></div>
```
opr, line 128: Do we need to do a check for if taxes are enabled here?

opr (commented):
Testing works well! 🙌🏼  Also the code changes look mostly fine, just added a question about the taxes block.

src/BlockTypes/Checkout.php
ralucaStan: Most definitely, but I don't trust myself to do this change. There is no issue if the wrapper appears when it shouldn't because it's handled on the CSS side, but we can add a ticket for this separately.

src/BlockTypes/Cart.php
```diff
@@ -105,6 +112,30 @@ protected function render( $attributes, $content ) {
...
+		 * Cart i3 added inner blocks for Order summary. We need to add them to Cart i2 templates.
+		 * The order needs to match the order in which these blocks were registered.
+		 */
+		$coupons_enabled                 = wc_coupons_enabled();
+		$order_summary_with_inner_blocks = '$0
+			<div data-block-name="woocommerce/cart-order-summary-heading-block" class="wp-block-woocommerce-cart-order-summary-heading-block"></div>
+			<div data-block-name="woocommerce/cart-order-summary-subtotal-block" class="wp-block-woocommerce-cart-order-summary-subtotal-block"></div>
+			<div data-block-name="woocommerce/cart-order-summary-fee-block" class="wp-block-woocommerce-cart-order-summary-fee-block"></div>
+			<div data-block-name="woocommerce/cart-order-summary-discount-block" class="wp-block-woocommerce-cart-order-summary-discount-block"></div>' .
+			( $coupons_enabled ? '<div data-block-name="woocommerce/cart-order-summary-coupon-form-block" class="wp-block-woocommerce-cart-order-summary-coupon-form-block"></div>' : '' ) .
```
senadir: This shouldn't be here and shouldn't be needed at all, the logic inside `cart-order-summary-coupon-form/block.tsx` should be enough. The logic to render those divs should match the logic we have inside `defaultTemplate` of `cart-order-summary-block/edit.tsx`, so if we don't have conditions inside that array, we shouldn't have logic here.

senadir: You can confirm that by inserting a V3 Cart and saving, the HTML from coupon form is always there, regardless of taxes being enabled or not.

src/BlockTypes/Checkout.php
senadir: We shouldn't check for taxes being enabled, not coupons as well. That logic lives inside the actual block.

senadir (approved):
Tested and reviewed functionality, this is looking great! 🚢

PR #6203, opened 2022-04-06
Update @woocmmerce/eslint-plugin to 2.0.0

alexflorisca:
This fixes the `npm run lint:js:report` error in CI. This was due to a conflict of eslint versions, where `@woocommerce/eslint-plugin` was using v7, and we (or another package) are using v8.

```
Oops! Something went wrong! :(

ESLint: 8.12.0

ESLint couldn't find the plugin "eslint-plugin-testing-library".
```

As part of this upgrade, `npm run lint:js` started to throw lots of errors. I've fixed most of these as part of this PR too

### Manual Testing

1. `npm ci --no-optional && npm run lint:js:report` should be successful 
2. `npm run lint:js` should show 0 errors and 5 warnings
3. Best to check all other linting scripts the depends on eslint run successfully as well


package.json
```diff
@@ -133,7 +133,7 @@
 		"@wordpress/dependency-extraction-webpack-plugin": "3.2.1",
 		"@wordpress/dom": "3.2.7",
 		"@wordpress/e2e-test-utils": "6.0.2",
-		"@wordpress/e2e-tests": "^3.1.1",
+		"@wordpress/e2e-tests": "3.1.1",
```
ralucaStan, line 136: Any reason this was also changed?

ralucaStan (approved):
👏 Amazing fix! Thank you for looking into this. All commands work as expected, file linting and commits also work.

package.json
alexflorisca: Every other dependency is pinned to a specific version, I thought I'd do the same here as I noticed the package-lock.json was being updated every time I ran `npm i` locally.

assets/js/atomic/blocks/product-elements/price/edit.js
```diff
@@ -10,7 +10,6 @@ import {
 import { __ } from '@wordpress/i18n';
 
 /**
- *
```
ralucaStan, line 13: Let's remove this all together

package.json
```diff
@@ -67,7 +67,7 @@
 		"storybook:deploy": "rimraf ./storybook/dist/* && npm run storybook:build && gh-pages -d ./storybook/dist",
 		"test": "wp-scripts test-unit-js --config tests/js/jest.config.json",
 		"test:debug": "ndb .",
-		"test:e2e": "npm run wp-env:config && cross-env NODE_CONFIG_DIR=tests/e2e/config wp-scripts test-e2e --config tests/e2e/config/jest.config.js --testPathIgnorePatterns performance",
```
ralucaStan: What is the reason behind removing these parameters?

packages/checkout/slot/index.js
```diff
@@ -2,13 +2,13 @@
  * External dependencies
  */
 import deprecated from '@wordpress/deprecated';
+import { CURRENT_USER_IS_ADMIN } from '@woocommerce/settings';
+import { Children, cloneElement } from '@wordpress/element';
 import {
 	createSlotFill as baseCreateSlotFill,
 	__experimentalUseSlot,
-	useSlot as __useSlot,
+	useSlot as __useSlot, //eslint-disable-line
```
ralucaStan, line 10: Without the comment this throws a warning, I think that's ok not to disable, it won't break anything.

`  10:2  warning  useSlot not found in 'wordpress-components'  import/named`

assets/js/base/context/hooks/cart/test/use-store-cart.js
```diff
@@ -165,14 +165,17 @@ describe( 'useStoreCart', () => {
...
+				shouldSelect: false,
+			} );
 
 			act( () => {
 				renderer = TestRenderer.create(
 					getWrappedComponents( TestComponent )
 				);
 			} );
 
+			//eslint-disable-next-line testing-library/await-async-query
```
ralucaStan, line 178: I propose to make a ticket to treat/replace these comments in the future, because at the end of the day, it is about using the [Testing library's API wrong](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/await-async-query.md).

ralucaStan (approved):
Looks good, I left some comments but I'll approve as it's nothing major. Great work and effort fixing this.

ralucaStan:
Heads up, there's also one conflict!

package.json
alexflorisca: That was an accident, good spot, thank you 🙏

assets/js/base/context/hooks/cart/test/use-store-cart.js
alexflorisca: In this case, it seems eslint is incorrectly applying these rules that should be for `testing-library/react` to the `react-teset-renderer` library, which from what I can tell is different. So i think these comments are legit, but I didn't look into why eslint is doing this

alexflorisca:
Thanks @ralucaStan, I've made some changes based on your suggestions and merged trunk in. Should be good to merge once the CI finishes 👍

PR #6206, opened 2022-04-06
Tidy up and organise e2e tests

alexflorisca:
This is an overdue cooldown PR which aims to tidy up all the Cart & Checkout E2E tests, More specifically:

### 1. Consistent test descriptions

There was a lot of tests with description block like `describe( 'Shopper → Cart → Can update product quantity', () => {` and test description `allows customer to update product quantity via the input field`. This is duplicating the same description so I simplified it into a simpler format

```describe(`Shopper > {Block name} > {Section (optional)}`)```

and 

``` it('user can {whatever is being tested}')```

### 2. Combine related tests. 

I've combined related tests into one file to benefit from less `before` and `after` hooks and improve the output when running the tests.

`shopper/specs/cart-checkout/cart.test.js` and `shopper/specs/cart-checkout/checkout.test.js` test functionality related to each block respectively. Some tests operate across both Cart and Checkout blocks, and these are kept in separate files within the `cart-checkout` folder.

The benefit of doing this over having each test in a separate file, is that reporting is a lot clearer:
Screenshot 2022-04-06 at 16 17 20

### 3. Standardised use of constants and other smaller inconsistencies 

Because the tests were written by different people at the same time, things were done a bit differently, some more verbose than others. I've tidied up the code a bit to be more consistent.

**What I haven't done**

The utils are a bit of a mess as well, but I've raised a separate [issue ](https://github.com/woocommerce/woocommerce-gutenberg-products-block/issues/6205)for a cooldown maybe.

### Manual Testing

1. `npm run test:e2e` passes

### CODE REVIEW

File: tests/e2e/specs/shopper/cart-checkout/translations.test.js [hunk 1]
```diff
@@ -1,16 +1,16 @@
 /**
  * Internal dependencies
  */
-import { cli } from '../../../utils/run-cli-from-test';
-import { merchant } from '../../../utils/merchant';
-import { shopper } from '../../../utils/shopper';
+import { cli } from '../../../../utils/run-cli-from-test';
+import { merchant } from '../../../../utils/merchant';
+import { shopper } from '../../../../utils/shopper';
```
nielslange: What do you think about changing 

```js
import { cli } from '../../../../utils/run-cli-from-test';
import { merchant } from '../../../../utils/merchant';
import { shopper } from '../../../../utils/shopper';
```

to 

```js
import { cli, merchant, shopper } from '../../../../utils';
```

and making sure that both `cli` and `merchant` are available within https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/trunk/tests/utils/index.js? I have the feeling that combining these dependencies makes the file a bit easier to read.

tests/e2e/specs/shopper/cart-checkout/translations.test.js
```diff
@@ -1,16 +1,16 @@
...
 
 if ( process.env.WOOCOMMERCE_BLOCKS_PHASE < 2 ) {
 	// eslint-disable-next-line jest/no-focused-tests
-	test.only( 'Skipping Checkout tests', () => {} );
+	test.only( 'Skipping Cart & Checkout tests', () => {} );
```
nielslange, line 8: Nitpick: I see that some files use `test.only( 'skipping ... tests', () => {} );` while others use `test.only( 'Skipping ... tests', () => {} );`, thus, `skipping` vs. `Skipping`. What about using one format and adjust all files accordingly?

tests/e2e/specs/shopper/cart-checkout/translations.test.js
alexflorisca: Sounds good, I've exported more files from `utils/index.js` for easier import 👍

tests/e2e/specs/shopper/cart-checkout/translations.test.js
alexflorisca: Good spot, adjusted so they all use uppercase 👍

nielslange (approved):
LGTM! 🚢

PR #6209, opened 2022-04-07
Prepare the Mini Cart block for WC Core release

### LINKED ISSUE #6200: Prepare Mini Cart for WC Core release
This issue is for tracking and preparing the Mini Cart block to release to [WooCommerce](https://wordpress.org/plugins/woocommerce) plugin.

## Pre-release steps

- [ ] Review #6167 and either merge or punt.
- [ ] Review #6188 and either fix or punt.

## Release steps

- [x] Branch: starting from `trunk`, create a new branch for block release. Don't use the `release/` prefix in the branch name because it will trigger the release workflow.
- [x] Changing the register block function from `registerFeaturePluginBlockType()` to `registerBlockType()`.
- [x] Moving the block out of the `is_feature_plugin_build()` check in `src/BlockTypesController.php`.

## Verification steps

- [x] Update `woocommerce_blocks_phase` to 1 in `.blocks.ini`.
- [x] Test the plugin with both block and classic themes to see if a new block is included and functions properly.
- [ ] If the block contains translation, test if the translation works with both block and classic themes.

_Note: This is meant to be updated along the Mini Cart block release process. Once closed, this can also be used as the template for future block releases._

dinhtungdu:
Fixes #6200 

This PR prepares the Mini Cart block to be merged in the WC Core release.

#### Automated Tests
  * [ ] Unit tests
  * [ ] E2E tests
### Manual Testing

1. Run `npm run build`.
2. Replace the content of`.blocks.ini` with the following config:
```
woocommerce_blocks_phase = 1
woocommerce_blocks_env = production
```
3. See the Mini Cart block and Mini Cart blocks content work as expected.

### User Facing Testing
* [x] Same as above, or
* [ ] See steps below.


src/AssetsController.php
```diff
@@ -62,9 +62,7 @@ public function register_assets() {
 		// The price package is shared externally so has no blocks prefix.
 		$this->api->register_script( 'wc-price-format', 'build/price-format.js', [], false );
 
-		if ( Package::feature()->is_feature_plugin_build() ) {
-			$this->api->register_script( 'wc-blocks-checkout', 'build/blocks-checkout.js', [] );
-		}
```
dinhtungdu, line 67: The reason for this removal is: Mini Cart Contents inner blocks use `registerCheckoutBlock()` from `@woocommerce/blocks-checkout` which is mapped to `wc-blocks-checkout`.

@ralucaStan Is there any issue if we make this script available in WC Core?

Aljullu (approved):
Approving from Woo FSE side, but as you said, let's wait until we get feedback on `wc-blocks-checkout` being included in core before merging.

src/AssetsController.php
ralucaStan: Without a lot of background around this, my thoughts go around two things:
- would it be a performance issue to add all the content of `@woocommerce/blocks-checkout`, assuming that the MiniCart doesn't use all the exports from the package?
- I see that `@woocommerce/blocks-checkout` also exports this: `export { SlotFillProvider } from 'wordpress-components';` @Aljullu do you know why this is the case?

dinhtungdu:
Update: the Payment Method icons are lost because the Payment script is [gated](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fix/6200/src/Domain/Bootstrap.php#L247-L256). @Aljullu Do you think we should make the payment scripts available as well?

image

_Notes: The screenshots of the Mini Cart Drawer with Phase 1 is on the left, and phase 3 is on the right._

src/AssetsController.php
Aljullu: > - would it be a performance issue to add all the content of `@woocommerce/blocks-checkout`, assuming that the MiniCart doesn't use all the exports from the package?

That's a good point, @ralucaStan! However, I don't know how much we could reduce the bundle size. Even though Mini Cart only imports `registerCheckoutBlock()` directly, it includes some Cart components that have other dependencies which are required by extensions that want to apply filters (ie: WC Subscriptions).

> - I see that `@woocommerce/blocks-checkout` also exports this: `export { SlotFillProvider } from 'wordpress-components';` @Aljullu do you know why this is the case?

Unfortunately, I don't recall it. @senadir do you have any clue? Could it be deprecated and import `SlotFillProvider` from `@wordpress/components` instead?

Aljullu:
> Update: the Payment Method icons are lost because the Payment script is [gated](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fix/6200/src/Domain/Bootstrap.php#L247-L256). @Aljullu Do you think we should make the payment scripts available as well?

I would say so, yes. @ralucaStan, thoughts on this one?

In any case, that's probably not blocking release, so we can work on it in a separate PR if needed.

ralucaStan:
> > Update: the Payment Method icons are lost because the Payment script is [gated](https://github.com/woocommerce/woocommerce-gutenberg-products-block/blob/fix/6200/src/Domain/Bootstrap.php#L247-L256). @Aljullu Do you think we should make the payment scripts available as well?
> 
> I would say so, yes. @ralucaStan, thoughts on this one?

@Aljullu  There is a lot of logic around Payments that is not relevant for the Mini Cart. It would be great if only the  Payment Method icons part could be exposed. This is PHP land, so I am not sure how/if that would be possible.

As this is not a blocker maybe we can investigate this in a separate issue?

Aljullu:
Ok, given the discussions above and in p1649320040573469-slack-C02FL3X7KR6. Let's merge this PR as-is and we can create follow-up issues for the pending tasks.

dinhtungdu:
Note that I'm merging this PR with failed workflows: The failed JS linting workflow is due to https://github.com/ataylorme/eslint-annotate-action/issues/39. Failed E2E tests are unrelated to this PR.

